PR #421, opened 2023-01-25
MWPW-123872 -  [Blog2Milo] Sticky Header Block

saugatmalla:
*  Block name changed to Promo
*  Sticky logic moved to section-metadata

Resolves: MWPW-123872

**Test URLs:**
- https://main--blog--adobecom.hlx.page/drafts/smalla/sticky-bottom?milolibs=sticky-header
- https://main--blog--adobecom.hlx.page/drafts/smalla/sticky?milolibs=sticky-header


libs/blocks/promo/promo.css
```diff
@@ -0,0 +1,83 @@
...
+        background-position: 0% 50%
+    }
+}
+
+.promo .button {
+    display: block;
+    margin: 0;
+    border-radius: 2rem;
+    border: 2px solid transparent;
+    padding: 6px 1rem;
```
JasonHowellSlavin, line 65: Why are there both `rem` and `px`?

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -42,4 +47,5 @@ export default function init(el) {
   if (metadata.style) handleStyle(metadata.style.text, section);
   if (metadata.background) handleBackground(metadata.background.content, section);
   if (metadata.layout) handleLayout(metadata.layout.text, section);
+  if (metadata.sticky) handleSticky(metadata.sticky.text, section);
```
JasonHowellSlavin, line 50: Not sure what was said in the other PR, but couldn't you just add `sticky-bottom`, `sticky-top` in section-metadata under styles? I understand if you want this to be explicit for AX purposes though.

chrischrischris: Great idea @JasonHowellSlavin

chrischrischris (commented):
Nice!  A few things:
1. We can't hardcode the gnav height as that can be variable.  I think we'll need to use js to get the `<header>` height and use that. (Though defaulting to the current height is fine)
2. For `sticky-top` the element *must* be the first child of main.  If it isn't we need to move it there.  Same for `sticky-bottom` except it needs to be the last child.

For 2, we could educate authors to let them know that the sticky sections must be the first or last, but I feel like many will forget and then wonder why it's not working as expected.  It would allow for sticky sections within the page but I don't think we need to support that use case.

@rgclayton @auniverseaway @JasonHowellSlavin Thoughts?

rgclayton:
> For 2, we could educate authors to let them know that the sticky sections must be the first or last, but I feel like many will forget and then wonder why it's not working as expected. It would allow for sticky sections within the page but I don't think we need to support that use case.

Hmm. I'm not sure how well that would go over with authors having to remember that. I need to think on that.

libs/blocks/section-metadata/section-metadata.js
rgclayton: Not a bad idea. I guess that would help with not "junking up" `section-metadata` with more properties. Sounds good to me.

rgclayton:
@saugatmalla looking good! Still reviewing but just wanted to say thanks :)

saugatmalla:
> Nice! A few things:
> 
> 1. We can't hardcode the gnav height as that can be variable.  I think we'll need to use js to get the `<header>` height and use that. (Though defaulting to the current height is fine)

@chrischrischris Should we listen to `resize` event or is it fine to set it only at first?

chrischrischris:
> @chrischrischris Should we listen to `resize` event or is it fine to set it only at first?

@saugatmalla Yes good idea.  Be sure to debounce it.

chrischrischris:
Re: Automatically moving sticky divs to top/bottom of `<main>`:  I think we should do it to avoid any author confusion.

rgclayton (commented):
Great work @saugatmalla! 

It looks like this is introducing some CLS that impact LH scores. You can see a shift in the subscribe button when the close button loads. It might help to predefine the width of those in css to stop the shifting and bring that score back up. 

image

saugatmalla:
> It looks like this is introducing some CLS that impact LH scores. You can see a shift in the subscribe button when the close button loads. It might help to predefine the width of those in css to stop the shifting and bring that score back up.
> 
@rgclayton I added width and height for the close icon in css, which should solve the CLS of subscribe.

rgclayton:
Confirmed that moving the promo and sticky section-metadata to be the first content on the page fixes the CLS issue I mentioned. This would require KT to authors to avoid performance penalties. 

[Test page](https://sticky-header--milo--adobecom.hlx.page/drafts/rclayton/sticky-promo?martech=off)

cc: @chrischrischris @auniverseaway

rgclayton (approved):
Promo and section-metadata are working and looking good. 
CLS issue for sticky-top can be fixed in authoring.

chrischrischris (approved):

PR #426, opened 2023-01-27
MWPW-125095 Adding iFrame and Consumer AutoBlocks

Brandon32:
* Add iFrame block using shared styles with YouTube and Adobe TV blocks.
* Allow consumers to add AutoBlocks via config.

Resolves: [MWPW-125095](https://jira.corp.adobe.com/browse/MWPW-125095)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/bmarshal/embed-product-tour?martech=off
- After: https://bmarshal-embed-product-tour--milo--adobecom.hlx.page/drafts/bmarshal/embed-product-tour?martech=off

**BACOM Test URLs:** 
- Before: https://bmarshal-embed-product-tour--bacom--adobecom.hlx.page/drafts/bmarshal/embed-product-tour?martech=off
- After: https://bmarshal-embed-product-tour--bacom--adobecom.hlx.page/drafts/bmarshal/embed-product-tour?milolibs=bmarshal-embed-product-tour&martech=off


libs/utils/utils.js
```diff
@@ -394,10 +396,12 @@ export function decorateSVG(a) {
 }
 
 export function decorateAutoBlock(a) {
+  const config = getConfig();
   const { hostname } = window.location;
   const url = new URL(a.href);
   const href = hostname === url.hostname ? `${url.pathname}${url.search}${url.hash}` : a.href;
-  return AUTO_BLOCKS.find((candidate) => {
+  const autoBLocks = AUTO_BLOCKS.concat(config.autoBlocks ?? [])
```
auniverseaway, line 403: I wouldn't concat twice. I'd to this all in the config once and be done.

```js
config.autoBlocks = [ ...AUTO_BLOCKS, ...conf.autoBlocks ];
```

auniverseaway (changes requested):

libs/blocks/iframe/iframe.js
```diff
@@ -0,0 +1,13 @@
+export default function init(a) {
+  const url = a?.href ?? a?.querySelector('a')?.href;
```
meganthecoder, line 2: I would suggest renaming `a` to be `link`. It seems like there's a scenario where the arg could be a div with an anchor in it. I think if the param name is a tag name, it really should only accept that tag.

meganthecoder, line 2: Nit: The amount of `?` operators in there makes it hard to read in my opinion
```suggestion
export default function init(a) {
  if (!a) return;
  const url = a.href ?? a.querySelector('a')?.href;
```

libs/blocks/iframe/iframe.js
```diff
@@ -0,0 +1,13 @@
...
+
+  if (!url) return;
+
+  const embed = `<div class="milo-iframe">
+    <iframe src="${url}">
+    </iframe>
+  </div>`;
```
meganthecoder, line 9: FYI there's also the milo util `createTag`

meganthecoder (commented):
Few suggestions

libs/blocks/iframe/iframe.js
Brandon32: I think we can assume the parameter exists, that is the convention for all other blocks.
```
const url = a.href ?? a.querySelector('a')?.href;
```

Brandon32: The convention seems to be `a` for autoBlocks and `el` for blocks, since this works for both `link` seems like a good option.

chrischrischris (approved):

libs/blocks/iframe/iframe.js
```diff
@@ -0,0 +1,13 @@
+import { createTag } from '../../utils/utils.js';
+
+export default function init(link) {
+  const url = link.href ?? link.querySelector('a')?.href;
```
JasonHowellSlavin, line 4: You could remove the optional chaining here I think as well. If the link does not have an href, it will return `undefined` which will lead ` if(!url)`  to evaluate to true, thus fulfilling your condition regardless of the chaining.

JasonHowellSlavin (approved):
LGTM

meganthecoder (approved):

libs/blocks/iframe/iframe.js
Brandon32: Optional chaining is required here, if a link is not authored then you will get an type error because it will resolve to `null.href`.

auniverseaway: This seems really weird here. Where is the scenario where a link is inside another link?

@Brandon32

Brandon32: To test this functionality I am creating a block with a link in it, so `link` is a `div > a`. Normally init takes the variable `a` for autoBlocks or `el` for blocks, I used `link`, maybe `el` is a better option?

Brandon32: @auniverseaway I've updated the code to use `el`, let me know if that work for you, thanks.

PR #431, opened 2023-01-31
MWPW-124483 Dynamic content roots

fullcolorcoder:
Adds ability to pass an array of folders to Milo that define an individual contentRoot for that folder. This allows authors to have custom gnav/footers for different sections of their site by authoring gnav/footer as they normally would, but in that folder. 

Resolves: [MWPW-124483](https://jira.corp.adobe.com/browse/MWPW-124483)

**Test URLs:**

**FedPub** (Passes array in this example) 
- Before: https://dynamic-gnav--fedpub--adobecom.hlx.page/kr/acrobat/hub/how-to/10-benefits-of-digitizing-documents
- After: https://dynamic-gnav--fedpub--adobecom.hlx.page/kr/acrobat/hub/how-to/10-benefits-of-digitizing-documents?milolibs=dynamic-content-roots

**FedPub homepage** - fallback example. 
- Before: https://dynamic-gnav--fedpub--adobecom.hlx.page
- After: https://dynamic-gnav--fedpub--adobecom.hlx.page?milolibs=dynamic-content-roots

The above example is the only situation currently where an array is being passed in contentRoot: 

`contentRoot: ['acrobat', 'documentcloud', 'creativecloud']`

The below test URLs should just not break as they are today. 

**Milo**
- Before: https://main--milo--adobecom.hlx.page/drafts/bradjohn/abb-case-study?martech=off
- After: https://dynamic-content-roots--milo--adobecom.hlx.page/drafts/bradjohn/abb-case-study?martech=off

**Bacom**
- Before: https://main--bacom--adobecom.hlx.page/customer-success-stories/hugo-boss-case-study?milolibs=dynamic-content-roots
- After: https://main--bacom--adobecom.hlx.page/customer-success-stories/hugo-boss-case-study?milolibs=dynamic-content-roots


libs/blocks/footer/footer.js
```diff
@@ -315,7 +315,12 @@ async function fetchFooter(url) {
 
 export default async function init(block) {
   const { prefix } = locale;
-  const url = getMetadata('footer-source') || `${prefix}/footer`;
+  let url = '';
+  if (contentRoot?.length) {
+    url = getMetadata('footer-source') || `${prefix}${contentRoot}/footer`;
+  } else {
+    url = getMetadata('footer-source') || `${prefix}/footer`;
+  }
```
seanchoi-dev, line 323:
```suggestion
  const url = getMetadata('footer-source') || `${prefix}${contentRoot}/footer`;
```
Since the default value of `contentRoot` is empty string, I think this would work. WDYT?

libs/utils/utils.js
```diff
@@ -144,6 +144,23 @@ export const [setConfig, getConfig] = (() => {
...
+            config.contentRoot = `/${pathSplit[2]}`;
+          } else {
+            config.contentRoot = `/${pathSplit[1]}/${pathSplit[2]}`;
+          }
+        } else if (!config.locale.prefix.length > 0 && !contentRoots.includes(pathSplit[1]) && pathSplit[2]) {
+          config.contentRoot = `/${pathSplit[1]}`;
+        } else {
+          config.contentRoot = '';
+        }
+      }
```
seanchoi-dev, line 163: This is complicated to read due to many if-else statements. If this is only way to achieve the logic, it will need some comments.

seanchoi-dev: It seems the new logic isn't covered by test.

seanchoi-dev (changes requested):

libs/utils/utils.js
```diff
@@ -144,6 +144,27 @@ export const [setConfig, getConfig] = (() => {
         // eslint-disable-next-line no-console
         console.log('Invalid or missing locale:', e);
       }
+      if (Array.isArray(config.contentRoot)) {
```
chrischrischris, line 147: Break out into another small func and make use of config.locale instead - something like:
```
const getContentRoot = (contentRoot, prefix, pathname) => {
  if (Array.isArray(contentRoot)) {
    const rootDir = pathname.split('/')[prefix ? 2 : 1];
    if (contentRoot.includes(rootDir)) {
      return `${prefix}/${rootDir}`;
    } else {
      return prefix;
    }
  }
  return contentRoot;
}

config.contentRoot = getContentRoot(config.contentRoot, config.locale.prefix, pathname);
```

chrischrischris (changes requested):
Does gnav also need to be updated?

libs/utils/utils.js
```diff
@@ -144,6 +144,27 @@ export const [setConfig, getConfig] = (() => {
...
+          }
+        } else if (!config.locale.prefix.length > 0 && !contentRoots.includes(pathSplit[1]) && pathSplit[2]) {
+          // page is not in contentRoot array, has no prefix (us-en), and isn't root index. ie: root page of /au/ site.
+          config.contentRoot = `/${pathSplit[1]}`;
+        } else {
+          // Fallback - ie: "/".
+          config.contentRoot = '';
+        }
+      }
       config.locale.contentRoot = `${origin}${config.locale.prefix}${config.contentRoot ?? ''}`;
```
chrischrischris, line 168: With your changes, will this become `/de/de/contentroooot` ?

fullcolorcoder: I don't think so. This one comes out to /kr/acrobat for example. 
https://dynamic-gnav--fedpub--adobecom.hlx.page/kr/acrobat/hub/how-to/10-benefits-of-digitizing-documents

libs/utils/utils.js
fullcolorcoder: It looks like coverage is there generically. Do you think we should write extensive tests for this?

seanchoi-dev (approved):

PR #435, opened 2023-02-02
share block: copy to clipboard button added & localization fixed

So1ene:
# share block: copy to clipboard button added & localization fixed #

**Test URLs:**
- Before: [https://...hlx.page/drafts/solene/archive/share](https://main--stock--adobecom.hlx.page/drafts/solene/archive/share?martech=off)
- After: [https://...hlx.page/drafts/solene/archive/share?milolibs=copy-to-clipboard](https://main--stock--adobecom.hlx.page/drafts/solene/archive/share?martech=off&milolibs=copy-to-clipboard)

Localization Test:
- Before: [https://...hlx.page/es/pages/drafts/solene/share](https://main--stock--adobecom.hlx.page/es/pages/drafts/solene/share?martech=off)
- After: [https://...hlx.page/es/pages/drafts/solene/share?milolibs=copy-to-clipboard](https://main--stock--adobecom.hlx.page/es/pages/drafts/solene/share?martech=off&milolibs=copy-to-clipboard)


libs/blocks/share/share.js
```diff
@@ -42,43 +42,62 @@ function getPlatforms(el) {
...
+  };
+  
+  const heading = toSentenceCase(await replaceKey('share-this-page', config));
+  el.append(createTag('p', null, ((heading))));
   const container = createTag('p', { class: 'icon-container' });
-  svgs.forEach((svg) => {
-    const details = getDetails(svg.name, url);
-    if (!details) return;
-    const shareLink = createTag('a', { target: '_blank', href: details.href, title: `Share to ${details.title}` }, svg.svg);
+  svgs.forEach(async (svg) => {
```
chrischrischris, line 75: I assume there is only ever one `clipboard` svg, so it may be cleaner to pull it out of the svgs.

```
const clipboardSvg = svgs.find((svg) => svg.name === 'clipboard');
if (!(clipboardSupport && clipboardSvg)) return;

...
```

libs/blocks/share/share.js
```diff
@@ -42,43 +42,62 @@ function getPlatforms(el) {
...
+    let clipboard = svg.name === 'clipboard';
+    const obj = getDetails(svg.name, url);
+    if (!obj && !clipboard) return;
+
+    let shareLink;
+    if (clipboard && clipboardSupport) {
+      const clipboardToolTip = toSentenceCase(await replaceKey('copy-to-clipboard', config));
+      const copiedTooltip = toSentenceCase(await replaceKey('copied', config));
+      shareLink = createTag('button', { title: clipboardToolTip, type:'button', class:'copy-to-clipboard', 'data-copy-to-clipboard': clipboardToolTip, 'data-copied': `${copiedTooltip}!` }, svg.svg);
```
chrischrischris, line 84: nit: use multiple lines

amauch-adobe:
@So1ene do there need to be more unit tests added for code coverage in the share.js block? It just looks like the coverage percentage went down.  So it made me wonder if we need to add any missing test cases.

amauch-adobe (approved):
@So1ene, everything looks good.  I've tested the functionality against all of the browsers.  My only question/concern was the dip in the share.js testing coverage.  If we can get it up to 100% then I would say this is approved, unless there is a valid reason for it being lower.

chrischrischris (approved):
Looks good! Just the fix in coverage (or ignore unreachable lines).

So1ene:
As discussed with @chrischrischris and Mairi (designer from Webistry), I updated the style of the tooltip to match the Adobe style guide convension

PR #438, opened 2023-02-02
[MWPW-123840] Gnav centralised placeholders & main style refactor

overmyheadandbody:
This PR includes:
* placeholder logic with fallback mechanism;
* use of placeholders in the Search block;
* refactor of the main global navigation styles - going away from grid and fixed positioning, since they weren't set up correctly, nor did they take all use-cases into consideration, such as elements that might be added before the navigation. The styles were also a risk for maintainability, since the same (or even slightly different) values were added in multiple places, with varying degrees of specificity;
* first steps for refactoring the brand block;
* first steps for breadcrumbs style refactoring.

Resolves: [MWPW-123840](https://jira.corp.adobe.com/browse/MWPW-123840) & [MWPW-125090](https://jira.corp.adobe.com/browse/MWPW-125090)

**Test URLs:**
- Before: https://feds-gnav--milo--adobecom.hlx.page/drafts/ramuntea/gnav-refactor?martech=off
- After: https://mwpw-123840-centralised-placeholders--milo--overmyheadandbody.hlx.page/drafts/ramuntea/gnav-refactor?martech=off

### DIFF
--- a/libs/blocks/global-navigation/blocks/search/gnav-search.js
+++ b/libs/blocks/global-navigation/blocks/search/gnav-search.js
@@ -4,6 +4,8 @@ import {
   getLocale,
   getCountry,
 } from '../../utilities.js';
+import { replaceKeyArray } from '../../../../features/placeholders.js';
+import { getConfig } from '../../../../utils/utils.js';
 
 const CONFIG = {
   suggestions: {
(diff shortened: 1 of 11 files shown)


libs/blocks/global-navigation/blocks/search/gnav-search.js
```diff
@@ -18,24 +20,31 @@ const CONFIG = {
...
     this.addEventListeners();
   }
 
+  async getLabels() {
+    const config = getConfig();
+    const sheet = 'feds';
+    this.labels = {};
+    [this.labels.search, this.labels.clearResults, this.labels.tryAdvancedSearch] = await replaceKeyArray(['search', 'clear-results', 'try-advanced-search'], config, sheet);
+  }
+
```
mokimo, line 43: Since we just use the function and variables once, IMO just having it all in the init is a bit cleaner - but it's only a slight preference, feel free to keep your code or use this suggestion:
```suggestion
  async init() {
    this. Labels = {};
    [this.labels.search, this.labels.clearResults, this.labels.tryAdvancedSearch] = await replaceKeyArray(['search', 'clear-results', 'try-advanced-search'], getConfig(), 'feds');
    this.decorate();
    this.addEventListeners();
  }
```

mokimo (approved):
Great PR 👏

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -7,11 +7,15 @@ import {
 } from '../../utils/utils.js';
 import { analyticsGetLabel } from '../../martech/attributes.js';
 import { toFragment } from './utilities.js';
+import { replaceKey } from '../../features/placeholders.js';
 
-const CONFIG = { search: { icon: '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" focusable="false"><path d="M14 2A8 8 0 0 0 7.4 14.5L2.4 19.4a1.5 1.5 0 0 0 2.1 2.1L9.5 16.6A8 8 0 1 0 14 2Zm0 14.1A6.1 6.1 0 1 1 20.1 10 6.1 6.1 0 0 1 14 16.1Z"></path></svg>' } };
+const CONFIG = {
+  icons: {
+    company: '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 133.46 118.11"><defs><style>.cls-1{fill:#fa0f00;}</style></defs><polygon class="cls-1" points="84.13 0 133.46 0 133.46 118.11 84.13 0"/><polygon class="cls-1" points="49.37 0 0 0 0 118.11 49.37 0"/><polygon class="cls-1" points="66.75 43.53 98.18 118.11 77.58 118.11 68.18 94.36 45.18 94.36 66.75 43.53"/></svg>',
```
narcis-radu, line 14: Nothing urgent or critical, but I think this icon can be optimized further - the viewbox size and styles can be removed from here and added in CSS

narcis-radu (approved):

libs/blocks/global-navigation/blocks/search/gnav-search.js
overmyheadandbody: I usually prefer the init method to just declare instance variables and call methods, so it's super-clear what's going on, without any sort of clutter. I think I'll leave it as-is for now and we can come back to it if we decide on another rule.

libs/blocks/global-navigation/global-navigation.js
overmyheadandbody: Yup, I did that for another icon that ended up not being used, but I can do it here as well. However, this is a fallback icon for the logo block. I'm not sure if we want to have a hardcoded image in our code anyway. Instead, it should probably be added from Sharepoint by authors.

PR #443, opened 2023-02-02
MWPW-125464 - Adding events to the modal on close and load

JasonHowellSlavin:
* Adds a 'milo:modal:closed' event on modal close
* Adds a 'milo:modal:loaded' event on modal open/load
* Adds a check to see if the modal item to be removed during close modal is the modal itself - resolves a bug @Brandon32  found while reviewing

Resolves: [MWPW-125464](https://jira.corp.adobe.com/browse/MWPW-125464)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/slavin/complex-modal?martech=off
- After: https://add-modal-events--milo--adobecom.hlx.page/drafts/slavin/complex-modal?martech=off

**To Test**
You can add the following functions in the console and observe that the console logs out when the modal is opened and closed:
```
window.addEventListener('milo:modal:closed', () => {
    console.log('closed');
})
window.addEventListener('milo:modal:loaded', () => {
    console.log('loaded');
})
```

### DIFF
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -18,6 +18,8 @@ function findDetails(hash, el) {
 
 function closeModal(modal) {
   const { id } = modal;
+  const closeEvent = new Event('modal:closed');
+  window.dispatchEvent(closeEvent);
 
   document.querySelectorAll(`#${id}`).forEach((mod) => {
     if (mod.nextElementSibling?.classList.contains('modal-curtain')) {
@@ -66,6 +68,7 @@ export async function getModal(details, custom) {
   const { id } = details || custom;
 
   const dialog = createTag('div', { class: 'dialog-modal', id });
+  const loadedEvent = new Event('modal:loaded');
 
   if (custom) getCustomModal(custom, dialog);
   if (details) await getPathModal(details.path, dialog);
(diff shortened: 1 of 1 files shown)


Brandon32:
This is unrelated to your update but the H1 is being removed after closing the modal?

Brandon32 (approved):

libs/blocks/modal/modal.js
```diff
@@ -18,6 +18,8 @@ function findDetails(hash, el) {
 
 function closeModal(modal) {
   const { id } = modal;
+  const closeEvent = new Event('modal:closed');
```
mokimo, line 21: The event could be a bit more specific `milo:modal:closed`? I feel like `modal: closed` could also be used by some 3rd party scripts leading to funky behaviour in the future

JasonHowellSlavin:
> This is unrelated to your update but the H1 is being removed after closing the modal?

The entirety of the modal is removed, if it was authored with an h1, then I suppose so, but removing the modal doesn't affect the DOM of the page on which the modal is placed.

Brandon32:
> > This is unrelated to your update but the H1 is being removed after closing the modal?
> 
> The entirety of the modal is removed, if it was authored with an h1, then I suppose so, but removing the modal doesn't affect the DOM of the page on which the modal is placed.

But that's what I'm seeing, tried this on Chrome, FireFox and Safari to double check.
image

libs/blocks/modal/modal.js
JasonHowellSlavin: great callout, will change it!

mokimo (approved):

chrischrischris (approved):

PR #447, opened 2023-02-06
MWPW-121123 - Aside Block Updates

Sartxi:
- Adds right alignment variant on Standard Asides
- Adjustments to button/link spacing and allowing links in action areas
- Adds Video support on Standard Asides for MP4 videos and AdobeTv Embed
- Adds ability to author custom content backgrounds for Split Asides
- Also contains style updates to the notification aside to match consonant designs

Resolves: [MWPW-121123](https://jira.corp.adobe.com/browse/MWPW-121123)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/sarchibeque/asides?martech=off
- After: https://consonant-asides--milo--adobecom.hlx.page/drafts/sarchibeque/asides?martech=off


libs/blocks/aside/aside.js
```diff
@@ -14,74 +14,72 @@
...
-      btnWrap.append(...[...buttons].map((button) => button.parentElement));
-    }
-    const desc = createTag('p', null, text.innerHTML);
-    text.innerHTML = '';
-    text.append(
-      iconArea || '',
-      desc,
-      btnWrap || '',
-    );
-  }
```
meganthecoder, line 53: If I'm remembering correctly, this was added because franklin doesn't add a `p` tag if the content in a block is only one line. Did that change or are you no longer supporting a single line (for the notification aside especially)?

libs/blocks/aside/aside.css
```diff
@@ -1,3 +1,4 @@
+/* stylelint-disable selector-class-pattern */
```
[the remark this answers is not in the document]
Sartxi: @meganthecoder this selector class pattern complains about our text content styles like `detail-M` not being kebab-case. We have quite a few of those. I am thinking we might want to update our stylelint config to leave those alone. Thoughts?

libs/blocks/aside/aside.js
Sartxi: @meganthecoder yes, notification asides should still supports a single line but it was easy enough to just style using the parent class name without having to wrap it in a `p` element. I was just trying to clean it up and streamline the function a bit and noticed it was easy enough to get the desired outcome without extra code, if that makes sense.

libs/blocks/aside/aside.css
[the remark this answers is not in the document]
ryanmparrish: I'd say we should not disable style-lint and fix the casing issues across the whole site styles. This can be a separate task but id rather fix the issue wholesale then pepper in style-disable comments. That'd be my take.

libs/blocks/aside/aside.css
```diff
@@ -7,9 +8,14 @@
   overflow: hidden;
 }
 
+.dark .aside {
+  color: white;
```
ryanmparrish, line 12: This probs isn't necessary. Our global styles should catch any .dark > elem w/ this already.

libs/blocks/aside/aside.css
```diff
@@ -7,9 +8,14 @@
...
+}
+
 .aside .detail-M {
   text-transform: uppercase;
   font-weight: 700;
```
ryanmparrish, line 17: Thinks we can omit these text styles. 
Any `detail-*` selector should already have these styles from the typography.css base styles.

libs/blocks/aside/aside.css
```diff
@@ -29,19 +35,29 @@
 }
 
 .aside.dark {
+  color: white;
```
ryanmparrish, line 38: nit: same as above. .dark is a global

libs/blocks/aside/aside.css
```diff
@@ -29,19 +35,29 @@
...
+
 .aside.dark h1,
 .aside.dark h2,
 .aside.dark h3,
 .aside.dark h4,
 .aside.dark h5,
 .aside.dark h6,
-.aside.dark p {
+.aside.dark p:not(.action-area) {
   color: white;
```
ryanmparrish, line 54: Hmm, Id check if this is necessary. Same as above.

libs/blocks/aside/aside.css
```diff
@@ -29,19 +35,29 @@
 }
 
 .aside.dark {
+  color: white;
   background-color: #1E1E1E;
 }
 
+.aside.light,
+.aside.split {
+  background-color: #f5f5f5;
```
ryanmparrish, line 44: Shouldn't this be an author-able first row decorator and not set in style?

libs/blocks/aside/aside.css
```diff
@@ -93,23 +122,44 @@
...
+.aside .foreground.container .text .action-area > a {
   margin-right: var(--spacing-s);
 }
 
+.aside .foreground.container .text .action-area > a:last-child {
+  margin-right: 0;
+}
+
+.aside .foreground.container .text .action-area > a:not(.con-button) {
+  text-decoration: underline;
```
ryanmparrish, line 134: We need global link underline styles in typography.css. I'd shy away from this deep of specificity.

ryanmparrish: Again probably another task to sweep cleanup global link styles.

libs/blocks/aside/aside.css
```diff
@@ -93,23 +122,44 @@
...
 .aside .foreground.container div {
   flex: 1 1 auto;
 }
 
-.aside .foreground.container .image {
+.aside .foreground.container .image,
+.aside .foreground.container div:has(.milo-video),
+.aside .foreground.container .milo-inline-video {
+  display: flex;
   margin-top: 32px;
```
ryanmparrish, line 145: This seems weird to have set pixel value on top and variable on bottom. Id use vars for both if possible

libs/blocks/aside/aside.css
```diff
@@ -140,11 +195,10 @@
 
 .aside.split {
   flex-direction: column;
-  background-color: #1E1E1E;
 }
 
-.aside.split.light {
-  background-color: white;
+.aside.split.dark {
+  background-color: #1E1E1E;
```
ryanmparrish, line 201: Same comment as above. Might want to check w/ the team if these should be author-able via the first row decorator. Hard coded colors aren't very forgiving.

libs/blocks/aside/aside.css
```diff
@@ -153,9 +207,24 @@
 
 .aside.split .foreground.container {
   width: 100%;
+  max-width: 100%;
+  flex-direction: column;
+  z-index: 1;
 }
 
 .aside.split .foreground.container .text {
+  padding: 40px 0;
```
ryanmparrish, line 216: Can we use spacing vars here? `--spacing-l`

libs/blocks/aside/aside.css
```diff
@@ -153,9 +207,24 @@
...
+}
+
+.aside.inline .foreground.container .image,
+.aside.inline .foreground.container div:has(.milo-video),
+.aside.inline .foreground.container .text {
+  padding: 0;
+}
+
+.aside.split.dark .foreground.container .text {
   color: white;
```
ryanmparrish, line 228: Omit if possible - global dark/light type styles.

libs/blocks/aside/aside.css
```diff
@@ -253,72 +281,59 @@
...
-  white-space: nowrap;
+  justify-content: center;
 }
 
 .aside.notification .foreground.container .text a:not(.con-button) {
   width: auto;
-  margin-left: 5px;
+  text-decoration: underline;
+  font-weight: normal;
+  color: #2c2c2c;
```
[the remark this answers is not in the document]
ryanmparrish: Yep, same comment above w/ using global link styles. Lets fall back where possible to those.

libs/blocks/aside/aside.css
```diff
@@ -253,72 +281,59 @@
...
 
-.aside.notification .foreground.container .text .action-area>a {
+.aside.notification .foreground.container .text .action-area > a {
   margin-right: 0;
 }
 
-.aside.notification .foreground.container .text .body-area a {
-  text-decoration: underline;
+.aside.notification.dark .foreground.container .text a:not(.con-button) {
+  color: white;
```
ryanmparrish, line 310: Seems too specific. use global styles

libs/blocks/aside/aside.css
```diff
@@ -253,72 +281,59 @@
...
 }
 
-.aside.notification .foreground.container .text .body-area a:first-of-type {
-  color: var(--text-color);
-  font-weight: normal;
+.aside.notification.extra-small .foreground.container .text a:last-child {
+  color: var(--link-color);
+  font-weight: bold;
```
ryanmparrish, line 315: Lets use global link styles instead of per block.

libs/blocks/aside/aside.css
```diff
@@ -253,72 +281,59 @@
...
 }
 
-.aside.notification.dark .foreground.container .text .body-area a:first-of-type {
+.aside.notification.extra-small.dark .foreground.container .text a:last-child {
   color: white;
```
ryanmparrish, line 319: Too specific. Global link .dark/.light styles should work here

libs/blocks/aside/aside.js
```diff
@@ -14,74 +14,72 @@
...
+import { decorateBlockBg, decorateBlockText } from '../../utils/decorate.js';
 import { createTag } from '../../utils/utils.js';
 
-const asideTypes = ['inline', 'notification'];
-const [INLINE, NOTIFICATION] = asideTypes;
-const asideSizes = ['extra-small', 'small', 'medium', 'large'];
-const [SIZE_XS, SIZE_S, SIZE_M, SIZE_L] = asideSizes;
+// standard/default aside uses same text sizes as the split
+const variants = ['split', 'inline', 'notification'];
+const sizes = ['extra-small', 'small', 'medium', 'large'];
```
ryanmparrish, line 22: Nit: should we use the same namespace 'xsmall' rather?

test/blocks/aside/mocks/body.html
```diff
@@ -251,23 +245,18 @@ <h2>Notification - Default (Large)</h2>
...
-          <source type="image/webp" srcset="./media_1c93cf2bdf612f082bbf51805ddb83ffbba17fbd1.jpeg?width=750&amp;format=webply&amp;optimize=medium">
-          <source type="image/jpeg" srcset="./media_1c93cf2bdf612f082bbf51805ddb83ffbba17fbd1.jpeg?width=2000&amp;format=jpeg&amp;optimize=medium" media="(min-width: 400px)">
-          <img loading="eager" alt="" type="image/jpeg" src="./media_1c93cf2bdf612f082bbf51805ddb83ffbba17fbd1.jpeg?width=750&amp;format=jpeg&amp;optimize=medium" width="640" height="480">
+          <img loading="eager" alt="" type="image/jpeg" src="" width="640" height="480">
         </picture>
       </div>
     </div>
   </div>
-</div>
+</div>
```
ryanmparrish, line 262: add new line at end.

ryanmparrish (changes requested):
Few things to address, mostly tapping into global styles and not overriding them. LMK if you have any issues/questions. Thanks

libs/blocks/aside/aside.js
Sartxi: @ryanmparrish This was the original way they are aurthored I am unsure if there are any pages already authored with that size. So we might want to verify that before changing it?

libs/blocks/aside/aside.css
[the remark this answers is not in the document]
Sartxi: I simply removed the comment and we can fix them everywhere in a separate task (being that this is not the only place where these exist)

libs/blocks/aside/aside.css
```diff
@@ -253,72 +281,59 @@
...
-
-.aside.notification .foreground.container .text[data-align=center],
-.aside.notification .foreground.container .text[data-align=center] > * {
-  justify-content: center;
-}
-
-.aside.notification .foreground.container .text[data-align=center] p {
+.aside.notification.center .foreground.container .text,
+.aside.notification.center .foreground.container .text > * {
   text-align: center;
```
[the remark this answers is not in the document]
Sartxi: I added the data align attribute selector back so this is compatible. Thanks @meganthecoder

libs/blocks/aside/aside.js
```diff
@@ -14,74 +14,72 @@
...
+  return variant && size && !Array.isArray(blockData) ? blockData[size] : blockData;
+}
+
+function decorateInlineVideo(el) {
+  const video = el.querySelector('a[href*=".mp4"]');
+  if (!video) return;
+  const videoHref = video.getAttribute('href');
+  const videoAttrs = videoHref.includes('autoplay=true') ? 'playsinline autoplay muted loop' : 'controls';
+  const parent = video.parentElement;
+  parent.innerHTML = `<video preload="metadata" ${videoAttrs}><source src="${video.href}" type="video/mp4" /></video>`;
```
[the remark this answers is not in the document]
Sartxi: cool I made it work with createTag!

libs/blocks/aside/aside.css
Sartxi: okay removed this one

libs/blocks/aside/aside.css
Sartxi: fixed

libs/blocks/aside/aside.css
Sartxi: it looks like we are doing this in other blocks so leaving this until we create the global styles

libs/blocks/aside/aside.css
Sartxi: werent needed!

libs/blocks/aside/aside.css
Sartxi: done

libs/blocks/aside/aside.css
Sartxi: We can actually override these but I think removing the defaults will have some issues backward compatibility

libs/blocks/aside/aside.css
Sartxi: I adjusted this a bit, but these were their final decisions on default light/dark colors. These are overrides by the first row authoring.

meganthecoder (approved):
LGTM

ryanmparrish (approved):

PR #456, opened 2023-02-08
MWPW-124568: Floodgate Admin Base Page

sukamat:
* The floodgate admin page follows the pattern of Milo Localization in terms of code/content setup and processing, especially given the overlap of functionalities and Sharepoint API usage.
* Creates a base admin page for Floodgate
* Authenticates with sharepoint
* Reads the Floodgate config and project excel file and renders the data on the page
* Checks for Source and Floodgated page in sharepoint and renders the link and status

Resolves: [MWPW-124568](https://jira.corp.adobe.com/browse/MWPW-124568)

**Test URLs:**
- Before: N/A
- After: [https://floodgate-base--milo--sukamat.hlx.page/tools/floodgate/index.html](https://floodgate-base--milo--sukamat.hlx.page/tools/floodgate/index.html?project=milo--adobecom&referrer=https%3A%2F%2Fadobe.sharepoint.com%2F%3Ax%3A%2Fr%2Fsites%2Fadobecom%2F_layouts%2F15%2FDoc.aspx%3Fsourcedoc%3D%257B4EBD6F30-D51F-419B-BD0A-7485D4081302%257D%26file%3Dsample_fg_project.xlsx%26action%3Ddefault%26mobileredirect%3Dtrue)

### DIFF
--- a/tools/floodgate/js/config.js
+++ b/tools/floodgate/js/config.js
@@ -1,0 +1,30 @@
+import { getUrlInfo } from '../../loc/utils.js';
+import {
+  fetchConfigJson,
+  getSharepointConfig,
+  getHelixAdminConfig,
+} from '../../loc/config.js';
+
+const FLOODGATE_CONFIG = '/drafts/floodgate/configs/config.json';
+
+let decoratedConfig;
+
+async function getConfig() {
+  if (!decoratedConfig) {
+    const urlInfo = getUrlInfo();
+    if (urlInfo.isValid()) {
+      const configPath = `${urlInfo.origin}${FLOODGATE_CONFIG}`;
+      const configJson = await fetchConfigJson(configPath);
+      decoratedConfig = {
+        sp: getSharepointConfig(configJson),
+        admin: getHelixAdminConfig(),
+      };
+    }
+  }
+  return decoratedConfig;
+}
+
+export {
+  getConfig,
+  FLOODGATE_CONFIG,
+};
--- a/tools/floodgate/js/project.js
+++ b/tools/floodgate/js/project.js
@@ -1,0 +1,128 @@
+import { getConfig } from './config.js';
+import {
+  getUrlInfo,
+  getDocPathFromUrl,
+  fetchProjectFile,
+} from '../../loc/utils.js';
+import {
+  getProjectFileStatus,
+  getHelixAdminApiUrl,
+  readProjectFile,
+} from '../../loc/project.js';
+import { getSpFiles } from '../../loc/sharepoint.js';
+import { getFloodgateUrl } from './utils.js';
+
+let project;
+
+/**
+ * Makes the sharepoint file data part of `projectDetail` per URL.
+ */
+function injectSharepointData(projectUrls, filePaths, docPaths, spBatchFiles, isFloodgate) {
+  spBatchFiles.forEach((spFiles) => {
+    if (!spFiles?.responses) return;
+    spFiles.responses.forEach(({ id, status, body }) => {
+      const filePath = docPaths[id];
+      const fileBody = status === 200 ? body : {};
+      const urls = filePaths.get(filePath);
+      urls.forEach((key) => {
+        const urlObjVal = projectUrls.get(key);
+        if (isFloodgate) {
+          urlObjVal.doc.fg.sp = fileBody;
+          urlObjVal.doc.fg.sp.status = status;
+        } else {
+          urlObjVal.doc.sp = fileBody;
+          urlObjVal.doc.sp.status = status;
+        }
+      });
+    });
+  });
+}
+
+async function updateProjectWithDocs(projectDetail) {
+  if (!projectDetail || !projectDetail?.filePaths) {
+    return;
+  }
+  const { filePaths } = projectDetail;
+  const docPaths = [...filePaths.keys()];
+  const spBatchFiles = await getSpFiles(docPaths);
+  injectSharepointData(projectDetail.urls, filePaths, docPaths, spBatchFiles);
+  const fgSpBatchFiles = await getSpFiles(docPaths, true);
+  injectSharepointData(projectDetail.urls, filePaths, docPaths, fgSpBatchFiles, true);
+}
+
+async function initProject() {
+  if (project) return project;
+  const config = await getConfig();
+  const urlInfo = getUrlInfo();
+  if (!urlInfo.isValid()) {
+    throw new Error('Invalid Url Parameters that point to project file');
+  }
+
+  // helix API to get the details/status of the file
+  const hlxAdminStatusUrl = getHelixAdminApiUrl(urlInfo, config.admin.api.status.baseURI);
+
+  // get the status of the project file
+  const projectFileStatus = await getProjectFileStatus(hlxAdminStatusUrl, urlInfo.sp);
+  if (!projectFileStatus || !projectFileStatus?.webPath) {
+    throw new Error('Project file does not have valid web path');
+  }
+
+  const projectPath = projectFileStatus.webPath;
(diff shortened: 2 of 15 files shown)


amauch-adobe (approved):
Only one comment or question about the drafts directory usage, otherwise looks good to me.

tools/floodgate/js/config.js
```diff
@@ -0,0 +1,32 @@
+import {
+  getUrlInfo,
+} from '../../loc/utils.js';
+import {
+  fetchConfigJson,
+  getSharepointConfig,
+  getHelixAdminConfig,
+} from '../../loc/config.js';
+
+const FLOODGATE_CONFIG = '/drafts/floodgate/configs/config.json';
```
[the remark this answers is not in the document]
sukamat: @amauch-adobe It is a temp location for now similar to [Milo Loc](https://github.com/adobecom/milo/blob/main/tools/loc/config.js#L15). It will most probably move to AIO in the near future once we have a better location for storing configs like these.

.eslintrc.js
```diff
@@ -33,6 +33,7 @@ module.exports = {
   ignorePatterns: [
     '/libs/deps/*',
     '/tools/loc/*',
+    '/tools/floodgate/*',
```
meganthecoder, line 36: Why are you omitting it from linting rules? Usually this is reserved for external libraries, etc

sukamat: @meganthecoder Thanks for taking a look.

I am mostly following the model followed by Milo localization.
AFAIK, some of these configs were updated to unblock the feature from being implemented on the client side. But it is not scalable on the client-side. Once the basic functionality is working for FG and Loc, most of the related code currently in the repo will move to AIO.

@janaki-r-bhagwath who worked on Milo Loc might have additional context and reasoning behind it.

sukamat: @janaki-r-bhagwath  Was the eslint config updated to avoid the helix libs? or is there more to it?
Could this be removed for FG?

janaki-r-bhagwath: I did not add the eslint skip for loc - usually when I do local development I ensure to enable it so all loc files are lint compliant. But yes loc has prebuilt helix libraries which are not needed in Floodgate. 
So don't think you need to skip linting in floodgate.

tools/floodgate/js/config.js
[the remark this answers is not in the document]
sukamat: @janaki-r-bhagwath Agree, both FG and loc configs should be moved to a better location.

In case of Floodgate, it is just the "helix-sp" sheet info that is needed with some FG specific stuff which is what I have added in a separate config excel file [here](https://adobe.sharepoint.com/:x:/r/sites/adobecom/_layouts/15/Doc.aspx?sourcedoc=%7B8215365C-10F4-48D7-A41D-FB4B239DF6C6%7D&file=config.xlsx&action=default&mobileredirect=true)

In case of FG, as most of the info in that sheet are just required to run the sharepoint APIs, ideally all of it can be moved to AIO (or perhaps in a central location that AIO can access). This may not apply to loc.

Perhaps we will have a better understanding of where this central location for loc and FG configs can be once we start exploring the AIO side of things.

.eslintrc.js
sukamat: @janaki-r-bhagwath @meganthecoder 
Removed the path from the lint configs. Could you guys take a look again? Thanks!

sukamat: @janaki-r-bhagwath @meganthecoder Lint errors fixed and pushed

tools/floodgate/js/ui.js
```diff
@@ -0,0 +1,103 @@
...
+function createTableWithHeaders() {
+  const $table = createTag('table');
+  const $tr = createTag('tr', { class: 'header' });
+  $tr.appendChild(createColumn('Source URL', 'header'));
+  $tr.appendChild(createColumn('Source File', 'header'));
+  $tr.appendChild(createColumn('Source File Info', 'header'));
+  $tr.appendChild(createColumn('Floodgated Content', 'header'));
+  $tr.appendChild(createColumn('Floodgated File Info', 'header'));
+  $table.appendChild($tr);
+  return $table;
```
meganthecoder, line 50: Great to see `createTag`. FYI in milo, we don't prefix our dom element variables with `$`. Not sure if there's a rule against it, but it's not a pattern we follow

tools/floodgate/js/ui.js
```diff
@@ -0,0 +1,103 @@
...
+  }
+  return fgDocStatus.msg;
+}
+
+async function updateProjectDetailsUI(projectDetail, config) {
+  if (!projectDetail || !config) {
+    return;
+  }
+
+  const container = document.getElementsByClassName('project-detail')[0];
```
meganthecoder, line 67: Super nit: Shorter
```suggestion
  const container = document.querySelector('.project-detail');
```

tools/floodgate/js/utils.js
```diff
@@ -0,0 +1,17 @@
+export function getFloodgateUrl(url) {
+  if (!url) {
+    return undefined;
```
meganthecoder, line 3: Nit: Redundant. By default it returns `undefined`
```suggestion
    return;
```

meganthecoder (commented):
Good stuff. I had a few suggestions

tools/floodgate/js/utils.js
sukamat: @meganthecoder Eslint throws  `consistent-return` error if I remove undefined. So I left it as-is. LMK if there is a better way to handle this.

tools/floodgate/js/ui.js
sukamat: Loc and Floodgate share similar UI and they will be ported over to use Preact soon. All of this UI specific code will be removed at that point. Jingle is currently working on creating the layout. In any case, I have removed the usage of `$`.

tools/floodgate/js/project.js
```diff
@@ -0,0 +1,132 @@
...
+}
+
+async function updateProjectWithDocs(projectDetail) {
+  if (!projectDetail || !projectDetail?.filePaths) {
+    return;
+  }
+  const { filePaths } = projectDetail;
+  const docPaths = [...filePaths.keys()];
+  const spBatchFiles = await getSpFiles(docPaths);
+  injectSharepointData(projectDetail, filePaths, docPaths, spBatchFiles);
```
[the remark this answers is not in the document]
sukamat: Thanks! Modified it slightly to pass just the urls instead of the entire `projectDetail` object.

tools/floodgate/js/project.js
```diff
@@ -0,0 +1,129 @@
...
+
+let project;
+
+/**
+ * Makes the sharepoint file data part of `projectDetail` per URL.
+ */
+function injectSharepointData(projectUrls, filePaths, docPaths, spBatchFiles, isFloodgate) {
+  spBatchFiles.forEach((spFiles) => {
+    if (spFiles && spFiles.responses) {
+      spFiles.responses.forEach(({ id, status, body }) => {
```
chrischrischris, line 23: nit:
```suggestion
    if (!spFiles?.responses) return;
    spFiles.responses.forEach(({ id, status, body }) => {
```

chrischrischris (approved):

meganthecoder (approved):

janaki-r-bhagwath:
Verified that the floodgate UI reads information from the project files and display status of the source and floodgate content. 
Currently only the view of floodgate UI is available with no actions like copy, promote - confirmed the view is rendered. 

image
image

Validated that there are no regression issues on loc due to the addition of the floodgate param and movement of functions to util file.

janaki-r-bhagwath (approved):

PR #457, opened 2023-02-08
MWPW-124262 PDF Viewer Analytics

meganthecoder:
* Send PDF analytics events to AEP

Resolves: [MWPW-124262](https://jira.corp.adobe.com/browse/MWPW-124262)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/methomas/pdf-viewsdk
- After: https://methomas-pdf-analytics--milo--adobecom.hlx.page/drafts/methomas/pdf-viewsdk

Note: The pdf will show an error on the test URL because the test page domain doesn't match the client id. This is expected and won't occur on prod.


meganthecoder:
I'm adding the verified label on behalf of Sunita and Uma from the enablement team. I worked with them prior to creating the PR because I needed to use a temporary client id for test page that shouldn't be added to the codebase. They gave their approval via slack.

sanjayms01 (approved):

chrischrischris (approved):

PR #458, opened 2023-02-08
Library Search UI Fix

rgclayton:
Fix for block list cutting off the last item when search is in view.

**Not fixed**
image

**Fixed**
image

**Test URLs:**
- After: https://library-search-update--milo--adobecom.hlx.page/tools/library?martech=off


chrischrischris (approved):

amauch-adobe (approved):
Verified against all browsers, and everything looks good.

PR #467, opened 2023-02-10
Share block updates, Reading Time (inline)

So1ene:
### 1. New platform support added: Reddit ###
- Test url: [before](https://main--stock--adobecom.hlx.page/drafts/solene/share-reddit?martech=off) / [after](https://main--stock--adobecom.hlx.page/drafts/solene/share-reddit?martech=off&milolibs=share-block-updates)

---

### 2. New variants added: `Share (Inline)` & `Reading Time (Inline)` ###

- XD Mockup ("Something like this"): https://xd.adobe.com/view/0701df61-661f-40a9-96f6-59ca74724616-5252/
I was given permission by the stock team to use the current share block's icons instead of the ones from the mockup

- Test url: [before](https://main--stock--adobecom.hlx.page/drafts/solene/share-reading-time?martech=off) / [after](https://main--stock--adobecom.hlx.page/drafts/solene/share-reading-time?martech=off&milolibs=share-block-updates)

### Authoring: ###
image

### Section Styles: ###

- Default (left-aligned):
   desktop
   image
   mobile
   image

- Section metadata style center (center-aligned):
   desktop
   image
   mobile
   image

Other test URLs for QA purposes
- Only reading-time (inline): https://main--stock--adobecom.hlx.page/drafts/solene/reading-time-inline?milolibs=share-block-updates
- Only share (inline): https://main--stock--adobecom.hlx.page/drafts/solene/share-inline?milolibs=share-block-updates
- Both: https://main--stock--adobecom.hlx.page/drafts/solene/share-reading-time?milolibs=share-block-updates
- No sections between:  https://main--stock--adobecom.hlx.page/drafts/solene/share-reading-time-no-sections?martech=off&milolibs=share-block-updates

- Localization test: https://main--stock--adobecom.hlx.page/es/pages/drafts/solene/share-reading-time?milolibs=share-block-updates


auniverseaway (changes requested):
I'm not a fan of this. Especially when we have other ways we can solve this.

Block variants should not be so tightly coupled to other blocks.

If we were to use this pattern for other blocks, we would have an explosion of block variants...

share (aside)
share (accordion)
share (z-pattern)

The list goes on.

I don't think we should mistake layout patterns with a variant of a block. The variant of a block should be how to control the block itself, not how it should be organized next to other blocks.

auniverseaway:
If we find that share needs to be next to another block, I think we either:

1. Use section-metadata to organize them using existing patterns (2-up)
2. Convert blocks that are more likely to be inline into a link block.

I don't want us introducing _another_ way to put two blocks next to each other. I'm pretty sure we already have two ways:
```style | 2-up```
or
```layout | 1, 1```

**Comment by So1ene** [comment]
> If we find that share needs to be next to another block, I think we either:
> 
> 1. Use section-metadata to organize them using existing patterns (2-up)
> 2. Convert blocks that are more likely to be inline into a link block.
> 
> I don't want us introducing _another_ way to put two blocks next to each other. I'm pretty sure we already have two ways: `style | 2-up` or `layout | 1, 1`

I had the same idea but when I discussed it in the milo-community channel I was convinced otherwise: https://adobestock.slack.com/archives/C03B0BUA75E/p1675707753596839
both Ryans seem to like the variant solution, and if thinking back to the 'priority list', if I remember correctly the author experience is the most important, and comparing these 2 author experiences, it's clear to see that one is much better than the other:
image
image

**Comment by williambsm** [comment]
@auniverseaway just a slight nudge here for feedback, we would love to keep the momentum going!

**Comment by rgclayton** [comment]
@So1ene correct me if I'm wrong here. But, I think I remember the main problem with using `style | two up` was that it wasn't able to achieve the desired layout. It does put the reading time next to the share icons but didn't look anything like the comps. I put together a [quick sample to illustrate this approach](https://main--milo--adobecom.hlx.page/drafts/rclayton/test/share-and-read).

The other option discussed was adding code to automagically adjust the layout if the two blocks were in a `two up` column and the share block was the first block and the reading time was the second block. 

To me this seemed like a bit of an authoring problem because there is no indication that that the reading time and share blocks needed to be in a certain order and that the code would just change the `two up` column layout with out any mention. Didn't seem very intuitive.

The variant option was discussed so authors could explicitly achieve this layout with out confusion.

Maybe we need to explore more ideas to fine tune this? 

cc: @auniverseaway @ryanmparrish

**Comment by rgclayton** [comment]
@So1ene Chris and I briefly chatted about this. The idea is adding an `inline` variant to the blocks to make them `display: inline-block`, or something related. Any blocks that get the inline variant would display on the same row space. This would create a pretty simple authoring pattern and dev pattern for blocks wanting to behave like this.

You would need to come up with something to get the border to the left of the read time. You could possibly create styles that look for the the read time block to be the direct sibling of the share block. Thinking out loud here. 

I made a quick POC using the developer tools [on this page](https://main--milo--adobecom.hlx.page/drafts/rclayton/test/share-and-read).

Screenshot of POC with
image


cc: @auniverseaway @williambsm

**Comment by So1ene** [comment]
> @So1ene Chris and I briefly chatted about this. The idea is adding an `inline` variant to the blocks to make them `display: inline-block`, or something related. Any blocks that get the inline variant would display on the same row space. This would create a pretty simple authoring pattern and dev pattern for blocks wanting to behave like this.
> 
> You would need to come up with something to get the border to the left of the read time. You could possibly create styles that look for the the read time block to be the direct sibling of the share block. Thinking out loud here.
> 
> I made a quick POC using the developer tools [on this page](https://main--milo--adobecom.hlx.page/drafts/rclayton/test/share-and-read).
> 
> Screenshot of POC with image
> 
> cc: @auniverseaway @williambsm

A fantastic solution.
I implemented this change, please take a look again!~

File: libs/blocks/reading-time/reading-time.css [hunk 1]
```diff
@@ -12,3 +12,45 @@ main > .section > .content.reading-time > span {
...
+.reading-time.inline {
+  font-style: italic;
+  font-weight: 700;
+  padding-left: 20px;
+  margin-left: 20px;
+  border-left: 1px solid #adadad;
+}
+
+.reading-time.inline > span {
+  color: #adadad;
```
rgclayton, line 42: Text color changes for inline variants?

libs/blocks/reading-time/reading-time.css
```diff
@@ -12,3 +12,45 @@ main > .section > .content.reading-time > span {
...
+
+main > .section > .content.inline-wrapper > .reading-time.inline {
+  margin: 0;
+  order: 1;
+  margin-left: 20px;
+}
+
+.reading-time.inline {
+  font-style: italic;
+  font-weight: 700;
```
rgclayton, line 35: Different text styling for inline variant?

rgclayton (commented):
Looking better. 

I do think the css is heavy handed to make these blocks inline. For the most part you can get the same affect by applying `display: inline-block` to both the share and reading time blocks. Not sure why all of the flexbox is needed. It seems overly complicated for what you are trying to achieve.

rgclayton:
Let me know if setting up quick call to chat would be helpful.

libs/blocks/share/share.css
```diff
@@ -119,3 +119,46 @@
 .share .copy-to-clipboard.copy-to-clipboard-copied::before {
   content: attr(data-copied);
 }
+
+.share.inline {
+  display: inline-block;
```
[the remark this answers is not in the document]
So1ene: I wasn't able to center them without this

libs/blocks/reading-time/reading-time.css
So1ene: Yes, to match the mockup

libs/blocks/share/share.css
```diff
@@ -119,3 +119,46 @@
...
+  padding: 8px 0 12px;
+}
+
+main > .section.center > .content.inline-wrapper {
+  justify-content: center;
+}
+
+main > .section > .content.inline-wrapper > .share.inline {
+  margin: 0;
+  order: -1;
```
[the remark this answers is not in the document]
So1ene: Why?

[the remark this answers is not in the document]
So1ene: Fixed

libs/blocks/reading-time/reading-time.css
So1ene: Fixed, made all variants the same

libs/blocks/reading-time/reading-time.css
So1ene: Fixed, made all variants the same

So1ene:
@rgclayton please check again, I updated the test url with the different authoring orders: https://main--stock--adobecom.hlx.page/drafts/solene/share-reading-time?milolibs=share-block-updates

libs/blocks/reading-time/reading-time.css
```diff
@@ -12,3 +14,52 @@ main > .section > .content.reading-time > span {
 .center .reading-time > span {
   text-align: center;
 }
+
+main > .section > .content.inline-wrapper {
+  display: flex;
+  align-items: center;
+  flex-wrap: wrap;
+  padding: 8px 0 12px;
+}
```
rgclayton, line 23: There is a duplicate of this in the [Share css](https://github.com/adobecom/milo/pull/467/files#diff-f5946602c5148d0f9a4f0705c1e71bcc2446191a670372af4fe68c519d494c8dR139-R144 ). Is it needed in both files?

So1ene: I figured I would put it in both blocks, in case one block is loaded but not the other or vice versa

So1ene: Only `reading-time (inline)` loaded: https://main--stock--adobecom.hlx.page/drafts/solene/reading-time-inline?milolibs=share-block-updates
Only `share (inline)` loaded: https://main--stock--adobecom.hlx.page/drafts/solene/share-inline?milolibs=share-block-updates

So1ene:
I'll add the tests too!

libs/blocks/reading-time/reading-time.css
rgclayton: Hmm. What are we gaining by only having one block set to the `inline` variant? 

The real magic happens when two blocks are set to `inline` and they sit in the same row. Which, means there are  unnecessary [DOM updates](https://github.com/adobecom/milo/pull/467/files/fbb582fc882c7a9ce5b784501bc0328f1938c2b2#diff-4995aeeb074dd406bdcb22a5460207f2228dda32b90545b35f1df684658e6760R34) happening if only one block is set to `inline`.  I think you can look for the existence of a sibling element with the `inline` variant and only update the DOM if there is.

So1ene: Good point! It's now fixed so that it won't affect the DOM if it's a single sibling.

libs/blocks/share/share.css
```diff
@@ -119,3 +119,72 @@
...
+  margin: 0 auto;
+  justify-content: center;
+}
+
+main > .section > .inline-wrapper {
+  display: flex;
+  align-items: center;
+  flex-wrap: wrap;
+  padding: 8px 0 12px;
+}
```
rgclayton, line 149: Still have a duplicate of this in `reading-time.css`.
If that container is only applied when there are inline siblings I think it makes sense to only include this snippet of css in one place. This will help avoid confusion in the event something needs to change or be adapted with this.

image

libs/blocks/reading-time/reading-time.js
```diff
@@ -28,5 +30,19 @@ export default async function init(el) {
...
+    if (inlineSiblinggs.length > 1) {
+      let inlineContainer = section.querySelector('.inline-wrapper');
+      if (!inlineContainer) {
+        inlineContainer = createTag('div', { class: 'inline-wrapper content' });
+        block.after(inlineContainer);
+      }
+      inlineSiblinggs.forEach((el) => inlineContainer.append(el));
+    } else {
+      block.classList.add('content');
+    }
```
rgclayton, line 46: This whole block of code is the exact same as [what's in share.js](https://github.com/adobecom/milo/pull/467/files#diff-71d2d1ba42c460262b54dc2d8c6a6195d8112dd9b65e509dcbb19d979eb22d83R122-R133).
So, the exact code runs multiple times every time there are sibling inline variants. 

We need a better way to handle this. Inline-utility.js or something.

rgclayton:
Looking better!

A few layout issues I noticed. If you just have the share block or the reading-time block there is a bunch of spacing/margin above it. If the `inline` variant is added that removes the spacing. But, it doesn't make sense to author the inline variant for only one block - share or reading-time.

image

image

rgclayton:
On mobile viewport if the share block is first, followed by the reading-time block, the divider line doesn't span the full width of the share block. Wanted to check if that's expected?

image

So1ene:
> On mobile viewport if the share block is first, followed by the reading-time block, the divider line doesn't span the full width of the share block. Wanted to check if that's expected?
> 

I don't mind it personally, it's due to selecting the "next element" with `+` in css:
So in this case the share block has the border-top
``` CSS
@media screen and (max-width: 600px) {
  main > .section > .inline-wrapper > .inline + .inline {
    border-left: none;
    border-top: 1px solid #adadad;
  }
}
```
Hope that makes sense

---

> 
> A few layout issues I noticed. If you just have the share block or the reading-time block there is a bunch of spacing/margin above it. If the `inline` variant is added that removes the spacing. But, it doesn't make sense to author the inline variant for only one block - share or reading-time.

Oh I think this is just a `stock` repo style because Artisthub uses its own template for creating section spacing, if you test the same issue from the milo repo it should be working correctly

rgclayton:
@So1ene This is how it looks in Milo. Is this correct?

https://share-block-updates--milo--adobecom.hlx.page/drafts/rclayton/test/share-and-read

libs/blocks/share/inline-utils.js
```diff
@@ -0,0 +1,18 @@
+import { createTag } from '../../utils/utils.js';
+
+export function inlineBlock(block) {
+  if (!block.classList.contains('inline')) return;
+
+  const section = block.closest('.section');
+  const inlineSiblinggs = section.querySelectorAll('.inline');
+  if (inlineSiblinggs.length > 1) {
```
chrischrischris, line 8: If there are multiple inline blocks in a section that are not next to each other, they will be moved to a new order and placed next to each other.  See:

https://share-block-updates--milo--adobecom.hlx.page/drafts/cpeyer/bleh

Screenshot 2023-03-20 at 10 13 17 AM

Where the 2 inline blocks are in different parts of the section.  Are we ok with merging into one location?

So1ene: Alright, it should be fixed now: 

https://share-block-updates--milo--adobecom.hlx.page/drafts/cpeyer/bleh

https://main--stock--adobecom.hlx.page/drafts/solene/share-reading-time-no-sections?martech=off&milolibs=share-block-updates

https://main--stock--adobecom.hlx.page/drafts/solene/share-reading-time?martech=off&milolibs=share-block-updates

libs/blocks/share/inline-utils.css
```diff
@@ -0,0 +1,45 @@
...
+    align-items: center;
+  }
+
+  main > .section > .inline-wrapper > .inline + .inline {
+    border-left: none;
+    padding-left: 0;
+    margin-left: 0;
+    border-top: 1px solid #adadad;
+  }
+}
```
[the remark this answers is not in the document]
ryanmparrish: I agree, This would be a great universal section style too 'inline' content and have the ability to pass justification as needed. Here's a future page example that could benefit from this. see the nemonic list. 

Screen Shot 2023-03-22 at 1 06 30 PM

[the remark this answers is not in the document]
So1ene: Sounds great~ Which folders should I move `inline-utils.css` and `inline-utils.js` to?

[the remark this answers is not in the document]
rgclayton: @chrischrischris Does it make sense to rename this to `inline.js` and put it in `/libs/utils/`? That's sorta where I'm thinking.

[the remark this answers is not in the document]
rgclayton: > I agree, This would be a great universal section style to 'inline' content and have the ability to pass justification as needed. Here's a future page example that could benefit from this. see the nemonic list.
> 
> Screen Shot 2023-03-22 at 1 06 30 PM

Agreed! I think this is a great use case for that new pattern that's popping up.

libs/utils/inline.css
```diff
@@ -0,0 +1,46 @@
...
+    align-items: center;
+  }
+
+  main > .section > .inline-wrapper > .inline + .inline {
+    margin-left: 0;
+    padding-left: 0;
+    margin-right: 0;
+    padding-right: 0;
+  }
+}
```
rgclayton, line 46: The CSS should go inside of `/libs/styles/`.  `libs/utils` is for .js files.

rgclayton (approved):
I think we've made it to the finish line!

ryanmparrish (approved):

rgclayton:
@auniverseaway You have a change request on this PR still. I think we have this PR in a much better state now. Mind taking one last look?

auniverseaway:
Screen Shot 2023-04-11 at 9 38 56 AM
This should not be here.

So1ene:
Oops I did it again 🙈 don't know how that got there, sorry!

honstar:
@auniverseaway , can you please review your last change request? This has been addressed and the PR is otherwise ready to go. Thx

PR #470, opened 2023-02-13
MWPW-125513 - Cookie preferences bug

meganthecoder:
* Fix cookie preferences trigger to work on all geos

Resolves: [MWPW-125513](https://jira.corp.adobe.com/browse/MWPW-125513)

**Test URLs:**
- Before: https://main--bacom--adobecom.hlx.page/kr/customer-success-stories?martech=off
- After: https://main--bacom--adobecom.hlx.page/kr/customer-success-stories?milolibs=methomas-cookie-modal-bug&martech=off
- After: https://main--bacom--adobecom.hlx.page/customer-success-stories?milolibs=methomas-cookie-modal-bug&martech=off
- After: https://methomas-cookie-modal-bug--milo--adobecom.hlx.page/?martech=off

Note: You must wait until `privacy-standalone.js` has loaded for the modal trigger to work (abt 3 secs)

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -592,14 +592,19 @@ export async function loadDeferred(area, blocks, config) {
   });
 }
 
-function loadPrivacy() {
+export function loadPrivacy() {
   window.fedsConfig = {
     privacy: {
       otDomainId: '7a5eb705-95ed-4cc4-a11d-0cc5760e93db',
-      footerLinkSelector: '[href="https://www.adobe.com/#openPrivacy"]',
     },
   };
   loadScript('https://www.adobe.com/etc.clientlibs/globalnav/clientlibs/base/privacy-standalone.js');
+
+  const privacyTrigger = document.querySelector('footer a[href$="#openPrivacy"]');
+  privacyTrigger?.addEventListener('click', (event) => {
+    event.preventDefault();
+    window.adobePrivacy?.showPreferenceCenter();
+  });
 }
 
 function initSidekick() {
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -600,7 +600,7 @@ export function loadPrivacy() {
   };
   loadScript('https://www.adobe.com/etc.clientlibs/globalnav/clientlibs/base/privacy-standalone.js');
 
-  const privacyTrigger = document.querySelector('footer a[href$="#openPrivacy"]');
+  const privacyTrigger = document.querySelector('footer a[href*="#openPrivacy"]');
   privacyTrigger?.addEventListener('click', (event) => {
     event.preventDefault();
     window.adobePrivacy?.showPreferenceCenter();
--- a/test/utils/utils.test.js
+++ b/test/utils/utils.test.js
@@ -9,18 +9,7 @@ const config = {
   locales: { '': { ietf: 'en-US', tk: 'hah7vzn.css' } },
 };
 
-document.head.innerHTML = await readFile({ path: './mocks/head.html' });
-document.body.innerHTML = await readFile({ path: './mocks/body.html' });
-
 describe('Utils', () => {
-  before(() => {
-    sinon.spy(console, 'log');
-  });
-
-  after(() => {
-    console.log.restore();
-  });
-
   before(async () => {
     const module = await import('../../libs/utils/utils.js');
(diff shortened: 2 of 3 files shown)


libs/utils/utils.js
```diff
@@ -592,14 +592,19 @@ export async function loadDeferred(area, blocks, config) {
   });
 }
 
-function loadPrivacy() {
+export function loadPrivacy() {
   window.fedsConfig = {
     privacy: {
       otDomainId: '7a5eb705-95ed-4cc4-a11d-0cc5760e93db',
-      footerLinkSelector: '[href="https://www.adobe.com/#openPrivacy"]',
```
vhargrave, line 599: I think instead of adding our own event listener, we just need to fix the footerLinkSelector to something like `footerLinkSelector: '[href*="#openPrivacy"]',`
This fixes the issue with the modal not opening - 
https://main--bacom--adobecom.hlx.page/kr/customer-success-stories?milolibs=vhargrave-privacy-test&martech=off

The link text is then being translated by the library, but maybe there's a config option we can pass so this doesn't happen. @narcis-radu & @overmyheadandbody - is something like that possible?

narcis-radu: @vhargrave - there's no option to skip translation, the only feature we have is to use page language (instead of browser language). If no `footerLinkSelector` is provided (or the associated DOM element can't be found), then the event listener has to be set manually, just like @meganthecoder did.
Since this is implemented in Milo, I would like to go one step further: instead of adding an event listener to a DOM element, let's listen to a hash change and allow the Preference Center modal to be opened from the page content as well. This will get used on pages like https://www.adobe.com/privacy/opt-out.html.

meganthecoder: @narcis-radu I prefer attaching the listener to the DOM element for performance reasons. Milo uses hashes for modals which are fairly popular content-wise. I'd rather not do a check for openPrivacy every time, especially  because we don't have a specific need to open the modal from anywhere other than the footer.

test/utils/utils.test.js
```diff
@@ -329,4 +329,16 @@ describe('Utils', () => {
     });
     expect(io instanceof IntersectionObserver).to.be.true;
   });
+
+  it('adds privacy trigger to cookie preferences link in footer', () => {
+    window.adobePrivacy = { showPreferenceCenter: sinon.spy() };
+    if (!document.querySelector('footer')) {
+      const footerHtml = '<footer><a href="https://www.adobe.com/#openPrivacy" id="privacy-link">Cookie preferences</a></footer>';
+      document.body.insertAdjacentHTML('beforeEnd', footerHtml);
+    }
```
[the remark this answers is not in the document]
meganthecoder: Thanks for pressing on this. I think I came to a better solution.

Basically the problem was that the `document.body` was being added at the top of the file, being shared by all tests, but one test removed the footer. Since the test ran async, I was getting different results and fails per run.

To fix this, I wrapped all of the tests sharing the same `document.body` in a `describe` and separated out my test with its own `document.body`.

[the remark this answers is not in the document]
JasonHowellSlavin: Awesome! Glad you found a good solution!

JasonHowellSlavin (approved):
Looks good to me. I know the discussion re: the listener may cause some refactoring, will take another look if things change.

libs/utils/utils.js
```diff
@@ -592,14 +592,19 @@ export async function loadDeferred(area, blocks, config) {
...
     },
   };
   loadScript('https://www.adobe.com/etc.clientlibs/globalnav/clientlibs/base/privacy-standalone.js');
+
+  const privacyTrigger = document.querySelector('footer a[href$="#openPrivacy"]');
```
vhargrave, line 603: nit: I'd make the selector look for a hashtag anywhere in the link 
`a[href*="#openPrivacy"]`
Just in case a descriptive attribute is added such as #_dns (do not style) or who knows what else might be used in the future. This makes everything a little more robust.

vhargrave (approved):
Rest looks great - nice job!

libs/utils/utils.js
narcis-radu: makes sense, go for it!

narcis-radu (approved):

PR #473, opened 2023-02-14
Refactor profile and placeholders

mokimo:
### Description
- Move sign in button logic to its own file, decouple the CSS from the other menu items
- Move the profile button and profile creation into it's own file 
- Use placeholders in the profile
- Create a FEDS-based placeholder utility function as we want placeholders to be authored centrally and not on a consumer basis. For this reason, https://github.com/adobecom/milo/pull/391 wouldn't be necessary anymore from a FEDS/GNAV perspective.

Resolves: [126117](https://jira.corp.adobe.com/browse/MWPW-126117)
Closes https://github.com/adobecom/milo/pull/391

**Test URLs:**
- Before: https://feds-gnav--milo--adobecom.hlx.page/ch_de/drafts/osahin/document?martech=off#
- After:  https://feds-improvements--milo--adobecom.hlx.page/ch_de/drafts/osahin/document?martech=off#


libs/blocks/global-navigation/blocks/profile/signIn.js
```diff
@@ -0,0 +1,30 @@
...
+  const signInWithDropDown = toFragment`<a href="#" daa-ll="${signInLabel}" class="feds-signin" role="button" aria-expanded="false" aria-controls="navmenu-profile" daa-lh="header|Close" aria-haspopup="true">${signInLabel}</a>`;
+  dropDown.classList.add('feds-signin-dropdown');
+  decoratedEl.insertAdjacentElement('beforeend', dropDown);
+
+  // TODO we don't have a good way of adding attributes to links
+  const dropDownSignIn = dropDown.querySelector('[href="https://adobe.com?sign-in=true"]');
+  if (dropDownSignIn) {
+    dropDownSignIn.addEventListener('click', (e) => {
+      e.preventDefault();
+      window.adobeIMS.signIn();
```
overmyheadandbody, line 22: Since the profile block might eventually be exposed to non-Milo consumers, maybe it makes sense to check that `adobeIMS.signIn` exists as a function before calling it?

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -113,15 +110,15 @@ class Gnav {
...
+        loadStyles('./blocks/profile/profile.css'),
+        loadStyles('./blocks/navMenu/menu.css'),
+        loadStyles('./blocks/search/gnav-search.css'),
       ]);
       this.menuControls = new MenuControls(this.curtain);
       this.decorateMenu = decorateMenu;
       this.decorateLargeMenu = decorateLargeMenu;
       this.appLauncher = appLauncher;
-      this.Profile = Profile;
+      this.blocks.profile.Profile = Profile;
```
overmyheadandbody, line 121: Not a big fan of `profile.Profile` 😬 Could we switch it up a bit?

libs/blocks/global-navigation/utilities.js
```diff
@@ -35,3 +38,71 @@ export function getLocale() {
...
+}
+
+const keyToStr = (key) => key.replaceAll('-', ' ');
+
+export async function getPlaceholder(key = '') {
+  if (typeof key !== 'string') return { value: '' };
+  const { locale } = getConfig();
+  const defaultLocale = 'en-US';
+  const placeholders = await fetchPlaceholders();
+  if (placeholders[key]) return placeholders[key];
```
overmyheadandbody, line 86: If the key is not valid, `{ value: '' }` is returned. If the key exists in the fetched placeholders, it looks like you're returning the value directly, not `{ value: 'string for placeholder' }`. Is this the case? Am I missing something from the data structure?

libs/blocks/global-navigation/blocks/profile/milo-wrapper.js
```diff
@@ -0,0 +1,39 @@
+import { toFragment, getPlaceholders } from '../../utilities.js';
+
+const initProfileButton = async ({
+  blockEl,
+  decoratedEl,
+  avatar,
+}) => {
+  const placeholders = await getPlaceholders(['profile-button']);
```
overmyheadandbody, line 8: Why `getPlaceholders` if we just need one?

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -152,77 +149,56 @@ class Gnav {
...
-        avatarImgEl,
-        sections,
-        profileRes,
-        profileButtonEl,
+    if (this.blocks.profile.profileRes.status !== 200) {
+      const [decorateSignIn] = await Promise.all([
+        loadBlock('./blocks/profile/signIn.js'),
+        loadStyles('./blocks/profile/signIn.css'),
+      ]);
+      const signInEl = await decorateSignIn({ blockEl, decoratedEl, toFragment, getPlaceholder });
```
overmyheadandbody, line 165: What's the reasoning behind sending utility methods as params rather than loading them where needed?

overmyheadandbody (commented):
Just a few comments for now

libs/blocks/global-navigation/utilities.js
mokimo: If it's not valid we return `{value: 'sign in'}` transforming the key (`sign-in`) 

If it is defined, we map whatever franklin delivers to us, so that this response
```json
{
 "total":5,
 "offset":0,
 "limit":5,
 "data":[{"key":"search","value":"Suchen"},{"key":"clear-results","value":"Ergebnisse löschen"},{"key":"try-advanced-search","value":"Erweiterte Suche"},{"key":"sign-in","value":"Anmelden", "link": "https://adobe.com/sign-in", analytics: "authored-analytics-label"},{"key":"view-account","value":"Account"}],
 ":type":"sheet"
}
```
turns to
```js
const placeholders = {
    "search": {
        "key": "search",
        "value": "Suchen"
    },
    "clear-results": {
        "key": "clear-results",
        "value": "Ergebnisse löschen"
    },
    "try-advanced-search": {
        "key": "try-advanced-search",
        "value": "Erweiterte Suche"
    },
    "sign-in": {
        "key": "sign-in",
        "value": "Anmelden",
        "analytics": "authored-analytics-label",
        "link": "https://adobe.com/sign-in"
    },
    "view-account": {
        "key": "view-account",
        "value": "Account"
    }
}
```
So `placeholders["sign-in"]` would be reference an object, so `{key: "sign-in", value: "Anmelden", ...rest}`

Originally I did that to retain the option to add more attributes, so if the placeholder sheet also contained a "link" column, or whatever attribute we want to pass, might be analytics related or any string, `placeholders["sign-in"]` would be `{key: "sign-in", value: "Anmelden", link: "https://adobe.com/sign-in", analytics: "special-authored-attribute"` instead of just referencing a string.

We don't use links now, but I don't think it hurts to stick to an object based structure so we can adapt to future changes more easily

libs/blocks/global-navigation/blocks/profile/milo-wrapper.js
mokimo: Frankly, didn't want to import it if we already have `getPlaceholders` - but there's no downside so I'll switch it up 😁

libs/blocks/global-navigation/blocks/profile/signIn.js
mokimo: So, there's a few different scenarios
1. Standalone Profile - In the old beagle-based profile, we'd only ship the profile itself - consumers cared for sign in themselves. (We can think about offering that too)
2. Standalone Nav - We strictly control IMS/Sign in flow, so it's available
3. Usual milo nav usecase - We strictly control IMS/Sign in flow, so it's available

So I'd like to keep it as is, just so that lana picks up any JS errors that we can monitor and fix (especially, but not only - when we launch)

libs/blocks/global-navigation/global-navigation.js
mokimo: Maybe `ProfileClass`? Open for suggestions!

It makes a bit more sense where it's used IMO
```js
const initProfileMenu = async ({ blockEl, ProfileClass, ...rest }) => {
  const placeholders = await getPlaceholders(['profile-button', 'sign-out', 'view-account', 'manage-teams', 'manage-enterprise']);
  const profile = new ProfileClass({
    localMenu: blockEl.querySelector('h5')?.parentElement,
    placeholders,
    ...rest,
  });
  return profile;
};
```

libs/blocks/global-navigation/blocks/profile/milo-wrapper.js
mokimo: Done!

libs/blocks/global-navigation/blocks/profile/signIn.js
```diff
@@ -0,0 +1,30 @@
...
+    });
+    return signIn;
+  }
+
+  const signInWithDropDown = toFragment`<a href="#" daa-ll="${signInLabel}" class="feds-signin" role="button" aria-expanded="false" aria-controls="navmenu-profile" daa-lh="header|Close" aria-haspopup="true">${signInLabel}</a>`;
+  dropDown.classList.add('feds-signin-dropdown');
+  decoratedEl.insertAdjacentElement('beforeend', dropDown);
+
+  // TODO we don't have a good way of adding attributes to links
+  const dropDownSignIn = dropDown.querySelector('[href="https://adobe.com?sign-in=true"]');
```
narcis-radu, line 18: I'm not sure I follow this selector; why are we using `?sign-in=true`?

libs/blocks/global-navigation/blocks/profile/signIn.js
narcis-radu: There's one more use-case: imslib is loaded through npm and `adobeIMS` object is not available. In that case we might need to trigger some events to get the lib instance. Not exactly something we want to focus on right now, but add a TODO

libs/blocks/global-navigation/global-navigation.js
narcis-radu: We should revert and spend some time investigating how to mock

libs/blocks/global-navigation/utilities.js
narcis-radu: Just to recap and make sure I understand:
`placeholders["sign-in"]` will return:
1. `{key: "sign-in", value: "Something"}` if the key is found
2. `{key: "sign-in", value: "sign in"}` if the key is not found

`placeholders["sign in"]` will return:
3. `{value: ""}` since the key is not valid

mokimo: Yes, almost.. the gist it it should always return `value: ""` so u never need to worry about access to something undefined
`placeholders["sign-in"]` will return:
1. `{key: "sign-in", value: "Something-language-specific"}` if the key is found
2. `{key: "sign-in", value: "Something-english-backup"}` if the key was not found for a specific language, we try to fetch the English version
3. `{value: "sign in"}` if the key is not found

`placeholders["sign in"]` will return:
4. `{value: ""}` since the key is not valid

libs/blocks/global-navigation/blocks/profile/signIn.js
mokimo: We use it to know which link is used for signing in (and add an eventListener for IMS) instead of following the link.

There's a TODO above `// TODO we don't have a good way of adding attributes to links`

narcis-radu: I saw the todo, but did not understand the use-case; now I get it, thanks!

libs/blocks/global-navigation/global-navigation.js
mokimo: Properly set this up now

libs/blocks/global-navigation/blocks/profile/signIn.js
mokimo: Added a todo

narcis-radu (approved):

narcis-radu (approved):

libs/blocks/global-navigation/blocks/search/gnav-search.js
```diff
@@ -35,10 +35,8 @@ class Search {
   }
 
   async getLabels() {
-    const config = getConfig();
-    const sheet = 'feds';
     this.labels = {};
-    [this.labels.search, this.labels.clearResults, this.labels.tryAdvancedSearch] = await replaceKeyArray(['search', 'clear-results', 'try-advanced-search'], config, sheet);
+    [this.labels.search, this.labels.clearResults, this.labels.tryAdvancedSearch] = await replaceKeyArray(['search', 'clear-results', 'try-advanced-search'], getFedsPlaceholderConfig());
```
overmyheadandbody, line 39: 👍

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -418,6 +382,14 @@ header .app-launcher {
 
 /* END APP LAUNCHER */
 
+.feds-profile {
```
overmyheadandbody, line 385: This should be moved up. Right now it's in the `/* TODO: Old styles, to be refactored */` section. These styles can go on L186 with a `/* Profile */` comment, for consistency purposes.

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -101,27 +98,28 @@ class Gnav {
     // eslint-disable-next-line no-async-promise-executor
     this.ready = this.ready || new Promise(async (resolve) => {
       this.el.removeEventListener('click', this.loadDelayed);
+      // TODO load placeholders here? is that too late?
```
overmyheadandbody, line 101: I don't think we should force-load placeholders, each component could request them as needed.

overmyheadandbody (approved):
Looks good 👍

libs/blocks/global-navigation/global-navigation.css
mokimo: Good catch, changed 👍

libs/blocks/global-navigation/global-navigation.js
mokimo: Ah yes, this is a very early relict before I looked deeper into that - removed 👍

PR #469, opened 2023-02-11
MWPW-121123 - Aside Block Updates

Sartxi:
- Adjustments to button/link spacing and allowing links in the action areas
- Adds ability to change background color behind text area on Split Asides
- Also contains style updates to the notification aside in order to match consonant designs
- Verified mp4 videos get decorated and laid out properly

Resolves: [MWPW-121123](https://jira.corp.adobe.com/browse/MWPW-121123)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/sarchibeque/asides?martech=off
- After: https://mwpw-121123-aside-block-updates--milo--adobecom.hlx.page/drafts/sarchibeque/asides?martech=off

PR #478, opened 2023-02-15
MWPW-123645 & MWPW-123653 - [Blog2Milo]

saugatmalla:
* Decorates  mp4s uploaded using Franklin slack bot  

Resolves: MWPW-123645 & MWPW-123653 - [Blog2Milo]

**Test URLs:**
- Before: https://main--blog--adobecom.hlx.page/drafts/rclayton/videos?martech=off
- After: https://main--blog--adobecom.hlx.page/drafts/rclayton/videos?milolibs=video&martech=off


libs/utils/utils.js
```diff
@@ -411,6 +411,20 @@ export function decorateSVG(a) {
...
+export function decorateVideo(a) {
+  const { pathname, hash } = a;
+  if (!pathname.match('media_.*.mp4$')) return;
+
+  const isAutoplay = !!(hash?.includes('autoplay'));
+
+  const attrs = isAutoplay ? 'playsinline autoplay loop muted' : 'playsinline controls';
+  const video = `<video ${attrs}>
+        <source src=".${pathname}" type="video/mp4" />
+      </video>`;
```
JasonHowellSlavin, line 423: Not sure, but I think `createTag` is typically preferred, though I understand with `${attrs}` and `${pathname}`this might be the more performant way.

test/utils/mocks/body.html
```diff
@@ -26,6 +26,10 @@
     <p><a href="https://milo.adobe.com/img/favicon.svg">https://milo.adobe.com/img/favicon.svg</a></p>
     <p><a href="https://www.adobe.com">https://milo.adobe.com/img/favicon.svg</a></p>
     <p><a class="bad-url" href="https://www.adobe.com">img/favicon.svg</a></p>
+    <!-- Video -->
+    <p><a href="https://main--blog--adobecom.hlx.page/media_17927691d22fe4e1bd058e94762a224fdc57ebb7b.mp4">
+      https://main--blog--adobecom.hlx.page/media_17927691d22fe4e1bd058e94762a224fdc57ebb7b.mp4
+    </a></p>
```
JasonHowellSlavin, line 32: You add the anchor to mocks, but I don't see a corresponding test for `decorateVideo`?

libs/utils/utils.js
saugatmalla: Yeah, video `attributes` was not working properly since `createTag` takes object for attributes. Also, it's much cleaner in this situation.

test/utils/mocks/body.html
saugatmalla: I added a test.

test/utils/utils.test.js
```diff
@@ -200,7 +200,23 @@ describe('Utils', () => {
...
 
-  describe ('rtlSupport', () => {
+  describe('video uploaded using franklin bot', () => {
+    it('decorates video', async () => {
+      const block = document.createElement('div');
+      const a = document.createElement('a');
+      const href = 'https://main--blog--adobecom.hlx.page/media_17927691d22fe4e1bd058e94762a224fdc57ebb7b.mp4';
+      a.setAttribute('href', href);
+      a.textContent = href;
+      block.append(a);
```
JasonHowellSlavin, line 210: You likely don't need to create your block on the fly like this since line 13 sets the mock you edited to the innerHTML of the document: 
`
document.body.innerHTML = await readFile({ path: './mocks/body.html' })
`

saugatmalla: The mock video `a` tag in body.html gets decorated on line 147 as
`loadArea` ->  `decorateSections` -> `decorateLinks` -> `decorateVideo`. 
To mock that again, will need to import the original `body.html`.

JasonHowellSlavin: Ah, right you are! Very cool.

Seems to speak to the need to break some of these tests up, but that is beyond the scope of this ticket.

JasonHowellSlavin (approved):

chrischrischris (approved):

rgclayton (approved):

rgclayton:
Adding the "do not merge" while I sort a few things out related to https://github.com/adobecom/milo/pull/469

rgclayton:
@saugatmalla Can we make this url pattern work in this PR?
`https://images-tv.adobe.com/mpcv3/5574/b0a1f104-a964-4c2d-93f0-04819f31a158_1664235386.854x480at800_h264.mp4?autoplay=true`

This would cover the other use case we are seeing with video's coming from AdobeTV (MPC)in Sean's ticket https://github.com/adobecom/milo/pull/469.

Sample page with that same AdobeTv url in it:
https://main--blog--adobecom.hlx.page/drafts/rclayton/videos?milolibs=video

saugatmalla:
> @saugatmalla Can we make this url pattern work in this PR? `https://images-tv.adobe.com/mpcv3/5574/b0a1f104-a964-4c2d-93f0-04819f31a158_1664235386.854x480at800_h264.mp4?autoplay=true`

@rgclayton Yeah, I think we can add that in [`adobetv` auto block](https://github.com/adobecom/milo/blob/main/libs/blocks/adobetv/adobetv.js)? And in this PR, to support the aside block.

libs/utils/utils.js
```diff
@@ -412,6 +413,20 @@ export function decorateSVG(a) {
   }
 }
 
+export function decorateVideo(a) {
```
auniverseaway, line 416: Why can't this be an autoblock? Seems like that's basically what we're doing already.

The great part about autoblocks / link blocks is that they can run inside other blocks... marquee > inline-video.

libs/utils/utils.js
```diff
@@ -58,6 +58,7 @@ const MILO_BLOCKS = [
 ];
 const AUTO_BLOCKS = [
   { adobetv: 'https://video.tv.adobe.com' },
+  { adobetv: 'https://images-tv.adobe.com' },
```
auniverseaway, line 61: Why do we have to support both URLs?

libs/utils/utils.js
saugatmalla: How would you use regex (`media_.*.mp4$`) in auto block though? https://github.com/adobecom/milo/blob/a0b6e00dc8e9c773665f4d4743b77af53500af4c/libs/utils/utils.js#L418

rgclayton (approved):
This is looking good. Use cases for being able to add slack uploaded .mp4's to other blocks is working.

@auniverseaway @chrischrischris Mind giving this another look?

auniverseaway:
@rgclayton @saugatmalla 

1. I really dislike having two ways to author adobe tv links. I don't think we should add another link to utils for what is essentially a training issue.
2. I really dislike the decorateVideo function. I don't understand why we cannot use a link block for this feature. That's code that ships to every single milo page.

Can we address both of these today?

rgclayton:
> I really dislike having two ways to author adobe tv links. I don't think we should add another line to utils for what is essentially a training issue.
Ok. This one is probably my fault. Makes sense. I think we can make sure authors use the embed code reference for AdobeTV videos. 

cc: @saugatmalla

libs/utils/utils.js
honstar: I agree with @auniverseaway, having this as an autoblock would be great. How about supporting regular expressions for autoblocks by extending the `decorateAutoBlocks` method?
https://github.com/adobecom/milo/blob/main/libs/utils/utils.js#L419

PR #486, opened 2023-02-17
MWPW-125923 Fix Footer Titles

Brandon32:
* Clicking on footer titles reloads the page

Resolves: [MWPW-125923](https://jira.corp.adobe.com/browse/MWPW-125923)

**Test URLs:**
- After: https://bmarshal-footer-titles--milo--adobecom.hlx.page/?martech=off

**BACOM Test URLs:**
- Before: https://main--bacom--adobecom.hlx.page/customer-success-stories/the-home-depot-case-study?martech=off
- After: https://main--bacom--adobecom.hlx.page/customer-success-stories/the-home-depot-case-study?martech=off&milolibs=bmarshal-footer-titles

### DIFF
--- a/libs/blocks/footer/footer.js
+++ b/libs/blocks/footer/footer.js
@@ -95,12 +95,8 @@ class Footer {
       const headings = column.querySelectorAll('h2');
       headings.forEach((heading) => {
         const titleId = heading.textContent.trim().toLowerCase().replace(/ /g, '-');
-        const title = createTag('a', {
-          class: 'footer-nav-item-title',
-          role: 'button',
-          'aria-expanded': this.desktop.matches,
-          'aria-controls': `${titleId}-menu`,
-        });
+        const title = createTag('p', { class: 'footer-nav-item-title', id: titleId });
+
         title.textContent = heading.textContent;
 
         const navItem = createTag('div', { class: 'footer-nav-item' });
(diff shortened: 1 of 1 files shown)


libs/blocks/footer/footer.js
```diff
@@ -95,7 +95,7 @@ class Footer {
       const headings = column.querySelectorAll('h2');
       headings.forEach((heading) => {
         const titleId = heading.textContent.trim().toLowerCase().replace(/ /g, '-');
-        const title = createTag('a', {
+        const title = createTag('p', {
           class: 'footer-nav-item-title',
           role: 'button',
           'aria-expanded': this.desktop.matches,
```
meganthecoder, line 101: Setting it to a `p` tag with a role of `button` seems very strange. Is there a case where this needs to be clicked, like on mobile?

chrischrischris: In mobile, the titles are clickable to expand/collapse the subitems.

Since the aria role should not be button on desktop, what about using matchMedia to toggle the aria role and properties?  Then using `<p>` makes more sense.

vhargrave: agree with the comments above, you'd probably also want to modify the role on a mediaChange within the setDesktopTitles and setMobileTitles methods. 🙌

Brandon32: Sounds good, added set and remove attributes to setDesktopTitles and setMobileTitles methods so they are updated on mediaChange.

chrischrischris (approved):

meganthecoder (approved):

PR #492, opened 2023-02-20
MWPW-126479 Adding support for gradient background in section

aishwaryamathuria:
* Fix to allow addition of support in section metadata to allow users to add gradient in section background
* Feature is mentioned in milo docs https://milo.adobe.com/docs/library/blocks/section-metadata but the gradients are not added in section when provided in background option.

Resolves: [MWPW-126479](https://jira.corp.adobe.com/browse/MWPW-126479)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/mathuria/section-metadata-gradient?martech=off
- After: https://sectiongradient--milo--aishwaryamathuria.hlx.page/drafts/mathuria/section-metadata-gradient?martech=off


libs/blocks/section-metadata/section-metadata.js
```diff
@@ -7,7 +7,12 @@ function handleBackground(div, section) {
   } else {
     const color = div.textContent;
     if (color) {
-      section.style.backgroundColor = color;
+      if (color.indexOf('gradient') !== -1) {
+        section.style.background = color;
+      }
+      else {
+        section.style.backgroundColor = color;
+      }
```
chrischrischris, line 15: Since setting `background: colorName;` is the same as `backgroundColor: colorName` we should probably just place the `background` string directly.  This allows for any kind of background strings.

```
const background = div.textContent;
if (background) {
  section.style.background = background;
}
```

@rgclayton @auniverseaway @ryanmparrish Thoughts?

aishwaryamathuria: @chrischrischris Thanks for sharing the feedback. I have updated the PR to set background property in place of backgroundColor to allow all background strings. Could you please review?

chrischrischris (approved):

libs/blocks/section-metadata/section-metadata.js
aishwaryamathuria: @chrischrischris Thanks for adding the approval for the PR and reviewing the latest changes. 
Marking the conversation as resolved.

salonijain3 (approved):
Looks good!

vhargrave (approved):
very nice!

aishwaryamathuria:
@vhargrave @chrischrischris 
Thanks for sharing the review on the PR. 
The codecov check seems to be off by 0.01% due to an unrelated utils file not changed as a part of this PR.
In case there are no more changes required could you please merge the PR?

auniverseaway (approved):

PR #496, opened 2023-02-21
MWPW-126633 Adding check in library blocks to not show duplicate variations

aishwaryamathuria:
* Adding check in library JS file to check if the block variation has already been listed in library.
* In authoring mode on sharepoint, the variations of blocks for author to copy comes from DOM of the documentation and the same variation is duplicated multiple times under some blocks like Media, Card Horizontal, Carousel.
* In some pages like media, cards, carousel one component with same class is authored multiple times hence multiple entries are added as variation is not verified for duplications
* Use the following steps for verification
1. Open the urls in edit mode 
2. Click on Library -> Blocks on franklin sidekick
3. Expand the blocks to check for duplicate variation in list

Resolves: [MWPW-126633](https://jira.corp.adobe.com/browse/MWPW-126633)

**Test URLs:**

- Before: https://main--milo--adobecom.hlx.page/docs/library/blocks/media
- After: https://main--milo--adobecom.hlx.page/docs/library/blocks/media

To verify After scenario use the following github url in for the milo project in franklin sidekick extension https://github.com/aishwaryamathuria/milo/tree/duplicateblocksinlibrary

Adding verification screenshots :

Screenshot 2023-02-21 at 10 56 47 PM       Screenshot 2023-02-21 at 10 57 44 PM


auniverseaway:
@rgclayton @ryanmparrish @elan-tbx @Sartxi 
Who is authoring the library this way? This is an incorrect use of the library.

We shouldn't have variations that are the same in block library.

I'm fine with the fix (minus some syntax nits), but curious how we got to this point.

aishwaryamathuria:
Closing to re-run CLA check

aishwaryamathuria:
> @rgclayton @ryanmparrish @elan-tbx @Sartxi Who is authoring the library this way? This is an incorrect use of the library.
> 
> We shouldn't have variations that are the same in block library.
> 
> I'm fine with the fix (minus some syntax nits), but curious how we got to this point.

@auniverseaway I have updated the syntax of the code. Could you please share your review for the same?

libs/blocks/library-config/lists/blocks.js
```diff
@@ -100,10 +100,16 @@ export default async function loadBlocks(blocks, list, query) {
     const doc = parser.parseFromString(html, 'text/html');
     const pageBlocks = doc.body.querySelectorAll('div[class]');
     let matchingBlockFound = false;
+    const classVariationsFound = [];
     pageBlocks.forEach((pageBlock) => {
       if (pageBlock.className === 'section-metadata') {
         return;
       }
+      const blockVariationName = [...pageBlock.classList].join('.');
```
vhargrave, line 108: I like your approach! 
after the initial classname, the order of the variants doesn't actually matter. 
That means that if there are two same block variants but with a different order of the attributes, they'll both be listed in the library with the current solution. We technically shouldn't be authoring this way, but since we're fixing the duplicate entries, we should probably cover all cases of duplicates. 
You can what I'm talking about with this example. 
https://main--milo--adobecom.hlx.page/drafts/vhargrave/example-variant-ordering

chrischrischris:
```suggestion
      const blockVariationName = [...pageBlock.classList].sort().join('.');
```

aishwaryamathuria: @vhargrave @chrischrischris Thanks for sharing the feedback. I have updated the statement to sort the classnames before joining them so that the order can be same fro the same classname.
Could you please review and confirm if there are any changes required?

auniverseaway:
This PR masks a problem of people (engineers) using the library incorrectly. Do we want to ship this?

The other piece is that with authorable block names, if for whatever reason there _is_ a valid reason to have two blocks of the same name, this doesn't allow an author to do that.

If it was my decision, I would not ship this. I don't like masking a problem we made for ourselves by creating bad content.

vhargrave:
> This PR masks a problem of people (engineers) using the library incorrectly. Do we want to ship this?
> 
> The other piece is that with authorable block names, if for whatever reason there _is_ a valid reason to have two blocks of the same name, this doesn't allow an author to do that.
> 
> If it was my decision, I would not ship this. I don't like masking a problem we made for ourselves by creating bad content.

It's a good point, and as you said on slack templates should actually go into kitchen sink documents. If I'm honest though I didn't know there was a process currently for templates. Where do these kitchen sink documents go, so that authors can find them?
If there's already a process for this, maybe we can document it here or somewhere else on the wiki? - https://github.com/adobecom/milo/wiki/The-Developer-Kitchen-Sink
With that all said my vote would then also go to not ship this.

chrischrischris:
Agree with @auniverseaway.  Let's not hide a poorly authored library doc.  We need to:
1. Make it so that `library-metadata` can control whether a section is displayed or not, and add ability to name it
2. Re-write the library files to not be broken and/or contain duplicates

salonijain3:
Closing this as per this discussion : https://github.com/orgs/adobecom/discussions/504
cc: @aishwaryamathuria @vhargrave

PR #502, opened 2023-02-22
Svg relative

seanchoi-dev:
- Updating all SVG urls as relative.

Resolves: [MWPW-126739](https://jira.corp.adobe.com/browse/MWPW-126739)

**Test URLs:**
- Before: https://main--dc--adobecom.hlx.page/acrobat/online/pdf-to-ppt?martech=off
- After: https://main--dc--adobecom.hlx.page/acrobat/online/pdf-to-ppt?martech=off&milolibs=svg-relative

QA note: Before URL also won't show the issue until you change it as .live or production URL.
Since `milolibs` will not work for .live nor production, this is tricky to test. When I tested locally, it seems solid.


libs/utils/utils.js
```diff
@@ -400,6 +400,10 @@ export function decorateSVG(a) {
   try {
     const textContentUrl = new URL(sanitizedTextContent);
     const hrefUrl = new URL(href);
+    
+    // Making sure all svg URLs are relative.
+    img.src = img.src.replace(textContentUrl?.origin, '');
```
auniverseaway, line 405: Why would we not fix img.src when we actually make the image?

seanchoi-dev: 1. I wanted to reuse `textContentUrl` that is defined at 401 line to make this easy.
2. I wanted to keep localizeLink() code so it is blessed by all the Janaki's logic.
3. There was js error when I was trying it out before `try` statement, I thought it was from `new URL(sanitizedTextContent)` but then I figured it out while I was writing this reply.. so forget about this.

I will update this as.
```
img.src = sanitizedTextContent.replace(new URL(sanitizedTextContent).origin, '');
```

libs/utils/utils.js
```diff
@@ -392,7 +392,7 @@ export function decorateSVG(a) {
     : textContent.substring(textContent.indexOf('|') + 1).trim();
   const img = document.createElement('img');
   img.setAttribute('loading', 'lazy');
-  img.src = localizeLink(sanitizedTextContent);
+  img.src = sanitizedTextContent.replace(new URL(sanitizedTextContent).origin, '');
```
chrischrischris, line 395: This will throw an error if `sanitizedTextContent` is not a valid url

chrischrischris (changes requested):

libs/utils/utils.js
seanchoi-dev: As I test with "wrong url" for the value of sanitizedTextContent, it does throw by the `catch` than just JS error from various places where using decorateSVG(). 
I still added `?` for the case of undefined. If we want to put it in under try/catch I could still move it under the try below : (https://github.com/adobecom/milo/blob/38c1a4828991964e5dba6e301df139bb90206e4b/libs/utils/utils.js#L400-L413)
Which I initially did then changed due to Chris M. comment above.

seanchoi-dev:
Closing this due to https://github.com/adobecom/milo/pull/508 from Chris M.

PR #503, opened 2023-02-22
[MWPW-126441] Unblock style-linting by lowercasing all t-shirt sizes in classnames

elan-tbx:
* Updates all CSS rules to use lowercase t-shirt sizes
* Updates all JS files that set classnames to use lowercase t-shirt sizes
* Should allow us to get a clearer picture of style-linting errors moving forward

Resolves: [MWPW-126441](https://jira.corp.adobe.com/browse/MWPW-126441)

**Test URLs:**
(Note that these should both look the same, but the After view uses lowercase t-shirt size classnames)
- Before: https://main--milo--adobecom.hlx.page/docs/library/blocks/media?martech=off
- After: https://ebartholomew-kebab-case-tshirts--milo--adobecom.hlx.page/docs/library/blocks/media?martech=off


meganthecoder:
Love it! However, we use some of these classes in bacom. What's your plan for updating the consumers without any disruption to the frontend?

chrischrischris (approved):

chrischrischris:
@meganthecoder
For this PR I think we're ok with doing a co-ordinated merge of PRs at the same time.  Looking at bacom only the stats and tree-view blocks are affected.

If it was more complicated, we probably would have to have the PR in Milo support both upper and lower cases temporarily until all consumers had been updated.

@elan-tbx We also need to check the other milo consumers before merging this.

meganthecoder (approved):
Good stuff! I'll leave the "do not merge" label on until the consumer PRs are out. Once those are ready, feel free to remove.

Another note is that there's a PR open for changes to the Aside block #469 by @Sartxi . Depending on who merges first, there'll need to be some rework

elan-tbx:
@meganthecoder Sounds good! Thanks for the heads up

elan-tbx:
@chrischrischris @meganthecoder Hey y'all, I've made a corresponding [PR in bacom](https://github.com/adobecom/bacom/pull/85) that aligns with this one. As far as I can tell, that's the only consumer repo that requires further updates... I've checked bacom, bacom-blog, dc, cc, and fedpub. Is there anything else I should do? Maybe a general heads up to devs in Slack about the changes?

rgclayton:
Bacom and DC/frictionless do have live pages that most likely use some of these blocks.

rgclayton (approved):

PR #507, opened 2023-02-23
MWPW-126724 - Support non-link text in copyright

meganthecoder:
* Allow footer copyright to be authored as unordered list `ul`
* If authored as `ul`, non-link text will remain visible in copyright
* Maintain support for existing single line `p` authoring

Resolves: [MWPW-126724](https://jira.corp.adobe.com/browse/MWPW-126724)

**Test URLs:**

New Authoring `ul`:
- Before: https://main--milo--adobecom.hlx.page/drafts/methomas/test-footer-text?martech=off
- After: https://methomas-footer-text--milo--adobecom.hlx.page/drafts/methomas/test-footer-text?martech=off

Screen Shot 2023-02-23 at 10 52 48 AM

Existing Authoring `p` (for regression):
- Before: https://main--milo--adobecom.hlx.page/drafts/methomas/checkmark?martech=off
- After: https://methomas-footer-text--milo--adobecom.hlx.page/drafts/methomas/checkmark?martech=off

Screen Shot 2023-02-23 at 10 53 22 AM

### DIFF
--- a/libs/blocks/footer/footer.js
+++ b/libs/blocks/footer/footer.js
@@ -197,23 +197,38 @@ class Footer {
 
   decoratePrivacy = () => {
     const copyrightEl = this.body.querySelector('div > p > em');
-    const links = copyrightEl?.parentElement.querySelectorAll('a');
-    if (!copyrightEl || !links) return null;
+    if (!copyrightEl) return null;
+    const container = copyrightEl.closest('div');
     const privacyWrapper = createTag('div', { class: 'footer-privacy' });
     const copyright = createTag('p', { class: 'footer-privacy-copyright' });
     const year = new Date().getFullYear();
     copyright.textContent = `Copyright © ${year} ${copyrightEl.textContent}`;
     privacyWrapper.append(copyright);
-    const infoLinks = createTag('ul', { class: 'footer-privacy-links' });
-    links.forEach((link) => {
-      const li = createTag('li', { class: 'footer-privacy-link' });
-      if (link.hash === '#interest-based-ads') {
-        link.insertAdjacentHTML('afterbegin', ADCHOICE_IMG);
-      }
-      li.append(link);
-      infoLinks.append(li);
-    });
-    privacyWrapper.append(infoLinks);
+
+    const adchoice = container.querySelector('a[href*="#interest-based-ads"]');
+    adchoice?.insertAdjacentHTML('afterbegin', ADCHOICE_IMG);
+
+    const ulClass = 'footer-privacy-links';
+    const liClass = 'footer-privacy-link';
+    let ul = container.querySelector('ul');
+
+    if (ul) {
+      ul.classList.add(ulClass);
+      const listItems = ul.querySelectorAll('li');
+      [...listItems].forEach((item) => {
+        item.classList.add(liClass);
+      });
+    } else {
+      const links = container.querySelectorAll('a');
+      if (!links) return null;
+      ul = createTag('ul', { class: ulClass });
+      links.forEach((link) => {
+        const li = createTag('li', { class: liClass });
+        li.append(link);
+        ul.append(li);
+      });
+    }
+    privacyWrapper.append(ul);
     return privacyWrapper;
   };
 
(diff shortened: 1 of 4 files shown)


JasonHowellSlavin (approved):

JasonHowellSlavin:
@meganthecoder Sorry, missed this in original overview. Is there a test to update?

libs/blocks/footer/footer.js
```diff
@@ -197,23 +197,38 @@ class Footer {
 
   decoratePrivacy = () => {
     const copyrightEl = this.body.querySelector('div > p > em');
-    const links = copyrightEl?.parentElement.querySelectorAll('a');
-    if (!copyrightEl || !links) return null;
+    if (!copyrightEl) return null;
```
Brandon32, line 200: Could get this out of the way so you don't have to check in your conditional
```suggestion
    const container = copyrightEl.closest('div');
    const links = container.querySelectorAll('a');
    let ul = container.querySelector('ul');
    if (!copyrightEl || !(ul || links)) return null;
```

Brandon32 (approved):

libs/blocks/footer/footer.js
meganthecoder: Not sure I see what you're saying. `container` will throw an error is `copyrightEl` is null

meganthecoder:
@JasonHowellSlavin No, I couldn't find any footer tests

libs/blocks/footer/footer.js
Brandon32: You return null in the `ul` conditional, lines 222-223, if there are no `links`, so checking at the top if there are ether `links` or a `ul` and return early if both are missing.
Optional chaining for `copyrightEl` and `container`
```
    const container = copyrightEl?.closest('div');
    const links = container?.querySelectorAll('a');
    let ul = container?.querySelector('ul');
```

meganthecoder: Chris M has pointed out before that he prefers a conditional rather than optional chaining if you need to check the same variable multiple times. I was trying to follow his guidelines

Brandon32: Returning early is also a good practice, and here I don't think we need to sacrifice one for another. It feels like optional chaining would be okay here in order to simplify the return condition, instead of how it's often over used to hunt for nested object properties but maybe that's just preference. 

If you wanted to avoid all nested chaining you could try this:
```suggestion
    if (!copyrightEl) return null;
    const container = copyrightEl.closest('div');
    if (!container) return null;
    const links = container.querySelectorAll('a');
    let ul = container.querySelector('ul');
    if !(ul || links) return null;
```
If you saying that using `container?.` twice is the issues you could try this:
```suggestion
    const container = copyrightEl?.closest('div');
    if (!copyrightEl || !container) return null;
    const links = container.querySelectorAll('a');
    let ul = container.querySelector('ul');
    if !(ul || links) return null;
```

fullcolorcoder (changes requested):
Originally I did this and tests weren't added, I apologize for that. 
However, given the footer is used widely on prod there should be a test when introducing changes. Even when we re-write this soon, the scaffolding of some tests would carry over to a re-write and be helpful.

meganthecoder:
@fullcolorcoder I'm not sure I see the use in writing tests on code that's going to be rewritten soon. Our tests are highly dependent on the dom structure, and anything I wrote would likely be useless after the rewrite

auniverseaway:
From a forward compatibility perspective related to what I'm working on, as long as you don't have an "h2" for copywright you should be good. I'm doing _extremely_ light touches on the outer div for copywright, so this should all work + a bit of css to support inline or list.

I wonder if we say something like, "copywright should always be the last section." from a content model perspective.

fullcolorcoder:
@meganthecoder given the additional feedback, I don't think we should skip tests in the footer because it has wider impact and will remain valid through refactoring.

meganthecoder:
Failing test isn't for footer

fullcolorcoder (approved):

PR #508, opened 2023-02-23
MWPW-126739 - SVG Updates

auniverseaway:
* Fail faster when not an SVG
* Simpler pipe detection for alt text
* Use FQDN if not a .hlx. link
* Better support for relative links

Resolves: [MWPW-126739](https://jira.corp.adobe.com/browse/MWPW-126739)

**Test URLs:**
- After: https://main--milo--adobecom.hlx.live/?martech=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -380,35 +380,34 @@ export async function loadBlock(block) {
 
 export function decorateSVG(a) {
   const { textContent, href } = a;
-  const altTextFlagIndex = textContent.indexOf('|');
-  const sanitizedTextContent = altTextFlagIndex === -1
-    ? textContent
-    : textContent?.slice(0, altTextFlagIndex).trim();
-  const ext = sanitizedTextContent?.substring(sanitizedTextContent.lastIndexOf('.') + 1);
-  if (ext !== 'svg') return;
-
-  const altText = altTextFlagIndex === -1
-    ? ''
-    : textContent.substring(textContent.indexOf('|') + 1).trim();
-  const img = document.createElement('img');
-  img.setAttribute('loading', 'lazy');
-  img.src = localizeLink(sanitizedTextContent);
-  img.alt = altText;
-  const pic = document.createElement('picture');
-  pic.append(img);
-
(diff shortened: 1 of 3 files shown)


libs/utils/utils.js
```diff
@@ -380,35 +380,34 @@ export async function loadBlock(block) {
...
-    ? ''
-    : textContent.substring(textContent.indexOf('|') + 1).trim();
-  const img = document.createElement('img');
-  img.setAttribute('loading', 'lazy');
-  img.src = localizeLink(sanitizedTextContent);
-  img.alt = altText;
-  const pic = document.createElement('picture');
-  pic.append(img);
-
+  if (!(textContent.includes('.svg') || href.includes('.svg'))) return a;
```
seanchoi-dev, line 383: nit: you could use `href.endsWith('.svg')` for href checking.

libs/utils/utils.js
```diff
@@ -380,35 +380,34 @@ export async function loadBlock(block) {
...
   try {
-    const textContentUrl = new URL(sanitizedTextContent);
-    const hrefUrl = new URL(href);
-    if (textContentUrl?.pathname === hrefUrl?.pathname) {
+    // Mine for URL and alt text
+    const splitText = textContent.split('|');
+    const textUrl = new URL(splitText.shift().trim());
```
seanchoi-dev, line 387: This will fail if `textContent` doesn't have proper URL text, therefore there will be no `pic` for that case.
(I guess we want that failure..?)

seanchoi-dev (approved):
Looks better and easier to read than before. I like it.
I left few comments to consider.

libs/utils/utils.js
auniverseaway: ...unless someone does: 
`https://.../my-icon.svg#_blank`
or
`https://.../my-icon.svg#_dnt`

libs/utils/utils.js
auniverseaway: Yup! That's the goal!

rgclayton (approved):
Looks good. 

[Sample page](https://svg-fixes--milo--adobecom.hlx.page/drafts/rclayton/svg-test) with options for reference

PR #509, opened 2023-02-23
MWPW-126238 Add Footer social links for CN

Brandon32:
* Add Footer social links for CN, Weibo and WeChat.
* WeChat icon links to a social media page with QR codes. 
  * https://www.adobe.com/cn/about-adobe/newsroom/social-media.html

Resolves: [MWPW-126238](https://jira.corp.adobe.com/browse/MWPW-126238)

**Test URLs:**
- Before: https://main--bacom--adobecom.hlx.page/cn/footer?martech=off
- After: https://main--bacom--adobecom.hlx.page/cn/footer?martech=off&milolibs=bmarshal-footer-social-links


JasonHowellSlavin (approved):

libs/blocks/footer/footer.js
```diff
@@ -17,7 +17,7 @@ const base = miloLibs || codeRoot;
...
 const SPECTRUM_CHEVRON = '<svg class="icon-chevron-down" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16"><defs><style>.spectrum-chevron-down{fill:none;stroke-linecap:round;stroke-linejoin:round;}</style></defs><path stroke="currentColor" class="spectrum-chevron-down" d="M3.47,5.74l4.53,4.53,4.54-4.53"/></svg>';
 const ADCHOICE_IMG = `<img class="footer-link-img" loading="lazy" alt="AdChoices icon" src="${base}/blocks/footer/adchoices-small.svg" height="9" width="9">`;
-const SUPPORTED_SOCIAL = ['facebook', 'instagram', 'twitter', 'linkedin', 'pinterest', 'discord', 'behance', 'youtube'];
+const SUPPORTED_SOCIAL = ['facebook', 'instagram', 'twitter', 'linkedin', 'pinterest', 'discord', 'behance', 'youtube', 'weibo', 'social-media'];
```
auniverseaway, line 20: What is "social-media" ? Seems generic.

Brandon32: It is, but I need an icon for https://www.adobe.com/cn/about-adobe/newsroom/social-media.html

meganthecoder (approved):

libs/blocks/footer/footer.js
auniverseaway: Wouldn't `adobe-newsroom` be better?

auniverseaway: @Brandon32 FWIW: "should be good enough" is almost never an acceptable answer.

If there's some weird URL check thing going on that forces the lookup, we fix the weird thing causing the poor choice. We don't just say, "good enough."

Names matter. Names should make sense. Now when someone goes to look at this, they're going to see `social-media` and wonder what the heck is going on.

Brandon32: @auniverseaway Should be enough to identify what icon we are talking about not the code quality, also are you saying there are no tradeoffs in programming? We are adding these icons for the CN go live today so a refactor can come later, and I understand is already underway. 
I talk about scope because there are a total of 10 icons and 1 has a generic name because we are identifying icons by URL and the URL is generic. The `adobe-newsroom` is also a bad name considering the icon is actually WeChat and so as-is this is an edge case from an outside decision to account for. There is not always a great relationship for the icon based on the URL provided but that is the initial assumption for the authoring experience.

Without re-working the footer AX I don't see a good fix for this without mapping the complete URL to identify the  icon needed. Changing `SUPPORTED_SOCIAL` to a key value object to map URLs so that 1 icon name can be more specific seems like it's not worth the tradeoff of added complexity and bytes but I would like to know your thoughts.

PR #510, opened 2023-02-23
MWPW-123646 & MWPW-123874 [Blog2Milo] - Article Header and Tags 

saugatmalla:
* Template to auto-generate article headers and article tags

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/saugat/articles/adobe-again-named-one-of-worlds-most-admired-companies-1
- After: https://smalla-article-header--milo--adobecom.hlx.page/drafts/saugat/articles/adobe-again-named-one-of-worlds-most-admired-companies-1?martech=off


auniverseaway (changes requested):
I think we need to change where loadTemplate is. We probably need to do some testing to see how bad the performance is by moving it in front of LCP.

rgclayton:
@saugatmalla Let me know if we need to sync up again after Chris' comments.

PR #377, opened 2023-01-09
MWPW-120935 - [SEO Structured Data] Event

jckautzmann:
- add SEO data when the page is set as Event in the metadata

Resolves: [MWPW-120935](https://jira.corp.adobe.com/browse/MWPW-120935)

Notes:
- mixed event is not supported (`"eventAttendanceMode": "https://schema.org/MixedEventAttendanceMode”,`)
- authoring (Word/gdoc): meta data key is written with ‘-‘  so that it’s easier to read in HTML:
    - Word/gdoc: `Offers-Price-Currency | USD`
    - HTML: `<meta name="offers-price-currency" content="USD”>`

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/jck/test-seo-event?martech=off
- After: https://seo-event--milo--adobecom.hlx.page/drafts/jck/test-seo-event?martech=off

Content model:

Screenshot 2023-01-09 at 17 59 23

JSON-LD in the rendered HTML:

Screenshot 2023-01-09 at 18 00 48

PR #511, opened 2023-02-24
MWPW-120935 - [SEO Structured Data] Event

jckautzmann:
- new block to add rich results for Event to a page (SEO)

Resolves: [MWPW-120935](https://jira.corp.adobe.com/browse/MWPW-120935)

Notes:
- see former implementation and discussions in https://github.com/adobecom/milo/pull/377 
- mixed event is not supported (`"eventAttendanceMode": "https://schema.org/MixedEventAttendanceMode”,`)
- authoring (Word/gdoc): meta data key is written with ‘-‘  so that it’s easier to read in HTML:
    - Word/gdoc: `Offers-Price-Currency | USD`
    - HTML: `<meta name="offers-price-currency" content="USD”>`

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/jck/seo-event/test-seo-event?martech=off
- After: https://seo-event-block--milo--adobecom.hlx.page/drafts/jck/seo-event/test-seo-event?martech=off

Content model:

Screenshot 2023-02-24 at 13 04 03

JSON-LD in the rendered HTML:

Screenshot 2023-01-09 at 18 00 48


jckautzmann:
The failing test seems unrelated to this PR:
```
test/blocks/fallback/fallback.test.js:

 ❌ Fallback > does not set error if block is a sub-block
      AssertionError: expected 'true' to be undefined
        at o.<anonymous> (test/blocks/fallback/fallback.test.js:17:39)
```

jckautzmann:
The `Check code compatibility` fails also on the `main` branch:
```
/home/runner/work/milo/milo/libs/blocks/fallback/fallback.js
  20:1  error  Definition for rule 'import/prefer-default-export' was not found  import/prefer-default-export
```

jckautzmann:
The LH score does not seem related to this PR: I tested with an empty page with code from `main` and the score was 68:
https://main--milo--adobecom.hlx.page/drafts/jck/seo-event/test-seo-event2?martech=off

libs/blocks/event-rich-results/event-rich-results.js
```diff
@@ -0,0 +1,78 @@
...
+    return {
+      '@type': 'Place',
+      name: data.get('location-name'),
+      address: {
+        '@type': 'PostalAddress',
+        streetAddress: data.get('location-address-street'),
+        addressLocality: data.get('location-address-locality'),
+        postalCode: data.get('location-address-postal-code'),
+        addressRegion: data.get('location-address-region'),
+        addressCountry: data.get('location-address-country'),
```
rgclayton, line 25: I know for metadata we only accept the `-` between keys like `gnav-source` and not a space. This being a block is it a concern that authors have to add `-` in place of spaces in the keys or the schema fails? Anyone have thoughts?

Otherwise looks good. 
Same [result URL](https://search.google.com/test/rich-results/result/r%2Fevents?id=IDy7h4Oh3HCEucXZ3JoOHw)

jckautzmann: @rgclayton I adapted the code at https://github.com/adobecom/milo/pull/511/commits/5d5cda6cb3209c6ec35a78006c27c5c4a534b7cf to allow authors to enter either whitespace or '-' in the property name.

rgclayton (approved):
Thanks for the update JC!

echampio-at-adobe (changes requested):
The code seems quite bulletproof with various sets of bogus data. The primary values that drive the code are Place and VirtualLocation. If they're missing or capitalized incorrectly, the data is just skipped. If other values are missing, the resulting null values are happily written into the scheme without console errors. In your buildData method, before you return the sanitized data, would it be possible to add some validation if some set of key values like Place or VirtualLocation are really required? If those values are missing, log an error with Lana before continuing as before and returning the data? That way, if an author enters "place" instead of "Place," or omits something entirely, there will be something to flag it if they never check to see how things are rendered.

libs/blocks/event-rich-results/event-rich-results.js
```diff
@@ -0,0 +1,78 @@
...
+      },
+    };
+  }
+  if (data.get('location-type') === 'VirtualLocation') {
+    return {
+      '@type': 'VirtualLocation',
+      url: data.get('location-url'),
+    };
+  }
+  return null;
```
JasonHowellSlavin, line 35: nit: could write a test for the null case from this function.

JasonHowellSlavin (approved):

libs/blocks/event-rich-results/event-rich-results.js
jckautzmann: @JasonHowellSlavin I updated the PR with a new test covering null/undefined properties at https://github.com/adobecom/milo/pull/511/commits/4136d46e818045fdc7fb50e45a57c1a9480f0b09

jckautzmann:
@echampio-at-adobe I updated the PR with some logic that logs in the console (based on Lana) null/undefined properties at https://github.com/adobecom/milo/pull/511/commits/4136d46e818045fdc7fb50e45a57c1a9480f0b09

JasonHowellSlavin (approved):

echampio-at-adobe (approved):
Thanks for adding the lana message.

honstar (approved):

PR #519, opened 2023-03-01
Adding event driven modals to decorateMeta in utils.js

JasonHowellSlavin:
* Moves the event driven modals code to the `decorateMeta` function in utils. 
* Fixes a race condition for modal loading
* Adding test for new code placement

Helps to resolves: [MWPW-125597](https://jira.corp.adobe.com/browse/MWPW-125597

**Test URLs:**
- Before: https://chat-modal--bacom--adobecom.hlx.page/drafts/slavin/chat-cta-examples/page-one?martech=off
- After: https://chat-modal--bacom--adobecom.hlx.page/drafts/slavin/chat-cta-examples/page-one?milolibs=event-modal-rc


auniverseaway (approved):

honstar (approved):

libs/blocks/modal/modal.js
```diff
@@ -146,12 +146,6 @@ export default function init(el) {
   return null;
 }
 
-// Event-based modal
-window.addEventListener('modal:open', (e) => {
```
honstar, line 150: Moving the event listener for the modal causes unit-test failures in `test/blocks/modals/modals.test.js`. Looks like the event listener is not registered when running the modal tests since it now is registered as part of `decorateMeta` in `libs/utils/utils.js`.

honstar (changes requested):

libs/blocks/modal/modal.js
JasonHowellSlavin: Thank you for catching. Removed the failing test as it's functionality is accounted for in utils.test.js

honstar (approved):

PR #522, opened 2023-03-02
MWPW-119230: Feds Utilities

mokimo:
### Relations
Resolves: [MWPW-119230](https://jira.corp.adobe.com/browse/MWPW-119230)

### Description
I've done an exhaustive review of all the FEDS related utilities that we currently offer in the aem based global navigation.
As a lot of the backwards compatibility isn't needed anymore (example: IMS does not need a wrapper anymore to support V1 and V2 with the same API, but we can just rely on the fact IMS v2 is used) only two crucial utilities will be exported, and users are free to consume them.

They do not occupy any space in the global navigation/FEDS anymore but serve as convenient wrapping functions anyone is free to consume.

### Usage
The scripts can either be consumed from a milo context or standalone when you import the absolute path equal to lana.
```js
import getUserEntitlements from './utilities/getUserEntitlements.js';
// Entitlements will be the unformatted response from the API
// https://www.adobe.com/aos-api/users/${profile.userId}/subscriptions
const entitlements = await getUserEntitlements().catch(err => console.log(err))
```

```js
import getUserEventHistory from './utilities/getUserEventHistory';
const historicallyRegistered = await getUserEventHistory().catch(e => console.log("handle error"))
```

**Test URLs:**
- Before: https://feds-gnav--milo--adobecom.hlx.page/?martech=off
- After: https://feds-gnav--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/libs/blocks/global-navigation/utilities/getUserEntitlements.js
+++ b/libs/blocks/global-navigation/utilities/getUserEntitlements.js
@@ -66,11 +66,8 @@ const getSubscriptions = async ({ queryParams, locale }) => {
  * @returns {object} JIL Entitlements
  */
 const getUserEntitlements = async ({ params, locale } = {}) => {
-  if (!window.adobeIMS?.isSignedInUser()) {
-    return new Promise((resolve, reject) => {
-      reject(new Error('User not signed in'));
-    });
-  }
+  if (!window.adobeIMS?.isSignedInUser()) return Promise.reject(new Error('User not signed in'));
+
   const queryParams = getQueryParameters(params);
   if (entitlements[queryParams]) return entitlements[queryParams];
 
--- a/libs/blocks/global-navigation/utilities/getUserEntitlements.js
+++ b/libs/blocks/global-navigation/utilities/getUserEntitlements.js
@@ -1,0 +1,92 @@
+import { getConfig } from '../../../utils/utils.js';
+
+const API_WAIT_TIMEOUT = '10000';
+const entitlements = {};
+
+const getAcceptLanguage = (locale) => {
+  let languages = [];
+  if (!locale || Object.keys(locale).length === 0) {
+    return languages;
+  }
+
+  languages = [
+    `${locale.language}-${locale.country.toUpperCase()}`,
+    `${locale.language};q=0.9`,
+    'en;q=0.8',
+  ];
+  return languages;
+};
+
+const getQueryParameters = (params) => {
+  let result = '';
+  const query = [];
+  if (Array.isArray(params) && params.length) {
+    params.forEach((parameter = {}) => {
+      const { name, value } = parameter;
+      if (typeof name === 'string' && name.length) {
+        query.push(`${encodeURIComponent(name)}=${encodeURIComponent(value)}`);
+      }
+    });
+
+    if (query.length) {
+      result = `?${query.join('&')}`;
+    }
+  }
+
+  return result;
+};
+
+const getSubscriptions = async ({ queryParams, locale }) => {
+  const profile = await window.adobeIMS.getProfile();
+  const apiUrl = getConfig().env.name === 'prod'
+    ? `https://www.adobe.com/aos-api/users/${profile.userId}/subscriptions`
+    : `https://www.stage.adobe.com/aos-api/users/${profile.userId}/subscriptions`;
+  const res = await window
+    .fetch(`${apiUrl}${queryParams}`, {
+      method: 'GET',
+      cache: 'no-cache',
+      credentials: 'same-origin',
+      headers: {
+        'Content-Type': 'application/json',
+        Authorization: `Bearer ${window.adobeIMS.getAccessToken().token}`,
+        'X-Api-Key': window.adobeIMS.adobeIdData.client_id,
+        Accept: 'application/json',
+        'Accept-Language': getAcceptLanguage(locale).join(','),
+      },
+    })
+    .then((response) => (response.status === 200 ? response.json() : null));
+  return res;
+};
+
+/**
+ * @description Return the JIL User Entitlements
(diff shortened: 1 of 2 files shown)


libs/blocks/global-navigation/utilities/getUserEntitlements.js
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * @description Return the JIL User Entitlements
+ * @param {object} object required params
+ * @param {array} object.params array of name value query parameters [{name: 'Q', value: 'PARAM'}]
+ * @param {object} object.locale {country: 'CH', language: 'de'}
+ * @returns {object} JIL Entitlements
+ */
+const getUserEntitlements = async ({ params, locale } = {}) => {
+  if (!window.adobeIMS?.isSignedInUser()) {
+    return new Promise((resolve, reject) => {
```
overmyheadandbody, line 70: `resolve` is not used, could you not return just a `Promise.reject(new Error('User not signed in'));`?

libs/blocks/global-navigation/utilities/getUserEventHistory.js
```diff
@@ -0,0 +1,31 @@
+/* eslint-disable no-promise-executor-return, no-async-promise-executor */
+import { getConfig } from '../../../utils/utils.js';
+
+let userEventHistory;
+const getUserEventHistory = async (id) => {
+  if (userEventHistory) return userEventHistory;
+  if (!window.adobeIMS?.isSignedInUser()) return new Promise((resolve, reject) => { reject(new Error('User not signed in')); });
```
overmyheadandbody, line 7: Same thing, maybe a `Promise.reject` would suffice.

libs/blocks/global-navigation/utilities/getUserEventHistory.js
```diff
@@ -0,0 +1,31 @@
...
+    resolve = _resolve;
+    reject = _reject;
+  });
+  const profile = await window.adobeIMS.getProfile();
+  const guid = profile.userId;
+  const apiUrl = getConfig().env.name === 'prod'
+    ? 'https://www.adobe.com/gating/api/realtime-profile'
+    : 'https://www.stage.adobe.com/gating/api/realtime-profile';
+  const res = await window
+    .fetch(`${apiUrl}?guid=${id || guid}`, {
```
overmyheadandbody, line 20: Completely trivial nit: you're using `guid` just once, only if an `id` hasn't been passed to the function. If an `id` is available, you're saving the `guid` without it ever being used, so I'd skip the variable assignment and just do `id || profile.userId` here.

overmyheadandbody (approved):

libs/blocks/global-navigation/utilities/getUserEntitlements.js
```diff
@@ -0,0 +1,92 @@
...
+    || new Promise((_resolve, _reject) => {
+      resolve = _resolve;
+      reject = _reject;
+    });
+  setTimeout(() => reject('Subscriptions API Call timeout'), API_WAIT_TIMEOUT);
+  // TODO we might need to format data for analytics
+  const data = await getSubscriptions({ queryParams, locale });
+  if (data) resolve(data);
+  reject('No response from the aos-api');
+  return entitlements[queryParams];
```
narcis-radu, line 89: how is this used?

mokimo: What exactly do you mean? 
We set it on [L79 and L87](https://github.com/adobecom/milo/pull/522/files/7dc9d4e0e32ca5d92ba6c25c43de07e9d41b0c62#diff-21d166afb348116f92a7a5d80b34913d94a5552b736df9320c634fbceb10fabbR79-R87)

And cache all the entitlement responses by queryParams

honstar:
@mokimo: just for my understanding: the reason why you're introducing a new `libs/blocks/global-navigation` folder is to make it clear that this is about the new gnav that will replace the existing milo one at `libs/blocks/gnav`, right?

mokimo:
@honstar yes that's one reason, we have a few more reasons for it:
1. it's a rewrite of the current milo `libs/blocks/gnav` that follows the same authoring patterns to be compatible
2. so that we don't have any merge conflicts with the current `libs/blocks/gnav` implementation
3. it's provides a convenient migration path as authors just need to change the document metadata from `header: gnav` to `header: global-navigation` and we can be sure to not break anything in production or be able to revert quickly

narcis-radu (approved):

libs/blocks/global-navigation/utilities/getUserEntitlements.js
```diff
@@ -0,0 +1,89 @@
...
+
+  let resolve;
+  let reject;
+  entitlements[queryParams] = entitlements[queryParams]
+    || new Promise((_resolve, _reject) => {
+      resolve = _resolve;
+      reject = _reject;
+    });
+  setTimeout(() => reject('Subscriptions API Call timeout'), API_WAIT_TIMEOUT);
+  // TODO we might need to format data for analytics
```
honstar, line 82: Is this comment still needed for future work?

mokimo: Yes it is!

mokimo: A bit more context:
We're currently adding a lot of todo's for anything that goes into our feds-gnav branch, we try to challenge everything that exists in the AEM gnav and how things have been done. Some things aren't 100% clear as it's a lot of different consumers using that code.

We try to omit any code that just exists for historic/compatibility reasons and fully embrace the milo landscape and the embrace the options we have from starting on a green field. (While still trying to maintain feature parity)

honstar (approved):

PR #524, opened 2023-03-02
Use optional chaining for lana calls

chrischrischris:
If a Milo consumer does not have lana setup (or setup wrong) then the `window.lana.log` call can fail and stop js execution.  Using optional chaining means it will fail gracefully.

Note that ALL `lana.log` calls in Milo need to use this.

**Test URLs:**
- After: https://lanaupdate--milo--adobecom.hlx.page/?martech=off


auniverseaway (approved):

libs/utils/lana.js
```diff
@@ -83,7 +85,7 @@
       queryParams.push('tags=' + encodeURI(o.tags));
     }
 
-    if (w.lana.debug || w.lana.localhost) console.log('LANA Msg: ', msg, '\nOpts:', o);
+    if (!isProdDomain || w.lana.debug || w.lana.localhost) console.log('LANA Msg: ', msg, '\nOpts:', o);
```
jckautzmann, line 88: shouldn't it be `window.lana?.log` instead of `console.log`?

test/blocks/fragment/fragment.test.js
```diff
@@ -18,13 +18,13 @@ document.body.innerHTML = await readFile({ path: './mocks/body.html' });
 const { default: getFragment } = await import('../../../libs/blocks/fragment/fragment.js');
 
 describe('Fragments', () => {
-  beforeEach(() => {
-    sinon.spy(console, 'log');
-  });
+  // beforeEach(() => {
```
jckautzmann, line 21: should the commented lines be removed?

libs/utils/lana.js
chrischrischris: This is inside the `lana.log` function.  When lana log is called on a non-prod domain, the message is output to the console

honstar (approved):

PR #526, opened 2023-03-03
MWPW-126169: authoring changes for hiding individual CTAs

sheridansunier:
Authoring would work by having a separate panel where they can list ids and for those IDs the primary CTA will be removed. (either css display:none or just never render html)

Here's a detail of UX requirements:

We want a new tab similar to the "Feature Cards" and "Exclude Cards".
The new tab will be called "Hide CTAs" and the user will be the prompted to add content ID or card path, with "Add" option for more cards (similar to the Exclude Cards UX).

We also have a "Hide all cards in collection" option. While this duplicates the current functionality in UI->"Collection Button Style", this will be good in case users go to the new "Hide CTAs" tab and don't know why they can't hide all CTAs for the collection (even though they can if they used the current method)

Requires: https://github.com/adobecom/caas/pull/17

Resolves: [[MWPW-126169](https://jira.corp.adobe.com/browse/MWPW-126169)]

Test URLs:

Before: [https://main--milo--adobecom.hlx.page?martech=off](https://main--milo--adobecom.hlx.page/?martech=off)
After: [https://hide-caas-ctas--milo--sheridansunier.hlx.page?martech=off](https://mwpw-126169--milo--sheridansunier.hlx.page/?martech=off)


auniverseaway (approved):

honstar:
@sheridansunier, I'm testing your change on the caas tool page with your branch: https://hide-caas-ctas--milo--sheridansunier.hlx.page/tools/caas

If you look at the following screenshot, I added the card id of the first 3 cards to the new _Hidden CTAs_ section. However, those 3 cards in question still show a CTA button.

Screenshot 2023-03-07 at 15 10 20

I do see the configured card ids getting sent in query parameter `hideCtaIds` though.

sheridansunier:
@honstar - it requires a PR to chimera-io that is on dev.  So you will need to switch to that endpoint on the advanced tab.

sheridansunier:
It also requires https://github.com/adobecom/caas/pull/17 (so making sure this is consuming the latest or at least that version of caas)

sanrai:
@sunier, leaving notes of what we discussed in slack in the PR for historical reference:

We discussed having the hide cta functionality for individual cards be 100% clientside versus sending the params to chimera io approach that is in the PR. 

The reason being is will easily exceed the URL request limit of 2048 chars using the original approach in this PR.

We already have cases where we are using 1600-1800 chars for URLs on prod and if they wanted to hide some CTAs it may become troublesome for us.

The other benefit is, by changing the approach, we don’t have to worry about having people sync chimera dev versions and dexter release versions when reviewing this PR.

honstar:
I agree with @sanrai's assessment, hiding CTAs should be done entirely on the clientside. The chimera-io backend does not need to be aware of it nor process this setting in any way.

mokimo (approved):
As suggestion, the tests all seem to have `hideCtaIds=` IMO it would make sense to have a test that actually sets it.

chrischrischris:
Is `Hide all cards in collection` supposed to be in this PR?

I understand giving the ability for authors to hide ctas this way in a collection, but shouldn't that also be doable when creating a card so the card always has the cta hidden?  With it setup this way authors have to remember all of the cards that shouldn't have CTAs and then add those ids in.

sheridansunier:
@chrischrischris - that's already enabled, you can hide all of the ctas by going to UI > Collection Button Style > Hide CTAs

honstar:
@sheridansunier : what's the status on this PR? It has been last updated 2 weeks ago and it has the `do not merge` label.

PR #532, opened 2023-03-06
Fixing bug when there is no firstelementchild on footer div.

seanchoi-dev:
This is for fixing small bug in case footer is empty.

image


mokimo (approved):
It would be nice if you include a link where the bug is being reproduced before/after in the description, even on trivial fixes so the reviewers can give it a quick glance!

seanchoi-dev:
> It would be nice if you include a link where the bug is being reproduced before/after in the description, even on trivial fixes so the reviewers can give it a quick glance!

I found this bug while I was testing personal project during the weekend. I verified this would fix the bug but didn't want to include personal project link here.

honstar (approved):

PR #533, opened 2023-03-06
[MWPW-125175] Nav Links refactor

overmyheadandbody:
This is a refactor of the Gnav links decoration, popups/dropdowns (sync or async, normal or mega-menu style), headliness and CTAs. It also scopes the CSS custom properties used by FEDS to avoid collision with other parts of the application.

A few additional improvements worth mentioning:
* the logic will now follow the structure as authored in Word, it is no longer opinionated on item order;
* the headlines now act as accordion headers for subsections on mobile.

This PR also fixes two bugs related to Search:
* https://jira.corp.adobe.com/browse/MWPW-127264 (ensure correct curtain styles when Search is open)
* https://jira.corp.adobe.com/browse/MWPW-127263 (clear Search input and results when trigger is clicked)

Resolves: [MWPW-125175](https://jira.corp.adobe.com/browse/MWPW-125175)

**Test URLs:**
- Before: https://feds-gnav--milo--overmyheadandbody.hlx.page/drafts/ramuntea/gnav-refactor?martech=off
- After: https://mwpw-125175-nav-links--milo--overmyheadandbody.hlx.page/drafts/ramuntea/gnav-refactor?martech=off

### DIFF
--- a/libs/blocks/global-navigation/blocks/navDropdown/dropdown.css
+++ b/libs/blocks/global-navigation/blocks/navDropdown/dropdown.css
@@ -13,6 +13,10 @@
   background-color: var(--feds-background-popup--light);
 }
 
+.feds-popup p {
+  margin: 0;
+}
+
 .feds-popup,
 .feds-popup-items {
   display: none;
--- a/libs/blocks/global-navigation/blocks/navDropdown/dropdown.css
+++ b/libs/blocks/global-navigation/blocks/navDropdown/dropdown.css
@@ -1,0 +1,169 @@
+:root {
+  --feds-background-popup--light: #fafafa;
+  --feds-color-navLink-description--light: #656565;
+  --feds-borderColor-popup--light: #e1e1e1;
+  --feds-color-headline--light: #505050;
+}
+
+.feds-navItem {
+  flex-direction: column;
+}
+
+.feds-popup {
+  background-color: var(--feds-background-popup--light);
+}
+
+.feds-popup,
+.feds-popup-items {
+  display: none;
+  padding: 12px 0;
+}
+
+.feds-popup-headline {
+  position: relative;
+  padding: 12px 44px 12px 32px;
+  border-bottom: solid 1px var(--feds-borderColor-popup--light);
+  color: var(--feds-color-headline--light);
+  font-weight: 600;
+  white-space: nowrap;
+  cursor: pointer;
+}
+
+[dir = "rtl"] .feds-popup-headline {
+  padding-right: 32px;
+  padding-left: 44px;
+}
+
+.feds-popup-headline:after {
+  position: absolute;
+  right: 30px;
+  top: 50%;
+  display: flex;
+  width: 6px;
+  height: 6px;
+  margin-top: -3px;
+  border-width: 0 1px 1px 0;
+  border-style: solid;
+  border-color: var(--feds-color-link--light);
+  transform-origin: 75% 75%;
+  transform: rotateZ(45deg);
+  transition: transform 0.1s ease;
+  content: "";
+}
+
+.feds-navLink--hoverCaret[aria-expanded = "true"] + .feds-popup,
+.feds-popup-headline[aria-expanded = "true"] + .feds-popup-items {
+  display: flex;
+  flex-direction: column;
+}
+
+.feds-navLink--hoverCaret[aria-expanded = "true"]:after,
+.feds-popup-headline[aria-expanded = "true"]:after {
+  transform: rotateZ(-135deg);
+}
+
+.feds-popup-items {
+  border-bottom: solid 1px var(--feds-borderColor-popup--light);
+}
+
+.feds-popup .feds-navLink,
+.feds-popup-items .feds-cta-wrapper {
+  padding: 6px 32px;
+}
+
+.feds-popup .feds-navLink:hover,
+.feds-popup .feds-navLink:focus {
+  background-color: var(--feds-background-link--hover--light);
+}
+
+.feds-navLink-image,
+.feds-navLink-description {
+  display: none;
+}
+
+.feds-popup ul {
+  margin: 0;
+  padding: 0;
+}
+
+@media (min-width: 900px) {
+  .feds-navItem {
+    flex-direction: initial;
+  }
+
+  .feds-navItem--megaMenu {
+    position: static;
+  }
+
+  .feds-popup {
+    position: absolute;
+    top: 100%;
+    left: 0;
+  }
+
+  [dir = "rtl"] .feds-popup {
+    left: auto;
+    right: 0;
+  }
+
+  .feds-navLink--hoverCaret[aria-expanded = "true"] + .feds-popup {
+    flex-direction: row;
+  }
+
+  .feds-navItem--megaMenu .feds-popup {
+    right: 0;
+    justify-content: space-around;
+    padding: 32px 0;
+  }
+
+  .feds-popup-headline {
+    position: static;
+    margin: 0 32px;
+    cursor: default;
+  }
+
+  .feds-popup-headline,
+  [dir = "rtl"] .feds-popup-headline {
+    padding: 12px 0;
+  }
+
+  .feds-popup-headline:after {
+    content: none;
+  }
+
+  .feds-popup-headline + .feds-popup-items {
+    display: flex;
+    flex-direction: column;
+    padding: 0;
+    border-bottom: none;
+  }
+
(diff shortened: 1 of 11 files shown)


libs/blocks/global-navigation/blocks/navDropdown/dropdown.css
```diff
@@ -0,0 +1,169 @@
...
+    flex-direction: column;
+    padding: 0;
+    border-bottom: none;
+  }
+
+  .feds-navLink-image {
+    display: flex;
+  }
+
+  .feds-navLink-image:nth-last-child(2):after {
```
mokimo, line 145: Do we have a use-case where the feds-navLink-image is not the second child? 
If yes, do we want to give it a different class name?

libs/blocks/global-navigation/blocks/navDropdown/dropdown.js
```diff
@@ -0,0 +1,179 @@
...
+  const headline = toFragment`<div
+    class="feds-popup-headline"
+    role="heading"
+    aria-level="2"
+    aria-haspopup="true"
+    aria-expanded="false">
+    ${elem.textContent.trim()}
+  </div>`;
+
+  // TODO: move proper logic to accessibility,
```
mokimo, line 17: As discussed this should rather live in the keyboard navigation 👍

libs/blocks/global-navigation/blocks/navDropdown/dropdown.js
```diff
@@ -0,0 +1,179 @@
...
+    } else {
+      headline.setAttribute('aria-expanded', 'false');
+    }
+  });
+
+  // Since heading is turned into a div, it can be safely removed
+  elem.remove();
+
+  return headline;
+};
```
mokimo, line 41: As discussed this should rather live in the keyboard navigation 👍

mokimo, line 41: As discussed this should rather live in the keyboard navigation 👍

libs/blocks/global-navigation/blocks/navDropdown/dropdown.js
```diff
@@ -0,0 +1,179 @@
...
+
+const decorateLinkGroup = (elem, index) => {
+  if (!(elem instanceof HTMLElement)) return null;
```
mokimo, line 44: Let's add a safeguard for `const link = elem.querySelector('a');` here and discard the link group if it doesn't have a link. Wonder if we also have any debugging logs we could add?
```suggestion
  if (!(elem instanceof HTMLElement) || !elem.querySelector('a')) return null;
```

libs/blocks/global-navigation/blocks/navDropdown/dropdown.js
```diff
@@ -0,0 +1,179 @@
...
+
+          // Decorate CTA
+          if (columnElem.querySelector('strong > a')) {
+            decoratedElem = decorateCta({ elem: columnElem, index: currIndex });
+          }
+
+          // If element has been decorated, remove it
+          // from the initial list of column elements
+          if (decoratedElem) {
+            columnElem.remove();
```
mokimo, line 148: What happens with the `while` loop if the columnElement not a link-group, a CTA or UL?

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -346,23 +478,48 @@ header .gnav-navitem.has-menu > a:after {
...
+  position: relative;
+  justify-content: end;
 }
 
-.gnav-cta .con-button {
-  line-height: 36px;
-  border-radius: 20px;
-  padding: 0 14px;
+/* TODO temporarily more specificity until we improve the CSS */
+header .gnav-navitem .feds-signin {
```
mokimo, line 498: Does this CSS need to be in the `global-navigations.css` if we have `signIn.css` ?

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -346,23 +478,48 @@ header .gnav-navitem.has-menu > a:after {
...
+  max-height: 30px;
+  max-width: 30px;
+  display: block;
+}
+
+.feds-profile-button {
+  padding: 5px var(--feds-spacing-utilityNav);
+  border-radius: var(--feds-radius-utilityIcon);
+  border: none;
+}
```
mokimo, line 520: Does this CSS need to be in the `global-navigations.css` if we have `milo-wrapper.css` ?

mokimo (approved):
There's still some unused styles like `class="gnav-promo image-only"` what's the plan with them?

libs/blocks/global-navigation/blocks/navDropdown/dropdown.js
mokimo: Btw on this note, I'll create a ticket that we have to add some lana logs - I want to observe any errors we see after launching.

libs/blocks/global-navigation/blocks/navDropdown/dropdown.js
overmyheadandbody: Just the function from the event listener should belong to the keyboard navigation, so lines 19-35. The rest should still remain here.

libs/blocks/global-navigation/blocks/navDropdown/dropdown.js
overmyheadandbody: Yup, good point with checking that the group has a link, will accept your change 👍

libs/blocks/global-navigation/blocks/navDropdown/dropdown.js
overmyheadandbody: At the moment - nothing. But thinking of it, we should definitely allow simple links (not defined in a link group or a list). Otherwise, maybe we should just add the element unprocessed?

libs/blocks/global-navigation/global-navigation.css
overmyheadandbody: Nope, this is likely a rebase conflict omission. I'll remove it from here.

libs/blocks/global-navigation/global-navigation.css
overmyheadandbody: Most likely a rebase conflict omission, I'll look into it, good catch!

libs/blocks/global-navigation/blocks/navDropdown/dropdown.js
mokimo: Let's keep it as is for now. It's easier to add functionalityy in the future rather than taking it away and potentially breaking stuff.

overmyheadandbody:
Regarding the promo, it will be implemented in another PR and any styled deemed unnecessary will be removed then.

libs/blocks/global-navigation/blocks/navDropdown/dropdown.js
overmyheadandbody: I retract my previous comment 😅 If the `columnElem` doesn't fit into a category where we know we need to add extra decoration, then we simply add it to the column/section as-is. However, it is very possible that a simple link is encountered, outside of a list, I will add the appropriate class and analytics attributes to it.

mokimo (approved):

honstar:
While testing the gnav menus, I noticed some layout issues with the gnav promo card that is part of the _Medium Dropdown_ menu on this test page: https://mwpw-125175-nav-links--milo--overmyheadandbody.hlx.page/drafts/ramuntea/gnav-refactor?martech=off

Screenshot 2023-03-07 at 12 17 06

Shouldn't this have a clear card-like appearance and constraints? On adobe.com in production, this looks much better:
Screenshot 2023-03-07 at 12 27 13

narcis-radu:
> Regarding the promo, it will be implemented in another PR and any styled deemed unnecessary will be removed then.

@honstar - I think @overmyheadandbody answered it here https://github.com/adobecom/milo/pull/533#issuecomment-1457913145

mokimo:
Yup, I've discussed this with @overmyheadandbody and it's something he will tackle next in a new PR - this one is already big enough as is and I'll build up on the logic in a coming feature so we want to merge this initial code.

narcis-radu (approved):
awesome work!

overmyheadandbody:
Hi @honstar! As already mentioned, the promo will be handled in a future commit, but we need the ground work from this PR to unblock other stories.

Also, this PR is open against a feature branch, `feds-gnav`, not the `main` one, so production will not be affected by this change.

honstar (approved):

PR #536, opened 2023-03-07
MWPW-127611 - Preflight Performance

auniverseaway:
* Display list of found URLs immediately
* Lazily load helix admin statuses

Resolves: [MWPW-127611](https://jira.corp.adobe.com/browse/MWPW-127611)

**Test URLs**
- Before: https://main--milo--adobecom.hlx.page/drafts/cmillar/preflight?martech=off
- After: https://preflight-perf--milo--adobecom.hlx.page/drafts/cmillar/preflight?martech=off

**Acrobat URLs**
- Before: https://main--dc--adobecom.hlx.page/acrobat/online/pdf-to-ppt
- After: https://main--dc--adobecom.hlx.page/acrobat/online/pdf-to-ppt?milolibs=preflight-perf

---

**Steps to reproduce**
1. Open any _before_ page.
2. Click `Preflight` in Sidekick.
3. Observe time it takes for URLs to appear.

**Steps to see fix**
1. Open any _after_ page.
2. Click `Preflight` in Sidekick.
3. Observe time it takes for URLs to appear.

### DIFF
--- a/libs/blocks/preflight/panels/general.js
+++ b/libs/blocks/preflight/panels/general.js
@@ -11,8 +11,7 @@ function getAdminUrl(url, type) {
   return type === 'status' ? `${base}?editUrl=auto` : base;
 }
 
-async function getStatus(suppliedPath) {
-  const url = new URL(suppliedPath);
+async function getStatus(url) {
   const adminUrl = getAdminUrl(url, 'status');
   const resp = await fetch(adminUrl);
   if (!resp.ok) return {};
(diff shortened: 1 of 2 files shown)


mokimo (approved):
Great performance/quality of life improvement. Visual feedback is there a lot faster and that feels a lot better as user

libs/blocks/preflight/panels/general.js
```diff
@@ -23,44 +22,42 @@ async function getStatus(suppliedPath) {
...
-      const url = new URL(dataPath);
-      dataPath = url.pathname;
-    } catch (e) {
-      // not a fully qualified URL
-    }
-    let path = dataPath ? `${window.location.origin}${dataPath}` : el.href;
-    path = path.endsWith('/') ? `${path}index` : path;
-    return getStatus(path);
+async function getStatuses() {
+  Object.keys(content.value).forEach((key) => {
```
overmyheadandbody, line 26: I wonder whether this is a good use case for `for...of`, since `forEach` might lead to unexpected results within `async/await` - https://gist.github.com/joeytwiddle/37d2085425c049629b80956d3c618971

auniverseaway: @overmyheadandbody normally, I would [agree](https://github.com/adobecom/milo/blob/main/libs/utils/utils.js#L671).

If we were concerned about doing something in a specific order, or if we do not want the next item to be fetched until the current is done, a for loop would be the best use. The other use case is if we are trying to set values outside of the loop.

In this case, we are using forEach's poor async/await support as an advantage. All the requests are sent nearly concurrently, and they come back when they come back (and out of order). We then set the content within the context of the item in the forEach. This has fairly substantial performance benefits since all requests are sent at nearly the same time.

Based on some quick calculations, i see:
1. ForEach - 1917ms
2. For loop - 20895ms

I'm inclined to keep this as is unless we:

a) find a race condition with setting the content inside the loop
b) find out helix admin cannot handle so many concurrent requests

Using for loop...
Screen Shot 2023-03-07 at 10 40 45 AM

Using forEach...
Screen Shot 2023-03-07 at 10 43 51 AM

chrischrischris: I agree that running in parallel is preferred here.  I think you should include a comment explaining this for future devs though.

Do not change this - but this is a case where the Promise.then(...) form is clearer to me.

chrischrischris (approved):

chrischrischris:
nit: I think the  `content.value = { ...content.value };` on line 68 is not needed as it's updated again on line 78

PR #537, opened 2023-03-07
Lana: Remove event handlers before import

chrischrischris:
If there is an unhandled error it will trigger a lana.log message, which tries importing ./lana.js, that will throw an error, which is then caught by the error listener which tries to log a lana message. 

This happens on DC when using the `milolibs` param as the DC CSP (Content Security Policy) doesn't allow loading of files from milobranch--main--adobecom.hlx.page.

**Test URLs:**

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -759,9 +759,9 @@ export function loadLana(options = {}) {
 
   window.lana = {
     log: async (...args) => {
-      await import('./lana.js');
       window.removeEventListener('error', lanaError);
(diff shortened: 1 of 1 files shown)


rgclayton (approved):

auniverseaway (approved):

amauch-adobe:
Not that it matters since this has been merged already, but I'm just adding my comments to Chris P. from my slack convo for posterity: 

> I checked the url and its good now with your lana loop branch fix.  I tested your branch as well as Victor's georouting branch you added the test fix too.

PR #540, opened 2023-03-08
MWPW-127701 CaaS Publish Date

chrischrischris:
Add support for `publishdate` metadata.
Adds a check for dates in day-month-year format

Resolves: [MWPW-127701](https://jira.corp.adobe.com/browse/MWPW-127701)

**Test URLs:**
- After: https://c3-caaspublishdate--milo--adobecom.hlx.page/?martech=off


tools/send-to-caas/send-utils.js
```diff
@@ -249,6 +249,14 @@ const getTags = (s) => {
 const getDateProp = (dateStr, errorMsg) => {
   if (!dateStr) return undefined;
   try {
+    if (dateStr.length === 10) {
+      const [m, d, y] = dateStr.split('-');
+      if (m > 12) {
```
rgclayton, line 254: Month greater than 12?

meganthecoder: That means that some dates will be misleading, but I suppose you don't have any way of knowing?

chrischrischris: Yep.  No way of knowing.  The alternative is to fail and force authors to fix it in the metadata.

meganthecoder (approved):

tools/send-to-caas/send-utils.js
rgclayton: Got it. Thanks for the explanation.

rgclayton (approved):

sukamat (approved):

PR #541, opened 2023-03-08
MWPW-123646 & MWPW-123874 [Blog2Milo]

saugatmalla:
* Add article-header block
* Add tags block

**Test URLs:**

- **Article header**
- - Before: https://main--blog--adobecom.hlx.page/drafts/smalla/article-header
- - After: https://main--blog--adobecom.hlx.page/drafts/smalla/article-header?milolibs=blog-article&martech=off

- **Tags**
- - Before: https://main--blog--adobecom.hlx.page/drafts/smalla/tags?martech=off
- - After: https://main--blog--adobecom.hlx.page/drafts/smalla/tags?milolibs=blog-article&martech=off


rgclayton:
Getting an "invalid link" error on the page.
image

rgclayton:
Something odd is going on with the section around the first article header. I noticed in the Word doc there is a malformed section flag (`---`). It's showing two dashes. If I correct that so it's a proper section the social share links disappear for that first article header.

To reproduce: 

- Edit the Word doc so that the outlined section is the proper `---`. Save and preview.
- After making that edit you'll notice the social share icons are no longer on the page for the first article header.

image

**After editing Word doc to fix section:**
image

saugatmalla:
> Getting an "invalid link" error on the page. Is this on purpose to call out bad links?
Yes, this is intentional.

rgclayton:
The share link copy button isn't working now. The little tooltip isn't displaying either when clicked on.
image

rgclayton (approved):

libs/blocks/article-header/article-header.js
```diff
@@ -0,0 +1,184 @@
...
+}
+
+async function copyToClipboard(button) {
+  try {
+    await navigator.clipboard.writeText(window.location.href);
+    button.setAttribute('title', await replaceKey('copied-to-clipboard', getConfig()));
+    button.setAttribute('alt', await replaceKey('copied-to-clipboard', getConfig()));
+    button.setAttribute('aria-label', await replaceKey('copied-to-clipboard', getConfig()));
+
+    const tooltip = createTag('div', { role: 'status', 'aria-live': 'polite', class: 'copied-to-clipboard' }, `${await replaceKey('copied-to-clipboard', getConfig())}`);
```
meganthecoder, line 55: Seems like this should be pulled out into a variable  
`const copiedToClipboard = await replaceKey('copied-to-clipboard', getConfig())`

libs/blocks/tags/tags.js
```diff
@@ -0,0 +1,25 @@
...
+  if (!tags) return;
+
+  if (!getTaxonomyModule()) {
+    await loadTaxonomy();
+  }
+
+  blockEl.innerHTML = '';
+  const tagsArray = tags.split(', ').map((tag) => tag.trim());
+
+  const articleTax = computeTaxonomyFromTopics(tagsArray);
```
meganthecoder, line 16: Remove unnecessary line
```suggestion
  const tagsArray = tags.split(', ').map((tag) => tag.trim());
  const articleTax = computeTaxonomyFromTopics(tagsArray);
```

meganthecoder (commented):
A few comments for you here. Also, it looks like you missed a few lines for the code coverage. Is it possible to add more tests for those?

libs/blocks/article-header/article-header.js
```diff
@@ -0,0 +1,184 @@
...
+  bylineContainer.firstElementChild.classList.add('article-byline-info');
+
+  const author = bylineContainer.firstElementChild.firstElementChild;
+  const authorEl = author.querySelector('a');
+  const authorURL = authorEl.href;
+  const authorName = author.textContent;
+  author.classList.add('article-author');
+
+  const date = bylineContainer.firstElementChild.lastElementChild;
+  date.classList.add('article-date');
```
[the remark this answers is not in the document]
saugatmalla: I added a small check.

libs/blocks/tags/tags.js
```diff
@@ -0,0 +1,26 @@
...
+  console.log(articleTax)
+  const tagsWrapper = createTag('p');
+
+  articleTax.visibleTopics?.forEach((topic) => {
+    const link = getLinkForTopic(topic);
+    tagsWrapper.insertAdjacentHTML('beforeend', link);
```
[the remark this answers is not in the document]
auniverseaway: @saugatmalla can you fix this?

auniverseaway (changes requested):

libs/blocks/article-header/article-header.js
```diff
@@ -0,0 +1,198 @@
...
+  const copyButton = sharing.querySelector('#copy-to-clipboard');
+  copyButton.addEventListener('click', async () => {
+    await copyToClipboard(copyButton);
+  });
+
+  return sharing;
+}
+
+async function validateDate(date) {
+  console.log(date)
```
chrischrischris, line 148: remove

libs/blocks/article-header/article-header.js
```diff
@@ -0,0 +1,198 @@
...
+      img.addEventListener('error', () => {
+        img.remove();
+      });
+    } else {
+      authorImg.style.backgroundImage = 'none';
+    }
+  }
+}
+
+async function copyToClipboard(button) {
```
chrischrischris, line 67: I wonder if this should be moved into a central file.  Something like /libs/utils/tools.js
```suggestion
async function copyToClipboard(button, copyTxt) {
```

libs/blocks/article-header/article-header.js
```diff
@@ -0,0 +1,198 @@
...
+  const anchorTags = platforms.map(async (platform) => {
+    const platformProperties = platformMap[platform];
+    if (platformProperties) {
+      return createTag('a', platformProperties, svgs[platform].cloneNode(true));
+    }
+    return null;
+  });
+
+  const sharing = createTag('div', { class: 'article-byline-sharing' });
+  const allAnchorTags = await Promise.all(anchorTags);
```
chrischrischris, line 130: I don't think you need to use promises as nothing in anchorTags actually awaits anything
```suggestion
  const allAnchorTags = platforms.map((platform) => {
    const platformProperties = platformMap[platform];
    if (platformProperties) {
      return createTag('a', platformProperties, svgs[platform].cloneNode(true));
    }
    return null;
  }).filter(Boolean);

  const sharing = createTag('div', { class: 'article-byline-sharing' });
```

libs/blocks/article-header/article-header.js
```diff
@@ -0,0 +1,198 @@
...
+import { loadTaxonomy, getLinkForTopic, getTaxonomyModule } from '../article-feed/article-helpers.js';
+import { replaceKey } from '../../features/placeholders.js';
+import { fetchIcons } from '../../features/icons.js';
+import { buildFigure } from '../figure/figure.js';
+
+async function validateAuthorUrl(url) {
+  if (!url) return null;
+
+  const resp = await fetch(`${url}.plain.html`);
+  if (!resp || !resp.ok) {
```
chrischrischris, line 11: nit
```suggestion
  if (!resp?.ok) {
```

libs/blocks/article-feed/article-helpers.js
```diff
@@ -115,8 +114,7 @@ export function getTaxonomyModule() {
 }
```
chrischrischris, line 114: You may want to use a closure wrapped var that self initializes instead of a global.  This way you will not need to init `taxonomyModule` in article-header.js and tags.js.  It will require a change to `loadTaxonomy` to return `taxonomyModule` and a change in getLinkForTopic to `await` any calls to `getTaxonomyModule`:

```
export const getTaxonomyModule = (() => {
  let taxonomyModule;
  return async () => {
    if (!taxonomyModule) {
      taxonomyModule = await loadTaxonomy();
    }
    return taxonomyModule;
  };
})();
```

meganthecoder (approved):

test/blocks/article-header/article-header.test.js
```diff
@@ -0,0 +1,85 @@
...
+const invalidDoc = await readFile({ path: './mocks/body-invalid.html' });
+
+describe('article header', () => {
+  const block = document.body.querySelector('.article-header');
+
+  it('creates article header block', async () => {
+    config.locale.contentRoot = '/test/blocks/article-header/mocks';
+
+    await init(block);
+    expect(document.body.querySelector('.article-category')).to.be.exist;
```
JasonHowellSlavin, line 25: can remove `be` from the assertion: `expect(document.body.querySelector('.article-category')).to.exist`;

this occurs elsewhere where you could remove it.

test/blocks/article-header/article-header.test.js
```diff
@@ -0,0 +1,85 @@
...
+    expect(document.body.querySelector('.article-date')).to.be.exist;
+    expect(document.body.querySelector('.article-byline-sharing')).to.be.exist;
+  });
+
+  it('should open link popup when share links are clicked', () => {
+    const shareLink = document.querySelector('.article-byline-sharing a');
+    const spy = sinon.spy(shareLink, 'click');
+    shareLink.click();
+    expect(spy.called).to.be.true;
+  });
```
JasonHowellSlavin, line 38: In this test, instead of spying on the function call, could you wait and query the pop up to see if it exists? That would be more closely testing the functionality surfaced to the end user.

test/blocks/article-header/article-header.test.js
```diff
@@ -0,0 +1,85 @@
...
+    copyLink.click();
+
+    const tooltip = await waitForElement('.copied-to-clipboard');
+
+    expect(tooltip).to.be.exist;
+    writeTextStub.restore();
+  });
+});
+
+describe('test the invalid block', () => {
```
JasonHowellSlavin, line 65: nit: rename to something that matches the describe block above, i.e. "invalid article header"

test/blocks/article-header/article-header.test.js
```diff
@@ -0,0 +1,85 @@
...
+  beforeEach(() => {
+    document.body.innerHTML = invalidDoc;
+  });
+
+  it('tests invalid url', async () => {
```
JasonHowellSlavin, line 70: change the string here to something more semantic, like: "does not init if the element is invalid"

test/blocks/article-header/article-header.test.js
```diff
@@ -0,0 +1,85 @@
...
+    const authorLink = document.querySelector('.article-author a');
+    expect(authorTextEl).to.be.exist;
+    expect(authorLink).to.not.exist;
+  });
+
+  it('adds invalid-date when invalid date is provided', async () => {
+    await init(document.body.querySelector('.article-header'));
+
+    const date = await waitForElement('.article-date-invalid');
+    console.log(date);
```
JasonHowellSlavin, line 82: do you still want this console log?

test/blocks/tags/tags.test.js
```diff
@@ -0,0 +1,27 @@
...
+
+  afterEach(() => {
+    document.body.innerHTML = ogDoc;
+  });
+
+  it('renders tags block', async () => {
+    document.body.innerHTML = await readFile({ path: './mocks/body.html' });
+    const block = document.querySelector('.tags');
+    await init(block);
+    expect(document.body.querySelector('.tags')).to.be.exist;
```
JasonHowellSlavin, line 25: `to.exist`

JasonHowellSlavin (approved):
Good stuff. Just some test cleanup, nothing blocking from my end.

libs/blocks/article-feed/article-helpers.js
saugatmalla: I like this idea. But, this is going to affect a lot of methods and files which are using taxonomy methods. 
We need to change many methods to async functions and use Promises for loops. I wonder if we should make all the changes in this PR.

test/blocks/article-header/article-header.test.js
saugatmalla: I updated the test. Could you take a look and LMK if it's good?

libs/blocks/article-feed/article-helpers.js
chrischrischris: It's fine to leave as is, but I'm only seeing it used in 2 places:
https://github.com/search?q=repo%3Aadobecom%2Fmilo%20getTaxonomyModule&type=code

saugatmalla: Yeah, but these changes are inside other methods which are inside other methods/modules, and so on.

PR #542, opened 2023-03-08
[MWPW-126441] Clean up typography.css

elan-tbx:
Out of an abundance of caution, #503 kept the old uppercase t-shirt classnames in typography.css in addition to the new lowercase classnames. After in-depth review across milo and its consumers, I think it's safe to say that all blocks have been transitioned to the lowercase version successfully. We should remove these old rules from typography.css to avoid any confusion with future code contribution.

This PR also fixes formatting of some rules in typography.css so that it passes style-linting.

Resolves: [MWPW-126441](https://jira.corp.adobe.com/browse/MWPW-126441)

**Test URLs:**
(Note that these should both look the same)
- Before: https://main--milo--adobecom.hlx.page/docs/library/blocks/marquee?martech=off
- After: https://ebartholomew-cleanup-typography--milo--adobecom.hlx.page/docs/library/blocks/marquee?martech=off


mokimo (approved):
Cool, always a big fan of keeping repo's clean and getting rid of stuff that isn't needed anymore! 👍

ryanmparrish (approved):

PR #554, opened 2023-03-10
MWPW-127043 Fix Region Selector US Link

meganthecoder:
* Removes extra ".html" in url path for US link in region selector (e.g. `https://business.adobe.com/.html/customer-success-stories.html`)
* Fixes all links to US home with hash or query params on .html

Resolves: [MWPW-127043](https://jira.corp.adobe.com/browse/MWPW-127043)

**Test URLs:**
- Before: https://business.stage.adobe.com/customer-success-stories.html?martech=off
- After: https://business.stage.adobe.com/customer-success-stories.html?milolibs=methomas-change-region-html&martech=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -248,12 +248,12 @@ export function appendHtmlPostfix(area = document) {
 
   const HAS_EXTENSION = /\..*$/;
   const shouldNotConvert = (href) => {
-    let url;
+    let url = { pathname: href };
 
     try { url = new URL(href, pageUrl) } catch (e) {}
 
     if (!(href.startsWith('/') || href.startsWith(pageUrl.origin))
-      || url?.pathname.endsWith('/')
+      || url.pathname?.endsWith('/')
       || href === pageUrl.origin
       || HAS_EXTENSION.test(href.split('/').pop())
       || htmlExclude?.some((excludeRe) => excludeRe.test(href))) {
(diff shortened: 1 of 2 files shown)


meganthecoder:
I'm not really sure how to verify this. I'm trying to get milolibs working on business.stage but would appreciate any other suggestions if you have them.

meganthecoder:
> I'm not really sure how to verify this. I'm trying to get milolibs param working on business.stage but would appreciate any other suggestions if you have them.

Got the milolibs param working https://github.com/adobecom/bacom/pull/87 and added test urls to the description

Brandon32 (approved):

libs/utils/utils.js
```diff
@@ -248,8 +248,12 @@ export function appendHtmlPostfix(area = document) {
 
   const HAS_EXTENSION = /\..*$/;
   const shouldNotConvert = (href) => {
+    let url;
```
chrischrischris, line 251: just in case `new URL` fails, should we init this href?
```suggestion
    let url = { pathname: href };
```

chrischrischris (approved):
Thanks for fixing this.  1 Possible change

rgclayton (approved):

meganthecoder:
Brandon pointed out that the "before" url doesn't actually reproduce the issue. It looks like that's because the link is authored as the prod url, so `shouldNotConvert` returns true. I thought we were supposed to be able to author stage urls and they'd be automatically converted to the correct environment. Could be remembering that wrong. I've reached out to GWP for clarification

JasonHowellSlavin (approved):

meganthecoder:
@Brandon32 This has been fixed now so please verify when you get a chance. 

The "before" link shows the extra `/.html/` and the "after" shows the correct url

PR #559, opened 2023-03-13
Bulk preview and publish tool

jckautzmann:
- adds support for bulk previewing and publishing of urls

**Features**

- displays the status for each URL
- supports resuming after a refresh
- detects and displays duplicate URLs
- persists a report in an Excel file for each operation with: timestamp, user, operation, site, number of urls, number of success
- persists urls between reloads
- supports IMS sign-in
- only authorised users can perform a bulk operation
- only supported sites can be bulk operated
- maximum batch size: 1000 urls
- throttling of 100 ms between actions
- the single bulk action (fetch) times out after 5 s

**Bulk UI**

Screenshot 2023-04-17 at 13 30 32

**Report (Excel file)**

Screenshot 2023-04-17 at 13 32 05

**Test URLs:**
- After: https://bulk-publish--milo--adobecom.hlx.page/tools/bulk-publish/index.html?martech=off


JasonHowellSlavin (approved):
Looks great. Would it be worth splitting some of the functions out into their own file? at all to separate the preact components from utility functions?

rgclayton:
When hitting the "run process" button, with no URLS added, I do get the message that tells me no URLS were processed and I need to add urls. When I add URLS and click "run process" again the page does nothing.
In the console I see:
image


Refreshing the page and adding my urls again works as expected - preview/publishing the urls.

jckautzmann:
Note: the bad PSI score (68) is achieved because IMS log in is enabled. When disabled the PSI is 98 (see above).

libs/blocks/bulk/bulk.js
```diff
@@ -0,0 +1,345 @@
...
+
+function StatusRow({ row }) {
+  const timeStamp = prettyDate();
+  const errorStyle = 'status-error';
+  const previewStatusError = row.status.preview !== 200
+    ? errorStyle
+    : '';
+  const publishStatusError = row.status.publish !== 200
+    ? errorStyle
+    : '';
```
mokimo, line 108: I'm always confused by having negative ternaries - it's clearer to swap the conditions around, something like:
```suggestion
  const previewStatusError = row.status.preview === 200 ? '' : errorStyle
  const publishStatusError = row.status.publish === 200 ? '' : errorStyle
```

libs/blocks/bulk/utils.js
```diff
@@ -0,0 +1,246 @@
...
+  }
+  return true;
+};
+
+export const signOut = (e) => {
+  e.preventDefault();
+  window.adobeIMS?.signOut();
+};
+
+export const getAuthorizedUsers = async () => {
```
mokimo, line 46: Security on the frontend? There is none. 
But this is good enough given the use-case?

mokimo (approved):
Looks great, I think it's a cool tool.

A few things on my mind after checking it out / improvement ideas:
- If you continue bulk publishing after navigating away from the page, you lose the history of the first few entries. It would make sense to also store that in localstorage so you see which URLs already have been previewed/published & their status
- filtered duplicate URLs could be logged, so an author can look why/which duplicates are even showing up
- consider adding LANA logs so you can monitor any JS errors encountered. Though I'd imagine authors would bug you otherwise, if something is not working!

libs/blocks/bulk/bulk.css
```diff
@@ -0,0 +1,451 @@
...
+}
+
+.bulk-urls-title,
+.bulk-user-header {
+  text-transform: uppercase;
+  font-weight: bold;
+  position: relative;
+}
+
+.bulk-urls-title::after {
```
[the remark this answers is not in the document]
jckautzmann: I'll let @rgclayton chime in for the UI/UX part :)

libs/blocks/bulk/bulk.js
jckautzmann: Addressed in https://github.com/adobecom/milo/pull/559/commits/3c8cc9387ed85003d8590024063603cba69b2b11

jckautzmann:
@mokimo thx for your feedback!

> If you continue bulk publishing after navigating away from the page, you lose the history of the first few entries. It would make sense to also store that in localstorage so you see which URLs already have been previewed/published & their status

good point, the results from the previous run are now stored in local storage and displayed when resuming: see https://github.com/adobecom/milo/pull/559/commits/3c8cc9387ed85003d8590024063603cba69b2b11

> filtered duplicate URLs could be logged, so an author can look why/which duplicates are even showing up

you're right, duplicate URLs are now displayed in the results with a `duplicate` status. See https://github.com/adobecom/milo/pull/559/commits/3c8cc9387ed85003d8590024063603cba69b2b11

> consider adding LANA logs so you can monitor any JS errors encountered. Though I'd imagine authors would bug you otherwise, if something is not working!

Which JS errors are you thinking about?

mokimo:
Cool, great to see the feedback addressed! 👍 

As for JS errors, any uncaught errors would be interesting to observe.
It's a lot of code so it might make sense to have an eye on that after the first few weeks when the tool is used, to catch any unexpected behavior.

libs/blocks/bulk/bulk.css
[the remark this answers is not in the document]
rgclayton: I could go either way with this. 
An easy change if everyone feels the same. Shouldn't hold up this PR.

libs/blocks/bulk/utils.js
```diff
@@ -0,0 +1,267 @@
...
+  let authorizedUsers = getLocalStorage(BULK_AUTHORIZED_USERS);
```
chrischrischris, line 74: Should there be some expiration on pulling from localstorage?

libs/blocks/bulk/utils.js
```diff
@@ -0,0 +1,267 @@
...
+  if (authorizedUsers) return authorizedUsers;
+  const resp = await fetch(BULK_CONFIG_FILE_PATH);
+  const json = await resp.json();
+  authorizedUsers = json.users.data.map((user) => user.user);
+  setLocalStorage(BULK_AUTHORIZED_USERS, authorizedUsers);
+  return authorizedUsers;
+};
+
+const getSupportedSites = async () => {
+  let supportedSites = getLocalStorage(BULK_SUPPORTED_SITES);
```
chrischrischris, line 84: same

libs/blocks/bulk/utils.js
```diff
@@ -0,0 +1,267 @@
...
+  }
+};
+
+export const getUrls = (element) => {
+  return element.current?.value.split('\n')
+    .filter((url) => url.length > 0)
+    .map((e) => e.trim());
+};
+
+export const getActionName = (action, gerund) => {
```
chrischrischris, line 115: What is `gerund`?

jckautzmann: @chrischrischris it's a flag to either return the verb or its present participe form (root verb + 'ing'): 
getActionName('preview') returns 'Preview'
getActionName('preview', true) returns 'Previewing'
...

jckautzmann: to make it clearer I could rename it `useGerund`

chrischrischris: Ah ok - I learned a new word today!

libs/blocks/bulk/utils.js
jckautzmann: @chrischrischris makes senses: the cache for authorized users and supported sites is cleared once the bulk operation is done: see 
https://github.com/adobecom/milo/pull/559/commits/d0c655477c907aa4565e2b3994d5066ba2ee59a5

vhargrave (commented):
really looking forward to using this in some projects ! :) 
Nearly everything is working for me. 
I did notice that when I click on the sign out button nothing happens. Does that need to be fixed?
Additionally, was it a deliberate decision to update the package-lock.json with this PR?

libs/blocks/bulk/bulk.js
```diff
@@ -0,0 +1,341 @@
...
+  sendReport,
+  signIn,
+  getUser,
+  getStoredOperation,
+  storeUrls,
+  storeOperation,
+} from './utils.js';
+
+// eslint-disable-next-line max-len
+// TODO: enable IMS sign in when the code is merged on the main branch, as IMS does not work on the PR branch
```
honstar, line 19: Is this accurate? I think I reviewed a couple of milo PRs that did have IMS enabled but I might be wrong.

libs/blocks/bulk/bulk.js
```diff
@@ -0,0 +1,341 @@
...
+      resumeModal=${resumeModal}
+      resume=${resume}
+      hideModal=${hideModal} />
+  </div>
+  `;
+}
+
+export default async function init(el) {
+  if (IMS_SIGN_IN_ENABLED) {
+    const signedIn = await signIn(IMS_SIGN_IN_ENABLED);
```
honstar, line 334: There is no need to pass in an argument here.

libs/blocks/bulk/utils.js
```diff
@@ -0,0 +1,269 @@
+import { getImsToken } from '../../../tools/send-to-caas/send-utils.js';
+import { fetchWithTimeout } from '../caas/utils.js';
+import { loadScript } from '../../utils/utils.js';
+import {
+  getLocalStorage,
+  setLocalStorage,
+} from '../review/utils/localStorageUtils.js';
```
honstar, line 7: Same as above, ideally shared utility functions should be located in a central location.

libs/blocks/bulk/utils.js
```diff
@@ -0,0 +1,269 @@
+import { getImsToken } from '../../../tools/send-to-caas/send-utils.js';
+import { fetchWithTimeout } from '../caas/utils.js';
```
honstar, line 2: This seems suboptimal to reuse a utility from another block. Instead, the common bits should be moved to a more central location IMO.

honstar (changes requested):

libs/blocks/bulk/utils.js
```diff
@@ -0,0 +1,269 @@
...
+  return name;
+};
+
+const executeAction = async (action, url) => {
+  const operation = (action === 'preview') ? 'preview' : 'live';
+  const siteAllowed = await siteIsSupported(url);
+  if (!siteAllowed) return UNSUPPORTED_SITE_STATUS;
+  const { hostname, pathname } = new URL(url);
+  const [branch, repo, owner] = hostname.split('.')[0].split('--');
+  const adminURL = `${ADMIN_BASE_URL}/${operation}/${owner}/${repo}/${branch}${pathname}`;
```
honstar, line 137: On a general note, executing potentially long-running operations (in this case subsequent actions up until 1k pages) in the browser are not really considered stable. The alternative is to offload this to a microservice, e.g. I/O runtime action. Both loc and floodgate are currently refactored to move the traditional backend processing to external microservices.

libs/blocks/bulk/bulk.js
mokimo: Certain branches are whitedlisted. Such as the `gnav` branch, `ims`, `main` .. currently it's a manual ask to get a branch whitelisted for the milo token

libs/blocks/bulk/bulk.css
```diff
@@ -0,0 +1,451 @@
+button {
+  color: #FFF;
+  cursor: pointer;
+  font-family: var(--body-font-family);
+  font-size: 16px;
```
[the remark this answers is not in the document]
jckautzmann: The `font-size` is defined in `px` for almost all the Milo blocks. Not sure if we want to change this. If so, we should tackle it in a separate PR, not in this one. @rgclayton WDYT?

libs/blocks/bulk/bulk.js
jckautzmann: @honstar Good catch! I addressed in with https://github.com/adobecom/milo/pull/559/commits/80e5403a54c434c52a6e93baa6ba71b80ea23c4a

libs/blocks/bulk/bulk.js
vhargrave: didn't know that either - thx for sharing @mokimo !

libs/blocks/bulk/utils.js
jckautzmann: @honstar I addressed your comment as follows:
- I moved the bulk publish code below milo/tools to make it a tool (like floodgate), not a block
- I moved the methods used from specific blocks/tools to some shared utils
See commit https://github.com/adobecom/milo/pull/559/commits/e3ee3cb327d5098c5f6dd1f18491484ff2e18c58

libs/blocks/bulk/utils.js
jckautzmann: Please see comment below.

libs/blocks/bulk/utils.js
rgclayton: > @honstar I agree. As management and authors expect this tool to be released soon, I would release the bulk publish feature as it is now. Moving the processing to an external service should be tackled in a separate task. @rgclayton WDYT?

I agree. I'm assuming there will be some refinement as it used and we get more feedback.

libs/blocks/bulk/bulk.css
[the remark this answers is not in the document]
rgclayton: > The `font-size` is defined in `px` for almost all the Milo blocks. Not sure if we want to change this. If so, we should tackle it in a separate PR, not in this one. @rgclayton WDYT?

I think we've had this conversation before. Each unit has pro's and con's.

We have been using `px` for font-sizes pretty much across the board so let's stick to that.

rgclayton (approved):

sukamat (approved):

honstar (approved):

PR #576, opened 2023-03-17
[FAAS] Auto submit disable on config page.

seanchoi-dev:
Resolves: [MWPW-128249](https://jira.corp.adobe.com/browse/MWPW-128249)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/tools/faas?martech=off
- after: https://faas-fix--milo--adobecom.hlx.page/tools/faas?martech=off

### DIFF
--- a/libs/blocks/faas-config/faas-config.js
+++ b/libs/blocks/faas-config/faas-config.js
@@ -425,8 +425,8 @@ const Configurator = ({ rootEl }) => {
 
   useEffect(() => {
     if (isFaasLoaded) {
-      initFaas(state, document.getElementsByClassName('faas')[0]);
       saveStateToLocalStorage(state);
+      initFaas(JSON.parse(localStorage.getItem(LS_KEY)), document.getElementsByClassName('faas')[0]);
     }
   }, [isFaasLoaded, state]);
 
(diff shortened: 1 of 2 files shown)


sukamat (approved):

honstar (approved):

libs/blocks/faas-config/faas-config.js
```diff
@@ -425,8 +425,8 @@ const Configurator = ({ rootEl }) => {
 
   useEffect(() => {
     if (isFaasLoaded) {
-      initFaas(state, document.getElementsByClassName('faas')[0]);
       saveStateToLocalStorage(state);
+      initFaas(JSON.parse(localStorage.getItem(LS_KEY)), document.getElementsByClassName('faas')[0]);
```
chrischrischris, line 429: Why the change to reading from localstorage?  Isn't `state` the exact same?

seanchoi-dev: @chrischrischris This PR is trying to set state.as as false if it is configuration page.
By this change, the faas config renders form as `state.as` false even though user sets as true.

PR #587, opened 2023-03-22
MWPW-127772: M@S price/CTAs as autoblock

yesil:
* externalized tacocat library
* added async script support to milo

Due to CORS, the **After URL** should be tested with the following CORS settings:

`Access-Control-Allow-Origin: *`
`Access-Control-Allow-Headers: x-api-key`

Resolves: [MWPW-127772](https://jira.corp.adobe.com/browse/MWPW-127772)

**Test URLs:** (corp VPN needed)
- Before: https://main--milo--adobecom.hlx.page/fragments/ilyas/merch?martech=off
- After: https://mwpw-127772--milo--yesil.hlx.page/fragments/ilyas/merch?martech=off


yesil:
Due to CORS, the after URL should be tested the following configs:

image

libs/blocks/merch/merch.js
```diff
@@ -1,57 +1,151 @@
...
+  if (target != null) {
+    Object.entries(target).forEach(([key, value]) => {
+      if (value == null) delete target[key];
+    });
+  }
+  return target;
+}
+
+const getTacocatEnv = (envName, ietf) => {
+  const scriptUrl = (envName === envProd
```
vhargrave, line 34: envProd can be moved inside this method. No reason to have it global.

libs/styles/styles.css
```diff
@@ -218,6 +218,12 @@ picture.bg-img img {
   visibility: hidden;
 }
 
+/* merch */
+[data-wcs-osi],
+a[href^="/tools/ost"] {
```
auniverseaway, line 223: This shouldn't be needed.

libs/blocks/merch/merch.js
```diff
@@ -1,57 +1,151 @@
...
+    ? 'https://www.adobe.com/special/tacocat/lib/1.12.0/tacocat.js'
+    : 'https://www.stage.adobe.com/special/tacocat/lib/1.12.0/tacocat.js');
+  // eslint-disable-next-line prefer-const
+  let [language, country = 'us'] = ietf.split('-', 2);
+  if (!supportedLanguages.includes(language)) {
+    language = 'en'; // default to english
+  }
+  const literalScriptUrl = (envName === envProd
+    ? `https://www.adobe.com/special/tacocat/literals/${language}.js`
+    : `https://www.stage.adobe.com/special/tacocat/literals/${language}.js`);
```
vhargrave, line 44: nit: 
```
const literalScriptUrl = envName === envProd
    ? `https://www.adobe.com/special/tacocat/literals/${language}.js`
    : `https://www.stage.adobe.com/special/tacocat/literals/${language}.js`;
```

libs/blocks/merch/merch.js
vhargrave, line 34: nit: you don't need to put this within parentheses -> () ;

honstar:
@yesil : there's a typo in your mod headers extension: it should read `Access-Control-Allow-Headers` (plural), otherwise you'll still get CORS errors.

vhargrave (commented):
Looking pretty good @yesil 
Found a few small things that you can look at.

Additionally I think expanding your tests to see these merch links get transformed within blocks would be helpful.

libs/styles/styles.css
yesil: @auniverseaway this is to prevent merch links being displayed while waiting for merch block logic and styles.
Otherwise it gives a flicker effect. The unprocessed merch links should not be visible to the visitor.

libs/blocks/merch/merch.js
```diff
@@ -1,57 +1,151 @@
...
+
+/**
+ * Removes undefined properties from an object
+ * @param {*} target Object with potentially undefined properties
+ * @returns new Object without undefined properties
+ */
+export function omitUndefined(target) {
+  if (target != null) {
+    Object.entries(target).forEach(([key, value]) => {
+      if (value == null) delete target[key];
```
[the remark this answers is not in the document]
yesil: @vhargrave this is on purpose.
We only want to omit undefined and null hence double equals instead of triple.

libs/blocks/merch/merch.js
yesil: this is to reach 100% test coverage. Otherwise it is not easy to get this method called with `env = 'prod'`

yesil:
@honstar regarding `Access-Control-Allow-Header` it is actually `Access-Control-Allow-Headers` but for some reason the UI cuts the `s`.
See when I disable the rule, it fails.
 image

yesil:
@vhargrave for the replacement of the prices/cta urls, I think we should consider e2e tests, as it depends on many conditions such as external lib, literal files, CORS setting, etc.   any take on this ? @auniverseaway @chrischrischris ?

libs/blocks/merch/merch.js
```diff
@@ -1,57 +1,151 @@
...
+function loadTacocat() {
+  if (initialized) {
+    return;
+  }
+  initialized = true;
+  const {
+    env,
+    locale: { ietf },
+  } = getConfig();
+  const { scriptUrl, literalScriptUrl, country, language } = getTacocatEnv(env.name, ietf);
```
3ch023, line 65: 'ietf' is a confusing name if used outside of the 'locale' object. it says nothing about what is in the value, pls consider using something like 'ietfLocale' or just 'locale'

libs/styles/styles.css
yesil: This is also needed to hide unresolved placeholders while merch.css is still being loaded.
Happened several times during development.

libs/blocks/merch/merch.js
```diff
@@ -1,57 +1,151 @@
...
+
+  Promise.all([
+    loadScript(literalScriptUrl, undefined, true),
```
3ch023, line 68: If the requested literals script is not available for some reason it will fail the whole tacocat price render right?
maybe we could have a fallback and try to load en literals instead?
or even have the hard version of en literals somewhere in the source code for the worst-case scenario.

if page has only CTA it doesn't even need literals, so taco for sure should load for it without literals js in place.

vhargrave:
> @vhargrave for the replacement of the prices/cta urls, I think we should consider e2e tests, as it depends on many conditions such as external lib, literal files, CORS setting, etc. any take on this ? @auniverseaway @chrischrischris ?

@yesil sorry, didn't mean unit tests, but what you're testing in your word file atm. I'd expand the content in your word file to demonstrate that links are transformed correctly within blocks :)

libs/blocks/merch/merch.js
vhargrave: ok 👍

libs/blocks/merch/merch.js
yesil: renamed.

libs/blocks/merch/merch.js
yesil: en literals are bundled with consonant-templates in tacocat as a fallback. Current tacocat implementation requires the literals to be provided before initialisation even if empty object.

What you suggest could be an improvement on tacocat first and then we can update milo codebase.

I have improved the error handling so if the literal resource cannot be loaded, it will not block the initialization.

test/utils/utils.test.js
```diff
@@ -116,6 +116,12 @@ describe('Utils', () => {
       }
     });
 
+    it('Load async script', async () => {
+      const asyncScript = await utils.loadScript('/test/utils/mocks/script.js', undefined, true);
+      expect(asyncScript).to.exist;
+      expect(asyncScript.getAttribute('async')).to.equal('true');
```
honstar, line 122: This test currently fails, can you please check @yesil ?

honstar: Looks good now 👍

libs/blocks/merch/merch.js
```diff
@@ -1,57 +1,146 @@
...
-    const price = buildPrice(osi, 'price');
-    el.append(price);
-  }
+  const promotionCode = (searchParams.get('promo')
+    ?? el.closest('[data-promotion-code]')?.dataset.promotionCode) || undefined;
 
-  const wcs = { apiKey: 'wcms-commerce-ims-ro-user-cc' };
-  window.tacocat({ environment: env.name, wcs });
+  if (isCTA(type)) {
+    const checkoutContext = {}; // TODO: load dynamically
```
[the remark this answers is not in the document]
yesil: this will be addressed in a future PR. 
I have seen similar TODO comments so followed the same approach.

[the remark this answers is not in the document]
honstar: Ok, as long as we're tracking this work somewhere.

libs/blocks/merch/merch.js
```diff
@@ -1,57 +1,151 @@
...
-  if (!a) return null;
+let initialized = false;
+
+// supported languages for the price literals
+const supportedLanguages = [
+  'ar', 'bg', 'cs', 'da', 'de', 'en', 'es', 'et', 'fi', 'fr', 'he', 'hu', 'it', 'ja', 'ko',
+  'lt', 'lv', 'nb', 'nl', 'pl', 'pt', 'ro', 'ru', 'sk', 'sl', 'sv', 'tr', 'uk', 'zh_CN', 'zh_TW',
+];
+
+export function omitUndefined(target) {
```
Axelcureno, line 19: Nit: remove export

libs/styles/styles.css
npeltier: shouldn't we in that case add a class for auto blocks sharing that requirement? @tuerkben @auniverseaway

Axelcureno (approved):

honstar (approved):
LGTM

libs/blocks/merch/merch.js
```diff
@@ -1,57 +1,152 @@
...
+  }
+  return target;
+}
+
+const getTacocatEnv = (envName, locale) => {
+  const scriptUrl = envName === envProd
+    ? `https://www.adobe.com/special/tacocat/lib/${tacocatVersion}/tacocat.js`
+    : `https://www.stage.adobe.com/special/tacocat/lib/${tacocatVersion}/tacocat.js`;
+  // eslint-disable-next-line prefer-const
+  let [language, country = 'us'] = locale.split('-', 2);
```
vhargrave, line 34: I'm not sure if the country should be 'us' for mena countries. I took a quick look on adobe.com and they also have different prices served to them.

vhargrave (commented):
see comment :)

libs/blocks/merch/merch.js
```diff
@@ -1,57 +1,152 @@
...
-  if (!a) return null;
+let initialized = false;
+
+// supported languages for the price literals
+const supportedLanguages = [
+  'ar', 'bg', 'cs', 'da', 'de', 'en', 'es', 'et', 'fi', 'fr', 'he', 'hu', 'it', 'ja', 'ko',
+  'lt', 'lv', 'nb', 'nl', 'pl', 'pt', 'ro', 'ru', 'sk', 'sl', 'sv', 'tr', 'uk', 'zh_CN', 'zh_TW',
+];
+
+function omitUndefined(target) {
```
chrischrischris, line 20: nit: `omitNullValues`?

libs/blocks/merch/merch.js
```diff
@@ -1,57 +1,152 @@
...
+    ? `https://www.adobe.com/special/tacocat/lib/${tacocatVersion}/tacocat.js`
+    : `https://www.stage.adobe.com/special/tacocat/lib/${tacocatVersion}/tacocat.js`;
+  // eslint-disable-next-line prefer-const
+  let [language, country = 'us'] = locale.split('-', 2);
+  if (!supportedLanguages.includes(language)) {
+    language = 'en'; // default to english
+  }
+  const literalScriptUrl = envName === envProd
+    ? `https://www.adobe.com/special/tacocat/literals/${language}.js`
+    : `https://www.stage.adobe.com/special/tacocat/literals/${language}.js`;
```
chrischrischris, line 40: nit

```suggestion
  // eslint-disable-next-line prefer-const
  let [language, country = 'us'] = locale.split('-', 2);
  if (!supportedLanguages.includes(language)) {
    language = 'en'; // default to english
  }
  
  const host = envName === envProd
    ? 'https://www.adobe.com'
    : 'https://www.stage.adobe.com';

  const scriptUrl = `${host}/special/tacocat/lib/${tacocatVersion}/tacocat.js`;
  const literalScriptUrl = `${host}/special/tacocat/literals/${language}.js`
```

chrischrischris (approved):

libs/blocks/merch/merch.js
yesil: good catch, addressed.

libs/blocks/merch/merch.js
```diff
@@ -1,57 +1,159 @@
...
+  getConfig,
+  createTag,
+} from '../../utils/utils.js';
 
-const { miloLibs, codeRoot, env } = getConfig();
-const base = miloLibs || codeRoot;
+const version = '1.12.0';
+const wcs = { apiKey: 'wcms-commerce-ims-ro-user-milo' };
+const envProd = 'prod';
+const ctaPrefix = /^CTA +/;
```
chrischrischris, line 10: one more nit: const values should be UPPERCASED_NAMES

chrischrischris: same for `SUPPORTED_LANGS` and `GEO_MAPPINGS` and those should be above `initialized`

PR #595, opened 2023-03-27
MWPW-120932: Org Logo JSON LD Schema

fullcolorcoder:
Adds organization as a type for JSON LD data in the metadata table. 

Resolves: [MWPW-120932](https://jira.corp.adobe.com/browse/MWPW-120932)

Rich Results Test: https://search.google.com/test/rich-results/result?id=V6Va-r_fW4q6sphEqj5ngA

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/bradjohn/org-logo-schema?martech=off
- After: https://org-schema--milo--adobecom.hlx.page/drafts/bradjohn/org-logo-schema?martech=off


libs/features/richresults.js
```diff
@@ -33,12 +33,23 @@ function getRichResultsForSiteSearchBox(getMetadata) {
...
+  };
+}
+
 function getRichResults(type, getMetadata) {
   switch (type) {
     case 'NewsArticle':
       return getRichResultsForNewsArticle(getMetadata);
     case 'SiteSearchBox':
       return getRichResultsForSiteSearchBox(getMetadata);
+    case 'organization':
```
hparra, line 51: nit: Uppercase to match Schema.org names but I realize this is passed in from `<meta name="richresults" content="organization">` which may already exist.

fullcolorcoder: Yea I agree. Word will auto capitalize this too it seems, so good call thanks :)

jckautzmann (approved):

fullcolorcoder:
@amauch-adobe I assigned this to you quick, if you have the bandwidth would you mind verifying for me?

hparra:
Related unit test is failing

hparra (commented):
Related Organization unit test is failing.
Is the case being changed case somewhere to reflect capitalization in switch?

amauch-adobe (changes requested):
Everything looks good with the test urls on all browsers, there were no console errors, and I found the JSON schema on the test pages.  The only concern I have is with what everyone else has said about the unit test failure.  Once that is working I'm okay with approving.

fullcolorcoder:
@amauch-adobe should be good now

amauch-adobe (commented):
Approved.

hparra:
We changed test. There is no indication that `getMetadata` normalizes casing and I think metadata block lives elsewhere (franklin?). Are we sure if someone uses lowercase "organization" in metadata block that switch case with catch? I understand this problem exists for the other cases as well.

PR #564, opened 2023-03-15
MWPW-127736 Fix Safari sign-in dropdown

meganthecoder:
* Fix issue where clicking "Sign in" on safari on mobile device didn't show dropdown
* Fix issue where "Sign in" button is inaccessible on Safari 14 mobile
* Fix issue where "Sign in" dropdown and toggle/hamburger menu can be open at the same time

Resolves: [MWPW-127736](https://jira.corp.adobe.com/browse/MWPW-127736)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/methomas/checkmark?martech=off
- After: https://methomas-sign-in-mobile--milo--adobecom.hlx.page/drafts/methomas/checkmark?martech=off

BACOM:
- Before: https://main--bacom--adobecom.hlx.page/customer-success-stories?martech=off
- After: https://main--bacom--adobecom.hlx.page/customer-success-stories?milolibs=methomas-sign-in-mobile&martech=off

Note: This only occurs on iOS, so you'll need to use browserstack to verify.

PR #538, opened 2023-03-07
MWPW-127324 - Second line copyright

meganthecoder:
* Support a second line of copyright text, such as in the bacom korean footer
* Fix stylelint errors in footer.css

Resolves: [MWPW-127324](https://jira.corp.adobe.com/browse/MWPW-127324)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/methomas/footer-kr-page?martech=off
- After: https://methomas-kr-copyright--milo--adobecom.hlx.page/drafts/methomas/footer-kr-page?martech=off

PR #515, opened 2023-02-28
MWPW-126915 Fix gnav link order

meganthecoder:
* Fixes gnav so that the cta remains in the order that it's authored
* For example, the telephone number can now be added to the right of the cta

Resolves: [MWPW-126915](https://jira.corp.adobe.com/browse/MWPW-126915)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/methomas/gnav-phone-right?martech=off
- After: https://methomas-gnav-phone--milo--adobecom.hlx.page/drafts/methomas/gnav-phone-right?martech=off

Regression (gnav with phone authored to the left of the cta still works):
- After: https://methomas-gnav-phone--milo--adobecom.hlx.page/drafts/methomas/checkmark?martech=off

PR #601, opened 2023-03-29
Bring in Global Navigation progress into 'main' branch

overmyheadandbody:
As discussed recently, this is the initial FEDS/Gnav PR into the `main` Milo branch. The goal is to keep all functionality as-is on production with regards to the Milo nav until our code is fully ready to be consumed.

This is to say - we don't think our code is production-ready, we still have critical features to implement, unit tests to write and a lot of TODO comments for things we need to revisit as we move forward with our development. However, merging to `main` does mean our code will be in production, so why are we opening this PR:
1. we want to get early feedback on the progress we've made so far and address areas of concern. This will also mean it will be easier for upcoming features to be reviewed by the extended Milo team, since future PRs will be limited to just one feature;
2. our code lives under a different path. The Milo nav uses `gnav`, we've made our changes under `global-navigation`. This means that the current Milo nav consumers will get the same experience as they do today. As soon as we're done with development, we can either switch everyone to the new path via a code change or let consumers decide when they're ready to make the transition by asking them to change one metadata entry.

We recently had a status update meeting where we summarised all of the changes we're bringing in with our PR. You can have a quick look over [a few slides with details](https://adobe.sharepoint.com/:p:/r/sites/federated_services/Shared%20Documents/Presentations/Milo%20migration/FEDS%20on%20Milo%20-%20March%202023%20status%20update.pptx?d=we797423dd5444e0b83a6f6e7921bb240&csf=1&web=1) or the entire [1 hour demo & discussion](https://adobe-my.sharepoint.com/:v:/p/narcis/EYAmNWTWEg9PuVpTrp81NNcBlxXHmgsCy01Az0dDJjOk9w).

As a final note - the code has been reviewed by the Eurovision team, nothing has been merged to this feature branch without peer approval. You can have a look over the PRs [here](https://github.com/adobecom/milo/pulls?q=is%3Apr+is%3Aclosed+base%3Afeds-gnav+base%3Agnav).

Resolves: [MWPW-128666](https://jira.corp.adobe.com/browse/MWPW-128666), epic with all tasks is [MWPW-115149](https://jira.corp.adobe.com/browse/MWPW-115149).

**Test URLs:**
- After: https://gnav--milo--adobecom.hlx.page/drafts/ramuntea/gnav-refactor?martech=off


libs/blocks/global-footer/footer.js
```diff
@@ -0,0 +1,334 @@
+import {
```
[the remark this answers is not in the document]
meganthecoder: When did you pull this code? We've made a few recent updates to the footer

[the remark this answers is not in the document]
narcis-radu: The feature branch was created a couple of months ago. We are working on rewriting the footer to support all use-cases and will make sure the recent updates are still working as expected. @meganthecoder - is there anything in particular we should look at?

libs/blocks/global-footer/footer.js
```diff
@@ -0,0 +1,334 @@
...
+  return { html: respText };
+}
+
+export default async function init(block) {
+  const { prefix } = locale;
+  const url = getMetadata('footer-source') || `${prefix}/footer`;
+  if (url) {
+    const { html, error } = await fetchFooter(url);
+    if (error) {
+      console.log(`Could not create footer: ${error}`);
```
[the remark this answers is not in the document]
mokimo: Yep agree, it's on my mental todo list. Let's tackle this in a separate PR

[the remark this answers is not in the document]
overmyheadandbody: For the time being, I'll just update the footer code to match the one from Milo until we have our refactoring in place. This being said, even the updated version has some console logs, I'll make sure to remove them during refactoring.

As @mokimo mentioned, as soon as we're done implementing MVP features, we should go through the code and add LANA logs. Maybe it's something that can be paired with unit tests.

libs/blocks/global-footer/footer.js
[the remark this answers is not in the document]
overmyheadandbody: I'll update the footer code to match the Milo version until we're done with our refactoring.

libs/blocks/global-navigation/blocks/navDropdown/dropdown.js
```diff
@@ -0,0 +1,275 @@
...
+
+    const openPopup = document.querySelector('.feds-popup-headline[aria-expanded = "true"]');
+
+    if (openPopup && openPopup !== headline) {
+      openPopup.setAttribute('aria-expanded', 'false');
+    }
+
+    const currentState = headline.getAttribute('aria-expanded');
+
+    if (currentState === 'false') {
```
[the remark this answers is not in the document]
overmyheadandbody: Yup, I got carried away with being too explicit :) I'll make the change.

libs/blocks/global-navigation/blocks/navDropdown/dropdown.js
```diff
@@ -0,0 +1,275 @@
...
+        </div>`;
+    }
+
+    // If the promo is set to 'image-only',
+    // replace the whole promo content with the decorated image
+    if (isImageOnly) {
+      elem.replaceChildren(promoImageElem);
+    } else {
+    // Otherwise, just replace the image container with the decorated image
+      imageWrapper.replaceWith(promoImageElem);
```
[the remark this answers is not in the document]
overmyheadandbody: I'll make the change to check for the element. However, I can't move the declaration of `imageWrapper`, since we need a reference to its initial state. Until the logic reaches this point, it will have been moved to another parent.

libs/blocks/global-navigation/blocks/navMenu/menu.js
```diff
@@ -0,0 +1,139 @@
+import {
```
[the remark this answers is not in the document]
overmyheadandbody: I will remove it, it was just a copy-paste of the initial Milo code.

libs/blocks/global-navigation/blocks/profile/dropdown.js
```diff
@@ -0,0 +1,188 @@
+import { getConfig } from '../../../../utils/utils.js';
+import { toFragment, getFedsPlaceholderConfig } from '../../utilities/utilities.js';
+import { replaceKeyArray } from '../../../../features/placeholders.js';
+
+const getLanguage = (ietfLocale) => {
+  if (!ietfLocale.length) return 'en';
+
+  const nonStandardLocaleMap = { 'no-NO': 'nb' };
```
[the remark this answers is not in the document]
overmyheadandbody: Comments suggest that overrides be kept in their respective block, which we're already doing here, since this is part of the Profile block.

libs/blocks/global-navigation/blocks/profile/dropdown.js
```diff
@@ -0,0 +1,188 @@
...
+    this.buttonElem.addEventListener('click', () => {
+      this.toggleDropdown();
+    });
+  }
+
+  toggleDropdown() {
+    const isOpen = this.buttonElem.getAttribute('aria-expanded') === 'true';
+
+    if (isOpen) {
+      this.buttonElem.setAttribute('aria-expanded', 'false');
```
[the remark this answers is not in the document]
overmyheadandbody: Will do

libs/blocks/global-navigation/blocks/search/gnav-search.js
```diff
@@ -0,0 +1,284 @@
...
+
+  getSuggestions(query = this.query) {
+    const { env } = getConfig();
+    const subdomain = env === 'prod' ? 'adobesearch' : 'adobesearch-stage';
+    const api = `https://${subdomain}.adobe.io/autocomplete/completions?q[locale]=${getLocale()}&scope=${CONFIG.suggestions.scope}&q[text]=${encodeURIComponent(query)}`;
+
+    return window.fetch(api, { headers: { 'x-api-key': CONFIG.suggestions.apiKey } })
+      .then((data) => data.json())
+      .catch(() => {
+        // do nothing
```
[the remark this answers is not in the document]
overmyheadandbody: I think I'd defer this to a bit later, maybe we can group it with unit tests, since that will be the time when we go over all of the possible errors.

auniverseaway:
I'm seeing intermittent TBT coming from global-navigation.js and global-footer.js. One of the main goals of this refactor was to eliminate all TBT coming from gnav and footer.

https://pagespeed.web.dev/analysis/https-gnav--milo--adobecom-hlx-page-drafts-ramuntea-gnav-refactor/nbrfk1r1zh?form_factor=mobile

Screen Shot 2023-03-31 at 12 31 35 PM

libs/blocks/global-navigation/utilities/getUserEntitlements.js
```diff
@@ -0,0 +1,89 @@
...
+ */
+const getUserEntitlements = async ({ params, locale } = {}) => {
+  if (!window.adobeIMS?.isSignedInUser()) return Promise.reject(new Error('User not signed in'));
+
+  const queryParams = getQueryParameters(params);
+  if (entitlements[queryParams]) return entitlements[queryParams];
+
+  let resolve;
+  let reject;
+  entitlements[queryParams] = entitlements[queryParams]
```
yesil, line 76: could you please format entitlements as follows similar to Dexter one?

```
{
    "clouds": {},
    "arrangment_codes": {},
    "fulfilled_codes": {},
    "offer_families": {},
    "list": {
        "fulfilled_codes": []
    }
}
```

libs/blocks/global-navigation/utilities/getUserEntitlements.js
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * @description Return the JIL User Entitlements
+ * @param {object} object required params
+ * @param {array} object.params array of name value query parameters [{name: 'Q', value: 'PARAM'}]
+ * @param {object} object.locale {country: 'CH', language: 'de'}
+ * @returns {object} JIL Entitlements
+ */
+const getUserEntitlements = async ({ params, locale } = {}) => {
+  if (!window.adobeIMS?.isSignedInUser()) return Promise.reject(new Error('User not signed in'));
```
yesil, line 69: don't reject and return empty response

libs/blocks/global-navigation/utilities/getUserEntitlements.js
mokimo: PR: https://github.com/adobecom/milo/pull/613

libs/blocks/global-navigation/utilities/getUserEntitlements.js
mokimo: PR: https://github.com/adobecom/milo/pull/613

overmyheadandbody:
TL;DR: on the `gnav` branch we’re seeing better overall TBT for code that has more functionality, so I’d say this is definitely an improvement. The long task time does not impact final TBT values and seems to be a side-effect of how the LH algorithm makes certain timing assumptions.

---

Hi @auniverseaway! Thanks for looking into this! We spent some time looking into the TBT reports, here are our findings so far:
* from our tests, you definitely got the exception - we never saw a task longer than `75ms`. Even then, the actual TBT value was between `0-20ms`. Here’s a [report](https://pagespeed.web.dev/analysis/https-gnav--milo--adobecom-hlx-page-drafts-ramuntea-gnav-refactor/h92ai6flq9?form_factor=mobile). The long task from _global-navigation.js_ is `54ms`, but the TBT value is `0`. Looking at the current Milo nav, we get similar values, but the TBT value is usually higher. A report of this as well [here](https://pagespeed.web.dev/analysis/https-main--milo--adobecom-hlx-page-drafts-ramuntea-prev-gnav-refactor/21zmf6m5fd?form_factor=mobile). The long task from _gnav.js_ is `61ms`, but TBT is `10ms`. The long task duration constantly changes between runs, but TBT is lower on the `gnav` branch more often than not;
* the previous reports, however, have an issue: the _gnav.js_ logic doesn’t know how to interpret new authoring capabilities added as part of _global-navigation.js_, thus it renders much less content in the nav. So for a fair comparison, I created two new pages that end up rendering the exact same navigation content. [Here](https://pagespeed.web.dev/analysis/https-gnav--milo--overmyheadandbody-hlx-page-drafts-ramuntea-limited-gnav/xsp6dr2ksh?form_factor=mobile) are the results for the `gnav` branch. Long task is still `54ms`, TBT is still `0`. [Reports](https://pagespeed.web.dev/analysis/https-main--milo--overmyheadandbody-hlx-page-drafts-ramuntea-prev-limited-gnav/kprb4gsdkx?form_factor=mobile) from the Milo nav show that the long task from _gnav.js_ is `80ms` and the TBT value is `40ms`. Again, long task values may differ, but the final TBT value is lower on the `gnav` branch more often than not;
* why the long task, then? @mokimo has a hypothesis - we mentioned in the past that we initially load just what is needed and the rest after 3 seconds or after the user interacts with the navigation. It would seem that there need to be 5 seconds of ‘quiet window’ time before final results are computed. Which we don’t currently get: minimal nav loads, 3 seconds pass, the rest of the Gnav logic is executed and counts towards the total value for the long task. We tested this hypothesis by creating another branch where the 3 second delay is replaced with a 10 seconds one. Here are the [results](https://pagespeed.web.dev/analysis/https-delay--milo--mokimo-hlx-page-drafts-ramuntea-gnav-refactor/5aic6vioqu?form_factor=mobile). As you can see, no more long task coming from _global-navigation.js_. However, we wouldn’t want to increase the delay just to get a better long task value that ultimately doesn’t affect the LH TBT result;
* regarding the footer - we haven’t made any changes to it compared to the Milo version, it’s an exact copy-paste. We’ve started work on it and we’re thinking of loading it either after some time (3 seconds seems to be our go-to) or based on an intersection observer.

We’ll continue to look into this, but it doesn’t look like a cause for concern at the moment.

libs/blocks/global-footer/footer.js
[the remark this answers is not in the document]
meganthecoder: #486 , #507 , #509 , #538 

Also, other bugs related to gnav: #515 , #564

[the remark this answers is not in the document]
overmyheadandbody: Thanks, @meganthecoder! We actually weren't aware of the multiple-line copyright for Korea, we'll keep that in mind during our refactoring.

narcis-radu (approved):

mokimo (approved):
We've individually reviewed small PRs leading to this big one. 
We are aware there are still some gaps in this feds-milo-gnav implementation, this is expected (for now!). We didn't cover this with unit tests just yet, but will do so in a follow up PR.

MO this PR/feds-milo-gnav implementation is in a good state and a great starting place for teams to experiment pulling our version in, while we do the finishing in the meantime.

mokimo:
@chrischrischris is there still anything we need to address before getting this first big PR merged?

Ideally this doesn't change any existing functionality that currently exists since 99% of the work has been done in a separate block. 

To use this code vs the old gnav, one would simply have to adapt the metadata of a document to use our global-navigation block. Which we don't recommend just yet, while we still deliver the last features and clean everything up and address small things here and there

nerrdo:
If all concerns are addressed, can this be merged ASAP? We're facing deadline concerns from our PMs.

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -0,0 +1,80 @@
...
+
+  return (...args) => {
+    clearTimeout(timeout);
+    timeout = setTimeout(() => callback(...args), time);
+  };
+}
+
+// TODO: should we replace these with proper Locale/country service(s)?
+export function getLocale() {
+  return document.documentElement.getAttribute('lang') || 'en-US';
```
vhargrave, line 34: you can just get the locale from the config . You don't need this method

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export function getCountry() {
```
vhargrave, line 37: not really sure we need this method either since it's only being used in one place (getting the helpx link)

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -0,0 +1,80 @@
...
+
+  Array.prototype.map.call(fragment.querySelectorAll('elem'), (replaceable) => {
+    const ref = replaceable.getAttribute('ref');
+    replaceable.replaceWith(values[ref]);
+  });
+
+  return fragment;
+}
+
+export function debounceCallback(callback, time = 200) {
```
vhargrave, line 21: do this have to be a util if it's only being used in one place?

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -0,0 +1,80 @@
+import { getConfig, localizeLink } from '../../../utils/utils.js';
+
+export function toFragment(htmlStrings, ...values) {
```
vhargrave, line 3: this is cool!

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -0,0 +1,80 @@
...
+
+// TODO this is just prototyped
+export const getFedsPlaceholderConfig = () => {
+  const { locale, miloLibs } = getConfig();
+  let libOrigin = 'https://adobe.com/libs';
+  if (window.location.origin.includes('localhost')) {
+    libOrigin = `${window.location.origin}/libs`;
+  }
+  if (window.location.origin.includes('.hlx.')) {
+    libOrigin = miloLibs?.replace('hlx.live', 'hlx.page') || 'https://main--milo--adobecom.hlx.page/libs';
```
vhargrave, line 49: we only want stage pages and never live pages?

libs/blocks/global-navigation/utilities/getUserEventHistory.js
```diff
@@ -0,0 +1,30 @@
...
+  let reject;
+  userEventHistory = userEventHistory || new Promise((_resolve, _reject) => {
+    resolve = _resolve;
+    reject = _reject;
+  });
+  const profile = await window.adobeIMS.getProfile();
+  const apiUrl = getConfig().env.name === 'prod'
+    ? 'https://www.adobe.com/gating/api/realtime-profile'
+    : 'https://www.stage.adobe.com/gating/api/realtime-profile';
+  const res = await window
```
vhargrave, line 18: can't we just call fetch() ?

vhargrave (changes requested):
ok, so this was a lot to look at, but I think I was able to start scratching the surface of the PR. I think it's great that you've built so much already, and it seems to be already be working quite well 🙌 
I do have a couple of general comments: 
1. It looks like you've made your own version of the footer, but haven't attached a link to test. Can we do that?
2. This is a very large PR. It's honestly too big and very hard to review. Can we aim for smaller and more iterative PRs for the future, as I feel that my review quality is not as high as I'd like it to otherwise be. 
3. I don't know what I'm supposed to test. In the jira task that's attached it just says that gnav is integrated with milo, but what does that mean? Until I know what I'm testing it's very hard for me to approve a PR 
4. Why is so much content in a separate global-navigation folder? This feels like a separate beast that's going to grow within milo instead of it being properly integrated within milo somehow. Can't we just have blocks within the blocks folder and utils if necessary also at the right place? 
5. There are quite a lot of open TODOs. I understand that you guys want to get your code merged so that milo has at least some of the current state of the new gnav, but usually that would be a reason to "deny" a PR imo. 
6. No unit tests :( - our real code coverage is going to drop drastically if we merge this. 

Personally I'm not sure what to do. I don't want you guys to be blocked, but things like unit tests are just things that shouln't be added later, because otherwise it just doesn't happen.

libs/blocks/global-footer/global-footer.css
```diff
@@ -0,0 +1,399 @@
+footer {
+  font-size: var(--type-body-xxs-size);
+  padding: 1rem 0;
```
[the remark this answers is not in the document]
overmyheadandbody: At the moment the footer is a 1:1 copy/paste of the current footer used in Milo, https://github.com/adobecom/milo/blob/main/libs/blocks/footer/footer.js.

We're currently working on refactoring the footer and it will come in another PR. For the time being, we just wanted the Milo footer to work with our refactored Gnav.

So I will mark the footer-related comments as resolved, since this is already in production and we're re-writing it as we speak.

libs/blocks/global-navigation/blocks/appLauncher/appLauncher.js
```diff
@@ -0,0 +1,64 @@
+import { createTag, localizeLink } from '../../../../utils/utils.js';
+
+const WAFFLE_ICON = '<svg xmlns="http://www.w3.org/2000/svg"  viewBox="0 0 36 36" focusable="false" aria-hidden="true" role="img" class="spectrum-Icon spectrum-Icon--sizeS"><path d="M10 10H2V3a1 1 0 0 1 1-1h7zm4-8h8v8h-8zm20 8h-8V2h7a1 1 0 0 1 1 1zM2 14h8v8H2zm12 0h8v8h-8zm12 0h8v8h-8zM10 34H3a1 1 0 0 1-1-1v-7h8zm4-8h8v8h-8zm19 8h-7v-8h8v7a1 1 0 0 1-1 1z"></path></svg>';
```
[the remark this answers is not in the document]
overmyheadandbody: App Launcher is in the same situation as the footer, it is an exact copy/paste of the Milo version, https://github.com/adobecom/milo/blob/main/libs/blocks/gnav/gnav-appLauncher.js.

There is a mismatch between what Milo is doing and what the business wants, we're still discussing the preferred approach with stakeholders. Until we have clarity, we'll not interfere much with this.

libs/blocks/global-navigation/utilities/getUserEventHistory.js
mokimo: Yup, we can change that ... @overmyheadandbody do you want to batch that with other smaller improvements for this PR?

libs/blocks/global-navigation/blocks/navDropdown/dropdown.js
```diff
@@ -0,0 +1,270 @@
...
+      ${imageElem}
+      ${contentElem}
+    </a>`;
+
+  return linkGroup;
+};
+
+// Decorate a list of simple links
+const decorateLinkList = ({ list, className = 'feds-navLink', includeCta = true } = {}) => {
+  list.querySelectorAll('a').forEach((link, index) => {
```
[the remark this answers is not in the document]
overmyheadandbody: Unfortunately the `decorateButtons` method adds a lot of classes and markup that aren't useful for our use-cases, so we'd rather keep it contained to what we actually need instead of re-using a method and then manipulating the result yet again.

libs/blocks/global-navigation/utilities/utilities.js
mokimo: Good catch, we should change that to properly include stage for stage and live for live - otherwise that might confuse authors that change placeholders and the wrong sheet is fetched.

@overmyheadandbody do you want to batch that with other smaller improvements for this PR?

libs/blocks/global-navigation/blocks/profile/button.js
```diff
@@ -0,0 +1,26 @@
+import { toFragment, getFedsPlaceholderConfig } from '../../utilities/utilities.js';
+import { replaceKey } from '../../../../features/placeholders.js';
+
+const decorateButton = async ({ avatar }) => {
```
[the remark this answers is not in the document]
overmyheadandbody: I'm not sure I follow. This is used in https://github.com/adobecom/milo/blob/gnav/libs/blocks/global-navigation/global-navigation.js#L213.

For the Profile block, we aim to serve only the minimum required logic. If the user is not logged in, we just load `blocks/profile/signIn.js` which contains logic for signed out users. Nothing else (such as avatar, profile dropdown, etc) is needed, so there's no point in loading it.

If the user is signed in, we skip loading `blocks/profile/signIn.js` and instead load this file, `blocks/profile/button.js` which just decorates the avatar that is visible in the top navigation. However, the Profile dropdown is not loaded instantly, we load it with a delay so we give the CPU time to handle more important things. After the delay, we proceed to loading `blocks/profile/dropdown.js` that handles rendering of the actual Profile dropdown.

So yes, they need to be in different files so we can load the bare minimum depending on the scenario.

libs/blocks/global-navigation/blocks/profile/button.js
```diff
@@ -0,0 +1,26 @@
...
+  const label = await replaceKey(
```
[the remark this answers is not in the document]
overmyheadandbody: You mean the whole file? `Profile` is the block added in the Word document, which then loads different pieces of logic depending on the use-case.

libs/blocks/global-navigation/blocks/profile/dropdown.js
```diff
@@ -0,0 +1,183 @@
...
+    });
+  }
+
+  toggleDropdown() {
+    const currentState = this.buttonElem.getAttribute('aria-expanded');
+    this.buttonElem.setAttribute('aria-expanded', currentState === 'false');
+  }
+}
+
+export default ProfileDropdown;
```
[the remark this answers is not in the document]
overmyheadandbody: Same comment as before, the `Profile` block loads either `signIn.js` or `button.js` and then, after a delay, `dropdown.js`. It is used in the main global-navigation file - https://github.com/adobecom/milo/blob/gnav/libs/blocks/global-navigation/global-navigation.js#L224

overmyheadandbody:
TL;DR - it was agreed with the Milo team that we would open this PR as-is to avoid an even larger monolithic PR. Merging this will have no effect on production, as our code resides in different folders. Consumers will be switched over when ready. We aim to address TODOs and unit tests by mid-May.

---

First of all, thanks a lot for looking at this! We're aware it's a tedious task and, as discussed with the Milo team a few weeks back, the aim is to get this merged so we can then proceed with opening PRs focused on just one feature.

I'll try to address your concerns:
- the footer is a copy/paste of the Milo version, currently used in production. We're currently refactoring it and will come back with another PR for it next week or so. Heads up - it's also a big change, hopefully we get this PR merged by then so they don't stack up;
- you're right with the JIRA item. I've added the link to our story where we track time for the effort to actually get this PR merged. I've updated the description with the actual epic, https://jira.corp.adobe.com/browse/MWPW-115149. A bit of a warning, though, the epic is as tedious to go through as this PR 🙂 The gist of it is that this should mimic the AEM version of the nav, but I get that this isn't enough. Rest assured that our QEs will be on top of testing the features/functionality, both manually and with automated tests. In fact, they've already started doing this;
- the aim with having a separate _global-navigation_ folder is to allow for the two versions to co-exist. And this is a tremendous help for this exact situation: we can merge our code as-is, even though we're still building a lot of features, but it won't be used on production. Instead of force-pushing our changes to all consumers, we're enabling them to test our version by updating a simple metadata tag and choose when they're ready to use our implementation. In the meantime, they can continue using the Milo navigation. This also gives us the chance to fix any issues reported by the consumers in case anything was missed. Ultimately, the current navigation logic from Milo will be deleted and just our version will remain;
- continuing the ideas from above regarding the open TODOs, we're aware this is work in progress, we still have a lot of work to do before using this on production. However, even if merged, it won't actually be used anywhere until we're ready to switch consumers over. Waiting to address all the current TODOs will delay this initial PR even more and will also considerably increase its size. Our deadline for MVP is mid-May, we aim to address all of the TODOs by then;
- same as the TODOs, we aim to add unit tests in future PRs by mid-May, when our MVP release is scheduled. Waiting for those would further push this PR out and increase its size. There are _some_ unit tests for the Milo `gnav` at the moment, but not such a considerable amount that merging this PR would drop numbers significantly.

libs/blocks/global-navigation/utilities/utilities.js
mokimo: Moved to search for now

libs/blocks/global-footer/global-footer.js
```diff
@@ -0,0 +1,358 @@
...
+  return { html: respText };
+}
+
+export default async function init(block) {
+  const { prefix } = locale;
+  const url = getMetadata('footer-source') || `${prefix}/footer`;
+  if (url) {
+    const { html, error } = await fetchFooter(url);
+    if (error) {
+      console.log(`Could not create footer: ${error}`);
```
honstar, line 344: Instead of logging to the `console`, have you considered using lana?

libs/blocks/global-footer/global-footer.js
```diff
@@ -0,0 +1,358 @@
...
+      });
+      title.addEventListener('blur', () => {
+        window.removeEventListener('keydown', this.toggleOnKey);
+      });
+    });
+  };
+
+  onMediaChange = (e) => {
+    const footerTitles = document.querySelectorAll('.footer-nav-item-title');
+    if (e.matches) {
```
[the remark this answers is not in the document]
overmyheadandbody: Hi @honstar! Thanks for having a look over this!

First of all, the footer logic is an exact replica of the Milo footer, currently used on production. We copied it over to our folder as-is for now to ensure we can render both the top and footer navigation in our refactored version. I'm currently working on a complete refactor of the footer, a new PR will be raised by the end of next week.

But to address you query on the Milo code, this `e.matches` actually refers to `window.matchMedia`, where the current viewport width is checked. `matches` in this context refers to whether the current viewport _matches_ a defined media query, in this case `'(min-width: 900px)'`.

libs/blocks/global-footer/global-footer.js
overmyheadandbody: Yep; again, the footer is identical to the one used in Milo today on production, we're re-writing it. Our todo list for MVP includes adding LANA logs wherever neccessary.

libs/blocks/global-navigation/blocks/profile/signIn.js
```diff
@@ -0,0 +1,53 @@
+import { toFragment, getFedsPlaceholderConfig } from '../../utilities/utilities.js';
+import { replaceKey } from '../../../../features/placeholders.js';
+
+const signIn = () => {
+  if (typeof window.adobeIMS?.signIn !== 'function') return;
```
honstar, line 5: So if ims is not loaded yet, a call to this function simply does nothing by returning? What does that mean for the end-user?

overmyheadandbody: The logic only gets to here if IMS is ready, as seen [here](https://github.com/adobecom/milo/pull/601/files/10037426414c6f9ff894179cd240d95f19ec3519#diff-628bf5a25903978c5b36da4943ec813da171346d24bd64f80907f012a840b098R165).

When IMS executes the `onReady` callback, we initiate the profile decoration logic. This condition is an extra safeguard in the off chance that `signIn` is not a function.

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -0,0 +1,80 @@
...
+  }
+  return {
+    locale: {
+      ...locale,
+      contentRoot: `${libOrigin}/feds${locale.prefix}`,
```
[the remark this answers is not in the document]
vhargrave: ok 🙌  - i actually am realising that I don't think that pathing is going to work at all. 🤔 
https://adobe.com/libs/ maps directly to the `libs` folder on the `main` branch of milo on github. 

For the time being I think you're better off just mapping directly to your feds folder in milo and if you get your own sharepoint folder then you can map to there. 

if you want to talk about it, then let's even have a short call.

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -0,0 +1,65 @@
...
+  return analyticsValue;
+}
+
+export function decorateCta({ elem, type = 'primaryCta', index } = {}) {
+  const modifier = type === 'secondaryCta' ? 'secondary' : 'primary';
+
+  return toFragment`
+    <div class="feds-cta-wrapper">
+      <a 
+        href="${localizeLink(elem.href)}"
```
vhargrave, line 59: do you really have to call localizeLink? 
i thought that all links are localized in the utils.js file

libs/blocks/global-footer/global-footer.js
narcis-radu: hey @honstar, here's the production code that @overmyheadandbody  mentioned https://github.com/adobecom/milo/blob/main/libs/blocks/footer/footer.js#L344. We do have plans to use lana for logging, our version of the footer is still work in progress.

libs/blocks/global-footer/global-footer.js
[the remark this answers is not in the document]
narcis-radu: Adding reference to production code - https://github.com/adobecom/milo/blob/main/libs/blocks/footer/footer.js#L319.

libs/blocks/global-navigation/utilities/getUserEventHistory.js
```diff
@@ -0,0 +1,29 @@
...
+  userEventHistory = userEventHistory || new Promise((_resolve, _reject) => {
+    resolve = _resolve;
+    reject = _reject;
+  });
+  const profile = await window.adobeIMS.getProfile();
+  const apiUrl = getConfig().env.name === 'prod'
+    ? 'https://www.adobe.com/gating/api/realtime-profile'
+    : 'https://www.stage.adobe.com/gating/api/realtime-profile';
+  const res = await fetch(`${apiUrl}?guid=${id || profile.userId}`, {
+    method: 'GET',
```
vhargrave, line 19: you don't need to pass the method to the fetch, as 'GET' is the default
I actually don't even think that you need to pass 'Content-Type' or Accept either. It should just work, so might as well save on those bytes :)

libs/blocks/global-navigation/utilities/getUserEventHistory.js
```diff
@@ -0,0 +1,29 @@
...
+    'Content-Type': 'application/json',
+    Accept: 'application/json',
+  })
+    .then((resp) => (resp.status === 200 ? resp.json() : null));
+  if (res) resolve(res);
```
vhargrave, line 24: why not put this resolve and reject stuff directly in your .then?

libs/blocks/global-navigation/blocks/navDropdown/dropdown.js
```diff
@@ -0,0 +1,270 @@
+import { getAnalyticsValue, toFragment, decorateCta } from '../../utilities/utilities.js';
```
vhargrave, line 1: can we rename the folder to dropdown or rename the files here to navDropdown.js & navDropdown.css?

narcis-radu:
hey @vhargrave - I see that @mokimo implemented your feedback https://github.com/adobecom/milo/pull/646. Can you take another look at the PR? Thanks!

vhargrave (changes requested):
ok, took another look 🙌 
right now a lot of the blocks are not consistent with how other blocks are designed. In general, it should be possible for an author to add a block in word and for the block to then be independently rendered on the page. 
For blocks like the dropdown that is currently not the case. All blocks should also follow a pattern, that their default method accepts an element, which they then transform and appropriately style. In my opinion these "blocks" must either be adjusted or we need to stop naming them blocks.
Furthermore, can we please keep the folder names the same as the blocks they contain for my OCD? It bugs me seeing a block folder named "search" but the files are named "gnav-search" 😄

vhargrave:
> hey @vhargrave - I see that @mokimo implemented your feedback #646. Can you take another look at the PR? Thanks!

@narcis-radu yes sir! just had to find some time again 🙌

libs/blocks/global-navigation/blocks/navDropdown/dropdown.js
overmyheadandbody: In my footer refactor work, I've moved the dropdown logic under a `utilities` folder and renamed it as 'menu'. This is because the footer uses the same logic to generate the menu, so it needed to be a bit more generic. Could this wait until we open the full footer PR?
dropdown renamed to menu

narcis-radu:
@vhargrave - regarding this:
> I don't fully understand - why not just use the milo footer if this is a 1:1 copy?
@auniverseaway expressed his desire to have these new features available as soon as possible in the main branch and allow consumers to switch between the navigation blocks (`gnav` and `global-navigation`) and test. We agreed to raise a PR as soon as possible with what we have and then create smaller PRs with remaining functionality. The footer is still work in progress, the plan is to update `global-footer` with custom code, but so far it's a copy of the existing block. We could have used the existing block, but wanted to make sure things don't break when consumers test things around.

vhargrave:
> 

@narcis-radu ok got it - makes sense thx!

libs/blocks/global-navigation/utilities/utilities.js
overmyheadandbody: The content for the global-navigation comes through another request, thus goes through another pipeline.

Assuming my page URL is https://www.adobe.com/testPath and I add a link with this `href` on the page, then keeping the call to `localizeLink` means the final `href` value will be `.testPath`. Removing the call to `localizeLink` would mean that the final `href` value will be `https://www.adobe.com/testPath`. What would be the desired effect between the two?

Keeping the call to `localizeLink` means that

libs/blocks/global-navigation/blocks/profile/button.js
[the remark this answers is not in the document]
overmyheadandbody: Oh an another reason for having this - the Profile will need to be exported as its own standalone module to be used by non-Milo consumers, so we want to separate its logic as much as possible to make it easier to make that transition post-MVP.

libs/blocks/global-navigation/blocks/navDropdown/dropdown.js
vhargrave: yeah - sure 👍

libs/blocks/global-navigation/utilities/utilities.js
vhargrave: why not have the content go through the normal pipeline though? That way there's much less chance of forgetting to decorate stuff properly ...japanese word wrap, link localization, etc. etc. 
You'd call the normal pipeline by calling loadArea on your loaded html

libs/blocks/global-navigation/utilities/getUserEntitlements.js
```diff
@@ -0,0 +1,162 @@
...
+  languages = [
+    `${locale.language}-${locale.country.toUpperCase()}`,
+    `${locale.language};q=0.9`,
+    'en;q=0.8',
+  ];
+  return languages;
+};
+
+const getQueryParameters = (params) => {
+  let result = '';
```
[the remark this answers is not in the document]
mokimo: https://github.com/adobecom/milo/pull/651

libs/blocks/global-navigation/utilities/getUserEntitlements.js
```diff
@@ -0,0 +1,162 @@
...
+      offer_families[family.toLowerCase()] = true;
+    }
+
+    if (offer.product_arrangement_code) {
+      arrangment_codes[offer.product_arrangement_code] = true;
+    }
+
+    if (Array.isArray(fulfilled_items)) {
+      fulfilled_items.forEach(({ code }) => {
+        if (code) {
```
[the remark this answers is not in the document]
mokimo: https://github.com/adobecom/milo/pull/651

libs/blocks/global-navigation/utilities/getUserEntitlements.js
```diff
@@ -0,0 +1,162 @@
...
+const entitlements = {};
+
+const getAcceptLanguage = (locale) => {
+  let languages = [];
+  if (!locale || Object.keys(locale).length === 0) {
+    return languages;
+  }
+
+  languages = [
+    `${locale.language}-${locale.country.toUpperCase()}`,
```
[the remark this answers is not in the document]
mokimo: It's an AEM relict, I added a todo in https://github.com/adobecom/milo/pull/651 to address this outside of the scope of this PR - so consumers actually see those changes, here they'd go under

libs/blocks/global-navigation/utilities/getUserEventHistory.js
mokimo: https://github.com/adobecom/milo/pull/651

honstar (approved):

auniverseaway:
> TL;DR: on the `gnav` branch we’re seeing better overall TBT for code that has more functionality, so I’d say this is definitely an improvement. The long task time does not impact final TBT values and seems to be a side-effect of how the LH algorithm makes certain timing assumptions.
> 
> Hi @auniverseaway! Thanks for looking into this! We spent some time looking into the TBT reports, here are our findings so far:
> 
> * from our tests, you definitely got the exception - we never saw a task longer than `75ms`. Even then, the actual TBT value was between `0-20ms`. Here’s a [report](https://pagespeed.web.dev/analysis/https-gnav--milo--adobecom-hlx-page-drafts-ramuntea-gnav-refactor/h92ai6flq9?form_factor=mobile). The long task from _global-navigation.js_ is `54ms`, but the TBT value is `0`. Looking at the current Milo nav, we get similar values, but the TBT value is usually higher. A report of this as well [here](https://pagespeed.web.dev/analysis/https-main--milo--adobecom-hlx-page-drafts-ramuntea-prev-gnav-refactor/21zmf6m5fd?form_factor=mobile). The long task from _gnav.js_ is `61ms`, but TBT is `10ms`. The long task duration constantly changes between runs, but TBT is lower on the `gnav` branch more often than not;
> * the previous reports, however, have an issue: the _gnav.js_ logic doesn’t know how to interpret new authoring capabilities added as part of _global-navigation.js_, thus it renders much less content in the nav. So for a fair comparison, I created two new pages that end up rendering the exact same navigation content. [Here](https://pagespeed.web.dev/analysis/https-gnav--milo--overmyheadandbody-hlx-page-drafts-ramuntea-limited-gnav/xsp6dr2ksh?form_factor=mobile) are the results for the `gnav` branch. Long task is still `54ms`, TBT is still `0`. [Reports](https://pagespeed.web.dev/analysis/https-main--milo--overmyheadandbody-hlx-page-drafts-ramuntea-prev-limited-gnav/kprb4gsdkx?form_factor=mobile) from the Milo nav show that the long task from _gnav.js_ is `80ms` and the TBT value is `40ms`. Again, long task values may differ, but the final TBT value is lower on the `gnav` branch more often than not;
> * why the long task, then? @mokimo has a hypothesis - we mentioned in the past that we initially load just what is needed and the rest after 3 seconds or after the user interacts with the navigation. It would seem that there need to be 5 seconds of ‘quiet window’ time before final results are computed. Which we don’t currently get: minimal nav loads, 3 seconds pass, the rest of the Gnav logic is executed and counts towards the total value for the long task. We tested this hypothesis by creating another branch where the 3 second delay is replaced with a 10 seconds one. Here are the [results](https://pagespeed.web.dev/analysis/https-delay--milo--mokimo-hlx-page-drafts-ramuntea-gnav-refactor/5aic6vioqu?form_factor=mobile). As you can see, no more long task coming from _global-navigation.js_. However, we wouldn’t want to increase the delay just to get a better long task value that ultimately doesn’t affect the LH TBT result;
> * regarding the footer - we haven’t made any changes to it compared to the Milo version, it’s an exact copy-paste. We’ve started work on it and we’re thinking of loading it either after some time (3 seconds seems to be our go-to) or based on an intersection observer.
> 
> We’ll continue to look into this, but it doesn’t look like a cause for concern at the moment.

Thanks for looking into this. I don't think comparisons to gnav make a ton of sense from my POV because we knew we didn't like what gnav was doing. The whole point is to be better.

This is something to keep an eye on. You get CLS and LCP numbers for free from my milo but you are now the proud owners of intermittent TBT. We have our own work to do in utils.js which can (infrequently) cause TBT.

I don't think the answer is delays, but to more strategically introduce breaks `setTimeout(doSomethingExpensive(), 1)`. For loops and promise.all can be your friend here.

I think @vhargrave has a point about the use of loadArea, but I'll let you folks duke that out. We know from experience, going your own path for decorators can be quite time consuming. In an ideal world, the fragments you pull in for your various pieces should use loadArea. I know gnav.js did some horrible things regarding looping over links more than it needed to.

overmyheadandbody:
Hi @auniverseaway! Thanks for chiming in.

I did a few quick tests integrating your suggestions for addressing TBT concerns and while they do seem to have a slight effect on my local environment, that's not the case when using PSI. The helix bot from this PR also reports good TBT. However, we can definitely look into a few improvements, @mokimo wanted to try a few things out, likely after our MVP deadline.

Regarding `loadArea`, we had a chat with @vhargrave today to go through the challenges of using this approach. To summarise, the Global Navigation is a block in Milo, but the Global Navigation is made out of its own blocks, such as Search, Gnav Promo, Link Group, etc. All these are specific to the Global Navigation and can't be used on Milo pages, nor the other way around, due to the different scopes of the two. Using `loadArea` on Global Navigation content will try to load all of those blocks as Milo blocks. We're not sure this would be the right approach, nor would we have time to integrate this and still make our deadline. For now, we're suggesting we just make a call to `decorateLinks` to handle link localisation, `target` values, etc..

Moving forward, post-MVP, we're thinking of doing some due diligence to see whether using Milo College would be a better fit for the Global Navigation, as we could have "regular" blocks and host content in a single Sharepoint location.

vhargrave (approved):
Looking good!

PR #607, opened 2023-03-30
Mwpw 128670 gnav logo is covering the page

zagi25:
The problem is that on our stage env (https://www.stage.adobe.com/acrobat/online/jpg-to-pdf.html) sometimes gnav.plain.html is downloaded before the gnav.css and that leads to raw gnav html displaying before it is styled as you can see on attached images. The best way to reproduce the problem, is to open chrome in incognito mode and check 'Disable cache' option in Network tab in dev tools and to reload the page few times.

Resolves: https://jira.corp.adobe.com/browse/MWPW-128670

logo

gnav menu

### DIFF
--- a/libs/blocks/gnav/gnav.js
+++ b/libs/blocks/gnav/gnav.js
@@ -662,7 +662,7 @@ export default async function init(header) {
   if (!html) return null;
 
   const link = document.head.querySelector(`link[href="${miloLibs}/blocks/gnav/gnav.css"]`);
-  if(!link.getAttribute('data-loaded') === 'true'){
+  if(link.getAttribute('data-loaded') !== 'true'){
     header.style.display = 'none';
     const isStyleLoaded = setInterval(() => {
       if(link.getAttribute('data-loaded') === 'true'){
--- a/libs/blocks/gnav/gnav.js
+++ b/libs/blocks/gnav/gnav.js
@@ -655,11 +655,26 @@ async function fetchGnav(url) {
 }
 
 export default async function init(header) {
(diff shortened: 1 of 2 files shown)


honstar:
FYI @narcis-radu @overmyheadandbody @mokimo

mokimo:
The way I read the [loadBlock](https://github.com/adobecom/milo/blob/main/libs/utils/utils.js#L358-L385) code, the JS/CSS always a running condition - IMO blocks shouldn't be initialized until the CSS is ready.  (For example adding `await styleLoaded` before calling `await init(block)` in the milo utils)

That doesn't seem to be a gnav specific issue though, any input @chrischrischris @auniverseaway ?

chrischrischris (changes requested):
@mokimo It is not an issue for other blocks in milo because the section the block is in is not made visible until after the css and js are loaded.

@zagi25 I think the better fix for this would be to hide the gnav and only display once loaded.  I think we can change `loadPostLCP` to await the loading of the header and making it visible once we know js and css is loaded:

```
  loadMartech(config);
  const header = document.querySelector('header');
  if (header) {
    header.style.opacity = 0;
    await loadBlock(header);
    header.style.opacity = '';
  }
  ...
```

Though we probably should use a class to control it

mokimo (approved):

mokimo:
@zagi25 make sure you add some helix based test links (before/after) so that all the checks can run through, here's an example PR  https://github.com/adobecom/milo/pull/604

honstar (approved):

PR #612, opened 2023-03-31
Remove carousel styles from typography css

rgclayton:
* Removes carousel related styles from typography.css
* Isolates all carousel styling to carousel.css
* Updated spacing from section-metadata to work better with blocks in general
* Reset containers when inside carousel slides to avoid content being blown out in some cases.

Effort to clean up LCP

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rclayton/carousel-css?martech=off
- After: https://carousel-clean-up--milo--adobecom.hlx.page/drafts/rclayton/carousel-css?martech=off

Bacom:
- Before: https://main--bacom--adobecom.hlx.live/customer-success-stories/coca-cola-case-study
- After: https://main--bacom--adobecom.hlx.live/customer-success-stories/coca-cola-case-study?milolibs=carousel-clean-up

### DIFF
--- a/libs/blocks/carousel/carousel.css
+++ b/libs/blocks/carousel/carousel.css
@@ -130,7 +130,6 @@
   transition: opacity;
 }
 
-
 .carousel-slide.reference-slide {
   order: 1;
 }
@@ -138,17 +137,6 @@
 .carousel-slide.active {
   opacity: 1;
 } 
-
-/* Corrects backgournd images coming from section metadata */
-.carousel .carousel-slide.has-background {
-  position: relative;
-}
-
-/* Rest for nested text blocks background image */
-.carousel .carousel-slide .foreground {
-  max-width: unset;
-  margin: unset;
-}
 
 .carousel-slides.is-ready,
 .carousel-slides.is-ready.is-reversing,
@@ -319,6 +307,7 @@
 
 .carousel .carousel-expand img {
   background-color: var(--carousel-expand-background-color);
+  box-sizing: content-box;
   border: 1px solid var(--carousel-expand-border-color);
   border-radius: 4px;
   padding: 2px 1px 2px 3px;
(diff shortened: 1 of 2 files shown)


libs/blocks/carousel/carousel.css
```diff
@@ -358,6 +347,54 @@
...
+
+.carousel.container .carousel-slide.l-spacing > div,
+.carousel.container .carousel-slide .l-spacing { padding: var(--spacing-l); }
+
+.carousel.container .carousel-slide.xl-spacing > div,
+.carousel.container .carousel-slide .xl-spacing { padding: var(--spacing-xl); }
+
+.carousel.container .carousel-slide.xxl-spacing > div
+.carousel.container .carousel-slide .xxl-spacing { padding: var(--spacing-xxl); }
+
```
ryanmparrish, line 397: Do we want to include the `.section.xxxl-spacing` token in this as well? 
I see that it's not in the `typography.css` file either so maybe it's OK, or we need to add it to both to cover the bases.

ryanmparrish (approved):
Looks good. One minor comment on adding style for `.section.xxxl-spacing`. 
GTG as is I think but it might be worth adding this while you're in there.

libs/blocks/carousel/carousel.css
rgclayton: Good catch!

Hmm. the part I cleaned up in `typography.css` I think relates to the Text block. Do we need to update that to include `xxxl-spacing`? Can you or @Sartxi confirm?

It does makes sense to include in `carousel.css` since that is targeting `section-metadata` spacing tokens.

seanchoi-dev (approved):

libs/blocks/carousel/carousel.css
ryanmparrish: Gotcha, Yea that `xxxl-spacing` isn't widely used but I feel like we should add it to `typography.css` as well just to cover all bases. It looks to have just been missed.

ryanmparrish: The only thing i see is we should probably add this `xxxl-spacing` to `typography.css` for any block to potentially use. 
`div[class*='-up'] .con-block.has-bg.xxxl-spacing { padding: var(--spacing-xxxl); }`

rgclayton: @ryanmparrish Added. Do you mind giving one more look over and adding the verified label is all looks good?

PR #613, opened 2023-04-03
MWPW-129129: Improve user entitlements

mokimo:
### Description
We got some feedback for https://github.com/adobecom/milo/pull/522 the default response for the getUserEntitlements utility now looks similar to what we provide in the dexter world with `feds.utilities.getUserEntitlements().then(data => console.log(data))` 

An empty response will be provided in any case of error instead of rejecting the promise.
```js
// Example output import getEntitlements from './utilities/getUserEntitlements.js';
const res = await getEntitlements({params, locale});
res = {
    "clouds": {
        "creative_cloud": false,
        "document_cloud": false,
    },
    "arrangment_codes": {
        "arrangment_codes-1": true,
        "arrangment_codes-2": true
    },
    "fulfilled_codes": {
        "fulfilled_codes-1": true,
        "fulfilled_codes-2": true,
    },
    "offer_families": {
        "offer_families_string": true
    },
    "list": {
        "fulfilled_codes": [
            "fulfilled-code-string-1",
            "fulfilled-code-string-2",
        ]
    }
}
```

A raw response can still be retrieved
```js
// Example output import getEntitlements from './utilities/getUserEntitlements.js';
// const res = await getEntitlements({params, locale, format: 'raw'});
// res = raw/unformatted output from the JIL api from https://www.adobe.com/aos-api/users/${profile.userId}/subscriptions
```
Resolves: [MWPW-129129](https://jira.corp.adobe.com/browse/MWPW-129129)

**Test URLs:**
Not required, no live changes

### DIFF
--- a/libs/blocks/global-navigation/utilities/getUserEntitlements.js
+++ b/libs/blocks/global-navigation/utilities/getUserEntitlements.js
@@ -1,6 +1,7 @@
+/* eslint-disable camelcase */
 import { getConfig } from '../../../utils/utils.js';
 
-const API_WAIT_TIMEOUT = '10000';
+const API_WAIT_TIMEOUT = 10000;
 const entitlements = {};
 
 const getAcceptLanguage = (locale) => {
@@ -36,7 +37,75 @@ const getQueryParameters = (params) => {
   return result;
 };
 
+const emptyEntitlements = () => ({
+  clouds: {},
+  arrangment_codes: {},
+  fulfilled_codes: {},
+  offer_families: {},
+  list: { fulfilled_codes: [] },
+});
+const CREATIVE_CLOUD = 'creative_cloud';
+const DOCUMENT_CLOUD = 'document_cloud';
+const EXPERIENCE_CLOUD = 'experience_cloud';
+
+/**
+ * Breaks JIL offers(subscriptions) into easily addressable flate data structure.
+ * @param {*} offers
+ */
+const mapSubscriptionCodes = (offers) => {
+  if (!Array.isArray(offers)) {
+    return emptyEntitlements();
+  }
+
+  const {
+    clouds,
+    arrangment_codes,
+    fulfilled_codes,
+    offer_families,
+    list,
+  } = emptyEntitlements();
+
+  offers.forEach(({ fulfilled_items, offer = {} }) => {
+    const cloud = offer.product_arrangement?.cloud;
+    clouds[CREATIVE_CLOUD] = clouds[CREATIVE_CLOUD] || cloud === 'CREATIVE';
+    clouds[DOCUMENT_CLOUD] = clouds[DOCUMENT_CLOUD] || cloud === 'DOCUMENT';
+    clouds[EXPERIENCE_CLOUD] = clouds[EXPERIENCE_CLOUD] || cloud === 'EXPERIENCE';
+
+    const family = offer?.product_arrangement?.family;
+    if (family) {
+      offer_families[family.toLowerCase()] = true;
+    }
+
+    if (offer.product_arrangement_code) {
+      arrangment_codes[offer.product_arrangement_code] = true;
+    }
+
+    if (Array.isArray(fulfilled_items)) {
+      fulfilled_items.forEach(({ code }) => {
+        if (code) {
+          fulfilled_codes[code] = true;
+          // Avoid duplicates
+          if (list.fulfilled_codes.indexOf(code) === -1) {
+            list.fulfilled_codes.push(code);
+          }
+        }
+      });
+    }
+  });
+
+  return {
+    clouds,
+    arrangment_codes,
+    fulfilled_codes,
+    offer_families,
+    list,
+  };
+};
+
 const getSubscriptions = async ({ queryParams, locale }) => {
+  const controller = new AbortController();
+  const { signal } = controller;
+  const timeout = setTimeout(() => controller.abort(), API_WAIT_TIMEOUT);
   const profile = await window.adobeIMS.getProfile();
   const apiUrl = getConfig().env.name === 'prod'
     ? `https://www.adobe.com/aos-api/users/${profile.userId}/subscriptions`
@@ -46,6 +115,7 @@ const getSubscriptions = async ({ queryParams, locale }) => {
       method: 'GET',
       cache: 'no-cache',
       credentials: 'same-origin',
+      signal,
       headers: {
         'Content-Type': 'application/json',
         Authorization: `Bearer ${window.adobeIMS.getAccessToken().token}`,
@@ -54,7 +124,10 @@ const getSubscriptions = async ({ queryParams, locale }) => {
         'Accept-Language': getAcceptLanguage(locale).join(','),
       },
     })
-    .then((response) => (response.status === 200 ? response.json() : null));
+    .then((response) => {
+      clearTimeout(timeout);
+      return response.status === 200 ? response.json() : null;
+    });
   return res;
 };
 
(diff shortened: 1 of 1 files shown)


mokimo:
@yesil would be lovely if you can have a look if that's in line with what we discussed / what you need.

libs/blocks/global-navigation/utilities/getUserEntitlements.js
```diff
@@ -63,27 +130,37 @@ const getSubscriptions = async ({ queryParams, locale }) => {
...
-      reject = _reject;
-    });
-  setTimeout(() => reject('Subscriptions API Call timeout'), API_WAIT_TIMEOUT);
+   || new Promise((_resolve) => { resolve = _resolve; });
+
+  setTimeout(() => resolve(emptyEntitlements()), API_WAIT_TIMEOUT);
   // TODO we might need to format data for analytics
-  const data = await getSubscriptions({ queryParams, locale });
+  const data = await getSubscriptions({ queryParams, locale })
+    .finally((res) => res || emptyEntitlements());
```
yesil, line 154: as per, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally

finally does not take any argument, `res` is not useful here.

libs/blocks/global-navigation/utilities/getUserEntitlements.js
```diff
@@ -63,27 +130,37 @@ const getSubscriptions = async ({ queryParams, locale }) => {
...
   const queryParams = getQueryParameters(params);
-  if (entitlements[queryParams]) return entitlements[queryParams];
+  if (entitlements[queryParams]) {
+    return format === 'raw'
+      ? entitlements[queryParams]
+      : entitlements[queryParams]
+        .then((res) => mapSubscriptionCodes(res));
+  }
 
   let resolve;
```
yesil, line 147: looks like concurrent calls might override resolve object.
Why not move `setTimeout` logic inside the promise and remove `resolve` variable?

libs/blocks/global-navigation/utilities/getUserEntitlements.js
mokimo: Indeed, changed that 👍

libs/blocks/global-navigation/utilities/getUserEntitlements.js
mokimo: Moved a bit into the promise itself and added an abort controller for the setTimeout, feel like that's a little cleaner

libs/blocks/global-navigation/utilities/getUserEntitlements.js
```diff
@@ -54,7 +124,10 @@ const getSubscriptions = async ({ queryParams, locale }) => {
         'Accept-Language': getAcceptLanguage(locale).join(','),
       },
     })
-    .then((response) => (response.status === 200 ? response.json() : null));
+    .then((response) => {
+      clearTimeout(timeout);
+      return response.status === 200 ? response.json() : null;
```
yesil, line 129: if the status is not 200, will `null` bill replaced with the `emptyEntitlements`?

yesil (commented):
Unit tests for `getUserEntitlements.js` missing. Given that 100% test coverage is needed on Milo. I think there are multiple flows that need to be covered.

libs/blocks/global-navigation/utilities/getUserEntitlements.js
mokimo: Nope, great catch - fixed.

mokimo:
@yesil Yeah, I'm very aware the tests are missing (and I'm biting myself not just writing this test-driven to begin with) but we still got a few high priority topics to cover for the nav migration MVP and after those are done, we will start covering the whole new feds based global-navigation with tests

libs/blocks/global-navigation/utilities/getUserEntitlements.js
```diff
@@ -63,27 +136,33 @@ const getSubscriptions = async ({ queryParams, locale }) => {
...
+const getUserEntitlements = async ({ params, locale, format } = {}) => {
+  if (!window.adobeIMS?.isSignedInUser()) return Promise.resolve(emptyEntitlements());
 
   const queryParams = getQueryParameters(params);
-  if (entitlements[queryParams]) return entitlements[queryParams];
+  if (entitlements[queryParams]) {
+    return format === 'raw'
+      ? entitlements[queryParams]
+      : entitlements[queryParams]
+        .then((res) => mapSubscriptionCodes(res));
```
overmyheadandbody, line 150: Would we also need a `catch` statement here?

mokimo: I'll try to revisit this comment when I write some tests in the future so I can actually check this exact scenario.
Added it for now

libs/blocks/global-navigation/utilities/getUserEntitlements.js
```diff
@@ -36,7 +37,75 @@ const getQueryParameters = (params) => {
...
+    if (offer.product_arrangement_code) {
+      arrangment_codes[offer.product_arrangement_code] = true;
+    }
+
+    if (Array.isArray(fulfilled_items)) {
+      fulfilled_items.forEach(({ code }) => {
+        if (code) {
+          fulfilled_codes[code] = true;
+          // Avoid duplicates
+          if (list.fulfilled_codes.indexOf(code) === -1) {
```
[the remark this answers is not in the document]
mokimo: 1:1 ported that, I might slightly change it once I write tests for this so I'm sure nothing breaks by changing it right now

libs/blocks/global-navigation/utilities/getUserEntitlements.js
```diff
@@ -36,7 +37,75 @@ const getQueryParameters = (params) => {
...
+    list,
+  } = emptyEntitlements();
+
+  offers.forEach(({ fulfilled_items, offer = {} }) => {
+    const cloud = offer.product_arrangement?.cloud;
+    clouds[CREATIVE_CLOUD] = clouds[CREATIVE_CLOUD] || cloud === 'CREATIVE';
+    clouds[DOCUMENT_CLOUD] = clouds[DOCUMENT_CLOUD] || cloud === 'DOCUMENT';
+    clouds[EXPERIENCE_CLOUD] = clouds[EXPERIENCE_CLOUD] || cloud === 'EXPERIENCE';
+
+    const family = offer?.product_arrangement?.family;
```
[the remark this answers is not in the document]
mokimo: It will default to an empty object if not available, so we can safely remove it 👍

libs/blocks/global-navigation/utilities/getUserEntitlements.js
```diff
@@ -36,7 +37,75 @@ const getQueryParameters = (params) => {
   return result;
 };
 
+const emptyEntitlements = () => ({
```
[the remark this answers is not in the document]
mokimo: Nope that part was just ported from the aem version - I don't see any reason for it to stay a function - changed 👍

narcis-radu (approved):

libs/blocks/global-navigation/utilities/getUserEntitlements.js
```diff
@@ -36,7 +37,75 @@ const getQueryParameters = (params) => {
...
+    offer_families,
+    list,
+  };
+};
+
 const getSubscriptions = async ({ queryParams, locale }) => {
+  const controller = new AbortController();
+  const { signal } = controller;
+  const timeout = setTimeout(() => controller.abort(), API_WAIT_TIMEOUT);
   const profile = await window.adobeIMS.getProfile();
```
honstar, line 109: While this line is not part of this PR, I'm wondering whether we should guard against `window.adobeIMS` being null at this point?

libs/blocks/global-navigation/utilities/getUserEntitlements.js
```diff
@@ -36,7 +37,75 @@ const getQueryParameters = (params) => {
...
+    fulfilled_codes,
+    offer_families,
+    list,
+  };
+};
+
 const getSubscriptions = async ({ queryParams, locale }) => {
+  const controller = new AbortController();
+  const { signal } = controller;
+  const timeout = setTimeout(() => controller.abort(), API_WAIT_TIMEOUT);
```
honstar, line 108: Maybe it makes more sense to use `AbortSignal.timeout(API_WAIT_TIMEOUT)` instead of setting a custom timeout? See https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout

libs/blocks/global-navigation/utilities/getUserEntitlements.js
mokimo: The `getUserEntitlements` method ensures IMS is available and the user is logged in 
```js
const getUserEntitlements = async ({ params, locale, format } = {}) => {
  if (!window.adobeIMS?.isSignedInUser()) return Promise.resolve(emptyEntitlements);
  ... 
  getSubscriptions()
```
`

overmyheadandbody (approved):

PR #616, opened 2023-04-04
MWPW-126637: add support for checkout ctx params

3ch023:
Provides required context parameters for checkout url generation:

* clientId - code config (scripts.js) -> default (adobe_com)
* checkoutType - merch link -> metadata -> default (UCv3)
* workflowStep - merch link -> default (email)
* marketSegment - merch link -> default (COM)

All default values provided by tacocat library, not stored in milo.

Resolves: [MWPW-126637](https://jira.corp.adobe.com/browse/MWPW-126637)

Due to CORS, the After URL should be tested with the following CORS settings:

Access-Control-Allow-Origin: *
Access-Control-Allow-Headers: x-api-key

**Test URLs:**
- Before: [https://main--milo--adobecom.hlx.page/fragments/ilyas/merch?martech=off](https://main--milo--adobecom.hlx.page/fragments/ilyas/merch?martech=off)
- After: [https://mwpw-126637-cta-ctx-params--milo--3ch023.hlx.page/fragments/ilyas/merch?martech=off](https://mwpw-126637-cta-ctx-params--milo--3ch023.hlx.page/fragments/ilyas/merch?martech=off)


JasonHowellSlavin (approved):

libs/blocks/merch/README.MD
```diff
@@ -0,0 +1,14 @@
+# Overwriting default Commerce Checkout values:
```
vhargrave, line 1: can you add this to the wiki on github instead? :)

libs/blocks/merch/checkout-context.js
```diff
@@ -0,0 +1,30 @@
+const DEFAULT_WORKFLOW_STEP = 'email';
```
vhargrave, line 1: since both these variables are only being used in one place you don't need to make them global. It's extra bytes for no reason .

libs/blocks/merch/checkout-context.js
```diff
@@ -0,0 +1,30 @@
...
+const DEFAULT_CLIENT_ID = 'adobe_com';
```
vhargrave, line 2: why have you actually created a separate file for the checkout context? All this can be moved to the merch.js block imo.

libs/blocks/merch/checkout-context.js
```diff
@@ -0,0 +1,30 @@
...
+
+function getWorkflowStep(searchParams, checkoutContext) {
+  const workflowStep = searchParams?.get('workflowStep') ?? checkoutContext?.workflowStep ?? DEFAULT_WORKFLOW_STEP;
```
vhargrave, line 5: that's cool - didn't know you could do that with the null coalescing operator

libs/blocks/merch/checkout-context.js
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Checkout Type, Workflow Step and Market Segment can be set in the OST link.
+ * If there is no value in search params, value is taken from global config.
+ * If there is no value defined in global config, default value is used.
+ * @param {*} searchParams link level overrides for checkout parameters
+ * @returns checkout context object required to build a checkout url
+ */
+export default function getCheckoutContext(searchParams, config) {
+  const { commerce } = config;
```
vhargrave, line 17: I'm wondering if config is really the right place for this cta configuration . Won't things like the checkoutContext be decided on a page by page basis? If so, these configurations should go in the metadata, and not the project config

vhargrave (changes requested):
looking pretty good! 
found a few things - let me know what you think.

libs/blocks/merch/checkout-context.js
yesil, line 1: defaults values are already provided in tacocat. Anything we pass should be non default values.

libs/blocks/merch/checkout-context.js
yesil: good point, each file is loaded separately, better to keep the logic concise and in merch.js

test/blocks/merch/merch.test.js
```diff
@@ -154,15 +154,39 @@ describe('Merch Block', () => {
...
       expect(dataset.template).to.equal('checkoutUrl');
       expect(dataset.promotionCode).to.equal(undefined);
+      expect(dataset.checkoutWorkflow).to.equal(undefined);
+      expect(dataset.checkoutWorkflowStep).to.equal('email');
+      expect(dataset.checkoutClientId).to.equal('adobe_com');
+      expect(dataset.checkoutMarketSegment).to.equal(undefined);
+    });
+
+    it('merch link to CTA, config values', async () => {
+      setConfig({
```
yesil, line 171: I believe merch block should have its own config system.

test/blocks/merch/merch.test.js
```diff
@@ -196,6 +220,16 @@ describe('Merch Block', () => {
       expect(dataset.template).to.equal('checkoutUrl');
       expect(dataset.promotionCode).to.equal('nicopromo');
     });
+
+    it('merch link to UCv2 cta with Link-level overrides.', async () => {
```
yesil, line 224: nit: `link-level` instead of `Link-level`

test/blocks/merch/mocks/body.html
```diff
@@ -56,6 +56,11 @@ <h2>Promo prices</h2>
     href="/tools/ost?osi=0RwUma15VBOveHgpEz6cKCxATwVkUO8WO3QkI97f4Sg&type=price&promo=nicopromo">Price -
     632B3ADD940A7FBB7864AA5AD19B8D28 - All Apps </a>
 </p>
+<h2>CTAs</h2>
+<p>CTA with config default: <a class="merch cta config"
+  href="/tools/ost?osi=0RwUma15VBOveHgpEz6cKCxATwVkUO8WO3QkI97f4Sg&type=checkoutUrl">CTA Buy Now</a></p>
+<p>CTA with link-level overrides: <a class="merch cta link-overrides"
+  href="/tools/ost?osi=0RwUma15VBOveHgpEz6cKCxATwVkUO8WO3QkI97f4Sg&type=checkoutUrl&checkoutType=UCv2&workflowStep=checkout_email&marketSegment=EDU">CTA Buy Now</a></p>
```
yesil, line 63: in `checkout_email` isn't `checkout` part implicit with `ucv2` ?  Are there other values?
if not, I recommend to keep only the step value, e.g: `email`.
Also, if we dont' keep the `checkoutType` in the OST url but as a context value, switch from `ucv2` to `ucv3` would be seamless.

libs/blocks/merch/README.MD
3ch023: @vhargrave i have no permissions to edit wiki, who should i ask for access?

libs/blocks/merch/checkout-context.js
3ch023: I find it easier to locate checkout logic in a dedicated file, but if it affects performance I am ok to move it back to merge.js.

libs/blocks/merch/checkout-context.js
3ch023: sure, will fix that

libs/blocks/merch/checkout-context.js
3ch023: >Won't things like the checkoutContext be decided on a page-by-page basis?
this was the case for Dexter indeed. However, in Milo, we are challenging the actual need for this requirement.
These parameters can be global + overwritten on the link level. 

We will need to use metadata for 'promo code', but this parameter is out of the scope of this PR.
Another possible use case for metadata might be the 'checkoutType' parameter, but it also can be that we will switch all CTAs at once, so there is no need to make it configurable in metadata right now.

libs/blocks/merch/README.MD
```diff
@@ -0,0 +1,14 @@
...
+in scripts.js
```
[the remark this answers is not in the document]
3ch023: scripts.js doesn't have commerce config. 
it _can_ have it, if milo consumer needs to overwrite a default value that is placed in merch block.

libs/blocks/merch/checkout-context.js
3ch023: Workflow Step default is not provided in Tacocat. This is something I wanted to discuss, shall I add it to the taco and remove it from here?
For workflowStep it does make sense.

However there probably should not be default value for client id, so we still will need one in the milo, wdyt?

yesil: it is though:
image
yes, `client_id` is **adobe_com** by default but I agree that it must be overriden in the consumers.

test/blocks/merch/merch.test.js
yesil: AFAIK `setConfig` is consumer level configuration and for merch, we may need contextual parameters.
Actually, we should start with metadata and see how it goes.
since each consumer has own metadata, setting the client_id and checkout workflow (only ucv2 ones as ucv3 by default) in metadata.xls is probably the simplest approach.

test/blocks/merch/mocks/body.html
yesil: I agree, the ost link should contain only exceptional values.
the checkout type and step could indeed be a context value.

libs/blocks/merch/checkout-context.js
3ch023: as discussed in call, might get an error because of an empty value being passed. i am going to check & fix it.

test/blocks/merch/merch.test.js
3ch023: the best part is that, if we will need a checkout type in metadata - we can add it at any moment without significant cost.

vhargrave: @3ch023 you can specify where metadata appears. Either on a page by page basis or in all pages in a folder etc.

libs/blocks/merch/README.MD
vhargrave: I'd ask in the milo-dev channel on slack 👍

libs/blocks/merch/checkout-context.js
vhargrave: Sure, but by that logic there's also no need to make it configurable over the config object either right now. If you don't want to make it configurable, then I'd least switch your code to check the page metadata instead of the config object.

3ch023: > then I'd least switch your code to check the page metadata instead of the config object.
added this for checkoutType

yesil (approved):

vhargrave (approved):
Looks good!

PR #617, opened 2023-04-04
MWPW-129009: Video Loop Enhancement

JasonHowellSlavin:
* Adds the functionality to have videos loop once.  Adding `#autoplay1` at the end of a video url will ensure the video only place once instead of looping. 

Resolves: [MWPW-129009](https://jira.corp.adobe.com/browse/129009)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/slavin/digital-trends-report?martech=off
- After: https://video-loop-options--milo--adobecom.hlx.page/drafts/slavin/digital-trends-report?martech=off

### DIFF
--- a/libs/blocks/video/video.js
+++ b/libs/blocks/video/video.js
@@ -2,7 +2,7 @@ export default function init(a) {
   const { pathname, hash } = a;
 
   const isAutoplay = !!(hash?.includes('autoplay'));
-  const isNotLooped = !!(hash?.includes('noloop'));
+  const isNotLooped = !!(hash?.includes('autoplay1'));
 
   const attrs = isAutoplay ? `playsinline autoplay ${isNotLooped ? '' : 'loop'} muted` : 'playsinline controls';
   const video = `<video ${attrs}>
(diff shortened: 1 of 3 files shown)


rgclayton (changes requested):
I think we can simplify this. 

- `#_autoplay` autoplays and loops infinitely - already a pattern in use
- `#_autoplay1` autoplays and only plays on time - simple update to our existing pattern.

*Noteworthy: all autoplay features need to have `muted` in order to be autoplayed by the browser.

rgclayton:
The Jira ticket linked to this is broken. So, I'm not sure of the requirements this was created from.

There is a [ticket added to the Milo backlog](https://jira.corp.adobe.com/browse/MWPW-129087) that I created last week.

After asking some questions in the dexter-authoring Slack channel it was mentioned that setting video's to loop a specified amount of times (loop four times then stop) isn't a pattern. So, I don't think we add that feature unless there has been a specific request related to that. I think there is more discovery related to that discussion going on.  

cc: @auniverseaway

rgclayton (approved):
Your test link (after) needs updated to reflect code updates. 
Description also needs updated to reflect changes for clarity. 

Otherwise looking good.

saugatmalla (approved):

PR #621, opened 2023-04-06
Improve Codecov config for unexpected coverage changes

hparra:
- PRs should have full coverage but allow wiggle room for [unexpected coverage changes](https://docs.codecov.com/docs/unexpected-coverage-changes).
- main should have same wiggle room, just don't let coverage drop

Resolves one issue causing red builds. See https://github.com/orgs/adobecom/discussions/626

Resolves: [MWPW-129597](https://jira.corp.adobe.com/browse/MWPW-129597)

Test URL:

- https://codecov-changes--milo--hparra.hlx.page/

Validation: `curl -X POST --data-binary @codecov.yaml https://codecov.io/validate`

Documentation: https://docs.codecov.com/docs/codecovyml-reference

### DIFF
--- a/codecov.yaml
+++ b/codecov.yaml
@@ -1,5 +1,5 @@
-codecov:
-  status:
+coverage:
+  status: 
     patch:
       default:
         target: 100%
--- a/codecov.yaml
+++ b/codecov.yaml
@@ -1,0 +1,10 @@
+codecov:
+  status:
+    patch:
+      default:
+        target: 100%
+        threshold: 0.1%
+    project:
+      default:
+        target: auto
+        threshold: 0.1%


codecov.yaml
```diff
@@ -0,0 +1,10 @@
+coverage:
+  status: 
+    patch:
+      default:
+        target: 100%
+        threshold: 0.1%
```
vhargrave, line 6: Are we adding this threshold because the coverage might be irregular? Imo we shouldn't be affected by coverage irregularities and allowing a threshold of 0.1% is not the way to fix it.
We're always uploading our coverage reports, our tests should not be time sensitive etc. If we are having irregularities I'd like to investigate what the root cause of them is instead of allowing for thresholds.

codecov.yaml
```diff
@@ -0,0 +1,10 @@
...
+    project:
+      default:
+        target: auto
+        threshold: 0.1%
```
vhargrave, line 10: I'd remove the threshhold here too.

vhargrave (changes requested):
I respect the initiative! 
Couple of comments: 
1. I'm not sure we should be setting a specific codecov config for milo. Milo kind of sets the baseline for our other adobecom projects and so we should probably make configuration changes to our codecov in our global config - https://docs.codecov.com/docs/codecov-yaml
2. I don't know the history for this PR, but in my opinion a Jira task and link is missing. This would show that this change had been verified with the milo team first before just building it.

hparra:
> I respect the initiative!

Concerns come from [discussion](https://github.com/orgs/adobecom/discussions/626) that was started by one of org's lead architects 64 days ago and no one had responded to until now.

> I'm not sure we should be setting a specific codecov config for milo. Milo kind of sets the baseline for our other adobecom projects and so we should probably make configuration changes to our codecov in our global config - https://docs.codecov.com/docs/codecov-yaml

I disagree strongly. Projects and teams have their own standards and exceptions. We should never couple. I have never seen coupling like that at scale (5000+ project repos) even after they were moved to a monorepo.

> I don't know the history for this PR, but in my opinion a Jira task and link is missing. This would show that this change had been verified with the milo team first before just building it. - Submittings PRs

I could not test this change without making the actual PR since codecov and other checks are not enabled on my fork and adding them may affect whatever non-determinism may be happening here.

Intent of these changes was mentioned in 1:1 with @auniverseaway and with Milo Contributors last week. I could not bring it up again this week because meeting was canceled. My hope was that [discussion](https://github.com/orgs/adobecom/discussions/626) would lead to creation of a new epic that would cover multiple changes to make checks green and improve developer velocity.

vhargrave:
> @vhargrave
> 
> > I respect the initiative!
> 
> Concerns come from [discussion](https://github.com/orgs/adobecom/discussions/626) that was started by one of org's lead architects 64 days ago and no one had responded to until now.
> 
> > I'm not sure we should be setting a specific codecov config for milo. Milo kind of sets the baseline for our other adobecom projects and so we should probably make configuration changes to our codecov in our global config - https://docs.codecov.com/docs/codecov-yaml
> 
> I disagree strongly. Projects and teams have their own standards and exceptions. We should never couple. I have never seen coupling like that at scale (5000+ project repos) even after they were moved to a monorepo.
> 
> > I don't know the history for this PR, but in my opinion a Jira task and link is missing. This would show that this change had been verified with the milo team first before just building it. - Submittings PRs
> 
> I could not test this change without making the actual PR since codecov and other checks are not enabled on my fork and adding them may affect whatever non-determinism may be happening here.
> 
> Intent of these changes was mentioned in 1:1 with @auniverseaway and with Milo Contributors last week. I could not bring it up again this week because meeting was canceled. My hope was that [discussion](https://github.com/orgs/adobecom/discussions/626) would lead to creation of a new epic that would cover multiple changes to make checks green and improve developer velocity.

Fair - you still should have a Jira issue though. 

Regarding the global config, good point. 👍

vhargrave (approved):
After discussion today we've agreed that we'll try this out and separate tasks will be created to investigate the flaky tests.

honstar (approved):

PR #628, opened 2023-04-11
MWPW-128106 FG handle rate limit

Nethraasivakumar:
https://jira.corp.adobe.com/browse/MWPW-128106 

This PR has the implementation of the fetchWithRetry function which would enable the status and retry-after value to be tracked. A timeout is set if status code 429 and retry-after value is observed during a fetch call. 

The PR does the following:
- fetchWithRetry function in sharepoint.js
- in order to use fetchWithRetry function only for floodgate, the fetchWithRetry function is enabled only in floodgate.js
- the appropriate fetch calls have been updated with the fetchWithRetry 

**Test URLs:**

**Before**: https://main--milo--adobecom.hlx.page/tools/floodgate/index.html?project=milo--adobecom&referrer=https%3A%2F%2Fadobe.sharepoint.com%2F%3Ax%3A%2Fr%2Fsites%2Fadobecom%2F_layouts%2F15%2FDoc.aspx%3Fsourcedoc%3D%257B59E5B572-A72C-40B8-96AF-4591908EA744%257D%26file%3Dtest_with_50.xlsx%26action%3Ddefault%26mobileredirect%3Dtrue%26cid%3D8244df40-9733-45b5-91bf-e912822c912c 
**After**: https://main--milo--nethraasivakumar.hlx.page/tools/floodgate/index.html?project=milo--adobecom&referrer=https%3A%2F%2Fadobe.sharepoint.com%2F%3Ax%3A%2Fr%2Fsites%2Fadobecom%2F_layouts%2F15%2FDoc.aspx%3Fsourcedoc%3D%257B59E5B572-A72C-40B8-96AF-4591908EA744%257D%26file%3Dtest_with_50.xlsx%26action%3Ddefault%26mobileredirect%3Dtrue%26cid%3D8244df40-9733-45b5-91bf-e912822c912c


tools/loc/sharepoint.js
```diff
@@ -84,6 +82,53 @@ function getAuthorizedRequestOption({
   return options;
 }
 
+let nextCallAfter = 0;
```
honstar, line 85: I don't think we should be adding the fetch-with-retry functionality to the `loc/sharepoint.js` file since there is nothing specific to sharepoint. I'd rather add this to `floodgate/js/utils.js` for now.
Please note that the complete loc and floodgate backend will be externalized as part of MWPW-122323 and MWPW-125048 anyway.

sukamat: Actually this check is directly related to the response provided by Sharepoint APIs. The fetchWithRetry checks for headers returned by Sharepoint APIs. This is not specific to Floodgate or Loc but applies to anything that will rely on Sharepoint APIs. So considering that, it makes more sense for this to be in sharpeoint.js. More in this [MS documentation](https://learn.microsoft.com/en-us/sharepoint/dev/general-development/how-to-avoid-getting-throttled-or-blocked-in-sharepoint-online#how-to-handle-throttling). That said, I do agree that this will move to AIO at some point and is being added here for now for the ease of testing.

honstar: Thanks for the context, guess I overlooked the status headers from sharepoint. I'm ok with adding this here for the time being until we move things to I/O runtime.

sukamat (approved):
@Nethraasivakumar Thanks for performing the tests. As discussed in the ticket, let's perform some more tests once we have the $batch and copy API changes. The changes look promising for now.

honstar (approved):

PR #631, opened 2023-04-11
MWPW-127711 Marquee play button

JasonHowellSlavin:
* Checks the link next to a marquee image for an .svg extension, and adds a clickable and authorable button on top of the image. 

Resolves: [MWPW-127711](https://jira.corp.adobe.com/browse/MWPW-127711)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/slavin/marquee-play-button
- After: https://marquee-play-button--milo--adobecom.hlx.page/drafts/slavin/marquee-play-button


JasonHowellSlavin:
> @JasonHowellSlavin , there are marquee related test failures. Can you please check?

Good looking out @honstar, thank you. Added an additional test for the functionality as well.

test/blocks/marquee/mocks/body.html
```diff
@@ -315,3 +315,28 @@ <h1 id="standing-out-amid-a-sea-of-labels-1">Standing out amid a sea of labels.<
...
+      <p>Body M Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation.</p>
+      <p><em><a href="https://www.adobe.com/">Lorem ipsum</a></em> <strong><a href="https://www.adobe.com/">Learn more</a></strong></p>
+    </div>
+    <div>
+      <p>
+        <picture>
+          <source type="image/webp" srcset="./media_12a67eee02f1aa5a0c2f2d2d1d540dbd1ac424b35.png?width=2000&amp;format=webply&amp;optimize=medium" media="(min-width: 600px)">
+          <source type="image/webp" srcset="./media_12a67eee02f1aa5a0c2f2d2d1d540dbd1ac424b35.png?width=750&amp;format=webply&amp;optimize=medium">
+          <source type="image/png" srcset="./media_12a67eee02f1aa5a0c2f2d2d1d540dbd1ac424b35.png?width=2000&amp;format=png&amp;optimize=medium" media="(min-width: 600px)">
+          <img loading="eager" alt="" type="image/png" src="./media_12a67eee02f1aa5a0c2f2d2d1d540dbd1ac424b35.png?width=750&amp;format=png&amp;optimize=medium" width="720" height="520">
```
Ruchika4, line 336: png path mentioned here are giving 404 and hence the test cases are failing
Screen Shot 2023-04-12 at 10 20 36 AM

Could you please check. Thanks.

Ruchika4 (changes requested):

test/blocks/marquee/mocks/body.html
JasonHowellSlavin: Hi @Ruchika4 ,  All of the tests I ran pass locally, I did notice the one that failed was in the utils.js file: 
Screen Shot 2023-04-12 at 11 13 38 AM

Also, I did not resolve the network 404s because it does not appear to be an issue that is focused on in  our unit tests:
Screen Shot 2023-04-12 at 11 10 57 AM
Do you think it would be better to remove the 404s in this particular instance? Maybe a good community topic? 

I am happy to look into the utils test, but they are all passing locally for me.

libs/blocks/marquee/marquee.js
```diff
@@ -80,8 +80,16 @@ const decorateImage = (media) => {
 
   const imageLink = media.querySelector('a');
   const picture = media.querySelector('picture');
+  const buttonImage = imageLink && imageLink.querySelector('img[src$=".svg"]');
```
rgclayton, line 83: I think you need to do a check if the link is to a fragment and opening a modal. Those are key points for this, yeah?

Valid use case for SVG's is to be able to link them, to any url.

Example page:
View the bottom marquee. I have the svg linking to adobe.com.
https://marquee-play-button--milo--adobecom.hlx.page/drafts/rclayton/marquee-play-button

JasonHowellSlavin: Do you think something like this would suffice:
```
  let buttonImage;

  if (imageLink && imageLink.dataset?.modalHash) {
    buttonImage = imageLink.querySelector('img[src$=".svg"]');
  }

  if (buttonImage) {
```

If the link links to anything in the fragments folder, it will get decorated with the correct modal data attributes, so it might be enough to just check that there is a modal hash? Or we could check the modal path? 

Also, this would prevent the overlay from occurring in your example, is that okay? It would default to the current `decorateImage` functionality, and replace the link contents with the image.

test/blocks/marquee/mocks/body.html
JasonHowellSlavin: Odd with the unit tests. I added the discussion here: https://github.com/orgs/adobecom/discussions/635. Thank you for the input!

libs/blocks/marquee/marquee.js
rgclayton: Yeah. Just thinking about if an author does want to add an SVG under the image for some reason, that it doesn't decorate the image. The video button only decorates when linking to/opening a modal.

@auniverseaway Any concerns with this I might be overlooking? I'm thinking about our conversation related to image linking and this feature.

JasonHowellSlavin:
Closing this PR. The ticket associated is closed. We are waiting for consonant to provide an overarching solution so we can enact this feature universally throughout milo.

PR #632, opened 2023-04-12
Bulk publish styling

rgclayton:
Updating bulk publisher work to address UI/UX in base PR https://github.com/adobecom/milo/pull/559 
* Creating UI for Milo bulk publishing
* Updating UX

**Test URLs:**
- Before: https://bulk-publish--milo--adobecom.hlx.page/tools/bulk?martech=off
- After: https://bulk-publish-styling--milo--adobecom.hlx.page/tools/bulk?martech=off


libs/blocks/bulk/bulk.js
```diff
@@ -67,117 +67,117 @@ function prettyDate() {
...
+function StatusTitle({ bulkTriggered, submittedAction, urlNumber }) {
   const name = getActionName(submittedAction, true);
   const URLS = (urlNumber > 1) ? 'URLS' : 'URL';
   return bulkTriggered && html`
-      <div class="bulk-status-head">
-          <div class="bulk-status-head-title">
-              STATUS ${name} ${urlNumber} ${URLS}
-          </div>
-          <div>
-              <button class="bulk-status-head-copy" onclick=${copyResults} title="Copy results"></button>
```
jckautzmann, line 92: what's the reason for removing the copy results button?

libs/blocks/bulk/bulk.js
```diff
@@ -67,117 +67,117 @@ function prettyDate() {
...
-            </tr>
-            ${result.map((row) => html`<${StatusRow} row=${row} />`)}
-        `}
+      ${result && html`
+        <tr class="bulk-status-header-row">
+          <th>URL</th>
+          <th class="${bulkPreviewed}">Previewed</th>
+          <th class="${bulkPublished}">Published</th>
+        </tr>
+        ${result.reverse().map((row) => html`<${StatusRow} row=${row} />`)}
```
jckautzmann, line 136: Why are the results reversed? Shouldn't they be displayed in the same order as the input?

libs/blocks/bulk/bulk.js
```diff
@@ -67,117 +67,117 @@ function prettyDate() {
...
-        <div class="bulk-status-footer-container">
-            <${StatusFooter}
-                completion=${completion} />
-        </div>
+      <div class="bulk-status-info">
+        <${StatusTitle}
+          bulkTriggered=${bulkTriggered}
+          submittedAction=${submittedAction}
+          urlNumber=${urlNumber} />
+        <${BulkStatus}
```
jckautzmann, line 172: To be in sync with `StatusTitle` and `StatusContent`, I'd rename it: `StatusCompletion`

libs/blocks/bulk/bulk.js
```diff
@@ -299,46 +288,47 @@ function Bulk({ user, storedOperation }) {
...
+        <div class="bulk-urls-title">Add URLS</div>
+        <div class="bulk-action">
+        <${SelectBtn}
+          actionElt=${actionElt}
+          onSelectChange=${onSelectChange}
+          storedOperationName=${storedOperation.name} />
+        <${SubmitBtn}
+          submit=${submit}
+          selectedAction=${selectedAction} />
+      </div>
```
jckautzmann, line 309: nit: add some indenting

libs/blocks/caas-config/caas-config.js
```diff
@@ -145,6 +145,7 @@ const defaultOptions = {
     incremental: 'Incremental',
   },
   paginationType: {
+    none: 'None',
```
jckautzmann, line 148: Are the changes in caas-config, caas/utils, and merch below related to this PR?

test/blocks/bulk/performance.test.js
jckautzmann, line 1: why should this test be removed?

test/blocks/bulk/utils.test.js
jckautzmann, line 1: why should this test be removed?

jckautzmann:
@rgclayton Really cool UI/UX! 

I added a few comments to the PR. Here some more comments/feedback:
1) when successful, the result for each URL is displayed in black, whereas the unsuccessful result is red: how about displaying successful result in green?
2) the select dropdown for selecting the bulk operation has no specific style and defaults to the browser one: should it have its own styling?
3) the error message when there is no URL or when there are too many URLS is barely visible
4) I added support for duplicate URLs in https://github.com/adobecom/milo/pull/559/commits/5596ac619f9d28a0feb3db2bfea739b34eb5715d
5) you noticed that the bulk publishing only works with the main branch otherwise an error is thrown: this can be configured in bulk-config.xlsx

PR #634, opened 2023-04-12
Bulk publish UI/UX updates

rgclayton:
Updating bulk publisher work to address UI/UX in base PR https://github.com/adobecom/milo/pull/559 
* Creating UI for Milo bulk publishing
* Updating UX

**Test URLs:**
- Before: https://bulk-publish--milo--adobecom.hlx.page/tools/bulk?martech=off
- After: https://bulk-publish-ui--milo--adobecom.hlx.page/tools/bulk?martech=off


libs/blocks/bulk/bulk.js
```diff
@@ -67,117 +65,117 @@ function prettyDate() {
...
+function StatusTitle({ bulkTriggered, submittedAction, urlNumber }) {
   const name = getActionName(submittedAction, true);
   const URLS = (urlNumber > 1) ? 'URLS' : 'URL';
   return bulkTriggered && html`
-      <div class="bulk-status-head">
-          <div class="bulk-status-head-title">
-              STATUS ${name} ${urlNumber} ${URLS}
-          </div>
-          <div>
-              <button class="bulk-status-head-copy" onclick=${copyResults} title="Copy results"></button>
```
rgclayton, line 92: @jckautzmann We decided to remove copy results button because it wasn't requested and questioned it getting used. Also, one less thing to maintain.

libs/blocks/bulk/bulk.js
```diff
@@ -67,117 +65,117 @@ function prettyDate() {
...
-            </tr>
-            ${result.map((row) => html`<${StatusRow} row=${row} />`)}
-        `}
+      ${result && html`
+        <tr class="bulk-status-header-row">
+          <th>URL</th>
+          <th class="${bulkPreviewed}">Previewed</th>
+          <th class="${bulkPublished}">Published</th>
+        </tr>
+        ${result.reverse().map((row) => html`<${StatusRow} row=${row} />`)}
```
rgclayton, line 134: @jckautzmann So the most recent previewed/published url sits at the top of the list.

jckautzmann: As a user I find it misleading that the ordering of the results is different from the input area.

jckautzmann:
@rgclayton Really cool UI/UX! 

I added a few comments to the PR. Here some more comments/feedback:
1) when successful, the result for each URL is displayed in black, whereas the unsuccessful result is red: how about displaying successful result in green?
2) the select dropdown for selecting the bulk operation has no specific style and defaults to the browser one: should it have its own styling?
3) FYI: I added support for duplicate URLs in https://github.com/adobecom/milo/pull/559/commits/5596ac619f9d28a0feb3db2bfea739b34eb5715d
4) you noticed that the bulk publishing only works with the main branch otherwise an error is thrown: this can be configured in `bulk-config.xlsx`

rgclayton:
> * when successful, the result for each URL is displayed in black, whereas the unsuccessful result is red: how about displaying successful result in green?

Just want highlight the errors. 
> * the select dropdown for selecting the bulk operation has no specific style and defaults to the browser one: should it have its own styling?

Styling for browser components can be a little combersome. This matches the dropdowns in CaaS tool.

> * FYI: I added support for duplicate URLs in [5596ac6](https://github.com/adobecom/milo/commit/5596ac619f9d28a0feb3db2bfea739b34eb5715d)
Awesome!
> * you noticed that the bulk publishing only works with the main branch otherwise an error is thrown: this can be configured in `bulk-config.xlsx`

Ok. Good to know.

Thanks @jckautzmann

libs/blocks/bulk/bulk.js
rgclayton: I think the plan is to go with this unless we get complaints. 

cc: @auniverseaway

rgclayton:
@jckautzmann If you are good with this let's merge it into your branch to finalize and get your PR approved and merged.

jckautzmann (approved):

PR #636, opened 2023-04-13
MWPW-125561 - UAR Results Block

colloyd:
* Initial commit
* Basic UAR results
* Nested UAR results
* Style support

Resolves: [MWPW-125561](https://jira.corp.adobe.com/browse/MWPW-125561)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/colloyd/uar-results-block/uar-results?quizKey=cc-uar-result&martech=off
- After: https://mwpw-125561-uar-results-block--milo--colloyd.hlx.page/drafts/colloyd/uar-results-block/uar-results?quizKey=cc-uar-result&martech=off

### DIFF
--- a/libs/blocks/uar-results/uar-results.js
+++ b/libs/blocks/uar-results/uar-results.js
@@ -1,0 +1,59 @@
+/* eslint-disable no-await-in-loop */
+import { createTag } from '../../utils/utils.js';
+import { getMetadata, handleStyle } from '../section-metadata/section-metadata.js';
+
+async function loadFragments(el, experiences) {
+  // eslint-disable-next-line no-restricted-syntax
+  for (const href of experiences) {
+    const a = createTag('a', { href });
+    el.append(a);
+    const { default: createFragment } = await import('../fragment/fragment.js');
+    await createFragment(a);
+  }
+}
+
+function redirectPage(quizUrl) {
+  const url = (quizUrl) ? quizUrl.text : 'https://adobe.com';
+  console.log(`Invalid Result, redirecting to: ${url}`);
+}
+
+export default async function init(el) {
+  const data = getMetadata(el);
+  const metaData = el.querySelectorAll('div');
+  const params = new URL(document.location).searchParams;
+  const localStoreKey = params.get('quizKey');
+  const quizUrl = data['quiz-url'];
+
+  let results = localStorage.getItem(localStoreKey);
+
+  for( const div of metaData){
+    div.remove();
+  }
+
+  if (!results) {
+    redirectPage(quizUrl);
+    return
+  }
+
+  results = JSON.parse(results);
+
+  if(data['nested-fragments'] && el.classList.contains('nested')) {
+    const nested = results[data['nested-fragments'].text];
+    if (nested) loadFragments(el, nested);
+  } else if (el.classList.contains('basic')) {
+    const structure = results['structure-fragments'];
+    if (!structure) { 
+      redirectPage(quizUrl);
+      return
+    }
+    loadFragments(el, structure);
+  } else {
+    return
+  }
+  
+  if(data.style) {
+    el.classList.add('section');
+    handleStyle(data.style.text, el);
+  }
+  el.classList.add('show');
+}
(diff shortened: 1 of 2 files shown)


colloyd:
To properly test this, hit the test url and open dev tools. Add the 'cc-uar-result' key to Local Storage and set the value to this:

`{"structure-fragments":["https://main--milo--adobecom.hlx.page/fragments/colloyd/sample-uar-fragments/uar-sample-marquee-cc","https://main--milo--adobecom.hlx.page/fragments/colloyd/sample-uar-fragments/uar-sample-card-list"],"marquee-product":["https://main--milo--adobecom.hlx.page/fragments/colloyd/sample-uar-fragments/uar-sample-marquee-product-au","https://main--milo--adobecom.hlx.page/fragments/colloyd/sample-uar-fragments/uar-sample-marquee-product-lr","https://main--milo--adobecom.hlx.page/fragments/colloyd/sample-uar-fragments/uar-sample-marquee-product-id"],"commerce-card":["https://main--milo--adobecom.hlx.page/fragments/colloyd/sample-uar-fragments/uar-sample-card-au","https://main--milo--adobecom.hlx.page/fragments/colloyd/sample-uar-fragments/uar-sample-card-lr","https://main--milo--adobecom.hlx.page/fragments/colloyd/sample-uar-fragments/uar-sample-card-id"]}`

echampio-at-adobe (approved):
It's a nice validation of Milo as a framework that so much can be accomplished with so little code. In the long run, might be nice to add a real simple regex to validate the fragment paths, just to make sure they're on the same domain under the fragments path. I'm sure createFragment will do it's own validation. Just don't like to propagate garbage. Might also be nice to add a logger method of our own that can use the console while we're banging away among our selves, but that we can later redirect to lana without changing a bunch of lines everywhere. These comments are just future ideas, though.

JackySun9 (approved):

echampio-at-adobe:
> I do see the pages with content after I setup local storage key, just want to know is it dynamic to generate marquee-product and commerce-card pages according to some excel sheets?

Hopefully, it will be. The list of fragments you see in local storage should eventually be based upon the matches made in results.xslx, Thus, dynamic.

libs/blocks/uar-results/uar-results.js
```diff
@@ -0,0 +1,59 @@
...
+    const a = createTag('a', { href });
+    el.append(a);
+    const { default: createFragment } = await import('../fragment/fragment.js');
+    await createFragment(a);
+  }
+}
+
+function redirectPage(quizUrl) {
+  const url = (quizUrl) ? quizUrl.text : 'https://adobe.com';
+  console.log(`Invalid Result, redirecting to: ${url}`);
```
rgclayton, line 17: Should tap into Lana

echampio-at-adobe: Hi @rgclayton. I mentioned lana in my comment too. I asked whether we should just define our own log wrapper so we can use the console for now, but send everything to lana when we're really ready to merge to main. Has anyone else written a method like that already that we could import?

echampio-at-adobe (commented):
Hi @colloyd,

Given your code, how will it be possible for a section like "More apps to explore" to drop out when there aren't any cards specified?

https://www.adobe.com/creativecloud/quiz-recommender-var/result.html#type=cc:app-reco&quiz=UARv21&selectedOptions=category_video/Flagship/Product_AN_Video/plan_personal
Screenshot 2023-04-13 at 2 26 08 PM

rgclayton:
> In the long run, might be nice to add a real simple regex to validate the fragment paths, just to make sure they're on the same domain under the fragments path. I'm sure createFragment will do it's own validation.

I think you can use `getConfig()` from `utils.js`.
[Here's an example ](https://github.com/adobecom/milo/blob/main/libs/blocks/carousel/carousel.js#L1-L9)

@colloyd @echampio-at-adobe

libs/blocks/uar-results/uar-results.js
rgclayton: > Has anyone else written a method like that already that we could import?

@echampio-at-adobe not that i know of.

colloyd: That's the plan, I don't want to have redirects happening when we're debugging all of this, hence the console.log()

PR #637, opened 2023-04-13
stabilize unit test for disable auto bloking

Ruchika4:
Unit test case for disable auto block has been a little flaky in github actions although it always passed on local for me. Added  a step to wait for the element to stabilize it.

Below is the screenshot of failure for this test in github actions
Screen Shot 2023-04-13 at 9 56 52 AM

**Test URLs:**
N/A


mokimo (approved):

vhargrave (approved):
Nice! :)

PR #638, opened 2023-04-13
MWPW-129529: consume latest tacocat

yesil:
Update country code from IMS country on co links
Map local environment to commerce stage
and stage+prod to commerce prod
Reduced tacocat.js  gzip size by 5kb (20% of gain).

Until the core milo IMS API is provided, we would like to use the `setInterval` based solution in this PR.

Due to CORS, the **After URL** should be tested with the following CORS settings:

`Access-Control-Allow-Origin: *`
`Access-Control-Allow-Headers: x-api-key`

Resolves: 
[MWPW-129529](https://jira.corp.adobe.com/browse/MWPW-129529)
[MWPW-129128](https://jira.corp.adobe.com/browse/MWPW-129128)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ilyas/merch?martech=off
- After: https://mwpw-129529--milo--yesil.hlx.page/drafts/ilyas/merch?martech=off


honstar:
@yesil : there are a lot of merch block unit-tests failing now. Can you please check?

libs/blocks/merch/merch.css
```diff
@@ -1,4 +1,5 @@
-span[data-wcs-osi],
+.bad-content,
```
honstar, line 1: Is this needed? AFAICT, this is only needed in unit-tests.

libs/blocks/merch/merch.js
```diff
@@ -40,70 +27,88 @@ function omitNullValues(target) {
...
+      clearInterval(check);
+      if (window.adobeIMS.isSignedInUser()) {
+        window.adobeIMS
+          .getProfile()
+          .then(({ countryCode }) => resolve(countryCode))
+          .catch(() => resolve());
+      } else {
+        resolve();
+      }
+    } else if (count > 25) {
```
honstar, line 44: Just curious how you came up with this number of retries? If I read this code correctly, you're trying get hold of `window.adobeIMS` with a 200ms wait interval in between for a max number of 25 retries = 5000ms. This seems a bit long imo.

libs/blocks/merch/merch.js
```diff
@@ -40,70 +27,88 @@ function omitNullValues(target) {
...
+  let language = wcsLocale[0];
+  const country = wcsLocale[1] || 'US';
   if (!SUPPORTED_LANGS.includes(language)) {
     language = 'en';
   }
-  const host = envName === ENV_PROD
-    ? 'https://www.adobe.com'
-    : 'https://www.stage.adobe.com';
-  const scriptUrl = `${host}/special/tacocat/lib/${VERSION}/tacocat.js`;
+  const host = envName === ENV_LOCAL
```
honstar, line 58: This means that every environment that is not `local` will default to `prod`, hence `stage` would be considered `prod` as well, no?

libs/blocks/merch/merch.js
```diff
@@ -40,70 +27,88 @@ function omitNullValues(target) {
...
+    ? 'https://www.stage.adobe.com'
+    : 'https://www.adobe.com';
+
   const literalScriptUrl = `${host}/special/tacocat/literals/${language}.js`;
-  return { scriptUrl, literalScriptUrl, country, language };
+  const scriptUrl = `${host}/special/tacocat/lib/${VERSION}/tacocat.js`;
+  const tacocatEnv = envName === ENV_LOCAL ? 'STAGE' : 'PRODUCTION';
```
honstar, line 64: Same as above, a non `local` env is automatically treated as `prod`.

libs/blocks/merch/merch.js
yesil: Until the Milo IMS supports is ready, I think this is a safe approach to wait for 5s.  It is not blocking so the UI thread.
Ideally, we will have a promise from milo core that will resolve as soon as imslibjs is loaded.

libs/blocks/merch/merch.js
yesil: @honstar yes, the reason behind this is the following:

When we test the "after URL" in the PRs, the environment is considered stage. If we load tacocat from stage environment, as soon as PR is merged, it will load from prod environment, which means we would have never tested with the prod url.

I consider tacocat stage more for tacocat devs to validate things while developing locally.

I'm still not 100% sure whether this is a good decision though. WDYT?

libs/blocks/merch/merch.js
yesil: as explained above, we can discuss this in the milo commerce meeting.

libs/blocks/merch/merch.js
honstar: If I'm not mistaken, you can force override the environment to be used by adding a query parameter `env`, see https://github.com/adobecom/milo/blob/main/libs/utils/utils.js#L110.

yesil: Indeed this comes handy. Let me update this part, thanks.

honstar (approved):

libs/blocks/merch/merch.js
yesil: @honstar  reverted back to prod/stage.

libs/blocks/merch/merch.js
```diff
@@ -24,86 +18,97 @@ const GEO_MAPPINGS = {
...
+      clearInterval(check);
+      if (window.adobeIMS.isSignedInUser()) {
+        window.adobeIMS
+          .getProfile()
+          .then(({ countryCode }) => resolve(countryCode))
+          .catch(() => resolve());
+      } else {
+        resolve();
+      }
+    } else if (count > 25) {
```
npeltier, line 46: can we somehow cover this use case? maybe stubbing isSignedInUser, why 25 btw :) ?

libs/blocks/merch/merch.js
```diff
@@ -24,86 +18,97 @@ const GEO_MAPPINGS = {
...
+      resolve();
+    }
+  }, 200);
+});
+window.tacocat.imsCountryPromise = imsCountryPromise();
+
+export const getTacocatEnv = (envName, locale) => {
+  const wcsLocale = (GEO_MAPPINGS[locale.prefix] ?? locale.ietf).split('-', 2);
+  let language = wcsLocale[0];
+  const country = wcsLocale[1] || 'US';
```
npeltier, line 57: can we do something like LOCALE_PATTERN=/(?<language>\w+)-(?<country>\w+)/ and match groups ? Talking of which, won't we have an issue with some languages that have - in their code?

libs/blocks/merch/merch.js
```diff
@@ -24,86 +18,97 @@ const GEO_MAPPINGS = {
...
-  });
-}
-
-function loadTacocat() {
-  if (initialized) {
-    return;
-  }
-  initialized = true;
+window.tacocat.loadPromise = new Promise((resolve) => {
+  const { env, locale } = getConfig();
```
npeltier, line 72: we need landscape here, and pass it along. May be making env some object?

libs/blocks/merch/merch.js
yesil: 25 * 200 = 5000 second before failing the promise.

libs/blocks/merch/merch.js
yesil: @npeltier this is simple enough.
also as per libs/scripts/scripts.js there is no language with `-` in it.

3ch023: it can happen that 'zh' will turn into 'zh-Hant' later, but we could guard against it just by splitting by the last found '-'. 
however it can be a future improvement for a next PR.

[https://github.com/adobecom/milo/blob/main/libs/scripts/scripts.js#L99](https://github.com/adobecom/milo/blob/main/libs/scripts/scripts.js#L99)

libs/blocks/merch/merch.js
3ch023: we need to decide first where the landscape=draft would be stored, before passing it. so probably shouldn't be in scope of this pr

libs/blocks/merch/merch.js
3ch023: could you add it as a comment there @yesil ?

3ch023 (approved):

libs/blocks/merch/merch.js
yesil: @3ch023 any unnecessary byte is bad for performance. reading the code it is obvious.
And this could should vanish as soon as milo ims support is ready.

cc: @chrischrischris

libs/blocks/merch/merch.js
yesil: yes, i tested this against the current scripts object.

libs/blocks/merch/merch.js
yesil: IMO we should address landscape in a separate PR.

PR #639, opened 2023-04-13
MWPW-126640 Offer Selector Tool in Milo

Axelcureno:
- Adds Offer Selector Tool link button to AEM Sidekick inside "Tools" under the name "Use Offer" to give authors an easier access to the tool when authoring prices and checkout links. 
- Offer Selector Tool implemented in Milo. Consumed as an external dependency in stage.adobe.com and which lets authors copy/paste both price and checkout links easily in Word documents. These links can also deeplink back to OST when clicked in Word.
- IMS User authentication enabled to prevent unauthorized access to the tool. 

Resolves: [MWPW-126640](https://jira.corp.adobe.com/browse/MWPW-126640)

**Test URLs:**
- After: https://ost--milo--adobecom.hlx.page/tools/ost?martech=off


libs/blocks/ost/ost.js
```diff
@@ -0,0 +1,49 @@
...
+  return window.adobeIMS?.getAccessToken()?.token;
+};
+
+export default async function init() {
+  const ostAosAccessToken = await getImsToken();
+  const ostCountry = 'US';
+  const ostLanguage = 'en';
+  const ostEnvironment = 'PROD';
+  const ostWcsApiKey = 'wcms-commerce-ims-ro-user-cc';
+  const ostAosApiKey = 'wcms-commerce-ims-user-prod';
```
yesil, line 30: nit: I don't think we need to prefix variables with ost, they are implicitly for OSt.

yesil (approved):

libs/blocks/ost/ost.js
```diff
@@ -0,0 +1,49 @@
+
```
honstar, line 1: Superfluous empty line at beginning of file.

honstar (approved):

libs/blocks/ost/ost.js
Axelcureno: removed

libs/blocks/ost/ost.js
Axelcureno: removed

yesil (approved):

3ch023 (approved):

PR #641, opened 2023-04-17
Keyboard support

mokimo:
### Description
This brings the keyboard navigation feature that current exists in AEM to the milo feds-gnav. This has been a hard requirement for the AEM-Feds-GNAV in the past for accessibility reasons, which is why we need to port this to the milo navigation. Resolves: [MWPW-125255](https://jira.corp.adobe.com/browse/MWPW-125255)

### Feature
A few of the notable use-cases
- RTL support
- Shift+Tab support
- Cycle through the navigation if the search is open
- Close any open menu when you open a new one (E.g. only search OR profile OR a popup can be open)
- Arrow key support on the main nav & popups, also on mobile
- enter/space/escape keys should work on any element now
- In total ~80 test cases based on the old implementation which had a lot of back and forth with the accessibility team, there's a lot of small use-cases - so we ported those use-cases.

**Test URLs:**
- After: https://keyboard-support--milo--adobecom.hlx.page/drafts/ramuntea/gnav-refactor?martech=off#

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -5,7 +5,14 @@ import {
   localizeLink,
   loadStyle,
 } from '../../utils/utils.js';
-import { toFragment, getFedsPlaceholderConfig, getAnalyticsValue, decorateCta } from './utilities/utilities.js';
+import {
+  toFragment,
+  getFedsPlaceholderConfig,
+  getAnalyticsValue,
+  decorateCta,
+  trigger,
+  closeAllDropdowns,
+} from './utilities/utilities.js';
 import { replaceKey } from '../../features/placeholders.js';
 
 const CONFIG = {
@@ -57,15 +64,9 @@ const decorateSignIn = async ({ rawElem, decoratedElem }) => {
   } else {
     signInElem = toFragment`<a href="#" daa-ll="${signInLabel}" class="feds-signIn" role="button" aria-expanded="false" aria-haspopup="true">${signInLabel}</a>`;
 
-    signInElem.addEventListener('click', () => {
-      const isOpen = signInElem.getAttribute('aria-expanded') === 'true';
-
-      if (isOpen) {
-        signInElem.setAttribute('aria-expanded', 'false');
-      } else {
-        signInElem.setAttribute('aria-expanded', 'true');
-      }
-    });
+    signInElem.addEventListener('click', () => trigger({ element: signInElem }));
+    signInElem.addEventListener('keydown', (e) => e.code === 'Escape' && closeAllDropdowns());
+    dropdownElem.addEventListener('keydown', (e) => e.code === 'Escape' && closeAllDropdowns());
 
     dropdownElem.classList.add('feds-signIn-dropdown');
 
@@ -130,7 +131,6 @@ class Gnav {
 
   init = () => {
     this.elements.curtain = toFragment`<div class="feds-curtain"></div>`;
-
     this.elements.navWrapper = toFragment`
       <div class="feds-nav-wrapper">
         ${this.isDesktop.matches ? '' : this.decorateBreadcrumbs()}
@@ -157,6 +157,7 @@ class Gnav {
       </div>`;
 
     this.el.addEventListener('click', this.loadDelayed);
+    this.el.addEventListener('keydown', this.loadDelayed);
     setTimeout(() => this.loadDelayed(), 3000);
     this.loadIMS();
     this.el.append(this.elements.curtain, this.elements.topnavWrapper);
@@ -195,16 +196,19 @@ class Gnav {
     // eslint-disable-next-line no-async-promise-executor
     this.ready = this.ready || new Promise(async (resolve) => {
       this.el.removeEventListener('click', this.loadDelayed);
+      this.el.removeEventListener('keydown', this.loadDelayed);
       const [
         decorateDropdown,
         { appLauncher },
         ProfileDropdown,
         Search,
+        KeyboardNavigation,
       ] = await Promise.all([
         loadBlock('./blocks/navDropdown/dropdown.js'),
         loadBlock('./blocks/appLauncher/appLauncher.js'),
         loadBlock('./blocks/profile/dropdown.js'),
         loadBlock('./blocks/search/gnav-search.js'),
+        loadBlock('./utilities/keyboard/index.js'),
         loadStyles('./blocks/profile/dropdown.css'),
         loadStyles('./blocks/navDropdown/dropdown.css'),
         loadStyles('./blocks/search/gnav-search.css'),
@@ -213,6 +217,7 @@ class Gnav {
       this.ProfileDropdown = ProfileDropdown;
       this.appLauncher = appLauncher;
       this.Search = Search;
+      this.keyboardNavigation = new KeyboardNavigation();
       resolve();
     });
     return this.ready;
@@ -468,28 +473,6 @@ class Gnav {
           <div class="feds-navItem${isSectionMenu ? ' feds-navItem--section' : ''}">
             ${dropdownTrigger}
           </div>`;
-        // TODO: move proper logic to accessibility,
-        // this is just for demo functionality purposes
-        dropdownTrigger.addEventListener('click', (e) => {
-          e.preventDefault();
-
-          const openPopup = document.querySelector('.feds-navLink[aria-expanded = "true"]');
-
-          if (openPopup && openPopup !== dropdownTrigger) {
-            openPopup.setAttribute('aria-expanded', 'false');
-            openPopup.setAttribute('daa-lh', 'header|Open');
-          }
-
-          const currentState = dropdownTrigger.getAttribute('aria-expanded');
-
-          if (currentState === 'false') {
-            dropdownTrigger.setAttribute('aria-expanded', 'true');
-            dropdownTrigger.setAttribute('daa-lh', 'header|Close');
-          } else {
-            dropdownTrigger.setAttribute('aria-expanded', 'false');
-            dropdownTrigger.setAttribute('daa-lh', 'header|Open');
-          }
-        });
         delayDropdownDecoration(triggerTemplate);
         return triggerTemplate;
       }
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -217,6 +217,7 @@ class Gnav {
       this.ProfileDropdown = ProfileDropdown;
       this.appLauncher = appLauncher;
       this.Search = Search;
+      // TODO we might only want to load the keyboard navigation on keydown when it's actually used
       this.keyboardNavigation = new KeyboardNavigation();
       resolve();
     });
--- a/libs/blocks/global-navigation/utilities/keyboard/index.js
+++ b/libs/blocks/global-navigation/utilities/keyboard/index.js
@@ -1,0 +1,72 @@
+/* eslint-disable class-methods-use-this */
+import { getNextVisibleItemPosition, getPreviousVisibleItemPosition, selectors } from './utils.js';
+import MainNav from './mainNav.js';
+import { closeAllDropdowns } from '../utilities.js';
+
+const cycleOnOpenSearch = ({ e, isDesktop }) => {
+  const withoutBreadcrumbs = [
+    ...document.querySelectorAll(`
+      ${selectors.brand}, 
+      ${selectors.mainNavToggle},
+      ${selectors.mainNavItems},
+      ${selectors.searchTrigger},
+      ${selectors.searchField},
+      ${selectors.signIn},
+      ${selectors.profileButton},
+      ${selectors.logo}
+  `),
+  ];
+  const first = getNextVisibleItemPosition(-1, withoutBreadcrumbs);
+  const last = getPreviousVisibleItemPosition(withoutBreadcrumbs.length, withoutBreadcrumbs);
+  const openSearch = isDesktop && document.querySelector(selectors.openSearch);
+  if (openSearch && document.activeElement === withoutBreadcrumbs[e.shiftKey ? first : last]) {
+    e.preventDefault();
+    withoutBreadcrumbs[e.shiftKey ? last : first].focus();
+  }
+};
+
+const closeOnClickOutside = (e) => {
+  if (
+    !e.target.closest(selectors.globalNav)
+    || e.target.classList.contains(selectors.curtain.replace('.', ''))
+  ) {
+    closeAllDropdowns();
+  }
+};
+
+class KeyboardNavigation {
+  constructor() {
+    this.addEventListeners();
+    this.mainNav = new MainNav();
+    this.desktop = window.matchMedia('(min-width: 900px)');
+  }
+
+  addEventListeners = () => {
+    document.addEventListener('click', (e) => closeOnClickOutside(e));
+
+    document.querySelector(selectors.globalNav).addEventListener('keydown', (e) => {
+      switch (e.code) {
+        case 'Tab': {
+          cycleOnOpenSearch({ e, isDesktop: this.desktop.matches });
+          break;
+        }
+        case 'Enter': {
+          e.stopPropagation();
+          e.preventDefault();
+          e.target.click();
+          break;
+        }
+        case 'Space': {
+          e.stopPropagation();
+          e.preventDefault();
+          e.target.click();
+          break;
+        }
+        default:
+          break;
+      }
+    });
+  };
+}
+
+export default KeyboardNavigation;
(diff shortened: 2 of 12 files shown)


libs/blocks/global-navigation/utilities/keyboard/mainNav.js
```diff
@@ -0,0 +1,143 @@
...
+          if (document.dir === 'ltr') {
+            if (this.prev === -1) break;
+            this.focusPrev({ focus: null });
+          } else {
+            if (this.next === -1) break;
+            this.focusNext({ focus: null });
+          }
+          break;
+        }
+        case 'ArrowUp': {
```
narcis-radu, line 55: I'm seeing some weird behavior with up and down arrows, here's how to reproduce:
1. go to your after page
2. scroll down
3. focus on the last element in the nav - "no dropdown"
4. press "up" arrow 3-4 times
5. press "down" arrow 3-4 times

desired behavior: no page scrolling
actual behavior: focus is changed to previous / next elements, but pressing the arrows will introduce page scrolling as well

libs/blocks/global-navigation/utilities/keyboard/index.js
```diff
@@ -0,0 +1,84 @@
...
+  ${selectors.profileButton},
+  ${selectors.logo}
+  `),
+  ];
+  const first = getNextVisibleItem(-1, withoutBreadcrumbs);
+  const last = getPreviousVisibleItem(withoutBreadcrumbs.length, withoutBreadcrumbs);
+  const openSearch = isDesktop && document.querySelector(selectors.openSearch);
+  if (openSearch && document.activeElement === withoutBreadcrumbs[e.shiftKey ? first : last]) {
+    e.preventDefault();
+    withoutBreadcrumbs[e.shiftKey ? last : first].focus();
```
overmyheadandbody, line 24: You're using this 2 times, maybe it could be stored in a variable?

libs/blocks/global-navigation/utilities/keyboard/index.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+};
+
+const closeOnClickOutside = (e) => {
+  if (
+    !e.target.closest('header')
```
overmyheadandbody, line 30: `header` might be too generic, I'd go for `.global-navigation`.

libs/blocks/global-navigation/utilities/keyboard/index.js
```diff
@@ -0,0 +1,84 @@
...
+  constructor() {
+    this.addEventListeners();
+    this.mainNav = new MainNav();
+    this.desktop = window.matchMedia('(min-width: 900px)');
+  }
+
+  addEventListeners = () => {
+    document.addEventListener('click', (e) => closeOnClickOutside(e));
+
+    document.querySelector('header').addEventListener('keydown', (e) => {
```
overmyheadandbody, line 47: Same comment with `header` vs `.global-navigation`.

libs/blocks/global-navigation/utilities/keyboard/index.js
```diff
@@ -0,0 +1,84 @@
...
+      if (e.shiftKey && e.code === 'Tab') {
+        cycleOnOpenSearch({ e, isDesktop: this.desktop.matches });
+        return;
+      }
```
overmyheadandbody, line 51: Why can't this be integrated into the switch statement? It looks very similar to the `Tab` case.

libs/blocks/global-navigation/utilities/keyboard/index.js
```diff
@@ -0,0 +1,84 @@
...
+          e.preventDefault();
+          e.target.click();
+          break;
+        }
+        case 'Escape': {
+          const profileBtn = document.querySelector(selectors.profileButton);
+          if (e.target === profileBtn && profileBtn.getAttribute('aria-expanded') === 'true') {
+            profileBtn.setAttribute('aria-expanded', 'false');
+          }
+          break;
```
overmyheadandbody, line 67: Could this be handled in the Profile block logic directly, as we currently do with the Search block escape listeners?

libs/blocks/global-navigation/utilities/keyboard/index.js
```diff
@@ -0,0 +1,84 @@
...
+        }
+        case 'Space': {
+          e.preventDefault();
+          e.target.click();
+          break;
+        }
+        case 'ArrowDown': {
+          break;
```
overmyheadandbody, line 75: What's the purpose of this being here, since it doesn't seem to achieve anything?

libs/blocks/global-navigation/utilities/keyboard/popup.js
```diff
@@ -0,0 +1,132 @@
...
+    const popupItems = [...(getOpenPopup()?.querySelectorAll(selectors.popupItems) || [])];
+    if (!popupItems.length) return;
+    const first = getNextVisibleItem(-1, popupItems);
+    const last = getPreviousVisibleItem(popupItems.length, popupItems);
+    if (focus === 'first') popupItems[first].focus();
+    if (focus === 'last') popupItems[last].focus();
+  }
+
+  addEventListeners = () => {
+    document.querySelector('header').addEventListener('keydown', (e) => {
```
overmyheadandbody, line 49: Same comment with `header` vs `.global-navigation`; and we'll also need to adapt once we have the footer.

libs/blocks/global-navigation/utilities/keyboard/popup.js
```diff
@@ -0,0 +1,132 @@
...
+      const { popupItems, prev, next, prevColumn, nextColumn } = getState({ e });
+
+      if (e.shiftKey && e.code === 'Tab') {
+        if (prev === -1) {
+          this.mainNav.items[this.mainNav.curr].focus();
+          return;
+        }
+        popupItems[prev].focus();
+        return;
+      }
```
overmyheadandbody, line 63: Why not include this in the `switch` statement for `Tab`?

libs/blocks/global-navigation/utilities/keyboard/mobilePopup.js
```diff
@@ -0,0 +1,195 @@
+/* eslint-disable class-methods-use-this */
+import {
+  getNextVisibleItem,
+  getPreviousVisibleItem,
+  getOpenPopup,
+  selectors,
+} from './utils.js';
+import { closeAllDropdowns } from '../utilities.js';
+
+const closeHeadline = () => {
```
overmyheadandbody, line 10: Maybe `closeHeadlines`, plural, would be more exact?

libs/blocks/global-navigation/utilities/keyboard/mobilePopup.js
```diff
@@ -0,0 +1,195 @@
...
+};
+
+const openHeadline = ({ headline, focus } = {}) => {
+  closeHeadline();
+  if (headline.getAttribute('aria-haspopup') === 'true') {
+    headline.setAttribute('aria-expanded', 'true');
+    const section = headline.closest(selectors.section)
+      || headline.closest(selectors.column);
+    const items = [...section.querySelectorAll(selectors.popupItems)]
+      .filter((el) => el.offsetHeight && el.offsetWidth);
```
overmyheadandbody, line 22: Could the `isVisible` method be used here instead?

libs/blocks/global-navigation/utilities/keyboard/mobilePopup.js
```diff
@@ -0,0 +1,195 @@
...
+  const popupEl = getOpenPopup();
+  if (!popupEl) return { popupItems: [] };
+  const popupItems = [...popupEl.querySelectorAll(selectors.popupItems)];
+  // In the markup either a section OR column can contain a expandable headline
+  // which is what we are interested in - so we can treat them both as sections.
+  const section = document.activeElement.closest(selectors.section)
+    || document.activeElement.closest(selectors.column);
+  let allSections = [...popupEl.querySelectorAll(selectors.section)];
+  if (!allSections.length) allSections = [...popupEl.querySelectorAll(selectors.column)];
+  const visibleSections = allSections.filter((el) => el.offsetHeight && el.offsetWidth);
```
overmyheadandbody, line 38: Could we use `isVisible` here?

libs/blocks/global-navigation/utilities/keyboard/mobilePopup.js
```diff
@@ -0,0 +1,195 @@
...
+      this.mainNav.open();
+      return;
+    }
+
+    // Case 3: Open the next headline
+    openHeadline({ headline: nextHeadline, focus: 'first' });
+  };
+
+  addEventListeners = () => {
+    document.querySelector('header').addEventListener('keydown', (e) => {
```
overmyheadandbody, line 124: `.global-navigation` might be better than `header`

libs/blocks/global-navigation/utilities/keyboard/mainNav.js
```diff
@@ -0,0 +1,143 @@
...
+      if (!e.target.closest(selectors.fedsNav) || e.target.closest(selectors.popup)) return;
+      const open = document.querySelector(selectors.expandedPopupTrigger);
+      closeAllDropdowns();
+      if (open !== e.target) this.open({ triggerEl: e.target });
+    });
+
+    document.querySelector('header').addEventListener('keydown', (e) => {
+      if (!e.target.closest(selectors.fedsNav) || e.target.closest(selectors.popup)) return;
+      this.setActive(e.target);
+      if (e.shiftKey && e.code === 'Tab') {
```
overmyheadandbody, line 26: Could this be moved to the `switch` statement?

libs/blocks/global-navigation/utilities/keyboard/mainNav.js
overmyheadandbody: Maybe `stopPropagation` was missed here, I saw it in the popups logic.

mokimo: In the AEM implementation, only for the main nav elements the event stops doing anything - e.g. pressing down on the search will still scroll.

I'll add it that way, I think that makes the most sense, since those are the only elements that interact with the arrow keys

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -78,3 +80,27 @@ export function decorateCta({ elem, type = 'primaryCta', index } = {}) {
...
+  if (!openElements) return;
+  if (e) e.preventDefault();
+  [...openElements].forEach((el) => {
+    el.setAttribute('aria-expanded', 'false');
+    el.setAttribute('daa-lh', 'header|Open');
+  });
+  document.querySelector(curtainSelector)?.classList.remove('is-open');
+}
+
+export function openOrClose({ trigger } = {}) {
```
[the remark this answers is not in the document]
mokimo: It always calls `closeAllDropdowns()` so it's always setting aria-expanded to false on all dropdowns (including itself)

libs/blocks/global-navigation/utilities/keyboard/utils.js
```diff
@@ -0,0 +1,57 @@
...
+`;
+
+const isElementVisible = (elem) => !!(
+  elem
+    && elem instanceof HTMLElement
+    && (elem.offsetWidth || elem.offsetHeight || elem.getClientRects().length)
+    && window.getComputedStyle(elem).getPropertyValue('visibility') !== 'hidden'
+);
+
+const getNextVisibleItem = (position, items) => {
```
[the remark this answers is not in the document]
mokimo: getNextVisibleItemPosition it is!

libs/blocks/global-navigation/utilities/keyboard/index.js
mokimo: Its two different conditions, that look similar `first : last` and `last: first`.. but with extra `if's` it becomes a lot worse, so that was the best way. (Or maybe I didn't get your point and need an example 😁)

libs/blocks/global-navigation/utilities/keyboard/popup.js
mokimo: I've been on that code way too long, thanks for the second pair of eyes! It didn't even come into my mind at this point anymore. Moved them all accordingly

mokimo:
Thanks for the feedback @overmyheadandbody & @narcis-radu ! Addressed all the proposed changes

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -123,26 +122,26 @@ class Gnav {
...
         ProfileDropdown,
         Search,
+        KeyboardNavigation,
       ] = await Promise.all([
-        loadBlock('./utilities/delayed-utilities.js'),
         loadBlock('./blocks/navDropdown/dropdown.js'),
         loadBlock('./blocks/appLauncher/appLauncher.js'),
         loadBlock('./blocks/profile/dropdown.js'),
         loadBlock('./blocks/search/gnav-search.js'),
+        loadBlock('./utilities/keyboard/index.js'),
```
narcis-radu, line 135: Do we need this to be loaded as soon as possible? I was thinking maybe we can delay until needed to avoid potential performance issues.

libs/blocks/global-navigation/utilities/keyboard/index.js
```diff
@@ -0,0 +1,72 @@
...
+  const openSearch = isDesktop && document.querySelector(selectors.openSearch);
+  if (openSearch && document.activeElement === withoutBreadcrumbs[e.shiftKey ? first : last]) {
+    e.preventDefault();
+    withoutBreadcrumbs[e.shiftKey ? last : first].focus();
+  }
+};
+
+const closeOnClickOutside = (e) => {
+  if (
+    !e.target.closest('.global-navigation')
```
narcis-radu, line 30: we could use `selectors.globalNav`, the value is available here https://github.com/adobecom/milo/blob/keyboard-support/libs/blocks/global-navigation/utilities/keyboard/utils.js#L24

libs/blocks/global-navigation/utilities/keyboard/index.js
```diff
@@ -0,0 +1,72 @@
...
+  constructor() {
+    this.addEventListeners();
+    this.mainNav = new MainNav();
+    this.desktop = window.matchMedia('(min-width: 900px)');
+  }
+
+  addEventListeners = () => {
+    document.addEventListener('click', (e) => closeOnClickOutside(e));
+
+    document.querySelector('.global-navigation').addEventListener('keydown', (e) => {
```
narcis-radu, line 47: `selectors.globalNav`

JasonHowellSlavin (approved):
Lots of great work in here, but was there no way to break it up into multiple PRs? It can be somewhat daunting to code review a PR of this size and be sure of the efficacy of the review.

mokimo:
@JasonHowellSlavin definitely agree, this grew too big! I'll make sure to raise smaller incremental PRs in the future

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -1,5 +1,8 @@
 import { getConfig, localizeLink } from '../../../utils/utils.js';
 
+const curtainSelector = '.feds-curtain';
+const navLinkClassName = 'feds-nav-link';
+const globalNavSelector = '.global-navigation';
```
narcis-radu, line 5: Can't we use the selectors from utils?

mokimo: Yes we could, but the utilities are "sync" (e.g. get loaded right away with the global-navigation)
The utils are delayed until we load the keyboard-navigation (3s or on interaction)

We can consider moving the selectors to the more general utilities file though

narcis-radu: Got it, no big deal, we can explore afterwards.

honstar (approved):
LGTM

honstar:
@overmyheadandbody , would you mind having a final look?

overmyheadandbody (approved):
Looking good!

libs/blocks/global-navigation/utilities/keyboard/mainNav.js
```diff
@@ -0,0 +1,156 @@
...
+    this.desktop = window.matchMedia('(min-width: 900px)');
+    this.popup = new Popup({ mainNav: this });
+    this.mobilePopup = new MobilePopup({ mainNav: this });
+    this.addEventListeners();
+  }
+
+  addEventListeners() {
+    document.querySelector(selectors.globalNav)
+      .addEventListener('click', (e) => {
+        if (!e.target.closest(selectors.fedsNav) || e.target.closest(selectors.popup)) {
```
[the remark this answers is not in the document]
mokimo: Nope we need both, giving it a quick test makes a lot of the test cases fail

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -195,16 +196,19 @@ class Gnav {
     // eslint-disable-next-line no-async-promise-executor
     this.ready = this.ready || new Promise(async (resolve) => {
       this.el.removeEventListener('click', this.loadDelayed);
+      this.el.removeEventListener('keydown', this.loadDelayed);
       const [
         decorateDropdown,
         { appLauncher },
         ProfileDropdown,
         Search,
+        KeyboardNavigation,
```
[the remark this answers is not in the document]
mokimo: adding the TODO is worth it to keep it on our radar, it's a big feature and it makes sense to only load it when required

PR #646, opened 2023-04-19
Address Viktor's gnav PR feedback

mokimo:
Address some of the PR feedback in https://github.com/adobecom/milo/pull/601

**Test URLs:**
- After: https://feedback-gnav--milo--mokimo.hlx.page/drafts/ramuntea/gnav-refactor?martech=off

### DIFF
--- a/libs/blocks/global-navigation/utilities/utilities.js
+++ b/libs/blocks/global-navigation/utilities/utilities.js
@@ -18,36 +18,21 @@ export function toFragment(htmlStrings, ...values) {
   return fragment;
 }
 
-export function debounceCallback(callback, time = 200) {
-  if (typeof callback !== 'function') return undefined;
-
-  let timeout = null;
-
-  return (...args) => {
-    clearTimeout(timeout);
-    timeout = setTimeout(() => callback(...args), time);
-  };
-}
-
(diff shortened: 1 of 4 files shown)


libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -18,36 +18,21 @@ export function toFragment(htmlStrings, ...values) {
...
-  const { locale, miloLibs } = getConfig();
+  const { locale, miloLibs, env } = getConfig();
   let libOrigin = 'https://adobe.com/libs';
   if (window.location.origin.includes('localhost')) {
     libOrigin = `${window.location.origin}/libs`;
   }
+
   if (window.location.origin.includes('.hlx.')) {
-    libOrigin = miloLibs?.replace('hlx.live', 'hlx.page') || 'https://main--milo--adobecom.hlx.page/libs';
+    const baseMiloUrl = env.name === 'prod'
```
overmyheadandbody, line 30: Small nit: why always compute the value if it is only the second option? Maybe this could be a function and L33 could be something like `libOrigin = miloLibs || getBaseMiloUrl();`, thus you compute the value only if needed.

overmyheadandbody (approved):

narcis-radu (approved):

PR #658, opened 2023-04-24
Card block authoring compatablity

ryanmparrish:
* Fixed the Cards block to work with "two up, three up, four up, five up" section-metadata styles so it operates the same as all the other blocks in `two-five up` sections. 

Adding any of the style tokens "two up, three up, four up, five up" work for all blocks in a section EXCEPT the Card block. The Card block only works with "2 up, 3 up, 4 up" etc. The Card block also gets auto columns if nothing is specified (3 columns).

Having two different naming conventions for this creates a lot of confusion and questions from authors.

Resolves: [MWPW-129763](https://jira.corp.adobe.com/browse/MWPW-129763)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rparrish/columns-test?martech=off
- After: https://rparrish-card--milo--adobecom.hlx.page/drafts/rparrish/columns-test?martech=off


mokimo:
The _before_ is broken, do you just want to compare it to the main branch or `main-card`?

libs/blocks/card/card.js
```diff
@@ -32,7 +32,15 @@ const addWrapper = (el, section, cardType) => {
...
-  const upClass = getUpFromSectionMetadata(section);
+  let upClass = getUpFromSectionMetadata(section);
+  // Authored w/ a typed out number reference... 'two-up' vs. '2-up'
+  const list = ['two-up', 'three-up', 'four-up', 'five-up'];
+  const ixd = list.findIndex(i => i.includes(upClass));
+  if (ixd > -1) {
+    upClass = `${ixd+2}-up`;
+    section.classList.remove(list[ixd]);
+    section.classList.add(upClass);
+  }
```
mokimo, line 43: This should have some test.

libs/blocks/card/card.js
```diff
@@ -32,7 +32,15 @@ const addWrapper = (el, section, cardType) => {
 
   if (prevGrid) return;
 
-  const upClass = getUpFromSectionMetadata(section);
+  let upClass = getUpFromSectionMetadata(section);
+  // Authored w/ a typed out number reference... 'two-up' vs. '2-up'
+  const list = ['two-up', 'three-up', 'four-up', 'five-up'];
+  const ixd = list.findIndex(i => i.includes(upClass));
+  if (ixd > -1) {
+    upClass = `${ixd+2}-up`;
```
mokimo, line 40: Given L44 `const up = upClass?.replace('-', '') || '3up';` shouldn't we set this without the  `-` to begin with?

libs/blocks/card/card.js
```diff
@@ -32,7 +32,15 @@ const addWrapper = (el, section, cardType) => {
...
+    section.classList.remove(list[ixd]);
+    section.classList.add(upClass);
```
mokimo, line 42: Why do we add `3-up` to the section? it seems on the gridClass on L45 sets the class

libs/blocks/card/card.js
ryanmparrish: Good call. I've added one

libs/blocks/card/card.js
ryanmparrish: Also un-necessary. I've removed this .add() redundancy. Thanks

ryanmparrish:
@mokimo - I've added those extra 'four-up' & 'five-up' examples in the doc and fixed the link on the 'before'. 
Thanks for the feedback.

mokimo (approved):
Cool 👍

libs/blocks/card/card.js
```diff
@@ -32,7 +32,14 @@ const addWrapper = (el, section, cardType) => {
 
   if (prevGrid) return;
 
-  const upClass = getUpFromSectionMetadata(section);
+  let upClass = getUpFromSectionMetadata(section);
+  // Authored w/ a typed out number reference... 'two-up' vs. '2-up'
+  const list = ['two-up', 'three-up', 'four-up', 'five-up'];
+  const ixd = list.findIndex(i => i.includes(upClass));
```
meganthecoder, line 38: What does `ixd` stand for?

meganthecoder (approved):
Glad to see this improvement!

libs/blocks/card/card.js
honstar: I feel this should be `idx` short for index :)

ryanmparrish: Oops. my bad on the typo. We'll clean that up in a future task. Thanks

PR #660, opened 2023-04-25
MWPW-130197 remove ost from AUTO_BLOCKS

VKniaz:
remove ost from AUTO_BLOCKS

Resolves: [MWPW-130197](https://jira.corp.adobe.com/browse/MWPW-130197)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ilyas/merch#
- After: https://mwpw-130197--milo--adobecom.hlx.page/drafts/ilyas/merch#

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -64,7 +64,6 @@ const AUTO_BLOCKS = [
   { gist: 'https://gist.github.com' },
   { caas: '/tools/caas' },
   { faas: '/tools/faas' },
-  { ost: '/tools/ost' },
   { fragment: '/fragments/' },
   { instagram: 'https://www.instagram.com' },
   { slideshare: 'https://www.slideshare.net' },


yesil (approved):

yesil:
@chrischrischris we would like to merge this asap as this is breaking commerce now, thanks for your help.

mokimo:
@VKniaz cool to see you starting to contribute more and more! 

One improvement around PR sanity would be adding Test URLs https://github.com/adobecom/milo/pull/631 so that
1) Lighthouse checks can run
2) The reviewer has a page where the fix can be tested on

vhargrave (changes requested):
I think we need to take another look at this 
this does fix the bug with the merch block, but I think it'll break the OST. 
Whenever I try to test https://mwpw-130197--milo--adobecom.hlx.page/tools/ost I keep getting redirected. 
@Axelcureno can you take a look?

yesil:
@vhargrave the redirection is not a consequence of this fix.  The redirection is needed for the IMS.
We need to unblock commerce.

vhargrave:
> @vhargrave the redirection is not a consequence of this fix. The redirection is needed for the IMS. We need to unblock commerce.

@yesil as said, I think this PR will break the OST. We need to make sure that doesn't happen.

vhargrave (approved):
After taking a proper look it should be fine. The OST block is not provided outside of milo, hence why it doesn't need to be declared 👍

PR #661, opened 2023-04-25
Add 'alternating' class name for z-pattern block and ambient video support for media blocks

drukelly:
* This PR features an update to the z-pattern by adding an 'alternating' class name to the z-pattern block.
* From an authoring POV, it rids having to rearrange the media + content copy to render the layout similar to what's shown on the After test URL.
* Tested on multiple viewports: Desktop, Tablet (Portrait and Landscape orientations), and Mobile.
* A current A.com use case can be seen past the marquee and Adobe Express blade: https://www.adobe.com/products/photoshop.html

Adds to: https://jira.corp.adobe.com/browse/MWPW-104749

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/druk/z-pattern-test?martech=off
- After: https://main--milo--drukelly.hlx.page/drafts/druk/z-pattern-test?martech=off


auniverseaway:
Thanks for the contribution!

Please take a moment to read: https://github.com/adobecom/milo/wiki/Submitting-PRs

rgclayton (changes requested):
Thanks Dru! I've outlined a few concerns.

ryanmparrish:
@drukelly  - Thanks for adding me in this review. We already offer an auto 'alternating' feature on the z-pattern if the document is structured like in the docs w/ the asset in the first table column and copy in the second. 
See this test page -> https://main--milo--adobecom.hlx.page/drafts/rparrish/zpattern-test

Also, I noticed your doc has a `light` variant. This likely isn't necessary.

libs/blocks/media/media.js
```diff
@@ -18,6 +18,14 @@ import { decorateBlockBg, decorateBlockText, getBlockSize } from '../../utils/de
...
 import { createTag } from '../../utils/utils.js';
 
+const decorateVideo = (container) => {
+  const link = container.querySelector('a[href$=".mp4"]');
+
+  container.innerHTML = `<video preload="metadata" playsinline autoplay muted loop>
+    <source src="${link.href}" type="video/mp4" />
+  </video>`;
+};
+
```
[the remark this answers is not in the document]
ryanmparrish: I agree, It might be better to pu this in a shared function `decorateVideoAsset()` or something in `utils/decorate.js` and then include it as a video asset decorator in media and marquee. General rule is if a util is needed across multiple blocks, create an import func

libs/blocks/z-pattern/z-pattern.css
```diff
@@ -41,3 +41,58 @@
...
 }
+
+/* Alternating Z-Pattern riverflow with Ambient Video */
+.ambient-video > video {
+  object-fit: fill;
+}
+
+@media screen and (min-width: 600px) {
+  .z-pattern.alternating > .media:nth-child(odd) > .container > .media-row {
+    flex-direction: row-reverse;
```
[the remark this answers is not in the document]
ryanmparrish: See the func `getReversedRowCount()` in 'z-pattern.js' to see how we're already doing this w/ the class 'media-reversed`. LMK if you have any questions.

libs/blocks/media/media.js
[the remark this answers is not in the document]
JasonHowellSlavin: +1 for placing in decorate JS. 

I wonder if we also wanted to take some of the authoring options for autoplay that might be useful here that are in `video/video.js`?

libs/blocks/media/media.js
```diff
@@ -46,7 +54,13 @@ export default function init(el) {
       decorateBlockText(text, blockTypeSizes[size]);
     }
     const image = row.querySelector(':scope > div:not([class])');
-    if (image) image.classList.add('image');
+    const video = row.querySelector(':scope > div > a[href$=".mp4"]');
+    if (video) {
+      decorateVideo(video.closest('div'));
+      image.classList.add('ambient-video');
+    } else {
```
JasonHowellSlavin, line 61: To keep current functionality, we would want to change this to an `else if (image) {...` since on line 49 we check for the image first before modifying `classList`

libs/blocks/z-pattern/z-pattern.css
```diff
@@ -41,3 +41,58 @@
...
+  }
+}
+
+/* Mobile: Ambient Video */
+@media screen and (max-width: 767px) {
+  .ambient-video > video {
+    width: 100%;
+    height: auto;
+  }
+}
```
JasonHowellSlavin, line 98: Not sure about the media query breakpoints here. The most commonly used breakpoints for media queries in Milo are `600 900 1200 and 1440`. `768` and `1024` seem very specific to device widths. @ryanmparrish @rgclayton is this something we should standardize throughout Milo? I am assuming those are consonant breakpoints.

ryanmparrish: Thats correct. We should probably keep to these unless we have a reason to need other ones. 

```
/* mobile only */
@media (max-width: 600px) {}

/* Tablet up */
@media screen and (min-width: 600px) {}

/* desktop up */
@media screen and (min-width: 1200px) {}

/* desktop large < 1440 */
@media screen and (min-width: 1440px) {}
```

ryanmparrish: It might be a good idea to move these media query values into the --variable, but maybe thats for another day. :)

libs/blocks/z-pattern/z-pattern.css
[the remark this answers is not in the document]
drukelly: Through this exercise, I discovered that `z-pattern (large)` achieves the desired layout I was looking for! Thank you @ryanmparrish for your guidance! I've created a separate test page as a reference: https://main--milo--adobecom.hlx.page/drafts/druk/z-pattern-test-v2

libs/blocks/z-pattern/z-pattern.css
ryanmparrish: Ah, I was just thinking the value. So 
```
@media screen and (min-width: var(--breakpoint-large)) {}
```
Looks kind of ugly and unnecessary though.

rgclayton:
> @rgclayton @ryanmparrish does this work accessibility / tab index wise?
> 
> I'd also like if these two things were separate PRs... 1x JIRA, 1x Commit, 1x PR.
> 
> This seems like two things: z-index quality of life and video stuff.

I don't think we need the alternating update that the title of this PR description mentions. @ryanmparrish mentioned that already exists or this can be authored to fit this use case. 

@drukelly do you want to update this PR to remove that alternating code and address the other issues Ryan P and Jason mentioned? 

I think that will clean this PR to only address the ambient video addition. Does that sound correct?

Related to accessibility. Videos longer than five seconds that loop should have some sort of pause or stop controls.

1. https://stackoverflow.com/questions/59848343/accessibility-rules-for-generic-lifestyle-background-video-with-no-audio-as-it-r
2. https://accessibleweb.com/question-answer/do-i-need-video-controls-for-background-videos/

libs/blocks/z-pattern/z-pattern.css
drukelly: I'll remove the `.ambient-video` CSS on z-pattern.css

libs/blocks/media/media.js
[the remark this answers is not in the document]
drukelly: I'll remove the decorateVideo function off the media.js file and update video.js to add on attributes to help accessibility, i.e. `aria-hidden="true"`

drukelly:
Per @ryanmparrish, I was able to create the alternating z-pattern build without the proposed milo modifications:
- https://main--cc--adobecom.hlx.page/drafts/druk/photoshop-riverflow-z-pattern

rgclayton:
@drukelly there's a bunch of 404's for you .mp4's on your test page.

drukelly:
@rgclayton an earlier iteration of my test page had images-tv.adobe.com mp4 links. I've updated them per this [documentation](https://milo.adobe.com/docs/authoring/videos) re: AdobeTV videos tab. Not sure why the 3rd ambient video is the only one rendering here: https://main--milo--adobecom.hlx.page/drafts/druk/z-pattern-test

1. https://main--cc--adobecom.hlx.page/media_1073508e08745368364dd4ea74f9fbc340c2564d0.mp4#_autoplay
2. https://main--cc--adobecom.hlx.page/media_199df46e9da7fe7c077f6fc01f365799b73e297ad.mp4#_autoplay
3. https://main--cc--adobecom.hlx.page/media_17ce8fd60c6a6ef9b5b8a0f1086183825006db811.mp4#_autoplay
4. https://main--cc--adobecom.hlx.page/media_1bc9aabf5ac7dd733d4ce157e1b950e7aca4bf69f.mp4#_autoplay
5. https://main--cc--adobecom.hlx.page/media_17cd00319df37df637a98c2b017e3192153d187d9.mp4#_autoplay

rgclayton:
> @rgclayton an earlier iteration of my test page had images-tv.adobe.com mp4 links. I've updated them per this [documentation](https://milo.adobe.com/docs/authoring/videos) re: AdobeTV videos tab. Not sure why the 3rd ambient video is the only one rendering here: https://main--milo--adobecom.hlx.page/drafts/druk/z-pattern-test

@drukelly You can't host videos out of CC and play them on Milo pages. You either need to create a CC related test page or use videos that are hosted from Milo.

auniverseaway:
@rgclayton @drukelly what's up with this PR?

auniverseaway:
Closing. Stale. Can re-open.

PR #664, opened 2023-04-26
MWPW-129798-Fix image stretch issue in fragment

Ruchika4:
* Fixes the issue of images in fragments getting stretched on a page

Resolves: [MWPW-129798](https://jira.corp.adobe.com/browse/MWPW-129798)

**Test URLs:**
Milo

- Before: https://main--milo--adobecom.hlx.page/drafts/ruchika/fragment-test?martech=off
- After: https://mwpw-129798-image-stretch-issue--milo--adobecom.hlx.page/drafts/ruchika/fragment-test?martech=off

CC

- Before: https://main--cc--adobecom.hlx.page/drafts/tkato/test-cc-overview?martech=off
- After: https://main--cc--adobecom.hlx.page/drafts/tkato/test-cc-overview?milolibs=MWPW-129798-image-stretch-issue&martech=off

### DIFF
--- a/libs/styles/styles.css
+++ b/libs/styles/styles.css
@@ -137,7 +137,7 @@ img {
 
 /* Not blocks */
 div:not([class]) > div > div img[src^="./media_"] {
-  width: 100%;
+  max-width: 100%;
 }
 
 .container {


meganthecoder:
I'm interested to know why the `width: 100%;` was added in the first place. We should make sure that we're not breaking the  original use case.

mokimo (approved):

honstar (approved):

libs/styles/styles.css
```diff
@@ -137,7 +137,7 @@ img {
 
 /* Not blocks */
 div:not([class]) > div > div img[src^="./media_"] {
-  width: 100%;
+  max-width: 100%;
```
overmyheadandbody, line 140: I think it would be valuable to get a bit more context on some use-cases around this, because the two can have quite different effects. A quick CodePen as a simulation - https://codepen.io/ramuntea/pen/eYPRJRe

If the image is larger than its wrapper, then the two declarations have the same effect. If, however, the image is smaller than its wrapper (which could have a set width), then the two declarations have different outcomes.

Ruchika4: @overmyheadandbody this issue is particularly about images in fragment. When previewed in fragment they load in their actual size. However with width: 100% when these fragments were added on the page, these images were getting stretched and distorted (see the before url). I did tried to add the specific selector like below and add max-width

```
/* fragment image */
.fragment > * img[src^="./media_"] {
  max-width: 100%;
}
```

but then both the css (the one for fragment and the more generic one too) got applied and image got stretched. For example the yellow flower image width stretched to 1274px from 183px
Screen Shot 2023-04-27 at 2 33 45 PM

Ruchika4:
Closing this PR. Discussed this with @auniverseaway and this issue is fixed now.

PR #667, opened 2023-04-27
Improve unit test output

mokimo:
### Description
No ticket / no test URLs as this purely touches unit tests

### Changes
- only include relevant folders for code coverage (and thus, exclude all 3rd party scripts)
- Improve the current web test runner setup & strip a few images from mocks that lead to 404s in the test runner logs

Reduces log output from unit tests from **1000 lines to 400** (tested on local)

### More context
We can be even better, but it would require looking into the info/errors a bit more deeply. Ideally a test shouldn't ever log anything, most notable candidates are:
- 404 network requests
- JS errors
- Promises being thrown (that might require mocking a bit more stuff...)

Having on eye on that allows to spot errors a lot quicker when running the tests on bulk for example on PRs if a test fails.


honstar (approved):
Definitely +1 to get rid of the superfluous 404s 👍

mokimo:
There's still some left, but it's a first step!

JasonHowellSlavin (approved):

overmyheadandbody (approved):

sukamat (approved):

PR #670, opened 2023-04-27
con-button styles have bad defaults

ryanmparrish:
The default `con-button` looks invisible without a `outline`, or `blue` style being set on the class list. 

* Updated the `con-button` default styles to match `outline` and removed any redundant styles
* Add to the `typography.css` file to support `aria-disabled="true"` styles and fixed sort order/css-linting

Resolves: [MWPW-130320](https://jira.corp.adobe.com/browse/MWPW-130320)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rparrish/buttons
- After: https://rparrish-button-default--milo--adobecom.hlx.page/drafts/rparrish/buttons

### DIFF
--- a/libs/styles/typography.css
+++ b/libs/styles/typography.css
@@ -162,8 +162,8 @@
 }
 
 .dark .con-button {
-  border-color: #fff;
-  color: #fff;
+  border-color: var(--color-white);
+  color: var(--color-white);
 }
 
 .light .con-button {
@@ -179,15 +179,15 @@
 }
 
 .dark .con-button:hover {
-  background-color: #fff;
-  color: #000;
+  background-color: var(--color-white);
+  color: var(--color-black);
   text-decoration: none;
 }
 
 .con-button.blue {
   background: var(--color-accent);
   border-color: var(--color-accent);
-  color: #FFF;
+  color: var(--color-white);
 }
 
 .con-button.blue:hover,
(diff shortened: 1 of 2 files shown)


libs/styles/typography.css
```diff
@@ -154,14 +161,33 @@
   width: 100%;
 }
 
+.dark .con-button {
+  border-color: #fff;
+  color: #fff;
```
overmyheadandbody, line 166: Could `--color-white` not be used here as well?

overmyheadandbody (approved):

libs/styles/typography.css
ryanmparrish: yea, might as well. I updated em. Thanks

Sartxi (approved):

yesil (approved):

yesil:
@ryanmparrish unrelated to this PR but the italic button doesn't look italic:

image

rgclayton (approved):
Thanks @ryanmparrish for the quick turn around.

Looks got be solving for https://rparrish-button-default--milo--adobecom.hlx.page/drafts/ilyas/merch

Checked on consumer sites too. Not seeing any issues.

rgclayton:
> @ryanmparrish unrelated to this PR but the italic button doesn't look italic:
> 
> image

@yesil The italic variant is the outline. You won't see italic text in the button. Bold created the sold blue button.

PR #671, opened 2023-04-27
Table sticky

vgoodric:
* removed code for highlight from table-metadata block and moved it to the table block
* removed some temp CSS from table.css. It was causing issues to leave it in and I don't think it was meant to be permanent.
* supports first row being left empty if the author wants to keep it as a placeholder to add a highlight row later
* highlight is now a variant of the table block.  If highlight is on, the first row with content is treated as the highlight row. The 2nd row is assumed to be the header row. Otherwise, if highlight is off, the first row with content is the header row.
* sticky top added as a variant of table block. Locks header row and highlight row (if it exists). Added default background color of white so you cannot see the content of the lower rows passing behind the words.
* update to ESLint to work on windows

Resolves: [MWPW-129574](https://jira.corp.adobe.com/browse/MWPW-129574)

**Test URLs:**
- Before: https://tableblock--milo--adobecom.hlx.page/drafts/vgoodric/sticky-top?martech=off
- After: https://table-sticky--milo--adobecom.hlx.page/drafts/vgoodric/sticky-top?martech=off

### DIFF
--- a/libs/blocks/table-metadata/table-metadata.js
+++ b/libs/blocks/table-metadata/table-metadata.js
@@ -1,26 +1,11 @@
-import { getMetadata, handleStyle } from '../section-metadata/section-metadata.js';
+import { getMetadata } from '../section-metadata/section-metadata.js';
 
 const getIndexedValues = (text) => text.split('\n').map((value) => value.split(/,(.*)/s).map((v) => v.trim()));
-
-function handleHighlight(text, table) {
-  if (!text) return;
-  const highLights = getIndexedValues(text);
-  highLights.forEach((hl) => {
-    const hlCol = hl[0];
-    const hlText = hl[1];
-    if (hlText) {
-      const highLightDiv = document.createElement('div');
-      highLightDiv.textContent = hlText;
-      table.querySelector(`.col-${hlCol}`).classList.add('col-highlight');
-      table.querySelector(`.col-${hlCol}`).prepend(highLightDiv);
-    }
-  });
-}
 
 function handleSectionHead(text, table) {
   if (!text) return;
   const sectionsHeads = text.split('\n');
-  sectionsHeads.forEach(sh => {
+  sectionsHeads.forEach((sh) => {
     const sectionHead = table.querySelector(`.row-${sh.trim()}`);
     sectionHead.classList.add('sectionHead');
   });
@@ -49,7 +34,7 @@ const handleColumnBgColor = (text, table, columnType) => {
 
 function handleCompare(text, table) {
   if (!(text || table)) return;
-  const comparisonGroup  = text.split('\n');
+  const comparisonGroup = text.split('\n');
   comparisonGroup.forEach((comp, i) => {
     const col = comp.trim().split(' ')[1];
     const comparable = table.querySelector(`.col-${col}`);
@@ -62,7 +47,6 @@ export default function init(el) {
   if (!table) return;
 
   const metadata = getMetadata(el);
-  if (metadata.highlight) handleHighlight(metadata.highlight.text, table);
   if (metadata.section) handleSectionHead(metadata.section.text, table);
   if (metadata.compare) handleCompare(metadata.compare.text, table);
   if (metadata['heading color']) handleColumnColor(metadata['heading color'].text, table, 'heading');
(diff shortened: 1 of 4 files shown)


honstar:
I find the experience a bit strange when the sticky top section of a table suddenly disappears when pushed up by a following table/block when scrolling down. Can this be changed to have a smoother transition somehow?

vgoodric:
@honstar I agree it's awkward, but I feel that's because the sticky header should never be so tall.  If you look at the last table with a small sticky header, I think it works great. 1 idea I had was to auto-generate compact headers from the first line in the header with text.  If I use the first text line as the assumed title, they'd always have a compact sticky row.  But I can't think of a way to make the large sticky headers less awkward.  I'm down for concrete suggestions if anyone has them though.

vgoodric:
Spoke with Sean and he feels we should leave it for now

libs/blocks/table-metadata/table-metadata.js
```diff
@@ -62,7 +47,6 @@ export default function init(el) {
   if (!table) return;
 
   const metadata = getMetadata(el);
-  if (metadata.highlight) handleHighlight(metadata.highlight.text, table);
```
honstar, line 65: Do existing tables need to be re-authored to account for this change, `highlight` being removed from table metadata and moved to the `table` block directly?

vgoodric: I'm not positive I understand your question.  But I think you're asking if the template of having a blank row for a highlight would still be in the template people copy from the library.  I would say you could leave the row blank, but it would make more sense to put placeholder text. If you're asking about the sample tables you may have on your test pages, yes.  If you're sample tables have a highlight row, you need to do 2 things to make them still work:
1. Move the text of the highlights from the meta data table block to the row/column of the table block where you want them
2. Add the highlight variant to the table block

If this block was already live and in use, this would be an issue.  But since the only people using it are the devs working on this block, now is the time to make this improvement.

honstar: > If this block was already live and in use, this would be an issue. But since the only people using it are the devs working on this block, now is the time to make this improvement.

That was actually my question but then I checked that there currently is no `table` block yet, so all good. Thanks for the info 👍

rgclayton:
>@honstar I agree it's awkward, but I feel that's because the sticky header should never be so tall. If you look at the last table with a small sticky header, I think it works great. 1 idea I had was to auto-generate compact headers from the first line in the header with text. If I use the first text line as the assumed title, they'd always have a compact sticky row. But I can't think of a way to make the large sticky headers less awkward. I'm down for concrete suggestions if anyone has them though.

I agree with @honstar. This is a jarring user experience. 

It does work better for the smaller header - even that looks like it could be refined a bit, it also jumps if you aren't scrolling very fast. 

Doesn't the design requirements call out the larger header being sticky? I think if this is a problem it needs to be communicated with design and a better solution is provided. Or, there needs to be a solve to make the larger header less jumpy.

@ryanmparrish Any thoughts?

ryanmparrish:
I agree that the sticky part of the header shouldn't ever be that tall and from a UX it would be better if the mega header collapsed (seamlessly) to a condensed 'headline only' view when it's stuck. I wonder if we could infer a hidden sticky sub-header from the content and use that as the stuck label when in that state?
I understand in a real world scenario the compare table content will be quite large so there likely won't be as much of a glaring issue so it might be ok as is but my gut would be to account for that on a UX level and set the pattern for sticky table headers to use minimal content where possible.
And yea. The mobile views are especially hard to manage w/ the limited screen space.

vgoodric:
@honstar @rgclayton @ryanmparrish @seanchoi-dev 
So I put a bunch of thought into how to make it more gradual, and I think the most natural choice is that it is sticky until only one more row is visible.  Then further scrolling starts to scroll the header and highlight off the screen.

Also, for mobile view, I have a suspicion this code will work once the mobile view is built.  But right now, the mobile view just stacks each column on top of each other with only one per.  The final design is supposed to hide excess columns and keep visible headers side by side.  So the sticky header can't work correctly until that is coded.

libs/blocks/table/table.js
```diff
@@ -1,11 +1,99 @@
+const positionStickyRows = (table) => {
+  const tableRec = table.getBoundingClientRect();
+  const tableBottom = tableRec.top + tableRec.height;
+  const gnav = document.querySelector('header');
+  const gnavHeight = gnav ? document.querySelector('header').getBoundingClientRect().height : 0;
```
ryanmparrish, line 5:
```suggestion
  const gnavHeight = gnav ? gnav.getBoundingClientRect().height : 0;
```

ryanmparrish:
@vgoodric  I think that sounds good. Stick till the last item then scroll away with it and don't worry about the <th> row height. Let us know what shakes w/ that and if you need anything.

seanchoi-dev (approved):

PR #673, opened 2023-04-28
MWPW-127384 Ability to anchor/focus background image at certain point…

suhjainadobe:
The PR gives the ability to set the anchor/focus point of background images for the marquee and for section metadata blocks so that the focal point of the imagery is always visible.

Resolves: [MWPW-127384](https://jira.corp.adobe.com/browse/MWPW-127384).

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/suhjain/background-image-test?martech=off
- After: https://focalpoint--milo--suhjainadobe.hlx.page/drafts/suhjain/background-image-test?martech=off


suhjainadobe:
Closing the PR

rgclayton:
Why do we have multiple PR's open for this?

I just finished commenting on this newer PR
https://github.com/adobecom/milo/pull/685

suhjainadobe:
@rgclayton : this PR is specifically for section-metadata and https://github.com/adobecom/milo/pull/685 is for the marquee.

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -1,17 +1,28 @@
 function handleBackground(div, section) {
-  const pic = div.querySelector('picture');
+  const pic = div.background.content.querySelector('picture');
   if (pic) {
     section.classList.add('has-background');
     pic.classList.add('section-background');
+    if (div.focalpoint) {
```
rgclayton, line 6: I think this should be a little more hardened. If an author enters "focal point" this breaks.

rgclayton:
Anyone have any opinions about "focalpoint" vs "focuspoint" ?

@ryanmparrish @chrischrischris @auniverseaway

rgclayton:
I noticed that since the background image for `section-metadata` is set to `object-fit: cover` the image is _always_ stretched the full width. Which means that the left/right settings for this don't change the image position. Only the vertical access gets changed.

Reference:
https://github.com/adobecom/milo/blob/main/libs/blocks/section-metadata/section-metadata.css#L72

Do we want to handle the background position different with this focal point setting so authors can adjust the horizontal access?

cc: @auniverseaway @ryanmparrish @chrischrischris @JasonHowellSlavin

JasonHowellSlavin:
> I noticed that since the background image for section-metadata is set to object-fit: cover the image is always stretched the full width[0]...Do we want to handle the background position different with this focal point setting so authors can adjust the horizontal access?

I don't particularly like authors directly authoring CSS properties. Is there an issue with just adding classes within the section-metadata that will target the background image and add the focus points we want? It would also likely allow for an override of object-fit, which would solve for the issue[0] you raised @rgclayton. 

Seems simpler to add four CSS classes like `focus-upper-left, focus-upper-right` etc. than to add a new key value pair to section metadata and two value words.

JasonHowellSlavin:
Apologies, did not mean to close the PR.

rgclayton:
> Seems simpler to add four CSS classes like `focus-upper-left, focus-upper-right` etc. than to add a new key value pair to section metadata and two value words.

If that's the case, something more simple, and I think how @suhjainadobe originally had it  `left, top`, `right, bottom` etc. Not a fan of long, multiple word key words as they are much easier to mess up.

suhjainadobe:
> Seems simpler to add four CSS classes like `focus-upper-left, focus-upper-right` etc. than to add a new key value pair to section metadata and two value words.

The authors will have to remember the specific keywords for adding the classes which might be an issue if they miss out on something.

rgclayton:
I just had an idea. I think it makes more sense to group the values with the background image - similar to how you are doing it for the Marquee block. The values are directly under the background image and we remove the `focalPoint` property.


image

I also think the background image property `object-fit: cover` needs to be addressed. With it currently set to `cover` the image is always full-width and will never adjust on the horizontal (x) axis. Which, will cause confusion for authors not being able to adequately adjust the position of the image.

JasonHowellSlavin:
> I just had an idea. I think it makes more sense to group the values with the background image - similar to how you are doing it for the Marquee block. The values are directly under the background image and we remove the `focalPoint` property.
> 
> image
> 
> I also think the background image property `object-fit: cover` needs to be addressed. With it currently set to `cover` the image is always full-width and will never adjust on the horizontal (x) axis. Which, will cause confusion for authors not being able to adequately adjust the position of the image.

I think that makes sense, and you could perhaps utilize css selectors to address and override the `object-fit: cover` issue. 

So the author just has to put `top, left` or `left, top` or whatever, and programmatically we could add whatever class name fits that. 

I think what you propose is also what Chris may have meant in this comment on the ticket: 
Screen Shot 2023-05-19 at 12 44 03 PM

salonijain3:
@suhjainadobe I believe this is what we discussed as well, what is the confusion here?

echampio-at-adobe:
Just from a functional level, the examples in the test doc don't quite seem to be working. 

https://focalpoint--milo--suhjainadobe.hlx.page/drafts/suhjain/background-image-test?martech=off

On Chrome on Mac Desktop, at least, the second example to ping the pink dot to the upper-right corner is the only one that really works for me when resizing the page. In  first example, the pink dot in the lower-left generally isn't visible for me. The third example with the centered pink dot also generally is lost offscreen to the left when resizing.

suhjainadobe:
> I also think the background image property `object-fit: cover` needs to be addressed. With it currently set to `cover` the image is always full-width and will never adjust on the horizontal (x) axis.

@rgclayton  @JasonHowellSlavin : I think object-fit:covers trim the image to maintain the aspect-ratio in different viewports. When we give object-position: left for eg, it will focus towards left and cut the right side image more to keep the aspect-ratio.  It works in all directions in the same way. Also, I will make the changes in the section metadata similar to the marquee to keep it uniform.

suhjainadobe:
@rgclayton @JasonHowellSlavin : I have made the changes for the authoring. Included the marquee block as well here as the functionality for both blocks was the same. I will close #685 as it is being included in this PR only. Kindly review the same.

suhjainadobe:
@echampio-at-adobe: Could you please take a look into it now? There were authoring changes that did not focus the image properly.

libs/blocks/section-metadata/section-metadata.js
suhjainadobe: This approach has been removed. Directly the value for focal point would be given under the image.

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -1,17 +1,32 @@
...
+  let text = '';
+  if (child.childElementCount == 2) {
+    text = child.querySelectorAll('p')[1].textContent;
+  } else if (child.textContent) {
+    text = child.textContent;
+  }
+  const image = pic.querySelector('img');
+  const directions = text.slice(text.indexOf(':') + 1).split(',');
+  const [x,y = ''] = directions
+  image.style.objectPosition = `${x.trim().toLowerCase()} ${y.trim().toLowerCase()}`;
```
JasonHowellSlavin, line 27: nit: might be nice to check for the image here as well:
 `if (image) image.style.objectPosition = '${x.trim().toLowerCase()} ${y.trim().toLowerCase()}';`

test/blocks/section-metadata/section-meta.test.js
```diff
@@ -20,6 +20,15 @@ describe('Section Metdata', () => {
     expect(sec.classList.contains('has-background')).to.be.true;
   });
 
+  it('Handles background image focal point', () => {
+    it('gets section metadata', () => {
```
JasonHowellSlavin, line 24: I don't think you need two `it` blocks here.

test/blocks/section-metadata/section-meta.test.js
```diff
@@ -20,6 +20,15 @@ describe('Section Metdata', () => {
...
+      const sec = document.querySelector('.section.image');
+      const sm = sec.querySelector('.section-metadata');
+      init(sm);
+      expect(sec.style.objectPosition).to.equal('center bottom');
```
JasonHowellSlavin, line 28: Does this test pass? Your mock has `Right,bottom`?

JasonHowellSlavin (commented):
A lot of good stuff in here. A few of the comments are nit (nit-picky) items.

libs/blocks/marquee/marquee.js
```diff
@@ -46,6 +48,11 @@ const decorateBlockBg = (block, node) => {
     if (child.querySelector('a[href$=".mp4"]')) {
       decorateVideo(child);
     }
+
+    if (child.querySelector('picture')) {
+      const pic = child.querySelector('picture');
+      handleFocalpoint(pic, child)
+    }
```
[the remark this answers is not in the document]
suhjainadobe: done

libs/blocks/section-metadata/section-metadata.js
suhjainadobe: done

test/blocks/section-metadata/section-meta.test.js
suhjainadobe: done

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -1,17 +1,32 @@
...
       section.style.background = color;
     }
   }
 }
 
+export function handleFocalpoint(pic, child) {
+  if (!(child)) return;
+  let text = '';
+  if (child.childElementCount == 2) {
+    text = child.querySelectorAll('p')[1].textContent;
```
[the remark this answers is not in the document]
suhjainadobe: This to accommodate when p tag that appears when the user hits enter after the image and authors the focal point. Else it directly appears as a textcontent for the div.

JasonHowellSlavin (approved):

echampio-at-adobe:
Hi. Unfortunately, if you've updated the "after" doc, I still don't see things working. 
https://focalpoint--milo--suhjainadobe.hlx.page/drafts/suhjain/background-image-test?martech=off

The pink dot that should remain in the viewport for the very first image happily scrolls off the left side of the window as soon as I resize. It's visible only when the window is almost full-screen on my MacBook. I see the same behavior in Safari and Firefox as well as Chrome.

suhjainadobe:
@echampio-at-adobe: Please check now.I think the points are not authored properly.

libs/blocks/marquee/marquee.js
```diff
@@ -13,10 +13,12 @@
 /*
  * Marquee - v6.0
  */
+import { handleFocalpoint } from '../section-metadata/section-metadata.js';
```
auniverseaway, line 16: This will create performance problems and this should not be done.

auniverseaway (changes requested):
This PR created performance problems. Marquee should not have a dependency on section-metadata.

libs/blocks/marquee/marquee.js
auniverseaway: This would mean that section-metadata loads as a dependency in every marquee. This will impact LCP in a not insignificant way. It's bad enough we have the others... decorate, attributes, etc.

suhjainadobe: I have tested the LCP score. It is not lowering the score, it remains the same. Also, @rgclayton suggested using the function if the same is present on section-metadata in one of my old PR. 
@rgclayton any suggestions for this?

auniverseaway: Can you share your draft docs that have your page speed? The ones in the PR are not sufficient.

rgclayton: >Also, @rgclayton suggested using the function if the same is present on section-metadata in one of my old PR.
@suhjainadobe My suggestion was to treat `section-metadata` focal point the same as marquee, in that there isn't a new key value pair introduced/different authoring experience.

auniverseaway: We only have so much LCP budget. You're adding bytes to every single marquee request which is wrong. We should be gating the use of the module...

```js
async function decorateBackground(el) {
  if (el.children.length > 1) {
    const { default: initPos } = await import('./your-bg-handler.js');
    initPos(el);
  }
}
```
Something like this (it's definitely not exact) would be much more preferable.

suhjainadobe: Actually, this particular function might be called more than one time depending on how many viewports are authored. So that would mean it will import the function multiple times. Can I directly write the function in the marquee rather than importing it from section-metadata.

suhjainadobe: @auniverseaway I have pushed the changes. Could you please review the PR again?

libs/blocks/marquee/marquee.js
```diff
@@ -54,6 +60,22 @@ const decorateBlockBg = (block, node) => {
...
+    text = child.textContent;
+  }
+  const image = pic.querySelector('img');
+  const directions = text.slice(text.indexOf(':') + 1).split(',');
+  const [x,y = ''] = directions
+  if (image) {
+    image.style.objectPosition = `${x.trim().toLowerCase()} ${y.trim().toLowerCase()}`;
+  }
+}
+
```
JasonHowellSlavin, line 78: We now have this method in two areas, here and `section-metadata.js` Is it necessary in both places? One would be preferable. 

I see you changed the authoring experience to reflect this change. Having this authored two ways could be confusing. If the requirement is only for the Marquee, then maybe remove this function from `section-metadata.js`. That being said, I believe we also want to keep `marquee.js` as small as possible since it is typically the first block to load above the fold and its images are typically set to loading eagerly prior to `loadArea`.

JasonHowellSlavin, line 78: We now have this method in two areas, here and `section-metadata.js` Is it necessary in both places? One would be preferable. 

I see you changed the authoring experience to reflect this change. Having this authored two ways could be confusing. If the requirement is only for the Marquee, then maybe remove this function from `section-metadata.js`. That being said, I believe we also want to keep `marquee.js` as small as possible since it is typically the first block to load above the fold and its images are typically set to loading eagerly prior to `loadArea`.

suhjainadobe: The authoring experience remains the same for both marquee and section-metadata.  The background images can be added in marquee and section-metadata; both are independent. We have to author the focal point under the background images in the same way as it was done previously. I have added the function in both blocks to reduce the load on marquee and not make it dependent on section-metadata.

suhjainadobe: For eg: https://focalpoint--milo--suhjainadobe.hlx.page/drafts/suhjain/background-image-test?martech=off
Here the authoring remains same for both section-metadata and marquee.

libs/blocks/marquee/marquee.js
auniverseaway: ESM de-dupes imports automatically. Also, if you build this correctly, you would intelligently figure out if the "row" of background has any positioning and import the function.

libs/blocks/marquee/marquee.js
```diff
@@ -17,6 +17,7 @@ import { decorateButtons, getBlockSize } from '../../utils/decorate.js';
 import { decorateBlockAnalytics, decorateLinkAnalytics } from '../../martech/attributes.js';
 import { createTag } from '../../utils/utils.js';
 
+
```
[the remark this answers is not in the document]
suhjainadobe: removed

libs/blocks/marquee/marquee.js
suhjainadobe: @auniverseaway : It is done. Please review.

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -1,19 +1,36 @@
 import { debounce } from '../../utils/utils.js';
 
 function handleBackground(div, section) {
-  const pic = div.querySelector('picture');
+  const pic = div.background.content.querySelector('picture');
   if (pic) {
     section.classList.add('has-background');
     pic.classList.add('section-background');
+    handleFocalpoint(pic, div.background.content);
```
ryanmparrish, line 8: Was it discussed to scrape this info from the image alt text description at all?  `(top, left) | desc...` Similar to the image_link PR. 
I'm just thinking of other context for other potential extra text rows w/ images like photo credit.

libs/blocks/marquee/marquee.js
```diff
@@ -38,14 +38,20 @@ const decorateBlockBg = (block, node) => {
...
     }
 
     if (child.querySelector('a[href$=".mp4"]')) {
       decorateVideo(child);
     }
+
+    const pic = child.querySelector('picture');
+    if (pic && (child.childElementCount == 2 || child.textContent)) {
+      const { handleFocalpoint } = await import ('../section-metadata/section-metadata.js');
+      handleFocalpoint(pic,child);
```
ryanmparrish, line 53:
```suggestion
      handleFocalpoint(pic, child);
```

libs/blocks/section-metadata/section-metadata.js
suhjainadobe: @ryanmparrish : Currently we only handling focal points on the background images. alt text description is not handled in this story.

ryanmparrish:
Great suggestions @vhargrave. @suhjainadobe, I'll give it a  ✅  once these are implemented. 
On the comment above about the left margin on the background cell when on tablet viewport. That looks to me like an intentional style - [library](https://main--milo--adobecom.hlx.page/docs/library/blocks/marquee). I can talk to the consonant team and address in a subsequent PR if needed.

rgclayton:
Since this is a new feature we are marking with the `new feature` label to broadcast to the community. [Please see the PR label guide](https://github.com/adobecom/milo/wiki/PR-Labels#communication-labels) for prepping this to be sent out once merged.

rgclayton:
> Also, I verified the 'left, right' string etc. are persisting as text artifact elements in the `.background` div.
> We should probably `.remove()` any DOM elements that aren't suppose to be in the view.

I agree with Ryan. These should get cleaned up.
image

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -1,19 +1,36 @@
...
-    const color = div.textContent;
+    const color = div.background.content.textContent;
     if (color) {
       section.style.background = color;
     }
   }
 }
 
+export function handleFocalpoint(pic, child) {
+  if (!(child)) return;
```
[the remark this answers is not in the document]
suhjainadobe: done.

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -1,19 +1,36 @@
...
+  let text = '';
+  if (child.childElementCount == 2) {
```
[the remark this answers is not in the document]
suhjainadobe: done

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -1,19 +1,36 @@
...
+    text = child.querySelectorAll('p')[1]?.textContent;
+  } else if (child.textContent) {
+    text = child.textContent;
+  }
+  const image = pic.querySelector('img');
+  const directions = text.split(',');
+  const [x,y = ''] = directions;
```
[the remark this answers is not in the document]
suhjainadobe: done

libs/blocks/marquee/marquee.js
```diff
@@ -38,14 +38,20 @@ const decorateBlockBg = (block, node) => {
     children[1].classList.add(viewports[2]);
   }
 
-  Array.from(children).forEach((child, index) => {
+  Array.from(children).forEach(async(child, index) => {
```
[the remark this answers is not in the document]
suhjainadobe: done

libs/blocks/marquee/marquee.js
```diff
@@ -38,14 +38,20 @@ const decorateBlockBg = (block, node) => {
...
       child.classList.add(viewports[index]);
     }
 
     if (child.querySelector('a[href$=".mp4"]')) {
       decorateVideo(child);
     }
+
+    const pic = child.querySelector('picture');
+    if (pic && (child.childElementCount == 2 || child.textContent)) {
+      const { handleFocalpoint } = await import ('../section-metadata/section-metadata.js');
```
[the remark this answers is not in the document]
suhjainadobe: done

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -1,19 +1,36 @@
...
+export function handleFocalpoint(pic, child) {
+  if (!(child)) return;
+  let text = '';
+  if (child.childElementCount == 2) {
+    text = child.querySelectorAll('p')[1]?.textContent;
+  } else if (child.textContent) {
+    text = child.textContent;
+  }
+  const image = pic.querySelector('img');
+  const directions = text.split(',');
```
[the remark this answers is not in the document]
suhjainadobe: nice suggestion..this is done. Thanks!

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -1,19 +1,36 @@
...
 
+export function handleFocalpoint(pic, child) {
+  if (!(child)) return;
+  let text = '';
+  if (child.childElementCount == 2) {
+    text = child.querySelectorAll('p')[1]?.textContent;
+  } else if (child.textContent) {
+    text = child.textContent;
+  }
+  const image = pic.querySelector('img');
```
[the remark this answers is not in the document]
suhjainadobe: yeah..I was only checking for the child element, but now checking for image as well

suhjainadobe:
@vhargrave @rgclayton @ryanmparrish : yes, the margin was coming because the element was getting added in DOM for marquee. I have removed the element specifically for the marquee and not for section-metadata to keep its format consistent in DOM.

vhargrave (approved):
looks good!

vhargrave (changes requested):
unit tests are now failing

mokimo:
@vhargrave flaky tests 😢 re-ran the workflow all green now

vhargrave (approved):

rgclayton (approved):

auniverseaway (approved):

PR #675, opened 2023-04-28
Global navigation TBT improvements

mokimo:
### Description 
After merging https://github.com/adobecom/milo/pull/601 we went back to the drawing board to re-architect a bit of how our global navigation is set up to avoid any TBT.

No single global navigation task should take above ~15-20ms now (tested on a 6x throttled CPU on MAC)

We might be able to optimize further by turning the global-navigation "features" (profile/search/dropdowns) into their own blocks. We currently don't want to explore this path, because those "features" wouldn't be useable standalone like any MILO block, but this is a first step and after the MVP we still have more performance related plans.

### Changes
- Improve TBT for the global navigation
- dropdowns should open faster
  - decouple logic from the keyboard navigation
  - reduce delay and improve TBT
- Search should get cleared when it's closed

### Relations
Resolves: [MWPW-130482](https://jira.corp.adobe.com/browse/MWPW-130482)
Resolves: [MWPW-130481](https://jira.corp.adobe.com/browse/MWPW-130481)

### Case study 1 - Global navigation
Before: one long init tasks that took care of a lot of logic
gnav-before

After: multiple small tasks, giving milo/browser/user input the chance to execute work as well.
Flow: init -> CPU BREAK -> init sub-feature (such as search) -> CPU BREAK -> init sub-feature (such as profile) -> CPU BREAK -> ... once the nav is completely ready we add it to the DOM causing only one layout task
gnav-after

### Case study 2 - dropdowns
Before: Initializing the dropdowns was just one big task with a lot of sub tasks
decorateDropdown-before

After: the dropdown tasks are scattered with each "sub task" taking only a few ms allowing the browser to process user input inbetween. (If they aren't interacted with for the first 800ms, this happens when the main thread is _hopefully_ idle.)
Flow: similar to gnav, only render the dropdown once it's fully ready.
decorateDropdown-after

**Test URLs:**
- After: https://mwpw-130482-gnav-tbt-improvements--milo--mokimo.hlx.page/drafts/ramuntea/gnav-refactor?martech=off#


mokimo:
We are NOT live with the (feds-) global navigation - so i'll apply the trivial label to make it easier on the reviewers.

auniverseaway:
This is incredible. Thank you.

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -109,6 +117,16 @@ const decorateProfileTrigger = async ({ avatar }) => {
   return buttonElem;
 };
 
+let keyboardNav;
+const setupKeyboardNav = async () => {
+  if (keyboardNav) return;
+  keyboardNav = keyboardNav || new Promise(async (resolve) => {
```
overmyheadandbody, line 123: If `keyboardNav` exists, then the method returns immediately, so I'm not sure whether the 'or' condition is needed here.

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -129,39 +147,50 @@ class Gnav {
...
-        ${this.isDesktop.matches ? this.decorateBreadcrumbs() : ''}
-      </div>`;
-
-    this.el.addEventListener('click', this.loadDelayed);
-    this.el.addEventListener('keydown', this.loadDelayed);
-    setTimeout(() => this.loadDelayed(), 3000);
-    this.loadIMS();
+    ${this.elements.topnav}
+    ${this.isDesktop.matches ? this.decorateBreadcrumbs() : ''}
+  </div>`;
```
overmyheadandbody, line 189: Indentation looks a bit off here, since next line has more spaces.

overmyheadandbody (approved):
Great job! Glad this could be implemented so quickly! 👍

honstar (approved):
💯

narcis-radu (approved):

overmyheadandbody:
Since this is marked as 'trivial', has 3 approvals and will not be used in production for the moment, could we go ahead and merge?

PR #676, opened 2023-04-28
MWPW-130217 Use localized content in FEDS search

TsayAdobe:
FEDS search links to Helpx global search and it needs the locale path prefix for content and the query parameter `country` for user location.

* It's assumed that adobe.com and Helpx has the same path prefix for every locale.
It takes the prefix of the current page and uses it for Helpx. (@overmyheadandbody,  @nkthakur48. Please review.)  

* This fix passes the HTML lang country to Helpx instead of the user geolocation. Finding user geolocation is the georouting feature, which needs refactoring for other modules to use. Need a new ticket for  @vhargrave 
 
* There was no unit test for gnav-search.js. Paid the unit-test debt. 100% coverage now.

Resolves: [MWPW-130217](https://jira.corp.adobe.com/browse/MWPW-130217)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/kr/drafts/tsay/test-feds-search?martech=off
- After: https://mwpw-130217--milo--tsayadobe.hlx.page/kr/drafts/tsay/test-feds-search?martech=off

**DC URLs:**
- Before: https://www.stage.adobe.com/fr/acrobat/online/pdf-to-ppt.html?martech=off
- After: https://www.stage.adobe.com/fr/acrobat/online/pdf-to-ppt.html?martech=off&milolibs=MWPW-130217--milo--tsayadobe


test/blocks/gnav/mocks/autocomplete..json
```diff
@@ -0,0 +1,6 @@
+{
+  "query_prefix": "",
+  "locale": "fr_FR",
+  "suggested_completions": [],
+  "elastic_search_time": 1.0
+}
```
honstar, line 6: Nit: newline missing at the end of the file. Also, the filename seems to have a superfluous `.`, not sure if it's needed.

test/blocks/gnav/mocks/autocomplete.photoshopx.json
```diff
@@ -0,0 +1,6 @@
+{
+  "query_prefix": "photoshopx",
+  "locale": "fr_FR",
+  "suggested_completions": [],
+  "elastic_search_time": 7.09057
+}
```
honstar, line 6: Nit: newline missing.

honstar (approved):

libs/blocks/gnav/gnav-search.js
```diff
@@ -18,10 +18,10 @@ const fetchResults = async (searchStr, locale = 'en_US') => {
   return null;
 };
 
-const getNoResultsEl = (value) => {
+const getNoResultsEl = (value, prefix, country = 'US') => {
```
overmyheadandbody, line 21: I'd suggest we either keep default values for both `prefix` and `country` or remove them for both, since the `getHelpxLink` method has defaults for both.

libs/blocks/gnav/gnav-search.js
```diff
@@ -74,9 +74,9 @@ const getSuggestions = (json) => {
 };
 
 const onSearchInput = async ({ value, resultsEl, locale, searchInputEl }) => {
-  const results = await fetchResults(value, locale);
+  const results = await fetchResults(value, window.adobeid?.locale);
```
overmyheadandbody, line 77: I'd add a comment here to document why `locale` isn't a good choice and `window.adobeid?.locale` needs to be used instead.

overmyheadandbody (commented):
The Search logic will be updated with the rollout of the FEDS Gnav, but we can move forward with this for the moment. We'll need a story on the Eurovision side to integrate the prefix as well.

libs/blocks/gnav/gnav-search.js
TsayAdobe: Some places use hyphen (en-US), and some use underscore (en_US). If we can't standardize it, we may keep both in the `locale` object.

overmyheadandbody: I'm still not sure I'd base this on another piece of logic running. `window.adobeid` is currently declared in the `decorateProfile` method [here](https://github.com/adobecom/milo/blob/ed5955c1ae1db4ab34f7bfad216893cdf58db230/libs/blocks/gnav/gnav.js#L464). But maybe the current page doesn't need a Profile, so that logic would never run, thus `adobeid.locale` would not get defined. We already have access to `locale` via the method's parameters, I think it would be more straightforward to use `locale?.ietf?.replace('-', '_')` here.

On another note, I made a few calls with both `en-US` and `en_US`, `de-CH` and `de_CH`. The results are identical, so I'm wondering why this change is actually needed.

libs/blocks/gnav/gnav.js
```diff
@@ -413,7 +413,9 @@ class Gnav {
     const searchResults = createTag('div', { class: 'gnav-search-results' });
     const searchResultsUl = createTag('ul');
     searchResults.append(searchResultsUl);
-    const locale = getLocale();
+    const { locale } = getConfig();
+
+    locale.geo = getCountry();
```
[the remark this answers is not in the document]
overmyheadandbody: I'd still argue that appending new properties to the `locale` outside of the `getConfig` scope would be a bad practice, but this version of the Milo gnav will be deprecated soon anyway and we went down another path with the whole Search logic, so this won't be an issue then.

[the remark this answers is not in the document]
vhargrave: Agree, don't think we should do that either, even as a temporary thing.

libs/blocks/gnav/gnav-search.js
TsayAdobe: Dexter uses locales with an underscore.

libs/blocks/gnav/gnav.js
```diff
@@ -427,7 +429,7 @@ class Gnav {
 
     searchInput.addEventListener('keydown', (e) => {
       if (e.code === 'Enter') {
-        window.open(this.getHelpxLink(e.target.value, getCountry()));
+        window.open(this.getHelpxLink(e.target.value, locale.prefix, locale.geo));
```
adeshmukh-1585, line 432: Instead of Hardcoding it to HelpX URL, is it possible to have helpx URL as a fallback and the author could pass that serp page url via configuration metadata on gnav ? That way we can have flexibility on A/B testing if the page undergoes some changes and we need to test the new / against the old and also some applications may need to send customers to helpx serp and would like to have their own serp ?

honstar: Sounds like a good idea 👍

TsayAdobe: It sounds like a story. I will file a ticket for Milo team to triage.

auniverseaway:
I'm just seeing the "helpX" stuff which is a bit concerning that it's already in Milo. We should never have consumer specific code like this in Milo.

That being said, gnav is going away soon so this is fine to merge in.

PR #684, opened 2023-05-03
mwpw-130571: Allow for easier QA of the new global-navigation

mokimo:
### Description
Allow consumers to test the new global-navigation/global-footer via a query parameter on top of changing the documents metadata. To offer as smooth of a transition as possible to as many teams as possible. `&headerqa=global-navigation` or `&footerqa=global-navigation`

Resolves: [MWPW-130571](https://jira.corp.adobe.com/browse/MWPW-130571)

### Benefits
- Allow QA to compare and test the same page using both navigations and raise potential bugs.
- Minimize duplicate content needed to test existing / new navigation using metadata. Right now QA would need to create new pages and enable the new navigation.
- Enable smooth transitions to new navigation for Milo consumers.

### Further context
This is currently also achievable by using the milolibs `&milolibs=mwpw-130571--milo--mokimo&headerQA=global-navigation` though that comes with drawbacks.
- Only useable on `hlx.live` `hlx.page` pages
- We need to ensure the branch is always synchronised

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/suhjain/background-image-test?martech=off&headerQA=global-navigation#
- After: https://mwpw-130571--milo--mokimo.hlx.page/drafts/suhjain/background-image-test?martech=off&headerQA=global-navigation#


narcis-radu (approved):

narcis-radu (approved):

honstar (approved):

mokimo:
@adobecom/admins could one of you have a look since this touches the utils?

chrischrischris (approved):
I'm not a big fan of adding code to utils just for QA, but I can see why this helps.  If we add this with the knowledge that we'll remove it once we all are using the new gnav then this should be fine.

Should the params be all lowercase?  Makes it easier to not mess it up.

mokimo:
@chrischrischris added it as mental TODO and added a reminder that will ping me on slack so I'm 100% sure we won't forget to remove it again!

PR #685, opened 2023-05-03
MWPW-127384 Ability to anchor/focus background image at certain point for marquee

suhjainadobe:
Ability to anchor/focus background image at certain point for marquee

Resolves: [MWPW-127384](https://jira.corp.adobe.com/browse/MWPW-127384)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/suhjain/background-image-test?martech=off
- After: https://focalPointMarquee--milo--suhjainadobe.hlx.page/drafts/suhjain/background-image-test?martech=off


rgclayton (changes requested):
Trying to unpack what's going on here.

Looking at the authoring environment I see `section-metadata` set up with `focalPoint` as a key. But, not seeing anything related to section-metadata updates in the code.

For the marquee I see there is a nested table inside the marquee background image row to define the focal points. This really isn't a pattern we've been supporting.
image

I think we need some more discussion around the authoring pattern for this. 

We can also leverage the `object-position` property to do a combination of `x` `y` coordinates with the use of `object-position: 0% 100%;` which translates to left and bottom.

libs/blocks/marquee/marquee.js
```diff
@@ -54,6 +60,16 @@ const decorateBlockBg = (block, node) => {
...
+export const getMetadata = (el) => [...el.rows].reduce((obj, row) => {
+  if (row.children) {
+    const key = row.children[0].textContent.trim().toLowerCase();
+    const content = row.children[1];
+    const text = content.textContent.trim().toLowerCase();
+    if (key && content) obj[key] = { content, text };
+  }
+  return obj;
+}, {});
+
```
rgclayton, line 72: This function shouldn't be added. It's already part of https://github.com/adobecom/milo/blob/main/libs/blocks/section-metadata/section-metadata.js#L59

You should just import it to use it.

suhjainadobe:
Hello @rgclayton ,
Changing the approach to author the focal point on the background image.
image
We have added an additional row below the background, where the author will mention focalPoint for the respective background images.

libs/blocks/marquee/marquee.js
suhjainadobe: Removed the function, as the approach to author the focal point changed.

rgclayton:
I think this can be even more simple to author. I don't think you need to add an extra row under the background images. Just add the text below each image and check for those options. You will probably need to check for textContent and <p> with the text. [Here's an example](https://main--milo--adobecom.hlx.page/drafts/rclayton/marquee-sample.plain.html) of what you might look for.

image


cc: @auniverseaway @ryanmparrish

rgclayton:
I tried authoring the focal points in a marquee and it's not working. Any ideas what's going wrong in my document?

https://focalpointmarquee--milo--suhjainadobe.hlx.page/drafts/rclayton/background-focal-point

rgclayton:
I still don't see it working on the page I've authored. Can you have a look and let me know if it's authored incorrectly or if there is another issue?

https://focalpointmarquee--milo--suhjainadobe.hlx.page/drafts/rclayton/background-focal-point

suhjainadobe:
> I still don't see it working on the page I've authored. Can you have a look and let me know if it's authored incorrectly or if there is another issue?

I can see that it is working for mobile and tablet. But for the desktop, somehow the image and text under it are rendering differently in DOM and not inside the P tag. Investigating it further.

suhjainadobe:
> I still don't see it working on the page I've authored. Can you have a look and let me know if it's authored incorrectly or if there is another issue?

I observed that for the desktop, after uploading the image, Enter was not hit and directly the text was written. Thus the DOM was taking this as a single element. I have the authoring changes on your page. Kindly look into it.

rgclayton:
> I observed that for the desktop, after uploading the image, Enter was not hit and directly the text was written. Thus the DOM was taking this as a single element. I have the authoring changes on your page. Kindly look into it.

Thanks for investigating @suhjainadobe! I think if this pattern needs to be accounted for. I would imagine our authors will run into the same issues.

suhjainadobe:
Closing this PR. The same has been included in #673 as the functionality and authoring were the same.

PR #691, opened 2023-05-08
[MWPW-117909] Footer block refactor

overmyheadandbody:
This is a complete refactor of the Footer block. It:
- refactors the menu logic to use the same algorithm as the one from the header. This should also be helpful for accessibility, which is pending implementation (already available in the top navigation);
- refactors the region selector, but keeps previous modal/fragment functionality intact;
- simplifies the social section logic;
- refactors privacy links. Previous use-cases, such as multiple rows have been taken into account. The privacy event listener from `utils.js` has been adapted to account for all elements on the page with an `openPrivacy` hash value, regardless of their parent type or render time;
- adds support for a 'Featured Products' section. This gets created based on the presence of 'Link Group' blocks;
- implements the `yieldToMain` workflow for decorating the footer in a performant manner. Additionally, the footer is loaded either after 3 seconds or when it is close to coming into view in the viewport;
- centralises all of the footer icons under a single file, since all pages that include a footer have at least the small version of it, which contains all of the icons anyway;
- renames the `blocks` folder under the `global-navigation` block to `features`, since we're not using standard Milo blocks (some discussions are already ongoing around this topic);
- moves logic for importing menu decoration to `utilities`, since it is used by both the top navigation and footer. Same goes for a new `base.css` file, which contains variables and styles relevant to both blocks, since it is not mandatory to load them both on the page, but, when loaded, they should share some styles;
- adds a `getExperienceName` method, since this is useful for analytics purposes;
- moves logic to load blocks and styles to the `utilities` file, since those methods are needed for both the top navigation and the footer;
- calls the `replaceText` placeholder method for dynamically fetched navigation content to ensure placeholders get replaced;
- enhances the Search block logic to pass in correct locale information;
- updates logic that closes open dropdown menus.

I suggest that `global-footer.js` and `global-footer.css` be reviewed as completely new files, since the previous logic was just copied over from the Milo Footer and has been mostly scrapped.

Resolves: [MWPW-117909](https://jira.corp.adobe.com/browse/MWPW-117909)
Also implements: [MWPW-118683](https://jira.corp.adobe.com/browse/MWPW-118683), [MWPW-118518](https://jira.corp.adobe.com/browse/MWPW-118518)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ramuntea/gnav-refactor?martech=off
- After (regular footer): https://mwpw-117909-footer-block-refactor--milo--overmyheadandbody.hlx.page/drafts/ramuntea/gnav-refactor?martech=off
- After (variant with dropdown region selector and two privacy rows): https://mwpw-117909-footer-block-refactor--milo--overmyheadandbody.hlx.page/drafts/ramuntea/footer-refactor-variant?martech=off
- After (small footer): https://mwpw-117909-footer-block-refactor--milo--overmyheadandbody.hlx.page/drafts/ramuntea/footer-refactor-small?martech=off


libs/blocks/global-footer/global-footer.js
```diff
@@ -1,358 +1,352 @@
...
-    return { error: respText }; // can pass additional Response info if needed
-  }
-
-  return { html: respText };
+  addEventListeners = () => {
+    this.isDesktop.addEventListener('change', this.setHeadlineAttributes);
+  };
 }
 
 export default async function init(block) {
```
mokimo, line 348: The init isn't async (anymore?)

mokimo (approved):
Awesome 👏

libs/blocks/global-footer/global-footer.js
overmyheadandbody: Makes sense, I removed it

narcis-radu (approved):

PR #693, opened 2023-05-08
Add support for Jarvis chat

narcis-radu:
Add Jarvis chat support in Milo
* enable Jarvis to be loaded via metadata
* render Jarvis chat based on custom configuration
* enable on-demand rendering

Resolves: [MWPW-118682](https://jira.corp.adobe.com/browse/MWPW-118682)

**Test URLs:**
- After: https://jarvis-chat--milo--narcis-radu.hlx.page/drafts/narcis/default?martech=off
- After (no jarvis): https://jarvis-chat--milo--narcis-radu.hlx.page/drafts/ramuntea/gnav-refactor?martech=off


libs/utils/utils.js
```diff
@@ -112,7 +114,7 @@ function getEnv(conf) {
   const query = location.searchParams.get('env');
 
   if (query) return { ...ENVS[query], consumer: conf[query] };
-  if (host.includes('localhost:')) return { ...ENVS.local, consumer: conf.local };
+  if (host.includes('localhost:') || host.includes('localhost.adobe.com:')) return { ...ENVS.local, consumer: conf.local };
```
narcis-radu, line 117: Adding support for running Milo locally with https

mokimo: That would nicely align with the docs: https://milo.adobe.com/docs/engineering/running-milo-locally 
I'd suggest if this gets merged, we slightly change the docs afterwards
`You will then be able to go to https://localhost.adobe.com:3000/` -> `You will then be able to go to https://localhost.corp.adobe.com:3000/`

`localhost.corp.adobe.com` is also whitelisted for IMS and should make testing anything that requires IMS easier.

libs/utils/utils.js
```diff
@@ -82,6 +82,8 @@ const AUTO_BLOCKS = [
 const ENVS = {
   local: {
     name: 'local',
+    ims: 'stg1',
+    adobeIO: 'cc-collab-stage.adobe.io',
```
mokimo, line 86: Can we also add `account: 'stage.account.adobe.com'` when we are on it? So that you can instantly log in with a stage account

libs/utils/utils.js
narcis-radu: Is your suggestion to check whether the host includes `localhost.corp.adobe.com` as well or just remove the host from the documentation?

mokimo: Edit the documentation to only include `localhost.corp.adobe.com` otherwise IMS will not work locally and ppl might be confused

narcis-radu: I will remove these changes for now and handle them separately afterwards.

libs/features/jarvis-chat.js
```diff
@@ -0,0 +1,344 @@
...
+  const instance = window.AdobeMessagingExperienceClient;
+  return instance?.isAdobeMessagingClientInitialized()
+    && instance?.getMessagingExperienceState()?.windowState !== 'hidden';
+};
+
+const openChat = (event) => {
+  if (!chatInitialized) redirectToSupport();
+  const open = window.AdobeMessagingExperienceClient?.openMessagingWindow;
+  if (!open) return;
+  if (isChatOpen()) return;
```
mokimo, line 231:
```suggestion
  if (typeof open !== "function" || isChatOpen()) return
```

libs/features/jarvis-chat.js
```diff
@@ -0,0 +1,344 @@
...
+  if (event) {
+    const sourceType = event?.target?.tagName?.toLowerCase();
```
mokimo, line 233: You already have a check if event exists. Target should always exist and should be a htmlElement you don't need any of the three checks.

Looks overly defensive to me, but there's no real downside

libs/utils/utils.js
```diff
@@ -654,6 +654,16 @@ export function loadPrivacy() {
   });
 }
 
+async function loadJarvisChat() {
+  const config = getConfig();
+  const jarvis = getMetadata('jarvis-chat');
+  if (!config.jarvis?.id || !config.jarvis?.version) return;
+  if (jarvis === 'on') {
```
mokimo, line 661: Mind giving a bit of background on how milo features work? Does a consumer now explicitly need to state the feature is on, or is it on by default?

mokimo (approved):
Cool 👍

libs/utils/utils.js
narcis-radu: The feature is disabled by default, it's only needed for certain projects. Milo consumers will have to provide a chat configuration and turn the feature on on specific pages (or all) via metadata.

libs/features/jarvis-chat.js
narcis-radu: You're correct, I can update. However, there are use-cases when the event is not set.

honstar (approved):

libs/features/jarvis-chat.js
```diff
@@ -0,0 +1,343 @@
...
+    campaignId: data['exp.campaign_id'],
+    containerId: data['exp.container_id'],
+    controlGroupId: data['exp.control_group_id'],
+    treatmentId: data['exp.treatment_id'],
+    surfaceId: data['exp.surface_id'],
+  };
+
+  Object.keys(sophiaData).forEach((key) => {
+    if (!sophiaData[key]) delete sophiaData[key];
+  });
```
chrischrischris, line 43: Is there a possibility of zero or empty string values that need to be preserved?
```suggestion
  Object.keys(sophiaData).forEach((key) => sophiaData[key] ?? delete sophiaData[key]);
```

libs/features/jarvis-chat.js
```diff
@@ -0,0 +1,343 @@
...
+      captured = (sophiaDataLayer?.campaignId === id);
+    }
+
+    return captured;
+  };
+
+  if (sophiaDataCaptured(sophiaData.campaignId)) return;
+
+  window.digitalData.sophiaResponse = window.digitalData.sophiaResponse || {};
+  window.digitalData.sophiaResponse.fromPage = window.digitalData.sophiaResponse.fromPage || [];
```
chrischrischris, line 63: nit
```suggestion
  window.digitalData.sophiaResponse ||= {};
  window.digitalData.sophiaResponse.fromPage ||= [];
```

libs/features/jarvis-chat.js
```diff
@@ -0,0 +1,343 @@
...
+
+const sendChatIconRenderEvent = (data) => {
+  window.digitalData.primaryEvent = {
+    eventInfo: {
+      eventName: 'chat:'
+        + `${data['event.workflow'] || ''}:`
+        + `${data['event.subcategory'] || ''}:`
+        + `${data['event.subtype'] || ''}:`
+        + `${data['content.name'] || ''}:`
+        + `${data['event.type'] || ''}`,
```
chrischrischris, line 98: Since this is used multiple times, maybe use a fn that can do the aggregation?

something like:
```
const getEventStr = (data, props = []) => 
  props.reduce((str, prop) => {
    str += `${str && ':'}${data[prop] || ''}`;
    return str;
  }, '');
```

usage:
```
eventName: `chat:${getEventStr(data, ['event.workflow', 'event.subcategory', 'event.subtype', ...])}`
```

libs/features/jarvis-chat.js
```diff
@@ -0,0 +1,343 @@
...
+  });
+};
+
+const initJarvisChat = async (config, loadScriptFunction, loadStyleFunction) => {
+  if (!config?.jarvis) return;
+
+  loadScript = loadScriptFunction;
+  loadStyle = loadStyleFunction;
+
+  document.addEventListener('click', async (event) => {
```
chrischrischris, line 325: Does jarvis also load late?  Really wonder if there's a better soln but don't know what that is

libs/utils/utils.js
```diff
@@ -654,6 +654,16 @@ export function loadPrivacy() {
   });
 }
 
+async function loadJarvisChat() {
```
chrischrischris, line 657: Since this is loaded after delay, this can be moved to it's own file.

chrischrischris (changes requested):

libs/features/jarvis-chat.js
narcis-radu: There's no exact information about the data types and values, but technically some properties can be `0` or empty string values.

libs/utils/utils.js
narcis-radu: We would end up loading two files (one with 7 lines of code and the other one with all jarvis chat logic). Do you really think this is necessary? The method is just 7 lines of code and does basic checks whether the jarvis chat logic module should be loaded of not.
The privacy feature is not part of this PR, but we do have some ideas to move it under features, provide the possibility of adding a custom configuration (needed for non *.adobe.com projects) and maybe then we can think about starting a delay file. I remember there have been some problems with this approach in the past in the `blog` project, but can't find the details now - https://github.com/adobe/blog/blob/main/scripts/delayed.js.

mokimo (approved):

overmyheadandbody (approved):
All the good comments have already been added, looks good to me 😀

libs/utils/utils.js
```diff
@@ -654,6 +654,16 @@ export function loadPrivacy() {
...
+  const config = getConfig();
+  const jarvis = getMetadata('jarvis-chat');
+  if (!config.jarvis?.id || !config.jarvis?.version) return;
+  if (jarvis === 'on') {
+    const { initJarvisChat } = await import('../features/jarvis-chat.js');
+    initJarvisChat(config, loadScript, loadStyle);
+  }
+}
```
auniverseaway, line 665: I get that this is delayed, but is there any reason to _not_ load this out of footer instead?

We're trying to reduce the footprint of utils since it is needed to be fully downloaded to get to LCP, and ideally we would get this out.

The alternative is to use this as an opportunity to remove loadDelayed out of utils completely because it shouldn't be here anyway.

narcis-radu: Jarvis chat can also be used on pages without header or footer; in the Milo world we would not be able to load Jarvis if the loading happens within the footer.

I do agree with the second approach, it's something that I previously mentioned in the comments, but we also want to tackle the Privacy component while we're at it and make some test. This will require additional effort that we're willing to do in an upcoming PR, scheduled for June.

chrischrischris (approved):

libs/utils/utils.js
auniverseaway: > Jarvis chat can also be used on pages without header or footer; in the Milo world we would not be able to load Jarvis if the loading happens within the footer.
💯 

Remove loadDelayed: https://jira.corp.adobe.com/browse/MWPW-131408

PR #697, opened 2023-05-09
MWPW-129942: Fix "rate your experience" string

zagi25:
Resolves: [MWPW-129942](https://jira.corp.adobe.com/browse/MWPW-129942)

**Test URLs:**
- Before: https://stage--dc--adobecom.hlx.page/jp/acrobat/online/ppt-to-pdf
- After: https://stage--dc--adobecom.hlx.page/jp/acrobat/online/ppt-to-pdf?milolibs=rate-your-experience-appearing-twice--milo--zagi25

### DIFF
--- a/libs/blocks/review/components/review/Review.js
+++ b/libs/blocks/review/components/review/Review.js
@@ -57,6 +57,11 @@ function Review({
   const [timeoutId, setTimeoutId] = useState(null);
 
   const beforeUnloadCallback = useRef(null);
+  const titleComponent = useRef(
+    html`
+      <h3 className="hlx-reviewTitle">${strings.reviewTitle}</h3>
+    `
+  );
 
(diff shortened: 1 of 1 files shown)


mokimo (approved):
Interesting, got any background/context why without the `useRef` it renders the title twice?

zagi25:
Hi @mokimo, in `loadDeferred()` for JP locale `controlLineBreaksJapanese()` is called and that function changes review title by breaking it into wbr tags. And when the title is re rendered, original title content is added after wbr tags and that is why the title is rendered twice.

TsayAdobe:
The Japanese word wrap feature seems to be dangerous. It could break any dynamic components which use h1 to h6.

narcis-radu (approved):

honstar (approved):

PR #701, opened 2023-05-09
Fix some flaky unit tests

mokimo:
No ticket, no before/after links **purely affects unit tests.**

- Fixes flakey instagram/tiktok/slideshare/twitter/vimeo tests that sporadically fail as they try to load real data / scripts from their respective endpoints. It's now all mocked properly, I don't think unit tests are the right place to test integration stuff, that's what we have nala for.

IMO everyone seeing failing/flaky tests should go ahead and fix them, whether written by oneself or not!


test/blocks/vimeo/vimeo.test.html
```diff
@@ -0,0 +1,39 @@
...
+    </script>
+  </head>
+
+  <body>
+    <a href="https://vimeo.com/636596252">https://vimeo.com/636596252</a>
+  </body>
+
+  <script type="module">
+    import { runTests } from '@web/test-runner-mocha';
+
```
JasonHowellSlavin, line 19: you have this newline after `import { runTests } from '@web/test-runner-mocha` in each test file. Any reason as to why?

test/blocks/tiktok/tiktok.test.html
```diff
@@ -0,0 +1,42 @@
...
+    runTests(() => {
+      describe('tiktok', () => {
+        it('renders embed video for tiktok link', async () => {
+          const tiktok = document.querySelector('a');
+
+          init(tiktok);
+
+          const blockquote = document.querySelector('.tiktok-embed');
+          const script = document.querySelector('script[src="https://www.tiktok.com/embed.js"]');
+
```
JasonHowellSlavin, line 34: every other test does not have a newline between theses `const` declarations and the expects below. I suppose it doesn't matter too much, but might be nice to be consistent across files.

JasonHowellSlavin (approved):
Good stuff, just a bunch of `nit` stuff. The newlines could use consistency between files. 

Is there any way to have the `utils.js` stub be in one place, or does it need to be in each mocks folder under the blocks? Adding 5 of the same file feels off.

test/blocks/tiktok/tiktok.test.html
mokimo: Made sure all the tests have consisntent spacing, at least all the embed ones I've touched

mokimo:
@JasonHowellSlavin They don't have a good place to live (yet) I've thrown it into the instagram mock and just consume it from there. 

Interesting enough, web test runner had some issues before and wouldn't run the tests with each file having their own utils - now it's only one file instead of copied over a dozen of times!

narcis-radu (approved):

mokimo:
@honstar you can provide import maps which allow us to mock es modules https://modern-web.dev/docs/test-runner/writing-tests/mocking/#mocking-es-modules

Benefit: Mock `loadScript` to avoid actual network requests to Instagram and co. which leads to flakiness if the embed providers don't respond in time

honstar (approved):

PR #705, opened 2023-05-10
[MWPW-117909] Footer block refactor

mokimo:
This is a complete refactor of the Footer block. It:
- refactors the menu logic to use the same algorithm as the one from the header. This should also be helpful for accessibility, which is pending implementation (already available in the top navigation);
- refactors the region selector, but keeps previous modal/fragment functionality intact;
- simplifies the social section logic;
- refactors privacy links. Previous use-cases, such as multiple rows have been taken into account. The privacy event listener from `utils.js` has been adapted to account for all elements on the page with an `openPrivacy` hash value, regardless of their parent type or render time;
- adds support for a 'Featured Products' section. This gets created based on the presence of 'Link Group' blocks;
- implements the `yieldToMain` workflow for decorating the footer in a performant manner. Additionally, the footer is loaded either after 3 seconds or when it is close to coming into view in the viewport;
- centralises all of the footer icons under a single file, since all pages that include a footer have at least the small version of it, which contains all of the icons anyway;
- renames the `blocks` folder under the `global-navigation` block to `features`, since we're not using standard Milo blocks (some discussions are already ongoing around this topic);
- moves logic for importing menu decoration to `utilities`, since it is used by both the top navigation and footer. Same goes for a new `base.css` file, which contains variables and styles relevant to both blocks, since it is not mandatory to load them both on the page, but, when loaded, they should share some styles;
- adds a `getExperienceName` method, since this is useful for analytics purposes;
- moves logic to load blocks and styles to the `utilities` file, since those methods are needed for both the top navigation and the footer;
- calls the `replaceText` placeholder method for dynamically fetched navigation content to ensure placeholders get replaced;
- enhances the Search block logic to pass in correct locale information;
- updates logic that closes open dropdown menus.

I suggest that `global-footer.js` and `global-footer.css` be reviewed as completely new files, since the previous logic was just copied over from the Milo Footer and has been mostly scrapped.

Resolves: [MWPW-117909](https://jira.corp.adobe.com/browse/MWPW-117909)
Also implements: [MWPW-118683](https://jira.corp.adobe.com/browse/MWPW-118683), [MWPW-118518](https://jira.corp.adobe.com/browse/MWPW-118518) [MWPW-130523](https://jira.corp.adobe.com/browse/MWPW-130523)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ramuntea/gnav-refactor?martech=off
- After (regular footer): https://gnav--milo--adobecom.hlx.page/drafts/ramuntea/gnav-refactor?martech=off
- After (variant with dropdown region selector and two privacy rows): https://gnav--milo--adobecom.hlx.page/drafts/ramuntea/footer-refactor-variant?martech=off
- After (small footer): https://gnav--milo--adobecom.hlx.page/drafts/ramuntea/footer-refactor-small?martech=off


mokimo:
Both @narcis-radu & me already reviewed the original PR at https://github.com/adobecom/milo/pull/691 which has been closed to avoid merge conflicts with new changes

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -488,8 +477,11 @@ class Gnav {
       const decorateDropdown = async () => {
         template.removeEventListener('click', decorateDropdown);
         clearTimeout(decorationTimeout);
-        await this.loadDecorateDropdown();
-        this.decorateDropdown({
+
+        const menuLogic = await loadDecorateMenu();
+        this.decorateMenu = menuLogic.decorateMenu;
+
+        this.decorateMenu({
```
chrischrischris, line 484: curious why the assignment to this.decorateMenu as it doesn't appear to be used elsewhere. Could you just use:
```suggestion
        menuLogic.decorateMenu({
```

libs/blocks/global-navigation/utilities/menu/menu.css
```diff
@@ -0,0 +1,175 @@
...
+    justify-content: space-between;
+    width: 100%;
+    max-width: var(--feds-maxWidth-nav);
+  }
+
+  .feds-menu-column {
+    padding: 12px 0;
+  }
+
+  /* Uncomment if border between simple dropdown columns are needed */
```
chrischrischris, line 91: Is there a jira associated with this?  Don't want it to become forgotten commented out code.

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -53,6 +53,48 @@ export function getAnalyticsValue(str, index) {
...
+    ]);
+
+    resolve({
+      decorateMenu,
+      decorateLinkGroup,
+    });
+  });
+
+  return cachedDecorateMenu;
+}
```
chrischrischris, line 96: If you don't want to use a global var:
```suggestion
export const loadDecorateMenu = (() => {
  let cachedDecorateMenu;
  return async () => {
    // eslint-disable-next-line no-async-promise-executor
    cachedDecorateMenu = cachedDecorateMenu || new Promise(async (resolve) => {
      const [{ decorateMenu, decorateLinkGroup }] = await Promise.all([
        loadBlock('./menu/menu.js'),
        loadStyles('utilities/menu/menu.css'),
      ]);

      resolve({
        decorateMenu,
        decorateLinkGroup,
      });
    });
    return cachedDecorateMenu;
  };
})();
```

chrischrischris: Also I think there's a way to write the Promise without async (and remove the eslint-disable) but not sure.

libs/utils/utils.js
```diff
@@ -647,10 +647,12 @@ export function loadPrivacy() {
   };
   loadScript('https://www.adobe.com/etc.clientlibs/globalnav/clientlibs/base/privacy-standalone.js');
 
-  const privacyTrigger = document.querySelector('footer a[href*="#openPrivacy"]');
-  privacyTrigger?.addEventListener('click', (event) => {
-    event.preventDefault();
-    window.adobePrivacy?.showPreferenceCenter();
+  // Privacy triggers can exist anywhere on the page and can be added at any time
+  document.addEventListener('click', (event) => {
```
chrischrischris, line 651: > can be added at any time

Does this mean after the page has loaded?  This seems heavy handed to intercept all clicks and look for the link, but yeah if links get added after load then i guess it's the only solution, or possibly use a mutation observer to look for when it gets added and then add the click handler?  (that may not be a better soln)

overmyheadandbody (approved):

chrischrischris (approved):

PR #711, opened 2023-05-10
[MWPW-126504] offer preview in sidekick

Axelcureno:
* Adds button in AEM sidekick to invoke Offer Preview experience.
* Lets an author paste an OST link in an iframe to fetch and preview offer details, checkout link within Word.

Resolves: [MWPW-126504](https://jira.corp.adobe.com/browse/MWPW-126504)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page//tools/commerce?martech=off
- After: [https://offer-preview--milo--adobecom.hlx.page/?martech=off](https://offer-preview--milo--adobecom.hlx.page/tools/commerce)


yesil (changes requested):
@Axelcureno your change on loadPromise breaks tacocat initialization and results wcs call promises are no more cached in-memory.

Compare the number of WCS calls below to see the difference:

before: https://main--milo--adobecom.hlx.page/drafts/ilyas/merch
after: https://offer-preview--milo--adobecom.hlx.page/drafts/ilyas/merch

libs/blocks/commerce/commerce.js
```diff
@@ -0,0 +1,122 @@
...
+    offerValue.classList.add('offer-value');
+    offerKey.textContent = `${formatOfferDetailKeys(key)}: `;
+    offerValue.textContent = value;
+    offerData.appendChild(offerKey);
+    offerData.appendChild(offerValue);
+    offerDetailsList.appendChild(offerData);
+  });
+  const checkoutLink = document.createElement('a');
+  checkoutLink.textContent = 'Checkout link';
+  searchParams.set('workflowStep', 'email');
```
honstar, line 61: As per #716, these default parameters are not needed. cc @3ch023

libs/blocks/commerce/commerce.js
```diff
@@ -0,0 +1,139 @@
+import { createTag, getConfig, loadScript } from '../../utils/utils.js';
+import { getTacocatEnv, runTacocat, buildCheckoutButton, getCheckoutContext, omitNullValues } from '../merch/merch.js';
```
honstar, line 2: @Axelcureno : can you please add unit tests for this new `commerce` block? We want to make sure it's not breaking in case these imported functions are modified in the `merch` block. Not sure if it makes sense to externalize these functions into a separate `utils` script.

honstar (changes requested):

libs/blocks/commerce/commerce.js
Axelcureno: Unit tests added ✅

libs/blocks/commerce/commerce.js
```diff
@@ -0,0 +1,139 @@
...
+  });
+  const checkoutUrl = buildCheckoutButton(checkoutLink, options);
+  checkoutUrl.target = '_blank';
+  offerDetailsList.appendChild(checkoutUrl);
+  el.append(offerDetailsList);
+};
+
+export const handleSearch = async (e, el) => {
+  let searchParams = {};
+  function displaySearchError() {
```
yesil, line 96: a normal function inside an arrow function looks a bit weird to me.
Can we make `displaySearchError` an arrow function too?

libs/blocks/commerce/commerce.js
```diff
@@ -0,0 +1,139 @@
...
+    wcsOsi: searchParams.get('osi'),
+    ...getCheckoutContext(searchParams, getConfig()),
+  });
+  const checkoutUrl = buildCheckoutButton(checkoutLink, options);
+  checkoutUrl.target = '_blank';
+  offerDetailsList.appendChild(checkoutUrl);
+  el.append(offerDetailsList);
+};
+
+export const handleSearch = async (e, el) => {
```
yesil, line 94: nit: suggestion to rename `e` to `event`
`async (event, el) => {`

libs/blocks/commerce/commerce.js
```diff
@@ -0,0 +1,139 @@
...
+  function displaySearchError() {
+    const notValidUrl = document.createElement('h4');
+    notValidUrl.classList.add('not-valid-url');
+    notValidUrl.textContent = 'Not a valid offer link';
+    el.append(notValidUrl);
+  }
+  el.textContent = '';
+  const search = e.target.value;
+  if (search === '') return undefined;
+  try {
```
yesil, line 105: if you use `try` to catch an exception from an async function, I recommend to `await` it.
Otherwise,  just chain the promise with .catch function at the line 114.

`window.tacocat.wcs.resolveOfferSelector(osi).then(([offerDetails]) => {}).catch(displaySearchError)`

libs/blocks/commerce/commerce.js
```diff
@@ -0,0 +1,139 @@
...
+  let searchParams = {};
+  function displaySearchError() {
+    const notValidUrl = document.createElement('h4');
+    notValidUrl.classList.add('not-valid-url');
+    notValidUrl.textContent = 'Not a valid offer link';
+    el.append(notValidUrl);
+  }
+  el.textContent = '';
+  const search = e.target.value;
+  if (search === '') return undefined;
```
yesil, line 104: do you need to return `undefined`?

libs/blocks/commerce/commerce.js
```diff
@@ -0,0 +1,139 @@
...
+};
+
+export const decorateSearch = (el) => {
+  const search = createTag('input', { class: 'offer-search', placeholder: 'Enter offer URL to preview' });
+  const icon = createTag('div', { class: 'offer-search-icon' });
+  const searchWrapper = createTag('div', { class: 'offer-search-wrapper' }, [search, icon]);
+  const offerDetailsWrapper = createTag('div', { class: 'offer-details-wrapper' });
+  el.append(searchWrapper);
+  el.append(offerDetailsWrapper);
+  search.addEventListener('keyup', (e) => { handleSearch(e, offerDetailsWrapper); });
```
yesil, line 128: nit: you can debounce the search request by 500ms or something like this to issue less search requests

libs/blocks/commerce/commerce.js
Axelcureno: Converted `displaySearchError ` into an arrow function ✅

libs/blocks/commerce/commerce.js
Axelcureno: Done ✅

libs/blocks/commerce/commerce.js
```diff
@@ -0,0 +1,152 @@
...
+      runTacocat(tacocatEnv, country, language);
+      resolve(false);
+    })
+    .catch((error) => {
+      console.error('Failed to load tacocat', error);
+      resolve(true);
+    });
+});
+
+export const formatOfferDetails = (o) => {
```
chrischrischris, line 27: Is the only change from `o` the `price`?  Could you instead just add `.price` to `o`?

libs/blocks/commerce/commerce.js
Axelcureno: Chained `.catch()` to promise ✅

libs/blocks/commerce/commerce.js
Axelcureno: It was meant to keep the content empty when input value is empty, not exactly required so I have removed it ✅

libs/blocks/commerce/commerce.js
Axelcureno: Great tip, added 500ms timeout ✅

libs/blocks/commerce/commerce.js
Axelcureno: > Is the only change from `o` the `price`?
It may seem so, but not exactly. The `o` object contains many other properties that are not needed to be displayed. The purpose of `formatOfferDetails` is to filter out the object, so renaming the function to `filterOfferDetails` to better reflect its purpose.

libs/blocks/commerce/commerce.js
```diff
@@ -0,0 +1,152 @@
...
+    searchParams = new URL(search).searchParams;
+    const osi = searchParams.get('osi');
+    if (!osi) {
+      displaySearchError();
+      return undefined;
+    }
+    window.tacocat.wcs.resolveOfferSelector(osi).then(([offerDetails]) => {
+      decorateOfferDetails(el, offerDetails, searchParams);
+    }).catch(displaySearchError);
+  } catch (error) {
```
yesil, line 118: do you still need this catch  here?

libs/blocks/commerce/commerce.js
```diff
@@ -0,0 +1,152 @@
...
+
+export const decorateSearch = (el) => {
+  const search = createTag('input', { class: 'offer-search', placeholder: 'Enter offer URL to preview' });
+  const icon = createTag('div', { class: 'offer-search-icon' });
+  const searchWrapper = createTag('div', { class: 'offer-search-wrapper' }, [search, icon]);
+  const offerDetailsWrapper = createTag('div', { class: 'offer-details-wrapper' });
+  el.append(searchWrapper);
+  el.append(offerDetailsWrapper);
+  search.addEventListener('keyup', (event) => {
+    setTimeout(() => {
```
yesil, line 132: seTimeout isn't mimicking debounce.

There seems to be at least 2 `debounce` functions in Milo codebase already.

libs/blocks/commerce/commerce.js
Axelcureno: Removed ✅

libs/blocks/commerce/commerce.js
Axelcureno: Now reusing debounce function ✅

yesil (approved):

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -12,7 +12,7 @@ function handleBackground(div, section) {
   }
 }
 
-function debounce(func, timeout = 300) {
+export function debounce(func, timeout = 300) {
```
honstar, line 15: I'm not a fan of simply exporting a function from an unrelated block so that it can be reused elsewhere in a totally different block. I'd rather see this function being moved someplace else (maybe `utils/utils.js`?).

libs/blocks/commerce/commerce.js
```diff
@@ -0,0 +1,144 @@
+import { createTag, getConfig, loadScript } from '../../utils/utils.js';
+import { debounce } from '../section-metadata/section-metadata.js';
```
honstar, line 2: See my other comment, this is not ideal imo.

test/blocks/commerce/commerce.test.js
```diff
@@ -0,0 +1,96 @@
+import { expect } from '@esm-bundle/chai';
+import { readFile } from '@web/test-runner-commands';
+import {
+  formatOfferDetails,
```
honstar, line 4: This function doesn't exist (anymore?) and breaks the test. Please remove it.

honstar (changes requested):

test/blocks/commerce/commerce.test.js
honstar: I think the initial function was renamed to `filterOfferDetails`, please adjust.

libs/blocks/section-metadata/section-metadata.js
Axelcureno: Moved the function to utils.js ✅

libs/blocks/commerce/commerce.js
Axelcureno: Fixed

test/blocks/commerce/commerce.test.js
Axelcureno: Done ✅

honstar (approved):

libs/utils/utils.js
```diff
@@ -782,3 +783,11 @@ export function loadLana(options = {}) {
   window.addEventListener('error', lanaError);
   window.addEventListener('unhandledrejection', lanaError);
 }
+
+export function debounce(func, timeout = 300) {
```
honstar, line 787: On second thought, adding the `debounce` function to `utils/utils.js` might not be such a good idea since we're trying to keep this file as small as possible.

@auniverseaway , wdyt?

vhargrave: I know I wasn't asked, but I agree with you @honstar 
that code doesn't belong in utils.

auniverseaway: I'm just seeing this now. This should not have been allowed in `utils.js`. There's no reason a visitor should pay for those bytes unless they need them. To my knowledge, no other function needs this.

@vhargrave @honstar @Axelcureno can one of you PR to change the location?

I will make a note somewhere about the criteria by which something should be allowed into utils.js.

When we first built utils.js, we had some spare bytes where we put things in there for simplicity sake. As utils.js has grown, we need to get back to it being a smaller file where every line should be justified. We set a bad example by having things like `utf8ToB64` in there.

We have since added more critical items like htmlPostFix and localizeLink and we should be kicking out non-critical functions (like @narcis-radu did with loadDelayed).

If memory serves, we should be able to get it down to 600 lines without making things less readable.

vhargrave: @auniverseaway done - created a new PR https://github.com/adobecom/milo/pull/930

PR #712, opened 2023-05-11
MWPW-129098 Add ability to disable autoblocking

chrischrischris:
A comma separated list of block classname selectors can be set in the `DO_NOT_AUTOBLOCK` (for Milo) or in the project config under `doNotAutoBlock` for consumers.

Any blocks specified there will not have any links processed within the block.

Block classnames should be entered like: `'.card-metadata,.other-block'`.

Resolves: [MWPW-129098](https://jira.corp.adobe.com/browse/MWPW-129098)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/iscandar/videoonlycardaem?martech=off
- After: https://c3-donotautoblock--milo--adobecom.hlx.page/drafts/iscandar/videoonlycardaem?martech=off

To verify:
* In the first link under `main > .section > .card-metadata > 5th div  > 2nd div` The video link will be wrapped in `.milo-video` and set in an `iframe`.
* In the second link, same location, the link will be a regular `a` link that has not been transformed for video.

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -423,15 +423,14 @@ export function decorateSVG(a) {
 }
 
 export function decorateAutoBlock(a) {
-  const { doNotAutoBlock = '' } = getConfig();
+  const { autoBlocks, doNotAutoBlock = null } = getConfig();
   if (a.closest(doNotAutoBlock)) {
     return false;
   }
-  const config = getConfig();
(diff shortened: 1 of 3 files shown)


mokimo (approved):
So card content

cmiqueo (approved):

PR #714, opened 2023-05-11
[MWPW-125908] [UAR] Step indicator addition to quiz

fullcolorcoder:
Adds step indicator to UAR quiz. 

Resolves: [MWPW-125908](https://jira.corp.adobe.com/browse/MWPW-125908)

**Test URLs:**
- Before: https://uar-integration--milo--adobecom.hlx.page/drafts/sabya/quiz-2/?martech=off
- After: https://uar-progress-indicator--milo--fullcolorcoder.hlx.page/drafts/sabya/quiz-2/?martech=off


colloyd (approved):

echampio-at-adobe (changes requested):
Hi @fullcolorcoder. Sabya notice this problem, which I'm just passing along. On line 97 of quiz.js, you've hardcoded behavior around certain question ids, which you can't do. The question Ids are completely authored, and may change at any time. This code will also be used to deliver several possible quiz workflows, which will use different questions and different ids accordingly.

echampio-at-adobe (changes requested):
It looks like we might need to add some padding below the progress bar element in the body. In chrome, at least in responsive view, the progress bar circles and the top of the footer touch.

JackySun9:
I think before should be https://uar-integration--milo--adobecom.hlx.page/drafts/sabya/quiz-2/?martech=off

libs/blocks/quiz/quiz.js
```diff
@@ -87,6 +94,11 @@ const App = () => {
       handleResultFlow(transformToFlowData(userSelection));
     } else {
       setUserFlow(nextQuizViews);
+      if (nextQuizViews[0] === 'q-rather' || nextQuizViews[0] === 'q-customer') {
```
sabyamon, line 97: @fullcolorcoder - `q-rather` and `q-customer` should not be hard coded. These are dynamic values that are authorable.

libs/blocks/quiz/quiz.js
```diff
@@ -145,8 +162,13 @@ const App = () => {
                       onOptionClick=${onOptionClick}
                       getOptionsIcons=${getOptionsIcons}
                       handleOnNextClick=${handleOnNextClick} />
+                      <${StepIndicator}
```
sabyamon, line 165: I might not have understood the reason, but can we not include the `StepIndicator` component once and show them either in the header or the footer with the breakpoint changes through CSS?

sabyamon (changes requested):

libs/blocks/quiz/quiz.js
fullcolorcoder: Resolving since we talked about this the other day.

echampio-at-adobe (approved):

echampio-at-adobe (changes requested):
@JackySun9 found a couple of problems that we should review

PR #704, opened 2023-05-09
Control search

karri-sek:
Hi,
     Based on this [comment here](https://github.com/adobecom/milo/pull/676/files) , we would like to control the search URL and option of open the results in a new tab from the author. A sample Doc contains the details like below.
  

Screenshot 2023-05-09 at 11 06 31

**Search url** : optional : a url to search the user text entered on the search bar.
**open in New Tab**:  'no' : append the search text entered on the search bar to the **Search url** and open in the current tab. 
**open in New Tab**:  'yes' : append the search text entered on the search bar to the **Search url** and open in new tab. 

**Test URLs:**

- Before:  https://main--milo--adobecom.hlx.page/en/

- After  : https://control-search--milo--karri-sek.hlx.page/en/

Resolves: [ABPDOCS-140](https://jira.corp.adobe.com/browse/ABPDOCS-140)

PR #715, opened 2023-05-11
feat(ABPDOCS-140): control search bar enter command through author me…

karri-sek:
Hi,
     Based on this [comment here](https://github.com/adobecom/milo/pull/676/files) , we would like to control the search URL and option of open the results in a new tab from the author. A sample Doc contains the details like below.
  

Screenshot 2023-05-09 at 11 06 31

**Search url** : optional : a url to search the user text entered on the search bar.
**open in New Tab**:  'no' : append the search text entered on the search bar to the **Search url** and open in the current tab. 
**open in New Tab**:  'yes' : append the search text entered on the search bar to the **Search url** and open in new tab. 

**Test URLs:**

- Before:  https://main--milo--adobecom.hlx.page/en/

- After  : https://control-search--milo--karri-sek.hlx.page/en/

Resolves: [ABPDOCS-140](https://jira.corp.adobe.com/browse/ABPDOCS-140)

### DIFF
--- a/libs/blocks/gnav/gnav.js
+++ b/libs/blocks/gnav/gnav.js
@@ -56,9 +56,6 @@ class Gnav {
     this.curtain = createTag('div', { class: 'gnav-curtain' });
     const nav = createTag('nav', { class: 'gnav', 'aria-label': 'Main' });
 
-    const mobileToggle = this.decorateToggle(nav);
-    nav.append(mobileToggle);
-
     const brand = this.decorateBrand();
     if (brand) {
       nav.append(brand);
@@ -67,7 +64,9 @@ class Gnav {
     const scrollWrapper = createTag('div', { class: 'mainnav-wrapper' });
 
     const mainNav = this.decorateMainNav();
-    if (mainNav) {
+    if (mainNav && mainNav.childElementCount) {
+      const mobileToggle = this.decorateToggle();
+      nav.prepend(mobileToggle);
       scrollWrapper.append(mainNav);
     }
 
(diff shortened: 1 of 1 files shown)


mokimo:
The branch is not correctly rebased with main

overmyheadandbody:
@mokimo already raised some concerns in the previous PR (https://github.com/adobecom/milo/pull/704) and I second that. I'm not convinced this is the way to go, this way of authoring can get very messy very fast. I'd go for a project config approach, since the behaviour should be consistent across one project's pages.

Also, I would not go ahead with this for now, as we're close to switching to the completely refactored FEDS Gnav soon. If you have a valid use-case for this feature, we'll need to take it through the usual planing process. Merging this now would mean that it would work for a few weeks until we switch to the FEDS Gnav.

honstar:
> @mokimo already raised some concerns in the previous PR (#704) and I second that. I'm not convinced this is the way to go, this way of authoring can get very messy very fast. I'd go for a project config approach, since the behaviour should be consistent across one project's pages.
> 
> Also, I would not go ahead with this for now, as we're close to switching to the completely refactored FEDS Gnav soon. If you have a valid use-case for this feature, we'll need to take it through the usual planing process. Merging this now would mean that it would work for a few weeks until we switch to the FEDS Gnav.

@karri-sek , based on the reviews and feedback from @mokimo and @overmyheadandbody , can you please work with the eurovision team to plan for this feature enhancement? This would be part of the FEDS gnav that the eurovision essentially owns. Thanks!

PR #716, opened 2023-05-12
Mwpw 130472 dont save defaults

3ch023:
Resolves: [MWPW-130472](https://jira.corp.adobe.com/browse/MWPW-130472)

**Test URLs:**
- After:  https://mwpw-130472-dont-save-defaults--milo--3ch023.hlx.page/tools/ost?martech=off

### Tech Details
Don't save default parameters:

- checkoutType=UCv3

- workflowStep=email

### Motivation
In case we need to change the default in the future we will do a small code change, but no content update will be required.

### DIFF
--- a/test/blocks/ost/ost.test.js
+++ b/test/blocks/ost/ost.test.js
@@ -9,8 +9,8 @@ const osi = 'cea462e983f649bca2293325c9894bdd';
 const offerId = 'aeb0bf53517d46e89a1b039f859cf573';
 const offerType = 'M2M';
 const placeholderOptions = {
-  workflow: 'UCv2',
-  workflowStep: 'email/checkout',
+  workflow: 'UCv3',
+  workflowStep: 'email',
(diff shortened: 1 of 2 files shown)


honstar (approved):

test/blocks/ost/ost.test.js
```diff
@@ -41,6 +58,9 @@ describe('test createLinkMarkup', () => {
   });
 
   it('create a "price" link', async () => {
+    const EXPECTED_PRICE_TEXT = `{{PRICE - ${offerType} - Stock}}`;
```
yesil, line 61: why do we keep `{{` around the price text?

yesil (approved):

test/blocks/ost/ost.test.js
honstar: > why do we keep `{{` around the price text?

There's a separate story to address pricing links from OST.

PR #718, opened 2023-05-15
MWPW-129539 : Update file overwrite logic during copy (COPY action)

Nethraasivakumar:
Link to the ticket: https://jira.corp.adobe.com/browse/MWPW-129539

The PR does the following:

Right now, in copy action, saveFile() is called if the file exists at the destination. saveFile() makes several calls in order to handle locked files. The updated code in this PR calls saveFile() only if the file is locked or the folder is absent at the destination. Otherwise, it directly copies the file. This reduces the number of calls significantly.

Test URLs:

Before: https://main--milo--adobecom.hlx.page/tools/floodgate/index.html?project=milo--adobecom&referrer=https%3A%2F%2Fadobe.sharepoint.com%2F%3Ax%3A%2Fr%2Fsites%2Fadobecom%2F_layouts%2F15%2Fdoc2.aspx%3Fsourcedoc%3D%257BB2136347-CE65-470B-842E-C1E55C7C86A4%257D%26file%3DfileOverwrite.xlsx%26action%3Ddefault%26mobileredirect%3Dtrue%26cid%3D9f332e29-5a13-43a3-b2d7-fe2b630f2c13 
After: https://main--milo--nethraasivakumar.hlx.page/tools/floodgate/index.html?project=milo--adobecom&referrer=https%3A%2F%2Fadobe.sharepoint.com%2F%3Ax%3A%2Fr%2Fsites%2Fadobecom%2F_layouts%2F15%2FDoc.aspx%3Fsourcedoc%3D%257BB2136347-CE65-470B-842E-C1E55C7C86A4%257D%26file%3DfileOverwrite.xlsx%26action%3Ddefault%26mobileredirect%3Dtrue%26cid%3D9f332e29-5a13-43a3-b2d7-fe2b630f2c13

### DIFF
--- a/tools/loc/sharepoint.js
+++ b/tools/loc/sharepoint.js
@@ -503,6 +503,7 @@ async function addWorksheetToExcel(excelPath, worksheetName) {
   throw new Error(`Failed to add worksheet ${worksheetName} to ${excelPath}.`);
 }
 
+
 export {
   connect,
   copyFile,
--- a/tools/loc/sharepoint.js
+++ b/tools/loc/sharepoint.js
@@ -503,7 +503,6 @@ async function addWorksheetToExcel(excelPath, worksheetName) {
   throw new Error(`Failed to add worksheet ${worksheetName} to ${excelPath}.`);
 }
 
-
 export {
   connect,
   copyFile,
--- a/tools/loc/sharepoint.js
+++ b/tools/loc/sharepoint.js
@@ -378,6 +378,7 @@ async function getMetadata(srcPath, file) {
 
 async function copyFile(srcPath, destinationFolder, newName, isFloodgate, isFloodgateLockedFile) {
   validateConnection();
+  await createFolder(destinationFolder, isFloodgate);
   const { sp } = isFloodgate ? await getFloodgateConfig() : await getConfig();
   const { baseURI, fgBaseURI } = sp.api.file.copy;
   const rootFolder = isFloodgate ? fgBaseURI.split('/').pop() : baseURI.split('/').pop();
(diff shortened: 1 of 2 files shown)


Nethraasivakumar:
Screenshot 2023-05-12 at 12 04 42 PM

The results from tests show that the number of calls required using the new logic are significantly lower than the current milo main logic.

tools/loc/sharepoint.js
```diff
@@ -378,10 +378,8 @@ async function getMetadata(srcPath, file) {
 
 async function copyFile(srcPath, destinationFolder, newName, isFloodgate, isFloodgateLockedFile) {
   validateConnection();
-  await createFolder(destinationFolder, isFloodgate);
```
sukamat, line 381: @Nethraasivakumar Removal of this might break localization as it calls the same copy function. 
https://github.com/adobecom/milo/blob/main/tools/loc/ui.js#L579
Was that validated with loc?

I would suggest keeping this line as-is in the milo repo but removing it from sharepoint.js that is part of the milo-fg repo as that will be floodgate only.

sukamat (changes requested):
Looks good. One change needed as it will create some issues in Loc.

tools/loc/sharepoint.js
Nethraasivakumar: @sukamat Had missed checking for the dependency of localisation on copyFile. Checked it now and found that you are right. Have reverted the code in order to include the createFolder line so that there's no discrepancy for the localisation flow. As mentioned by you, I'll remove the line when I raise a PR for the milo-fg repo.

sukamat (approved):

sirivuri (approved):

PR #723, opened 2023-05-16
WIP: Fix failing unit tests

mokimo:
Two tests consistently fail, there's a bunch that inconsistently fail that [require a bit more time to fix. ](https://github.com/orgs/adobecom/discussions/635)

Trying to find & smash the 2 tests that currently _consistently_ fail on the CI.


test/blocks/tabs/tabs.test.js
```diff
@@ -1,14 +1,6 @@
-import { readFile, sendMouse, sendKeys } from '@web/test-runner-commands';
+import { readFile, sendKeys } from '@web/test-runner-commands';
```
mokimo, line 1: 3 : Code removal of the `getMiddleOfElement` function and `sendMouse` decreases risks with potential unpredictable results caused by manual mouse events simulations. However, be aware that the test simulation might behave differently due to the difference between a user clicking an element and programmatically invoking the click event. Also, note the removal of the `async` keyword. Be sure that the remainder of the test does not rely on the `async` operation.

test/utils/utils.test.js
```diff
@@ -188,7 +188,7 @@ describe('Utils', () => {
       window.dispatchEvent(event);
       await waitForElement('#milo');
       expect(document.getElementById('milo')).to.exist;
-    })
+    });
```
mokimo, line 191: 1 : The correction to add a semicolon at the end of the statement is a good practice in Javascript, which can help prevent potential bugs related to automatic semicolon insertion (ASI). It also increases readability.

web-test-runner.config.mjs
```diff
@@ -1,5 +1,97 @@
 import { importMapsPlugin } from '@web/dev-server-import-maps';
+import { summaryReporter, defaultReporter } from '@web/test-runner';
```
mokimo, line 2: 2 : `customReporter` was added, presumably for custom test results logging. Although there should be no immediate impact, this could lead to problems if not correctly implemented, and it may increase the complexity of the codebase. Be sure to test this thoroughly. The addition of the `preserveSymlinks` and `nodeResolve` options could potentially affect how modules are resolved and have unintended side effects if other parts of the codebase rely on the previous behaviour.

PR #724, opened 2023-05-16
Fix two failing tests and add a custom test reporter

mokimo:
- Fix two failing tests
- Use a custom test reporter along with the defaultReporter that can miss to log some tests. 

Ticket: https://jira.corp.adobe.com/browse/MWPW-131198
Note: A PR to fix this bug in the defaultReporter has been opened as well https://github.com/modernweb-dev/web/pull/2278


overmyheadandbody (approved):

web-test-runner.config.mjs
```diff
@@ -1,5 +1,19 @@
 import { importMapsPlugin } from '@web/dev-server-import-maps';
+import { defaultReporter } from '@web/test-runner';
 
+function customReporter() {
+  return {
+    async reportTestFileResults({ logger, sessionsForTestFile }) {
+      sessionsForTestFile.forEach((session) => {
+        session.testResults.tests.forEach((test) => {
+          if (!test.passed && !test.skipped) {
+            logger.log(test);
```
narcis-radu, line 10: 👍

narcis-radu (approved):

honstar (approved):

auniverseaway (changes requested):
We want to fix the root problem and not introduce a custom reporter that we have to maintain.

mokimo:
@auniverseaway The root problem is a bug/oversight in the custom reporter, for which I've opened a PR.

https://github.com/modernweb-dev/web/pull/2278 - since we can't control when/if that is getting merged, there's a lot of value in having that customReporter in the meantime IMO.

auniverseaway:
@mokimo My concern was that we were introducing a _replacement_ reporter, not an _additional_ reporter.

We don't want to be on the hook for maintaining a custom reporter over a couple of incorrectly written tests. Re-looking at what you have, I think we're fine... assuming my comment above is accurate.

Thanks for looking into this and sending a PR. Really appreciate it.

auniverseaway (approved):

PR #728, opened 2023-05-17
MWPW-131190: Global-navigation correctly fetch placeholders for consumer .hlx. pages

mokimo:
Correctly fetch placeholders from `.hlx.`-pages from the milo content path and not `/libs or miloLibs`

Resolves: [MWPW-131190](https://jira.corp.adobe.com/browse/MWPW-131190)

### Test instructions
- Observe the "sign in" label to be lowercase before, and capitalised after.
- And/or observe `placeholders` in the network tab to correctly return a 200 fetching from https://main--milo--adobecom.hlx.page/placeholders.json?sheet=feds

**Test URLs:**
- Before: https://main--stockpoc--adobecom.hlx.page/blog/?martech=off&headerqa=global-navigation
- After: https://main--stockpoc--adobecom.hlx.page/blog/?martech=off&headerqa=global-navigation&milolibs=mwpw-131190--milo--mokimo

### DIFF
--- a/libs/blocks/global-navigation/utilities/utilities.js
+++ b/libs/blocks/global-navigation/utilities/utilities.js
@@ -24,19 +24,20 @@ export function toFragment(htmlStrings, ...values) {
   return fragment;
 }
 
-// TODO this is just prototyped
 export const getFedsPlaceholderConfig = () => {
-  const { locale, miloLibs, env } = getConfig();
+  const { locale } = getConfig();
   let libOrigin = 'https://milo.adobe.com';
+
   if (window.location.origin.includes('localhost')) {
     libOrigin = `${window.location.origin}`;
   }
 
-  if (window.location.origin.includes('.hlx.')) {
-    const baseMiloUrl = env.name === 'prod'
-      ? 'https://main--milo--adobecom.hlx.live'
-      : 'https://main--milo--adobecom.hlx.page';
-    libOrigin = miloLibs || `${baseMiloUrl}`;
+  if (window.location.origin.includes('.hlx.page')) {
+    libOrigin = 'https://main--milo--adobecom.hlx.page';
+  }
+
+  if (window.location.origin.includes('.hlx.live')) {
+    libOrigin = 'https://main--milo--adobecom.hlx.live';
   }
 
   return {


libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -24,19 +24,20 @@ export function toFragment(htmlStrings, ...values) {
...
     libOrigin = `${window.location.origin}`;
   }
 
-  if (window.location.origin.includes('.hlx.')) {
-    const baseMiloUrl = env.name === 'prod'
-      ? 'https://main--milo--adobecom.hlx.live'
-      : 'https://main--milo--adobecom.hlx.page';
-    libOrigin = miloLibs || `${baseMiloUrl}`;
+  if (window.location.origin.includes('.hlx.page')) {
+    libOrigin = 'https://main--milo--adobecom.hlx.page';
```
honstar, line 36: I'm a bit confused since the PR title mentions `consumer` but we're essentially hardcoding `https://main--milo--adobecom.hlx.*` as the source for placeholders. Is it correct to assume that gnav placeholders are only managed on milo? Theoretically, every consumer can have its own `placeholders.json`, e.g. https://main--bacom--adobecom.hlx.page/placeholders.json

mokimo: Yes, we specifically want to hardcode the path for the placeholders and always fetch them from 
`milo.adobe.com/placeholders.json` or `https://main--milo--adobe.com.hlx.page/placeholders.json` for every consumer. 

There have been plenty of discussions about generally centralising vs. not - so far we have a hybrid model. 
Content can be anywhere, but placeholders will live within `sharepoint/milo/placeholders.json` and the language variants.

We only use the placeholders for very generic stuff, such as "Sign in" or a "Search" placeholder. There's 0 reason to burden every consumer to create and maintain that file, also it could lead to a lot of inconsistency across consumers.

narcis-radu (approved):

libs/blocks/global-navigation/utilities/utilities.js
overmyheadandbody: Another thing to add: we're using a separate sheet for the navigation placeholders, named 'feds'. This is because we want to keep things consistent, as Okan mentioned. We want to avoid cases where CC uses 'Sign In', but DC uses 'Log in'. It would also be a burden for authors to create and constantly sync these labels. And it will also be problematic for localisation.

honstar (approved):

PR #736, opened 2023-05-18
Address TODOs

narcis-radu:
* Allow Milo consumers to configure ims custom scope
* Create `onImsReady` utility instead of dispatching an event; this should be more suitable for use-cases where logic is loaded **after** ims is ready.
* Call Subscriptions API using Milo configured language / locale
* Remove TODO for closing the search menu on scroll, this is not needed for sticky navigation
* There's a dedicated story for App Launcher, no need for an additional TODO - [MWPW-119227](https://jira.corp.adobe.com/browse/MWPW-119227)
* There's a dedicated story for refactoring the navigation curtain, no need for an additional TODO - [MWPW-131186](https://jira.corp.adobe.com/browse/MWPW-131186)
* Remove TODO for additional checks in Profile Menu dropdown, they are only needed in the standalone implementation which we're not doing anymore.
* Remove TODO from keyboard navigation logic, it has been already implemented.
* Update TODO for global navigation centralised content use-case; we need to check whether this is still required.

Resolves: [MWPW-131196](https://jira.corp.adobe.com/browse/MWPW-131196)

**Test URLs:**
- After: https://quickfix--milo--narcis-radu.hlx.page/drafts/narcis/default?martech=off

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -256,14 +256,11 @@ class Gnav {
   };
 
   loadIMS = () => {
-    const { locale, imsClientId, env } = getConfig();
+    const { locale, imsClientId, imsScope, env } = getConfig();
     if (!imsClientId) return null;
-    // TODO-1 scopes should be defineable by the consumers
-    // We didn't have a use-case for that so far
-    // TODO-2 we should emit an event after the onReady callback
     window.adobeid = {
       client_id: imsClientId,
-      scope: 'AdobeID,openid,gnav',
+      scope: imsScope,
       locale: locale?.ietf?.replace('-', '_') || 'en_US',
       autoValidateToken: true,
       environment: env.ims,
@@ -337,7 +334,6 @@ class Gnav {
   };
 
   decorateAppLauncher = () => {
-    // TODO: review App Launcher component
     // const appLauncherBlock = this.body.querySelector('.app-launcher');
     // if (appLauncherBlock) {
     //   await this.loadDelayed();
@@ -633,7 +629,7 @@ class Gnav {
 
 export default async function init(header) {
   const { locale } = getConfig();
-  // TODO locale.contentRoot is not the fallback we want
+  // TODO locale.contentRoot is not the fallback we want if we implement centralized content
   const url = getMetadata('gnav-source') || `${locale.contentRoot}/gnav`;
   const resp = await fetch(`${url}.plain.html`);
   const html = await resp.text();
--- a/libs/blocks/global-navigation/utilities/onImsReady.js
+++ b/libs/blocks/global-navigation/utilities/onImsReady.js
@@ -1,0 +1,31 @@
+let ready = false;
+let existingCall = null;
+const onInstance = 'onImsLibInstance';
+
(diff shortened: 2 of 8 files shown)


libs/scripts/scripts.js
```diff
@@ -109,6 +109,7 @@ const config = {
   fallbackRouting: 'on',
   links: 'on',
   imsClientId: 'milo',
+  imsScope: 'AdobeID,openid,gnav',
```
auniverseaway, line 112: I wish there was a better place to put this stuff. In five years, the config object is going to be huge.

auniverseaway (approved):

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -337,7 +334,6 @@ class Gnav {
   };
 
   decorateAppLauncher = () => {
-    // TODO: review App Launcher component
```
overmyheadandbody, line 340: We actually do need to review this, albeit after MVP. We do have a story for it, so I guess it's ok to remove the comment for now 🤷

libs/blocks/global-navigation/utilities/onImsReady.js
```diff
@@ -0,0 +1,31 @@
...
+      }
+
+      ready = true;
+      window.removeEventListener(onInstance, onSuccess);
+      clearTimeout(waitTimeout);
+      resolve(instance);
+    };
+
+    window.addEventListener(onInstance, onSuccess);
+    window.dispatchEvent(new window.CustomEvent('getImsLibInstance'));
```
overmyheadandbody, line 25: What's the scenario where it's helpful to know when the logic starts looking for the instance?

overmyheadandbody (approved):
Cool! I'm also preparing a PR that removes a few other TODOs, the list should be pretty limited soon.

libs/blocks/global-navigation/global-navigation.js
narcis-radu: Correct, this is something I mentioned in the description, we can handle the app launcher as part of the upcoming story (post mvp).

libs/blocks/global-navigation/utilities/onImsReady.js
overmyheadandbody: I was not aware that this is needed for IMS to return the instance, my previous comment is irrelevant.

libs/blocks/global-navigation/utilities/onImsReady.js
```diff
@@ -0,0 +1,31 @@
+let ready = false;
+let existingCall = null;
+const onInstance = 'onImsLibInstance';
+
+const onImsReady = (timeout = 3000) => {
```
chrischrischris, line 5: where is this imported / called?

chrischrischris (approved):

libs/blocks/global-navigation/utilities/onImsReady.js
narcis-radu: It's currently not used anywhere, but we want to avoid using `adobeid.onReady` callback given the async nature of blocks. The utility can be used for personalisation, API calls that need user information, tokens etc.

PR #740, opened 2023-05-18
Fix failing tests

Brandon32:
* Fix three failing tests, see https://github.com/adobecom/milo/pull/724

**Test URLs:**
- After: https://bmarshal-fix-failing-tests--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/libs/blocks/gnav/gnav.js
+++ b/libs/blocks/gnav/gnav.js
@@ -520,7 +520,7 @@ class Gnav {
 
   setBreadcrumbSEO = () => {
     const seoEnabled = getMetadata('breadcrumb-seo') !== 'off';
-    if (!seoEnabled) return;
+    if (!seoEnabled || !this.el) return;
     const breadcrumb = this.el.querySelector('.breadcrumbs');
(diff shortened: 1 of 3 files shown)


rgclayton (approved):
Thanks Brandon!

mokimo:
Thanks for taking the time to raise this 🙏 Seems like https://github.com/adobecom/milo/pull/724/files was merged so we can close this

libs/blocks/gnav/gnav.js
```diff
@@ -520,7 +520,7 @@ class Gnav {
 
   setBreadcrumbSEO = () => {
     const seoEnabled = getMetadata('breadcrumb-seo') !== 'off';
-    if (!seoEnabled) return;
+    if (!seoEnabled || !this.el) return;
```
mokimo, line 523: This might be interesting to take in anyway, but the gnav is getting closer and closer to get removed and the Eurovision's global-navigation will take over very soon.
So not sure if it's worth a separate PR in those 1-2-3 weeks until that happened.

PR #743, opened 2023-05-19
MWPW-126707: Global navigation lana logs

mokimo:
Add lana logs for the global-navigation and a few of its complicated event listeners.

Resolves: [MWPW-127607](https://jira.corp.adobe.com/browse/MWPW-127607)

### Context
In the global-navigation we can't rely on implicitly caught errors / promises, as we exist on every page.
We need to ensure (most) of the errors that are originating from our code are also caught by us and our lana clientId. 

Due to the way different modules are dynamically loaded and the very async nature of our code, trying to effectively catch errors originating from _everywhere_ is non-trivial. 

This should provide a good baseline and it might make sense to check consumer lana logs from time to time if we see any uncaught global-navigation related errors, that we didn't catch.

**Test URLs:**
- After: https://mwpw-127607--milo--mokimo.hlx.page/drafts/ramuntea/gnav-refactor?martech=off#

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -20,6 +20,8 @@ import {
   loadBaseStyles,
   yieldToMain,
   selectors,
+  logErrorFor,
+  lanaLog,
 } from './utilities/utilities.js';
 
 import { replaceKey } from '../../features/placeholders.js';
@@ -36,14 +38,6 @@ const CONFIG = {
     keyboardNav: 8000,
   },
 };
-
-function getBlockClasses(className) {
-  const trimDashes = (str) => str.replace(/(^\s*-)|(-\s*$)/g, '');
-  const blockWithVariants = className.split('--');
-  const name = trimDashes(blockWithVariants.shift());
-  const variants = blockWithVariants.map((v) => trimDashes(v));
-  return { name, variants };
-}
 
 // signIn, decorateSignIn and decorateProfileTrigger can be removed if IMS takes over the profile
 const signIn = () => {
@@ -145,13 +139,9 @@ class Gnav {
     this.body = body;
     this.isDesktop = window.matchMedia('(min-width: 900px)');
     this.elements = {};
-    body.querySelectorAll('[class$="-"]').forEach((block) => {
-      const { name, variants } = getBlockClasses(block.className);
-      block.classList.add(name, ...variants);
-    });
   }
 
-  init = async () => {
+  init = () => logErrorFor(async () => {
     this.elements.curtain = toFragment`<div class="feds-curtain"></div>`;
 
     // Order is important, decorateTopnavWrapper will render the nav
@@ -174,7 +164,7 @@ class Gnav {
     }
 
     document.addEventListener('click', closeOnClickOutside);
-  };
+  }, 'Error in global navigation init');
 
   decorateTopNav = () => {
     this.elements.mobileToggle = this.mobileToggle();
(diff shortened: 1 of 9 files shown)


libs/blocks/global-navigation/global-navigation.js
```diff
@@ -145,13 +139,9 @@ class Gnav {
     this.body = body;
     this.isDesktop = window.matchMedia('(min-width: 900px)');
     this.elements = {};
-    body.querySelectorAll('[class$="-"]').forEach((block) => {
```
narcis-radu, line 148: is this not needed anymore?

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -153,3 +158,16 @@ export function expandTrigger({ element } = {}) {
 }
 
 export const yieldToMain = () => new Promise((resolve) => { setTimeout(resolve, 0); });
+
+export const lanaLog = ({ message, e = '' }) => window.lana.log(`${message} ${e.reason || e.error || e.message || e}`, {
+  clientId: 'feds-milo',
+  sampleRate: 1,
+});
+
+export const logErrorFor = async (fn, message) => {
```
chrischrischris, line 167: Do you think we should add this directly to lana.js?  Also wonder if we should have a sync and async version.

chrischrischris (commented):
Would it be possible to use the try/catch surrounding `gnav.init()` to catch errors and log?

libs/blocks/global-navigation/utilities/utilities.js
mokimo: Might be overkill, since usually lana consumers can rely on implicit logging.

All this wrapping just makes the code more complex, and we shouldn't encourage that. GNAV doesn't have much of a choice. Milo blocks should be able to rely on this https://github.com/adobecom/milo/blob/main/libs/utils/utils.js#L376

mokimo:
> Would it be possible to use the try/catch surrounding gnav.init() to catch errors and log?

It's all a bit tedious. A lot of promises, callbacks (IMS), etc. - So only the `gnav.init()` alone doesn't catch a lot.
For now it's a compromise of getting logs and trying to introduce only a few new try/catch blocks

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -233,23 +223,28 @@ class Gnav {
...
+          loadBlock('../features/search/gnav-search.js'),
+          loadStyles('features/profile/dropdown.css'),
+          loadStyles('features/search/gnav-search.css'),
+        ]);
+        this.ProfileDropdown = ProfileDropdown;
+        this.appLauncher = appLauncher;
+        this.Search = Search;
+        resolve();
+      } catch (e) {
+        lanaLog({ message: 'Error within loadDelayed', e });
```
overmyheadandbody, line 245: This [line](https://github.com/adobecom/milo/blob/f3012c8045c629f83a2ed41d91c54761c6adf23a/libs/blocks/global-navigation/global-navigation.js#L274) has a message specific to 'Global Navigation'. I know that we have our own client ID, but since multiple platforms might have a `loadDelayed` method and there might be error spillover, I'd suggest we make this a bit more particular to 'Global Navigation' as well.

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -83,19 +85,22 @@ export function loadBlock(path) {
...
-    ]);
-
-    resolve({
-      decorateMenu,
-      decorateLinkGroup,
-    });
+  if (cachedDecorateMenu) return cachedDecorateMenu;
+
+  let resolve;
+  cachedDecorateMenu = cachedDecorateMenu || new Promise((_resolve) => {
```
overmyheadandbody, line 91: If `cachedDecorateMenu` exists, wouldn't the function have returned in `if (cachedDecorateMenu) return cachedDecorateMenu;`? So the OR might not be necessary.
```suggestion
  cachedDecorateMenu = new Promise((_resolve) => {
```

overmyheadandbody (approved):
I think this is a good start. I can't say I'm excited about having init calls wrapped inside other methods, but I also don't see a better solution for the moment

chrischrischris (approved):

PR #744, opened 2023-05-19
Serve non-critical features from delayed

narcis-radu:
* Move all non-critical and delayed code from `utils.js` to `delayed.js`
* Ensure `loadDelayed` is automatically called from `utils.js` via dynamic import
* Update milo-college to remove `loadDelayed` from `scripts.js` - https://github.com/adobecom/milo-college/pull/2

Resolves: [MWPW-131408](https://jira.corp.adobe.com/browse/MWPW-131408)

**Test URLs:**
- After: https://load-delayed--milo--narcis-radu.hlx.page/drafts/narcis/default?martech=off


libs/scripts/delayed.js
```diff
@@ -10,11 +10,47 @@
...
+const loadDelayed = ([
+  getConfigFunc,
+  getMetadataFunc,
+  loadScriptFunc,
+  loadStyleFunc,
+], delay = 3000) => {
+  const getConfig = getConfigFunc;
+  const getMetadata = getMetadataFunc;
+  const loadScript = loadScriptFunc;
+  const loadStyle = loadStyleFunc;
```
chrischrischris, line 39: Why not just name the param `getConfig` and not need the assignment?

libs/scripts/delayed.js
```diff
@@ -10,11 +10,47 @@
...
-  const { default: interlinks } = await import('../features/interlinks.js');
-  interlinks('/keywords.json');
-}
-delayed();
+const loadDelayed = ([
+  getConfigFunc,
+  getMetadataFunc,
+  loadScriptFunc,
+  loadStyleFunc,
+], delay = 3000) => {
```
chrischrischris, line 35: nit: move to const: `const DELAY = 3000`

libs/utils/utils.js
```diff
@@ -766,23 +726,12 @@ export async function loadArea(area = document) {
 
   // Load everything that can be deferred until after all blocks load.
   await loadDeferred(area, areaBlocks, config);
+  const { default: delayed } = await import('../scripts/delayed.js');
```
chrischrischris, line 729: This should be wrapped in a `if (isDoc)`

libs/utils/utils.js
```diff
@@ -766,23 +726,12 @@ export async function loadArea(area = document) {
...
-        const path = `${getConfig().locale.contentRoot}/keywords.json`;
-        import('../features/interlinks.js').then((mod) => { mod.default(path); resolve(mod); });
-      } else {
-        resolve(null);
-      }
-      import('./samplerum.js').then(({ sampleRUM }) => sampleRUM('cwv'));
-    }, delay);
-  });
+export function loadDelayed() {
+  // TODO: remove after all consumers have stopped calling this method
```
chrischrischris, line 734: Should this still load the delayed.js file so consumers that call it still get that functionality?  Maybe move the import and call in loadArea into here and call it from loadArea?

narcis-radu: Consumers would still get the functionality as part of `loadArea`, which they execute in `loadPage`.
If I move the logic to `utils.loadDelayed` we would get the logic executed twice on consuming projects: once by `loadArea` and once from the consumers - https://github.com/adobecom/milo-college/blob/main/scripts/scripts.js#L64-L65

auniverseaway: I think this is exactly how we want to do this. Consumers won't know the difference. They continue to call the function, but it doesn't run. Once consumers update (we can do this today), we can safely remove this.

auniverseaway (approved):

libs/features/privacy.js
```diff
@@ -0,0 +1,23 @@
...
+    documentLanguage: true,
+  };
+  loadScript('https://www.adobe.com/etc.clientlibs/globalnav/clientlibs/base/privacy-standalone.js');
+
+  // TODO: remove once gnav block is decommissioned
+  const privacyTrigger = document.querySelector('footer a[href*="#openPrivacy"]');
+  privacyTrigger?.addEventListener('click', (event) => {
+    event.preventDefault();
+    window.adobePrivacy?.showPreferenceCenter();
+  });
```
overmyheadandbody, line 20: This has been recently updated, is there a reason for it being reverted to its previous state? https://github.com/adobecom/milo/blob/main/libs/utils/utils.js#L653-L658

overmyheadandbody (approved):

libs/scripts/delayed.js
```diff
@@ -10,11 +10,40 @@
...
+  const jarvis = getMetadata('jarvis-chat');
+  if (!config.jarvis?.id || !config.jarvis?.version) return;
+  if (jarvis === 'on') {
+    const { initJarvisChat } = await import('../features/jarvis-chat.js');
+    initJarvisChat(config, loadScript, loadStyle);
+  }
+};
+
+export const loadPrivacy = async (getConfig, loadScript) => {
+  const { default: initPrivacy } = await import('../features/privacy.js');
```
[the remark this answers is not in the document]
narcis-radu: good point, we can do that!

mokimo (approved):

PR #745, opened 2023-05-19
FEDS Gnav critical bug squashing

overmyheadandbody:
This aims to close most of the critical bugs found while testing the new FEDS Gnav on consumers currently using the Milo Gnav. It solves:
- [MWPW-131240](https://jira.corp.adobe.com/browse/MWPW-131240) - changes the approach of loading the footer icons to lower the risk of policy issues. Icons are added to the footer as a (hidden) element, instead of including them directly via `href`. Icon names have been updated to lower the risk of duplicate IDs on the page;
- [MWPW-131193](https://jira.corp.adobe.com/browse/MWPW-131193) - adapted several CSS `box-sizing` values to ensure styles remain consistent even if a consumer explicitly sets `body, body *, body *:before, body *:after {box-sizing: border-box}` on their pages;
- [MWPW-131235](https://jira.corp.adobe.com/browse/MWPW-131235) - ensure clicking navigation items doesn't scroll the page to the top. `event` properties have been passed to the `trigger` method config. If an `event` exists, its default behaviour is prevented;
- [MWPW-131438](https://jira.corp.adobe.com/browse/MWPW-131438) - we usually consider any `h2` elements to be main navigation items. However, there might be scenarios where `h2`s are added in other features (aka 'blocks' in the usual Milo workflow). We now check whether `h2`s are nested inside known features to avoid element duplication;
- [MWPW-131182](https://jira.corp.adobe.com/browse/MWPW-131182) - revisit brand decoration logic:
    - try image decoration via `decorateSVG`. This also attaches `alt` text, if provided (via pipe: `<image source> | <alt text>`);
    - if image could not be decorated as an SVG, try decorating it as png/jpg/jpeg using a similar approach to the SVG one;
    - if an image is not authored or could not be decorated, provide default fallback to Adobe logo with an `alt` value of 'Adobe, Inc.', as it is currently set on production;
    - look for image path in both its `href` and `textContent` values to support backwards compatibility with how different consumers initially authored this block;
    - allow an icon-only brand via the use of the `image-only` modifier in the feature name.
- [MWPW-131185](https://jira.corp.adobe.com/browse/MWPW-131185) - streamline menu decoration logic, combining `decorateMenuElement` and `decorateLinkList` into a single method to avoid any overlap between the two;
- [MWPW-131189](https://jira.corp.adobe.com/browse/MWPW-131189) - ensure the element position from the analytics `daa-ll` values are correct in their context. The position is now reset when a new column starts;
- address TODO - clicking anywhere outside an open dropdown should close it;
- small style improvements:
    - copy the navigation dropdown styles to the 'Sign In' one for consistency;
    - ensure the bottom navigation border is always visible when a dropdown is open.

**Analytics position screenshots**

| Before | After |
| Analytics positions - before | Analytics positions - after |

**Test URLs:**
- Before: https://main--milo--overmyheadandbody.hlx.page/drafts/ramuntea/gnav-refactor?martech=off
- After: https://feds-address-critical-bugs--milo--overmyheadandbody.hlx.page/drafts/ramuntea/gnav-refactor?martech=off


libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -132,7 +132,8 @@ export function closeAllDropdowns({ e } = {}) {
  * @param {*} param0.element - the DOM element of the trigger to expand
```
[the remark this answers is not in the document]
overmyheadandbody: Good catch, I can update it. However, it's one of the very few places where we actually have JSDoc comments. Is it worth keeping in just these few select places or should we remove it altogether?

auniverseaway (approved):

libs/blocks/global-navigation/utilities/utilities.js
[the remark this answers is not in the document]
narcis-radu: I'm fine with removing it to be honest

overmyheadandbody:
Fixed issue surfaced by failing unit test and amended method documentation.

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -405,19 +443,21 @@ class Gnav {
   };
 
   decorateBrand = () => {
-    const brandBlock = this.body.querySelector('[class^="gnav-brand"]');
-    if (!brandBlock) return null;
-    const imgRegex = /(\.png|\.svg|\.jpg|\.jpeg)$/;
+    const brandBlock = this.body.querySelector('.gnav-brand');
```
mokimo, line 446: The selector would be good in `utilities.(features.)selectors`

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -405,19 +443,21 @@ class Gnav {
...
+    if (!brandBlock) return '';
+
+    const imgRegex = /(\.png|\.svg|\.jpg|\.jpeg)/;
     const brandLinks = [...brandBlock.querySelectorAll('a')];
-    const image = brandLinks.find((brandLink) => imgRegex.test(brandLink.href));
-    const link = brandLinks.find((brandLink) => !imgRegex.test(brandLink.href));
+    const image = brandLinks.find((brandLink) => imgRegex.test(brandLink.href)
+      || imgRegex.test(brandLink.textContent));
```
mokimo, line 452: Wdyt about checking if the image is an SVG? And then conditionally calling either `getSVG` or `getIMG` instead of putting both in one method?

What about `feds-navLink-image` and the logo - do they need that logic as well or is the brand special?

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -35,6 +36,14 @@ const CONFIG = {
     loadDelayed: 2000,
     keyboardNav: 8000,
   },
+  features: [
```
mokimo, line 39: Should we put those into `utilities.selectors` ? Wouldn't mind having a nested object, as long as we have all of the selectors in a central place.

mokimo (approved):

libs/blocks/global-navigation/global-navigation.js
overmyheadandbody: These features only refer to the initial raw HTML, we're using them just to know how to decorate a section, they have no use beyond this. It's basically a list of features supported by the global-navigation.

The selectors we have in our files, however, refer to the already decorated elements. So to me, the two have different purposes, that's why I didn't really want to mix them. Currently, their main purpose is [this](https://github.com/adobecom/milo/pull/745/files#diff-628bf5a25903978c5b36da4943ec813da171346d24bd64f80907f012a840b098R497), to ensure we don't decorate an element as a main navigation item if it is nested inside another known feature.

I could be persuaded to move them over, if you think the different nature of their purpose can co-exist and might bring a benefit for the future.

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -121,14 +130,33 @@ const setupKeyboardNav = async () => {
...
+
+  try {
+    // Try to decorate image as SVG
+    const decoratedSvg = decorateSVG(image);
+    // 'decorateSVG' might return the original element if decoration fails
+    // or the picture wrapped in an anchor element in certain cases
+    const svg = decoratedSvg instanceof HTMLPictureElement
+      ? decoratedSvg : decoratedSvg.querySelector('picture');
+    if (svg) return svg;
+  } catch (e) {
```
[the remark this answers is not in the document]
overmyheadandbody: `decorateSVG` can return either the original anchor element or a decorated anchor element with a `picture` child element or a `picture` element directly, so this is the safest option as opposed to performing checks every step of the way.

libs/blocks/global-navigation/global-navigation.js
overmyheadandbody: The brand block is definitely a special use-case. Not only is it already used by all consumers, but each have authored it in a slightly different way. We promised backwards compatibility, so this aims to be as inclusive as possible without being too particular. The link images are usually added directly as PNG/JPG. The logo needs to be refactored (https://jira.corp.adobe.com/browse/MWPW-131183) and it will be based on logic from this brand feature.

I actually did try to split up the code, but due to how the authoring is done today by different consumers, that approach actually ended up having more checks scattered over multiple functions, so I decided to keep it all in one place. Additionally, since the Logo feature needs to be refactored too, we might be able to leverage `getBrandImage` there too.

libs/blocks/global-navigation/global-navigation.js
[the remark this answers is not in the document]
mokimo: But `decorateSVG` would never throw an error, just return al element, so you'd never get in the catch block.
What am I missing here? 🤔

libs/blocks/global-navigation/global-navigation.js
mokimo: Oh yes, it's `gnav-brand` instead of `feds-brand` - you're correct.
In this case mixing them would lead to more confusion than help. Thanks for explaining, let's keep it as is!

libs/blocks/global-navigation/global-navigation.js
mokimo: Since this is also `gnav-brand` let's keep it here.

libs/blocks/global-navigation/global-navigation.js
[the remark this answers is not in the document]
overmyheadandbody: It can return either an `a` tag or a `picture` one. `const svg = decoratedSvg instanceof HTMLPictureElement ? decoratedSvg : decoratedSvg.querySelector('picture');` would fail if we get back the original anchor. Instead of performing a series of checks, since we're using an external method which might change anyway, I preferred to go for the best case scenario and wrap it in a `try/catch`

narcis-radu (approved):

libs/blocks/global-navigation/utilities/menu/menu.js
```diff
@@ -68,29 +68,56 @@ const decorateLinkGroup = (elem, index) => {
   return linkGroup;
 };
 
-// Decorate a list of simple links
-const decorateLinkList = ({ list, className = 'feds-navLink', includeCta = true } = {}) => {
-  list.querySelectorAll('a').forEach((link, index) => {
+const decorateElements = ({ elem, className = 'feds-navLink', parseCtas = true, index = { position: 0 } } = {}) => {
```
chrischrischris, line 71: Why use `index.position` vs just `index`?  Doesn't look like the index obj has any other properties

chrischrischris (approved):

honstar:
@overmyheadandbody : there is a merge conflict, can you please check?

overmyheadandbody:
Hey @honstar! Merging https://github.com/adobecom/milo/pull/743 caused a conflict in this PR, I've fixed it now, should be good to go

libs/blocks/global-navigation/utilities/menu/menu.js
overmyheadandbody: Because the index needs to be synced across blocks used as part of a menu, I pass the object wherever it's needed, thus all blocks can update its `index` property when they need to and all of the other blocks get the update via the object. It's sort of a shared context without declaring a global variable (which I've also tried, but since there is so much async logic, it leads to inconsistent values).

overmyheadandbody:
Since this has the appropriate approvals, could it be merged? Thanks!

mokimo:
I went ahead and merged it so we can enable the navigation for milo internal consumers and it's not used on production now.

PR #751, opened 2023-05-19
MWPW-130474: remove placeholder syntax from price links

Axelcureno:
Removes placeholder syntax from price links generated by OST

Resolves: [MWPW-130474](https://jira.corp.adobe.com/browse/MWPW-130474)

**Test URLs:**
- After: https://mwpw-130474--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/libs/blocks/ost/ost.css
+++ b/libs/blocks/ost/ost.css
@@ -1,0 +1,1 @@
+/* created due of block nature in Milo */
--- a/libs/blocks/ost/ost.css
+++ b/libs/blocks/ost/ost.css
@@ -1,1 +1,30 @@
-/* created due of block nature in Milo */
+body.tool-ost .ost {
+  position: fixed;
+  display: flex;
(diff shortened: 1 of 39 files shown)


yesil (approved):

mokimo (approved):
Please fix the failing OST unit test before merging, otherwise looks good 👍

libs/blocks/ost/ost.css
```diff
@@ -0,0 +1 @@
+/* created due of block nature in Milo */
```
honstar, line 1: FWIW, I started #752 to discuss this requirement.

Axelcureno: Since the conversation is still ongoing may I suggest we leave this for now for the sake of removing the 404?

Axelcureno: File is no longer empty :)

PR #756, opened 2023-05-22
MWPW-129575 Authorable Tooltip

TsayAdobe:
* An author can add a tooltip on a text or icon with the format
  ```
  This is a [hover text](## "my tooltip text") example. 
  ```
  or
  ```
  This is an icon [:info:](## "Tooltip") with a tooltip.
  ```
* The tooltip position can be specified with `top`, `right`, `bottom`, and `left` following the tooltip text. The default is `right`.
  ```
  This is an icon [:info:](## "Tooltip text" top) with a tooltip.
  ``` 
* On a desktop browser, mouse hovering and unhovering will show and hide the tooltip. On a mobile browser, click on the target text or icon and somewhere else.
* [Adobe Spectrum](https://spectrum.adobe.com/page/tooltip/) tooltip design.
* The GitHub discussion [thread](https://github.com/orgs/adobecom/discussions/650).

Resolves: [MWPW-129575](https://jira.corp.adobe.com/browse/MWPW-129575)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/tsay/mwpw-129575-test?martech=off
- After: https://mwpw-129575--milo--tsayadobe.hlx.page/drafts/tsay/mwpw-129575-test?martech=off

**DocCloud Test URLs:**
- Before: https://main--dc--adobecom.hlx.page/drafts/tsay/pdf-to-ppt?martech=off
- After: https://main--dc--adobecom.hlx.page/drafts/tsay/pdf-to-ppt?martech=off&milolibs=mwpw-129575--milo--tsayadobe


seanchoi-dev (approved):
Looks good to me.

auniverseaway (changes requested):
I think this is too complicated and we should re-think the authoring paradigm.

We don't want to introduce markdown in word.

I would much prefer we use existing patterns that have already been established.

Something like:

```html

<a href="{any-page}#_left">:tooltip: | My tooltip contents</a>

```

auniverseaway:
I think my suggestion is inline with @vgoodric's proposal as well: https://github.com/orgs/adobecom/discussions/650#discussioncomment-5713181

Blainegunn:
@auniverseaway we looked into the existing options. And we'd be happy to try something similar. 
A few issues we wanted your take on. 
1. Should we not use the icon placeholders if tooltips are needed since they would rely on hyperlinks to the image e?
2. If we go the route of the alt text should we use || to indicate tooltips as to not confuse it with alt text? Or what would the team like to see? 

Also. I think this approach from @TsayAdobe works and could help mitgate some of the confusion people may have with alt text v tooltips and the hyperlinks in Word. How strong is your change request, like do we have to change or is there room for discussion?

TY!

mokimo:
I'd second @auniverseaway opinion - sticking to existing authoring paradigms will also make it easier codewise to standardize things.

If we use the `text | context` way of doing things, I'd love to see a util that extracts the text and context which can be used both for alt texts and tooltips and potentially other stuff in the future too.

Blainegunn:
@mokimo 

>IMO no, just stick to `|` and document that in the authoring docs on how to author a tooltip that authors can be pointed to.

If both tooltips and alt texts use `|` how will we know how to differentiate between which is which? They would both be images with hyperlinks with a `|` in them... how can we determine which is a tooltip versus alt text. 
I'm not trying to be difficult I just want to get a better understanding before this is reworked.

thank you

Blainegunn:
ok ok 
so.. `link to svg#_tooltip | tooltip text blah blah`.
but if the tooltip is need at the bottom `link to svg#_tooltip_bottom | tooltip text blah blah`.
top; `link to svg#_tooltip_top | tooltip text blah blah`. 
versus 
[:info:](## "Tooltip" bottom) tooltip text blah blah.

imo, seems more confusing, complicated and like more work to build the experience... but we shall do it this the proposed way

Thank you for all the feedback PR update on the way!

auniverseaway:
Here's my suggestion: https://github.com/adobecom/milo/pull/775/files

TsayAdobe:
> Here's my suggestion: https://github.com/adobecom/milo/pull/775/files

The icon is hard-coded. What if an author wants a different icon. The link "https://www.adobe.com" serves nothing. How about using the address to identify this is a tooltip and then there is a flexibility for the text. e.g. 

Hyperlink Text: `:info: | This is the content of my tooltip`
Hyperlink Address:  `tel:tooltip#top`

auniverseaway:
@TsayAdobe I've updated the PR. I've removed the link requirement as it didn't make sense.

As for different icons, I think we start with "tooltip" icon and see how it's used in the wild. We do not like to build for hypotheticals. When someone comes to us with a real-world design that uses something different than, "i" we iterate.

honstar:
@TsayAdobe : looks like we're gravitating towards @auniverseaway's proposal #775. It'd be great to close this PR in favor of #775.

cc: @Blainegunn

honstar:
In general, I think a lot of discussions around the actual design of authorable tooltips could have been held in #650. Not trying to be difficult here, just sayin... :)

auniverseaway:
@honstar you're 100% correct. The suggestions @vgoodric made were not followed and this PR was raised instead.

TsayAdobe:
Move the improvement to https://github.com/adobecom/milo/pull/775

PR #763, opened 2023-05-24
Highlight expanded navigation menu on mobile

narcis-radu:
* Highlight expanded accordion menu on mobile devices
* Based on @overmyheadandbody's suggestion I have created a follow-up story to handle all use-cases for padding inside accordion sections on mobile devices.

Resolves: [MWPW-131194](https://jira.corp.adobe.com/browse/MWPW-131194)

**Test URLs:**
- After: https://mwpw-131194--milo--narcis-radu.hlx.page/drafts/narcis/default?martech=off


mokimo:
On AEM, we only highlight we active section, if that is open

Milo
Screenshot 2023-05-24 at 14 52 03

AEM
Screenshot 2023-05-24 at 14 52 09

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -561,7 +562,11 @@ class Gnav {
           <div class="feds-navItem${isSectionMenu ? ' feds-navItem--section' : ''}">
             ${dropdownTrigger}
           </div>`;
-        dropdownTrigger.addEventListener('click', (e) => trigger({ element: dropdownTrigger, event: e }));
+        dropdownTrigger.addEventListener('click', (e) => {
+          trigger({ element: dropdownTrigger, event: e });
+          const isOpen = dropdownTrigger.getAttribute('aria-expanded') === 'true';
+          if (isOpen) triggerTemplate.classList.add(selectors.activeDropdown.replace('.', ''));
```
mokimo, line 568: `trigger()` actually returns `true|false` based on if it was opened or not, so you don't need the isOpen check

```suggestion
          const hasBeenOpened = trigger({ element: dropdownTrigger, event: e });
          if (hasBeenOpened) triggerTemplate.classList.add(selectors.activeDropdown.replace('.', ''));
```

mokimo (approved):
Cool, would love this check to be added to the one or other unit test. There should also be some tests opening a headline on mobile - so maybe you can just add a check weather the activeClass is added

mokimo (approved):

overmyheadandbody (approved):
Thanks for integrating all the suggestions! Looks good now! 👍

PR #770, opened 2023-05-24
Let region nav to use prefix instead content root.

seanchoi-dev:
- Bug fix for region nav when content root is used.

Resolves: [MWPW-131680](https://jira.corp.adobe.com/browse/MWPW-131680)

**Test URLs:**
CC:
- Before: https://main--cc--adobecom.hlx.live/creativecloud/video/hub/ideas/what-is-a-contrast-cut#langnav
- After: https://main--cc--adobecom.hlx.live/creativecloud/video/hub/ideas/what-is-a-contrast-cut?milolibs=region-nav-fix#langnav

DC:
- Before: https://main--dc--adobecom.hlx.live/fr/acrobat/online/word-to-pdf#langnav
- After: https://main--dc--adobecom.hlx.live/fr/acrobat/online/word-to-pdf?milolibs=region-nav-fix#langnav

BACOM (site doesn't have issue now)
- Before: https://main--bacom--adobecom.hlx.page/kr/customer-success-stories
- After: https://main--bacom--adobecom.hlx.page/kr/customer-success-stories?milolibs=region-nav-fix

### DIFF
--- a/libs/blocks/region-nav/region-nav.js
+++ b/libs/blocks/region-nav/region-nav.js
@@ -22,7 +22,7 @@ function decorateLink(link, config, path) {
   const prefix = linkParts[1] || 'us';
   let { href } = link;
   if (href.endsWith('/')) href = href.slice(0, -1);
-  link.href = `${href}${config.contentRoot || ''}${path}`;
+  link.href = `${href}${path}`;
   link.addEventListener('click', (e) => {
     /* c8 ignore next 2 */
     e.preventDefault();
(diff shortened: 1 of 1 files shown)


Blainegunn (approved):
ty!

Blainegunn (approved):
ty!

auniverseaway (approved):

vhargrave (changes requested):
We're still using the contentRoot on line 12 for the fallback route if a route we try navigating to doesn't exist. As said in slack, I don't see how we can cleanly define a fallback route without the contentRoot 🙁

seanchoi-dev:
I discussed with @vhargrave and this is good to merge for now, since it is breaking production for dc and cc, but we definitely want to revisit specially for fallback redirection on handleEvent() since it should redirect the site's root instead adobe.com/{locale prefix}.  We may need something like `config.siteRoot` for this kind of cases.

PR #773, opened 2023-05-24
MWPW-131454 Use config.locale.dir instead textInfo

TsayAdobe:
* Firefox doesn't support textInfo. Add `dir` in `config.locale` for rtl languages.
* If `dir` is in config.locale, use it. Otherwise, `dir` is default to `ltr`.
* Please use Firefox to test the before and after links.

Resolves: [MWPW-131454](https://jira.corp.adobe.com/browse/MWPW-131454)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/mena_ar/drafts/tsay/mwpw-131454?martech=off
- After: https://mwpw-131454--milo--tsayadobe.hlx.page/mena_ar/drafts/tsay/mwpw-131454?martech=off

PR #777, opened 2023-05-25
MWPW-125482 Overriding text flow direction via metatag

aishwaryamathuria:
Give an override option via metatag for dir option only via content-direction metatag.
https://github.com/orgs/adobecom/discussions/746

Resolves: [MWPW-125482](https://jira.corp.adobe.com/browse/MWPW-125482)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/mena_ar/drafts/mathuria/test?martech=off
- After: https://rtlfix--milo--aishwaryamathuria.hlx.page/mena_ar/drafts/mathuria/test?martech=off


aishwaryamathuria:
Did not add the change to add dir property in config.locales list as they are already being added in the PR https://github.com/adobecom/milo/pull/773

narcis-radu:
I see there is another open PR that handles this issue - https://github.com/adobecom/milo/pull/773

salonijain3:
yes @aishwaryamathuria , probably @TsayAdobe is trying to fix the safari issue. Can you build upon his PR?

narcis-radu:
This PR introduces a way to override the content direction via metadata, but it does not fix the Firefox issue:
`Uncaught TypeError: (new Intl.Locale(...)).textInfo is undefined`. I think we need to combine @TsayAdobe's logic for handling Firefox and this enhancement to override the content direction via metadata.

libs/utils/utils.js
```diff
@@ -151,7 +151,8 @@ export const [setConfig, getConfig] = (() => {
       config.autoBlocks = conf.autoBlocks ? [...AUTO_BLOCKS, ...conf.autoBlocks] : AUTO_BLOCKS;
       document.documentElement.setAttribute('lang', config.locale.ietf);
       try {
-        document.documentElement.setAttribute('dir', (new Intl.Locale(config.locale.ietf)).textInfo.direction);
+        const contentDir = getMetadata('content-direction');
+        document.documentElement.setAttribute('dir', contentDir || config.locale.dir || (new Intl.Locale(config.locale.ietf)).textInfo.direction);
```
narcis-radu, line 155:
```suggestion
        document.documentElement.setAttribute('dir', contentDir || config.locale.dir || 'ltr';
```

test/utils/utils.test.js
```diff
@@ -229,8 +229,8 @@ describe('Utils', () => {
         config.locales = {
           '': { ietf: 'en-US', tk: 'hah7vzn.css' },
           africa: { ietf: 'en', tk: 'pps7abe.css' },
-          il_he: { ietf: 'he', tk: 'nwq1mna.css' },
-          mena_ar: { ietf: 'ar', tk: 'dis2dpj.css' },
+          il_he: { ietf: 'he', tk: 'nwq1mna.css', dir: 'rtl' },
+          mena_ar: { ietf: 'ar', tk: 'dis2dpj.css', dir: 'rtl' },
```
narcis-radu, line 233: make sure your tests cover both `ltr` and `rtl` use-cases

libs/scripts/scripts.js
```diff
@@ -75,10 +75,10 @@ const locales = {
...
   ru: { ietf: 'ru-RU', tk: 'aaz7dvd.css' },
   ua: { ietf: 'uk-UA', tk: 'aaz7dvd.css' },
-  il_he: { ietf: 'he', tk: 'nwq1mna.css' },
-  ae_ar: { ietf: 'ar', tk: 'nwq1mna.css' },
-  mena_ar: { ietf: 'ar', tk: 'dis2dpj.css' },
-  sa_ar: { ietf: 'ar', tk: 'nwq1mna.css' },
+  il_he: { ietf: 'he', tk: 'nwq1mna.css', dir: 'rtl' },
+  ae_ar: { ietf: 'ar', tk: 'nwq1mna.css', dir: 'rtl' },
+  mena_ar: { ietf: 'ar', tk: 'dis2dpj.css', dir: 'rtl' },
+  sa_ar: { ietf: 'ar', tk: 'nwq1mna.css', dir: 'rtl' },
```
narcis-radu, line 81: Consumers will have to be notified about this change, they need to implement the `dir` property in their config

libs/utils/utils.js
mokimo, line 155: // @TsayAdobe Either you can add the contentDir based the [discussion](https://github.com/orgs/adobecom/discussions/746#discussioncomment-5981042) and close https://github.com/adobecom/milo/pull/773/files or we use this PR. I'm fine with either. 
```suggestion
        document.documentElement.setAttribute('dir', contentDir || config.locale.dir || 'ltr')
```

aishwaryamathuria:
> This PR introduces a way to override the content direction via metadata, but does not fix the Firefox issue: `Uncaught TypeError: (new Intl.Locale(...)).textInfo is undefined`. I think we need to combine @TsayAdobe's logic for handling Firefox and this enhancement to override the content direction via metadata.

@narcis-radu Sure.. I can pull the logic into this PR and resolve the conflicts...

test/utils/utils.test.js
narcis-radu: Let's add tests for `rtl` as well

aishwaryamathuria:
@auniverseaway @chrischrischris @narcis-radu @TsayAdobe @salonijain3 @amitbikram @dstrong23 @Ruchika4 
Could you please add your review?

libs/utils/utils.js
aishwaryamathuria: @mokimo @TsayAdobe I have made the requested changes. Could you pelase review

mokimo (approved):
Just FYI on what happens with this merge: this will break RTL all non-firefox browsers for consumers until they add the RTL config in one form or another.
This should definitely be communicated with all the consuming teams so that they can adapt their configs.

libs/utils/utils.js
```diff
@@ -151,7 +151,8 @@ export const [setConfig, getConfig] = (() => {
       config.autoBlocks = conf.autoBlocks ? [...AUTO_BLOCKS, ...conf.autoBlocks] : AUTO_BLOCKS;
       document.documentElement.setAttribute('lang', config.locale.ietf);
       try {
-        document.documentElement.setAttribute('dir', (new Intl.Locale(config.locale.ietf)).textInfo.direction);
+        const contentDir = getMetadata('content-direction');
+        document.documentElement.setAttribute('dir', contentDir || config.locale.dir || 'ltr');
```
aishwaryamathuria, line 155: @mokimo @narcis-radu Initially the change here was following 
document.documentElement.setAttribute('dir', contentDir || config.locale.dir || (new Intl.Locale(config.locale.ietf)).textInfo.direction);

This was as discussed in the github discussion https://github.com/orgs/adobecom/discussions/746 and did not break the dir property for browsers other that firefox

This was later changed to  contentDir || config.locale.dir || 'ltr' as per the last comment.

Could you please confirm if I should change it to fallback to Intl.Locale(config.locale.ietf)).textInfo.direction so that the teams have time to add the support for Firefox by adding the dir property?

aishwaryamathuria: @chrischrischris @auniverseaway Could you please share your review on the same?

test/utils/utils.test.js
chrischrischris: still need rtl test

libs/utils/utils.js
chrischrischris: So that we can merge without waiting for consumers:
```suggestion
        document.documentElement.setAttribute('dir', contentDir || config.locale.dir || Intl.Locale(config.locale.ietf))?.textInfo?.direction) || 'ltr');
```

We can remove Intl.Locale once all consumers have updated.

test/utils/utils.test.js
aishwaryamathuria: @chrischrischris The rtl and ltr test are already there in the Unit tests.. we have added the dir property for the same in that config .

libs/utils/utils.js
aishwaryamathuria: @chrischrischris I have made the suggested changes

libs/scripts/scripts.js
aishwaryamathuria: @narcis-radu Yes.. If there are no changes required in the PR.. We will share a notification with the teams on milo-community channel before merge

test/utils/utils.test.js
narcis-radu: You are correct! But I think one existing test would fail, maybe you can check - https://github.com/adobecom/milo/blob/main/test/utils/utils.test.js#L258

```js
setConfigWithPath('/ua/solutions');
expect(document.documentElement.getAttribute('dir')).null;
``` 

Based on what you set - `document.documentElement.setAttribute('dir', contentDir || config.locale.dir || new Intl.Locale(config.locale.ietf)?.textInfo?.direction || 'ltr');`, this will return `ltr` and we're expecting to be null

aishwaryamathuria: @narcis-radu The test was completing successfully because it was only checking that if the dir attribute set fails it should not throw error and the value should be null.
I have added a check to verify that the Intl should run only if config.locale.ietf is set and if not it should not try to find textinfo from it and fallback to 'ltr'
Could you please review?

narcis-radu (approved):

chrischrischris (approved):

PR #780, opened 2023-05-25
MWPW-131645 Linking images in milo via alt text

ryanmparrish:
* Added the ability to link an image via the alt text input field in word. ex. `https://adobe.com#_blank | Alt text description`
* Supports #_blank in the url string to set the link target ⬆️ 

Resolves: [MWPW-131645](https://jira.corp.adobe.com/browse/MWPW-131645)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rparrish/image-links
- After: https://rparrish-img-links--milo--adobecom.hlx.page/drafts/rparrish/image-links


ryanmparrish:
I just realized I need to write tests for this feature.... Ill get those included asap.

libs/utils/utils.js
```diff
@@ -419,6 +419,40 @@ export function decorateSVG(a) {
...
+  const images = a.querySelectorAll('img');
+  if (!images.length) return a;
+  [...images].forEach((img) => {
+    // Check if the image has an alt tag with a url & pipe - 'https... |'
+    if (!(hasUrlWithPipe(img.alt))) return a;
+    try {
+      // //Mine for URL and alt text
+      const splitText = img.alt.split('|');
+      const textUrl = new URL(splitText.shift().trim());
+      const altText = splitText.join('|').trim();
```
honstar, line 437: This looks wrong, joining the single element array at this point using the pipe `|` character. `splitText.trim()` should suffice.

test/utils/mocks/body.html
```diff
@@ -9,6 +9,9 @@
     <!-- Picture Decorating -->
     <p><em><picture><img src="/test/mocks/media_.png"/></picture></em></p>
     <p><strong><picture><img src="/test/mocks/media_.png"/></picture></strong></p>
+    <!-- Picture Link Decorating -->
+    <p><picture><img src="/test/mocks/media_.png" alt="https://www.adobe.com#_blank | Image description"/></picture></p>
```
honstar, line 13: There should be specific unit-tests for checking image links.

honstar (changes requested):

libs/utils/utils.js
```diff
@@ -419,6 +419,40 @@ export function decorateSVG(a) {
...
   }
 }
+function hasUrlWithPipe(str) {
+  const regex = /^https?:\/\/.*\|.*/;
+  return regex.test(str);
+}
+
+export function decorateImageLinks(a) {
+  const images = a.querySelectorAll('img');
+  if (!images.length) return a;
```
narcis-radu, line 429: Is there an use-case for returning the element?

libs/utils/utils.js
```diff
@@ -419,6 +419,40 @@ export function decorateSVG(a) {
...
+      // //Mine for URL and alt text
+      const splitText = img.alt.split('|');
+      const textUrl = new URL(splitText.shift().trim());
+      const altText = splitText.join('|').trim();
+      if (altText) img.alt = altText;
+
+      const pic = img.closest('picture');
+      const picParent = pic.parentElement;
+      const aTag = createTag('a', { href: textUrl, class: 'desc-link' }, pic);
+      if (aTag.href.includes('#_blank')) {
```
narcis-radu, line 443:
```suggestion
      if (aTag.hash === '#_blank') {
```

libs/utils/utils.js
```diff
@@ -419,6 +419,40 @@ export function decorateSVG(a) {
...
+        aTag.setAttribute('target', '_blank');
+        aTag.href = aTag.href.replace('#_blank', '');
+      }
+      picParent.append(aTag);
```
narcis-radu, line 447:
```suggestion
      pic.parentElement?.append(aTag);
```
I don't think there's a need to cache the parent element, it's only used once

libs/utils/utils.js
```diff
@@ -419,6 +419,40 @@ export function decorateSVG(a) {
...
+    try {
+      // //Mine for URL and alt text
+      const splitText = img.alt.split('|');
+      const textUrl = new URL(splitText.shift().trim());
+      const altText = splitText.join('|').trim();
+      if (altText) img.alt = altText;
+
+      const pic = img.closest('picture');
+      const picParent = pic.parentElement;
+      const aTag = createTag('a', { href: textUrl, class: 'desc-link' }, pic);
```
rgclayton, line 442: Would it make more sense for the class to be `image-link`?  Not sure how "description" relates here.

libs/utils/utils.js
rgclayton: Maybe a utility function is needed for this since the same condition is [used here too](https://github.com/adobecom/milo/blob/27100d99661e0c8237e82be6bbf50849f97a2bb5/libs/utils/utils.js#L505-L508)

rgclayton: See my comment lower where you are calling `decorateImageLinks(area)`.

libs/utils/utils.js
ryanmparrish: `desc-` was because this link is derived from the image description. I like `image-link` better though. 👍

test/utils/mocks/body.html
ryanmparrish: 👍 I added these unit tests and moved them into their own test files.

libs/utils/utils.js
ryanmparrish: In testing it turns out this was necessary to declare the picParent as a valid DOM element.
I tried this suggestion and it resulted in a runtime error.

libs/utils/utils.js
```diff
@@ -419,6 +419,40 @@ export function decorateSVG(a) {
...
+
+export function decorateImageLinks(a) {
+  const images = a.querySelectorAll('img');
+  if (!images.length) return a;
+  [...images].forEach((img) => {
+    // Check if the image has an alt tag with a url & pipe - 'https... |'
+    if (!(hasUrlWithPipe(img.alt))) return a;
+    try {
+      // //Mine for URL and alt text
+      const splitText = img.alt.split('|');
```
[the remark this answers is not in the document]
ryanmparrish: @overmyheadandbody  - I refactored this logic and addressed the comments above w/ a simpler solution. Take a peek if you don't mind. Thanks

libs/utils/utils.js
ryanmparrish: @rgclayton  - I refactored this and called the 'decorateImageLinks()' in the `decorateLinks` func and thus inherited the target functionality. Checks out from my testing, I could see moving this to a single util if needed for multiple functions down the road. LMK if you have any thoughts.

libs/utils/utils.js
```diff
@@ -428,30 +429,16 @@ export function decorateImageLinks(a) {
...
-        aTag.href = aTag.href.replace('#_blank', '');
-      }
-      picParent.append(aTag);
-      return a;
-    } catch (e) {
-      /* c8 ignore next 3 */
-      console.log('Failed to decorate image links', e.message);
-      return a;
-    }
+    if (!img.alt || !(hasUrlWithPipe(img.alt))) return;
```
overmyheadandbody, line 432: Since the default return seems to be the initial `a`, should it also be returned here?

libs/utils/utils.js
```diff
@@ -428,30 +429,16 @@ export function decorateImageLinks(a) {
...
+    const [source, alt] = img.alt.split('|');
+    const textUrl = new URL(source);
+    if (alt) img.alt = alt;
```
overmyheadandbody, line 435: You might still want to apply a `.trim()`. A value like `a | b` will make the source `a ` and the alt ` b`. Just some extra spaces, but the output might be nicer without them. Also, if the value is `a | `, then an alt is still added as empty. So the alt value might do with both a trim and a length check.

libs/utils/utils.js
ryanmparrish: This looks right to me since this check is in the `images` loop and just returns if condition isn't met. The function param `a` does get returned at the end of the func. LMK if i'm off on this though. Thanks

overmyheadandbody: Yup, you're right, resolving conversation 👍

libs/utils/utils.js
```diff
@@ -420,6 +420,27 @@ export function decorateSVG(a) {
...
+}
+
+export function decorateImageLinks(a) {
+  const images = a.querySelectorAll('img');
+  if (!images.length) return a;
+  [...images].forEach((img) => {
+    if (!img.alt || !(hasUrlWithPipe(img.alt))) return;
+    const [source, alt] = img.alt.split('|');
+    const textUrl = new URL(source.trim());
+    if (alt && alt.trim().length > 0) img.alt = alt.trim();
```
overmyheadandbody, line 435: Nit: you could remove the `> 0` part, if the length is 0, it will be falsy anyway.

overmyheadandbody (approved):

libs/utils/utils.js
ryanmparrish: Good call ✔️

ryanmparrish (commented):
resolve feedback

rgclayton (approved):

rgclayton:
nit: Turn martech off for your page samples. That'll adjust your psi scores.

rgclayton:
@ryanmparrish Correct me if I'm wrong here, but I think the label should be `new-feature` instead of `high-impact`. I'm not seeing any breaking changes or anything that needs client set up details for engineers. Lemme know if I'm missing something though.


>high-impact = Any change that all Milo consumers should be aware of. E.g: breaking changes or a new feature that requires client setup. This will send an email to Grp-Milo-Dev. Any instructions of what consumers need to do should be included in the PR description.

honstar (approved):

honstar:
@ryanmparrish : can you please adjust the following 2 things that @rgclayton has already pointed out:
- add `?martech=off` to test urls
- change label `high-impact` to `new-feature`

rgclayton:
@ryanmparrish Is this ready to be merged? If not what else is needed?

libs/utils/utils.js
```diff
@@ -424,6 +424,27 @@ export function decorateSVG(a) {
   }
 }
 
+function hasUrlWithPipe(str) {
+  const regex = /^https?:\/\/.*\|.*/;
```
npeltier, line 428: to be "cleaner" here i'd consider using the "non pipe" range in 1+ instead of any character so something like `/^https?[^\|]+\|.*/`

libs/utils/utils.js
```diff
@@ -424,6 +424,27 @@ export function decorateSVG(a) {
...
+  return regex.test(str);
+}
+
+export function decorateImageLinks(a) {
```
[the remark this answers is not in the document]
ryanmparrish: yes, this is updated

libs/utils/utils.js
```diff
@@ -424,6 +424,27 @@ export function decorateSVG(a) {
...
+  const images = a.querySelectorAll('img');
```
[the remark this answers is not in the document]
ryanmparrish: Thanks, Good idea. This is resolved.

libs/utils/utils.js
ryanmparrish: the hasUrlWithPipe() func is no longer needed w/ the above `'img[alt*="|"]'` suggestion. 👍

libs/utils/utils.js
```diff
@@ -424,6 +424,27 @@ export function decorateSVG(a) {
...
+  if (!images.length) return a;
+  [...images].forEach((img) => {
+    if (!img.alt || !(hasUrlWithPipe(img.alt))) return;
+    const [source, alt] = img.alt.split('|');
+    const textUrl = new URL(source.trim());
```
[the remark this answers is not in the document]
ryanmparrish: Good call. I've added this. @chrischrischris, LMK if you think we need to check for the http(s) prefix?

libs/utils/utils.js
```diff
@@ -424,6 +424,27 @@ export function decorateSVG(a) {
...
+    if (alt && alt.trim().length) img.alt = alt.trim();
+    const pic = img.closest('picture');
+    const picParent = pic.parentElement;
+    const aTag = createTag('a', { href: textUrl, class: 'image-link' }, pic);
+    picParent.append(aTag);
```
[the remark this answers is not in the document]
ryanmparrish: I tested this out and it looks like the `const picParent` does needs to be declared otherwise the pic.parentElement cant append to itself and throws an error.

ryanmparrish:
Thanks for all the feedback. Everything discussed has been addressed. One more quick look from @chrischrischris and I think we can ship it. LMK thanks

honstar:
@ryanmparrish : can you please append the `martech=off` query param to your test URLs in the description? The LH check is currently failing.

chrischrischris (approved):

PR #784, opened 2023-05-26
MWPW-125482 Fixing line length and function defined before used eslint issue

aishwaryamathuria:
Fixing following eslint issue in PR https://github.com/adobecom/milo/pull/777
1. getMetadata used before defined
2. Line length

Resolves: [MWPW-125482](https://jira.corp.adobe.com/browse/MWPW-125482)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/mena_ar/drafts/mathuria/test?martech=off
- After: https://linelen--milo--aishwaryamathuria.hlx.page/mena_ar/drafts/mathuria/test?martech=off
Another After link with content-direction set to 'ltr' https://linelen--milo--aishwaryamathuria.hlx.page/mena_ar/drafts/mathuria/test1?martech=off


libs/utils/utils.js
```diff
@@ -152,7 +158,9 @@ export const [setConfig, getConfig] = (() => {
       document.documentElement.setAttribute('lang', config.locale.ietf);
       try {
         const contentDir = getMetadata('content-direction');
-        document.documentElement.setAttribute('dir', contentDir || config.locale.dir || (config.locale.ietf ? (new Intl.Locale(config.locale.ietf)?.textInfo?.direction) : null) || 'ltr');
+        const locietf = config.locale.ietf;
+        const ietfDir = locietf ? (new Intl.Locale(locietf)?.textInfo?.direction) : null;
+        document.documentElement.setAttribute('dir', contentDir || config.locale.dir || ietfDir || 'ltr');
```
auniverseaway, line 163: This seems backwards. Why would we do a calculation for something we don't know we will use?

```js
        const dir = getMetadata('dir')
          || config.locale.dir
          || new Intl.Locale(config.locale.ietf).textInfo.direction;
        documentElement.setAttribute('dir', dir || 'ltr');
```
Does this work?

auniverseaway (changes requested):
See above.

libs/utils/utils.js
chrischrischris, line 163: Not a big deal at all, but I think the var names could be better

```suggestion
        const { ietf } = config.locale;
        const intlLocaleDir = ietf && new Intl.Locale(ietf)?.textInfo?.direction;
        document.documentElement.setAttribute('dir', contentDir || config.locale.dir || intlLocaleDir || 'ltr');
```

Another option would be:
```
document.documentElement.setAttribute(
  'dir',
  contentDir
  || config.locale.dir
  || (config.locale.ietf && new Intl.Locale(config.locale.ietf)?.textInfo?.direction)
  || 'ltr',
);
```

aishwaryamathuria: @auniverseaway Directly using `new Intl.Locale(config.locale.ietf).textInfo.direction;` will throw an error if config.locale.ietf is not set for a locale.
Could you please confirm if I should change the lines to following? -
```
        let dir = getMetadata('content-direction') || config.locale.dir;
        if (!dir && config.locale.ietf) {
          dir = new Intl.Locale(config.locale.ietf)?.textInfo?.direction;
        }
        document.documentElement.setAttribute('dir', dir || 'ltr');
```

chrischrischris: oops, just saw @auniverseaway comment.  use that, but textInfo needs to use `?.`

aishwaryamathuria: Sure @chrischrischris I can change the var names or add the shared snippet.
@auniverseaway @chrischrischris Could you please confirm If I should include the snippet or change the variable names?

aishwaryamathuria: @chrischrischris Directly using new Intl.Locale(config.locale.ietf)?.textInfo?.direction will throw an error if ietf property is missing for a locale in the config.. 
One of the Unit test checks that if ietf and dir is not set it should fallback to 'ltr'
Could you please confirm if following can be used? -
```
        let dir = getMetadata('dir') || config.locale.dir;
        if (!dir && config.locale.ietf) {
          dir = dir || new Intl.Locale(config.locale.ietf)?.textInfo?.direction;
        }
        document.documentElement.setAttribute('dir', dir || 'ltr');
```

honstar:
@aishwaryamathuria : can you please adjust the title and add more details to the description? This PR includes an actual code change, the PR title and description should match that. Thanks!

aishwaryamathuria:
> @aishwaryamathuria : can you please adjust the title and add more details to the description? This PR includes an actual code change, the PR title and description should match that. Thanks!

Hi @honstar 
We are only fixing the 2 eslint issues added in the description in the PR - 
1. getMetadata used before defined.. The getMetadata only had to be shifted before the following as the function was after this below function
 `export const [setConfig, getConfig] = (() => {`
2. The line length was longer for 'dir' setting in document.dir so we split this into multiple lines to perform following check in below order (which is same as before)
- If metatag is set for 'content-direction' use it for 'dir'
- or if config.locale.dir is set use it for dir
- or if config.locale.ietf is defined and dir is not yet set then try to get the textinfo direction from it
- In the end check if 'dir' has a property then use that else fallback to 'ltr'

libs/utils/utils.js
3ch023: @chrischrischris i think it's ok if we always set the 'dir' explicitly because it might be risky to rely on the browser to select proper direction.
e.g. in the [use case of this ticket](https://main--milo--adobecom.hlx.page/mena_ar/drafts/mathuria/test?martech=off), the 'lang' attr of the HTML tag says 'ar'. but the actual page content is in English, not sure what will be prioritised by the browser, but we want to set it to 'LTR'.

3ch023:
> Fixing following eslint issue in PR #777
> 
> 1. getMetadata used before defined
> 2. Line length
> 
> Resolves: [MWPW-125482](https://jira.corp.adobe.com/browse/MWPW-125482)
> 
> **Test URLs:**
> 
> * Before: https://main--milo--adobecom.hlx.page/mena_ar/drafts/mathuria/test?martech=off
> * After: https://linelen--milo--aishwaryamathuria.hlx.page/mena_ar/drafts/mathuria/test?martech=off

I know this PR is only to fix the styling, but just noticed that the Before and After links in the original PR #777  both have dir='RTL'.
Isn't the expected behavior for the After link to have the LTR?

aishwaryamathuria:
> xpected behavior for the After link to have the LTR?

Hi @3ch023 I have added another after link in the description for the same page which has content-direction set to 'ltr'

libs/utils/utils.js
mokimo: I like this option. It's a lot more scannable and shows the hierarchy of importance what decides the direction.
```js
document.documentElement.setAttribute(
  'dir',
  contentDir
  || config.locale.dir
  || (config.locale.ietf && new Intl.Locale(config.locale.ietf)?.textInfo?.direction)
  || 'ltr',
);
```

libs/utils/utils.js
```diff
@@ -157,10 +157,9 @@ export const [setConfig, getConfig] = (() => {
...
       document.documentElement.setAttribute('lang', config.locale.ietf);
       try {
-        let dir = getMetadata('content-direction') || config.locale.dir;
-        if (!dir && config.locale.ietf) {
-          dir = dir || new Intl.Locale(config.locale.ietf)?.textInfo?.direction;
-        }
+        const dir = getMetadata('content-direction')
+          || config.locale.dir
+          || (config.locale.ietf && (new Intl.Locale(config.locale.ietf)?.textInfo?.direction));
         document.documentElement.setAttribute('dir', dir || 'ltr');
```
mokimo: Since we already chain everything to `const dir`, I'd also add 'ltr' onto that instead of having a second conditional assignment within setAttribute.

mokimo (approved):

salonijain3 (approved):

salonijain3:
@auniverseaway any other changes that are needed on this PR?

aishwaryamathuria:
@auniverseaway 
We have made the changes in the PR. Could you please review the same.
In case all looks good could you please merge as well?

PR #737, opened 2023-05-18
MWPW-131381 Move flags to sharepoint

chrischrischris:
This moves the loading of the country flag svgs from within the milo codebase:
https://milo.adobe.com/libs/features/georoutingv2/img/flag_ch.svg

to within sharepoint where authors can maintain the flag svgs:
https://milo.adobe.com/libs/img/georouting/flag-ch.svg

Note that _ have been changed to - since Franklin automatically converts underscores to dash.  I've also renamed all the flag files in sharepoint to use dashes instead of underscores in sharepoint/milo/libs/img/georouting/*.svg

Once we've verified the fix after merging we will be able to delete all of the flag files in /libs/features/georoutingv2/img

Resolves: [MWPW-131381](https://jira.corp.adobe.com/browse/MWPW-131381)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/cpeyer/experiment-test?akamaiLocale=ch&martech=off
- After: https://c3-flags--milo--adobecom.hlx.page/drafts/cpeyer/experiment-test?akamaiLocale=ch&martech=off

Testing Notes:
Inspect the country flag icon in the button.

For the "before" the img src will be to /libs/features/georoutingv2/img/flag_ch.svg

For the "after" the img src will be to /libs/img/georouting/flag-ch.svg

PR #786, opened 2023-05-26
MWPW-129912 New Feature Nav Marquee 

trixyfox:
* New Block Nav Marquee
* this is a combination of the marquee and the table of content (toc) block. 
* Authors wanted a singular authoring experience, instead of using 3 different blocks to create the marquee with toc. 
*  The figma design is here: https://www.figma.com/file/O4oM8f4Jgln5s6t05Oqfnw/Milo_Migration_Blocks-(EXT)?type=design&node-id=8-2&t=ZgEsbsxzntdDhTyq-0

**Acceptance Criteria** 

- Supports drop link/ anchor links on the page 
- Buttons are optional
- small subtext beneath the drop link boxes is optional
- this block will always be used as a Marquee (as far as page hierarchy/placement)
    - - If possible, use/create the variant (toc) as one of marquee variants. This way it will reduce confusion to authors of the block usage.
- panel with drop links has a background color that should be authored however, the default should be the grey with white boxes as shown in figma.
- 

Resolves: MWPW-129912 (https://jira.corp.adobe.com/browse/MWPW-129912)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/steadams/document
- After: https://mwpw-129912--milo--trixyfox.hlx.page/drafts/steadams/document


honstar:
@trixyfox , can you please update the PR title to include what this PR is about? Usually, the title of the corresponding JIRA issue suffices, thanks!

libs/blocks/nav-marquee/nav-marquee.css
```diff
@@ -0,0 +1,275 @@
...
+  background-color: #FFF;
+  border: 1px solid #4B4B4B;
+  border-radius: 4px;
+  padding: 24px;
+  cursor: pointer;
+  line-height: 1.5;
+}
+
+.nav-marquee .toc-item:hover {
+  box-shadow: 2px 4px 5px #0000001A;
```
narcis-radu, line 50: I don't think this is a proper hex code`#0000001A`

libs/blocks/nav-marquee/nav-marquee.css
```diff
@@ -0,0 +1,275 @@
...
+.nav-marquee .toc-item:focus-within {
+  border: 1px solid #1473E6;
+  background-color: #F5F9FD;
+}
+
+.nav-marquee .marquee p {
+  margin: 0 0 var(--spacing-s);
+}
+
+.nav-marquee .marquee .foreground .icon-area img{
```
narcis-radu, line 63: space before `{`

libs/blocks/nav-marquee/nav-marquee.css
```diff
@@ -0,0 +1,275 @@
...
+  .nav-marquee .marquee .foreground .text h3 {
+    font-size: var(--type-heading-m-size);
+    text-transform: none;
+  }
+
+  .nav-marquee .marquee .foreground .media {
+    margin: 0;
+  }
+}
+@media screen and (max-width: 600px) {
```
narcis-radu, line 140: I think the preferred approach is to build your way up:
- styles
- min-width: 600px
- min-width: 1200px
- min-width: 1440px

This particular media query can be removed, all styles should be added before the other media queries

libs/blocks/nav-marquee/nav-marquee.css
```diff
@@ -0,0 +1,275 @@
...
+  .nav-marquee .toc-item {
+    padding: 16px;
+  }
+
+  .nav-marquee .section-description {
+    font-size: 12px;
+    margin-top: 16px;
+  }
+
+  .nav-marquee .toc-footer{
```
narcis-radu, line 154: space before `{`

libs/blocks/nav-marquee/nav-marquee.css
```diff
@@ -0,0 +1,275 @@
...
+    min-width: 200px;
+    display: flex;
+    position: relative;
+    justify-content: space-between;
+    background-color: #FFF;
+    border: 1px solid #4B4B4B;
+    border-radius: 4px;
+    padding: 24px;
+    cursor: pointer;
+    line-height: 1.5;
```
narcis-radu, line 225: you have a similar definition for `.nav-marquee .toc-item`. The previous one would get inherited on min 1200px
https://github.com/adobecom/milo/pull/786/files#diff-3052973bc600cc270df3e74fc90fefef74cfa0de6424aea73b1b415ea91e58a7R46

Also, you're setting `line-height: 1.5` for the whole section, but then you specify different line-height for the `section-title` which is a child of `toc-item`. Maybe you can add specific line height directly to elements.

libs/blocks/nav-marquee/nav-marquee.css
```diff
@@ -0,0 +1,275 @@
...
+  cursor: pointer;
+  line-height: 1.5;
+}
+
+.nav-marquee .toc-item:hover {
+  box-shadow: 2px 4px 5px #0000001A;
+  border: 1px solid #1473E6;
+}
+
+.nav-marquee .toc-item:focus-within {
```
narcis-radu, line 54: I can't see the styles being applied on desktop. The desktop styles override these declarations.
https://github.com/adobecom/milo/pull/786/files#diff-3052973bc600cc270df3e74fc90fefef74cfa0de6424aea73b1b415ea91e58a7R220-R221

Also, I don't see a reason for using focus-within, hover is good enough for this use-case.

libs/blocks/nav-marquee/nav-marquee.css
```diff
@@ -0,0 +1,275 @@
...
+      flex-direction: row;
+  }
+
+  .nav-marquee .content-table {
+    background-color: #FFF;
+    border: none;
+  }
+
+  .nav-marquee .section-title {
+    line-height: 22.5px;
```
narcis-radu, line 175: Mixed usage for line-height:
1. unitless value - https://github.com/adobecom/milo/pull/786/files#diff-3052973bc600cc270df3e74fc90fefef74cfa0de6424aea73b1b415ea91e58a7R225
2. px value - https://github.com/adobecom/milo/pull/786/files#diff-3052973bc600cc270df3e74fc90fefef74cfa0de6424aea73b1b415ea91e58a7R175

test/blocks/nav-marquee/mocks/body.html
```diff
@@ -0,0 +1,108 @@
...
+        </div>
+      </div>
+    </div>
+  </div>
+  <div class="content-table">
+    <div class="toc-container">
+      <p class="toc-title">This is the title of the section</p>
+      <nav aria-label="Table of contents">
+        <ul class="toc-list">
+          <li class="toc-item">
```
narcis-radu, line 35: I am not able to TAB through the toc items, this breaks keyboard navigation / accessibility patterns.

libs/blocks/nav-marquee/nav-marquee.js
```diff
@@ -0,0 +1,129 @@
...
+ */
+import { createTag } from '../../utils/utils.js';
+import { decorateButtons, getBlockSize } from '../../utils/decorate.js';
+
+const DOWN_ARROW_ICON = '<svg aria-hidden="true" xmlns="http://www.w3.org/2000/svg" width="10.875" height="13.323" viewBox="0 0 10.875 13.323"><title>down arrow</title><g fill="none" stroke="#1473e6" stroke-linecap="round" stroke-width="2"><path d="M5.478 1v10.909"/><g><path d="m9.461 7.885-4.023 4.023m0 0L1.414 7.885"/></g></g></svg>';
+
+function findAnchorTarget(text) {
+  let linkText = text.toLowerCase();
+  if (linkText.charAt(0) === '_') {
+    linkText = linkText.charAt(0) === '_' ? linkText.substring(1) : linkText;
```
narcis-radu, line 24: you know for sure that the first character is `_`, you can just create the new string

libs/blocks/nav-marquee/nav-marquee.js
```diff
@@ -0,0 +1,129 @@
...
+    linkText = linkText.replaceAll('_', '-');
+    linkText = linkText.replaceAll(/[ /|&;$%@"<>()+,.]/g, '');
+    linkText = `#${linkText}`;
+    //return document.querySelector(`[id^="${linkText}"]`);
+  }
+  return linkText;
+}
+
+function getItem(title, description, target) {
+  if (!title) {
```
narcis-radu, line 34:
```suggestion
  if (!title) return null;
```

libs/blocks/nav-marquee/nav-marquee.js
```diff
@@ -0,0 +1,129 @@
...
+    return null;
+  }
+  const item = createTag('li', { class: 'toc-item' });
+  const linkText = createTag('div', { class: 'toc-link-text' });
+  const pageTop = document.querySelector('header')?.offsetHeight ?? 0;
+
+  if (title) {
```
narcis-radu, line 41: there is already a check in place - https://github.com/adobecom/milo/pull/786/files#diff-48db90833512b765bffd11f28f093b0f2e7c92a3ec8a0294d99f99c5faaf2b4bR34

libs/blocks/nav-marquee/nav-marquee.js
```diff
@@ -0,0 +1,129 @@
...
+    const link = createTag('a', { class: 'section-title', href: target, target: '_self' }, title);
+    linkText.append(link);
+    item.addEventListener('click', () => {
+      const isTextSelected = window.getSelection().toString();
```
narcis-radu, line 45: I noticed some issues with this approach:
1. select text inside the item
2. once you're done selecting, it will trigger the click and `getSelection` will return selected text
3. with text already selected, click on the item

desired behaviour: the click is executed
actual behaviour: the click is not executed because `getSelection` returns text

I thinks the approach needs to be refactored and the item should be wrapped inside an anchor. This way:
* you won't need to check for selection
* you'll get TAB support out of the box
* errors like `target?.getBoundingClientRect is not a function` are avoided
* you don't need to force a click on the element inside of the item - https://github.com/adobecom/milo/pull/786/files#diff-48db90833512b765bffd11f28f093b0f2e7c92a3ec8a0294d99f99c5faaf2b4bR48

libs/blocks/nav-marquee/nav-marquee.js
```diff
@@ -0,0 +1,129 @@
...
+
+      if (!isTextSelected) {
+        link.click();
+      }
+    });
+    link.addEventListener('click', (e) => {
+      const targetPosition = target?.getBoundingClientRect().top ?? 0;
```
narcis-radu, line 52:
```suggestion
      const targetPosition = target?.getBoundingClientRect()?.top ?? 0;
```

libs/blocks/nav-marquee/nav-marquee.js
```diff
@@ -0,0 +1,129 @@
...
+
+function decorateText(el, size) {
+  const headings = el.querySelectorAll('h1, h2, h3, h4, h5, h6');
+  const heading = headings[headings.length - 1];
+  const decorate = (headingEl, headingSize, bodySize, detailSize) => {
+    headingEl.classList.add(`heading-${headingSize}`);
+    headingEl.nextElementSibling?.classList.add(`body-${bodySize}`);
+    const sib = headingEl.previousElementSibling;
+    if (sib) {
+      sib.querySelector('img, .icon') ? sib.classList.add('icon-area') : sib.classList.add(`detail-${detailSize}`);
```
narcis-radu, line 76:
```suggestion
      sib.classList.add(sib.querySelector('img, .icon') ? 'icon-area' : `detail-${detailSize}`);
```

libs/blocks/nav-marquee/nav-marquee.js
```diff
@@ -0,0 +1,129 @@
...
+    if (section.firstElementChild.textContent === 'footer') {
+      footer.append(section.lastElementChild);
+      section.remove();
+      return;
+    }
+
+    const sectionTags = Array.from(section.querySelectorAll('p'));
+    const sectionTitle = section.querySelector('strong');
+    const link = section.querySelector('a');
+    const subtitle = sectionTags.find((element) => element.childElementCount === 0);
```
narcis-radu, line 115:
```suggestion
    const subtitle = [...section.querySelectorAll('p')].find((element) => element.childElementCount === 0);
```
no need for the `sectionTags` const

narcis-radu (changes requested):

libs/blocks/nav-marquee/nav-marquee.css
```diff
@@ -0,0 +1,275 @@
+.nav-marquee{
+  flex: 1 1 auto;
+  background-color: black;
```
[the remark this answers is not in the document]
trixyfox: fixed in more recent version.

libs/blocks/nav-marquee/nav-marquee.css
```diff
@@ -0,0 +1,275 @@
...
+  border: 1px solid #E8E8E8;
+  background-color: #FAFAFA;
+}
+
+.nav-marquee .content-table .toc-container {
+  flex: 1 1 auto; 
+  padding: 24px;
+}
+
+.nav-marquee .content-table .toc-title {
```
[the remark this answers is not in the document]
trixyfox: There was conflict with the marquee but I've since refactored toc to offer and I can reduce specificity.

libs/blocks/nav-marquee/nav-marquee.js
trixyfox: It's not always an anchor tag, there is a possibility of a normal link.

libs/blocks/nav-marquee/nav-marquee.js
trixyfox: 2 line reduction. great.

libs/blocks/nav-marquee/nav-marquee.css
trixyfox: This is actually valid as the third value pair is transparency. 
https://www.digitalocean.com/community/tutorials/css-hex-code-colors-alpha-values

test/blocks/nav-marquee/mocks/body.html
trixyfox: Added tabindex

libs/blocks/nav-marquee/nav-marquee.css
```diff
@@ -0,0 +1,278 @@
+.nav-marquee {
+  flex: 1 1 auto;
+}
+
+.nav-marquee.light {
```
ryanmparrish, line 5: What is the default color-scheme on this? Dark or light. Which ever one is default shouldn't need a helper `.light` class.

libs/blocks/nav-marquee/nav-marquee.css
```diff
@@ -0,0 +1,278 @@
...
+  color: var(--text-color);
+  background-color: #FFF;
+}
+
+.nav-marquee.dark {
+ background-color: #000;
```
ryanmparrish, line 11: To fully support `.dark` we need to consider the section so
```
.dark .nav-marquee, 
.nav-marquee.dark
```
for when it's nested in a .section.dark
Also shouldn't the font color be included here?

libs/blocks/nav-marquee/nav-marquee.css
```diff
@@ -0,0 +1,278 @@
...
+}
+
+.nav-marquee .content-table {
+  border: 1px solid #E8E8E8;
+  background-color: #FAFAFA;
```
ryanmparrish, line 16: Please use vars when specifying specific background or border colors. `--color-gray-100` && `--color-gray-200`

libs/blocks/nav-marquee/nav-marquee.css
```diff
@@ -0,0 +1,278 @@
...
+}
+
+.nav-marquee .marquee p {
+  margin: 0 0 var(--spacing-s);
+}
+
+.nav-marquee .marquee .foreground .icon-area img {
+  height: 20px;
+  float: left;
+  padding-right: 16px;
```
ryanmparrish, line 69: This looks like a good place for a spacing token `--spacing-xs` and float feels kind of outdated. Consider using flex-box w/ gap

libs/blocks/nav-marquee/nav-marquee.css
```diff
@@ -0,0 +1,278 @@
...
+
+  .nav-marquee .marquee .foreground {
+    margin-right: 0;
+    align-content: space-between;
+    flex-flow: row wrap;
+    gap: 0;
+  }
+
+  .nav-marquee .offer-container {
+    padding: 32px;
```
ryanmparrish, line 197: use spacing tokens `--spacing-m`

ryanmparrish (changes requested):
Looking good. I added a few optimization notes. We should support the color-scheme options our sections and blocks. LMK if you have any questions around this. Thanks

ryanmparrish (changes requested):
Looking good. I added a few optimization notes. We should support the color-scheme options our sections and blocks. LMK if you have any questions around this. Thanks

libs/blocks/nav-marquee/nav-marquee.css
trixyfox: Default is light, so I will remove helper

libs/blocks/nav-marquee/nav-marquee.css
```diff
@@ -0,0 +1,278 @@
...
+    box-sizing: border-box;
+    margin-left: 0;
+    padding-right: 0;
+    background-color: #FFF;
```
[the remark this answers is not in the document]
trixyfox: light and dark don't effect the offer-container.

libs/blocks/nav-marquee/nav-marquee.css
rgclayton: Mobile first approach please.

libs/blocks/nav-marquee/nav-marquee.css
rgclayton: We typically prefer the unit less value. It tends to be less fussy and MDN recommends it.

libs/blocks/nav-marquee/nav-marquee.js
```diff
@@ -0,0 +1,133 @@
...
+ * governing permissions and limitations under the License.
+ */
+
+/*
+ * Navigation Marquee
+ */
+import { createTag } from '../../utils/utils.js';
+import { decorateButtons, getBlockSize } from '../../utils/decorate.js';
+
+const DOWN_ARROW_ICON = '<svg aria-hidden="true" xmlns="http://www.w3.org/2000/svg" width="10.875" height="13.323" viewBox="0 0 10.875 13.323"><title>down arrow</title><g fill="none" stroke="#1473e6" stroke-linecap="round" stroke-width="2"><path d="M5.478 1v10.909"/><g><path d="m9.461 7.885-4.023 4.023m0 0L1.414 7.885"/></g></g></svg>';
```
rgclayton, line 19: I think we'd like to move away from this approach of adding images.
Either:

- An `/img/` folder in the block with a reference to it. 
- Or, the image is [added here](https://adobe.sharepoint.com/sites/adobecom/Shared%20Documents/Forms/AllItems.aspx?id=%2Fsites%2Fadobecom%2FShared%20Documents%2Fmilo%2Flibs%2Fimg&viewid=d776cf70%2D9b7e%2D4ab7%2Db9da%2D9e0f8e03a7d2) and then referenced. This makes any changes related to the image a non engineering effort. Here's a [nice reference](https://github.com/adobecom/milo/pull/737/files#diff-5353f6a7bd38256954f04062d9effabd975b061e55049968316ce24db587e768R187) for this approach. 

@chrischrischris What do you think?

libs/blocks/nav-marquee/nav-marquee.js
```diff
@@ -0,0 +1,122 @@
...
+
+function findAnchorTarget(text) {
+  let linkText = text.toLowerCase();
+  if (linkText.charAt(0) === '_') {
+    linkText = linkText.charAt(0) === '_' ? linkText.substring(1) : linkText;
+    linkText = linkText.replaceAll('_', '-');
+    linkText = linkText.replaceAll(/[ /|&;$%@"<>()+,.]/g, '');
+    linkText = `#${linkText}`;
+  }
+  return linkText;
```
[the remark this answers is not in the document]
rgclayton: I think we should bake in`scroll-margin-top` to the linking or the `<h>` tags. Currently when I click on any links to jump to that section the title for that section gets cut off by the global nav. 

[This will help sort that out](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-margin-top).

libs/blocks/nav-marquee/nav-marquee.js
```diff
@@ -0,0 +1,133 @@
...
+ * Navigation Marquee
+ */
+import { createTag } from '../../utils/utils.js';
+import { decorateButtons, getBlockSize } from '../../utils/decorate.js';
+
+const DOWN_ARROW_ICON = '<svg aria-hidden="true" xmlns="http://www.w3.org/2000/svg" width="10.875" height="13.323" viewBox="0 0 10.875 13.323"><title>down arrow</title><g fill="none" stroke="#1473e6" stroke-linecap="round" stroke-width="2"><path d="M5.478 1v10.909"/><g><path d="m9.461 7.885-4.023 4.023m0 0L1.414 7.885"/></g></g></svg>';
+
+function findAnchorTarget(text) {
+  let linkText = text.toLowerCase();
+  if (linkText.charAt(0) === '_') {
```
rgclayton, line 23: Using the underscore character seems less obvious to me. Is there are reason the flag isn't `#`? 
The hash is much more universal when creating jump links on web pages. 

@Blainegunn @auniverseaway

trixyfox: from my understanding, this was how the authors were accustomed to creating anchor links.

narcis-radu (changes requested):
There are still some tweaks needed:
1. be consistent with values for `line-height` - @rgclayton's suggestion was to use unit less values - https://github.com/adobecom/milo/pull/786/files#r1212627379 - see https://github.com/adobecom/milo/pull/786/files#diff-3052973bc600cc270df3e74fc90fefef74cfa0de6424aea73b1b415ea91e58a7R193
2. mobile first approach - no need for `@media screen and (max-width: 600px)` - see https://github.com/adobecom/milo/pull/786/files#diff-3052973bc600cc270df3e74fc90fefef74cfa0de6424aea73b1b415ea91e58a7R150
3. accessibility - I was able to TAB through items, but can't see the active one, it should be highlighted with a focus ring.

libs/blocks/nav-marquee/nav-marquee.js
```diff
@@ -0,0 +1,134 @@
...
+  const detailSize = size === 'large' ? 'l' : 'm';
+  decorate(heading, headingSize, bodySize, detailSize);
+}
+
+export default function init(el) {
+  const size = getBlockSize(el);
+  const children = Array.from(el.querySelectorAll(':scope > div'));
+  // const tone = (el.classList.contains('dark')) ? 'dark' : 'light';
+  // el.classList.add(tone);
+  const marquee = createTag('div', { class: 'marquee' }, children.shift());
```
ryanmparrish, line 93: I was wondering if your intentionally adopting and using .marquee styles? 
Otherwise we should probably name this different so there isn't a naming collision.

libs/blocks/nav-marquee/nav-marquee.js
Blainegunn: @rgclayton, @trixyfox is following the previous convention. That said if you are in favor of `#` that's easy enough to change! 

@trixyfox please give `#` a try instead of the the `_`

libs/blocks/nav-marquee/nav-marquee.css
```diff
@@ -0,0 +1,271 @@
...
+
+
+.nav-marquee .marquee .foreground {
+  width: var(--grid-container-width);
+  margin: 0 auto;
+  padding-top: var(--spacing-xxl);
+  padding-bottom: 0;
+}
+
+.nav-marquee .marquee .media {
```
ryanmparrish, line 126: `.media` doesn't appear to be used. These can be removed right?

libs/blocks/nav-marquee/nav-marquee.js
ryanmparrish: There appears to be an un necessary amount of createTag() usage going on. We should try to look first at the natural dom structure and see if we can work w/ that where possible. I have this branch pulled down locally and am testing a few things if you wanna sync up in a few @trixyfox

PR #787, opened 2023-05-26
Add divider-inherit css to support different color divider

JackySun9:
* Adding a divider style row just adds a divider class.
* Extent the syntax to add an inherit option to not specify the color.
* That way, the divider would just follow the text color and give us what we would like. 

Resolves: [MWPW-131449](https://jira.corp.adobe.com/browse/MWPW-131449)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/fragments/xiasun/divider?martech=off
- After: https://main--milo--JackySun9.hlx.page/fragments/xiasun/divider?martech=off

I need to add an usage example to the section metadata documentation after the PR is merged.

### DIFF
--- a/libs/blocks/section-metadata/section-metadata.css
+++ b/libs/blocks/section-metadata/section-metadata.css
@@ -88,6 +88,10 @@
   border-bottom: 1px solid #d8d8d8;
 }
 
+.section.divider-inherit {   
+  border-bottom: 1px solid; 
+}
+
 .section.center[class*='-up'] {
   justify-items: center;
 }


echampio-at-adobe (approved):
Thanks, Jacky. To me, the change looks like what the story had in mind. Since it just adds support for another class, I don't see that an unit tests need to be updated, just the documentation for the block after the PR is eventually merged.

mokimo (approved):

fullcolorcoder (approved):

PR #790, opened 2023-05-29
MWPW-129909-inline variation of icon block

SonjaPopovic:
Resolves: [MWPW-129909](https://jira.corp.adobe.com/browse/MWPW-129909)

Test URLs:

Before: https://main--milo--adobecom.hlx.page/drafts/sonja/tab-icon-inline-3-up?martech=off
After: https://iconblockinline--milo--sonjapopovic.hlx.page/drafts/sonja/tab-icon-inline-3-up?martech=off


puruadobe (approved):

ryanmparrish:
This is looking good. There's a few things we should probably address. (sorry the first one doesn't look related to your work)

In the after view

1st tab, the .action-area should have a 24px (spacing m) top padding. 
Screen Shot 2023-05-30 at 2 05 37 PM

2nd tab, the block .foreground is getting a 800px width set on desktop viewport.
Screen Shot 2023-05-30 at 2 06 31 PM

ryanmparrish:
@elan-tbx  - Do you have any thoughts on the things I noted above? Feels like we should have a default .icon-block .action-area w/ top margin and flex gap? && the 800 width being set on the item foreground might need a `*-up` selector for override otherwise maybe we can refactor that need all together and rely on .con-block max-width-8 as an opt in way to tack that constraint? Any thoughts?

SonjaPopovic:
@ryanmparrish  Thanks for taking look at this.
For 1st item:
On ticket we have this design:
iconBlock

Please notice that here we  have body-s size required (which is why padding-24 px is not inserted on element with body-s (comparing to body-m on 'before' preview). On the ticket there is no design for case with buttons, so i am not sure which padding should be used here.  So please add design or requirement on the ticket if you have information which padding is required for this.

For 2nd item, since this issue exist even on preview 'before', can we have this fixed on separate ticket? I would be happy to contribute. Thanks!

elaineskpt (approved):
LGTM

we are checking this for a design requirement for our uar-marquee
https://xd.adobe.com/view/ff56e29c-fa5f-44f5-99f7-80ed62e0151e-aa05/screen/3fc66a56-d253-4735-ae4b-f713abd8268d

elaineskpt:
added PR labels, checking was failling

libs/blocks/icon-block/icon-block.js
```diff
@@ -15,13 +15,15 @@
 */
 
 import { decorateBlockText, getBlockSize } from '../../utils/decorate.js';
+import { createTag } from '../../utils/utils.js';
 
-const variants = ['fullwidth', 'vertical', 'bio'];
+const variants = ['fullwidth', 'vertical', 'bio', 'icon-inline'];
```
honstar, line 20: The variant label `icon-inline` seems redundant, why not simply use `inline`?

libs/blocks/icon-block/icon-block.js
```diff
@@ -15,13 +15,15 @@
...
 const iconBlocks = {
   small: {
     [variants[0]]: ['m', 'm'],
     [variants[1]]: ['s', 'm'],
     [variants[2]]: ['s', 's'],
+    [variants[3]]: ['s', 's'],
```
honstar, line 26: Is this really needed? If yes, an additional entry would be missing for both `medium` and `large`.

honstar:
The code coverage for the patch diff indicates missing unit-tests for the added feature. @SonjaPopovic , can you please add a unit-test for the new `inline` feature? Thanks

libs/blocks/icon-block/icon-block.js
SonjaPopovic: yes, it is really needed, since we can not use existing variants[2] for inline variation.. Added additional entry for medium and large.

honstar (approved):

elaineskpt (approved):

ryanmparrish (approved):

honstar:
Looks like the CLA issue has been resolved in the meantime, thx @SonjaPopovic 👍

PR #792, opened 2023-05-29
Set proper styles for top navigation menus

narcis-radu:
Set correct text color and background color when interacting with top menu items.

Resolves: [MWPW-130810](https://jira.corp.adobe.com/browse/MWPW-130810)

**Test URLs:**
- After: https://mwpw-130810--milo--adobecom.hlx.page/drafts/narcis/default?martech=off

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.css
+++ b/libs/blocks/global-navigation/global-navigation.css
@@ -172,6 +172,11 @@ header.global-navigation {
   background-color: var(--feds-background-link--hover--light);
 }
 
+.feds-navItem > .feds-navLink--hoverCaret:hover,
+.feds-navItem > .feds-navLink--hoverCaret:focus {
+  color: var(--feds-color-link--light);
+}
+
 .feds-cta-wrapper {
   display: flex;
 }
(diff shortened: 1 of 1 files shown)


mokimo (approved):

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -172,6 +172,11 @@ header.global-navigation {
   background-color: var(--feds-background-link--hover--light);
 }
 
+.feds-navItem > .feds-navLink--hoverCaret:hover,
+.feds-navItem > .feds-navLink--hoverCaret:focus {
+  color: var(--feds-color-link--light);
+}
+
```
overmyheadandbody, line 179: I think this is more complex than it needs to be. `.feds-navLink--hoverCaret` represent top menu items which have a caret, I don't see a need for the `.feds-navItem > ` part.

I also think this belongs to `base.css`, maybe somewhere around [here](https://github.com/adobecom/milo/blob/main/libs/blocks/global-navigation/base.css#L55).

Of course, same comments apply for the desktop styles.

overmyheadandbody (changes requested):

overmyheadandbody (approved):

honstar (approved):

PR #800, opened 2023-05-30
Adding the current locale code to the local storage key

sabyamon:
Adding the current locale code to the local storage key.

Resolves: https://jira.corp.adobe.com/browse/MWPW-130650

**Test URLs:**
- Before: https://uar-integration--milo--adobecom.hlx.page/drafts/sabya/quiz-2/?martech=off
- After: https://quiz-locale-key--milo--sabyamon.hlx.page/drafts/sabya/quiz-2/?martech=off


fullcolorcoder (approved):
LGTM - locale coming through. 
image

JackySun9:
@sabyamon I do see the local set with the key, but when I hit 'get your results', it will redirect to invalid path, I think because when it redirects, it is still using uar-integration endpoints, we can test more after this merging

JackySun9:
The performance is about mobile and SEO, we need to log some tickets for them

JackySun9 (approved):

libs/blocks/quiz/utils.js
```diff
@@ -1,10 +1,12 @@
 import { getMetadata } from '../section-metadata/section-metadata.js';
+import { getConfig } from '../../utils/utils.js';
 
 const QUESTIONS_EP_NAME = 'questions.json';
 const STRINGS_EP_NAME = 'strings.json';
 const RESULTS_EP_NAME = 'results.json';
 
 let getConfigPath; let getQuizKey; let getAnalyticsType; let getAnalyticsQuiz; let metaData;
```
honstar, line 8: Just a minor code smell: please do not use `get` prefixed variable names, it's confusing to see getter function naming for regular variables.

sukamat (approved):

libs/blocks/quiz/utils.js
honstar: > Just a minor code smell: please do not use `get` prefixed variable names, it's confusing to see getter function naming for regular variables.

@sabyamon , are you planning to address the requested change? I think it would make the code much more readable.

sabyamon: Hi @honstar - Thanks much for the detailed review. It absolutely makes sense and there are more occurrences of similar naming issues in the question flow. We discussed this in our today's planning session and created a follow-up ticket to address this. Hope that works.

https://jira.corp.adobe.com/browse/MWPW-132229

PR #810, opened 2023-06-01
MWPW-123036 [Milo] Custom filter panel options

cmiqueo:
Adds custom filters panel and functionality to the CaaS card collection configurator.

In addition to the automatically generated filter panels, now authors will be able to manually specify custom names for the filter groups and each individual filter.

Resolves: [MWPW-123036](https://jira.corp.adobe.com/browse/MWPW-123036)

Output:
image

**Test URLs:**
- After:  https://mwpw-123036--milo--adobecom.hlx.page/tools/caas?martech=off


libs/blocks/caas-config/caas-config.css
```diff
@@ -36,3 +36,72 @@ dd.content .sort-options > .field {
   top: 12px;
 }
 
+.config-panel dd.content.is-expanded {
+  max-height: 6000px;
```
mokimo, line 40: 6000px max-height looks a little weird, is it intended?

libs/blocks/caas-config/caas-config.css
```diff
@@ -36,3 +36,72 @@ dd.content .sort-options > .field {
...
+}
+
+.multifield .multifield {
+  padding: 0 0 15px;
+  margin: 0 -15px 9px 0;
```
mokimo, line 45: Negative margins are usually bad practice, if you can avoid them anyhow.

libs/blocks/caas-config/caas-config.css
```diff
@@ -36,3 +36,72 @@ dd.content .sort-options > .field {
...
+}
+
+.multifield .multifield h3 {
+  font-size: 16px;
+  font-weight: 300;
+  margin: 0;
+}
+
+.multifield .multifield > .multifield-set > .multifield-fields {
+  width: 190px;
```
mokimo, line 76: Fixed width is usually bad practice, as things are static rather than responsive.

libs/blocks/caas-config/caas-config.js
```diff
@@ -639,7 +679,7 @@ const getInitialState = () => {
   // /* c8 ignore next 2 */
   if (!state) {
     const lsState = localStorage.getItem(LS_KEY);
-    if (lsState) {
+    if (lsState?.includes('filtersCustom')) {
```
mokimo, line 682: Do we really only want to parse the state if it has customfilters?

libs/blocks/caas/utils.js
```diff
@@ -208,15 +208,43 @@ const getFilterObj = ({ excludeTags, filterTag, icon, openedOnLoad }, tags, stat
...
+  const useCustomFilters = state.filterBuildPanel === 'custom';
+
+  let filters = [];
+  if (!useCustomFilters) {
+    filters = state.filters
+      .map((filter) => getFilterObj(filter, tags, state))
+      .filter((filter) => filter !== null);
+  } else {
+    filters = state.filtersCustom.length > 0
+      ? state.filtersCustom.map((filter) => getCustomFilterObj(filter, state))
```
mokimo, line 244: You only destructure the first argument on `getCustomFilterObj` do you need to pass state?

libs/blocks/caas-config/caas-config.css
```diff
@@ -36,3 +36,72 @@ dd.content .sort-options > .field {
...
+
+/* sticky panel */
+.config-panel {
+  position: sticky;
+  top: 60px;
+  align-self: start;
+  overflow: scroll;
+  max-height: 95vh;
+  background: #ddd;
+  -ms-overflow-style: none;  /* Internet Explorer 10+ */
```
narcis-radu, line 101: not sure that's needed

libs/blocks/caas-config/caas-config.css
cmiqueo: Since we don't know how many groups of filters and filters will be added I hope 6000 pixels is enough. the original 1800pixes was not enough.

libs/blocks/caas-config/caas-config.css
cmiqueo: We try to avoid negative margins when possible,

libs/blocks/caas-config/caas-config.css
cmiqueo: In this case the container is inside another container with a fix width and constrain dimensions, so I figure it will work since the parent element is not fluid/responsive.

libs/blocks/caas-config/caas-config.js
cmiqueo: Yes. This is needed only to support legacy users, with and older localStorage formatted state.

libs/blocks/caas/utils.js
cmiqueo: Good catch! :)

libs/blocks/caas-config/caas-config.css
```diff
@@ -36,3 +36,72 @@ dd.content .sort-options > .field {
...
 
+.config-panel dd.content.is-expanded {
+  max-height: 6000px;
+}
+
+.multifield .multifield {
+  padding: 0 0 15px;
+  margin: 0 -15px 9px 0;
+  border-bottom: solid 1px #999;
+  border-bottom: none;
```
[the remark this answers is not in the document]
cmiqueo: Good catch! : )

libs/blocks/caas-config/caas-config.css
```diff
@@ -36,3 +36,72 @@ dd.content .sort-options > .field {
...
+.multifield .multifield > .multifield-header > .multifield-add {
+  font-size: 16px;
+}
+
+.config-panel button {
+  cursor: pointer;
+}
+
+.accordion-item:hover dt.title span,
+.accordion-item:hover  dt.title button {
```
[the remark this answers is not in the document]
cmiqueo: Good catch! : )

libs/blocks/caas-config/caas-config.js
```diff
@@ -21,6 +21,7 @@ import TagSelect from '../../ui/controls/TagSelector.js';
 import MultiField from '../../ui/controls/MultiField.js';
 import '../../utils/lana.js';
 
+const DEBUG = window.location.search.includes('debug=true');
```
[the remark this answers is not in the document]
cmiqueo: Nuked!

libs/blocks/caas-config/caas-config.js
```diff
@@ -648,6 +688,7 @@ const getInitialState = () => {
   }
 
   if (!state) state = {};
+  if (DEBUG) window.state = state;
```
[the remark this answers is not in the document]
cmiqueo: Double nuked!

libs/blocks/caas-config/caas-config.css
narcis-radu: I checked in Edge are there's no difference.
Screenshot 2023-06-02 at 18 58 37

cmiqueo: removed!

libs/blocks/caas-config/caas-config.js
chrischrischris: Let's add a comment here explaining why we're doing this

chrischrischris (approved):

libs/blocks/caas-config/caas-config.js
cmiqueo: @chrischrischris 
The CAAS configurator saves the current state with the user's selected options in `localStorage `for persistency. 
Since the current implementation does not use the `filtersCustom `option, that entry is missing from the saved state and the preact code fails without this check. Originally we just checked if the localStorage state (lsstate) existed. not we check if it exists AND if it contains the new property filtersCustom before using it.

PR #813, opened 2023-06-01
Configure Jarvis chat for mobile / desktop

narcis-radu:
* Allow authors to specify where Jarvis chat can be rendered - mobile or desktop

Resolves: [MWPW-132006](https://jira.corp.adobe.com/browse/MWPW-132006)

**Test URLs:**
- After: https://mwpw-132006--milo--narcis-radu.hlx.page/drafts/narcis/default?martech=off


libs/scripts/delayed.js
```diff
@@ -12,9 +12,14 @@
 
 export const loadJarvisChat = async (getConfig, getMetadata, loadScript, loadStyle) => {
   const config = getConfig();
-  const jarvis = getMetadata('jarvis-chat');
+  const jarvis = getMetadata('jarvis-chat')?.toLowerCase();
   if (!config.jarvis?.id || !config.jarvis?.version) return;
-  if (jarvis === 'on') {
+
+  if (jarvis) {
```
mokimo, line 18: How about adding jarvis to the early return statement?

libs/scripts/delayed.js
```diff
@@ -12,9 +12,14 @@
...
+    const desktopViewport = window.matchMedia('(min-width: 900px)').matches;
+    if (jarvis === 'mobile' && desktopViewport) return;
```
mokimo, line 20: What needs to be configured so it's available for both desktop and mobile? Just any string that is not 'mobile' or 'desktop' ?

mokimo (approved):

libs/scripts/delayed.js
narcis-radu: it just needs a value, we'll document it as `on` - `mobile, desktop, on`

libs/scripts/delayed.js
```diff
@@ -12,12 +12,15 @@
...
-  if (jarvis === 'on') {
-    const { initJarvisChat } = await import('../features/jarvis-chat.js');
-    initJarvisChat(config, loadScript, loadStyle);
-  }
+  const jarvis = getMetadata('jarvis-chat')?.toLowerCase();
+  if (!jarvis || !config.jarvis?.id || !config.jarvis?.version) return;
+
+  const desktopViewport = window.matchMedia('(min-width: 900px)').matches;
+  if (jarvis === 'mobile' && desktopViewport) return;
+  if (jarvis === 'desktop' && !desktopViewport) return;
```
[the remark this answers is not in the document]
overmyheadandbody: But there's no explicit value checking, so someone could set it to 'jarvis | testing1234' and Jarvis would be activated on that page. It's not terrible, but I'm not sure it's something we want to allow.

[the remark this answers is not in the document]
mokimo: Yeah it might be good to limit it to `mobile|desktop|on` and if it's not on, we just return.. we can always allow more options later but taking away the behaviour might otherwise break some sites in the future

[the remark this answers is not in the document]
narcis-radu: > someone could set it to 'jarvis | testing1234' and Jarvis would be activated on that page
That's actually acceptable, it's the same for other metadata. However, if you set `jarvis-chat|off`, then it would still be on, so I do agree that additional checks need to be performed

libs/scripts/delayed.js
```diff
@@ -12,12 +12,16 @@
...
-    const { initJarvisChat } = await import('../features/jarvis-chat.js');
-    initJarvisChat(config, loadScript, loadStyle);
-  }
+  const jarvis = getMetadata('jarvis-chat')?.toLowerCase();
+  if (!jarvis || !['mobile', 'desktop', 'on'].includes(jarvis)
+    || !config.jarvis?.id || !config.jarvis?.version) return;
+
+  const desktopViewport = window.matchMedia('(min-width: 900px)').matches;
+  if (jarvis === 'mobile' && desktopViewport) return;
+  if (jarvis === 'desktop' && !desktopViewport) return;
```
overmyheadandbody, line 21: Going deep into nitpicking territory, this could be written as a single check:
```
const jarvis = getMetadata('jarvis-chat')?.toLowerCase();
const desktopViewport = window.matchMedia('(min-width: 900px)').matches;

if (!jarvis
  || !['mobile', 'desktop', 'on'].includes(jarvis)
  || !config.jarvis?.id
  || !config.jarvis?.version
  || (jarvis === 'mobile' && desktopViewport)
  || (jarvis === 'desktop' && !desktopViewport)) return;
```

overmyheadandbody (approved):
I nit, therefore I am 😅 Code should work as expected, though, so no big objections 👍

libs/scripts/delayed.js
narcis-radu: That's correct, but it would require an additional (and useless) computation even if metadata is not defined.

mokimo (approved):

honstar (approved):

honstar:
FYI: codecov complains that the patch test coverage is not sufficient. @narcis-radu , can you please add corresponding tests? Thanks

narcis-radu:
@honstar - done

chrischrischris (approved):

PR #815, opened 2023-06-01
Surface unit test errors and point to a solution source

mokimo:
### Description
I played around with directly supplying an import map for all tests and mocking [all the utils](https://github.com/mokimo/milo/blob/import-maps-TODO/web-test-runner.config.mjs#L42-L62) but the major drawback of import maps in this specific case are that they aren't compatible with dynamic imports. It leads to a lot of headache and confusion, so I've abandoned that approach.

Instead, we can just log unit-test smells and provide developers with enough guidance https://github.com/orgs/adobecom/discussions/814

Resolves: [MWPW-129653](https://jira.corp.adobe.com/browse/MWPW-129653)

### More logs for when external scripts are loaded
I'm not the biggest fan of having a lot of logs BUT it's the only viable way we can provide guidance to milo core developers on what to look out for when writing unit tests.
When running tests in isolation it's also not a tremendous amount of logs to watch out for.
Screenshot 2023-06-01 at 15 00 30

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "test": "wtr --config ./web-test-runner.config.mjs \"./test/**/*.test.(js|html)\" --node-resolve --port=2000 --coverage --concurrent-browsers 4",
     "test:watch": "npm test -- --watch",
     "test:file": "wtr --config ./web-test-runner.config.mjs --node-resolve --port=2000 --coverage",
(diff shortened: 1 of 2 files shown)


honstar (approved):
+1

Ruchika4 (approved):
LGTM !

PR #819, opened 2023-06-01
MWPW-132078: hide ctas by id (feature parity with caas on dexter)

sheridansunier:
Authoring would work by having a separate panel where they can list ids and for those IDs the primary CTA will be removed. (either css display:none or just never render html). We have already completed this work in dexter and are implementing this in milo to have feature parity between the two authoring experiences.

Here's a detail of UX requirements from the original implementation in dexter:

We want a new tab similar to the "Feature Cards" and "Exclude Cards".
The new tab will be called "Hide CTAs" and the user will be the prompted to add content ID or card path, with "Add" option for more cards (similar to the Exclude Cards UX).

We also have a "Hide all cards in collection" option. While this duplicates the current functionality in UI->"Collection Button Style", this will be good in case users go to the new "Hide CTAs" tab and don't know why they can't hide all CTAs for the collection (even though they can if they used the current method)

Resolves: [MWPW-132078](https://jira.corp.adobe.com/browse/MWPW-132078)

**Test URLs:**
- After: https://MWPW-132078--milo--sheridansunier.hlx.page/tools/caas?martech=off


test/blocks/caas/utils.test.js
```diff
@@ -183,6 +183,7 @@ describe('getConfig', () => {
         reservoir: { sample: 3, pool: 1000 },
       },
       featuredCards: ['a', 'b'],
+      hideCtaIds: [""],
```
chrischrischris, line 186: nit
```suggestion
      hideCtaIds: [''],
```

libs/blocks/caas/utils.js
```diff
@@ -254,6 +254,7 @@ export const getConfig = async (state, strs = {}) => {
   const { country, language } = getCountryAndLang(state);
   const featuredCards = state.featuredCards && state.featuredCards.reduce(getContentIdStr, '');
   const excludedCards = state.excludedCards && state.excludedCards.reduce(getContentIdStr, '');
+  const hideCtaIds = state.hideCtaIds ? state.hideCtaIds.reduce(getContentIdStr, '') : "";
```
chrischrischris, line 257: nit: use `'`

chrischrischris (approved):

mokimo (approved):

test/blocks/caas/utils.test.js
sheridansunier: done - sorry for delay was on PTO lol

honstar (approved):

gmirijan:
@honstar The "Hide all CTAs in collection" implemented as part of https://jira.corp.adobe.com/browse/MWPW-123040

PR #820, opened 2023-06-01
MWPW-132076-fix css issues with georouting in rtl

Ruchika4:
* Fix issues with buttons and tabs in georouting when in rtl

Resolves: [MWPW-132076](https://jira.corp.adobe.com/browse/MWPW-132076)

**Test URLs:**
- Before: https://main--milo--ruchika4.hlx.page/ae_ar/drafts/ruchika/ar-page-us-ip?martech=off&akamaiLocale=ch
- After: https://georouting_rtl--milo--ruchika4.hlx.page/ae_ar/drafts/ruchika/ar-page-us-ip?martech=off&akamaiLocale=ch

Sharing URL for georouting with ltr text also as before after to make sure nothing broke these post this update. Before and after should be same in below scenario

- Before: https://main--milo--ruchika4.hlx.page/drafts/ruchika/us-page-ar-ip?akamaiLocale=ch&martech=off
- After: https://georouting_rtl--milo--ruchika4.hlx.page/drafts/ruchika/us-page-ar-ip?akamaiLocale=ch&martech=off


mokimo (approved):

ryanmparrish (approved):

overmyheadandbody (commented):
This doesn't really look like a sturdy solution to me. Having a look at this screenshot from the updated implementation:
Screenshot 2023-06-02 at 18 08 38

I don't get why:
* the body text (highlighted item and the description underneath it) is not right-aligned too, since all other elements are;
* why the 'Deutsch' element doesn't align with the outer margin of the body text, as is the case for LTR;
* why the caret next to the icon is not aligned with the icon.

I think the styles are over-complicated for their context. I'd advise using `column-gap` ([MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/column-gap)) since flexbox is used, it should make things a lot easier

Ruchika4:
> This doesn't really look like a sturdy solution to me. Having a look at this screenshot from the updated implementation: Screenshot 2023-06-02 at 18 08 38
> 
> I don't get why:
> 
> * the body text (highlighted item and the description underneath it) is not right-aligned too, since all other elements are;
> * why the 'Deutsch' element doesn't align with the outer margin of the body text, as is the case for LTR;
> * why the caret next to the icon is not aligned with the icon.
> 
> I think the styles are over-complicated for their context. I'd advise using `column-gap` ([MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/column-gap)) since flexbox is used, it should make things a lot easier

Hi @overmyheadandbody ,
1. Body text alignment was done as part of https://jira.corp.adobe.com/browse/MWPW-131201
2. I have fixed this as per xd it should be 56 px and now it is same for rtl as for ltr
3. Caret alignment was implemented as per the xd design https://xd.adobe.com/view/1088565a-caf4-4d87-90fb-4f276c82371e-45db/screen/74f2c64c-a839-4cb0-ac55-393e3e5ad86a/ provided during initial implementation of georoutingV2 and is not specifc to rtl design

elaineskpt (approved):

overmyheadandbody:
Thanks for the reply, @Ruchika4! My comments:

1. I still think this is a bit weird from a user perspective, but I get it's still being discussed, so this should be fine for now
2. Thanks for the update, looks good now!
3. I'm attaching 2 screenshots, I still think the caret is not positioned correctly. In the XD document, if you were to draw a line on top and below a small character, the caret will fit inside these lines. In the implementation, this doesn't seem to be the case:

| XD | Implementation |
| XD - caret fits inside small character lines | Implementation - XD - caret is outside the small character lines |

4. Another thing I noticed looking at the above screenshots: LTR item order is icon - text - caret, while on RTL it's caret - icon - text. Shouldn't these be consistent?

overmyheadandbody:
Seems that #4 from my previous comment might be better suited for @vhargrave: item order is changed between LTR and RTL:
| LTR - correct | RTL - seems incorrect |
| LTR - order correct | RTL - order incorrect |

Is there any reason for this or just an omission? Had a look on adobe.com and the implementation is correct there:

| LTR - correct | RTL - seems incorrect |
| Prod LTR | Prod RTL |

Maybe this doesn't need to be implemented as part of this PR, but a follow-up story should be created.

Ruchika4:
> Thanks for the reply, @Ruchika4! My comments:
> 
> 1. I still think this is a bit weird from a user perspective, but I get it's still being discussed, so this should be fine for now
> 2. Thanks for the update, looks good now!
> 3. I'm attaching 2 screenshots, I still think the caret is not positioned correctly. In the XD document, if you were to draw a line on top and below a small character, the caret will fit inside these lines. In the implementation, this doesn't seem to be the case:
> 
> XD	Implementation
> XD - caret fits inside small character lines	Implementation - XD - caret is outside the small character lines
> 4. Another thing I noticed looking at the above screenshots: LTR item order is icon - text - caret, while on RTL it's caret - icon - text. Shouldn't these be consistent?

Hi @overmyheadandbody , point 3 has been fixed both for LTR and RTL pages. 
For point 4 as you mentioned once we have confirmation about the expected behavior confirmed, I can look into implementing that followup story

overmyheadandbody:
Thanks for looking into the caret alignment as well, looks good now!

As for the item order, I'm pretty sure it's an omission in the original implementation. If you have a look on https://www.adobe.com/mena_ar/, for example, you'll see that the item order is correct (the caret is the last item, not the first, in an RTL context). Same thing holds true for the navigation for example, the carets are reflected. So it's pretty safe to say that a story can be created and get this addressed.

For the moment, we can leave this PR as-is (although I still think `column-gap` is better than setting margins 😅 ).

vhargrave (approved):
looks good to me 
@Ruchika4 & @overmyheadandbody - regarding point #4 , I would agree. The down arrow should be at the far left when rtl

PR #822, opened 2023-06-02
Remove `overflow: visible` from img to avoid undesired behaviour

narcis-radu:
Context: https://github.com/WICG/shared-element-transitions/blob/main/debugging_overflow_on_images.md

Resolves: [MWPW-132118](https://jira.corp.adobe.com/browse/MWPW-132118)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rparrish/icons?martech=off
- After: https://mwpw-132118--milo--narcis-radu.hlx.page/drafts/rparrish/icons?martech=off


overmyheadandbody (approved):
Had a quick look to see whether there are more such declarations in the codebase, but this seems to be the only one, otherwise, I might have suggested an explicit `clip`.

Cool find 👍

mokimo (approved):

rgclayton (approved):
Nice! Thank you!

PR #832, opened 2023-06-05
Global footer keyboard navigation

mokimo:
### Description
[Accessibility improvements] Enables keyboard navigation for the profile and footer.

Resolves: [MWPW-130524](https://jira.corp.adobe.com/browse/MWPW-130524) & [MWPW-132084](https://jira.corp.adobe.com/browse/MWPW-132084)

### How to test
Go to the global footer or profile - wait a few seconds until the keyboard navigation has fully loaded, use the arrow keys, shift, tab, space, enter, escape to navigate through on desktop and mobile. 

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/osahin/document?martech=off#
- After: https://footer-keyboard--milo--mokimo.hlx.page/drafts/osahin/document?martech=off#


libs/blocks/global-navigation/utilities/keyboard/index.js
```diff
@@ -25,6 +25,47 @@ const cycleOnOpenSearch = ({ e, isDesktop }) => {
...
+  if (!focusableElements.length) return { next: -1, prev: -1, items: [] };
+  const items = [...focusableElements];
+  const curr = items.findIndex((element) => element === e.target);
+  const next = curr + 1;
+  const prev = curr - 1;
+  return { next, prev, items };
+};
+const focusNextProfileItem = ({ e }) => {
+  const { items, next } = getProfileItems({ e });
+  if (items[next]) {
```
narcis-radu, line 49:
```suggestion
  if (items[next]) items[next].focus();
```

test/blocks/global-navigation/keyboard/mocks/global-nav-mobile.html
```diff
@@ -590,3 +590,141 @@
...
+    </a>
+  </li></ul>
+    </div>
+    <div class="feds-footer-legalWrapper"><p class="feds-footer-privacySection"><span class="feds-footer-copyright">
+  Copyright © 2023 All rights reserved.
+</span> / <a href="https://www.adobe.com/privacy.html" class="feds-footer-privacyLink">Privacy</a> / <a href="https://www.adobe.com/legal/terms.html" class="feds-footer-privacyLink">Terms of Use</a> / <a href="https://www.adobe.com/#openPrivacy" class="feds-footer-privacyLink">Cookie preferences</a> / <a href="https://www.adobe.com/privacy/ca-rights.html" class="feds-footer-privacyLink">Do not sell my personal information</a> / <a href="https://www.adobe.com/privacy/opt-out.html#interest-based-ads" class="feds-footer-privacyLink"><svg xmlns="http://www.w3.org/2000/svg" class="feds-adChoices-icon" focusable="false">
+  <use href=""></use>
+</svg>AdChoices</a></p></div>
+  </div>
+</div></footer>
```
narcis-radu, line 730: new line here

libs/blocks/global-navigation/utilities/keyboard/mobilePopup.js
```diff
@@ -80,116 +79,157 @@ class Popup {
...
-              this.mainNav.open();
-            } else {
-              this.mainNav.items[this.mainNav.curr].focus();
-            }
-            break;
+        openHeadline({ headline, focus: 'first' });
+        break;
+      }
+      case 'ArrowUp': {
+        this.mobileArrowUp({ prev, curr, element, isFooter });
```
narcis-radu, line 172: I think we need to add some CSS as well to highlight the active element, similar to the header.

header:
Screenshot 2023-06-06 at 11 30 01

footer:
Screenshot 2023-06-06 at 11 30 16

libs/blocks/global-navigation/utilities/keyboard/index.js
```diff
@@ -25,6 +25,47 @@ const cycleOnOpenSearch = ({ e, isDesktop }) => {
   }
 };
 
+const openProfile = ({ e, el }) => {
+  const button = e.target.closest(`${selectors.signIn}, ${selectors.profileButton}`);
+  if (button && button.getAttribute('aria-expanded') === 'false') {
+    e.target.click();
+    el.querySelector(selectors.profileDropdown).focus();
```
narcis-radu, line 32:
```suggestion
    el.querySelector(selectors.profileDropdown)?.focus();
```
just to be extra safe

libs/blocks/global-navigation/utilities/keyboard/index.js
```diff
@@ -25,6 +25,47 @@ const cycleOnOpenSearch = ({ e, isDesktop }) => {
...
+};
+
+const getProfileItems = ({ e }) => {
+  const focusableElements = document.querySelectorAll(selectors.profileDropdown);
+  if (!focusableElements.length) return { next: -1, prev: -1, items: [] };
+  const items = [...focusableElements];
+  const curr = items.findIndex((element) => element === e.target);
+  const next = curr + 1;
+  const prev = curr - 1;
+  return { next, prev, items };
```
narcis-radu, line 45:
```suggestion
  return { curr + 1, curr - 1, items };
```

libs/blocks/global-navigation/utilities/keyboard/mobilePopup.js
```diff
@@ -80,116 +79,157 @@ class Popup {
...
-          break;
+  focusMainNav = (isFooter) => {
+    if (isFooter) return;
+    this.mainNav.items[this.mainNav.curr].focus();
+  };
+
+  focusMainNavNext = (isFooter) => {
+    if (isFooter) return;
+    this.mainNav.focusNext();
+    this.mainNav.open({});
```
chrischrischris, line 130: nit: no need for {} since it's the default arg

libs/blocks/global-navigation/utilities/keyboard/popup.js
```diff
@@ -45,87 +44,114 @@ class Popup {
...
-              this.mainNav.focusNext();
-              this.mainNav.open({});
-              break;
-            }
-            popupItems[next].focus();
-          }
+  focusMainNavNext = (isFooter) => {
+    if (isFooter) return;
+    this.mainNav.focusNext();
+    this.mainNav.open({});
```
chrischrischris, line 55: same here

chrischrischris (approved):

auniverseaway (approved):

libs/blocks/global-navigation/utilities/keyboard/mobilePopup.js
mokimo: That should work now!

libs/blocks/global-navigation/utilities/keyboard/index.js
```diff
@@ -25,6 +25,43 @@ const cycleOnOpenSearch = ({ e, isDesktop }) => {
...
+const getProfileItems = ({ e }) => {
+  const focusableElements = document.querySelectorAll(selectors.profileDropdown);
+  if (!focusableElements.length) return { next: -1, prev: -1, items: [] };
+  const items = [...focusableElements];
+  const curr = items.findIndex((element) => element === e.target);
+  return { next: curr + 1, prev: curr - 1, items };
+};
+const focusNextProfileItem = ({ e }) => {
+  const { items, next } = getProfileItems({ e });
+  if (items[next]) items[next].focus();
```
overmyheadandbody, line 47: It's not so straightforward to test profile-related code from this branch, so I might have some silly questions: I see that `focusPrevProfileItem` closes the profile if a previous element is not found. Should this happen here as well? What happens after tab-ing on the last focusable element of the profile dropdown? Does it close and focus the next navigation item?

libs/blocks/global-navigation/utilities/keyboard/index.js
```diff
@@ -25,6 +25,43 @@ const cycleOnOpenSearch = ({ e, isDesktop }) => {
...
+  if (button?.getAttribute('aria-expanded') === 'false') {
+    e.target.click();
+    el.querySelector(selectors.profileDropdown)?.focus();
+    return true;
+  }
+  return false;
+};
+
+const getProfileItems = ({ e }) => {
+  const focusableElements = document.querySelectorAll(selectors.profileDropdown);
```
overmyheadandbody, line 39: Would naming this something like `profileDropdownLinks` or `profileDropdownButtons` make a bit more sense? I got a bit confused initially, expecting for this to just represent a dropdown, not focusable items inside it.

libs/blocks/global-navigation/utilities/keyboard/index.js
```diff
@@ -37,30 +74,46 @@ class KeyboardNavigation {
...
+              break;
+            }
+            case 'Enter': {
+              if (e.target.closest(selectors.searchField)) return;
+              e.stopPropagation();
+              e.preventDefault();
+              e.target.click();
+              break;
+            }
+            case 'Space': {
```
overmyheadandbody, line 92: Space and Enter seem identical, you can just do
```
case 'Enter':
case 'Space': {
  // shared code
}
```

libs/blocks/global-navigation/utilities/keyboard/mobilePopup.js
```diff
@@ -6,17 +6,20 @@ import {
...
   selectors,
 } from './utils.js';
-import { closeAllDropdowns, logErrorFor } from '../utilities.js';
+import { closeAllDropdowns, logErrorFor, setActiveDropdown } from '../utilities.js';
 
 const closeHeadlines = () => {
   const open = [...document.querySelectorAll(`${selectors.headline}[aria-expanded="true"]`)];
   open.forEach((el) => el.setAttribute('aria-expanded', 'false'));
+  // Shift active class back to the parent of the first closed headline
+  setActiveDropdown(open[0]);
```
overmyheadandbody, line 15: After the merge of https://github.com/adobecom/milo/pull/823, it would be good to double-check that this is still working as expected. Same for all other newly added `setActiveDropdown` calls from this PR.

test/blocks/global-navigation/keyboard/keyboard.test.js
```diff
@@ -485,7 +488,7 @@ describe('keyboard navigation', () => {
       keyboardNavigation.mainNav.setActive(trigger);
       keyboardNavigation.mainNav.open();
       navLinks = getNavLinks(trigger);
-      firstPopupItem = navLinks[0];
+      [firstPopupItem] = navLinks;
```
overmyheadandbody, line 491: 👍

overmyheadandbody (commented):
Awesome work! As per my MO, I do have some comments 😅 

I also noticed that the keyboard navigation is also working on desktop. We didn't really get this requirement, since all elements are already visible and thus accessible. But I guess it doesn't hurt to have it either.

Another note is that we should open an enhancement story to ensure the outlines are visible on mobile when the keyboard navigation is shifting focus.

libs/blocks/global-navigation/utilities/keyboard/mobilePopup.js
mokimo: Actually only integrated `setActiveDropdown` after rebasing to #823  yesterday

libs/blocks/global-navigation/utilities/keyboard/index.js
mokimo: Hm interestingly enough...
- last element of a profile -> tab -> next item but popup not closed
- last element of a popup (no next main nav item) -> tab -> does nothing

That's wrong in the current (aem/Eurovision milo gnav) version too you shouldn't get stuck using tab. I changed that.

I'll change that to
- last element of a profile -> tab -> close + focus next item
- last element of a popup (no next main nav item) -> tab -> close popup -> tab -> next item (if you wonder why close first and then another tab is needed... well otherwise things get a little messy)

mokimo: Thanks for raising this question, I uncovered some bigger issues when the keyboard navigation isn't loaded or nav items are clicked and the internal state is a bit off.

I refactored a bit of the logic and made some improvements where it made sense such as the flows described above

overmyheadandbody (approved):
Looking better, thanks for addressing things so quickly!

PR #846, opened 2023-06-07
Add aria label to profile thumbnail

narcis-radu:
Fix a small accessibility issue by adding `aria-label` to profile thumbnail.

**Test URLs:**
- After: https://profile-accessibility--milo--narcis-radu.hlx.page/drafts/narcis/default?martech=off

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -25,7 +25,7 @@ import {
   lanaLog,
 } from './utilities/utilities.js';
 
-import { replaceKey } from '../../features/placeholders.js';
+import { replaceKey, replaceKeyArray } from '../../features/placeholders.js';
 
 const CONFIG = {
   icons: {
@@ -93,8 +93,8 @@ const decorateSignIn = async ({ rawElem, decoratedElem }) => {
 };
 
 const decorateProfileTrigger = async ({ avatar }) => {
-  const label = await replaceKey(
-    'profile-button',
+  const [label, profileAvatar] = await replaceKeyArray(
+    ['profile-button', 'profile-avatar'],
     getFedsPlaceholderConfig(),
     'feds',
   );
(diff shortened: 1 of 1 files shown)


mokimo (approved):

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -108,7 +108,7 @@ const decorateProfileTrigger = async ({ avatar }) => {
       daa-ll="Account"
       aria-haspopup="true"
     >
-      <img class="feds-profile-img" src="${avatar}"></img>
+      <img class="feds-profile-img" src="${avatar}" aria-label="${profileAvatar}"></img>
```
overmyheadandbody, line 111: From [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label):

> By default, [...] , an image's accessible name is the content of its alt attribute, [...]

I'm wondering why this needs an `aria-label` instead of a simple `alt`?

narcis-radu: There are two reasons:
1. be consistent with existing code: https://github.com/adobecom/milo/blob/main/libs/blocks/global-navigation/features/profile/dropdown.js#L108
2. documentation mentions either `alt`, `aria-label` or `aria-labelledby` can be used

overmyheadandbody: For 2.:

> If none of these options are available, or if the default accessible name is not appropriate, use the aria-label attribute to define the accessible name of an element.

But `alt` is an available option, since it's an `img` tag. With this in mind, I'd argue that 1. is wrong and it should also be swapped with an `alt`. It's more accurate from a syntax perspective and we save a total of 14 characters 😅

narcis-radu: It's a small change, I can do it. The documentation mentions either is good for images, but `alt` is a native attribute.

mokimo (approved):

overmyheadandbody (approved):

PR #851, opened 2023-06-08
MWPW-124973-jump to section on tab change

SonjaPopovic:
* if there is a section with properties tab and tab-jump, on tab change document will be scrolled to that section
Feature demonstrated in AEM :  
https://dc-author.dev01.corp.adobe.com/content/dx-dc/us/en/test-page/jump-to-section-navigation-list.html?wcmmode=disabled 

Same behaviour implemented here: 
- on initial page load, do not jump to section, 
- on change of tab, jump to connected section 

Resolves: [MWPW-124973](https://jira.corp.adobe.com/browse/MWPW-124973)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/sonja/tabs?martech=off
- After: https://jumptosection--milo--sonjapopovic.hlx.page/drafts/sonja/tabs?martech=off


Blainegunn:
There are some parts of this functionality that need to be confirmed. 
Right now the section being jumped to is hidden when the corresponding tab isn't selected. I'm going to close this PR until we get more info. When we do I'll update this code to work, @SonjaPopovic thank you

PR #853, opened 2023-06-09
[MWPW-132468] Legal Block Content Not rendered in italics - Bug Fix

drashti1712:
Resolves: [MWPW-132468](https://jira.corp.adobe.com/browse/MWPW-132468)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/Soujanya/cc/textblock
- After: https://legalblockfix--milo--drashti1712.hlx.page/drafts/Soujanya/cc/textblock

### DIFF
--- a/libs/blocks/text/text.css
+++ b/libs/blocks/text/text.css
@@ -109,10 +109,11 @@
 }
 
 /* Legal */
-.legal.text-block .foreground {
+.legal.text-block {
   padding: var(--spacing-xl) 0;
 }
 
+.legal.text-block .foreground,
 .legal.text-block p {
   font-style: italic;
   font-size: var(--type-body-xxs-size);


libs/blocks/text/text.css
```diff
@@ -109,10 +109,11 @@
 }
 
 /* Legal */
-.legal.text-block .foreground {
+.legal.text-block {
```
elaineskpt, line 112: Hi @drashti1712,

why does the .foreground need to be removed here?
Saw the padding was working properly.

drashti1712: Hi @elaineskpt, I removed it as we are also considering an authoring change as mentioned in the comments for this issue: https://jira.corp.adobe.com/browse/MWPW-132398
So if we add another row to accommodate this use-case then, we would need padding on the outer div and not on each row (foreground). 
And even if we do not accommodate this change, adding a padding to the outer div instead of on the row seemed logical.

elaineskpt (approved):

overmyheadandbody (approved):

mokimo (approved):

PR #854, opened 2023-06-09
Remove for..of, await in loop eslint warnings

chrischrischris:
We use await inside of for..of loops enough that we don't need to have an eslint warning for it.  Saves a bit of file size too.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,26 +10,37 @@ module.exports = {
     requireConfigFile: false,
   },
   rules: {
-    // allow reassigning param
-    'no-param-reassign': [2, { props: false }],
-    'linebreak-style': ['error', 'unix'],
+    'chai-friendly/no-unused-expressions': 2,
     'import/extensions': ['error', { js: 'always' }],
+    'linebreak-style': ['error', 'unix'],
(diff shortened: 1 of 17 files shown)


.eslintrc.js
```diff
@@ -10,26 +10,37 @@ module.exports = {
...
-    'no-unused-expressions': 0,
-    'chai-friendly/no-unused-expressions': 2,
-
   },
   overrides: [
-    {
-      files: ['test/**/*.js'],
-      rules: { 'no-console': 'off' },
-    },
+    { files: ['test/**/*.js'] },
```
overmyheadandbody, line 43: Since there's no longer an override for this path, is it worth keeping it?

overmyheadandbody (approved):
Good idea, no use in having rules that just make developers mute them in so many cases.

sukamat (approved):

mokimo (approved):
Cool like the initiative of customizing the eslint rules as it fits the code base based on how we actually write code. 👍

.eslintrc.js
chrischrischris: eh probably not but I'm going to leave it just in case at some point we want to add other test folders

PR #855, opened 2023-06-09
MWPW-129907 [Blog Migration] Allow Custom Article Feed

rgclayton:
Updated to allow custom path for article feed.

**Context**
Milo defaults english to the root of a site. Sub folders for additional locales/regions.
Blog.adobe.com does not. English content is contained in `/en/` similar to all the additional languages `/de/, /fr/, /ko/, /jp/, /br/`, etc. 

The english homepage (Index) is located at the root of the site. 
All this created a challenge to provide a correct `query-index.json` path for the article feed.

The solution was to add a new key/value row to the Article Feed block to supply a custom feed for the english homepage articles.
image

Resolves: [MWPW-129907](https://jira.corp.adobe.com/browse/MWPW-129907)

**Test URLs:**
- After: https://blog-article-feed--milo--adobecom.hlx.page/?martech=off

**Blog Test URLs:**
- After: https://main--blog--adobecom.hlx.page/?milolibs=blog-article-feed&martech=off

**Testing**
* Click on the **blog before** url and notice the article feed - below the marquee area - and you will see the feed shows the most recent article posted is from 4/21/23. This is incorrect. 
* Click on the **blog after** url will show the most recent article published (as of today 6/9/23) is from 6/9/2023. This can be confirmed by looking at [blog.adobe.com](https://blog.adobe.com/) and seeing the same article dates.


libs/blocks/article-feed/article-feed.js
```diff
@@ -520,14 +526,13 @@ async function decorateFeedFilter(articleFeedEl, config) {
...
-  const config = readBlockConfig(articleFeed);
-  clearBlock(articleFeed);
+export default async function init(el) {
+  const config = readBlockConfig(el);
+  blogIndex.config = config;
+  el.innerHTML = '';
   await loadTaxonomy();
   if (config.filters) {
-    decorateFeedFilter(articleFeed, config);
+    decorateFeedFilter(el, config);
```
overmyheadandbody, line 535: Since you're now caching the config in `blogIndex.config`, is it still worth passing the `config` between multiple functions? It might be easier to just read `blogIndex.config` wherever needed and lower the number of arguments where `config` is used.

overmyheadandbody (approved):

mokimo (approved):

JasonHowellSlavin (approved):

libs/blocks/article-feed/article-feed.js
rgclayton: Great point! On it.

PR #862, opened 2023-06-12
Personalization

chrischrischris:
Adds personalization to Milo.

More info coming here soon...

Resolves: [MWPW-124495](https://jira.corp.adobe.com/browse/MWPW-124495)

**Test URLs:**
- After: https://cpeyer-manifest--milo-target--adobecom.hlx.page/?martech=off


libs/scripts/preview/preview.js
```diff
@@ -0,0 +1,296 @@
+/*
```
chrischrischris, line 1: FYI - This file is a port from Franklin and still has work that needs to be done to make it compatible with Milo.  For now it's ok to ignore this file

libs/blocks/fragment-personalization/fragment-personalization.js
```diff
@@ -0,0 +1,15 @@
...
+export default function init(el) {
+  el.classList.add('contained');
+  const rows = el.querySelectorAll(':scope > div');
+  rows.forEach((row, rdx) => {
+    row.className = `row row-${rdx + 1}`;
+    const cols = row.querySelectorAll(':scope > div');
+    cols.forEach((col, cdx) => {
+      col.className = `col col-${cdx + 1}`;
+    });
+    // row.addEventListener('click', (e) => console.log(e.target.closest('.row')));
```
rgclayton, line 12: Remove comment?

libs/utils/utils.js
```diff
@@ -365,16 +390,21 @@ export async function loadTemplate() {
 
 export async function loadBlock(block) {
   const name = block.classList[0];
-  const { miloLibs, codeRoot } = getConfig();
+  const { miloLibs, codeRoot, p13nBlocks } = getConfig();
```
rgclayton, line 393: Curious about the "p13nBlocks" name. Just a reference for personalization?

libs/utils/utils.js
```diff
@@ -437,13 +467,22 @@ export function decorateAutoBlock(a) {
...
+        if (url.hash === '') {
+          const { parentElement } = a;
+          const { nodeName, innerHTML } = parentElement;
+          const noText = innerHTML === a.outerHTML;
+          if (noText && nodeName === 'P') {
+            const div = createTag('div', null, a);
+            parentElement.parentElement.replaceChild(div, parentElement);
+          }
+        } else {
+          // Modals
```
rgclayton, line 480: Will this change impact in page jump links (TOC)? Those aren't a modal or a fragment.

libs/scripts/personalization.js
```diff
@@ -0,0 +1,374 @@
...
+    const names = variant.split(RE_SPACE_COMMA).filter(Boolean);
+    if (names.some((name) => (
+      name === variantLabel || (tagNames.includes(name) && PERSONALIZATION_TAGS[name]())))
+    ) {
+      return variant;
+    }
+    return false;
+  });
+  return matchingVariant;
+}
```
mokimo, line 221: This can be simplified (untested code below)
```js
function getPersonalizationVariant(variantNames = [], variantLabel = null) {
  const tagNames = Object.keys(PERSONALIZATION_TAGS);
  const matchingVariant = variantNames.find((variant) => {
    const names = variant.split(RE_SPACE_COMMA).filter(Boolean);
    // Simplified conditional logic by using implicit return
    return names.some((name) => name === variantLabel || (tagNames.includes(name) && PERSONALIZATION_TAGS[name]()));
  });
  return matchingVariant;
}
```

libs/scripts/personalization.js
```diff
@@ -0,0 +1,374 @@
...
+      console.log('error loading manifest:', resp);
+      return null;
+    }
+    const json = await resp.json();
+    return json.data;
+  } catch (e) {
+    console.log(`error loading manifest: ${path}`, e);
+  }
+  return null;
+}
```
mokimo, line 236: If we don't need the extra console.log we could also do
```suggestion
async function fetchManifest(path) {
  try {
    const resp = await fetch(path);
    const json = await resp.json();
    return json.data;
  } catch (e) {
    console.error(`Error loading manifest: ${path}`, e);
    return null;
  }
}
```

libs/scripts/personalization.js
```diff
@@ -0,0 +1,374 @@
...
+
+export async function fragmentPersonalization(doc) {
+  const fpTableRows = doc.querySelectorAll('.fragment-personalization > div');
+  doc.querySelector('.fragment-personalization').remove();
+  if (!fpTableRows) return doc;
+
+  const variantNames = [];
+  const info = [...fpTableRows].reduce((obj, row) => {
+    if (row.children[0]?.innerText?.toLowerCase() === 'action') {
```
mokimo, line 269: Wdyt about giving the element names using destructuring instead of accessing the indexes?

So something like:
```js
  const info = [...fpTableRows].reduce((obj, row) => {
    const [actionEl, selectorEl, variantEl, fragment] = row.children
    if (actionEl?.innerText?.toLowerCase() === 'action') {
```

libs/scripts/personalization.js
```diff
@@ -0,0 +1,374 @@
...
+        selectedEl.replaceWith(cmd.htmlFragment);
+      } else if (['insertbefore', 'insertcontentbefore'].includes(cmd.action)) {
+        selectedEl.insertAdjacentElement('beforebegin', cmd.htmlFragment);
+      } else if (['insertafter', 'insertcontentafter'].includes(cmd.action)) {
+        selectedEl.insertAdjacentElement('afterend', cmd.htmlFragment);
+      } else if (['remove', 'removecontent'].includes(cmd.action)) {
+        selectedEl.remove();
+      }
+    });
+  }
```
mokimo, line 301: Not a huge fan of if/else statements, maybe a switch?

Something like:
```js
const selectedVariant = getPersonalizationVariant(variantNames);
if (selectedVariant) {
  info[selectedVariant].forEach((cmd) => {
    const selectedEl = doc.querySelector(cmd.selector);
    if (!selectedEl) return;

    // Define the location for the insertAdjacentElement function
    let location;
    switch (cmd.action) {
      case 'replace':
      case 'replacecontent':
        selectedEl.replaceWith(cmd.htmlFragment);
        break;
      case 'insertbefore':
      case 'insertcontentbefore':
        selectedEl.insertAdjacentElement('beforebegin', cmd.htmlFragment);
        break;
      case 'insertafter':
      case 'insertcontentafter':
        selectedEl.insertAdjacentElement('afterend', cmd.htmlFragment);
        break;
      case 'remove':
      case 'removecontent':
        selectedEl.remove();
        return;
      default:
        console.warn(`Unknown action: ${cmd.action}`);
        return;
    }
  });
}
```

libs/scripts/preview/preview.css
```diff
@@ -0,0 +1,134 @@
...
+    background-color: #666;
+}
+
+.hlx-variant > div {
+    flex: 1 1 auto;
+}
+
+.hlx-variant > div.hlx-button {
+    flex: 0 0 auto;
+}
```
mokimo, line 134: EOL

libs/utils/utils.js
```diff
@@ -86,6 +87,7 @@ const AUTO_BLOCKS = [
 const ENVS = {
   local: {
     name: 'local',
+    adobeIO: 'cc-collab.adobe.io',
```
mokimo, line 90: We use the stage edge config & pdfViewer, should we just copy all the stage configs to local?
It would also make sense for ims etc. to properly test locally using stage data

mokimo (commented):
Would be awesome if you can add some setup/test instructions or (wip-)documentation -  I'd love to check this out and play around with it a bit and documentation will be needed at some point anyway 😁 

Left some comments / suggestions / questions, feel free to just resolve something if u don't agree with it!

libs/scripts/personalization.js
```diff
@@ -0,0 +1,374 @@
...
+function handleCommands(commands, rootEl) {
+  const mainEl = rootEl || document.querySelector('main');
+
+  commands.forEach((cmd) => {
+    if (VALID_COMMANDS.includes(cmd.action)) {
+      let targetEl = mainEl.querySelector(cmd.selector);
+
+      if (!targetEl) return;
+
+      if (targetEl.classList[0] === 'section-metadata') {
```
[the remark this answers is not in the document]
chrischrischris: The block class will always be the first class.

libs/utils/utils.js
chrischrischris: This was included by mistake and I will delete.

libs/utils/utils.js
chrischrischris: yep, just an abbreviation for personalization.  Similar to l10n = localization and a11y = accessibility

libs/utils/utils.js
chrischrischris: This was a merge messup - will revert back to current utils.js code

libs/blocks/fragment-personalization/fragment-personalization.css
```diff
@@ -0,0 +1,55 @@
...
+  padding: 10px;
+  margin-top: 16px !important;
+}
+
+.fragment-personalization > .row {
+  display: grid;
+  gap: var(--spacing-s);
+  margin-bottom: var(--spacing-xs);
+  grid-template-columns: repeat(auto-fit, minmax(200px, 1fr));
+  border-bottom: 1px solid gray;
```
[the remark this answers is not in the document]
chrischrischris: This is a straight copy from columns.css, which makes me realize I should just add the `.columns` class to this and not copy the code

chrischrischris:
Closing as the forked branch has been moved to milo and a new PR has been raised.  All PR feedback has been incorporated into the new PR #874

PR #866, opened 2023-06-13
Mwpw 132409 - Optional Headings for Link Farms

drashti1712:
* This extends the functionality of link farm variant of text blocks by allowing the authors to add optional Headings for the link columns. The solution, as discussed in the comments and with the consonant team, would be to add a [No-Heading] placeholder in the authoring for blocking a space for the heading as shown below. 
* Related to - https://jira.corp.adobe.com/browse/MWPW-129911, https://jira.corp.adobe.com/browse/MWPW-132409, https://jira.corp.adobe.com/browse/MWPW-132427
* Update: The solution was revised and we no longer need to add a [No-Heading] placeholder in authoring.

Resolves: [MWPW-132409](https://jira.corp.adobe.com/browse/MWPW-132409)

Authoring (updated):
image

Before:
image

After: 
image

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/Drashti/document
- After: https://mwpw-132409--milo--drashti1712.hlx.page/drafts/Drashti/document1#

### DIFF
--- a/libs/blocks/text/text.css
+++ b/libs/blocks/text/text.css
@@ -141,7 +141,7 @@
   line-height: var(--type-heading-xs-lh);
 }
 
-.link-farm.text-block [id^="no-heading"] {
+.link-farm.text-block .no-heading {
   margin: 0;
 }
 
@@ -190,7 +190,7 @@
     grid-template-columns: repeat(2, minmax(0, 1fr));
   }
 
-  .link-farm.text-block [id^="no-heading"] {
+  .link-farm.text-block .no-heading {
     height: var(--type-heading-m-size);
     margin: 0 0 var(--spacing-l);
   }
--- a/libs/blocks/text/text.css
+++ b/libs/blocks/text/text.css
@@ -141,6 +141,10 @@
   line-height: var(--type-heading-xs-lh);
 }
 
+.link-farm.text-block [id^="no-heading"] {
+  margin: 0;
+}
+
 .link-farm.text-block h2 {
   margin: 0 0 var(--spacing-m);
 }
@@ -181,9 +185,14 @@
   .inset.text-block .foreground > div {
     padding-left: var(--spacing-l);
   }
-  
+
   .link-farm.text-block .foreground {
     grid-template-columns: repeat(2, minmax(0, 1fr));
+  }
+
+  .link-farm.text-block [id^="no-heading"] {
+    height: var(--type-heading-m-size);
+    margin: 0 0 var(--spacing-l);
   }
 }
 
@@ -191,5 +200,10 @@
 @media screen and (min-width: 1200px) {
   .link-farm.text-block .foreground {
     grid-template-columns: repeat(4, minmax(0, 1fr));
+  }
+
+  .link-farm.text-block [id^="no-heading"] {
+    height: var(--type-heading-m-size);
+    margin: 0 0 var(--spacing-l);
   }
 }
--- a/libs/blocks/text/text.css
+++ b/libs/blocks/text/text.css
@@ -201,9 +201,4 @@
   .link-farm.text-block .foreground {
     grid-template-columns: repeat(4, minmax(0, 1fr));
   }
-
-  .link-farm.text-block [id^="no-heading"] {
-    height: var(--type-heading-m-size);
-    margin: 0 0 var(--spacing-l);
-  }
 }
--- a/libs/blocks/text/text.js
+++ b/libs/blocks/text/text.js
@@ -51,8 +51,14 @@ export default function init(el) {
     if (elAction) elAction.classList.add('body-s');
   }
   if (el.classList.contains('link-farm')) {
-    el.querySelectorAll('[id^="no-heading"]').forEach(elem => {
-      elem.innerHTML = '';
+    const foregroundDiv = el.querySelectorAll('.foreground')[1];
+    const count = foregroundDiv.querySelectorAll('h3').length;
+    foregroundDiv.querySelectorAll('div').forEach(divElem => {
+      if (!divElem.querySelector('h3') && count) {
+        let headingElem = document.createElement('h3');
+        headingElem.setAttribute('id', 'no-heading');
+        divElem.insertBefore(headingElem, divElem.firstChild);
+      }
     });
   }
   el.classList.add(...helperClasses);
--- a/libs/blocks/text/text.js
+++ b/libs/blocks/text/text.js
@@ -1,4 +1,5 @@
 import { decorateBlockBg, decorateBlockText, getBlockSize, decorateTextOverrides } from '../../utils/decorate.js';
+import { createTag } from '../../utils/utils.js';
 
 // size: [heading, body, ...detail]
 const blockTypeSizes = {
@@ -55,8 +56,7 @@ export default function init(el) {
     const count = foregroundDiv.querySelectorAll('h3').length;
     foregroundDiv.querySelectorAll('div').forEach(divElem => {
       if (!divElem.querySelector('h3') && count) {
-        let headingElem = document.createElement('h3');
-        headingElem.setAttribute('id', 'no-heading');
+        const headingElem = createTag('h3', { class: 'no-heading'});
         divElem.insertBefore(headingElem, divElem.firstChild);
       }
     });
--- a/libs/blocks/text/text.js
+++ b/libs/blocks/text/text.js
@@ -50,6 +50,11 @@ export default function init(el) {
     const elAction = el.querySelector('.action-area');
     if (elAction) elAction.classList.add('body-s');
   }
+  if (el.classList.contains('link-farm')) {
+    el.querySelectorAll('[id^="no-heading"]').forEach(elem => {
+      elem.innerHTML = '';
+    });
+  }
   el.classList.add(...helperClasses);
   decorateTextOverrides(el);
 }
--- a/test/blocks/text/text.test.js
+++ b/test/blocks/text/text.test.js
@@ -31,4 +31,33 @@ describe('text block', () => {
       expect(body).to.exist;
     });
   });
+
+  describe('Link Farm', () => {
+    it('should check if the "link-farm" class is present', () => {
+      const element = document.querySelector('.link-farm');
+      expect(element).to.exist;
+    });
+  
+    it('should check if the "h3" elements are added when necessary', () => {
+      const element = document.querySelector('.link-farm');
+      const foregroundDiv = element.querySelectorAll('.foreground')[1];
+      const headingElem = foregroundDiv.querySelectorAll('h3');
+      expect(headingElem.length).to.equal(4);
+    });
+    it('should add no-heading id to the h3 element', () => {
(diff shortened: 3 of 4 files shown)


chrischrischris:
Instead of having to explicitly specify `[No-Heading]` could you instead apply an H2 Style to the column headers which "promotes" them into column headers?

If the link farm will always be links in the columns, you could just treat any non-link text at the top as the heading.

libs/blocks/text/text.css
```diff
@@ -181,15 +185,25 @@
...
 /* Desktop up */
 @media screen and (min-width: 1200px) {
   .link-farm.text-block .foreground {
     grid-template-columns: repeat(4, minmax(0, 1fr));
   }
+
+  .link-farm.text-block [id^="no-heading"] {
+    height: var(--type-heading-m-size);
+    margin: 0 0 var(--spacing-l);
+  }
```
JasonHowellSlavin, line 208: I don't think you need this block since you add the same selectors and properties in the 1min-width: 600` media query.

JasonHowellSlavin:
Should update tests or add a new test as well.

libs/blocks/text/text.js
```diff
@@ -50,6 +50,11 @@ export default function init(el) {
     const elAction = el.querySelector('.action-area');
     if (elAction) elAction.classList.add('body-s');
   }
+  if (el.classList.contains('link-farm')) {
+    el.querySelectorAll('[id^="no-heading"]').forEach(elem => {
+      elem.innerHTML = '';
+    });
+  }
```
ryanmparrish, line 57: Does the `no-heading` selector consider localization? The final doc example might just need a `#_dns` in the label. Just a thought

ryanmparrish (approved):

ryanmparrish:
Another thought for authoring is have an additional header row in the doc table so you could just leave cells blank that don't need a headline.

drashti1712:
> Another thought for authoring is have an additional header row in the doc table so you could just leave cells blank that don't need a headline.

@ryanmparrish 
We had considered this authoring, but since these are text blocks and not columns, this type of authoring gives rise to another issue as raised here: https://jira.corp.adobe.com/browse/MWPW-132427
An additional header row will create a div element for each heading which are inside a single row div. Due to this, the headings cannot be linked to the links in the corresponding column for tablet and mobile view.

libs/blocks/text/text.css
drashti1712: Resolved. Thanks for pointing out!

libs/blocks/text/text.js
```diff
@@ -50,6 +50,11 @@ export default function init(el) {
     const elAction = el.querySelector('.action-area');
     if (elAction) elAction.classList.add('body-s');
   }
+  if (el.classList.contains('link-farm')) {
+    el.querySelectorAll('[id^="no-heading"]').forEach(elem => {
```
aishwaryamathuria, line 54: @drashti1712 
In place of adding [No Heading] and query selecting that .. Can we do the following? - 
1. Query select column divs inside second foreground div of text block
2. Look for ones without a heading tag
3. Append a heading tag at the top of that div?
We will not have to make any authoring changes for the same....

@chrischrischris @ryanmparrish @salonijain3

ryanmparrish: This seems like a good solve for authoring ease. We might want to just check if there is a header or <p> w/ out a child <a> tag(s) incase the author does not apply a heading style to the top label

drashti1712: Hi @aishwaryamathuria, @ryanmparrish 
I went ahead with this approach and made the changes. No placeholder will be required in the authoring for this approach. Please review.

drashti1712:
> Should update tests or add a new test as well.

@JasonHowellSlavin I have updated the tests for the new logic.

libs/blocks/text/text.css
```diff
@@ -181,10 +185,15 @@
   .inset.text-block .foreground > div {
     padding-left: var(--spacing-l);
   }
-  
+
   .link-farm.text-block .foreground {
     grid-template-columns: repeat(2, minmax(0, 1fr));
   }
+
+  .link-farm.text-block [id^="no-heading"] {
```
ryanmparrish, line 193:
```suggestion
  .link-farm.text-block .no-heading {
```

libs/blocks/text/text.js
ryanmparrish, line 1: add...
`import { createTag } from '../../utils/utils.js';`

libs/blocks/text/text.js
```diff
@@ -50,6 +50,17 @@ export default function init(el) {
     const elAction = el.querySelector('.action-area');
     if (elAction) elAction.classList.add('body-s');
   }
+  if (el.classList.contains('link-farm')) {
+    const foregroundDiv = el.querySelectorAll('.foreground')[1];
+    const count = foregroundDiv.querySelectorAll('h3').length;
+    foregroundDiv.querySelectorAll('div').forEach(divElem => {
+      if (!divElem.querySelector('h3') && count) {
+        let headingElem = document.createElement('h3');
```
ryanmparrish, line 58:
```suggestion
        const headingElem = document.createTag('h3', { class: 'no-heading' ));
```

libs/blocks/text/text.js
```diff
@@ -50,6 +50,17 @@ export default function init(el) {
...
+        headingElem.setAttribute('id', 'no-heading');
```
ryanmparrish, line 59:
```suggestion
```

libs/blocks/text/text.css
```diff
@@ -141,6 +141,10 @@
   line-height: var(--type-heading-xs-lh);
 }
 
+.link-farm.text-block [id^="no-heading"] {
```
ryanmparrish, line 144:
```suggestion
.link-farm.text-block .no-heading {
```

test/blocks/text/text.test.js
```diff
@@ -31,4 +31,33 @@ describe('text block', () => {
...
+      expect(element).to.exist;
+    });
+  
+    it('should check if the "h3" elements are added when necessary', () => {
+      const element = document.querySelector('.link-farm');
+      const foregroundDiv = element.querySelectorAll('.foreground')[1];
+      const headingElem = foregroundDiv.querySelectorAll('h3');
+      expect(headingElem.length).to.equal(4);
+    });
+    it('should add no-heading id to the h3 element', () => {
```
ryanmparrish, line 47:
```suggestion
    it('should add no-heading class to the h3 element', () => {
```

test/blocks/text/text.test.js
```diff
@@ -31,4 +31,33 @@ describe('text block', () => {
...
+      const element = document.querySelector('.link-farm');
+      const foregroundDiv = element.querySelectorAll('.foreground')[1];
+      const headingElem = foregroundDiv.querySelector('#no-heading');
```
ryanmparrish, line 50:
```suggestion
      const headingElem = foregroundDiv.querySelector('.no-heading');
```

ryanmparrish (changes requested):
Looks good. I think we should use a class selector however instead of an ID so we don't get duplicate ID's on different elements in the body.

test/blocks/text/text.test.js
```diff
@@ -31,4 +31,33 @@ describe('text block', () => {
...
+  describe('Link Farm', () => {
+    it('should check if the "link-farm" class is present', () => {
+      const element = document.querySelector('.link-farm');
+      expect(element).to.exist;
+    });
+  
+    it('should check if the "h3" elements are added when necessary', () => {
+      const element = document.querySelector('.link-farm');
+      const foregroundDiv = element.querySelectorAll('.foreground')[1];
+      const headingElem = foregroundDiv.querySelectorAll('h3');
```
JasonHowellSlavin, line 44:
```suggestion
      const headingElem = document.querySelectorAll('.link-farm .foreground h3');
```

test/blocks/text/text.test.js
```diff
@@ -31,4 +31,33 @@ describe('text block', () => {
...
     });
   });
+
+  describe('Link Farm', () => {
+    it('should check if the "link-farm" class is present', () => {
+      const element = document.querySelector('.link-farm');
+      expect(element).to.exist;
+    });
+  
+    it('should check if the "h3" elements are added when necessary', () => {
```
JasonHowellSlavin, line 41: nit `adds h3 elements when necessary`

test/blocks/text/text.test.js
JasonHowellSlavin: Could likely also make this selector one line `const headingElem = document.querySelectorAll('.link-farm .foreground .no-heading')`

test/blocks/text/text.test.js
```diff
@@ -31,4 +31,33 @@ describe('text block', () => {
...
+      const headingElem = foregroundDiv.querySelectorAll('h3');
+      expect(headingElem.length).to.equal(4);
+    });
+    it('should add no-heading id to the h3 element', () => {
+      const element = document.querySelector('.link-farm');
+      const foregroundDiv = element.querySelectorAll('.foreground')[1];
+      const headingElem = foregroundDiv.querySelector('#no-heading');
+      expect(headingElem).to.exist;
+    });
+    it('should add h3 as the first element in the div', () => {
```
JasonHowellSlavin, line 53: nit `adds h3 as the first element in the div`

test/blocks/text/text.test.js
```diff
@@ -31,4 +31,33 @@ describe('text block', () => {
...
+      const element = document.querySelector('.link-farm');
+      const foregroundDiv = element.querySelectorAll('.foreground')[1];
+      const divElements = foregroundDiv.querySelectorAll('div');
```
JasonHowellSlavin, line 56: could combine the selectors here as well

JasonHowellSlavin (commented):
Lot of great stuff. Being a bit nitpicky in the tests, but good work!

test/blocks/text/text.test.js
drashti1712: We need only the second foreground div here, so made changes accordingly.

test/blocks/text/text.test.js
drashti1712: Done, thanks!

test/blocks/text/text.test.js
drashti1712: Done, thanks!

drashti1712:
> Lot of great stuff. Being a bit nitpicky in the tests, but good work!

Modified the code accordingly. Thank you for these insights!

drashti1712:
> Looks good. I think we should use a class selector however instead of an ID so we don't get duplicate ID's on different elements in the body.

Yes sure, I modified the code accordingly. Thank you for these inputs.

ryanmparrish (approved):

spadmasa:
Verified after the pr changes able to see the heading aligned
https://mwpw-132409--milo--drashti1712.hlx.page/drafts/Drashti/document1# 
when heading not present in all columns
image
when heading is present in all columns
image
When there is no headings
image

PR #870, opened 2023-06-14
MWPW-129660 Override Marketo Form Params

Brandon32:
* Allow setting or overriding specific from configurations.
* Decorating destination URL for form success redirect.

Resolves: [MWPW-129660](https://jira.corp.adobe.com/browse/MWPW-129660)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/bmarshal/mcz/marketo-block?martech=off
- After: https://bmarshal-marketo-overrides--milo--adobecom.hlx.page/drafts/bmarshal/mcz/marketo-block?martech=off

**BACOM URLs:**
- Before: https://main--bacom--adobecom.hlx.live/uk/resources/reports/state-of-marketing-automation-b?martech=off
- After: https://main--bacom--adobecom.hlx.live/uk/resources/reports/state-of-marketing-automation-b?milolibs=bmarshal-marketo-overrides&martech=off

### DIFF
--- a/libs/blocks/marketo/marketo.js
+++ b/libs/blocks/marketo/marketo.js
@@ -19,10 +19,30 @@ const FORM_ID = 'form id';
 const BASE_URL = 'marketo host';
 const MUNCHKIN_ID = 'marketo munckin';
 
+const FORM_MAP = {
+  'destination url': 'form.success.content',
+  'co-partner names': 'program.copartnernames',
+  'sfdc campaign id': 'program.campaignids.sfdc',
+};
+
 export const formValidate = (form) => {
   const formEl = form.getFormElem().get(0);
   formEl.classList.remove('hide-errors');
   formEl.classList.add('show-warnings');
+};
+
+export const decorateURL = (destination, baseURL = window.location) => {
+  let destinationUrl = new URL(destination, baseURL.origin);
+
+  if (destinationUrl.hostname.includes('.hlx.')) {
+    destinationUrl = new URL(`${destinationUrl.pathname}${destinationUrl.search}${destinationUrl.hash}`, baseURL.origin);
+  }
+
+  if (baseURL.pathname.endsWith('.html') && !destinationUrl.pathname.endsWith('.html')) {
+    destinationUrl.pathname = `${destinationUrl.pathname}.html`;
+  }
+
+  return destinationUrl;
 };
 
 export const formSuccess = (form) => {
@@ -72,9 +92,7 @@ const setPreference = (key, value) => {
 
 export const setPreferences = (formData) => {
   window.mcz_marketoForm_pref = window.mcz_marketoForm_pref || {};
-  Object.entries(formData).forEach(([key, value]) => {
-    setPreference(key, value);
-  });
+  Object.entries(formData).forEach(([key, value]) => setPreference(key, value));
 };
 
 const init = (el, loadScriptFunc = loadScript) => {
@@ -91,19 +109,32 @@ const init = (el, loadScriptFunc = loadScript) => {
   children.forEach((element) => {
     const key = element.children[0]?.textContent.toLowerCase();
     const value = element.children[1]?.textContent;
-    if (key && value) { formData[key] = value; }
+    if (key && value) {
+      if (key in FORM_MAP) {
+        formData[FORM_MAP[key]] = value;
+      } else if (key && value) {
+        formData[key] = value;
+      }
+    }
   });
 
   const formID = formData[FORM_ID];
   const baseURL = formData[BASE_URL];
   const munchkinID = formData[MUNCHKIN_ID];
 
-  setPreferences(formData);
-
   if (!formID || !baseURL || !munchkinID) {
     el.style.display = 'none';
     return;
   }
+
+  if (formData['form.success.content']) {
+    const destinationUrl = decorateURL(formData['form.success.content']);
+
+    formData['form.success.type'] = 'redirect';
+    formData['form.success.content'] = destinationUrl.href;
+  }
+
+  setPreferences(formData);
 
   const fragment = new DocumentFragment();
   const formWrapper = createTag('section', { class: 'marketo-form-wrapper' });
--- a/libs/blocks/marketo/marketo.js
+++ b/libs/blocks/marketo/marketo.js
@@ -18,11 +18,10 @@ import { parseEncodedConfig, loadScript, createTag } from '../../utils/utils.js'
 const FORM_ID = 'form id';
 const BASE_URL = 'marketo host';
 const MUNCHKIN_ID = 'marketo munckin';
-
 const FORM_MAP = {
-  'destination url': 'form.success.content',
-  'co-partner names': 'program.copartnernames',
-  'sfdc campaign id': 'program.campaignids.sfdc',
+  'destination-url': 'form.success.content',
+  'co-partner-names': 'program.copartnernames',
+  'sfdc-campaign-id': 'program.campaignids.sfdc',
 };
 
 export const formValidate = (form) => {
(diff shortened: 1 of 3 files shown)


libs/blocks/marketo/marketo.js
```diff
@@ -19,12 +19,32 @@ const FORM_ID = 'form id';
 const BASE_URL = 'marketo host';
 const MUNCHKIN_ID = 'marketo munckin';
 
+const FORM_MAP = {
```
meganthecoder, line 22: Unnecessary extra line
```suggestion
const MUNCHKIN_ID = 'marketo munckin';
const FORM_MAP = {
```

libs/blocks/marketo/marketo.js
```diff
@@ -91,20 +109,33 @@ const init = (el, loadScriptFunc = loadScript) => {
   children.forEach((element) => {
     const key = element.children[0]?.textContent.toLowerCase();
     const value = element.children[1]?.textContent;
-    if (key && value) { formData[key] = value; }
+    if (key && value) {
+      if (key in FORM_MAP) {
+        formData[FORM_MAP[key]] = value;
+      } else if (key && value) {
```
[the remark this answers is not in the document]
Brandon32: Yes that is redundant, thank you.

libs/blocks/marketo/marketo.js
```diff
@@ -18,13 +18,32 @@ import { parseEncodedConfig, loadScript, createTag } from '../../utils/utils.js'
 const FORM_ID = 'form id';
 const BASE_URL = 'marketo host';
 const MUNCHKIN_ID = 'marketo munckin';
+const FORM_MAP = {
+  'destination url': 'form.success.content',
+  'co-partner names': 'program.copartnernames',
+  'sfdc campaign id': 'program.campaignids.sfdc',
```
mokimo, line 24: These look like very odd strings, but I guess that's why this map exist

mokimo (approved):

libs/blocks/marketo/marketo.js
Brandon32: I have created an authoring guide https://milo.adobe.com/docs/authoring/marketo.
We are converting author friendly names to values that are added to `window.mcz_marketoForm_pref`

mokimo: Since those are specific strings and the spaces etc. need to be set, maybe `destination-url` `co-partner-names` and `sfdc-campaign-id` ? 

Not blocking the PR, but it might be more obvious it's something "pasteable" to authors then, but I'll leave it up to you

Brandon32: @mokimo That's a good idea, I'll string replace spaces with dashes so we're more flexible with authoring.

test/blocks/marketo/marketo.test.js
```diff
@@ -95,4 +99,26 @@ describe('marketo', () => {
     formSuccess(window.MktoForms2);
     expect(window.mktoSubmitted).to.be.true;
   });
+
+  it('decorate destination urls', () => {
```
JasonHowellSlavin, line 103: super nit `decorates`

JasonHowellSlavin (approved):

honstar:
The PR looks good to me 👍 The only thing is the LH score that worries me a bit. @Brandon32, any idea what is causing this?

Brandon32:
@honstar We are loading the external Marketo forms js and this is causing the biggest impact. There are still changes happening on the Marketo side so LH should improve but I'll check on that.

honstar:
@overmyheadandbody @mokimo @narcis-radu : do you think there is a chance to take advantage of the delayed functionality that is being employed for the global-navigation? @Brandon32 : would you agree that marketo forms are typically below the fold? That might give us a chance to delay the 3rd party marketo dependency.

mokimo:
@honstar delaying doesn't really help TBT it's about introducing strategic delays to not block the main thread within the 3rd party scripts, so in this case that would be marketo script providing team making improvements.

Here's what I had shared on slack:
> A few web performance topics I’ve been looking at lately that really helped me to understand TBT and how to avoid it:
https://www.youtube.com/watch?v=7Rrv9qFMWNM
https://web.dev/long-tasks-devtools/
https://web.dev/optimize-long-tasks/
https://philipwalton.com/articles/idle-until-urgent/
https://developer.chrome.com/blog/using-requestidlecallback/
TBT accounts for 30% of the total LH score https://googlechrome.github.io/lighthouse/scorecalc/ - so it really makes sense to get a deep understanding of it
e: Applying the knowledge from above, we broke up a few long running tasks causing TBT in this PR for the global-navigation and it includes a two case studies of the results: https://github.com/adobecom/milo/pull/675 - might be interesting to skim as it is hands on and applied to adobe’ code

Brandon32:
@honstar I would like to give them time to adjust TBT from the Marketo side. I am open to using an intersection observer like we did with charts, most of the time forms are below the fold, but I would want to do that in a separate PR.

libs/blocks/marketo/marketo.js
```diff
@@ -18,13 +18,32 @@ import { parseEncodedConfig, loadScript, createTag } from '../../utils/utils.js'
...
   const formEl = form.getFormElem().get(0);
   formEl.classList.remove('hide-errors');
   formEl.classList.add('show-warnings');
 };
 
+export const decorateURL = (destination, baseURL = window.location) => {
+  let destinationUrl = new URL(destination, baseURL.origin);
+
+  if (destinationUrl.hostname.includes('.hlx.')) {
+    destinationUrl = new URL(`${destinationUrl.pathname}${destinationUrl.search}${destinationUrl.hash}`, baseURL.origin);
```
chrischrischris, line 37: nit: use destructuring (also shortens line that is > 100)
```suggestion
    const { pathname, search, hash } = destinationUrl;
    destinationUrl = new URL(`${pathname}${search}${hash}`, baseURL.origin);
```

chrischrischris: Looking at code below you prob want to destructure outside the if

libs/blocks/marketo/marketo.js
```diff
@@ -72,39 +91,54 @@ const setPreference = (key, value) => {
...
-    const value = element.children[1]?.textContent;
-    if (key && value) { formData[key] = value; }
+    const key = element.children[0]?.textContent.trim().toLowerCase().replaceAll(' ', '-');
+    const value = element.children[1]?.href ?? element.children[1]?.textContent;
+    if (key && value) {
+      if (key in FORM_MAP) {
+        formData[FORM_MAP[key]] = value;
+      } else {
+        formData[key] = value;
+      }
```
chrischrischris, line 120: nit
```suggestion
    if (!key || !value) return;
    if (key in FORM_MAP) {
      formData[FORM_MAP[key]] = value;
    } else {
      formData[key] = value;
    }
```

libs/blocks/marketo/marketo.js
```diff
@@ -18,13 +18,32 @@ import { parseEncodedConfig, loadScript, createTag } from '../../utils/utils.js'
...
+};
 
 export const formValidate = (form) => {
   const formEl = form.getFormElem().get(0);
   formEl.classList.remove('hide-errors');
   formEl.classList.add('show-warnings');
 };
 
+export const decorateURL = (destination, baseURL = window.location) => {
+  let destinationUrl = new URL(destination, baseURL.origin);
```
chrischrischris, line 34: Is destination guaranteed to be a valid url?  If it isn't this will throw

libs/blocks/marketo/marketo.js
Brandon32: Sounds good, thanks!

libs/blocks/marketo/marketo.js
Brandon32: Thanks!

libs/blocks/marketo/marketo.js
Brandon32: Good call, refactored this to a try/catch and made sure we have a hostname, added a test case to validate.

mokimo:
@Brandon32 if it's below the fold then an intersection observer makes a lot of sense 👍

PR #874, opened 2023-06-15
Personalization

chrischrischris:
Adds personalization to Milo.

Though this is a new feature, I am only announcing it to Grp-Milo-Dev in the small chance that this affects a consumer.  It should not and has been tested against bacom and cc.

The general announcement of Personalization will be released after the break, but we are merging now to allow for a limited group of authors to start testing it.

If you see any issues or have questions please contact me.

Resolves: [MWPW-124495](https://jira.corp.adobe.com/browse/MWPW-124495)

**Test URLs:**
- After: https://personalization--milo--adobecom.hlx.page/?martech=off


libs/utils/utils.js
```diff
@@ -366,16 +374,21 @@ export async function loadTemplate() {
...
 export async function loadBlock(block) {
   const name = block.classList[0];
-  const { miloLibs, codeRoot } = getConfig();
+  const { miloLibs, codeRoot, p13nBlocks } = getConfig();
+
   const base = miloLibs && MILO_BLOCKS.includes(name) ? miloLibs : codeRoot;
+  let blockPath = `${base}/blocks/${name}/${name}`;
+  if (p13nBlocks?.[name]) {
+    blockPath = `${p13nBlocks[name]}/${name}`;
+  }
```
auniverseaway, line 383: I would probably prefer expBlocks over `p13n`. As in, experiments blocks...
```suggestion
  if (expBlocks?.[name]) blockPath = `${expBlocks[name]}/${name}`;
```

libs/utils/utils.js
```diff
@@ -593,11 +606,76 @@ function decorateFooterPromo(config) {
...
+      createTag, getConfig, getMetadata, loadLink, loadScript, updateConfig,
+    },
+  });
+
+  return true;
+}
+
+async function checkForPageMods() {
+  const persMd = getMetadata('personalization');
+  let persManifests = [];
```
auniverseaway, line 654: Do we have an early return opportunity here?

libs/utils/utils.js
chrischrischris: I've had trouble deciding between experiments and personalization.  Ultimately when doing self-serve manifests that's not an experiment, so decided to go with personalization.  I could rename to `persBlocks` or `personalizationBlocks`

auniverseaway, line 383:
```suggestion
const path = expBlocks?.[name] ? `${expBlocks[name]}` : `${base}/blocks/${name}`;
const blockPath = `${path}/${name}`;
```
Does this work?

auniverseaway (approved):
Looks great. Mostly only focused on utils.js..

I get there's things in here we'd rather not do, but really appreciate the care you took to reduce the blast radius of the changes.

auniverseaway:
@chrischrischris pznBlocks? Something about the number still rubs me the wrong way. 😂

rgclayton (approved):
Should I be getting different content in the red outline area on Firefox?
image

libs/scripts/personalization.js
```diff
@@ -0,0 +1,406 @@
...
+
+export const PERSONALIZATION_TAGS = {
+  chrome: () => navigator.userAgent.includes('Chrome') && !navigator.userAgent.includes('Mobile'),
+  firefox: () => navigator.userAgent.includes('Firefox') && !navigator.userAgent.includes('Mobile'),
+  android: () => navigator.userAgent.includes('Android'),
+  ios: () => /iPad|iPhone|iPod/.test(navigator.userAgent),
+  loggedout: () => !window.adobeIMS?.isSignedInUser(),
+  loggedin: () => window.adobeIMS?.isSignedInUser(),
+  darkmode: () => window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches,
+  lightmode: () => !PERSONALIZATION_TAGS.darkmode(),
```
narcis-radu, line 15: wouldn't this throw a lint error - `PERSONALIZATION_TAGS was used before it was defined`?!

libs/scripts/personalization.js
```diff
@@ -0,0 +1,406 @@
...
+  },
+};
+
+const VALID_COMMANDS = Object.keys(COMMANDS);
+
+const GLOBAL_CMDS = [
+  'insertscript',
+  'replacefragment',
+  'replacepage',
+  'updatemetadata',
```
narcis-radu, line 53: maybe we can use camel case

libs/scripts/preview/preview.css
```diff
@@ -0,0 +1,134 @@
...
+    bottom: 64px;
+    right: 0;
+    background-color: #444;
+    min-width: 300px;
+    border-radius: 16px;
+    box-shadow: 0 0 10px #000;
+    font-size: 12px;
+}
+
+.hlx-popup a:any-link {
```
narcis-radu, line 62: Maybe `any-link` is not needed

libs/scripts/preview/preview.js
```diff
@@ -0,0 +1,296 @@
...
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+/* eslint-disable no-console */
+
+import {
+  toClassName,
+  getMetadata,
+  getConfigForFullExperiment,
+  getConfigForInstantExperiment,
```
narcis-radu, line 18: I was not able to find this method in `personalization.js`

libs/scripts/preview/preview.js
```diff
@@ -0,0 +1,296 @@
...
+  // checkTesting,
+} from '../personalization.js';
+
+const checkTesting = () => true;
+const getExperimentConfig = (experiment, instantExperiment) => {
+  if (instantExperiment) {
+    return getConfigForInstantExperiment(experiment, instantExperiment);
+  }
+  return getConfigForFullExperiment(experiment);
```
narcis-radu, line 27: not sure how `getConfigForFullExperiment` works, but we should be able to do `getConfigForFullExperiment(experiment, instantExperiment)` and check whether `instantExperiment` is defined inside the method

libs/utils/utils.js
chrischrischris: Can't early return as we also need to check target (line 666)

chrischrischris:
@rgclayton The next block is the firefox personalized one.  The latest code is at https://personalization--milo--adobecom.hlx.page/drafts/cpeyer/pers/pers-demo.

libs/scripts/personalization.js
chrischrischris: everything is lowercased so that user capitalization doesn't matter

libs/scripts/preview/preview.css
chrischrischris: preview is non-functional and not used right now and is going to be worked on after this merge.

libs/scripts/personalization.js
chrischrischris: The alternative is:
`lightmode: function () { return !this.darkmode(); },`

but I prefer the current one.  Thoughts?

chrischrischris:
@narcis-radu fyi the preview files have been removed.

chrischrischris:
fyi - removed the "new feature" tag as Rob Woblesky is going to do a formal detailed release email.  Still sending this out to the consumers via high-impact just in case it does change some behavior on a consumer that we didn't catch.

yesil:
@chrischrischris  I see the following error at https://personalization--milo--adobecom.hlx.page/drafts/cpeyer/pers/pers-demo

image

PR #876, opened 2023-06-20
MWPW-132671: Append region (country) display name to document title

hparra:
Resolves: [MWPW-132671](https://jira.corp.adobe.com/browse/MWPW-132671)

**Description**

Appends the geo's region display name (RDN) to `document.title`. This ensures that titles across Adobe sites are unique, which is critical for SEO. This feature already existed in Dexter, so this PR is a fix for a migration regression. See issue for details.

The RDN is the name of the geo's region/country in the geo's language. Examples:

- be_en => Belgium
- be_fr => Belgique

Code checks if RDN is already there before attempting to append again. The `locales` SSOT is modified to add explicit overrides for RFC5646-incompatible geos (e.g. africa, mena_en) using the `rdn` key, and fix ambiguous `ietf` values that are incorrect mappings of the geo to RFC5646 tag using `tag` values. I did not "fix" the offending `ietf` values because I am unsure what these values are used for.

Each Cloud (e.g. DC, CC, BACOM) MUST do one of two things:

1. copy changes to `locales` into their own repos, or
2. modified their utils.js to load locales from Milo: https://github.com/adobecom/milo/pull/880

**Test URLs:**

- Before: https://main--milo--hparra.hlx.page/
- Before: https://main--dc--adobecom.hlx.live/ar/acrobat/online/word-to-pdf?martech=off
- After: https://seo-title-region--milo--hparra.hlx.page/
- After: https://main--dc--adobecom.hlx.live/ar/acrobat/online/word-to-pdf?milolibs=seo-title-region--milo--hparra&martech=off


mokimo:
There should be no PRs in draft: https://github.com/adobecom/milo/wiki/Submitting-PRs#drafts-prs

libs/scripts/scripts.js
```diff
@@ -32,7 +32,7 @@ const locales = {
   pe: { ietf: 'es-PE', tk: 'oln4yqj.css' },
```
hparra, line 32: Line 30 `la: { ietf: 'es-LA', tk: 'oln4yqj.css' },` is completely wrong. The `la` geo is for Latin America, while the `LA` subtag is for Laos. There may be other preexisting mistakes like this.

libs/scripts/scripts.js
```diff
@@ -27,12 +27,12 @@ const locales = {
   ca_fr: { ietf: 'fr-CA', tk: 'vrk5vyv.css' },
   cl: { ietf: 'es-CL', tk: 'oln4yqj.css' },
   co: { ietf: 'es-CO', tk: 'oln4yqj.css' },
-  la: { ietf: 'es-LA', tk: 'oln4yqj.css' },
```
hparra, line 30: This is completely wrong. The la geo is for Latin America, while the LA subtag is for Laos. There may be other preexisting mistakes like this.

auniverseaway (changes requested):
Blocking for now. I see several issues. WIll address when Milo Academy is over.

libs/scripts/scripts.js
```diff
@@ -137,6 +138,8 @@ const config = {
   privacyId: '7a5eb705-95ed-4cc4-a11d-0cc5760e93db', // valid for *.adobe.com
 };
 
+export default config;
```
hparra, line 141: TODO: Delete

hparra:
@auniverseaway Added config feature gate and full test coverage. All checks pass except enforce PR labels. Prefer to land https://github.com/adobecom/milo/pull/880 first and then refactor this to use those changes.

auniverseaway:
1. The existing IETFs in Milo are wrong and should be updated to your tag properties. This should make this new "tag" property no longer needed. We will have to check the other consumers, but my guess would be they are more accurate since it's production content.
2. I would opt for a human readable property for the name instead of `rdn`. My first thought is that `name` seems to make the most sense since we need this for both language and region (aka locale). Another option would be to make it authorable in metadata.
3. We don't want milo opted into this in its config. This whole feature _as a default_ was previously shot down at a VP level several years ago. You have the right idea, though: config > metadata.xlsx > page metadata is usually how we like to handle overriding w/ granularity. I personally don't see any site implementing this site-wide, but I'm also not opposed to having it supported in config.
4. Given that the majority of content will **_not_** use this feature, it shouldn't take so much precedence in the call stack... At a high level: it shouldn't happen before LCP and the code shouldn't live in utils. I would gate this into a separate file similar to our other opt-in features (georouting, rich results, etc.). My general rule is, "is this needed to get us to LCP?" if the answer is no, it should fire after LCP. Things like decorating links run the risk of being interacted with so we leave those CPU cycles in front of LCP.

hparra:
> 1. The existing IETFs in Milo are wrong and should be updated to your tag properties. This should make this new "tag" property no longer needed. We will have to check the other consumers, but my guess would be they are more accurate since it's production content.
> 2. I would opt for a human readable property for the name instead of `rdn`. My first thought is that `name` seems to make the most sense since we need this for both language and region (aka locale). Another option would be to make it authorable in metadata.

See [Changes to IETF Values in Milo](https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=seoteam&title=Changes+to+IETF+Values+in+Milo). I'm not going to change the IETF values. There is a lot of pushback here even when the IETF value is wrong, e.g. https://github.com/adobecom/bacom/pull/123, https://github.com/adobecom/cc/pull/72, https://github.com/adobecom/dc/pull/270. Instead, I prefer to use the `locales` object I updated in https://github.com/adobecom/milo/pull/915. 

> 3. We don't want milo opted into this in its config. This whole feature _as a default_ was previously shot down at a VP level several years ago. You have the right idea, though: config > metadata.xlsx > page metadata is usually how we like to handle overriding w/ granularity. I personally don't see any site implementing this site-wide, but I'm also not opposed to having it supported in config.

Not sure this is relevant. Each cloud defines it's own config, so the fact that Milo opts-in only affects the actual Milo website, and not the libs' consumers. I will have to make the `locales` change for each cloud. That's why https://github.com/adobecom/milo/pull/880 was proposed.

> 4. Given that the majority of content will **_not_** use this feature, it shouldn't take so much precedence in the call stack... At a high level: it shouldn't happen before LCP and the code shouldn't live in utils. I would gate this into a separate file similar to our other opt-in features (georouting, rich results, etc.). My general rule is, "is this needed to get us to LCP?" if the answer is no, it should fire after LCP. Things like decorating links run the risk of being interacted with so we leave those CPU cycles in front of LCP.

OK. We're changing the `<title>`, so the concern here what that this will fire too late for googlebot to pick up change.

hparra:
Use-case solved with #955 ✌️

PR #880, opened 2023-06-21
MWPW-132671: Load locales from their own file

hparra:
Resolves: [MWPW-133060](https://jira.corp.adobe.com/browse/MWPW-133060).

👀  **Update:** `locales.js` is now `adobecom.json` . No changes in performance reported below. 

Move locales to their own file. This enables locales to be loaded by the other clouds from here, instead of using copies they have added to their own repos. This creates a single source of truth (SSOT) of geo to RFC5646 (locale, hreflang, etc) data to be maintained in Milo.

**Test URLs:**
- After: https://import-locales--milo--hparra.hlx.page/?martech=off

**Performance**

There are always performance concerns regarding any `import` so early in the Milo call stack.
In the following example, we trace an initial (no cache) load of test URL above throttled to mimic a fast 3G connection.

Screen Shot 2023-06-20 at 17 18 19

KPIs:

- DOMContentLoaded fired after 1.85 seconds
- Loaded fired after 6s
- Total javascript loaded is 79.6 kb
- Total content loaded is 385 kb

The import of `locales.js` (locales) loads 1.1 kB in 590ms. The size accounts for ~1% of our javascript and even less so for overall content.

The imports `utils.js` and `locales.js` happen concurrently, with utils taking 655ms to load, so there is only a negligible increase in time due to additional bandwidth usage.

An argument may be made about DNS Lookup costs once a cloud loads this from Milo. If the cloud is already loading utils from Milo, then the DNS Lookup is already being paid. We just need to re-enable import-style concurrency. For example:

```js
  // updated example for adobecom.json
  const [
    adobecomConfig,
    { loadArea, setConfig, loadLana },
  ] = await Promise.all([
    fetch('${miloLibs}/configs/adobecom.json`).then((r) => r.json()),
    import(`${miloLibs}/utils/utils.js`),
  ]);
  setConfig({ ...adobecomConfig, ...config });
```

Performance is similar with equivalent concurrency:

Screen Shot 2023-06-20 at 18 46 39

There is minimal performance cost to gain a "hot" SSOT in Milo. Changes to this SSOT will immediately benefit all clouds and reduce the engineering toil of copy/pasting a configuration for each cloud each time a change is made.


auniverseaway (changes requested):
I like the idea of moving the config object out scripts from a cleanliness perspective, but the fundamental premise is a bit off: "We need a source of truth for Adobe.com geos and how the map to hreflang." we don't need this, and I would probably say we don't want this either when taken into real world scenarios.

**_Every_** project goes live with different regions / locales / languages at different times. We don't want Milo be the source of truth for this as it would require everyone to go live at the same time... which we do not want. Even with the recent geo-expansion, business.adobe.com is not live with the new geos. Frictionless is another example: they went live with US, then Tier 1, then ROW... completely staggered and free to support the regions they want, _when_ they want.

This doesn't even get into the complexities that some projects are **language-based**, and some projects are **region-based**. This PR seems to propose that everyone uses locales.json from milo which I would never encourage. Then we would have maybe www using milo's locales because they agree with our list, and business.adobe.com and blog.adobe.com using a completely different locale list because they all have different locale needs. Now we have some projects doing things one way and other projects doing things another way... which, again, we do not want.

The current method is, "Every project lists the locales they want to support."
This PR seem to propose, "Some projects will use Milo's locale list, some projects use their own."

The projects that would use Milo's locale list would be in the overwhelming minority, and that's not a good enough reason to ship this.

The whole premise of Milo (and Gnav, Dexter) with the consumer model is that there's a shared set of functions and a de-coupled set of configurations for every project. I would not argue that having individual configs _can_ create duplicate work, but the alternative is not desired and creates complexity we do not want to maintain. Sites should be independently in charge of the locales they want to go live with, even if it's similar (or identical) to their siblings.

Milo is not just www (DC & CC). Milo is:

clio
HelpX
HelpX Internal
Stock
business
business blog
blog
etc.

hparra:
@auniverseaway This doesn't block https://github.com/adobecom/milo/pull/876. Please submit your feedback on that PR as soon as you can.

> I like the idea of moving the config object out scripts from a cleanliness perspective

This is about divergence, software regressions associated with it, and the toil needed to fix them.
> fundamental premise is a bit off: "We need a source of truth for Adobe.com geos and how the map to hreflang." we don't need this, and I would probably say we don't want this either when taken into real world scenarios.

*We do need it* for actual real world scenarios affecting production right now.

> **_Every_** project goes live with different regions / locales / languages at different times.

Not relevant. This mapping is a source of truth, not a set of feature flags. It should not suggest particular support, or be tied to migration or launches. I haven't seen anything in Milo that suggests it does this or that it has been used in this manner before.

> We don't want Milo be the source of truth for this as it would require everyone to go live at the same time...

> This doesn't even get into the complexities that some projects are **language-based**, and some projects are **region-based**.

Only Express is language based. ACOM (DC, CC, FedPub), BACOM, and HelpX all use the same geo-based directory structure expressed by this version of locales.

We are giving each cloud the *option* to use a common config that lives in Milo, because it's already there, and these clouds are already hot-loading files from this repo anyway. I can always put it somewhere else, but the problem remains that the core code and blocks that use `locales` live in Milo. That's why `locales` here there to begin with, because this is where everyone is developing these features. The Milo website itself doesn't need these locales at all!

> The projects that would use Milo's locale list would be in the overwhelming minority, and that's not a good enough reason to ship this.

ACOM (DC, CC, FedPub) + BACOM + HelpX compose the majority of Adobe.com marketing. Are you saying the primary Milo customers should find somewhere else to put Milo-specific data that they share between them?

auniverseaway:
@hparra I think we touched on this in our call yesterday, but I still do not want to ship this in Milo. I do not want to encourage anyone to think Milo's locales or configs are best practice and should be used. We get a lot of things wrong and projects should not be misled into thinking Milo is right. They should be able to go-live and do their own testing for their own uses... including their own Typekit decisions. More importantly, I think this sets us up for hard dependencies between milo and consumers we don't want.

To your point, yes you can say, "you don't have to use this." but I don't think that's the right approach given how many sites would actually use it. It adds complexity when there are really only two projects that need this centralized.

> Only Express is language based.

This is not correct. blog.adobe.com (the biggest Milo consumer by far) is language based... with a weird small blend of region just for the UK. stock.adobe.com/pages/artisthub is language based.

> We do need it for actual real world scenarios affecting production right now.

I disagree with this. If it was needed, sites wouldn't be live without it. Is it cumbersome for the person that has to change 4 configs? Yeah, it is. Is it better than the hell we had in 2016 where a single config could not go live until every single sub-project did a full regression on their own content? Yes, I think it is. 

We're already starting to creep into that territory now when we caused 2 different CSOs on DC because we didn't think a change affected them.

PR #881, opened 2023-06-21
MWPW-132693: Icon block inline variant style fix

zagi25:
Changes made to inline icon-block:

- Changed icon size (32 -> 40, 64 -> 56)
-  Fixed icons elongation
-  Changed Heading font size from S to XS
-  Removed width from bio inline blocks that are part of xx-up, so it can fill width of the column
- Added spacing between CTAs in action-area
- Changed font size of link in action-area to XS

After PR is approved, kitchen sink and sidekick lib should be updated (I can update the kitchen sink, but I don't have permission to edit library/block directory)

Resolves: [MWPW-132693](https://jira.corp.adobe.com/browse/MWPW-132693)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ratko/icon-block-inline?martech=off
- After: https://icon-block-inline--milo--zagi25.hlx.page/drafts/ratko/icon-block-inline?martech=off


libs/blocks/icon-block/icon-block.js
```diff
@@ -39,8 +39,18 @@ const iconBlocks = {
   },
 };
 
+//checking if block is part of xx-up and is inline variant
+function upAndInline(el) {
+  const isInline = el.classList.contains('inline');
+  if(!isInline) return false;
+  const sectionMetadata = el.parentElement?.querySelector('.section-metadata');
+  if(!sectionMetadata) return false;
```
JasonHowellSlavin, line 47: nit, could make a bit more concise.  Maybe something like? 

```suggestion
  const isInlineAndSectionMetadata = el.classList.contains('inline') && el.parentElement?.querySelector('.section-metadata')
  if(!isInlineAndSectionMetadata) return false;
```

libs/blocks/icon-block/icon-block.js
```diff
@@ -57,7 +67,7 @@ function decorateContent(el) {
     }
     const size = getBlockSize(el, 2);
     const variant = [...variants].filter((v) => el.classList.contains(v))?.[0] ?? 'fullwidth';
-    decorateBlockText(el, iconBlocks[size][variant]);
+    decorateBlockText(el, isUpAndInline ? ['xs', 's'] : iconBlocks[size][variant]);
```
JasonHowellSlavin, line 70: Can we instantiate this ternary somewhere else? I don't think it being a param is particularly clear.

JasonHowellSlavin (commented):
Good stuff. Could you add some tests for your changes, maybe also try this with `two-up` etc. on your authored pages?

libs/blocks/icon-block/icon-block.css
```diff
@@ -103,14 +103,16 @@
...
 }
 
-.icon-block.inline.icon-height-32 .icon-area,
-.icon-block.inline.icon-height-32 .icon-area picture img {
-  height: 32px;
+.icon-block.inline.icon-height-40 .icon-area,
+.icon-block.inline.icon-height-40 .icon-area picture img {
+  min-width:40px;
+  height: 40px;
 }
```
rgclayton, line 110: This change will require re-authoring of any page that uses the `icon-height-32` variant. Has this been communicated?

libs/blocks/icon-block/icon-block.css
```diff
@@ -103,14 +103,16 @@
...
 
-.icon-block.inline.icon-height-64 .icon-area,
-.icon-block.inline.icon-height-64 .icon-area picture img {
-  height: 64px;
+.icon-block.inline.icon-height-56 .icon-area,
+.icon-block.inline.icon-height-56 .icon-area picture img {
+  min-width: 56px;
+  height: 56px;
```
rgclayton, line 115: Same with this. Any content that is using `icon-height-64` will need re-authored with this change.

libs/blocks/icon-block/icon-block.css
```diff
@@ -201,16 +207,16 @@
...
 
-.icon-block.inline.icon-height-32.bio .icon-area,
-.icon-block.inline.icon-height-32.bio .icon-area picture img {
-  height: 32px;
-  width: 32px;
+.icon-block.inline.icon-height-40.bio .icon-area,
+.icon-block.inline.icon-height-40.bio .icon-area picture img {
+  height: 40px;
+  width: 40px;
 }
```
rgclayton, line 214: Same here.

libs/blocks/icon-block/icon-block.css
```diff
@@ -201,16 +207,16 @@
...
 
-.icon-block.inline.icon-height-64.bio .icon-area,
-.icon-block.inline.icon-height-64.bio .icon-area picture img {
-  height: 64px;
-  width: 64px;
+.icon-block.inline.icon-height-56.bio .icon-area,
+.icon-block.inline.icon-height-56.bio .icon-area picture img {
+  height: 56px;
+  width: 56px;
 }
```
rgclayton, line 220: Same here.

rgclayton (changes requested):
Icons for `inline` variants with out specific `icon-heights` are still being smooshed.

image

libs/blocks/icon-block/icon-block.css
ryanmparrish, line 1: on line 160: 
You should include an additional `.icon-block .foreground .text-content .heading-s` selector since this changed and also needs headline bottom margin.

libs/blocks/icon-block/icon-block.css
```diff
@@ -331,6 +338,14 @@
     max-width: 300px;
   }
 
+  .two-up .icon-block.bio.inline .foreground,
```
ryanmparrish, line 341: you could use a single selector - `[class*=-up] .icon-block.bio.inline .foreground`

libs/blocks/icon-block/icon-block.css
```diff
@@ -172,10 +174,14 @@
   margin-bottom: 0;
 }
 
-.icon-block.vertical:not(.center) .foreground .text-content .action-area > a {
+.icon-block.vertical:not(.center) .foreground .text-content .action-area > a,
+.icon-block.inline:not(.center) .foreground .text-content .action-area > a {
   margin-right: var(--spacing-s);
   white-space: nowrap;
 }
+.icon-block.inline:not(.center) .foreground .text-content .action-area > a:first-child {
```
ryanmparrish, line 182: nit: extra line

libs/blocks/icon-block/icon-block.css
```diff
@@ -172,10 +174,14 @@
   margin-bottom: 0;
 }
 
-.icon-block.vertical:not(.center) .foreground .text-content .action-area > a {
+.icon-block.vertical:not(.center) .foreground .text-content .action-area > a,
+.icon-block.inline:not(.center) .foreground .text-content .action-area > a {
   margin-right: var(--spacing-s);
   white-space: nowrap;
 }
```
ryanmparrish, line 181: nit: extra line for linting

ryanmparrish (changes requested):

libs/blocks/icon-block/icon-block.css
```diff
@@ -172,10 +174,14 @@
...
+.icon-block.inline:not(.center) .foreground .text-content .action-area > a:first-child {
+  margin-bottom: var(--spacing-xxs);
+}
```
ryanmparrish, line 185: Rather than targeting the <a> tags margin(s). 
Can you try adding some styles to the .action-area using flex with a gap? 
```
.icon-block .foreground .text-content .action-area {
    display: flex;
    gap: var(--spacing-s);
}
'''
If we were to approach it at the inline block level it would be good to use the 'margin-block' properties for better localization text-orientation support. Just a thought on that.

File: libs/blocks/icon-block/icon-block.js (hunk 1 above)
Comment (zagi25) [reply libs/blocks/icon-block/icon-block.js:47]: Yeah, this looks better. Thank you.

File: libs/blocks/icon-block/icon-block.js (hunk 2 above)
Comment (zagi25) [reply libs/blocks/icon-block/icon-block.js:70]: `const textSize = upAndInline(el) ? ['xs', 's'] : iconBlocks[size][variant];`
`decorateBlockText(el, textSize);`
Something like this ?

File: libs/blocks/icon-block/icon-block.css (hunk 10 above)
Comment (zagi25) [reply libs/blocks/icon-block/icon-block.css:185]:
```
  column-gap: var(--spacing-s);
  row-gap: var(--spacing-xs);
```
I used flex and this gaps instead of margins. Is this what you meant ?

**Comment by zagi25** [comment]
@rgclayton you are right any content using icon-height will need to be re-authored, as you saw icon-height changes were requested on the ticket. I don't think that anyone used this because it was broken ( but of course I am not 100% sure ), maybe we can leave old classes and notify authors to replace them with new ones, and then just remove the old ones ? 

I updated my code based on the comments and requests and also added somethings that I missed for five-up, I just need to write tests and also test it more manually. 

Before I push my changes I have question for you all @JasonHowellSlavin @ryanmparrish @rgclayton is there a need for this rule  https://github.com/adobecom/milo/blob/main/libs/blocks/section-metadata/section-metadata.css#L148 ? 

Screenshot 2023-06-22 at 15 30 22

As you can see, currently this rule is breaking the five-up inline blocks (and this is for 40px icon) ? Is there a way to fix this that I am not aware of, without erasing this line ?

**Comment by zagi25** [comment]
@rgclayton @ryanmparrish @JasonHowellSlavin PR is updated, please take a look. Thank you.

File: libs/blocks/section-metadata/section-metadata.css [hunk 11]
```diff
@@ -145,7 +145,7 @@ main>.section[class*='-up']>.content {
   padding-right: var(--grid-margins-width);
 }
 
-.section.five-up {
+.section.five-up :not(.icon-block.inline) {
```
ryanmparrish, line 148: We probably shouldn't add block specific selectors in section-metadata.css even in the :not() case. 
Can we make this an override in the icon-block.css rather?

test/blocks/icon-block/icon-block.test.js
```diff
@@ -20,15 +20,27 @@ describe('icon blocks', () => {
...
     });
-    describe('icon block inline has 2 columns', () => {
-      it('has 2 columns', () => {
-        const inlineBlock = document.querySelector('.icon-block.inline');
-        expect(inlineBlock).to.exist;
-        const firstColumn = inlineBlock.querySelector('.text-content .icon-area');
-        const secondColumn = inlineBlock.querySelector('.text-content .second-column');
-        expect(firstColumn).to.exist;
-        expect(secondColumn).to.exist;
+    if(block.classList.contains('inline')){
```
ryanmparrish, line 23:
```suggestion
    if (block.classList.contains('inline')) {
```

Check your linting

ryanmparrish (changes requested):
This is looking good. I'd suggest we move that section-meta added style out and into the icon-block styles and double checking the icon var sizes are correct as the jira ticket was a little hard to follow. Thanks.

zagi25:
@ryanmparrish @rgclayton @JasonHowellSlavin @Blainegunn  PR will be updated when I get sign off on class naming, please check [MWPW-132693](https://jira.corp.adobe.com/browse/MWPW-132693) . Thanks.

libs/blocks/icon-block/icon-block.js
```diff
@@ -39,6 +39,13 @@ const iconBlocks = {
   },
 };
 
+//checking if block is part of xx-up and is inline variant
+function upAndInline(el) {
+  const isInlineAndSectionMetadata = el.classList.contains('inline') && el.parentElement?.querySelector('.section-metadata')
+  if(!isInlineAndSectionMetadata) return false;
```
seanchoi-dev, line 45: nit: our lint standard requires a space between if and (
```suggestion
  if (!isInlineAndSectionMetadata) return false;
```

seanchoi-dev (changes requested):

libs/styles/variables.css
```diff
@@ -104,6 +104,11 @@
   --grid-margins-width-10: calc((100vw - (var(--grid-column-width) * 10)) / 2);
   --grid-margins-width-8: calc((100vw - (var(--grid-column-width) * 8)) / 2);
   --grid-margins-width-6: calc((100vw - (var(--grid-column-width) * 6)) / 2);
+
+  /* icon sizes */
+  --l-icon: 56px;
+  --m-icon: 40px;
+
```
ryanmparrish, line 111: We should follow the same naming convention as other size vars.  elem+attr+unit
`--icon-size-m`

libs/blocks/icon-block/icon-block.css
```diff
@@ -74,6 +73,16 @@
   width: auto;
 }
 
+.icon-block.inline .icon-area,
+.icon-block.inline .icon-area picture img {
+  min-width:80px;
```
ryanmparrish, line 78:
```suggestion
  min-width: var(--xxl-icon);
```

libs/styles/variables.css
```diff
@@ -104,6 +104,11 @@
   --grid-margins-width-10: calc((100vw - (var(--grid-column-width) * 10)) / 2);
   --grid-margins-width-8: calc((100vw - (var(--grid-column-width) * 8)) / 2);
   --grid-margins-width-6: calc((100vw - (var(--grid-column-width) * 6)) / 2);
+
+  /* icon sizes */
+  --l-icon: 56px;
```
ryanmparrish, line 109:
```suggestion
  --xxl-icon: 80px;
  --xl-icon: 64px;
  --l-icon: 56px;
```

libs/styles/variables.css
```diff
@@ -104,6 +104,11 @@
...
+  --m-icon: 40px;
```
ryanmparrish, line 110:
```suggestion
  --m-icon: 40px;
  --s-icon: 32px;
  --xs-icon: 24px;
  --xxs-icon: 16px;
```

libs/blocks/icon-block/icon-block.css
```diff
@@ -74,6 +73,16 @@
...
+}
+
+.icon-block.inline.bio .icon-area,
+.icon-block.inline.bio .icon-area picture img {
+  min-width:60px;
```
ryanmparrish, line 83:
```suggestion
  min-width: var(--xl-icon);
```

libs/blocks/icon-block/icon-block.css
```diff
@@ -103,14 +112,16 @@
   width: 60px;
```
ryanmparrish, line 112:
```suggestion
  width: var(--xl-icon);
```

libs/blocks/icon-block/icon-block.css
ryanmparrish: Technically the xl-icon is 64px but maybe we need that re-alignment on consistency anyways

libs/blocks/icon-block/icon-block.css
```diff
@@ -103,14 +112,16 @@
...
+  min-width:var(--m-icon);
+  height: var(--m-icon);
 }
 
-.icon-block.inline.icon-height-64 .icon-area,
-.icon-block.inline.icon-height-64 .icon-area picture img {
-  height: 64px;
+.icon-block.inline.l-icon .icon-area,
+.icon-block.inline.l-icon .icon-area picture img {
+  min-width: var(--l-icon);
```
ryanmparrish, line 123:
```suggestion
  min-width: var(--xl-icon);
```

libs/blocks/icon-block/icon-block.css
```diff
@@ -103,14 +112,16 @@
...
+  height: var(--l-icon);
```
ryanmparrish, line 124:
```suggestion
  height: var(--xl-icon);
```

libs/blocks/icon-block/icon-block.css
```diff
@@ -144,6 +155,27 @@
   max-width: 147px;
 }
 
+.five-up .icon-block.inline .foreground .icon-area,
+.five-up .icon-block.inline .foreground .icon-area img {
+  max-height: unset;
+  max-width: 80px;
```
ryanmparrish, line 161:
```suggestion
  max-width: var(--xxl-icon);
```

libs/blocks/icon-block/icon-block.css
```diff
@@ -267,11 +311,25 @@
     height: 80px;
```
ryanmparrish, line 311:
```suggestion
    height: var(--xxl-icon)
```

libs/blocks/icon-block/icon-block.css
```diff
@@ -267,11 +311,25 @@
...
   }
 
+  .icon-block.inline.bio .icon-area,
+  .icon-block.inline.bio .icon-area picture img {
+    min-width:80px;
```
ryanmparrish, line 316:
```suggestion
    min-width: var(--xxl-icon);
```

libs/blocks/icon-block/icon-block.css
```diff
@@ -295,6 +353,33 @@
     height: 80px;
```
ryanmparrish, line 353:
```suggestion
    height: var(--xxl-icon);
```

zagi25:
@ryanmparrish main ticket (https://jira.corp.adobe.com/browse/MWPW-129909) was moved to 2nd review . I notified them that I will be on PTO next week and told them that if this is urgent that they should ask somebody from my team to takeover. If not I will address your comments when I return from PTO. Thank you.

zagi25:
@ryanmparrish @seanchoi-dev @rgclayton @Blainegunn PR is updated, please check it out. Thanks.

libs/blocks/icon-block/icon-block.css
ryanmparrish: Thanks @zagi25 
One more similar spacing issue i'm seeing on the first icon block w regards to needing bottom margin on the Icon asset. Can you extend these styles to include additional `.icon-block.full-width.small .foreground .icon-area` ex as well? 

https://github.com/adobecom/milo/pull/881/files#diff-cd28b64751555fd1edeb00e5befdd09a237354424c9880ae5621ee81a13fec68L124-L130

```
.icon-block.full-width.small .foreground .icon-area,
.icon-block.vertical.small .foreground .icon-area {
}

.icon-block.full-width.medium .foreground .icon-area,
.icon-block.vertical.medium .foreground .icon-area {
  margin-bottom: var(--spacing-m);
}
```

ryanmparrish:
@zagi25 This is looking good, Thanks for those updates. I left one tiny comment in the css. 
Do you have your eslint package installed? I'm seeing a few linting errors that we should try to clean up.

libs/blocks/icon-block/icon-block.js
```diff
@@ -39,6 +39,13 @@ const iconBlocks = {
   },
 };
 
+//checking if block is part of xx-up and is inline variant
+function upAndInline(el) {
+  const isInlineAndSectionMetadata = el.classList.contains('inline') && el.parentElement?.querySelector('.section-metadata')
+  if (!isInlineAndSectionMetadata) return false;
+  return /(two|three|four|five)[- ]?up/i.test(el.parentElement.querySelector('.section-metadata').textContent);
```
seanchoi-dev, line 46: nit: missing semicolon and simplifying the test object.
```suggestion
  const isInlineAndSectionMetadata = el.classList.contains('inline') && el.parentElement?.querySelector('.section-metadata');
  if (!isInlineAndSectionMetadata) return false;
  return /(two|three|four|five)[- ]?up/i.test(el.parentElement.className);
  ```

seanchoi-dev (approved):
Looks good to me. I commented with a small suggestion, but approving it.

rgclayton:
Looking much better! Thank you @zagi25 .

Some clarification questions related to icon sizing. 
New icon size variants:

1. `m-icon` (40px)
2. `l-icon` (56px)
3.  `default` (80px)

These variants only work with the `icon-block.inline` variation, correct? 

cc: @Blainegunn @ryanmparrish

libs/blocks/icon-block/icon-block.css
rgclayton: I think the impact is low, but did this ever get communicated to the community?

rgclayton: I posted in #milo-community relaying the breaking change.

rgclayton (approved):
Thanks for the updates @zagi25 

I think we are good once the requests from @ryanmparrish are added.

libs/blocks/icon-block/icon-block.css
zagi25: @ryanmparrish  `var(--spacing-m)` is default bottom margin https://github.com/adobecom/milo/blob/main/libs/blocks/icon-block/icon-block.css#L68, but it has "fullwidth" class instead of "full-width". "fullwidth" is used throughout the whole file, is this correct class or should I change all occurrences of "fullwidth" to "full-width" ?

libs/blocks/icon-block/icon-block.js
zagi25: Thanks, this looks better.

zagi25:
PR is updated. I added css rule that @ryanmparrish requested and fixed lint errors. Thank you all for your reviews and feedback.
cc: @rgclayton @seanchoi-dev @Blainegunn @JasonHowellSlavin

ryanmparrish (approved):

PR #884, opened 2023-06-22
MWPW-129654: Add 'needs-verification' to PR label check

hparra:
Resolves: [MWPW-129654](https://jira.corp.adobe.com/browse/MWPW-129654)

I added 'needs-verification' label: https://github.com/adobecom/milo/issues/labels and changed enforce PR label workflow to accept it.

**Test URLs:**
- Before: https://main--milo--hparra.hlx.page/?martech=off
- After: https://needs-verification--milo--hparra.hlx.page/?martech=off

### DIFF
--- a/.github/workflows/enforce-labels.yml
+++ b/.github/workflows/enforce-labels.yml
@@ -9,6 +9,6 @@ jobs:
     steps:
     - uses: yogevbd/enforce-label-action@2.1.0
       with:
-        REQUIRED_LABELS_ANY: "verified,trivial"
-        REQUIRED_LABELS_ANY_DESCRIPTION: "PR must be labeled with 'verified-fixed' or 'trivial'"
+        REQUIRED_LABELS_ANY: "verified,trivial,needs-verification"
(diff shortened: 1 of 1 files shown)


hparra:
✅ Green Check appeared!

mokimo (approved):
As a dev, really dislike the big red "X" which indicates my PR is not ready. Cool addition 👍

overmyheadandbody (approved):
I quite like the idea, it's annoying to get notifications that the proper labels haven't been added when opening a PR

Ruchika4 (approved):

auniverseaway:
I want to make sure I understand this...

As an engineer I create a PR and add "needs verification" before hitting the create button so that I don't get a red X because of labels?

For people that do not have the ability to add labels they still get the red X?
If someone adds a PR and they don't know if they need "verification" they also get a red X?

If I do understand this correctly, I would almost rather get rid of the verification related label requirements altogether as this seems like a work around and busy work for people who can add labels.

Is there a scenario where a PR does not need verification? Do we merge things in without verifying them?

PR #886, opened 2023-06-23
Replace code coverage tool

hparra:
Resolves: [MWPW-129597](https://jira.corp.adobe.com/browse/MWPW-129597)

**Test URLs:**

- https://bye-codecov--milo--hparra.hlx.page/?martech=off

**Description**

In the past 7 days, there have been at least 7 times when the codecov report for a PR was not uploaded. 5 of these PRs have already been merged without their codecov being uploaded. This breaks coverage reporting for all subsequent PRs. In each of those 5 cases, there were check failures and in some cases the UT coverage was skipped.

The codecov tool and workflow is creating more noise than signal. Developers are paying cost of poor or broken coverage in unrelated parts of the code, and are being conditioned to ignore errors.

See snapshot of https://app.codecov.io/gh/adobecom/milo/pulls below.

Screen Shot 2023-06-22 at 18 17 29

**Challenges**

It is only possible for this workflow to comment on the PR when the PR comes from an adobecom/milo branch. This is because workflow requires write permission from its automatically created `secrets.GITHUB_TOKEN` and for security reasons the permission is always read-only when request comes from a public PR. For this reason this change will fail tests when coverage drops. It is the responsibility of the developer to check for failure.


hparra:
This works in my own fork: https://github.com/hparra/milo/pull/1

It does not work here because the scope of `secrets.GITHUB_TOKEN` is always `read` for PRs from public forks, regardless of `permissions` settings in workflow. The only way to make this work is for people to always make PR from a adobecom/milo branch, which will have correct permissions. See https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/

honstar:
> It is only possible for this workflow to comment on the PR when the PR comes from an adobecom/milo branch. This is because workflow requires write permission from its automatically created secrets.GITHUB_TOKEN and for security reasons the permission is always read-only when request comes from a public PR.

I think this is a blocker for accepting the change of this PR. We actually want folks to fork the `milo` repo and raise PRs from their forks, especially newly onboarded `Milo Contributors` won't have any write permissions to the `milo` repo.

hparra:
> I think this is a blocker for accepting the change of this PR. We actually want folks to fork the `milo` repo and raise PRs from their forks, especially newly onboarded `Milo Contributors` won't have any write permissions to the `milo` repo.

They can still do that. The problem is under this configuration we won't get the update comment on the PR, instead if coverage fails then the Unit Tests will fail.

chrischrischris (approved):
>  if coverage fails then the Unit Tests will fail.

I think that's fine (or even better).  If we don't like this change it's easy to revert so I think we should give it a try.

hparra:
@chrischrischris @auniverseaway Ironically, codecov seems to be causing less problems lately. I'm on the fence with this one. 🤷‍♀️

auniverseaway (approved):
I'm not in love this this, but I get it.

```
# Only works for PRs coming from adobecom/milo branches
```

Am I understanding it correctly that PRs from forks will break on this?

hparra:
> I'm not in love this this, but I get it.

Me neither. Maybe we should hold on this.

> Only works for PRs coming from adobecom/milo branches
> Am I understanding it correctly that PRs from forks will break on this?

I confused a few ppl with that sentence. What "breaks" is the comment on PR to show coverage report -- I have disabled that so PRs from forks don't error out. The check still runs and UT check will fail if coverage fails.

Unrelated to PR, seems this test is erring consistently:

 ❌ marquee > supports media credits > has a media credit with element content
      AssertionError: expected null to exist
        at o.<anonymous> (test/blocks/marquee/marquee.test.js:48:29)

hparra:
If we continue improving UTs (flapping, coverage) and ensure green checked PRs are landing on main then codecov will continue to behave. I can reopen this if it starts acting up again.

PR #893, opened 2023-06-26
[MWPW-132444] Save up nav space when content is overflowing

overmyheadandbody:
Certain Gnav consumers add a lot of navigation content despite [Consonant's recommendations](https://consonant.adobe.com/1975e5ba1/p/48adfd-navigation/b/815684):
> Global nav items | Use five items max.

There have been dozens of discussions on the topic over the past years, but a perfect solution was never found to satisfy marketing/content requirements, ease of authoring, UX and performance concerns.

This PR aims to 'squeeze' a few extra pixels from all navigation items when the content is overflowing the viewport. This does not cover all possible use-cases, but is a good compromise for the moment. If the content overflows, the logic will add a modifier which in turn will:
- remove the brand label;
- remove the (right side) logo entirely;
- lower paddings and font-sizes for all main navigation items.

**Screenshots** @ 1200px viewport width:
| Before | After |
| AEC - before | AEC - after |

Resolves: [MWPW-132444](https://jira.corp.adobe.com/browse/MWPW-132444)

**Test URLs:**
- Before: https://main--milo--overmyheadandbody.hlx.page/drafts/ramuntea/gnav-refactor-long?martech=off
- After: https://enhance-overflowing-gnav--milo--overmyheadandbody.hlx.page/drafts/ramuntea/gnav-refactor-long?martech=off

### DIFF
--- a/libs/blocks/global-navigation/base.css
+++ b/libs/blocks/global-navigation/base.css
@@ -103,6 +103,12 @@
     padding: 0 12px;
   }
 
+  .feds-topnav--overflowing .feds-navLink,
+  .feds-topnav--overflowing .feds-navLink--hoverCaret,
+  [dir = "rtl"] .feds-topnav--overflowing .feds-navLink--hoverCaret {
+    padding: 0 8px;
+  }
+
   .feds-navLink--hoverCaret {
     position: static;
   }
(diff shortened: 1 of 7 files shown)


mokimo (approved):
Very cool! That should really help on cluttered navigations, since we can't enforce that only 5 items are added.

overmyheadandbody:
As a note for other reviewers - to avoid resize listeners and debouncing, this mostly treats a real-life scenario - if you resize the browser manually, you might not always get the best layout. However, if you load the page and the content would normally overflow, then the styles from this PR will apply. Switching from portrait to landscape would achieve the same result.

honstar (approved):

narcis-radu (approved):

overmyheadandbody:
Failing unit tests come from Personalisation, for which @mokimo already opened a PR - https://github.com/adobecom/milo/pull/920. I will merge this as-is for now.

PR #497, opened 2023-02-21
MWPW-121136 add georoutingV2

vhargrave:
Adds the new georouting modal for adobe.com as defined here - https://xd.adobe.com/view/1088565a-caf4-4d87-90fb-4f276c82371e-45db/

- The v2 UX presents a single language by default, with tabbed access to other options.​ 
- We have 16 geos w/ content in >=2 languages. (e.g. India, Brussels, Canada)
- Details on default language for these geo​ is being finalized [here](https://adobe.sharepoint.com/:x:/r/sites/AdotcomEnterprisePersonalization/_layouts/15/Doc.aspx?sourcedoc=%7BA35309EE-6280-4BAF-96FB-D5B280D4020E%7D&file=Geo%20Routing%20v2%20UX%20multi-language%20defaults.xlsx&wdLOR=cCBCDED0E-9136-6048-A90B-420DCEBA868E&action=default&mobileredirect=true ) and will be relevant for authors

Resolves: [MWPW-121136](https://jira.corp.adobe.com/browse/MWPW-121136)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/about/?akamaiLocale=CH
- After: https://vhargrave-georouting-v2--milo--adobecom.hlx.page/about/?akamaiLocale=CH

- Before: https://main--milo--adobecom.hlx.page/about/?akamaiLocale=DE
- After: https://vhargrave-georouting-v2--milo--adobecom.hlx.page/about/?akamaiLocale=DE

**Bacom (still using the old georouting so do not expect to see the new modal)**
- https://main--bacom--adobecom.hlx.page/?akamaiLocale=CH
- https://main--bacom--adobecom.hlx.page/?akamaiLocale=CH&milolibs=vhargrave-georouting-v2

**Milo Coffee (has georoutingV2 and georoutingV1 activated)**
- https://main--locale-testing--vhargrave.hlx.page/milo-coffee/subpage/?akamaiLocale=CH
- https://main--locale-testing--vhargrave.hlx.page/milo-coffee/subpage/?milolibs=vhargrave-georouting-v2&akamaiLocale=CH

PR #897, opened 2023-06-26
MWPW-133063-Fix for tabs in a fragment not loading on a page

Ruchika4:
* Fix for tabs in a fragment not loading on a page

Resolves: [MWPW-133063](https://jira.corp.adobe.com/browse/MWPW-133063)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ruchika/fragmentpage?martech=off
- After: https://tab-in-fragment--milo--ruchika4.hlx.page/drafts/ruchika/fragmentpage?martech=off

As I am updating a code change that was done to make sure tabs load fine in georouting modal as part of this [PR](https://github.com/adobecom/milo/pull/497/files) hnece adding urls to verify that georouting modal with tabs also works fine after this PR

Georouting URLs to verify tabs in modal

Milo

- Before: https://main--milo--adobecom.hlx.page/drafts/ruchika/fragmentpage?akamaiLocale=in
- After: https://tab-in-fragment--milo--ruchika4.hlx.page/drafts/ruchika/fragmentpage?akamaiLocale=in

Bacom
- Before: https://main--bacom--adobecom.hlx.page/?akamaiLocale=CH
- After: https://main--bacom--adobecom.hlx.page/?akamaiLocale=CH&milolibs=tab-in-fragment--milo--ruchika4

### DIFF
--- a/libs/blocks/tabs/tabs.js
+++ b/libs/blocks/tabs/tabs.js
@@ -99,7 +99,7 @@ const handleDeferredImages = (block) => {
 }
 
 const init = (block) => {
-  const rootElem = block.closest('.georouting-wrapper.fragment') || document;
+  const rootElem = block.closest('.fragment') || document;
   const rows = block.querySelectorAll(':scope > div');
   const parentSection = block.closest('.section');
   /* c8 ignore next */
@@ -180,7 +180,7 @@ const init = (block) => {
         const values = row.children[1].textContent.split(',');
         id = values[0];
         val = getStringKeyName(String(values[1]));
-        assocTabItem = rootElem.getElementById(`tab-panel-${id}-${val}`);
+        assocTabItem = rootElem.querySelector(`#tab-panel-${id}-${val}`);
       }
       if (assocTabItem) {
         const section = sectionMetadata.closest('.section');
--- a/libs/blocks/tabs/tabs.js
+++ b/libs/blocks/tabs/tabs.js
@@ -99,7 +99,7 @@ const handleDeferredImages = (block) => {
 }
 
 const init = (block) => {
-  const rootElem = block.closest('.fragment') || document;
+  const rootElem = block.closest('.georouting-wrapper.fragment') || document;
   const rows = block.querySelectorAll(':scope > div');
   const parentSection = block.closest('.section');
   /* c8 ignore next */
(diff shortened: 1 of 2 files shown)


sukamat (approved):

overmyheadandbody (approved):

libs/blocks/tabs/tabs.js
```diff
@@ -99,7 +99,7 @@ const handleDeferredImages = (block) => {
 }
 
 const init = (block) => {
-  const rootElem = block.closest('.fragment') || document;
+  const rootElem = block.closest('.georouting-wrapper.fragment') || document;
```
vhargrave, line 102: I remember that this change was done because tabs weren't loading properly within some fragments, so I'm a little worried to just change this without doing some thorough testing. 
Addtionally, I do not think it's a good change to have georouting specific selectors within the tabs block. Blocks should be independent of one another, so doing this looks like a bandage of the problem and is leading to an anti-pattern. 
I'd like to have more time to look, but don't really have any  atm. 
Why don't we just fix the selector on line 183 which is causing the current problem ? 
`assocTabItem = rootElem.querySelector(`#tab-panel-${id}-${val}`);`

vhargrave (changes requested):

libs/blocks/tabs/tabs.js
Ruchika4: Thanks for the review @vhargrave . I agree to your point blocks should be independent. Only reason I did this change because there was already a block (fragment) added here so with existing approach we already have that block dependency issue. Also I could only find georouting modal working with existing fix hence the change all the other fragments I checked were broken with tab. Please do share if you remember which other fragments needed the initial fix. In the meantime I have gone ahead and updated the selector. Please review and let me know if any concern.  We might also need to fix the other issue of dependency of 2 blocks (tab and fragment) which already exists (not related to this PR).

Ruchika4:
Fixing the unit tests.

Ruchika4:
> Fixing the unit tests.



> Fixing the unit tests.

Unit tests are fixed now

ryanmparrish (approved):
Looks good. Thanks 🙏

sukamat (approved):

libs/blocks/tabs/tabs.js
vhargrave: If I get some time to look I'll send them on. 
the change now looks good to me though, thanks!

vhargrave (approved):
looking good!

PR #215, opened 2022-09-28
MWPW-115387 - Table Of Contents

Brandon32:
* Adding table of contents to be used with the marquee.
* This needed to be a separate block to fill the full width without breaking marquee containers and image layout.

Resolves: [MWPW-115387](https://jira.corp.adobe.com/browse/MWPW-115387)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/bmarshal/table-of-contents?martech=off
- After: https://table-of-contents--milo--adobecom.hlx.page/drafts/bmarshal/table-of-contents?martech=off

PR #899, opened 2023-06-26
marquee-anchors - new block

ryanmparrish:
* Added a new `marquee-anchors` block. 

This layout was initially implemented with a `marquee`, `toc` and related `section-metadata` blocks. 
This new block is standalone and is intended to replace/deprecate current usage.  See this [earlier PR](https://github.com/adobecom/milo/pull/215) for more info

Resolves: [MWPW-129912](https://jira.corp.adobe.com/browse/MWPW-129912)

**Test URLs:**
- Before [https://main--milo--adobecom.hlx.page/drafts/rparrish/marquee-nav](https://main--milo--adobecom.hlx.page/drafts/rparrish/marquee-nav)
- After [https://rparrish-marquee-nav--milo--adobecom.hlx.page/drafts/rparrish/marquee-nav](https://rparrish-marquee-nav--milo--adobecom.hlx.page/drafts/rparrish/marquee-nav)

**Authoring notes:**
Document anchor list rows start in the 3rd table row. (1. background color, 2. marquee content, 3, 4, 5... list rows)
List rows, header and footer are optional. 
The list rows link identifier is the 'display text' starts with a `#`. 
In page anchor links should have identical 'display text' and 'link' fields. see screenshot...
Screen Shot 2023-06-26 at 3 36 03 PM

The [down arrow svg](https://main--milo--adobecom.hlx.page/libs/img/ui/arrow-down.svg) location is in /milo/libs/img/ui/arrow-down.svg ->  https://adobe.sharepoint.com/sites/adobecom/Shared%20Documents/Forms/AllItems.aspx?FolderCTID=0x012000F36D5B4C46F81741BCAC9F03FA9F93D1&id=%2Fsites%2Fadobecom%2FShared%20Documents%2Fmilo%2Flibs%2Fimg%2Fui&viewid=d776cf70%2D9b7e%2D4ab7%2Db9da%2D9e0f8e03a7d2


libs/blocks/marquee-anchors/marquee-anchors.css
```diff
@@ -0,0 +1,175 @@
+/*
+[marquee-anchors]
+variants: (dark, transparent)
+*/
```
JasonHowellSlavin, line 4: meant to keep this comment?

libs/blocks/marquee-anchors/marquee-anchors.js
```diff
@@ -0,0 +1,80 @@
...
+// size: [heading, body, ...detail]
+const blockTypeSizes = {
+  default: {
+    xsmall: ['xs', 'xs', 'xs'],
+    small: ['m', 's', 's'],
+    medium: ['l', 'm', 'm'],
+    large: ['xl', 'm', 'l'],
+    xlarge: ['xxl', 'l', 'xl'],
+  },
+};
```
JasonHowellSlavin, line 16: Looks like there is a whole system of sizes being used in this block and elsewhere. Might be a good to create a discussion for it.

test/blocks/marquee-anchors/marquee-anchors.test.js
```diff
@@ -0,0 +1,23 @@
+import { readFile } from '@web/test-runner-commands';
+import { expect } from '@esm-bundle/chai';
+
+document.body.innerHTML = await readFile({ path: './mocks/body.html' });
+const { default: init } = await import('../../../libs/blocks/marquee-anchors/marquee-anchors.js');
+describe('marquee-anchors', () => {
+  const marquees = document.querySelectorAll('.marquee-anchors');
+  marquees.forEach((marquee) => {
+    init(marquee);
```
JasonHowellSlavin, line 9: maybe in a `beforeEach` and have one `describe` block?

JasonHowellSlavin (commented):
Looks good. Do we know where the bacom TOC is used? Plan to replace? Also there does seem to be a style difference between what is in https://github.com/adobecom/milo/pull/215 and what is here.

elan-tbx:
@ryanmparrish should there be a gap below the image here?

Screenshot 2023-06-27 at 1 16 02 PM

elan-tbx (commented):
@ryanmparrish Should there be a gap below the image here?

Screenshot 2023-06-27 at 1 16 02 PM

rgclayton:
Just had another thought.. When adding a link that leaves the current page, the icon is still using the down arrow - which from a UX perspective is telling me I'm going to visit content on the page below when clicked. What do think about changing the icon when the link is pointed to content outside of the current page?

libs/blocks/marquee-anchors/marquee-anchors.css
ryanmparrish: I scrapped this unnecessary comment above and fixed/added a few, ex.  //variants:dark comments next to their class position.

ryanmparrish:
> Just had another thought.. When adding a link that leaves the current page, the icon is still using the down arrow - which from a UX perspective is telling me I'm going to visit content on the page below when clicked. What do think about changing the icon when the link is pointed to content outside of the current page?

Love this idea. I added some logic to remove the icon when linking external. Do you think we should change it however to a different icon? This could be discussed and added as an enhancement

test/blocks/marquee-anchors/marquee-anchors.test.js
ryanmparrish: @JasonHowellSlavin  - I updated the test file adding a beforeEach etc similar to other tests. 
I don't think this snippet however should be in that func. I also have just one describe block now. 
Anyhow I ran it,  and it checks out. LMK if this looks right? Thanks

ryanmparrish:
> Looks good. Do we know where the bacom TOC is used? Plan to replace? Also there does seem to be a style difference between what is in #215 and what is here.

I don't know all the potential docs the TOC is used. I can coordinate a plan w/ the PMs at BACOM to help resolve any updates. Upon a quick search it looks to be fairly minimal.  I was following the XD in the new design for this block styles, I didn't do a close cross comparison from old to new, but if we need to put on the breaks to give that feedback some more time, we certainly can. I know JohnSpencer was involved however he's out on Sabbatical 🤷  

Screen Shot 2023-06-27 at 4 25 17 PM

libs/blocks/marquee-anchors/marquee-anchors.js
```diff
@@ -0,0 +1,84 @@
...
+  decorateBlockText(links, blockTypeSizes.default.xsmall);
+  foreground.append(links);
+  el.append(foreground);
+
+  [...list].forEach((i) => {
+    const aTag = i.querySelector('a');
+    if (aTag?.textContent.charAt(0) === '#') {
+      const content = i.querySelector(':scope > div');
+      // (href === origin+path) - url is an anchor
+      const hrefPathEqual = (aTag.href.split('?')[0] === window.location.origin + window.location.pathname);
```
Brandon32, line 63: nit: you could save `window.location.origin + window.location.pathname` to a `const originPath` or similar since it's not going to change. Maybe you could use `window.location.href.split('?')[0]` and remove the comment if that's more clear?

Brandon32:
Since you are using headers in authoring the Marquee Anchor block the header text has to be different than the headers they link to since both headers will get the same ID.
Have you seen content that uses the same titles in both places? You might think about remove the IDs from the Marquee Anchor link content.

libs/blocks/marquee-anchors/marquee-anchors.css
```diff
@@ -0,0 +1,179 @@
+html {
+	scroll-behavior: smooth;
+}
+
+:target::before {
```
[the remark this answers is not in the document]
Brandon32: If this adjusts for the gnav then it seems it would be better to add this there and fix the root cause.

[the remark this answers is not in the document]
ryanmparrish: This is to give the target anchor element a padding offset above its scroll position so the content isn't right at the top edge. It's not critical, lmk if you think it's overstepping or any concerns w/ other block conflicts.  This selector could also benefit accessibility by styling the targeted headline.

libs/blocks/marquee-anchors/marquee-anchors.css
```diff
@@ -0,0 +1,179 @@
+html {
+	scroll-behavior: smooth;
+}
```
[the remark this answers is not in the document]
ryanmparrish: That's a good thought, I suppose it is possible it could apply to any block w/ a contained `overflow: scroll` property. I'll test tabs, etc. and see if any others might be impacted. 

Do we want this, scroll-behavior and/or anchor offset 1) as a milo default? 2) have it at all?

libs/blocks/marquee-anchors/marquee-anchors.css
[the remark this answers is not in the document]
chrischrischris: I like that it pushes it down a bit.  One very minor thing is that anchor links will behave differently when an anchor-marquee is on the page vs when that css hasn't loaded.  However anchor links in Franklin/Milo are pretty broken anyways so for now I don't think it's a concern

libs/blocks/marquee-anchors/marquee-anchors.js
ryanmparrish: Good call. That is more clear. I've updated this. Thanks

ryanmparrish:
> You might think about remove the IDs from the Marquee Anchor link content.

@Brandon32 Good thought, I've addressed this issue but appending a namespace to headers IDs in the links. Thanks

libs/blocks/marquee-anchors/marquee-anchors.css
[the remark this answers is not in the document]
rgclayton: We should maybe use this this property `scroll-margin-top` on the header elements that have the id on them. Super simple.

https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-margin-top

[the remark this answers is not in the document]
rgclayton: >This is to give the target anchor element a padding offset above its scroll position so the content isn't right at the top edge. It's not critical, lmk if you think it's overstepping or any concerns w/ other block conflicts. This selector could also benefit accessibility by styling the targeted headline.

I think we definitely need something like this, otherwise the header/title element will go behind the gnav/breadcrumbs.

I think maybe applying it to the `<h>` element, like I mentioned above, might be a better solution.

rgclayton:
> > Just had another thought.. When adding a link that leaves the current page, the icon is still using the down arrow - which from a UX perspective is telling me I'm going to visit content on the page below when clicked. What do think about changing the icon when the link is pointed to content outside of the current page?
> 
> Love this idea. I added some logic to remove the icon when linking external. Do you think we should change it however to a different icon? This could be discussed and added as an enhancement

I think we should show a different icon. There's a fairly universal icon for this. Something like this:
image

Where did the down arrow come from? Consonant? Spectrum?

ryanmparrish:
@rgclayton  - sounds good. no prob on the Icon swap. I can add that one to the external links. The arrow-down icon came from the design file in the Jira ticket.

ryanmparrish:
@rgclayton  - I made a minor modification and am now fetching 2 icons for anchor and external links. Take a peek when you get a chance and let me know what you think. Thanks

elaineskpt (approved):

PR #900, opened 2023-06-27
MWPW-132724 POC Lottie Support

TsayAdobe:
This is a POC of supporting Lottie. Do not merge.

* An author should use a Word hyperlink to add a lottie. e.g. (where `Link` could be any valid link. It's not used.)
  ```
   Display text:  :lottie: | /libs/img/lotties/Compress_PDF_60fps.json
  Link: https://www.adobe.com/ 
  ``` 
* If the width and height are needed for style.
  ```
   Display text:  :lottie: | /libs/img/lotties/Compress_PDF_60fps.json | width: 300px; height: 300px
  Link: https://www.adobe.com/ 
  ``` 

* The lottie file can be a .json or .lottie file and stored in the code repo or a media storage.
* This POC uses `dotlottie-player` (420KB), which can play both .json and .lottie files. There is a .json only player `lottie-player`, which size is smaller (367KB). If authors will not use .lottie files, we can switch to the player.

Resolves: [MWPW-132724](https://jira.corp.adobe.com/browse/MWPW-132724)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/tsay/mwpw-132724?martech=off
- After: https://mwpw-132724--milo--tsayadobe.hlx.page/drafts/tsay/mwpw-132724?martech=off


rgclayton:
This introduces a bunch of performance issues. The performance score is one of the worst I've seen in Milo so far. Couple that with real content on a page and I think there is a bunch of performance adjustments that need to happen for this to be viable. We currently get much better performance using a video or even an animated gif.

honstar:
Since this is a POC, it should *not* go into `main` and rather into a feature branch for further review if at all. Looks like there is an active discussion and performance concerns that need to be addressed IMO.

sukamat:
Agree with @honstar . You can also change it to a draft PR to avoid accidental merges.

honstar:
> Agree with @honstar . You can also change it to a draft PR to avoid accidental merges.

Draft PRs are discouraged, see https://github.com/adobecom/milo/wiki/Submitting-PRs#drafts-prs

auniverseaway:
I'm going to close this, but a couple thoughts:

1. I would have expected this to be a block, not embedded into the fabric of Milo.
2. Performance is rough... the JIRA ticket says, "it's vector, that means performant!" and that's not at all true since moving vectors is CPU intensive.
3. We strongly encourage PRs to be attached to go-live dates in addition to what others have said. This PR takes eyeballs away from code that needs to ship. We love experimenting and welcome it, but that's why we have branches attached to HLX urls... for experimenting. If you want automatic PR checks, you can submit a PR to yourself or use pagespeed directly.

PR #845, opened 2023-06-06
MWPW-130971 - Centralized 404

auniverseaway:
**Features**
* Maintains legacy compatibility
* Defaults to local 404 using 404 - local meta property
* Allows for federated 404 content via libs (using 404 - feds)
* Ability to have any design

**Under the hood**
* Added `base` to config
* Fragment correctly respects default content via css.

Resolves: [MWPW-130971](https://jira.corp.adobe.com/browse/MWPW-130971)

**Milo URLs (Federated)**
https://main--milo--adobecom.hlx.page/does-not-exist?martech=off
https://central-404--milo--adobecom.hlx.page/does-not-exist?martech=off
https://central-404--milo--adobecom.hlx.page/de/does-not-exist?martech=off

**CC URLs (Federated)**
https://milo-central-404--cc--aishwaryamathuria.hlx.page/this-doesnt-exist?milolibs=central-404
(from local repo)

https://main--cc--adobecom.hlx.live/this-doesnt-exist?milolibs=central-404
https://main--cc--adobecom.hlx.live/de/this-doesnt-exist?milolibs=central-404

**Bacom URLs (Legacy)**
https://main--bacom--adobecom.hlx.live/this-doesnt-exist?milolibs=central-404
https://main--bacom--adobecom.hlx.live/de/this-doesnt-exist?milolibs=central-404 (still pulls English)

**Milo College URLs (Local)**
https://main--milo-college--adobecom.hlx.page/does-not-exist/?milolibs=central-404

PR #903, opened 2023-06-27
MWPW-132987 Adjusting Modal Fragment Widths 

JasonHowellSlavin:
* Adding a max-width for modal fragments to override changes to width that was done for 404s
* [Changes made in this PR](https://github.com/adobecom/milo/pull/845/files#diff-c685e2a5191e8c907b9958b7f267d04cc48fdce1491da2c38461955f5d2bd913R203) to fragment widths were effecting videos inside of modals due to content width being set.

Resolves: [MWPW-132987](https://jira.corp.adobe.com/browse/MWPW-132987)

**Test URLs:**
- Before: https://main--bacom--adobecom.hlx.page/customer-success-stories/hugo-boss-case-study?martech=off
- Before: https://main--milo--adobecom.hlx.page/drafts/slavin/video-modal-one?martech=off
- After: https://main--bacom--adobecom.hlx.page/customer-success-stories/hugo-boss-case-study?milolibs=video-scroll-bar-issue&?martech=off
- After https://video-scroll-bar-issue--milo--adobecom.hlx.page/drafts/slavin/video-modal-one?martech=off

edit: Removing the hash param from the url to get an accurate LH score.


sukamat (approved):

mokimo (approved):

libs/blocks/modal/modal.css
```diff
@@ -49,6 +49,10 @@
   outline: 2px solid var(--modal-focus-color);
 }
 
+.dialog-modal > .fragment > .section > .content {
```
mokimo, line 52: I'm thinking the `.dialog-modal > .fragment > .section` shouldn't have such a deep selector which you then need to overwrite with yet another level...

JasonHowellSlavin: Interesting issue we have here because `dialog-modal` and `fragment` are technically the namespaces for these two autoblocks. Do you think I should change the selector in the fragments file to `.fragment > .content` and this one to `.dialog-modal > .content` @mokimo ?

Brandon32 (approved):

elaineskpt (approved):

libs/blocks/modal/modal.css
mokimo: If that works, that would seem better yeah!

JasonHowellSlavin: @mokimo I made the specificity a bit better. I decided not to change the fragment styles in `style.css` since there might be edge cases of which I am unaware. Let me know what you think!

rgclayton (approved):

rgclayton:
Late to the party. But, I think I like this solution better too. Since the issue is related to modals it makes sense to only correct the modal experience and leave the regular content experience -  fragment > section > content - to have the max-width of 1200px desktop and 83.4% tablet and mobile.

PR #904, opened 2023-06-27
MWPW-131607: Extend feds.entitlements with the offer ID

mirafedas:
* Added the `offers` field in the `entitlements` object. This field will hold the offer IDs and offers data, which will be used on Titans to form the URL for the switch plan flow.

Resolves: [MWPW-131607](https://jira.corp.adobe.com/browse/MWPW-131607)

**Test URLs:**
https://mwpw-131607-entitlements--milo--adobecom.hlx.page/?martech=off - link doesn't work, will add correct later

### DIFF
--- a/libs/blocks/global-navigation/utilities/getUserEntitlements.js
+++ b/libs/blocks/global-navigation/utilities/getUserEntitlements.js
@@ -29,6 +29,7 @@ const emptyEntitlements = {
   arrangment_codes: {},
   fulfilled_codes: {},
   offer_families: {},
+  offers: {},
   list: { fulfilled_codes: [] },
 };
 const CREATIVE_CLOUD = 'creative_cloud';
@@ -37,10 +38,10 @@ const EXPERIENCE_CLOUD = 'experience_cloud';
 
 /**
  * Breaks JIL offers(subscriptions) into easily addressable flat data structure.
- * @param {*} offers
+ * @param {*} allOffers
  */
-const mapSubscriptionCodes = (offers) => {
-  if (!Array.isArray(offers)) {
+const mapSubscriptionCodes = (allOffers) => {
+  if (!Array.isArray(allOffers)) {
     return emptyEntitlements;
   }
 
@@ -49,10 +50,11 @@ const mapSubscriptionCodes = (offers) => {
     arrangment_codes,
     fulfilled_codes,
     offer_families,
+    offers,
     list,
   } = emptyEntitlements;
 
-  offers.forEach(({ fulfilled_items, offer = {} }) => {
+  allOffers.forEach(({ fulfilled_items, offer = {} }) => {
     const cloud = offer.product_arrangement?.cloud;
     clouds[CREATIVE_CLOUD] = clouds[CREATIVE_CLOUD] || cloud === 'CREATIVE';
     clouds[DOCUMENT_CLOUD] = clouds[DOCUMENT_CLOUD] || cloud === 'DOCUMENT';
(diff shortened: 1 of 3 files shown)


libs/blocks/global-navigation/utilities/getUserEntitlements.js
```diff
@@ -75,13 +77,19 @@ const mapSubscriptionCodes = (offers) => {
         }
       });
     }
+
+    if (Object.prototype.hasOwnProperty.call(offer, 'offer_id')) {
```
mokimo, line 81: Why not just `if( offer["offer_id"] )`?

libs/blocks/global-navigation/utilities/getUserEntitlements.js
```diff
@@ -75,13 +77,19 @@ const mapSubscriptionCodes = (offers) => {
...
+      const { offer_id, ...rest } = offer;
+      offers[offer_id] = rest;
```
mokimo, line 83: I just noticed a bug in the code, as we mutate the empty entitlements, on subsequent calls they still have the mutated values. Would be nice if you fix this in your PR the code is here: https://github.com/mokimo/milo/commit/6598d6340941e914f59979f68a3b596a4a392979

Otherwise I'll open a follow up PR

mokimo (approved):

mokimo:
You will need to Sign the Adobe CLA, see the failing check in the PR

libs/blocks/global-navigation/utilities/getUserEntitlements.js
mirafedas: Thanks a lot @mokimo ! I included this change

VKniaz (approved):

libs/blocks/global-navigation/utilities/getUserEntitlements.js
```diff
@@ -24,35 +24,37 @@ const getQueryParameters = (params) => {
   return query.length ? `?${query.join('&')}` : '';
 };
 
-const emptyEntitlements = {
+const emptyEntitlements = () => ({
```
overmyheadandbody, line 27: I'm not sure I see a reason for this to be a function rather than a simple object. It doesn't return anything, it doesn't have arguments, it doesn't get updated, etc. So I'd vote for this to be a simple object. Test file also uses a simple object, so we should update here to be consistent.

overmyheadandbody (approved):

mokimo:
No one uses this on prod yet and this has been changed for the AEM entitlements too, hence I'll take the freedom to merge this

libs/blocks/global-navigation/utilities/getUserEntitlements.js
narcis-radu: I agree with @overmyheadandbody - we should add some unit testing to cover this, I had the same question around using an object vs function.

npeltier:
@mokimo @overmyheadandbody @mirafedas sorry to spot this only now, but we should probably not use offerid anymore. Happy to meet tomorrow to discuss that 😗

mokimo:
@npeltier maybe you can group with @mirafedas since she handled the requirement? It was also added in hawks a few days ago and then ported here

overmyheadandbody:
Just spotted this. Yes, a discussion would be appropriate, as this will go in to both Hawks and Milo. We already started some discussions with @yesil around this over e-mail (👀), maybe he can provide some additional context.

yesil:
@npeltier @overmyheadandbody @mokimo In the Hawks PR, the offer ID is not authored, but received from the JIL request so it is dynamic. Same should be the case here.
Similar to how we generate checkout URLs with offer ID in it, here with this PR and Hawks PR, we will generate switch offer URLs with offer IDs in it because the URL points to an external service which expects an offer ID.

I agree that authors should not use these offer IDs for personalization use cases.

PR #909, opened 2023-06-29
Personalization martech fix: also check for window._satellite presence

chrischrischris:
Before running martech a check for `window.marketingtech.adobe.launch` is done to see if it's already loaded.  With the new setup we also need a check for `window._satellite`

**Test URLs:**
- After: https://martechfix--milo--adobecom.hlx.page/?martech=off


rgclayton (approved):

sukamat (approved):

Blainegunn:
@chrischrischris thank you!

PR #910, opened 2023-06-29
Fix broken scripts test

meganthecoder:
* Fixes broken test in scripts.js > Decorating > martech test

Resolves: https://github.com/adobecom/milo/actions/runs/5414607828/jobs/9841911472

**Test URLs:**
N/A

### DIFF
--- a/test/scripts/scripts.test.js
+++ b/test/scripts/scripts.test.js
@@ -6,29 +6,6 @@ import { waitForElement } from '../helpers/waitfor.js';
 document.head.innerHTML = await readFile({ path: './mocks/head.html' });
 document.body.innerHTML = await readFile({ path: './mocks/body.html' });
 
-const EXTERNAL_SCRIPTS = [
-  'https://www.adobe.com/marketingtech/main.standard.min.js',
-];
-
-// Prevents loading of EXTERNAL_SCRIPTS
-const observer = new MutationObserver((mutations) => {
-  mutations.forEach(({ addedNodes }) => {
-    addedNodes.forEach((node) => {
-      if (node.nodeType === 1 && node.tagName === 'SCRIPT') {
-        if (EXTERNAL_SCRIPTS.includes(node.src)) {
-          node.setAttribute('type', 'javascript/blocked');
-        }
-      }
-    });
-  });
-});
-
-// Starts the monitoring
-observer.observe(document.head, {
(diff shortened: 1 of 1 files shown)


meganthecoder:
Also getting error from global-navigation.test.js of exceeding time limit. Likely due to loadIms changes

test/scripts/scripts.test.js
```diff
@@ -6,29 +6,6 @@ import { waitForElement } from '../helpers/waitfor.js';
...
-  mutations.forEach(({ addedNodes }) => {
-    addedNodes.forEach((node) => {
-      if (node.nodeType === 1 && node.tagName === 'SCRIPT') {
-        if (EXTERNAL_SCRIPTS.includes(node.src)) {
-          node.setAttribute('type', 'javascript/blocked');
-        }
-      }
-    });
-  });
-});
```
mokimo, line 24: This is very interesting; do you know if that worked, and it prevented loading the script? 

I'd assume the moment something is added to the page it would start loading - and the intersection observer would be too late to block loading it.

mokimo:
> Also getting error from global-navigation.test.js of exceeding time limit. Likely due to loadIms changes

Addressed here: https://github.com/adobecom/milo/pull/911

chrischrischris (approved):
Other than Okans question 👍

test/scripts/scripts.test.js
meganthecoder: I'm not sure, sorry. The marketingtech script was moved into the libs folder, breaking the test. Since it's no longer an external script, I removed that code, but I'm not familiar enough with the test previously to know if it worked before.

sukamat (approved):

Ruchika4:
@meganthecoder unit test check is still failing, although don't see any test failing in logs.

meganthecoder:
> @meganthecoder unit test check is still failing, although don't see any test failing in logs.

Yes, that's what Okan has a PR for #911

Ruchika4 (approved):

test/scripts/scripts.test.js
mokimo: To answer the question, it will in any case _download_ script but should stop _execution_

mokimo:
Thanks for the PR! Given all the approvals and as this just touches tests, I'll merge it

PR #913, opened 2023-06-30
MWPW-133172 - Add Floodgate option in Caas Configurator and Pass floodgate param to Caas

arshadparwaiz:
* Add Floodgate option in Caas Configurator for Author Preview of Cards served from Floodgate
* When "Copy Link" operation is performed, Remove the FG option from the copied URL, so that when the URL is pasted on a Card Collection page under FG tree, and later the same page is promoted to Default tree, the FG link should be not included, hence not include the FG param in the copied Card Collection URL.
* In a separate Ticket, the Caas Libs will determine if the Card Collection page is under FG tree, then served cards will be from FG tree, otherwise from Default tree.

Resolves: [MWPW-124986](https://jira.corp.adobe.com/browse/MWPW-124986)

**Test URLs:**
- Before: https://milo.adobe.com/tools/caas
- After: https://main--milo--arshadparwaiz.hlx.page/tools/caas?caasver=0.6.38


mokimo:
Same here, no draft PRs: https://github.com/adobecom/milo/wiki/Submitting-PRs#drafts-prs

libs/blocks/caas-config/caas-config.js
```diff
@@ -673,7 +689,8 @@ const CopyBtn = () => {
 
   const getUrl = () => {
     const url = window.location.href.split('#')[0];
-    return `${url}#${utf8ToB64(JSON.stringify(state))}`;
+    state.fetchCardsFromFloodgateTree = 'default';
+    return `${url}#${utf8ToB64(JSON.stringify(state, fgKeyReplacer))}`;
```
sukamat, line 693: Could we use the delete operator here and avoid the additional replacer function call?

```
delete state.fetchCardsFromFloodgateTree;
return `${url}#${utf8ToB64(JSON.stringify(state))}`;
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete

May be @chrischrischris or someone else can confirm as I am not fully aware of the side effects of using it.

arshadparwaiz: @sukamat I didn't delete it because every time Authors click on "Copy Link", they will loose their FG checkbox selection in the configurator, causing the card results on the Configurator page to show Default cards instead of Floodgate cards. Hence deleting it from the JSON while performing copy link. If we prefer this option, will update this. Will discuss this with you.

libs/blocks/caas-config/caas-config.js
```diff
@@ -706,6 +707,21 @@ const saveStateToLocalStorage = (state) => {
...
 
+/**
+ * Removes the JSON key "fetchCardsFromFloodgateTree" from the Copied URL to Caas.
+ * Caas Collection will determine if the content should be served from floodgate
+ * based on  metadata.xslx logic in caas-libs
+ * @param {*} key jsonKey
+ * @param {*} value jsonValue
+ * @returns replacedJson
+ */
+const fgKeyReplacer = (key, value) => {
```
narcis-radu, line 718:
```suggestion
const fgKeyReplacer = (key, value) => key === 'fetchCardsFromFloodgateTree' ? undefined : value;
```

honstar:
@arshadparwaiz : please adjust `caas` unit-tests, there are 10 failures for `test/blocks/caas/utils.test.js`.

honstar:
Please also remember to populate the `before` and `after` test URLs, thanks.

honstar (changes requested):
CaaS unit tests are failing, please fix.

libs/blocks/caas/utils.js
```diff
@@ -266,6 +266,8 @@ export function getCountryAndLang({ autoCountryLang, country, language }) {
   };
 }
 
+const getfloodGateColor = ({ fetchCardsFromFloodgateTree }) => (fetchCardsFromFloodgateTree ? 'pink' : 'default');
```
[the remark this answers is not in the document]
arshadparwaiz: Sure @chrischrischris, will update this

arshadparwaiz:
> @arshadparwaiz : please adjust `caas` unit-tests, there are 10 failures for `test/blocks/caas/utils.test.js`.

@honstar Sure will resolve these. Currently there was a technical discussion causing change in code, am working on the updates, will ensure these as well. Thanks.

cmiqueo (approved):

arshadparwaiz:
@honstar Implemented enhancements based on architectural discussions, incorporated review comments & fixed unit tests. Thanks for checking.

chrischrischris (approved):

sukamat (approved):

honstar:
> @honstar Implemented enhancements based on architectural discussions, incorporated review comments & fixed unit tests. Thanks for checking.

Thanks @arshadparwaiz 👍 Can you please address the outstanding linting issues? Then we should be good to go.

arshadparwaiz:
> Thanks @arshadparwaiz 👍 Can you please address the outstanding linting issues? Then we should be good to go.

Thanks for checking @honstar . The Lint issues are pre-existing from the files part of this PR.
I have now fixed 2 out of 4 issues now. The remaining 2 issues could potentially impact the functionality, and I don't know the impact of this on code & features.

PR #915, opened 2023-07-01
Use only language subtag for html lang attribute

hparra:
Use only language subtag for html lang attribute based on various recommendations, including Google and WebAIM:

- https://www.youtube.com/watch?v=isW-Ke-AJJU&t=3354s
- https://www.seroundtable.com/google-still-does-not-use-html-lang-attribute-34207.html
- https://adrianroselli.com/2015/01/on-use-of-lang-attribute.html
- https://webaim.org/techniques/language/#langtags

To accomplish this easily we added `lang` for language subtag to `locales` config. We also add `reg` for region subtag based on the geo and not the `ietf` value.

Resolves: [MWPW-133115](https://jira.corp.adobe.com/browse/MWPW-133115)

**Test URLs:**

- After: https://seo-html-lang--milo--hparra.hlx.page/?martech=off


hparra:
Tests are currently broken. See #910 and #911.

chrischrischris (changes requested):
Though our use of the `lang` attribute may be non-standard, it is used for determining lang-region in some places (e.g: [gnav](https://github.com/adobecom/milo/blob/main/libs/blocks/gnav/gnav.js#L23)). There is a historical context to this as it was used in dexter to determine lang-country.

Setting "request-changes" as I do not want this merged until we've done a thorough review.

I also would like to understand why this is a required change.

In the references you refer to I don't see anything saying why including the country variant is bad.
* On https://webaim.org/techniques/language/#region it actually says to include the region.
* Looking at the lang attribute on https://html.spec.whatwg.org/multipage/dom.html it's set to `en-US-x-hixie`.
* The list of langs on https://developers.google.com/recaptcha/docs/language include regions.

We will have to wait until after summer break to vet this.

chrischrischris (approved):

libs/scripts/scripts.js
```diff
@@ -21,103 +21,103 @@ const prodDomains = ['milo.adobe.com'];
...
-  at: { ietf: 'de-AT', tk: 'vin7zsi.css' },
-  cz: { ietf: 'cs-CZ', tk: 'aaz7dvd.css' },
-  bg: { ietf: 'bg-BG', tk: 'aaz7dvd.css' },
-  ru: { ietf: 'ru-RU', tk: 'aaz7dvd.css' },
-  ua: { ietf: 'uk-UA', tk: 'aaz7dvd.css' },
-  il_he: { ietf: 'he', tk: 'nwq1mna.css', dir: 'rtl' },
-  ae_ar: { ietf: 'ar', tk: 'nwq1mna.css', dir: 'rtl' },
-  mena_ar: { ietf: 'ar', tk: 'dis2dpj.css', dir: 'rtl' },
-  sa_ar: { ietf: 'ar', tk: 'nwq1mna.css', dir: 'rtl' },
+  africa: { ietf: 'en', lang: 'en', reg: '002', tk: 'pps7abe.css' },
```
overmyheadandbody, line 35: I was just curious where you got the '002' value, I don't see a mention of it on https://developers.google.com/recaptcha/docs/language

libs/scripts/scripts.js
```diff
@@ -21,103 +21,103 @@ const prodDomains = ['milo.adobe.com'];
...
+  il_en: { ietf: 'en-IL', lang: 'en', reg: 'IL', tk: 'pps7abe.css' },
+  it: { ietf: 'it-IT', lang: 'it', reg: 'IT', tk: 'bbf5pok.css' },
+  lv: { ietf: 'lv-LV', lang: 'lv', reg: 'LV', tk: 'aaz7dvd.css' },
+  lt: { ietf: 'lt-LT', lang: 'lt', reg: 'LT', tk: 'aaz7dvd.css' },
+  lu_de: { ietf: 'de-LU', lang: 'de', reg: 'LU', tk: 'vin7zsi.css' },
+  lu_en: { ietf: 'en-LU', lang: 'en', reg: 'LU', tk: 'pps7abe.css' },
+  lu_fr: { ietf: 'fr-LU', lang: 'fr', reg: 'LU', tk: 'vrk5vyv.css' },
+  hu: { ietf: 'hu-HU', lang: 'hu', reg: 'HU', tk: 'aaz7dvd.css' },
+  mt: { ietf: 'en-MT', lang: 'en', reg: 'MT', tk: 'pps7abe.css' },
+  mena_en: { ietf: 'en', lang: 'en', reg: '', tk: 'pps7abe.css' },
```
overmyheadandbody, line 56: What's the reasoning behind leaving `mena_en` and `mena_ar` without a region? I think in the past we had `mena`.

libs/scripts/scripts.js
hparra: `002` is the [UN M49](https://en.wikipedia.org/wiki/UN_M49) code for Africa. You can use UN M49 for region subtag [BCP 47](https://en.wikipedia.org/wiki/IETF_language_tag).

libs/scripts/scripts.js
hparra: TBD. Current `ietf` codes are just the lang (ar). I did not change these. `mena` is not a valid region and there isn't a valid code to represent MENA either. Northern Africa is `015` but there is no Middle East region -- those countries are part of Western Asia (`145`).

Ruchika4 (approved):

PR #916, opened 2023-07-01
Use locale from getConfig (gnav)

hparra:
Get locale (ietf) from `getConfig` instead of from  `<html lang>`.

Both `config.locale` and `<html lang>` are set by `setConfig`. The value of `<html lang>` will change with #915.

**Test URLs:**
- After: https://fix-html-lang-personalization--milo--hparra.hlx.page/?martech=off


hparra:
Tests are currently broken. See https://github.com/adobecom/milo/pull/910 and https://github.com/adobecom/milo/pull/911.

overmyheadandbody (approved):
New gnav has already ported to using the `getConfig` locale, so this should be fine

mokimo (approved):

honstar (approved):

PR #923, opened 2023-07-05
MWPW-133444 use wcs prod as default

npeltier:
- introduces new commerceEnv configuration, to differ from general one. No usage yet to make of it.
    - change boolean logic for having prod by default
    - adapt OST calls,
    - adapt tests accordingly

Resolves: [MWPW-133444](https://jira.corp.adobe.com/browse/MWPW-133444)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.live/drafts/ilyas/merch
- After: https://mwpw-133444--npeltier--adobecom.hlx.live/drafts/ilyas/merch


npeltier:
might add some stage unit test as we provide that feature without using it

web-test-runner.config.mjs
```diff
@@ -1,5 +1,6 @@
 import { importMapsPlugin } from '@web/dev-server-import-maps';
 import { defaultReporter } from '@web/test-runner';
+import { chromeLauncher } from '@web/test-runner-chrome';
```
mokimo, line 3: Did you want to commit this? Sneaky way of getting this in! 😁

honstar: In light of #922 , I don't think we should add this.

npeltier: damn

npeltier: (sorry about that, just me doing yet another round of test and forgot to remove it again :/)

libs/blocks/ost/ost.js
```diff
@@ -125,7 +125,7 @@ export async function loadOstEnv() {
     }
   }
 
-  ({ country, language } = getTacocatEnv(ENV_PROD, locale));
+  ({ country, language } = getTacocatEnv(true, locale));
```
yesil, line 128: can we remove `env` parameter all together since it will be always `PRODUCTION`?

libs/blocks/merch/merch.js
```diff
@@ -32,15 +34,15 @@ export function getTacocatLocale(locale) {
   return { country, language };
 }
 
-export function getTacocatEnv(envName, locale) {
+export function getTacocatEnv(isProd = true, locale) {
```
yesil, line 37: arguments with default values should be at the end in the list.
but do we still need env parameter to start with?

libs/blocks/ost/ost.js
npeltier: not sure it will always be it: we need a way to test new wcs / aos / whatever commerce change. We need 
1. our codebase to be able to point to stage from time to time 
2. our codebase to be able to use draft landscape from time to time
1 & 2 will be handled later, now we are just not closing the door to 1.

libs/blocks/merch/merch.js
npeltier: discussed above, will change the order

yesil:
@npeltier there is a conflict in merch.js

mokimo (approved):

npeltier:
looks like @vladen fixed it in https://github.com/adobecom/milo/pull/929. We discussed with @vladen & @yesil about better communication for next time, and i'll close this :)

PR #924, opened 2023-07-05
MWPW-133465: Localise gnav mega menus

mokimo:
Correctly localise global navigation menus. Open the first link on both pages and see if the large menu for `PDF` is localised

Resolves: [MWPW-133465](https://jira.corp.adobe.com/browse/MWPW-133465)

**Test URLs:**
- Before: https://main--dc--adobecom.hlx.live/jp/acrobat/online/compress-pdf?headerqa=global-navigation
- After: https://main--dc--adobecom.hlx.live/jp/acrobat/online/compress-pdf?headerqa=global-navigation&milolibs=mwpw-133465--milo--mokimo


overmyheadandbody (approved):
Good catch! I once looked into placeholders being replaced, but this slipped through the cracks 👍

narcis-radu (approved):

honstar:
Weird, the `after` link doesn't render a globalnav at all for me. Getting a few errors in the browser console:

Screenshot 2023-07-06 at 15 26 00

mokimo:
@honstar sometimes not every file is correctly updated, I've manually updated the `utils.js` so the error doesn't happen and the nav renders .. there are a few other errors in the console but that's related to not all files correctly matching `adobecom:main` I think it can be ignored

honstar:
> it might be related to UPPERCASE branch names not being supported

We should double-check with the franklin team if this is really the case. I've seen many branches using our JIRA ticket IDs as branch name, i.e. `MWPW-xxxx` which worked just fine.

mokimo:
@honstar Asked in the franklin chat yesterday and saw a 2-year-old message saying uppercase branches are not supported 🤷 

Since I've faced this issue a couple of times already and locally the errors don't exist, we are safe.
If there's no objections, I'll merge the branch soon - the code purely touches the global-navigation which we don't have any prod consumers for yet

honstar (approved):

PR #927, opened 2023-07-07
[MWPW-131409] Update .html postfix

overmyheadandbody:
For performance reasons, we should ideally iterate over a given list of elements just once and group all relevant logic inside this single loop.

Right now, both [decorateLinks](https://github.com/adobecom/milo/blob/main/libs/utils/utils.js#L504) and [appendHtmlPostfix](https://github.com/adobecom/milo/blob/main/libs/utils/utils.js#L300) iterate over the same links.

This PR splits the `appendHtmlPostfix` logic into two parts:
- one for the canonical URL replacement, `appendHtmlToCanonicalUrl`; this is called in `loadArea` if the `area` parameter is a document;
- another for link URL update, `appendHtmlToLink`; this is called in the link loop from `decorateLinks`.

Unit tests have been added to ensure that `decorateLinks` adds the `.html` extension to a set of links if needed and that `loadArea` updates the canonical URL and adds the `.html` extension to document links that need it.

Milo consumers shouldn't be affected by this change, since the logic was and still is encapsulated in `loadArea`. Having a look over `appendHtmlPostfix` method calls [1], I see some references in 'milo-stage', 'milo-pink' and 'milo-target', I'm not sure what they represent and whether we need to make any changes there, input from @auniverseaway would be appreciated for this. We will also need to get in touch with the DC team, as they seem to [mock](https://github.com/adobecom/dc/blob/8037674416415f09ed570bc1667127de2249ea4a/test/mocks/milo/libs/utils/utils.js#L250) the Milo functionality for unit testing purposes.

[1] - https://github.com/search?q=org%3Aadobecom%20appendHtmlPostfix(&type=code

Resolves: [MWPW-131409](https://jira.corp.adobe.com/browse/MWPW-131409)

**Test URLs:**
- Before: https://main--milo--overmyheadandbody.hlx.page/drafts/ramuntea/gnav-refactor?martech=off
- After: https://update-htmlpostfix--milo--overmyheadandbody.hlx.page/drafts/ramuntea/gnav-refactor?martech=off


libs/utils/utils.js
```diff
@@ -257,63 +257,57 @@ export function loadStyle(href, callback) {
   return loadLink(href, { rel: 'stylesheet', callback });
 }
 
-export function appendHtmlPostfix(area = document) {
+export function appendHtmlToCanonicalUrl() {
   const pageUrl = new URL(window.location.href);
   if (!pageUrl.pathname.endsWith('.html')) return;
+  const canonEl = document.head.querySelector('link[rel="canonical"]');
+  if (!canonEl) return;
+  const { href } = canonEl;
```
narcis-radu, line 265: I know the code is not new, but was wondering whether we really need the property;
```suggestion
  const canonUrl = new URL(canonEl.href);
  if (canonUrl.pathname.endsWith('/') || canonUrl.pathname.endsWith('.html')) return;
  const pagePath = pageUrl.pathname.replace('.html', '');
  if (pagePath !== canonUrl.pathname) return;
  canonEl.setAttribute('href', `${canonUrl.href}.html`);

```

overmyheadandbody: Good point! I tried to commit your change from the GUI, but it mangled the code, I've pushed it from my local

test/utils/appendHtmlToLink.test.html
```diff
@@ -0,0 +1,206 @@
...
+          expect(hrefs).to.eql([
+          'http://localhost:2000/testA.html',
+          'http://localhost:2000/endswithslash/',
+          'http://localhost:2000/has/qp/and/hash.html?r=34&hello=world#blah',
+          'http://localhost:2000/has/extension.html',
+          'http://localhost:2000/test/relative/link.html',
+          'http://localhost:2000/test/endswith/slash/',
+          'http://localhost:2000/has/qp/and/hash.html?r=34&hello=world#blah',
+          'http://localhost:2000/test/relative/linkwith.html',
+          'http://adobe.com/test',
```
mokimo, line 199: shouldn't this be `test.html` ?

test/utils/appendHtmlToLink.test.html
```diff
@@ -0,0 +1,206 @@
...
+        <a href="/test/endswith/slash/"></a>
+        <a href="/has/qp?a=1&b=2"></a>
+        <a href="/has/hash#blah"></a>
+        <a href="/has/qp/and/hash?r=34&hello=world#blah"></a>
+        <a href="/test/relative/linkwith.html"></a>
+        <a href="/"></a>
+        <a href=""></a>
+      </div>
+      <div id="externallinks">
+        <a href="http://adobe.com/test"></a>
```
mokimo, line 38: adobe.com might be a bad example for an external link, on first sight it looks like an internal one!

overmyheadandbody: This was picked up from the initial tests - https://github.com/adobecom/milo/blob/main/test/utils/htmlpostfix.test.html#L35-L37, didn't see a good reason to change it.

mokimo (approved):

narcis-radu (approved):

auniverseaway (approved):
This is awesome!

libs/utils/utils.js
```diff
@@ -257,63 +257,57 @@ export function loadStyle(href, callback) {
   return loadLink(href, { rel: 'stylesheet', callback });
 }
 
-export function appendHtmlPostfix(area = document) {
-  const pageUrl = new URL(window.location.href);
-  if (!pageUrl.pathname.endsWith('.html')) return;
+const pageUrl = new URL(window.location.href);
+const pathEndsWithHtml = pageUrl.pathname.endsWith('.html');
```
chrischrischris, line 261: Move to top of file if we're going to set these as "globals"

After that `pageUrl` can also be used on lines 116 & 646

chrischrischris (approved):

libs/utils/utils.js
overmyheadandbody: Thanks for the feedback! Implemented the suggestions in a new commit.

chrischrischris:
Verified on https://business.adobe.com/customer-success-stories.html

PR #929, opened 2023-07-07
MWPW-133483: Milo Commerce

vladen:
This PR creates "Milo Commerce" feature and replicates minimum amount of Dexter Commerce code, sufficient to retrieve and display product prices, build checkout URLs on Milo pages.

In present architecture, product prices are obtained from offers served by [Web Commerce Service](https://developers.corp.adobe.com/wcs/docs/api/openapi/wcs/latest.yaml#/).

To communicate with this service, Milo Commerce uses [@pandora/data-source-wcs](https://git.corp.adobe.com/PandoraUI/odm-core/tree/master/packages/data-source-wcs) npm package along with some other proprietary packages hosted on corporate Artifactory.
These custom dependencies define the need for special `package.json` file, internal tests and build step.

Commerce dependencies should be installed separately and require Artifactory access token for `@pandora` scope to be stored on developer workstation (see `.npmrc` file).

During build, JS bundle is saved to `libs/deps/commerce.js` and should be imported by consumers from that location. New build have to be triggered for every git commit so the `deps` folder would contain the most recent version of bundle.
This PR includes simple git hook performing tests and build along with instruction how to install it in `readme.md`. For now, hook installation is manual.

The new feature has a set of internal tests, that can only run after installing all dependencies. Also, built bundle can tested externally without installing proprietary dependencies, via tests of `merch` and `ost` blocks - they consume Milo Commerce.

Resolves: [MWPW-133483](https://jira.corp.adobe.com/browse/MWPW-133483)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.live/drafts/ilyas/merch?martech=off
- After: https://commerce--milo--vladen.hlx.page/drafts/ilyas/merch?martech=off

"Before" page requires corporate VPN.
"After" page requires manual modification of CORS headers, this can be done with help of [modHeader](https://chrome.google.com/webstore/detail/modheader-modify-http-hea/idgpnmonknjnojddfkpgkljpfnnfcklj) Chrome extension.
Required header modifications are:
- `Access-Control-Allow-Origin: *`
- `Access-Control-Allow-Headers: x-api-key`

image


libs/deps/literals/price/ar.json
```diff
@@ -0,0 +1,12 @@
+{
```
mokimo, line 1: Should all of those really live in-code? The franklin/milo way would be to store those in sharepoint in a excel sheet and retrieve them from there

yesil: @mokimo they will be generated from an endpoint, they are not author-able.
Having said that, we might imagine generating and saving them in excel in the future, not sure if it is better.

libs/features/commerce/src/ims.js
```diff
@@ -0,0 +1,32 @@
+// @ts-nocheck
+export const IMS_POLL_INTERVAL = 200;
+export const IMS_POLL_MAX_ATTEMPTS = 25;
+
+/** @type {Commerce.pollImsCountry} */
+export function pollImsCountry({
```
yesil, line 6: it is time to check whether gnav can provide a better implementation to this?

libs/features/commerce/src/log.js
```diff
@@ -0,0 +1,128 @@
...
+    debug: createHandler(Level.debug),
+    error: createHandler(Level.error),
+    info: createHandler(Level.info),
+    warn: createHandler(Level.warn),
+  });
+
+  return log;
+}
+
+const common = createLog('milo');
```
yesil, line 77: initialize `common` this close to exports ? suggestion to rename: `main` or `mainLogger`

libs/features/commerce/src/placeholder.js
```diff
@@ -0,0 +1,153 @@
...
+
+  init() {
+    buckets.set(this, {
+      promises: [],
+      state: undefined,
+      version: 0,
+    });
+  },
+
+  onceSettled() {
```
yesil, line 45: I would avoid using a new term. We call placeholders resolved on success, so `onceResolved` sounds more natural to me.

libs/features/commerce/src/placeholder.js
```diff
@@ -0,0 +1,153 @@
...
+        new CustomEvent(RESOLVED, { bubbles: true }),
+      );
+    });
+    return true;
+  },
+
+  toggleFailed(version, error) {
+    const bucket = buckets.get(this);
+    // skip obsolete asyncs
+    if (version !== bucket.version) return false;
```
yesil, line 86: will this lead having unresolved promises for ever?
When we toggle, all the promises should be cleaned up properly.

libs/features/commerce/src/placeholder.js
```diff
@@ -0,0 +1,153 @@
...
+      bucket.timer = setImmediate(() => {
+        delete bucket.timer;
+        this.render();
+      });
+    }
+  }
+};
+
+// TODO: old name for backward compatibility, needs to be removed somewhen
+HTMLPlaceholderMethods.onceResolved = HTMLPlaceholderMethods.onceSettled;
```
yesil, line 136: any reason to prefer onceSettled to onceResolved ?

libs/features/commerce/src/service.js
```diff
@@ -0,0 +1,126 @@
...
+        /* c8 ignore next */
+        log.error('Price literals not loaded:', error);
+      },
+    );
+
+  // light up all DOM placeholders
+  document
+    .querySelectorAll('[is="checkout-link"],[is="inline-price"]')
+    // @ts-ignore
+    .forEach((el) => el.render());
```
yesil, line 75: Ideally it should also initialise custom placeholders.
Maybe we should dispatch an event instead?

libs/features/commerce/src/service.js
```diff
@@ -0,0 +1,126 @@
...
+/** @type {Commerce.init} */
+function init(callback, force) {
+  if (force) reset();
+  // eslint-disable-next-line no-return-assign
+  return promise ??= activate(callback);
+}
+
+/** @type {Commerce.Internal.Instance} */
+export default {
+  get checkout() {
```
yesil, line 104: can we initialize all members before export and, keep the export section clean and simple?

libs/features/commerce/test/checkoutLink.test.html
```diff
@@ -0,0 +1,29 @@
+<html>
+  <head>
+    <meta name="wcs-debounce-delay" content="1">
+    <meta name="wcs-offer-selector-limit" content="1">
+    <script type="importmap">
+      {
+        "imports": {
+          "@pandora/fetch": "./mocks/pandora-fetch.js",
```
yesil, line 8: +1

libs/features/commerce/test/checkoutLink.test.html
```diff
@@ -0,0 +1,29 @@
+<html>
+  <head>
+    <meta name="wcs-debounce-delay" content="1">
```
yesil, line 3: `wcs-buffer-delay`?

libs/features/commerce/test/mocks/lana.js
```diff
@@ -0,0 +1,15 @@
+import { sinon } from '../utils.js';
```
yesil, line 1: this could be a core milo mock. cc: @chrischrischris wdyt?

libs/features/commerce/test/mocks/literals.json
```diff
@@ -0,0 +1,12 @@
+{
```
yesil, line 1: do we need mock literals?
Since we already have all the literals copied in Milo?

libs/features/commerce/test/utils.js
```diff
@@ -0,0 +1,35 @@
...
+    return String(val).trim()
+      .replace(/>\s*</g, '><')
+      .replace(/>\s*/g, '>')
+      .replace(/\s*</g, '<')
+      .replace(/"\s*>/g, '">')
+      .replace(/"\s*\/>/g, '/>')
+      .replace(/\s+/g, ' ');
+  }
+
+  chai.Assertion.addMethod('html', function(snapshot) {
```
yesil, line 20: +1

test/blocks/merch/merch.test.js
```diff
@@ -1,266 +1,268 @@
...
+    setConfig(config);
   });
 
-  it('Doesnt decorate merch with bad content', async () => {
+  beforeEach(async () => {
+    const { init, Log } = await import('../../../libs/deps/commerce.js');
+    await init(() => config);
+    Log.reset();
+    Log.use(Log.quietFilter);
+    Merch = (await import('../../../libs/blocks/merch/merch.js')).default;
```
yesil, line 33: merch being a block, we use lower cases, why capitalize first letter?

libs/features/commerce/test/inlinePrice.test.html.js
```diff
@@ -0,0 +1,173 @@
+import Log from '../src/log.js';
```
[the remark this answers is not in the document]
vladen: @yesil 
this is because this file is loaded from **html** test in `inlinePrice.test.html` and should be never selected by web test runner for direct execution (because requires hosting `html` file)
per current configuration, web test runner selects all files matching pattern `./test/**/*.test.(js|html)`
`.test.html.js` does not match this pattern

libs/features/commerce/scripts/fetch-price-literals.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
+
+urlPattern="https://www.adobe.com/special/tacocat/literals/{lang}.js"
```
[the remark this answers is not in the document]
vladen: this is temporary script and should be removed
actually, I can remove it now
but ideally, it should be replaced with a working script requesting literals from Odin

libs/features/commerce/scripts/pre-commit
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env sh
+
+echo "Building libs/features/commerce"
+cd libs/features/commerce
+npm run test && npm run build:bundle && npm run build:types
```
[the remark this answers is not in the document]
vladen: yes, there is git hook for that and explanation in `readme.md` how to install it

libs/features/commerce/src/log.js
vladen: @yesil

This logger defines root logging namespace (`milo`) and consumers refer to it as `Log.common`. Renaming it as `mainLogger` would make it `Log.mainLogger`. The `Logger` suffix looks redundant to me here. 
There is also another pre-defined logging namespace: `Log.commerce` which is used by all Milo Commerce modules and outputs messages prefixed with `milo/commerce`

Alternative suggestion for renaming `Log.common`: `Log.milo`.

libs/features/commerce/src/placeholder.js
vladen: @yesil 

The `resolved` term has strong association with `Promise.resolve` static method.
But former `onceResolved` method returns promises that can be also `rejected` when associated placeholder fails. This can be confusing for consumers because `rejected` is an opposite to `resolved`.
The new `onceSettled` name is proposed for analogy with [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) static method.
For backward compatibility, the old `onceResolved` name is also kept.

libs/features/commerce/src/placeholder.js
vladen: please, see me answer above

libs/features/commerce/src/service.js
vladen: Why shouldn't we?
Attributes `is="checkout-link"` and `is="inline-price"` are introduced exactly in Dexter Commerce and are well known to this module. And it looks natural that just initialised service fires up all placeholders.
Maybe this reduces flexibility on consumer requiring to perform some extra attribution of DOM placeholder elements.
Let's discuss this offline.

libs/features/commerce/src/service.js
vladen: no problem

libs/features/commerce/test/checkoutLink.test.html
vladen: @yesil 

👍 
What if we simplify `wcs-offer-selector-limit` as well, say `wcs-buffer-limit`?

libs/features/commerce/test/mocks/literals.json
vladen: @yesil

Having them mocked gives more flexibility to tests. It makes easier to find literals used in tests and customise them.

test/blocks/merch/merch.test.js
vladen: renamed

libs/features/commerce/test/inlinePrice.test.html
```diff
@@ -0,0 +1,29 @@
+<html>
+  <head>
+    <meta name="wcs-debounce-delay" content="1">
```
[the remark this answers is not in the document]
vladen: @yesil please check my reply to same question below

libs/features/commerce/test/checkoutLink.test.html
vladen: @yesil, I renamed both, now it looks like:

```
<meta name="wcs-buffer-delay" content="1">
<meta name="wcs-buffer-limit" content="1">
```

libs/features/commerce/test/service.test.html.js
```diff
@@ -0,0 +1,63 @@
...
+
+  it('uses IMS country code', async () => {
+    await mockIms('CH');
+    const instance = await init();
+    await delay();
+    unmockIms();
+    expect(instance.ims.country).to.be.eventually.equal('CH');
+  });
+
+  it('call to reset de-initialises the service', async () => {
```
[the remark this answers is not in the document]
vladen: rephrased

libs/features/commerce/scripts/fetch-price-literals.sh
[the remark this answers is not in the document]
yesil: it not working let's remove for now. We can fix it add add it back.

libs/features/commerce/test/mocks/lana.js
yesil, line 1: this could be a core milo mock. cc: @chrischrischris wdyt?

libs/features/commerce/src/inlinePrice.js
```diff
@@ -0,0 +1,118 @@
...
+      this.placeholder.toggleFailed(version, error);
+    };
+  }
+
+  /**
+   * Render price offer info into this element.
+   * @param {Commerce.Wcs.Offer} offer 
+   * @param {Record<string, any>} overrides
+   */
+  renderOffer(offer, overrides = {}, version) {
```
yesil, line 74: default values should be last.

libs/features/commerce/.eslintrc.js
yesil, line 1: I think this should be renamed to .eslintrc.cjs

otherwise eslint complains on my local.

libs/features/commerce/src/service.js
```diff
@@ -0,0 +1,126 @@
...
+  document
+    .querySelectorAll('[is="checkout-link"],[is="inline-price"]')
+    // @ts-ignore
+    .forEach((el) => el.render());
+
+  log.debug('Initialised:', instance);
+  return instance;
+}
+
+function demand() {
```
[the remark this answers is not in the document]
vladen: `assertExists`?
this function just throws if instance is null with no attempt to create it

libs/features/commerce/src/inlinePrice.js
vladen: @yesil 
`version` arg is not supposed to be actively used by consumers, not that frequently as `overrides`
and it has its default :) which is `undefined`

libs/features/commerce/.eslintrc.js
vladen: I found that custom eslint config isn't needed in fact. Existing Milo config works perfectly.
Removed the file.

libs/features/commerce/scripts/fetch-price-literals.sh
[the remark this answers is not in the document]
vladen: removed

libs/features/commerce/src/ims.js
vladen: added todo comment in the code

libs/deps/commerce.d.ts
```diff
@@ -0,0 +1,316 @@
...
+         * associated with this osi.
+         * 
+         * If `multiple` is set to false (this is default value), resolved array will contain only one
+         * offer, selected by country/language-perpetual algorithm.
+         * Otherwise. all responded offers are returned.
+         * 
+         * If `taxExclusive` is set to true (default value defined in settings),
+         * then returned prices are transformed into tax exclusive variant.
+         */
+        resolveOfferSelectors(options: {
```
npeltier, line 232: as discussed offline, i really love that way of documenting APIs without bundling any extra byte to the outside payload. Once this is merged, let's discuss to make that more spread ❤️

libs/deps/commerce.d.ts
```diff
@@ -0,0 +1,316 @@
...
+      interface Instance extends Omit<Commerce.Instance, 'providers'> {
+        providers: Providers;
+      }
+
+      interface Providers {
+        price: Set<Commerce.providePriceOptions>
+      }
+    }
+
+    module Log {
```
npeltier, line 140: as discussed offline, looks like this module is not commerce specific, as well as `debug` query parameter / variable. I would love not to have a commerce specific logging system, and we should probably discuss how best to reuse existing or integrate that if we need more than existing. 
cc @chrischrischris @auniverseaway

libs/features/commerce/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "build:types": "cp types.d.ts ../../deps/commerce.d.ts",
+    "build:watch": "npm run build:bundle --watch",
+    "test": "wtr --config ../../../web-test-runner.config.mjs \"./test/**/*.test.(js|html)\" --node-resolve --port=2001 --coverage --concurrent-browsers 4",
+    "test:watch": "npm test -- --watch",
+    "test:file": "wtr --config ../../../web-test-runner.config.mjs --coverage --node-resolve --port=2001",
+    "test:file:watch": "wtr --config ../../../web-test-runner.config.mjs --coverage --node-resolve --port=2001 --watch"
+  },
+  "dependencies": {
+    "@dexter/tacocat-consonant-templates": "^1.9.7",
+    "@dexter/tacocat-core": "^1.10.3",
```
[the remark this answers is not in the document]
vladen: @npeltier, I agree that this would simplify Milo Commerce deps, however `tacocat-core` package was made free of any dependencies by design
instead, I'd recommend to create a new, specially crafted tacocat npm package for Milo, give it a self-explanatory name like `@dexter/tacocat-milo` and reexport all necessary dependencies from there
this could be done in a separate PR to prevent this one from bloating more

libs/features/commerce/src/inlinePrice.js
yesil: still, checkstyle will complain until we fix it.

libs/features/commerce/src/inlinePrice.js
```diff
@@ -0,0 +1,123 @@
...
+class HTMLInlinePriceElement extends HTMLSpanElement {
+  static get observedAttributes() {
+    return [
+      'data-wcs-osi',
+      'data-template',
+      'data-display-recurrence',
+      'data-display-per-unit',
+      'data-display-tax',
+      'data-promotion-code',
+      'data-display-price',
```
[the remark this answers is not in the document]
vladen: @yesil this attribute has been migrated from tacocat and does not seem to be used anywhere, removing it

yesil (approved):

libs/deps/literals/price/ar.json
3ch023: @mokimo @yesil  I am not sure which will be the advantage of putting them in a excel sheet. It could be that it will just increase the complexity.

auniverseaway (changes requested):
I see a lot of things in this PR that are Milo anti-patterns.

vladen:
Thank you, @auniverseaway for taking decision on the most disputable aspects this PR has.
It was designed as a proposal and can be easily moved to another repository.

Its build artifact, the `commerce.js` file can be either provided to Milo consumers as dependency from `libs/deps/` folder (and committed there whenever updated) or downloaded by browser from a CDN.
Or alternatively, all commerce-related blocks can be moved to the same repository for better co-location. This is something yet to decide.

libs/features/commerce/README.md
```diff
@@ -0,0 +1,26 @@
...
+## Development
+
+1. Obtain corporate artifactory access token and make it accessible to npm.
+  - you may want to check `.npmrc` file for list of required npm scopes
+2. Install npm dependencies
+  - `cd libs/features/commerce`
+  - `npm i`
+3. Run Adobe Helix Cli and open Commerce test page
+  - `hlx up`
+  - `http://localhost:3000/drafts/ilyas/merch?martech=off`
```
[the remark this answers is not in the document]
3ch023: what is the concern to use an actual page? it might change in future, but at it exists and we all use it as a template, so for a Dev following readme it will be nice to see some prices fast

PR #933, opened 2023-07-12
Ensure multiple places can load IMS from the milo utils

mokimo:
### Description
Refactor the milo core utils `loadIms` function so that any consumer can use the it as follows:
```js
loadIms()
  .then(() => {}) // ims available and instance has loaded
  .catch(() => {}) // error loading IMS and the instance is not ready
```

This is **NOT** a breaking change, as no consumer uses this [function yet]( https://github.com/search?q=org%3Aadobecom%20loadims&type=code)
This provides a more reliable way of knowing when IMS has finished loading and is ready to be used, especially when called from multiple places. (E.g. martech loads IMS, global-navigation loads IMS, consumer code loads IMS)

### Test instructions
Observe the sign in button to exist afterwards in the global navigation

Resolves: [MWPW-133665](https://jira.corp.adobe.com/browse/MWPW-133665)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/osahin/gnav-1
- After: https://mwpw-133665--milo--mokimo.hlx.page/drafts/osahin/gnav-1

- Before: https://main--bacom--adobecom.hlx.live/customer-success-stories
- After: https://main--bacom--adobecom.hlx.live/customer-success-stories?milolibs=mwpw-133665--milo--mokimo

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -48,7 +48,6 @@ const CONFIG = {
     'app-launcher',
     'adobe-logo',
   ],
-  imsReady: 'onImsLibInstance',
 };
 
 // signIn, decorateSignIn and decorateProfileTrigger can be removed if IMS takes over the profile
@@ -210,6 +209,8 @@ class Gnav {
     isDesktop.addEventListener('change', closeAllDropdowns);
   }, 'Error in global navigation init');
 
+  ims = async () => loadIms().then(() => this.imsReady()).catch(() => {});
+
   decorateTopNav = () => {
     this.elements.mobileToggle = this.decorateToggle();
     this.elements.topnav = toFragment`
@@ -233,14 +234,6 @@ class Gnav {
       </div>`;
 
     this.el.append(this.elements.curtain, this.elements.topnavWrapper);
-  };
-
-  ims = () => {
-    if (window.adobeIMS) {
-      if (window.adobeIMS.initialized) return this.imsReady();
-      return window.addEventListener(CONFIG.imsReady, this.imsReady, { once: true });
-    }
-    return loadIms(this.imsReady.bind(this));
   };
 
   addChangeEventListeners = () => {
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -209,7 +209,11 @@ class Gnav {
     isDesktop.addEventListener('change', closeAllDropdowns);
   }, 'Error in global navigation init');
 
-  ims = async () => loadIms().then(() => this.imsReady()).catch(() => {});
+  ims = async () => loadIms()
+    .then(() => this.imsReady())
+    .catch((e) => {
+      lanaLog({ message: 'GNAV: Error with IMS', e });
+    });
 
   decorateTopNav = () => {
     this.elements.mobileToggle = this.decorateToggle();
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -48,6 +48,7 @@ const CONFIG = {
     'app-launcher',
     'adobe-logo',
   ],
+  imsReady: 'onImsLibInstance',
 };
 
 // signIn, decorateSignIn and decorateProfileTrigger can be removed if IMS takes over the profile
@@ -193,7 +194,7 @@ class Gnav {
       this.decorateMainNav,
       this.decorateTopNav,
       this.decorateTopnavWrapper,
-      () => loadIms(this.imsReady.bind(this)),
+      this.ims,
       this.addChangeEventListeners,
     ];
     this.el.addEventListener('click', this.loadDelayed);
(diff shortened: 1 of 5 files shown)


libs/utils/utils.js
```diff
@@ -620,10 +620,13 @@ function decorateFooterPromo(config) {
 }
 
 export function loadIms(onReadyFn) {
-  if (window.adobeIMS) return;
+  if (window.adobeIMS) {
```
overmyheadandbody, line 623: I'm not sure this is enough - IMS is not necessarily ready if `adobeIMS` exists on the page. It is ready if `window.adobeid.onReady` has been called at least once. So one way to make sure we don't call `onReadyFn` prematurely would be to set a flag as part of the `window.adobeid.onReady` function body and then check here if that flag has already been set.

Another approach would be to use some event listeners to check when IMS is ready, something like
```
window.addEventListener('onImsLibInstance', (data) => {
    if(data?.detail?.instance) {
        console.log('IMS instance exists, IMS ready methods can be executed');
    }
});
window.dispatchEvent(new window.CustomEvent('getImsLibInstance'));
```

You'd likely need to use a combination of the two - if a flag has already been set in `window.adobeid.onReady`, execute `onReadyFn`. Otherwise, add the event listener logic and execute `onReadyFn` when the IMS instance is available.

overmyheadandbody (commented):
Not sure this solution is the final one.

Also, a description nit, I think 'we can just execute the logged in callback right away.' should actually be 'we can just execute the ready callback right away.', as we don't check logged in state just yet.

libs/utils/utils.js
mokimo: Thanks for the details, I was wondering about that 👍

mokimo: As discussed on Slack, I'll add the code to the `global-navigation` to not bulk up the utils even further.

We can still move it over to the utils if more consumers use the loadIms & run into the same issues

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -232,6 +233,14 @@ class Gnav {
     this.el.append(this.elements.curtain, this.elements.topnavWrapper);
   };
 
+  ims = () => {
+    if (window.adobeIMS) {
+      if (window.adobeIMS.initialized) return this.imsReady();
+      return window.addEventListener(CONFIG.imsReady, this.imsReady, { once: true });
+    }
+    return loadIms(this.imsReady.bind(this));
```
mokimo, line 241: It might make sense to eventually move this to the utils, but as the global-navigation(+gnav) are the only ones that pass a callback to loadIms so far - we wanted to keep this contained for now, to not further bloat the utils

mokimo:
TBT expected due to having martech on to load IMS from two different places. (Martech + global navigation)

overmyheadandbody (approved):

narcis-radu (approved):
We might need to iterate over this based on usage. The Global Navigation and all other blocks should still render if IMS is not ready (loaded but not ready due to a CORS error or blocked in the browser) and this implementation takes care of that.

vladen (approved):

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -208,6 +208,8 @@ class Gnav {
     isDesktop.addEventListener('change', closeAllDropdowns);
   }, 'Error in global navigation init');
 
+  ims = async () => loadIms().then(() => this.imsReady()).catch(() => {});
```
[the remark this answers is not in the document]
mokimo: I had it initially, removed it though because IMO it can cause some noise.
We can add it though and see what logs we get

chrischrischris (approved):
Verified on milo with personalization workflow.  One small nit

chadsunvice (approved):

libs/utils/utils.js
```diff
@@ -621,22 +621,31 @@ function decorateFooterPromo(config) {
...
-  loadScript('https://auth.services.adobe.com/imslib/imslib.min.js');
+let imsLoaded;
+export async function loadIms() {
+  imsLoaded = imsLoaded || new Promise((resolve, reject) => {
+    const { locale, imsClientId, imsScope, env } = getConfig();
+    if (!imsClientId) {
+      reject(new Error('Missing IMS Client ID'));
+      return;
+    }
+    const timeout = setTimeout(() => reject(new Error('IMS timeout')), 3000);
```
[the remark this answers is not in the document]
overmyheadandbody: I would say that the IMS timeout should be standardised across consumers, not a configurable value. This will impact Gnav and personalisation, which should be consistent across pages. True to the Milo spirit, I'd say we keep this 'static' until a need for a different behaviour arises.

I'd also argue that a constant might not be needed, since we use the value just once, so we'd end up adding extra characters to `utils` for no real benefit.

One change I would suggest, though, is to raise the value. With a slow 3G connection, Chrome downloads `imslib.js` in about 3.5s (granted, Chrome automatically adds 2s delay in this case if I'm not mistaken), so maybe `4000` would be a better option here.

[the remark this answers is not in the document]
mokimo: I'll keep as is for now and raised the timeout to 5s. We should observe how this behaves in the real world, also with personalization.
For the GNAV we will have some lana logs coming through seeing when/if/how often this will fail to load on time

PR #935, opened 2023-07-12
MWPW-133412 Pass correct country/lang to filters

chrischrischris:
When auto-country/lang was enabled, it was not updating the filter code with the auto-detected values.

Resolves: [MWPW-133412](https://jira.corp.adobe.com/browse/MWPW-133412)

**Test URLs:**
- Before: https://main--bacom--adobecom.hlx.page/fr/resources/main?martech=off
- After: https://main--bacom--adobecom.hlx.page/fr/resources/main?milolibs=filtersautolang&martech=off


mokimo (approved):

libs/blocks/caas/utils.js
```diff
@@ -362,7 +366,7 @@ export const getConfig = async (originalState, strs = {}) => {
       eventFilter: state.filterEvent,
       type: state.showFilters ? state.filterLocation : 'left',
       showEmptyFilters: state.filtersShowEmpty,
-      filters: await getFilterArray(state, strs),
+      filters: await getFilterArray(state, strs, country, language),
```
narcis-radu, line 369: `getFilterArray` will take `state`, `country` and `lang` as parameters but you're calling the method with an additional param - `strs`; I don't think this will produce the results you're expecting.

overmyheadandbody: Agree, it seems that `strs` wasn't used previously anyway, so it should be safe to simply remove.

auniverseaway (approved):

PR #940, opened 2023-07-14
[MWPW-133763] Long breadcrumbs UI

overmyheadandbody:
This solves two issues for breadcrumbs:

### Mobile UI when too many breadcrumb elements are added
Each breadcrumb element would flow on multiple rows depending on available space. The logic has now been adapted to: 
- keep elements on just one row;
- allow scrolling if available space is exceeded;
- replace all but the first and last two breadcrumb elements with an ellipsis, if more than 3 breadcrumb elements are available. Similar behaviour is currently implemented on the AEM-based Global Navigation.

| Before | After |
| Before breadcrumb UI updates | After breadcrumb UI updates |

### Remove hyphenation when breadcrumbs are inferred from the URL
If the URL is used to build the breadcrumbs, path pieces such as 'customer-success-stories' are added as-is. We're now simply replacing hyphens with spaces.

| Before | After |
| URL breadcrumbs before | URL breadcrumbs after |

---

**NOTE**: The `6px` top and bottom padding added to `.feds-breadcrumbs` should not cause CLS, since desktop remains unchanged, as seen [here](https://github.com/adobecom/milo/blob/main/libs/blocks/global-navigation/global-navigation.css#L565) and on mobile the breadcrumbs are nested inside the hamburger menu.

---

Resolves: [MWPW-133763](https://jira.corp.adobe.com/browse/MWPW-133763)

**Test URLs:**
- Before: https://main--milo--overmyheadandbody.hlx.page/drafts/ramuntea/Breadcrumbs/breadcrumbs-set-from-url?martech=off
- After: https://long-breadcrumbs-ui--milo--overmyheadandbody.hlx.page/drafts/ramuntea/Breadcrumbs/breadcrumbs-set-from-url?martech=off

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.css
+++ b/libs/blocks/global-navigation/global-navigation.css
@@ -303,8 +303,9 @@ header.global-navigation {
 }
 
 .feds-breadcrumbs {
-  padding: 0 12px;
+  padding: 6px 12px;
   font-size: 12px;
+  overflow-y: auto;
 }
 
 .feds-breadcrumbs ul {
@@ -316,6 +317,22 @@ header.global-navigation {
 .feds-breadcrumbs li {
   display: flex;
   align-items: center;
+  flex-shrink: 0;
+}
+
+/* Hide all breadcrumbs except the first and last two */
+.feds-breadcrumbs li:nth-last-child(n+3):not(:first-child) {
+  display: none;
+}
+
+/* If first breadcrumb is not third to last, add ellipsis after it */
+.feds-breadcrumbs li:first-child:not(:nth-last-child(-n+3)):after {
+  content: '/\2003...';
+  padding: 0 0 0 12px;
+}
+
+[dir = "rtl"] .feds-breadcrumbs li:first-child:not(:nth-last-child(-n+3)):after {
+  padding: 0 12px 0 0;
 }
 
 .feds-breadcrumbs a,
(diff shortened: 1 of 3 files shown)


overmyheadandbody:
Hey @JasonHowellSlavin! Thanks for the feedback, just added a quick unit test for the functionality.

JasonHowellSlavin (approved):

narcis-radu (approved):

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -316,6 +317,22 @@ header.global-navigation {
...
+}
+
+/* Hide all breadcrumbs except the first and last two */
+.feds-breadcrumbs li:nth-last-child(n+3):not(:first-child) {
+  display: none;
+}
+
+/* If first breadcrumb is not third to last, add ellipsis after it */
+.feds-breadcrumbs li:first-child:not(:nth-last-child(-n+3)):after {
+  content: '/\2003...';
```
fullcolorcoder, line 330: In Firefox and Safari the spacing doesn't come out quite right with this. What about breaking out the "/" and the "...", with the "/" going in :before and "..." in :after so you can control the spacing more consistently? 
image

fullcolorcoder (changes requested):

SilviuLCF (approved):

libs/blocks/global-navigation/global-navigation.css
SilviuLCF: Just a small add, inconsistent spacing replicate for Safari and FF on MAC desktop resizing browser.  
On real devices (Android and Iphone ) works properly for all browsers. 
Safari_IOS
FF_android
FF_IOS

fullcolorcoder (approved):

chrischrischris (approved):

PR #945, opened 2023-07-17
Use locale from getConfig (caas)

hparra:
Use locale (ietf) from getConfig instead of from <html lang>.

Same type of change as #916. Needed to prevent adverse effects from https://github.com/adobecom/milo/pull/915. Discovered in https://github.com/adobecom/bacom/pull/130.

We want to maintain existing behavior for CaaS so I've left edge-cased lang/cntry split alone. I also tested using new locale values (lang/reg) but Chimera endpoint returns no cards when `country` parameter is `undefined` or unsupported, eg. new `la` value (419). This will have to be migrated later.

**Test URLs:**

- Before: https://main--milo--hparra.hlx.page/
- Before: https://main--bacom--adobecom.hlx.page/customer-success-stories?martech=off
- Before: https://main--bacom--adobecom.hlx.page/la/customer-success-stories?martech=off
- After: https://fix-caas-lang-reg--milo--hparra.hlx.page/
- After: https://main--bacom--adobecom.hlx.page/customer-success-stories?martech=off&milolibs=fix-caas-lang-reg--milo--hparra
- After: https://main--bacom--adobecom.hlx.page/la/customer-success-stories?martech=off&milolibs=fix-caas-lang-reg--milo--hparra

### DIFF
--- a/libs/blocks/caas/utils.js
+++ b/libs/blocks/caas/utils.js
@@ -248,12 +248,9 @@ const getFilterArray = async (state) => {
   return filters;
 };
 
-export function getCountryAndLang(
-  { autoCountryLang, country, language },
-  miloConfig = pageConfig,
-) {
+export function getCountryAndLang({ autoCountryLang, country, language }) {
   if (autoCountryLang) {
-    const htmlLang = miloConfig?.locale?.ietf?.toLowerCase() || 'en-us';
+    const htmlLang = pageConfigHelper()?.locale?.ietf?.toLowerCase() || 'en-us';
     const [lang, cntry] = htmlLang.split('-');
     return {
       country: cntry,
--- a/libs/blocks/caas/utils.js
+++ b/libs/blocks/caas/utils.js
@@ -250,9 +250,11 @@ const getFilterArray = async (state) => {
 
 const getCountryAndLang = ({ autoCountryLang, country, language }) => {
   if (autoCountryLang) {
+    const htmlLang = pageConfig.locale?.ietf;
+    const [lang, cntry] = htmlLang.split('-');
     return {
-      country: pageConfig?.locale?.reg,
-      language: pageConfig?.locale?.lang,
+      country: cntry,
+      language: lang,
     };
   }
   return {
--- a/libs/blocks/caas/utils.js
+++ b/libs/blocks/caas/utils.js
@@ -250,11 +250,9 @@ const getFilterArray = async (state) => {
 
 const getCountryAndLang = ({ autoCountryLang, country, language }) => {
   if (autoCountryLang) {
-    const htmlLang = document.documentElement.getAttribute('lang')?.toLowerCase() || 'en-us';
-    const [lang, cntry] = htmlLang.split('-');
     return {
-      country: cntry,
-      language: lang,
+      country: pageConfig?.locale?.reg,
+      language: pageConfig?.locale?.lang,
     };
   }
   return {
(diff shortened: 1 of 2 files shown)


meganthecoder:
Your test urls need a couple updates. For the "before" tests, they shouldn't include the new milolibs. For the "after" tests, they shouldn't include your patch branch. We need to make sure that these changes won't affect anything on production, so you should only be changing the milolibs branch, not the bacom branch. You can include the patch links additionally if you want.
Before: https://main--bacom--adobecom.hlx.page/customer-success-stories?martech=off
Before: https://main--bacom--adobecom.hlx.page/la/customer-success-stories?martech=off

After: https://main--bacom--adobecom.hlx.page/customer-success-stories?martech=off&milolibs=fix-caas-lang-reg--milo--hparra
After: https://main--bacom--adobecom.hlx.page/la/customer-success-stories?martech=off&milolibs=fix-caas-lang-reg--milo--hparra

hparra:
> Your test urls need a couple updates. For the "before" tests, they shouldn't include the new milolibs. For the "after" tests, they shouldn't include your patch branch. We need to make sure that these changes won't affect anything on production, so you should only be changing the milolibs branch, not the bacom branch. You can include the patch links additionally if you want.

Lazy copy/paste from https://github.com/adobecom/bacom/pull/130. Fixed. Thank you.

libs/blocks/caas/utils.js
```diff
@@ -248,9 +248,12 @@ const getFilterArray = async (state) => {
   return filters;
 };
 
-const getCountryAndLang = ({ autoCountryLang, country, language }) => {
+export function getCountryAndLang(
+  { autoCountryLang, country, language },
+  miloConfig = pageConfig,
```
meganthecoder, line 253: Are you only using the `miloConfig` param for the tests? If so, that's not a great practice. What about using `setConfig` in the test instead?

hparra: I think what you're really asking is for me to use `getConfig` in body of func and not the `pageConfig` global param that probably shouldn't exist. That's the only way a test with `setConfig` would work. The pattern I'm using here is the one that already exists in the code (see top of file). There was no test coverage for this func previously.

hparra: Fixed at [23223b4](https://github.com/adobecom/milo/pull/945/commits/23223b44ae00cd76f6de9b031c00adb79f2fcc8c)

meganthecoder (approved):

meganthecoder:
@hparra I can do the verification, but your milo links don't have caas on them. Can you update them so they demonstrate the code?

hparra:
> I can do the verification, but your milo links don't have caas on them. Can you update them so they demonstrate the code?

Not sure I understand. This is why I had the BACOM links. Is there an example of CaaS on Milo?

meganthecoder:
> > I can do the verification, but your milo links don't have caas on them. Can you update them so they demonstrate the code?
> 
> Not sure I understand. This is why I had the BACOM links. Is there an example of CaaS on Milo?

Yeah, I guess you don't have access to sharepoint to create a draft page. The only non-draft example I can think of is the caas configurator page, but that wouldn't really help you here. In that case, just having the bacom links is fine, but you should remove the milo links then since they're not accurate.

hparra:
> In that case, just having the bacom links is fine, but you should remove the milo links then since they're not accurate.

Ah! I'm doing this because it's necessary for `helix-psi-check` to run (properly). It checks that the URL maps to the repo, if not I get an error. Related: [MWPW-130837](https://jira.corp.adobe.com/browse/MWPW-130837).

> *Also, in my previous comment, I used the hlx.page bacom links. We should be using the hlx.live links because hlx.page isn't public. Just a note for next time.

👍

PR #953, opened 2023-07-18
[UAR] - Alternate quiz data

fullcolorcoder:
- Allows query param to drive json data sources 
- Refactored setting window.location.search params with layout effect for async 
- Fixes step indicator bug
 
Resolves:  [MWPW-126143](https://jira.corp.adobe.com/browse/MWPW-126143)
Resolves: [MWPW-131592](https://jira.corp.adobe.com/browse/MWPW-131592)

**Test URLs:**
Before: https://uar-integration--milo--adobecom.hlx.page/drafts/bradjohn/quiz-22/?martech=off
After: https://alternate-quiz-data--milo--fullcolorcoder.hlx.page/drafts/bradjohn/quiz-22/?quiz-data=alternate-data&martech=off

### DIFF
--- a/libs/blocks/quiz/utils.js
+++ b/libs/blocks/quiz/utils.js
@@ -11,13 +11,13 @@ const { locale } = getConfig();
 
 export const initConfigPath = (rootElm) => {
   const link = rootElm.querySelector('.quiz > div > div > a');
-  const quizConfigPath = link?.text.toLowerCase();
+  const quizConfigPath = link?.text?.toLowerCase() || '';
   const urlParams = new URLSearchParams(window.location.search);
   const stringsPath = urlParams.get('quiz-data');
   if (stringsPath) {
     return (filepath) => `${stringsPath}/${filepath}`;
   }
-  return (filepath) => `${quizConfigPath}${filepath}`;
+  return (filepath) => `${quizConfigPath}/${filepath}`;
 };
 
 const initQuizKey = () => {
--- a/libs/blocks/quiz/utils.js
+++ b/libs/blocks/quiz/utils.js
@@ -311,6 +311,7 @@ export const handleNext = (questionsData, selectedQuestion, userInputSelections,
   const allcards = Object.keys(userInputSelections);
   let nextQuizViews = [];
   let hasResultTigger = false;
+  let lastStopValue;
 
   allcards.forEach((selection) => {
     // for each elem in current selection, find its coresponding
@@ -342,11 +343,11 @@ export const handleNext = (questionsData, selectedQuestion, userInputSelections,
             nextQuizViews = nextQuizViews.filter((val) => val !== skip);
           });
         }
-
         // RESET the queue and add only the next question.
         if (flowStepsList.includes('RESET')) { // Reset to intial question
           nextQuizViews = []; // Resetting the nextQuizViews
           userFlow = []; // Resetting the userFlow as well
+          lastStopValue = 'RESET';
         }
 
         if (!hasResultTigger) {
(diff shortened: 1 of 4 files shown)


libs/blocks/quiz/utils.js
```diff
@@ -9,9 +9,14 @@ const RESULTS_EP_NAME = 'results.json';
 let configPath; let quizKey; let analyticsType; let analyticsQuiz; let metaData;
 const { locale } = getConfig();
 
-const initConfigPath = (roolElm) => {
-  const link = roolElm.querySelector('.quiz > div > div > a');
-  const quizConfigPath = link?.text.toLowerCase();
+export const initConfigPath = (rootElm) => {
+  const link = rootElm.querySelector('.quiz > div > div > a');
+  const quizConfigPath = link?.text?.toLowerCase() || '';
```
echampio-at-adobe, line 14: It's a preexisting condition, but reading quizconfigpath from the first arbitrary link we find is pretty fragile. We should use the quizurl value from the existing global metadata object. Also, since we have this existing parameter in the quiz block, I think the parameter name that we use in the URL to override it should be the same. We should use quizurl all the way through, rather than quizurl and quiz-data. 

Screenshot 2023-07-19 at 9 37 01 AM

libs/blocks/quiz/utils.js
```diff
@@ -9,9 +9,13 @@ const RESULTS_EP_NAME = 'results.json';
 let configPath; let quizKey; let analyticsType; let analyticsQuiz; let metaData;
 const { locale } = getConfig();
 
-const initConfigPath = (roolElm) => {
-  const link = roolElm.querySelector('.quiz > div > div > a');
-  const quizConfigPath = link?.text.toLowerCase();
+export const initConfigPath = (quizMetaData) => {
```
JackySun9, line 12: I got enhance code from AI like this, do we really need the validation here or we are good?
/**
 * Initializes the configuration path based on the given quizMetaData.
 *
 * @param {Object} quizMetaData - The metadata of the quiz.
 * @returns {Function} The function to get the path of the file.
 */
export const initConfigPath = (quizMetaData) => {
  if (!quizMetaData || !quizMetaData.quizurl || typeof quizMetaData.quizurl.text !== 'string') {
    throw new Error('Invalid quizMetaData. Make sure it has a "quizurl.text" property of type string.');
  }

  const quizConfigPath = quizMetaData.quizurl.text.toLowerCase();
  const urlParams = new URLSearchParams(window.location.search);
  const stringsPath = urlParams.get('quiz-data');

  return (filepath) => {
    if (!filepath || typeof filepath !== 'string') {
      throw new Error('Invalid filepath. It should be a non-empty string.');
    }

    const baseConfigPath = stringsPath || quizConfigPath;
    return `${baseConfigPath}${filepath}`;
  };
};

JackySun9: and When you use quizMetaData, it is the same result as before, right?

JackySun9: I close this after sync with Erich, because we will have a story for the validation part

echampio-at-adobe (approved):
Works with other path variations as well:

https://alternate-quiz-data--milo--fullcolorcoder.hlx.page/drafts/bradjohn/quiz-22/?quiz-data=/drafts/delta/app-recommender-test&martech=off

https://alternate-quiz-data--milo--fullcolorcoder.hlx.page/drafts/bradjohn/quiz-22/?quiz-data=https://alternate-quiz-data--milo--fullcolorcoder.hlx.page/drafts/delta/app-recommender-test&martech=off

PR #955, opened 2023-07-19
Add title-append feature

hparra:
The title-append feature allows you to append an arbitrary string to an HTML document's title by populating a "title-append" column for any given row in [Franklin's bulk-metadata](https://www.hlx.live/docs/bulk-metadata). See README for more details.

Resolves: [MWPW-132671](https://jira.corp.adobe.com/browse/MWPW-132671)

**Test URLs:**
- Before: https://main--milo--hparra.hlx.page/drafts/hgpa/test/title-append/title-append?martech=off
- After: https://title-append--milo--hparra.hlx.page/drafts/hgpa/test/title-append/title-append?martech=off

Tip: You can see change happen in realtime -- open the "After" test URL in it own tab so you can see page title and refresh.

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -844,8 +844,7 @@ export async function loadArea(area = document) {
     }
     const appendage = getMetadata('title-append');
     if (appendage) {
-      const { default: titleAppend } = await import('../features/title-append/title-append.js');
-      titleAppend(appendage);
+      import('../features/title-append/title-append.js').then((module) => module.default(appendage));
     }
     const richResults = getMetadata('richresults');
     if (richResults) {
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -842,6 +842,11 @@ export async function loadArea(area = document) {
       const { default: loadGeoRouting } = await import('../features/georoutingv2/georoutingv2.js');
       loadGeoRouting(config, createTag, getMetadata, loadBlock, loadStyle);
     }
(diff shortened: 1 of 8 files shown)


fullcolorcoder (approved):

libs/utils/utils.js
```diff
@@ -840,6 +840,11 @@ export async function loadArea(area = document) {
       const { default: loadGeoRouting } = await import('../features/georoutingv2/georoutingv2.js');
       loadGeoRouting(config, createTag, getMetadata, loadBlock, loadStyle);
     }
+    const appendage = getMetadata('title-append');
+    if (appendage) {
+      const { default: titleAppend } = await import('../features/title-append/title-append.js');
+      titleAppend(appendage);
```
chrischrischris, line 846: In this case should we use a `.then` instead of `await` since there are no downstream dependencies?
```suggestion
      import('../features/title-append/title-append.js')
        .then((module) => module.default(appendage));
```

chrischrischris (approved):
Noodle!  One small change otherwise 👍

libs/utils/utils.js
hparra: That would make tests flake. I would have to check.

hparra: Fixed at [e1ff268](https://github.com/adobecom/milo/pull/955/commits/e1ff268924ad2945a8e09644efd29c2791c4bd26)

PR #959, opened 2023-07-20
Mwpw 132793 Icon and Quotes Blocks RTL Alignment Issue.

sharath-kannan:
* Icon and quote block rtl bug fix
Resolves: [MWPW-132793](https://jira.corp.adobe.com/browse/MWPW-132793)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/ae_ar/drafts/shrinidhivp/icon-blocks-1?akamaiLocale=en&martech=off
- After: https://mwpw-132793--milo--sharath-kannan.hlx.page/ae_ar/drafts/shrinidhivp/icon-blocks-1?akamaiLocale=en&martech=off

- Before: https://main--milo--adobecom.hlx.page/eg_ar/drafts/tat28144/blocks/quote?akamaiLocale=en&martech=off
-After: https://mwpw-132793--milo--sharath-kannan.hlx.page/eg_ar/drafts/tat28144/blocks/quote?akamaiLocale=en&martech=off


chrischrischris:
@sharath-kannan Please update the PR title to be more descriptive, thanks!

libs/blocks/icon-block/icon-block.css
```diff
@@ -42,6 +42,16 @@
   width: 100%;
 }
 
+.five-up .foreground .text-content {
```
chrischrischris, line 45: This is not specific to .icon-block - so this might affect other parts of the page.  Can you scope it to the icon block?

chrischrischris: just saw that it's existing css code that was moved, but I'm still curious as to why

sharath-kannan: The code was moved up to resolve StyleLint errors.

Ruchika4:
@sharath-kannan Have we discussed this ticket with GWP as requested in related discussion here https://github.com/orgs/adobecom/discussions/878 ?

sharath-kannan:
@Ruchika4 

> > @sharath-kannan Have we discussed this ticket with GWP as requested in related discussion here https://github.com/orgs/adobecom/discussions/878 ?

libs/blocks/icon-block/icon-block.css
```diff
@@ -168,6 +178,10 @@
   margin-bottom: var(--spacing-s);
 }
 
+html:not([dir='rtl']).icon-block .foreground .text-content p[class*="body-m"]:last-of-type {
```
overmyheadandbody, line 181: This would target `html` tags that have an `icon-block` class name, I'm not sure this is the desired behaviour here.

libs/blocks/icon-block/icon-block.css
```diff
@@ -42,6 +42,16 @@
...
 }
 
+.five-up .foreground .text-content {
+  text-align: left;
+  padding-bottom: var(--spacing-xs);
+}
+
+
+html[dir='rtl'] .icon-block .foreground .text-content {
+  align-content: right;
```
overmyheadandbody, line 52: `right` is not a valid property for `align-content` - https://developer.mozilla.org/en-US/docs/Web/CSS/align-content. Looking at previous declarations, maybe you wanted to use `text-align`?

libs/blocks/icon-block/icon-block.css
sharath-kannan: Block has been removed after reviewing and re-evaluating the code.

libs/blocks/icon-block/icon-block.css
sharath-kannan: This has been fixed, however the whole block was needed.

libs/blocks/icon-block/icon-block.css
```diff
@@ -346,6 +346,10 @@
   gap: 16px;
 }
 
+html[dir='rtl'] .icon-block.inline .foreground .text-content {
```
overmyheadandbody, line 349: L200 uses double quotes, this and the declaration in quote.css use single quotes. I'd suggest we stick to just one of them.

On another note, is the `html` tag selector actually necessary? We did have situations in the past where certain components had their own `dir` attribute, different than that of the `html` tag. So removing the `html` saves a few characters, should have the same effect and it might prove useful for some future use-cases.

overmyheadandbody (approved):

libs/blocks/icon-block/icon-block.css
sharath-kannan: okay,I have changed it to double quotes.
However i have noticed that all the tags do not have the dir tag set to rtl or ltr, but i still have the need to differentiate and set specific rules for both of the alignments and hence i have used html to differentiate.

libs/blocks/icon-block/icon-block.css
chrischrischris: All of the other `.five-up` selectors are still scoped to the icon block - e.g:
`.five-up .icon-block.fullwidth {`
but this one isn't, and could affect parts of the page unrelated to the icon-block

libs/blocks/icon-block/icon-block.css
```diff
@@ -346,6 +346,10 @@
   gap: 16px;
 }
 
+html[dir="rtl"] .icon-block.inline .foreground .text-content {
+  text-align: right;
+}
+
```
rgclayton, line 352: I think a better approach to all of this is [to correct the usage of `text-align: left`](https://github.com/adobecom/milo/blob/7c206180637964553aaa4e59ba6701b93a694fbf/libs/blocks/icon-block/icon-block.css#L345) to be `text-align: start`

That way when the `dir` attribute is applied the direction is automatically adjusted based on `dir`.
https://developer.mozilla.org/en-US/docs/Web/CSS/text-align

sharath-kannan: @rgclayton I have updated the code.

libs/blocks/icon-block/icon-block.css
sharath-kannan: @chrischrischris i have scoped this to icon-block.

chrischrischris (approved):

PR #963, opened 2023-07-20
Use "content-language" metadata for html lang attribute

hparra:
Use "content-language" metadata value as highest priority option for html lang attribute. This allows `<html lang>` to be driven from metadata.xlsx and avoid coupling with `locales`. Named "content-language" because it is similar to adjacent "content-direction" feature. 

Resolves: [MWPW-133115](https://jira.corp.adobe.com/browse/MWPW-133115)

**Test URLs:**
- Before: https://main--milo--hparra.hlx.live/drafts/hgpa/test/content-language/content-language?martech=off
- After: https://lang-metadata--milo--hparra.hlx.live/drafts/hgpa/test/content-language/content-language?martech=off

PR #965, opened 2023-07-20
Soft revert of lang/reg #915

hparra:
Soft revert of #915.

Resolves: [MWPW-133741](https://jira.corp.adobe.com/browse/MWPW-133741)

The initial driving feature was implemented without using these additions (#955) and the ietf-value bug that was found during dev will be resolved using #963 instead.

**Test URLs:**
- After: https://revert-lang-reg--milo--hparra.hlx.live/?martech=off


libs/scripts/delayed.js
```diff
@@ -63,7 +63,7 @@ const loadDelayed = ([
     if (getMetadata('interlinks') === 'on') {
       const { locale } = getConfig();
       const path = `${locale.contentRoot}/keywords.json`;
-      const language = locale.lang ?? locale.ietf?.split('-')[0];
```
hparra, line 66: @Ruchika4 I think you are the only other person to use this feature. Is this ok?

Ruchika4: Yes this looks good.

hparra:
@chrischrischris @Ruchika4 Asking for your reviews since you approved original PR.

chrischrischris (approved):

rgclayton (approved):

PR #984, opened 2023-07-21
SKMS integration

amauch-adobe:
- Adds auto-approved CMR integration on Milo releases for SKMS.
- New GitHub action for merged pull requests against the production branch.

Resolves: [MWPW-131002](https://jira.corp.adobe.com/browse/MWPW-131002)

**Test URLs:**
- After: https://skms-integration--milo--adobecom.hlx.page/?martech=off


mokimo (approved):
Since this should also be reporting merged PRs of consumer projects, I assume every consumer needs to replicate this workflow?

amauch-adobe:
> Since this should also be reporting merged PRs of consumer projects, I assume every consumer needs to replicate this workflow?

Yes, after this is merged into Milo we will move it to Milo College's repository so that new projects will need to use it.  You are correct it will need to be moved into consumer projects as well.

jedjedjedM:
@amauch-adobe Can you remove some of the boilerplate comments such as:
_# This is a basic workflow to help you get started with Actions_

JasonHowellSlavin (approved):

amauch-adobe:
> @amauch-adobe Can you remove some of the boilerplate comments such as: _# This is a basic workflow to help you get started with Actions_

@jedjedjedM thanks for the remark, I have it updated now.

.github/workflows/skms.yaml
```diff
@@ -0,0 +1,83 @@
...
+
+            curl https://${url}/web_api --data-urlencode "_parameters=${params}"
+            echo ''
+          }
+
+          end_date=$(TZ=":America/Los_Angeles" date '+%Y-%m-%d %H:%M')
+          release_details="${{ github.event.pull_request.body }}"
+          release_details="${release_details//$'\r\n'/ }"
+
+          skms_request '${{ secrets.SKMS_USER }}' '${{ secrets.SKMS_STAGE_PASS }}' CmrDao createCmrFromPreapprovedChangeModel '{"change_executor":"${{ github.actor }}","maintenance_window_end_time":"${end_date}","maintenance_window_start_time":"${start_date}","preapproved_change_model_id":"847787","summary":"Release Title:${{ github.event.pull_request.title }}::${release_details}::Pull Request #${{ github.event.pull_request.number }}::Pull Request Created at:${{ github.event.pull_request.created_at }}::Pull Request Merged at:${{ github.event.pull_request.merged_at }}"}'
```
jedjedjedM, line 83: _secrets.SKMS_STAGE_PASS_
I assume this is the prod password?  If so, the name should reflect that.

amauch-adobe: @jedjedjedM fixed, thanks for catching that one.  I switched the API endpoint but not the secrets, it's fixed now.

jedjedjedM:
If the PR title has a quote ("), will that potentially break the script?

.github/workflows/skms.yaml
```diff
@@ -0,0 +1,83 @@
...
+            fi
+
+            if ! type "curl" >/dev/null; then
+              echo "curl is required but not installed"
+            fi
+
+            local params="{\"_username\":\"${username}\",\"_passkey\":\"${passkey}\",\"_object\":\"${object}\",\"_method\": \"${method}\"}"
+            params=$(echo "$params $method_params" | jq -s add)
+
+            curl https://${url}/web_api --data-urlencode "_parameters=${params}"
```
JackySun9, line 75: Maybe you can do this to handle errors:
response=$(curl --write-out "%{http_code}" --silent --output /dev/null https://${url}/web_api --data-urlencode "_parameters=${params}")
            if [ "$response" -ne 200 ]; then
              echo "SKMS API returned HTTP status $response"
              exit 1
            fi

.github/workflows/skms.yaml
```diff
@@ -0,0 +1,83 @@
...
+
+            if [ -z "$method_params" ]; then
+              method_params='{}'
+            fi
+
+            if [ -z "$url" ]; then
+              url=$DEFAULT_SKMS_URL
+            fi
+
+            if ! type "jq" >/dev/null; then
```
JackySun9, line 64: You can move this part to check dependencies before you run this function:
# Check dependencies
      - name: Check dependencies
        run: |
          if ! type "jq" >/dev/null; then
            echo "jq is required but not installed"
            exit 1
          fi
          if ! type "curl" >/dev/null; then
            echo "curl is required but not installed"
            exit 1
          fi

.github/workflows/skms.yaml
```diff
@@ -0,0 +1,83 @@
...
+
+          function skms_request() {
+            local username=$1
+            local passkey=$2
+            local object=$3
+            local method=$4
+            local method_params=$5
+            local url=$6
+
+            if [ -z "$username" ]; then
```
JackySun9, line 36: If you handle the errors as below, you don't need these "if"s

JackySun9 (changes requested):

.github/workflows/skms.yaml
amauch-adobe: The if statements are there to catch the errors before you even try to send the curl statement, they are required for it to be able to be sent.  That being said I do like the idea of having a wrapper to retain the response back and to properly provide an exit code instead of just firing off the execution and expecting it to work.  I'll add it anyways for that reason.  Thanks, @JackySun9.

.github/workflows/skms.yaml
amauch-adobe: see my comment above.

.github/workflows/skms.yaml
```diff
@@ -0,0 +1,101 @@
...
+            echo "curl is required but not installed"
+            exit 1
+          fi
+
+          echo "Dependencies check was successful"
+
+      # Run command for creating CMR in SKMS system
+      - run: |
+          start_date=$(TZ=":America/Los_Angeles" date '+%Y-%m-%d %H:%M')
+          DEFAULT_SKMS_URL='api.skms.adobe.com'
```
TsayAdobe, line 41: For testing, it's better to use the stage env `stage.dev.skms.adobe.com`.

amauch-adobe: I've switched to Production on purpose since there is an error that is blocking these machines with the production environment.  Thanks though for the reminder.

jedjedjedM (approved):

.github/workflows/skms.yaml
```diff
@@ -0,0 +1,115 @@
...
+            str="${str//$'\r\n\t'/ }"
+
+            echo "$str"
+          }
+
+          end_date=$(TZ=":America/Los_Angeles" date '+%Y-%m-%d %H:%M')
+          release_title=$(sanitizeStr "${{ github.event.pull_request.title }}")
+          release_details=$(sanitizeStr "${{ github.event.pull_request.body }}")
+
+          skms_request '${{ secrets.SKMS_USER }}' '${{ secrets.SKMS_PASS }}' CmrDao createCmrFromPreapprovedChangeModel '{"change_executor":"${{ github.actor }}","maintenance_window_end_time":"${end_date}","maintenance_window_start_time":"${start_date}","preapproved_change_model_id":"847787","summary":"Release Title:${release_title}::${release_details}::Pull Request Number:${{ github.event.pull_request.number }}::Pull Request Created at:${{ github.event.pull_request.created_at }}::Pull Request Merged at:${{ github.event.pull_request.merged_at }}"}'
```
JackySun9, line 115: hardcode for preapproved_change_model_id

JackySun9 (changes requested):

.github/workflows/skms.yaml
amauch-adobe: @JackySun9, every team will have to create their own preapproved change model id and then add it to their site's repo for this github action.  This one is strictly for Milo so having the hardcoded value is okay.  I will be adding this to Milo College's repo once this is merged and working with Milo's repository and then the one in Milo College will be the one other sites currently on Milo and future sites will need to pull and update accordingly.  That is where I will remove the hardcoded id and leave it as a placeholder.  There really isn't a use case for a specific site repo to have different change model id's since it will fall under the same service/application that is having a change/release happening.  Each change model in the SKMS system is supposed to associate what services/apps are affected or potentially affected by a release coming from the change/deployment.

JackySun9: ok

JackySun9 (approved):

amauch-adobe:
I'm going to close this PR for now and reopen it when it is ready again for testing.

PR #986, opened 2023-07-21
Fix flakey scripts.test.js

hparra:
The `Decorating > Loads lana.js upon calling lana.log the first time` test is a common offender.

Both Milo's `loadArea` and LANA's `lana.js` are IIFEs, which means that awaiting a file's import does not necessarily mean that the necessary deps have loaded. We add a `loaded` param to LANA and use `waitFor` test helper. We also use `waitForElement` for other tests.

Addresses: [MWPW-129653](https://jira.corp.adobe.com/browse/MWPW-129653)

**Test URLs:**
- After: https://fix-scripts-test--milo--hparra.hlx.live/?martech=off


chrischrischris (approved):

sukamat (approved):

test/scripts/scripts.test.js
```diff
@@ -31,10 +38,11 @@ describe('Decorating', () => {
   });
 
   it('Loads lana.js upon calling lana.log the first time', async () => {
+    await delay(200); // wait for non-blocking imports in loadArea()
```
mokimo, line 41: Can we prevent the delay by having an intersection observer, or waiting until window.lana.log actually exists?
This might still pop up as failing on a slow CI day otherwise

mokimo (approved):
👍

libs/utils/lana.js
```diff
@@ -111,6 +111,7 @@
     debug: false,
     log,
     options: mergeOptions(w.lana && w.lana.options),
+    loaded: true,
```
mokimo, line 114: Does it need this? IMO it should be enough to tell if `window.lana` exists, if it does you can assume it has loaded

hparra: This is not true. Try it. `window.lana` is defined twice, once in `loadLana` again in `lana.js`. The file load only happens on call to log, but the iife inside `lana.js` may not have finished yet.

test/scripts/scripts.test.js
```diff
@@ -31,10 +38,11 @@ describe('Decorating', () => {
   });
 
   it('Loads lana.js upon calling lana.log the first time', async () => {
+    await waitFor(() => window.lana?.loaded);
```
chrischrischris, line 41: This is only working due to a bug in the test - lana.log hasn't been called explicitly.  It is called when catching a ReferenceError thrown by the martech script.  That needs to be fixed/removed so that this test runs correctly.

I think we can remove the `loaded` property, and instead do a reference check after calling lana the first time (once above is fixed):
```
expect(window.lana.log).to.exist;
const initialLana = window.lana.log;
sinon.spy(console, 'log');
await window.lana.log('test', { clientId: 'myclient', sampleRate: 0 });
await waitFor(() => initialLana !== window.lana.log);
```

I will look into the martech issue tomorrow.

test/helpers/waitfor.js
```diff
@@ -103,3 +103,28 @@ export const delay = (timeOut, cb) => new Promise((resolve) => {
...
 });
+
+/**
+ * Waits for predicate function to be true or times out.
+ * @param {function} predicate Callback that returns boolean
+ * @param {number} timeout Timeout in milliseconds
+ * @param {number} interval Interval delay in milliseconds
+ * @returns {Promise}
+ */
+export function waitFor(predicate, timeout = 1000, interval = 100) {
```
chrischrischris, line 114: 🎉

chrischrischris (changes requested):

test/scripts/scripts.test.js
hparra: OK. So you will fix the martech issue before we can update this?

chrischrischris:
Using https://github.com/adobecom/milo/pull/1012 to fix this issue.

PR #987, opened 2023-07-21
Autofix subset of eslint errors

hparra:
Result of `eslint . --fix`

Since I'm touching so many files, the "codecov patch" check fails when those files have insufficient tests.

There are still 120+ errors and 70+ warnings. Persistent errors include:

- import/no-cycle
- import/no-named-default
- import/prefer-default-export
- import/no-unresolved
- no-redeclare
- no-unused-vars
- no-continue
- no-param-reassign
- no-use-before-define
- etc.

Addresses: [MWPW-134063](https://jira.corp.adobe.com/browse/MWPW-134063)

**Test URLs:**
- After: https://chore-eslint-yikes--milo--hparra.hlx.live/?martech=off


mokimo (approved):
We might want to consider new defaults for things that we can't or don't want to fix/agree on, which we currently still have warnings & errors for.

JasonHowellSlavin (commented):
Looks good, agree with @mokimo as there should likely be a review of the linter rules.

JasonHowellSlavin (approved):

chrischrischris (approved):

PR #990, opened 2023-07-24
Make sure to always display the breadcrumbs when a base is set

mokimo:
### Description
Ensure to show the breadcrumbs element if there is just a "base-breadcrumbs"

Resolves: [MWPW-134085](https://jira.corp.adobe.com/browse/MWPW-134085)

**Test URLs:**
- Before: - https://mwpw-134085--milo--mokimo.hlx.live/
- Before: https://main--bacom--adobecom.hlx.page/resources/leader-in-experiences-breadcrumb-in-url-test
- After: https://main--bacom--adobecom.hlx.page/resources/leader-in-experiences-breadcrumb-in-url-test?milolibs=mwpw-134085--milo--mokimo

### DIFF
--- a/libs/blocks/global-navigation/features/breadcrumbs/breadcrumbs.js
+++ b/libs/blocks/global-navigation/features/breadcrumbs/breadcrumbs.js
@@ -78,7 +78,8 @@ const createWithBase = async (element = toFragment`<div><ul></ul></div>`) => {
     const base = new DOMParser().parseFromString(text, 'text/html').body;
     element.querySelector('ul')?.prepend(...base.querySelectorAll('li'));
     return createBreadcrumbs(element);
-  } catch (error) {
+  } catch (e) {
+    lanaLog({ e, message: 'Breadcrumbs failed fetching base' });
     return null;
   }
 };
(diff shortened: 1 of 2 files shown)


libs/blocks/global-navigation/features/breadcrumbs/breadcrumbs.js
```diff
@@ -99,7 +100,9 @@ const fromUrl = () => {
 
 export default async function init(el) {
   try {
-    const breadcrumbsEl = (await createWithBase(el)) || createBreadcrumbs(el) || fromUrl();
+    const breadcrumbsEl = (await createWithBase(el || undefined))
```
mokimo, line 103: Small note since this line might seem weird: In the global-navigation.js we init with `init(this.el.querySelector('.breadcrumbs'))` 

An empty querySelector will return null, which in this case will overwrite the default parameter we set for the `createWithBase` - hence we need the "undefined" in the case that `el` is actually null.

JasonHowellSlavin (approved):

libs/utils/utils.js
```diff
@@ -564,9 +564,10 @@ function decorateHeader() {
   const metadataConfig = getMetadata('breadcrumbs')?.toLowerCase()
   || getConfig().breadcrumbs;
   if (metadataConfig === 'off') return;
+  const baseBreadcrumbs = getMetadata('breadcrumbs-base')?.length;
```
narcis-radu, line 567: might be over engineering, but should we check whether the content is an actual URL?

narcis-radu (approved):

overmyheadandbody (approved):
Thanks for implementing the suggestion, I think this is a bit cleaner 👍

libs/utils/utils.js
mokimo: Let's trust authors for now and keep an eye on the lana logs

SilviuLCF (approved):

PR #991, opened 2023-07-25
MWPW-131707 Get correct region for caas autolang

chrischrischris:
* Correctly determine country for caas auto-country/lang

Resolves: [MWPW-131707](https://jira.corp.adobe.com/browse/MWPW-131707)

**Test URLs:**
- Bacom Before:
  - No Thai cards displayed
  - https://main--bacom--adobecom.hlx.page/th_th/resources/main
- Bacom After:
	- Thai cards displayed	
	- https://main--bacom--adobecom.hlx.page/th_th/resources/main?milolibs=caaslocale
- After: https://caaslocale--milo--adobecom.hlx.page/?martech=off


Ruchika4 (approved):

mokimo (approved):

mokimo:
FYI a caas test is failing, that needs to be adapted before merging

libs/utils/utils.js
```diff
@@ -146,6 +146,7 @@ export function getLocale(locales, pathname = window.location.pathname) {
     return locale;
   }
   locale.prefix = locale.ietf === 'en-US' ? '' : `/${localeString}`;
+  [locale.region] = localeString.split('_');
```
overmyheadandbody, line 149: Does this cover all use-cases? Say we have these pathname examples:

1. `/products/photoshop`
2. `/ch_de/products/photoshop`
3. `/ro/products/photoshop`
4. `/`

Looking at the code, the full formula for `locale.region` seems to be `pathname.split('/')[1].split('_')[0]`. For the use-cases above, the results would be:

1. `'products'`
2. `'ch'`
3. `'ro'`
4. `''`

So to me it seems that examples 2 & 3 are working as expected, but I'm not so sure about 1 & 4. 4 might be ok for the purposes of `caas/utils.js`, but 1 still stands out. What would be the desired outcome?

overmyheadandbody (commented):
I second @mokimo 's observation around unit tests and have a quick question.

libs/utils/utils.js
chrischrischris: Thanks for catching that.  Fixed.

PR #995, opened 2023-07-25
[MWPW-133763] Breadcrumbs mobile alignment

overmyheadandbody:
This is a slight style enhancement for the mobile breadcrumbs. It ensures:
- the ellipsis element has equal spacing on both desktop browsers in responsive view and actual devices when browsing on Firefox or Safari;
- the last breadcrumb has lateral space if the breadcrumbs section is scrollable.

| Before | After |
| Ellipsis spacing on FF - before | Ellipsis spacing on FF - after |
| Last element spacing on FF - before | Last element spacing on FF - after |

Resolves: [MWPW-133763](https://jira.corp.adobe.com/browse/MWPW-133763)

**Test URLs:**
- Before: https://main--milo--overmyheadandbody.hlx.page/drafts/ramuntea/footer-refactor-variant?martech=off
- After: https://breadcrumbs-mobile-spacing--milo--overmyheadandbody.hlx.page/drafts/ramuntea/footer-refactor-variant?martech=off

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.css
+++ b/libs/blocks/global-navigation/global-navigation.css
@@ -320,6 +320,15 @@ header.global-navigation {
   flex-shrink: 0;
 }
 
+.feds-breadcrumbs li:last-child:not(:first-child) {
+  padding-right: 12px;
+}
+
+[dir = "rtl"] .feds-breadcrumbs li:last-child:not(:first-child) {
+  padding-right: 0;
+  padding-left: 12px;
+}
+
 /* Hide all breadcrumbs except the first and last two */
 .feds-breadcrumbs li:nth-last-child(n+3):not(:first-child) {
   display: none;
@@ -327,7 +336,7 @@ header.global-navigation {
 
 /* If first breadcrumb is not third to last, add ellipsis after it */
 .feds-breadcrumbs li:first-child:not(:nth-last-child(-n+3)):after {
-  content: '/\2003...';
+  content: '/\3000…';
   padding: 0 0 0 12px;
 }
 


mokimo (approved):
Nice enhancementè

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -320,14 +320,23 @@ header.global-navigation {
...
+
 /* Hide all breadcrumbs except the first and last two */
 .feds-breadcrumbs li:nth-last-child(n+3):not(:first-child) {
   display: none;
 }
 
 /* If first breadcrumb is not third to last, add ellipsis after it */
 .feds-breadcrumbs li:first-child:not(:nth-last-child(-n+3)):after {
-  content: '/\2003...';
+  content: '/\3000…';
```
hparra, line 339: TIL https://en.wikipedia.org/wiki/Template:Whitespace_(Unicode)

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -320,14 +320,23 @@ header.global-navigation {
   flex-shrink: 0;
 }
 
+.feds-breadcrumbs li:last-child:not(:first-child) {
+  padding-right: 12px;
+}
+
+[dir = "rtl"] .feds-breadcrumbs li:last-child:not(:first-child) {
```
hparra, line 327: Was there a test URL example for this one?

hparra (approved):
Confirmed LTR example in Safari. Would be nice to double-check RTL.

narcis-radu (approved):

libs/blocks/global-navigation/global-navigation.css
overmyheadandbody: I had to research it too. The ideographic space seems to be the only one which is consistent on all browsers, so went for that one

libs/blocks/global-navigation/global-navigation.css
overmyheadandbody: We don't yet have any RTL locales ported over to the FEDS Gnav yet, I test by changing the attribute manually for the moment. But to your point, we always check RTL locales when making changes.

PR #1004, opened 2023-07-26
Mwpw 134073 table alignment fixed in mobile windows.

sharath-kannan:
*Min-width for mobile window has been added. 

Resolves: [MWPW-134073](https://jira.corp.adobe.com/browse/MWPW-134073)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/shrinidhivp/tables-test
- After: https://mwpw-134073--milo--sharath-kannan.hlx.page/drafts/shrinidhivp/tables-test?martech=off


libs/blocks/table/table.css
```diff
@@ -443,3 +443,19 @@ header.global-navigation {
   width: 30px;
   margin-right: 16px;
 }
+
+/* Start mobile styles */
+@media (max-width: 400px) {
+  .section{
```
mokimo, line 449: .section sounds really generic, I'm not sure the table should set something like that

sharath-kannan: Have changed it to a specific selector now.

libs/blocks/table/table.css
```diff
@@ -443,3 +443,16 @@ header.global-navigation {
   width: 30px;
   margin-right: 16px;
 }
+
+/* Start mobile styles */
```
chrischrischris, line 447: No need for this comment

chrischrischris (approved):

libs/blocks/table/table.css
sharath-kannan: done.

mokimo (approved):

PR #1005, opened 2023-07-26
Fix flakey title-append test

hparra:
Fix my own flakey title-append test using `waitFor` introduced in #986.

Using an explicit delay is a last resort that we can avoid here.

**Test URLs:**
- After: https://fix-title-append-flap--milo--hparra.hlx.live/?martech=off

### DIFF
--- a/test/helpers/waitfor.js
+++ b/test/helpers/waitfor.js
@@ -103,3 +103,28 @@ export const delay = (timeOut, cb) => new Promise((resolve) => {
     resolve((cb && cb()) || null);
   }, timeOut);
 });
+
+/**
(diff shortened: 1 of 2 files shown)


hparra:
UT fails due to "marquee > supports media credits > has a media credit with element content" that is broken on main.

chrischrischris (approved):

chrischrischris:
FYI - the failing test is not related to this PR and is fixed by skipping in this PR:
https://github.com/adobecom/milo/pull/1006

chrischrischris (approved):

vgoodric (approved):

PR #1006, opened 2023-07-26
Disable failing media credit test until it can be fixed

chrischrischris:
No details required.

Test URL:
- https://disablemediacredit--milo--adobecom.hlx.page/


hparra (approved):
@chrischrischris I keep forgetting that disabling test is an option. I avoid it because this will affect coverage, BUT notice in this case the codecov check DID NOT run, even though they should, because skipping the test would change coverage.

JasonHowellSlavin (approved):
Is there a ticket we can link to that can track fixing this test?

hparra:
CC @williambsm

PR #1008, opened 2023-07-26
MWPW-133232 Remove default Article tag

chrischrischris:
Previously if a page had no caas tags defined, it would add a default Article tag.  That has now been removed.

Also if the page does not have any tags defined it will not be pushed to CaaS.

Resolves: [MWPW-133232](https://jira.corp.adobe.com/browse/MWPW-133232)

**Test URLs:**
- After: https://removedefaulttag--milo--adobecom.hlx.page/?martech=off


mokimo (approved):
Should we add a eslint ignore for the `console.log`, since we want to see the error for the tool?

tools/sidekick/config.json
```diff
@@ -42,6 +42,14 @@
       "excludePaths": [ "/**" ],
       "includePaths": [ "**/:x**" ]
     },
+    {
+      "containerId": "tools",
+      "title": "Send to CaaS",
+      "id": "sendtocaas",
+      "environments": ["dev","preview", "live", "prod"],
```
overmyheadandbody, line 49: Smallest nit in the universe, space between `dev` and `preview`

overmyheadandbody (approved):

PR #1009, opened 2023-07-26
Improve linting: only warn on cycles, ignore console in tests

hparra:
Addresses: [MWPW-134063](https://jira.corp.adobe.com/browse/MWPW-134063)

Only warn on cycles, ignore console in tests and add `npm run lint:js:nibble`.

We have some deep dependency cycles, but expecting others to fix these as part of their PR is too much -- they already exist and can be difficult to unravel. We warn instead of error.

Ignore `console` in tests. These are usually needed to test that an error was sent to LANA.

Add `npm run lint:js:nibble`

Screen Shot 2023-07-26 at 12 54 58

**Test URLs:**
- After: https://chore-eslint-improve--milo--hparra.hlx.live/?martech=off

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,6 +12,7 @@ module.exports = {
   rules: {
     'chai-friendly/no-unused-expressions': 2,
     'import/extensions': ['error', { js: 'always' }],
+    'import/no-cycle': 1,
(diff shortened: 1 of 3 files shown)


chrischrischris (approved):
👍

sanrai (approved):

PR #1010, opened 2023-07-26
Chore: lana.js linting

hparra:
Fix eslint errors/warnings found in lana.js via #986

```
hgpa@noface milo % npx eslint-nibble libs/utils/lana.js
Browserslist: caniuse-lite is outdated. Please run:
  npx update-browserslist-db@latest
  Why you should do it regularly: https://github.com/browserslist/update-db#readme

1 file checked.  0 passed.  1 failed.  4 warnings.  6 errors.

? Which rule would you like to view? (Use arrow keys)
❯ consistent-return:    1|  
  func-names:           1|  
  no-console:           3|    
  no-param-reassign:    4|     
  no-use-before-define: 1|  
```

**Test URLs:**
- After: https://chore-lana-lint--milo--hparra.hlx.live/?martech=off

### DIFF
--- a/libs/utils/lana.js
+++ b/libs/utils/lana.js
@@ -1,4 +1,5 @@
-(function () {
+/* eslint no-console: 0 */
+(function lanaiife() {
   const MSG_LIMIT = 2000;
 
   const defaultOptions = {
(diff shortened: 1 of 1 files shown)


chrischrischris:
lana.js is the one file in the milo repo that has to be es5 / IE11 compatible since it is also loaded by Dexter which still supports the old browsers.

Instead I think we should add the file to the .eslint.js ignorePatterns

PR #1020, opened 2023-07-27
MWPW-134309: Milo to always use Prod WCS

3ch023:
Switch Milo to only use 'wcs.adobe.com'.
'wcs.stage.adobe.com' should only be to test experimental features provided by the commerce team.

Resolves: [MWPW-134309](https://jira.corp.adobe.com/browse/MWPW-134309)

**Test URLs:**
- Before: [https://main--milo--adobecom.hlx.live/drafts/ilyas/merch?martech=off](https://main--milo--adobecom.hlx.live/drafts/ilyas/merch?martech=off)
- After: [https://mwpw-134309-wcs--milo--3ch023.hlx.page/drafts/ilyas/merch?martech=off](https://mwpw-134309-wcs--milo--3ch023.hlx.page/drafts/ilyas/merch?martech=off)

### DIFF
--- a/libs/blocks/merch/merch.js
+++ b/libs/blocks/merch/merch.js
@@ -40,7 +40,8 @@ export function getTacocatEnv(envName, locale) {
 
   const literalScriptUrl = `${host}/special/tacocat/literals/${language}.js`;
   const scriptUrl = `${host}/special/tacocat/lib/${VERSION}/tacocat.js`;
-  return { country, language, literalScriptUrl, scriptUrl, tacocatEnv: 'PRODUCTION' };
+  const tacocatEnv = getMetadata('tacocat-env') ?? 'PRODUCTION';
+  return { country, language, literalScriptUrl, scriptUrl, tacocatEnv };
 }
 
 export const omitNullValues = (target) => {
@@ -119,7 +120,7 @@ window.tacocat.loadPromise = new Promise((resolve) => {
       resolve(false);
     })
     .catch((error) => {
-      console.error('Failed to load tacocat', error);
+      window.lana.log(`Failed to load tacocat. ${error?.message}`);
       resolve(true);
     });
 });
--- a/libs/blocks/merch/merch.js
+++ b/libs/blocks/merch/merch.js
@@ -40,8 +40,7 @@ export function getTacocatEnv(envName, locale) {
 
   const literalScriptUrl = `${host}/special/tacocat/literals/${language}.js`;
   const scriptUrl = `${host}/special/tacocat/lib/${VERSION}/tacocat.js`;
-  const tacocatEnv = envName === ENV_PROD ? 'PRODUCTION' : 'STAGE';
-  return { country, language, literalScriptUrl, scriptUrl, tacocatEnv };
+  return { country, language, literalScriptUrl, scriptUrl, tacocatEnv: 'PRODUCTION' };
 }
 
 export const omitNullValues = (target) => {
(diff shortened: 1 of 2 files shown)


libs/blocks/merch/merch.js
```diff
@@ -40,8 +40,7 @@ export function getTacocatEnv(envName, locale) {
 
   const literalScriptUrl = `${host}/special/tacocat/literals/${language}.js`;
   const scriptUrl = `${host}/special/tacocat/lib/${VERSION}/tacocat.js`;
-  const tacocatEnv = envName === ENV_PROD ? 'PRODUCTION' : 'STAGE';
-  return { country, language, literalScriptUrl, scriptUrl, tacocatEnv };
+  return { country, language, literalScriptUrl, scriptUrl, tacocatEnv: 'PRODUCTION' };
```
overmyheadandbody, line 43: Since this is always 'PRODUCTION', does it even make sense to define it here? Could it not be just a constant at the top of the file, which you can then just reference when needed? This would avoid passing a static value as an argument through a chain of functions

overmyheadandbody (commented):
You might need to substitute the `console.error` for a LANA log in order for all the PR checks to pass.

test/blocks/merch/merch.test.js
```diff
@@ -307,11 +307,12 @@ describe('Merch Block', () => {
 
   describe('Tacocat config', () => {
     it('falls back to en for unsupported languages', async () => {
-      const { literalScriptUrl, language } = getTacocatEnv('local', { ietf: 'xx-US' });
+      const { literalScriptUrl, language, tacocatEnv } = getTacocatEnv('local', { ietf: 'xx-US' });
       expect(literalScriptUrl).to.equal(
         'https://www.stage.adobe.com/special/tacocat/literals/en.js',
       );
       expect(language).to.equal('en');
+      expect(tacocatEnv).to.equal('PRODUCTION');
```
JasonHowellSlavin, line 315: if you define this as a const as rares said above, could import it into the test and use the var here to keep things dry.

JasonHowellSlavin (approved):

chrischrischris (approved):
Approved as long as @overmyheadandbody comment is addressed

3ch023:
hi @overmyheadandbody , actually after thinking about it the second time I've left a way for us to set tacocat env to Stage. with a help of metadata. Could you please check again? thx
also addressed console.log issue

libs/blocks/merch/merch.js
```diff
@@ -120,7 +120,7 @@ window.tacocat.loadPromise = new Promise((resolve) => {
       resolve(false);
     })
     .catch((error) => {
-      console.error('Failed to load tacocat', error);
```
mokimo, line 123: window.lana is sure to be defined at this point, right?

mokimo (approved):

libs/blocks/merch/merch.js
overmyheadandbody: Was commenting on this too 😄 I see this is used across a lot of files, but I'd argue `window.lana?.log` might be a bit safer

overmyheadandbody (approved):
The metadata approach seems like a better option to me 👍

libs/blocks/merch/merch.js
chrischrischris: Yes `window.lana` should be defined.  It's defined before `loadArea` fires for the document for every consumer:
https://github.com/adobecom/bacom/blob/main/scripts/scripts.js#L181

However it's possible a consumer forgets to add or removes `loadLana` so we should add the `?`

PR #1025, opened 2023-07-27
[MWPW-133904] Fixing duplicated icon issue and other table issues

seanchoi-dev:
* Issue fixed with duplicated icon rendered on fragments
* Not loading icon issues due to async.
* Other table style issue on mobile fixed when it has only 2 columns.

Resolves: 
- [MWPW-134676](https://jira.corp.adobe.com/browse/MWPW-134676)
- [MWPW-133904](https://jira.corp.adobe.com/browse/MWPW-133904)
- [MWPW-134073](https://jira.corp.adobe.com/browse/MWPW-134073)
- https://adobe-mwp.slack.com/archives/C03B0BUA75E/p1690483368480649

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rclayton/double-icon?martech=off
- After: https://MWPW-133904--milo--adobecom.hlx.page/drafts/rclayton/double-icon?martech=off

- Before: https://main--bacom--adobecom.hlx.page/products/product-analytics/pricing?martech=off
- After: https://main--bacom--adobecom.hlx.page/products/product-analytics/pricing?milolibs=MWPW-133904&martech=off

### DIFF
--- a/libs/blocks/table/table.css
+++ b/libs/blocks/table/table.css
@@ -21,16 +21,26 @@
 
 .table .col {
   min-height: 27px;
+  border-left: 1px var(--border-color) solid;
+  border-top: 1px var(--border-color) solid;
 }
 
-.table .col p {
-  margin: 0;
+.table .col:last-child {
+  border-right: 1px var(--border-color) solid;
+}
+
+.table .row:last-child .col {
+  border-bottom: 1px var(--border-color) solid;
 }
 
 .table .col,
 .table .col p {
   font-size: var(--type-body-s-size);
   line-height: var(--type-body-s-lh);
+}
+
+.table .col p {
+  margin: 0;
 }
 
 .table .col p picture {
@@ -47,21 +57,8 @@
   border-right: 1px var(--border-color) solid;
 }
 
-.table:not(.merch) .col {
-  border-left: 1px var(--border-color) solid;
-  border-top: 1px var(--border-color) solid;
-}
-
 .table .row-highlight .col.col-highlight {
   border-color: transparent;
-}
-
-.table .col:last-child {
-  border-right: 1px var(--border-color) solid;
-}
-
-.table .row:last-child .col {
-  border-bottom: 1px var(--border-color) solid;
 }
 
 .table.collapse .section-head-collaped .col {
--- a/libs/blocks/table/table.css
+++ b/libs/blocks/table/table.css
@@ -37,15 +37,23 @@
   margin-right: 8px;
 }
 
-.table.merch div[class*="col-"] {
+.table.merch .row-heading .col {
   border-top: 1px var(--border-color) solid;
+}
+
+.table.merch div[class*="col-"] {
+  border-bottom: 1px var(--border-color) solid;
   border-left: 1px var(--border-color) solid;
   border-right: 1px var(--border-color) solid;
 }
 
-.table div[class*="col-"] {
+.table:not(.merch) .col {
   border-left: 1px var(--border-color) solid;
   border-top: 1px var(--border-color) solid;
+}
+
+.table .row-highlight .col.col-highlight {
+  border-color: transparent;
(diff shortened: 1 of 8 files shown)


libs/blocks/table/table.css
```diff
@@ -323,6 +331,11 @@ header.global-navigation {
     margin: 0 30px;
   }
 
+  .table:not(.merch) .row .section-head-title,
+  .table:not(.merch) .row .section-row-title {
+    border-right: 1px solid var(--border-color);
```
overmyheadandbody, line 336: Is this being reflected for RTL?

libs/blocks/table/table.css
```diff
@@ -37,17 +37,25 @@
   margin-right: 8px;
 }
 
-.table.merch div[class*="col-"] {
+.table.merch .row-heading .col {
   border-top: 1px var(--border-color) solid;
+}
+
+.table.merch div[class*="col-"] {
```
overmyheadandbody, line 44: Does this need the extra specificity or could you just use the attribute, like `.table.merch [class*="col-"]`?

libs/utils/utils.js
```diff
@@ -826,10 +826,13 @@ export async function loadArea(area = document) {
...
     areaBlocks.push(...section.blocks);
 
+    await decorateIcons(section.el, config);
+
     // Only move on to the next section when all blocks are loaded.
     await Promise.all(loaded);
 
-    await decorateIcons(section.el, config);
+    const lcpLoadedEvent = new Event('milo:LCP:loaded');
+    window.dispatchEvent(lcpLoadedEvent);
```
overmyheadandbody, line 835: Nit: since you're using `lcpLoadedEvent` just once, you could combine the two lines, avoid allocating memory for a variable and trim a few characters. Since this is utils, every little bit counts :)

overmyheadandbody (approved):

libs/blocks/table/table.css
seanchoi-dev: This is for fixing mobile issue:
image

overmyheadandbody (commented):
I just noticed a small issue. Sometimes the sticky header gets a `.top-border-transparent` applied. On mobile, this border will overlay other elements. Notice the white border on top of the "User engagement trends" cell:
White border over element
Culprit

I'd also like to ask why the right side of the table doesn't have any borders. Is it something addressable from authoring? To me it feels like the rows containing the checkmarks should also expand to the full width of the table in this scenario

seanchoi-dev:
> I just noticed a small issue. Sometimes the sticky header gets a `.top-border-transparent` applied. On mobile, this border will overlay other elements. Notice the white border on top of the "User engagement trends" cell: White border over element Culprit
> 
> I'd also like to ask why the right side of the table doesn't have any borders. Is it something addressable from authoring? To me it feels like the rows containing the checkmarks should also expand to the full width of the table in this scenario

@overmyheadandbody Realized that there were some issues due to the author makes the table a way never expected in mockup- one combined heading with only two columns. 
I added some code to support this use case.

test/blocks/table/table.test.js
```diff
@@ -9,7 +9,7 @@ describe('table and tablemetadata', () => {
   beforeEach(() => {
     const tables = document.querySelectorAll('.table');
     tables.forEach((t) => init(t));
-    window.dispatchEvent(new Event('milo:icons:loaded'));
+    window.dispatchEvent(new Event('milo:LCP:loaded'));
   });
```
JasonHowellSlavin, line 13: should you add any tests for some of the additional functionality added? re: table.js 311-319?

rgclayton:
Getting some sticky header gap issues with the gnav on screens smaller than 900px.
image

rgclayton (approved):

fullcolorcoder (approved):

PR #1026, opened 2023-07-27
MWPW-129918: Masonry styles for section metadata

vgoodric:
* `masonry` variant added to section-metadata block. JS adds "masonry-up" to the block to leverage the CSS for two-up, three-up etc.

Resolves: [MWPW-129918](https://jira.corp.adobe.com/browse/MWPW-129918)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/vgoodric/masonry?martech=off
- After: https://masonry--milo--adobecom.hlx.page/drafts/vgoodric/masonry?martech=off

### DIFF
--- a/libs/blocks/section-metadata/section-metadata.js
+++ b/libs/blocks/section-metadata/section-metadata.js
@@ -56,6 +56,7 @@ export async function handleStyle(text, section) {
   const styles = text.split(', ').map((style) => style.replaceAll(' ', '-'));
   const sticky = styles.find((style) => style === 'sticky-top' || style === 'sticky-bottom');
   if (sticky) await handleStickySection(sticky, section);
+  if (styles.includes('masonry')) styles.push('twelve-up');
   section.classList.add(...styles);
 }
 
--- a/libs/blocks/section-metadata/section-metadata.js
+++ b/libs/blocks/section-metadata/section-metadata.js
@@ -56,7 +56,7 @@ export async function handleStyle(text, section) {
   const styles = text.split(', ').map((style) => style.replaceAll(' ', '-'));
   const sticky = styles.find((style) => style === 'sticky-top' || style === 'sticky-bottom');
   if (sticky) await handleStickySection(sticky, section);
-  if (styles.includes('masonry')) styles.push('twelve-up');
+  if (styles.includes('masonry')) styles.push('masonry-up');
   section.classList.add(...styles);
 }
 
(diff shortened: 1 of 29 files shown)


rgclayton:
This is the ticket @auniverseaway mentioned we need to workshop some of the authoring in section-metadata, correct? 

cc: @ryanmparrish

rgclayton:
Related to 
image

Does the properties `min-height-450`, `full width desktop`, `align-items-stretch`, `border radius s` only work when the `masonry` property is added? Or will those still work independently?

rgclayton:
This is pretty cool though. 99% css doing the work 😄 . 

With that said i think we are asking authors to basically know the basics of CSS Grid, which will most likely be a bit confusing. This would need a lot of documentation.

vgoodric:
Based on feedback from show and tell and discussing with Danielle Marriott, I am limiting this PR to the 1 line of JS and removing all CSS updates.  CSS is being moved to homepage-brick and will therefore be repository specific.  This will give us more time to decide on how to implement masonry sitewide.  But I am hoping that this code shows how much of the work can be done by pure CSS.

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -56,6 +56,7 @@ export async function handleStyle(text, section) {
   const styles = text.split(', ').map((style) => style.replaceAll(' ', '-'));
   const sticky = styles.find((style) => style === 'sticky-top' || style === 'sticky-bottom');
   if (sticky) await handleStickySection(sticky, section);
+  if (styles.includes('masonry')) styles.push('masonry-up');
```
mokimo, line 59: Why not just add `masonry-up` to the section metadata styles?

rgclayton (approved):

libs/blocks/section-metadata/section-metadata.js
vgoodric: I thought about that, but I feel like "masonry up" is not what anyone would call it and I'm trying to keep it author-friendly.  So I really wanted to call it "masonry".  And 1 line of JS saves us a bunch of lines of CSS.

chrischrischris (approved):

vgoodric:
> This is the ticket @auniverseaway mentioned we need to workshop some of the authoring in section-metadata, correct?
> 
> cc: @ryanmparrish

Correct.  That is why I moved the CSS to the repository-specific block "homepage-brick".

PR #1030, opened 2023-07-28
Add analytics tracking to modal

Ruchika4:
* Add analytics tracking for:
Modal close
Geourouting load
On click of link on georouting pop up when selecting country other than your region (Visiting jp page from US and selecting JP from georouting modal)

All above events have been added to make it same as we see right now on dexter pages as per discussion with analytics team

Resolves: [MWPW-134229](https://jira.corp.adobe.com/browse/MWPW-134229)
[MWPW-134229](https://jira.corp.adobe.com/browse/MWPW-130145)

**Test URLs:**

To verify georouting

- Before: https://modal-analytics--milo--adobecom.hlx.page/de/
- After: https://modal-analytics--milo--ruchika4.hlx.page/de/

To verify tracking on close for other modal (Click on Learn More button to open the modal):

- Before: https://modal-analytics--milo--adobecom.hlx.page/drafts/ruchika/iframe
- After: https://modal-analytics--milo--ruchika4.hlx.page/drafts/ruchika/iframe

CC:

- Before: https://main--cc--adobecom.hlx.page/creativecloud/animation/discover?akamaiLocale=ru
- After: https://main--cc--adobecom.hlx.page/creativecloud/animation/discover?milolibs=modal-analytics--milo--ruchika4&akamaiLocale=ru

Bacom:
- Before: https://main--bacom--adobecom.hlx.page/de/customer-success-stories
- After: https://main--bacom--adobecom.hlx.page/de/customer-success-stories?milolibs=modal-analytics--milo--ruchika4

### DIFF
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -30,11 +30,12 @@ export function sendAnalytics(event) {
 
 function closeModal(modal) {
   const { id } = modal;
-  const localeModal = id?.includes('locale-modal') ? 'localeModal' : 'milo';
-  const eventName = window.location.hash ? window.location.hash.replace('#', '') : localeModal;
-  const closeEvent = new Event(`${eventName}:modalClose:buttonClose`);
+  const closeEvent = new Event('milo:modal:closed');
   window.dispatchEvent(closeEvent);
-  sendAnalytics(closeEvent);
+  const localeModal = id?.includes('locale-modal') ? 'localeModal' : 'milo';
+  const analyticsEventName = window.location.hash ? window.location.hash.replace('#', '') : localeModal;
+  const closeEventAnalytics = new Event(`${analyticsEventName}:modalClose:buttonClose`);
+  sendAnalytics(closeEventAnalytics);
 
   document.querySelectorAll(`#${id}`).forEach((mod) => {
     if (mod.nextElementSibling?.classList.contains('modal-curtain')) {
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -1,3 +1,4 @@
+/* eslint-disable import/no-cycle */
 import { createTag, getMetadata, localizeLink, loadStyle, getConfig } from '../../utils/utils.js';
 
 const FOCUSABLES = 'a, button, input, textarea, select, details, [tabindex]:not([tabindex="-1"]';
@@ -29,7 +30,8 @@ export function sendAnalytics(event) {
 
 function closeModal(modal) {
   const { id } = modal;
-  const eventName = window.location.hash ? window.location.hash.replaceAll('#', '') : 'localeModal';
+  const localeModal = id?.includes('locale-modal') ? 'localeModal' : 'milo';
+  const eventName = window.location.hash ? window.location.hash.replace('#', '') : localeModal;
   const closeEvent = new Event(`${eventName}:modalClose:buttonClose`);
   window.dispatchEvent(closeEvent);
   sendAnalytics(closeEvent);
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -16,10 +16,23 @@ export function findDetails(hash, el) {
   return { id, path, isHash: hash === window.location.hash };
 }
 
+export function sendAnalytics(event) {
+  // eslint-disable-next-line no-underscore-dangle
+  window._satellite?.track('event', {
+    xdm: {},
+    data: {
+      web: { webInteraction: { name: event.type } },
+      _adobe_corpnew: { digitalData: event.data },
+    },
+  });
+}
+
 function closeModal(modal) {
   const { id } = modal;
-  const closeEvent = new Event('milo:modal:closed');
+  const eventName = window.location.hash ? window.location.hash.replaceAll('#', '') : 'localeModal';
+  const closeEvent = new Event(`${eventName}:modalClose:buttonClose`);
   window.dispatchEvent(closeEvent);
+  sendAnalytics(closeEvent);
 
   document.querySelectorAll(`#${id}`).forEach((mod) => {
     if (mod.nextElementSibling?.classList.contains('modal-curtain')) {
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -22,8 +22,8 @@ export function sendAnalytics(event) {
   window._satellite?.track('event', {
     xdm: {},
     data: {
-      web: { webInteraction: { name: event.type } },
-      _adobe_corpnew: { digitalData: event.data },
+      web: { webInteraction: { name: event?.type } },
+      _adobe_corpnew: { digitalData: event?.data },
     },
   });
 }
--- a/libs/features/georoutingv2/georoutingv2.js
+++ b/libs/features/georoutingv2/georoutingv2.js
@@ -309,12 +309,12 @@ export default async function loadGeoRouting(
     // Show modal when url and cookie disagree
     if (urlLocale.split('_')[0] !== storedLocale.split('_')[0]) {
       const localeMatches = json.georouting.data.filter(
-        (d) => d.prefix === storedLocale
+        (d) => d.prefix === storedLocale,
       );
       const details = await getDetails(
         urlGeoData,
         localeMatches,
-        json.geos.data
+        json.geos.data,
       );
       if (details) {
         await showModal(details);
@@ -322,8 +322,8 @@ export default async function loadGeoRouting(
           new Event(
             `Load:${urlLocale.split('_')[0]}-${
               storedLocale.split('_')[0]
-            }|Geo_Routing_Modal`
-          )
+            }|Geo_Routing_Modal`,
+          ),
         );
       }
     }
@@ -339,7 +339,7 @@ export default async function loadGeoRouting(
       await showModal(details);
       const modUrlLocale = urlLocale || 'us';
       sendAnalytics(
-        new Event(`Load:${akamaiCode}-${modUrlLocale}|Geo_Routing_Modal`)
+        new Event(`Load:${akamaiCode}-${modUrlLocale}|Geo_Routing_Modal`),
       );
     }
   }
--- a/libs/features/georoutingv2/georoutingv2.js
+++ b/libs/features/georoutingv2/georoutingv2.js
@@ -315,10 +315,8 @@ export default async function loadGeoRouting(
       const details = await getDetails(urlGeoData, localeMatches, json.geos.data);
       if (details) {
         await showModal(details);
-        const modUrlLocaleGeo = urlLocaleGeo || 'us';
-        const modStoredLocaleGeo = storedLocaleGeo || 'us';
         sendAnalyticsFunc(
-          new Event(`Load:${modUrlLocaleGeo}-${modStoredLocaleGeo}|Geo_Routing_Modal`),
+          new Event(`Load:${urlLocaleGeo || 'us'}-${storedLocaleGeo || 'us'}|Geo_Routing_Modal`),
         );
       }
     }
@@ -332,9 +330,8 @@ export default async function loadGeoRouting(
     const details = await getDetails(urlGeoData, localeMatches, json.geos.data);
     if (details) {
       await showModal(details);
-      const modUrlLocale = urlLocale || 'us';
       sendAnalyticsFunc(
-        new Event(`Load:${akamaiCode}-${modUrlLocale}|Geo_Routing_Modal`),
+        new Event(`Load:${akamaiCode || 'us'}-${urlLocale || 'us'}|Geo_Routing_Modal`),
       );
     }
   }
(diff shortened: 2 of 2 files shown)


libs/blocks/modal/modal.js
```diff
@@ -16,10 +17,24 @@ export function findDetails(hash, el) {
...
+    data: {
+      web: { webInteraction: { name: event.type } },
+      _adobe_corpnew: { digitalData: event.data },
+    },
+  });
+}
+
 function closeModal(modal) {
   const { id } = modal;
-  const closeEvent = new Event('milo:modal:closed');
```
overmyheadandbody, line 21: Renaming this will require consumers to update their implementations, this event is currently used in BACOM, Express and the footer, as can be seen [here](https://github.com/search?q=org%3Aadobecom%20milo%3Amodal%3Aclosed&type=code).

Is there any way we can keep the event name the same? Otherwise, a cross-team sync will be needed to implement this change as soon as it is merged in.

libs/features/georoutingv2/georoutingv2.js
```diff
@@ -293,10 +308,23 @@ export default async function loadGeoRouting(conf, createTagFunc, getMetadataFun
...
-      const localeMatches = json.georouting.data.filter((d) => d.prefix === storedLocale);
-      const details = await getDetails(urlGeoData, localeMatches, json.geos.data);
+      const localeMatches = json.georouting.data.filter(
+        (d) => d.prefix === storedLocale,
+      );
+      const details = await getDetails(
+        urlGeoData,
+        localeMatches,
+        json.geos.data,
+      );
```
overmyheadandbody, line 318: Is this formatting being done by a code editor extension or personal preference? I'd argue it creates a bit of noise in the PR and adds a few more lines without a real benefit. Same comment for L275 & L291.

libs/features/georoutingv2/georoutingv2.js
```diff
@@ -293,10 +308,23 @@ export default async function loadGeoRouting(conf, createTagFunc, getMetadataFun
...
       if (details) {
         await showModal(details);
+        sendAnalytics(
+          new Event(
+            `Load:${urlLocale.split('_')[0]}-${
+              storedLocale.split('_')[0]
```
overmyheadandbody, line 324: You're using `urlLocale.split('_')[0]` and `storedLocale.split('_')[0]` in the conditional on L310 too, maybe these could become variables.

libs/features/georoutingv2/georoutingv2.js
```diff
@@ -110,14 +113,18 @@ function getGeoroutingOverride() {
   return hideGeorouting === 'on';
 }
 
-function decorateForOnLinkClick(link, prefix) {
+function decorateForOnLinkClick(link, prefix, currPrefix) {
   link.addEventListener('click', () => {
     const modPrefix = prefix || 'us';
+    const modCurrPrefix = currPrefix || 'us';
```
overmyheadandbody, line 119: Since this is used only if the condition on L125 is met, maybe it can be moved inside that `if`

libs/features/georoutingv2/georoutingv2.js
```diff
@@ -110,14 +113,18 @@ function getGeoroutingOverride() {
...
     // set cookie so legacy code on adobecom still works properly.
     const domain = window.location.host === 'adobe.com'
       || window.location.host.endsWith('.adobe.com') ? 'domain=adobe.com' : '';
     document.cookie = `international=${modPrefix};path=/;${domain}`;
     link.closest('.dialog-modal').dispatchEvent(new Event('closeModal'));
+    if (currPrefix !== undefined) {
```
overmyheadandbody, line 125: If `currPrefix` is `null` this condition is met. Could we instead check for `.length`?

overmyheadandbody (commented):
My main concern here is around updating the event name. What's the strategy to avoid issues on the consumer side?

libs/blocks/modal/modal.js
Ruchika4: I have updated the event name based on the live pages. You can verify that here https://www.adobe.com/ru (click close of georouting modal).
I'll ask this question in #milo-community to get confirmation from Benjamin and Uma

libs/features/georoutingv2/georoutingv2.js
Ruchika4: This was done by IDE extension for formatting. I have reverted it. However didn't reverted the one on line 275 as that was done to fix below linting issue
Screen Shot 2023-07-28 at 6 36 09 AM

libs/features/georoutingv2/georoutingv2.js
Ruchika4: done

libs/features/georoutingv2/georoutingv2.js
Ruchika4: done

libs/features/georoutingv2/georoutingv2.js
Ruchika4: So this check is to to verify is currPrefix has been passed to the function then only do the next step. I have verified if the currPrefix (it is the third parameter in the function definition) is not passed in the function and it comes as undefined when not passed.

Also length cannot be used as for when we visit US page from say France prefix is passed as '' but we still want to call analytics page. We only don't want it to be called when it is not passed in the function

libs/blocks/modal/modal.js
mokimo: I'd probably stick to the JS event `milo:modal:closed` and then a different name for the Analytics event

libs/features/georoutingv2/georoutingv2.js
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable import/no-cycle */
+import { sendAnalytics } from '../../blocks/modal/modal.js';
```
JasonHowellSlavin, line 2: I know this is to keep the code dry, but I do not think georoutingv2 should have a dependency on modal.js / block code.  Could we move the send analytics function somewhere else that bot georoutingv2 and modal.js could consume?

JasonHowellSlavin (dismissed):

libs/blocks/modal/modal.js
Ruchika4: Thanks for the suggestion @mokimo . I have incorporated it in my code now.

libs/features/georoutingv2/georoutingv2.js
Ruchika4: @JasonHowellSlavin thanks for your review. However modal.js is already being imported in georouting file so it already has dependency on modal block code which I feel is needed and cannot be removed.

Ruchika4:
> My main concern here is around updating the event name. What's the strategy to avoid issues on the consumer side?

@overmyheadandbody 
This should be resolved now as we have separate event name for analytics now. Please review once more and let me know if any concern

libs/features/georoutingv2/georoutingv2.js
```diff
@@ -309,6 +331,10 @@ export default async function loadGeoRouting(conf, createTagFunc, getMetadataFun
     const details = await getDetails(urlGeoData, localeMatches, json.geos.data);
     if (details) {
       await showModal(details);
+      const modUrlLocale = urlLocale || 'us';
+      sendAnalytics(
+        new Event(`Load:${akamaiCode}-${modUrlLocale}|Geo_Routing_Modal`),
```
mokimo, line 336:
```suggestion
      sendAnalytics(
        new Event(`Load:${akamaiCode}-${urlLocale || 'us'}|Geo_Routing_Modal`),
```

libs/features/georoutingv2/georoutingv2.js
```diff
@@ -291,12 +304,21 @@ export default async function loadGeoRouting(conf, createTagFunc, getMetadataFun
...
+      const localeMatches = json.georouting.data.filter(
+        (d) => d.prefix === storedLocale,
+      );
       const details = await getDetails(urlGeoData, localeMatches, json.geos.data);
       if (details) {
         await showModal(details);
+        const modUrlLocaleGeo = urlLocaleGeo || 'us';
+        const modStoredLocaleGeo = storedLocaleGeo || 'us';
+        sendAnalytics(
+          new Event(`Load:${modUrlLocaleGeo}-${modStoredLocaleGeo}|Geo_Routing_Modal`),
```
mokimo, line 320: The variables introduce more confusion than clarity IMO, so I'd just inline them
```suggestion
        sendAnalytics(
          new Event(`Load:${urlLocaleGeo || 'us'}-${storedLocaleGeo || 'us'}|Geo_Routing_Modal`),
```

mokimo (approved):

libs/features/georoutingv2/georoutingv2.js
JasonHowellSlavin: Right you are! 

If we're importing at the top, does it make sense then to import `getModal` at the same time, and refactor the `showModal` function?

Ruchika4: Hi @JasonHowellSlavin I went ahead and did the other way round. Removed import from top of the file and added sendAnalytics function to the existing import call. Please let me know if that looks good

Axelcureno:
A few questions:

- Is AEP enabled by default? If so, where it is being initialized? 
- Shouldn't there be a check if AEP is enabled before making the tracking call?
- Is the window object needed? I believe if launch and martech scripts are part of the page DOM, we can simply call _satellite?.track directly without the window object.

Ruchika4:
> * https://modal-analytics--milo--ruchika4.hlx.page/drafts/ruchika/iframe

Hi @Axelcureno thanks for taking out time to review. I'll try to answer the questions to the best of my knowledge. I just started exploring analytics for few tickets I got last week.
1. As far as I know it is enabled by default unless martech is set as false in query param or in metadata. Here is code in utils where this check is done https://github.com/adobecom/milo/blob/main/libs/utils/utils.js#L664 
2. I do use optional chaining for that. _satellite is undefined when martech is set to off and that gets handled by this. 
3. I couldn't use directly _satellite?.track in my code and got error that _satellite is not defined. Hence used window object.
I'll tag @chrischrischris also here as he has extensively worked on martech in milo to confirm if above answers are correct.

libs/features/georoutingv2/georoutingv2.js
JasonHowellSlavin: could you then use undefined as a default param? 
`function decorateForOnLinkClick(link, prefix, currPrefix = undefined) {`

JasonHowellSlavin (approved):

libs/features/georoutingv2/georoutingv2.js
Ruchika4: It'll be undefined when not passed so was wondering if assigning the same default value will matter. Let me know if you think we should assign some other default value to it.

libs/features/georoutingv2/georoutingv2.js
Ruchika4: This is done.

libs/features/georoutingv2/georoutingv2.js
Ruchika4: This is done.

libs/blocks/modal/modal.js
```diff
@@ -16,10 +17,25 @@ export function findDetails(hash, el) {
   return { id, path, isHash: hash === window.location.hash };
 }
 
+export function sendAnalytics(event) {
+  // eslint-disable-next-line no-underscore-dangle
+  window._satellite?.track('event', {
+    xdm: {},
+    data: {
+      web: { webInteraction: { name: event.type } },
```
overmyheadandbody, line 25: It might be a a good idea to check whether `event` exists before accessing its properties. A check at the start pf the method's logic or optional chaining here and on L26 should do it.

libs/features/georoutingv2/georoutingv2.js
```diff
@@ -110,14 +111,18 @@ function getGeoroutingOverride() {
   return hideGeorouting === 'on';
 }
 
-function decorateForOnLinkClick(link, prefix) {
+function decorateForOnLinkClick(link, prefix, currPrefix) {
```
overmyheadandbody, line 114: I think naming here is a bit confusing. Looking at a call of this method, ` decorateForOnLinkClick(altAction, currentPage.prefix, locale.prefix);`, `currentPage.prefix` corresponds to `prefix`, while `locale.prefix` corresponds to `currPrefix`. If I isolated this method, I would have no idea what the difference between `prefix` and `currPrefix` is. Maybe some different naming can be found. `urlLocale` and `pageLocale`? `pagePrefix` and `urlPrefix`? Something along those lines, you probably know the use-case better 😀

Ruchika4: this is fixed

overmyheadandbody (approved):

Axelcureno (approved):

spadmasa:
Verified the flow in below urls, able to see the tracking rendered on modal close .

https://main--cc--adobecom.hlx.page/creativecloud/animation/discover?milolibs=modal-analytics--milo--ruchika4&akamaiLocale=ru
Screenshot 2023-07-31 at 8 38 12 PM




Screenshot 2023-07-31 at 8 39 34 PM





https://main--bacom--adobecom.hlx.page/de/customer-success-stories?milolibs=modal-analytics--milo--ruchika4

PR #1031, opened 2023-07-28
iframe window opened in mobile view is too small 

sharath-kannan:
* iframe window opened in mobile view is too small , not able to see the complete details to do the payment flow.

Resolves: [MWPW-133983](https://jira.corp.adobe.com/browse/MWPW-133983)

**Test URLs:**
- Before: https://stage--cc--adobecom.hlx.page/drafts/ruchika/discover1?martech=off#animate
- After: https://stage--cc--adobecom.hlx.page/drafts/ruchika/discover1?martech=off&milolibs=MWPW-133983--milo--sharath-kannan#animate

before

This is how the modal should look after the PR merge.
After

### DIFF
--- a/libs/blocks/modal/modal.css
+++ b/libs/blocks/modal/modal.css
@@ -166,3 +166,11 @@
     max-width: calc((100% - 6px) - 2em);
   } 
 }
+
+@media (max-width: 1200px) {
+  .dialog-modal {
+    width:100%;
+    max-width:100%;
+    max-height:100%;
+  }
+}
--- a/libs/blocks/modal/modal.css
+++ b/libs/blocks/modal/modal.css
@@ -168,9 +168,9 @@
 }
 
 @media (max-width: 1200px) {
-  .dialog-modal {
-    width:100%;
-    max-width:100%;
-    max-height:100%;
+  .dialog-modal.commerce-frame {
+    width: 100%;
+    max-width: 100%;
+    max-height: 100%;
   }
 }
--- a/libs/blocks/modal/modal.css
+++ b/libs/blocks/modal/modal.css
@@ -141,6 +141,7 @@
   margin: 0;
 }
 
+/* stylelint-disable-next-line no-descending-specificity */
 .region-selector .content {
   column-count: 1;
 }
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -8,6 +8,8 @@ const CLOSE_ICON = `<svg xmlns="http://www.w3.org/2000/svg" width="20" height="2
     <line x1="8" y1="8" transform="translate(10506 -3397)" fill="none" stroke="#fff" stroke-width="2"/>
   </g>
 </svg>`;
+const MOBILE_MAX = 599;
+const TABLET_MAX = 1199;
 
 export function findDetails(hash, el) {
   const id = hash.replace('#', '');
@@ -160,5 +162,17 @@ window.addEventListener('hashchange', (e) => {
   } else {
     const details = findDetails(window.location.hash, null);
     if (details) getModal(details);
+  }
+});
+
+function sendViewportDimensionsToiFrame(source) {
+  const viewportWidth = Math.max(document.documentElement.clientWidth || 0, window.innerWidth || 0);
+  source.postMessage({ mobileMax: MOBILE_MAX, tabletMax: TABLET_MAX, viewportWidth }, '*');
+}
+
+window.addEventListener('message', (t) => {
+  if (t.data === 'viewportWidth') {
+    sendViewportDimensionsToiFrame(t.source);
+    window.addEventListener('resize', () => sendViewportDimensionsToiFrame(t.source));
   }
 });
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -70,6 +70,17 @@ async function getPathModal(path, dialog) {
   const { default: getFragment } = await import('../fragment/fragment.js');
   await getFragment(block);
 }
+function sendViewportDimensionsToiFrame(source) {
+  const viewportWidth = Math.max(document.documentElement.clientWidth || 0, window.innerWidth || 0);
+  source.postMessage({ mobileMax: MOBILE_MAX, tabletMax: TABLET_MAX, viewportWidth }, '*');
+}
+
+function sendViewportDimensionsOnRequest(messageInfo) {
+  if (messageInfo.data === 'viewportWidth') {
+    sendViewportDimensionsToiFrame(messageInfo.source);
+    window.addEventListener('resize', () => sendViewportDimensionsToiFrame(messageInfo.source));
+  }
+}
 
 export async function getModal(details, custom) {
   if (!(details?.path || custom)) return null;
@@ -139,7 +150,12 @@ export async function getModal(details, custom) {
     [...document.querySelectorAll('header, main, footer')]
       .forEach((element) => element.setAttribute('aria-disabled', 'true'));
   }
-
+  if (dialog.classList.toString().includes('commerce-frame')) {
+    const { debounce } = await import('../../utils/action.js');
+    window.addEventListener('message', (messageInfo) => {
+      debounce(sendViewportDimensionsOnRequest(messageInfo));
+    });
+  }
   return dialog;
 }
 
@@ -162,17 +178,5 @@ window.addEventListener('hashchange', (e) => {
   } else {
     const details = findDetails(window.location.hash, null);
     if (details) getModal(details);
-  }
-});
-
-function sendViewportDimensionsToiFrame(source) {
-  const viewportWidth = Math.max(document.documentElement.clientWidth || 0, window.innerWidth || 0);
-  source.postMessage({ mobileMax: MOBILE_MAX, tabletMax: TABLET_MAX, viewportWidth }, '*');
-}
-
-window.addEventListener('message', (t) => {
-  if (t.data === 'viewportWidth') {
-    sendViewportDimensionsToiFrame(t.source);
-    window.addEventListener('resize', () => sendViewportDimensionsToiFrame(t.source));
   }
 });
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -9,6 +9,7 @@ const CLOSE_ICON = `<svg xmlns="http://www.w3.org/2000/svg" width="20" height="2
     <line x1="8" y1="8" transform="translate(10506 -3397)" fill="none" stroke="#fff" stroke-width="2"/>
   </g>
 </svg>`;
+let isInitailPageLoad = true;
 const MOBILE_MAX = 599;
 const TABLET_MAX = 1199;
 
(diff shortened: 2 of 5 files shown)


libs/blocks/modal/modal.css
```diff
@@ -166,3 +166,11 @@
     max-width: calc((100% - 6px) - 2em);
   } 
 }
+
+@media (max-width: 1200px) {
+  .dialog-modal {
+    width:100%;
+    max-width:100%;
+    max-height:100%;
```
overmyheadandbody, line 174: I'm not sure we want to expand all modals to these dimensions on viewports widths lower than `1200px`. Also, as a nit, you should leave a space between property and value for better readability.

libs/blocks/modal/modal.js
```diff
@@ -162,3 +164,15 @@ window.addEventListener('hashchange', (e) => {
...
+
+function sendViewportDimensionsToiFrame(source) {
+  const viewportWidth = Math.max(document.documentElement.clientWidth || 0, window.innerWidth || 0);
+  source.postMessage({ mobileMax: MOBILE_MAX, tabletMax: TABLET_MAX, viewportWidth }, '*');
+}
+
+window.addEventListener('message', (t) => {
+  if (t.data === 'viewportWidth') {
+    sendViewportDimensionsToiFrame(t.source);
+    window.addEventListener('resize', () => sendViewportDimensionsToiFrame(t.source));
```
overmyheadandbody, line 176: This should probably be debounced and it should be ensured that it's added just once.

overmyheadandbody (commented):
Could you add a valid 'after' URL for testing, please?

libs/blocks/modal/modal.css
sharath-kannan: This could be an alternative solution.

/*parent selector alternative for dialog-modal*/
.dialog-modal:has(.fragment[data-path*="fragments/trial-modals"]){
     width: 100%;
    max-width: 100%;
    max-height: 100%;
}

sharath-kannan:
> Could you add a valid 'after' URL for testing, please?

We do not have a valid test page live due to cors errors,you can refer to the After image i have attached in the PR.

libs/blocks/modal/modal.js
```diff
@@ -162,3 +164,15 @@ window.addEventListener('hashchange', (e) => {
     if (details) getModal(details);
   }
 });
+
+function sendViewportDimensionsToiFrame(source) {
+  const viewportWidth = Math.max(document.documentElement.clientWidth || 0, window.innerWidth || 0);
+  source.postMessage({ mobileMax: MOBILE_MAX, tabletMax: TABLET_MAX, viewportWidth }, '*');
+}
+
+window.addEventListener('message', (t) => {
```
chrischrischris, line 173: This is listening for the `message` event which would only fire if the page is inside an iFrame.  Is that the intent?

libs/blocks/modal/modal.css
Ruchika4: Hi @sharath-kannan as discussed please add this css to only commerce-frame css class and not to the dialog-modal class

Ruchika4 (changes requested):

libs/blocks/modal/modal.css
sharath-kannan: changed this to a much specific selector.

chrischrischris:
From what I understand the main reason for the iframe message passing around is that the dexter page is not setup to handle iframe resize without getting the message from the containing milo page.

I'm pretty sure some changes to the dexter page could fix this issue without needing to pass messages, but since this is for Tue and there's no changing that dexter page now I understand.

However, I'd rather not have every single modal add these event listeners ... can we add a condition so that the message stuff is only added if the iframe url contains "fragments/trial-modals"  or some other way to detect?

Ruchika4:
> From what I understand the main reason for the iframe message passing around is that the dexter page is not setup to handle iframe resize without getting the message from the containing milo page.
> 
> I'm pretty sure some changes to the dexter page could fix this issue without needing to pass messages, but since this is for Tue and there's no changing that dexter page now I understand.
> 
> However, I'd rather not have every single modal add these event listeners ... can we add a condition so that the message stuff is only added if the iframe url contains "fragments/trial-modals" or some other way to detect?

I think we can add a check that if the modal has a css class commerce-frame which we are using for this use case then add the event as in future authors might put fragments like this in other folders too.

Ruchika4:
@sharath-kannan could you please fix the eslint issues too

Ruchika4 (approved):

libs/blocks/modal/modal.js
```diff
@@ -65,9 +68,21 @@ async function getPathModal(path, dialog) {
...
 }
+function sendViewportDimensionsToiFrame(source) {
+  const viewportWidth = Math.max(document.documentElement.clientWidth || 0, window.innerWidth || 0);
+  source.postMessage({ mobileMax: MOBILE_MAX, tabletMax: TABLET_MAX, viewportWidth }, '*');
+}
+
+export function sendViewportDimensionsOnRequest(messageInfo) {
+  if (messageInfo.data === 'viewportWidth') {
+    sendViewportDimensionsToiFrame(messageInfo.source);
+    window.addEventListener('resize', () => sendViewportDimensionsToiFrame(messageInfo.source));
```
mokimo, line 83: You need the debounce here

libs/blocks/modal/modal.js
```diff
@@ -137,7 +152,12 @@ export async function getModal(details, custom) {
     [...document.querySelectorAll('header, main, footer')]
       .forEach((element) => element.setAttribute('aria-disabled', 'true'));
   }
-
+  if (dialog.classList.contains('commerce-frame')) {
+    const { debounce } = await import('../../utils/action.js');
+    window.addEventListener('message', (messageInfo) => {
+      debounce(sendViewportDimensionsOnRequest(messageInfo));
+    });
```
mokimo, line 159: You only want to set this up once, otherwise u will end up creating multiple resize event listeners

libs/blocks/modal/modal.js
sharath-kannan: The debounce has been added in the caller function.

mokimo, line 83:
```suggestion
    window.addEventListener('resize', debounce(() => sendViewportDimensionsToiFrame(messageInfo.source), 50));
```

mokimo: Yes that won't work because the resize event recieves a lot of calls - the caller function is irrelevant. You can test it by just checking how often the `sendViewportDimensionsToiFrame` function is called when you resize the window

libs/blocks/modal/modal.js
mokimo: The code still doesn't look correct. Everytime you get a `message` event you will call the `sendViewportDimensionsOnRequest` method, which will set up a new resize event listener without discarding the old one.

If you receive 10 messages, you have 10 resize event listeners and call `sendViewportDimensionsToiFrame` 10 times.

mokimo: You have a couple of options to fix that. Here are two suggestions:
1. Remove the 'message' event listener after the initial set up OR
2. Check if there is already a resize event listener, if yes - don't set it up again

libs/blocks/modal/modal.js
```diff
@@ -137,7 +152,12 @@ export async function getModal(details, custom) {
     [...document.querySelectorAll('header, main, footer')]
       .forEach((element) => element.setAttribute('aria-disabled', 'true'));
   }
-
+  if (dialog.classList.contains('commerce-frame')) {
+    const { debounce } = await import('../../utils/action.js');
+    window.addEventListener('message', (messageInfo) => {
+      debounce(sendViewportDimensionsOnRequest(messageInfo));
```
mokimo, line 158: You can remove the debounce from here, it does nothing. If you don't need this event listener after the initial setup, anymore consider cleaning it up and remove it - otherwise you continue to listen to uninteresting messages.
```suggestion
     sendViewportDimensionsOnRequest(messageInfo)
```

libs/blocks/modal/modal.js
```diff
@@ -65,9 +68,21 @@ async function getPathModal(path, dialog) {
...
+  const viewportWidth = Math.max(document.documentElement.clientWidth || 0, window.innerWidth || 0);
+  source.postMessage({ mobileMax: MOBILE_MAX, tabletMax: TABLET_MAX, viewportWidth }, '*');
+}
+
+export function sendViewportDimensionsOnRequest(messageInfo) {
+  if (messageInfo.data === 'viewportWidth') {
+    sendViewportDimensionsToiFrame(messageInfo.source);
+    window.addEventListener('resize', () => sendViewportDimensionsToiFrame(messageInfo.source));
+  }
+}
```
mokimo, line 85: This would be the quickest solution - though if you can move the `messageInfo.data === 'viewportWidth'` condition to the eventListener on L157, that's cleaner and you don't need the extra boolean check if the event listener already exists.
```suggestion
let resizeListenerAdded = false;
export function sendViewportDimensionsOnRequest(messageInfo) {
  if (messageInfo.data === 'viewportWidth') {
    sendViewportDimensionsToiFrame(messageInfo.source);
    if (!resizeListenerAdded) {
      window.addEventListener('resize', debounce(() => sendViewportDimensionsToiFrame(messageInfo.source), 50));
      resizeListenerAdded = true;
    }
  }
}
```

libs/blocks/modal/modal.js
sharath-kannan: if (dialog.classList.contains('commerce-frame')) {
    window.addEventListener('message', (messageInfo) => {
      if (messageInfo.data === 'viewportWidth') {
        sendViewportDimensionsOnRequest(messageInfo);
      }
    });
  }
 Moving the condition (messageInfo.data === 'viewportWidth') into the event listener would add the message event  listener every time a dialog modal is opened.

libs/blocks/modal/modal.js
```diff
@@ -137,7 +158,14 @@ export async function getModal(details, custom) {
     [...document.querySelectorAll('header, main, footer')]
       .forEach((element) => element.setAttribute('aria-disabled', 'true'));
   }
-
+  if (dialog.classList.contains('commerce-frame')) {
+    if (isInitialPageLoad) {
```
mokimo, line 162: The `isInitialPageLoad` does not do anything - as the event listener for the message is only set up once anyway.

libs/blocks/modal/modal.js
```diff
@@ -137,7 +158,14 @@ export async function getModal(details, custom) {
...
+      window.addEventListener('message', (messageInfo) => {
+        sendViewportDimensionsOnRequest(messageInfo);
+      });
+      isInitialPageLoad = false;
+    }
+  }
```
mokimo, line 168: Does that work?
```suggestion
  if (dialog.classList.contains('commerce-frame')) sendViewportDimensionsOnRequest(messageInfo);
```

libs/blocks/modal/modal.js
sharath-kannan: No,Without that i see that message event listeners are added multiple times.
Screenshot 2023-07-31 at 5 15 44 PM

mokimo (approved):
As a follow up, maybe we can see if we can simplify the code a bit // see if we can tackle this from a dexter perspective - it seems like this is doing a lot of checks and I'm not convinced all of them are needed for resizing.

Should be good enough for the release though

test/blocks/modals/mocks/body.html
```diff
@@ -5,6 +5,13 @@
   .modal-curtain.is-open {
     background: rgb(50 50 50 / 80%);
   }
+  @media (max-width: 1200px) {
+  .dialog-modal.commerce-frame {
```
chrischrischris, line 9: nit: indent

rgclayton:
> Could you add a valid 'after' URL for testing, please?

Did the after url get updated? I'm still seeing the cross origin error. I think you'll have to use a stage published page to test against in order to see the content in the iframe. 

cc: @Ruchika4

Ruchika4:
> > Could you add a valid 'after' URL for testing, please?
> 
> Did the after url get updated? I'm still seeing the cross origin error. I think you'll have to use a stage published page to test against in order to see the content in the iframe.
> 
> cc: @Ruchika4

Hi @rgclayton , here is the valid url https://www.stage.adobe.com/creativecloud/animation/discover.html#animate

PR #1065, opened 2023-08-03
MWPW-131747 - Support structured data API for Video SEO

hparra:
Authors can automatically add structured data to a page for YouTube videos by assigning the video URL to the `seotech-video-url` metadata property.

Resolves: [MWPW-129764](https://jira.corp.adobe.com/browse/MWPW-129764) > [MWPW-131747](https://jira.corp.adobe.com/browse/MWPW-131747)

**Test URLs:**
- Before: https://main--milo--hparra.hlx.live/drafts/hgpa/test/seotech-video/youtube?martech=off
- After: https://seotech-video--milo--hparra.hlx.live/drafts/hgpa/test/seotech-video/youtube?martech=off

Verification:

```js
// check for metadata
document.querySelector('meta[name="seotech-video-url"]')

// check for VideoObject
document.querySelector('script[type="application/ld+json"]')
```


libs/features/seotech/seotech.js
```diff
@@ -0,0 +1,34 @@
+export const SEOTECH_API_URL_PROD = 'https://14257-seotech.adobeioruntime.net';
+export const SEOTECH_API_URL_STAGE = 'https://14257-seotech-stage.adobeioruntime.net';
+
+export function logError(msg) {
+  window.lana.log(`SEOTECH: ${msg}`, {
+    debug: false,
+    implicitSampleRate: 100,
+    sampleRate: 100,
```
mokimo, line 8: Is this used on pages that don't see a lot of traffic? A sample rate of 100 sounds too high - usually 1 is enough

libs/utils/utils.js
```diff
@@ -849,13 +849,18 @@ export async function loadArea(area = document) {
...
     if (georouting === 'on') {
+      // eslint-disable-next-line import/no-cycle
       const { default: loadGeoRouting } = await import('../features/georoutingv2/georoutingv2.js');
       loadGeoRouting(config, createTag, getMetadata, loadBlock, loadStyle);
     }
     const appendage = getMetadata('title-append');
     if (appendage) {
       import('../features/title-append/title-append.js').then((module) => module.default(appendage));
     }
+    const seotechVideoUrl = getMetadata('seotech-video-url');
```
mokimo, line 860: Why have this as metadata property, rather than just defining it as block?

libs/utils/utils.js
```diff
@@ -849,13 +849,18 @@ export async function loadArea(area = document) {
   if (isDoc) {
     const georouting = getMetadata('georouting') || config.geoRouting;
     if (georouting === 'on') {
+      // eslint-disable-next-line import/no-cycle
       const { default: loadGeoRouting } = await import('../features/georoutingv2/georoutingv2.js');
       loadGeoRouting(config, createTag, getMetadata, loadBlock, loadStyle);
     }
     const appendage = getMetadata('title-append');
```
mokimo, line 856: Since this `getMetadata`-> loadFeature pattern is repeated a few times now - should we think about having this a bit more generic?

Something such as `metafeature-seotech-video-url` and everything that's prefixed as "metafeature" will try to load the feature `seotech-video-url`

Maybe a bit outside of the scope of this PR but we could shave a few LOC of the utils and future-proof this. Wdyt? //cc @chrischrischris

overmyheadandbody (approved):

test/utils/utils.test.js
```diff
@@ -400,4 +400,13 @@ describe('Utils', () => {
       expect(document.title).to.equal(expected);
     });
   });
+
+  describe('seotech', async () => {
+    beforeEach(async () => {
+      document.head.innerHTML = await readFile({ path: './mocks/head-seotech-video.html' });
+    });
+    it('should import and call default with falsey value', async () => {
+      await utils.loadArea();
```
JasonHowellSlavin, line 409: Should there be an expect statement in here?

libs/utils/utils.js
hparra: I noticed this as well but it's out of scope.

I would wait until this area gets bigger. I think it's too early to create an abstraction here. For example, in some cases more than one flag (feature on + feature data) could trigger the branch.

libs/features/seotech/seotech.js
hparra: This feature won't be used at scale until late Q4/Q1. We want all the errors logged so I can get some usefulness from reporting. This also makes it easier to dev an alert, e.g. "/this/page can't find /this/video" x 10. I can always come back later and lower these.

Slightly OT, but I've checked LANA and I'm concerned Milo is sampling far too little by default.

libs/features/seotech/seotech.js
```diff
@@ -0,0 +1,34 @@
...
+  });
+}
+
+export async function getVideoObject(url, seotechAPIUrl) {
+  if (!url) throw new Error('URL undefined');
+  const videosUrl = `${seotechAPIUrl}/api/v1/web/seotech/getVideoObject?url=${url}`;
+  const resp = await fetch(videosUrl, { headers: { 'Content-Type': 'application/json' } });
+  const body = await resp?.json();
+  if (!resp.ok) {
+    throw new Error(body?.error);
```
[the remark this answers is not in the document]
hparra: [97cb0e8](https://github.com/adobecom/milo/pull/1065/commits/97cb0e8ccb8534511fcb2802ed9650c1d5ec5997)

test/utils/utils.test.js
hparra: [cde384d](https://github.com/adobecom/milo/pull/1065/commits/cde384da3990ca7678cb02703f6b24819b72dccb)

mokimo (approved):

libs/features/seotech/seotech.js
```diff
@@ -0,0 +1,34 @@
...
+  window.lana?.log(`SEOTECH: ${msg}`, {
+    debug: false,
+    implicitSampleRate: 100,
+    sampleRate: 100,
+    tags: 'errorType=seotech',
+  });
+}
+
+export async function getVideoObject(url, seotechAPIUrl) {
+  const videoUrl = new URL(url)?.href;
```
overmyheadandbody, line 14: Although this is called from within a `try/catch` block and it definitely works, if you move some things around in the future and this gets taken out of the `try/catch` block, calling this method will throw errors if `url` is not defined or is invalid. It's definitely a matter or preference, but I usually go for checking the issue where it might occur.

overmyheadandbody (approved):

libs/features/seotech/seotech.js
mokimo: Yeah, as long as you have your use-case in mind that's good 👍 

As for the OT: Milo currently gets a lot of traction and a lot of projects that go live / will go live very soon. On the AEM lana logs things got flooded really fast, so it's just a matter of time until milo will see equal amounts of logs

libs/features/seotech/seotech.js
hparra: If that happens then the related test will fail (see L18 in seotech.test.js), but more to your point, the test will fail because there is a test to begin with -- it's easy to forget to write it since coverage tools count try/catch once, even though it wraps two statements that will branch. I'm not the biggest exception fan either for this reason, but it was useful here.

PR #1066, opened 2023-08-03
Mep button on target only

vgoodric:
If a page has the mep parameter, the MEP button should show.  This was not happening if the following conditions were both true:
* Target was on
* There were no manifests
If there were manifests or Target was off, the button properly showed.

**Test URLs:**
- Before: https://main--homepage--adobecom.hlx.page/?mep
- After: https://main--homepage--adobecom.hlx.page/?milolibs=mep-button-on-target-only&mep

### DIFF
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -475,10 +475,16 @@ function cleanManifestList(manifests) {
 }
 
 export async function applyPers(manifests) {
-  if (!manifests?.length) return;
-  const cleanedManifests = cleanManifestList(manifests);
-
   const config = getConfig();
+
+  if (!manifests?.length) {
+    if (config.mep?.preview) {
+      const { default: decoratePreviewMode } = await import('./preview.js');
+      decoratePreviewMode([]);
+    }
+    return;
+  }
+  const cleanedManifests = cleanManifestList(manifests);
 
   let results = [];
   for (const manifest of cleanedManifests) {
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -474,14 +474,18 @@ function cleanManifestList(manifests) {
   return cleanedList;
 }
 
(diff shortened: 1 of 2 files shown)


libs/features/personalization/personalization.js
```diff
@@ -475,11 +475,17 @@ function cleanManifestList(manifests) {
 }
 
 export async function applyPers(manifests) {
-  if (!manifests?.length) return;
-  const cleanedManifests = cleanManifestList(manifests);
-
   const config = getConfig();
 
+  if (!manifests?.length) {
+    if (config.mep?.preview) {
```
mokimo, line 481: Should we check if the query parameter is set? The button also shows without it https://main--homepage--adobecom.hlx.page/?milolibs=mep-button-on-target-only

overmyheadandbody: I think this happens here:

https://github.com/adobecom/milo/blob/31810376b5856facde23a3654591383dfe7c1178/libs/features/personalization/add-preview-to-config.js#L10-L14

Seems that on preview pages, the presence of the `mep` query parameter is totally irelevant. You would need to set `mepButton=off` to avoid loading the button: https://main--homepage--adobecom.hlx.page/?milolibs=mep-button-on-target-only&mepButton=off. I also see that it doesn't load on `.live` pages.

overmyheadandbody (approved):

vgoodric:
Sorry, I should have prefaced this with the intended behavior:

- The button should always appear if they use the mep parameter.
- It should always be off if they have mepButton=off (even if they also have the mep parameter.)
- It auto appears if it is
1. preview page
2. has personalization or Target enabled in the page's metadata.

The logic is that the MEP might be altering the page in a way you do not realize without the button.  So we want to be clear if MEP is currently applying changes.  But if neither Target nor Personalization are on, you do not need to worry about MEP.

chrischrischris (approved):

PR #1069, opened 2023-08-04
MWPW-134664: Safeguard autoblocks

mokimo:
### Description
Faulty links can lead to the whole page crashing. While it makes the issue visible, so no "faulty" links should be authored - with all the migrations and automatic imports going on this is hard to control.  We don't want end users to end up on white pages due to JS errors.

Resolves: [MWPW-134664](https://jira.corp.adobe.com/browse/MWPW-134664)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page?martech=off
- Before: https://main--helpx-internal--adobecom.hlx.page/content/help/en/aem-forms/internal-kb/not-able-to-edit-adaptive-form-on-aem-forms
- After: https://main--helpx-internal--adobecom.hlx.page/content/help/en/aem-forms/internal-kb/not-able-to-edit-adaptive-form-on-aem-forms?milolibs=do-not-error-autoblocks--milo--mokimo


overmyheadandbody (approved):
This does the job, but I'm still debating whether we need the `try/catch` around the whole body of the method. I'd have just wrapped the `url = new URL(a?.href);` part in a `try/catch` block, declaring a `let url` on top and return/log to LANA if that fails. Did you see other areas of the logic that might throw errors? I'm just not sure I'm a fan of this pattern and I fear it might spread to more methods - if one line could potentially fail, wrap the whole thing in a `try/catch`. I'd rather focus on improving the area that can cause issues, but I get it's a personal preference :)

mokimo:
I'd agree that just wrapping the URL would be preferable - but in this instance the whole method handles a lot of logic, so IMO it's safer to just wrap the whole function. Especially given that this method failing leads to a potentially completely white page.

I changed the lana log message to reflect that _anything_ in the method could fail.

narcis-radu (commented):
I second @overmyheadandbody's feedback and think we should only wrap the URL declaration in try / catch.
We can add more safeguards in the code for optional chaining if really needed.

mokimo:
Okie, since both of you agree on that I changed the check accordingly

narcis-radu (approved):

libs/utils/utils.js
```diff
@@ -461,7 +461,13 @@ export function decorateImageLinks(el) {
...
   const config = getConfig();
   const { hostname } = window.location;
-  const url = new URL(a.href);
+  let url;
+  try {
+    url = new URL(a.href);
+  } catch (e) {
+    window.lana?.log(`Cannot make URL from decorateAutoBlock - ${a?.href}: ${e.toString()}`);
+  }
+  if (!url) return false;
```
hparra, line 470: nit: save a branch by returning in catch. new URL() will not return null: https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

hparra (approved):

PR #1071, opened 2023-08-04
MWPW-134269 - Adding nested fragments based upon the primary product as well as secondary

sabyamon:
* Adding nested fragments based upon the primary product as well as secondary
* Allow comma-separated fragment lists in result-fragments

Resolves:  https://jira.corp.adobe.com/browse/MWPW-134269 and https://jira.corp.adobe.com/browse/MWPW-134273

**Test URLs:**
- Before: https://uar-integration--milo--adobecom.hlx.page/drafts/sabya/quiz-2/?martech=off&hideGeorouting=on
- After: https://uar-nested-frag-by-primary--milo--sabyamon.hlx.page/drafts/sabya/quiz-2/?martech=off&hideGeorouting=on


libs/blocks/quiz/utils.js
```diff
@@ -92,20 +93,21 @@ export const handleResultFlow = async (answers = []) => {
...
 ) => {
-  const nestedFrags = resultData.matchedResults[0]['nested-fragments'];
+  const nestedFragsPrimary = resultData.matchedResults[0]['nested-fragments-primary'];
+  const nestedFragsSecondary = resultData.matchedResults[0]['nested-fragments-secondary'];
   const structureFrags = resultData.matchedResults[0]['basic-fragments'];
 
   const structureFragsArray = structureFrags?.split(',');
-  const nestedFragsArray = nestedFrags?.split(',');
-
+  const nestedFragsPrimaryArray = nestedFragsPrimary?.split(',');
```
echampio-at-adobe, line 109: When splitting, would it be possible to trim at the same time to ditch any whitespace? For example:
const nestedFragsPrimaryArray = nestedFragsPrimary?.split(',').map(item => item.trim());

echampio-at-adobe (approved):
Seems to be working when I step through various products. I had a minor question about when to trim the fragment strings. If you prefer, you can leave things as they are, where you're trimming whitespace as a separate step.

rgclayton (approved):

libs/blocks/quiz/utils.js
```diff
@@ -152,41 +156,73 @@ const structuredFragments = (
...
+
+    function insertFragment() {
+      if (row[fragKey]) {
+        row[fragKey].split(',').forEach((val) => {
+          fragArray.push(val.trim());
+        });
+      }
+    }
+
+    function isMatchingProductForFrag() {
```
mokimo, line 211: I'd move the two helper functions out of the `getNestedFragments` one benefit is that you don't re-declare those functions within the `getNestedFragments` scope every time u call it.

mokimo (approved):
There's some quiz related tests failing, would be good if you can look at those before merging.

Also, there are a lot of linter issues - would be cool if you can see if you can fix some when you already working on those files

libs/blocks/quiz/utils.js
```diff
@@ -152,41 +156,73 @@ const structuredFragments = (
...
-      } else if (secondaryProductCodes.length > 0 && secondaryProductCodes.includes(row.product)) {
-        // Get nested frags for all the secondary products.
-        if (row[fragKey]) {
-          fragArray.push(row[fragKey]);
-        }
-      }
-    });
-    nestedObject[fragKey] = fragArray;
+    // eslint-disable-next-line max-len
+    nestedObject[fragKey] = getNestedFragments(resultResources, isUmbrella, primaryProducts, secondaryProducts, fragKey, true);
```
overmyheadandbody, line 172: Since this has 6 arguments, it might be better to use an object. Changing anything becomes difficult, since there's a strict order and it's difficult to understand what certain arguments refer to. What does `true` or `false` achieve? Something like
```
getNestedFragments({
  resultResources,
  isUmbrella,
  primaryProducts,
  secondaryProducts,
  fragKey,
  fetchForPrimaryProducts: true,
});
```
is a bit more clear and more maintainable, as you could omit certain arguments, change their order, etc.

libs/blocks/quiz/utils.js
```diff
@@ -152,41 +156,73 @@ const structuredFragments = (
...
+  });
+  nestedFragsSecondaryArray?.forEach((frag) => {
+    if (!frag) return;
```
overmyheadandbody, line 175: Can `nestedFragsSecondaryArray` contain `undefined` or `null` values? Otherwise this looks like it might not be needed. Or you could check whether it's a string, since on the next line you're calling `trim`

libs/blocks/quiz/utils.js
```diff
@@ -152,41 +156,73 @@ const structuredFragments = (
...
+const getNestedFragments = (
+  resultResources,
+  isUmbrella,
+  primaryProducts,
+  secondaryProducts,
+  fragKey,
+  fetchForPrimaryProducts,
+) => {
+  const fragArray = [];
+  resultResources.data.forEach((row) => {
```
overmyheadandbody, line 192: Would this need some optional chaining in case `resultResources` or its `data` property don't exist?

libs/blocks/quiz/utils.js
```diff
@@ -152,41 +156,73 @@ const structuredFragments = (
...
+    if (isUmbrella) {
+      // Get nested frags for all the primary products.
+      if (primaryProducts.length > 0 && primaryProducts.includes(row.product)) {
+        insertFragment();
+      }
+    } else if (isMatchingProductForFrag()) {
+      // Get nested frags for all the primary and secondary products.
+      insertFragment();
+    }
```
overmyheadandbody, line 201: In the end, you're calling `insertFragment` for each of the branches. You could go for something like
```
if ((isUmbrella && primaryProducts?.length && primaryProducts.includes(row.product))
  || isMatchingProductForFrag()) {
    insertFragment();
}
```
You might also want to check if `row.product` actually exists, in case both `primaryProducts` and `row.products` contain `undefined` entries

libs/blocks/quiz/utils.js
```diff
@@ -152,41 +156,73 @@ const structuredFragments = (
...
+      }
+    }
+
+    function isMatchingProductForFrag() {
+      return (!fetchForPrimaryProducts
+              && secondaryProducts.length > 0
+              && secondaryProducts.includes(row.product))
+            || (fetchForPrimaryProducts
+              && primaryProducts.length > 0
+              && primaryProducts.includes(row.product));
```
overmyheadandbody, line 217: Indenting looks a bit off or it may be just Git styling it this way, but it might be worth a check

libs/blocks/quiz/utils.js
```diff
@@ -242,11 +278,11 @@ const parseResultData = async (answers) => {
   return rObj;
 };
 
-const getRecomandationResults = (selectedDestination, deafult) =>
-  (selectedDestination.length ? selectedDestination : deafult);
+// eslint-disable-next-line max-len
+const getRecomandationResults = (selectedDestination, deafult) => (selectedDestination.length ? selectedDestination : deafult);
```
overmyheadandbody, line 282: Not sure whether this grants an ESLint exception, this could be simply split up like
```
const getRecomandationResults = (selectedDestination, deafult) => selectedDestination.length
  ? selectedDestination : default;
```
Also, there's a typo, it should probably be _default_, not _deafult_

libs/blocks/quiz/utils.js
```diff
@@ -378,7 +414,7 @@ export const getAnalyticsDataForBtn = (selectedQuestion, selectedCards) => {
   }
 };
 
-const getAnalyticsDataForLocalStorage = (answers) => {
+export const getAnalyticsDataForLocalStorage = (answers) => {
```
overmyheadandbody, line 417: I see a bunch of `export` declarations, but no files where these are imported. Should they actually be exported in the first place?

fullcolorcoder (approved):

libs/blocks/quiz/utils.js
sabyamon: I love that idea of trimming when splitting itself. The only reason I did not go that route is to not have to iterate the `nestedFragsPrimaryArray` array twice. First with `.map(item => item.trim());` and second in `nestedFragments` function `nestedFragsPrimaryArray?.forEach((frag)`. While I know the size of the `nestedFragsPrimaryArray` will be small in most of the cases and the performance difference should be negligible.

libs/blocks/quiz/utils.js
sabyamon: Thanks, Rares for checking. Fixed the typo and the eslint issue.

libs/blocks/quiz/utils.js
sabyamon: Export statements are for the upcoming PR for the UT for this file. No effect for this file tbh.

libs/blocks/quiz/utils.js
sabyamon: hi @mokimo, these two helper functions are logically very much tied to `getNestedFragments` which is called once in its lifetime (and only when the quiz flow ends and gets redirected to the result page). That's the main reason, I have declared them inside the `getNestedFragments`. We can sync up offline if you strongly feel against it. Thanks a lot for reviewing.

libs/blocks/quiz/utils.js
sabyamon: `nestedFragsSecondaryArray` can contain `undefined` for the empty cells in `results.xlsx`.

libs/blocks/quiz/utils.js
sabyamon: For sure. Thanks for catching that. Updated.

libs/blocks/quiz/utils.js
sabyamon: Updated.

libs/blocks/quiz/utils.js
sabyamon: Refactored the entire way we are fetching nested frags which has surely simplified the flow a lot. Would be great if you want to have a second look.

PR #1072, opened 2023-08-04
[Workflow] MWPW-134543 Dispatch DC workflow when main is changed

TsayAdobe:
* It creates a workflow dispatch to DC repo when there is a change to main
* DC repo will trigger necessary tests to ensure the change in Milo doesn't break DC pages
* Will add PRs later when shift-left is needed
* Other consuming teams can add their dispatches to jobs

Resolves: [MWPW-134543](https://jira.corp.adobe.com/browse/MWPW-134543)

**Test URLs:**
- N/A

### DIFF
--- a/.github/workflows/dispatch.yml
+++ b/.github/workflows/dispatch.yml
@@ -1,0 +1,32 @@
+name: Dispatch workflows
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  dispatch-dc:
+    name: Dispatch DC workflow
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v3    
+      - uses: dorny/paths-filter@v2
+        id: changes
+        with:
+          filters: |
+            src:
+              - 'libs/**'    
+      - if: steps.changes.outputs.src == 'true'
+        name: Trigger DC Workflow
+        uses: actions/github-script@v6
+        with:
+          github-token: ${{ secrets.DC_PAT }}
+          script: |
+            github.rest.actions.createWorkflowDispatch({
+              owner: 'adobecom',
+              repo: 'dc',
+              workflow_id: 'test-milo.yml',
+              ref: 'main',
+            }) 


mokimo (approved):
Cool, do you think other consumers would benefit from this as well?

.github/workflows/dispatch.yml
```diff
@@ -0,0 +1,32 @@
...
+      - main
+
+jobs:
+  dispatch-dc:
+    name: Dispatch DC workflow
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v3    
+      - uses: dorny/paths-filter@v2
```
JackySun9, line 15: do you have v3 for this

.github/workflows/dispatch.yml
```diff
@@ -0,0 +1,32 @@
...
+          filters: |
+            src:
+              - 'libs/**'    
+      - if: steps.changes.outputs.src == 'true'
+        name: Trigger DC Workflow
+        uses: actions/github-script@v6
+        with:
+          github-token: ${{ secrets.DC_PAT }}
+          script: |
+            github.rest.actions.createWorkflowDispatch({
```
JackySun9, line 27: do you need error handling for this

.github/workflows/dispatch.yml
TsayAdobe: There is only v2.
https://github.com/dorny/paths-filter

.github/workflows/dispatch.yml
TsayAdobe: The action library actions/github-script@v6 should have handled most errors. Not sure what is not handled. We will handle it when we see it.

TsayAdobe:
> Cool, do you think other consumers would benefit from this as well?

Yes, if they want to do some actions or tests in their repo when Milo is changed. They can add a workflow dispatch in this script.

JackySun9 (approved):

PR #1080, opened 2023-08-05
MWPW-133055 Sticky bottom section functionality enhancement

aishwaryamathuria:
Enhance the sticky bottom section to have following capabilities -
1. Make the section visible once scrolled past a particular part from top of the page.
2. The enhancement is required with bottom promobar.
3. Section hides once user scrolls to the footer.

Resolves: [MWPW-133055](https://jira.corp.adobe.com/browse/MWPW-133055)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/mathuria/promobar/sticky-bottom?martech=off
- After: https://stickypromo--milo--aishwaryamathuria.hlx.page/drafts/mathuria/promobar/sticky-bottom?martech=off


libs/blocks/section-metadata/section-metadata.js
```diff
@@ -34,6 +36,37 @@ function handleTopHeight(section) {
   const headerHeight = document.querySelector('header').offsetHeight;
   section.style.top = `${headerHeight}px`;
 }
+
+export function promoIntersectObserve(el, startPoint, options = {}) {
+  let abovePromoStart = false;
+  const io = new IntersectionObserver((entries, observer) => {
+    entries.forEach(async (entry) => {
```
mokimo, line 43: You aren't doing any async operations, you can remove the `async` keyword

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -34,6 +36,37 @@ function handleTopHeight(section) {
...
+      if (el.classList.contains('close-sticky-section')) return observer.unobserve(entry.target);
+      const isPromoStart = startPoint ? entry.target === startPoint : null;
+      abovePromoStart = isPromoStart
+        ? (entry.isIntersecting || entry.boundingClientRect.y > 0)
+        : abovePromoStart;
+      if (entry.isIntersecting || (!entry.isIntersecting && isPromoStart && abovePromoStart)) el.classList.add('hide-sticky-section');
```
mokimo, line 49: Instead of using a ternary operation that sets a boolean or null, you can simply use a logical AND
The whole condition for isIntersecting is a bit hard to read, maybe this simplifies it a bit:
```suggestion
      const isPromoStart = startPoint && entry.target === startPoint;
      abovePromoStart = isPromoStart ? (entry.isIntersecting || entry.boundingClientRect.y > 0) : abovePromoStart;
      if (entry.isIntersecting || (isPromoStart && abovePromoStart)) el.classList.add('hide-sticky-section');
```

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -34,6 +36,37 @@ function handleTopHeight(section) {
...
+      else if (!entry.isIntersecting && !abovePromoStart) el.classList.remove('hide-sticky-section');
+      return el;
```
mokimo, line 51: You can remove the return el, it's doing nothing other than making the linter happy.
Either disable the rule or change `if (el.classList.contains('close-sticky-section')) return observer.unobserve(entry.target);` to
```
if (el.classList.contains('close-sticky-section')) {
   observer.unobserve(entry.target);
   return 
}
```

libs/blocks/section-metadata/section-metadata.js
overmyheadandbody: And anyway, `forEach` and `async` don't work together, so this definitely doesn't achieve anything.

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -34,6 +36,37 @@ function handleTopHeight(section) {
...
+    });
+  }, options);
+  return io;
+}
+
+async function handleStickyPromobar(section) {
```
overmyheadandbody, line 57: To @mokimo 's earlier point, I don't see any `await` declarations, `async` is probably redundant. The `await` on L81 might also be redundant. Am I missing something? Is there some async code going on somewhere?

test/blocks/section-metadata/section-meta.test.js
```diff
@@ -70,6 +70,16 @@ describe('Section Metdata', () => {
     expect(main.firstElementChild).to.be.eql(sec);
   });
 
+  it('add promobar behaviour to section', async () => {
+    const main = document.querySelector('main');
+    const sec = document.querySelector('.section.sticky-bottom .promobar').closest('.section');
+    const sm = sec.querySelector('.section-metadata');
+    await init(sm);
+    console.log(main.lastElementChild);
```
overmyheadandbody, line 78: This pollutes the test runner console, it should be removed.

libs/blocks/section-metadata/section-metadata.js
aishwaryamathuria: @mokimo @overmyheadandbody I have made the requested changes.

libs/blocks/section-metadata/section-metadata.js
aishwaryamathuria: @mokimo I have made the requested changes.

libs/blocks/section-metadata/section-metadata.js
aishwaryamathuria: @mokimo I have made the requested changes.

libs/blocks/section-metadata/section-metadata.js
aishwaryamathuria: @mokimo @overmyheadandbody I have made the requested changes.

test/blocks/section-metadata/section-meta.test.js
aishwaryamathuria: @overmyheadandbody I have made the requested changes.

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -34,6 +36,39 @@ function handleTopHeight(section) {
   const headerHeight = document.querySelector('header').offsetHeight;
   section.style.top = `${headerHeight}px`;
 }
+
+export function promoIntersectObserve(el, startPoint, options = {}) {
+  let abovePromoStart = false;
```
overmyheadandbody, line 41: Since you don't need this outside the `io` scope, this can be removed from here and then L49-L51 could simply be
```
const abovePromoStart = isPromoStart
        ? (entry.isIntersecting || entry.boundingClientRect.y > 0)
        : false;
```

libs/blocks/section-metadata/section-metadata.css
```diff
@@ -17,6 +17,11 @@
   color: rgb(34 34 34);
 }
 
+.section.hide-sticky-section,
+.section.close-sticky-section {
```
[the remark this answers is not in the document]
aishwaryamathuria: @overmyheadandbody The promo sections can have a close button as per the proposed designs and in that case the observer would need to be removed. When the promobar element close action is triggerd it can add the class close-sticky-section to the section. This would hide the section and the intersection observer would disable the observers if the class is applied to the section.
hide sticky section would not enforce the observers to unobserve and the class would only toggle in case of scrolling.

libs/blocks/section-metadata/section-metadata.js
aishwaryamathuria: @overmyheadandbody The abovePromoStart is state of the scroll that is set only when the intersection observer is triggered by promo placeholder. 
The state set by promo placeholder is used when the observer is triggered for the footer.
For case when the promobar needs to be always visible this needs to be false always. Hence we are using one variable to save the state of scroll to be used by intersection observer.
The change would make the section visible every time the page loads because this would mean that the abovePromoStart is true which would indicate to the footer that the user has scrolled past the blocks after which the section should be made visible and it would set the display block for the section on load .

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -34,6 +36,39 @@ function handleTopHeight(section) {
   const headerHeight = document.querySelector('header').offsetHeight;
   section.style.top = `${headerHeight}px`;
 }
+
+export function promoIntersectObserve(el, startPoint, options = {}) {
```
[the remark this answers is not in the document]
aishwaryamathuria: @overmyheadandbody I have made the requested change.

spadmasa:
Validation done by QA  on the below urls 👍 
Before :
https://main--milo--adobecom.hlx.page/drafts/mathuria/promobar/sticky-bottom?martech=off
After : https://stickypromo--milo--aishwaryamathuria.hlx.page/drafts/mathuria/promobar/sticky-bottom?martech=off

In CC branch 👍 
sticky bottom : https://main--cc--adobecom.hlx.page/drafts/tat28144/promobar/promos?milolibs=stickypromo--milo--aishwaryamathuria

sticky bottom with multiple sections :
https://main--cc--adobecom.hlx.page/drafts/tat28144/discover?milolibs=stickypromo--milo--aishwaryamathuria

Stickybottom always show :
https://main--cc--adobecom.hlx.page/drafts/tat28144/promobar/sticky-bottom-always-show?milolibs=stickypromo--milo--aishwaryamathuria

Stickypromo at the top
https://main--cc--adobecom.hlx.page/drafts/tat28144/promobar/sticky-top?milolibs=stickypromo--milo--aishwaryamathuria

mokimo (approved):
Any documentation that needs to be updated? https://milo.adobe.com/docs/authoring/metadata

libs/blocks/section-metadata/section-metadata.js
overmyheadandbody: I wasn't questioning the logic, just that a variable is declared here, only to be updated on L49. But it's not a showstopper, just a minor improvement :)

overmyheadandbody (approved):

libs/blocks/section-metadata/section-metadata.js
chrischrischris: Agree with @overmyheadandbody - I don't see why `abovePromoStart` shouldn't just be declared on line 49.

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -34,6 +36,39 @@ function handleTopHeight(section) {
...
+
+function promoIntersectObserve(el, startPoint, options = {}) {
+  let abovePromoStart = false;
+  const io = new IntersectionObserver((entries, observer) => {
+    entries.forEach((entry) => {
+      if (el.classList.contains('close-sticky-section')) {
+        observer.unobserve(entry.target);
+        return;
+      }
+      const isPromoStart = startPoint && entry.target === startPoint;
```
chrischrischris, line 48: I don't think it's possible for `entry.target` to be `null` so this can be simplified to
```suggestion
      const isPromoStart = entry.target === startPoint;
```

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -34,6 +36,39 @@ function handleTopHeight(section) {
...
+  const main = document.querySelector('main');
+  section.classList.add('hide-sticky-section');
+  let startPoint = null;
+  if (main.children[0] !== section) {
+    startPoint = createTag('div', { class: 'section show-sticky-section' });
+    main.insertBefore(startPoint, section);
+  }
+  const io = promoIntersectObserve(section, startPoint);
+  if (startPoint) io.observe(startPoint);
+  io.observe(document.querySelector('footer'));
```
chrischrischris, line 69: If `startPoint === null` so there is no observation of `startPoint`, do we still need to observe the footer?

Also might be a bit clearer to name startPoint something like `stickySectionEl`

aishwaryamathuria: @chrischrischris We have 2 use cases of sticky bottom promobar shared with us -
1. The promobar will be visible after a section is scrolled up and hides on footer
2. The promobar will be visible always and hides on footer
If our promobar is authored at the top then in condition 2 the startPoint will be null but we will still need to observe the footer to hide the promobar.

I have made the changes in the PR to change variable name from startPoint to stickySectionEl

libs/blocks/section-metadata/section-metadata.js
aishwaryamathuria: @chrischrischris 

1. Comment 1
The promobars can have close buttons for some variations. Since the block is called after section-metadata sticky promobar handler, we cannot observe the elements inside the block and add an event listener to a possible close button in block. In case the block has a close button, it can add the event listener to the close button and mark the section it is inside to close the promobar behaviour via class 'close-sticky-promobar'

2. Comment 2
I have moved the assignment of abovePromoStart inside intersection observer function and removed it from outside.

3. Comment 3
entry.target cannot be null but it can have 2 values - startPoint or footer. If it was triggered due to startPoint we do not need to check the scroll point and can directly hide if it is intersecting.

chrischrischris (changes requested):
I hate to pivot late in the PR review process, but after thinking and discussing with the milo team, lets move the sticky code out of section-metadata.js into it's own file.  The file can live in blocks/section-metadata and I'm pretty sure can be lazy init'd.  Something like:

line 93:
if (sticky) {
  import('./sticky.js').then(({ default: handleStickySection }) => handleStickySection(sticky, section));
}

The main reason for breaking it out is that section-metadata.js is loaded on every single page, so it'd be preferable to keep it as small as possible.

I think that the sticky css in section-metadata.css can stay there as it's not many lines.

aishwaryamathuria:
> I hate to pivot late in the PR review process, but after thinking and discussing with the milo team, lets move the sticky code out of section-metadata.js into it's own file. The file can live in blocks/section-metadata and I'm pretty sure can be lazy init'd. Something like:
> 
> line 93:
> 
> ```
> if (sticky) {
>   import('./sticky.js').then(({ default: handleStickySection }) => handleStickySection(sticky, section));
> }
> ```
> 
> The main reason for breaking it out is that section-metadata.js is loaded on every single page, so it'd be preferable to keep it as small as possible.
> 
> I think that the sticky css in section-metadata.css can stay there as it's not many lines.

@chrischrischris I have made the requested changes and created a new section-sticky.js file with all the code to apply sticky functionality to a section. It will be dynamically imported in section-metadata to apply thw functionality when specified.

chrischrischris (approved):

PR #1081, opened 2023-08-07
Anchor links

suhjainadobe:
If the URL has anchor links, the browser is not scrolling automatically to that particular element.

Resolves:
 https://jira.corp.adobe.com/browse/MWPW-131971
https://jira.corp.adobe.com/browse/MWPW-133331

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/suhjain/blade-runner-2049s-oscar-winning-texturing-workflow-at-framestore#the-solar-farm
- After: https://anchorlinks--milo--suhjainadobe.hlx.page/drafts/suhjain/blade-runner-2049s-oscar-winning-texturing-workflow-at-framestore#the-solar-farm


libs/scripts/scripts.js
```diff
@@ -149,7 +150,11 @@ const eagerLoad = (img) => {
 
 (async function loadPage() {
   performance.mark('loadpage');
+  window.addEventListener('load', (e) => {
+    e.preventDefault();
+  });
   setConfig(config);
   loadLana({ clientId: 'milo' });
   await loadArea();
+  scrollToHashedElement();
```
mokimo, line 159: The scripts are defined on a per-consumer basis - so every consumer would need to redefine this, I don't think this is the right place to call this function, otherwise each consumer would need to repeat this

libs/utils/utils.js
```diff
@@ -932,3 +932,22 @@ export function loadLana(options = {}) {
...
+  const hash = window.location.hash;
+  const gnavElement = document.querySelector('.global-navigation');
+  if (hash) {
+    const elementId = hash.slice(1);
+    const targetElement = document.querySelector(`#${elementId}:not(.dialog-modal)`);
+    if (targetElement) {
+      if (gnavElement) {
+        let bufferHeight = gnavElement.offsetHeight;
+        targetElement.style.paddingTop = `${bufferHeight}px`;
+        targetElement.style.marginTop = `-${bufferHeight}px`;
```
mokimo, line 946: Why is that needed? That seems like an evil hack

mokimo (commented):
The linter has a lot of complaints make sure to fix those and check if your vscode/lint setup is running correctly

libs/utils/utils.js
```diff
@@ -932,3 +932,23 @@ export function loadLana(options = {}) {
...
+    const position = targetElement.getBoundingClientRect();
+    let bufferHeight = 0;
+    if (targetElement) {
+      if (gnavElement) {
+        bufferHeight = gnavElement.offsetHeight;
+      }
+      window.scrollTo({
+        top: position.top - bufferHeight,
+        behavior: 'smooth',
+      });
```
mokimo, line 951:
```suggestion
  if (hash) {
    const elementId = hash.slice(1);
    const targetElement = document.querySelector(`#${elementId}:not(.dialog-modal)`);
    if (targetElement) {
      const position = targetElement.getBoundingClientRect();
      let bufferHeight = document.querySelector('.global-navigation')?.offsetHeight || 0
      window.scrollTo({
        top: position.top - bufferHeight,
        behavior: 'smooth',
      });
```

suhjainadobe: done

libs/scripts/scripts.js
```diff
@@ -14,6 +14,7 @@ import {
   loadArea,
   loadLana,
   setConfig,
+  scrollToHashedElement,
```
mokimo, line 17: Unused import

libs/utils/utils.js
```diff
@@ -880,6 +897,7 @@ export async function loadArea(area = document) {
 
   // Load everything that can be deferred until after all blocks load.
   await loadDeferred(area, areaBlocks, config);
+  scrollToHashedElement();
```
mokimo, line 900: I'm a bit concerned about the timing of this, but I'll leave it up to the milo core team to have the final call on this

test/utils/utils.test.js
```diff
@@ -417,4 +417,44 @@ describe('Utils', () => {
...
+      document.body.appendChild(div);
+      window.location.hash = '#not-block';
+      window.scrollTo = () => {};
+      document.querySelector = () => ({ getBoundingClientRect: () => ({ top: 100 }) });
+    });
+
+    it('should scroll to the hashed element', () => {
+      window.scrollTo = (options) => {
+        expect(options.behavior).to.equal('smooth');
+      };
```
mokimo, line 434: Should this test setup be equal to the other ones setting a bool?
```
      window.scrollTo = () => {
        scrollToCalled = true;
      };
```

mokimo (approved):
Some last comments and would like for the milo core team to have some input on this

libs/utils/utils.js
suhjainadobe: Sorry @mokimo , this will not work, as load eventListener does not get trigger here and Safari is not working as expected. Will have to add it in script.js. Looking for other ways to handle this problem. scrollToHashedElement() is fine, it is working as expected but figuring out how to handle load eventListener without calling it in script.js

suhjainadobe:
Closing to check some changes of the load eventListener.

suhjainadobe:
@mokimo and @aishwaryamathuria : Could you please review it again.

libs/utils/utils.js
```diff
@@ -878,6 +891,9 @@ export async function loadArea(area = document) {
     delayed([getConfig, getMetadata, loadScript, loadStyle]);
   }
 
+  // placing the hash back to scroll the screen.
+  window.location.hash = currentHash;
```
chrischrischris, line 895: nit: only set this if `currentHash` has a value

chrischrischris (changes requested):
This is breaking modals - see:
https://anchorlinks--milo--suhjainadobe.hlx.page/drafts/suhjain/blade-runner-2049s-oscar-winning-texturing-workflow-at-framestore#playmodal

The hash keeps getting added and removed, causing modal to flicker / reload.  Closing the modal opens it up right away again.

libs/utils/utils.js
suhjainadobe: done

suhjainadobe:
@chrischrischris : The modal flickering is resolved. Please review again.

seanchoi-dev (approved):

suhjainadobe:
> On safari, it scrolls past the anchor title

@chrischrischris : Could you please let me know when is it not working? For me it is scrolling down before the title only.

chrischrischris:
Probably out of scope for this, but should a Header that's inside a tab that's not currently displayed activate that tab and scroll to it?

ryanmparrish:
@suhjainadobe  Just an FYI, I tested this page w/ a recently added block `marquee-anchors` which has some anchor link similarities. I don't see any issues w/ compatibility. 
Here's the test page https://anchorlinks--milo--suhjainadobe.hlx.page/drafts/rparrish/marquee-nav#how-to-block

suhjainadobe:
> @suhjainadobe Just an FYI, I tested this page w/ a recently added block `marquee-anchors` which has some anchor link similarities. I don't see any issues w/ compatibility.

@ryanmparrish : Here too I could see the issue intermittently. It was scrolling past the title in Safari.

libs/utils/utils.js
```diff
@@ -805,11 +818,15 @@ function decorateMeta() {
 }
 
 export async function loadArea(area = document) {
+  const currentHash = window.location.hash;
```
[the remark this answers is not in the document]
suhjainadobe: done

suhjainadobe:
> Probably out of scope for this, but should a Header that's inside a tab that's not currently displayed activate that tab and scroll to it?

If the header is not present in the active tab, then the page will not scroll down and activate the tab. We yet are not supporting this feature.

suhjainadobe:
@chrischrischris , @ryanmparrish : I have made some changes in the style.css file which accommodates the default scrolling in Safari. It also resolved [MWPW-133331](https://jira.corp.adobe.com/browse/MWPW-133331).

libs/styles/styles.css
```diff
@@ -86,31 +86,37 @@ header.global-navigation {
 h1 {
   font-size: var(--type-heading-xl-size);
   line-height: var(--type-heading-xl-lh);
+  scroll-margin-top: 94px;
```
mokimo, line 89: I don't think a hardcoded value will work long term

suhjainadobe: done

libs/styles/styles.css
```diff
@@ -86,31 +90,37 @@ header.global-navigation {
 h1 {
   font-size: var(--type-heading-xl-size);
   line-height: var(--type-heading-xl-lh);
+  scroll-margin-top: var(--feds-totalheight-nav);
```
mokimo, line 93: That's definitely better than a static value, though one Issue I forsee is when things are on top of the navigation.

There are already such requirements https://github.com/orgs/adobecom/discussions/1166

I don't think we necessarily need to fix this right now but we should keep it in mind, otherwise the scroll-location might be off in those cases.

mokimo: Can you explain a bit why the `scroll-margin-top` is needed and the top-offset isn't enough?
```
  window.scrollTo({
    top: topOffset - bufferHeight,
    behavior: 'smooth',
  });
```

suhjainadobe: On safari, the scrollbar is automatically initiated even before the scrollToHashedElement is initiated, so when it first scrolls down, the nav bar hides the title. Also to resolve [MWPW-133331](https://jira.corp.adobe.com/browse/MWPW-133331), we need this CSS property.

chrischrischris (approved):

chrischrischris (commented):
One nit: combine header selectors to save some file size:
```
h1, h2, h3, h4, h5, h6 {
  scroll-margin-top: var(--feds-totalheight-nav);
}
```

PR #1082, opened 2023-08-07
ost tool overlaps with sidekick.

sharath-kannan:
* Ost overlap with sidekick

Resolves: [MWPW-133826](https://jira.corp.adobe.com/browse/MWPW-133826)

**Test URLs:**
- After: https://mwpw-133826_ost_overlap--milo--sharath-kannan.hlx.page/tools/ost?martech=off&token=your_token

PS:kindly replace your_token by following the steps given in the below link
https://github.com/adobecom/milo/tree/main/libs/blocks/ost#development


libs/blocks/ost/ost.css
```diff
@@ -5,6 +5,7 @@ body.tool-ost .ost {
   left: 0;
   bottom: 0;
   right: 0;
+  margin-top: 5%;
```
yesil, line 8: can we use a definite value here,  50px should do it.

image

sharath-kannan: I have updated it now.

mokimo (approved):

yesil (approved):

PR #1086, opened 2023-08-07
[MWPW-129920] Marquee Large Updates 

drashti1712:
* Implementation of new updates for Marquee Large.
* Two new variants with min heights 560px (called standard variant) and 360px (called compact variant) are created. 
* Quiet Marquee updates will be pushed as a separate PR.
* XD link: https://www.dropbox.com/sh/ote8tkbnhpggkpn/AADvJ5JF-7zmm-yEA8ZbquA8a?dl=0

Resolves: [MWPW-129920](https://jira.corp.adobe.com/browse/MWPW-129920)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/drashti/marquee-std/marquee-large-final?martech=off
- After: https://marquee-block--milo--drashti1712.hlx.page/drafts/drashti/marquee-std/marquee-large-final?martech=off


ryanmparrish (changes requested):
There are a number of css linting issues i'm seeing and if at all possible can we avoid using taxing css selectors like `:not()`?

drashti1712:
> There are a number of css linting issues i'm seeing and if at all possible can we avoid using taxing css selectors like `:not()`?

The linting issues were resolved when the branch was updated, and I also removed the :not() selectors wherever possible.
Thanks!

Sartxi (approved):

ryanmparrish (changes requested):
@drashti1712 Thanks for looking at those scoped selectors. Good idea to have better performance first w/ regards to css selectors. 

I'm still seeing linting errors on the marquee.css. Big picture is the format of that file should follow somewhat the dom structure of the block. Example. the first line in the `marquee.css` file should be simply `.marquee` etc and cascade down the selector specificity. This contribution goes against this paradigm and makes it hard to maintain and read. A block css file should never start w/ a very specific selector scope. ie `.marquee.large .foreground.container .text`
See linting errors and order of selectors. 
Screen Shot 2023-08-09 at 2 07 26 PM

libs/blocks/marquee/marquee.css
```diff
@@ -1,9 +1,45 @@
...
+.marquee.large .container {
+  width: 100%;
+  padding: 0;
+  gap:0;
+}
+
+.marquee .icon-area.icon-area-multiple > :not(:last-child) img {
+  margin: 0 var(--spacing-s) 0 0;
+}
+
```
aishwaryamathuria, line 15: Can we use flex with gap and no shrink in place of margins?

libs/blocks/marquee/marquee.css
```diff
@@ -1,9 +1,45 @@
...
+.marquee.large .text .heading-xxl {
+  margin-bottom: var(--spacing-s);
+}
+
+.marquee.large .text .body-xl:not(.supplemental-text){
+  margin-bottom: 0;
```
aishwaryamathuria, line 21: supplemental-text is already overriding margin bottom to 0 below in this css. We do not neet additional :not operator

libs/blocks/marquee/marquee.css
```diff
@@ -1,9 +1,45 @@
...
+}
+
+.marquee.large .body-xl .icon {
+  margin: var(--spacing-xs) var(--spacing-xs) var(--spacing-xs) 0;
```
aishwaryamathuria, line 25: If this being used somewhere? icon-area does not have a body-xl attached to its classlist and I could not find .icon class on any element. Am I missing something?

libs/blocks/marquee/marquee.css
```diff
@@ -201,6 +232,7 @@
 
 .marquee.split .foreground,
 .marquee.small.split .foreground {
+  width: var(--grid-container-width);
```
aishwaryamathuria, line 235: Can we remove this?

libs/blocks/marquee/marquee.css
```diff
@@ -481,7 +507,37 @@
     display: flex;
   }
 
+  .marquee.large:not(.compact) .foreground.container .text {
```
aishwaryamathuria, line 510: :not(.compact) not required here

libs/blocks/marquee/marquee.js
```diff
@@ -147,7 +147,9 @@ export default function init(el) {
   if (firstDivInForeground.classList.contains('media')) el.classList.add('row-reversed');
 
   const size = getBlockSize(el);
-  decorateButtons(text, size === 'large' ? 'button-xl' : 'button-l');
+  decorateButtons(text, (size === 'large')
+  && !(el.classList.contains('standard')
+  || el.classList.contains('compact')) ? 'button-xl' : 'button-l');
```
JasonHowellSlavin, line 152: Ternary statements in function params can be confusing, and here there is at least 3 logical checks before the string assignment. Can we break it out to make it more readable, maintainable? Below is a suggestion, but there may be an even batter way to do it. 

```suggestion
  let buttonClass = 'button-l'
  if (size === 'large' && !el.classList.contains('standard') || el.classList.contains('compact)) {
    buttonClass = 'button-xl'
  }
  decorateButtons(text, buttonClass);
```

libs/blocks/marquee/marquee.css
drashti1712: Fixed this, thanks!

libs/blocks/marquee/marquee.css
drashti1712: Yes, removed!

drashti1712:
> @drashti1712 Thanks for looking at those scoped selectors. Good idea to have better performance first w/ regards to css selectors.
> 
> I'm still seeing linting errors on the marquee.css. Big picture is the format of that file should follow somewhat the dom structure of the block. Example. the first line in the `marquee.css` file should be simply `.marquee` etc and cascade down the selector specificity. This contribution goes against this paradigm and makes it hard to maintain and read. A block css file should never start w/ a very specific selector scope. ie `.marquee.large .foreground.container .text` See linting errors and order of selectors.
> 
> Screen Shot 2023-08-09 at 2 14 29 PM

@ryanmparrish Thanks for clarifying this. I misunderstood the previous comment. Fixed these issues now!

libs/blocks/marquee/marquee.js
drashti1712: > Ternary statements in function params can be confusing, and here there is at least 3 logical checks before the string assignment. Can we break it out to make it more readable, maintainable? Below is a suggestion, but there may be an even batter way to do it.

Changes as done as requested, thanks!

libs/blocks/marquee/marquee.js
```diff
@@ -147,7 +147,10 @@ export default function init(el) {
   if (firstDivInForeground.classList.contains('media')) el.classList.add('row-reversed');
 
   const size = getBlockSize(el);
-  decorateButtons(text, size === 'large' ? 'button-xl' : 'button-l');
+  const buttonClass = (size === 'large')
+  && !(el.classList.contains('standard')
+  || el.classList.contains('compact')) ? 'button-xl' : 'button-l';
```
chrischrischris, line 152: nit: formatting
```suggestion
  const buttonClass = (size === 'large')
    && !(el.classList.contains('standard') || el.classList.contains('compact'))
    ? 'button-xl'
    : 'button-l';
```

drashti1712: Fixed this. Thanks!

libs/blocks/marquee/marquee.css
drashti1712: > If this being used somewhere? icon-area does not have a body-xl attached to its classlist and I could not find .icon class on any element. Am I missing something or does it support library icons only and not any svg link?

This is implementing the use case when the body has 18px icons. We are currently waiting for the consonant team to confirm whether the icons will be added from the library or does it support svg links.
Thank you.

ryanmparrish (changes requested):
@drashti1712  - Can we take a stab at refactoring your css so that the selectors are ordered in a natural cascading order. Im seeing some dense selectors right at the top. Let's try to keep like selectors next to each other to maintain readability.

ryanmparrish (approved):

drashti1712:
Hello everyone,
This PR is blocked due to some confirmations pending from the consonant team regarding the designs. Hence, closing this PR for now. I will be raising another PR once the designs are revised.
Thank you for your valuable inputs.

PR #1101, opened 2023-08-10
MWPW-134968 - [Blog2Milo] Broken links on Homepage

jckautzmann:
- introduce a configuration called 'lang-root' in the page metadata
- append the lang-root config to the category links

Resolves: [MWPW-134968](https://jira.corp.adobe.com/browse/MWPW-134968)

**Test URLs:**
- After: https://appendlangroot--milo--adobecom.hlx.page/?martech=off

The Blog Homepage including the above changes:
https://main--blog--adobecom.hlx.page/?milolibs=appendlangroot


rgclayton (approved):

libs/blocks/utils/utils.js
```diff
@@ -26,4 +28,12 @@ const fetchWithTimeout = async (resource, options = {}) => {
   return response;
 };
 
-export { getLocalStorage, setLocalStorage, fetchWithTimeout };
+const updateLinkWithConfiguredLangRoot = (link) => {
```
chrischrischris, line 31: nit: name is a bit long, would `updateLinkWithLangRoot` work?

libs/blocks/utils/utils.js
```diff
@@ -26,4 +28,12 @@ const fetchWithTimeout = async (resource, options = {}) => {
...
+  const langRoot = getMetadata('lang-root');
+  if (!langRoot) return link;
+  const url = new URL(link);
```
chrischrischris, line 34: if `link` is an invalid URL it will throw a TypeError

libs/blocks/utils/utils.js
```diff
@@ -1,3 +1,5 @@
+import { getMetadata } from '../../utils/utils.js';
```
chrischrischris, line 1: I had not noticed this file before.  I don't think /blocks/utils is the proper place for this.  I know we unfortunately have /libs/utils/utils.js that actually isn't meant as a general util file.

I started a PR but haven't finished where we pull these types of util functions into /libs/utils/helpers.js.  What do you think about moving this file to that?

libs/blocks/utils/utils.js
```diff
@@ -26,4 +28,12 @@ const fetchWithTimeout = async (resource, options = {}) => {
   return response;
 };
 
-export { getLocalStorage, setLocalStorage, fetchWithTimeout };
+const updateLinkWithConfiguredLangRoot = (link) => {
+  const langRoot = getMetadata('lang-root');
```
chrischrischris, line 32: Curious: Do you foresee other milo consumers using `lang-root` ?

chrischrischris (approved):
If you don't want to refactor /blocks/utils/utils.js we can do that in a followup PR.

✅ with nits (and possibly add try/catch around URL)

libs/blocks/utils/utils.js
rgclayton: @chrischrischris Maybe? 
Probably not many unless their root content is different from the root (like the Blog). That's the main reason for this. It basically allows for a custom root.

rgclayton: Would `customRoot`/`cusom-root` make better sense?  `lang-root` is my fault :)

libs/blocks/utils/utils.js
rgclayton: `customLinkRoot`?

libs/blocks/utils/utils.js
jckautzmann: I'd keep it `lang-root` as it points to the language root of the site. For Blog, the language root is `/en`, not `/` which is the default language root for a Franklin/Milo based site.

libs/blocks/utils/utils.js
jckautzmann: I'll change it to `updateLinkWithLangRoot `

jckautzmann: addressed in https://github.com/adobecom/milo/pull/1101/commits/e2149877f76d990b95a22a56b442badd24f5c7b6

libs/blocks/utils/utils.js
jckautzmann: good catch, addressed in https://github.com/adobecom/milo/pull/1101/commits/e2149877f76d990b95a22a56b442badd24f5c7b6

libs/blocks/utils/utils.js
jckautzmann: I created `/libs/utils/helpers.js` and moved `updateLinkWithLangRoot` there (addressed in https://github.com/adobecom/milo/pull/1101/commits/e2149877f76d990b95a22a56b442badd24f5c7b6). I didn't move the other utility functions as they might
 be referenced in consumer sites. Moving them should be announced ahead so that consumers can adapt their code if needed.

libs/blocks/utils/utils.js
jckautzmann: addressed in https://github.com/adobecom/milo/pull/1101/commits/e2149877f76d990b95a22a56b442badd24f5c7b6

PR #1102, opened 2023-08-10
Browser data

zagi25:
On DC, we need the ability to detect browser and its version. For this we currently use `bowser.js` and we want to remove this dependency. So I was asked to add this feature to Milo.

Added `getBrowserData`:

- This will extract browser data (name, version and isMobile) from the navigator.userAgent string.

Resolves: [MWPW-127443](https://jira.corp.adobe.com/browse/MWPW-127443)

**Test URLs:**
- After: https://browser-data--milo--zagi25.hlx.page/?martech=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -949,3 +949,55 @@ export function loadLana(options = {}) {
   window.addEventListener('error', lanaError);
   window.addEventListener('unhandledrejection', lanaError);
 }
+
+export function getBrowserData(userAgent) {
+  if (!userAgent) {
+    return {};
+  }
+  const browser = {
+    ua: userAgent,
+    isMobile: userAgent.includes('Mobile'),
+  };
+
+  const regex = [
+    {
+      browserReg: /edg([ae]|ios)?/i,
+      versionReg: /edg([ae]|ios)?[\s/](\d+(\.?\d+)+)/i,
+      name: 'Microsoft Edge',
+    },
+    {
+      browserReg: /chrome|crios|crmo/i,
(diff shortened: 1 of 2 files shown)


mokimo:
https://developer.mozilla.org/en-US/docs/Web/API/Navigator/userAgent
> Note: The specification asks browsers to provide as little information via this field as possible. Never assume that the value of this property will stay the same in future versions of the same browser. Try not to use it at all, or only for current and past versions of a browser. New browsers may start using the same UA, or part of it, as an older browser: you really have no guarantee that the browser agent is indeed the one advertised by this property.
Also keep in mind that users of a browser can change the value of this field if they want (UA spoofing).

mokimo (changes requested):
There's several issues with this
1. The use of the userAgent is heavily discouraged
2. The function currently lives in the utils and which are loaded for _everyone_ - I'd either create a new file that contains this helper OR move it into the DC repo completely, unless really every milo consumer should always have access to a specific method.
3. Since this uses a discouraged pattern, using the userAgent, I don't think this belongs into milo core we shouldn't encourage the use of the browser checks.

An alternative to the userAgent, is doing feature checks ... MDN has a great article about the whole thing here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent

auniverseaway (changes requested):
I would rather not have this in the path of LCP and the additional bytes for every single request. Can you move this to your own project? We can elevate this to Milo when needed.

chrischrischris:
Since this is a feature that will only be used in DC, please implement it in your repo.  I would heed the advice from @mokimo too about not using userAgent if possible.  In general you should be doing feature detection instead of browser detection, but I don't know your use case.

PR #1104, opened 2023-08-10
MWPW-131211 Personalization Entitlement Support

chrischrischris:
* Entitlements can be defined in 2 ways:
  * using the `ent-` prefix and the entitlement name
  * using custom defined entitlements (`photoshop` and `lightroom` are provided as examples)

Resolves: [MWPW-131211](https://jira.corp.adobe.com/browse/MWPW-131211)

**Test URLs:**
Note that IMS does not work on milo non-main branches.  To see this PR in action see:
https://main--bacom--adobecom.hlx.page/drafts/cpeyer/pers/pers-ref?milolibs=persents
The top marquee will switch to an acrobat one if the signed in user is entitled to acrobat pro.

Milo Urls:
- Before: https://main--milo--adobecom.hlx.page/drafts/cpeyer/pers/pers-ref?martech=off
- After: https://persents--milo--adobecom.hlx.page/drafts/cpeyer/pers/pers-ref?martech=off


sukamat (approved):

libs/features/personalization/personalization.js
```diff
@@ -271,30 +269,87 @@ function parsePlaceholders(placeholders, config, selectedVariantName = '') {
...
+      const { default: getUserEntitlements } = await import('../../blocks/global-navigation/utilities/getUserEntitlements.js');
+      await loadIms();
+      entitlements = getUserEntitlements();
+    }
+    return entitlements;
+  });
+})();
+
+// remove underscores from entitlement names
+const rmUs = (obj) => Object.entries(obj).reduce((newObj, [key, val]) => {
```
JasonHowellSlavin, line 285: could you remove the comment if this was named more robustly? `removeEnttitlementUnderscore`?

libs/features/personalization/personalization.js
```diff
@@ -271,30 +269,87 @@ function parsePlaceholders(placeholders, config, selectedVariantName = '') {
...
+
+  const hasEntitlementPrefix = variantInfo.allNames.some((name) => name.startsWith('ent-'));
+  const hasEntitlementTag = entitlementTags.some((tag) => variantInfo.allNames.includes(tag));
+
+  let entitlements = [];
+  if (hasEntitlementPrefix || hasEntitlementTag) {
+    entitlements = await getFlatEntitlements();
+  }
+
+  const matchingVariant = variantNames.find((variant) => variantInfo[variant].some((name) => {
```
JasonHowellSlavin, line 344: This is a bit hard to parse. You could break out the function passed to `some` to its own named function to increase readability.

libs/utils/utils.js
```diff
@@ -685,6 +685,7 @@ export async function loadIms() {
       onError: reject,
     };
     loadScript('https://auth.services.adobe.com/imslib/imslib.min.js');
+    console.log('ims load start', performance.now() - window.utilsloaded);
```
JasonHowellSlavin, line 688: not sure if appropriate, but did we want to add lana logging here?

chrischrischris: doh - left in some perf timing I was testing. Thanks

libs/features/personalization/personalization.js
```diff
@@ -271,30 +272,91 @@ function parsePlaceholders(placeholders, config, selectedVariantName = '') {
...
+  let entitlements;
+  return (async () => {
+    if (!entitlements) {
+      const { default: getUserEntitlements } = await import('../../blocks/global-navigation/utilities/getUserEntitlements.js');
+      await loadIms();
+      entitlements = getUserEntitlements();
+    }
+    return entitlements;
+  });
+})();
```
mokimo, line 285: Strong pattern, but hard to grasp if someone isn't familiar with it.

I'd suggest something that feels more familiar:
```suggestion
async function fetchEntitlements() {
    const { default: getUserEntitlements } = await import('../../blocks/global-navigation/utilities/getUserEntitlements.js');
    await loadIms();
    return getUserEntitlements();
}
let entitlements;
export const loadEntitlements = () => {
    if (!entitlements) entitlements = fetchEntitlements();
    return entitlements;
};
```

mokimo (approved):

libs/features/personalization/personalization.js
```diff
@@ -271,30 +272,91 @@ function parsePlaceholders(placeholders, config, selectedVariantName = '') {
   return config;
 }
 
-function getPersonalizationVariant(manifestPath, variantNames = [], variantLabel = null) {
+export const loadEntitlements = (() => {
+  let entitlements;
+  return (async () => {
+    if (!entitlements) {
+      const { default: getUserEntitlements } = await import('../../blocks/global-navigation/utilities/getUserEntitlements.js');
+      await loadIms();
```
overmyheadandbody, line 280: Would it bring any value to have these coupled together in a `Promise.all`?

libs/martech/martech.js
```diff
@@ -92,7 +92,7 @@ const getDtmLib = (env) => ({
     env.name === 'prod'
       ? env.consumer?.marTechUrl || 'https://assets.adobedtm.com/d4d114c60e50/a0e989131fd5/launch-5dd5dd2177e6.min.js'
       // TODO: This is a custom launch script for milo-target - update before merging to main
-      : env.consumer?.marTechUrl || 'https://assets.adobedtm.com/d4d114c60e50/a0e989131fd5/launch-2c94beadc94f-development.min.js',
+      : env.consumer?.marTechUrl || 'https://assets.adobedtm.com/d4d114c60e50/a0e989131fd5/launch-a27b33fc2dc0-development.min.js',
```
overmyheadandbody, line 95: Comment says tho update this before merging to `main`. Is this the correct value to be used?

libs/utils/utils.js
```diff
@@ -965,3 +965,5 @@ export function loadLana(options = {}) {
   window.addEventListener('error', lanaError);
   window.addEventListener('unhandledrejection', lanaError);
 }
+
+window.utilsloaded = performance.now();
```
overmyheadandbody, line 969: Are we using this anywhere else? Name feels a bit generic to me, maybe we can scope it to something related to 'milo'

overmyheadandbody (approved):

libs/features/personalization/personalization.js
yesil, line 280: I think the ims logic should be handled in `getUserEntitlements` so that all consumers of this method don't duplicate the code.
Since `getUserEntitlements` is async, it should first wait for ims to be ready, then retrieve the entitlements.

yesil (commented):
A readme file with the logic implemented would be nice.
I see some logic like https://github.com/adobecom/milo/pull/1104/files#diff-50ce4be0b2dbdd422eabf9b3745f48792ee9142a1cd2c974104a69e5f13f2270R352 but it is hard to guess what it does exactly.

libs/features/personalization/personalization.js
chrischrischris: We use the pattern in a few places, and in general prefer for vars to be encapsulated.  I do like the idea of breaking out `fetchEntitlements` so will change that.

I think it's better that people learn this pattern - even if it's harder to grasp.

libs/features/personalization/personalization.js
chrischrischris: 👍 Good idea.  Really ims should already have been (or started) loading, this just makes sure before firing getUserEntitlements.

libs/utils/utils.js
chrischrischris: ugh thanks for seeing that.  Leftover debug code.

libs/features/personalization/personalization.js
chrischrischris: I don't think there's a clean way to add it there.  We don't have any of the ims config data, so we'd have to pass that in.

The majority of the time (all the time?) when this is called, ims will already be loaded due to gnav or other code needing IMS already.

3ch023 (approved):

libs/features/personalization/personalization.js
```diff
@@ -271,30 +272,98 @@ function parsePlaceholders(placeholders, config, selectedVariantName = '') {
...
+  const [{ default: getUserEntitlements }] = await Promise.all([
+    import('../../blocks/global-navigation/utilities/getUserEntitlements.js'),
+    loadIms(),
+  ]);
+  return getUserEntitlements();
+};
+
+export const getEntitlements = (() => {
+  let ents;
+  return (async () => {
```
yesil, line 285: if `ents` already defined, we can return it without returning an async function.

libs/features/personalization/personalization.js
```diff
@@ -271,30 +272,98 @@ function parsePlaceholders(placeholders, config, selectedVariantName = '') {
   return config;
 }
 
-function getPersonalizationVariant(manifestPath, variantNames = [], variantLabel = null) {
+const fetchEntitlements = async () => {
+  const [{ default: getUserEntitlements }] = await Promise.all([
+    import('../../blocks/global-navigation/utilities/getUserEntitlements.js'),
```
narcis-radu, line 277: There's one caveat with entitlements: the IMS client ID needs to be subscribed to JIL API. Maybe we should add some documentation to ensure all teams are aware of this step.

narcis-radu (approved):

libs/features/personalization/personalization.js
chrischrischris: The `return` async function here is assigned to `getEntitlements` (it's an IIFE)

libs/features/personalization/personalization.js
```diff
@@ -271,30 +272,98 @@ function parsePlaceholders(placeholders, config, selectedVariantName = '') {
...
+  newObj[key.replaceAll('_', '')] = val;
+  return newObj;
+}, {});
+
+const getFlatEntitlements = async () => {
+  if (window.adobeIMS && !window.adobeIMS.isSignedInUser()) return {};
+
+  const ents = await getEntitlements();
+  return {
+    ...removeUnderscores(ents.arrangement_codes),
```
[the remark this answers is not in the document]
chrischrischris: 👍  I forgot I was scrubbing underscores out of the key values initially.  Fixed.

yesil (approved):

vgoodric (approved):

PR #1105, opened 2023-08-10
MWPW-134888-marquee-video-autoplay

vgoodric:
stops replacing existing video elements and instead
* adds #autoplay to any video links without any anchor (thus making the default autoplay and not "controls"). This allows us to use the video auto block
* updates existing video elements with autoplay to address race conditions firing the video block before the marquee block

This should preserve existing behavior in marquees that have no anchor while allowing #autoplay1 videos to not loop.

Resolves: [MWPW-134888](https://jira.corp.adobe.com/browse/MWPW-134888)

**Test URLs:**
- Before: https://main--homepage--adobecom.hlx.page/drafts/vgoodrich/marquee-demo
- After: https://main--homepage--adobecom.hlx.page/drafts/vgoodrich/marquee-demo?milolibs=MWPW-134888-marquee-video-autoplay

URL for testing:

- https://mwpw-134888-marquee-video-autoplay--milo--adobecom.hlx.page/


libs/blocks/marquee/marquee.js
```diff
@@ -40,7 +40,7 @@ const decorateBlockBg = (block, node) => {
     if (childCount === 3) {
       child.classList.add(viewports[index]);
     }
-    const videoElement = child.querySelector('a[href$=".mp4"], video source[src$=".mp4"]');
```
auniverseaway, line 43: I think this is here due to a race condition.

If the video block loads before marquee, you will have a video element and not just a link.

ryanmparrish: This is correct.

ryanmparrish: Maybe address this in the dup func in this block decorateVid otherwise big picture we should refactor this to use a shared video decorator and import it rather. 
```
  const isAutoplay = !!(hash?.includes('autoplay'));
  const isNotLooped = !!(hash?.includes('autoplay1'));

  const attrs = isAutoplay ? `playsinline autoplay ${isNotLooped ? '' : 'loop'} muted` : 'playsinline controls';
  const video = `<video ${attrs}>
        <source src=".${pathname}" type="video/mp4" />
      </video>`;

**Comment by vgoodric** [comment]
Ok.  That makes sense.  One issue I thought of on the weekend is that the current system destroys information on looping/vs non looping.  But it also assumes auto play vs controls.  Which I think is smart for marquees.  And that means many marquees that are already authored won't have #autoplay on (because it was assumed). I'll tweak my code to keep looking for video links but "fix" existing videos to be autoplay vs controls if needed (while keeping the loop versus non-loop info).

**Review by seanchoi-dev — CHANGES_REQUESTED** [review]

File: libs/blocks/marquee/marquee.js [hunk 2]
```diff
@@ -133,16 +121,6 @@ export default function init(el) {
...
-    media.classList.add('media');
-
-    if (media.querySelector('a[href$=".mp4"]')) {
-      decorateVideo(media);
-    } else {
-      decorateImage(media);
-    }
-  }
-
   const firstDivInForeground = foreground.querySelector(':scope > div');
```
seanchoi-dev, line 124: If you are removing media class and above logic, you may want to check the css as well those using `media` class.

seanchoi-dev (changes requested):

vgoodric:
Updated the marquee block again in accordance with the discussion in Show and Tell.  #autoplay is added to video links without an anchor. This gives the marquee autoplay default behavior but then lets the video auto-block process the link.  This allows links with #autoplay1 or any future supported anchors to be honored. 
If a video element is found with a "controls" attribute, this means that the video block fired first and found a link with no anchor.  Video element attributes are fixed to account to enforce autoplay.

libs/blocks/marquee/marquee.js
vgoodric: I realized that the decorateImage function is still needed as well.  Added back in.

libs/blocks/marquee/marquee.js
```diff
@@ -99,25 +87,25 @@ function decorateMultipleIconArea(iconArea) {
...
 }
 
+function fixVideoAutoPlay(el) {
+  el.querySelectorAll('a[href$=".mp4"]').forEach((link) => {
+    link.href = `${link.href}#autoplay`;
+  });
+  el.querySelectorAll('video[controls]').forEach((video) => {
+    video.removeAttribute('controls');
+    video.setAttribute('autoplay');
+    video.setAttribute('muted');
```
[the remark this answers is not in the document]
vgoodric: Yeah, because I didn't want a value I wasn't sure if I could just do the name.  But adding an empty string got me the desired effect.

vgoodric:
resolved conflict with new code for hover behavior

auniverseaway (approved):

Sartxi (approved):

libs/blocks/marquee/marquee.js
```diff
@@ -132,18 +130,12 @@ export default function init(el) {
   const text = headline.closest('div');
   text.classList.add('text');
   const media = foreground.querySelector(':scope > div:not([class])');
-
   if (media) {
     media.classList.add('media');
+    decorateImage(media);
```
seanchoi-dev, line 135: Looks like you are calling `decorateImage()` even if it is video. I see it won't append any element if there is no `picture` element so, it should work find for now. Though, I feel this can go `else` of `if (video)` below so we don't call it at all when it is not necessary.

seanchoi-dev (approved):
@vgoodric  Approving this with a small concern. See the comment below.

chrischrischris:
Closing this as https://github.com/adobecom/milo/pull/1129 takes care of this.

PR #1107, opened 2023-08-14
Mwpw 134916 next button is not visible.

sharath-kannan:
* The next button is not visible after the recent changes in ost overlap with sidekick fix.
* Issue is replicable only in window systems.
* Margin top has been reduced to fix this.

Resolves: [MWPW-134916](https://jira.corp.adobe.com/browse/MWPW-134916)

**Test URLs:**\
- Before: https://main--milo--adobecom.hlx.page/tools/ost/?martech=off
- After: https://mwpw-134916--milo--sharath-kannan.hlx.page/tools/ost?martech=off


libs/blocks/ost/ost.css
```diff
@@ -31,3 +31,7 @@ body.tool-ost .ost>div {
 .price-unit-type::before {
   content: " ";
 }
+
+.spectrum-TabsPanel-tabpanel_3e140b > div > div>:last-child{
```
chrischrischris, line 35: This selector seems way too specific - are we guarenteed that `spectrum-TabsPanel-tabpanel_3e140b` will always exist?

Looks like you could instead target the button directly:
`spectrum-Button_e2d99e` or `data-testid="productsTab/nextButton"`

sharath-kannan: I have updated the code.

chrischrischris (approved):

narcis-radu (changes requested):
I understand this is a quick fix, but I don't agree with the workaround, it might cause some issues long term: the spectrum button class might change and using negative margins + important is not recommended.
A better approach would be to remove the grid gap from the second row, this would enable the "next" button to stay directly below the cards.

sharath-kannan:
> I understand this is a quick fix, but I don't agree with the workaround, it might cause some issues long term: the spectrum button class might change and using negative margins + important is not recommended. A better approach would be to remove the grid gap from the second row, this would enable the "next" button to stay directly below the cards.

I agree that using negative margin might not be the correct way to resolve this issue and i will make the required changes, but how do i override inline styles without using "Important"?

narcis-radu:
@sharath-kannan - I'm not seeing any inline styles on the button so I think `!important` can be avoided. Furthermore, if you remove the row gap you don't even need to set a negative margin. If you still think the margin is needed and you have to deal with inline styles, then using `!important` is the only way to override.

sharath-kannan:
This pull request has been closed because the requested changes has to be made in the offer selector tool react project.

PR #1109, opened 2023-08-15
MWPW-135105 Consumer Determined Post LCP

Blainegunn:
* Add your
* Specific
* Features or fixes

Resolves: [MWPW-135105](https://jira.corp.adobe.com/browse/MWPW-135105)

**Test URLs:**
- After: https://mwpw-135105--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -876,8 +876,6 @@ export async function loadArea(area = document) {
         loadPostLCP(config);
       });
     } else if (isDoc && section.el.dataset.idx === '0') { loadPostLCP(config); }
-
-
     // Show the section when all blocks inside are done.
     delete section.el.dataset.status;
     delete section.el.dataset.idx;
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -871,7 +871,12 @@ export async function loadArea(area = document) {
     window.dispatchEvent(new Event('milo:LCP:loaded'));
 
     // Post LCP operations.
-    if (isDoc && section.el.dataset.idx === '0') { loadPostLCP(config); }
+    if (window.PostLCPevnt && isDoc && section.el.dataset.idx === '0') {
+      window.addEventListener(config.PostLCPevnt, () => {
+        loadPostLCP(config);
+       });
+    } else if (isDoc && section.el.dataset.idx === '0') { loadPostLCP(config); }
+
 
     // Show the section when all blocks inside are done.
     delete section.el.dataset.status;
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -874,7 +874,7 @@ export async function loadArea(area = document) {
     if (window.PostLCPevnt && isDoc && section.el.dataset.idx === '0') {
       window.addEventListener(config.PostLCPevnt, () => {
         loadPostLCP(config);
-       });
+      });
     } else if (isDoc && section.el.dataset.idx === '0') { loadPostLCP(config); }
 
 


libs/utils/utils.js
```diff
@@ -871,8 +871,11 @@ export async function loadArea(area = document) {
     window.dispatchEvent(new Event('milo:LCP:loaded'));
 
     // Post LCP operations.
-    if (isDoc && section.el.dataset.idx === '0') { loadPostLCP(config); }
-
+    if (window.PostLCPevnt && isDoc && section.el.dataset.idx === '0') {
+      window.addEventListener(config.PostLCPevnt, () => {
```
peller, line 875: `window.PostLCPevnt`?

libs/utils/utils.js
```diff
@@ -871,8 +871,11 @@ export async function loadArea(area = document) {
...
+        loadPostLCP(config);
+      });
+    } else if (isDoc && section.el.dataset.idx === '0') { loadPostLCP(config); }
```
peller, line 878:
```suggestion
    if (isDoc && section.el.dataset.idx === '0') {
      if (window.postLCPevent) {
        window.addEventListener(window.postLCPevent, () => {
          loadPostLCP(config);
        });
      } else {
        loadPostLCP(config);
      }
    }
```

libs/utils/utils.js
```diff
@@ -871,8 +871,11 @@ export async function loadArea(area = document) {
     window.dispatchEvent(new Event('milo:LCP:loaded'));
 
     // Post LCP operations.
-    if (isDoc && section.el.dataset.idx === '0') { loadPostLCP(config); }
-
+    if (window.PostLCPevnt && isDoc && section.el.dataset.idx === '0') {
```
overmyheadandbody, line 874: How about something like `delayPostLCP`, `PostLCPevnt` sounds like an event triggered after LCP, so it might be a bit confusing.

Also, is there a connected consumer PR that actually declares this property?

libs/utils/utils.js
overmyheadandbody, line 875: Might be worth checking whether `config.PostLCPevnt` exists before adding the event listener, maybe not all projects declare it. Also, properties from `config` are usually lowercase, maybe something like `customPostLCPEvent` or similar might be better. There is already a `milo:LCP:loaded` event, that's why it might make sense to make the difference between the Milo one and a consumer one.

libs/utils/utils.js
peller: well, it's mostly whitespace changes. If you prefer, remove the newlines or even the curly braces around lines 877 and 880. Don't know what your style is, but it should minify to roughly the same thing.
Not repeating the if condition seems like better DRY.

libs/utils/utils.js
peller: I don't think we need a config property here. I think we can just use a single value `window.customPostLCPEvent` to check and listen for the event.

chrischrischris: Why not in config?  Everything else the consumer declares is in that object, so seems non-standard to set a global window property.  I'd rather we only use `config.postLCPevent`.

nit: no need to save one char with `evnt`, just use full `event`

chrischrischris (commented):
I wonder if instead we should have an event that triggers postLCP:
```
    // Post LCP operations.
    window.addEventListener('milo:loadPostLcp', () => loadPostLCP(config));
    if (!config.deferPostLcp && isDoc && section.el.dataset.idx === '0') {
		window.dispatchEvent(new Event('milo:loadPostLcp'));
    }
```

The consumer then can set `config.deferPostLcp = true` and then fire that event themselves.

Might need to tweak the naming of these things too.

peller:
@chrischrischris needed to use a global reference on window because, as I understand it, the consumer (dc-conversion-widget) would not have access to the config object without making additional network requests.

I very much like your event-based solution. It's loosely coupled and might not have all the same issues, but we'd still need a way for the consumer to set the delay flag to true without additional imports or network requests. Thoughts?

chrischrischris:
Closing for now as the correct approach is probably to wait to call `loadArea` until the consumer prereqs have finished.

PR #1111, opened 2023-08-15
[MWPW-134658] Localize meta links

JasonHowellSlavin:
* Chris' code to localize links that are authored in metadata
* Specifically solving for localized gnav

Resolves: [MWPW-134658](https://jira.corp.adobe.com/browse/MWPW-134658)

To verify: Open the pages and the network tab. Look for the call to gnav. Ensure the request url matches the region in which the page resides. 

**Test URLs:**
You should see no change in the network requests for the below pages, should stay the same
- Before: https://main--bacom--adobecom.hlx.page/ae_ar/customer-success-stories/adobe-experience-cloud?martech=off
- After: https://main--bacom--adobecom.hlx.page/ae_ar/customer-success-stories/adobe-experience-cloud?milolibs=localize-meta&martech=off

In the below pages, the before should not localize to a gnav in the network request, the after should have the local in the request:
- Before: https://main--bacom--adobecom.hlx.page/ch_fr/slavin-test?martech=off
- After: https://main--bacom--adobecom.hlx.page/ch_fr/slavin-test?martech=off&milolibs=localize-meta
- Before: https://main--bacom--adobecom.hlx.page/ae_ar/mukesh-test?martech=off
- After: https://main--bacom--adobecom.hlx.page/ae_ar/mukesh-test?martech=off&milolibs=localize-meta


mokimo (approved):

libs/utils/utils.js
```diff
@@ -803,7 +803,8 @@ function decorateMeta() {
     if (meta.getAttribute('property') === 'hlx:proxyUrl') return;
     try {
       const url = new URL(meta.content);
-      meta.setAttribute('content', `${origin}${url.pathname}${url.search}${url.hash}`);
+      const localizedPath = localizeLink(`${origin}${url.pathname}`);
```
overmyheadandbody, line 806: Is it certain that this always returns a path? `localizeLink` can return the whole URL in certain scenarios. If this happens, then the next line, `${origin}${localizedPath}...`, might be in an odd format.

overmyheadandbody (approved):

chrischrischris (approved):

meganthecoder:
I see the correct regional gnav.plain.html being loaded in your "before" links as well. Is that the wrong file to look at or is there something missing in the examples? Also, do we know of any metadata use cases that this affects other than gnav?

JasonHowellSlavin:
> I see the correct regional gnav.plain.html being loaded in your "before" links as well. Is that the wrong file to look at or is there something missing in the examples? Also, do we know of any metadata use cases that this affects other than gnav?

Great find. Looking in to this. Those pages don't seem to have gnav source authored in metadata, nor a gnav source authored in metadata.xlsx.

libs/utils/utils.js
JasonHowellSlavin: @overmyheadandbody I think because full paths are typically utilized in metadata.xlsx and the metadata block, I do not believe this will be an issue, but do you think it would be preferable to add a check like this:
```
const localizedURL = localizeLink(`${origin}${url.pathname}`);
const localizedPath = localizedURL.includes(origin) ? localizedURL : `${origin}${localizedURL}`;
```

overmyheadandbody: That's what I would personally go for, yes. But if you weren't able to find any such use-cases, I'm fine with having it as-is, I'm keeping my approval of the PR 😄 

If you do decide to implement the suggestion, though, I'd choose slightly different variable naming, as `localizedPath` would actually be `origin` + `pathname`, so not just a path. And of course, next line can just use this new constant rather than concatenating `${origin}${localizedPath}`

JasonHowellSlavin: Done!

sukamat (approved):

PR #1114, opened 2023-08-16
MWPW-135191 Hide caas link that has a query param

chrischrischris:
When copying a caas link from the configurator, it was also including any search params in the url (e.g: if the user had gone to another page, `?page=3` would be added)

This PR hides any caas links that have a search param, and fixes the configurator to not include those anymore.

Resolves: [MWPW-135191](https://jira.corp.adobe.com/browse/MWPW-135191)

**Test URLs:**
- After: https://hidecaaslink--milo--adobecom.hlx.page/?martech=off


mokimo (approved):

overmyheadandbody (approved):

honstar (approved):

libs/blocks/caas-config/caas-config.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react-hooks/exhaustive-deps */
```
honstar, line 1: The code compat check currently fails: https://github.com/adobecom/milo/actions/runs/5881564701/job/15970376456?pr=1114#step:3:30

Looks like you need to explicitly add `react-hooks` to the `plugins` array in `.eslintrc-code-compatibility.js` (found in https://stackoverflow.com/questions/59611822/definition-for-rule-react-hooks-exhaustive-deps-was-not-found).

chrischrischris: 👍 Thx.  I'm not actually sure we need to run this code compat step as it should be covered by the eslint pass, but will have to defer investigating for now.

PR #1122, opened 2023-08-17
MWPW-131211 Clear entitlements from localStorage on logout

chrischrischris:
Adds event for gnav logout: `feds:signOut`

Resolves: [MWPW-131211](https://jira.corp.adobe.com/browse/MWPW-131211)

**Test URLs:**
- After: https://clearls--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/libs/blocks/global-navigation/features/profile/dropdown.js
+++ b/libs/blocks/global-navigation/features/profile/dropdown.js
@@ -161,10 +161,9 @@ class ProfileDropdown {
       </li>
     `;
 
-    // TODO consumers might want to execute their own logic before a sign out
-    // we might want to provide them a way to do so here
     signOutLink.addEventListener('click', (e) => {
       e.preventDefault();
+      window.dispatchEvent(new Event('feds:signOut'));
       window.adobeIMS.signOut();
     });
 
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -313,6 +313,7 @@ const clearEntLocalStorage = () => {
 
 export const getEntitlements = (() => {
   let ents;
+  let logoutEventSet;
   return (async () => {
     if (window.adobeIMS && !window.adobeIMS.isSignedInUser()) {
       clearEntLocalStorage();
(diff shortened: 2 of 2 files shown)


libs/blocks/global-navigation/features/profile/dropdown.js
```diff
@@ -161,10 +161,9 @@ class ProfileDropdown {
       </li>
     `;
 
-    // TODO consumers might want to execute their own logic before a sign out
-    // we might want to provide them a way to do so here
     signOutLink.addEventListener('click', (e) => {
       e.preventDefault();
+      window.dispatchEvent(new Event('feds:signOut'));
```
overmyheadandbody, line 166: This might still cause some race conditions, so in the future we might want to look into allowing consumers to define a method in the global namespace or their config that would be called here.

libs/features/personalization/personalization.js
```diff
@@ -313,6 +313,7 @@ const clearEntLocalStorage = () => {
 
 export const getEntitlements = (() => {
   let ents;
+  let logoutEventSet;
```
overmyheadandbody, line 316: Should this be outside `getEntitlements` in case the method is called multiple times? Otherwise `logoutEventSet` is reset every time, right?

overmyheadandbody (approved):

mokimo (approved):

libs/features/personalization/personalization.js
chrischrischris: Since this is a IIFE `ents` and `logoutEventSet` are encapsulated by it.  The return value is the function assigned to `getEntitlements`.

This follows the "Revealing Module Pattern":
* https://gist.github.com/zcaceres/bb0eec99c02dda6aac0e041d0d4d7bf2

rgclayton (approved):

libs/features/personalization/personalization.js
overmyheadandbody: I missed a few extra parenthesis, my bad 🙈

PR #1124, opened 2023-08-17
fix video loading issue in figure block

Ruchika4:
* fix issue of video not loading in figure block

Resolves: [MWPW-129910](https://jira.corp.adobe.com/browse/MWPW-129910)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ruchika/figure?martech=off
- After: https://figureblock--milo--ruchika4.hlx.page/drafts/ruchika/figure?martech=off

### DIFF
--- a/libs/blocks/figure/figure.js
+++ b/libs/blocks/figure/figure.js
@@ -1,3 +1,5 @@
+import { applyHoverPlay } from '../../utils/decorate.js';
+
 function buildCaption(pEl) {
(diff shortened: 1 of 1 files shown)


chrischrischris (approved):
I know that this is a temp fix until the timing issues can be figured out.

chrischrischris:
Pagespeed score is down because the page is just a video and the footer which is late loaded.

PR #1126, opened 2023-08-17
Modify Hide Georouting Cookie

Ruchika4:
* Modify hideGeorouting cookie 
* More discussion regarding the request can be seen here https://adobe-mwp.slack.com/archives/C03HAU5SU2E/p1686757083623149

Resolves: [MWPW-128543](https://jira.corp.adobe.com/browse/MWPW-128543)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/au/drafts/ruchika/document16?georouting=off
- After: https://hide-georouting--milo--ruchika4.hlx.page/au/drafts/ruchika/document16?georouting=off

### DIFF
--- a/libs/features/georoutingv2/georoutingv2.js
+++ b/libs/features/georoutingv2/georoutingv2.js
@@ -102,9 +102,9 @@ function getGeoroutingOverride() {
   const urlParams = new URLSearchParams(window.location.search);
   const param = urlParams.get('georouting');
   const georouting = param || getCookie('georouting');
-  const domain = window.location.host === 'adobe.com'
-      || window.location.host.endsWith('.adobe.com') ? 'domain=adobe.com' : '';
   if (param === 'off') {
+    const domain = window.location.host === 'adobe.com'
+      || window.location.host.endsWith('.adobe.com') ? 'domain=adobe.com' : '';
     const d = new Date();
     d.setTime(d.getTime() + (365 * 24 * 60 * 60 * 1000));
     const expires = `expires=${d.toUTCString()}`;
--- a/libs/features/georoutingv2/georoutingv2.js
+++ b/libs/features/georoutingv2/georoutingv2.js
@@ -100,15 +100,17 @@ async function getAvailableLocales(locales) {
 
 function getGeoroutingOverride() {
   const urlParams = new URLSearchParams(window.location.search);
-  const param = urlParams.get('hideGeorouting');
-  const hideGeorouting = param || getCookie('hideGeorouting');
-  if (param === 'on') {
+  const param = urlParams.get('georouting');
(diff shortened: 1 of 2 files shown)


libs/features/georoutingv2/georoutingv2.js
```diff
@@ -100,15 +100,17 @@ async function getAvailableLocales(locales) {
 
 function getGeoroutingOverride() {
   const urlParams = new URLSearchParams(window.location.search);
-  const param = urlParams.get('hideGeorouting');
-  const hideGeorouting = param || getCookie('hideGeorouting');
-  if (param === 'on') {
+  const param = urlParams.get('georouting');
+  const georouting = param || getCookie('georouting');
+  const domain = window.location.host === 'adobe.com'
```
overmyheadandbody, line 105: Looks like you're using `domain` only if `param` is `off`, so this could be moved inside the L107 conditional

overmyheadandbody (approved):
NALA uses the old parameter quite a lot - https://github.com/search?q=repo%3Aadobecom%2Fnala%20hidegeorouting&type=code Did you reach out so they're aware and they can prepare a PR on their side? Otherwise adding the `run-nala` label to Milo PRs will likely fail until this change is addressed.

Approving these changes, but there should be a sync with the NALA team before merging this in.

libs/features/georoutingv2/georoutingv2.js
Ruchika4: Thanks for the review @overmyheadandbody . I have updated the code

libs/features/georoutingv2/georoutingv2.js
```diff
@@ -100,15 +100,17 @@ async function getAvailableLocales(locales) {
...
+    const domain = window.location.host === 'adobe.com'
+      || window.location.host.endsWith('.adobe.com') ? 'domain=adobe.com' : '';
     const d = new Date();
     d.setTime(d.getTime() + (365 * 24 * 60 * 60 * 1000));
     const expires = `expires=${d.toUTCString()}`;
-    document.cookie = `hideGeorouting=${hideGeorouting};${expires};path=/;`;
-  } else if (param === 'off') document.cookie = 'hideGeorouting=; expires= Thu, 01 Jan 1970 00:00:00 GMT';
-  return hideGeorouting === 'on';
+    document.cookie = `georouting=${georouting};${expires};path=/;${domain}`;
+  } else if (param === 'on') document.cookie = 'georouting=; expires= Thu, 01 Jan 1970 00:00:00 GMT';
```
hparra, line 112: nit: can this `else if` be a block since it is nested in if-block already? Or does eslint complain?

hparra (approved):

chadsunvice (approved):

PR #1127, opened 2023-08-18
Do not lana log IMS timeouts for the global nav

mokimo:
Do not lana log IMS timeouts as we can't fix those and they unnecessarily pollute our lana logs [MWPW-134861](https://jira.corp.adobe.com/browse/MWPW-134861)

Instead we want to ensure we try to show the sign in/profile button even if the initial `loadIms()` call was rejected if it eventually loads.

**Test URLs:**
- After: https://reduce-gnav-lana-logs--milo--mokimo.hlx.page/drafts/ramuntea/gnav-refactor?martech=off#

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -213,7 +213,11 @@ class Gnav {
   ims = async () => loadIms()
     .then(() => this.imsReady())
     .catch((e) => {
-      if (e?.message !== 'IMS timeout') lanaLog({ message: 'GNAV: Error with IMS', e });
+      if (e?.message === 'IMS timeout') {
+        window.addEventListener('onImsLibInstance', () => this.imsReady());
+        return;
+      }
+      lanaLog({ message: 'GNAV: Error with IMS', e });
     });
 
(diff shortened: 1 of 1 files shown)


overmyheadandbody (approved):

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -213,6 +213,10 @@ class Gnav {
   ims = async () => loadIms()
     .then(() => this.imsReady())
     .catch((e) => {
+      if (e?.message === 'IMS timeout') {
```
overmyheadandbody, line 216: Is it always exactly this message? No suffix/prefix/other capitalisation/etc.?

mokimo: Yup, we throw it ourselves https://github.com/adobecom/milo/blob/main/libs/utils/utils.js#L673 so if that changes we either consider it.. or see it popping up in the lana logs again

overmyheadandbody: Oh, right, I thought it's coming from IMS

mokimo:
blznm25i2o5

PR #1128, opened 2023-08-18
MWPW-132277 - Breakpoint content for Quiz Results

elaineskpt:
* Add breakpoints classes in order to allow Authoring add device specific content. Results might will show different parts of content for mobile, tablet or desktop.

Resolves: [MWPW-132277](https://jira.corp.adobe.com/browse/MWPW-132277)

**Test URLs:**
- Before: https://uar-integration--milo--adobecom.hlx.page/drafts/borges/end-to-end/results?primary=express-edu&quizKey=cc-quiz-en-US
- After: https://breakpoint-content--milo--elaineskpt.hlx.page/drafts/borges/end-to-end/results?primary=express-edu&quizKey=cc-quiz-en-US


libs/blocks/quiz-results/quiz-results.css
```diff
@@ -7,3 +7,59 @@
 .quiz-results.show {
   visibility: unset;
 }
+
+.section.phone-only,
```
colloyd, line 11: These generic ".section" selectors will likely not be acceptable to Milo core. They begrudgingly accepted my addition of the section class to quiz-results as a necessity, but they may draw the line at us adding css that can affect other sections in a page. Maybe we need to scope these rules to .quiz-results?

echampio-at-adobe: Rubber-stamping Cody's suggestion, I think scoping them to .quiz-results should do the trick. 

.quiz-results .section.phone-only …

libs/blocks/quiz-results/quiz-results.css
```diff
@@ -7,3 +7,59 @@
...
+.section.tablet-portrait-up,.section.tablet-landscape-up, 
+.section.desktop-up,.section.big-desktop-up {
+  display: none;
+}
+
+@media screen and (max-width: 599px) { 
+  .section.section.phone-only {
+    display: block;
```
narcis-radu, line 19: We have a mobile first approach, meaning that default (mobile) properties can be written without a media query.

libs/blocks/quiz-results/quiz-results.css
colloyd: Just to be clear here, you can just swap the .section selectors with .quiz-results, since .section is added manually in the JS to quiz-results anyway. Except in instances in which you are trying to target sections inside of fragments loaded by quiz-results - though that situation could potentially be very confusing for authors.

libs/blocks/quiz-results/quiz-results.css
```diff
@@ -7,3 +7,59 @@
...
+  }
+
+  .section.tablet-portrait-up,.section.tablet-landscape-up, 
+  .section.big-desktop-up {
+    display: none;
+  }
+}
+
+@media screen and (min-width: 1800px) {
```
fullcolorcoder, line 56: Where is 1800px coming from?

elaineskpt: was using the same ref we use in Dexter 
https://www.freecodecamp.org/news/the-100-correct-way-to-do-css-breakpoints-88d6a5ba1862#:~:text=to%20CSS%20or-,Less%20just%20the%20same.,-%40mixin%20for%2Dphone

libs/blocks/quiz-results/quiz-results.css
elaineskpt: in this case we need to display some titles only on mobile or switch content for mobile view.
this is why we are working with section metadata do add some phone-only classes.

Style side we are creating the quiz mobile first

cc: @echampio-at-adobe  @fullcolorcoder @colloyd

fullcolorcoder: @elaineskpt I think this is mostly about readability, and keeping it clean for future maintenance. It's always cleaner to see CSS inside of a media query and know that this is always non-mobile style. Max-width could save a few lines for now maybe, but we could say .phone-only is set to display: block, then set to none in the min-width:600 section to keep in alignment with the mobile first coding style. 

Bootstrap does a similar approach https://getbootstrap.com/docs/4.0/utilities/display/#hiding-elements

libs/blocks/quiz-results/quiz-results.css
echampio-at-adobe: Hi @fullcolorcoder, I think you're correct about it being beneficial to have this sort of thing available in Milo in a generic fashion for handling breakpoint specific content. But, when I've discussed the idea earlier, the reply was that Milo wasn't going there yet. For now, at least, I think we should finish this PR so we're done, and afterwards, raise another milo-community thread to ask about moving it so it's generic.

echampio-at-adobe (approved):
To test, I started here and went through the photo, template, edu questions:
https://breakpoint-content--milo--elaineskpt.hlx.page/drafts/borges/end-to-end/

Here's the info for local storage:
localStorage.setItem('cc-quiz-en-US', '{"primaryProducts":["express-edu"],"secondaryProducts":["cc-all-edu","lr-edu"],"umbrellaProduct":"","basicFragments":["https://uar-integration--milo--adobecom.hlx.page/drafts/borges/end-to-end/fragments/basic/uar-sample-whole-marquee-ps","https://uar-integration--milo--adobecom.hlx.page/drafts/borges/end-to-end/fragments/basic/uar-sample-card-list"],"nestedFragments":{"marquee-product-single":["https://uar-integration--milo--adobecom.hlx.page/drafts/borges/end-to-end/fragments/nested/marquee-product-single-ps"],"marquee-product-plan":["https://uar-integration--milo--adobecom.hlx.page/drafts/borges/end-to-end/fragments/nested/marquee-product-plan-lrpr"],"commerce-card":["https://uar-integration--milo--adobecom.hlx.page/drafts/borges/end-to-end/fragments/nested/uar-sample-card-ps"]},"pageloadHash":"type=cc:app-reco&quiz=uarv3&selectedOptions=q-category/photo|q-rather/template|q-customer/educational"}');

In the resulting page, the headings marked for different breakpoints seem to switch in and our as they should:
https://breakpoint-content--milo--elaineskpt.hlx.page/drafts/borges/end-to-end/results?debug=uar?primary=express-edu&quizKey=cc-quiz-en-US

libs/blocks/quiz-results/quiz-results.css
elaineskpt: Just added an update for it @fullcolorcoder @narcis-radu

echampio-at-adobe (approved):
After addressing comments from Narcis, the test page still seems to function well for me.

fullcolorcoder (approved):
Awesome!

narcis-radu (approved):

chrischrischris:
@elaineskpt Can you fix this linting error:
```
223:3   error    React Hook "useEffect" is called conditionally. React Hooks must be called in the exact same order in every component render. Did you accidentally call a React Hook after an early return?                                     react-hooks/rules-of-hooks
```

and the unit test failures?
Once that's done we can merge.

chrischrischris:
@elaineskpt I didn't see that there are no js changes, so failures are not due to this PR.  Let's create a followup PR to address those failures in the `uar-integration` branch.

JackySun9 (approved):

echampio-at-adobe:
> @elaineskpt I didn't see that there are no js changes, so failures are not due to this PR. Let's create a followup PR to address those failures in the `uar-integration` branch.

Hi @chrischrischris. We have these stories logged to specifically deal with unit tests: MWPW-133119 and MWPW-134984

PR #1129, opened 2023-08-18
MWPW-135276 Fix video selector

chrischrischris:
Selector was looking for video links that end in .mp4, but if any hash was appended to video link then it would not match.

Changed selector to `*=` which will match any part of the string

This also adds support for `autoplay1` for marquee background videos (used to always loop).

Resolves: [MWPW-135276](https://jira.corp.adobe.com/browse/MWPW-135276)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rclayton/poc/video-samples?martech=off
- After: https://videofix--milo--adobecom.hlx.page/drafts/rclayton/poc/video-samples?martech=off

### DIFF
--- a/libs/blocks/marquee/marquee.js
+++ b/libs/blocks/marquee/marquee.js
@@ -17,7 +17,7 @@ const decorateVideo = (container, src) => {
     const { pathname, hash } = src;
     const attrs = getVideoAttrs(hash);
     container.innerHTML = `<video ${attrs}>
-          <source src=".${pathname}" type="video/mp4" />
+          <source src="${pathname}" type="video/mp4" />
         </video>`;
   }
   return container.firstChild;
--- a/libs/blocks/marquee/marquee.js
+++ b/libs/blocks/marquee/marquee.js
@@ -14,10 +14,10 @@ const decorateVideo = (container, src) => {
     </video>`;
     container.classList.add('has-video');
   } else {
-    const { pathname, hash } = src;
+    const { href, hash } = src;
     const attrs = getVideoAttrs(hash);
     container.innerHTML = `<video ${attrs}>
-          <source src="${pathname}" type="video/mp4" />
+          <source src="${href}" type="video/mp4" />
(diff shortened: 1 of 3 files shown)


Sartxi (approved):
Nice catch!

Ruchika4 (approved):

libs/blocks/marquee/marquee.js
```diff
@@ -137,8 +137,8 @@ export default function init(el) {
     media.classList.add('media');
     const video = media.querySelector('video');
     if (video) applyHoverPlay(video);
-    if (media.querySelector('a[href$=".mp4"]')) {
-      decorateVideo(media);
+    if (media.querySelector('a[href*=".mp4"]')) {
+      decorateVideo(media, media.querySelector('a').href);
```
Ruchika4, line 141: hoverplay and autoplay1 is not working intermittently in marquee block. This appears to be because of the race situation for video autoblock in a block we discussed yesterday. With this change if video.js is not called first video block is not created. Hence hoverplay event is not added at line 139. However at line 141 we are creating video tag in marquee.js itself for mp4 videos with hash value and video.js is not called and events not added. In marquee.js (decorateVideo function) we are hardcoding video block with autoplay on loop. Hence any hash value other than autoplay is not working as expected

Test page to see the behavior: https://videofix--milo--adobecom.hlx.page/drafts/ruchika/marquee1

Ruchika4 (changes requested):

libs/blocks/marquee/marquee.js
chrischrischris: fixed

rgclayton (approved):

meganthecoder (approved):
Looks good for a quick bugfix: https://main--bacom--adobecom.hlx.live/uk/resources/digital-trends-report?milolibs=videofix 

As discussed via slack, we'll refactor this in the future for better defaults

PR #1131, opened 2023-08-21
Mwpw 135206 geo routing popup is displayed on top of the commerce iframe modal.

sharath-kannan:
* Geo routing popup overlaps commerce frame modal.
* To Match the functionality in adobe.com, The geo routing popup has been sent behind the commerce-frame modal.

Resolves: [MWPW-135206](https://jira.corp.adobe.com/browse/MWPW-135206)

**Test URLs:**
- Before: https://stage--cc--adobecom.hlx.page/drafts/ruchika/discover1#animate
- After: https://stage--cc--adobecom.hlx.page/drafts/ruchika/discover1?milolibs=MWPW-135206--milo--sharath-kannan#animate

### DIFF
--- a/libs/blocks/modal/modal.css
+++ b/libs/blocks/modal/modal.css
@@ -22,7 +22,7 @@
 }
 
 .dialog-modal.commerce-frame {
-  z-index: 110;
+  z-index: 103;
 }
 
 .modal-curtain.is-open {
--- a/libs/blocks/modal/modal.css
+++ b/libs/blocks/modal/modal.css
@@ -21,6 +21,10 @@
   z-index: 102;
 }
 
+.dialog-modal.commerce-frame {
+  z-index: 110;
+}
+
 .modal-curtain.is-open {
   position: fixed;
   top: 0;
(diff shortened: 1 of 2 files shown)


narcis-radu (changes requested):
I don't agree with this approach, we are rendering all modals based on the existence of commerce modals (which are not available on every single page).
There are two things here:
1. does the commerce modal really need to sit on top of geo modal? I would say no, I think the geo modal makes more sense to be on top since it provides a localised experience for the user. If the user has already selected a value, the modal will no longer be visible.
2. if the commerce modal is really needed on top of geo modal, let's make use of `z-index`.

libs/blocks/modal/modal.css
```diff
@@ -21,6 +21,10 @@
   z-index: 102;
 }
 
+.dialog-modal.commerce-frame {
+  z-index: 110;
```
narcis-radu, line 25: I think we can safely use `103` in this case.

sharath-kannan: done.

narcis-radu (approved):

PR #1145, opened 2023-08-22
MWPW-130835 search osi in OST

VKniaz:
* Add your
* Specific
* Features or fixes

Resolves: [MWPW-130835](https://jira.corp.adobe.com/browse/MWPW-130835 )

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/tools/ost?milolibs=local#?martech=off
- After: https://MWPW-130835--milo--adobecom.hlx.page/tools/ost?martech=off

### DIFF
--- a/libs/blocks/ost/ost.js
+++ b/libs/blocks/ost/ost.js
@@ -4,7 +4,7 @@ import { ENV_PROD, getTacocatEnv } from '../merch/merch.js';
 
 const IMS_COMMERCE_CLIENT_ID = 'aos_milo_commerce';
 const IMS_PROD_URL = 'https://auth.services.adobe.com/imslib/imslib.min.js';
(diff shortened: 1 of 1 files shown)


mokimo (approved):

overmyheadandbody (approved):
I see there's a unit test error in `test/blocks/caas-config/caas-config.test.html`. It doesn't sound like these are related, but it might be better to check before merging.

honstar:
> I see there's a unit test error in `test/blocks/caas-config/caas-config.test.html`. It doesn't sound like these are related, but it might be better to check before merging.

Updating the branch to the latest state fixed the unit test failure, might have been a temporary issue. All green now.

@VKniaz : please remember adding a description to the PR, regardless how trivial the change might be :)

PR #1149, opened 2023-08-23
[UAR] Quiz URL States

fullcolorcoder:
Resolves: 
[MWPW-126142](https://jira.corp.adobe.com/browse/MWPW-126142)
[MWPW-132262](https://jira.corp.adobe.com/browse/MWPW-132262)
[MWPW-125568](https://jira.corp.adobe.com/browse/MWPW-125568)

- Allows for state to be populated via url params. 
- Allows for back button to restore previous state
- Manages delivery of url to results page

**Test URLs:**
- Before: https://uar-integration--milo--adobecom.hlx.page/drafts/bradjohn/quiz-22/?martech=off
- After: https://quiz-url-states--milo--fullcolorcoder.hlx.page/drafts/bradjohn/quiz-22/?martech=off

### DIFF
--- a/libs/blocks/quiz/quiz.js
+++ b/libs/blocks/quiz/quiz.js
@@ -55,7 +55,9 @@ const App = () => {
       const hasInValidParam = Object.keys(initialUrlParams)
         .some((key) => !VALID_QUESTIONS.includes(key) && !KNOWN_PARAMS.includes(key));
 
-      const lastParamKey = Object.keys(initialUrlParams)[Object.keys(initialUrlParams).length - 1];
+      const filteredParams = Object.keys(initialUrlParams)
+        .filter((key) => !KNOWN_PARAMS.includes(key));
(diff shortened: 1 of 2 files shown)


echampio-at-adobe (approved):
In its current state, I think this PR makes things much better. From a behavioral aspect, leaving url params like martech=off that aren't related to the quiz will cause problems when trying to rebuild the page, but we should be able to deal with that problem in MWPW-132262.  The remaining lint warnings about react dependencies in useEffect methods will need some discussion between us as a group, and probably brand new stories, to fix the right way.

JackySun9:
@fullcolorcoder  Please check https://jira.corp.adobe.com/browse/MWPW-126142

echampio-at-adobe:
From what I can see, I think this PR takes care of all of the stories related to preserving the quiz state in the URL and preserving other URL params that aren't related to the quiz.

PR #1150, opened 2023-08-23
MWPW-135352 Adding checkboxes for Marketo

Brandon32:
* Adding checkbox styles for Marketo forms.
  * Note: Use a country like Korea, South to see privacy checkboxes.
* Update mobile focus fix to only activate on select field types.
  * Necessary for testing checkboxes on mobile.

Resolves: [MWPW-135352](https://jira.corp.adobe.com/browse/MWPW-135352)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/bmarshal/mcz/marketo?martech=off
- After: https://bmarshal-marketo-checkbox--milo--adobecom.hlx.page/drafts/bmarshal/mcz/marketo?martech=off

**BACOM URLs:**
- Before: https://main--bacom--adobecom.hlx.live/uk/resources/reports/state-of-marketing-automation-b?martech=off
- After: https://main--bacom--adobecom.hlx.live/uk/resources/reports/state-of-marketing-automation-b?martech=off&milolibs=bmarshal-marketo-checkbox

### DIFF
--- a/libs/blocks/marketo/marketo.css
+++ b/libs/blocks/marketo/marketo.css
@@ -168,8 +168,6 @@
   border: 0;
   border-radius: 2px;
   padding: 0;
-  width: 14px!important;
-  height: 14px;
   position: relative;
   top: 2px;
   margin-right: 10px;
@@ -191,9 +189,6 @@
   height: 10px;
   border: 2px solid var(--marketo-form-border);
   border-radius: 2px;
-  position: absolute;
-  top: 0;
-  left: 0;
 }
 
 .marketo .mktoForm .mktoCheckboxList.mktoInvalid input::before {
--- a/libs/blocks/marketo/marketo.js
+++ b/libs/blocks/marketo/marketo.js
@@ -191,8 +191,8 @@ export default function init(el) {
 
   createIntersectionObserver({
     el,
-    callback: (el) => {
-      loadMarketo(el, formData);
+    callback: (target) => {
+      loadMarketo(target, formData);
     },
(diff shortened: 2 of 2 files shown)


libs/blocks/marketo/marketo.css
```diff
@@ -162,11 +163,63 @@
   font-weight: 400;
 }
 
+.marketo .mktoForm .mktoCheckboxList input {
+  appearance: none;
+  border: 0;
+  border-radius: 2px;
+  padding: 0;
+  width: 14px!important;
```
overmyheadandbody, line 171: I wasn't loving this `!important` so I had a quick look - you can skip declaring the `width` and `height` here altogether. Then, remove absolute positioning (and `top`/`left` declarations as well, of course) from `.marketo .mktoForm .mktoCheckboxList input::before`, which will then fit perfectly into its parent. It already has a width and height of `10px` + `2px` border on each side, which totals `14px`, exactly as this element. This leads to the similar UI, but avoids the `!important` declaration

JasonHowellSlavin (approved):
Agree with Rares, but otherwise LGTM

libs/blocks/marketo/marketo.css
Brandon32: Thanks!

puruadobe (approved):

mokimo (approved):

overmyheadandbody (approved):
Thanks for implementing the suggestion!

PR #1154, opened 2023-08-24
MWPW-135537 Add ability to hide a block

chrischrischris:
Authors can add `hide block` to the block classlist for it to be ignored and hidden (js & css for the block will not be loaded and the block will not be decorated).

Resolves: [MWPW-135537](https://jira.corp.adobe.com/browse/MWPW-135537)

**Test URLs:**
There are 2 quote blocks on the ben & jerry page that have "hide block" set, so all quotes will be hidden in the after branch.

- Before: https://main--milo--adobecom.hlx.page/drafts/cpeyer/ben-and-jerrys-case-study?martech=off
- After: https://hideblock--milo--adobecom.hlx.page/drafts/cpeyer/ben-and-jerrys-case-study?martech=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -389,7 +389,11 @@ function checkForExpBlock(name, expBlocks) {
 }
 
 export async function loadBlock(block) {
-  if (block.classList.contains('hide-block')) return block;
+  if (block.classList.contains('hide-block')) {
+    block.remove();
+    return null;
+  }
+
   let name = block.classList[0];
   const { miloLibs, codeRoot, expBlocks } = getConfig();
 
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -389,6 +389,7 @@ function checkForExpBlock(name, expBlocks) {
 }
 
 export async function loadBlock(block) {
+  if (block.classList.contains('hide-block')) return block;
   let name = block.classList[0];
   const { miloLibs, codeRoot, expBlocks } = getConfig();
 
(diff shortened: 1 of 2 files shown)


sukamat (approved):

libs/utils/utils.js
```diff
@@ -389,6 +389,7 @@ function checkForExpBlock(name, expBlocks) {
 }
 
 export async function loadBlock(block) {
+  if (block.classList.contains('hide-block')) return block;
```
overmyheadandbody, line 392: Does it even make sense to keep it? Could it not be removed entirely? DOM would be a bit cleaner, there wouldn't be a need for the CSS class, it would avoid issues if someone decides that `.quote.center` needs a `display:flex !important` for whatever reason. Just my 2 cents

overmyheadandbody (approved):

chrischrischris:
Updated to remove the block from the DOM

Brandon32 (approved):

sukamat (commented):
@chrischrischris Looks like some unit test is failing.

mokimo (approved):
Quick and simple, nice 👍 
A unit test would make a lot of sense for this

narcis-radu (approved):

honstar (approved):

honstar:
Looks like `test/utils/richresults.test.js` doesn't like this change, all those unit-tests are failing currently.

sukamat (approved):

PR #1156, opened 2023-08-24
Only run when MEP or MEP param is enabled

chrischrischris:
`addPreviewToConfig` should only run when either:
1. the `mep` param is added to url
2. We are not on prod and target or MEP is activated for the page

Resolves: No Jira Ticket

**Test URLs:**
- After: https://meppreview--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -739,16 +739,14 @@ async function checkForPageMods() {
       .filter((path) => path?.trim());
   }
 
-  const { mep: mepOverride } = Object.fromEntries(PAGE_URL.searchParams);
   const { env } = getConfig();
-  const previewPage = env?.name === 'stage' || env?.name === 'local';
-  if (mepOverride || mepOverride === '' || previewPage) {
+  const { mep } = Object.fromEntries(PAGE_URL.searchParams);
+  if (mep !== undefined || (env.name !== 'prod' && (persEnabled || targetEnabled))) {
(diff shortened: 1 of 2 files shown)


auniverseaway (approved):

rgclayton (approved):

libs/utils/utils.js
```diff
@@ -739,16 +739,14 @@ async function checkForPageMods() {
       .filter((path) => path?.trim());
   }
 
-  const { mep: mepOverride } = Object.fromEntries(PAGE_URL.searchParams);
   const { env } = getConfig();
-  const previewPage = env?.name === 'stage' || env?.name === 'local';
-  if (mepOverride || mepOverride === '' || previewPage) {
+  const mep = PAGE_URL.searchParams.get('mep');
+  if (mep !== null || (env.name !== 'prod' && (persEnabled || targetEnabled))) {
```
mokimo, line 744: Seems like we got some unit tests failing @chrischrischris `env?..name` or need to properly set the configs for the failing tests

> Rich Results > add the NewsArticle rich results
      TypeError: Cannot read properties of undefined (reading 'name')
        at checkForPageMods (libs/utils/utils.js:744:28)
        at loadArea (libs/utils/utils.js:844:11)
        at o.<anonymous> (test/utils/richresults.test.js:9:11)

PR #1157, opened 2023-08-25
Add support for Google Yolo

narcis-radu:
### Description
Implement Google Yolo (you only login once) in Milo. The feature will be enabled only if `google-login` metadata is set to `on`. Google scripts are loaded from `delayed.js` to ensure there's no impact on performance.
An additional parameter can be used to specify a redirect page after a successful login.

Screenshot 2023-08-25 at 10 46 23

### Engineering
For security reasons the Google web application is allowed to render the dropdown and exchange tokens in IMS only on certain origins, wildcards are not supported. If you want to test the Google Yolo implementation in your project and your origin is not allowed, please reach out to @narcis-radu, @overmyheadandbody, @mokimo. 
The Google web application will throw an error if the origin is not allowed.
```html
[GSI_LOGGER]: The given origin is not allowed for the given client ID.
```
### Authoring
The following parameters are available:
* `google-login` - set it to `on` to enable the feature.
* `google-login-redirect` - specify an absolute URL where the user should be redirected after a successful login.

Resolves: [MWPW-135093](https://jira.corp.adobe.com/browse/MWPW-135093)

**Test URLs:**
- After: https://yolo--milo--narcis-radu.hlx.page/drafts/narcis/default?martech=off

**Note:** the _after_ test URL is not allowed by IMS to authenticate, a full workflow test might not be possible until the code is merged.

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.css
+++ b/libs/blocks/global-navigation/global-navigation.css
@@ -442,8 +442,6 @@ header.global-navigation {
   right: auto;
 }
 
-
-
 /* Desktop styles */
 @media (min-width: 900px) {
   /* General */
--- a/libs/blocks/global-navigation/global-navigation.css
+++ b/libs/blocks/global-navigation/global-navigation.css
@@ -431,6 +431,19 @@ header.global-navigation {
   background-color: var(--feds-background-link--hover--light);;
 }
 
+#feds-googleLogin {
+  position: absolute;
+  top: 100%;
+  right: 0;
+}
+
+[dir = "rtl"] #feds-googleLogin {
+  left: 0;
+  right: auto;
+}
+
+
+
 /* Desktop styles */
 @media (min-width: 900px) {
   /* General */
--- a/libs/features/google-login.js
+++ b/libs/features/google-login.js
@@ -1,0 +1,53 @@
+const GOOGLE_SCRIPT = 'https://accounts.google.com/gsi/client';
+const GOOGLE_ID = '530526366930-l874a90ipfkn26naa71r010u8epp39jt.apps.googleusercontent.com';
+const PLACEHOLDER = 'feds-googleLogin';
+const WRAPPER = 'feds-profile';
+
+const onToken = async (getMetadata, data) => {
+  let destination;
+  try {
+    destination = new URL(getMetadata('google-login-redirect'))?.href;
+  } catch {
+    // Do nothing
+  }
+
+  await window.adobeIMS.socialHeadlessSignIn({
+    provider_id: 'google',
+    idp_token: data?.credential,
+    client_id: window.adobeid?.client_id,
+    scope: window.adobeid?.scope,
+  }).then(() => {
+    if (window.DISABLE_PAGE_RELOAD === true) return;
+    // Existing account
+    if (destination) {
+      window.location.assign(destination);
+    } else {
+      window.location.reload();
+    }
+  }).catch(() => {
+    // New account
+    window.adobeIMS.signInWithSocialProvider('google', { redirect_uri: destination || window.location.href });
+  });
+};
+
+export default async function initGoogleLogin(loadIms, getMetadata, loadScript) {
+  try {
+    await loadIms();
+  } catch {
+    return;
+  }
+  if (window.adobeIMS?.isSignedInUser()) return;
+
+  await loadScript(GOOGLE_SCRIPT);
+  const placeholder = document.createElement('div');
+  placeholder.id = PLACEHOLDER;
+  document.querySelector(`.${WRAPPER}`)?.append(placeholder);
+
+  window.google?.accounts?.id?.initialize({
+    client_id: GOOGLE_ID,
+    callback: (data) => onToken(getMetadata, data),
+    prompt_parent_id: PLACEHOLDER,
+    cancel_on_tap_outside: false,
+  });
+  window.google?.accounts?.id?.prompt();
+}
--- a/libs/scripts/delayed.js
+++ b/libs/scripts/delayed.js
@@ -48,6 +48,16 @@ export const loadPrivacy = async (getConfig, loadScript) => {
     }
   });
 };
+
+export const loadGoogleLogin = async (getMetadata, loadIms, loadScript) => {
+  const googleLogin = getMetadata('google-login')?.toLowerCase();
+  if (googleLogin !== 'on') return;
+  if (window.adobeIMS?.isSignedInUser()) return;
+
+  const { default: initGoogleLogin } = await import('../features/google-login.js');
+  initGoogleLogin(loadIms, getMetadata, loadScript);
+};
+
 /**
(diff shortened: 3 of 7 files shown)


libs/features/google-login.js
```diff
@@ -0,0 +1,53 @@
...
+    window.adobeIMS.signInWithSocialProvider('google', { redirect_uri: destination || window.location.href });
+  });
+};
+
+export default async function initGoogleLogin(loadIms, getMetadata, loadScript) {
+  try {
+    await loadIms();
+  } catch {
+    return;
+  }
```
mokimo, line 38: Since you have the isSignedInUser check anyway, you could alternatively do this:
```suggestion
    await loadIms().catch(() => {})
```

mokimo (approved):

libs/features/google-login.js
mokimo: Oh you return if the user is already logged in, not NOT logged in. My bad! No in this case I think it's fine as is

narcis-radu: Also, I added one more check in `delayed` - https://github.com/adobecom/milo/pull/1157/files#diff-c21f6b181685a27721a54ce2b519823f6186157499b08ea290437d3fdff64473R55 - I wanted to make sure we do a fast fail and avoid loading the `google-login` module altogether if the user is logged-in already.

test/features/google-login/google-login.test.js
```diff
@@ -0,0 +1,84 @@
...
+    window.adobeid = {
+      client_id: 'milo',
+      scope: 'gnav',
+    };
+  });
+
+  afterEach(() => {
+    document.body.innerHTML = '';
+    initializeSpy.restore();
+    promptSpy.restore();
```
overmyheadandbody, line 30: Should we also delete `window.adobeid` to not pollute other tests?

test/features/google-login/google-login.test.js
```diff
@@ -0,0 +1,84 @@
...
+    await initGoogleLogin(sinon.stub(), sinon.stub(), sinon.stub());
+    onToken = initializeSpy.getCall(0).args[0].callback;
+    window.DISABLE_PAGE_RELOAD = true;
+    signInWithSocialProviderSpy.resetHistory();
+    await onToken(sinon.stub(), sinon.stub());
+    expect(signInWithSocialProviderSpy.called).not.to.be.true;
+    expect(document.getElementById('feds-googleLogin')).to.exist;
+    signInWithSocialProviderSpy.restore();
+    socialHeadlessSignInStub.restore();
+    window.DISABLE_PAGE_RELOAD = false;
```
overmyheadandbody, line 72: Could we not just `delete window.DISABLE_PAGE_RELOAD`, since it wasn't there to begin with?

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -431,6 +431,19 @@ header.global-navigation {
...
+  right: 0;
+}
+
+[dir = "rtl"] #feds-googleLogin {
+  left: 0;
+  right: auto;
+}
+
+
+
```
overmyheadandbody, line 446: Extra lines could probably be deleted

overmyheadandbody (approved):
I see some unit tests failing in `utils.js`, are they somehow related to this PR or are they coming from somewhere else?

libs/features/google-login.js
```diff
@@ -0,0 +1,53 @@
+const GOOGLE_SCRIPT = 'https://accounts.google.com/gsi/client';
+const GOOGLE_ID = '530526366930-l874a90ipfkn26naa71r010u8epp39jt.apps.googleusercontent.com';
```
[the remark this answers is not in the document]
narcis-radu: Yes, this is the Google web application ID, it's not a secret. Also, this web application is federated and will be used for all Milo consumers, we don't need to add it as part of a config.

libs/features/google-login.js
```diff
@@ -0,0 +1,53 @@
...
+    await loadIms();
+  } catch {
+    return;
+  }
+  if (window.adobeIMS?.isSignedInUser()) return;
+
+  await loadScript(GOOGLE_SCRIPT);
+  const placeholder = document.createElement('div');
+  placeholder.id = PLACEHOLDER;
+  document.querySelector(`.${WRAPPER}`)?.append(placeholder);
```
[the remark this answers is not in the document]
narcis-radu: That's correct.

libs/features/google-login.js
```diff
@@ -0,0 +1,53 @@
...
+
+  window.google?.accounts?.id?.initialize({
+    client_id: GOOGLE_ID,
+    callback: (data) => onToken(getMetadata, data),
+    prompt_parent_id: PLACEHOLDER,
+    cancel_on_tap_outside: false,
```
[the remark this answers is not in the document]
narcis-radu: That was the desired behaviour based on previous A/B tests.

test/features/google-login/google-login.test.js
narcis-radu: I did not find it necessary, but I can add it for peace of mind.

test/features/google-login/google-login.test.js
narcis-radu: Good call

libs/blocks/global-navigation/global-navigation.css
narcis-radu: I missed that, thank you

narcis-radu:
> I see some unit tests failing in `utils.js`, are they somehow related to this PR or are they coming from somewhere else?

@overmyheadandbody - looks like the failures are not related to this PR
Screenshot 2023-08-25 at 13 23 16
Screenshot 2023-08-25 at 13 23 31

libs/features/google-login.js
```diff
@@ -0,0 +1,53 @@
...
+    // Do nothing
+  }
+
+  await window.adobeIMS.socialHeadlessSignIn({
+    provider_id: 'google',
+    idp_token: data?.credential,
+    client_id: window.adobeid?.client_id,
+    scope: window.adobeid?.scope,
+  }).then(() => {
+    if (window.DISABLE_PAGE_RELOAD === true) return;
```
chrischrischris, line 20: super nit: do we need the `=== true` part?

libs/scripts/delayed.js
```diff
@@ -48,6 +48,16 @@ export const loadPrivacy = async (getConfig, loadScript) => {
     }
   });
 };
+
+export const loadGoogleLogin = async (getMetadata, loadIms, loadScript) => {
+  const googleLogin = getMetadata('google-login')?.toLowerCase();
+  if (googleLogin !== 'on') return;
+  if (window.adobeIMS?.isSignedInUser()) return;
```
chrischrischris, line 55: nit:
```suggestion
  if (googleLogin !== 'on' || window.adobeIMS?.isSignedInUser()) return;
```

chrischrischris (approved):

PR #1158, opened 2023-08-25
Mwpw 135324 continue button is not seen in iphone

sharath-kannan:
* The padding bottom has been changed to dvh.
* A fallback property for padding bottom has also been added.

Resolves: [MWPW-135324](https://jira.corp.adobe.com/browse/MWPW-135324)

**Test URLs:**
-before: https://main--milo--adobecom.hlx.page/drafts/tat28144/discover?martech=off
-After: https://main--milo--adobecom.hlx.page/drafts/tat28144/discover?martech=off

-Test: https://stage--cc--adobecom.hlx.page/drafts/ruchika/discover1?martech=off&milolibs=MWPW-135324--milo--sharath-kannan#animate


libs/styles/iframe.css
```diff
@@ -28,6 +28,8 @@
 
 @media (max-width: 1200px) {
   .dialog-modal.commerce-frame .milo-iframe{
-    padding-bottom:100vh;
+    padding-bottom: 100vh;
```
aishwaryamathuria, line 31: We can remove this if we are overriding this in next line

aishwaryamathuria (changes requested):

libs/styles/iframe.css
sharath-kannan: This has been added as a fallback for those browsers that do not support dvh.

spadmasa:
Validation done in saucelabs as we are not able to test in milo and cc due to cors error  : before the fix 
image

after the fix update 
image
image

Ipad before 
image
After 
image

libs/styles/iframe.css
sharath-kannan: @aishwaryamathuria I have tested dvh in all standard browsers and update the code with the required changes.

libs/styles/iframe.css
```diff
@@ -28,6 +28,7 @@
 
 @media (max-width: 1200px) {
   .dialog-modal.commerce-frame .milo-iframe{
-    padding-bottom:100vh;
+    /* stylelint-disable-next-line unit-no-unknown */
```
chrischrischris, line 31: I'm not getting a lint error when removing this line.

sharath-kannan: image
This is the error i'm getting when i remove the comment.

Ruchika4: I am getting the eslint error too.

Ruchika4 (approved):

aishwaryamathuria (approved):

chrischrischris (approved):

PR #1159, opened 2023-08-25
Fix failing tests

narcis-radu:
**Test URLs:**
- After: https://fix-failing-tests--milo--narcis-radu.hlx.page/drafts/narcis/default?martech=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -741,7 +741,7 @@ async function checkForPageMods() {
 
   const { env } = getConfig();
   const mep = PAGE_URL.searchParams.get('mep');
-  if (mep !== null || (env.name !== 'prod' && (persEnabled || targetEnabled))) {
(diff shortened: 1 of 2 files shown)


libs/utils/utils.js
```diff
@@ -741,7 +741,7 @@ async function checkForPageMods() {
 
   const { env } = getConfig();
   const mep = PAGE_URL.searchParams.get('mep');
-  if (mep !== null || (env.name !== 'prod' && (persEnabled || targetEnabled))) {
+  if (mep !== null || (env?.name !== 'prod' && (persEnabled || targetEnabled))) {
```
narcis-radu, line 744: This might not be needed since we have an environment name for every use-case.

honstar (approved):

overmyheadandbody (approved):

mokimo (approved):

PR #979, opened 2023-07-21
MWPW-129917 - Milo Aside block enhancements

aishwaryamathuria:
MWPW-129917 - Milo Aside block enhancements

* Product icon stack in large and medium split asides. Desktop mode will have 2 columns and mobile and tablet will have stack in a row. Author can add only icon, linked icon, icon with plain text or linked text.
* Static links to inherit link color same as the text. Adding style option in section metadata to change plain link color and inherit the same color as text.
* Option to set the aspect ratio of image/media for large and medium split asides. 
* Add an avatar icon for aside having contributor bio. Add a bio variation for aside that converts icon-area to avatar image.
* Aside center variation. No side image or media. Large variation has icon and supplemental text, Medium variation has icon or supplemental text, Small variation does not have either icon or supplemental text. Icon max sizes differ as per the XD for screen sizes.
* ~~Add play button to image having a modal link for all Aside variations (standard and split). Using Milo alt text to convert image to link and then formatting the link to have a play button.~~ This is removed from this PR and will be covered as a part of generic implementation in https://jira.corp.adobe.com/browse/MWPW-127223
* Support video in background and foreground for standard and split aside.
* Add a 24px padding before supplemental text.
* Spacing in Asides - For all aside (split and standard) the spacing is 40px between text and image, and 80px from margin. For split aside the image does not have any spacing from edge.

Discussion thread - https://github.com/orgs/adobecom/discussions/894 

Resolves: [MWPW-129917](https://jira.corp.adobe.com/browse/MWPW-129917)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/mathuria/aside/mwpw-129917?martech=off
- After: https://miloaside--milo--aishwaryamathuria.hlx.page/drafts/mathuria/aside/mwpw-129917?martech=off

PR #1162, opened 2023-08-25
[MWPW-131763] Variable Aside backgrounds based on media breakpoints

elan-tbx:
* Recent updates to the Aside block did not allow for CSS values to be used in the multi-background stacking order. This PR addresses that.
* The `decorateBlockBg()` function in `decorate.js` has been optimized/simplified, and now replaces the `decorateBlockBg()` definition that was within `aside.js`
* The requisite CSS for handling variable backgrounds has been added to `aside.css`
* Stylelinting errors have been fixed in `aside.css`

Resolves: [MWPW-131763](https://jira.corp.adobe.com/browse/MWPW-131763)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ebartholomew/aside?martech=off
- After: https://ebartholomew-aside-bg-breakpoints--milo--adobecom.hlx.page/drafts/ebartholomew/aside?martech=off

NOTE: Resize your browser window when viewing test URLs to test variable breakpoints.


meganthecoder:
Are you sure the `decorateBlockBg` export from decorate.js wasn't being used anywhere? That doesn't seem likely

test/blocks/aside/mocks/body.html
```diff
@@ -314,12 +315,14 @@ <h2 id="heading-xl-aside-standard-small">Heading XL Aside Standard Small</h2>
     <div class="aside split large">
       <div>
         <div data-valign="middle">
-          <a
-            href="https://images-tv.adobe.com/mpcv3/5286/b4401687-407b-444a-b9f1-9cd169952992_1660079495.1920x1080at3000_h264.mp4#autoplay1">https://images-tv.adobe.com/mpcv3/5286/b4401687-407b-444a-b9f1-9cd169952992_1660079495.1920x1080at3000_h264.mp4#autoplay1</a>
+          <video playsinline="" autoplay="" loop="" muted="">
+            <source src="./drafts/mathuria/aside/media_1491fc79163bea6b7f6daeeb266898344ff827142.mp4" type="video/mp4">
```
JasonHowellSlavin, line 319: These sources in the mock will result in 404s in our tests. Is there a way to remove them or are they needed for testing?

libs/blocks/aside/aside.css
```diff
@@ -426,7 +452,7 @@
 
 .aside.notification .foreground.container [data-align=center],
 .aside.notification.center .foreground.container,
-.aside.notification.center .foreground.container > * {
+.aside.notification.center .foreground.container>* {
```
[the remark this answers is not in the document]
elan-tbx: @JasonHowellSlavin I think this set off my subconscious "mental linter", sorry. I can revert it.

test/blocks/aside/mocks/body.html
elan-tbx: @JasonHowellSlavin These are holdovers from #979 that I just copy pasted to better align with the new background video functionality, I'm totally fine with removing the sources throughout to clean things up 👍

libs/blocks/aside/aside.css
```diff
@@ -92,43 +101,75 @@
...
   display: flex;
-  flex-wrap: wrap;
-  align-content: center;
 }
 
-.aside .foreground.container .image {
-  position: relative;
+.aside.inline .foreground.container .image,
+.aside.inline .foreground.container div:has(.milo-video),
+.aside.inline .foreground.container .text {
```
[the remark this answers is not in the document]
elan-tbx: @JasonHowellSlavin Honestly I have no idea if the specificity is necessary here -- the re-ordering I did in this file was the main work I did to resolve style-linting errors. Someone else wrote those styles a long time ago, but the style linter was throwing [no-descending-specificity](https://stylelint.io/user-guide/rules/no-descending-specificity/) errors all over the file so I moved things around and tested the aside variants again -- seems to have gone smoothly.

elan-tbx:
> Are you sure the `decorateBlockBg` export from decorate.js wasn't being used anywhere? That doesn't seem likely
> 
> Edit: I just checked and I see it in 4 different blocks. Can you check that your new code works with all of those blocks and provide examples for testing?

@meganthecoder Oof. Sorry. Apparently the `Find References` feature in VS Code does not work like I thought it did 🤦🏻‍♂️ I'll make sure to test them all.

ryanmparrish:
@elan-tbx  - Looking good so far. 
I was wondering if we should think about making some of those styles shareable to support the viewport styles needed across multiple blocks that use this feature. If we can make the classes generic we could probably put these into `typography.css`. Maybe even scope it to  `.con-block .background` 🤷 

Some blocks might wan't different props for a specific reason and may need to audit all these block and add any block style tweaks for any specific support. 

https://ebartholomew-aside-bg-breakpoints--milo--adobecom.hlx.page/drafts/rparrish/media-breakpoints
I created a draft w/ all the various block using it. Id say if we can get these all to play w/ the shared bgDecorator etc were GTG. I know the `marquee` original is an outlier and looks like `marquee-anchors` needs some '.background' support in general. LMK how I can help. Thanks

libs/utils/decorate.js
```diff
@@ -77,20 +77,20 @@ export function decorateBlockText(el, config = ['m', 's', 'm'], type = null) {
...
-        /* c8 ignore next */
-        e.classList.add(viewports[i]);
-      });
-    }
-  }
-  if (!node.querySelector(':scope img')) {
+  const childCount = node.childElementCount;
+  if (node.querySelector('img, video, a[href*=".mp4"]') || childCount > 1) {
+    node.classList.add('background');
+    const binaryVP = [['mobile-only', 'tablet-only'], ['desktop-only']];
```
rgclayton, line 83: To me this seems overly complicated. Maybe I don't understand the goal here? 

The behavior should be:

1. If only 1 breakpoint is authored (default) then mobile, tablet, desktop get that background.
2. If 2 breakpoints are authored (mobile, tablet) Mobile only shows on mobile, tablet displays and desktop inherits tablet.
3. If 3 breakpoints are authored each breakpoint gets is own background

Something is not working correctly in the second scenario.

Ideally the only breakpoints that are added are the ones that are defined/authored and the mobile first cascade of css does its thing. 

[Example page ](https://ebartholomew-aside-bg-breakpoints--milo--adobecom.hlx.page/drafts/rclayton/test/background-per-breakpoint)

In the first aside on the page, i have 2 breakpoints defined. Mobile (Red), Tablet (image). I would expect the image would display on tablet and desktop. But, what happens is the image only shows on desktop and the Red color is displayed on mobile and tablet.

rgclayton (changes requested):
I'm seeing some breakpoint issues if only two breakpoints are defined.

libs/utils/decorate.js
elan-tbx: @rgclayton this is just a refactoring of the same behavior this function had before -- both versions do the same thing. If you look at the Marquee and its proprietary `decorateBlockBg()`, it also combines mobile and tablet into the same effective "breakpoint", with desktop being separate (not inheriting from tablet).

I'm happy to change it up though, if `(mobile, (tablet, desktop))` is indeed the new direction we want to go in?

Screenshot 2023-09-06 at 12 07 07 PM

elan-tbx: @rgclayton Here's an example page showing current behavior on the `main` branch

https://main--milo--adobecom.hlx.page/drafts/ebartholomew/two-breakpoints

rgclayton: Ahh. Got it. I don't think this needs fixed in this ticket then if it was already existing. To me it seems like a bug with how the viewports are set up.

@ryanmparrish can you provide any clarity here?

rgclayton (commented):
@ryanmparrish Can you help clarify about the viewport set up?

libs/utils/decorate.js
ryanmparrish: @rgclayton - You're right, this is set up wrong IMO for the case w/ just 2 bg cells. 
It should be
`[mobile][tablet/desktop]`
It currently is. 
`[mobile/tablet][desktop]`

I'm open as to wether we fix this here or in a separate PR. But I do think we should address it and come correct.. (that probably means looking at potential usage in the wild if thats even searchable?) 
Otherwise we leave it as is and deal w/ the weird anti-patern in this small edge case. I looked at my, [original draft](https://main--milo--adobecom.hlx.page/drafts/rparrish/marquee) and see it wasn't quite right. ex. '2 viewport images (mobile/tablet, desktop)' :(

rgclayton: >It should be
[mobile][tablet/desktop]
It currently is.
[mobile/tablet][desktop]

Yep. 

@elan-tbx if you have the bandwidth I think we fix it in this PR and squash the issue. Otherwise code is just getting added on to it that tangles it more. Sooner it can get fixed the better.  

cc: @ryanmparrish

elan-tbx: @rgclayton yep, I have the bandwidth. Just to make sure -- since the Marquee has its own function that also does this, should I also make the change in that block in this PR? Feels kind of out of scope of this PR's original purpose, but might be worth making an exception for this.

cc: @ryanmparrish

libs/blocks/aside/aside.js
```diff
@@ -73,38 +73,6 @@ function decorateMedia(el) {
   });
 }
 
-function decorateVideo(container) {
-  const link = container.querySelector('a[href*=".mp4"]');
```
aishwaryamathuria, line 77: @elan-tbx Currently milo utils only treat fragment and franklin or slack uploaded videos as autoblock and hence this was added to make sure that if any link remains in background that matched a video format, it should be converted to a video tag.

images-tv.adobe.com BG applied on main - https://main--milo--adobecom.hlx.page/drafts/mathuria/aside/asidebackground 

images-tv.adobe.com BG failing on the PR branch - https://ebartholomew-aside-bg-breakpoints--milo--adobecom.hlx.page/drafts/mathuria/aside/asidebackground

In case if this needs to be removed the following line in utils.js decorateAutoBlock can be updated to handle all videos in place of only uploaded videos 
`if (key === 'video' && !a.textContent.match('media_.*.mp4')) `

Also the following line in video.js would need to be updated to add the full href in source pathname if the href domain does not match the current domain so that it does not convert the external video links (for instance from images-tv.adobe.com) into relative paths -
<source src=".${pathname}" type="video/mp4" />

elan-tbx: @aishwaryamathuria thanks for the context!

@chrischrischris what do you think would be the best way to proceed here? I deferred to the auto-block in this case to avoid the race condition we were seeing, but not sure if we want to open up video.js to non-slack uploaded videos.

chrischrischris: Aside and Marquee are the only 2 blocks that support .mp4 background videos.  We don't want authors using .mp4 except for these 2 use cases.  We may be able to consolidate the code into decorate.js however I'm not sure if there are minor differences between the 2 block bg videos.

elan-tbx: @chrischrischris That makes sense. In that case, how would your recommend avoiding the video.js race condition once I bring back the function @aishwaryamathuria is talking about here? Is there a way I can opt out of auto-blocking with the background specifically?

elan-tbx: I just read through #1242 -- think I understand now. Thanks all.

libs/utils/decorate.js
```diff
@@ -78,20 +78,20 @@ export function decorateBlockText(el, config = ['m', 's', 'm'], type = null) {
...
-        e.classList.add(viewports[i]);
-      });
-    }
-  }
-  if (!node.querySelector(':scope img')) {
+  const childCount = node.childElementCount;
+  if (node.querySelector('img, video, a[href*=".mp4"]') || childCount > 1) {
+    node.classList.add('background');
+    const binaryVP = [['mobile-only'], ['tablet-only', 'desktop-only']];
+    const allVP = [['mobile-only'], ['tablet-only'], ['desktop-only']];
```
ryanmparrish, line 85:
```suggestion
```

libs/utils/decorate.js
```diff
@@ -78,20 +78,20 @@ export function decorateBlockText(el, config = ['m', 's', 'm'], type = null) {
...
-        /* c8 ignore next */
-        e.classList.add(viewports[i]);
-      });
-    }
-  }
-  if (!node.querySelector(':scope img')) {
+  const childCount = node.childElementCount;
+  if (node.querySelector('img, video, a[href*=".mp4"]') || childCount > 1) {
+    node.classList.add('background');
+    const binaryVP = [['mobile-only'], ['tablet-only', 'desktop-only']];
```
ryanmparrish, line 84:
```suggestion
```

libs/utils/decorate.js
```diff
@@ -78,20 +78,20 @@ export function decorateBlockText(el, config = ['m', 's', 'm'], type = null) {
...
+    const allVP = [['mobile-only'], ['tablet-only'], ['desktop-only']];
+    const viewports = childCount === 2 ? binaryVP : allVP;
```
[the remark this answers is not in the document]
ryanmparrish: Do you think we should follow the same naming convention as the other media queries?
`mobile-only`, `tablet-up`, `desktop-up`?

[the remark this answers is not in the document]
elan-tbx: @ryanmparrish this is how I originally had it, but @JasonHowellSlavin requested I break it up 3 weeks ago: https://github.com/adobecom/milo/pull/1162#discussion_r1309081018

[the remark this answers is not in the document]
elan-tbx: totally fine with changing the names, though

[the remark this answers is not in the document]
ryanmparrish: Ah I see, Im good w/ it as is. the '-only' makes sense in this case since we're just toggling the display.

libs/styles/styles.css
```diff
@@ -496,12 +496,45 @@ div[class*='-up'] .con-block.has-bg.xxl-spacing { padding: var(--spacing-xxl); }
...
+  top: 0;
+  bottom: 0;
+}
+
+.con-block .background>div { height: 100%; }
+
+.con-block .background img,
+.con-block .background video {
+  object-fit: cover;
+  min-height: 700px;
```
ryanmparrish, line 521: I'm seeing an issue w/ this height being set all the time. 

```suggestion
```
If we need a min height on .background assets, aka `.marquee.large`, we should add that in the block specific css. 
This is causing some assets to get clipped in smaller height views. 

Screen Shot 2023-09-19 at 3 45 41 PM
Screen Shot 2023-09-19 at 3 45 56 PM

ryanmparrish (approved):

libs/utils/decorate.js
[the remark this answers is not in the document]
Sartxi:
```suggestion
  const viewports = [['mobile-only'], ['tablet-only'], ['desktop-only']];
  if (childCount === 2) viewports[1] = [...viewports[1], ...viewports.pop()];
```

you could also simplify this by assigning all viewports and simply moving things around after

aishwaryamathuria (approved):

Sartxi (approved):

PR #1171, opened 2023-08-28
[MWPW-135217] Use decorateLinks before decorating Gnav

overmyheadandbody:
This PR aims to better follow Milo patterns for link handling in Gnav, to allow authors to use `#_blank`, `#_dnt`, etc. in a consistent way and to ensure all links are localised properly and receive the `.html` extension if needed.

Some additional notes:
- for the footer, `decorateLinks` was called after our decoration logic, we're now calling it before, to follow the same pattern as the (updated) Gnav logic. This causes two loose ends, for which we're opening a follow-up story to address as soon as there is bandwidth (https://jira.corp.adobe.com/browse/MWPW-135656);
- social links in the footer now receive a `target="_blank"` attribute, as these links always lead outside of adobe.com pages;
- dropdown triggers have been transformed into buttons. These shouldn't count towards analytics link hits and this should be better in terms of accessibility. This also reduces the scope of links to be decorated and avoids potential localisation issues;
- the breadcrumb links are no longer localised on their own, as the main 'global-navigation.js' file decorates the links beforehand;
- the logic that handled SVG decoration has been removed from `getBrandImage`, as the SVG should have been decorated previously by the `decorateLinks` call;
- the `createFullGlobalNavigation` method from 'test-utilities.js' has been updated to have arguments for a custom config and breadcrumbs element to be used when creating a full navigation for unit testing purposes;
- a 'large-menu.plain.html' file has been created, as a 404 message was being logged to the console;

### Diffs
To make sure all links are decorated properly and no new issues are introduced, I cross-referenced diffs between the decorated header and footer between the current logic and the one from this PR. I'm attaching them here as well

 | Header | Footer |
| Header diff | Footer diff |

Resolves: [MWPW-135217](https://jira.corp.adobe.com/browse/MWPW-135217), [MWPW-135073](https://jira.corp.adobe.com/browse/MWPW-135073)

**Test URLs:**
- Before: https://main--milo--overmyheadandbody.hlx.page/ch_de/drafts/ramuntea/gnav-refactor?martech=off
- After: https://hash-in-gnav-links--milo--overmyheadandbody.hlx.page/ch_de/drafts/ramuntea/gnav-refactor?martech=off

### DIFF
--- a/libs/blocks/global-footer/global-footer.js
+++ b/libs/blocks/global-footer/global-footer.js
@@ -68,6 +68,16 @@ class Footer {
 
     // TODO: log to LANA if Footer content could not be found
     if (!this.body) return;
+    // TODO: revisit region picker and social links decoration logic
+    const regionAnchor = this.body.querySelector('.region-selector a');
+    if (regionAnchor?.href) {
+      regionAnchor.setAttribute('href', `${regionAnchor.getAttribute('href')}#_dnt#_dnb`);
+    }
+    const socialLinks = document.querySelectorAll('.social a');
+    socialLinks.forEach((socialLink) => {
+      socialLink.setAttribute('href', `${socialLink.getAttribute('href')}#_dnb`);
+    });
+    decorateLinks(this.body);
 
     // Order is important, decorateFooter makes use of elements
     // which have already been created in previous steps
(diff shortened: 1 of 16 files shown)


mokimo (approved):
Nice 👍 
Good to solve this properly at the right location(s) rather than the occasional quick fixes that we had done before. Always good to solve the issue at root rather than treating the symptoms

narcis-radu (approved):

SilviuLCF (approved):
Verified ready to merge 
https://hash-in-gnav-links--milo--overmyheadandbody.hlx.page/drafts/sghimpos/gnav-1

PR #1176, opened 2023-08-29
Add code block

hparra:
We add a `code` block. See README below.

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.live/drafts/hgpa/test/code/code?martech=off
- After: https://feat-code--milo--hparra.hlx.page/drafts/hgpa/test/code/code?martech=off


hparra:
`helix-psi-check` seems  to be lagging today.

libs/blocks/code/code.js
```diff
@@ -0,0 +1,44 @@
+import * as utils from '../../utils/utils.js';
+
+export const HIGHLIGHT_JS = 'https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.8.0/highlight.min.js';
```
mokimo, line 3: Let's store this in libs/deps

mokimo:
Does this have any Ticket assosciated with it?

libs/blocks/code/code.js
hparra: I've copied over JS to libs/deps. I've also refactored theme support to light/dark variants and embedded that CSS instead.

libs/blocks/code/README.md
```diff
@@ -0,0 +1,41 @@
+# Code
+
+The `code` block allows you create a formatted block of programming code, complete with language specific syntax highlighting.
+
+## Use
+
+Paste code into block and format code as monospaced text. Block will not render and error if text is not monospaced. See "Tips" for more details.
```
JasonHowellSlavin, line 7:
```suggestion
Paste code into the block and format the code as monospaced text. The block will not render and will log an error if text is not monospaced. See "Tips" for more details.
```

libs/styles/typography.css
```diff
@@ -129,6 +132,27 @@
...
+  font-size: var(--type-code-s-size);
+  line-height: var(--type-code-s-lh);
+}
+
+.code-l,
+.code-m,
+.code-s {
+  font-weight: var(--type-code-all-weight);
+}
+
```
JasonHowellSlavin, line 155: Would it be better to put these in `code.css`? In typography even though minimal, still adds bytes to pages.

libs/styles/variables.css
```diff
@@ -97,6 +98,15 @@
...
 
+  /* Code */
+  --type-code-all-weight: 400;
+  --type-code-l-size: 18px;
+  --type-code-l-lh: 27px;
+  --type-code-m-size: 16px;
+  --type-code-m-lh: 24px;
+  --type-code-s-size: 14px;
+  --type-code-s-lh: 21px;
+
```
JasonHowellSlavin, line 109: same with the vars here, could they go into `code.css`?

JasonHowellSlavin (commented):
small nitpicky stuff. Looks good!

libs/styles/typography.css
hparra: Maybe? Consonant has typography for code which we have not implemented, so I thought it best to add it the same way the other typography is defined. See surrounding code and https://consonant.adobe.com/1975e5ba1/p/216c7e-typography/t/244b88

It can be argued that any `<pre><code>` on page should have typography styling, including the font, but that's potentially a high impact CSS change and loading another font all the time would definitely be slower.

JasonHowellSlavin: 🤷  Either way for me, it is 33 (including vars below) lines of CSS, so not a lot. If it follows consonant and this is the best spot for it, why not.

hparra: I made https://github.com/adobecom/milo/pull/1187 to address this separately.

hparra:
Please see #1187. If that lands first then this PR will become smaller once rebased.

hparra:
It looks like HelpX made a similar code block https://github.com/adobecom/helpx-internal/pull/11

Problem is that these blocks have the same name so there may be a conflict. I need to confirm this.

auniverseaway (changes requested):
See the gist autoblock...

https://milo.adobe.com/developer/academy/list

hparra:
> See the gist autoblock...
> 
> https://milo.adobe.com/developer/academy/list

@auniverseaway I mentioned it README. I don't want to depend on github.

PR #1182, opened 2023-08-30
Use constants for milo events

chrischrischris:
Prevents possible typos from breaking things.

Also fixes some linting errors.

**Test URLs:**
- After: https://miloevents--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -115,6 +115,8 @@ ENVS.local = {
   name: 'local',
 };
 
+export const EVENT_MILO_DEFERRED = 'milo:deferred';
+export const EVENT_MILO_LCP_LOADED = 'milo:LCP:loaded';
 const LANGSTORE = 'langstore';
 const PAGE_URL = new URL(window.location.href);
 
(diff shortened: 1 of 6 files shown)


mokimo (approved):

VKniaz (approved):

sukamat (approved):

libs/utils/utils.js
```diff
@@ -115,6 +115,8 @@ ENVS.local = {
   name: 'local',
 };
 
+export const EVENT_MILO_DEFERRED = 'milo:deferred';
+export const EVENT_MILO_LCP_LOADED = 'milo:LCP:loaded';
```
overmyheadandbody, line 119: If this list grows with time, would it not make sense to have a single `EVENTS` object containing all events? Then you would just export one constant and you would import the same constant name wherever events are needed (instead of multiple ones if more events are required in the file).

overmyheadandbody (approved):
Good idea to start consolidating things!

libs/utils/utils.js
chrischrischris: Good idea, will update to do that

PR #1186, opened 2023-08-30
MWPW-134986 - Rendering nested fragments in the order of products

sabyamon:
With the existing logic, the rows were getting traversed and on the way, the secondary products were matched.
Hence, if the product code is defined in a different order in result resources compared to the product codes array, the fragments are picked in that order.

With this PR, we will be traversing the product array first and will match that with every result resource row.
It's obviously less performant as for each entry in the product code array (which is very small in size), we will be iterating all the rows of the result resources (which is relatively a much bigger array). But, this is a tradeoff we will have to bear to maintain the matched product code order.

Note - This does not fix the order for basic fragments for more than one primary product without an umbrella product.

Resolves: https://jira.corp.adobe.com/browse/MWPW-134986

**Test URLs:**
- After: https://uar-nested-frag-in-order--milo--sabyamon.hlx.page/drafts/xiasun/quiz-2/?martech=off

### DIFF
--- a/libs/blocks/quiz/utils.js
+++ b/libs/blocks/quiz/utils.js
@@ -202,18 +202,20 @@ export const nestedFragments = (
 
 const getNestedFragments = (resultResources, productCodes, fragKey) => {
   const fragArray = [];
-  resultResources?.data?.forEach((row) => {
-    if (productCodes.length > 0 && productCodes.includes(row.product)) {
-      insertFragment();
-    }
-
-    function insertFragment() {
-      if (row[fragKey]) {
-        row[fragKey].split(',').forEach((val) => {
-          fragArray.push(val.trim());
-        });
+  productCodes?.forEach((product) => {
+    resultResources?.data?.forEach((row) => {
+      if (product && product === row?.product) {
+        insertFragment();
       }
-    }
+
+      function insertFragment() {
+        if (row[fragKey]) {
+          row[fragKey]?.split(',').forEach((val) => {
+            fragArray.push(val.trim());
+          });
+        }
+      }
+    });
   });
   return fragArray;
 };


fullcolorcoder (approved):

libs/blocks/quiz/utils.js
```diff
@@ -202,18 +202,20 @@ export const nestedFragments = (
...
-        });
+      function insertFragment() {
+        if (row[fragKey]) {
+          row[fragKey]?.split(',').forEach((val) => {
+            fragArray.push(val.trim());
+          });
+        }
       }
-    }
+    });
```
JasonHowellSlavin, line 218: can we move this function declaration above line 207 `if (product && product === row?.product) {`

JasonHowellSlavin (commented):
Might be outside of the scope of this ticket, but is there a way to flatten `getNestedFragments`? By the time we get into `insertFragment` with it's `row[fragKey]?.split(...)` we are in the 3rd level of loops.

JackySun9:
for some reason, I cannot test this, any suggestion on this?

libs/blocks/quiz/utils.js
echampio-at-adobe: Hi @sabyamon, 

As far as I can see, when I set a breakpoint in storeResultInLocalStorage(), I believe the fragments are coming in the correct order. For some reason the results page doesn't load from the uar-nested-frag-in-order branch, but I'm approving based upon what I see in that last method call before it redirects. 

https://uar-nested-frag-in-order--milo--adobecom.hlx.live/drafts/xiasun/uar-results-block/uar-results?primary=lr-edu,pr-edu&quizKey=cc-quiz-en-US

If you can work on the comments from @JasonHowellSlavin, I think we should be good to go afterwards.

sabyamon: hi @JasonHowellSlavin, I am a bit unclear about why you want me to move the `insertFragment` function above line 207. Can you please explain?

sabyamon:
> Might be outside of the scope of this ticket, but is there a way to flatten `getNestedFragments`? By the time we get into `insertFragment` with it's `row[fragKey]?.split(...)` we are in the 3rd level of loops.

Hi @JasonHowellSlavin, you are right about the 3rd level of iteration. But we are dealing with 2 levels of inter-dependent data with tuples inside it which need to be separated and processed. I do not think we can reduce the complexity of this without making a major refactoring of either the code or the data model or both.

JackySun9 (approved):

libs/blocks/quiz/utils.js
JasonHowellSlavin: Sure. `insertFragment` is used on 208, but defined on 211. If running eslint, this should set off an error: `no-use-before-defined`. Though the function will be hoisted, it is preferable to not rely on hoisting.

PR #1187, opened 2023-08-31
Add support for Consonant code typography

hparra:
Load font and styling to support [Consonant's code typography](https://consonant.adobe.com/1975e5ba1/v/32165/p/216c7e-typography/t/244b88) when document contains any inlined or blocked monospaced text.

Screenshot 2023-08-30 at 19 17 02

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.live/drafts/hgpa/test/code/code?martech=off
- After: https://feat-code-noblock--milo--hparra.hlx.live/drafts/hgpa/test/code/code?martech=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -784,8 +784,12 @@ async function loadPostLCP(config) {
     header.classList.remove('gnav-hide');
   }
   loadTemplate();
-  const { default: loadFonts } = await import('./fonts.js');
-  loadFonts(config.locale, loadStyle);
+  import('./fonts.js').then((module) => {
+    module.default(config.locale, loadStyle);
+    if (document.querySelector('code')) {
+      module.loadTypeKitFont('vih2anh.css', loadStyle);
+    }
+  });
 }
 
 export async function loadDeferred(area, blocks, config) {
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -775,6 +775,14 @@ async function checkForPageMods() {
   }
 }
 
+export async function loadFonts(config, { loadStyle: utilsLoadStyle }) {
+  const { default: loadTypeKitFonts, loadTypeKitFont } = await import('./fonts.js');
+  return Promise.allSettled([
+    loadTypeKitFonts(config.locale, utilsLoadStyle),
+    document.querySelector('code') && loadTypeKitFont('vih2anh.css', utilsLoadStyle),
+  ]);
+}
+
 async function loadPostLCP(config) {
   loadMartech(config);
   const header = document.querySelector('header');
@@ -784,12 +792,7 @@ async function loadPostLCP(config) {
     header.classList.remove('gnav-hide');
   }
   loadTemplate();
-  import('./fonts.js').then((module) => {
-    module.default(config.locale, loadStyle);
-    if (document.querySelector('code')) {
-      module.loadTypeKitFont('vih2anh.css', loadStyle);
-    }
-  });
+  loadFonts(config, { loadStyle });
 }
 
 export async function loadDeferred(area, blocks, config) {
--- a/libs/styles/styles.css
+++ b/libs/styles/styles.css
@@ -100,6 +100,13 @@
   --type-detail-s-size: 10px;
   --type-detail-s-lh: 12px;
 
+  /* Code */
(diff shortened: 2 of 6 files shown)


hparra:
I also made a code block #1176 that has most of this code, but the Consonant styling shouldn't depend on whether the code is in a block or not.

libs/styles/variables.css
```diff
@@ -30,6 +30,7 @@
 
   /* Fonts */
   --body-font-family: 'Adobe Clean', adobe-clean, 'Trebuchet MS', sans-serif;
+  --code-font-family: 'Source Code Pro', source-code-pro, monospace;
   --fixed-font-family: 'Roboto Mono', menlo, consolas, 'Liberation Mono', monospace;
```
hparra, line 34: I checked all repos under org and nothing is using `--fixed-font-family`. Delete?

libs/styles/variables.css
```diff
@@ -97,6 +98,15 @@
...
   --type-detail-s-lh: 12px;
 
+  /* Code */
+  --type-code-all-weight: 400;
+  --type-code-l-size: 18px;
+  --type-code-l-lh: 27px;
+  --type-code-m-size: 16px;
+  --type-code-m-lh: 24px;
+  --type-code-s-size: 14px;
+  --type-code-s-lh: 21px;
```
overmyheadandbody, line 108: All line heights are actually `1.5` when compared to the fonts. Do we even need them defined explicitly?

libs/utils/utils.js
```diff
@@ -778,8 +778,12 @@ async function loadPostLCP(config) {
...
   }
   loadTemplate();
-  const { default: loadFonts } = await import('./fonts.js');
-  loadFonts(config.locale, loadStyle);
+  import('./fonts.js').then((module) => {
+    module.default(config.locale, loadStyle);
+    if (document.querySelector('code')) {
+      module.loadTypeKitFont('vih2anh.css', loadStyle);
+    }
+  });
```
overmyheadandbody, line 786: Maybe a unit test would be valuable here to check that requests are made for the respective TypeKit fonts

libs/styles/variables.css
overmyheadandbody: HelpX seems to use it in their [styles](https://github.com/adobecom/helpx/blob/37b7eb6956405970321e53657b140e32c6784a1a/styles/styles.css#L108), but they also define it [themselves](https://github.com/adobecom/helpx/blob/37b7eb6956405970321e53657b140e32c6784a1a/styles/styles.css#L24), so this does seem redundant to me.

libs/styles/typography.css
```diff
@@ -129,6 +132,27 @@
...
+
+.code-s {
+  font-size: var(--type-code-s-size);
+  line-height: var(--type-code-s-lh);
+}
+
+.code-l,
+.code-m,
+.code-s {
+  font-weight: var(--type-code-all-weight);
```
overmyheadandbody, line 153: `line-height: 1.5` could go here directly.

overmyheadandbody (approved):

libs/styles/variables.css
hparra: I did not notice that. I’m following the conventions of the surrounding typography, which do not follow a consistent scale. Should we be explicit like Consonant specs or are we free to optimize? I’m ok with either way.

hparra:
TODO: Rebase after https://github.com/adobecom/milo/pull/1194

auniverseaway:
What shipping page needs this? I'm really struggling with all the additions we keep piling on to utils.js.

hparra:
@auniverseaway It's part of Consonant. It should already be there whenever a page uses monospaced text.

auniverseaway (changes requested):
1. We need a JIRA.
2. I don't think we should ship this until there's an actual need. I dislike shipping features that no one is actually asking for or is not tied to a page going live. "It's in Consonant" in my mind is not good enough... there's a lot of stuff in Consonant that doesn't exist in Milo.
3. If we are going to ship, I would not make utils more complicated, I would move the complexity into fonts.js.

hparra:
Thanks for your review @overmyheadandbody. Closing.

PR #1189, opened 2023-08-31
mwpw-133639: anchor tags should be buttons

mokimo:
### Remove undesirable links in search engine results

Due to how google indexes pages log in pages might end up being indexed where they are not interesting. For this reason, IMS related Sign in "Anchor" tags should really be buttons.

Resolves: [MWPW-133639](https://jira.corp.adobe.com/browse/MWPW-133639)

### Test instructions
The "sign in" button should be a button instead of an anchor tag (first link osahin/gnav-1) - as standalone button as well as within a sign-in dropdown (second link ramuntea/gnav-refactor)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/osahin/gnav-1?martech=off
- After: https://mwpw-133693--milo--mokimo.hlx.page/drafts/osahin/gnav-1?martech=off
- Before: https://main--milo--adobecom.hlx.page/drafts/osahin/gnav-1?martech=off
- After: https://mwpw-133693--milo--mokimo.hlx.page/drafts/ramuntea/gnav-refactor?martech=off


libs/blocks/global-navigation/global-navigation.js
```diff
@@ -63,25 +63,26 @@ const decorateSignIn = async ({ rawElem, decoratedElem }) => {
...
 
     dropdownElem.classList.add('feds-signIn-dropdown');
 
-    const dropdownSignIn = dropdownElem.querySelector('[href$="?sign-in=true"]');
-
-    if (dropdownSignIn) {
-      dropdownSignIn.addEventListener('click', (e) => {
+    const dropdownSignInAnchor = dropdownElem.querySelector('[href$="?sign-in=true"]');
+    if (dropdownSignInAnchor) {
+      const dropdownSignInButton = toFragment`<button qa="gnav-ims-sign-in">${dropdownSignInAnchor.innerText}</button>`;
```
overmyheadandbody, line 83: I'd suggest the use of `textContent` instead of `innerText`. The latter needs layout information, the former is more performant and standartdised.

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -63,25 +63,26 @@ const decorateSignIn = async ({ rawElem, decoratedElem }) => {
...
-    signInElem = toFragment`<a href="#" daa-ll="${signInLabel}" class="feds-signIn">${signInLabel}</a>`;
+    signInElem = toFragment`<button qa="gnav-ims-sign-in" daa-ll="${signInLabel}" class="feds-signIn">${signInLabel}</button>`;
 
     signInElem.addEventListener('click', (e) => {
       e.preventDefault();
       signIn();
     });
   } else {
-    signInElem = toFragment`<a href="#" daa-ll="${signInLabel}" class="feds-signIn" role="button" aria-expanded="false" aria-haspopup="true">${signInLabel}</a>`;
+    signInElem = toFragment`<button qa="gnav-ims-sign-in" daa-ll="${signInLabel}" class="feds-signIn" aria-expanded="false" aria-haspopup="true">${signInLabel}</button>`;
```
overmyheadandbody, line 73: I have to say I'm not in love with this attribute. Attribute name is pretty generic, its value is prefixed with _gnav_, whereas all our prefixes are _feds_, it can represent either a sign in link OR both a dropdown trigger and a sign in link and its only purpose is to make unit tests a bit more convenient. I'm not convinced this is the right approach.

Since we decide what gets turned into a button and what remains a link, wouldn't just checking for the button be enough?

overmyheadandbody (commented):
I do have a couple of concerns before going forward with this

libs/blocks/global-navigation/features/profile/dropdown.js
```diff
@@ -157,7 +157,7 @@ class ProfileDropdown {
   decorateSignOut() {
     const signOutLink = toFragment`
       <li>
-        <a href="#" class="feds-profile-action" daa-ll="${this.placeholders.signOut}">${this.placeholders.signOut}</a>
+        <button class="feds-profile-action" daa-ll="${this.placeholders.signOut}">${this.placeholders.signOut}</button>
```
narcis-radu, line 160: The computed font is now Arial for me, we should use Adobe's font family.
Screenshot 2023-08-31 at 16 32 04

narcis-radu: Other things to look out for after you change the `font-family`: `font-size` and `line-height`; you need to set them explicitly otherwise browser defaults will be applied.

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -421,13 +426,19 @@ header.global-navigation {
   padding: 12px 0;
 }
 
-.feds-signIn-dropdown li > a {
+.feds-signIn-dropdown li > a,
+.feds-signIn-dropdown li > button {
```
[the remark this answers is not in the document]
mokimo: This will lead to issues as `a` has precedence, which is the reason I've added `li > a` in the first place 🤔 
Open for suggestions! 
Screenshot 2023-09-01 at 12 09 35

libs/blocks/global-navigation/global-navigation.js
mokimo: Adding the `signIn` class to both now to share some styles. But you are right we could also just use the button to see which is the ims sign in!

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -421,13 +426,19 @@ header.global-navigation {
...
   display: block;
   color: var(--feds-color-link--light);
   padding: 6px 32px;
+  border: none;
+  width: 100%;
+  text-align: left;
+  line-height: 1.4;
```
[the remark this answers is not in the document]
mokimo: I went with a different approach and just inherit the line-height

chrischrischris (approved):

chadsunvice (approved):

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -391,6 +391,16 @@ header.global-navigation {
   border-radius: var(--feds-radius-utilityIcon);
   color: #4B4B4B;
   white-space: nowrap;
+  border: none;
+  text-align: start;
+  font-family: inherit;
+  font-size: inherit;
+  line-height: inherit;
```
overmyheadandbody, line 398: Did you try with just `font: inherit`? Since it's a shorthand, that should inherit all 3 of these properties + the `font-size`.

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -421,15 +431,17 @@ header.global-navigation {
...
+.feds-signIn-dropdown li > a,
+.feds-signIn-dropdown .feds-signIn {
   display: block;
   color: var(--feds-color-link--light);
   padding: 6px 32px;
 }
 
-.feds-signIn-dropdown li > a:hover {
+.feds-signIn-dropdown li > a:hover,
+.feds-signIn-dropdown li > button:hover {
```
overmyheadandbody, line 442: I see this just uses `button`, while L435 uses `.feds-signIn`. Do they need to be different?

SilviuLCF (approved):
Changes tested ready to merge
Testing details https://jira.corp.adobe.com/browse/MWPW-133639

libs/blocks/global-navigation/global-navigation.js
chadsunvice: Too bad this attribute didn't make it as it would have given SDETs a way to target the button via attribute, instead of class, which isn't as reliable cause classes are not unique. Also, it would have started a best practice to add anchors for automation.

PR #1191, opened 2023-08-31
MWPW-135814 Add underline in static link

aishwaryamathuria:
Link and text is not differentiable in case of static links. Currently the default behaviour of links involve no-underline with underline on hover. When static link class is applied the text and links have same color and no underline makes it difficult to differentiate from the text. 
Adding underline in static link by default to be in-sync with notification static link style of a.static.(https://github.com/aishwaryamathuria/milo/blob/main/libs/styles/styles.css). 

Resolves: [MWPW-135814](https://jira.corp.adobe.com/browse/MWPW-135814)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/mathuria/aside/documentation/sectionmetadatadoc?martech=off
- After: https://staticlink--milo--aishwaryamathuria.hlx.page/drafts/mathuria/aside/documentation/sectionmetadatadoc?martech=off

### DIFF
--- a/libs/blocks/section-metadata/section-metadata.css
+++ b/libs/blocks/section-metadata/section-metadata.css
@@ -19,6 +19,7 @@
 
 .section.static-links a:not([class]) {
   color: inherit;
+  text-decoration: underline;
 }
 
 .section.hide-sticky-section,


ryanmparrish (approved):

Ruchika4 (approved):

Sartxi (approved):

rgclayton:
I think on hover the underline should disappear. 
@ryanmparrish Any thoughts from the Consonant side?

mokimo (approved):

libs/blocks/section-metadata/section-metadata.css
```diff
@@ -19,6 +19,7 @@
 
 .section.static-links a:not([class]) {
   color: inherit;
+  text-decoration: underline;
```
overmyheadandbody, line 22: As @rgclayton mentioned, there should definitely be a hover indicator, whatever that may be from a Consonant perspective. This will also come up during an accessibility review, a visual indicator should exist between states.

ryanmparrish:
@aishwaryamathuria @rgclayton, I'm checking with the consonant team and will let you know exacts. My gut says we should use the var blue-hover color for the :hover style and keep the underline.

ryanmparrish:
@aishwaryamathuria  @rgclayton  - I spoke with the consonant team and they want to follow the same static link guidelines from spectrum. Static links have inherit text color, are underline and the only :hover style is the cursor is a pointer. So w/ that said this is GTG as is. LMK if you have any questions/concerns. Thanks

PR #1196, opened 2023-08-31
Merch Card block - UAR enhancement

ryanmparrish:
* Added the following variants to the **merch-card** block
-- `background-opacity-70` - This sets the card background white to 70% opacity
-- `inline-heading` - This allows the icon and header to display inline

* Added a 'decorate block hr' function to author `<hr>` color
-- This is authored w/ the following syntax `--- #ff0000` in the **merch-card** block content area
-- Note: The color value can be any css color or referenced placeholder value. ex `--- {{color-photoshop}}`

Screen Shot 2023-08-31 at 5 09 06 PM

Resolves: [MWPW-133315](https://jira.corp.adobe.com/browse/MWPW-133315)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rparrish/cards/special-offers?martech=off
- After: https://rparrish-merch-card-hr--milo--adobecom.hlx.page/drafts/rparrish/cards/special-offers?martech=off


libs/blocks/merch-card/merch-card.css
```diff
@@ -238,6 +238,34 @@
...
+.merch-card.has-divider hr {
+  margin: var(--spacing-xxs) 0;
+}
+
+.merch-card.has-divider .consonant-CardFooter hr {
+  display: none;
+}
+
+.merch-card.background-opacity-70 {
+  background-color: rgba(255 255 255 / 70%);
```
yesil, line 250: could this be a variable as well?

narcis-radu (approved):
Looks good, but I noticed some card placement issues in Safari which seem to have been introduced before this PR.

Axelcureno (approved):

VKniaz (approved):

libs/blocks/merch-card/merch-card.css
ryanmparrish: @yesil - This definitely could be a var. I'm open, but I think till we have a use case for other blocks to use this i'll likely keep it as is for now.

elan-tbx (approved):

PR #1198, opened 2023-09-01
Chrome devtools breaks on parsing $`

chrischrischris:
When a line contains a template literal that ends with a $:
`` `^${pattern}$` ``
Chrome devtools will show all the code after this literal as being part of a string literal:

Screenshot 2023-09-01 at 9 06 03 AM

Adding an end-of-line comment that ends in a backtick fixes it:

Screenshot 2023-09-01 at 9 07 29 AM

This is the only relevant chrome bug I could find:
https://bugs.chromium.org/p/chromium/issues/detail?id=659515

Resolves: N/A

**Test URLs:**
- After: https://devtoolssucks--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -136,7 +136,7 @@ function normalizePath(p) {
 
 const matchGlob = (searchStr, inputStr) => {
   const pattern = searchStr.replace(/\*\*/g, '.*');
-  const reg = new RegExp(`^${pattern}$`, 'i');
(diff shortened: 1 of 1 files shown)


overmyheadandbody (approved):

mokimo (approved):

narcis-radu:
Could not find other template literals ending with `$`, also could not reproduce with other characters.

sukamat (approved):

PR #1200, opened 2023-09-01
MWPW-135442 - Fixing the issue of NOT operator when the order of selection is changed

sabyamon:
- Gathering all the next steps for the selected cards first and then filtering on the entire set to make sure ordering of card selection does not make any difference when it comes to the `NOT` operator.

Resolves: https://jira.corp.adobe.com/browse/MWPW-135442

**Test URLs:**
- After: https://uar-skip-order-of-selection--milo--sabyamon.hlx.page/drafts/xiasun/quiz-2/?martech=off

### DIFF
--- a/libs/blocks/quiz/utils.js
+++ b/libs/blocks/quiz/utils.js
@@ -355,6 +355,7 @@ export const handleNext = (questionsData, selectedQuestion, userInputSelections,
   let nextQuizViews = [];
   let hasResultTigger = false;
   let lastStopValue;
+
   allcards.forEach((selection) => {
     // for each elem in current selection, find its coresponding
     // next element and push it to the next array.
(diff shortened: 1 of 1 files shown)


colloyd (approved):

JackySun9 (approved):

fullcolorcoder (approved):

echampio-at-adobe (approved):
Spying on userFlow in this method, the flow of events seems good now. It's no longer different, for example, if 3d is selected before or after another content type in the first question. 

  useEffect(() => {
    if (userFlow.length) {
      const currentFlow = userFlow.shift();
      if (!currentFlow.length) {
        return;
      }
      setSelectedQuestion(questionList[currentFlow] || []);
    }
  }, [userFlow, questionList]);

PR #1212, opened 2023-09-05
Mwpw 135924 fallback css property added to iframe

sharath-kannan:
* A fallback property for iframe is added as dvh unit is not supported in few devices.(ipad 7th gen)
*The fallback value is set to 90vh to resolve the initial [issue](https://jira.corp.adobe.com/browse/MWPW-135324).

Resolves: [MWPW-[135924](https://jira.corp.adobe.com/browse/MWPW-135924)]

Test URLs:
-before: https://stage--cc--adobecom.hlx.page/drafts/ruchika/discover1?martech=off
-After:https://stage--cc--adobecom.hlx.page/drafts/ruchika/discover1?martech=off&milolibs=MWPW-135924--milo--sharath-kannan#animate

### DIFF
--- a/libs/styles/iframe.css
+++ b/libs/styles/iframe.css
@@ -28,7 +28,7 @@
 
 @media (max-width: 1200px) {
   .dialog-modal.commerce-frame .milo-iframe {
-    padding-bottom: 100vh;
+    padding-bottom: 90vh;
     padding-bottom: 100dvh;
   }
 }
--- a/libs/styles/iframe.css
+++ b/libs/styles/iframe.css
@@ -28,6 +28,7 @@
 
 @media (max-width: 1200px) {
   .dialog-modal.commerce-frame .milo-iframe {
+    padding-bottom: 100vh;
     padding-bottom: 100dvh;
   }
 }
--- a/libs/styles/iframe.css
+++ b/libs/styles/iframe.css
@@ -28,6 +28,7 @@
 
 @media (max-width: 1200px) {
   .dialog-modal.commerce-frame .milo-iframe {
+    /* fallback to vh if dvh is not supported */
     padding-bottom: 90vh;
     padding-bottom: 100dvh;
   }


overmyheadandbody (changes requested):
Please add different links in the 'before' and 'after' sections and make sure that the change from the PR is applied to the 'after' link. The link from the 'test' section of this PR also doesn't have the CSS update applied. Also, please ensure that you're using the appropriate labels when opening PRs, details are available at https://github.com/adobecom/milo/wiki/PR-Labels

sharath-kannan:
> Please add different links in the 'before' and 'after' sections and make sure that the change from the PR is applied to the 'after' link. The link from the 'test' section of this PR also doesn't have the CSS update applied. Also, please ensure that you're using the appropriate labels when opening PRs, details are available at https://github.com/adobecom/milo/wiki/PR-Labels

This is the required format of the test urls.
- https://mwpw-135924--milo--sharath-kannan.hlx.page/
I have added it in a similar format to clear the helix-spi-check.
I have update the test urls with : https://stage--cc--adobecom.hlx.page this format.

overmyheadandbody:
The initial links added to the PR were:
> -before: https://main--milo--adobecom.hlx.page/drafts/tat28144/discover?martech=off
> -After: https://main--milo--adobecom.hlx.page/drafts/tat28144/discover?martech=off
which were both referencing the `main` branch. The 'after' one should contain the code proposed by the PR. You can keep both Milo and CC-specific links for the purposes of PSI, but the 'after' links should always include the code from the PR.

sharath-kannan:
> The initial links added to the PR were:
> 
> > -before: https://main--milo--adobecom.hlx.page/drafts/tat28144/discover?martech=off
> > -After: https://main--milo--adobecom.hlx.page/drafts/tat28144/discover?martech=off
> 
> which were both referencing the `main` branch. The 'after' one should contain the code proposed by the PR. You can keep both Milo and CC-specific links for the purposes of PSI, but the 'after' links should always include the code from the PR.

okay , Got it.
Are you able to verify it now?

libs/styles/iframe.css
```diff
@@ -28,6 +28,7 @@
 
 @media (max-width: 1200px) {
   .dialog-modal.commerce-frame .milo-iframe {
+    padding-bottom: 90vh;
```
narcis-radu, line 31: Wouldn't this get overridden by the next line?

overmyheadandbody: It's used as a fallback in case the value from the next line can not be applied, which seems to be the case for older iPads and other devices/browsers. At least that's my understanding

sharath-kannan: Yes,That's correct.
The dvh unit is not supported by ipad 7th gen in my case and the fallback scenario has been tested in the same device.

narcis-radu: got it, thank you. It would be great to add a small comment for this use-case.

sharath-kannan: I have added the comment.

overmyheadandbody (approved):
This should be tested along with https://github.com/adobecom/milo/pull/1215, since it touches on the same logic

overmyheadandbody:
I would argue that Milo links should still be added, since this touches on Milo functionality, to ensure PSI checks run and the change isn't affecting other areas of the website.

PR #1219, opened 2023-09-05
MWPW-135296 - Quiz css and related markup cleanup

colloyd:
* Drastically simplifies quiz css, removing unused css
* Simplified markup, removing unused classes and standardizing the rest
* Remove caas related css and classes that were creating linting errors
* add icon and cover-image types to cards
* add top and bottom classes to progress indicators
* set all styling to match the comps from consonant for all devices

Resolves: [MWPW-135296](https://jira.corp.adobe.com/browse/MWPW-135296)

**Test URLs:**
- Before: https://uar-integration--milo--adobecom.hlx.page/drafts/delta/app-recommender-test/?martech=off
- After: https://mwpw-135296-quiz-css-cleanup--milo--colloyd.hlx.page/drafts/delta/app-recommender-test/?martech=off


colloyd:
Just some stats on what these changes have done for quiz.css specifically: 

- From 980 lines to 385 lines.
- From 19k file size to 7k file size

libs/blocks/quiz/quizcontainer.js
```diff
@@ -9,12 +9,12 @@ export const DecorateBlockBackgroundCmp = ({ background = '' }) => html`<div dat
       </picture>
   </div>`;
 
-export const DecorateBlockForeground = ({ heading, subhead }) => html`<div class="foreground container">
-    <div data-valign="middle" class="text" daa-lh="${heading}">
-      <h2 id="heading-xl-marquee-standard-medium-left" class="heading-XL">
+export const DecorateBlockForeground = ({ heading, subhead }) => html`<div class="quiz-foreground">
+    <div class="text" daa-lh="${heading}">
```
overmyheadandbody, line 13: I see this class name was there before, but `.text` sounds a bit generic, it might lead to issues when other blocks are present on the page

libs/blocks/quiz/quizoption.js
```diff
@@ -3,25 +3,23 @@ import { html } from '../../deps/htm-preact.js';
 export const OptionCard = ({
   text, title, coverImage, cardIcon, options, disabled, selected,
 }) => {
-  const cardIconHtml = html`<div class="consonant-OneHalfCard-image">
-    <img src="${cardIcon}" alt="Video" class="icon" />
+  const cardIconHtml = html`<div class="quiz-option-icon">
+    <img src="${cardIcon}" alt="Quiz Option Icon" class="icon" />
```
overmyheadandbody, line 7: Same generic class question, `.icon` might have conflicts with other blocks

libs/blocks/quiz/quiz.css
```diff
@@ -1,980 +1,384 @@
-.quiz { 
-  min-height: 1000px;
-  display: flex;
-  align-items: center;
-  flex-direction: column;
+.quiz-page * {
```
overmyheadandbody, line 1: `*` does not account for `:before` and `:after` pseudoelements, do they need to be included?

libs/blocks/quiz/quiz.css
```diff
@@ -1,980 +1,384 @@
...
 }
 
-.quiz-container {
-  display: flex;
-  align-items: center;
-  justify-content: flex-start;
-  flex-direction: column;
-  color: var(--color-white);
-  width: 100%;
+div[data-valign="middle"] {
```
overmyheadandbody, line 9: Should this not be scoped to the quiz logic?

libs/blocks/quiz/quiz.css
```diff
@@ -1,980 +1,384 @@
...
-  margin-top: var(--spacing-m);
-  margin-bottom: 0;
-}
-
-.quiz-container.light,
-.quiz-container.quiet,
-.quiz-container.inline {
-  color: var(--text-color);
+.quiz {
+  min-height: calc(100vh - 131px);
```
overmyheadandbody, line 18: `131px` sounds a bit arbitrary. Could you add a comment or replace the value with a variable that has an appropriate name?

libs/blocks/quiz/quiz.css
```diff
@@ -1,980 +1,384 @@
...
-
-.quiz-container .background.has-video {
-  bottom: 0;
-  left: 0;
-  overflow: hidden;
-  position: absolute;
-  right: 0;
-  top: 0;
+.quiz-background {
+  z-index: 0;
```
overmyheadandbody, line 29: Is this actually needed, since 0 should be similar to the default value?

libs/blocks/quiz/quiz.css
```diff
@@ -1,980 +1,384 @@
...
-  min-width: 100%;
-  max-width: 100%;
-  margin: 0 0 16px;
-}
-
-.quiz-container .consonant-CardsGrid--5up .card div div {
-  text-align: right;
+.quiz-option.selected .quiz-option-title {
+  color: var(--link-hover-color-dark);
+  font-weight: bold;
```
overmyheadandbody, line 135: Earlier I saw `900` was used as the value for a `font-weight`. Maybe it would be a good idea to follow a pattern of just using number values?

libs/blocks/quiz/quiz.css
```diff
@@ -1,980 +1,384 @@
...
-
-.dot-indicators:last-of-type {
   display: flex;
-  visibility: visible;
+  justify-content: space-between;
+  width: 300px;
+  z-index: 15;
 }
 
 .dot {
```
overmyheadandbody, line 199: I'd argue that `.dot` might be a bit too generic, maybe it could be scoped to the quiz block

libs/blocks/quiz/quiz.css
```diff
@@ -1,980 +1,384 @@
...
+  margin-left: 20px;
   position: absolute;
   top: 6px;
-  margin-left: 20px;
-}
-
-/* Visited lines */
-.dot.prev:not(.current)::after {
-  border-bottom: 2px solid #676767;
+  width: calc(100% - 23px);
```
overmyheadandbody, line 221: Again, it might make sense to add a comment or switch to a relevant variable name for the `23px` value

libs/blocks/quiz/quiz.css
```diff
@@ -1,980 +1,384 @@
...
 
-.dot:not(.prev).current::before {
+.dot.current::before {
   background-color: #000;
   border-color: #000;
 }
 
-.dot:not(.current).prev::before {
+.dot.prev::before {
   background-color: #676767;
```
overmyheadandbody, line 230: `#676767` is used a few times, maybe it could be turned into a variable

libs/blocks/quiz/quiz.css
```diff
@@ -1,980 +1,384 @@
...
-  }
+.dot-indicators.bottom .dot.current::before {
+  background-color: var(--color-white);
+  border: 2px solid var(--color-white);
+}
 
-  .quiz-container.split .button-justified-mobile {
-    width: auto;
-  }
+.dot-indicators.bottom .dot.prev::before {
```
overmyheadandbody, line 266: Seems to me that the styles are similar to the ones for `.dot-indicators.bottom .dot.current::before`, maybe they can be grouped together

libs/blocks/quiz/quiz.css
```diff
@@ -1,980 +1,384 @@
...
-    padding-bottom: 0;
-  }
-
-  .milo-card-wrapper {
-    padding: 0;
-  }
-
-  .quiz-container .card:not(.quiz-container .card.selected) {
+  .quiz-option {
+    align-items: stretch ;
```
overmyheadandbody, line 308: Extra space before semicolon can be removed

overmyheadandbody (commented):
Always good to see technical debt and/or improvements being brought in 🙂 It's a rather large PR to follow, but I did spot a few things that could be improved.

libs/blocks/quiz/quizoption.js
```diff
@@ -3,25 +3,23 @@ import { html } from '../../deps/htm-preact.js';
...
-  return html`<div class="card half-card border consonant-Card consonant-OneHalfCard quiz-option ${disabled} ${selected}" 
-                    data-card-name="${options}">
-        <div class="consonant-OneHalfCard-inner">
-          ${cardIcon && cardIconHtml}
-          ${coverImage && coverImageHtml}
-        <div data-valign="middle">  
-          <h3 class="consonant-OneHalfCard-title">${title}</h3>
-          <p class="consonant-OneHalfCard-text">${text}</p>
+  return html`<div class="quiz-option ${disabled}${selected}" data-option-name="${options}" 
+        data-option-type="${cardIcon ? 'icon' : ''}${coverImage ? 'cover-image' : ''}">
```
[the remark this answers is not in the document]
colloyd: Image and icon are exclusive and there should never be both, so the lack of space here is deliberate to avoid an odd trailing space in this property

libs/blocks/quiz/quiz.css
```diff
@@ -1,980 +1,384 @@
...
-  align-items: center;
-  flex-direction: column;
+.quiz-page * {
+  box-sizing: border-box;
 }
 
-.quiz > div:not([class]) {
-  display:none;
+.quiz-page .feds-topnav-wrapper {
+  padding: 0 20px;
```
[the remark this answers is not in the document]
colloyd: This was a hold over from the old stuff that I persisted. It's there because without it, our customized nav has the logo and sign-in hugging the edge uncomfortably. If you think this should be removed because it could interfere with gnav in some way, I can certainly do it

echampio-at-adobe:
Hi @colloyd. I notice the card background images can overlap the card boundaries. You can see the problem on the video options screen because of the contrast between the different images and the page background. 
https://mwpw-135296-quiz-css-cleanup--milo--colloyd.hlx.page/drafts/xiasun/quiz-2/?q-category=video&q-rather=custom
image

libs/blocks/quiz/quiz.css
colloyd: No, this is not needed for pseudoelements, which only appear on the progress indicators.

libs/blocks/quiz/quiz.css
colloyd: Completely unnecessary, as it turns out. It was only used in the background > div and it didn't actually apply. Just some old stuff that I had missed, good catch.

libs/blocks/quiz/quizoption.js
colloyd: no need for this class at all.

libs/blocks/quiz/quizcontainer.js
colloyd: No need for this class, or the wrapping div, for that matter. I've moved the daa-lh to the h3.

libs/blocks/quiz/quiz.css
colloyd: I'll add a comment, but this is basically 100vh - gnav and gfooter heights to ensure that we avoid odd whitespace below gfooter. (see the before page with a tall browser window to see what I'm talking about here).

libs/blocks/quiz/quiz.css
colloyd: Nope! good catdh

libs/blocks/quiz/quiz.css
colloyd: added  a comment on this, it's basically the total width of .dot minus the width of the next indicator dot to prevent the dashed lines from spilling out over the next dot.

libs/blocks/quiz/quiz.css
colloyd: Adjusted to use --color-gray-600 and --color-black for the #000.

libs/blocks/quiz/quiz.css
colloyd: renamed .dot-indicators to quiz-step-container and .dot to .quiz-step. This should guarantee that we don't have any potential collisions

colloyd:
@echampio-at-adobe added `border-radius: 0.5rem 0.5rem 0 0;` to the images to address how they were spilling out. Nice catch.

echampio-at-adobe:
> @echampio-at-adobe added `border-radius: 0.5rem 0.5rem 0 0;` to the images to address how they were spilling out. Nice catch.

Looks great now. Thanks!

echampio-at-adobe (approved):

JackySun9 (approved):

JackySun9:
Great job!!!

libs/blocks/quiz/quiz.css
overmyheadandbody: It does work right now, but it might not be an ideal solution long-term. Once we have site-wide alerts on top of the navigation, this calculation will fail. Footers for Korea and China can span multiple rows (even the limited footer present on this page) due to legal requirements, so this approach will not yield expected results.

This shouldn't be blocker for the purposes of this CSS refactoring PR, but you might want to open a follow-up/technical debt story to investigate a better approach to this.

overmyheadandbody (commented):
Thanks for integrating the changes, this is looking much better!

I see unit tests are currently failing for this feature, will this be addressed by another commit to this PR or as a separate issue? I wouldn't normally ask, but I see this goes to a feature branch, not `main`, so you could potentially open a new PR for unit tests without affecting the checks from the `main` branch.

libs/blocks/quiz/quiz.css
echampio-at-adobe: Thanks @overmyheadandbody. I create this story for later: https://jira.corp.adobe.com/browse/MWPW-136135

PR #1222, opened 2023-09-06
MWPW-135967: card container to support fragments

3ch023:
When MEP is used for Cards, they are inserted into the Cards Container as Fragments.
This code change is to fix the cards container not rendering fragments correctly, resulting in cards displayed in the wrong order/position.

Resolves: [MWPW-135967](https://jira.corp.adobe.com/browse/MWPW-135967)
Test page: [Special Offers](https://adobe.sharepoint.com/:w:/r/sites/adobecom/_layouts/15/Doc.aspx?sourcedoc=%7BB56B2688-2141-4C5E-AF47-C4DBC79E581E%7D&file=special-offers.docx&action=default&mobileredirect=true)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/mariia/promodemo/special-offers/special-offers?martech=off
- After: https://mwpw-135967--milo--adobecom.hlx.page/drafts/mariia/promodemo/special-offers/special-offers?martech=off

Before:
image
After (Card positioned as 2nd, as dictated by the MEP manifest):
image


libs/blocks/card/card.js
```diff
@@ -57,8 +57,6 @@ const init = (el) => {
   const base = miloLibs || codeRoot;
   loadStyle(`${base}/deps/caas.css`);
 
-  const section = el.closest('.section');
-  section.classList.add('milo-card-section');
```
yesil, line 61: way the class `milo-card-section` needs to be removed?

libs/blocks/card/cardUtils.js
```diff
@@ -36,10 +36,17 @@ export const addFooter = (links, container, merch) => {
   });
 };
 
-export const addWrapper = (el, section, cardType) => {
+export const addWrapper = (el, cardType) => {
+  const fragment = el.closest('.fragment');
+  if (fragment) {
```
yesil, line 41: suggestion to combine `fragment` and `fragmentSection` selectors.
`const fragmentSection = el.closest('.fragment')?.closest('.section');`

overmyheadandbody: Why not just `.closest('.section .fragment')`?

libs/blocks/card/card.js
3ch023: @VKniaz and @Axelcureno might be able to clarify, I just removed code duplication, not sure what is it needed for.

libs/blocks/card/cardUtils.js
3ch023: i need fragment too, to set a flex display on it 
should I change it to something like
```
el.closest('.fragment')?.style.display = 'flex';
el.closest('.fragment .section')?.replaceWith(el);
```
?

it will be shorter but el.closest('.fragment') is then computed twice.

yesil: @3ch023 if you need `fragment` as well keep it, I thought it wasn't used.

yesil (approved):

libs/blocks/card/cardUtils.js
overmyheadandbody: Yeah, I just reacted to @yesil's comment, but if you need to set properties on both, then it should be fine. And yeah, order was not correct, just highlighting the other possible approach.

libs/blocks/card/cardUtils.js
```diff
@@ -36,10 +36,17 @@ export const addFooter = (links, container, merch) => {
...
+    const fragmentSection = el.closest('.section');
+    fragmentSection?.replaceWith(el);
+    fragment.style.display = 'flex';
```
overmyheadandbody, line 44: Could this be achieved via a CSS modifier class? Inline styles are a bit more difficult to maintain in the long run and it makes sense to have styles grouped together

libs/blocks/card/cardUtils.js
```diff
@@ -36,10 +36,17 @@ export const addFooter = (links, container, merch) => {
   });
 };
 
-export const addWrapper = (el, section, cardType) => {
+export const addWrapper = (el, cardType) => {
+  const fragment = el.closest('.fragment');
+  if (fragment) {
+    const fragmentSection = el.closest('.section');
```
overmyheadandbody, line 42: This is called `fragmentSection`, but you're looking for the closest `.section` ancestor of the `el` element. Should `el` be switched to `fragment` or should the variable name be updated to `elSection` or whatever fits best? In the unit tests, it seems that `fragment` is used in this scenario, not `el`.

This might also be helpful in case the structure is sometimes `<fragment><section>` and other times `<section><fragment>` or other combinations of these elements.

test/blocks/merch-card/merch-card.test.js
```diff
@@ -60,6 +60,21 @@ describe('Merch Card', () => {
...
   });
 
+  it('Supports Card as a Fragment', async () => {
+    document.body.innerHTML = await readFile({ path: './mocks/special-offers.html' });
+    await init(document.querySelector('.special-offers-black-friday'));
+    const fragment = document.querySelector('.special-offers-fragment');
+    const fragmentSection = fragment.querySelector('.section');
+    const card = fragment.querySelector('.merch-card');
+
+    expect(fragment).to.be.exist;
```
overmyheadandbody, line 70: Wouldn't `to.exist` be enough?

test/blocks/merch-card/merch-card.test.js
```diff
@@ -60,6 +60,21 @@ describe('Merch Card', () => {
...
+    expect(fragment.parentElement.children.length).to.be.equal(3);
```
overmyheadandbody, line 71: You could also use `childElementCount` on the parent. And I think `.be` can be removed (valid for the next few checks as well)

VKniaz (approved):

libs/blocks/card/card.js
Axelcureno: milo-card-section class is used by consonant styles and should not be removed. I see @3ch023 is just moving this changes to addWrapper in cardUtils.js and updating the usage in card.js so I'm good with this 👍🏼

Axelcureno:
@3ch023 I created [this page with your changes](https://mwpw-135967--milo--adobecom.hlx.page/drafts/axel/merch-card-fragments-test) to test the use case of fragments that contain merch cards and I noticed that in this example while the fragment contains three cards, only one is displayed on the page. 
I also noticed your before/after link doesn't contain any fragments (I might be missing a step to test this with personalization).

libs/blocks/card/cardUtils.js
3ch023: the structure is like this (card being the element):
```
<section>
   <fragment>
      <section>
         <card>
      <section>
   <fragment>
<section>
```

I agree 'elSection' is more clear, will rename, thx

libs/blocks/card/cardUtils.js
overmyheadandbody: Any chance this can still be moved to a modifier class instead of an inline style?

libs/blocks/fragment/fragment.css
```diff
@@ -0,0 +1,3 @@
+.fragment-flex {
+  display: flex;
+}
```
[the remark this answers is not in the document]
3ch023: trueeee, addressed!

Axelcureno (approved):

Axelcureno:
> No further concerns, but I noticed that CTAs are not aligned across cards. I think there was a recent PR trying to tackle this. Is this behaviour expected for now? Screenshot 2023-09-13 at 14 44 03

Correct, that is being fixed [here](https://github.com/adobecom/milo/pull/1241)

PR #1227, opened 2023-09-06
MWPW-136001 fix for safari

VKniaz:
Fix the Safari race condition issue when trying to mutate section classList

Resolves: [MWPW-136001](https://jira.corp.adobe.com/browse/MWPW-136001)

**Test URLs:**

- Before: https://main--milo--adobecom.hlx.page/drafts/vkniaziev/document6?martech=off
- After: https://mwpw-136001--milo--adobecom.hlx.page/drafts/vkniaziev/document6?martech=off

Before:
Screenshot 2023-09-06 at 1 10 39 PM
After:
Screenshot 2023-09-06 at 1 10 57 PM


libs/blocks/card/cardUtils.js
```diff
@@ -35,8 +35,19 @@ export const addFooter = (links, container, merch) => {
     if (parentElement && document.body.contains(parentElement)) parentElement.remove();
   });
 };
-
 export const addWrapper = (el, section, cardType) => {
+  const observeClassListChanges = (classToRemove) => {
+    const callback = (mutationsList) => {
+      for (const mutation of mutationsList) {
+        const { classList } = section;
+        if (mutation.type === 'attributes' && mutation.attributeName === 'class' && classList.contains(classToRemove)) classList.remove(classToRemove);
```
vladen, line 43: `mutation.type === 'attributes'` is implied in `attributes: true` below and might be redundant

mutation.attributeName === 'class'` condition is implied in `attributeFilter: ['class']` and might be redundant

`classList.remove(classToRemove)` does nothing if CSS class being removed is not present in the list, `classList.contains(classToRemove)` condition can be omitted

libs/blocks/card/cardUtils.js
```diff
@@ -35,8 +35,19 @@ export const addFooter = (links, container, merch) => {
     if (parentElement && document.body.contains(parentElement)) parentElement.remove();
   });
 };
-
 export const addWrapper = (el, section, cardType) => {
+  const observeClassListChanges = (classToRemove) => {
```
vladen, line 39: this function is called from exactly one place in the code below and can be inlined for simplicity

vladen (approved):

libs/blocks/card/cardUtils.js
VKniaz: classList.contains(classToRemove) condition is added not to have an infinite loop.

libs/blocks/card/cardUtils.js
VKniaz: inlined

overmyheadandbody (approved):

libs/blocks/card/cardUtils.js
```diff
@@ -35,7 +35,6 @@ export const addFooter = (links, container, merch) => {
     if (parentElement && document.body.contains(parentElement)) parentElement.remove();
   });
 };
-
```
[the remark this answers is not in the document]
VKniaz: added

vladen (approved):

libs/blocks/card/cardUtils.js
```diff
@@ -48,7 +48,8 @@ export const addWrapper = (el, section, cardType) => {
   const idx = list.findIndex((i) => i.includes(upClass));
   if (idx > -1) {
     upClass = `${idx + 2}-up`;
-    section.classList.remove(list[idx]);
+    const classToRemove = list[idx];
+    new MutationObserver(() => { if (section.classList.contains(classToRemove)) section.classList.remove(classToRemove); }).observe(section, { attributes: true, attributeFilter: ['class'] });
```
yesil, line 52: this looks like hacking the section block. It feels this must be addressed in the section block it-self.
accessing a class name by index and removing it might unexpected issues in the future when the section block code changes.

yesil: given that, the section is a grid, and you have the following content tree, shouldn't you use `display: contents` on `milo-card-wrapper, consonant-Wrapper-inner, consonant-Wrapper-collection, consonant-CardsGrid` elements
image

yesil (changes requested):
MutationObserver looks overkill for this, let's discuss alternative options.

yesil (commented):
We have discussed with @VKniaz who will track discussed refactoring in a MWPW ticket.

yesil (approved):

ryanmparrish:
@VKniaz - Looking good. With some of the concerns brought up by @yesil around the mutationObserver and along the same lines suggested, it might be worth revisiting a simple css approach. What do you think?
```
.milo-card-section.two-up,
.milo-card-section.three-up,
.milo-card-section.four-up,
.milo-card-section.five-up {
  grid-template-columns: none;
  padding-left: 0;
  padding-right: 0;
}

File: libs/blocks/card/cardUtils.js [hunk 5]
```diff
@@ -48,7 +47,8 @@ export const addWrapper = (el, section, cardType) => {
...
```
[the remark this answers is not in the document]
VKniaz: There are no warnings from ESLint.

VKniaz:
> @VKniaz - Looking good. With some of the concerns brought up by @yesil around the mutationObserver and along the same lines suggested, it might be worth revisiting a simple css approach. What do you think?
> 
> ```
> .milo-card-section.two-up,
> .milo-card-section.three-up,
> .milo-card-section.four-up,
> .milo-card-section.five-up {
>   grid-template-columns: none;
>   padding-left: 0;
>   padding-right: 0;
> }
> ```

From my point of view, it's better to have the logic with Observer. In this case, we have one place that takes care of style adjustment. The approach of resetting CSS is more complicated in terms of maintenance. 
Illyas proposes to adjust the logic for section-metadata parameters. An appropriate Jira ticket will be initiated.

PR #1228, opened 2023-09-06
{viewpoint}-only-{theme} feature for marquee

seanchoi-dev:
{viewpoint}-only-{theme} feature for marquee.

Resolves: [MWPW-135840](https://jira.corp.adobe.com/browse/MWPW-135840)

This will allow authors to have different themes (light or dark) for each veiwpoint (mobile, tablet, desktop) by simply adding theme text in the background cell

Example:
image

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/seanchoi/marquees?martech=off
- After: https://marquee-update--milo--adobecom.hlx.page/drafts/seanchoi/marquees?martech=off

*I also checked library to make sure that this doesn't break existing blocks
(https://marquee-update--milo--adobecom.hlx.page/docs/library/blocks/marquee)

### DIFF
--- a/libs/blocks/marquee/marquee.css
+++ b/libs/blocks/marquee/marquee.css
@@ -16,8 +16,12 @@
   margin-bottom: 0;
 }
 
-.marquee.light, .marquee.quiet, .marquee.inline {
+.marquee.light, .marquee.mobile-only-light, .marquee.quiet, .marquee.inline {
   color: var(--text-color);
+}
+
+.marquee.mobile-only-dark {
+  color: #FFF;
 }
 
 .marquee .foreground {
@@ -159,6 +163,10 @@
   background-color: black;
 }
 
+.marquee.large.mobile-only-light {
+  background-color: initial;
+}
+
 .marquee.large .text {
   display: flex;
   flex-direction: column;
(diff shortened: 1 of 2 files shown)


auniverseaway:
I _think_ this may be incompatible with background positioning.

seanchoi-dev:
> positioning

Can you think of an example?

vgoodric (approved):

meganthecoder (approved):

ryanmparrish:
Just thinking thru the logic w/ mobile first and the marquee defaults. A marquee is .dark by default unlike other blocks.
So with the variant structure above... 
`large, desktop dark, mobile light, tablet light`
This works, but is kind of seems more natural to have it be like this, no?
`large, light, desktop dark`

The 'mobile-light' cascades up, but we already have 'light' that does the same, so is that even necessary?

seanchoi-dev:
> 
@ryanmparrish @vgoodric , Though I agree with you but I added those mobile-* lines at last for two reason:
1. there are existing issue with main branch that there is weird behavior when there are two viewpoints (mobile, tablet) and with those code it will prevent any potential issue completely if author simply add more specific variant.
 you can see more details from this PR's Ryan C's comment: https://github.com/adobecom/milo/pull/1162#discussion_r1316307823
 
 2. We are familiar with mobile first concept but authors and some some engineers are not. I think it is lower cost to add few lines than train all people who would use Marquee branch. I just don't want to see slack question or ticket that saying mobile-* isn't working then someone goes like "You can just put default theme for mobile".

libs/blocks/marquee/marquee.css
```diff
@@ -296,6 +304,22 @@
...
+  .marquee.large.tablet-light {
+    background-color: initial;
+  }
+
+  .marquee.tablet-light {
+    color: var(--text-color);
+  }
+
+  .marquee.tablet-dark {
+    color: #FFF;
```
overmyheadandbody, line 320: I see hex, variables and named color values are used interchangeably, it might not be a bad idea to standardise and stick to a rule. If an update is needed, it will be easier to check for just `#fff`, `var(--whatever)` or `white` instead of all of them.

PR #1241, opened 2023-09-08
Mwpw-135989 - Special Offers Merch Card styling fixes 

Axelcureno:
Fixes Special Offers Merch card variant styles to match Consonant specs.

Screenshot 2023-09-08 at 10 53 53 AM

This PR fixes the following:

- Introduces `full-width` property to expand card to 3-row width.
- Introduces `center` property to center align text inside the card.
- Content alignment across rows. Top aligned: Text, icons, body copy. Bottom aligned: Footer (CTAs and `hr`)

Resolves: [MWPW-135989](https://jira.corp.adobe.com/browse/MWPW-135989)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/axel/special-offers-merch-card-fixes?martech=off
- After: https://mwpw-135989--milo--adobecom.hlx.page/drafts/axel/special-offers-merch-card-fixes?martech=off


vladen (approved):

libs/blocks/merch-card/merch-card.css
```diff
@@ -103,6 +111,16 @@
   border-radius: 16px;
 }
 
+.merch-card.special-offers.full-width {
+  min-width: 100%;
+  max-width: 100%;
+  width: 1200px;
```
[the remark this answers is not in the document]
Axelcureno: Updated to use var of full container width intead

libs/blocks/merch-card/merch-card.css
```diff
@@ -103,6 +111,16 @@
   border-radius: 16px;
 }
 
+.merch-card.special-offers.full-width {
```
ryanmparrish, line 114: Just a thought.
Is there any reason to scope these too .special-offers / would other types benefit from having them? 
```suggestion
.merch-card.full-width {
```
Below desktop VP the `--grid-container-width` is 83%. Is that the intended view for tablet? I see this goes full-width on mobile and desktop but the tablet does not? In the jira ticket it says to be a section property. Is this right. There is a .con-block.full-width style that we use for our ups grid override, but I know the cards do their own thing. 
> Use/develop “full width” instead of “one-up” for full size cards (this will be a section property)

libs/blocks/merch-card/merch-card.css
```diff
@@ -103,6 +111,16 @@
...
+  min-width: 100%;
+  max-width: 100%;
+  width: var(--grid-container-width);
+}
+
+.merch-card.special-offers.center {
```
ryanmparrish, line 120:
```suggestion
.merch-card.center {
```
if we want this on other card types? 🤷

ryanmparrish: Any author-able alignment considerations for the cta row? if we we might need a few more flex align props.

libs/blocks/merch-card/merch-card.css
Axelcureno: -As per Consonant guide, only the Special Offers card will use `full-width`. 
-No, 83% is not intended for tablet, this is a good catch. If you think it's ok, I can use 1200px hardcoded which is the default for containers in Milo. I have tested with this value and the card adjusts accordingly as expected in mobile/tablet.
-About this being a section metadata property, I have asked Fred (GWP) to update the description since we have agreed to keep it as a block property instead to avoid impacting other blocks.

libs/blocks/merch-card/merch-card.css
Axelcureno: -Only Special Offers Card was requested to have this property. 
-CTA row alignment is not in scope as per Consonant XD style guide so far. We can always introduce the change if there is a requirement.

libs/blocks/merch-card/merch-card.css
Axelcureno: ticket description has been updated, full-width is expected to be a block property instead of section-metadata

libs/blocks/merch-card/merch-card.css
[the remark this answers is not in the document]
Axelcureno: changed back to 1200px because it's the default width for containers in Milo. Tested in mobile/tablet and working as expected.

VKniaz (approved):

ryanmparrish (approved):

PR #1242, opened 2023-09-08
MWPW-135676 Inline Fragments

chrischrischris:
This introduces "inline fragments" aka "unwrapped fragments".

Currently, fragment links are not supported inside of blocks due to how they are inserted into the DOM - A fragment is wrapped with multiple divs and is decorated upon insertion.

The new inline fragment inserts just the content of the fragment into the block, as if it was part of the original block.

When a fragment inside a block is detected, the fragment will be loaded and inserted into the block DOM.  The block decoration will then happen after this.

In order to support this, autoblocks are now all decorated before the block itself is decorated.  Previous decoration of blocks and autoblocks within them would happen async, but this leads to race conditions that the blocks have to account for (see video in marquee and aside).  Think of this as decorating a block from the inside -> out.

There will be followup PRs to remove the race condition code from some of the blocks since the load order is now deterministic.

To recap decorating order:

1. Fragment link inside block: fragment is loaded and inserted
2. All other autoblock links inside the block are decorated
3. Block is decorated

Inline fragments ONLY support a single section in the fragment (or no section at all).  If there are multiple sections in the fragment, only the first one will be inserted.  The idea is that these fragments will be small things like buttons or text that need to be changed within the block.

Inline fragments can be manually used by appending `#_inline` to the fragment url, but this is NOT to be something that authors are to use.  Internally the code appends `#_inline` to all fragment links inside a block.  It can also be used for dev testing, and there may be uses for it in the future though we have not come up with any use cases.

NOTE: I also refactored a bunch of the code, which leads to a large diff.  I recommend reviewing with whitespace ignored though there is still a lot of code just moved.

The sample page has fragment content inserted into the marquee.  First one is using a fragment link in the block, 2nd one is using MEP to swap out the CTA:

Resolves: [MWPW-135676](https://jira.corp.adobe.com/browse/MWPW-135676)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/cpeyer/fragment-test/marquee-with-frag?martech=off
- After: https://inlinefrag--milo--adobecom.hlx.page/drafts/cpeyer/fragment-test/marquee-with-frag?martech=off


libs/utils/utils.js
```diff
@@ -27,7 +27,6 @@ const MILO_BLOCKS = [
   'featured-article',
   'figure',
   'fragment',
-  'fragment-personalization',
```
chrischrischris, line 30: This was leftover from functionality that has been removed.

libs/utils/utils.js
```diff
@@ -91,6 +90,11 @@ const AUTO_BLOCKS = [
   { merch: '/tools/ost?' },
   { 'offer-preview': '/tools/commerce' },
 ];
+const DO_NOT_INLINE = [
```
chrischrischris, line 93: Open to better naming suggestions

libs/utils/utils.js
```diff
@@ -179,6 +183,9 @@ export const [setConfig, updateConfig, getConfig] = (() => {
       config.base = config.miloLibs || config.codeRoot;
       config.locale = pathname ? getLocale(conf.locales, pathname) : getLocale(conf.locales);
       config.autoBlocks = conf.autoBlocks ? [...AUTO_BLOCKS, ...conf.autoBlocks] : AUTO_BLOCKS;
+      config.doNotInline = conf.doNotInline
+        ? [...DO_NOT_INLINE, ...conf.doNotInline]
```
overmyheadandbody, line 187: Couldn't this lead to duplicates in the resulting array? Maybe something like `config.doNotInline = [...new Set([...DO_NOT_INLINE, ...conf.doNotInline || []])]` might do the trick?

overmyheadandbody (commented):
No big issues spotted so far, but it is a rather large PR, so I'll have another go whenever I get some time 🙂

echampio-at-adobe (commented):
Hi @chrischrischris. I'm sure it's not a surprise to you, but just to confirm, I copied your updates to fragments.js and utils.js into a copy of the uar-integration branch and the code that @colloyd wrote to dynamically load fragments in the quiz-results block still seems to work as it should.

vgoodric (approved):
Did some cursory QA and it is working as I'd expect.

libs/utils/utils.js
chrischrischris: Consumers will never have blocks that have the same name as a Milo block since the Milo block will take precedence and the consumer block will be ignored.

narcis-radu (approved):

narcis-radu (approved):

libs/utils/utils.js
```diff
@@ -655,17 +666,47 @@ async function loadFooter() {
...
+function decorateSection(section, idx) {
+  let links = decorateLinks(section);
+  decorateDefaults(section);
+  const blocks = section.querySelectorAll(':scope > div[class]:not(.content)');
+
+  const { doNotInline } = getConfig();
+  const blockLinks = [...blocks].reduce((blkLinks, block) => {
+    const blockName = block.classList[0];
+    links.forEach((link) => {
+      if (!block.contains(link)) return;
```
vhargrave, line 678: nit: 
I'd recommend filtering first so you don't need that `if` in the beginning -> links.filter((link) => 
`links.filter((link) => block.contains(link)).forEach((link) => { ...`

just seems a little nicer

libs/utils/utils.js
```diff
@@ -655,17 +666,47 @@ async function loadFooter() {
...
+
+      if (link.classList.contains('fragment')
+        && MILO_BLOCKS.includes(blockName) // do not inline consumer blocks (for now)
+        && !doNotInline.includes(blockName)) {
+        link.href = `${link.href}#_inline`;
```
vhargrave, line 683: not sure how much it matters, but you might already have #_inline on some inline fragment links here.

libs/utils/utils.js
```diff
@@ -863,6 +904,84 @@ function decorateMeta() {
...
+    config.experiment.selectedVariant.scripts.forEach((script) => loadScript(script));
+  }
+  initSidekick();
+
+  const { default: delayed } = await import('../scripts/delayed.js');
+  delayed([getConfig, getMetadata, loadScript, loadStyle, loadIms]);
+}
+
+async function processSection(section, config, isDoc) {
+  const inlineFrags = [...section.el.querySelectorAll('a[href*="#_inline"]')];
```
vhargrave, line 951: could we not return the inlineFrags as part of the return object in `decorateSection` so we don't need to select them here again / hit the DOM?

libs/utils/utils.js
```diff
@@ -863,6 +904,84 @@ function decorateMeta() {
...
+  if (inlineFrags.length) {
+    const { default: loadInlineFrags } = await import('../blocks/fragment/fragment.js');
+    const fragPromises = inlineFrags.map((link) => loadInlineFrags(link));
+    await Promise.all(fragPromises);
+    const newlyDecoratedSection = decorateSection(section.el, section.idx);
```
vhargrave, line 956: ouch - Could we not refactor this entire `if` block to `decorateSection`, to prevent this rerun of `decorateSection`?

vhargrave (commented):
Gave it a couple of tests and couldn't find any issues :) 
Found a few things in the code. Curious what you think 💪

libs/utils/utils.js
```diff
@@ -500,50 +507,54 @@ export function decorateAutoBlock(a) {
...
+    }
 
-      // Modals
-      if (key === 'fragment' && url.hash !== '') {
-        a.dataset.modalPath = url.pathname;
-        a.dataset.modalHash = url.hash;
-        a.href = url.hash;
-        a.className = 'modal link-block';
-        return true;
+    const isInlineFrag = url.hash.includes('#_inline');
```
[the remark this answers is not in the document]
chrischrischris: The if block on line 526 needs to be run before the `media_.*mp4` check since the fragment could contain an mp4 in the path.  However I think it may be better to wrap the 3 `if` checks inside a `if (key === 'fragment')` block.  I'll give that a try.

Honestly I don't know if that is a valid use case, but I don't want to change the execution order.

libs/utils/utils.js
chrischrischris: No we have to run it again:
1. Inline fragment(s) found
2. Download all of the fragments and insert them into the DOM as if they were part of the original page
3. Now DOM has changed, but that new DOM hasn't been processed
4. Load `decorateSection` again to decorate the updated DOM

However if you see a better way to do this please lmk

libs/utils/utils.js
chrischrischris: How it's written now, decorateSection adds `#_inline` to all fragment links inside of blocks (excluding DO_NOT_INLINE blocks) but does not look for `#_inline` fragments outside of blocks.

You're right though that we could have `decorateLinks` return any `#_inline` links, though we'd also have to check that it's a fragment path.

I'll give it a try.

chrischrischris:
> Hi @chrischrischris. I'm sure it's not a surprise to you, but just to confirm, I copied your updates to fragments.js and utils.js into a copy of the uar-integration branch and the code that @colloyd wrote to dynamically load fragments in the quiz-results block still seems to work as it should.

@echampio-at-adobe quiz-results may be able to make use of inline fragments to simplify things.  Once this is merged we can go over and see if that would help you guys.

libs/utils/utils.js
chrischrischris: Did some perf testing, the selector takes < 1ms to run, so I think I'd rather leave as is instead of having logic in `decorateLinks` and `decorateSection` to find and pass around the inline frags.

libs/utils/utils.js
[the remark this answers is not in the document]
vhargrave: fair

ryanmparrish:
@chrischrischris Looks good on all my testing. 

I have a related [issue](https://jira.corp.adobe.com/browse/MWPW-133742) w/ fragment previews auto blocking in [marquee-anchor](https://inlinefrag--milo--adobecom.hlx.page/fragments/consonant/marquee/marquee-anchors) usage, we already discussed a fix for this adding another conditional in `decorateAutoBlock`.. 
```
if (a.href === window.location.href)  return false;
```
Just sharing w/ the group what we discovered. Thanks

test/blocks/fragment/fragment.test.js
```diff
@@ -44,4 +44,17 @@ describe('Fragments', () => {
     expect(document.querySelector('h4')).to.exist;
     expect(window.lana.log.args[2][0]).to.equal('ERROR: Fragment Circular Reference loading http://localhost:2000/test/blocks/fragment/mocks/fragments/frag-a');
   });
+
+  it('Inlines fragments inside a block', async () => {
+    const marquee = document.querySelector('.marquee-section');
+    await loadArea(marquee);
+    // await waitForUpdate(marquee);
```
overmyheadandbody, line 51: If this is no longer needed, it should be removed, as well as the `waitForUpdate` import, so ESLint stops complaining about it.

overmyheadandbody (approved):

vhargrave (approved):

PR #1246, opened 2023-09-08
Fix hidden IG and /reel/ links

chrischrischris:
Instagrams that load with IO after the first IG is loaded were not displaying, this uses the instagram api to manually display them (`window.instgrm.Embeds.process`)

Also `/reel/` links were not working, so we replace that with `/p/`

**Test URLs:**
- After: https://instagfix--milo--adobecom.hlx.page/?martech=off


rgclayton (approved):

rgclayton:
Verified with blog link - https://main--blog--adobecom.hlx.page/en/topics/nwsl?milolibs=instagfix

PR #1252, opened 2023-09-11
MWPW-136094 Fix links authored using placeholders

puruadobe:
Links when authored using placeholders were broken due to Franklin "url encoding the placeholder". Updated regex and the placeholder replace method to handle url encoded placeholders.

Resolves: [MWPW-136094](https://jira.corp.adobe.com/browse/MWPW-136094)

**Test URLs:**
- Before:  https://main--milo--puruadobe.hlx.live/drafts/cmillar/placeholder-demo?martech=off
- After: https://puru-mwpw-136094--milo--puruadobe.hlx.live/drafts/cmillar/placeholder-demo?martech=off


libs/blocks/global-footer/global-footer.js
```diff
@@ -108,7 +108,7 @@ class Footer {
 
     if (!html) return null;
 
-    const parsedHTML = await replaceText(html, getFedsPlaceholderConfig(), /{{(.*?)}}/g, 'feds');
+    const parsedHTML = await replaceText(html, getFedsPlaceholderConfig(), /{{(.*?)}}|%7B%7B(.*?)%7D%7D/g, 'feds');
```
overmyheadandbody, line 111: Did you give `decodeURIComponent(html)` a shot? That might be a bit more straightforward than basically duplicating all the regular expressions.

overmyheadandbody: This is a question for other reviewers: are there use-cases where encoded strings are encountered in a Word document? `%20, %25`, etc.? It doesn't seem highly likely, but there was talk of some code blocks which might contradict this assumption. If there are no concerns, could `unescape` be safely used for all content?

If not, I'd at least suggest removing the explicit regex from `global-navigation.js`, `global-footer.js` and `menu.js`, `undefined` should suffice so we don't duplicate the definition from `placeholders.json`.

puruadobe: Isn't it replacing unescape entire dom/html maybe expensive operation for pages containing a lot of blocks/content.    

I agree that if there is no requirement on global-*.js, menu.js  we can remove the regex for url-encoded placeholders.

puruadobe: @overmyheadandbody  Sorry, I didn't catch the undefined part of your statement.

overmyheadandbody: `unescape` should not be an expensive operation, I had a look over a few pages and added some performance markers and the operation seems to take less than `1ms`, usually `0.1ms`. I'm more worried about other possible side-effects this might have, other developers might have good reasons for not wanting to do this.

Regarding `undefined`, because of how the placeholder logic is built, it takes multiple parameters, not a single object. Because we need a specific placeholder sheet, `feds`, which is the last argument of the method, we'll need to make the regex `undefined` so the default value is used, something like `const parsedHTML = await replaceText(html, getFedsPlaceholderConfig(), undefined, 'feds');`.

puruadobe: @overmyheadandbody there was a related discussion on 'unescape' content. https://adobe-mwp.slack.com/archives/C03B0BUA75E/p1694546087448259. Seems like encoding of certain characters in content is on purpose.

libs/features/placeholders.js
```diff
@@ -105,7 +105,7 @@ export async function replaceText(text, config, regex = /{{(.*?)}}/g, sheet = 'd
   if (!matches.length) {
     return text;
   }
-  const keys = Array.from(matches, (match) => match[1]);
+  const keys = Array.from(matches, (match) => (match[1] ? match[1] : match[2]));
```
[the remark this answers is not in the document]
puruadobe: Done

libs/features/placeholders.js
```diff
@@ -98,14 +98,14 @@ export async function replaceKeyArray(keys, config, sheet = 'default') {
...
-export async function replaceText(text, config, regex = /{{(.*?)}}/g, sheet = 'default') {
+export async function replaceText(text, config, regex = /{{(.*?)}}|%7B%7B(.*?)%7D%7D/g, sheet = 'default') {
   if (typeof text !== 'string' || !text.length) return '';
 
   const matches = [...text.matchAll(new RegExp(regex))];
   if (!matches.length) {
     return text;
   }
-  const keys = Array.from(matches, (match) => match[1]);
+  const keys = Array.from(matches, (match) => (match[1] ? match[1] : match[2]));
```
overmyheadandbody, line 108: Wouldn't `match[1] || match[2]` do the trick as well instead of the ternary operator?

overmyheadandbody (approved):
Approving for now, but please add Milo test links to the PR's description as well, so the PSI checks can run

@overmyheadandbody Done!

libs/features/placeholders.js
puruadobe: yes, it does :) Done

puruadobe:
> Approving for now, but please add Milo test links to the PR's description as well, so the PSI checks can run

Updated with Milo live page for PSI checks. Thanks.

Blainegunn (approved):

PR #1253, opened 2023-09-12
MWPW-126974: extracts URL string from html links in caas card-metadata

cmiqueo:
Sometimes users accidentally convert URLs into html-links by entering space or return after the URL in SharePoint.

This PR extracts URL string from the html links in the caas card-metadata block during the send to caas workflow.
 
Resolves: [MWPW-126974](https://jira.corp.adobe.com/browse/MWPW-126974)

**Test URLs:**
- After: https://126974--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/tools/send-to-caas/send-utils.js
+++ b/tools/send-to-caas/send-utils.js
@@ -143,9 +143,16 @@ const prefixHttps = (url) => {
   return url;
 };
 
+const flattenLink = (link) => {
+  var htmlElement = document.createElement('div');
+  htmlElement.innerHTML = link;
+  return htmlElement.querySelector('a').getAttribute('href');
+};
+
 const checkUrl = (url, errorMsg) => {
   if (url === undefined) return url;
-  return isValidUrl(url) ? prefixHttps(url) : { error: errorMsg };
+  const flatUrl = (url.indexOf('href=')) ? flattenLink(url) : url;
+  return isValidUrl(flatUrl) ? prefixHttps(flatUrl) : { error: errorMsg };
 };
 
 // Case-insensitive search through tag name, path, id and title for the searchStr


cmiqueo:
> Hi @cmiqueo - I think there are a couple of things you should address here:
> 
> * the _after_ URL is not working. It would be great to showcase the changes addressed in this PR
> * unit tests are missing

@narcis-radu I have updated the "After" link.
 But in any case I not sure what the best way to test this is since this is. 
@chrischrischris do you have any suggestions?

mokimo:
@cmiqueo for the unit tests there are two options

1. set the tests up in a way that you can pass the link as html-link or as url and check if the output is correct either way.
2. export the `checkUrl` method and import it in a unit test so you can test it in isolation and pass the input as it would come either as html-link or url

tools/send-to-caas/send-utils.js
```diff
@@ -143,9 +143,16 @@ const prefixHttps = (url) => {
...
+const flattenLink = (link) => {
+  const htmlElement = document.createElement('div');
+  htmlElement.innerHTML = link;
+  return htmlElement.querySelector('a').getAttribute('href');
+};
+
 const checkUrl = (url, errorMsg) => {
   if (url === undefined) return url;
-  return isValidUrl(url) ? prefixHttps(url) : { error: errorMsg };
+  const flatUrl = (url.indexOf('href=')) ? flattenLink(url) : url;
```
chrischrischris, line 154: `-1` would eval to true

```suggestion
  const flatUrl = url.includes('href=') ? flattenLink(url) : url;
```

chrischrischris (approved):

sheridansunier (approved):

PR #1258, opened 2023-09-12
[MWPW-135720] Decorate video only when the element is visible.

seanchoi-dev:
Resolves: [MWPW-135720](https://jira.corp.adobe.com/browse/MWPW-135720)

- Added intersection observer for video auto block (with hiding the a link before decorating it)
- Added a logic for Marquee not to run the fallback decorateVideo() if current screen size doesn't meet the viewport that includes video.

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/seanchoi/marquee-video?martech=off
- After: https://MWPW-135720-marquee-video--milo--adobecom.hlx.page/drafts/seanchoi/marquee-video?martech=off

### DIFF
--- a/libs/blocks/marquee/marquee.js
+++ b/libs/blocks/marquee/marquee.js
@@ -24,27 +24,34 @@ const decorateVideo = (container, src) => {
 };
 
 const decorateBlockBg = (block, node) => {
-  const viewports = ['mobile-only', 'tablet-only', 'desktop-only'];
+  const viewports = { 'mobile-only': 0, 'tablet-only': 600, 'desktop-only': 1200 };
+  const viewportsKeys = Object.keys(viewports);
+  const viewportWidth = Math.max(document.documentElement.clientWidth || 0, window.innerWidth || 0);
   const childCount = node.childElementCount;
   const { children } = node;
 
   node.classList.add('background');
 
   if (childCount === 2) {
-    children[0].classList.add(viewports[0], viewports[1]);
-    children[1].classList.add(viewports[2]);
+    children[0].classList.add(viewportsKeys[0], viewportsKeys[1]);
+    children[1].classList.add(viewportsKeys[2]);
   }
 
   [...children].forEach(async (child, index) => {
     if (childCount === 3) {
-      child.classList.add(viewports[index]);
+      child.classList.add(viewportsKeys[index]);
     }
-    const videoElement = child.querySelector('a[href*=".mp4"], video source[src$=".mp4"]');
-    if (videoElement) {
-      const video = decorateVideo(child, videoElement.href || videoElement.src);
-      const hash = video.firstElementChild?.src.split('#')[1];
-      if (hash?.includes('autoplay1')) {
-        video.removeAttribute('loop');
+
+    // Skip the fallback if current screen size isn't matching the child's viewport.
+    if (viewportWidth >= viewports[viewportsKeys[index]]) {
+      // decorateVideo as fallback of video autoblock.
+      const videoElement = child.querySelector('a[href*=".mp4"], video source[src$=".mp4"]');
+      if (videoElement) {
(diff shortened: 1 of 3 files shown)


vgoodric (approved):
Fulfills feature request

libs/blocks/video/video.js
```diff
@@ -11,4 +14,16 @@ export default function init(a) {
   const videoElem = document.body.querySelector(`source[src=".${pathname}"]`)?.parentElement;
   applyHoverPlay(videoElem);
   a.remove();
+};
+
+export default async function init(a) {
+  if (a.textContent.includes('no-lazy')) {
+    loadVideo(a);
+  } else {
+    createIntersectionObserver({
```
narcis-radu, line 23: It looks to me that the intersection observer is not always needed; I think it makes more sense to load it dynamically similar to focal point in marquee - https://github.com/adobecom/milo/blob/main/libs/blocks/marquee/marquee.js#L53

libs/blocks/marquee/marquee.js
```diff
@@ -24,27 +24,36 @@ const decorateVideo = (container, src) => {
...
+    // Skip the fallback if current screen size isn't matching the child's viewport.
+    const nextChildViewportWidth = viewports[viewportsKeys[index + 1]] || 99999;
+    if (nextChildViewportWidth > viewportWidth
+      && viewportWidth >= viewports[viewportsKeys[index]]) {
+      // decorateVideo as fallback of video autoblock.
+      const videoElement = child.querySelector('a[href*=".mp4"], video source[src$=".mp4"]');
+      if (videoElement) {
+        const video = decorateVideo(child, videoElement.href || videoElement.src);
+        const hash = video.firstElementChild?.src.split('#')[1];
+        if (hash?.includes('autoplay1')) {
```
narcis-radu, line 54: no need for curly brackets

libs/blocks/video/video.js
seanchoi-dev: This ticket is for performance. I think Video will be still loaded background if it is not handled in video.js.
Do you see any downside to have intersection observer on video?

narcis-radu: I don't see a downside to having an intersection observer for video, all I'm saying is that `createIntersectionObserver` is only used when video is lazy loaded.
```js
  if (a.textContent.includes('no-lazy')) {
    loadVideo(a);
  } else {
    const { createIntersectionObserver } = await import('../../utils/utils.js'');
    createIntersectionObserver({
      el: a,
      options: { rootMargin: `${ROOT_MARGIN}px` },
      callback: loadVideo,  
    });
  }
```

seanchoi-dev: I think having video lazy loaded (w/ intersection observer) as default is a good way to improve our performance in overall.
And author can add `no-lazy` in case author doesn't want with the current logic so, it is good to go as it is.

narcis-radu: I agree that video should be lazy loaded and the intersection observer is good for performance but what I mentioned and highlighted with the code was that maybe we should import the `createIntersectionObserver` method only when authors want to load the video lazy.

seanchoi-dev: I am not sure if I am understanding your comment 100% but, I believe it should be almost no impact on page performance having another import `utils.js` since it should be loaded before it got here (at scripts.js). And what I am trying to say is authors should want to load the video lazy as default, and in case they want no-lazy, it won't hit import createIntersectionObserver anyway since it is only under `else` condition.

narcis-radu: Hey Sean, no worries, this was mainly a suggestion to showcase best practices. We've had performance problems in the past and since there weren't any existing good examples in the code, the PRs were without awareness.

Sartxi (approved):

PR #1263, opened 2023-09-14
MWPW-135672 analytics

vgoodric:
* Add impression tracking to the main element so that all its content automatically sends impression-tracking
* Add default link tracking to all links and buttons (even when not in a block) while not overriding custom tracking added by a block
* export decorateDefaultLinkAnalytics for use in blocks in case it's needed during decoration
* Add a section number link header to sections that are direct children of the main element
* Add a block number link header to each block
* Add default link tracking to all links and buttons
* All milo core blocks checked for and updated to not need decorateBlockAnalytics and decorateLinkAnalytics.  Functions left in until blocks in other servers can be updated
* Tested against all library pages in Milo core. See here for a summary: https://adobe.sharepoint.com/:x:/r/sites/adobecom/_layouts/15/Doc.aspx?sourcedoc=%7B0CB55C19-EB88-4562-955B-FDECD21C9849%7D&file=impression-tracking-summary.xlsx&action=default&mobileredirect=true

Resolves: [https://jira.corp.adobe.com/browse/MWPW-135672](https://jira.corp.adobe.com/browse/MWPW-135672)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/docs/library/blocks/marquee/?martech=off
- After: https://mwpw-135672-analytics--milo--adobecom.hlx.page/docs/library/blocks/marquee/?martech=off


Ruchika4:
Thanks for raising the PR for analytics @vivgoodrich . Could you please look at the unit test and lint failures. Also please add a relevant label to the PR.

libs/utils/decorate.js
```diff
@@ -1,6 +1,3 @@
-import { createTag } from './utils.js';
```
Ruchika4, line 1: createTag is used in decorate,js file. I believe we shouldn't be removing it's import as this will cause error. It is also causing failure of unit tests
Screen Shot 2023-09-13 at 10 40 25 PM

Ruchika4 (changes requested):

libs/utils/decorate.js
vgoodric: yeah, it seems odd to me that it's importing from decorate, but safer to restore it

Ruchika4:
Thanks for addressing the changes requested @vgoodric . Everything looks good. However I am not completely familiar with changes done in personalization.js Tagging @chrischrischris to take a look at that once.

Ruchika4 (approved):

libs/features/personalization/personalization.js
```diff
@@ -555,6 +556,11 @@ export async function applyPers(manifests) {
   deleteMarkedEls();
 
   const experiments = results.map((r) => r.experiment);
+  if (experiments.length) {
+    const manifestLh = experiments.map((e) => e.manifest.split('/').pop().replace('.json', ''));
+    const variantLh = experiments.map((e) => e.selectedVariantName?.replace('target-', '') || 'default');
+    document.body.dataset.mep = `${variantLh.join('--')}|${manifestLh.join('--')}`;
```
chrischrischris, line 562: Edge case: If there are manifests but no experiments, then `document.body.dataset.mep` will not be defined, which means that there will be a trailing `|` in the analytics string.

libs/martech/analytics.js
```diff
@@ -0,0 +1,44 @@
...
+
+export function decorateDefaultLinkAnalytics(block) {
+  if (block.classList.length
+    && !block.className.includes('metadata')
+    && !block.classList.contains('link-block')
+    && !block.classList.contains('section')
+    && block.nodeName === 'DIV') {
+    let header = '';
+    let linkCount = 1;
+    block.querySelectorAll('h1, h2, h3, h4, h5, h6, a:not(.video.link-block), button, .heading-title').forEach((item) => {
```
chrischrischris, line 14: nit: line length > 100 here and L19

libs/martech/attributes.js
```diff
@@ -1,26 +1,9 @@
...
-    const { classList } = link;
-    if (classList.contains('con-button') && classList.contains('blue')) { linkType = 'filled'; }
-    if (classList.contains('con-button') && classList.contains('outline')) { linkType = 'outline'; }
-    const str = `${linkType}|${link.innerText} ${i + 1}`;
-    link.setAttribute('daa-ll', str.replace(/\s+/g, ' ').trim());
-  });
-}
+// decorateBlockAnalytics & decorateLinkAnalytics are being sunset
+export function decorateBlockAnalytics() { return false; }
+export function decorateLinkAnalytics() { return false; }
```
chrischrischris, line 6: Looks like `analyticsDecorateList` is only used in footer.js.  I think it would be safe to move there and remove this file (after checking to make sure no consumers are using it)

libs/utils/utils.js
```diff
@@ -959,6 +962,10 @@ async function documentPostSectionLoading(config) {
 
   const { default: delayed } = await import('../scripts/delayed.js');
   delayed([getConfig, getMetadata, loadScript, loadStyle, loadIms]);
+
+  import('../martech/analytics.js').then((analytics) => {
```
chrischrischris, line 966: Since this is in `documentPostSectionLoading`, analytics will not be run on any fragments that are loaded.

libs/utils/utils.js
```diff
@@ -603,6 +603,7 @@ function decorateContent(el) {
   const block = document.createElement('div');
   block.className = 'content';
   block.append(...children);
+  block.dataset.block = 'true';
```
chrischrischris, line 606: Where is this used?

test/blocks/accordion/accordion.test.js
```diff
@@ -33,15 +33,15 @@ describe('Accordion', () => {
     // handleClick()
     const firstAccordionButton = document.body.querySelector('dt button');
     expect(firstAccordionButton.getAttribute('aria-expanded')).to.equal('false');
-    expect(firstAccordionButton.getAttribute('daa-ll')).to.equal(`open-1|${'What if my dough didn\'t rise?'.trim().slice(0, 30)}|accordion`);
+    expect(firstAccordionButton.getAttribute('daa-ll')).to.equal(`open-1|${'What if my dough did'.trim().slice(0, 30)}`);
```
chrischrischris, line 36: Do we need to keep the `.trim().slice`?  The string is already trimmed and only 20 chars long.

test/martech/analytics.test.js
```diff
@@ -0,0 +1,21 @@
...
+import { expect } from '@esm-bundle/chai';
+
+describe('Analytics', async () => {
+  beforeEach(async () => {
+    await readFile({ path: './mocks/body.html' });
+    const analytics = await import('../../libs/martech/analytics.js');
+    document.body.outerHTML = await readFile({ path: './mocks/body.html' });
+    document.querySelectorAll('main > div').forEach((section, idx) => analytics.decorateSectionAnalytics(section, idx));
+  });
+  it('should decorate w attributes', async () => {
```
chrischrischris, line 11: w?

libs/features/personalization/personalization.js
vgoodric: Well I set it in 3 places.  Line 562, line 543: https://github.com/adobecom/milo/blob/476d83100b3ed1a4e9822bdd891b2aa3c54cb175/libs/features/personalization/personalization.js#L543
and 827 in utils https://github.com/adobecom/milo/blob/476d83100b3ed1a4e9822bdd891b2aa3c54cb175/libs/utils/utils.js#L827

So I think it will set the value correctly if there are no experiments.  But if you think there is a scenario where those don't cover it, I can add an else clause here.

That being said, a trailing `|` is preferred when there is no value.  It still creates the level but with an empty string, thus keeping all of the other levels in their correct place.

libs/martech/attributes.js
vgoodric: I see it in footer.js and gnav.js.  But this file is now really just for gnav and footer.  None of the other blocks will be using this file. so since they have it in 2 places, I figured I'd leave it in this file.

libs/utils/utils.js
vgoodric: Hmm.  It's working on homepage and we have a ton of fragments in there.
preview link: [https://analytics--homepage--adobecom.hlx.page/homepage/index-loggedout?milolibs=mwpw-135672-analytics](https://analytics--homepage--adobecom.hlx.page/homepage/index-loggedout?milolibs=mwpw-135672-analytics)
source page: [https://adobe.sharepoint.com/:w:/r/sites/adobecom/_layouts/15/Doc.aspx?sourcedoc=%7BE04B5FF7-CEAA-465C-B44E-539B8062B9F1%7D&file=index-loggedout.docx&action=default&mobileredirect=true](https://adobe.sharepoint.com/:w:/r/sites/adobecom/_layouts/15/Doc.aspx?sourcedoc=%7BE04B5FF7-CEAA-465C-B44E-539B8062B9F1%7D&file=index-loggedout.docx&action=default&mobileredirect=true)

But I'm totally willing to move it if there's a better location.  Where do you recommend?

libs/utils/utils.js
vgoodric: analytics.js L35.  It helps me add an accurate sequence number to the block.
https://github.com/adobecom/milo/blob/476d83100b3ed1a4e9822bdd891b2aa3c54cb175/libs/martech/analytics.js#L35

test/blocks/accordion/accordion.test.js
vgoodric: hmmm, I just kept what was in there and updated the value. As a unit test, it doesn't need either IMO.  Updated to simple string.

test/martech/analytics.test.js
vgoodric: not sure what you mean

3ch023: I believe the test message, it says 'should decorate w attributes'. typo?

libs/blocks/share/share.js
```diff
@@ -76,39 +76,34 @@ export default async function decorate(block) {
...
-          'daa-lh': 'reddit:click',
         };
       default:
         return null;
     }
   };
   if (!block.classList.contains('inline')) {
     const heading = toSentenceCase(await replaceKey('share-this-page', config));
-    block.append(createTag('p', null, heading));
+    block.append(createTag('p', { class: 'heading-title' }, heading));
```
overmyheadandbody, line 106: I see the table block has a similar class - https://github.com/search?q=repo%3Aadobecom%2Fmilo%20heading-title&type=code. I'm guessing this is just for having a consolidated selector, right?

libs/martech/analytics.js
```diff
@@ -0,0 +1,48 @@
...
+    && !block.classList.contains('section')
+    && block.nodeName === 'DIV') {
+    let header = '';
+    let linkCount = 1;
+    block.querySelectorAll('h1, h2, h3, h4, h5, h6, a:not(.video.link-block), button, .heading-title')
+      .forEach((item) => {
+        if (item.nodeName === 'A' || item.nodeName === 'BUTTON') {
+          if (!item.hasAttribute('daa-ll')) {
+            let label = item.textContent;
+            if (label.trim() === '') {
```
overmyheadandbody, line 19: Why not trim it directly when defining the `label` above?

libs/martech/attributes.js
overmyheadandbody: Just as an FYI, those files are the previous versions of the navigations, currently the blocks are `global-navigation.js` and `global-footer.js`. So `gnav.js` and `footer.js` will be deprecated quite soon

test/martech/analytics.test.js
overmyheadandbody: Well, usually `w/` translates to `with`; since it's not so clear, maybe replace with either of those

libs/martech/analytics.js
```diff
@@ -0,0 +1,48 @@
...
+                || item.getAttribute('aria-label')
+                || item.querySelector('img')?.getAttribute('alt')
+                || 'no label';
+            }
+            label = processTrackingLabels(label);
+            item.setAttribute('daa-ll', `${label}-${linkCount}|${header}`);
+          }
+          linkCount += 1;
+        } else {
+          header = processTrackingLabels(item.textContent);
```
overmyheadandbody, line 30: Is this being used just for the `daa-ll` values for the `item`? Or should it define a `daa-lh` one as well?

libs/martech/analytics.js
```diff
@@ -0,0 +1,48 @@
...
+  }
+}
+
+export async function decorateSectionAnalytics(section, idx) {
+  document.querySelector('main')?.setAttribute('daa-im', 'true');
+  section.setAttribute('daa-lh', `s${idx + 1}`);
+  section.querySelectorAll('[data-block="true"] [data-block="true"]').forEach((block) => {
+    block.removeAttribute('data-block');
+  });
+  section.querySelectorAll('[data-block="true"]').forEach((block, blockIdx) => {
```
overmyheadandbody, line 42: Maybe these two could be grouped together. In the `section.querySelectorAll('[data-block="true"]')` items, you could check if it has a `closest` `'[data-block="true"]'` parent and remove the attribute

rgclayton:
Should content/not a block be marked with the `b` identifier? On the after page posted the first content div is getting that ID
image

libs/blocks/share/share.js
chrischrischris: Maybe rename the class to something specific to analytics if it's only being used for that?

libs/martech/attributes.js
vgoodric: Yeah, so since I've moved all of the functions we're using in utils to the analytics.js file, the functions living there until after gnav and footer are deprecated seems fine to me.

libs/utils/utils.js
chrischrischris: nit: since the value is always true:
```suggestion
  block.dataset.block = '';
```

and in analytics.js the selector can be changed to `[data-block]`

libs/utils/utils.js
chrischrischris: ok, I must be wrong there - if we encounter it later it can be fixed then.

libs/blocks/share/share.js
vgoodric: Yes, same situation.  They didn't choose to use an actual header, but it was clear what was the "header" in terms of analytics.

libs/martech/analytics.js
vgoodric: The header only goes on the daa-ll.

vgoodric:
> Should content/not a block be marked with the `b` identifier? On the after page posted the first content div is getting that ID image

@rgclayton I did that very purposely.  Here is my reasoning:
1. every button and link should be tracked even if outside of blocks
2. if they are tracked, they need to match the taxonomy of their counterparts inside blocks. And that means having a block number and name etc.

libs/martech/analytics.js
vgoodric: moved

test/martech/analytics.test.js
vgoodric: updated to "with"

libs/blocks/share/share.js
vgoodric: updated to tracking-header

libs/utils/utils.js
vgoodric: updated to empty string

chrischrischris (approved):

skumar09:
@vivgoodrich : One question on media blocks present z-pattern 

will `daa-lh` attribute stay for media blocks present in  z-pattern ( below image) or no 

image

vgoodric:
> @vivgoodrich : One question on media blocks present z-pattern
> 
> will `daa-lh` attribute stay for media blocks present in z-pattern ( below image) or no
> 
> image

no, they are removed. Please refer to the z-pattern tab in the comparison spreadsheet: [https://adobe.sharepoint.com/:x:/r/sites/adobecom/_layouts/15/Doc.aspx?sourcedoc=%7B0CB55C19-EB88-4562-955B-FDECD21C9849%7D&file=impression-tracking-summary.xlsx&action=default&mobileredirect=true&CID=059E64F1-53E8-446B-999D-1006EE37A219&wdLOR=c8EA57F8F-424F-467B-ACE5-38F3083768A5](https://adobe.sharepoint.com/:x:/r/sites/adobecom/_layouts/15/Doc.aspx?sourcedoc=%7B0CB55C19-EB88-4562-955B-FDECD21C9849%7D&file=impression-tracking-summary.xlsx&action=default&mobileredirect=true&CID=059E64F1-53E8-446B-999D-1006EE37A219&wdLOR=c8EA57F8F-424F-467B-ACE5-38F3083768A5)

Here is a screenshot of the 2nd z-pattern button:
image

Each button/link is properly numbered.  In the old version, they were all listed as "1"

rgclayton (approved):

fullcolorcoder (approved):

ryanmparrish (approved):

libs/features/personalization/personalization.js
```diff
@@ -180,16 +180,15 @@ function normalizeKeys(obj) {
...
-      try {
-        let selectorEl = rootEl.querySelector(cmd.selector);
-        if (!selectorEl) return;
-        if (selectorEl.classList[0] === 'section-metadata') {
-          selectorEl = selectorEl.parentElement || selectorEl;
-        }
-        COMMANDS[cmd.action](selectorEl, cmd.target, manifestId);
-      } catch (e) {
-        console.log('Invalid selector: ', cmd.selector);
+      let selectorEl = rootEl.querySelector(cmd.selector);
```
overmyheadandbody, line 183: I think the `try/catch` block might've made sense here, since I see that unit tests are failing
> SyntaxError: Failed to execute 'querySelector' on 'Document': ''/.marqueee' is not a valid selector.

The actual failing test is https://github.com/adobecom/milo/blob/main/test/features/personalization/personalization.test.js#L145

3ch023: right, we need to catch an error in case someone accidentally will use an invalid selector in MEP manifest.
Please see [https://github.com/adobecom/milo/pull/1348](https://github.com/adobecom/milo/pull/1348)

3ch023 (changes requested):
`let selectorEl = rootEl.querySelector(cmd.selector); ` needs a try/catch wrapping it, I belive it was accidentally removed while merging with 'main' branch.
could you please bring it back?
see [this PR](https://github.com/adobecom/milo/pull/1348) for more details

3ch023 (approved):

PR #1266, opened 2023-09-14
MWPW-136294: authoring support for second source/content-type query in caas

sheridansunier:
**Problem Statement**
"As an author, we need the ability to set two data sources (example, BACOM and Northstar) without pulling duplicate content based on content types displaying in my card collections on the Resource Center homepage. Currently, the card collection is pulling data from the BACOM (MILO) data source, but some content such as Event Sessions are not migrated to MILO and exist only in AEM or the Northstar data source. This can result in duplicates in my card collections if I use both data sources. This is because resource center pulls content types; reports, video, webinar, infographic, ebook, demo, guide, courses which will be served out of Milo and event sessions will be served out of Northstar."
...
This is likely to be necessary for other teams that make the migration to Milo, and not just the Northstar/BACOM migration.  So we decided that this kind of advanced tag configuration would be something that we support

**This PR addresses the above by...** adding the authoring features to milo that would enable a user to add two additional query parameters to a chimera-io request.  For the desired behavior to be observable the following chimera-io PR is required: https://git.corp.adobe.com/wcms/chimera-io/pull/396

Resolves: [MWPW-136294](https://jira.corp.adobe.com/browse/MWPW-136294)

**Test URLs:**
- After: https://MWPW-136294--milo--sheridansunier.hlx.page/tools/caas?martech=off


libs/blocks/caas-config/caas-config.js
```diff
@@ -408,6 +412,9 @@ const TagsPanel = ({ tagsData }) => {
     >
       <${TagSelect} id="orTags" options=${allTags} label="Tags"
     /><//>
+    <label>Advanced Tag Configurations</label>
```
cmiqueo, line 415: Minor (subjective) edit:
Change label to "Advanced Tag Configuration"

libs/blocks/caas-config/caas-config.js
```diff
@@ -408,6 +412,9 @@ const TagsPanel = ({ tagsData }) => {
...
+    <${Input} label="Use a secondary source for some content types" prop="showSecondarySource" type="checkbox" />
```
cmiqueo, line 416: Minor (subjective) edit: 
Change label to "Use a secondary source"

libs/blocks/caas-config/caas-config.js
```diff
@@ -377,6 +377,10 @@ const TagsPanel = ({ tagsData }) => {
     });
   };
 
+  const secondarySourcePanel = html`
+    <${DropdownSelect} options=${defaultOptions.source} prop="secondSource" label="Secondary Source" />
```
cmiqueo, line 381: Minor (subjective) edit: 
For consistency, should this be prop be named `secondarySource`?

cmiqueo (approved):

libs/blocks/caas/utils.js
```diff
@@ -358,7 +358,9 @@ export const getConfig = async (originalState, strs = {}) => {
         state.endpoint
       }${targetActivity}?originSelection=${originSelection}&contentTypeTags=${state.contentTypeTags.join(
         ',',
-      )}&collectionTags=${collectionTags}&excludeContentWithTags=${excludeContentWithTags}&language=${language}&country=${country}&complexQuery=${complexQuery}&excludeIds=${excludedCards}&currentEntityId=&featuredCards=${featuredCards}&environment=&draft=${
+      )}&secondSource=${state.showSecondarySource ? state.secondSource.join(',') : []}&secondaryTags=${state.showSecondarySource ? state.secondaryTags.join(
+        ',',
```
mokimo, line 362: The spacing seems a little off

libs/blocks/caas-config/caas-config.js
mokimo, line 416: Is it intended that the input is always visible, while the panel is only visible when context.state.showSecondarySource is set?

mokimo (approved):

libs/blocks/caas-config/caas-config.js
sheridansunier: I think since you can use multiple sources in the normal configurator - the clarification that this is specifically for when you want to restrict content types for this source is necessary

libs/blocks/caas-config/caas-config.js
sheridansunier: since the chimera-io changes accepting "secondSource" have already merged I am going to push back on changing this to be "secondarySource" to avoid having to change the property value downstream.  But this is a good point and I probably should've named it that all along

libs/blocks/caas/utils.js
sheridansunier: every time I try to mess around with the way this is spaced I get linting errors - so maybe I haven't tried the correct way to space it yet?

libs/blocks/caas-config/caas-config.js
sheridansunier: The panel should only be visible when showSecondarySource is set - I wanted it to be kind of annoying to use this configuration to avoid an author setting it without knowing what it does

sheridansunier: ok I decided to be less of a baby and just change everything except the parameter sent to chimera-io

libs/blocks/caas/utils.js
chrischrischris: I dealt with that too - essentially there aren't any places for line breaks except the join statements.  I think it's weird but fine.

chrischrischris (approved):

honstar:
Does this PR need verification? If so, please apply the corresponding `needs-verification` label. Otherwise, it should be `verified` by someone so it can be merged.

PR #1268, opened 2023-09-14
MWPW-136171 - Icon block consistant variant convention from "fullwidth" to "full-width"

Sartxi:
- Update the icon block variant `fullwidth` to follow convention and be consistent with the identical text block variant `full-width` (with a space/dash). 
- Added a check for `fullwidth` variant to be updated in icon-block.js for backwards compatibility 
- Moving our grid full-width style from icon-block.css to style.css to work with text block full width variant when they are nested in a section grid. 
- I also fixed some style lint errors I saw in style.css

Resolves: [MWPW-136171](https://jira.corp.adobe.com/browse/MWPW-136171)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/sarchibeque/icon-block-full-width?martech=off
- After: https://sarchibeque-mwpw-136171-full-width--milo--adobecom.hlx.page/drafts/sarchibeque/icon-block-full-width?martech=off


ryanmparrish (approved):
This is great. I tested on an [old draft](https://sarchibeque-mwpw-136171-full-width--milo--adobecom.hlx.page/drafts/rparrish/text) and added some use cases and all looks good. 
We can now author grid items w/ an optional full width row and all contained in one section background. 👍

fullcolorcoder (approved):

libs/styles/styles.css
```diff
@@ -2,6 +2,8 @@
  /*** Variables ***/
 
  :root {
+  --feds-totalheight-nav: calc(var(--feds-height-nav) + var(--feds-height-breadcrumbs));
```
overmyheadandbody, line 5: Just seeing this now (and I get that it was simply moved around for the purposes of this PR) and I'd advise caution when using this custom property. `--feds-height-nav` and `--feds-height-breadcrumbs` become available post-LCP, after the Gnav content has been fetched and a call to fetch the Gnav base styles has completed. So post-LCP + 2 calls.

Right now the value is used for setting `scroll-margin-top` for heading elements, which should not cause any issues, but using it for margins, paddings or other layout-changing properties could lead to increased CLS.

It might be safer to add some hardcoded fallbacks, in case someone decides to use this for other purposes: `--feds-totalheight-nav: calc(var(--feds-height-nav, 64px) + var(--feds-height-breadcrumbs, 33px));`. Or even go a step further and turn these fallback values into variables, since we already use them in a few places in this file.

overmyheadandbody (approved):

cmiqueo (approved):

libs/styles/styles.css
Sartxi: @overmyheadandbody I added some variables for the global navigation height and breadcrumb height values and added them as fallbacks on the `--feds-totalheight-nav` and also added the new variables to the places I thought appropriate in the file. Please let me know if you see any issues with my updates or have any other feedback. Thanks for the review!

overmyheadandbody: Thanks for implementing the changes, looking good 👍

PR #1279, opened 2023-09-18
[MWPW-136611] Replace default experience name with IMS client ID

overmyheadandbody:
## Description
The default _gnav_ experience name is not of much use for Analytics purposes, thus we want to replace such strings with the IMS client ID. What is now `gnav|gnav` will become `gnav|acrobatmilo` for an Acrobat page, for example.

## Related Issue
Resolves: [MWPW-136611](https://jira.corp.adobe.com/browse/MWPW-136611)

## Testing instructions
On a test page, run `document.querySelector('.global-navigation').getAttribute('daa-lh')` to check the Analytics value being used. Ideally, there shouldn't be duplicated _gnav_ strings being used.

From the test URLs shared below, just the Acrobat page should see an update in the Analytics tracking attribute.

## Test URLs
**Acrobat:**
- After: https://www.stage.adobe.com/acrobat/online/sign-pdf.html?martech=off&milolibs=gnav-analytics-naming--milo--overmyheadandbody

**BACOM:**
- After: https://business.stage.adobe.com/fr/customer-success-stories.html?martech=off&milolibs=gnav-analytics-naming--milo--overmyheadandbody

**CC:**
- After: https://main--cc--adobecom.hlx.live/?martech=off&milolibs=gnav-analytics-naming--milo--overmyheadandbody

**Milo:**
- Before: https://main--milo--adobecom.hlx.page/ch_de/drafts/ramuntea/gnav-refactor?martech=off
- After: https://gnav-analytics-naming--milo--overmyheadandbody.hlx.page/ch_de/drafts/ramuntea/gnav-refactor?martech=off

### DIFF
--- a/libs/blocks/global-navigation/utilities/utilities.js
+++ b/libs/blocks/global-navigation/utilities/utilities.js
@@ -60,7 +60,8 @@ export function getAnalyticsValue(str, index) {
 export function getExperienceName() {
   const experiencePath = getMetadata('gnav-source');
   const explicitExperience = experiencePath?.split('/').pop();
-  if (explicitExperience?.length) return explicitExperience;
+  if (explicitExperience?.length
+    && explicitExperience !== 'gnav') return explicitExperience;
 
   const { imsClientId } = getConfig();
   if (imsClientId?.length) return imsClientId;
--- a/test/blocks/global-navigation/utilities/utilities.test.js
+++ b/test/blocks/global-navigation/utilities/utilities.test.js
@@ -134,6 +134,19 @@ describe('global navigation utilities', () => {
     expect(experienceName).to.equal(config.imsClientId);
   });
 
+  it('getExperienceName replaces default experience name with client ID', () => {
+    // If the experience name is the default one (gnav), the imsClientId should be used instead
+    const gnavSourceMeta = toFragment`<meta name="gnav-source" content="http://localhost:2000/ch_de/libs/feds/gnav">`;
+    document.head.append(gnavSourceMeta);
+    let experienceName = getExperienceName();
+    expect(experienceName).to.equal(config.imsClientId);
+    // If the experience name is not the default one, the custom name should be used
+    gnavSourceMeta.setAttribute('content', 'http://localhost:2000/ch_de/libs/feds/custom-gnav');
+    experienceName = getExperienceName();
+    expect(experienceName).to.equal('custom-gnav');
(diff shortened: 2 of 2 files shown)


mokimo:
LGTM to on Chrome / FF / Safari. 👍 

It would make sense to include this somewhere in the documentation that having a different file name than "gnav" _replaces_ the clientId

narcis-radu (approved):

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -60,7 +60,8 @@ export function getAnalyticsValue(str, index) {
 export function getExperienceName() {
   const experiencePath = getMetadata('gnav-source');
   const explicitExperience = experiencePath?.split('/').pop();
-  if (explicitExperience?.length) return explicitExperience;
+  if (explicitExperience?.length
+    && explicitExperience !== 'gnav') return explicitExperience;
```
narcis-radu, line 64: Is there a risk of uppercase / capitalisation?

overmyheadandbody: I'd say it's unlikely. If a consumer uses the default, then `gnav` is used - https://github.com/adobecom/milo/blob/main/libs/blocks/global-navigation/global-navigation.js#L676

They would have to explicitly define a _Gnav.docx_ document and update their metadata to use it. Even then, Franklin will make it lowercase. Such an example is the Milo page used in the test URLs, _/ch_de/drafts/ramuntea/gnav-refactor_ - its document is actually named _Gnav refactor_.

mokimo (approved):

overmyheadandbody:
@mokimo - I agree on the documentation part, but I'm not sure if this is a high impact or breaking change PR. Most consumers will keep on getting the same values they do now.

If @narcis-radu or others think this grants a more important label, I can definitely make the changes, though 🙂

overmyheadandbody:
@mokimo - I've added a new section to the FAQ part of the Global Navigation documentation page to details the 3 possible scenarios of how the `daa-lh` value is generated based on what `gnav-source` is used on the page - https://milo.adobe.com/docs/authoring/global-navigation#accordion-2-trigger-2

3ch023 (approved):

SilviuLCF (approved):
Verified , testing details 
https://jira.corp.adobe.com/browse/MWPW-136611

PR #1285, opened 2023-09-19
Restrict changes done for MWPW-136057 to commerce modal

Ruchika4:
* Restrict modal-open css class to commerce modal for CC

Resolves: [MWPW-136656](https://jira.corp.adobe.com/browse/MWPW-136656)

**Test URLs:**
Milo
- Before:  https://main--milo--adobecom.hlx.page/drafts/ruchika/discover1?martech=off
- After: https://modal--milo--ruchika4.hlx.page/drafts/ruchika/discover1?martech=off

DC
- Before:  https://main--dc--adobecom.hlx.page/acrobat/online/convert-pdf?martech=off
- After: https://main--dc--adobecom.hlx.page/acrobat/online/convert-pdf?milolibs=modal--milo--ruchika4&martech=off

CC
- Before: https://main--cc--adobecom.hlx.page/creativecloud/animation/discover?martech=off#after-effects
- After: https://main--cc--adobecom.hlx.page/creativecloud/animation/discover?martech=off&milolibs=modal--milo--ruchika4#after-effects

### DIFF
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -34,7 +34,7 @@ export function sendAnalytics(event) {
 function closeModal(modal) {
   const { id } = modal;
   const closeEvent = new Event('milo:modal:closed');
-  if(document.body.classList.contains('modal-open')) {
+  if (document.body.classList.contains('modal-open')) {
     document.body.classList.remove('modal-open');
   }
   window.dispatchEvent(closeEvent);
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -34,9 +34,7 @@ export function sendAnalytics(event) {
 function closeModal(modal) {
   const { id } = modal;
   const closeEvent = new Event('milo:modal:closed');
-  if (document.body.classList.contains('modal-open')) {
-    document.body.classList.remove('modal-open');
-  }
+  document.body.classList.remove('commerce-modal-open');
   window.dispatchEvent(closeEvent);
   const localeModal = id?.includes('locale-modal') ? 'localeModal' : 'milo';
   const analyticsEventName = window.location.hash ? window.location.hash.replace('#', '') : localeModal;
@@ -176,7 +174,7 @@ export async function getModal(details, custom) {
       .forEach((element) => element.setAttribute('aria-disabled', 'true'));
   }
   if (dialog.classList.contains('commerce-frame')) {
-    document.body.classList.add('modal-open');
+    document.body.classList.add('commerce-modal-open');
     if (isInitialPageLoad) {
       window.addEventListener('message', (messageInfo) => {
         sendViewportDimensionsOnRequest(messageInfo);
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -34,7 +34,7 @@ export function sendAnalytics(event) {
 function closeModal(modal) {
   const { id } = modal;
   const closeEvent = new Event('milo:modal:closed');
-  if(document.body.classList.contains('modal-open') {
+  if(document.body.classList.contains('modal-open')) {
     document.body.classList.remove('modal-open');
   }
   window.dispatchEvent(closeEvent);
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -34,7 +34,9 @@ export function sendAnalytics(event) {
 function closeModal(modal) {
   const { id } = modal;
   const closeEvent = new Event('milo:modal:closed');
-  document.body.classList.remove('modal-open');
+  if(document.body.classList.contains('modal-open') {
+    document.body.classList.remove('modal-open');
+  }
   window.dispatchEvent(closeEvent);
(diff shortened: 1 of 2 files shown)


TsayAdobe:
Tested with the link: 
https://main--dc--adobecom.hlx.page/acrobat/online/convert-pdf?milolibs=modal--milo--ruchika4

Ruchika4:
Adding the slack conversation link about the issue and it's validation done by DC team 
https://adobe-mwp.slack.com/archives/C03HAU5SU2E/p1695085781970119

cc @TsayAdobe @JFernandezAdobe

spadmasa:
Verified the flow in chrome,edge,firefox and brave browser for the CC url  : 
https://main--cc--adobecom.hlx.page/creativecloud/animation/discover?martech=off&milolibs=modal--milo--ruchika4#after-effects
image
In chrome :
https://main--cc--adobecom.hlx.page/creativecloud/animation/discover?martech=off&milolibs=modal--milo--ruchika4#animate
image
image

In edge : 
https://main--cc--adobecom.hlx.page/creativecloud/animation/discover?martech=off&milolibs=modal--milo--ruchika4#animate
image
image

In firefox : 
image

image


In brave : 
image

image


https://modal--milo--ruchika4.hlx.page/drafts/ruchika/discover1?martech=off#animate

sharath-kannan (approved):

libs/blocks/modal/modal.js
```diff
@@ -34,7 +34,9 @@ export function sendAnalytics(event) {
 function closeModal(modal) {
   const { id } = modal;
   const closeEvent = new Event('milo:modal:closed');
-  document.body.classList.remove('modal-open');
+  if (document.body.classList.contains('modal-open')) {
```
overmyheadandbody, line 37: Is this being used in an intersection observer context? Why exactly is the check needed? Otherwise `document.body.classList.remove('mumbo-jumbo')` will not result in an error, it will just not achieve anything.

libs/blocks/modal/modal.js
```diff
@@ -175,6 +176,7 @@ export async function getModal(details, custom) {
       .forEach((element) => element.setAttribute('aria-disabled', 'true'));
   }
   if (dialog.classList.contains('commerce-frame')) {
+    document.body.classList.add('modal-open');
```
overmyheadandbody, line 179: If this is commerce-specific behaviour, shouldn't the class naming reflect that in some way? The generic _modal-open_ naming would lead me to believe that this class is there whenever any modal is displayed.

Ruchika4: update the css class name

libs/blocks/modal/modal.js
```diff
@@ -34,9 +34,7 @@ export function sendAnalytics(event) {
 function closeModal(modal) {
   const { id } = modal;
   const closeEvent = new Event('milo:modal:closed');
-  if (document.body.classList.contains('modal-open')) {
-    document.body.classList.remove('modal-open');
-  }
+  document.body.classList.remove('commerce-modal-open');
```
mokimo, line 37: Should the general modal close the commerce modal?
I would imagine you listen to the event and close the modal based on that

libs/blocks/modal/modal.js
```diff
@@ -176,7 +174,7 @@ export async function getModal(details, custom) {
       .forEach((element) => element.setAttribute('aria-disabled', 'true'));
   }
   if (dialog.classList.contains('commerce-frame')) {
```
mokimo, line 176: I'm also not sure if the commerce-frame actually belongs into the generic modal code

Ruchika4: It was done as part of another ticket to fix modal size in mobile and tablet if it is opening a commerce iframe in the modal. PR https://github.com/adobecom/milo/pull/1031

Ruchika4: commerce iframe is using the same modal to open and close just applies a specific css class to define its height and width.

libs/blocks/modal/modal.js
overmyheadandbody: Come to think of it, @mokimo does have a point. Why is this logic here instead of the commerce code? There's already a `milo:modal:closed` event that could be listened to on the commerce side. When it fires, you can remove the commerce-specific class. Same for adding the class, you should be able to dispatch an event from the general modal logic here and tie into it in the commerce logic.

libs/blocks/modal/modal.js
mokimo: PR #1031 should have had the logic tied into commerce as well. The modal code is loaded in a lot of use-cases, commerce being a very specific one.

IMO very specific code that solves for _one_ use case shouldn't belong into milo core - but the originating logic, in this case commerce.

I'd love to see someone from the @adobecom/admins weigh in on this

mokimo:
The fix is fine and tested and has a production issue tied to it, so I'm fine if you go ahead and merge this (given another approval) - but I'm not sure the fix lives in the correct place.

Might be worth to merge this to fix the issue and then see if the code can be moved

Ruchika4:
As discussed over slack and here, we'll be addressing the location of the code for commerce modal in a later PR and will be merging this PR for now to address the production issue.

PR #1301, opened 2023-09-19
Remove acomsis cookie on sign out.

seanchoi-dev:
This is needed for homepage while it adds `acomsis` cookie when sign-in happens.
There should be removal for the cookie when sign-out.

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/seanchoi/hhs-next-best-door-whitepaper?martech=off
- After: https://acomsis-signout-remove--milo--adobecom.hlx.page/drafts/seanchoi/hhs-next-best-door-whitepaper?martech=off

### DIFF
--- a/libs/blocks/global-navigation/features/profile/dropdown.js
+++ b/libs/blocks/global-navigation/features/profile/dropdown.js
@@ -66,6 +66,10 @@ class ProfileDropdown {
     if (this.openOnInit) trigger({ element: this.buttonElem });
 
     this.decoratedElem.append(this.dropdown);
+
+    window.addEventListener('feds:signOut', () => {
+      document.cookie = 'acomsis=; expires=Thu, 01 Jan 1970 00:00:00 UTC; path=/;';
+    });
   }
 
   async getData() {


libs/blocks/global-navigation/features/profile/dropdown.js
```diff
@@ -66,6 +66,10 @@ class ProfileDropdown {
     if (this.openOnInit) trigger({ element: this.buttonElem });
 
     this.decoratedElem.append(this.dropdown);
+
+    window.addEventListener('feds:signOut', () => {
+      document.cookie = 'acomsis=; expires=Thu, 01 Jan 1970 00:00:00 UTC; path=/;';
```
yesil, line 71: This code must be in Milo homepage, the reason is, a user can sign-in and sign-out on domains that are not managed on Dexter or Milo platforms.

The following logic is already implemented in Dexter Homepage, and I think the same should be implemented in Milo Homepage.

- As a logged-in user, if you sign out on a Milo page or on some other adobe domains and visit www.stage.adobe.com **with** `acomsis` cookie, Akamai will first deliver the logged-in page, then before the page is displayed
- check IMS status,
- if the user is logged-out, delete acomsis cookie
- force refresh in the page

Same can  be said for the inverse;
- As a logged-out user, if you sign in on a Milo page or on some other adobe domains and visit www.stage.adobe.com **without** `acomsis` cookie, Akamai will first deliver the logged-out page, then before the page is displayed
- check IMS status,
- if the user is logged-in, set `acomsis` cookie
- force refresh in the page

yesil (changes requested):

libs/blocks/global-navigation/features/profile/dropdown.js
yesil: please check the homepage version for different edge cases as well.

https://git.corp.adobe.com/wcms/titan/blob/04b986e866f796fb3afb6e78d3ba3bb68a0c518f/app/jcr_root/apps/homepage/components/structure/page/partials/ims.htm

seanchoi-dev: I didn't think about the case of "a user can sign-in and sign-out on domains that are not managed on Dexter or Milo platforms." Closing this PR.

PR #1302, opened 2023-09-19
MWPW 135990 evergreen cards contnent bg

VKniaz:
Added a 'decorate content ' function to author content background color
-- This is authored w/ the following syntax:
/-- #ff0000 
in the merch-card block content area 
--/
-- Note: The color value can be any css color or referenced placeholder value. ex --- {{color-photoshop}}

* Add your
* Specific
* Features or fixes

Resolves: [MWPW-135990](https://jira.corp.adobe.com/browse/MWPW-135990)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/vkniaziev/document4?martech=off
- After: https://MWPW-135990_evergreen_card--milo--adobecom.hlx.page/drafts/vkniaziev/document4?martech=off


libs/utils/decorate.js
```diff
@@ -117,6 +117,29 @@ export const decorateBlockHrs = (el) => {
   if (hasHr && els.length) el.classList.add('has-divider');
 };
 
+export const decorateBgContent = (el) => {
```
3ch023, line 120: do we need the flexibility for authors to choose the color? maybe we can just offer a few types of background (tied to classes) and save the js processing?

3ch023 (changes requested):
Hi, 
In [https://jira.corp.adobe.com/browse/MWPW-135990](https://jira.corp.adobe.com/browse/MWPW-135990) I don't see anything mentioning changing the background of the card descr section. 
For which card [of these](https://wiki.corp.adobe.com/display/WP4/Commerce+cards+on+adobe.com) is this change needed? 

we probably need to update the ticket description (or even create a separate ticket for this change)

libs/utils/decorate.js
```diff
@@ -117,6 +117,29 @@ export const decorateBlockHrs = (el) => {
...
+  const els = el.querySelectorAll('p');
+  let insidePattern = false;
+  let decoratedBlock;
+  [...els].forEach((e) => {
```
vladen, line 124: For evergreen browsers, there's no need to convert `NodeList` into an array in order to use `forEach` method: https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach

PR #1303, opened 2023-09-19
MWPW-133600 - Floodgate for loading Caas Card Collections

arshadparwaiz:
- Fetch Caas from Floodgate, based on Metadata floodgatecolor attribute with pink as value
- Swap the Chimera endpoint to use AdobeIO endpoint instead of Adobe.com's chimera endpoint, to retain the FG header we attach (Akamai removes the header if it is invoked from a non Adobe.com URL)
- Added Draft=true for hlx.page URLs & for Local servers, to pull from Chimera's Draft container
- Remove the Dev Chimera Endpoint from Caas Configurator endpoints' list

Resolves: [MWPW-133600](https://jira.corp.adobe.com/browse/MWPW-133600)

Test URLs:

Before: https://main--milo--adobecom.hlx.page/drafts/arshad/fg/fg-caas-test?caasver=0.6.38&martech=off
After: https://main--milo--arshadparwaiz.hlx.page/drafts/arshad/fg/fg-caas-test?caasver=0.6.38&martech=off


arshadparwaiz:
Lighthouse score is already low with only one Caas block added on a page through adobecom Prod code. This PR is not impacting the score.

sukamat (approved):
@arshadparwaiz Validated with some sample pages. Seems to be working as desired. We have to test this again on milo.adobe.com with FG cookie once the PR is merged.

libs/blocks/caas/caas.js
```diff
@@ -30,6 +32,30 @@ const loadCaas = async (a) => {
...
+
+  if (host.includes('stage.adobe') || env?.name === 'local') {
+    chimeraEndpoint = S_CAAS_AIO;
+  } else if (host.includes('.hlx.')) {
+    // If invoking URL is not an Acom URL, then switch to AIO
+    chimeraEndpoint = P_CAAS_AIO;
+  }
+
+  if (host.includes('hlx.page') || env?.name === 'local') {
+    // Set draft=true if URL consisting hlx.page (if published url or hlx.live, then draft is false)
```
chrischrischris, line 52: nit: no need for this comment and line 56 comment

libs/blocks/caas/caas.js
```diff
@@ -30,6 +32,30 @@ const loadCaas = async (a) => {
...
   a.remove();
 
+  const floodGateColor = getMetadata('floodgatecolor') || '';
+  if (floodGateColor === fgHeaderValue) {
+    state.fetchCardsFromFloodgateTree = true;
+  }
+
+  const { env } = getConfig();
+  const { host } = window.location;
+  let chimeraEndpoint = 'www.adobe.com/chimera-api/collection';
```
chrischrischris, line 42: The advanced panel allows users to select an endpoint:
Screenshot 2023-09-22 at 1 18 14 PM

This now makes it so that only the prod endpoint is used.   I think that's fine, but if that's what we want we should rip out the Advanced Endpoint option.

sirivuri (approved):

libs/blocks/caas/caas.js
sanrai: I think we should keep the Advanced Endpoint option and not hard code it to prod. People would want to be able to point to stage content.

libs/blocks/caas/caas.js
arshadparwaiz: Updated, thanks

libs/blocks/caas/caas.js
arshadparwaiz: @chrischrischris as part of this PR, we are using both Stage & Prod endpoints, as discussed the endpoints are base64 encoded. Hence we can keep the Stage & Prod both in the configurator, we have removed the Dev endpoint from the configurator.

tools/send-to-caas/send-utils.js
```diff
@@ -416,7 +416,7 @@ const props = {
   cardimage: () => getCardImageUrl(),
   cardimagealttext: (s) => s || getCardImageAltText(),
   contentid: (_, options) => {
-    const floodGateColor = getMetadata('floodGateColor') || '';
+    const floodGateColor = getMetadata('floodgatecolor') || '';
```
honstar, line 419: Do we need to think about potential regression, i.e. existing metadata on pages using the camel-case variant `floodGateColor`? Maybe it makes sense to make the `getMetadata` function more robust in that it ignores case altogether and always does a lowercase before accessing properties from DOM elements.

arshadparwaiz: @honstar Thanks for looking, the property floodgatecolor is still not in use by consumers, is part of the feature work happening now. 
I think it’s a good idea to make the getmetadata function to ignore case for all properties, however other metadata properties are in use by consumers and can have potential impact, hence do you think it will be better to leave the generic function in current state.

chrischrischris (approved):

PR #1323, opened 2023-09-21
MWPW-136143: Autodetect not working, Milo not sending correct lang to CaaS

sanrai:
When using `autoDetect` (where Milo dynamically tries to figure out the correct country/lang to send to CaaS) it turns out it fails for every locale for CaaS.

The idea for this was to prevent authors from having to manually author the `country/lang` for every collection.

But the way this currently works in production is either:

- Authors are manually setting the lang/country at the collection level using tags (This is occurring for all Tier 1 locales and is actually how https://business.adobe.com/customer-success-stories.html is live right now)

image

- For ones that aren't manually set, autoDetect sends a nonsense lang. Luckily CaaS's backend is able to a generate some search results for this BUT it gives us undefined behavior like a mix of English and translated content

image

The way this is fixed is I added a map in the caas block with all the milo regions as keys and their values being the actual correct locales that should be sent to CaaS. 

This map:
- Is in sync with the one milo uses AND also maps to every locale that is used when someone sends to CaaS
- Is only loaded when a CaaS block is loaded on the page. I was originally going to put it in `scripts.js` but realized that would load unnecessary data for pages that aren't using CaaS.

Resolves: https://jira.corp.adobe.com/browse/MWPW-136143

**Test URLs:**
- Before: https://main--bacom--adobecom.hlx.live/la/customer-success-stories
- After: https://main--bacom--adobecom.hlx.live/la/customer-success-stories?milolibs=MWPW-136143

### DIFF
--- a/libs/blocks/caas/utils.js
+++ b/libs/blocks/caas/utils.js
@@ -261,32 +261,33 @@ const getFilterArray = async (state, country, lang, strs) => {
 
 export function getCountryAndLang({ autoCountryLang, country, language }) {
   const locales = getMetadata('caas-locales');
-  if(locales){
+  if (locales) {
+    /* eslint-disable-next-line object-shorthand */
     return {
       country: '',
       language: '',
-      locales: locales
-    }
+      locales: locales,
+    };
   }
   if (autoCountryLang) {
-    let locale = pageConfigHelper()?.locale?.ietf || 'en-us';
+    const locale = pageConfigHelper()?.locale?.ietf || 'en-us';
     const region = pageConfigHelper()?.locale?.region || '';
-    let country = locale.split('-')[1];
-    if(!country){
+    country = locale.split('-')[1];
+    if (!country) {
       country = region;
     }
-    const language = locale.split('-')[0];
+    language = locale.split('-')[0];
 
     return {
-      country: country,
-      language: language,
-      locales: ''
+      country,
+      language,
+      locales: '',
     };
   }
   return {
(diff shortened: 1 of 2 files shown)


overmyheadandbody (commented):
Same PR pollution comment, PRs should be opened only after they're ready for others to review.

cmiqueo (approved):

libs/blocks/caas/utils.js
```diff
@@ -8,6 +8,92 @@ const pageConfig = pageConfigHelper();
 const pageLocales = Object.keys(pageConfig.locales || {});
 const requestHeaders = [];
 
+const LOCALES = {
```
chrischrischris, line 11: Can we combine this and the LOCALES object in send-utils?  I don't think we should have two copies within the caas files.

chrischrischris: I know this was originally copied from the milo locales object, but I don't think we need the `ietf` subobject:
```
const LOCALES = {
  ar: 'es-AR',
  br: 'pt-BR',
  ...
};
```

chrischrischris: nit: Please alpha sort within each region

chrischrischris (changes requested):
Adding request changes to prevent from merging until the LOCALES obj location can be figured out.

sanrai:
As discussed, moved code to one PR: https://github.com/adobecom/milo/pull/1324

PR #1324, opened 2023-09-21
MWPW-136766, MWPW-136143: Milo not supporting sending to CaaS all locales & Autodetect not working

sanrai:
This PR fixes issues with both the send to caas button and the bulk publisher as well as all the autodetect.

When using the send to caas button, the following locales were mapped incorrectly and would send the wrong data to CaaS.
```
ae_ar, ae_en, africa, cr, ec, eg_ar, el, gr_el, gt, id_en, id_id, il_he, in_hi, kw_ar, mena_ar, mena_en, my_ms, ph_en, qa_ar, sa_ar, sa_en, th_en, th_th, vn_vi
```

For example: this page, https://milo.adobe.com/qa_ar/drafts/sanrai/mwpw-136065, it'll return the wrong value, ar, which then sends the wrong lang/country combination to CaaS (ar_US instead of ar_QA)

This is because the API the code was previously using, `document.documentElement.lang` was incorrect.

Here is it failing for `my_ms`:
image

See this bug for more information:
https://jira.corp.adobe.com/browse/MWPW-136832

When trying sending content from the following set of locales 
```
cr, ec, el, gt, pr, langstore, eg_ar, eg_en, gr_el, il_he, kw_ar, kw_en, ng, qa_ar, qa_en, za
``` 
to CaaS using the bulk publisher (or the Send to CaaS button),  because these locales were undefined, these locales would wrongly be sent into `en_US` to CaaS.

Here is an example of it failing for `qa_ar`:
image

See https://jira.corp.adobe.com/browse/MWPW-136766 for more information on the latter issue.

Resolves: 
https://jira.corp.adobe.com/browse/MWPW-136766, 
https://jira.corp.adobe.com/browse/MWPW-136832

**Test URLs:**
- Before: https://MWPW-136143-2--milo--adobecom.hlx.page/drafts/sanrai/testing-bulk-publish?martech=off
- After: https://MWPW-136143-2--milo--adobecom.hlx.page/drafts/sanrai/testing-bulk-publish?martech=off

When using `autoDetect` (where Milo dynamically tries to figure out the correct country/lang to send to CaaS) it turns out it fails for every locale for CaaS.

The idea for this was to prevent authors from having to manually author the `country/lang` for every collection.

But the way this currently works in production is either:

- Authors are manually setting the lang/country at the collection level using tags (This is occurring for all Tier 1 locales and is actually how https://business.adobe.com/customer-success-stories.html is live right now)

image

- For ones that aren't manually set, autoDetect sends a nonsense lang. Luckily CaaS's backend is able to a generate some search results for this BUT it gives us undefined behavior like a mix of English and translated content

image

The way this is fixed is I added a map in the caas block with all the milo regions as keys and their values being the actual correct locales that should be sent to CaaS. 

This map:
- Is in sync with the one milo uses AND also maps to every locale that is used when someone sends to CaaS
- Is only loaded when a CaaS block is loaded on the page. I was originally going to put it in `scripts.js` but realized that would load unnecessary data for pages that aren't using CaaS.

Resolves: https://jira.corp.adobe.com/browse/MWPW-136143

**Test URLs:**
- Before: https://main--bacom--adobecom.hlx.live/la/customer-success-stories
- After: https://main--bacom--adobecom.hlx.live/la/customer-success-stories?milolibs=MWPW-136143-2


overmyheadandbody (approved):
Same PR pollution comment 🙂

tools/send-to-caas/send-utils.js
```diff
@@ -30,27 +36,34 @@ const LOCALES = {
...
   lt: { ietf: 'lt-LT' },
   lu_de: { ietf: 'de-LU' },
   lu_en: { ietf: 'en-LU' },
   lu_fr: { ietf: 'fr-LU' },
   hu: { ietf: 'hu-HU' },
   mt: { ietf: 'en-MT' },
-  mena: { ietf: 'en-mena' },
   mena_en: { ietf: 'en-mena' },
   mena_ar: { ietf: 'ar-mena' },
-  mena_fr: { ietf: 'fr-mena' },
```
sanrai, line 49: Deleted as `mena_fr` isn't supported when rolled out and to keep parity with the 93 values supported by Milo's `getLocale()` api

cmiqueo (approved):

chrischrischris (approved):

PR #1325, opened 2023-09-21
MWPW-136763: Fixes wrong origins being used by consumers

sanrai:
When sending content to CaaS, milo was using the adobecom github repository name as the origin for the content source in CaaS. 

CaaS has it's own source values (legacy from dexter) that should be used, otherwise each site will have separate buckets where their content exists.

This PR creates a hard coded map for where to pull the content from which will update as new consumers onboard onto CaaS and milo.

Resolves: https://jira.corp.adobe.com/browse/MWPW-136763

**Test URLs:**
- Before: https://milo.adobe.com/tools/send-to-caas/bulkpublisher.html?martech=off
- After: https://mwpw-136763--milo--adobecom.hlx.page/tools/send-to-caas/bulkpublisher.html?martech=off


overmyheadandbody (commented):
Same comment as your other PR, having these draft PRs can be misleading to reviewers going through all the open PRs.

tools/send-to-caas/send-utils.js
```diff
@@ -253,7 +253,20 @@ const getDateProp = (dateStr, errorMsg) => {
...
+    dc: 'doccloud',
+    bacom: 'bacom',
+    experienceleague: 'experienceleague',
+    magento: 'magento',
+    marketo: 'marketo',
+    milo: 'milo',
+    northstar: 'northstar',
+    workfront: 'workfront',
+  };
+  if (mappings[origin]) {
```
cmiqueo, line 267: New line before 'if' statement?

cmiqueo (approved):
just a cosmetic comment

cmiqueo (approved):

tools/send-to-caas/send-utils.js
```diff
@@ -253,7 +253,21 @@ const getDateProp = (dateStr, errorMsg) => {
...
+    experienceleague: 'experienceleague',
+    magento: 'magento',
+    marketo: 'marketo',
+    milo: 'milo',
+    northstar: 'northstar',
+    workfront: 'workfront',
+  };
+
+  if (mappings[origin]) {
+    return mappings[origin] || origin;
```
chrischrischris, line 269: Nit: Since the default is `origin` we can remove all mappings that have key/values that are identical.

tools/send-to-caas/send-utils.js
```diff
@@ -253,7 +253,14 @@ const getDateProp = (dateStr, errorMsg) => {
...
 const getOrigin = () => {
   const origin = getConfig().project || getConfig().repo;
-  if (origin) return origin;
+  const mappings = {
+    cc: 'hawks',
+    dc: 'doccloud',
+  };
+
+  if (mappings[origin]) {
+    return mappings[origin] || origin;
```
overmyheadandbody, line 262: Since you're checking that `mappings[origin]` exist in the `if` statement, this will always return `mappings[origin]` and never `origin`. I think the `if` statement can be removed entirely.

tools/send-to-caas/send-utils.js
```diff
@@ -253,7 +253,14 @@ const getDateProp = (dateStr, errorMsg) => {
 
 const getOrigin = () => {
   const origin = getConfig().project || getConfig().repo;
-  if (origin) return origin;
+  const mappings = {
+    cc: 'hawks',
```
narcis-radu, line 257: I see additional mappings in the JIRA ticket but not in the code. 
One other thing: you should convert `mappings` to UPPERCASE and move it to the top.

mokimo:
@sanrai is this still valuable? There has been no movement since a month.

auniverseaway:
@sanrai this is the second oldest PR. Any update here?

arshadparwaiz (approved):
Looks good. Do we have more mappings for bacom and other consumers, or are they same on milo & chimera ends, hence don’t require a mapping

sanrai:
@arshadparwaiz 

For bacom they were supposed to use northstar, but they created their own bucket in CaaS so we just left it as is.

For other CaaS consumers, we will add as they onboard

> Looks good. Do we have more mappings for bacom and other consumers, or are they same on milo & chimera ends, hence don’t require a mapping

chrischrischris (approved):
Please address PR feedback before merging.

PR #1328, opened 2023-09-21
Exclude locales from automated breadcrumb URLs

mokimo:
## Description
Exclude locale folders as part of the automated breadcrumbs by URLs

## Related Issue
Resolves: [MWPW-136202](https://jira.corp.adobe.com/browse/MWPW-136202)

## Testing instructions
The breadcrumbs should not contain the localised folder

### Before
Screenshot 2023-09-21 at 15 44 35

### After
Screenshot 2023-09-21 at 15 44 38

## Test URLs
**Milo:**
- Before: https://main--milo--adobecom.hlx.page/ch_de/libs/feds/drafts/qa/breadcrumbs/feds-breadcrumbs-no-hidden-links
- After: https://mwpw-136202--milo--mokimo.hlx.page/ch_de/libs/feds/drafts/qa/breadcrumbs/feds-breadcrumbs-no-hidden-links

No changes should be observed here, but for completeness - things should still run...:
**Acrobat:**
- After: https://www.stage.adobe.com/acrobat/online/sign-pdf.html?martech=off&milolibs=mwpw-136202--milo--mokimo

**BACOM:**
- After: https://business.stage.adobe.com/fr/customer-success-stories.html?martech=off&milolibs=mwpw-136202--milo--mokimo


honstar (approved):

test/blocks/global-navigation/ch_de/breadcrumbs-from-file.test.html
```diff
@@ -6,20 +6,39 @@
...
-          const gnavBreadcrumb = breadcrumb.querySelector('a[href $= "/global-navigation"]');
+          const gnavBreadcrumb = breadcrumb.querySelector(
+            'a[href $= "/global-navigation"]'
+          );
           expect(gnavBreadcrumb.textContent).to.equal('global navigation');
         });
 
-        it('should not create a breadcrumb from the url if the config isn\'t on', async () => {
-          document.head.innerHTML = ''
+        it("should not create a breadcrumb from the url if the config isn't on", async () => {
```
overmyheadandbody, line 40: I see the test on L12 uses single quotes, is there a reason for diverging from that?

libs/blocks/global-navigation/features/breadcrumbs/breadcrumbs.js
```diff
@@ -85,7 +85,10 @@ const fromUrl = () => {
   if (getMetadata(metadata.fromUrl) !== 'on') return null;
   const list = toFragment`<ul></ul>`;
   const paths = document.location.pathname.split('/').filter((n) => n);
+  const locales = getConfig().locales || {};
   for (let i = 0; i < paths.length; i += 1) {
+    // eslint-disable-next-line no-continue
+    if (locales[paths[i]]) continue;
```
overmyheadandbody, line 91: I'm not sure I want to disable the `no-continue` rule for this. Initially I was going to suggest we simply wrap the logic inside an `if` clause, but then I figured: why do this check for all breadcrumb entries? The locale should be the first one, right? So how about: outside the `for` loop, check whether the first path matches a known locale. If so, initialise `i` as `1` instead of `0`.

mokimo: Is it guaranteed to be the first path? In theory, it could as well be `stories/ch_de/...` although if a consumer opted for that - that might have other side effects

test/blocks/global-navigation/ch_de/breadcrumbs-from-file.test.html
mokimo: Prettier ran over it! I'll revert those changes

libs/blocks/global-navigation/features/breadcrumbs/breadcrumbs.js
overmyheadandbody: The general rule has the locale as the first item, so that sounds like a good bet to me.

Another question: doesn't getConfig have a `prefix` property under the `locale` object? Maybe we can use just that and do a string replace in the initial `document.location.pathname`, before doing the split

mokimo: Yeah bunch of ways to solve this, prefix sounds good 👍

narcis-radu (approved):

chadsunvice (approved):

PR #1343, opened 2023-09-22
MWPW-135720 - Marquee Video

seanchoi-dev:
Reopening PR (https://github.com/adobecom/milo/pull/1258) with fixed autoplay issue.

Resolves: [MWPW-135720](https://jira.corp.adobe.com/browse/MWPW-135720)
- Also this will make our video rendering experience more consistent (See this [Slack conversation](https://adobe-mwp.slack.com/archives/C03HAU5SU2E/p1695833363239599))
- All default video rendering will be under intersectionObserver which will help our performance and specially help mobile experience when background video doesn't need to be loaded.
- Marquee block to use decorateBlockBg() in decorate.js instead its own.
- decorate.js - decorateBlockBg() now set default .mp4 link's hash with `autoplay`.
- decorate.js - decorateBlockBg() now supports handleFocalpoint().

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/seanchoi/marquee-video?martech=off
- After: https://MWPW-135720-marquee-video--milo--adobecom.hlx.page/drafts/seanchoi/marquee-video?martech=off

`autoplay` test:
- Before: https://main--milo--adobecom.hlx.page/drafts/cpeyer/marq-video
- After: https://mwpw-135720-marquee-video--milo--adobecom.hlx.page/drafts/cpeyer/marq-video

Other marquee test URL:
- https://main--bacom--adobecom.hlx.page/resources/state-of-cx-research-report?milolibs=MWPW-135720-marquee-video

Video test URL:
- Before: https://main--milo--adobecom.hlx.page/drafts/rclayton/poc/video-samples
- After: https://MWPW-135720-marquee-video--milo--adobecom.hlx.page/drafts/rclayton/poc/video-samples

Aside block with video:
- Before: https://main--milo--adobecom.hlx.page/drafts/seanchoi/aside-video
- After: https://mwpw-135720-marquee-video--milo--adobecom.hlx.page/drafts/seanchoi/aside-video
- Other: https://main--cc--adobecom.hlx.page/drafts/dmosier/aside-vs-aside-split?milolibs=MWPW-135720-marquee-video


chrischrischris (changes requested):
This breaks marquee background videos that don't have `autoplay` specified.

See bottom video on:
https://main--milo--adobecom.hlx.page/drafts/cpeyer/marq-video
vs
https://mwpw-135720-marquee-video--milo--adobecom.hlx.page/drafts/cpeyer/marq-video

seanchoi-dev:
> This breaks marquee background videos that don't have `autoplay` specified.
> 
> See bottom video on: https://main--milo--adobecom.hlx.page/drafts/cpeyer/marq-video vs https://mwpw-135720-marquee-video--milo--adobecom.hlx.page/drafts/cpeyer/marq-video

Good catch! I've updated code to handle this issue. Updated code includes:
```
Marquee block to use decorateBlockBg() in decorate.js instead its own.
decorate.js - decorateBlockBg() now set default .mp4 link's hash with autoplay.
decorate.js - decorateBlockBg() now supports handleFocalpoint().
```
I added them in the PR description and your test URLs as well.

libs/utils/utils.js
```diff
@@ -74,7 +74,7 @@ const MILO_BLOCKS = [
   'reading-time',
 ];
 const AUTO_BLOCKS = [
-  { adobetv: 'https://video.tv.adobe.com' },
+  { adobetv: 'tv.adobe.com' },
```
rgclayton, line 77: 👍 

I didn't see a test for this in any of your pages. So, I added [a .mp4 MPC video](https://mwpw-135720-marquee-video--milo--adobecom.hlx.page/drafts/rclayton/video/videos) to a sample page to verify

rgclayton (approved):
Everything appears to be working as outlined. 

Thanks Sean.

Here's some test pages with many video options I used to test.

- https://mwpw-135720-marquee-video--milo--adobecom.hlx.page/drafts/rclayton/video/videos

- https://mwpw-135720-marquee-video--milo--adobecom.hlx.page/drafts/rclayton/video/videos-mp4

libs/blocks/video/video.css
```diff
@@ -1,3 +1,7 @@
+a[href*='.mp4'].hide {
+  visibility: hidden !important;
```
yesil, line 2: `!important` should not be necessary.

libs/blocks/video/video.js
```diff
@@ -22,4 +24,17 @@ export default function init(a) {
...
   applyHoverPlay(videoElem);
   a.remove();
+};
+
+export default function init(a) {
+  a.classList.add('hide');
+  if (a.textContent.includes('no-lazy')) {
+    loadVideo(a);
+  } else {
+    createIntersectionObserver({
```
yesil, line 34: ideally there should be 1 intersection observer created for all the videos not one for each video.

test/blocks/marquee/mocks/video.html
```diff
@@ -23,7 +23,7 @@ <h2 id="heading-xl-marquee-standard-medium-left">Heading XL Marquee standard med
 <div class="marquee" id="multiple-background">
   <div>
     <div><a href="https://business.adobe.com/content/dam/dx/us/en/product-tours/workfront/video/Chapter%201_Video01.mp4">https://business.adobe.com/content/dam/dx/us/en/product-tours/workfront/video/Chapter%201_Video01.mp4</a></div>
-    <div><a href="/drafts/methomas/media_16e79d59a997bc94e5f1b2a518acb152a14f97490.mp4">https://main--milo--adobecom.hlx.page/media_16e79d59a997bc94e5f1b2a518acb152a14f97490.mp4</a></div>
+    <div><a href="https://main--milo--adobecom.hlx.page/media_16e79d59a997bc94e5f1b2a518acb152a14f97490.mp4">https://main--milo--adobecom.hlx.page/media_16e79d59a997bc94e5f1b2a518acb152a14f97490.mp4</a></div>
```
yesil, line 26: test files should not have external resources as dependencies.

libs/blocks/video/video.css
seanchoi-dev: This is for same pattern of hiding other auto blocks:
https://github.com/adobecom/milo/blob/main/libs/blocks/caas/caas.css
https://github.com/adobecom/milo/blob/main/libs/blocks/faas/faas.css
This is little more general since this is no parent class for the a link and I think it is safe to have !important specially with `.hide`. (like bootstrap does.).
If you still disagree having !important for this case,  I can remove it.

libs/blocks/video/video.js
seanchoi-dev: That is kinda difficult approach with current milo structure.
Maybe we can come up a solution adding option at autoblock stage and get the logic there, but probably not with this PR.
https://github.com/adobecom/milo/blob/main/libs/utils/utils.js#L76

test/blocks/marquee/mocks/video.html
seanchoi-dev: Unfortunately, web test runner is not running base of hlx localhost which means, all relative URLs pointing our content base are getting 404. 
(See this example https://github.com/adobecom/milo/actions/runs/6540291246/job/17759947301?pr=1343#step:6:1491)
This is sometimes frustrating when you need to debug your test deeply. There are many other test mockup body using external URLs. 

If you think this is really bugging you, I can still undo my test URLs, I don't think it will really affect existing coverage at this point but adding little more 404s on test.

libs/blocks/video/video.js
yesil: I addressed the video issue as at https://github.com/adobecom/milo/blob/perfbs/libs/blocks/video/video.js, not sure which one is more convenient.

libs/blocks/video/video.css
yesil: strange that without `!important` it doesn't work, anyway not a blocker for me.

test/blocks/marquee/mocks/video.html
yesil: actually I'm probably wrong. What I said is probably for js/css files, so it should be ok.

Sartxi (approved):

PR #1349, opened 2023-09-25
MWPW-135730 - Results card-list alignment correction

elaineskpt:
For quiz-results we use consonants card, they have already grid set.
In our case we need an extra class to correct the alignment as each card will be hold by a fragment, causing that the grid consonant do not get being used in our case.

https://jira.corp.adobe.com/browse/MWPW-135730

**Test URLs:**
- Before: https://uar-integration--milo--adobecom.hlx.page/drafts/delta/app-recommender-test/?martech=off
- After: https://mwpw-135730--milo--elaineskpt.hlx.page/drafts/delta/app-recommender-test/?martech=off

To test in your own folder please update the card-list nested-fragment
Instead style center, three up
Screenshot 2023-09-25 at 18 22 57
to style quiz-card-list

Screenshot 2023-09-25 at 18 39 12


JackySun9:
I want to test this, but for some reason, it doesn't work to me

fullcolorcoder (approved):

JackySun9 (approved):

PR #1348, opened 2023-09-25
MWPW-136982: fix invalid selector

3ch023:
The invalid selector in MEP manifest fails whole page render.
Selector
`4".merch-card1.mep-stock`
[Page Doc](https://adobe.sharepoint.com/:w:/r/sites/adobecom/_layouts/15/Doc.aspx?sourcedoc=%7B8379913A-7C3F-43DC-BDAC-1AE86293F1E0%7D&file=special-offers.docx&action=default&mobileredirect=true)
[Manifest](https://adobe.sharepoint.com/:x:/r/sites/adobecom/_layouts/15/Doc.aspx?sourcedoc=%7B8CC13F11-7811-442C-830B-DA6F47437343%7D&file=bts-manifest.xlsx&action=default&mobileredirect=true)

Resolves: [MWPW-136982](https://jira.corp.adobe.com/browse/MWPW-136982)

**Test URLs:**
- Before: [https://main--milo--adobecom.hlx.page/drafts/mariia/promo/special-offers?martech=off](https://main--milo--adobecom.hlx.page/drafts/mariia/promo/special-offers?martech=off)
- After: https://mwpw-136982--milo--adobecom.hlx.page/drafts/mariia/promo/special-offers?martech=off

image

PR #1353, opened 2023-09-26
safeguard milo modals from invalid query selectors

mokimo:
### Description
We have a bunch of issues with querySelectors which were fixed here https://github.com/adobecom/milo/pull/1348 or here https://github.com/adobecom/milo/pull/1287 

**Question:** Should we find a more generic solution for this querySelector syntax errors ?
**Issue:** We ran into querySelector syntax errors 3x in a brief time frame now
**Proposal:** A milo utility wrapper for `querySelector/querySelectorAll` that adds a try/catch and you can just use `import {safeQuerySelector} from './action.js'` then

Resolves: [MWPW-136672](https://jira.corp.adobe.com/browse/MWPW-136672)

## Test instructions
1. Open a milo page with a modal
2. Change the hash to by pasting  `window.location.hash = '';` in the console
3. No error should appear on the **after** test URLs anymore.
Screenshot 2023-09-26 at 17 32 43

## Test URLs
**BACOM:**
- Before: https://business.stage.adobe.com/fr/customer-success-stories.html?martech=off#milo=&
- After: https://business.stage.adobe.com/fr/customer-success-stories.html?martech=off&milolibs=ims--milo--adobecom#milo=&

**CC:**
- Before: https://main--cc--adobecom.hlx.live/?martech=off#milo=&
- After: https://main--cc--adobecom.hlx.live/?martech=off&milolibs=ims--milo--adobecom#milo=&

**Milo:**
- Before: https://main--milo--adobecom.hlx.page/ch_de/drafts/ramuntea/gnav-refactor?martech=off#milo=&
- After: https://ims--milo--adobecom.hlx.page/ch_de/drafts/ramuntea/gnav-refactor?martech=off#milo=&

### DIFF
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -202,10 +202,10 @@ window.addEventListener('hashchange', (e) => {
     let dialog;
     try {
       dialog = document.querySelector(`.dialog-modal${url.hash}`);
+      if (dialog) closeModal(dialog);
     } catch (error) {
       /* do nothing */
     }
-    if (dialog) closeModal(dialog);
   } else {
     const details = findDetails(window.location.hash, null);
     if (details) getModal(details);
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -198,10 +198,9 @@ export default function init(el) {
 // Click-based modal
 window.addEventListener('hashchange', (e) => {
   if (!window.location.hash) {
-    const url = new URL(e.oldURL);
-    let dialog;
     try {
-      dialog = document.querySelector(`.dialog-modal${url.hash}`);
+      const url = new URL(e.oldURL);
+      const dialog = document.querySelector(`.dialog-modal${url.hash}`);
       if (dialog) closeModal(dialog);
     } catch (error) {
       /* do nothing */
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -199,7 +199,12 @@ export default function init(el) {
 window.addEventListener('hashchange', (e) => {
   if (!window.location.hash) {
     const url = new URL(e.oldURL);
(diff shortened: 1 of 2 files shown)


mokimo:
@3ch023 had this fix open on my local as well 😬 Same issue you just ran into

narcis-radu (approved):
Looks good to me, maybe we can explore `CSS.escape` in the future.

libs/blocks/modal/modal.js
```diff
@@ -199,7 +199,12 @@ export default function init(el) {
...
   if (!window.location.hash) {
     const url = new URL(e.oldURL);
-    const dialog = document.querySelector(`.dialog-modal${url.hash}`);
+    let dialog;
+    try {
+      dialog = document.querySelector(`.dialog-modal${url.hash}`);
+    } catch (error) {
+      /* do nothing */
+    }
     if (dialog) closeModal(dialog);
```
overmyheadandbody, line 208: I guess this can be moved inside the `try` block and have `dialog` as a `const` directly

test/blocks/modals/modals.test.js
```diff
@@ -170,4 +170,10 @@ describe('Modals', () => {
     dialogmodal.classList.remove('commerce-frame');
     expect(window.innerWidth).not.equal(dialogmodal.offsetWidth);
   });
+
+  it('does not error for a modal with a non-querySelector compliant hash', () => {
+    // Test passing, means there was no error thrown
+    window.location.hash = '#milo=&';
```
overmyheadandbody, line 176: Should we not `await` for any modal logic to execute after calling this?

mokimo: Hmm, I tested this, and it should work but given event listeners are async let's make the test more resilient and wait for the hashchange event listener to be triggered.

We can't really expect an event listener to not throw, but if it does throw - the test will fail.

chrischrischris (approved):

PR #1355, opened 2023-09-26
MWPW-136621 | Send to Caas IMS error (Single page Send to Caas from sidekick)

raga-adbe-gh:
Send To CaaS from sidekick triggers a postXDM call with user access token.This token is validated by by the service, with prod as default. However for hlx.page or hlx.live the access token is of stage (based on window.adobeid being currently set). A user-ims-env header needs to be passed appropriately for the service to validate against stage. Change is to set the header as stage in case adobeid.environment is non prod.

Due to another jira MWPW-136658 tested this with localhost by updating below in libs/utils/sidekick.js

```
        const sendToCaasListener = async (e) => {
         const { host, project, ref: branch, repo, owner } = e.detail.data.config;
    -    const { sendToCaaS } = await import('https://milo.adobe.com/tools/send-to-caas/send-to-caas.js');
    +    const { sendToCaaS } = await import('../../tools/send-to-caas/send-to-caas.js');
         sendToCaaS({ host, project, branch, repo, owner }, loadScript, loadStyle);
       };
```

Resolves: [MWPW-136621](https://jira.corp.adobe.com/browse/MWPW-136621)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.live/tools/send-to-caas/bulkpublisher.html
- After: https://mwpw-136658_movesendtocaas--milo--raga-adbe-gh.hlx.live/tools/send-to-caas/bulkpublisher.html
- Before: https://main--cc-pink--adobecom.hlx.live/drafts/raga/caas-fg-card-3
- After: https://main--cc-pink--adobecom.hlx.live/drafts/raga/caas-fg-card-3?milolibs=mwpw-136658_movesendtocaas--milo--raga-adbe-gh


tools/send-to-caas/send-utils.js
```diff
@@ -589,21 +589,24 @@ const getCardMetadata = async (options) => {
 };
 
 const postDataToCaaS = async ({ accessToken, caasEnv, caasProps, draftOnly }) => {
+  const uie = window.adobeid?.environment?.toLowerCase
```
sukamat, line 592: nit: descriptive var name like `userImsEnv` might be better

sukamat (approved):

arshadparwaiz (approved):

tools/send-to-caas/send-utils.js
```diff
@@ -589,21 +589,24 @@ const getCardMetadata = async (options) => {
...
+    && window.adobeid.environment.toLowerCase() !== 'prod' ? { 'user-ims-env': 'stage' } : {};
+
   const options = {
     method: 'POST',
     body: JSON.stringify(caasProps),
     headers: {
       Authorization: `Bearer ${accessToken}`,
       draft: !!draftOnly,
       'caas-env': caasEnv,
+      ...uie,
```
narcis-radu, line 602: The whole part can be written as:
```js
'user-ims-env': window.adobeid?.environment?.toLowerCase() === 'prod' ? '' : 'stage',
```

narcis-radu: I would even pass `'prod'` instead of `''` for the first use-case but now sure whether this is handled in CaaS.

raga-adbe-gh: Thank you @narcis-radu , In order keep existing call unchanged, we were adding user-ims-env for stage only. Can recheck and add as suggested.

tools/send-to-caas/send-utils.js
```diff
@@ -589,21 +589,24 @@ const getCardMetadata = async (options) => {
 };
 
 const postDataToCaaS = async ({ accessToken, caasEnv, caasProps, draftOnly }) => {
+  const userImsEnv = window.adobeid?.environment?.toLowerCase
```
narcis-radu, line 592: No need to check for existence of `toLowerCase`. I have provided some suggestions in my previous comments.

raga-adbe-gh:
Closing this for testing couple scenarios and will reopen post that.

PR #1360, opened 2023-09-27
MWPW-135672 no changes fix

vgoodric:
The value for no manifest is "default|default" but the value for default when you have a manifest is "no changes|manifest name".  This changes "no changes" to "default" to be consistent.

For the before and after links, you can look at the data-mep attribute on `body` or the `daa-lh` values on `header` and all blocks to see the difference.

Resolves: [MWPW-135672](https://jira.corp.adobe.com/browse/MWPW-135672)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/vgoodric/pers-demo?martech=off
- After: https://mwpw-135672-no-changes-fix--milo--adobecom.hlx.page/drafts/vgoodric/pers-demo?martech=off

### DIFF
--- a/libs/blocks/global-footer/global-footer.js
+++ b/libs/blocks/global-footer/global-footer.js
@@ -98,7 +98,7 @@ class Footer {
       await task();
     }
 
-    this.footerEl.setAttribute('daa-lh', `gnav|${getExperienceName()}|footer`);
+    this.footerEl.setAttribute('daa-lh', `gnav|${getExperienceName()}|footer|${document.body.dataset.mep}`);
 
     this.footerEl.append(this.elements.footer);
   };
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -440,7 +440,7 @@ export async function getPersConfig(name, variantLabel, manifestData, manifestPa
     config.selectedVariantName = selectedVariantName;
     config.selectedVariant = config.variants[selectedVariantName];
   } else {
-    /* c8 ignore next */
+    /* c8 ignore next 2 */
     config.selectedVariantName = 'default';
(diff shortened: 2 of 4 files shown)


overmyheadandbody (commented):
I'm assuming there's no way to update _no changes_ to _default_ at the root? And I'm guessing preview stays as is? https://github.com/adobecom/milo/blob/4bf541a73ab87ee10c7b226978407fe33d4a1c12/libs/features/personalization/preview.js#L159-L160

Lastly, I see that the `daa-lh` attribute on the `header` element for the _After_ link is `gnav|homepage_milo|undefined`, am I missing some testing step?

chrischrischris (approved):

vgoodric:
> I'm assuming there's no way to update _no changes_ to _default_ at the root? And I'm guessing preview stays as is?
> 
> https://github.com/adobecom/milo/blob/4bf541a73ab87ee10c7b226978407fe33d4a1c12/libs/features/personalization/preview.js#L159-L160
> 
> Lastly, I see that the `daa-lh` attribute on the `header` element for the _After_ link is `gnav|homepage_milo|undefined`, am I missing some testing step?

Changing the preview feature to use `default` is a more extensive change, but as I thought about it, I think you're right.  It's better to just use default everywhere.

Apologies, I realized that not having martech=off on the test links was causing it to fail the aem-psi-check.  But martech=off also turns off MEP and therefore the impression tracking.  So to test this feature, we need to keep martech on.

vgoodric:
@overmyheadandbody Actually, martech=off does seem to work.  Not sure what the issue was, but I can't reproduce anymore.  Are you still seeing it?

narcis-radu (approved):

PR #1381, opened 2023-10-03
MWPW-136372: Removing the loading screen to improve performance scores

echampio-at-adobe:
* Removing the loading screen to improve performance scores. 

Resolves: [MWPW-136372](https://jira.corp.adobe.com/browse/MWPW-136372)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/delta/app-recommender-test/?martech=off
- After: https://mwpw-136372--milo--echampio-at-adobe.hlx.page/drafts/xiasun/quiz-2/?martech=off

### DIFF
--- a/libs/blocks/quiz/quiz.css
+++ b/libs/blocks/quiz/quiz.css
@@ -1,3 +1,5 @@
+@import url('../../styles/styles.css');
+
 .quiz-page * {
   box-sizing: border-box;
 }
@@ -11,7 +13,7 @@
 }
 
 .quiz {
-  min-height: calc(100vh - 64px); /* window height minus global nav height */
+  min-height: calc(100vh - var(--global-height-nav)); /* window height minus global nav height from styles.css */
 }
 
 .quiz-container {
--- a/libs/blocks/quiz/quiz.css
+++ b/libs/blocks/quiz/quiz.css
@@ -11,7 +11,7 @@
 }
 
 .quiz {
-  min-height: calc(100vh - 131px); /* window height minus global nav + global footer heights */
+  min-height: calc(100vh - 64px); /* window height minus global nav height */
 }
 
 .quiz-container {
--- a/libs/blocks/quiz/quiz.js
+++ b/libs/blocks/quiz/quiz.js
@@ -253,7 +253,7 @@ const App = ({
 
   useEffect(() => {
     const getStringValue = (propName) => {
-      if (!selectedQuestion) return '';
+      if (!selectedQuestion || !selectedQuestion.questions) return '';
       const question = stringQList[selectedQuestion.questions];
       return question?.[propName] || '';
     };
@@ -262,10 +262,6 @@ const App = ({
       loadFragments(fragmentURL);
     }
   }, [selectedQuestion, stringQList]);
-
-  if (!isDataLoaded || !selectedQuestion) {
-    return html``;
-  }
 
   const getStringValue = (propName) => {
     const question = stringQList[selectedQuestion.questions];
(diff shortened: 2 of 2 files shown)


libs/blocks/quiz/quiz.css
```diff
@@ -11,7 +11,7 @@
 }
 
 .quiz {
-  min-height: calc(100vh - 131px); /* window height minus global nav + global footer heights */
+  min-height: calc(100vh - 64px); /* window height minus global nav height */
```
mokimo, line 14: Do we have a var for this?
Also the global navigation can be either with or without breadcrumbs, which changes the height

libs/blocks/quiz/quiz.js
```diff
@@ -254,7 +254,7 @@ const App = ({
 
   useEffect(() => {
     const getStringValue = (propName) => {
-      if (!selectedQuestion) return '';
+      if (!selectedQuestion || !selectedQuestion.questions) return '';
```
mokimo, line 257:
```suggestion
      if (!selectedQuestion?.questions) return '';
```

libs/blocks/quiz/quiz.js
```diff
@@ -264,11 +264,8 @@ const App = ({
     }
   }, [selectedQuestion, stringQList]);
 
-  if (!isDataLoaded || !selectedQuestion) {
-    return html`<div class="quiz-load">Loading</div>`;
-  }
-
   const getStringValue = (propName) => {
+    if (!selectedQuestion || !selectedQuestion.questions) return '';
```
mokimo, line 268:
```suggestion
      if (!selectedQuestion?.questions) return '';
```

libs/blocks/quiz/quiz.js
```diff
@@ -279,43 +276,48 @@ const App = ({
...
-                      onOptionClick=${onOptionClick}
-                      getOptionsIcons=${getOptionsIcons}
-                      handleOnNextClick=${handleOnNextClick}
-                      btnAnalyticsData=${btnAnalytics}/>
+                    <${DecorateBlockForeground} 
+                        heading=${getStringValue('heading')} 
+                        subhead=${getStringValue('sub-head')} 
+                        btnText=${getStringValue('btn')} />
+                  
+                    ${selectedQuestion && selectedQuestion.questions && html`<${GetQuizOption}
```
mokimo, line 296:
```suggestion
                    ${selectedQuestion?.questions && html`<${GetQuizOption} 
```

libs/blocks/quiz/quiz.js
```diff
@@ -279,43 +276,48 @@ const App = ({
     return optionItem && optionItem[prop] ? optionItem[prop] : '';
   };
 
-  return html`<div class="quiz-container">
-                  <${StepIndicator} 
-                    currentStep=${currentStep} 
-                    totalSteps=${totalSteps} 
-                    prevStepIndicator=${prevStepIndicator}
-                    top="${true}" />
+  if (isDataLoaded) {
```
mokimo, line 279: If you do any early return, you don't need to put this into an else/if
```suggestion
  if(!isDataLoaded) return ''
```

mokimo (approved):
Looks good after a couple of nits are addressed!

libs/blocks/quiz/quiz.css
overmyheadandbody: There are actually some variables that could be used, defined here - https://github.com/adobecom/milo/blob/main/libs/styles/styles.css#L4-L6

mokimo:
FYI consider the blackout period before merging.

libs/blocks/quiz/quiz.css
echampio-at-adobe: Thanks for the pointer to styles.css. I imported it take advantage of the variable you mentioned. The quiz designs always request an absolutely minimal header, so I believe we're on with the assumption we're making about the height, especially not that we can take advantage of that shared variable.

libs/blocks/quiz/quiz.js
echampio-at-adobe: Done. Thanks!

fullcolorcoder (approved):

echampio-at-adobe:
> FYI consider the freeze period before merging.

Hi @mokimo. We're just merging to the uar-integration branch for the near future, not to main, so we won't perturb the freeze for MAX.

libs/blocks/quiz/quiz.css
```diff
@@ -11,7 +13,7 @@
 }
 
 .quiz {
-  min-height: calc(100vh - 131px); /* window height minus global nav + global footer heights */
+  min-height: calc(100vh - var(--global-height-nav)); /* window height minus global nav height from styles.css */
```
colloyd, line 16: I'm seeing a permanent scrollbar with these values in chrome, however, if I add a -1px after the var, it seems to resolve it. Maybe we need to adjust this a bit. 

Also, for what it's worth, Chris Millar removed this min-height stuff in his quiz-perf branch due to some cls: 

https://github.com/adobecom/milo/blob/quiz-perf/libs/blocks/quiz/quiz.css#L12

colloyd (changes requested):

PR #1383, opened 2023-10-03
MWPW-137106 use ost 1.14.1

npeltier:
all options formerly attached to a link were not passed through anymore, using latest version of OST fixing similar bug for RTE version fixed it.

Resolves: [MWPW-137106](https://jira.corp.adobe.com/browse/MWPW-137106)

**Test URLs:**
- Before: https://milo.adobe.com/tools/ost?excluded=true
- After: https://mwpw-137106--milo--npeltier.hlx.page/tools/ost?token=...&excluded=true
(you need to copy & paste token in ... token parameter from Before OST console, using `copy(adobeIMS.getAccessToken().token)` command)


mokimo (approved):
Don't forget the blackout period in regards to merging.

overmyheadandbody (approved):
As @mokimo mentioned, this might have to wait until post-Max

npeltier:
i'll wait :) thanks for the reviews

Ruchika4 (approved):

PR #1389, opened 2023-10-05
MWPW-130327 Tag-Based Footer Promo

meganthecoder:
* Update the Footer Promo functionality to allow taxonomy-based promos
* Author with new metadata "footer-promo-type" of "taxonomy"

Resolves: [MWPW-130327](https://jira.corp.adobe.com/browse/MWPW-130327)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/methomas/footer-promo-page
- After: https://methomas-blog-footer-promo--milo--adobecom.hlx.page/drafts/methomas/footer-promo-page?martech=off

Regression:
- Before: https://main--milo--adobecom.hlx.page/drafts/cmillar/footer-fragment-promo-thing
- After: https://methomas-blog-footer-promo--milo--adobecom.hlx.page/drafts/cmillar/footer-fragment-promo-thing?martech=off

**Bacom blog:**
Promo by taxonomy: 
- Before: https://main--bacom-blog--adobecom.hlx.page/drafts/methomas/article-with-footer-promo
- After: https://main--bacom-blog--adobecom.hlx.page/drafts/methomas/article-with-footer-promo?milolibs=methomas-blog-footer-promo

Promo by url (no change): 
- Before: https://main--bacom-blog--adobecom.hlx.page/drafts/methomas/connecting-external-data-sources-to-adobe-experience-manager-guides
- After: https://main--bacom-blog--adobecom.hlx.page/drafts/methomas/connecting-external-data-sources-to-adobe-experience-manager-guides?milolibs=methomas-blog-footer-promo

**DC (regression):**
- Before: https://main--dc--adobecom.hlx.page/acrobat/hub/how-to-convert-heic-to-jpg-on-a-mac
- After: https://main--dc--adobecom.hlx.page/acrobat/hub/how-to-convert-heic-to-jpg-on-a-mac?milolibs=methomas-blog-footer-promo


meganthecoder:
Failing unit test isn unrelated

libs/utils/utils.js
```diff
@@ -721,13 +721,47 @@ function decorateSections(el, isDoc) {
...
+    return false;
+  }
+  return false;
+}
+
+export async function decorateFooterPromo({ locale: { contentRoot } }) {
+  const footerPromoMeta = getMetadata('footer-promo-tag');
+  const tagBased = getMetadata('tag-based-footer-promo');
+  if ((!footerPromoMeta || footerPromoMeta === 'off') && tagBased !== 'on') return;
+  let href;
```
mokimo, line 752: As this is becoming increasingly larger, let's turn this into a separate function and dynamically load it.
```suggestion
export async function decorateFooterPromo({ locale: { contentRoot } }) {
  const footerPromoMeta = getMetadata('footer-promo-tag');
  const tagBased = getMetadata('tag-based-footer-promo');
  if ((!footerPromoMeta || footerPromoMeta === 'off') && tagBased !== 'on') return;
  const { decorateFooterPromo } = await import('../features/footerPromo.js');
  decorateFooterPromo({ locale: { contentRoot } });
```

JasonHowellSlavin (commented):
Looks great, i'll agree with Okan on the module part. The tests look good as well. 

If CC is using it, maybe we could ping them to add some test URLs as well?

meganthecoder:
@Blainegunn If I recall correctly, you were interested in the footer promo functionality, right? Do you have a live page that I could test against for regression?

Blainegunn:
> @Blainegunn If I recall correctly, you were interested in the footer promo functionality, right? Do you have a live page that I could test against for regression?

https://www.adobe.com/acrobat/hub/how-to-convert-heic-to-jpg-on-a-mac.html

meganthecoder:
> If CC is using it, maybe we could ping them to add some test URLs as well?

@JasonHowellSlavin Not sure about CC, but I added the DC link that Blaine gave me. Looks good

libs/features/footer-promo.js
```diff
@@ -0,0 +1,41 @@
...
+    return false;
+  }
+  return false;
+}
+
+export default async function initFooterPromo(contentRoot, urlBasedPromo, tagBasedPromo) {
+  let href;
+  if (urlBasedPromo && urlBasedPromo !== 'off') href = `${contentRoot}/fragments/footer-promos/${urlBasedPromo}`;
+
+  if (tagBasedPromo === 'on') {
```
mokimo, line 31: So tagBasedPromos can overwrite urlBasedPromos?

libs/features/footer-promo.js
```diff
@@ -0,0 +1,41 @@
...
+  } catch (error) {
+    window.lana.log(`Footer Promo - Taxonomy error: ${error}`);
+    return false;
+  }
+  return false;
+}
+
+export default async function initFooterPromo(contentRoot, urlBasedPromo, tagBasedPromo) {
+  let href;
+  if (urlBasedPromo && urlBasedPromo !== 'off') href = `${contentRoot}/fragments/footer-promos/${urlBasedPromo}`;
```
mokimo, line 29:
```suggestion
  let href = (urlBasedPromo && urlBasedPromo !== 'off') && `${contentRoot}/fragments/footer-promos/${urlBasedPromo}`
```

libs/features/footer-promo.js
```diff
@@ -0,0 +1,41 @@
...
+  if (tagBasedPromo === 'on') {
+    const linkFromTag = await getFooterPromoByTag(contentRoot);
+    if (linkFromTag) href = linkFromTag;
+  }
+
+  if (href) {
+    const para = createTag('p', {}, createTag('a', { href }, href));
+    const section = createTag('div', null, para);
+    document.querySelector('main > div:last-of-type').insertAdjacentElement('afterend', section);
+  }
```
mokimo, line 40:
```suggestion
  if (!href) return
  const para = createTag('p', {}, createTag('a', { href }, href));
  const section = createTag('div', null, para);
  document.querySelector('main > div:last-of-type').insertAdjacentElement('afterend', section);
```

libs/features/footer-promo.js
```diff
@@ -0,0 +1,41 @@
+import { createTag } from '../utils/utils.js';
+
+async function getFooterPromoByTag(contentRoot) {
+  const NAME = 'Name';
+  const FOOTER_PROMO_LINK = 'Footer Promo Link';
+  const taxonomyUrl = `${contentRoot}/taxonomy.json`;
+  const tagsOnPage = [...document.head.querySelectorAll('meta[property="article:tag"]')].map((el) => el.content);
+
+  if (!tagsOnPage.length) return false;
```
mokimo, line 9: Since the expected return value is `primaryTag[FOOTER_PROMO_LINK];` and not a Boolean (true) - I would just `return` (undefined) or `null` if you really want to indicate nothing was returned.

libs/utils/utils.js
```diff
@@ -923,10 +923,10 @@ function decorateMeta() {
   });
 }
 
-function decorateDocumentExtras(config) {
+async function decorateDocumentExtras(config) {
   decorateMeta();
   decorateHeader();
-  decorateFooterPromo(config);
+  await decorateFooterPromo(config);
```
mokimo, line 929: Since this is usually pushed out to the button of the page, does it really make sense to await this?

mokimo (commented):
Couple of comments and questions!

libs/features/footer-promo.js
meganthecoder: Yes, when it's turned on, it will supersede the urlBasedPromo

libs/utils/utils.js
meganthecoder: Yeah it needs to finish before the fragments are decorated. (You'll notice that the outcome of `decorateFooterPromo` is that it adds a fragment link onto the page). If you think it's a problem, I'll need to do something custom to load the fragment and avoid a race condition

mokimo: This runs before the LCP.
Doing something custom is worth it IMO, as is happening in the most crucial period of the whole rendering setup.

meganthecoder: @auniverseaway @chrischrischris @rgclayton Thoughts on this?

rgclayton: This comes with the same performance penalty as a regular fragment, correct? Meaning, this would be a known use case.

One potential thought, since this is loaded at the bottom of the page, would be to use [loadDeferred](https://github.com/adobecom/milo/blob/main/libs/utils/utils.js#L864). That means your fragments are all ready decorated so another call would need to be made. 

Another thought is taking a similar approach as inline fragments and using  [processSection](https://github.com/adobecom/milo/blob/main/libs/utils/utils.js#L973C16-L973C30)

mokimo: A difference is that we now also load the `taxonomy.json` before LCP

meganthecoder: @mokimo @rgclayton Thanks for the feedback! I talked with @chrischrischris and I think we've gotten to a good solution. I moved `decorateFooterPromo` into `documentPostSectionLoading` so it won't block LCP. Then I'm importing `fragment.js` and loading the fragment.

Please take another look and let me know what you think

mokimo (approved):

libs/utils/utils.js
mokimo: Perfect! 👍 Thanks for following up on this - looking great

JasonHowellSlavin (approved):
Great stuff!

overmyheadandbody (approved):

meganthecoder:
@chrischrischris @rgclayton I'd appreciate if you could take another look at this. I need an Admin's approval as a code owner

Blainegunn (approved):
looks great @meganthecoder thanks for considering dc

auniverseaway:
I'm not in love with a few things in this PR, but it's probably fine. I think it could be simpler and more consistent. Maybe I'm also not understanding the use cases. A few observations:

```
footer-promo-tag
tag-based-footer-promo
```
From CC's perspective, they are using a "tag". In reality, yours is "taxonomy" based.

In this light I would expect:

```
footer-promo-tag
footer-promo-type | <tag>, <taxonomy>
```

The expectation here would be that `footer-promo-tag` can be populated with a singular "tag" and if you want a different style (taxonomy based), you opt into that using the type. I don't know what I'm supposed to put in `tag-based-footer-promo` but all the examples show it as a toggle of off / on.

The other way you could do this is that if it truly is on / off. I would probably try to cut down to one meta tag and have a special keyword that means taxonomy lookup...

```
footer-promo-tag | your-specific-tag
OR
footer-promo-tag | taxonomy (implying a taxonomy.json lookup)
```

Super nit:
I appreciate the "off" addition to `footer-promo-tag` but I would avoid adding this complexity until someone actually says, "I have an entire tree of footer promos, but this _one_ page needs to not have it." I believe `""` can handle this and the absence of the meta tag should be the guide for "off" in this case. I say this mostly so the utils feature flag can be simpler.

libs/utils/utils.js
```diff
@@ -940,6 +939,9 @@ async function documentPostSectionLoading(config) {
     const { default: loadGeoRouting } = await import('../features/georoutingv2/georoutingv2.js');
     loadGeoRouting(config, createTag, getMetadata, loadBlock, loadStyle);
   }
+
+  await decorateFooterPromo(config);
```
chrischrischris, line 943: Since you now load fragment.js directly, I think we can remove the `await`?   It's preferable for the code below to keep running while footer-promo.js loads.

libs/features/footer-promo.js
```diff
@@ -0,0 +1,43 @@
+import { createTag } from '../utils/utils.js';
```
chrischrischris, line 1: Since we already import from utils.js, I think it's better to also import `getConfig` and get the contentRoot that way instead of having to pass it in.

chrischrischris (changes requested):

meganthecoder:
> I'm not in love with a few things in this PR, but it's probably fine. I think it could be simpler and more consistent. Maybe I'm also not understanding the use cases. A few observations:

@auniverseaway I definitely understand the confusion since "tag" sort of means two different things here. What if I changed `tag-based-footer-promo` to `taxonomy-based-footer-promo` or `taxonomy-footer-promo`?

Your second suggestion is elegant, but I don't think it works for our use case. The bacom blog stakeholders want to have a url-based option (ie `footer-promo-tag`) as a backup if the taxonomy tag doesn't have a footer promo associated with it. That doesn't really work if there's only one metadata tag for both. It's an either-or situation.

The first suggestion doesn't really make sense to me. If we're going to use two metadata tags anyway, why not keep it how I wrote it? The "taxonomy" version of the footer promo is only an on/off value, so it'd never use `footer-promo-tag` and then there'd be the extra step of having to add `footer-promo-type | <tag>` to existing pages.

Renaming it makes more sense to me but let me know if you have another suggestion.

libs/utils/utils.js
meganthecoder: Good catch! Updated

chrischrischris (approved):

auniverseaway:
As mentioned, I would prefer we change the name to be more consistent with the existing naming convention...

```
footer-promo-tag
footer-promo-taxonomy
```

This is in spirit of how css props are classified...

```
border-radius
border-width
border-style
```

CSS doesn't do:

```
border-radius
width-border
style-border
```

I still dislike `on/off` vs `type` as it is not future proof for other decision engines... are we going to have `locale-based-footer-promo = on/off` next? I think this paints us into a corner, and as far as fallback to path, I think using type doesn't preclude you from doing that. If you say, `type = taxonomy` and there's `footer-promo-tag,` it falls back to it. If there's not, it doesn't. I'm not saying working with the existing content (`footer-promo-tag` only) isn't clumsy, but I think there are ways to mitigate the clumsyness and infer intent with that existing content.

Feel free to ship this as is, but do know my observations above.

meganthecoder:
@auniverseaway @JasonHowellSlavin @mokimo @overmyheadandbody @chrischrischris @Blainegunn I've updated a few things based on the Chris M's authoring experience feedback. Please take a second look, thanks!

mokimo (approved):

mokimo:
Don't forget to add this to the docs once merged: https://milo.adobe.com/docs/authoring/metadata

JasonHowellSlavin (approved):

PR #1393, opened 2023-10-06
MWPW-137448 - Point Stage to Acom Reviews on www.stage.adobe.com

draganatrajkovic:
* With this fix pages in stage will fetch Acom Reviews data from www.stage.adobe.com

Resolves: [MWPW-137448](https://jira.corp.adobe.com/browse/MWPW-137448)

**Test URLs:**
- Before: https://www.adobe.com/acrobat/online/crop-pdf.html
- After: https://www.adobe.com/acrobat/online/crop-pdf.html?milolibs=mwpw-137448-review-from-stage--milo--draganatrajkovic

### DIFF
--- a/libs/blocks/review/review.js
+++ b/libs/blocks/review/review.js
@@ -121,6 +121,11 @@ export default async function init(el) {
   loadStyle(`${base}/ui/page/page.css`);
   const metaData = getMetaData(el);
   const strings = getStrings(metaData);
+
+  if (window.location.host === 'www.stage.adobe.com') {
+    strings.postUrl = strings.postUrl.replace('www.adobe', 'www.stage.adobe');
+  }
+
   removeMetaDataElements(el);
 
   const app = html` <${App} rootEl=${el} strings="${strings}" /> `;


mokimo (commented):
Should this work just for stage.adobe.com? What about `hlx.page` / `hlx.live` and other consumers such as business.adobe.com?

Also:
- There are some ESlint errors in the file, would be great if you can fix them
- This should have a unit test

narcis-radu:
Good catch, @mokimo! It looks like https://main--dc--adobecom.hlx.page/acrobat/online/crop-pdf?milolibs=mwpw-137448-review-from-stage--milo--draganatrajkovic is still sending reviews to www.adobe.com.
@draganatrajkovic - I think we need to fix for all lower environments.

libs/blocks/review/review.js
```diff
@@ -121,6 +121,11 @@ export default async function init(el) {
   loadStyle(`${base}/ui/page/page.css`);
   const metaData = getMetaData(el);
   const strings = getStrings(metaData);
+
+  if (window.location.host === 'www.stage.adobe.com') {
```
Blainegunn, line 125: @narcis-radu @mokimo @draganatrajkovic 
would this make everyone happy?
```
  if (window.location.host != 'www.adobe.com') { ... }

Comment (mokimo) [reply libs/blocks/review/review.js:125]: If `stage.adobe.com` and `adobe.com` is enough sure. I'm don't have the context whether that works for all use-cases/consumers.

Who's the code-owner for the review block anyway? @nkthakur48 Do you know?

Comment (nkthakur48) [reply libs/blocks/review/review.js:125]: @mokimo Could we explore other approaches (like specifying an optional stage submission URL through authoring; or swapping domain if Milo somewhere has config details for all consuming domains, stage and prod)? Personally I think harcoding checks for domain name will reduce the re-usability of this component

CC: @Blainegunn @mokimo

Comment (draganatrajkovic) [reply libs/blocks/review/review.js]: @nkthakur48 @mokimo @Blainegunn 
Can you please check my fix?

File: libs/blocks/review/utils/utils.js [hunk 2]
```diff
@@ -9,4 +9,8 @@
     && ev.nativeEvent.webkitForce !== undefined
     && ev.nativeEvent.webkitForce === 0);
 
-export { addToAverage, isKeyboardNavigation };
+const checkPostUrl = (postUrl, env) => (
+  env?.name !== 'prod' && postUrl?.includes('adobe.com') ? postUrl.replace('adobe.com', 'stage.adobe.com') : postUrl
```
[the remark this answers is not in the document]
nkthakur48: Minor: @draganatrajkovic Looks to be a valid string sanitisation error. Could you maybe try `postUrl.origin.match(/adobe\.com$)/)` as check before replacing the substring. Also ensuring that you are only replacing the origin and not any other parts of the URL could be done, something like:
```
const dataStoreEndPoint = new URL(postUrl)
const replacedOrigin = dataStoreEndPoint.origin.replace(/adobe\.com$/, 'stage.adobe.com')
const newDataStoreEndPoint = replacedOrigin + dataStoreEndPoint.pathname + dataStoreEndPoint.search + dataStoreEndPoint.hash
```

libs/blocks/review/review.js
```diff
@@ -54,9 +55,6 @@ const App = ({ rootEl, strings }) => html`
       visitorId=${getVisitorId()}
       reviewPath=${getReviewPath(strings.postUrl)}
       initialValue=${strings.initialValue}
-      onRatingSet=${({ rating, comment }) => {}}
-      onRatingHover=${({ rating }) => {}}
-      onReviewLoad=${({ hasRated, rating }) => {}}
```
mokimo, line 59: If we remove those and they are never used, should we also remove them from the rest of the children?

mokimo (approved):
Looks good code wise, would be nice if you can address Nishants feedback.

Also, the PR description doesn't have any `hlx.page` `hlx.live` or production URLs so ensure that those receive proper QA before merging.

libs/blocks/review/review.js
draganatrajkovic: Thank you @mokimo @nkthakur48
I fixed eslint by commenting out argument that are not used because I am not sure if these props are planed for some future usage.
If you think they should be removed, I can do that right away.

mokimo:
Would be nice if you also find out and add the respective CODEOWNERS of this block: https://github.com/adobecom/milo/blob/main/CODEOWNERS

draganatrajkovic:
> Would be nice if you also find out and add the respective CODEOWNERS of this block: https://github.com/adobecom/milo/blob/main/CODEOWNERS

Done :)

libs/blocks/review/review.js
narcis-radu: If the callbacks will be used in the future we can always check the code history and add them. Please remove them if they're not used.

libs/blocks/review/review.js
```diff
@@ -115,15 +116,16 @@ const removeMetaDataElements = (el) => {
...
 
 export default async function init(el) {
-  const { miloLibs, codeRoot } = getConfig();
+  const { miloLibs, codeRoot, env } = getConfig();
   const base = miloLibs || codeRoot;
 
   loadStyle(`${base}/ui/page/page.css`);
   const metaData = getMetaData(el);
   const strings = getStrings(metaData);
+  strings.postUrl = strings.postUrl ? checkPostUrl(strings.postUrl, env) : strings.postUrl;
```
narcis-radu, line 125: I'm trying to understand this line: if the `strings.postUrl` is undefined or empty then you are assigning it back. Do we even need to init the application in this case? I see that `getReviewPath` will handle the error, but don't think we need to go that far if we know the URL is not valid.

libs/blocks/review/utils/utils.js
```diff
@@ -9,4 +9,14 @@ const isKeyboardNavigation = (ev) => (ev.clientX === 0 && ev.clientY === 0)
     && ev.nativeEvent.webkitForce !== undefined
     && ev.nativeEvent.webkitForce === 0);
 
-export { addToAverage, isKeyboardNavigation };
+const checkPostUrl = (postUrl, env) => {
+  const url = new URL(postUrl);
```
narcis-radu, line 13: maybe we need some try/catch here to ensure we catch errors when URLs are not valid fomat.

libs/blocks/review/utils/utils.js
```diff
@@ -9,4 +9,14 @@ const isKeyboardNavigation = (ev) => (ev.clientX === 0 && ev.clientY === 0)
...
+  if (env?.name !== 'prod' && url.origin.match(/adobe\.com$/)) {
```
narcis-radu, line 14: url pattern can be cached in a constant

libs/blocks/review/utils/utils.js
```diff
@@ -9,4 +9,14 @@ const isKeyboardNavigation = (ev) => (ev.clientX === 0 && ev.clientY === 0)
...
+    const newOrigin = url.origin.replace(/adobe\.com$/, 'stage.adobe.com');
+    const newUrl = newOrigin + url.pathname + url.search + url.hash;
```
narcis-radu, line 16: you can return directly

draganatrajkovic:
> 



> There's a failing unit test after the latest changes
> 
> > Review Comp > could be initialized (with all missing fields)
> > Error: Invalid postUrl. Initialization aborted.
> > at init (libs/blocks/review/review.js:131:11)
> > at o. (test/blocks/review/review.test.js:23:11)

@mokimo Right. Yes, I left the comment here: https://github.com/adobecom/milo/pull/1393#discussion_r1376523760.
Wasn't sure if the approach without initialization is approved, but I will remove that part of the test right away.

PR #1397, opened 2023-10-09
MWPW-134799: add scheduled MEP preview capability

3ch023:
Add 'preview on' and 'preview off' columns to the MEP manifest.
This will enhance the MEP manifest with the ability to preview scheduled changes.
The scheduled on/off time can be set using [the later expression](https://bunkat.github.io/later/index.html), always in the UTC timezone. 
An automation tool, that will be developed later, can use on/off time in manifests to create the crontab.xlsx schedule.

Resolves: [MWPW-134799](https://jira.corp.adobe.com/browse/MWPW-134799)
Discussion: [MEP to show scheduled manifest](https://github.com/orgs/adobecom/discussions/1384)

[Sample Page doc](https://adobe.sharepoint.com/:w:/r/sites/adobecom/_layouts/15/Doc.aspx?sourcedoc=%7BF1E673D9-A70D-4FBD-9AA9-5673A3929FA5%7D&file=special-offers.docx&action=default&mobileredirect=true)
[Sample manifest](https://adobe.sharepoint.com/:x:/r/sites/adobecom/_layouts/15/Doc.aspx?sourcedoc=%7BA15D6A6D-257A-45DB-B80C-777AD698DE80%7D&file=mep-manifest1.xlsx&action=default&mobileredirect=true)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/mariia/testpage?martech=off
- After: https://mwpw-134799-mep-preview--milo--adobecom.hlx.page/drafts/mariia/testpage?martech=off

image

image


libs/features/personalization/personalization.js
```diff
@@ -246,6 +248,26 @@ export function parseConfig(data) {
       variants[vn] = { commands: [] };
     });
 
+    const laterToDate = (laterExpression) => {
+      if (!laterExpression) return null;
+      const regex = /at (\d{1,2}):(\d{1,2}) on the (\d{1,2}) day of (\w+) in (\d{4})/;
```
npeltier, line 253: could you call it with more obvious name of what it means, not what type it is? `scheduleFormat` ?

libs/features/personalization/personalization.js
```diff
@@ -246,6 +248,26 @@ export function parseConfig(data) {
...
+      const matches = laterExpression.match(regex);
+      if (matches && matches.length === 6) {
```
npeltier, line 255: what is the purpose of `matches.length === 6`? to access matches[X] with a peaceful mind? normally regex matching is enough as it implies 5 matching groups.

libs/features/personalization/personalization.js
```diff
@@ -246,6 +248,26 @@ export function parseConfig(data) {
...
+        const monthNumber = new Date(Date.parse(`${matches[4]}1, 2012`)).getMonth();
```
npeltier, line 256: wouldn't we be better with an array of month string somewhere, or easier, with a format asking for a simpler date format?

libs/features/personalization/personalization.js
npeltier: could we make that format shorter and more standard ? [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601)? YYYY-mm-ddThh:mm? This should save us a few bytes and be simpler

3ch023: >could you call it with more obvious name of what it means, not what type it is? scheduleFormat ?
sure, will rename it

>could we make that format shorter and more standard ? [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601)? YYYY-mm-ddThh:mm? This should save us a few bytes and be simpler
I can, but then we have 1 format in crontab.xlsx and another format in the manifest file. I chose to use same format that the Franklin team is using because then you can compare 2 dates in 2 places more quickly. 
not sure what is better.

libs/features/personalization/personalization.js
3ch023: true, its redundant, will remove it

libs/features/personalization/personalization.js
npeltier: will the users that edit both be the same? one is fetched on client side, one not. I think this makes sense not aligning

vgoodric (changes requested):
@3ch023 I want to review this feature with you more closely.  Might be easier to meet.

3ch023:
> @3ch023 I want to review this feature with you more closely. Might be easier to meet.

sure, let's sync offline

3ch023:
closed, since we decided on a different solution - will create a discussion for that.

PR #1399, opened 2023-10-10
MWPW-126972: Removes SendToCaas option from dev and preview

cmiqueo:
Removes _Send To Caas_ option from the _Sidekick > Tools_ menu in Dev and Preview environments

Resolves: [MWPW-126972](https://jira.corp.adobe.com/browse/MWPW-126972)

Before URL: n/a
After URL: n/a


overmyheadandbody (commented):
Even though it's a small PR, we should still have proper labels and before/after links, even if the change shouldn't have any impact on PSI scores. Keeping checks green is always a good idea 🙂

mokimo:
To add to Rares' point, more info here: https://github.com/orgs/adobecom/discussions/997

chrischrischris (approved):

sheridansunier (approved):

PR #1401, opened 2023-10-11
MWPW-135873 Compress the caas hash strings

chrischrischris:
The hash string is now compressed using gzip: the hash length is reduced by ~60%

The new compressed hashes start with `~~` in order to differentiate them from the original non-compressed format.

The configurator at /tools/caas will now only generate the v2 hashes, but pages can use links with either format.

Resolves: [MWPW-135873](https://jira.corp.adobe.com/browse/MWPW-135873)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/cpeyer/caas-hash-link?martech=off
- After: https://caaslink-135873--milo--adobecom.hlx.page/drafts/cpeyer/caas-hash-link?martech=off


libs/blocks/caas-config/caas-config.js
```diff
@@ -36,13 +43,15 @@ const updateObj = (obj, defaultObj) => {
 
 const isValidUuid = (id) => /^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$/.test(id);
 
-const getHashConfig = () => {
+const getHashConfig = async () => {
```
JasonHowellSlavin, line 46: I don't see an await in this function, `async` needed?

libs/blocks/caas-config/caas-config.js
```diff
@@ -904,6 +946,17 @@ const Configurator = ({ rootEl }) => {
...
+    const setInitialState = async () => {
+      const initialState = await getInitialState();
+      dispatch({ type: 'SET_STATE', value: initialState });
+    };
+
+    if (!Object.keys(state).length) {
+      setInitialState();
+    }
+  }, [state]);
+
```
JasonHowellSlavin, line 959: Could you simplify this by putting an empty `[]` instead of `[state]` so the `useEffect` runs only once? Might be able to omit the check for state length on line 955 if so.

libs/blocks/caas-config/caas-config.js
chrischrischris: `getHashConfig` now can either be async (for v2 hashes) or sync (for v1 hashes).  I think for clarity, since it's possible that it will be async this function should be marked as async?  Otherwise you have to just know.  Thoughts?

chrischrischris: Nevermind, am updating the code to show why it's async.

rgclayton (approved):
Is the only way to verify that the truncation is/isn't happening is to go through the loc rollout process?

test/blocks/caas/caas.test.html
```diff
@@ -87,6 +99,16 @@
           expect(loadStyle.callCount).to.equal(1);
           expect(caasEl).to.equal(document.getElementById('caas'));
         });
+
+        it('it can use the v2 hash format', async () => {
+          const a = document.getElementById('v2-caaslink');
+          await init(a);
+          await waitForElement('#caas');
+          await delay(5);
```
mokimo, line 107: Shoul we use a fake timer instead? I'm fine with the delay as well if it's deterministic such as waiting for an eventListener execution and doesn't lead to flakiness.

mokimo (approved):
Fancy, one test failing ensure it's running through before merging 
Culprit test: `Can copy the config to a hash`

PR #1403, opened 2023-10-11
MWPW-135119: Redirect to result pages without loading results.json

JackySun9:
1. With this change, author can add URL into question.json, when hit the next button, the page will redirect to that URL directly and skip loading results.json
2. This change will shortcut the question flow and avoid any unnecessary steps 

Resolves: [MWPW-135119](https://jira.corp.adobe.com/browse/MWPW-135119)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/dxtreme/dx-solution-finder/?martech=off
- After: https://mwpw-135119--milo--jackysun9.hlx.page/drafts/xiasun/dx-solution-finder/?martech=off

Before:
Screenshot 2023-10-11 at 4 38 28 PM

After:
Screenshot 2023-10-16 at 8 54 25 AM

Author has to provide a valid URL like the following examples:
http://example.com
https://example.net
http://test-site.org/path?query=value#fragment


libs/blocks/quiz/utils.js
```diff
@@ -5,6 +5,7 @@ import { getConfig } from '../../utils/utils.js';
 const QUESTIONS_EP_NAME = 'questions.json';
 const STRINGS_EP_NAME = 'strings.json';
 const RESULTS_EP_NAME = 'results.json';
+const VALID_URL_RE = /(http(s)?:\/\/.)?(www\.)?[-a-zA-Z0-9@:%._+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_+.~#?&//=]*)/;
```
fullcolorcoder, line 8: We should either expect this to be true from authors or have checks in more meaningful places like before publish.

fullcolorcoder (changes requested):

libs/blocks/quiz/utils.js
echampio-at-adobe: Hi @JackySun9. I would ditch the prefixHttps method. As @fullcolorcoder suggests, just expect a valid URL. To that end, if you update the regex as follows, I believe it will enforce https and lowercase characters. 
`
/^(http(s):\/\/.)[-a-z0-9@:%._+~#=]{2,256}\.[a-z]{2,6}\b([-a-z0-9@:%_+.~#?&//=]*)/
`

echampio-at-adobe (changes requested):

echampio-at-adobe (approved):

fullcolorcoder (approved):

colloyd (approved):

PR #1411, opened 2023-10-13
MWPW-137755 Fix RTL text issue -how-to ordered lists

puruadobe:
Fixing how to ordered lists for RTL language pages.

Resolves: [MWPW-137755](https://jira.corp.adobe.com/browse/MWPW-137755)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.live/mena_ar/drafts/doreswam/how-to?martech=off
- After: https://mwpw-137755--milo--puruadobe.hlx.live/mena_ar/drafts/doreswam/how-to?martech=off

### DIFF
--- a/libs/blocks/how-to/how-to.css
+++ b/libs/blocks/how-to/how-to.css
@@ -44,7 +44,7 @@
 
 html[dir="rtl"] .how-to ol > li::before {
   margin-left: 20px;
-  margin-right: inherit;
+  margin-right: 0;
 }
 
 .how-to-heading {
--- a/libs/blocks/how-to/how-to.css
+++ b/libs/blocks/how-to/how-to.css
@@ -42,6 +42,11 @@
   line-height: 42px;
 }
 
+html[dir="rtl"] .how-to ol > li::before {
+  margin-left: 20px;
+  margin-right: inherit;
+}
+
 .how-to-heading {
   grid-area: heading;
 }


puruadobe:
Please note that the Lighthouse score is low because of the region selector modal. This is just an example page in Milo for RTL language fix of how to block, https://main--milo--adobecom.hlx.live/mena_ar/drafts/doreswam/how-to?martech=off 

DC page where the fix is required --> https://www.stage.adobe.com/mena_ar/acrobat/online/ppt-to-pdf.html?milolibs=mwpw-137755--milo--puruadobe.

arshadparwaiz (approved):

echampio-at-adobe (approved):
The result looks good to me in your mena_ar test page. If I remove the rtl attribute from the root html element, the regular padding comes back into force, which is also as it should be. 
Screenshot 2023-10-15 at 4 47 22 PM

puruadobe:
@JFernandezAdobe please verify the fix and approve.

libs/blocks/how-to/how-to.css
```diff
@@ -42,6 +42,11 @@
   line-height: 42px;
 }
 
+html[dir="rtl"] .how-to ol > li::before {
+  margin-left: 20px;
+  margin-right: inherit;
```
overmyheadandbody, line 47: I'd argue that this should be set to its implicit value, which in this case is `0`. Otherwise, adding margins to parents might lead to unexpected results here.

overmyheadandbody (approved):

mokimo (approved):

libs/blocks/how-to/how-to.css
echampio-at-adobe: Hi @overmyheadandbody. Using the initial keyword, rather than specifying 0, would work too, wouldn't it?

puruadobe: Thanks @overmyheadandbody & @echampio-at-adobe. I will update the initial value with `0` and close this conversation.

JFernandezAdobe:
> @JFernandezAdobe please verify the fix and approve.

The fix is looking good in DC pages. I left a comment in MWPW-137755.

puruadobe:
@ryanmparrish need your approval on this, since you are the code owner, please. :)

ryanmparrish (commented):
What about updating the current property to 'margin-inline-end'?
This does the same thing w/ less code. 
```
.how-to ol > li::before {
  content: counter(step-counter);
  margin-inline-end: 20px;
  font-size: 28px;
  line-height: 42px;
}

ryanmparrish (approved):
Works good as is. My comment above is food for thought. Good w/ either approach.

PR #1416, opened 2023-10-17
MWPW-137970: make MEP prelinks relative

yesil:
Makes the prelink href relative for personalization fragments.

Resolves: [MWPW-137970](https://jira.corp.adobe.com/browse/MWPW-137970)

**Test URLs:**
- After: https://main--homepage--adobecom.hlx.page/homepage/index-loggedout?milolibs=mwpw-137970--milo--yesil&mep=%2Fhomepage%2Ffragments%2Fmep%2Fhp.json--target-dc

### DIFF
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -50,14 +50,21 @@ const DATA_TYPE = {
 };
 
 const createFrag = (el, url, manifestId) => {
-  const a = createTag('a', { href: url }, url);
+  let href = url;
+  try {
+    const { pathname, hash } = new URL(url);
+    href = `${pathname}${hash}`;
+  } catch {
+    // ignore
+  }
+  const a = createTag('a', { href }, url);
   if (manifestId) a.dataset.manifestId = manifestId;
   let frag = createTag('p', undefined, a);
   const isSection = el.parentElement.nodeName === 'MAIN';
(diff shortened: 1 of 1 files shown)


libs/features/personalization/personalization.js
```diff
@@ -50,14 +50,21 @@ const DATA_TYPE = {
 };
 
 const createFrag = (el, url, manifestId) => {
-  const a = createTag('a', { href: url }, url);
+  let href = url;
+  try {
+    const { pathname, search, hash } = new URL(url);
```
yesil, line 55: `normalizePath` in the manifest would drop the search and hash in the url.

chrischrischris: Fragment links don't use hash & search, so we should use normalizePath

chrischrischris (approved):

overmyheadandbody (approved):
The 'after' link is still fetching resources from .page, am I missing a testing step?

chrischrischris:
I've updated the after link to show the fixed code path.   Verified that it is now a relative link instead of absolute.

yesil:
@overmyheadandbody tested with a local override, otherwise it is not possible to test on adobe.com directly AFAIK.

honstar (approved):

PR #1421, opened 2023-10-17
MWPW-137993 Setup global deferred promise

chrischrischris:
For external code (martech) they want to know when the deferred event fires, but sometimes it fires before their code loads.

This adds a global promise at `window.milo.deferredPromise` that will resolve at the same time as the milo deferred event fires.

Resolves: [MWPW-137993](https://jira.corp.adobe.com/browse/MWPW-137993)

**Test URLs:**
- After: https://137993-deferred-promise--milo--adobecom.hlx.page/?martech=off


mokimo (approved):
We could consider deprecating the deferred event in this case, since the promise provides a more reliable approach

libs/utils/utils.js
```diff
@@ -1006,10 +1021,13 @@ async function processSection(section, config, isDoc) {
 
 export async function loadArea(area = document) {
   const isDoc = area === document;
+  let resolveDeferredFn;
 
   if (isDoc) {
+    window.milo = {};
```
overmyheadandbody, line 1027: Nit: should we declare this inside `setupDeferredPromise`, since we don't really use it here? Also, should it be future-proofed a bit by declaring it as `window.milo = window.milo || {}`? Maybe this global namespace property will be used by other pieces of logic than happen before, this would just override everything.

libs/utils/utils.js
```diff
@@ -861,9 +861,24 @@ export function scrollToHashedElement(hash) {
...
+
+export async function loadDeferred(area, blocks, config, resolveDeferred) {
   const event = new Event(MILO_EVENTS.DEFERRED);
   area.dispatchEvent(event);
+
+  if (area !== document) {
+    return;
+  }
+
+  resolveDeferred(true);
```
overmyheadandbody, line 880: Do we need this differentiation between event and Promise based on whether the area is a document? Could we not call `loadDeferred` inside the `loadArea` method only if the `isDoc` condition is met?

test/utils/utils.test.js
```diff
@@ -234,6 +234,15 @@ describe('Utils', () => {
       expect(newTabLink.href).to.equal('https://www.adobe.com/test');
     });
 
+    it('Sets up milo.deferredPromise', async () => {
+      window.milo = {};
```
overmyheadandbody, line 238: If you decide to move the `window.milo` declaration as previously suggested, this would no longer be needed. Instead, you could check for its existence after calling `setupDeferredPromise`

overmyheadandbody (approved):

libs/utils/utils.js
chrischrischris: I encountered an issue with doing `window.milo ??= {}` because I was testing on the milo homepage which has:

`<h1 id="milo">milo</h1>`

This means that `window.milo` actually is populated and points to that element.  So now when running the above code, `window.milo` isn't an empty object, but an h1 element.

By declaring it here, I'm pretty sure that there wouldn't be anyplace that declares `window.milo` before this runs.

Also we want to be very strict about what we allow in this object.  It should *only* be used for values that have to be shared cross-script.  In this case it's for martech to be able to get the deferred status.

libs/utils/utils.js
rgclayton: [Tabs](https://github.com/adobecom/milo/blob/main/libs/blocks/tabs/tabs.js#L95) follows a similar pattern and uses this event too.

rgclayton:
Verified on [bacom](https://main--bacom--adobecom.hlx.page/?milolibs=137993-deferred-promise) and [milo](https://137993-deferred-promise--milo--adobecom.hlx.page/).

image

PR #1423, opened 2023-10-17
MWPW-137999 Marquee deskop only issue

seanchoi-dev:
This will fix the js error that leading missing desktop-only class

Resolves: [MWPW-137999](https://jira.corp.adobe.com/browse/MWPW-137999)

**Test URLs:**
- Before: https://main--bacom--adobecom.hlx.page/drafts/iscandar/adobe-experience-manager
- After: https://main--bacom--adobecom.hlx.page/drafts/iscandar/adobe-experience-manager?milolibs=marquee-deskop-only-issue

Issue with tablet:
image

### DIFF
--- a/libs/blocks/marquee/marquee.js
+++ b/libs/blocks/marquee/marquee.js
@@ -71,7 +71,8 @@ export default function init(el) {
   const foreground = children[children.length - 1];
   if (children.length > 1) {
     children[0].classList.add('background');
-    decorateBlockBg(el, children[0], true);
+    const useHandleFocalpoint = true;
+    decorateBlockBg(el, children[0], { useHandleFocalpoint });
   }
   foreground.classList.add('foreground', 'container');
   const headline = foreground.querySelector('h1, h2, h3, h4, h5, h6');
--- a/libs/blocks/marquee/marquee.js
+++ b/libs/blocks/marquee/marquee.js
@@ -71,7 +71,7 @@ export default function init(el) {
   const foreground = children[children.length - 1];
   if (children.length > 1) {
     children[0].classList.add('background');
-    decorateBlockBg(el, children[0]);
+    decorateBlockBg(el, children[0], true);
   }
   foreground.classList.add('foreground', 'container');
   const headline = foreground.querySelector('h1, h2, h3, h4, h5, h6');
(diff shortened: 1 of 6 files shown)


auniverseaway (approved):

libs/utils/decorate.js
```diff
@@ -88,12 +88,13 @@ export function handleFocalpoint(pic, child, removeChild) {
...
     if (removeChild) childData.forEach((c) => c.nodeType === Node.TEXT_NODE && c.remove());
   }
+  if (!text) return;
   const directions = text.trim().toLowerCase().split(',');
   const [x, y = ''] = directions;
   image.style.objectPosition = `${x} ${y}`;
 }
 
-export async function decorateBlockBg(block, node) {
+export async function decorateBlockBg(block, node, addHandleFocalpoint = false) {
```
chrischrischris, line 97: I know I initially suggested this, but let's change it to be an object instead of bool. Also nit change to be `useHandleFocalpoint`
```suggestion
export async function decorateBlockBg(block, node, { useHandleFocalpoint = false } = {}) {
```

chrischrischris (changes requested):

libs/blocks/marquee/marquee.js
```diff
@@ -71,7 +71,8 @@ export default function init(el) {
   const foreground = children[children.length - 1];
   if (children.length > 1) {
     children[0].classList.add('background');
-    decorateBlockBg(el, children[0]);
+    const useHandleFocalpoint = true;
+    decorateBlockBg(el, children[0], { useHandleFocalpoint });
```
chrischrischris, line 75:
```suggestion
    decorateBlockBg(el, children[0], { useHandleFocalpoint: true });
```

libs/utils/decorate.js
```diff
@@ -105,7 +106,8 @@ export async function decorateBlockBg(block, node) {
       if (videoLink && !videoLink.hash) videoLink.hash = 'autoplay';
       if (childCount > 1) child.classList.add(...viewports[i]);
       const pic = child.querySelector('picture');
-      if (pic && (child.childElementCount === 2 || child.textContent?.trim())) {
+      if (useHandleFocalpoint
+        && pic && (child.childElementCount === 2 || child.textContent?.trim())) {
```
chrischrischris, line 110: super nit
```suggestion
      if (useHandleFocalpoint && pic 
        && (child.childElementCount === 2 || child.textContent?.trim())) {
```

chrischrischris (approved):

PR #1429, opened 2023-10-17
MWPW-136372 - Fixing stuffs to make performance great again..

sabyamon:
Thanks to @auniverseaway for the guidance on this.

Resolves: https://jira.corp.adobe.com/browse/MWPW-136372

**Test URLs:**
- After: https://quiz-performance--milo--sabyamon.hlx.live/drafts/xiasun/quiz-2/?martech=off


sanrai (approved):

libs/blocks/quiz/quiz.js
```diff
@@ -294,41 +294,42 @@ const App = ({
...
-                  <${StepIndicator} 
+                  ${selectedQuestion.questions && html`<${StepIndicator}
                     currentStep=${currentStep} 
                     totalSteps=${totalSteps} 
                     prevStepIndicator=${prevStepIndicator}
                     top="${true}" />
+                  `}
 
-                  <div class="quiz-background">
+                  ${selectedQuestion.questions && html`<div class="quiz-background">
```
echampio-at-adobe, line 304: The background element should be contingent on whether one has actually been specified in the json. Otherwise, we just create a picture element with an empty src, which renders as a broken image. 

`${selectedQuestion.questions && getStringValue('background') !== '' && html`<div class="quiz-background">`

The two instances of getStringValue also need to be made more robust. Respectively, I think they need to look like the following:

`const getStringValue = (propName) => {
  if (!selectedQuestion?.questions) return '';
  const question = stringQList[selectedQuestion.questions];
  return question?.[propName] || '';
};`

`const getStringValue = (propName) => {
  if (!selectedQuestion?.questions) return '';
  const question = stringQList[selectedQuestion.questions];
  return question?.[propName] || '';
};`

Otherwise, the methods can happily be called when selectedQuestion might be valid, but selectedQuestion.questions isn't, and they error in their current forms.

echampio-at-adobe (changes requested):

fullcolorcoder:
Nice!

JackySun9 (approved):

libs/blocks/quiz/quiz.js
sabyamon: Done. Updated. Thanks for the feedback.

echampio-at-adobe (approved):

PR #1431, opened 2023-10-18
MWPW-137805 - update to latest version of rum collection

chicharr:
* Update `samplerum.js` to the latest version, pulling it via npm module `@adobe/helix-rum-js: 1.6.0`
* Init RUM collection eagerly.
* Track `top` checkpoint.

Resolves: [MWPW-137805](https://jira.corp.adobe.com/browse/MWPW-137805)

**Test URLs:**
- After: https://update-rum--milo--chicharr.hlx.page/?martech=off


package.json
```diff
@@ -39,6 +39,7 @@
     "@web/dev-server-import-maps": "^0.0.6",
     "@web/test-runner": "0.13.27",
     "@web/test-runner-commands": "0.6.1",
+    "@adobe/helix-rum-js": "1.6.0",
```
mokimo, line 42: If we just copy the code anyway, is there any benefit in adding this as a dev dependency?

chicharr: At the moment no benefit, it just shows that the file is a copy from `node-modules/@adobe/helix-rum-js` .
If as next step you would like use `rollup`  to include the file in your code, then it would be relevant.

libs/utils/utils.js
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable no-console */
+import { sampleRUM } from './samplerum.js';
```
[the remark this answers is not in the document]
auniverseaway: We should not put rum in front of LCP or as part of a non-dynamic import.

mokimo (changes requested):
Given Chris' comment, let's move back to loading the samplerum the way it was before and just update the version.

libs/utils/utils.js
```diff
@@ -144,6 +145,21 @@ function getEnv(conf) {
   return { ...ENVS.prod, consumer: conf.prod };
   /* c8 ignore stop */
 }
+export function initRUM() {
```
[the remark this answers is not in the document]
chicharr: Without knowing the internals, I would also go for option B) as:
* You don't need to modify the `samplerum.js` module
* It enables RUM collection in all consumers.

Fyi, in the boilerplate this is the very first code [loaded](https://github.com/adobe/aem-boilerplate/blob/2d4cfa73a9c35c065e9dd2096a9bab3637ddd9b4/scripts/aem.js#L24) and [executed ](https://github.com/adobe/aem-boilerplate/blob/2d4cfa73a9c35c065e9dd2096a9bab3637ddd9b4/scripts/aem.js#L678)

libs/scripts/scripts.js
```diff
@@ -138,6 +139,8 @@ const eagerLoad = (img) => {
   img?.setAttribute('loading', 'eager');
   img?.setAttribute('fetchpriority', 'high');
 };
+// Initializes Real User Monitoring
+initRUM();
```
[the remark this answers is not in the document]
chicharr: Thank you @mokimo I was not aware that this is just a sample file. 
So if I got it right basically all the consumers are actually importing `libs/utils.js`, but they have their own entry point `scripts.js`,  correct?

libs/utils/utils.js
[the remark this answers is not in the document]
yesil: @auniverseaway  do you mean it can be added as non blocking import?
Where do you suggest this to be moved to?

libs/scripts/delayed.js
```diff
@@ -82,7 +82,7 @@ const loadDelayed = ([
     } else {
       resolve(null);
     }
-    import('../utils/samplerum.js').then(({ sampleRUM }) => sampleRUM('cwv'));
+    import('../utils/utils.js').then(({ sampleRUM }) => sampleRUM('cwv'));
```
mokimo, line 85: This leads to a circular dependency, you can just pass sampleRUM as a dependency to loadDelayed in the utils if we stick with this approach

mokimo (commented):
This approach an improvement, though we do have a general issue with a steadily growing `utils.js`

package.json
mokimo, line 42: This seems to be an omission, we want to pull this from the public npm instance, correct? https://github.com/adobe/helix-rum-js

mokimo: Currently this points to artifcatory, which is definitely nothing we want to include in the setup

mokimo: We use microbundle for preact, which might be an option. But IMO just copy/pasting seems to be the more straight forward way for now

chicharr: you are right. 
I have updated it

libs/scripts/delayed.js
chicharr: done!

mokimo:
There's a PR to upgrade the `samplerum.js` https://github.com/adobecom/milo/pull/1448
We don't have alignment AND/OR a use-case that benefits us as a platform on moving loading RUM to pre-LCP yet.
For this reason, I'll close this PR.

If you think this is still valuable, to move to loading the `samplerum.js` tracking to pre LCP - feel free to get some discussion going on https://github.com/orgs/adobecom/discussions

PR #1433, opened 2023-10-18
MWPW-137906 Round media block

suhjainadobe:
Ability to create rounded corners from a defined set of options.
An author can create a media block. The changes on brick block will done in separate PR.

Authoring has been given in [MWPW-137562](https://jira.corp.adobe.com/browse/MWPW-137562)

Resolves: [MWPW-137906](https://jira.corp.adobe.com/browse/MWPW-137906)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/suhjain/round-media-block?martech=off
- After: https://round-media-block--milo--suhjainadobe.hlx.page/drafts/suhjain/round-media-block?martech=off


libs/styles/styles.css
```diff
@@ -124,6 +124,13 @@
   --icon-size-s: 32px;
   --icon-size-xs: 24px;
   --icon-size-xxs: 16px;
+
+    /* border-radius sizes */
+    --border-radius-s: 4px;
+    --border-radius-l: 8px;
+    --border-radius-xl: 16px;
+    --full-border-radius: 50%;
```
overmyheadandbody, line 132: Are we certain authors understand that this would create a circle only if the element has an aspect ratio of 1:1? Here's a quick example of what I mean - [link](https://jsbin.com/nekizuhuvi/1/edit?html,css,output). I think this might be misleading if not properly explained.

libs/styles/styles.css
```diff
@@ -124,6 +124,13 @@
...
+    --no-border-radius: 0;
```
overmyheadandbody, line 133: I'd argue that the meaning of 'no border radius' will always be '0', so I'd save on a few dozen characters and just use `0` where needed.

libs/blocks/section-metadata/section-metadata.css
```diff
@@ -137,6 +137,46 @@
...
+.section.border-radius-block-l > div .background picture img {
+  border-radius: var(--border-radius-l);
+}
+
+.section.border-radius-block-xl > div,
+.section.border-radius-block-xl > div .background picture img {
+  border-radius: var(--border-radius-xl);
+}
+
+.section.no-border-block-radius > div,
```
overmyheadandbody, line 155: Since this is supposed to negate the previous class names, shouldn't it be `no-border-radius-block`? I see the same pattern has been applied for the image use case.

libs/blocks/section-metadata/section-metadata.css
```diff
@@ -137,6 +137,46 @@
...
+.section.no-border-block-radius > div .background picture img {
+  border-radius: var(--no-border-radius) !important;
```
overmyheadandbody, line 157: Why is the `!important` needed? Same question for L173. What other rule might override this one?

overmyheadandbody (commented):
Could you edit your document to include the new variations? This would be helpful both for QA purposes and future reference.

I also suggest adding some PR labels - https://github.com/adobecom/milo/wiki/PR-Labels#labels

Lastly, what's the reason for the very low LH score? It's likely not due to the changes proposed by this PR, but it's definitely worth investigating.

libs/styles/styles.css
suhjainadobe: This was the assumption we has taken that the authors would give images with aspect ratio 1:1 only. Because if we apply aspect-ratio property to the image, it may get a little distorted and authors may not be able to debug it.

libs/styles/styles.css
suhjainadobe: This was to keep similarity between the class names used by the authors and In case if the same is used in some other block in future. But you are correct, we can directly use 0.

libs/blocks/section-metadata/section-metadata.css
suhjainadobe: done. That was a mistake

libs/blocks/section-metadata/section-metadata.css
suhjainadobe: Removed it. There is no rule to remove it right now.

suhjainadobe:
> Could you edit your document to include the new variations? This would be helpful both for QA purposes and future reference.
> 
> I also suggest adding some PR labels - https://github.com/adobecom/milo/wiki/PR-Labels#labels
> 
> Lastly, what's the reason for the very low LH score? It's likely not due to the changes proposed by this PR, but it's definitely worth investigating.

I have updated the document and added the PR label. Looking into LH score.

vgoodric:
@suhjainadobe 
Perhaps I'm missing it, but can you spell out in your description what you are adding?  I'm inferring these are variants to section metadata only, not to blocks and the new variant names are:
border-radius-block-s
border-radius-block-m
border-radius-block-l
border-radius-block-xl
border-radius-image-s
border-radius-image-m
border-radius-image-l
border-radius-image-xl

Is that accurate?

auniverseaway (changes requested):
I think this will create un-expected behaviors and it's a bit of an anti-pattern.

Sections should not influence how child blocks look. 
A section should _only_ influence how blocks are laid out.. 3-up, etc. 

This should be a variation on the blocks that need it. Authors should not be required to use section-metadata to make a block have a rounded border.

auniverseaway:
Another piece of feedback, and to echo Milo Academy...

We should be speaking in human terms when talking variations.

```
border-radius-block-s
```

This is not ideal. Suggestions:

```
media (s-rounded-corners)
```
Implying: "I want a media block with small rounded corners."

rgclayton:
I'm not seeing the block border radius working on the first media blocks in your example page. I see the image border radius works, but not the block border radius. 

image

suhjainadobe:
> * https://round-media-block--milo--suhjainadobe.hlx.page/drafts/suhjain/round-media-block

Yes, these are the styles added to section-metadata only and not to individual blocks. There separate styles for rounded border radius and rounded images.

suhjainadobe:
> I'm not seeing the block border radius working on the first media blocks in your example page. I see the image border radius works, but not the block border radius.
> 
> image

@rgclayton : Please try some background color. Border-radius will be visible.
image

libs/styles/styles.css
```diff
@@ -453,6 +459,38 @@ span.icon.margin-inline-end { margin-inline-end: 8px; }
...
+  border-radius: var(--s-rounded-corners);
+}
+
+.m-image-rounded-corners .foreground picture img {
+  border-radius: var(--m-rounded-corners);
+}
+
+.l-image-rounded-corners .foreground picture img {
+  border-radius: var(--l-rounded-corners);
+}
```
overmyheadandbody, line 488: For brevity purposes, since this is the main style file and we want to keep it small, could we not group these with their non-image counterparts? As in
```
.s-rounded-corners,
.s-rounded-corners > .background img,
.s-image-rounded-corners .foreground picture img {
  border-radius: var(--s-rounded-corners);
}
```

overmyheadandbody (commented):
This looks a lot cleaner, I have just one more question

libs/styles/styles.css
rgclayton: +1 this.
I think grouping like this would also help if/when an update is needed.

rgclayton:
Adding `?martech=off` to your testing urls should help the "aem-psi-check". 
Performance score is pretty bad though. These css updates shouldn't have much of an impact.

auniverseaway (changes requested):
This should not be in styles.css.

This would impact every single block... marquee, aside, card, faas, etc.

I would rather start with a small blast radius of where this is used for the validated use cases.

Blindly saying, "Allow rounded corners on **_every_** block..." is going to create a lot of un-predictability.

I would drive back to what blocks this is truly needed for and scope this file to those blocks. These styles, in theory, could be imported as a dependency for the required blocks.

Delivering all this extra stuff to every visitor does not seem right.

salonijain3:
This ticket is currently blocked by design where we dont know the targeted blocks. 
In the original [JIRA](https://jira.corp.adobe.com/browse/MWPW-137906) description, it said "I want the ability to create rounded corners from a defined set of options (requirements below) on any image or block container in Milo."

In the functional acceptance, it is mentioned that "An author can create a media block, text block, or icon block and the image or container can have these options: "

Then, when discussed further, It was added that we can potentially "keep it to blocks that have containers (like bricks) and images in media blocks, subject images of marquees...things like that and maybe we need to scope a what a rounded corner theming looks like? "

I am overall confused as to which blocks really need this and implementation of this small block of css would depend on which blocks actually need it, once we have that clarity, we can implement it either in the styles.css i.e. if we need it at a broad level which was the original ask and this PR is a result of that or we can import it in the blocks that actually need it. 

I would like to ensure greater clarity on the implementation requirements from the Design/Consonant side and thoroughly understand the use cases before initiating the creation of ambiguous JIRA tickets, which may lead to confusion among engineers. 

cc: @rgclayton @suhjainadobe @dstrong23

suhjainadobe:
Curretly this is blocked by merging of PR for bricks block.(https://github.com/adobecom/milo/pull/1515). Once that is merged, I can raise the PR for rounded corner story.
@ryanmparrish @rgclayton @salonijain3

suhjainadobe:
> @suhjainadobe Do we need a `do not merge` label on this ticket? Or do we need to close this and reopen later?

Ill close it for now, as there is dependency on brick block. Anyways the implementation will change, so will reopen it later.

Patil31:
Verified the media block for rounded corners 

https://round-media-block--milo--suhjainadobe.hlx.page/drafts/shrinidhivp/media-story/media-4th-dec

libs/blocks/media/media.css
```diff
@@ -1,4 +1,6 @@
 /* media - consonant v5.1 */
+@import '../../styles/rounded-corner.css';
```
chrischrischris, line 2: Instead of importing for all media blocks, what about checking for the existance of a `rounded-corners` class and only import rounded-corner.css (via loadStyle) when needed.

libs/styles/rounded-corner.css
```diff
@@ -0,0 +1,29 @@
+:root {
```
chrischrischris, line 1: nit: Since the classnames are `x-rounded-corners` the filename should probably be `rounded-corners.css`

libs/styles/rounded-corner.css
```diff
@@ -0,0 +1,29 @@
...
+  /* border-radius sizes */
+  --s-rounded-corners: 4px;
+  --m-rounded-corners: 8px;
+  --l-rounded-corners: 16px;
+  --full-rounded-corners: 50%;
+}
+
+.s-rounded-corners,
+.s-rounded-corners > .background img,
+.s-image-rounded-corners .foreground .image picture img {
```
ryanmparrish, line 11: Few thoughts.. 

nit: I like the name `border-radius` better for the file name, css class names and vars. Authoring may prefer `*-rounded-corners` however, so disregard if that's been discussed otherwise Id lean towards css prop names if in doubt. 
ex. 
```
  --s-border-radius: 4px;
  --m-border-radius: 8px;
  --l-border-radius: 16px;
  --50-border-radius: 50%;
  ```
I do think we should update the variant 
  `*-image-rounded-corners` to... 
  `*-rounded-corners-image` 
to keep the names unified and have an optional type convention when incrementing utility css options `${unit-property-type}`

Comment (ryanmparrish) [reply libs/styles/rounded-corner.css:11]: Also, Is there a default size for just `rounded-corners` || `border-radius`?

Comment (rgclayton) [reply libs/styles/rounded-corner.css:11]: @ryanmparrish [See this feedback](https://github.com/adobecom/milo/pull/1433#issuecomment-1781579239) related to the naming

Comment (suhjainadobe) [reply libs/styles/rounded-corner.css:11]: The naming was done based on Chris's suggestion. Is there a need to still change it?

File: libs/blocks/media/media.css (hunk 6 above)
Comment (suhjainadobe) [reply libs/blocks/media/media.css:2]: Would it be fine to dynamically loadStyle in every js file that needs the styling? Right now is needed only for media and Brick block, but it might be extended to more blocks .The css file is not big, thus directly imported it on block's css file.
@chrischrischris @rgclayton @salonijain3

File: libs/styles/rounded-corner.css (hunk 8 above)
Comment (ryanmparrish) [reply libs/styles/rounded-corner.css:11]: I guess it's OK. I still hate it for the record. 
If we're talking about readability, When I read `s-image...` I immediately think image size, not what is the border radius of an image. Hence the suggestion `s-border-radius-image` etc. Anyhow thats my 2 cents.

File: libs/styles/rounded-corner.css [hunk 9]
```diff
@@ -0,0 +1,29 @@
...
+  border-radius: var(--s-rounded-corners);
+}
+
+.m-rounded-corners,
```
ryanmparrish, line 15:
```suggestion
.rounded-corners,
.m-rounded-corners,
```
We should set defaults for no size passed `rounded-corners` && `rounded-corners-image`

libs/styles/rounded-corner.css
```diff
@@ -0,0 +1,29 @@
...
+.m-rounded-corners > .background img,
+.m-image-rounded-corners .foreground .image picture img {
```
ryanmparrish, line 17:
```suggestion
.rounded-corners-image .foreground .image picture img,
.m-rounded-corners-image .foreground .image picture img {
```
I still think we should update these to keep to a convention
`${size-property-type}`

libs/styles/rounded-corner.css
```diff
@@ -0,0 +1,29 @@
...
+  border-radius: var(--m-rounded-corners);
+}
+
+.l-rounded-corners,
+.l-rounded-corners > .background img,
+.l-image-rounded-corners .foreground .image picture img {
```
ryanmparrish, line 23:
```suggestion
.l-rounded-corners-image .foreground .image picture img {
```

libs/styles/rounded-corner.css
```diff
@@ -0,0 +1,29 @@
...
+  border-radius: var(--l-rounded-corners);
+}
+
+.full-image-rounded-corners .foreground .image picture img {
```
ryanmparrish, line 27:
```suggestion
.full-rounded-corners-image .foreground .image picture img {
```

ryanmparrish (changes requested):
Looking good. Few nit naming conventions I think we should update. Also to what @chrischrischris said, update the file name to `rounded-corners.css` to match. ;)

qq: Is this going to be used in other files? Is that why it's separate?

libs/styles/rounded-corner.css
suhjainadobe: done

libs/styles/rounded-corner.css
suhjainadobe: done

libs/styles/rounded-corners.css
```diff
@@ -1,29 +1,36 @@
...
   --l-rounded-corners: 16px;
   --full-rounded-corners: 50%;
 }
 
+.rounded-corners,
+.rounded-corners > .background img,
+.rounded-corners-image .foreground .image picture img {
+  border-radius: var(--rounded-corners);
+}
+
```
ryanmparrish, line 15: Thanks for the updates on this. 
I was thinking the default of `rounded-corners` to have a value and just added to either the `s-` or `m-` instances. It doesn't make sense to have it at all if the value is just 0. My thinking is that the most often use case shouldn't have to denote a size, but just rounded-corner: default 8. etc.

ryanmparrish (approved):

salonijain3:
All the changes have been accomodated, @auniverseaway can you re-review so that this can be merged. 
P.S. This cannot be merged until you resolve your "requested changes".

cc : @ccongerp  @audi5

libs/blocks/media/media.js
```diff
@@ -20,6 +20,11 @@ function decorateAvatar(el) {
 }
 
 export default function init(el) {
+  const { miloLibs, codeRoot } = getConfig();
+  const regex = /rounded-corners/g;
+  if ([...el.classList].some((c) => regex.test(c))) {
+    loadStyle(`${miloLibs || codeRoot}/styles/rounded-corners.css`);
```
auniverseaway, line 26: Use "base"

auniverseaway (approved):

PR #627, opened 2023-04-10
Aside Notification Centered Block Update

Sartxi:
The aside notification "centered" is authored to read the `data-align` attribute added by word when you set the text contents alignment. But when copying the aside from the sidekick library those attributes aren't being set. This update will make it so you can also set the centered text as a style variant in order to make the centered variation available when copying the block from the sidekick library.

Resolves: https://jira.corp.adobe.com/browse/MWPW-129469

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/sarchibeque/notification-centering?martech=off
- After: https://aside-notification-centered--milo--adobecom.hlx.page/drafts/sarchibeque/notification-centering?martech=off

PR #1450, opened 2023-10-25
[MWPW-117914] Allow promobar use on top of Gnav

overmyheadandbody:
## Description
This allows authors to include a promo bar on top of the Global Navigation by defining a new `gnav-promo-source` metadata property with its value leading to a fragment which contains an `aside` block. For design purposes, it's important that authors use the `promobar` variation of the block, although we don't currently have a restriction for this from a code perspective.

Some additional notes:
- `utils.js` has been slightly adapted to add a modifier class to the header in case a Gnav promo was authored on a page. Its use can be seen in `style.css` where we adapt the nav height;
- some minimal changes have been made to the `aside` block CSS to ensure content alignment follows Word indentation and that the CTA does not wrap on multiple rows. The latter might be something useful for all `.con-button` elements, although I'm not aware of all the use cases;

Documentation will be added soon at https://milo.adobe.com/docs/authoring/global-navigation.

## Related Issue
Resolves: [MWPW-117914](https://jira.corp.adobe.com/browse/MWPW-117914)

## Testing instructions
As an FYI, it was decided that promos will not be available on mobile devices for the moment. Rendering will only occur if the viewport width is larger than `900px`. If a user then resizes their viewport below `900px`, the promo will be hidden.

One thing to look into is any CLS impact - the promo will be added on top of the navigation, thus pushing down all other page content. We assign a minimal value to the promo to apply the same principle as breadcrumbs - assign space early on to avoid content jumping around later. Tablet devices will likely experience _some_ CLS, since we can't guesstimate the actual length of the copy, but we're trying to keep the impact to a minimum.

## Screenshots:
**Dc Gnav with Promo**
DC page with Gnav Promo

## Test URLs
**Acrobat:**
- Before: https://main--dc--adobecom.hlx.page/drafts/ramuntea/page-with-gnav-promo-fragment?martech=off
- After: https://main--dc--adobecom.hlx.page/drafts/ramuntea/page-with-gnav-promo-fragment?milolibs=mwpw-117914-gnav-promo--milo--overmyheadandbody&martech=off

**BACOM:**
- Before: https://main--bacom--adobecom.hlx.page/drafts/ramuntea/page-with-gnav-promo-fragment
- After: https://main--bacom--adobecom.hlx.page/drafts/ramuntea/page-with-gnav-promo-fragment?milolibs=mwpw-117914-gnav-promo--milo--overmyheadandbody&martech=off

**CC:**
- Before: https://main--cc--adobecom.hlx.page/drafts/ramuntea/page-with-gnav-promo-fragment?martech=off
- After: https://main--cc--adobecom.hlx.page/drafts/ramuntea/page-with-gnav-promo-fragment?milolibs=mwpw-117914-gnav-promo--milo--overmyheadandbody&martech=off

**Milo:**
- Before: https://main--milo--overmyheadandbody.hlx.page/drafts/ramuntea/gnav-with-metadata-promo?martech=off
- After: https://mwpw-117914-gnav-promo--milo--overmyheadandbody.hlx.page/drafts/ramuntea/gnav-with-metadata-promo?martech=off

### DIFF
--- a/libs/blocks/aside/aside.css
+++ b/libs/blocks/aside/aside.css
@@ -481,6 +481,14 @@
   height: var(--icon-size-m);
 }
 
+.aside.promobar .promo-text[data-align="center"] {
+  justify-content: center;
+}
+
+.aside.promobar .action-area .con-button {
+  flex-shrink: 0;
+}
+
 .aside.promobar .promo-text.desktop-up,
 .aside.promobar .promo-text.tablet-up {
   display: none;
@@ -600,7 +608,7 @@
 
   .aside.split.no-media:not(.notification) .foreground.container .text {
     padding: var(--spacing-xxxl) 0;
-  }  
+  }
 
   .aside.promobar .mobile-up.hide-block ~ .promo-close {
     display: none;
@@ -611,15 +619,15 @@
   .aside.small {
     min-height: 420px;
   }
-  
+
   .aside.medium {
     min-height: 560px;
   }
-  
+
   .aside.large {
     min-height: 700px;
   }
-  
+
   .aside .foreground.container {
     align-items: center;
     flex-direction: row;
@@ -889,7 +897,7 @@
     align-items: center;
     text-align: center;
   }
-  
+
   .aside.promobar.popup .promo-text .action-area {
     justify-content: center;
     padding: 0 0 var(--spacing-xxs);
@@ -1106,7 +1114,7 @@
     right: var(--spacing-s);
     top: calc(50% - 10px);
   }
-  
+
   .aside.promobar .desktop-up.hide-block ~ .promo-close {
     display: none;
   }
--- a/libs/blocks/global-navigation/features/aside/aside.js
+++ b/libs/blocks/global-navigation/features/aside/aside.js
@@ -1,0 +1,26 @@
+import { loadBlock, decorateAutoBlock } from '../../../../utils/utils.js';
+import { toFragment, lanaLog, getAnalyticsValue } from '../../utilities/utilities.js';
+
+export default async function decorateAside({ headerElem, promoPath } = {}) {
+  const onError = () => {
+    headerElem.classList.remove('has-promo');
+    lanaLog({ message: 'Gnav Promo fragment not replaced, potential CLS' });
+    return '';
+  };
+
+  const fragLink = toFragment`<a href="${promoPath}">${promoPath}</a>`;
+  const fragTemplate = toFragment`<div>${fragLink}</div>`;
+  decorateAutoBlock(fragLink);
+  if (!fragLink.classList.contains('fragment')) return onError();
+  await loadBlock(fragLink).catch(() => onError());
+  const aside = fragTemplate.querySelector('.aside');
+  if (fragTemplate.contains(fragLink) || !aside) return onError();
+
+  aside.removeAttribute('data-block');
+  aside.setAttribute('daa-lh', 'Promo');
+  aside.querySelectorAll('a').forEach((link, index) => {
+    link.setAttribute('daa-ll', getAnalyticsValue(link.textContent, index + 1));
+  });
+
+  return aside;
+}
--- a/libs/blocks/global-navigation/features/aside/aside.js
+++ b/libs/blocks/global-navigation/features/aside/aside.js
@@ -3,7 +3,7 @@ import { toFragment, lanaLog, getAnalyticsValue } from '../../utilities/utilitie
 
 export default async function decorateAside({ headerElem, promoPath } = {}) {
   const onError = () => {
-    headerElem.classList.remove('has-promo');
+    headerElem?.classList.remove('has-promo');
     lanaLog({ message: 'Gnav Promo fragment not replaced, potential CLS' });
     return '';
   };
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -4,6 +4,8 @@ import {
   getMetadata,
   loadIms,
   decorateLinks,
+  loadBlock as loadMiloBlock,
+  decorateAutoBlock,
 } from '../../utils/utils.js';
 import {
   toFragment,
@@ -182,6 +184,7 @@ class Gnav {
       loadBaseStyles,
       this.decorateMainNav,
       this.decorateTopNav,
+      this.decorateAside,
       this.decorateTopnavWrapper,
       this.ims,
       this.addChangeEventListeners,
@@ -231,7 +234,7 @@ class Gnav {
         ${breadcrumbs}
       </div>`;
 
-    this.el.append(this.elements.curtain, this.elements.topnavWrapper);
+    this.el.append(this.elements.curtain, this.elements.aside, this.elements.topnavWrapper);
   };
 
   addChangeEventListeners = () => {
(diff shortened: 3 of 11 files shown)


libs/blocks/aside/aside.css
```diff
@@ -480,6 +480,14 @@
   height: var(--icon-size-m);
 }
 
+.aside.promobar .promo-text[data-align="center"] {
```
Sartxi, line 483: Just so you are aware, we saw issues copying these Word data attributes from the sidekick library. I am not sure if that is still a problem, but we had to go with `center` as a block variant as a result. More context [here](https://github.com/adobecom/milo/pull/627).

rgclayton:
There is also a `promo' block. It was migrated from blog.adobe.com. This should probably be included?
https://github.com/adobecom/milo/tree/main/libs/blocks/promo

[Sample page](https://main--milo--adobecom.hlx.page/drafts/rclayton/blog-migration/sticky)

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -479,6 +482,35 @@ class Gnav {
     return decoratedElem;
   };
 
+  decorateAside = async () => {
```
mokimo, line 485: Should we have a dynamic import for this, since it's not always used by everyone?

Or at least leave a comment+ticket so we can refactor this if we decide on a different path later on?

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -479,6 +482,35 @@ class Gnav {
...
+      this.el.classList.remove('has-promo');
+      lanaLog({ message: 'Gnav Promo fragment not replaced, potential CLS' });
+      return this.elements.aside;
+    };
+
+    const fragLink = toFragment`<a href="${promoPath}">${promoPath}</a>`;
+    const fragTemplate = toFragment`<div>${fragLink}</div>`;
+    decorateAutoBlock(fragLink);
+    if (!fragLink.classList.contains('fragment')) return onError();
+    await loadMiloBlock(fragLink);
```
mokimo, line 500: We should add a catch either here, or even better around the whole block.

As our logic currently stands, if this promise doesn't resolve the page doesn't have a nav at all.

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -182,6 +184,7 @@ class Gnav {
       loadBaseStyles,
       this.decorateMainNav,
       this.decorateTopNav,
+      this.decorateAside,
       this.decorateTopnavWrapper,
       this.ims,
       this.addChangeEventListeners,
```
mokimo, line 190: We should create a Ticket to investigate a graceful failover strategy if any of those tasks fail - the nav will be gone

libs/styles/styles.css
```diff
@@ -1,8 +1,9 @@
 /*** Variables ***/
 
  :root {
-  --global-height-nav: 64px;
+  --global-height-nav: 65px;
```
mokimo, line 4: Some consumers might rely on the fact it's 64px and get a tiny amount of CLS https://github.com/search?q=org%3Aadobecom+64px&type=code&p=1

Should we shrink the nav itself by 1px so we are at 64px with the border?

mokimo (approved):
A few comments & questions, but looks good 👍

overmyheadandbody:
Hey @rgclayton! During the due diligence phase I also tried out the Promo block, but it lacked background image support, device-specific copy capabilities and has the close button which we don't need (and is also a bit annoying, since state is not maintained once a promo is closed), so we decided to go with the Aside block for now.

Looking into the future, this PR tailors to current Black Friday needs, but our stakeholders are still gathering up information and data to figure out whether Promos have a future on top of the Global Navigation anyway. Depending on this outcome, we'll decide how to expand support of Aside/Promo blocks inside the Global Navigation.

libs/blocks/aside/aside.css
overmyheadandbody: Thanks for the context, Sean! I'm not sure this would be copied over using the Library, but it also doesn't hurt to have it. I'll look into adding the `center` variant for the `promobar` variant as well.

libs/blocks/global-navigation/global-navigation.js
overmyheadandbody: We'd still need some base logic here, since we need to check the metadata property and decide whether we need to import the Aside module. So essentially we'd be making an additional call for the code on L490-L509. I'm not sure it's completely worth it, wdyt?

libs/blocks/global-navigation/global-navigation.js
overmyheadandbody: Good idea, I'll see what makes sense to wrap so we avoid any issues 👍

libs/blocks/global-navigation/global-navigation.js
overmyheadandbody: Well, they rely on each other anyway, so essentially if anything before `this.decorateTopnavWrapper` fails we wouldn't show a navigation. It wouldn't make sense to just show some bits and pieces that actually succeeded, like a nav with just Search and Breadcrumbs, either. IMS has its own catch, while `this.addChangeEventListeners` has all the safety checks in place. Lastly, the whole `init` method is wrapped in a `logErrorFor` call, so I think there should be enough safety mechanisms in place. If we start seeing `"Error in global navigation init"` in our logs, we can look into this more.

libs/blocks/global-navigation/global-navigation.js
mokimo: It's a though one, if it grows certainly own module - for now we can keep it

libs/styles/styles.css
overmyheadandbody: I need to look into this a bit more, good callout

libs/blocks/aside/aside.css
overmyheadandbody: Adding the `center` variation skews the layout because of https://github.com/adobecom/milo/blob/main/libs/blocks/aside/aside.css#L48-L50 interacting with https://github.com/adobecom/milo/blob/main/libs/blocks/aside/aside.css#L634-L636.

I'm not aware of all the Aside use-cases, as it seems highly complex, so I think I'll stay away from additional logic for the moment. One could argue that if `:not(.center)` is used on L634 (`.aside:not(.notification):not(.inline):not(.center) .foreground.container> :last-child`), then it could also be used on L48 (`.aside:not(.notification):not(.inline) .foreground> :last-child`), which would fix the layout issues I'm seeing, but I wouldn't want to make assumptions.

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -182,6 +184,7 @@ class Gnav {
       loadBaseStyles,
       this.decorateMainNav,
       this.decorateTopNav,
+      this.decorateAside,
```
narcis-radu, line 187: I was looking at @mokimo's suggestion to dynamically import `decorateAside` logic, maybe we can use something like
```js
const tasks = [ loadBaseStyles, this.decorateMainNav, thisDecorateTopNav, this.decorateTopNavWrapper, ... ];
if (getMetadata('gnav-promo-source')) {
  this.decorateAside = await import('.../path');
  tasks.splice(2, 0, this.decorateAside);
}
```

narcis-radu: I know we would be making an additional call, but might be worth it vs. loading the code and never use it.

overmyheadandbody: I kept a hybrid approach, load minimal stuff here, but move the decoration logic to a module, https://github.com/adobecom/milo/pull/1450/commits/f82e56982923d0f4e62ad972db8f646e2a45fd68

libs/styles/styles.css
overmyheadandbody: I think this would grant a bit more time to investigate, since no one should be manually adding `64px` in their projects and instead use `var(--global-height-nav)`. However, there are too many occurrences to enforce this rule, so I'm 'shrinking' the nav by `1px` for now.

This actually addresses a production bug, where the page content overlaps the Global Navigation for `1px.`

mokimo (approved):

narcis-radu (approved):

libs/blocks/global-navigation/features/aside/aside.js
```diff
@@ -0,0 +1,26 @@
+import { loadBlock, decorateAutoBlock } from '../../../../utils/utils.js';
+import { toFragment, lanaLog, getAnalyticsValue } from '../../utilities/utilities.js';
+
+export default async function decorateAside({ headerElem, promoPath } = {}) {
+  const onError = () => {
+    headerElem.classList.remove('has-promo');
```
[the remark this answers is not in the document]
overmyheadandbody: Added in https://github.com/adobecom/milo/pull/1450/commits/4f5823c3e3c5ddf47e2768955391c5919fbcd8c6

overmyheadandbody:
Hi @Sartxi! Could you please have another look over this? Its merge is pending an aside block code owner approval. Thanks!

SilviuLCF (approved):

Sartxi (approved):

PR #345, opened 2022-12-08
MWPW-116104 - New Window Links

vhargrave:
v2 of opening #_blank links in new windows and using the links.xmls file to specify how to decorate certain links

Resolves: [MWPW-116104](https://jira.corp.adobe.com/browse/MWPW-116104)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/vhargrave/links-in-new-window?martech=off
- After: https://vhargrave-author-links-new-windows--milo--adobecom.hlx.page/drafts/vhargrave/links-in-new-window?martech=off

PR #1455, opened 2023-10-25
MWPW-136383 - Add off-origin ability to links.js

rgclayton:
Adds functionality to open all off-origin links into a new tab via `links.js`

* "off-origin" keyword is added to [links.xlsx](https://main--milo--adobecom.hlx.page/seo/links.json)

**Context:**
Links.js offers the ability to call out specific links and add nofollow, noopener, noreferrer, and/or _blank to the link. More details about the original implementation can be found in https://github.com/adobecom/milo/pull/345

"Links" is opt in and needs to be [enabled per site](https://github.com/adobecom/milo/blob/main/libs/scripts/scripts.js#L121)

**Resolves:** [MWPW-136383](https://jira.corp.adobe.com/browse/MWPW-136383)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rclayton/links?martech=off
- After: https://links-off-origin--milo--adobecom.hlx.page/drafts/rclayton/links?martech=off

**Testing notes:**
Any off-origin links should should open in a new tab (target: "_blank"). 
E.g. In the example page, any link that points away from Milo pages will open in a new tab when clicked.

If a link is pointed to a Milo page and contains `#_blank` the link will work as expected and open in a new tab.

### DIFF
--- a/libs/features/links.js
+++ b/libs/features/links.js
@@ -1,7 +1,7 @@
 let fetched = false;
 let linkData = null;
 
-const getLinks = async (path) => {
+const fetchSeoLinks = async (path) => {
   if (!path) return null;
   if (!fetched) {
     const resp = await fetch(path);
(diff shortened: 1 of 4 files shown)


vhargrave (approved):
Looks good !
Small nit that I'd add a unit test to cover this case, but asides from that, 👍

JasonHowellSlavin (approved):
Looks good to me as well!

chrischrischris (approved):

PR #1468, opened 2023-10-31
MWPW-138570: dont await martech loading

yesil:
Resolves: [MWPW-138570](https://jira.corp.adobe.com/browse/MWPW-138570)

it is loaded without `await` in `loadPostLCP` method.
**Test URLs:**

- Before: https://main--milo--adobecom.hlx.live
- After: https://main--milo--adobecom.hlx.live?milolibs=mwpw-138570

*** Homepage ***
- Before: https://main--homepage--adobecom.hlx.live/homepage/index-loggedout
- After: https://main--homepage--adobecom.hlx.live/homepage/index-loggedout?milolibs=mwpw-138570


libs/utils/utils.js
```diff
@@ -822,7 +822,7 @@ async function checkForPageMods() {
   }
 
   if (targetEnabled) {
-    await loadMartech({ persEnabled: true, persManifests, targetMd });
+    loadMartech({ persEnabled: true, persManifests, targetMd });
```
overmyheadandbody, line 825: Wouldn't this have the potential of causing flicker? Without `await`, page rendering continues normally, adding, say, a 'Buy Now' promo as the first element on the page. But personalisation then figures out that the user is already entitled to the product and that the promo should be removed, causing flicker and CLS. Did you set up a personalisation manifest with such a use-case and see that everything is working as expected?

yesil: martech has alloy-hiding mechanism to fight flicker by setting the opacity to 0.01 then target is enabled.
To test against more advanced cases, I think it is better if the work is done by Target/MEP devs.
This came up as a clear impact factor on the LCP on Homepage.
I hope @chrischrischris  and @vgoodric can confirm whether this is doable or a no go.

vgoodric:
The reason you need to use await is that the response from martech may alter the page before you load fragments and decorate.  It's a necessary evil.

When I look at https://main--homepage--adobecom.hlx.live/homepage/index-loggedout?milolibs=mwpw-138570 it looks super broken. The entire body is missing: 
image

@chrischrischris is back from PTO tomorrow, so we can get a 2nd opinion.

yesil:
@vgoodric if you can reproduce the super broken page, can you share your console output?
I get the normal experience.

yesil:
Looking at the waterfalls below, if we compare when the video blocks gets loaded in before and after, we can see that there is around 1.5s of LCP impact. I obtained the same number also via Web Vitales extension and I believe it is a very realistic number.
In the first waterfall, I see plenty of google tag manager related JS files getting loaded before the LCP blocks which is I doubt intended.

> the response from martech may alter the page before you load fragments and decorate

Target on Dexter can alter content anytime, is not needed at the very beginning of the processing of the components.
On Milo, I think it should be able to alter already processed blocks and re-decorate them, it can keep the body hiding or section hiding applied during this time.
I think the current approach sacrifices the performance a lot, even for visitors who are not qualified for a test.

before:
image

after:
image

yesil:
This is causing a flicker currently, keeping it open for the discussion.

yesil:
Discussed with @vgoodric, closing for now.

PR #1470, opened 2023-10-31
MWPW-138223 -  Aside - max width on large notifications

Sartxi:
The `aside.notification.large` has a max width of 1000px, and should be reduced to 800px when centered or when it doesn't have media in the foreground. The correct styles are currently provided when no media is present, but is not taking the specific `notification.large.center` rules into consideration. So when a background is provided to the block, the max width is rendered at 1000px wide. I also discovered that when you render less text content the content elements are displayed inline.

**Solution**
1. Apply the proper max width to `notification.large.center` variations. _See aside.css line 437_
2. Apply the missing flex direction style on `notification.large` variations. _See aside.css line 367_

**Note:** You may notice there are a lot of edits to the file and that is because I fixed all the style lint errors

Resolves: [MWPW-138223](https://jira.corp.adobe.com/browse/MWPW-138223)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/sarchibeque/aside-notification-width?martech=off
- After: https://sarchibeque-mwpw-138223-large-notification--milo--adobecom.hlx.page/drafts/sarchibeque/aside-notification-width?martech=off


libs/blocks/aside/aside.css
```diff
@@ -41,11 +41,11 @@
   gap: var(--spacing-l);
 }
 
-.aside:not(.notification):not(.inline) .foreground> :first-child {
+.aside:not(.notification):not(.inline) .foreground >  :first-child {
```
overmyheadandbody, line 44: L44, L48, L506 have double spaces after the `>`, they could be removed.

libs/blocks/aside/aside.css
```diff
@@ -585,24 +595,16 @@
...
     gap: 0;
   }
 
   .aside.split.no-media:not(.notification) .foreground.container .text {
     padding: var(--spacing-xxxl) 0;
-  }  
+  }
 
-  .aside.promobar .mobile-up.hide-block ~ .promo-close {
+  .aside.promobar .mobile-up.hide-block~.promo-close {
```
overmyheadandbody, line 607: Is the close button available for simple promobars or just the `popup` variant ones? Would it make sense to make it specific to the `popup` variant if that is the case?

overmyheadandbody (approved):
I also checked the upcoming promobar integration in Gnav with these new styles and things continue to work as expected. That PR also contains some minor indentation/empty spaces updates, so there might be some conflicts between the two.

libs/blocks/aside/aside.css
Sartxi: I will be sure to fix those 🙏🏻

meganthecoder (approved):
Looks good! It'd probably be good to double check the promo bar, etc to make sure that moving around the styles didn't affect anything because of inheritance

libs/blocks/aside/aside.css
Sartxi: @overmyheadandbody I added the popup variant to these

PR #1471, opened 2023-11-01
MWPW-137634 footer unit tests

robert-bogos:
## Description
This adds unit tests for the global footer

## Related Issue
Resolves: [MWPW-137634](https://jira.corp.adobe.com/browse/MWPW-137634)

## Testing instructions
1. change `web-test-runner.config.mjs` by adding `  files: ['./test/blocks/global-footer/global-footer.test.js'],` to the default exported json (this will help target the `global-footer.test.js` for tests)
2. run `npm run test:file:watch `

## Screenshots:
image

## Test URLs
**Acrobat:**
- Before: https://main--dc--adobecom.hlx.page/drafts/rbogos/default?martech=off
- After: https://main--dc--adobecom.hlx.page/drafts/rbogos/default?milolibs=mwpw-138031-footer-tests--milo--rbogos&martech=off

**BACOM:**
- Before: https://main--bacom--adobecom.hlx.page/drafts/rbogos/default
- After: https://main--bacom--adobecom.hlx.page/drafts/rbogos/default?milolibs=mwpw-138031-footer-tests--milo--rbogos&martech=off

**CC:**
- Before: https://main--cc--adobecom.hlx.page/drafts/rbogos/default?martech=off
- After: https://main--cc--adobecom.hlx.page/drafts/rbogos/default?milolibs=mwpw-138031-footer-tests--milo--rbogos&martech=off

**Milo:**
- Before: https://main--milo--robert-bogos.hlx.page/drafts/rbogos/default?martech=off
- After: https://footer-tests--milo--robert-bogos.hlx.page/drafts/rbogos/default?martech=off


test/blocks/global-footer/global-footer.test.js
```diff
@@ -0,0 +1,266 @@
...
+    it('should have footer', async () => {
+      const footer = await createFullGlobalFooter({ waitForDecoration: true });
+      expect(footer).to.exist;
+      expect(document.querySelector('footer')).to.exist;
+    });
+
+    it('should have menu', async () => {
+      const footer = await createFullGlobalFooter({ waitForDecoration: true });
+
+      expect(footer).to.exist;
```
mokimo, line 66: Since you've got a test checking if this exists, it can IMO be skipped in other tests.

test/blocks/global-footer/global-footer.test.js
```diff
@@ -0,0 +1,266 @@
...
+    });
+
+    it('should render the whole footer', async () => {
+      const footer = await createFullGlobalFooter({ waitForDecoration: true });
+      expect(footer).to.exist;
+
+      for (const selectorKey of allSelectorsKeys) {
+        const targetEl = document.querySelector(selectors[selectorKey]);
+        expect(!!targetEl).to.equal(true);
+      }
```
mokimo, line 130: Since all the tests up until here create the footer & then test if elements are visible, it might make sense to squash the tests together into the first "basic sanity test"

That should also help with overall testing speed, since creating the whole (nav/footer) is quite expensive from a unit testing perspective they should in theory only take 1-5ms.

The gnav tests are sadly not great in that regard, but we can be better here!

test/blocks/global-footer/global-footer.test.js
```diff
@@ -0,0 +1,266 @@
...
+    });
+  });
+
+  describe('conditional render tests', () => {
+    it('should render the footer when in viewport', async () => {
```
mokimo, line 135: This is the same test as your first ones, correct? If yes, we can skip on this first one since we already tested the functionality

mokimo (commented):
Very cool, clean structure & covers a good batch of tests.
A few first comments & a test for a tablet / mobile footer would be appreciated!

test/blocks/global-footer/global-footer.test.js
```diff
@@ -0,0 +1,266 @@
...
+      const viewportHeight = window.innerHeight;
+      insertDummyElementOnTop({ height: viewportHeight + 400 });
+
+      const footer = await createFullGlobalFooter({ waitForDecoration: false });
+
+      expect(footer).to.exist;
+
+      for (const selectorKey of allSelectorsKeys) {
+        expect(!!document.querySelector(selectors[selectorKey])).to.equal(
+          false,
```
JasonHowellSlavin, line 162: not sure why you split these expect statements on to new lines. same with 170,186.202. Does the `,` necessitate it?

skumar09:
@robert-bogos : 
A few nala tests corresponding to `Text block` are failing on your PR branch, ( they are passing on other branches)
I just wanted to check do you have latest milo code base on your branch also 

@rgclayton @AddumOne

test/blocks/global-footer/global-footer.test.js
robert-bogos: I'm using `forEach` now, so we can say this is solved. It was because of my VScode formatter extension

test/blocks/global-footer/test-utilities.js
```diff
@@ -0,0 +1,125 @@
...
+  const waitForElements = [];
+
+  const selectorsKeys = Object.keys(givenSelectors);
+  for (const selectorKey of selectorsKeys) {
+    const targetEl = parentEl.querySelector(givenSelectors[selectorKey]);
+    waitForElements.push(isElementVisible(targetEl));
+  }
+
+  return waitForElements.every((el) => el);
+};
```
mokimo, line 85: You could also chain the methods:
```suggestion
export const allElementsVisible = (selectors, parentEl) => Object
  .keys(selectors)
  .map((key) => isElementVisible(parentEl.querySelector(selectors[key])))
  .every((el) => el);
```

test/blocks/global-footer/test-utilities.js
```diff
@@ -0,0 +1,125 @@
...
+export const waitForFooterToDecorate = async () => {
+  const waitForElements = [];
+
+  const selectorsKeys = Object.keys(selectors);
+  for (const selectorKey of selectorsKeys) {
+    waitForElements.push(waitForElement(selectors[selectorKey]));
+  }
+  const elementsAreRendered = await Promise.all(waitForElements);
+  return !!elementsAreRendered;
+};
```
mokimo, line 96: You could shorten this:
```suggestion
export const waitForFooterToDecorate = () => Promise.all(
    Object
    .keys(selectors)
    .map(key => waitForElement(selectors[key]))
  );
```

And then just use
```js
// you don't need any checks, as the waitForElement method should timeout/reject and hence fail the Promise.all chain completely. If this runs through, you should be able to be confident all elements are now visible. (Or your checks in the test would fail anyway)
await waitForFooterToDecorate()
```

test/blocks/global-footer/test-utilities.js
```diff
@@ -0,0 +1,125 @@
...
+
+  await Promise.all([
+    loadStyles('../../../../libs/styles/styles.css'),
+    loadStyles('../../../../libs/blocks/global-footer/global-footer.css'),
+  ]);
+
+  const instance = initFooter(document.querySelector('footer'));
+  if (waitForDecoration) {
+    await waitForFooterToDecorate();
+  }
```
mokimo, line 115: nit: we slightly prefer one line if's in the codebase
```suggestion
  if (waitForDecoration) await waitForFooterToDecorate();
```

mokimo (commented):
Few nits and suggestions!

test/blocks/global-footer/global-footer.test.js
```diff
@@ -0,0 +1,263 @@
...
+
+      Object.keys(selectors).forEach((key) => expect(
+        document.querySelector(selectors[key]) instanceof HTMLElement,
+      ).to.equal(true));
+
+      const elementsAreVisible = allElementsVisible(
+        alwaysVisibleSelectorsDesktop,
+        document.querySelector(containerSelector),
+      );
+      expect(elementsAreVisible).to.equal(true);
```
mokimo, line 66: I personally prefer skipping on variables if they don't provide context or are used multiple times.
```suggestion
      expect(allElementsVisible(
        alwaysVisibleSelectorsDesktop,
        document.querySelector(containerSelector),
      )).to.equal(true);
```

test/blocks/global-footer/global-footer.test.js
```diff
@@ -0,0 +1,263 @@
...
+  afterEach(() => {
+    clock.restore();
+    window.fetch = originalFetch;
+  });
+
+  describe('wide screen', async () => {
+    it('should render the footer on wide screen', async () => {
+      const footer = await createFullGlobalFooter({ waitForDecoration: true, viewport: 'wide' });
+
+      expect(footer).to.exist;
```
mokimo, line 56: Since you already have some checks in the `createFullGlobalFooter` do you need the variable? The promise would reject and fail the test, potentially you can also just do:
```suggestion
      wait createFullGlobalFooter({ waitForDecoration: true, viewport: 'wide' });
```

mokimo:
@skumar09 this purely touches unit tests, so nala checks shouldn't fail. With the latest additions to the PR that seems to have cleared 👍

test/blocks/global-footer/test-utilities.js
robert-bogos: Yes, much better 👍

test/blocks/global-footer/test-utilities.js
robert-bogos: Got it 👍

test/blocks/global-footer/global-footer.test.js
robert-bogos: True, if they are not used multiple times they just add more lines of code for no reason

mokimo (approved):
Nice 👍 

Last thing you might want to look at, is if you run this in the browser and check the network tab/console it has 3 404s.
> http://localhost:2000/favicon.ico
> http://localhost:2000/libs/blocks/undefined/undefined.css
> http://localhost:2000/libs/blocks/undefined/undefined.js

robert-bogos:
> Nice 👍
> 
> Last thing you might want to look at, is if you run this in the browser and check the network tab/console it has 3 404s.
> 
> > http://localhost:2000/favicon.ico
> > http://localhost:2000/libs/blocks/undefined/undefined.css
> > http://localhost:2000/libs/blocks/undefined/undefined.js

- the `/favicon.ico ` seems to only show on first load, before the tests start. A solution would be to mock this call, but it will add more code and no extra functionality.
- the `/undefined` calls come from the test where the region picker's `href` has an empty `hash` and that causes these 404s in the `await loadBlock(regionSelector)` from `global-footer.js`, but it looks like the expected behaviour. Could mock these calls just to get rid of the 404s.

What do you think? @mokimo

test/blocks/global-footer/global-footer.test.js
```diff
@@ -0,0 +1,247 @@
...
+  waitForFooterToDecorate,
+  allSelectors,
+} from './test-utilities.js';
+import baseFooter from './mocks/base-footer.js';
+import fetchedFooter from './mocks/fetched-footer.js';
+import icons from './mocks/icons.js';
+import { mockRes } from '../global-navigation/test-utilities.js';
+import placeholders from './mocks/placeholders.js';
+
+const originalFetch = window.fetch;
```
overmyheadandbody, line 20: I know we use this approach in the global-navigation testing as well, but couldn't we just use `.restore()` in the `afterEach` call?

test/blocks/global-footer/global-footer.test.js
```diff
@@ -0,0 +1,247 @@
...
+      it('should render the footer after 3s when outside of the 300px range of the viewport', async () => {
+        insertDummyElementOnTop({ height: window.innerHeight + 400 });
+
+        await createFullGlobalFooter({ waitForDecoration: false });
+
+        Object.keys(allSelectors).forEach((key) => expect(
+          document.querySelector(allSelectors[key]) instanceof HTMLElement,
+        ).to.equal(false));
+
+        clock.tick(3000);
```
overmyheadandbody, line 108: Do we need to be this specific or would `await clock.runAllAsync();` also do the trick?

test/blocks/global-footer/global-footer.test.js
```diff
@@ -0,0 +1,247 @@
...
+      if (url.includes('/placeholders')) return mockRes({ payload: placeholders });
+
+      if (url.includes('icons.svg')) return mockRes({ payload: icons });
+
+      return null;
+    });
+  });
+
+  afterEach(() => {
+    clock.restore();
```
overmyheadandbody, line 49: Should we also clear the body content here to ensure we're not polluting any other tests after the footer suite is completed?

test/blocks/global-footer/global-footer.test.js
```diff
@@ -0,0 +1,247 @@
...
+          document.querySelector(containerSelector),
+        )).to.equal(true);
+      });
+    });
+
+    describe('region picker tests', () => {
+      it('should handle non-empty hash', async () => {
+        await createFullGlobalFooter({ waitForDecoration: true });
+
+        const regionPickerElem = document.querySelector('.feds-regionPicker');
```
overmyheadandbody, line 162: I see this `.feds-regionPicker` is being used a few times, could it be added to the selectors list?

test/blocks/global-footer/global-footer.test.js
```diff
@@ -0,0 +1,247 @@
...
+        regionPickerElem.dispatchEvent(new Event('click'));
+
+        expect(regionPickerElem.getAttribute('href') === '#langnav').to.equal(true);
+        expect(regionPickerElem.hasAttribute('aria-expanded')).to.equal(true);
+        expect(regionPickerElem.getAttribute('aria-expanded') === 'true').to.equal(true);
```
overmyheadandbody, line 167: Could these not be grouped together and check `getAttribute('aria-expanded').to.equal('true')`? Same for L192-193

test/blocks/global-footer/test-utilities.js
```diff
@@ -0,0 +1,93 @@
...
+    'featuredProducts',
+    'featuredProductsLabel',
+    'navLink',
+    'navLinkImage',
+    'navLinkContent',
+    'navLinkTitle',
+    'menuItems',
+    'menuHeadline',
+  ].includes(key)),
+);
```
overmyheadandbody, line 43: I think the logic is a bit too convoluted. What if we went for something more like
```
const getMobileVisibleSelectors = () => {
    const { footerIcons, featuredProducts, featuredProductsLabel, navLink, navLinkImage, navLinkContent, navLinkTitle, menuItems, menuHeadline, ...mobileSelectors } = allSelectors;
    return mobileSelectors;
}
export const visibleSelectorsMobile = getMobileVisibleSelectors();
```

overmyheadandbody (commented):
Looking good overall, but I do have a few notes. Also, could you add the coverage report to the PR description in the Screenshots section?

I'm glad to see we're finally adding tests for the footer, good job!

test/blocks/global-footer/test-utilities.js
```diff
@@ -0,0 +1,93 @@
...
+
+  const instance = initFooter(document.querySelector('footer'));
+  if (waitForDecoration) await waitForFooterToDecorate();
+  return instance;
+};
+
+export const insertDummyElementOnTop = ({ height }) => {
+  const dummyElement = document.createElement('div');
+  dummyElement.style.height = `${height}px`;
+  document.body.insertBefore(dummyElement, document.body.firstChild);
```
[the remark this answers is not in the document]
robert-bogos: Done 👍

test/blocks/global-footer/test-utilities.js
```diff
@@ -0,0 +1,93 @@
...
+export const isElementVisible = (elem) => {
+  const computedStyle = window.getComputedStyle(elem);
+  return !!(
+    (elem.offsetWidth && elem.offsetHeight)
+  && elem.getClientRects().length
+  && computedStyle.getPropertyValue('visibility') !== 'hidden')
+  && computedStyle.getPropertyValue('opacity') > 0;
+};
+
+export const allElementsVisible = (givenSelectors, parentEl) => Object
```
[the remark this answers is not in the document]
robert-bogos: Done 👍

test/blocks/global-footer/global-footer.test.js
robert-bogos: Done 👍 Yes, it's cleaner with `.restore()` in the `afterEach`

test/blocks/global-footer/global-footer.test.js
robert-bogos: In here I think we need to be specific because the test is checking if the footer is waiting for 3s before rendering, when outside of the 300px range of the viewport. This 3s seconds wait is a feature of the footer.

test/blocks/global-footer/global-footer.test.js
robert-bogos: Yes, makes sense

test/blocks/global-footer/global-footer.test.js
robert-bogos: Done 👍

test/blocks/global-footer/global-footer.test.js
robert-bogos: Done 👍

mokimo:
> the /favicon.ico  seems to only show on first load, before the tests start. A solution would be to mock this call, but it will add more code and no extra functionality.

Can you just add a `/favicon.ico` mock file under `/tests` ? Or mocking the fetch, whatever is the fastest route, since there's no real benefit other than cleaner logs

> the /undefined calls come from the test where the region picker's href has an empty hash and that causes these 404s in the await loadBlock(regionSelector) from global-footer.js, but it looks like the expected behaviour. Could mock these calls just to get rid of the 404s.

Preferably no 404s, if it takes too long to mock it's fine to leave it as is IMO.

test/blocks/global-footer/test-utilities.js
```diff
@@ -0,0 +1,93 @@
+import { setViewport } from '@web/test-runner-commands';
+import { setConfig } from '../../../libs/utils/utils.js';
+import { config, loadStyles, viewports } from '../global-navigation/test-utilities.js';
+import { waitForElement } from '../../helpers/waitfor.js';
+
+export const containerSelector = '.global-footer';
```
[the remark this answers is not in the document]
robert-bogos: This came with another small change in the logic, but it's done 👍

test/blocks/global-footer/test-utilities.js
```diff
@@ -0,0 +1,93 @@
...
+
+export const allSelectors = {
```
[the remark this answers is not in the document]
robert-bogos: The best solution I found, to avoid refactoring too much, is to import the values from the gnav utilities in the footer utilities and use them. In this way, if `.feds-navLink` now needs to be `.feds-nav-link`, as in your example, there's just one place that this change needs to be done.

`import { selectors as keyboardSelectors } from '../../../libs/blocks/global-navigation/utilities/keyboard/utils.js';

export const allSelectors = {
  container: keyboardSelectors.globalFooter,
  footerIcons: '.feds-footer-icons',
.......`

web-test-runner.config.mjs
```diff
@@ -44,6 +44,7 @@ export default {
   testRunnerHtml: (testFramework) => `
     <html>
       <head>
+      <link rel="icon" href="/libs/img/favicons/favicon.ico" size="any">
```
mokimo, line 47: indentation is off

test/blocks/global-footer/mocks/placeholders.js
```diff
@@ -0,0 +1,106 @@
+const placeholders = {
```
overmyheadandbody, line 1: Do we actually need this redefined, since it should already be available in https://github.com/adobecom/milo/blob/main/test/blocks/global-navigation/mocks/placeholders.js? Could we not re-use that one?

test/blocks/global-footer/test-utilities.js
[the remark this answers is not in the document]
overmyheadandbody: I think this is a good approach overall, as we don't have to duplicate selectors 👍

test/blocks/global-footer/test-utilities.js
```diff
@@ -0,0 +1,97 @@
...
+    container,
+    footerIcons,
+    featuredProducts,
+    featuredProductsLabel,
+    navLink,
+    navLinkImage,
+    navLinkContent,
+    navLinkTitle,
+    menuItems,
+    menuHeadline, ...mobileSelectors
```
overmyheadandbody, line 44: Could we move `...mobileSelectors` to  a new line? Does ESLint not complain because of the lack of a final comma?

overmyheadandbody (commented):
Looking good, just a few more notes. Also, looking at the coverage report, I see that just a bit over 50% of the branches have been covered. For the global-navigation tests, that number is closer to 90%. Is there any way we could bring that number up or are there some non-obvious limitations?

test/blocks/global-footer/test-utilities.js
robert-bogos: Missed that one, weird that ESLint did not complain

overmyheadandbody (approved):
Awesome, I'm glad we're in the green with all the coverage metrics now

mokimo (approved):

PR #1472, opened 2023-11-01
MWPW-138106: b.a.com Blog category titles fix

JasonHowellSlavin:
* Checks for the meta property of category for card titles in the article feed
* Sanitizes taxonomy properties and params passed into get method to avoid case sensitivity issues. 

Additional context: 
We may want to revisit this code since there are now two business logics being used by it, not to mention the new taxonomy pattern introduced with the tag selector. 

There should be no impact on brand blog's recommended articles or cards 

Resolves: [MWPW-138106](https://jira.corp.adobe.com/browse/MWPW-138106)

**Test URLs:**

*Dx Blog*
- Before: https://main--bacom-blog--adobecom.hlx.page/uk/drafts/methomas/?martech=off
- Before: https://main--bacom-blog--adobecom.hlx.page/blog/?martech=off
- After: https://main--bacom-blog--adobecom.hlx.page/uk/drafts/methomas/?milolibs=blog-category-fix&martech=off
- After: https://main--bacom-blog--adobecom.hlx.page/blog/?milolibs=blog-category-fix&martech=off

*Brand Blog*
- Before: https://main--blog--adobecom.hlx.page/en/publish/2023/10/26/celebrating-those-setting-trend-in-digital-inclusivity?martech=off
- Before: https://main--blog--adobecom.hlx.page/en?martech=off
- After: https://main--blog--adobecom.hlx.page/en/publish/2023/10/26/celebrating-those-setting-trend-in-digital-inclusivity?milolibs=blog-category-fix&martech=off
- After: https://main--blog--adobecom.hlx.page/en?milolibs=blog-category-fix&martech=off

### DIFF
--- a/libs/blocks/recommended-articles/recommended-articles.js
+++ b/libs/blocks/recommended-articles/recommended-articles.js
@@ -20,8 +20,7 @@ async function getArticleDetails(article) {
   const ending = trimEndings.find((el) => title.endsWith(el));
   [title] = title.split(ending);
 
-  const metaCategory = getMetadata('category', doc);
-  const category = metaCategory || getMetadata('article:tag', doc);
+  const category = getMetadata('category', doc) || getMetadata('article:tag', doc);
 
   return {
     title,
--- a/libs/blocks/recommended-articles/recommended-articles.js
+++ b/libs/blocks/recommended-articles/recommended-articles.js
@@ -20,10 +20,13 @@ async function getArticleDetails(article) {
   const ending = trimEndings.find((el) => title.endsWith(el));
   [title] = title.split(ending);
 
+  const metaCategory = getMetadata('category', doc);
+  const category = metaCategory || getMetadata('article:tag', doc);
+
   return {
     title,
     path,
-    category: getMetadata('article:tag', doc),
+    category,
     description: getMetadata('description', doc),
     imageEl: doc.querySelector('picture'),
     date: getMetadata('publication-date', doc),
--- a/libs/scripts/taxonomy.js
+++ b/libs/scripts/taxonomy.js
@@ -68,6 +68,7 @@ async function fetchTaxonomy(target) {
 }
 
 function parseTaxonomyJson(data, root, route) {
+  console.log(data, route, root)
   let level1; let level2;
   return data?.reduce((taxonomy, row) => {
     const level3 = removeLineBreaks(row[TAXONOMY_FIELDS.level3]);
@@ -83,7 +84,7 @@ function parseTaxonomyJson(data, root, route) {
       : (level2 ? LEVEL_INDEX.level2
         : LEVEL_INDEX.level1);
 
-    const name = level3 || level2 || level1;
+    const name = level3?.toLowerCase() || level2?.toLowerCase() || level1?.toLowerCase();
     const category = row[TAXONOMY_FIELDS.type]?.trim().toLowerCase() || INTERNALS;
 
     // skip duplicates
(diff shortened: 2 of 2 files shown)


libs/scripts/taxonomy.js
```diff
@@ -83,7 +83,7 @@ function parseTaxonomyJson(data, root, route) {
       : (level2 ? LEVEL_INDEX.level2
         : LEVEL_INDEX.level1);
 
-    const name = level3 || level2 || level1;
+    const name = level3?.toLowerCase() || level2?.toLowerCase() || level1?.toLowerCase();
```
meganthecoder, line 86: Nit: seems cleaner to lowercase `name` after it's set but not a big deal

meganthecoder (approved):

libs/blocks/recommended-articles/recommended-articles.js
```diff
@@ -20,10 +20,13 @@ async function getArticleDetails(article) {
   const ending = trimEndings.find((el) => title.endsWith(el));
   [title] = title.split(ending);
 
+  const metaCategory = getMetadata('category', doc);
+  const category = metaCategory || getMetadata('article:tag', doc);
```
mokimo, line 24:
```suggestion
  const category = getMetadata('category', doc) || getMetadata('article:tag', doc);
```

libs/blocks/recommended-articles/recommended-articles.js
```diff
@@ -20,10 +20,13 @@ async function getArticleDetails(article) {
   const ending = trimEndings.find((el) => title.endsWith(el));
   [title] = title.split(ending);
 
+  const metaCategory = getMetadata('category', doc);
```
mokimo, line 23: If `category` is a new metadata field, I'd make it more specific to this block, to avoid naming clashes with other blocks

libs/scripts/taxonomy.js
mokimo, line 86:
```suggestion
    const name = (level3 || level2 || level1).toLowerCase()
```

mokimo (approved):
Ensure the metadata is properly documented: https://milo.adobe.com/docs/authoring/metadata

libs/scripts/taxonomy.js
JasonHowellSlavin: Good catch, thank you both!

libs/blocks/recommended-articles/recommended-articles.js
JasonHowellSlavin: Much cleaner, thank you.

libs/blocks/recommended-articles/recommended-articles.js
JasonHowellSlavin: I definitely agree, but this is kind of an interesting issue. So this metadata property does not exist in `brand-blog`'s content. They use the first meta article tag on each page to set labels. The `category` metadata property exists in `dx-blog` on every article currently authored and was used for their labels, meaning it is already a locked in convention. 

I am not so sure recommended articles is or will be used elsewhere. Would also like @rgclayton to weigh in since he knows more about `brand-blog`

rgclayton: @JasonHowellSlavin
Correct, blog.adobe.com uses the first tag authored/on the page. 

I understand @mokimo's point.  While I don't know of recommended articles being used elsewhere, I think it might be safe to give `category` a bit more meaning to avoid possible naming clash. What about something like `article-category` to help identify what's going on?

rgclayton (approved):

libs/blocks/recommended-articles/recommended-articles.js
JasonHowellSlavin: After speaking to Ryan, we have the go ahead for merging this. Essentially, every article that business-blog has already uses the `category` metadata property, and we cannot at this juncture bulk update all of those pages. 

However, I have created a ticket to address the issue, and discuss switching to a different property name and paradigm to address Okan's and Ryan's points.

PR #1473, opened 2023-11-01
MWPW-137301 XXL Button Style

meganthecoder:
* Add styles for new "xxl-button" variant
* Usage: adding the ""xxl-button" variant to the block (e.g. Marquee, Aside) will override the default button styles in favor of the XXL button styles
* (Note: This is colloquially called the "mega" button, but we decided that using a t-shirt size was better for continuity)

Resolves: [MWPW-137301](https://jira.corp.adobe.com/browse/MWPW-137301)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/methomas/mega-button?martech=off
- After: https://methomas-mega-button--milo--adobecom.hlx.page/drafts/methomas/mega-button?martech=off


libs/styles/styles.css
```diff
@@ -345,6 +345,14 @@
   width: 100%;
 }
 
+.xxl-button .con-button {
```
JasonHowellSlavin, line 348: the button classes usually have the button first, i.e. `button-l`, `button-xl`, may want to follow that pattern?

JasonHowellSlavin (commented):
I know this muddies the waters, but the addition of this class reveals perhaps the need to  modify how `decorateButtons` is called in marquee.js, since we have a new size. If you did refactor there, you could follow the same class convention `.con-button.button-xxl` instead of `.xxl-button .con-button`. But I suppose that depends on the use cases for this.

libs/styles/styles.css
meganthecoder: `.xxl-button` is what Ryan Clayton, Ryan Parrish, and I agreed on. The reasoning is because that follows the pattern for other block variants `.xs-spacing`, `.xxs-gap`, `.l-heading`, etc

Brandon32 (approved):

ryanmparrish (approved):

meganthecoder:
> > I know this muddies the waters, but the addition of this class reveals perhaps the need to modify how `decorateButtons` is called in marquee.js, since we have a new size. If you did refactor there, you could follow the same class convention `.con-button.button-xxl` instead of `.xxl-button .con-button`. But I suppose that depends on the use cases for this.
> 
> @ryanmparrish I believe this is more like your original idea, whereas I thought it was simpler just to have the css override. I'm curious what other people think CC @rgclayton @chrischrischris

@JasonHowellSlavin Since Ryan approved, I'm going to move forward with this approach

PR #1479, opened 2023-11-02
MWPW-135425 - Remove icon width limit on small notif aside

elan-tbx:
* This PR removes the fixed-width rule set on icons in small notification asides above 600px (see test URLs below for visual of the problem)
* This PR also fixes various style-linting errors in aside.css

Resolves: [MWPW-135425](https://jira.corp.adobe.com/browse/MWPW-135425)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ebartholomew/aside-notification?martech=off
- After: https://ebartholomew-aside-notif-icon-fix--milo--adobecom.hlx.page/drafts/ebartholomew/aside-notification?martech=off


Sartxi (approved):

libs/blocks/aside/aside.css
```diff
@@ -859,7 +859,6 @@
   }
 
   .aside.notification.small .foreground.container .text .icon-area {
-    width: 40px;
```
overmyheadandbody, line 862: Should there be a max-width, though?

overmyheadandbody (approved):

libs/blocks/aside/aside.css
elan-tbx: @overmyheadandbody there actually is one that cascades to this variant, rule is on line 199 of aside.css

PR #1493, opened 2023-11-03
MWPW-129984 Contextual Search without Helpx

Brandon32:
* Prevent Enter from creating a new tab when advanced search is not set.
* Add no results placeholder.

Resolves: [MWPW-129984](https://jira.corp.adobe.com/browse/MWPW-129984)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/bmarshal/search/contextual-search?martech=off
- After: https://bmarshal-search-results--milo--adobecom.hlx.page/drafts/bmarshal/search/contextual-search?martech=off

**Bacom Blog**
- Before: https://main--bacom-blog--adobecom.hlx.page/drafts/bmarshal/?martech=off
- After: https://main--bacom-blog--adobecom.hlx.page/drafts/bmarshal/?martech=off&milolibs=bmarshal-search-results

**Blog**
- Before: https://main--blog--adobecom.hlx.page/?martech=off
- After: https://main--blog--adobecom.hlx.page/?martech=off&milolibs=bmarshal-search-results

### DIFF
--- a/libs/blocks/gnav/gnav-contextual-search.js
+++ b/libs/blocks/gnav/gnav-contextual-search.js
@@ -1,6 +1,7 @@
 import { fetchBlogArticleIndex } from '../article-feed/article-feed.js';
 import { getArticleTaxonomy, buildArticleCard } from '../article-feed/article-helpers.js';
-import { createTag } from '../../utils/utils.js';
+import { createTag, getConfig } from '../../utils/utils.js';
+import { replaceKey } from '../../features/placeholders.js';
 
 let abortController;
 let articles = [];
@@ -97,7 +98,8 @@ export default async function onSearchInput({ value, resultsEl, searchInputEl, a
 
   if (currentSearch === lastSearch) {
     if (!hits.length) {
-      const emptyMessage = createTag('p', {}, 'No results');
+      const noResults = await replaceKey('no-results', getConfig());
+      const emptyMessage = createTag('p', {}, noResults);
       let emptyList = createTag('li', null, emptyMessage);
       if (advancedSearchEl) {
         const advancedLink = advancedSearchEl.querySelector('a');
--- a/libs/blocks/gnav/gnav-contextual-search.js
+++ b/libs/blocks/gnav/gnav-contextual-search.js
@@ -97,11 +97,16 @@ export default async function onSearchInput({ value, resultsEl, searchInputEl, a
 
   if (currentSearch === lastSearch) {
     if (!hits.length) {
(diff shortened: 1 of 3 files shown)


libs/blocks/gnav/gnav-contextual-search.js
```diff
@@ -97,11 +97,16 @@ export default async function onSearchInput({ value, resultsEl, searchInputEl, a
 
   if (currentSearch === lastSearch) {
     if (!hits.length) {
-      const advancedLink = advancedSearchEl.querySelector('a');
-      const href = new URL(advancedLink.href);
-      href.searchParams.set('q', value);
-      advancedLink.href = href.toString();
-      resultsEl.replaceChildren(advancedSearchEl);
+      const emptyMessage = createTag('p', {}, 'No results');
```
meganthecoder, line 100: Should "no results" be a placeholder value so it can be translated?

Brandon32: Updated, thanks!

meganthecoder (approved):

mokimo (approved):

libs/blocks/gnav/gnav-contextual-search.js
```diff
@@ -97,11 +98,17 @@ export default async function onSearchInput({ value, resultsEl, searchInputEl, a
...
-      advancedLink.href = href.toString();
-      resultsEl.replaceChildren(advancedSearchEl);
+      const noResults = await replaceKey('no-results', getConfig());
+      const emptyMessage = createTag('p', {}, noResults);
+      let emptyList = createTag('li', null, emptyMessage);
+      if (advancedSearchEl) {
+        const advancedLink = advancedSearchEl.querySelector('a');
+        const href = new URL(advancedLink.href);
+        href.searchParams.set('q', value);
+        advancedLink.href = href.toString();
```
overmyheadandbody, line 108: Would using the `href` property not achieve the same result?

overmyheadandbody (commented):
I know that `gnav.js` is sort of in maintenance mode, but the ESLint errors seem quite straightforward to address, I'd vote to fix those so the PR checks pass

libs/blocks/gnav/gnav-contextual-search.js
saugatmalla: @overmyheadandbody are you suggesting something like this?
```
const advancedLink = advancedSearchEl.querySelector('a');
const { href } = advancedLink
advancedLink.href = `${href}?q=${value}`
```
In this case, we would be ignoring the existing params.

PR #1494, opened 2023-11-06
fix: enforce origin of experiment blocks

ramboz:
When running experiments that modify blocks, the resulting path for the new block is relative to the origin the import is run on, which means that we always try to load from the milo origin even when the block is in the project, unless we specify the whole URL in the config… but that leads to possible CORS issues during dev/testing since we would try to load from different hosts.

To fix the issue, we explicitly re-use the `base` path that contains either the milo or the project origin so the import happens from the right origin.

Resolves: [MWPW-138857](https://jira.corp.adobe.com/browse/MWPW-138857)

**Test URLs:**
- Before: https://franklin-experimentation--homepage--ramboz.hlx.page/drafts/ramboz/index-loggedout?experiment=demo%2Fchallenger-1
- After: https://franklin-experimentation--homepage--ramboz.hlx.page/drafts/ramboz/index-loggedout?experiment=demo%2Fchallenger-1&milolibs=fix-expblocks (check the homepage-brick-alt block resolution)

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -421,7 +421,7 @@ export async function loadBlock(block) {
   const expBlock = checkForExpBlock(name, expBlocks);
   if (expBlock) {
     name = expBlock.blockName;
-    path = expBlock.blockPath;
+    path = `${base}${expBlock.blockPath}`;
   }
 
   const blockPath = `${path}/${name}`;
--- a/test/features/personalization/personalization.test.js
+++ b/test/features/personalization/personalization.test.js
@@ -118,6 +118,7 @@ describe('Functional Test', () => {
     setFetchResponse(manifestJson);
 
     await applyPers([{ manifestPath: '/path/to/manifest.json' }]);
+
     expect(getConfig().expBlocks).to.deep.equal({ myblock: '/test/features/personalization/mocks/myblock' });
     const myBlock = document.querySelector('.myblock');
     expect(myBlock.textContent?.trim()).to.equal('This block does not exist');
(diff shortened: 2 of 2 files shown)


ramboz:
Screenshot 2023-11-06 at 11 16 22 AM
Screenshot 2023-11-06 at 11 15 59 AM

yesil:
@ramboz some personalization tests are failing now.

ramboz:
@yesil 👍🏼 Should be fixed now

auniverseaway (dismissed):

auniverseaway:
@ramboz I think this opens up a question about how personalization works in general.

I think part of the original requirements were that all experiments were centralized so they could be rolled out instantly to all consumers.

Your PR flips this behavior in that milo is assumed to never have the experiment. This means cross-project tests don't work any more unless that experiment lives in each project... which we don't want.

I think we need _something_ to determine if the test lives in milo or if the test lives in a consumer. This is why we maintain a list of what blocks milo supports... so we can say what are project blocks and what are milo blocks.

ramboz:
@auniverseaway I didn't change the existing logic that detects if the block that is targeted is Milo or project-level. I just properly handled the case when it is a project-level block experiment. The default global Milo behavior should still be in place and working.

auniverseaway:
@ramboz I think you misunderstand. With this change, **_all_** experiments are now coming from the consumer... which they should not be. I can see that vanilla non-experiment blocks are untouched.

ramboz:
@auniverseaway Indeed, I think I don't understand your point. The existing code properly separated milo blocks from project blocks and was adjusting the mapping accordingly for blocks that were part of an experiment.

That part didn't change.

Is your concern that until now project blocks had to be "experimented" on the milo codebase and will now be able to be customized at the project level with the change?

honstar:
@ramboz , what's the status on this PR? The last update from you was last year November. Please review, thanks!

PR #1496, opened 2023-11-06
MWPW-137073: Github action to sync repos

sukamat:
- Uses newly created [Floodgate Repo Sync](https://github.com/apps/floodgate-repo-sync) github app to generate the auth token and uses that to sync the source and the equivalent FG repo as setup in the script
- The workflow runs only on demand and can be triggered from the Actions tab as and when needed
- More details on setup and usage in the wiki -> [Floodgate Repo Sync Github App](https://wiki.corp.adobe.com/display/WP4/Floodgate+Repo+Sync+Github+App)

Resolves: [MWPW-137073](https://jira.corp.adobe.com/browse/MWPW-137073)


arshadparwaiz (approved):

raga-adbe-gh (approved):

skumar09 (approved):
Looks good...

@sukamat : Just a thought, since it is a `workflow_dispatch`, can it be made generic to push (sync) to any repository by parameterizing the repo

.github/workflows/fg-sync-repos.yml
```diff
@@ -0,0 +1,46 @@
...
+        run: |
+          cd ../milo-pink
+          git config user.email "$FG_SYNC_BOT_EMAIL"
+          git config user.name "$FG_SYNC_BOT_NAME"
+          git status          
+          git remote set-url origin https://oauth2:$GITHUB_TOKEN@github.com/adobecom/milo-pink.git
+          git remote -v
+          git add .
+          git commit -m "Syncing milo to milo-pink"
+          git push origin main --force
```
mokimo, line 42: Should we force-push milo pink or create PRs and have a manual step in between to accept/merge those PRs?

mokimo (approved):

sukamat:
> Looks good...
> 
> @sukamat : Just a thought, since it is a `workflow_dispatch`, can it be made generic to push (sync) to any repository by parameterizing the repo

That will be convenient but it will also need another intermediate manual step to raise a PR instead of merging directly to avoid accidental syncs. At this time there is no use-case or need to have that manual step in place but perhaps in the near future if things change.

.github/workflows/fg-sync-repos.yml
sukamat: There is no requirement/need to have that intermediate manual step right now. Perhaps in the near future.

PR #1497, opened 2023-11-06
MWPW-138488 - Arabic font family

rgclayton:
Updated Arabic pages to use `adobe-arabic` font family.

* Corrected AR to point to Arabic languages.
* Updated typekit for `adobe-arabic`
* Updated Argentina country code to "ARG" and lang code to "ar_es" for Spanish speaking Argentina
* Bumped up font size for Arabic to be on par with legacy pages.

Resolves: [MWPW-138488](https://jira.corp.adobe.com/browse/MWPW-138488)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/ar/font-test?martech=off
- After: https://arabic--milo--adobecom.hlx.page/ar/font-test?martech=off
- Before: https://main--milo--adobecom.hlx.page/ae_ar/font-test?martech=off
- After: https://arabic--milo--adobecom.hlx.page/ae_ar/font-test?martech=off


libs/styles/styles.css
```diff
@@ -203,97 +212,173 @@
   line-height: var(--type-heading-xxxl-lh);
 }
 
+:lang(ar) .heading-xxxl  {
```
overmyheadandbody, line 215: Do we need these changes exclusively for this subset of elements, as in `.heading-<size>`, `.body-<size>`, `.detail-<size>`? Or should these updates happen for all elements on the page using these variables? If the latter is true, then could we not define all the font variable values based on the multiplier? Something like:
```
/* Declare the font size multiplier for any locale that needs it */
:root:lang(ar) {
  --font-size-multiplier: 1.25;
}

/* Adapt all font sizes, not just for a subset of elements */
:root {
  --type-heading-xxxl-size: calc(var(--font-size-multiplier, 1) * 80px); /* previously the value was just 80px */
  ...
}

/* Nothing changes compared to the current declarations, because the variable itself has been updated */
.heading-xxxl {
  font-size: var(--type-heading-xxxl-size);
  line-height: var(--type-heading-xxxl-lh);
}
```

This might be useful if other fonts need a multiplier. And this will ensure all elements on the page are consistent, not just a subset. However, since we'd be adding `calc` to all typography values, this might lead to an undesirable overhead, it should be tested out.

rgclayton: I decided to implement this approach. 

[Doing some research related to calc() and performance](https://blog.jiayihu.net/css-custom-properties-performance-in-2018/#:~:text=The%20several%20ways%20of%20using%20calc%20with%20CSS%20variables%20don%27t%20have%20any%20actual%20performance%20difference!%20Why%20are%20the%20metrics%20contrasting%20on%20jsperf.org%20and%20Codepen%3F). There doesn't appear to be an impact.
This is also a good read and something to keep in mind while using CSS Custom Properties (CSS vars)

echampio-at-adobe (approved):
Fonts certainly appear readable to me. When I copied content from a sample page on helpx, for example, content viewed from your arabic branch seems much better to my eye than the original. 

https://helpx.adobe.com/mena_ar/photoshop/using/workspace-basics.html

https://arabic--milo--adobecom.hlx.page/drafts/echampio/arabic-test

libs/styles/styles.css
rgclayton: @overmyheadandbody Bumping this for any follow up you might have.

rgclayton:
> Fonts certainly appear readable to me. When I copied content from a sample page on helpx, for example, content viewed from your arabic branch seems much better to my eye than the original.
> 
> https://helpx.adobe.com/mena_ar/photoshop/using/workspace-basics.html
> 
> https://arabic--milo--adobecom.hlx.page/drafts/echampio/arabic-test

Thanks for taking a look @echampio-at-adobe!

overmyheadandbody (approved):
Looks good, thanks for looking into this! Other languages that require font size scaling should now have an easier time doing it across the board with just a small update.

ryanmparrish (approved):

ryanmparrish:
@rgclayton  - TGTM `verified`

PR #1500, opened 2023-11-07
MWPW-138185 Update Marketo Intersection Observer

Brandon32:
* Improve form performance by loading earlier.

Resolves: [MWPW-138185](https://jira.corp.adobe.com/browse/MWPW-138185)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/bmarshal/marketo?martech=off
- After: https://bmarshal-marketo-loading--milo--adobecom.hlx.page/drafts/bmarshal/marketo?martech=off

**BACOM**
- Before: https://main--bacom--adobecom.hlx.page/resources/state-of-cx-research-report?martech=off
- After: https://main--bacom--adobecom.hlx.page/resources/state-of-cx-research-report?milolibs=bmarshal-marketo-loading

### DIFF
--- a/libs/blocks/marketo/marketo.js
+++ b/libs/blocks/marketo/marketo.js
@@ -15,6 +15,7 @@
  */
 import { parseEncodedConfig, loadScript, createTag, createIntersectionObserver } from '../../utils/utils.js';
 
+const ROOT_MARGIN = 1000;
 const FORM_ID = 'form id';
 const BASE_URL = 'marketo host';
 const MUNCHKIN_ID = 'marketo munckin';
@@ -199,5 +200,6 @@ export default function init(el) {
     callback: (target) => {
       loadMarketo(target, formData);
     },
+    options: { rootMargin: `${ROOT_MARGIN}px` },
   });
 }


libs/blocks/marketo/marketo.js
```diff
@@ -199,5 +200,6 @@ export default function init(el) {
     callback: (target) => {
       loadMarketo(target, formData);
     },
+    options: { rootMargin: `${ROOT_MARGIN}px` },
```
JasonHowellSlavin, line 203: If root margin is not used anywhere else could store it as just a string and omit the template literal?

JasonHowellSlavin (approved):

libs/blocks/marketo/marketo.js
Brandon32: You could, however there's 5 other blocks using this format and it makes sense to keep `ROOT_MARGIN` as an int.

meganthecoder (approved):

libs/blocks/marketo/marketo.js
```diff
@@ -15,6 +15,7 @@
  */
 import { parseEncodedConfig, loadScript, createTag, createIntersectionObserver } from '../../utils/utils.js';
 
+const ROOT_MARGIN = 1000;
```
mokimo, line 18: Any insight on how this was chosen?

mokimo (approved):

libs/blocks/marketo/marketo.js
overmyheadandbody: Since the footer is the last element on the page, with low interaction, and we load it after a set timeout anyway, we had the luxury to hold off a bit more and use a lower rootMargin value. But it is a valid question, a high `rootMargin` value could potentially load things a bit eagerly. Where on the page is this being generally used? How long does it take to load it? That may also justify a higher value, to ensure the section is decorated before the user reaches that point.

overmyheadandbody (approved):

libs/blocks/marketo/marketo.js
Brandon32: This is being used on resource pages and we will be migrating FaaS forms to Marketo forms so value this was taken from the FaaS block. Most of these forms appear above the fold or at the bottom of the page. There is some load time for the form and this value may be too high but it's a good place to start until we have more data.

PR #1502, opened 2023-11-07
MWPW-138178: Block grouping for Milo Loc

sukamat:
- Allows authors to group blocks like tabs, carousel etc. in Docx file so that it is identified as a single entity when the docx is diff'ed for localization

Resolves: [MWPW-138178](https://jira.corp.adobe.com/browse/MWPW-138178)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/sukamat/dash/block-group?martech=off
- After: https://block-group--milo--adobecom.hlx.page/drafts/sukamat/dash/block-group?martech=off


mokimo:
With this change, every page that has this authored would essentially make 4 extra requests, 2 for the start block (html,css) 2 for the end block (html,css) 

I'm not sure this is a great option

sukamat:
@mokimo This was discussed with @auniverseaway offline. There is no better option to achieve this at this point. The alternative is for Franklin to support this at the markdown level which is not in their roadmap.

mokimo (approved):
If you got the blessing from @auniverseaway & @chrischrischris and there are no better ways to use currently, from a PR perspective that's fine then. There's not a lot happening anyway. 

Make sure this is properly documented in the milo docs

honstar:
> With this change, every page that has this authored would essentially make 4 extra requests, 2 for the start block (html,css) 2 for the end block (html,css)
> 
> I'm not sure this is a great option

I agree with @mokimo that these overhead requests just for the sake of adding markers in the DOM is not optimal. Can we please make sure to create a follow-up ticket in JIRA for https://github.com/adobe/helix-docx2md/issues/350?

honstar (approved):

mokimo:
As an idea: should we at least exclude loading those blocks from a milo perspective? If that is straight forward, it might be worth to investigate. We know that a user will never be interested in those blocks.

overmyheadandbody (approved):
Having the extra calls is definitely not ideal, since we'll be affecting the end user's performance to support authoring use-cases. I get this is time-sensitive and the current approach was agreed on, but we should ideally create a follow-up story to investigate other, more performant, options in the future.

Is there any logic on the loc side to ensure a group has both a start and an end and if not, fail gracefully? Has nesting been considered if authors start a new group without ending the previous one?

overmyheadandbody:
I think @mokimo does have a good point. Could we not adapt `loadBlock`, https://github.com/adobecom/milo/blob/main/libs/utils/utils.js#L410, to simply avoid those extra calls to begin with?

mokimo:
In fact, as @overmyheadandbody pointed out adding the `hide-block` modifier is enough to prevent the blocks from ever loading. I've added this for the `block-group-start` and it does not load anymore opposed to the `block-group-end`

Screenshot 2023-11-08 at 12 00 32

We have three good options:
1. Instruct authors to always add the block with the modifier, or add this into the block library with the modifier to begin with
2. Add some lana logs in case the blocks end up being loaded (might be available in the RUM data?)
3. Adapt loadbock to just early return for those two blocks.

nishantka (approved):

sukamat:
Thanks for all the inputs.

> Could we not adapt loadBlock, https://github.com/adobecom/milo/blob/main/libs/utils/utils.js#L410

I like this. Let me see if this works and I will update this PR if it does. 
If not, there is already a Jira ticket in the backlog to follow up on the Franklin request. I will update that to explore some of the other options discussed here.

auniverseaway:
To piggyback on the other comments, rather than make two blocks, I would probably lean into variations...

```
block-group (start, hide-block)

block-group (end, hide-block)
```

My understanding is that you should still be able to find these in your mdast / regional edit flows.

sukamat:
@auniverseaway 
Similar to how "hide-block" is handled in [utils.js](https://github.com/adobecom/milo/blob/main/libs/utils/utils.js#L410), when I add the below in `loadBlock(block)`, it avoids calling the JS/CSS and it also continues to retain that info in the markdown.

```
    ...
    let name = block.classList[0];
    if (name === 'block-group-start' || name === 'block-group-end') {
        block.remove();
	return null;
    }
    ...
}
```

This will keep the group marker simple to add for authors without setting additional info in the block name/header area (especially when adding manually).

I will run some more tests but LMK if there are any objections with this approach. 

Thanks @mokimo @overmyheadandbody

auniverseaway:
@sukamat correct me if I'm wrong, but we're now going to need three checks, right?

```
if (name === 'hide-block' || name === 'block-group-start' || name === 'block-group-end')
```
(obviously this could be cleaned up to look in an array of names)

Utils is precious from a character count standpoint, so my hope would be we could leverage what is already there... i.e.

```
if (name === 'hide-block')
```

And then implied is...
```
hide-block (group-start)
hide-block (group-end)
```

I don't know that I'm in love with it, but I kind of like a "hide block" is actually a block we can do things with. As in there's multiple uses for a hide block.

auniverseaway:
The more I think about this, the more I kind of like it...

"What kind of hide block is this?" 
"A block group"

sukamat:
@auniverseaway hide-block is not a block. It is a class that can be set on any block to remove it from the DOM. But given that authors need something that shows it as start or end, that's why the need for the additional markers.

sukamat:
@auniverseaway As discussed offline, we will go with the below
```
block-group-start (hide-block)
block-group-end (hide-block)
```

This removes the need of creating any new blocks and the hide-block class takes care of removing it from the DOM.

I will work on updating the diff'ing logic in rollout.js to use the above format.

PR #1503, opened 2023-11-07
MWPW-137903 - Geo Routing LCP

joaquinrivero:
* Geo Routing LCP

Resolves: [MWPW-137903](https://jira.corp.adobe.com/browse/MWPW-137903)

**Test URLs:**
- Milo: https://mwpw-137903-b--milo--joaquinrivero.hlx.live?martech=off 
- Before: https://main--homepage--adobecom.hlx.live/homepage/index-loggedout?martech=off&akamaiLocale=MX
- After: https://main--homepage--adobecom.hlx.live/homepage/index-loggedout?&milolibs=mwpw-137903-B--milo--joaquinrivero&martech=off&akamaiLocale=MX

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -38,6 +38,7 @@
 /libs/blocks/quiz/ @colloyd @sabyamon @fullcolorcoder @JackySun9 @elaineskpt @echampio-at-adobe
 /libs/blocks/quiz-results/ @colloyd @sabyamon @fullcolorcoder @JackySun9 @elaineskpt @echampio-at-adobe
 /libs/blocks/quote/ @ryanmparrish
+/libs/blocks/review/ @nkthakur48 @chrischrischris @auniverseaway
 /libs/blocks/table-of-contents/ @Brandon32
 /libs/blocks/tabs/ @ryanmparrish
 /libs/blocks/tag-selector/ @meganthecoder
--- a/libs/blocks/review/review.js
+++ b/libs/blocks/review/review.js
@@ -2,6 +2,7 @@ import { html, render } from '../../deps/htm-preact.js';
 
 import { getMetadata, loadStyle, getConfig } from '../../utils/utils.js';
 import HelixReview from './components/helixReview/HelixReview.js';
+import { checkPostUrl } from './utils/utils.js';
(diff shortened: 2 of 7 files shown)


Axelcureno (approved):

auniverseaway (changes requested):
This PR puts Footer, delayed.js, GeoRouting.js, geo2.adobe.com json, Favicon, etc. in front of LCP and **_makes every other page_** on adobe.com significantly slower.

The premise in the JIRA is a bit off, too. We should never be optimizing for GeoRouting to be LCP as it is a **_major edge case_** in browsing the site for both Google's indexer as well as humans. This is born out in CWV, CRuX and even LH data.

I'm totally fine making a couple images lower res so they load faster, but I think we should not make every page slower for some non-ideal testing conditions.

You only get the GeoRouting pop up when visiting pages cross locale. We have `?geoRouting=off` for testing purposes or people who are out of US should test within the region that matches their locale if they do not want to add the query param. This is no different than `?martech=off`.

This is what this PR moves in front of LCP even with GeoRouting turned off...
Screenshot 2023-11-07 at 5 26 09 PM

@yesil can you share details on why you would want to treat GeoRouting as a valid LCP use case to optimize when Google's same-locale PSI (and CWV) does not treat it as a problem? If this is the case, we are doing all projects wrong and we should only have to optimize for the modal.

**Compare the waterfalls here:**
https://main--bacom--adobecom.hlx.live/?&milolibs=mwpw-137903-B--milo--joaquinrivero&martech=off
VS
https://main--bacom--adobecom.hlx.live/?martech=off

BACOM's homepage LCP went down to 2.9s from 2.2s...
https://pagespeed.web.dev/analysis/https-main--bacom--adobecom-hlx-live/f9fvabi8sd?form_factor=mobile (2.9s)

https://pagespeed.web.dev/analysis/https-main--bacom--adobecom-hlx-live/7n08my6zcd?form_factor=mobile (2.2s)

Axelcureno (changes requested):
Given the impact on `utils.js` pointed out, let's revert the changes in that file and just leave the lower res images.

PR #1504, opened 2023-11-08
MWPW-137866: Web component-based Merch Card in Milo

Axelcureno:
MWPW-137866: Introduces an all-new merch card block built using Lit Web Components.
Introduces Lit library and merch-card lit-based dependencies inside `/deps` folder in Milo.

What's new:
- Faster performance in Merch Card.
- A new text style-based authoring approach for easier and more consistent authoring of Merch Cards.
- Backward compatibility with legacy Merch Card.
- New Catalog Card.
- New column grid designation in section metadata (one merch card, two merch cards, three merch cards, four merch cards).
- More consistent layout.
- Fully responsive.
- Supports cards inside Fragments.
- Fixes other minor issues in Segment, Special Offers and Plans card.

The differentiator between the old merch card and the new one will be the section metadata column grid designation.
Old merch card uses: One up, two up, three up, four up.
New merch card uses: one merch card, two merch cards, three merch cards, four merch cards.

**New text-style based approach**
To better align with Consonant designs and have authored cards be more consistent, this change introduces a new text-style based approach mapping Sharepoint headings (Heading 1, Heading 2, Heading 3, Heading 4, Heading 5) mapped to [Merch Card Consonant styles](https://xd.adobe.com/view/48c7ee2a-60d7-48de-a6d3-d9ecf3bf8092-1d94/screen/3aef0f04-d89e-405b-92fa-8100320cb4f9).

The mapping is the following:
- H5: Detail M
- H4: Body XXS
- H3: Heading XS
- H2: Heading M
- H1: heading L
- Paragraph: Body XS

Catalog Card as an example:

Screenshot 2023-11-07 at 8 28 29 PM

- Backward compatibility scenario to show no impact on existing merch card block: https://mwpw-137866--milo--axelcureno.hlx.page/docs/library/kitchen-sink/merch-card

Resolves: [MWPW-137866](https://jira.corp.adobe.com/browse/MWPW-137866)

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/drafts/axel/merch-card-docs/merch-card
- After: https://main--cc--adobecom.hlx.page/drafts/axel/merch-card-docs/merch-card?milolibs=mwpw-137866--milo--axelcureno&martech=off


libs/blocks/merch-card/merch-card.js
```diff
@@ -1,184 +1,186 @@
...
+  if (upClass) {
+    initLegacyMerchCard(el);
+  } else {
+    if (section.parentElement.classList.contains('fragment')) {
+      section.style.display = 'contents';
+      const fragment = section.parentElement;
+      fragment.style.display = 'contents';
+      const fragmentParent = fragment.parentElement;
+      fragmentParent.style.display = 'contents';
+      section = fragmentParent.parentElement;
```
3ch023, line 106: this is to support fragments, right? 
we could see later if inline fragments can be used instead, but i think it can be a future optimization since we are in a bit of a hurry with this PR.
in case you want already to try it out its just about adding '#_inline' in the end of referenced fragment.

3ch023 (changes requested):
PR needs some brush-ups:
- lint fixes
- both Before/After links are broken ('?' is there twice) so perf check didn't run

@Axelcureno also just noticed that prices look much smaller in live pages with new changes:
[Before](https://main--cc--adobecom.hlx.page/creativecloud/file-types/image)
[After](https://main--cc--adobecom.hlx.page/creativecloud/file-types/image?milolibs=mwpw-137866--milo--axelcureno)
image
image

libs/blocks/merch-card/merch-card.js
```diff
@@ -1,184 +1,186 @@
 /* eslint-disable prefer-destructuring */
 import { decorateButtons, decorateBlockHrs } from '../../utils/decorate.js';
 import { loadStyle, getConfig, createTag } from '../../utils/utils.js';
-import { addBackgroundImg, addWrapper, addFooter } from '../card/cardUtils.js';
+ import { getUpFromSectionMetadata } from '../card/cardUtils.js';
+import { decorateLinkAnalytics } from '../../martech/attributes.js';
 import { replaceKey } from '../../features/placeholders.js';
+import '../../deps/commerce.js';
+import '../../deps/merch-card.js';
+import { initLegacyMerchCard } from './legacy-merch-card.js';
```
yesil, line 9: let's import legacy-merch-card only if needed. Otherwise it will impact performance.

yesil (changes requested):

libs/blocks/merch-card/merch-card.js
Axelcureno: Yes, it's for supporting fragments. I will git it a try.

libs/blocks/merch-card/merch-card.js
Axelcureno: Fixed

Axelcureno:
> PR needs some brush-ups:
> 
> * lint fixes
> * both Before/After links are broken ('?' is there twice) so perf check didn't run
> 
> @Axelcureno also just noticed that prices look much smaller in live pages with new changes: [Before](https://main--cc--adobecom.hlx.page/creativecloud/file-types/image) [After](https://main--cc--adobecom.hlx.page/creativecloud/file-types/image?milolibs=mwpw-137866--milo--axelcureno) image image

@3ch023 all three items fixed.

VKniaz (approved):

3ch023 (approved):

yesil:
Prices are not in the correct slot and do not have the font sizes etc...
image

vs

image

plans card body-xxs has extra bottom-padding
image


prices in different variants such as merch pods are mis placed.

image

cc: @Axelcureno

libs/blocks/merch-card/merch-card.js
```diff
@@ -1,184 +1,188 @@
...
+    const styles = [...el.classList];
+    const cardType = getPodType(styles);
+    const ribbonMetadata = rows[0].children?.length === 2 ? rows[0].children : null;
+    const actionMenuContent = cardType === 'catalog' ? getActionMenuContent(el, ribbonMetadata) : null;
+    const row = getMerchCardRows(rows, ribbonMetadata, cardType, actionMenuContent);
+    const altCta = rows[rows.length - 1].children?.length === 2
+      ? rows[rows.length - 1].children : null;
+    const allPElems = row.querySelectorAll('p');
+    const ctas = allPElems[allPElems.length - 1];
+    images.forEach((img) => {
```
honstar, line 125: We'll probably optimize this in a subsequent PR once this is merged to fetch an image from Fastly that is optimized for the merch-card specs. Currently, it uses the original image that an author has added which might be over-dimensioned and potentially unnecessarily large in size.

honstar:
@Axelcureno : can you please review the code that codecov is complaining about? Also, I think the LH score is being run against `milo` only URLs found in the description. Can you please provide `milo` before and after test URLs with `?martech=off` to run LH scores against an actual test page?

PR #1513, opened 2023-11-08
MWPW-138962:  Bacom-blog Article Feed Issue

JasonHowellSlavin:
* Adds logic looking into the config to determine if the site is bacom blog before reassigning the category prop

Resolves: [MWPW-138962](https://jira.corp.adobe.com/browse/MWPW-138962)

**Test URLs:**
_bacom-blog_
- Before: https://main--bacom-blog--adobecom.hlx.page/drafts/staged-content/blog/tags/adobe-analytics?martech=off
- After: https://main--bacom-blog--adobecom.hlx.page/drafts/staged-content/blog/tags/adobe-analytics?martech=off&milolibs=article-feed-category

_brand blog_
- Before: https://main--blog--adobecom.hlx.page/en/topics/customer-stories?martech=off
- After: https://main--blog--adobecom.hlx.page/en/topics/customer-stories?martech=off&milolibs=article-feed-category

### DIFF
--- a/libs/blocks/article-feed/article-helpers.js
+++ b/libs/blocks/article-feed/article-helpers.js
@@ -90,7 +90,11 @@ function loadArticleTaxonomy(article) {
 
     const articleTax = computeTaxonomyFromTopics(topics, path);
 
-    clonedArticle.category = articleTax.category;
+    const { productionDomain, contentRoot } = getConfig();
+
+    if (productionDomain !== 'business.adobe.com' && contentRoot !== '/blog') {
+      clonedArticle.category = articleTax.category;
+    }
 
     // topics = tags as an array
     clonedArticle.topics = topics;
--- a/libs/blocks/article-feed/article-helpers.js
+++ b/libs/blocks/article-feed/article-helpers.js
@@ -90,11 +90,7 @@ function loadArticleTaxonomy(article) {
 
     const articleTax = computeTaxonomyFromTopics(topics, path);
 
-    const { productionDomain, contentRoot } = getConfig();
-
-    if (productionDomain !== 'business.adobe.com' && contentRoot !== '/blog') {
-      clonedArticle.category = articleTax.category;
-    }
+    clonedArticle.category ??= articleTax.category;
 
     // topics = tags as an array
     clonedArticle.topics = topics;


meganthecoder (approved):
As we discussed, there are no great solutions for this. Checking the domain isn't ideal, but neither is most of the taxonomy.js related code. It could all do with an overhaul at some point, but this will get us past the post for the bacom blog migration.

libs/blocks/article-feed/article-helpers.js
```diff
@@ -90,7 +90,11 @@ function loadArticleTaxonomy(article) {
 
     const articleTax = computeTaxonomyFromTopics(topics, path);
 
-    clonedArticle.category = articleTax.category;
+    const { productionDomain, contentRoot } = getConfig();
+
+    if (productionDomain !== 'business.adobe.com' && contentRoot !== '/blog') {
```
overmyheadandbody, line 95: Since this seems consumer-specific, could it not be sent as a param instead of having a consumer check in Milo?

JasonHowellSlavin: I'd have to look into that, but currently this `article-feed `is tuned for `brand-blog` alone necessitating this change.

The code in this function mutates the original taxonomy data, so maybe there is somewhere farther upstream I could change it. This seemed to be the least intrusive, and now that there is likely another consumer who will be using these blocks, we may want to overhaul this code entirely.

JasonHowellSlavin: @overmyheadandbody based on this discussion: https://github.com/orgs/adobecom/discussions/1507 these files are likely to get highly refactored at a future date to implement any changes to made to the core tagging systems. 

Magma has another requirement that will likely result in a similar addition to these files like the one above: https://jira.corp.adobe.com/browse/MWPW-138963. 

MWPW-138963 and MWPW-138962 are required for the bacom-blog launch happening later this week. I have informed our stakeholders that this will cause tech debt and I will make a ticket based on your observation above. 

@rgclayton do you see any issue adding the above code?

fullcolorcoder: @JasonHowellSlavin Just want to point out that the original issue that caused this ticket was initially done on purpose. From 138963 it says

> the titles in our milo consuming repo are pulled from the meta title

That was done so authors can have control and can show a different title in the card feed vs the page itself. Some titles are way too long and needed rephrased. 

This probably should be handled a different way on import instead, and hopefully those intentions aren't lost.

JasonHowellSlavin: @fullcolorcoder this ticket deals with the tag headers on the article cards:

Screen Shot 2023-11-14 at 3 08 53 PM

 which seem to default to the first meta `article:tag` on the page. The code in question actually reassigns the category property to a modified version of the taxonomy. In this case, what occurs is `bacom-blog`'s authored pages set the `category` meta property, and this gets ignored on the cards in favor of the first `article:tag`, which is not the desired functionality for our bacom blog. 

> That was done so authors can have control and can show a different title in the card feed vs the page itself. Some titles are way too long and needed rephrased.

This is a great point about https://github.com/adobecom/milo/pull/1528 and I will bring it up with our PMs and Authoring. Essentially, half of the content on `bacom-blog` has the truncated title, half doesn't so maybe it is an authoring retouch. 

This PR, the tags, blocks our launch, the other does not, I don't believe.

Brandon32 (approved):

JasonHowellSlavin:
@Brandon32 suggested using the null coalescing operator.

Please re review when you have a second @overmyheadandbody and @fullcolorcoder. I believe this solution is much cleaner, better, and non consumer specific!

libs/blocks/article-feed/article-helpers.js
JasonHowellSlavin: Resolving, changed the code.

meganthecoder (approved):

fullcolorcoder (approved):
@JasonHowellSlavin @Brandon32 nice!

auniverseaway (approved):

rgclayton (approved):

PR #1515, opened 2023-11-09
MWPW-137717 | New block - Basic Brick & Masonry 

aishwaryamathuria:
This is a new Milo block that allows us to create a grid with bricks of size (full, half, thwo-third or one third) widths.
These are arranged into a masonry layout once section metadata is applied. 
The bricks can be authored as fragments or directly on docx and will be arranged via section metadata.
1. The bricks will be authored as individual blocks and then arranged via the masonry layout.
2. The layout is not fixed, the different section layouts can be paired together (like 1:1:1 followed by 1:2 ).
3. Background option for color + image (multiple viewport).
6. Option to author multiple background image for different media ports.
7. Text alignment - left, center, align-center.
8. Default text sizes with override option.
9. Clickable brick (Only if brick has single link)

Adding poc links -
https://masonry--milo--aishwaryamathuria.hlx.page/drafts/mathuria/masonry/poc/poc-1
https://masonry--milo--aishwaryamathuria.hlx.page/drafts/mathuria/masonry/poc/poc-2

Resolves: [MWPW-137717](https://jira.corp.adobe.com/browse/MWPW-137717)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/mathuria/masonry/poc/poc-2?martech=off
- After: https://masonry--milo--aishwaryamathuria.hlx.page/drafts/mathuria/masonry/poc/poc-2?martech=off

### DIFF
--- a/libs/blocks/brick/brick.css
+++ b/libs/blocks/brick/brick.css
@@ -20,17 +20,6 @@
   color: var(--color-white);
 }
 
-.brick .foreground a:not([class]),
-.brick .foreground span.first-link {
-  font-weight: 700;
-}
-
-.brick.rounded-corners,
-.brick.rounded-corners .background,
-.brick.rounded-corners .foreground {
-  border-radius: var(--spacing-xs);
-}
-
 .brick .background {
   position: absolute;
   bottom: 0;
@@ -45,6 +34,12 @@
   display: flex;
   flex-grow: 1;
   padding: var(--spacing-m);
+}
+
+.brick.rounded-corners,
+.brick.rounded-corners .background,
+.brick.rounded-corners .foreground {
+  border-radius: var(--spacing-xs);
 }
 
 .brick.align-center .foreground,
@@ -86,10 +81,6 @@
   display: none;
 }
 
-.brick .foreground .icon-area picture {
-  display: flex;
-}
-
 .brick .foreground p {
   padding: 0;
   margin: 0;
@@ -122,6 +113,10 @@
   margin: 0;
   padding: 0;
   flex-shrink: 0;
+}
+
+.brick .foreground .icon-area picture {
+  display: flex;
 }
 
 .brick .icon-stack-area li img {
@@ -161,6 +156,11 @@
   align-items: center;
   gap: var(--spacing-xs);
   text-align: left;
+}
+
+.brick .foreground a:not([class]),
+.brick .foreground span.first-link {
+  font-weight: 700;
 }
 
 [dir="rtl"] .brick .icon-stack-area li,


auniverseaway (commented):
I'll let others chime in more, but I think `basic-brick` should just be `brick`.

This would be like calling a product, "iPhone Basic" or "MacBook Basic"

I think the omission of "advanced" or "pro" implies this is basic... and all blocks should have basic implied.

libs/blocks/basic-brick/basic-brick.js
```diff
@@ -0,0 +1,97 @@
...
+  if (lastP) lastP.className = 'supplemental-text';
+}
+
+function handleObjectFit(bgRow) {
+  const bgConfig = bgRow.querySelectorAll('div');
+  [...bgConfig].forEach((r) => {
+    const image = r.querySelector('img');
+    if (!image) return;
+    const pchild = r.querySelectorAll(':scope > p');
+    if (pchild.length < 2) return;
```
seanchoi-dev, line 41: You may want to check `<p>` **_only_** if it has _**content**_ to avoid empty `<p>` case.

aishwaryamathuria:
> ould be like calling a product, "iPhone Basic" or "MacBoo

@auniverseaway I have made the required changes for suggestion. The block name is now be 'brick' in place of 'basic-brick'

libs/blocks/basic-brick/basic-brick.js
aishwaryamathuria: @seanchoi-dev I have made the suggested change and am now targetting p:not(:empty) for further processing.

ryanmparrish:
@aishwaryamathuria  - Just a thought, have you considered using the section-metadata `layout: masonry` rather than `style: masonry-layout`? The usage of the layout key is only used w/ 2-5 up fractions, but this seems like a great candidate for this section-meta key to separate a style from a layout. Thoughts?

libs/blocks/section-metadata/section-metadata.css
```diff
@@ -266,3 +266,38 @@ main > .section[class*='-up'] > .content {
...
+@media screen and (min-width: 1200px) {
+  .section.masonry-layout {
+    grid-template-columns: repeat(12, 1fr);
+  }
+
+  .section.masonry-layout > div { grid-column: span 4; }
+  .section.masonry-layout .full-grid {grid-column: span 12; }
+  .section.masonry-layout .two-third-grid { grid-column: span 8; }
+  .section.masonry-layout .half-grid { grid-column: span 6; }
+}
```
JasonHowellSlavin, line 303: nit: keep media queries together, can move these to line 221

libs/blocks/section-metadata/section-metadata.css
```diff
@@ -266,3 +266,38 @@ main > .section[class*='-up'] > .content {
...
+  padding-right: var(--grid-margins-width);
+}
+
+.section.masonry-layout > div:not([class]),
+.section.masonry-layout > div:not([class]) > div.fragment,
+.section.masonry-layout > div:not([class]) > div.fragment > div.section {
+  display: grid;
+}
+
+@media screen and (min-width: 768px) and (max-width: 1199px) {
```
JasonHowellSlavin, line 284: these are pretty specific `px` widths for the queries. Though I can guess the need for `max-width: 1199`, `768` seems awfully device specific.

JasonHowellSlavin (commented):
A few nits, looks good!

libs/blocks/brick/brick.css
```diff
@@ -0,0 +1,233 @@
...
+  text-size-adjust: none;
+  min-height: 300px;
+}
+
+.brick,
+.brick.click a.foreground {
+  color: inherit;
+}
+
+.light.brick,
```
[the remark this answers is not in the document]
aishwaryamathuria: @JasonHowellSlavin I have made the suggested changes.

libs/blocks/brick/brick.css
```diff
@@ -0,0 +1,233 @@
...
+.light.brick.click a.foreground {
+  color: var(--text-color);
+}
+
+.dark.brick,
```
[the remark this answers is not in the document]
aishwaryamathuria: @JasonHowellSlavin I have made the suggested changes.

libs/blocks/brick/brick.css
```diff
@@ -0,0 +1,233 @@
...
+
+@media screen and (min-width: 1200px) {
+  .brick .desktop-only {
+    display: block;
+  }
+
+  .brick.full-grid {
+    min-height: 500px;
+  }
+}
```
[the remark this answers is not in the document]
aishwaryamathuria: @JasonHowellSlavin The display is set to block to make the device specific background visible. Initially the display for `.brick .<device>-only` is set to none and is only the background for that viewport is made visible in media queries.

libs/blocks/section-metadata/section-metadata.css
aishwaryamathuria: @JasonHowellSlavin I have made the suggested changes.

libs/blocks/section-metadata/section-metadata.css
aishwaryamathuria: @JasonHowellSlavin As per the XD the grid layout needs to have a 2 column view only for 768px to 1199px viewport and for smaller tablet screen sizes and mobile screens it needs to be 1 column view

ryanmparrish: @aishwaryamathuria  - It looks like the Figma file is wrong and doesn't use our set breakpoints. I'd talk to Kate and the consonant team and double check this. I'm w/ @JasonHowellSlavin on this one. I wouldn't recommend we use these additional breakpoints. I think as engineers we have to talk some of the designs w/ a grain of salt and resolve the discrepancies between design and reality.

libs/blocks/brick/brick.css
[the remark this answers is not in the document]
ryanmparrish: @aishwaryamathuria  - What about support for a .dark  or .light section? I think we want o add support for this right? 
`.dark .brick`, `.light .brick` etc.

honstar:
Without having looked too deep into the code, I was wondering the following from an authoring perspective: assuming the image is the same across `mobile`, `tablet` and `desktop`, wouldn't it be more authoring friendly if only a single (high-res desktop) image needs to be added to the `brick`? The `brick` block could then leverage the Fastly Image Optimizer service to dynamically request optimized images for each surface. All other surfaces would be optional.

This would also save on the number of DOM elements.

aishwaryamathuria:
> d also save on the number of DOM elemen

@honstar The brick currently supports image for multiple or combined viewports bots. In case same image is required across all viewports the author can in that case add only one image as well.

honstar:
> @honstar The brick currently supports image for multiple or combined viewports both. In case same image is required across all viewports the author can in that case add only one image as well.

Perfect, thanks a lot for pointing this out @aishwaryamathuria 👍 Please make sure to add this to documentation.

libs/blocks/brick/brick.js
```diff
@@ -0,0 +1,105 @@
...
+  bgEl.append(document.createTextNode(focusOption.join(',')));
+}
+
+function handleObjectFit(bgRow) {
+  const bgConfig = bgRow.querySelectorAll('div');
+  [...bgConfig].forEach((r) => {
+    const pic = r.querySelector('picture');
+    if (!pic) return;
+    let text = '';
+    const pchild = r.querySelectorAll('p:not(:empty)');
```
rgclayton, line 49: Should this be trimmed? [According to MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/:empty) whitespace counts as a child.

rgclayton:
> Without having looked too deep into the code, I was wondering the following from an authoring perspective: assuming the image is the same across `mobile`, `tablet` and `desktop`, wouldn't it be more authoring friendly if only a single (high-res desktop) image needs to be added to the `brick`? The `brick` block could then leverage the Fastly Image Optimizer service to dynamically request optimized images for each surface. All other surfaces would be optional.

Authors typically want a whole new image with an aspect ratio that better matches the viewport, not just a better optimized image.

libs/blocks/section-metadata/section-metadata.css
```diff
@@ -189,6 +189,20 @@ main > .section[class*='-up'] > .content {
   margin: 0;
 }
 
+.section.masonry-layout {
```
rgclayton, line 192: I think along the same lines as the `basic-brick` conversation from Chris M, I think `masonry-layout` can be simplified to just `masonry`. Layout is implied unless there are more options related to `masonry`.

rgclayton:
Is it not possible to have more than three bricks on one row? Doing some testing and if I just use the `brick` variant, with no sizing, I can only get three on one row. 

I don't know if we currently need to support all 12 options but I would think support for more than three blocks on a row (up to 6?) would be needed. Maybe I'm wrong. Happy to hear others thoughts here.

cc: @ryanmparrish @auniverseaway

rgclayton:
I also think if we are doing a masonry/grid block we need to come up with more scaleable brick size variants. If we need to support more than 3 elements on a row the 1/3, 2/3 naming starts to get awkward. 

If we are supporting a 12 col grid I think size variants like `brick-1-column` through `brick-11-column`, with `brick-full-width` equalling the full 12 columns, is easy to author and combine. The goal is to equal 12. So, I could easily author a layout like `brick-4-column` with `brick-8-column` to create a 2/3's layout

ryanmparrish:
@aishwaryamathuria  - I agree w/ @rgclayton, we should have good defaults when authoring any number of bricks in a section-style: masonry. I would think that if 4 `bricks` are in a masonry section, they will fill the space equally w out any additional brick variants and if any of the `bricks` are declared to be bigger than the grid will respect that. Im all for the best option, the cooking measurement of `one-third` etc doesn't scale well when we're talking 1-12 potential width options. I feel like authors are used to the 12 grid system so it should come naturally to say `grid-4-column` when in a masonry, or maybe `grid-4-of-12` by clearly stating the bounds. Just some thoughts as Im not totally sold on the [authoring pattern](https://jira.corp.adobe.com/browse/MWPW-137718) just yet.

libs/blocks/section-metadata/section-metadata.css
```diff
@@ -222,6 +236,16 @@ main > .section[class*='-up'] > .content {
   }
 }
 
+@media screen and (min-width: 768px) and (max-width: 1199px) {
```
ryanmparrish, line 239: Shouldn't this be `min-width: 600px`? `768` isn't one of our breakpoint as far as I can remember

aishwaryamathuria: @ryanmparrish As per the XD shared the layout will have a two column grid for media screens 768px to 1199px and below 768px it is aligned to have a one column layout only hence we have used the explicit values.

libs/blocks/brick/brick.js
aishwaryamathuria: @rgclayton I have made the suggested change

rgclayton:
Thanks @salonijain3. Comments inline. 

> Here is my two cents on this matter:
> 
> The [Jira ticket](https://jira.corp.adobe.com/browse/MWPW-137717) explicitly indicates that the masonry block was initially constructed for the homepage, and the objective is to replicate the same functionality in Milo. The authoring format on the homepage masonry block aligns with `two-third` and `full` grid brick layouts.

Sometimes these tickets need to be interpreted with better clarity. They are not definite and they often need more questions asked to get on the right track. Here's what I'm reading from the ticket:

> "As an author in Milo, I want the ability to use Masonry on a **12 column grid** with the [new] basic-brick blocks. "

> Masonry was developed for the priority use case of the homepage and supports the homepage-brick block variants. It remains on the homepage repo **but needs to be contributed to Milo core for extended use cases**. 

The ticket mentions a **12 col grid** that needs to be developed for **extended use cases**.  I would think that's where this additional brick sizing comes in to play (more than three elements on a row). 

> It is worth considering the potential for confusion among authors if we adopt different authoring approaches for the same block. Maintaining consistency in how we author this block can enhance clarity and ease of use.

This is exactly what we are trying to avoid. The homepage has a small handful of authors. Milo has many, many, many, (20x?) more authors. Maintaining consistency for a few and creating difficult patterns for the many isn't a good plan.  

> From a design standpoint, current masonry figma is currently constrained to a minimum one-third grid. Expanding the grid size could pose challenges, as the existing font sizes may not translate seamlessly. Creating smaller grids would necessitate distinct design specifications, adding complexity.

Again, the ticket says to extend the use cases beyond the homepage.  If we are going to do a 12 col masonry/grid block, lets make a 12 col masonry block. 

> Additionally, following Milo Academy's approach, it is important not to address issues until they become tangible problems. Currently, there are no use cases for the masonry block beyond those developed here, and the design for additional grids has not been established.

I agree with this. However, again we are extending the use cases for this and we are also setting up patterns to grow with. If we have bad patterns to start with it makes growing much more difficult. Lets spend the time now to create good patterns to move forward with.

> P.S. From naming perspective for masonry, if we name it as masonry instead of masonry-layout, the css will start conflicting with the homepage css.

I thought the Homepage masonry block is prefixed with `homepage-masonary`? Is that not the case? `masonry-layout` is clumsy. It's already a layout, no need to say it's a layout again.

auniverseaway:
I'm not able to spend a ton of time on this to get in the weeds, but I think there's some general guidance that can be applied here. Again, I'm not even sure where things stand right now.

How homepage built the solution should have _zero_ impact on the implementation in Milo. Why?

1. Homepage specifically has a very small number of pages (yes, including geos & fragments). If there is a conflict with Milo principles (more on this later), homepage has a smaller blast radius than shipping bad patterns in milo.
2. If a consuming team builds something in isolation, or gets feedback from milo and doesn't follow it, that's not Milo's problem when that team tries to contribute it back (not placing blame here, I know Brahmos is trying to port something they didn't originally build).

**Other reminders:**
1. We (Milo) will never ship something we don't agree with, or are not willing to support. Every feature contributed to Milo impacts hundreds of authors, tens of sites, and hundreds of thousands of pages. The overhead on training a non-ideal solution is extremely high. We start creating bad and inconsistent patterns that we then need to train on why things do not follow our previously documented guidelines. We won't always get this right, but we should always try.
2. We should never be in a rush to ship. The cost of being in a rush means bad content, bad docs, bad patterns, bad performance. I've seen this soooo many times.
3. Rule of Chris: never trust JIRA. It's likely wrong.
4. With a few exceptions, blocks should not have dependencies on other blocks. The block should just work and not look weird if you put it on a page with no training. Not sure if that is or isn't the case here.
5. Section-metadata should only impact the layout of blocks, not the block content itself. (not sure if this is the case here)

---

@salonijain3 @aishwaryamathuria Seeing what I'm seeing, it doesn't look to far off in the word doc. I like the brick / masonry connection... despite knowing we're going to ship `section metadata (advanced)` that will do something similar. The background placement stuff seems pretty heavy... I would expect good defaults there, but the samples don't show that. We shouldn't require:

1. Per-breakpoint images (to @honstar's point)
2. Per-breakpoint image sizing / positioning

For this to work. Can you put some simpler examples together @aishwaryamathuria?

On the naming stuff, we can also "polyfill" on the Milo side inside the block to avoid collisions. If we want authors to say, "masonry" instead of "masonry-layout" (as an example) we can easily polyfill these naming conventions in the block so they map to non-conflicting CSS in Milo.

aishwaryamathuria:
> FWIW, this is probably how I would build masonry. The complexity of **_layout_** should be deferred to the parent section as it is what should be saying how big the items are. Items (bricks) should only care about their internal contents.
> 
> I think when you put layout on blocks, you will run into issues because you change one block, and then the other one below falls out. You end up scrolling up and down your document while authoring to change each block. What is below is a little more stable because it's centrally managed by the parent section.
> 
> I just kinda think, if you want this kind of super flexible control, lets not do half measures and lets lean into it. Give something powerful enough that it crushes all these use cases.
> 
> Screenshot 2023-11-29 at 12 52 48 PM

@auniverseaway @ryanmparrish @rgclayton 
We will go ahead and syncup with the XD to review the block design to support the mentioned requirements and media breakpoints for supporting the suggested authoring.

Once we have the final requirements we will work on the section metadata - advanced variant for arranging the blocks inside it with the mentioned style.
We will check in the gap align and justify fields as currently they are not authorable and if they are in the scope. 
As for the row field, if we support mobile tablet and desktop we will have some fields with null values as number of rows in mobile will be larger (if it goes for one column) and in desktop it will be smaller if it goes for more that 2 blocks in one row.

**_If the scope requires only one column layout in mobile and 2 column layout in tablet then probably we can have it fixed for these screens and have row styling requested for desktop? thoughts_**

CC: @salonijain3 @dstrong23

rgclayton:
> 1. Creating a new js section-metadata-advanced.js to contain the code for advanced variant that would handle this grid layout.

After talking with Chris about this one the answer is no. No new `section-metadata-advanced.js`. Keep code in `section-metadata.js` to more easily leverage other functionality already in place in this file. 

cc: @aishwaryamathuria @salonijain3 @auniverseaway

auniverseaway:
> Creating a new js section-metadata-advanced.js to contain the code for advanced variant that would handle this grid layout.

FWIW, section-metadata.js is 52 lines. I think it's overkill to add another file (request headers, etc.) to add this.

auniverseaway:
From my email on this topic...

> Hopefully, I understand the details correctly…
>
> When I said, “… this is probably how I would build masonry.” about advanced section metadata, it was really a guide for exploration at a later date. Given the research needed, I would not tightly couple these ideas.
>
> If we think we can ship an architecturally elegant solution with simple authoring patterns in the [URGENT] timeline, great… but I have my doubts. I think doing section-metadata (advanced) correctly will take more research than we have time for.
>
> This means we are likely shipping a non-ideal solution in brick / masonry now, but I think that’s the lesser of evils. I don’t want poor authoring patterns in the wild more than we have to and I think my advanced section metadata idea needs to bake more. And yes, this comes at the expense of having overlap between what brick / masonry does today and what section-metadata will do tomorrow.

In short, I don't think you should try to ship advanced section-metadata and I think it requires more research than the teams (both Brahmos and Milo) have time for.

auniverseaway:
> In that case should we go with the masonry layout with the current authoring?

I think putting the layout widths on the items is problematic and causes a lot of extra authoring and confusion. 

I would maybe use a custom row in section-metadata that masonry owns and you can play in that space for your layouts.

When you put layout on the individual blocks, it's creates problems in syncing content...

You're scrolling up and down the page to see what the siblings are. "Did I put 1/3 on the block above or 2/3?"

Here's my ideal:

1. Ship a custom key for masonry with your layout options (feel free to steal ideas from advanced section).
2. If 1 cannot be done in a reasonable time, ship what you have above.

rgclayton:
> > > In that case should we go with the masonry layout with the current authoring?
> > 
> > 
> > I think putting the layout widths on the items is problematic and causes a lot of extra authoring and confusion.
> > I would maybe use a custom row in section-metadata that masonry owns and you can play in that space for your layouts.
> > When you put layout on the individual blocks, it's creates problems in syncing content...
> > You're scrolling up and down the page to see what the siblings are. "Did I put 1/3 on the block above or 2/3?"
> > Here's my ideal:
> > 
> > 1. Ship a custom key for masonry with your layout options (feel free to steal ideas from advanced section).
> > 2. If 1 cannot be done in a reasonable time, ship what you have above.
> 
> Thanks @auniverseaway We can work on point 1 by pulling inspiration from the advanced section. Have a row field for masonry which will be processed and applied only in case the layout is set to masonry. Rows will be processed and used for brick size evaluation only is layout is set to masonry.
> 
> Screenshot 2023-12-05 at 12 12 08 AM
> CC: @salonijain3 @dstrong23 @ccongerp @rgclayton @ryanmparrish


I think this is what Chris meant by a "custom row" in `section-metadata`

An idea for the width values - the width values in my example are easy to expand upon when needed and to me are easier to understand the layout I would be getting (the max equaling 12 columns per row). The order of the widths would correspond to the order the bricks are in the section.

This layout would create a 1/3 - 2/3 layout for the the first two bricks. 
A half width row for the second brick
A full width row for the  last brick
image

vgoodric:
2 tiny notes:
1. in the homepage version, the two-third-grid variant is called two-thirds-grid. That personally sounds more natural to me
2. I'm curious what the purpose of the new brick block is.  Is it basically just the text block or does it have special functionality?  The section CSS doesn't seem to need the children to be "bricks" to control the width, so just curious.

aishwaryamathuria:
> > > > In that case should we go with the masonry layout with the current authoring?
> > > 
> > > 
> > > I think putting the layout widths on the items is problematic and causes a lot of extra authoring and confusion.
> > > I would maybe use a custom row in section-metadata that masonry owns and you can play in that space for your layouts.
> > > When you put layout on the individual blocks, it's creates problems in syncing content...
> > > You're scrolling up and down the page to see what the siblings are. "Did I put 1/3 on the block above or 2/3?"
> > > Here's my ideal:
> > > 
> > > 1. Ship a custom key for masonry with your layout options (feel free to steal ideas from advanced section).
> > > 2. If 1 cannot be done in a reasonable time, ship what you have above.
> > 
> > 
> > Thanks @auniverseaway We can work on point 1 by pulling inspiration from the advanced section. Have a row field for masonry which will be processed and applied only in case the layout is set to masonry. Rows will be processed and used for brick size evaluation only is layout is set to masonry.
> > Screenshot 2023-12-05 at 12 12 08 AM
> > CC: @salonijain3 @dstrong23 @ccongerp @rgclayton @ryanmparrish
> 
> I think this is what Chris meant by a "custom row" in `section-metadata`
> 
> An idea for the width values - the width values in my example are easy to expand upon when needed and to me are easier to understand the layout I would be getting (the max equaling 12 columns per row). The order of the widths would correspond to the order the bricks are in the section.
> 
> This layout would create a 1/3 - 2/3 layout for the the first two bricks. A half width row for the second brick A full width row for the last brick image

@rgclayton We can use the proposed authoring for arranging the bricks in place of a multiple row format. This authoring looks different from @auniverseaway suggestion of row wise arrangement. I believe that this change have already been reviewed by you and Chris, We can implement this in place of using the layout+rows both.
I will make the required changes for this authoring in the current PR.

CC: @salonijain3 @dstrong23 @ccongerp

aishwaryamathuria:
@auniverseaway @rgclayton @ryanmparrish 
We have made the suggested change as mentioned in https://github.com/adobecom/milo/pull/1515#issuecomment-1840064577 in the PR. Section metadata would now directly control the size and layout and it would not need to be specified at the brick level. 
Author can add new key 'masonry' and specify the div spans (ranging from 1 to 12) as required. The sizes can be added in multiple or single line as required.
Could you re-review the PR ?

Screenshot 2023-12-05 at 4 35 45 PM

Screenshot 2023-12-05 at 4 46 37 PM

CC: @salonijain3 @dstrong23 @ccongerp

rgclayton:
> 2\. I'm curious what the purpose of the new brick block is.  Is it basically just the text block or does it have special functionality?  The section CSS doesn't seem to need the children to be "bricks" to control the width, so just curious.

From my screenshot? If so, that was just a simple example of blocks. No new brick block. My main point revolved around the `section metatdata`

cc: @vivgoodrich

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -27,6 +27,17 @@ export async function handleStyle(text, section) {
   section.classList.add(...styles);
 }
 
+function handleMasonry(text, section) {
+  section.classList.add(...['masonry-layout', 'course-up']);
```
rgclayton, line 31: What is `course-up`?

libs/blocks/section-metadata/section-metadata.css
```diff
@@ -269,4 +291,23 @@ main > .section[class*='-up'] > .content {
...
+  .section.masonry-layout .grid-span-3 {grid-column: span 3; }
+  .section.masonry-layout .grid-span-4 {grid-column: span 4; }
+  .section.masonry-layout .grid-span-5 {grid-column: span 5; }
+  .section.masonry-layout .grid-span-6 {grid-column: span 6; }
+  .section.masonry-layout .grid-span-7 {grid-column: span 7; }
+  .section.masonry-layout .grid-span-8 {grid-column: span 8; }
+  .section.masonry-layout .grid-span-9 {grid-column: span 9; }
+  .section.masonry-layout .grid-span-10 {grid-column: span 10; }
+  .section.masonry-layout .grid-span-11 {grid-column: span 11; }
+  .section.masonry-layout .grid-span-12 {grid-column: span 12; }
```
rgclayton, line 312: Do we need both `full-width` and `span-12`?

rgclayton (commented):
Adding a few comments.

libs/blocks/section-metadata/section-metadata.css
aishwaryamathuria: @rgclayton Removed the extra span 12 from the PR

libs/blocks/section-metadata/section-metadata.js
aishwaryamathuria: @rgclayton The class is not adding any new CSS. However it allows us to use the customizable gap customization option already available in the section metadata for masonry as well - `.section[class*='-up'].<size>-gap`

vgoodric:
> > 2. I'm curious what the purpose of the new brick block is.  Is it basically just the text block or does it have special functionality?  The section CSS doesn't seem to need the children to be "bricks" to control the width, so just curious.
> 
> From my screenshot? If so, that was just a simple example of blocks. No new brick block. My main point revolved around the `section metatdata`
> 
> cc: @vivgoodrich

When I look at files changed there is a js and CSS file for blocks/brick
image

aishwaryamathuria:
> > > 2. I'm curious what the purpose of the new brick block is.  Is it basically just the text block or does it have special functionality?  The section CSS doesn't seem to need the children to be "bricks" to control the width, so just curious.
> > 
> > 
> > From my screenshot? If so, that was just a simple example of blocks. No new brick block. My main point revolved around the `section metatdata`
> > cc: @vivgoodrich
> 
> When I look at files changed there is a js and CSS file for blocks/brick image



> > > 2. I'm curious what the purpose of the new brick block is.  Is it basically just the text block or does it have special functionality?  The section CSS doesn't seem to need the children to be "bricks" to control the width, so just curious.
> > 
> > 
> > From my screenshot? If so, that was just a simple example of blocks. No new brick block. My main point revolved around the `section metatdata`
> > cc: @vivgoodrich
> 
> When I look at files changed there is a js and CSS file for blocks/brick image

@vgoodric The brick block is not similar to text block. It is a new block that has custom requirements (requirements listed in [MWPW-137717](https://jira.corp.adobe.com/browse/MWPW-137717)) . And it can be arranged via masonry layout in multiple row and column format.

libs/blocks/section-metadata/section-metadata.css
aishwaryamathuria: @rgclayton Adding the update to the comment - masonry-layout is a polyfill and the authoring does not require the authors to add this as a style. As per the latest authoring format this would only need new masonry key row in the section-metadata with layout info.

libs/blocks/section-metadata/section-metadata.js
aishwaryamathuria: @rgclayton I have update the classname to make it more intuitive.

vgoodric:
So I feel bad because it looks like the work to move sizing information off of the bricks and into the section metadata has already been completed.  But I checked with Mimi and Christine and we don't want to use this system on homepage.  To be fair, if you had content where bricks were designed to be interchangeably used at multiple sizes, I see the advantage to this new system.  But all of our blocks only look right at the specific size they were built for.  So being able to look at a brick's source and know what size it's supposed to be to look right is really helpful.  And then everywhere we put the brick, it's automatically the right size.

As the 2 that have spent the most time building and authoring content in a masonry layout, I put a lot of stock in Christine and Mimi's opinion.  And we feel strongly that we don't want to lose the option to just put the size right on the brick (and have a default size so most bricks don't need it). But maybe we should open it up to some author feedback?  Maybe we want to support both?

@auniverseaway @aishwaryamathuria

rgclayton:
> So I feel bad because it looks like the work to move sizing information off of the bricks and into the section metadata has already been completed. But I checked with Mimi and Christine and we don't want to use this system on homepage. To be fair, if you had content where bricks were designed to be interchangeably used at multiple sizes, I see the advantage to this new system. But all of our blocks only look right at the specific size they were built for. So being able to look at a brick's source and know what size it's supposed to be to look right is really helpful. And then everywhere we put the brick, it's automatically the right size.
> 
> As the 2 that have spent the most time building and authoring content in a masonry layout, I put a lot of stock in Christine and Mimi's opinion. And we feel strongly that we don't want to lose the option to just put the size right on the brick (and have a default size so most bricks don't need it). But maybe we should open it up to some author feedback? Maybe we want to support both?
> 
> @auniverseaway @aishwaryamathuria

My 2 cents.

@vivgoodrich Can the homepage not continue to operate with it's current set up once this is merged into Milo?

It makes sense to have the homepage tailored to a more specific set up since it's the homepage. It being added to the Milo repo, with a broader consumer base, does that specificity still make sense? Or does it make sense to extend it for more use cases?

vgoodric:
> My 2 cents.
> 
> @vivgoodrich Can the homepage not continue to operate with it's current set up once this is merged into Milo?
> 
> It makes sense to have the homepage tailored to a more specific set up since it's the homepage. It being added to the Milo repo, with a broader consumer base, does that specificity still make sense? Or does it make sense to extend it for more use cases?

Absolutely.  If you guys decide not to support sizing on the bricks we will just not switch over and continue to do our own thing.  But I don't think the viewpoints should be discounted as homepage only.  I suspect that most blocks in a masonry design will be designed for a specific size and will not look good in others.  I tried to view the figma doc in Jira but can't get it to work so maybe I just don't know how it will be used on the rest of the site.  But in general if a brick is using a background image, it will likely look awful at the wrong size.

Snehayt:
Verified the block-Basic brick and Masonry .Works as expected.

In Milo:
https://masonry--milo--aishwaryamathuria.hlx.page/drafts/sneha/poc1/poc-11
https://masonry--milo--aishwaryamathuria.hlx.page/drafts/sneha/poc1/poc-4
https://masonry--milo--aishwaryamathuria.hlx.page/drafts/sneha/poc1/poc-2
https://masonry--milo--aishwaryamathuria.hlx.page/drafts/sneha/poc1/poc-1

In CC:
https://main--cc--adobecom.hlx.page/drafts/Sneha/syt/poc1/poc-2?milolibs=masonry--milo--aishwaryamathuria
https://main--cc--adobecom.hlx.page/drafts/Sneha/syt/poc1/poc-4?milolibs=masonry--milo--aishwaryamathuria
https://main--cc--adobecom.hlx.page/drafts/Sneha/syt/poc1/poc-11?milolibs=masonry--milo--aishwaryamathuria

rgclayton (approved):
I feel like we have a pretty good authoring pattern now with this.

ryanmparrish (approved):
Thanks for all the re-working on this. 
I think it's looking good. Approved, but can you maybe take a quick look at the stylelint on your brick.css. I'm seeing a few linting issues. (see suggestion below)
My hope is we can expand on this in the future to possibly offer the best of both worlds w/ how HP is being authored. 

`.brick {
  position: relative;
  display: flex;
  text-size-adjust: none;
  min-height: 300px;
}

.brick,
.brick.click a.foreground {
  color: inherit;
}

.brick.click > a {
  text-decoration: none;
}

.brick.light,
.brick.light.click a.foreground {
  color: var(--text-color);
}

.brick.dark,
.dark.brick.click a.foreground {
  color: var(--color-white);
}

.brick .icon-stack-area li,
.brick .icon-stack-area li a {
  display: flex;
  align-items: center;
  gap: var(--spacing-xs);
  text-align: left;
}

.brick .foreground a:not([class]),
.brick .foreground span.first-link {
  font-weight: 700;
}

.brick .background {
  position: absolute;
  bottom: 0;
  left: 0;
  right: 0;
  top: 0;
  overflow: hidden;
}

.brick .foreground {
  position: relative; 
  display: flex;
  flex-grow: 1;
  padding: var(--spacing-m);
}

.brick.rounded-corners,
.brick.rounded-corners .background,
.brick.rounded-corners .foreground {
  border-radius: var(--spacing-xs);
}

.brick.align-center .foreground,
.brick.align-center .foreground .action-area,
.brick.center .foreground,
.brick.center .foreground .action-area {
  align-items: center;
  text-align: center;
  justify-content: center;
}

.brick.center .foreground {
  align-items: flex-start;
}

.brick .background div,
.brick .background p,
.brick .background picture {
  height: 100%;
  margin: 0;
  padding: 0;
}

.brick .background p,
.brick .background picture {
  display: block;
}

.brick .background img {
  object-fit: contain;
  object-position: bottom center;
  width: 100%;
  height: 100%;
}

.brick .mobile-only,
.brick .tablet-only,
.brick .desktop-only {
  display: none;
}

.brick .icon-stack-area li picture {
  display: flex;
  margin: 0;
  padding: 0;
  flex-shrink: 0;
}

.brick .foreground .icon-area picture {
  display: flex;
}

.brick .foreground p {
  padding: 0;
  margin: 0;
}

.brick .foreground div > * {
  margin-top: var(--spacing-xs);
}

.brick .foreground p:first-child,
.brick .foreground p.icon-area,
.brick .foreground p.icon-area + p {
  margin-top: 0;
}

.brick .foreground p.icon-area {
  display: inline-block;
}

.brick .foreground p.action-area {
  display: flex;
  flex-wrap: wrap;
  gap: 24px;
  margin-top: var(--spacing-s);
}

.brick .icon-stack-area li img {
  width: var(--icon-size-s);
  height: auto;
}

.brick .foreground .icon-area img {
  height: var(--icon-size-l);
  width: auto;
}

.brick .icon-stack-area {
  display: flex;
  flex-flow: row wrap;
  flex-direction: column;
  gap: var(--spacing-xs);
  width: 100%;
  margin: 0;
  padding: 0;
  list-style-type: none;
}

.brick.center .icon-stack-area,
.brick.align-center .icon-stack-area {
  display: inline-flex;
  width: auto;
}

[dir="rtl"] .brick .icon-stack-area li,
[dir="rtl"] .brick .icon-stack-area li a {
  text-align: right;
}

.brick.click a.foreground .first-link:not([class*="button"]) {
  color: var(--link-color);
  text-decoration: none;
}

.brick.click:hover a.foreground .first-link:not([class*="button"]) {
  text-decoration: underline;
  color: var(--link-hover-color);
}

.static-links .brick.click a.foreground .first-link:not([class*="button"]),
.static-links .brick.click a.foreground a:not([class*="button"]),
.brick.static-links.click a.foreground .first-link:not([class*="button"]),
.brick.static-links.click a.foreground a:not([class*="button"]) {
  color: inherit;
  text-decoration: underline;
}

.brick.click:hover .first-link.con-button.blue,
.brick.click:active .first-link.con-button.blue {
  background: var(--color-accent-hover);
  border-color: var(--color-accent-hover);
  color: var(--color-white);
}

.brick.click:hover .first-link.con-button,
.brick.click:active .first-link.con-button,
.brick.light.click:hover .first-link.con-button,
.brick.light.click:active .first-link.con-button,
.light .brick.click:hover .first-link.con-button,
.light .brick.click:active .first-link.con-button {
  background-color: var(--color-black);
  border-color: var(--color-black);
  color: var(--color-white);
}

.dark .brick.click:hover .first-link.con-button,
.brick.dark.click:active .first-link.con-button {
  background-color: var(--color-white);
  color: var(--color-black);
  text-decoration: none;
}


@media screen and (max-width: 600px) {
  .brick .mobile-only {
    display: block;
  }
}

@media screen and (min-width: 600px) {
  .brick {
    min-height: 384px;
  }

  .brick .foreground {
    padding: var(--spacing-l);
  }
}

@media screen and (min-width: 600px) and (max-width: 1199px) {
  .brick .tablet-only {
    display: block;
  }
}

@media screen and (min-width: 1200px) {
  .brick .desktop-only {
    display: block;
  }

  .brick.large {
    min-height: 500px;
  }
}
`

aishwaryamathuria:
> he future to possibly offer the best of both worlds

@ryanmparrish I have made the suggested changes

PR #1531, opened 2023-11-14
Mwpw-138750 analytics update 2.1

vgoodric:
The Analyst team has requested changes to the daa-ll and daa-lh values to accommodate their reporting needs. 

NOTE: nala tests will fail because I am updating values nala tests check.  I am coordinating with Santoshkumar Sharanappa Nateekar to merge this PR at the same time he pushes his updates.

OLD:
link name(20)-#|Last header before link(20)|s#|b#|blockname|experience name or default--experience name or default2|manifest file name--manifest file2 name

NEW:
link name(20)-#--Last header before link(20)|s#|b#|blockname(15)|pzn experience name or "nopzn"(15)|pzn manifest filename or "nopzn"(20)

- Number in () is character limit.
- in the pzn experience name, don't pass "target-" 
- Add tab to MEP Manifest to indicate PZN vs Test/Promo and an override name
- If tab indicates PZN, then pass manifest name/experience name in Analytics string.
- If tab indicates Test/Promo, then pass "nopzn|nopzn" (We need to use "nopzn" instead of "default" because default means something very specific in Analytics)
- If an override name is provided to a PZN manifest, use that name in analytics
- Clean symbols out of analytic strings
- Stop replacing spaces with underscores in the GNAV analytics value
- Updated preview button feature to include the analytics info
- consolidated manifest-utils into personalization to reduce importing new files
- fixed a couple preview button bugs found when QA the analytics
- change marketingtech.adobe.launch.controlPageLoad to true in martech.js L122.  Request came from Ben Bytheway because if we are explicitly firing the _satellite.track('pageload')  call (line 133). This instructs marketingtech to not fire the default _satellite.track('pageload')  call within their code.  So right when it's set to false it duplicates.

Resolves: [MWPW-138750](https://jira.corp.adobe.com/browse/MWPW-138750)
Also resolves: [MWPW-137512](https://jira.corp.adobe.com/browse/MWPW-137512)

Additional QA links with Target on:
Old: https://main--milo--adobecom.hlx.page/drafts/vgoodric/qa/analytics/both?test=on&pzn2=on&clickflag=true
New: https://mwpw-138750-analytics2--milo--adobecom.hlx.page/drafts/vgoodric/qa/analytics/both?test=on&pzn2=on&clickflag=true

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/vgoodric/qa/analytics/none?clickflag=true&martech=off
- After: https://mwpw-138750-analytics2--milo--adobecom.hlx.page/drafts/vgoodric/qa/analytics/none?clickflag=true&martech=off


libs/martech/analytics.js
```diff
@@ -1,9 +1,16 @@
-export function processTrackingLabels(text, charLimit = 20) {
-  return text?.trim().replace(/\s+/g, ' ').split('|').join(' ')
+export function processTrackingLabels(text, config, charLimit = 20) {
+  if (!config) {
```
overmyheadandbody, line 2: What exactly is the purpose of the `config` argument and its re-assignment, since it doesn't seem to be used to generate the `analyticsValue`?

vgoodric: That's in preparation for the feature to preserve the analytics in English during localization.  I've already started work on it and I know I'm going to need the config in this function.

auniverseaway:
@vivgoodrich this flew under my radar, but the file `analytics.js` should be renamed. Browsers like brave and ad blockers like AdBlock will block files with these names. We originally called `martech.js` `analytics.js` and had to change for this very reason.

@chrischrischris

auniverseaway:
@vivgoodrich do you want to point this to stage so you can test on stage.adobe.com? Seems like it may be a good idea.

@Blainegunn

vgoodric:
> @vivgoodrich this flew under my radar, but the file `analytics.js` should be renamed. Browsers like brave and ad blockers like AdBlock will block files with these names. We originally called `martech.js` `analytics.js` and had to change for this very reason.
> 
> @chrischrischris

Interesting... easy enough to fix.  Should I put the functions into martech?  Or just rename the file? Maybe martech-utils or aa-utils?

vgoodric:
> @vivgoodrich do you want to point this to stage so you can test on stage.adobe.com? Seems like it may be a good idea.
> 
> @Blainegunn

Maybe that would have been helpful on the first big overhaul, but this is not nearly as big and I've QA'd a bunch.  Including on https://main--homepage--adobecom.hlx.page/homepage/index-loggedout?milolibs=mwpw-138750-analytics2

But if you want me to, I can.

libs/features/personalization/personalization.js
```diff
@@ -11,6 +11,47 @@ const ENT_CACHE_EXPIRE = 1000 * 60 * 60 * 3; // 3 hours
...
 const PAGE_URL = new URL(window.location.href);
 
+export const appendJsonExt = (path) => (path.endsWith('.json') ? path : `${path}.json`);
+
+export const normalizePath = (p) => {
+  let path = p;
+
+  if (!path.includes('/')) {
+    return path;
+  }
```
overmyheadandbody, line 21: You could save a couple of bits and just do `if (!path.includes('/')) return path;`

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -51,7 +52,7 @@ export const getFedsPlaceholderConfig = () => {
 export function getAnalyticsValue(str, index) {
   if (typeof str !== 'string' || !str.length) return str;
 
-  let analyticsValue = str.trim().replace(/[^\w]+/g, '_').replace(/^_+|_+$/g, '');
+  let analyticsValue = processTrackingLabels(str, false, 30);
```
overmyheadandbody, line 55: Just as an FYI, underscores will still be used for the AEM-based Gnav tracking that is still available on a lot of our pages. I'm hoping that analysts are aware that there will be some mismatch between the two platforms

libs/features/personalization/personalization.js
```diff
@@ -515,6 +540,7 @@ export async function runPersonalization(info, config) {
     experiment,
     blocks: selectedVariant.useblockcode,
     fragments: selectedVariant.replacefragment,
+
```
overmyheadandbody, line 543: Potential redundant empty space

overmyheadandbody (approved):
Things seem good from a Gnav perspective, but I do share Chris' concern around module naming, 'analytics', 'tracking' and such terms should ideally be avoided

libs/features/personalization/personalization.js
vgoodric: agreed. updated.

libs/blocks/global-navigation/utilities/utilities.js
vgoodric: We've warned them and no objections were raised.  But still wouldn't be shocked if they object later.

libs/features/personalization/personalization.js
vgoodric: not sure what you mean here.

vgoodric:
Decided to use the attributes.js file that was originally used for these kinds of functions.  I'm going to leave analytics.js in so that I can have this published first, and update the couple of references to analytics.js in other clouds.  Then I'll come back to this and remove the analytics.js file next week.

libs/features/personalization/personalization.js
chrischrischris: I think he means "delete this empty line"

libs/martech/attributes.js
```diff
@@ -1,13 +1,66 @@
...
+  const analyticsValue = text?.replace(/[^\w\s]+/g, ' ').replace(/\s+/g, ' ').replace(/^_+|_+$/g, '').trim()
+    .slice(0, charLimit);
+  return analyticsValue;
+}
 
-// decorateBlockAnalytics & decorateLinkAnalytics are being sunset
+export function decorateDefaultLinkAnalytics(block, config) {
+  if (block.classList.length
+    && !block.className.includes('metadata')
+    && !block.classList.contains('link-block')
```
chrischrischris, line 16: Why `className` && `classList`?  Are you looking for any class containing `metadata`?

libs/martech/attributes.js
```diff
@@ -1,13 +1,66 @@
...
+
+// below functions are being sunset
 export function decorateBlockAnalytics() { return false; }
 export function decorateLinkAnalytics() { return false; }
 
 export const analyticsGetLabel = (txt) => txt.replaceAll('&', 'and')
-  .replace(RE_ALPHANUM, '_')
-  .replace(RE_TRIM_UNDERSCORE, '');
+  .replace(/[^0-9a-z]/gi, '_')
+  .replace(/^_+|_+$/g, '');
```
chrischrischris, line 63: why remove the consts?

libs/features/personalization/personalization.js
vgoodric: line removed

libs/features/personalization/personalization.js
```diff
@@ -572,9 +587,19 @@ export async function applyPers(manifests) {
...
-  const trackingVariants = results.map((r) => r.experiment.selectedVariantName);
-  document.body.dataset.mep = `${trackingVariants.join('--')}|${trackingManifests.join('--')}`;
-
-  decoratePreviewCheck(config, experiments);
+  const personalizedManifests = results.filter((r) => (r.experiment.manifestType !== 'test or promo'));
+  if (!personalizedManifests.length) {
+    document.body.dataset.mep = 'nopzn|nopzn';
+  } else {
+    const trackingVariants = personalizedManifests.map((r) => {
+      const val = r.experiment.selectedVariantName.replace('target-', '').trim().slice(0, 15);
```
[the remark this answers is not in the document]
vgoodric: Yup, just what they decided.  To get the string under a 100 characters, not everything could be 20.  So they have a couple at 15 limit. 
link name(20)-#--Last header before link(20)|s#|b#|blockname(15)|pzn experience name or "nopzn"(15)|pzn manifest filename or "nopzn"(20)

If the section, block and link numbers are single digits, and everything that is so long it has to be truncated, the above string is 103 characters.  So the fear was that if those 2 items were limited 20 characters, we'd have more issues.  I'm actually wondering now if we should make the limit on the manifest name 15 as well.  That puts us at 98, allowing for a couple of double digit numbers.
image

libs/martech/attributes.js
vgoodric: line 15 is className to catch different metadata blocks: section-metadata, table-metadata etc. I think this future proofs it more
line 16 is classList because the exact name of the class I'm checking for is "link-block"

libs/martech/attributes.js
vgoodric: I can put this back.  My logic was that it's only used once and we're deprecating those functions.  So I figured I'd save on characters.

libs/features/personalization/add-preview-to-config.js
```diff
@@ -17,14 +17,23 @@ export default async function addPreviewToConfig({
...
+        const url = new URL(manifest);
+        return url.pathname;
+      } catch (e) {
+        return manifest;
+      }
+    });
+
     config.mep.override.split(',').forEach((manifestPair) => {
-      persManifests.push(manifestPair.trim().toLowerCase().split('--')[0]);
+      const manifestPairManifest = manifestPair.trim().toLowerCase().split('--')[0];
```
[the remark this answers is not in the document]
vgoodric: changed to manifestTitle

libs/features/personalization/personalization.js
```diff
@@ -572,9 +587,19 @@ export async function applyPers(manifests) {
     expBlocks: consolidateObjects(results, 'blocks'),
     expFragments: consolidateObjects(results, 'fragments'),
   });
-  const trackingManifests = results.map((r) => r.experiment.manifest.split('/').pop().replace('.json', ''));
-  const trackingVariants = results.map((r) => r.experiment.selectedVariantName);
-  document.body.dataset.mep = `${trackingVariants.join('--')}|${trackingManifests.join('--')}`;
-
-  decoratePreviewCheck(config, experiments);
+  const personalizedManifests = results.filter((r) => (r.experiment.manifestType !== 'test or promo'));
```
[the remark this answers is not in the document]
vgoodric: done

libs/features/personalization/personalization.js
```diff
@@ -572,9 +587,19 @@ export async function applyPers(manifests) {
...
+  if (!personalizedManifests.length) {
+    document.body.dataset.mep = 'nopzn|nopzn';
```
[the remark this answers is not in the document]
vgoodric: done

libs/features/personalization/personalization.js
```diff
@@ -572,9 +587,19 @@ export async function applyPers(manifests) {
...
+  } else {
+    const trackingVariants = personalizedManifests.map((r) => {
+      const val = r.experiment.selectedVariantName.replace('target-', '').trim().slice(0, 15);
+      return val === 'default' ? 'nopzn' : val;
+    });
+    const trackingManifests = personalizedManifests.map((r) => {
+      const val = r.experiment?.manifestOverrideName || r.experiment?.manifest;
+      return val.split('/').pop().replace('.json', '').trim()
+        .slice(0, 20);
```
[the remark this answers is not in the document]
vgoodric: based on the math from the above comment, making this one 15

chrischrischris (approved):

chrischrischris:
Removing `run-nala` label as there are updates in nala to accomodate the changes in this script that will be merged at the same time as this PR.

3ch023 (approved):

PR #1542, opened 2023-11-15
MWPW-137938 - Inline video image fallback

rgclayton:
Adds the ability to include a fallback image for inline .mp4 videos. The image displays while the video loads, or if the video fails to load.

An author defined image is added via the `poster` attribute on a video. The image/poster displays while the video loads. Or, if the path to the video is incorrect.

Resolves: [MWPW-137938](https://jira.corp.adobe.com/browse/MWPW-137938)

**Test URLs: Sample page using a marquee fragment with video fallback image**
- After: https://video-fallback--milo--adobecom.hlx.page/drafts/rclayton/test-template?martech=off

Sample page with several block variations for regression testing:
- Before: https://main--milo--adobecom.hlx.page/drafts/rclayton/video/video-fallback?martech=off
- After: https://video-fallback--milo--adobecom.hlx.page/drafts/rclayton/video/video-fallback?martech=off

**Test URLs: BACOM **
-Before: https://main--bacom--adobecom.hlx.page/drafts/rclayton/test/rtcdp?martech=off
-After: https://main--bacom--adobecom.hlx.page/drafts/rclayton/test/rtcdp?milolibs=video-fallback&martech=off

**Testing notes:**
Videos with fallback image will have a poster attribute on the video tag.

Bacom page example is under the first tab "Unified Profiles". Video in media block with fallback image/poster.

**Authoring via image alt text**
Use image and alt text property to author a fallback image to inline video's.
image


echampio-at-adobe (approved):

rgclayton:
Will look into the unit test for utils that is erroring.

rgclayton:
Failing unit test appears to be coming from caas-config, which is outside of this PR.

libs/utils/decorate.js
```diff
@@ -168,20 +168,21 @@ export function decorateTextOverrides(el, options = ['-heading', '-body', '-deta
...
   if (playOnHover && isAutoplayOnce) {
-    return 'playsinline autoplay muted data-hoverplay';
+    return `playsinline autoplay muted data-hoverplay ${setPoster}`;
   }
   if (isAutoplayOnce) {
-    return 'playsinline autoplay muted';
+    return `playsinline autoplay muted ${setPoster}`;
   }
-  return 'playsinline controls';
+  return `playsinline controls ${setPoster}`;
```
chrischrischris, line 185: nit: rename `setPoster` to `poster`

Also I wonder if we should set a "global attribute" to use - something like:
```
 const poster = dataset?.videoPoster ? `poster='${dataset.videoPoster}'` : '';
 const globalAttrs = `playsinline ${poster}`;
  if (isAutoplay && !isAutoplayOnce) {
    return `${globalAttrs} autoplay loop muted`;
  }
  if (playOnHover && isAutoplayOnce) {
    return `${globalAttrs} autoplay muted data-hoverplay`;
  }
  ....
```

chrischrischris (approved):

libs/utils/decorate.js
rgclayton: Updated. Thanks Chris.

ryanmparrish (approved):

PR #1543, opened 2023-11-15
MWPW-137835 - Tabs new 'pill' variant with sizes

Sartxi:
As an author in Milo, I want the ability to add a **Pill** variant to the Tabs block. 

* Add new pill variant to Tabs block that match spectrum 2 examples
* Add pill size variations `s pill or small pill` and `l pill or large pill` (medium is the default)

Resolves: [MWPW-137835](https://jira.corp.adobe.com/browse/MWPW-137835)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/sarchibeque/pill-tabs?martech=off
- After: https://sarchibeque-pill-tabs--milo--adobecom.hlx.page/drafts/sarchibeque/pill-tabs?martech=off


ryanmparrish (approved):
Good looks 👍

fullcolorcoder (approved):

JackySun9 (approved):

elan-tbx (approved):
Lookin' good. Small nit from me -- the Figma designs include a small gap between the :focus outline and the pill, which appears to be missing on the demo page.
Screenshot 2023-11-21 at 10 38 46 AM
Screenshot 2023-11-21 at 10 38 55 AM

Other than that, everything looks good

libs/blocks/tabs/tabs.css
```diff
@@ -172,7 +180,35 @@
   margin-inline-start: 16px;
 }
 
-.tabs.quiet div[role="tablist"] button:first-of-type {
+.tabs[class*='pill'] {
```
narcis-radu, line 183: I'm just curious why are you using this approach instead of `.tabs.pill`.
The following element will match your selectors, it might create some weird side-effects in the future.
```
<div class="tabs pillow"></div>
```

Sartxi: @narcis-radu the reason we are using that selector is because the `pill` variant can have sizes `pill-small` and `pill-large`

PR #1567, opened 2023-11-20
MWPW-138654 - Quiz breakpoint icons and background

colloyd:
* Icon and image columns renamed in strings sheet from image and cover for clarity
* icon-background-color support for strings > questions sheet
* icon-tablet and icon-desktop support added for each option

Resolves: [MWPW-138654](https://jira.corp.adobe.com/browse/MWPW-138654)

**Test URLs:**
- After: https://mwpw-138654-quiz-breakpoint-icons--milo--colloyd.hlx.page/drafts/colloyd/quiz/?martech=off

This makes some core changes to strings.xlsx for quizzes. All instances in strings.xlsx where the columns named "**image**" will need to be changed to "**icon**", all instances of "**cover**" will need to be renamed to "**image**". Support has also been added for "**icon-background color**" in the questions sheet inside of strings.xlsx. This column can take and for of css color, be that rgba, hsla, hex, etc. Hex values are recommended. Support for tablet and desktop overrides for icons has also been added and can be implemented by adding the "**icon-tablet**" and "**icon-desktop**" columns to each additional sheet in strings.xslx

See this sheet that powers the "After" URL for examples of the details outlined above: https://main--milo--adobecom.hlx.page/drafts/colloyd/quiz/strings.json

Mobile views of quiz options has changed. If there is an image with no default icon as a fall back, no image will display. If there is an image with either a tablet or default icon fallback, the icon will be displayed at appropriate breakpoints and if there is a background, it will stretch to the edge of the option as is seen in the current dexter dx quizzes. Images will never be displayed at mobile widths.


JackySun9 (approved):

libs/blocks/quiz/quiz.js
```diff
@@ -258,6 +258,10 @@ const App = ({
     if (fragmentURL) {
       loadFragments(fragmentURL);
     }
+    const iconBg = getStringValue('icon-background-color');
```
fullcolorcoder, line 261: Could this be done after page load?

libs/blocks/quiz/quiz.css
```diff
@@ -62,44 +62,48 @@
 }
 
 .quiz-option {
-  align-items: center;
+  align-items: stretch;
```
fullcolorcoder, line 65: Does this cause layout shift?

fullcolorcoder:
I think it's looking great, although the performance drop concerns me.

libs/blocks/quiz/quizoption.js
```diff
@@ -1,40 +1,62 @@
...
+    if (icon || iconTablet || iconDesktop) className += 'has-icon ';
+    if (image) className += 'has-image ';
+    if (background) className += 'has-background ';
+    if (disabled) className += disabled;
+    if (selected) className += selected;
+    return className;
+  };
+
+  const getIconHtml = (iconUrl, iconClass) => html`<div class="quiz-option-icon ${iconClass}">
+    <img src="${iconUrl}" alt="Quiz Option Icon" />
```
fullcolorcoder, line 17: "Quiz Option Icon" - can you get the alt from the image or do authors have to do the replace thing for loc?

echampio-at-adobe: We should likely set alt text to the card title or description, rather than something hardcoded like this value.

echampio-at-adobe:
FYI: Pagespeed on live with the current quiz data is still 100. 

https://pagespeed.web.dev/analysis/https-mwpw-138654-quiz-breakpoint-icons--milo--colloyd-hlx-live-drafts-quiz-quiz-2/vcrm1h3sta?form_factor=mobile

libs/blocks/quiz/quizoption.js
colloyd: Good call on these. I hadn't touched alt tags at all in all of this,  but we should probably at least add some checks for title, or description before defaulting to the current generic text.

colloyd:
> FYI: Pagespeed on live with the current quiz data is still 100.
> 
> https://pagespeed.web.dev/analysis/https-mwpw-138654-quiz-breakpoint-icons--milo--colloyd-hlx-live-drafts-quiz-quiz-2/vcrm1h3sta?form_factor=mobile

@echampio-at-adobe So how much do I need to agonize over this then, if it's at 100 on live?

libs/blocks/quiz/quiz.css
colloyd: @fullcolorcoder So I've been testing a bunch of changes manually using a test branch that was created from this PR's branch. I notice no significant difference at all when I've set this back to center. I think this isn't having any impact. I've tried a couple of other things, like removing the background-color code altogether - no difference at all. I have noticed that when I get lower scores, I'm getting TBT from typekit, though.

libs/blocks/quiz/quiz.js
colloyd: I've placed it here so authors have the option of different backgrounds per question - not sure if there's a use-case for that or not.

libs/blocks/quiz/quizoption.js
colloyd: Set this to use text or title before resorting to the generic default

libs/blocks/quiz/quizoption.js
```diff
@@ -1,40 +1,62 @@
...
+
+  const getIconHtml = (iconUrl, iconClass) => html`<div class="quiz-option-icon ${iconClass}">
+    <img src="${iconUrl}" alt="${text || title || 'Quiz Option Icon'}" />
   </div>`;
 
-  const coverImageHtml = html`
+  const imageHtml = html`
     <div class="quiz-option-image" 
-      style="background-image: url('${coverImage}')">
+      style="background-image: url('${image}')">
```
fullcolorcoder, line 22: I'm wondering if it would be more performant and consistent with other Milo features to use srcset. 

image

libs/blocks/quiz/quiz.js
```diff
@@ -258,6 +258,10 @@ const App = ({
...
+    if (iconBg) {
+      document.documentElement.style.setProperty('--quiz-icon-bg', iconBg);
```
fullcolorcoder, line 263: This could be causing re-rendering in the browser, would it make sense to scope it lower to the component instead of as an inline style on the HTML el?

fullcolorcoder (changes requested):
I'm still having an issue with the performance drop. 

Here I've created a quiz that has 1 option on the 1st question, and no values for icon/tablet/dektop etc. 

https://mwpw-138654-quiz-breakpoint-icons--milo--colloyd.hlx.page/drafts/bradjohn/quiz/?martech=off

The results are consistently in the 74 area, and I'm still seeing about that locally on your test quiz and in pagespeed.

image

I added in a few spots to explore. I vote we figure it out, but understand if we need to run my request changes over and iterate.

libs/blocks/quiz/quiz.js
echampio-at-adobe: The srcset idea does seem like a good one. As to the backgrounds, each button can be identified by its data-option-name attribute, so perhaps logic similar to what would write the picture srcset could also write a style block for the backgrounds.

colloyd: Changed scope of background and added srcset

libs/blocks/quiz/quizoption.js
colloyd: changed to using picture, media and srcset for icons

PR #1573, opened 2023-11-21
MWPW-136867: Merch-Card Offer Selection

3ch023:
Add Offer Selection to the Merch Card block.
User can switch between Offers radio buttons and trigger description, badge CTA and Price updates.

Resolves: [MWPW-136867](https://jira.corp.adobe.com/browse/MWPW-136867)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.live/drafts/mariia/PR/acrobatcard?martech=off
- After: https://offerselection--milo--adobecom.hlx.live/drafts/mariia/PR/acrobatcard?martech=off

### DIFF
--- a/libs/blocks/merch-card/merch-card.js
+++ b/libs/blocks/merch-card/merch-card.js
@@ -282,13 +282,13 @@ const init = async (el) => {
       ),
     );
   }
+  const footer = createTag('div', { slot: 'footer' });
   const ctas = el.querySelector('p > strong a, p > em a')?.closest('p');
   if (ctas) {
-    const footer = createTag('div', { slot: 'footer' });
     decorateButtons(ctas);
     footer.append(ctas);
-    merchCard.appendChild(footer);
   }
+  merchCard.appendChild(footer);
   if (image !== undefined) {
     const imageSlot = createTag('div', { slot: 'bg-image' });
     imageSlot.appendChild(image);
@@ -327,6 +327,11 @@ const init = async (el) => {
   merchCard.setAttribute('filters', categories.join(','));
   merchCard.setAttribute('types', types.join(','));
   parseContent(el, merchCard);
+  const offerSelection = cardType === 'plans' ? el.querySelector('ul') : null;
+  if (offerSelection) {
+    const { initOfferSelection } = await import('./offer-selection.js');
+    initOfferSelection(merchCard, offerSelection);
+  }
   decorateBlockHrs(merchCard);
   if (merchCard.classList.contains('has-divider')) {
     merchCard.setAttribute('custom-hr', true);
--- a/libs/blocks/merch-card/offer-selection.js
+++ b/libs/blocks/merch-card/offer-selection.js
@@ -1,0 +1,49 @@
+import { createTag } from '../../utils/utils.js';
+import { decorateButtons } from '../../utils/decorate.js';
+import '../../deps/merch-offers.js';
+
+function createDynamicSlots(el, bodySlot) {
+  const price = createTag('h5', { class: 'merch-card-price' });
+  price.append(createTag('span', { slot: 'price', is: 'inline-price' }));
+  bodySlot.append(price);
+
+  const p = createTag('p', { class: 'action-area' });
+  p.append(createTag('a', { slot: 'cta', is: 'checkout-link' }));
+  const footer = el.querySelector('div[slot="footer"]');
+  footer.append(p);
+  bodySlot.querySelector('p')?.setAttribute('slot', 'description');
+}
+function createMerchOffer(option) {
+  const merchOffer = createTag('merch-offer', { text: option.childNodes[0].textContent.trim() });
+  [...option.querySelector('ul').children].forEach((li, index) => {
+    const override = li.childNodes[0];
+    if (override.nodeName === '#text') {
+      if (index === 0) {
+        merchOffer.setAttribute('badge-text', override.textContent.trim());
+      } else {
+        const desc = createTag('p', { slot: 'description' });
+        desc.textContent = override.textContent.trim();
+        merchOffer.append(desc);
+      }
+    } else {
+      merchOffer.append(override);
+    }
+  });
+  decorateButtons(merchOffer);
+  return merchOffer;
+}
+
+export const initOfferSelection = (merchCard, offerSelection) => {
+  const bodySlot = merchCard.querySelector('div[slot="body-xs"]');
+  if (!bodySlot) return;
+  createDynamicSlots(merchCard, bodySlot);
+  const merchOffers = createTag('merch-offers', { container: 'merch-card' });
+  [...offerSelection.children].forEach((option) => {
+    merchOffers.append(createMerchOffer(option));
+  });
+  merchOffers.querySelectorAll('a[is="checkout-link"]').forEach((link) => { link.setAttribute('slot', 'cta'); });
+  merchOffers.querySelectorAll('span[is="inline-price"]').forEach((price) => { price.setAttribute('slot', 'price'); });
+  bodySlot.append(merchOffers);
+};
+
+export default initOfferSelection;
--- a/test/blocks/merch-card/mocks/acrobat-card.html
+++ b/test/blocks/merch-card/mocks/acrobat-card.html
@@ -1,7 +1,7 @@
 <div class="fragment">
   <div class="section">
     <div class="before"></div>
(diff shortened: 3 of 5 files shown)


libs/blocks/merch-card/merch-card.js
```diff
@@ -161,12 +161,13 @@ const init = (el) => {
       merchCard.setAttribute('action-menu', true);
       merchCard.append(createTag('div', { slot: 'action-menu-content' }, actionMenuContent.innerHTML));
     }
-    if (ctas) {
-      const footer = createTag('div', { slot: 'footer' });
+
+    const footer = createTag('div', { slot: 'footer' });
```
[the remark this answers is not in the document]
3ch023: hmm good point, let me check if we have cards without footer

libs/blocks/merch-card/merch-card.js
```diff
@@ -194,6 +195,13 @@ const init = (el) => {
       }
     }
     parseContent(el, altCta, cardType, merchCard);
+    const offerSelection = cardType === 'plans' ? el.querySelector('ul') : null;
+    if (offerSelection) {
+      import('./offer-selection.js')
```
yesil, line 200: shouldn't we await for the import here?

`const {default: {initOfferSelection} } = await import('./offer-selection.js');` or something in line.

test/blocks/merch-card/offer-selection.test.js
```diff
@@ -0,0 +1,70 @@
...
+import { expect } from '@esm-bundle/chai';
+
+const { default: initCard } = await import('../../../libs/blocks/merch-card/merch-card.js');
+
+describe('Offer Selection', () => {
+  before(async () => {
+    document.body.innerHTML = await readFile({ path: './mocks/acrobat-card.html' });
+    document.body.innerHTML = await readFile({ path: './mocks/acrobat-card.html' });
+    await initCard(document.querySelector('.acrobat'));
+    await new Promise((resolve) => { setTimeout(resolve, 100); });
```
yesil, line 11: you can declare a delay function and re-use it.
```
const delay = (duration = 100) => new Promise((resolve) => { setTimeout(resolve, duration); });
...
await delay();
```

libs/blocks/merch-card/merch-card.js
3ch023: you mean to avoid 'offer selection' radios appearing after the merch card was attached to dom?
i can make this change if we are ok with having card 'init' async

3ch023: updated

test/blocks/merch-card/offer-selection.test.js
3ch023: yep, did that

test/blocks/merch-card/mocks/acrobat-card.html
```diff
@@ -0,0 +1,63 @@
...
+                <li>New Description</li>
+              </ul>
+            </li>
+            <li>Monthly, without commitment
+              <ul>
+                <li><span is="inline-price" data-display-per-unit="true" data-display-recurrence="true" data-display-tax="false" data-force-tax-exclusive="true" data-wcs-osi="BWGlzrQG6jgkf-_zPJm55M5QpAyb4skYi05BizQIJ3U" class="placeholder-resolved" ><span class="price" aria-label="US$29.99 per month per license"><span class="price-currency-symbol">US$</span><span class="price-currency-space disabled"></span><span class="price-integer">29</span><span class="price-decimals-delimiter">.</span><span class="price-decimals">99</span><span class="price-recurrence">/mo</span><span class="price-unit-type">per license</span><span class="price-tax-inclusivity disabled"></span></span></span></li>
```
yesil, line 53: I think you should use unprocessed placeholders, here it is already resolved (class`="placeholder-resolved"`)
same for links, they should not have href set yet.

3ch023: doesn't it mean we'd need to copy all the mocking from merch and load merch itself?
otherwise they don't resolve and look quite ugly.
but copying all that stuff seemed also a bit too much to me, especially taking into account that merch-card itself also doesn't use any of the real placeholders

libs/blocks/merch-card/merch-card.js
```diff
@@ -194,6 +195,11 @@ const init = (el) => {
       }
     }
     parseContent(el, altCta, cardType, merchCard);
+    const offerSelection = cardType === 'plans' ? el.querySelector('ul') : null;
```
Axelcureno, line 198: cardtype should be 'product' instead of 'plans', but we have not build that card yet. Just mentioning that we will need to change it in the future

libs/blocks/merch-card/offer-selection.js
```diff
@@ -0,0 +1,48 @@
...
+      }
+    } else {
+      merchOffer.append(override);
+    }
+  });
+  decorateButtons(merchOffer);
+  return merchOffer;
+}
+
+export const initOfferSelection = (el, merchCard, offerSelection) => {
```
Axelcureno, line 36: It doesn't seem we are using `el` here, we may remove it.

3ch023: fixed, thanks for noticing

yesil (approved):

honstar (approved):

libs/deps/merch-offers.js
```diff
@@ -0,0 +1,95 @@
+// Tue, 21 Nov 2023 12:17:48 GMT
```
narcis-radu, line 1: Should we take the whole file content for granted or it's up for discussion?

libs/blocks/merch-card/merch-card.js
```diff
@@ -161,12 +161,13 @@ const init = (el) => {
       merchCard.setAttribute('action-menu', true);
       merchCard.append(createTag('div', { slot: 'action-menu-content' }, actionMenuContent.innerHTML));
     }
-    if (ctas) {
-      const footer = createTag('div', { slot: 'footer' });
+
+    const footer = createTag('div', { slot: 'footer' });
+    if (ctas && ctas.querySelector('a')) {
```
narcis-radu, line 166:
```suggestion
    if (ctas?.querySelector('a')) {
```

libs/blocks/merch-card/offer-selection.js
```diff
@@ -0,0 +1,48 @@
+import { createTag } from '../../utils/utils.js';
+import { decorateButtons } from '../../utils/decorate.js';
+import '../../deps/merch-offers.js';
+
+function createDynamicSlots(el, bodySlot) {
+  const price = createTag('h5', { style: 'margin-top: 8px; margin-bottom: 16px;' });
```
narcis-radu, line 6: what about this?
```js
const price = createTag('h5', { class: 'merchCard-title' });
```

```css
.merchCard-title {
  margin: 8px 0 16px;
}
```

libs/blocks/merch-card/offer-selection.js
```diff
@@ -0,0 +1,48 @@
...
+    } else {
+      merchOffer.append(override);
+    }
+  });
+  decorateButtons(merchOffer);
+  return merchOffer;
+}
+
+export const initOfferSelection = (merchCard, offerSelection) => {
+  const bodySlot = merchCard.querySelector('div[slot="body-xs"]');
```
narcis-radu, line 37: I see this element is very particular. If the element is not found, `createDynamicSlots` will throw errors since you're appending some elements. Maybe you want to handle this particular use-case. If the element is always there you can ignore the message :)

libs/deps/merch-offers.js
yesil: These are built using esbuild. So take for granted.
Feel free to review our merch tacocat PRs at https://git.corp.adobe.com/wcms/tacocat.js/pulls :)

narcis-radu: Thank you!

libs/blocks/merch-card/offer-selection.js
3ch023: fixed

libs/blocks/merch-card/offer-selection.js
3ch023: I will check for its presence, to make the code more stable, thx @narcis-radu .
this element is indeed always there, but nobody knows what will happen in future changes, so it's a good point.

PR #1568, opened 2023-11-20
MWPW-135194: gnav mega menu layout adjustments

robert-bogos:
## Description
This PR introduces the following changes to the global navigation mega menu:
1. the maximum width of the mega menu container is `1200px`;
2. the promo box from the mega menu has a padding of `24px` top-bottom, `32px` left-right;
3. `40px` padding between the top edge of the promo box and the top edge of the mega menu;

## Related Issue
Resolves: [MWPW-135194](https://jira.corp.adobe.com/browse/MWPW-135194)

## Screenshots:
**_1. the maximum width of the mega menu container is `1200px`;_**

Screenshot 2023-11-22 at 21 48 23

**_2. the promo box from the mega menu has a padding of `24px` top-bottom, `32px` left-right;_**

Screenshot 2023-11-22 at 21 45 28

**_3. `40px` padding between the top edge of the promo box and the top edge of the mega menu;_**

Screenshot 2023-11-22 at 21 50 11

## Test URLs
**Acrobat:**
- Before: https://main--dc--adobecom.hlx.page/drafts/rbogos/gnav-layout?martech=off
- After: https://main--dc--adobecom.hlx.page/drafts/rbogos/gnav-layout?milolibs=gnav-layout-adjustments--milo--robert-bogos&martech=off

**BACOM:**
- Before: https://main--bacom--adobecom.hlx.page/drafts/rbogos/gnav-layout
- After: https://main--bacom--adobecom.hlx.page/drafts/rbogos/gnav-layout?milolibs=gnav-layout-adjustments--milo--robert-bogos&martech=off

**CC:**
- Before: https://main--cc--adobecom.hlx.page/drafts/rbogos/gnav-layout?martech=off
- After: https://main--cc--adobecom.hlx.page/drafts/rbogos/gnav-layout?milolibs=gnav-layout-adjustments--milo--robert-bogos&martech=off

**Milo:**
- Before: https://main--milo--robert-bogos.hlx.page/drafts/rbogos/gnav-layout?martech=off
- After: https://gnav-layout-adjustments--milo--robert-bogos.hlx.page/drafts/rbogos/gnav-layout?martech=off

PR #1586, opened 2023-11-28
MWPW-131195: gnav cloud links section

robert-bogos:
## Description
This PR adds a cloud links section on the megamenu from the global navigation.

## Related Issue
Resolves: [MWPW-131195](https://jira.corp.adobe.com/browse/MWPW-131195)

## Screenshots:
Screenshot 2023-11-28 at 09 10 29

## Test URLs
**Acrobat:**
- Before: https://main--dc--adobecom.hlx.page/drafts/rbogos/gnav-layout?martech=off
- After: https://main--dc--adobecom.hlx.page/drafts/rbogos/gnav-layout?milolibs=gnav-cloud-section--milo--robert-bogos&martech=off

**BACOM:**
- Before: https://main--bacom--adobecom.hlx.page/drafts/rbogos/gnav-layout
- After: https://main--bacom--adobecom.hlx.page/drafts/rbogos/gnav-layout?milolibs=gnav-cloud-section--milo--robert-bogos&martech=off

**CC:**
- Before: https://main--cc--adobecom.hlx.page/drafts/rbogos/gnav-layout?martech=off
- After: https://main--cc--adobecom.hlx.page/drafts/rbogos/gnav-layout?milolibs=gnav-cloud-section--milo--robert-bogos&martech=off

**Milo:**
- Before: https://main--milo--robert-bogos.hlx.page/drafts/rbogos/gnav-layout?martech=off
- After: https://gnav-cloud-section--milo--robert-bogos.hlx.page/drafts/rbogos/gnav-layout?martech=off


libs/blocks/global-navigation/utilities/menu/menu.css
```diff
@@ -197,4 +201,51 @@
...
     text-decoration: underline;
   }
+
+
+  .feds-links-strip-wrapper {
+    width: 100%;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    height: 100%;
```
overmyheadandbody, line 211: This probably has no effect, since it doesn't track its parent's height in any way.

libs/blocks/global-navigation/utilities/menu/menu.css
```diff
@@ -197,4 +201,51 @@
   .feds-promo--dark a:hover {
     text-decoration: underline;
   }
+
+
+  .feds-links-strip-wrapper {
+    width: 100%;
+    display: flex;
+    justify-content: center;
+    align-items: center;
```
overmyheadandbody, line 210: This probably has no effect, since is contains just one child.

libs/blocks/global-navigation/utilities/menu/menu.css
```diff
@@ -197,4 +201,51 @@
...
+    height: 100%;
+    border-top: solid 1px #f3f3f3;
+  }
+
+  .feds-links-strip {
+    width: 100%;
+    padding: 20px 0;
+    max-width: var(--feds-maxWidth-nav);
+    display: flex;
+    justify-content: flex-start;
```
overmyheadandbody, line 220: This should happen by default, it probably doesn't need to be declared explicitly.

libs/blocks/global-navigation/utilities/menu/menu.css
```diff
@@ -197,4 +201,51 @@
...
+    align-items: center;
+  }
+
+  .feds-links-strip > *:not(:last-child)::after {
```
overmyheadandbody: We should ideally avoid `*` selectors, as they are non-performant. In this case, we know the children's structure, you could simply do `.feds-linksStrip-item:not(:last-child):after`

libs/blocks/global-navigation/utilities/menu/menu.css
```diff
@@ -197,4 +201,51 @@
...
+    content: '|';
+    margin: 0;
```
overmyheadandbody, line 226: This should be the default value, is another one inherited from some other place? This could probably be omitted.

libs/blocks/global-navigation/utilities/menu/menu.css
```diff
@@ -197,4 +201,51 @@
...
+  }
+
+  .feds-links-strip-item {
+    display: flex;
+    justify-content: center;
```
overmyheadandbody, line 231: Since these items act like they're inline, justifying their content should not achieve anything, this could probably be removed.

libs/blocks/global-navigation/utilities/menu/menu.css
```diff
@@ -197,4 +201,51 @@
...
+    align-items: center;
```
overmyheadandbody, line 232: I see that the `linksStrip-item` contains a decorated `navLink`, which already has this setting, this might be redundant.

libs/blocks/global-navigation/utilities/menu/menu.css
```diff
@@ -197,4 +201,51 @@
...
+  }
+
+  .feds-popup .feds-navLink:hover {
+    background-color: transparent;
```
overmyheadandbody, line 236: This is likely an error, we definitely want background colours for dropdown links, it's an accessibility (and therefore legal) requirement. You probably meant to cancel the background colour just for navLink elements from the linksStrip?

libs/blocks/global-navigation/utilities/menu/menu.css
```diff
@@ -197,4 +201,51 @@
...
+      display: flex;
+      justify-content: center;
+      align-items: center;
+    }
+
+    & img {
+      width: 25px;
+      object-fit: contain;
+    }
+  }
```
overmyheadandbody: This is LESS/SASS syntax and achieves nothing, this whole section could probably be removed.

libs/blocks/global-navigation/utilities/menu/menu.css
```diff
@@ -82,6 +82,10 @@
   display: none;
 }
 
+.feds-links-strip-wrapper {
```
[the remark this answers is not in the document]
robert-bogos: I had just fixed this before you submitted the review 😂

libs/blocks/global-navigation/utilities/menu/menu.js
```diff
@@ -267,6 +267,28 @@ const decorateColumns = async ({ content, separatorTagName = 'H5' } = {}) => {
   }
 };
 
+const decorateLinksStrip = (content) => {
+  const linksStripEl = content.querySelector('.links-strip');
+  if (!linksStripEl) return;
+  linksStripEl.parentElement.removeChild(linksStripEl);
+  const decorateLink = (el) => {
```
overmyheadandbody, line 274: Since this resembles the markup of link groups, would it not be possible to reuse `decorateLinkGroup`?

libs/blocks/global-navigation/utilities/menu/menu.js
```diff
@@ -267,6 +267,28 @@ const decorateColumns = async ({ content, separatorTagName = 'H5' } = {}) => {
...
+    return toFragment`<div class="feds-linksStrip-item">
+      <a href="${link.href}" class="feds-navLink">
+      ${imageEl}
+      ${contentEl}
+      </a>
+    </div>`;
+  };
+
+  const linksStripWrapper = toFragment`<div class="feds-linksStrip-wrapper"><div class="feds-linksStrip"></div></div>`;
+  linksStripWrapper.querySelector('.feds-linksStrip').append(...[...linksStripEl.children].map(decorateLink));
```
overmyheadandbody, line 288: We could probably create 2 different fragments and use each for its specific purpose, rather than querying what we create ourselves.

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -542,7 +542,16 @@ header.global-navigation {
...
     justify-content: center;
-    padding: 32px 12px;
+    padding: 32px 0 0 0;
+  }
+
+  .feds-navItem--megaMenu .feds-popup .feds-menu-content {
+    padding-bottom: 32px;
+  }
+
+  .feds-topnav .feds-nav-wrapper .feds-nav .feds-popup {
```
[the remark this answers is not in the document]
robert-bogos: For the first one, it works only with `.feds-navItem--megaMenu .feds-menu-content`, but for the second one, I added the extra specificity to override the styles from menu.css. Do you think there's a better way to handle this ?

libs/blocks/global-navigation/utilities/menu/menu.css
```diff
@@ -197,4 +201,51 @@
   .feds-promo--dark a:hover {
     text-decoration: underline;
   }
+
+
```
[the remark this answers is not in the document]
robert-bogos: Fixed 👍

libs/blocks/global-navigation/utilities/menu/menu.css
robert-bogos: True, removed it

libs/blocks/global-navigation/utilities/menu/menu.css
robert-bogos: True, removed it

libs/blocks/global-navigation/utilities/menu/menu.css
robert-bogos: True, removed it

libs/blocks/global-navigation/utilities/menu/menu.css
robert-bogos: Done

libs/blocks/global-navigation/utilities/menu/menu.css
robert-bogos: True, removed the margin

libs/blocks/global-navigation/utilities/menu/menu.css
robert-bogos: True, removed it

libs/blocks/global-navigation/utilities/menu/menu.css
robert-bogos: This is necessary for aligning the `|` pseudo-element with the `navLink` inside the `linksStrip-item`. Do you think this can be done with less css ?

libs/blocks/global-navigation/utilities/menu/menu.css
robert-bogos: Yes, it's an error. I wanted to cancel the background only for the elements inside the linksStrip

libs/blocks/global-navigation/utilities/menu/menu.css
robert-bogos: The `&` nesting selector is also valid CSS syntax. I used it to handle the image coming from the `*.docx`, but I'll probably need to remove this if I switch to svgs

libs/blocks/global-navigation/utilities/menu/menu.js
```diff
@@ -267,6 +267,28 @@ const decorateColumns = async ({ content, separatorTagName = 'H5' } = {}) => {
   }
 };
 
+const decorateLinksStrip = (content) => {
+  const linksStripEl = content.querySelector('.links-strip');
+  if (!linksStripEl) return;
+  linksStripEl.parentElement.removeChild(linksStripEl);
```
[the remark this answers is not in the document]
robert-bogos: Done

libs/blocks/global-navigation/utilities/menu/menu.js
```diff
@@ -267,6 +267,28 @@ const decorateColumns = async ({ content, separatorTagName = 'H5' } = {}) => {
   }
 };
 
+const decorateLinksStrip = (content) => {
```
mokimo, line 270: Should this be only loaded when actually used? Same for the CSS. @overmyheadandbody knows more, but I think this is only used on a small sub-section of pages

overmyheadandbody: In theory, this should be used in a wide range of places, except the homepage, where we already have all the clouds; and BACOM, where they don't seem to be showing the cloud menu.
* https://www.adobe.com/products/photoshop.html - CC subpages have it
* https://www.adobe.com/acrobat.html - DC subpages have it

However, this won't be authored in the near future, I'd assume, so it might be dead code for a while. I wouldn't necessarily do a dynamic import, as it's already part of `menu.js`, which is loaded later in the flow anyway. I'd avoid longer chains of dynamic imports and instead focus on failing fast and having a low code footprint for the new functionality.

libs/blocks/global-navigation/utilities/menu/menu.css
overmyheadandbody: Regarding SVGs, I think I misunderstood the intent. For now, it should be fine to keep just .jpg/.png support. Hopefully we can re-use the linkGroup decoration logic so that when we add SVG support there, it trickles down here as well.

Regarding the CSS nesting - I still wouldn't use it right now, as browsers with full support have started rolling it out in October/November this year. I'd wait a bit before commiting to using it

libs/blocks/global-navigation/utilities/menu/menu.js
mokimo: Sounds reasonable! No action required in this case

libs/blocks/global-navigation/utilities/menu/menu.js
robert-bogos: Yes, we can reuse `decorateLinkGroup`. I modified the code

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -29,10 +29,11 @@ import {
...
 
-const CONFIG = {
+export const CONFIG = {
   icons: {
     company: '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 133.46 118.11" alt="Adobe, Inc."><defs><style>.cls-1{fill:#fa0f00;}</style></defs><polygon class="cls-1" points="84.13 0 133.46 0 133.46 118.11 84.13 0"/><polygon class="cls-1" points="49.37 0 0 0 0 118.11 49.37 0"/><polygon class="cls-1" points="66.75 43.53 98.18 118.11 77.58 118.11 68.18 94.36 45.18 94.36 66.75 43.53"/></svg>',
     search: '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" focusable="false"><path d="M14 2A8 8 0 0 0 7.4 14.5L2.4 19.4a1.5 1.5 0 0 0 2.1 2.1L9.5 16.6A8 8 0 1 0 14 2Zm0 14.1A6.1 6.1 0 1 1 20.1 10 6.1 6.1 0 0 1 14 16.1Z"></path></svg>',
+    home: '<svg xmlns="http://www.w3.org/2000/svg" height="25" viewBox="0 0 18 18" width="25"><path fill="#6E6E6E" d="M17.666,10.125,9.375,1.834a.53151.53151,0,0,0-.75,0L.334,10.125a.53051.53051,0,0,0,0,.75l.979.9785A.5.5,0,0,0,1.6665,12H2v4.5a.5.5,0,0,0,.5.5h4a.5.5,0,0,0,.5-.5v-5a.5.5,0,0,1,.5-.5h3a.5.5,0,0,1,.5.5v5a.5.5,0,0,0,.5.5h4a.5.5,0,0,0,.5-.5V12h.3335a.5.5,0,0,0,.3535-.1465l.979-.9785A.53051.53051,0,0,0,17.666,10.125Z"/></svg>',
```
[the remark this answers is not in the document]
robert-bogos: It's used in the `decorateCrossCloudMenu` function inside `menu.js`

libs/blocks/global-navigation/utilities/menu/menu.css
```diff
@@ -197,4 +201,31 @@
   .feds-promo--dark a:hover {
     text-decoration: underline;
   }
+
+  .feds-linksStrip-wrapper {
```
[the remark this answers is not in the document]
robert-bogos: Done 👍

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -550,7 +550,15 @@ header.global-navigation {
   .feds-navItem--megaMenu .feds-popup {
     right: 0;
     justify-content: center;
-    padding: 40px 12px;
+    padding: 40px 0 0 0;
+  }
+
+  .feds-navItem--megaMenu .feds-menu-content {
+    padding-bottom: 40px;
```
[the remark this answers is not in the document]
robert-bogos: I made this change to keep the bottom padding added from https://github.com/adobecom/milo/pull/1568.

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -550,7 +550,15 @@ header.global-navigation {
   .feds-navItem--megaMenu .feds-popup {
     right: 0;
     justify-content: center;
-    padding: 40px 12px;
+    padding: 40px 0 0 0;
```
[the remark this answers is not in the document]
robert-bogos: This looks like an use case for the colorful navigation that I was not aware of. 
Thanks for pointing out, I've fixed it 👍

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -550,7 +550,15 @@ header.global-navigation {
...
+  }
+
+  .feds-navItem--megaMenu .feds-menu-content {
+    padding-bottom: 40px;
+  }
+
+  .feds-topnav .feds-nav-wrapper .feds-nav .feds-popup {
+    flex-direction: column;
```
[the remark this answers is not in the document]
robert-bogos: This has a higher level of specificity than the rule from `menu.css` and it overrides it. 
It's only meant for the global navigation; in other cases, like for the footer, the rule from `menu.css` applies

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -673,4 +681,8 @@ header.global-navigation {
   .feds-menu-container {
     min-width: var(--feds-minWidth-menu);
   }
+
+  .feds-topnav .feds-nav-wrapper .feds-nav .feds-popup {
+    align-items: center;
```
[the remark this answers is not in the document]
robert-bogos: This keeps the `.feds-menu-container` center aligned inside the `.feds-popup`; otherwise it would default to `align-items: stretch;`

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -550,7 +551,11 @@ header.global-navigation {
   .feds-navItem--megaMenu .feds-popup {
     right: 0;
     justify-content: center;
-    padding: 40px 12px;
+    padding: 40px 0 0 0;
+  }
+
+  #feds-popup {
```
mokimo, line 557: As discussed on Slack, let's find a different way around higher specificity rather than using IDs. You have a bunch of options.
- A longer selector (2-3 +- a pseudo selector)
- Refactoring the CSS that leads to a longer selector to begin with to keep styles flat, you shouldn't ever need to go more than 2 or 3 selectors deep.
- Figuring out why you need to overwrite this to begin with, overwriting something as the flex direction creates more branches the code can go into & thus is harder to test/catch and might lead to more issues

libs/blocks/global-navigation/utilities/menu/menu.js
```diff
@@ -276,6 +277,21 @@ const decorateColumns = async ({ content, separatorTagName = 'H5' } = {}) => {
...
+  const crossCloudMenuContent = toFragment`<div class="feds-crossCloudMenu"></div>`;
+  crossCloudMenuContent.append(...[...crossCloudMenuEl.children].map((el, index) => {
+    const decoratedLink = decorateLinkGroup(el, index);
+    if (index === 0) decoratedLink.prepend(toFragment`${CONFIG.icons.home}`);
+    return toFragment`<div class="feds-crossCloudMenu-item">${decoratedLink}</div>`;
+  }));
+
+  content.append(toFragment`<div class="feds-crossCloudMenu-wrapper">${crossCloudMenuContent}</div>`);
+  crossCloudMenuEl.remove();
+};
```
mokimo, line 293:
```suggestion
const decorateCrossCloudMenu = (content) => {
  const crossCloudMenuEl = content.querySelector('.cross-cloud-menu');
  if (!crossCloudMenuEl) return;

  const crossCloudMenu = toFragment`<div class="feds-crossCloudMenu-wrapper"<div class="feds-crossCloudMenu"></div></div>`;
  // I'm not a big fan of the double destructing assignment
  crossCloudMenu.append(...[...crossCloudMenuEl.children].map((el, index) => {
    const decoratedLink = decorateLinkGroup(el, index);
    // Does this need to be nested?
    if (index === 0) decoratedLink.prepend(toFragment`${CONFIG.icons.home}`);
    return toFragment`<div class="feds-crossCloudMenu-item">${decoratedLink}</div>`;
  }));

  content.append(crossCloudMenu);
  crossCloudMenuEl.remove();
};
```

libs/blocks/global-navigation/utilities/menu/menu.js
```diff
@@ -303,11 +319,12 @@ const decorateMenu = (config) => logErrorFor(async () => {
     const content = await res.text();
     const parsedContent = await replaceText(content, getFedsPlaceholderConfig(), undefined, 'feds');
     const menuContent = toFragment`<div class="feds-menu-content">${parsedContent}</div>`;
-    menuTemplate = toFragment`<div class="feds-popup">
+    menuTemplate = toFragment`<div class="feds-popup" id="feds-popup">
```
mokimo, line 322: Since you add the id for styling purposes, you probably can remove this again if you find a different solution.

test/blocks/global-navigation/test-utilities.js
```diff
@@ -128,7 +131,7 @@ export const createFullGlobalNavigation = async ({
   window.fetch = stub().callsFake((url) => {
     if (url.includes('profile')) { return mockRes({ payload: defaultProfile }); }
     if (url.includes('placeholders')) { return mockRes({ payload: placeholders || defaultPlaceholders }); }
-    if (url.endsWith('large-menu.plain.html')) { return mockRes({ payload: largeMenuMock }); }
+    if (url.endsWith('large-menu.plain.html')) { return mockRes({ payload: hasCrossCloudMenu ? largeMenuMock : largeMenuWithoutCrossCloudMenu }); }
```
mokimo, line 134: Should this not be two different menu's and a different url if one has a cross cloud menu and the other doesn't ?

libs/blocks/global-navigation/global-navigation.css
robert-bogos: Solved 👍 Looks like the rule I was trying to override was not needed to begin with

libs/blocks/global-navigation/utilities/menu/menu.js
robert-bogos: 1. The use of 'forEach' here is a tradeoff. It makes the code slightly longer, but also makes it more readable. Made the change 👍
2. On the AEM version of the Global Navigation I saw it's nested, so I tried to keep the consistency

libs/blocks/global-navigation/utilities/menu/menu.js
robert-bogos: Done 👍

test/blocks/global-navigation/test-utilities.js
robert-bogos: Correct. Made the change 👍

test/blocks/global-navigation/global-navigation.test.js
```diff
@@ -63,6 +64,54 @@ describe('global navigation', () => {
...
+      it('should render the Cross Clound Menu', async () => {
+        await createFullGlobalNavigation({ globalNavigation: globalNavigationCrossCloud });
+        const crossCloudMenu = document.querySelector(selectors.crossCloudMenuWrapper);
+
+        expect(crossCloudMenu).to.exist;
+        expect(isElementVisible(crossCloudMenu)).to.equal(false);
+
+        document.querySelector(`${selectors.largeMenu} ${selectors.navLink}`).click();
+
+        [...crossCloudMenu.querySelectorAll(selectors.navLink)].forEach((el) => {
```
mokimo, line 78: NodeLists already expose a `forEach` method https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach

I don't think the destructuring here is neccesary

test/blocks/global-navigation/global-navigation.test.js
```diff
@@ -63,6 +64,54 @@ describe('global navigation', () => {
...
+    describe('small desktop', () => {
+      it('should render the Cross Clound Menu', async () => {
+        await createFullGlobalNavigation({ globalNavigation: globalNavigationCrossCloud, viewport: 'smallDesktop' });
+        const crossCloudMenu = document.querySelector(selectors.crossCloudMenuWrapper);
+
+        expect(isElementVisible(crossCloudMenu)).to.equal(false);
+
+        document.querySelector(`${selectors.largeMenu} ${selectors.navLink}`).click();
+
+        [...crossCloudMenu.querySelectorAll(selectors.navLink)].forEach((el) => {
```
mokimo, line 98: NodeLists already expose a `forEach` method https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach

I don't think the destructuring here is neccesary

mokimo (approved):
Two minor comments

mokimo:
On safari the menu isn't centered
Screenshot 2023-12-14 at 12 10 56

mokimo:
On mobile this leads to a piece of extra spacing 
Screenshot 2023-12-14 at 12 12 22

test/blocks/global-navigation/global-navigation.test.js
robert-bogos: Good catch! I fixed it 👍

test/blocks/global-navigation/global-navigation.test.js
robert-bogos: Good catch! I fixed it 👍

robert-bogos:
> On safari the menu isn't centered Screenshot 2023-12-14 at 12 10 56

It was an inconsistency with AEM, fixed now 👍

robert-bogos:
> On mobile this leads to a piece of extra spacing Screenshot 2023-12-14 at 12 12 22

That's an already existing problem, but will be fixed in my next PR about padding consistency on mobile

narcis-radu (approved):

mokimo (approved):

libs/blocks/global-navigation/utilities/menu/menu.js
```diff
@@ -13,6 +13,7 @@ import {
 } from '../utilities.js';
 import { decorateLinks } from '../../../../utils/utils.js';
 import { replaceText } from '../../../../features/placeholders.js';
+import { CONFIG } from '../../global-navigation.js';
```
overmyheadandbody, line 16: Why add the icon to the main file, when it can be added here directly? Maybe a navigation doesn't include dropdown menus, such as the plans pages. Then you have that icon with no use in the main file. Are there other reasons for it being there?

libs/blocks/global-navigation/utilities/menu/menu.js
```diff
@@ -273,6 +274,22 @@ const decorateColumns = async ({ content, separatorTagName = 'H5' } = {}) => {
   }
 };
 
+const decorateCrossCloudMenu = (content) => {
+  const crossCloudMenuEl = content.querySelector('.cross-cloud-menu');
+  if (!crossCloudMenuEl) return;
+
+  const crossCloudMenuContent = toFragment`<div class="feds-crossCloudMenu"></div>`;
+
+  [...crossCloudMenuEl.querySelector('ul').children].forEach((el, index) => {
```
overmyheadandbody, line 283: Could you not simply select `ul a` instead of children?

libs/blocks/global-navigation/utilities/menu/menu.js
```diff
@@ -273,6 +274,22 @@ const decorateColumns = async ({ content, separatorTagName = 'H5' } = {}) => {
...
+    const decoratedLink = decorateLinkGroup(el, index);
```
overmyheadandbody, line 284: Continuing the previous comment, if an author forgets to add a link and leaves a simple text list item, then this will return an empty string. Trying to prepend things to it in the next step might result in issues

libs/blocks/global-navigation/utilities/menu/menu.js
```diff
@@ -273,6 +274,22 @@ const decorateColumns = async ({ content, separatorTagName = 'H5' } = {}) => {
...
+    if (index === 0) decoratedLink.prepend(toFragment`${CONFIG.icons.home}`);
+    crossCloudMenuContent.append(toFragment`<div class="feds-crossCloudMenu-item">${decoratedLink}</div>`);
+  });
```
overmyheadandbody, line 287: Come to think of it, why not use `decorateElements({ elem: crossCloudMenuEl })` directly? We get a bit more markup, but we avoid swapping out entire elements for newly built ones

libs/blocks/global-navigation/utilities/menu/menu.js
```diff
@@ -273,6 +274,22 @@ const decorateColumns = async ({ content, separatorTagName = 'H5' } = {}) => {
...
+
+  content.append(toFragment`<div class="feds-crossCloudMenu-wrapper">${crossCloudMenuContent}</div>`);
```
overmyheadandbody, line 289: I see that the unit tests use a `selectors.crossCloudMenuWrapper`, could we not use it here as well?

test/blocks/global-navigation/global-navigation.test.js
```diff
@@ -63,6 +64,49 @@ describe('global navigation', () => {
     });
   });
 
+  describe('Cross Clound Menu', () => {
```
overmyheadandbody, line 67: Nit: typo in 'Cloud'. Same for each `it` declaration

overmyheadandbody (commented):
I still have a few suggestions 😅

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -536,6 +536,7 @@ header.global-navigation {
 
   .feds-menu-container {
     min-width: 100%;
+    padding: 0 12px 40px 12px;
```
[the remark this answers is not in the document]
robert-bogos: Done 👍

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -550,7 +551,7 @@ header.global-navigation {
   .feds-navItem--megaMenu .feds-popup {
     right: 0;
     justify-content: center;
-    padding: 40px 12px;
+    padding: 40px 0 0 0;
```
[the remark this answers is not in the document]
robert-bogos: Done 👍

libs/blocks/global-navigation/utilities/menu/menu.js
robert-bogos: Changed 👍

libs/blocks/global-navigation/utilities/menu/menu.js
robert-bogos: Doesn't quite work in here because `selectors.crossCloudMenuWrapper` has a dot at the beginning -> `crossCloudMenuWrapper: '.feds-crossCloudMenu-wrapper'`

test/blocks/global-navigation/global-navigation.test.js
robert-bogos: Fixed 👍

libs/blocks/global-navigation/utilities/menu/menu.js
overmyheadandbody: One strategy we're using in other places is simply calling `.slice(1)`

libs/blocks/global-navigation/utilities/menu/menu.css
```diff
@@ -203,3 +201,45 @@
...
+    content: '|';
+  }
+
+  .feds-crossCloudMenu-item {
+    display: flex;
+    align-items: center;
+    list-style: none;
+  }
+
+  .feds-crossCloudMenu-item u {
```
overmyheadandbody, line 234: Would this be a common use-case? We usually cancel it for anchor elements, but I haven't seen any underline ones

overmyheadandbody (commented):
Dropdown menus are not expanding on https://gnav-cloud-section--milo--robert-bogos.hlx.page/drafts/rbogos/page-cross-cloud-menu?martech=off and one unit test is failing, not sure if they're related.

robert-bogos:
> ~Dropdown menus are not expanding on https://gnav-cloud-section--milo--robert-bogos.hlx.page/drafts/rbogos/page-cross-cloud-menu?martech=off and~ one unit test is failing, not sure if they're related.
> 
> LE: dropdown expansion seems to have been a temporary issue, it might be that I had connectivity issues. It is no longer happening.

Looks like that unit test it's not related to this PR, it also fails on other PRs

overmyheadandbody (approved):
Looks good, thanks for integrating everything!

SilviuLCF (approved):
Ready to deploy 

Testing details : 
https://jira.corp.adobe.com/browse/MWPW-131195

PR #1588, opened 2023-11-28
MWPW-137905 Integrate Heartbeat Video Tracking on YouTube Videos

suhjainadobe:
This request is in anticipation of future YouTube livestreams being added to A.com Firefly launch pages.
Requirement wiki: https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=marketingtech&title=YouTube+Video+Tracking+Set+Up

Resolves: [[MWPW-137905](https://jira.corp.adobe.com/browse/MWPW-137905)

**Test URLs:**
- Before: https://main--milo--suhjainadobe.hlx.page/drafts/suhjain/document2?martech=off
- After: https://youtube--milo--suhjainadobe.hlx.page/drafts/suhjain/document2?martech=off


auniverseaway:
Please turn on linting in your editor so your PRs don't show up with linting errors. You can also run `npm run lint`.

libs/blocks/youtube/youtube.js
```diff
@@ -1,11 +1,11 @@
 import { createIntersectionObserver, isInTextNode } from '../../utils/utils.js';
 
 export default function init(a) {
+  const searchParams = new URLSearchParams(a.search);
```
mokimo, line 4: Why move it out of the embedVideo?

libs/blocks/youtube/youtube.js
```diff
@@ -14,12 +14,19 @@ export default function init(a) {
...
+        id='player-${id}'
         title="${title}">
       </iframe>
     </div>`;
     a.insertAdjacentHTML('afterend', embedHTML);
     a.remove();
   };
-
   createIntersectionObserver({ el: a, callback: embedVideo });
+  document.onreadystatechange = () => {
```
mokimo, line 25: Should we not move this within the `embedVideo`? 
Otherwise, this might get tracked while it is never shown on the page.

overmyheadandbody, line 25: Couldn't a simple event listener be used instead of overhauling the functionality altogether? Like `document.addEventListener('readystatechange')`. Otherwise any other pieces of logic depending on this event might get impacted.

I also agree with @mokimo that we wouldn't want this tracked if the video is not actuallt embedded.

suhjainadobe: moved it up

suhjainadobe: done

libs/blocks/youtube/youtube.js
```diff
@@ -14,12 +14,18 @@ export default function init(a) {
         webkitallowfullscreen mozallowfullscreen allowfullscreen
         allow="encrypted-media; accelerometer; gyroscope; picture-in-picture"
         scrolling="no"
+        id='player-${id}'
```
overmyheadandbody, line 17: Nit: other attribute values are surrounded by double quotes, this is surrounded by single quotes, we should follow the same pattern here as well.

auniverseaway: +1. this is HTML. Should be double quotes.

auniverseaway (commented):
See notes.

libs/blocks/youtube/youtube.js
suhjainadobe: done

libs/blocks/youtube/youtube.js
```diff
@@ -14,12 +14,18 @@ export default function init(a) {
...
         allow="encrypted-media; accelerometer; gyroscope; picture-in-picture"
         scrolling="no"
+        id="player-${id}"
         title="${title}">
       </iframe>
     </div>`;
     a.insertAdjacentHTML('afterend', embedHTML);
     a.remove();
+    document.addEventListener('readystatechange', () => {
+      if (document.readyState === 'complete') {
```
mokimo, line 24: Does it need the `readystatechange` complete event listener? I would be under the impression the page is already ready and analytics have been added by the time we render this block

mokimo (approved):

libs/blocks/youtube/youtube.js
suhjainadobe: @mokimo : Sometimes Launch's load is not completed and it fails to trigger the call. So to make sure that everything is loaded, this EventListener is added.

nishantka (approved):

libs/blocks/youtube/youtube.js
```diff
@@ -14,12 +14,18 @@ export default function init(a) {
         webkitallowfullscreen mozallowfullscreen allowfullscreen
         allow="encrypted-media; accelerometer; gyroscope; picture-in-picture"
         scrolling="no"
+        id="player-${id}"
         title="${title}">
       </iframe>
     </div>`;
     a.insertAdjacentHTML('afterend', embedHTML);
     a.remove();
+    document.addEventListener('readystatechange', () => {
```
overmyheadandbody, line 23: Is it possible that the event is added after the `readyState` is already `complete`? If so, it should first be checked if the `document.readyState` is already `complete`. If so, call the satellite method. If not, add the event listener.

suhjainadobe: I all the scenarios we have not seen this case when tested at our end. As this is the loading state of the document, it is unlikely this will happen.

libs/blocks/youtube/youtube.js
```diff
@@ -14,12 +14,22 @@ export default function init(a) {
...
         allow="encrypted-media; accelerometer; gyroscope; picture-in-picture"
         scrolling="no"
+        id="player-${id}"
         title="${title}">
       </iframe>
     </div>`;
     a.insertAdjacentHTML('afterend', embedHTML);
     a.remove();
+    if (document.readyState === 'complete') {
+      window._satellite?.track('trackYoutube');
```
overmyheadandbody, line 24: You should also add the ESLint exception here.

overmyheadandbody (approved):

libs/blocks/youtube/youtube.js
```diff
@@ -14,12 +14,23 @@ export default function init(a) {
...
     a.insertAdjacentHTML('afterend', embedHTML);
     a.remove();
+    if (document.readyState === 'complete') {
+      /* eslint-disable-next-line no-underscore-dangle */
+      window._satellite?.track('trackYoutube');
+    } else {
+      document.addEventListener('readystatechange', () => {
+        if (document.readyState === 'complete') {
+          /* eslint-disable-next-line no-underscore-dangle */
+          window._satellite?.track('trackYoutube');
```
chrischrischris, line 30: Should a named callback be used so the event listener can be removed?

overmyheadandbody: I was going to suggest the `once` approach, but now I have another question - can we have multiple YouTube videos on a page? If so, would all videos be considered by this satellite call?

spadmasa:
PR has been verified by Atanu and details are updated in Jira
image

libs/blocks/youtube/youtube.js
suhjainadobe: Pasting Atanu's comment who requested this functionality:
image

suhjainadobe: @overmyheadandbody : Does this answer your request? Is there anything else you need from my side?

overmyheadandbody: Hey @suhjainadobe! PR was approved from my side; this comment shows that follow-ups may be needed in the future since this seems to be a POC

suhjainadobe:
Merging the PR as I have the required approvals. May need to refer back in future as the analytics is in POC

libs/blocks/youtube/youtube.js
suhjainadobe: Yes, we may need to refer it in future. Merging the PR as it is needed by the analytics team.

PR #1591, opened 2023-11-29
MWPW-139077 - New Form Block

elan-tbx:
* This PR adds a new Form block, which utilizes Franklin's forms architecture to produce a form from, and save submissions within, the same Excel document. [More info here](https://www.aem.live/developer/forms).
* There is a basic "Rules" system in place that allows for classes to be added to a field based on comparisons of text, numbers, or dates. The operator shorthands used here are based off of the [Python operator function names](https://docs.python.org/3/library/operator.html). Any recommendations for alternatives are welcome.
* The block itself is pretty simple -- just a row with the JSON link for the Excel doc, and a row with either a "thank you" message or a URL to redirect to after submission. You can view the JSON data being used in the URLs below here: https://main--milo--adobecom.hlx.page/drafts/ebartholomew/forms/fresh-form1.json
* This is just a jumping off point. Please beat this up/add any suggestions you can think of.

Resolves: [MWPW-139077](https://jira.corp.adobe.com/browse/MWPW-139077)

**Test URLs:**

Milo:
- Before: https://main--milo--adobecom.hlx.page/drafts/ebartholomew/elan-form3?martech=off
- After: https://ebartholomew-form-block--milo--adobecom.hlx.page/drafts/ebartholomew/elan-form3?martech=off

Blog:
- Before: https://main--blog--adobecom.hlx.page/en/drafts/ebartholomew/form?martech=off
- After: https://main--blog--adobecom.hlx.page/en/drafts/ebartholomew/form?milolibs=ebartholomew-form-block&martech=off


rgclayton:
If i try to submit with an empty required field, I do get the error message. But, the submit button stays grayed out even after I add to the required field. I have to refresh the page to get the submit button re-enabled.

image

rgclayton:
I'm seeing the `hidden` field even though the `favorite number` hasn't been populated.

Not sure what I did to get that. I might have refreshed the page after the submit button got stuck in the disabled state. See previous comment.
image

rgclayton:
Nit. Maybe add the image(s) to an image folder `img`. Or serve them out of `/libs/`

elan-tbx:
> I'm seeing the `hidden` field even though the `favorite number` hasn't been populated.
> 
> Not sure what I did to get that. I might have refreshed the page after the submit button got stuck in the disabled state. See previous comment.

@rgclayton Yeah, that's actually working as expected (notice the "or empty" in parentheses). I think it's a matter of debate -- do we always want to force the rule comparisons to evaluate to `true` when the target field is empty? Might be an opportunity for enhancement on how this part works so an author has control over that.

For now though, considering that `.hidden` is the only class that is handled by the block's CSS, I see your point; makes sense to force `true` in this case.

libs/blocks/form/form.js
```diff
@@ -0,0 +1,378 @@
...
+  [...form.elements].forEach((fe) => {
+    if (fe.type.match(/(?:checkbox|radio)/)) {
+      if (fe.checked) {
+        payload[fe.name] = payload[fe.name] ? `${fe.value}, ${payload[fe.name]}` : fe.value;
+      } else if (fe.closest('.group-container').classList.contains('required')) {
+        payload[fe.name] = 'required-not-checked';
+      }
+    } else if (fe.id) {
+      payload[fe.id] = fe.value;
+    }
```
rgclayton, line 47: Any way to flatten some of this?

libs/blocks/form/form.js
```diff
@@ -0,0 +1,378 @@
...
+          form.append(thanksText);
+          setTimeout(() => thanksText.remove(), 2000);
+          /* c8 ignore next 3 */
+        } else {
+          window.location.href = handleThankYou;
+        }
+      }
+    });
+  } else if (fd.Type === 'clear') {
+    button.classList.add('outline');
```
rgclayton, line 120: Nit. Does this need to be an else if? 
If submit do submit stuff.
If clear do clear stuff.

rgclayton (approved):
A few nits for simplicity and readability. 
Working well though.

libs/blocks/form/form.js
```diff
@@ -0,0 +1,378 @@
+import sanitizeComment from '../../utils/sanitizeComment.js';
+
+const RULE_OPERATORS = {
+  equal: 'eq',
```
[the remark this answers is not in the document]
elan-tbx: Good point, that makes a lot more sense. I'll implement this and update the [RFC](https://github.com/orgs/adobecom/discussions/1596)

libs/blocks/form/form.js
```diff
@@ -0,0 +1,379 @@
...
+    if (fe.type.match(/(?:checkbox|radio)/)) {
+      fe.checked = false;
+    } else {
+      fe.value = '';
+    }
+  });
+}
+
+function createButton(fd, thankYou) {
+  const button = document.createElement('button');
```
ryanmparrish, line 93:
```suggestion
  const button = createTag('button', { class: 'button' }, fd.Label)
```
You could condense a number of lines thru out this file w/ the createTag func.

elan-tbx:
Thanks @ryanmparrish 

> * In `form.js` it looks like you could import and use our `createTag` util instead of `createElement` and save some lines adding classes and content in the parameters.

Can't believe I forgot about this. Thanks. Reduced quite a few lines.

> * In `form.css` we may want to load those image assets from [libs/img/ui](https://adobe.sharepoint.com/sites/adobecom/Shared%20Documents/Forms/AllItems.aspx?id=%2Fsites%2Fadobecom%2FShared%20Documents%2Fmilo%2Flibs%2Fimg%2Fui&viewid=d776cf70%2D9b7e%2D4ab7%2Db9da%2D9e0f8e03a7d2) similar to other blocks w/ ui icons inserted (carousel, marquee-anchor)

For now I'm going to keep these as base64. They were taken directly from Spectrum, so AFAIK they should already be in the shared library for others to use.

> * When there is a select w/ 'other' as an option, should another input field appear to add the other?

This should be up to the authors. They have this ability via the "Rules" feature of this block.

> * The resizable fields may want some max-widths or only Y styles. (You can grab that handle and really throw things around)

Good catch -- I've set it to only allow vertical resizing.

> * The redirect transition seems a little un-informative. (fields go white and no message of being redirected) - This may be a future UI enhancement thing. IDK, just pointing it out.

I think this would be a good thing to hammer out for the next version of the block. We'd want to establish how it's authored first (currently the final row expects a "thank you" message OR a redirect URL. We'd need to agree on a plan on how to best include both)

ryanmparrish (approved):
Looks good. Thanks for the deets. 
FYI, There is one lint dog barking around 'hideTitleOnReload'

chrischrischris (approved):

arumsey:
@elan-tbx What happens if a milo-based site already has its own `form` block that pre-dates this `form` block? Ever since this PR was merged the form block on https://aem-trial.adobe.com/headless no longer works. Is there a confguration that forces custom blocks with the same name to be used? Isn't this a big risk introducing new blocks that may have the same name as custom blocks?

- https://github.com/adobe/aem-trials/tree/main/blocks/form

PR #1532, opened 2023-11-14
[MWPW-136943] Colorful Gnav headings for A/B test

overmyheadandbody:
## Description
This enables authors to add gradient backgrounds to _Link Group_ blocks in Gnav navigation documents. This is needed to support an upcoming UI A/B test for the Creativity & Design dropdown menu content. Considering this, we've implemented a minimal set of changes; the features will be expanded or completely removed depending on the test outcome. A few notes, in order of files:

- gradient backgrounds can be added by using some special class names. Since we don't know whether this feature has a future, we didn't yet want to enable authors to define any background possibility on their own. The class names are defined based on the proposed design and which elements require the gradient backgrounds: `photo-gradient, design-gradient, 3d-gradient, pdf-gradient, video-gradient, ai-gradient`. I've added the design specs below;
- the selectors for the Keyboard Navigation have been slightly updated. Some anchor elements were switched to buttons in a previous PR, but the change wasn't reflected in the Keyboard Navigation logic as well. This was affecting production behaviour;
- some padding and margin values have been updated to ensure a more consistent layout across all of the dropdown content use cases;
- the `75%` for wide columns has been chosen with the following logic: only if the section contains 3 columns does it get the `--wide` modifier. In the mega menu dropdown we usually have 4 columns, so 75% has been chosen to reflect this. We should not have all 4 columns nested under a single headline, since it would be redundant, given that we have the dropdown trigger above. If this feature is a winner, we'll likely need to get back to it and refine it to fit all of the possible use-cases. Without this 75% limitation, the design looks too stretched and strays away from the suggested specs;
- a new _Column Break_ block has been created to separate the content of a column in multiple sections. This is needed to be able to group multiple columns under a single headline. In the design spec below, 3 columns are nested under a single _Featured Products_ headline. We wanted to avoid nesting tables and over-fragmentising content, so this is the best compromise in the context of an A/B test;

Screenshot 2023-09-29 at 12 35 13 PM

## Related Issue
Resolves: [MWPW-136943](https://jira.corp.adobe.com/browse/MWPW-136943)

## Testing instructions
Just open the **first** dropdown and notice the colourful headings on desktop devices.

## Screenshots:
Screenshot 2023-11-14 at 14 46 24

## Test URLs
**Homepage:**
- Before: https://main--homepage--adobecom.hlx.page/drafts/ramuntea/index-regular?martech=off
[...]

PR #1595, opened 2023-12-01
[MWPW-131197] Gnav active element detection

overmyheadandbody:
## Description
To reach better feature parity between the AEM and Milo versions of the Global Navigation, this PR implements active element detection for navigation elements. A detailed Wiki is available in the JIRA issue, describing all use-cases.

A few developer notes, in order of files:
- test URLs for the Homepage have been added to the _gnav.md_ custom PR template;
- the `column-break` synthetic block added for https://github.com/adobecom/milo/pull/1532 has been added to _fallback.js_ to avoid false alerts when previewing navigation files;
- link groups that don't contain links will be transformed into empty strings to avoid `null` elements in the final markup;
- you might notice that on top of the `.feds-navItem--active` class, another `.feds-navItem--activeDeferred` is used. The latter is used for when no items loaded synchronously contain an active element and the navigation contains multiple asynchronously-loaded section menus. Since we depend on async code to mark the active element, we want to adapt the behaviour to avoid potential CLS issues, thus we use this extra modifier. It gets removed once the viewport switches to mobile or the content of the navigation is overflowing, since CLS should no longer be impacted.

## Related Issue
Resolves: [MWPW-131197](https://jira.corp.adobe.com/browse/MWPW-131197)

## Testing instructions
Go on a page whose URL is referenced by a link from the Global Navigation. The top navigation item that contains that link should have its text bolded and receive a bottom border.

Feel free to edit some experiences and add multiple items with the potential of being active. There should obviously be just one active item in the navigation. If multiple links match the criteria, the Wiki describes which gets priority.

## Screenshots:
Screenshot 2023-12-01 at 15 04 58

## Test URLs
**Acrobat** (link in dropdown):
- After: https://www.stage.adobe.com/acrobat/online/sign-pdf.html?martech=off&milolibs=gnav-active-elem--milo--overmyheadandbody

**BACOM** (simple top navigation link):
- Before: https://main--bacom--adobecom.hlx.page/products/real-time-customer-data-platform/rtcdp?martech=off
- After: https://main--bacom--adobecom.hlx.page/products/real-time-customer-data-platform/rtcdp?martech=off&milolibs=gnav-active-elem--milo--overmyheadandbody

**CC** (cloud menu):
- After: https://main--cc--adobecom.hlx.live/?martech=off&milolibs=gnav-active-elem--milo--overmyheadandbody

**Homepage** (no item):
- After: https://main--homepage--adobecom.hlx.page/homepage/index-loggedout?martech=off&milolibs=gnav-active-elem--milo--overmyheadandbody

**Milo** (multiple cloud menus):
- Before: https://main--milo--overmyheadandbody.hlx.page/drafts/ramuntea/gnav-refactor-colourful?foo=bar#baz&martech=off
- After: https://gnav-active-elem--milo--overmyheadandbody.hlx.page/drafts/ramuntea/gnav-refactor-colourful?martech=off

### DIFF
--- a/libs/blocks/global-navigation/base.css
+++ b/libs/blocks/global-navigation/base.css
@@ -115,14 +115,51 @@
     padding: 0 12px;
   }
 
+  .feds-navItem--active > .feds-navLink {
+    position: relative;
+    font-weight: 700;
+  }
+
+  .feds-navItem--active > .feds-navLink:only-child {
+    cursor: default;
+  }
+
+  .feds-navItem--active > .feds-navLink:only-child:hover,
+  .feds-navItem--active > .feds-navLink:only-child:focus {
+    color: var(--feds-color-link--light);
+  }
+
+  .feds-navItem--active > .feds-navLink:before {
+    position: absolute;
+    bottom: 0;
+    left: 0;
+    right: 0;
+    border-bottom: 2px solid #2c2c2c;
+    content: "";
+  }
+
+  .feds-navItem--active > .feds-navLink:before {
+    margin: 0 12px;
+  }
+
+  .feds-navItem--activeDeferred > .feds-navLink {
+    padding: 0;
+    width: 100%;
+    justify-content: center;
+  }
+
   .feds-topnav--overflowing .feds-navLink,
   .feds-topnav--overflowing .feds-navLink--hoverCaret,
   [dir = "rtl"] .feds-topnav--overflowing .feds-navLink--hoverCaret {
     padding: 0 8px;
   }
 
-  .feds-navLink--hoverCaret {
-    position: static;
+  .feds-topnav--overflowing .feds-navItem--active > .feds-navLink:before {
+    margin: 0 8px;
+  }
+
+  .feds-topnav--overflowing .feds-navItem--activeDeferred > .feds-navLink {
+    padding: 0;
   }
 
   .feds-navLink--hoverCaret:hover,
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -22,6 +22,9 @@ import {
   isDesktop,
   isTangentToViewport,
   setCurtainState,
+  hasActiveLink,
+  setActiveLink,
+  getActiveLink,
   selectors,
   logErrorFor,
   lanaLog,
@@ -272,6 +275,7 @@ class Gnav {
         && this.elements.topnav.scrollWidth > document.body.clientWidth;
 
       this.elements.topnav.classList.toggle(selectors.overflowingTopNav.slice(1), isOverflowing);
+      window.dispatchEvent(new CustomEvent('feds:navOverflow', { detail: { isOverflowing } }));
     };
 
     toggleContraction();
@@ -533,6 +537,15 @@ class Gnav {
       this.elements.mainNav.appendChild(this.decorateMainNavItem(item, index));
     }
 
+    if (!hasActiveLink()) {
+      const sections = this.elements.mainNav.querySelectorAll('.feds-navItem--section');
+
+      if (sections.length === 1) {
+        sections[0].classList.add(selectors.activeNavItem.slice(1));
+        setActiveLink(true);
+      }
+    }
+
     return this.elements.mainNav;
   };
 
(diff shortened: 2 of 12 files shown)


mokimo (approved):
Cool 👍

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -531,6 +535,15 @@ class Gnav {
       this.elements.mainNav.appendChild(this.decorateMainNavItem(item, index));
     }
 
+    if (!hasActiveLink()) {
+      const sections = this.elements.mainNav.querySelectorAll('.feds-navItem--section');
+
+      if (sections.length === 1) {
```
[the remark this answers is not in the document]
overmyheadandbody: If a match is not found in any of the navigation links and just a single cloud is used, then it's a safe assumption that the cloud is _sort of_ active. This is also mentioned in the Wiki. Even if it's not, we're still giving our users some sort of context as to what cloud they're navigating through. If we have multiple clouds, then we really need to check and there is a risk of not having any active element.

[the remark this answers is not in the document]
mokimo: Ah yes, that's this use-case - yes, we've talked about it a couple of times, thanks for clarifying! 👍

robert-bogos (approved):

libs/blocks/global-navigation/base.css
```diff
@@ -115,14 +115,51 @@
     padding: 0 12px;
   }
 
+  .feds-navItem--active > .feds-navLink {
+    position: relative;
+    font-weight: 700;
```
narcis-radu, line 120: In the AEM implementation we're highlighting the active element with **bold**, but I see this does not get applied in Milo.

AEM:
Screenshot 2023-12-06 at 14 10 32

Milo:
Screenshot 2023-12-06 at 14 11 38

libs/blocks/global-navigation/global-navigation.js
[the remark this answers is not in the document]
narcis-radu: @SilviuLCF - please make sure this use-case is properly tested, we'll need to have multiple navigations to test around.

libs/blocks/global-navigation/utilities/menu/menu.js
```diff
@@ -311,6 +308,26 @@ const decorateMenu = (config) => logErrorFor(async () => {
...
+    if (getActiveLink(menuTemplate) instanceof HTMLElement) {
+      // Special handling on desktop, as content is loaded async;
+      // bolding the item text would normally push the content
+      // to the right, potentially causing CLS
+      const resetActiveState = () => {
+        config.template.style.width = '';
+        config.template.classList.remove(selectors.deferredActiveNavItem.slice(1));
+      };
+
+      if (isDesktop.matches) {
```
narcis-radu, line 321: Noticed your message 
> It gets removed once the viewport switches to mobile or the content of the navigation is overflowing

but I think we should keep the bold item in mobile to be consistent with existing implementation.

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -162,6 +166,30 @@ export function setActiveDropdown(elem) {
...
+      if (hasActiveLink() || !(area instanceof HTMLElement)) return null;
+      const { origin, pathname } = window.location;
+      let activeLink;
+
+      [`${origin}${pathname}`, pathname].forEach((path) => {
+        if (activeLink) return;
+        activeLink = area.querySelector(`a[href = '${path}'], a[href ^= '${path}?'], a[href ^= '${path}#']`);
+      });
+
+      if (!activeLink) return null;
```
narcis-radu, line 185: I don't think there's a risk here but still asking - would we need to set `activeLinkFound` as false for this scenario?

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -162,6 +166,30 @@ export function setActiveDropdown(elem) {
...
+    () => activeLinkFound,
+    (val) => { activeLinkFound = !!val; },
+    (area) => {
+      if (hasActiveLink() || !(area instanceof HTMLElement)) return null;
+      const { origin, pathname } = window.location;
+      let activeLink;
+
+      [`${origin}${pathname}`, pathname].forEach((path) => {
+        if (activeLink) return;
+        activeLink = area.querySelector(`a[href = '${path}'], a[href ^= '${path}?'], a[href ^= '${path}#']`);
```
narcis-radu, line 182: cool 😬

libs/blocks/global-navigation/base.css
narcis-radu: I'm fine with keeping things as is as long as everything has been discussed. Maybe we should add a note in the JIRA ticket for future reference.

SilviuLCF (approved):
Testing details https://jira.corp.adobe.com/browse/MWPW-131197

PR #1599, opened 2023-12-04
Mwpw 139503: Merch Card link & block analytics

Axelcureno:
Adds analytics daa-ll to all links inside merch card and daa-lh to the merch card block.

This PR is trivial for Milo Commerce release 3

284465877-19b4b08e-fd96-423d-bc0b-1ba07a94c5e3

Resolves: [MWPW-139503](https://jira.corp.adobe.com/browse/MWPW-139503)

Test URLs:

Before: https://main--milo--adobecom.hlx.page/drafts/axel/document2
After: https://mwpw-139503--milo--axelcureno.hlx.page/drafts/axel/document2


libs/blocks/merch-card/merch-card.js
```diff
@@ -102,6 +102,14 @@ function addMerchCardGridsIfMissing(section) {
   }
 }
 
+const decorateMerchCardLinkAnalytics = (el) => {
+  [...el.querySelectorAll('a')].forEach((link, index) => {
+    const heading = el.querySelector('h3');
+    const analyticsString = `${processTrackingLabels(link.textContent)}-${index + 1}|${heading.textContent}`;
+    link.setAttribute('daa-ll', analyticsString);
```
overmyheadandbody, line 109: Is it certain that links will always contain this `h3`? Otherwise, a check might be beneficial, since `heading.textContent` is referenced

overmyheadandbody:
Also, please add the proper PR labels - https://github.com/adobecom/milo/wiki/PR-Labels

libs/blocks/merch-card/merch-card.js
```diff
@@ -102,6 +102,15 @@ function addMerchCardGridsIfMissing(section) {
   }
 }
 
+const decorateMerchCardLinkAnalytics = (el) => {
+  [...el.querySelectorAll('a')].forEach((link, index) => {
+    const heading = el.querySelector('h3');
+    const headingText = heading.textContent ? heading.textContent : '';
```
overmyheadandbody, line 108: Optional chaining might also help, in case a heading is not defined, just so you don't get the 'can't read properties of undefined' error

honstar (approved):

libs/blocks/merch-card/merch-card.js
Axelcureno: Fixed

overmyheadandbody (approved):
L108 from _merch-card.js_ might exceed 100 line length, but I think ESLint tolerates it for strings. Might be worth checking, though.

libs/blocks/merch-card/merch-card.js
```diff
@@ -102,6 +102,14 @@ function addMerchCardGridsIfMissing(section) {
   }
 }
 
+const decorateMerchCardLinkAnalytics = (el) => {
+  [...el.querySelectorAll('a')].forEach((link, index) => {
+    const heading = el.querySelector('h3');
+    const analyticsString = `${processTrackingLabels(link.textContent)}-${index + 1}|${heading?.textContent || ''}`;
```
yesil, line 108: @Axelcureno  I found an NPE with this line an fixed it here:https://github.com/adobecom/milo/blob/commerce-v3/libs/blocks/merch-card/merch-card.js#L190C1-L197C1
 can you review if the fix makes sense?

Axelcureno: Yes, it does. I've applied the fix.

VKniaz (approved):

yesil (approved):

PR #1604, opened 2023-12-04
MWPW-140042: Merch Card V3

yesil:
Removed legacy merch-card
No dependency on CaaS
Accessible via arrow keys

card-layout & styles contributed by @Axelcureno & @VKniaz 
`merch-offer-select` feature is contributed by @3ch023.

Stock checkbox for plans card depends on https://github.com/adobecom/milo/pull/1600

Resolves: [MWPW-140042](https://jira.corp.adobe.com/browse/MWPW-140042)

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/creativecloud/file-types/image/comparison/jpeg-vs-pdf?martech=off
- After: https://main--cc--adobecom.hlx.page/creativecloud/file-types/image/comparison/jpeg-vs-pdf?milolibs=mwpw-140042--milo--yesil&martech=off

- Before: https://main--bacom--adobecom.hlx.page/products/adobe-experience-cloud-products?martech=off
- After: https://main--bacom--adobecom.hlx.page/products/adobe-experience-cloud-products?milolibs=mwpw-140042--milo--yesil&martech=off

- Before: https://main--cc--adobecom.hlx.page/products/special-offers?martech=off
- After: https://main--cc--adobecom.hlx.page/products/special-offers?milolibs=mwpw-140042--milo--yesil&martech=off

- Before: https://main--dc--adobecom.hlx.page/acrobat/how-to/share-pdf-online?martech=off
- After: https://main--dc--adobecom.hlx.page/acrobat/how-to/share-pdf-online?milolibs=mwpw-140042--milo--yesil&martech=off

- Before: https://main--milo--adobecom.hlx.page/fragments/docs/commerce/plans-cards
- After: https://mwpw-140042--milo--yesil.hlx.page/fragments/docs/commerce/plans-cards

PR #1606, opened 2023-12-04
MWPW-140076 Update merch-card to match new analytics taxonomy

vgoodric:
Fix for merch card analytics PR was developed during code freeze at the same time as the analytics 2.1 PR.  Block update to:
- use attributes.js instead of analytics.js
- use processTrackingLabels on header text, not just link
- send config (will be used in upcoming translation feature.

Also changed | to -- in interlinks

Resolves: [MWPW-140076](https://jira.corp.adobe.com/browse/MWPW-140076)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/vgoodrich/fragments/analytics/merch-card?martech=off
- After: https://analytics-merch-140076--milo--adobecom.hlx.page/drafts/vgoodrich/fragments/analytics/merch-card?martech=off


Axelcureno (approved):

libs/blocks/merch-card/merch-card.js
```diff
@@ -104,10 +104,11 @@ function addMerchCardGridsIfMissing(section) {
...
 const decorateMerchCardLinkAnalytics = (el) => {
   [...el.querySelectorAll('a')].forEach((link, index) => {
+    const config = getConfig();
     const heading = el.querySelector('h3');
-    const linkText = `${processTrackingLabels(link.textContent)}-${index + 1}`;
-    const headingText = heading ? `${heading.textContent}` : '';
-    const analyticsString = heading ? `${linkText}|${headingText}` : linkText;
+    const linkText = `${processTrackingLabels(link.textContent, config)}-${index + 1}`;
+    const headingText = heading ? `${processTrackingLabels(heading.textContent, config)}` : '';
+    const analyticsString = heading ? `${linkText}--${headingText}` : linkText;
```
overmyheadandbody, line 111: Judging by the unit test, it seems the `--` is intentional? I think one of the purposes of the recent analytics PR was to remove some special characters, as well as limiting the final string length. With this, you'll be defining something like `Link text-1--Heading text`. But if you apply the `processTrackingLabels` method to it, you'd be left with just `Link text 1 Heading `. Notice the string length is limited to 20 characters and the special characters are removed. Are you certain this is what your analysts want? Or should the `processTrackingLabels` be used for consistency?

vgoodric: @overmyheadandbody Great questions, but to clarify, this PR is not about what analytics format we should use.  That has been debated and agreed to and already merged yesterday.  This is just about getting 2 blocks to conform to the new system.  That being said, I'm more than happy to answer your questions.

This is the link to the full documentation: https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=marketingtech&title=Analytics+in+Milo

And this is a 1 line summary of the syntax:
link name(20)-#--Last header before link(20)|s#|b#|blockname(15)|pzn experience name or "nopzn"(15)|pzn manifest filename or "nopzn"(20)

The text before the first | character is on the link itself: link name(20)-#--Last header before link(20)

processTrackingLabels won't remove the dashes because the link text and header are sent through separately, and the 2 outputs are then assembled with the proper dashes.  But yes, that function cleans the string and limits to 20 characters to enforce consistency. Also, I have another PR that will soon update the function to also check for EN equivalents of the string. https://github.com/adobecom/milo/pull/1609

The removal of special characters is something the gnav was already doing to their analytics and it reduces the chance of sending characters that AA can't process.

honstar (approved):

yesil:
@vgoodric possible to wait https://github.com/adobecom/milo/pull/1604 for merging this one?
We have limited QA and we are almost done,  planning to merge later today.

PR #1608, opened 2023-12-05
MWPW-140080: Merch card grid with fragments

Axelcureno:
Fixes Merch cards grids with nested fragments.

Currently merch cards are inheriting the style display: block from Milo styles, whereas we use display: grid for our grids that are children of main. This change adds the use case when the grid is a children of a fragment.

Resolves: [MWPW-140080](https://jira.corp.adobe.com/browse/MWPW-140080)

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/drafts/axel/segment-page
- After: https://main--cc--adobecom.hlx.page/drafts/axel/segment-page?milolibs=MWPW-140080--milo--axelcureno

### DIFF
--- a/libs/blocks/merch-card/merch-card.css
+++ b/libs/blocks/merch-card/merch-card.css
@@ -1,5 +1,7 @@
 main > div[class*="-merch-card"],
-main > div[class*="-merch-cards"] {
+main > div[class*="-merch-cards"],
+.fragment > div[class*="-merch-card"],
+.fragment > div[class*="-merch-cards"] {
   display: grid;
 }
 


yesil:
@Axelcureno this page works fine with https://main--cc--adobecom.hlx.page/drafts/axel/segment-page?milolibs=MWPW-140042--milo--yesil

In this PR it is still missing the segment token in the section. 
I recommend to proceed with the other PR.

libs/blocks/merch-card/merch-card.css
```diff
@@ -1,5 +1,7 @@
 main > div[class*="-merch-card"],
-main > div[class*="-merch-cards"] {
+main > div[class*="-merch-cards"],
+.fragment > div[class*="-merch-card"],
+.fragment > div[class*="-merch-cards"] {
```
overmyheadandbody, line 4: Is there actually a need for both? `[class*="-merch-card"]` matches both `.something-merch-card` and `.something-merch-cards`, since both _contain_ `merch-card`. Am I missing something? This would make sense if instead of _contains_ (`*=`) you would use _ends with_ (`$=`), otherwise it could be trimmed a bit.

PR #1610, opened 2023-12-05
MWPW-139177: TWP modal over Milo (Home) page is not auto re-sized

mirafedas:
This change adds the automatic height adjustment for modals if the two following conditions are met:
1. The modal must have both the `commerce-frame` and `height-fit-content` classes.
2. The modal must contain an iFrame that sends a postMessage request with the content height of the page inside the iFrame.

On the desktop screen size:
If the content height is lower than the modal window height, the modal window height will be reduced to eliminate any blank space at the bottom of the page. 
Conversely, if the content height exceeds the modal window height, the modal window will expand to fill the entire available viewport height, with 20px vertical margins.

On the tablet and mobile screen sizes:
The height will stay the same, its value will be 100%.

The information about this class is added in the authoring documentation here:
https://main--milo--adobecom.hlx.page/docs/authoring/create-modal

Before:
https://github.com/adobecom/milo/assets/30750556/a59babe9-0de7-4ed8-b846-04d7d09d5335

After:
https://github.com/adobecom/milo/assets/30750556/1338f5d5-e04d-4535-9ec1-e18c7c0a2d7c

Also fixed the issue (not mentioned in the Jira) with the two cards layout on the smaller desktop screen sizes by setting a minimal modal width to 1000px for the desktop.
Before:
Screenshot 2023-11-24 at 17 47 35
After:
Screenshot 2023-11-24 at 17 49 11

**Test URLs:**
- Before: https://main--homepage--adobecom.hlx.live/homepage/index-loggedout?martech=off
- After:  https://mwpw-139177-twp-modal-height-adjustment--milo--mirafedas.hlx.page/drafts/mirafedas/twp

Note for testers:
In the ModHeaders please add in the CSP response modifiers the header 'frame-ancestors' with the value '*', otherwise on my test page there will be a CSP violation and the content inside the modal will not be displayed (and the height will not be adjusted).


libs/blocks/modal/modal.js
```diff
@@ -110,6 +110,29 @@ export async function sendViewportDimensionsOnRequest(messageInfo) {
   }
 }
 
+/** For the modal height adjustment to work the following conditions must be met:
+ * 1. The modal must have classes 'commerce-frame height-fit-content';
+ * 2. The iframe inside must send a postMessage with the contentHeight (a number of px or '100%);
+ */
+function adjustModalHeight(contentHeight) {
+  if (!contentHeight || !window.location.hash) return;
+  const dialog = document.querySelector(window.location.hash);
```
vladen, line 119: Passing `window.location.hash` to the `querySelector` method as is, seems to be  very fragile to me.
For instance, Milo Commerce code uses `window.location.hash` to store some state in format of a query string (e.g. `#a=1&b=2`).

vladen (changes requested):

libs/blocks/modal/modal.js
mirafedas: Thanks, changed it to not operate the window.location.hash

vladen (approved):

honstar (approved):

Axelcureno (approved):

PR #1624, opened 2023-12-07
MWPW-139379 Improve Tab Scrolling for Mobile

Brandon32:
* Improve mobile experience for tab scrolling by adding scroll indicator buttons.

Resolves: [MWPW-139379](https://jira.corp.adobe.com/browse/MWPW-139379)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/bmarshal/tabs-scrolling?martech=off
- After: https://bmarshal-tabs-scrolling--milo--adobecom.hlx.page/drafts/bmarshal/tabs-scrolling?martech=off
- Before: https://main--milo--adobecom.hlx.page/drafts/bmarshal/tabs-scrolling-dark?martech=off
- After: https://bmarshal-tabs-scrolling--milo--adobecom.hlx.page/drafts/bmarshal/tabs-scrolling-dark?martech=off

**BACOM URLs:**
- Before: https://main--bacom--adobecom.hlx.page/?martech=off
- After: https://main--bacom--adobecom.hlx.page/?martech=off&milolibs=bmarshal-tabs-scrolling

### DIFF
--- a/libs/blocks/tabs/tabs.css
+++ b/libs/blocks/tabs/tabs.css
@@ -59,6 +59,7 @@
   box-shadow: 0 -1px 0 inset var(--tabs-border-color);
   display: flex;
   z-index: 2;
+  margin: 0 32px;
 
   /* ScrollProps - If content exceeds height of container, overflow! */
   overflow: auto;
@@ -81,12 +82,11 @@
 }
 
 .tabs.center .tab-headline {
-    text-align: center;
+  text-align: center;
 }
 
 .tabs div[role="tablist"] .tab-list-container {
   display: flex;
-  padding: 0 8.3%;
   width: var(--grid-container-width);
   margin: 0 auto;
   box-sizing: content-box;
@@ -101,7 +101,6 @@
 
 .tabs.center div[role="tablist"] .tab-list-container {
   width: auto;
-  margin: 0;
 }
 
 .tabs .tab-content {
@@ -264,6 +263,61 @@
   height: 24px;
 }
 
+.tabs .tab-paddles {
+  position: relative;
+  z-index: 3;
+}
+
+.tabs .tab-paddles .paddle {
+  position: absolute;
+  width: 32px;
+  height: 48px;
+  bottom: 0;
+  background: #fff;
+  border: 0;
+}
+
+.tabs .tab-paddles .paddle:disabled {
+  box-shadow: 0 -1px 0 inset var(--tabs-border-color);
+  background: var(--tabs-list-bg-gradient);
+}
+
+.tabs .tab-paddles .paddle-left {
+  box-shadow: 0 -1px 0 inset var(--tabs-border-color), 4px 0 4px -2px #0000001a;
+  left: 0;
+}
+
+.tabs .tab-paddles .paddle-right {
+  box-shadow: 0 -1px 0 inset var(--tabs-border-color), -4px 0 4px -2px #0000001a;
+  right: 0;
+}
+
+.tabs .tab-paddles .paddle::after {
+  content: "";
+  position: absolute;
+  top: 15px;
+  left: 9px;
+  right: 17px;
+  opacity: 100;
+  transition: opacity 150ms ease;
+}
+
+.tabs .tab-paddles .paddle:disabled::after {
+  opacity: 0;
+}
+
+.tabs .tab-paddles .paddle-left::after {
+  background: url("data:image/svg+xml,%3Csvg width='8' height='14' viewBox='0 0 8 14' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath fill-rule='evenodd' clip-rule='evenodd' d='M6.5 0.750003C6.77979 0.750003 7.06006 0.856933 7.27344 1.0708C7.70069 1.49805 7.70069 2.19043 7.27344 2.61768L2.89063 7L7.27344 11.3823C7.70069 11.8096 7.70069 12.502 7.27344 12.9292C6.84668 13.3569 6.15332 13.3569 5.72656 12.9292L0.570308 7.77344C0.365228 7.56836 0.249998 7.29004 0.249998 7C0.249998 6.70996 0.365228 6.43164 0.570308 6.22656L5.72656 1.0708C5.93994 0.856933 6.22021 0.750493 6.5 0.750003Z' fill='%236D6D6D'/%3E%3C/svg%3E%0A") no-repeat center center;
+  width: 8px;
+  height: 14px;
+}
+
+.tabs .tab-paddles .paddle-right::after {
+  background: url("data:image/svg+xml,%3Csvg width='8' height='14' viewBox='0 0 8 14' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath fill-rule='evenodd' clip-rule='evenodd' d='M1.50001 13.25C1.22022 13.25 0.939945 13.1431 0.726565 12.9292C0.299315 12.5019 0.299315 11.8096 0.726565 11.3823L5.10938 7L0.726565 2.61768C0.299315 2.19043 0.299315 1.49805 0.726565 1.0708C1.15333 0.643068 1.84669 0.643068 2.27345 1.0708L7.4297 6.22656C7.63478 6.43164 7.75001 6.70996 7.75001 7C7.75001 7.29004 7.63478 7.56836 7.4297 7.77344L2.27345 12.9292C2.06007 13.1431 1.7798 13.2495 1.50001 13.25Z' fill='%236D6D6D'/%3E%3C/svg%3E%0A") no-repeat center center;
+  width: 8px;
+  height: 14px;
+}
+
 /* Section Metadata */
 .tabs-background-transparent .tabs,
 .tabs-background-transparent .tabs div[role="tablist"],
@@ -289,5 +343,13 @@
   .tabs.quiet div[role="tablist"] button {
     padding-top: 18px;
     padding-bottom: 18px;
+  }
+
+  .tabs .tab-paddles .paddle {
+    height: 62px;
+  }
+
+  .tabs .tab-paddles .paddle::after {
+    top: 24px;
   }
 }
--- a/libs/blocks/tabs/tabs.css
+++ b/libs/blocks/tabs/tabs.css
@@ -12,6 +12,7 @@
   --tabs-pill-bg-color-hover: #dadada;
   --tabs-pill-bg-color-active: #c6c6c6;
   --tabs-pill-text-color: #131313;
+  --tabs-paddle-bs-color: #0000001a;
 }
 
 :root .dark {
@@ -25,6 +26,7 @@
   --tabs-pill-bg-color-hover: #666;
   --tabs-pill-bg-color-active: #444;
   --tabs-pill-text-color: #fff;
+  --tabs-paddle-bs-color: #ffffff1a;
 }
 
 .tabs {
@@ -273,8 +275,15 @@
   width: 32px;
   height: 48px;
   bottom: 0;
-  background: #fff;
+  background: var(--tabs-active-bg-color);
   border: 0;
+  padding: 14px 0 12px;
+  display: flex;
+  align-items: center;
+}
+
+.tabs .tab-paddles .paddle:hover {
+  cursor: pointer;
 }
 
 .tabs .tab-paddles .paddle:disabled {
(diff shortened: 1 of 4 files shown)


libs/blocks/tabs/tabs.js
```diff
@@ -3,37 +3,39 @@
...
-    && rect.bottom <= (window.innerHeight || /* c8 ignore next */ docEl.clientHeight)
-    && rect.right <= (window.innerWidth || /* c8 ignore next */ docEl.clientWidth)
-  );
+const isTabInTabListView = (tab, partial = false) => {
+  const tabList = tab.closest('[role="tablist"]');
+  const tabRect = tab.getBoundingClientRect();
+  const tabListRect = tabList.getBoundingClientRect();
+  if (partial) {
+    return Math.round(tabRect.left) < Math.round(tabListRect.right)
+      && Math.round(tabRect.right) > Math.round(tabListRect.left);
```
JasonHowellSlavin, line 14: Saving the `Math.round(...)` method calls as variables might help readability, but I imagine naming this is going to be a pain lol.

libs/blocks/tabs/tabs.css
```diff
@@ -264,6 +263,61 @@
...
+  position: relative;
+  z-index: 3;
+}
+
+.tabs .tab-paddles .paddle {
+  position: absolute;
+  width: 32px;
+  height: 48px;
+  bottom: 0;
+  background: #fff;
```
ryanmparrish, line 276: Looking good. Few things i'm seeing. 
1) We should consider the dark theme. both `.dark .tabs` and `.tabs.dark` and style the overflow arrows accordingly. 
2) There should be a :hover state for these including a cursor: pointer, and some type of icon color hover indication. Probably just follow the tab button text hover colors.

libs/blocks/tabs/tabs.js
```diff
@@ -84,6 +86,68 @@ function initTabs(elm, config, rootElem) {
...
 
+function previousTab(item, i, arr) {
+  const next = arr[i + 1];
+  return (!isTabInTabListView(item) && next && isTabInTabListView(next, true));
+}
+
+function nextTab(item, i, arr) {
+  const previous = arr[i - 1];
+  return (!isTabInTabListView(item) && previous && isTabInTabListView(previous, true));
+}
```
JasonHowellSlavin, line 97: odd to have `nextTab()` contain and check for the const `previous` and vice versa with `previousTab()` and `next`

libs/blocks/tabs/tabs.css
```diff
@@ -264,6 +263,61 @@
...
+  opacity: 100;
+  transition: opacity 150ms ease;
+}
+
+.tabs .tab-paddles .paddle:disabled::after {
+  opacity: 0;
+}
+
+.tabs .tab-paddles .paddle-left::after {
+  background: url("data:image/svg+xml,%3Csvg width='8' height='14' viewBox='0 0 8 14' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath fill-rule='evenodd' clip-rule='evenodd' d='M6.5 0.750003C6.77979 0.750003 7.06006 0.856933 7.27344 1.0708C7.70069 1.49805 7.70069 2.19043 7.27344 2.61768L2.89063 7L7.27344 11.3823C7.70069 11.8096 7.70069 12.502 7.27344 12.9292C6.84668 13.3569 6.15332 13.3569 5.72656 12.9292L0.570308 7.77344C0.365228 7.56836 0.249998 7.29004 0.249998 7C0.249998 6.70996 0.365228 6.43164 0.570308 6.22656L5.72656 1.0708C5.93994 0.856933 6.22021 0.750493 6.5 0.750003Z' fill='%236D6D6D'/%3E%3C/svg%3E%0A") no-repeat center center;
```
ryanmparrish, line 310: If your just using a `` asset. Can we just reference the [UI icons from here](https://adobe.sharepoint.com/sites/adobecom/Shared%20Documents/Forms/AllItems.aspx?id=%2Fsites%2Fadobecom%2FShared%20Documents%2Fmilo%2Flibs%2Fimg%2Fui&viewid=d776cf70%2D9b7e%2D4ab7%2Db9da%2D9e0f8e03a7d2)? 
`url('./libs/img/ui/...)` - That way all block can reference the same icon and updated or changes can be made in authoring. Otherwise we may want to include an inline SVG so they can be styled: hover etc.

JasonHowellSlavin (commented):
A lot of good work. I had intermittent success with clicking the tabs myself though. The first click would often work, but the second would not.

libs/blocks/tabs/tabs.css
```diff
@@ -264,6 +263,61 @@
...
+}
+
+.tabs .tab-paddles .paddle::after {
+  content: "";
+  position: absolute;
+  top: 15px;
+  left: 9px;
+  right: 17px;
+  opacity: 100;
+  transition: opacity 150ms ease;
```
ryanmparrish, line 302: Nit: Can we hone in on the alignment a bit. These icons appear to sit higher, line-height wise then the tab-list labels. 
Screen Shot 2023-12-07 at 9 42 49 AM

libs/blocks/tabs/tabs.js
Brandon32: I thought about documenting this as the logic is a little hard to parse from the expression.
previousTab is looking for the first tab not in view with the next tab in view, example tab1 not in view with tab2 in view.
nextTab is looking for the last tab not in view with the previous tab in view, example tab2 not in view with tab1 in in view.

Brandon32:
> A lot of good work. I had intermittent success with clicking the tabs myself though. The first click would often work, but the second would not.

I found this really difficult to debug, what browser and screen size did you see this on?

libs/blocks/tabs/tabs.css
Brandon32: I believe this is expected from the designs.
image
image

libs/blocks/tabs/tabs.css
Brandon32: Updated this to an inline SVG.

libs/blocks/tabs/tabs.css
meganthecoder: Maybe we should check with design? I agree that it looks a little weird not centered

libs/blocks/tabs/tabs.js
```diff
@@ -168,6 +237,16 @@ const init = (block) => {
...
   }
 
+  // Tab Paddles
+  const tabPaddles = createTag('div', { class: 'tab-paddles', role: 'scrollbar' });
+  const paddleLeft = createTag('button', { class: 'paddle paddle-left', disabled: '' });
+  const paddleRight = createTag('button', { class: 'paddle paddle-right', disabled: '' });
+  paddleLeft.innerHTML = LEFT_PADDLE;
+  paddleRight.innerHTML = RIGHT_PADDLE;
+  tabPaddles.append(paddleLeft, paddleRight);
+  tabList.insertAdjacentElement('afterend', tabPaddles);
```
overmyheadandbody, line 247: Would `.after` not work here?

libs/blocks/tabs/tabs.css
Brandon32: It looks like the arrows should align with the label text.

libs/blocks/tabs/tabs.js
```diff
@@ -3,37 +3,42 @@
...
+const RIGHT_PADDLE = '<svg viewBox="0 0 8 14" fill="none" xmlns="http://www.w3.org/2000/svg"><path fill-rule="evenodd" clip-rule="evenodd" d="M1.50001 13.25C1.22022 13.25 0.939945 13.1431 0.726565 12.9292C0.299315 12.5019 0.299315 11.8096 0.726565 11.3823L5.10938 7L0.726565 2.61768C0.299315 2.19043 0.299315 1.49805 0.726565 1.0708C1.15333 0.643068 1.84669 0.643068 2.27345 1.0708L7.4297 6.22656C7.63478 6.43164 7.75001 6.70996 7.75001 7C7.75001 7.29004 7.63478 7.56836 7.4297 7.77344L2.27345 12.9292C2.06007 13.1431 1.7798 13.2495 1.50001 13.25Z" fill="currentColor"/></svg>';
+
+const isTabInTabListView = (tab) => {
+  const tabList = tab.closest('[role="tablist"]');
+  const tabRect = tab.getBoundingClientRect();
+  const tabListRect = tabList.getBoundingClientRect();
```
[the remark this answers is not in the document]
Brandon32: We can use `IntersectionObserver` to show and hide the tabs but we still need this for click scrolling.

libs/blocks/tabs/tabs.js
Brandon32: Refactored this section and now using consts.

libs/blocks/tabs/tabs.css
Brandon32: There is no official dark theme for tabs implemented yet but I used the existing dark tabs colors.

libs/blocks/tabs/tabs.js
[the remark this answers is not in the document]
overmyheadandbody: What I'm saying is that the `IntersectionObserver` can have uses other than intersection detection. Here's what I mean, this can be executed on this page, as it references one of the comments:
```
let getElementBounds = element => new Promise((resolve, reject) => {
    if (!(element instanceof HTMLElement)) reject();

    const observer = new window.IntersectionObserver((entries) => {
        entries.forEach((entry) => {
            resolve(entry.boundingClientRect);
        });

        observer.disconnect();
    });

    observer.observe(element);
});

let elem = document.querySelector('#discussion_r1423259142');
getElementBounds(elem)
    .then((bounds) => console.log('bounds are', bounds))
    .catch(() => console.log('Something went wrong'));
```

This can be used to get positioning data in an async manner in order to avoid reflow latencies. I'm not saying it should be used everywhere, but performance analysis could show that such an approach might be beneficial.

libs/blocks/tabs/tabs.css
Brandon32: Updated so arrows are aligned with text.

meganthecoder (approved):

libs/blocks/tabs/tabs.js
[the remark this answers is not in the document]
Brandon32: I understand we can invoke an observation on an element, however I don't want to introduce lag on an interaction. Also the current code take 0.1 ms to execute, I wouldn't want to add the complexity of converting the code to an async solution before we notice a performance issue.

libs/blocks/tabs/tabs.css
```diff
@@ -264,6 +265,58 @@
...
+
+.tabs .tab-paddles .paddle-right {
+  box-shadow: 0 -1px 0 inset var(--tabs-border-color), -4px 0 4px -2px var(--tabs-paddle-bs-color);
+  right: 0;
+}
+
+.tabs .tab-paddles .paddle svg {
+  position: absolute;
+  top: 15px;
+  left: 12px;
```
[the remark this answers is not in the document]
Brandon32: I couldn't find or create any examples with RTL but testing in my browser it seems to work fine.

libs/blocks/tabs/tabs.css
```diff
@@ -264,6 +265,59 @@
...
+  left: 0;
+}
+
+.tabs .tab-paddles .paddle-right {
+  box-shadow: 0 -1px 0 inset var(--tabs-border-color), -4px 0 4px -2px var(--tabs-paddle-bs-color);
+  right: 0;
+}
+
+.tabs .tab-paddles .paddle svg {
+  opacity: 100;
```
overmyheadandbody, line 305: I think this should be either `1` or `100%` - https://developer.mozilla.org/en-US/docs/Web/CSS/opacity#alpha-value

libs/blocks/tabs/tabs.js
```diff
@@ -4,36 +4,40 @@
...
-  );
+const LEFT_PADDLE = '<svg viewBox="0 0 8 14" fill="none" xmlns="http://www.w3.org/2000/svg"><path fill-rule="evenodd" clip-rule="evenodd" d="M6.5 0.750003C6.77979 0.750003 7.06006 0.856933 7.27344 1.0708C7.70069 1.49805 7.70069 2.19043 7.27344 2.61768L2.89063 7L7.27344 11.3823C7.70069 11.8096 7.70069 12.502 7.27344 12.9292C6.84668 13.3569 6.15332 13.3569 5.72656 12.9292L0.570308 7.77344C0.365228 7.56836 0.249998 7.29004 0.249998 7C0.249998 6.70996 0.365228 6.43164 0.570308 6.22656L5.72656 1.0708C5.93994 0.856933 6.22021 0.750493 6.5 0.750003Z" fill="currentColor"/></svg>';
+const RIGHT_PADDLE = '<svg viewBox="0 0 8 14" fill="none" xmlns="http://www.w3.org/2000/svg"><path fill-rule="evenodd" clip-rule="evenodd" d="M1.50001 13.25C1.22022 13.25 0.939945 13.1431 0.726565 12.9292C0.299315 12.5019 0.299315 11.8096 0.726565 11.3823L5.10938 7L0.726565 2.61768C0.299315 2.19043 0.299315 1.49805 0.726565 1.0708C1.15333 0.643068 1.84669 0.643068 2.27345 1.0708L7.4297 6.22656C7.63478 6.43164 7.75001 6.70996 7.75001 7C7.75001 7.29004 7.63478 7.56836 7.4297 7.77344L2.27345 12.9292C2.06007 13.1431 1.7798 13.2495 1.50001 13.25Z" fill="currentColor"/></svg>';
```
overmyheadandbody, line 8: Out of curiosity - are the two significantly different? Or could the left one be rotated 180 degrees to get the right one? Maybe you could skip specifying both

libs/blocks/tabs/tabs.js
```diff
@@ -4,36 +4,40 @@
...
+
+const isTabInTabListView = (tab) => {
+  const tabList = tab.closest('[role="tablist"]');
+  const tabRect = tab.getBoundingClientRect();
+  const tabListRect = tabList.getBoundingClientRect();
```
overmyheadandbody, line 13: Should some safety mechanisms be added? Is `tab` sure to exist and an element? Is it certain it will have a parent with an `[role="tablist"]` attribute? Otherwise the `getBoundingClientRect` calls would result in errors

test/blocks/tabs/tabs.test.js
```diff
@@ -35,4 +42,58 @@ describe('tabs', () => {
     await sendKeys({ press: 'Space' });
     expect(unSelectedBtn2.ariaSelected).to.equal('false');
   });
+
+  it('paddles are disabled when tabList is not scrollable', async () => {
+    setViewport({ width: DESKTOP_WIDTH, height: HEIGHT });
+    window.dispatchEvent(new Event('resize'));
+    await delay(200);
```
overmyheadandbody, line 49: Could some `useFakeTimers` variation be used instead of actually waiting for a set time? A lot of small delays can lead to long run times for unit tests

Brandon32: Unfortunately `useFakeTimers` doesn't work here.

libs/blocks/tabs/tabs.css
Brandon32: Updated

libs/blocks/tabs/tabs.js
Brandon32: No, the tabs are in a tablist and without a tablist scrolling doesn't make any sense. 
The tablist is core to the block and is always created from init. If for some reason there wasn't a tablist then the tabs themselves couldn't be initialized.

libs/blocks/tabs/tabs.css
```diff
@@ -59,6 +61,7 @@
   box-shadow: 0 -1px 0 inset var(--tabs-border-color);
   display: flex;
   z-index: 2;
+  margin: 0 32px;
```
ryanmparrish, line 64:
```suggestion
  margin: 0 var(--spacing-m);
```

libs/blocks/tabs/tabs.js
```diff
@@ -168,6 +243,16 @@ const init = (block) => {
     tabListItems[0].parentElement.remove();
   }
 
+  // Tab Paddles
+  const tabPaddles = createTag('div', { class: 'tab-paddles', role: 'scrollbar' });
+  const paddleLeft = createTag('button', { class: 'paddle paddle-left', disabled: '' });
```
ryanmparrish, line 248:
```suggestion
  const paddleLeft = createTag('button', { class: 'paddle paddle-left', disabled: '' }, LEFT_PADDLE);
```
You can likely simplify these and just insert the paddle element on create.

libs/blocks/tabs/tabs.css
Brandon32: Updated

libs/blocks/tabs/tabs.js
Brandon32: Updated to rotate 180

PR #1628, opened 2023-12-07
MWPW-138221 - Add .m4v support

elan-tbx:
* The Franklin bot (Slack uploaded videos) now creates .m4v files instead of .mp4 files. Milo needs to support both file extensions moving forward.

Resolves: [MWPW-138221](https://jira.corp.adobe.com/browse/MWPW-138221)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ebartholomew/document2?martech=off
- After: https://ebartholomew-m4v-support--milo--adobecom.hlx.page/drafts/ebartholomew/document2?martech=off

### DIFF
--- a/libs/blocks/video/video.js
+++ b/libs/blocks/video/video.js
@@ -6,7 +6,7 @@ const ROOT_MARGIN = 1000;
 const loadVideo = (a) => {
   const { pathname, hash } = a;
   let videoPath = `.${pathname}`;
-  if (pathname.match('media_.*.mp4')) {
+  if (pathname.match('media_.*.(?:mp4|m4v)')) {
     const { codeRoot } = getConfig();
     const root = codeRoot.endsWith('/')
       ? codeRoot
--- a/libs/utils/decorate.js
+++ b/libs/utils/decorate.js
@@ -94,13 +94,13 @@ export function handleFocalpoint(pic, child, removeChild) {
 
 export async function decorateBlockBg(block, node, { useHandleFocalpoint = false } = {}) {
   const childCount = node.childElementCount;
-  if (node.querySelector('img, video, a[href*=".mp4"]') || childCount > 1) {
+  if (node.querySelector('img, video, a[href*=".mp4"], a[href*=".m4v"]') || childCount > 1) {
     node.classList.add('background');
     const binaryVP = [['mobile-only'], ['tablet-only', 'desktop-only']];
     const allVP = [['mobile-only'], ['tablet-only'], ['desktop-only']];
     const viewports = childCount === 2 ? binaryVP : allVP;
     [...node.children].forEach((child, i) => {
-      const videoLink = child.querySelector('a[href*=".mp4"]');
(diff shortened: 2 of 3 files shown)


mokimo (approved):
Here's also a relevant conversation in the EDS Discord https://discord.com/channels/1131492224371277874/1144495462255185971/1164610632453210112 - it's less officially supported and more of a possibility. I guess we might run into other extensions too if those are uploaded

libs/blocks/video/video.js
```diff
@@ -6,7 +6,7 @@ const ROOT_MARGIN = 1000;
 const loadVideo = (a) => {
   const { pathname, hash } = a;
   let videoPath = `.${pathname}`;
-  if (pathname.match('media_.*.mp4')) {
+  if (pathname.match('media_.*.(?:mp4|m4v)')) {
```
npeltier, line 9: can't we enforce character `\.` rather than using wildcard '.' ?
you are not using matching group down here so i would suggest not to use `?:`

npeltier (commented):
shouldn't we have a util generating video selector? this will be a one line change in the next video format we support :)

libs/blocks/video/video.js
overmyheadandbody: I agree with @npeltier, I think we could use something like `media_.*\.(mp4|m4v)`.

Or, looking at Nicolas' other suggestion, we may be able to get the values of the `video` auto blocks and build a `.mp4|.m4v)` string out of them. We would then just have to update _utils.js_ once and then everything trickles down automatically.

rgclayton:
Checking to see if we still need this due to the recent changes in Slack bot video uploads.

libs/blocks/video/video.js
JackySun9: how about this: if (pathname.match(/^\/media_.*\.(mp4|m4v)$/))

rgclayton:
With Slack no longer supporting Franklin bot video uploads and Sharepoint not converting .m4v to a working URL I don't see an need for this PR any more. 

Any differing thoughts?

image

libs/blocks/video/video.js
npeltier: this matches /mediawhatevermp4 and /mediaxwhateverm4v

npeltier:
> With Slack no longer supporting `Franklin bot` video uploads and Sharepoint not converting .m4v to a working URL I don't see an need for this PR any more.
> 
> Any differing thoughts?
> 
> image

true! we could still though constrain regexps a bit more and eventually factorize queries or selectors in one place

libs/blocks/video/video.js
overmyheadandbody: Adding some visuals to this, @npeltier is right:

- initial implementation, `media_.*.(?:mp4|m4v)`
Screenshot 2023-12-14 at 15 36 16

- my suggestion from above, `media_.*\.(mp4|m4v)`
Screenshot 2023-12-14 at 15 39 52

- the latest suggestion is invalid, so it would just lead to an error

PR #1637, opened 2023-12-11
MWPW-139968: Catalog Marquee

Axelcureno:
- Introduces a Catalog Marquee variant which includes a list of product mnemonics and titles for catalog page.

Resolves: [MWPW-139968](https://jira.corp.adobe.com/browse/MWPW-139968)
Xd: https://xd.adobe.com/view/432e3b75-3e56-4ea2-bc4b-b5e13c9ee2cd-167d/screen/83d63683-6649-47d4-951d-444ef03bff12/specs/
Screenshot 2023-12-11 at 3 13 26 PM

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/products/catalog?martech=off
- After: https://main--cc--adobecom.hlx.page/products/catalog?milolibs=MWPW-139968--milo--axelcureno&martech=off

URL for testing:

- https://main--cc--adobecom.hlx.page/products/catalog?milolibs=MWPW-139968--milo--axelcureno&martech=off

### DIFF
--- a/libs/blocks/marquee/marquee.css
+++ b/libs/blocks/marquee/marquee.css
@@ -718,3 +718,37 @@
     max-width: var(--grid-container-width);
   }
 }
+
+/* catalog marquee */
+
+@media screen and (min-width: 800px) {
+  .marquee.catalog .foreground .text {
+    max-width: 800px;
+  }
+}
+
+.marquee.catalog .foreground .catalog-product-list {
+  display: inline-flex;
+  flex-wrap: wrap;
+  gap: var(--spacing-xs);
+  justify-content: center;
+  width: 100%;
+  text-align: center;
+  position: relative;
+  margin: 0 0 var(--spacing-s);
+}
+
+.marquee.catalog .foreground .catalog-product {
+  display: flex;
+  flex: none;
+  gap: var(--spacing-xxs);
+}
+
+.marquee.catalog .foreground .catalog-product-list .catalog-product strong {
+  font-size: var(--type-heading-xxs-size);
+  line-height: var(--type-heading-xs-lh);
+}
+
+.marquee.catalog .foreground .catalog-product-list .catalog-product picture img {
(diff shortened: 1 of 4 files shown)


npeltier:
shouldn't that CSS at least be in CC css? or could we somehow inherit that marquee in CC in a special catalog-marquee? i find it a bit cumbersome to add all that code to all pages using marquee (which is a fair number)

honstar:
> shouldn't that CSS at least be in CC css? or could we somehow inherit that marquee in CC in a special catalog-marquee? i find it a bit cumbersome to add all that code to all pages using marquee (which is a fair number)

While reviewing the PR, I agree with @npeltier : ideally, we should not add catalog specific marquee features to the generic `marquee` block in milo. How about starting a discussion and involve other consumers to see if there is a need outside of the catalog page and cc?

vladen:
IMO, marquee should know nothing about catalogs and any other blocks and I'd rather make catalog block responsible in enhancing any marquee related to it. Because catalog needs this enhancement and knows how to do it best.

vladen (changes requested):
IMO, marquee should know nothing about catalogs and any other blocks and I'd rather make catalog block responsible in enhancing any marquee related to it. Because catalog is the one who needs this enhancement and knows how to do it best.

yesil:
@Axelcureno we should move the css and js logic in separate files in the marquee folder and load them only if it is marquee with an include section.
There should be no reference to catalog, but it should look and be generic.

vladen:
> @Axelcureno we should move the css and js logic in separate files in the marquee folder and load them only if it is marquee with an include section. There should be no reference to catalog, but it should look and be generic.

I'd move new files to `catalog` folder, not `marquee`

yesil:
@vladen 

> I'd move new files to catalog folder, not marquee

The includes section is not necessarily specific to catalog. We can keep it generic.

I would call it `includes.js` and `includes.css`

PR #1643, opened 2023-12-12
MWPW-139507:Allows the CaaS configurator to a use caas from a developer's host

cmiqueo:
We want to remove different risk factors in our e2e development process. One is that we need a way to test locally running caas ui code within milo without changing the milo code to support localhost. 

By using the URLParam "`caas-host=localhost`" or "`caas-host=sunier`" the developer can use the localversion of the configurator to poing to their local version of CaaS

Resolves: [MWPW-139507](https://jira.corp.adobe.com/browse/MWPW-139507)

**Test URLs:**
_(For illustration purposes only)_
- Before: https://main--milo--adobecom.hlx.page/tools/caas
- After: https://main--milo--adobecom.hlx.page/tools/caas?caas-host=locahost


sheridansunier (approved):

auniverseaway (approved):

mokimo (approved):

libs/blocks/caas/utils.js
```diff
@@ -199,12 +199,22 @@ export const decodeCompressedString = async (txt) => {
...
 
 export const loadCaasFiles = async () => {
-  const version = new URL(document.location.href)?.searchParams?.get('caasver') || 'stable';
+  const searchParams = new URLSearchParams(document.location.search);
+  const version = searchParams?.get('caasver') || 'stable';
+  let cssFile = `https://www.adobe.com/special/chimera/caas-libs/${version}/app.css`;
+  let jsFile = `https://www.adobe.com/special/chimera/caas-libs/${version}/main.min.js`;
+
+  // for caas local development
+  const host = searchParams?.get('caas-host');
```
overmyheadandbody, line 208: Nit: one query param is `caasver` (no hyphen), the other one is `caas-host` (with hyphen). Could we use the same naming rule for both?

cmiqueo: Good point. Updated

overmyheadandbody (approved):

PR #1649, opened 2023-12-13
MWPW-138459 XLG Entitlement support

chrischrischris:
* Adds support for XLG entitlement data returned by martech
* Sets the user entitlements into `config.entitlements` function which returns a Promise so it can be await'd
*  entitlements are just a simple array with the entitlements listed.  e.g: `[ 'photoshop', 'lightroom' ]`
* XLG ids are mapped to the entitlement names in entitlements.js in the `ENTITLEMENT_MAP` object.
* Consumers can add additional entitlements via `config.entitlements`, which needs to be an object with `segmentId: entitlementName`

Resolves: [MWPW-138459](https://jira.corp.adobe.com/browse/MWPW-138459)

**Test URLs:**
- After: https://cpeyer-aep--milo--adobecom.hlx.page/


libs/features/personalization/entitlements.js
```diff
@@ -0,0 +1,17 @@
+export const ENTITLEMENT_MAP = {
+  '9565ef55-faad-430b-b661-596ba7a036c4': 'all-apps',
+  'c8d50cc2-491e-48df-a1b0-1509f0ca7323': 'photoshop',
```
yesil, line 3: Can we generate this list for all the available products already?

libs/features/personalization/entitlements.js
```diff
@@ -0,0 +1,17 @@
...
+};
```
mokimo, line 4: Should we leave some info here who's setting up those segments - or a wiki link?

Maybe more of a future discussion, but should the segment ID mappings be authorable? I can imagine there to be a lot of experiments/tests if a user is in segment X, button shows Y

libs/features/personalization/personalization.js
```diff
@@ -316,78 +304,6 @@ function parsePlaceholders(placeholders, config, selectedVariantName = '') {
   }
   return config;
 }
-
-const fetchEntitlements = async () => {
```
mokimo, line 320: Since xLG doesn't yet support all entitlements, are there prod use-cases that would warrant keeping this?

libs/martech/martech.js
```diff
@@ -97,49 +98,91 @@ const getDtmLib = (env) => ({
...
+    const handleEntitlements = (e) => {
+      window.removeEventListener(ALLOY_SEND_EVENT, handleEntitlements);
+
+      if (e.detail?.result?.destinations?.length) {
+        resolve(setEntitlements(e.detail.result.destinations));
+      } else {
+        resolve([]);
+      }
+    };
+    window.addEventListener(ALLOY_SEND_EVENT, handleEntitlements);
```
mokimo, line 117: Since there are multiple interact calls, are there also multiple alloy event calls or is it enough to just wait for the first one?

// @narcis-radu didn't you just have something similar with alloy/gnav tests on AEM?

libs/features/personalization/entitlements.js
chrischrischris: I'm still waiting for the official list, these were the test values I was given.  Will use a followup PR once we get the values.

libs/features/personalization/entitlements.js
chrischrischris: We considered using an excel sheet to map this, but decided against it due to performance implications when personalizing a page.  We need this info asap as it holds up LCP until we know what personalization needs to be done.

libs/martech/martech.js
chrischrischris: From my understanding, the first time this event fires that call will contain the XLG data.

libs/features/personalization/personalization.js
chrischrischris: I asked around and was told it wasn't being used

libs/martech/martech.js
```diff
@@ -97,49 +98,91 @@ const getDtmLib = (env) => ({
...
-export default async function init({ persEnabled = false, persManifests }) {
+const setupEntitlementCallback = () => {
+  const setEntitlements = async (destinations) => {
+    const { default: parseEntitlements } = await import('../features/personalization/entitlements.js');
+    return parseEntitlements(destinations);
+  };
+
+  const getEntitlements = () => new Promise((resolve) => {
+    const handleEntitlements = (e) => {
+      window.removeEventListener(ALLOY_SEND_EVENT, handleEntitlements);
```
overmyheadandbody, line 109: Would setting `{ once: true }` not be an option when adding the event listener, thus eliminating the need to remove it?

libs/features/personalization/entitlements.js
overmyheadandbody, line 3: So any and all segments for all Milo consumers would need to go in here, no matter how specific to a particular consumer they are? This does make sense for entitlements, but might need to be expanded for more specific segments, say 'user generated more than 15 Firefly images'.

libs/martech/martech.js
chrischrischris: Good point

libs/features/personalization/entitlements.js
chrischrischris: Added support for consumers to defined entitlements via `config.entitlements`

libs/features/personalization/entitlements.js
chrischrischris: Still code based but I added support for consumers to defined entitlements via `config.entitlements`

libs/martech/martech.js
chrischrischris: done

libs/features/personalization/personalization.js
narcis-radu: I have been asked about this feature recently, adding @suhjainadobe for visibility in case this will have an impact.

libs/martech/martech.js
narcis-radu: We should add some guardrails to ensure that's the case. We use a similar approach in FEDS and have to check against a parameter that has been used to perform the call.

libs/features/personalization/entitlements.js
narcis-radu: We would end up with a lot of duplicate configuration in that case, right? I imagine pages like homepage, CC, events to use the same segments and they'd need to be added in every project. I know performance is a reason why we are not doing this through a spreadsheet, but we have to think of ways to solve this when multiple folks are using the same segments.

mokimo: It's consumer core milo entitlements enhanced with consumer defined entitlements https://github.com/adobecom/milo/pull/1649/commits/4cb04a7efc523a761a8869be85dd7061af004067#diff-ae750b175c471cc57e8d8b5e55bbfd271e4678b7bab2c8aa94587939c2d998d7R10

Just that either way a developer would need to add them

libs/martech/martech.js
```diff
@@ -97,49 +98,89 @@ const getDtmLib = (env) => ({
...
+
+  const getEntitlements = () => new Promise((resolve) => {
+    const handleEntitlements = (e) => {
+      if (e.detail?.result?.destinations?.length) {
+        resolve(setEntitlements(e.detail.result.destinations));
+      } else {
+        resolve([]);
+      }
+    };
+    window.addEventListener(ALLOY_SEND_EVENT, handleEntitlements, { once: true });
```
narcis-radu, line 115: A similar approach has been used in FEDS and noticed this event is fired with a small delay. Would it be possible to call `alloy('sendEvent')` directly and get the response? It might be faster, but I don't understand the exact use-case or limitations.

libs/martech/martech.js
```diff
@@ -97,49 +98,89 @@ const getDtmLib = (env) => ({
...
 });
 
-export default async function init({ persEnabled = false, persManifests }) {
+const setupEntitlementCallback = () => {
+  const setEntitlements = async (destinations) => {
+    const { default: parseEntitlements } = await import('../features/personalization/entitlements.js');
+    return parseEntitlements(destinations);
+  };
+
+  const getEntitlements = () => new Promise((resolve) => {
```
narcis-radu, line 107: I think there are use-cases where the Promise would fail, we need to handle those as well. I checked https://cpeyer-aep--milo--adobecom.hlx.page/ in Incognito and I'm stuck with a pending promise. We need to set a timeout or fail directly if `alloy` object is not available or other Martech logic is not running because of extension or hosts blockers

libs/martech/martech.js
```diff
@@ -97,49 +98,89 @@ const getDtmLib = (env) => ({
...
+  });
+
+  const { miloLibs, codeRoot } = getConfig();
+  loadLink(
+    `${miloLibs || codeRoot}/features/personalization/entitlements.js`,
+    { as: 'script', rel: 'modulepreload' },
+  );
+
+  window.milo ||= {};
+  window.milo.entitlements = getEntitlements();
```
narcis-radu, line 125: Would it make more sense to export this method and have consuming projects import it instead of using global objects? Using this system will open the door for making other properties available and my concern is that at some point a fallback value will not be provided and personalisation / usage might get affected.

chrischrischris: I considered that, but I didn't want consumers to have to import the whole martech file just to get the entitlement values.  I could possibly stuff it into utils.js config which would be better, but it just seemed simpler to use `milo.entitlements`.  Definitely open to discussion though.

> my concern is that at some point a fallback value will not be provided and personalisation / usage might get affected

Yes that's possible, but I'd argue that's a dev bug if that happens.

libs/martech/martech.js
chrischrischris: I don't know yet, I think it's worth revisiting once this is in prod and we've checked perf

libs/martech/martech.js
chrischrischris: Very good point.  I've added a timeout to it where it will just set it to an empty array in that case.  I think that's better than a rejection, but would like your opinion on it.  Timeout is currently set to 3s.

libs/features/personalization/entitlements.js
chrischrischris: @narcis-radu The idea is that most of the entitlement mappings would go into Milo, but any that are consumer specific (or need to be added asap) could also be added in the consuming project via `config.entitlements`.

libs/martech/martech.js
chrischrischris: Can you send me more info on this?

libs/martech/martech.js
narcis-radu: A default value should be enough, we've used the same approach in FEDS code. Thanks!

libs/martech/martech.js
narcis-radu: This pattern has been used by FEDS to expose multiple utilities, data attributes and we ended up with race conditions because there was no order in which scripts are loaded. This might not be the case with Milo, `window.milo` should always be available before consumers use it, but if `martech=off` is used, executing `window.milo.entitlements.then(console.log)` will throw an error.
I don't have a strong opinion on this, it's just based on previous experience with a similar pattern.

libs/utils/utils.js
```diff
@@ -176,6 +176,27 @@ export function getMetadata(name, doc = document) {
...
+    if (resolveVal !== undefined) {
+      entResolve(resolveVal);
+    }
+    return entPromise;
+  };
+})();
+
+function setupMiloObj(config) {
+  window.milo ||= {};
+  window.milo.deferredPromise = new Promise((resolve) => {
```
narcis-radu, line 195: is `window.milo.deferredPromise` used?

libs/utils/utils.js
```diff
@@ -176,6 +176,27 @@ export function getMetadata(name, doc = document) {
...
+  return (resolveVal) => {
+    if (resolveVal !== undefined) {
+      entResolve(resolveVal);
+    }
+    return entPromise;
+  };
+})();
+
+function setupMiloObj(config) {
+  window.milo ||= {};
```
narcis-radu, line 194: one issue I noticed on https://cpeyer-aep--milo--adobecom.hlx.page/ - `window.milo` will return an element with ID 'milo' instead of an Object with `deferredPromise`

libs/utils/utils.js
chrischrischris: yeah it's used by martech to wait to start some of their calls

libs/utils/utils.js
chrischrischris: Ugh - looks like if you have a title or anchor named `#milo` then `window.milo` will point to that.  This is an existing issue.  Since entitlements is now not using the window.milo object, and `window.milo.deferredPromise` still exists and works I think we can leave as is.

libs/martech/martech.js
chrischrischris: After offline conversation with @narcis-radu I've moved the entitlements into the config object as a function.
If entitlements are needed, you can `await config.entitlements()`

mokimo (approved):

overmyheadandbody (approved):

libs/martech/martech.js
```diff
@@ -97,49 +103,93 @@ const getDtmLib = (env) => ({
...
+  const getEntitlements = (resolve) => {
+    const handleEntitlements = (detail) => {
+      if (detail?.result?.destinations?.length) {
+        resolve(setEntitlements(detail.result.destinations));
+      } else {
+        resolve([]);
+      }
+    };
+    waitForEventOrTimeout(ALLOY_SEND_EVENT, ENTITLEMENT_TIMEOUT, [])
+      .then(handleEntitlements);
```
narcis-radu, line 121: If `waitForEventOrTimeout` gets rejected (https://github.com/adobecom/milo/pull/1649/files#diff-01f2571b06af2ef537675645367e6758c40e1d7baec6d783e6c6ac9512119a35R26) then `handleEntitlements` is not executed anymore and `getEntitlements` would not get resolved.

vgoodric (approved):

libs/martech/martech.js
chrischrischris: 👍 added catch

chrischrischris:
PR set to "trivial" as it will be announced by the PM once the martech side of things is working.

narcis-radu (approved):

PR #1658, opened 2023-12-14
MWPW-140345-Fix issues in media QR block variant

Ruchika4:
* Fixes issue of QR code being visible on mobile and tablet in media large
* Sets the dimension of QR code to 140 x 140 as requested 
* Also set the spacing between the QR code image and text to 24px
* Added unit tests to test it in different viewports

Resolves: [MWPW-140345](https://jira.corp.adobe.com/browse/MWPW-140345)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/fragments/consonant/media/media-qrcode?martech=off
- After: https://media-qr--milo--ruchika4.hlx.page/fragments/consonant/media/media-qrcode?martech=off


Ruchika4:
Worked with Kyungwha Lee, and got it verified on feature branch.

ryanmparrish (approved):

test/blocks/media/media.test.js
```diff
@@ -67,6 +68,33 @@ describe('media', () => {
       const appStoreCta = medias[5].querySelector('a.app-store');
       expect(appStoreCta).to.exist;
     });
+    it('desktop view has visibile qr code image and no google-play and app-store CTA', async () => {
+      await setViewport({ width: 1200, height: 100 });
+      const qrCodeImg = medias[5].querySelector('img.qr-code-img');
+      const qrCTA = medias[5].querySelectorAll('.qr-button-container');
+      expect(window.getComputedStyle(qrCodeImg).getPropertyValue('display')).not.to.be.equal('none');
+      qrCTA.forEach(async (cta) => {
```
overmyheadandbody, line 76: Why does this function need to be `async`? `forEach` and `async` can interact in unexpected ways, so I'd remove it if it doesn't serve any purpose. Same for L85 and L94.

Ruchika4: fixed

libs/blocks/media/media.css
```diff
@@ -341,8 +342,9 @@ div[class*="-up"] .media .foreground > .media-row {
 
   .media.qr-code img.qr-code-img {
```
overmyheadandbody, line 343: Is the added tag specificity actually needed? Would `.media.qr-code .qr-code-img` not achieve the same result?

overmyheadandbody (approved):

libs/blocks/media/media.css
Ruchika4: I believe it will. However as it was not required to update for the changes requested in the ticket, I didn't touched it. I can follow up with original contributor to check the reason and may be address this in a separate PR.

test/blocks/media/media.test.js
```diff
@@ -67,6 +68,33 @@ describe('media', () => {
       const appStoreCta = medias[5].querySelector('a.app-store');
       expect(appStoreCta).to.exist;
     });
+    it('desktop view has visibile qr code image and no google-play and app-store CTA', async () => {
+      await setViewport({ width: 1200, height: 100 });
+      const qrCodeImg = medias[5].querySelector('img.qr-code-img');
+      const qrCTA = medias[5].querySelectorAll('.qr-button-container');
+      expect(window.getComputedStyle(qrCodeImg).getPropertyValue('display')).not.to.be.equal('none');
```
JasonHowellSlavin, line 75: nit: do you need the `be` in the matchers here and below? Can it be:
`expect(window.getComputedStyle(qrCodeImg).getPropertyValue('display')).not.to.equal('none')`

JasonHowellSlavin (approved):

test/blocks/media/media.test.js
Ruchika4: Thanks for the review, @JasonHowellSlavin Both worked but I have updated to remove .be

PR #1671, opened 2023-12-18
MWPW-140623 - Auto-merge main to stage

elan-tbx:
* This PR adds a github actions workflow that automatically merges `main` into `stage` upon the merging of a PR into `main`

Resolves: [MWPW-140623](https://jira.corp.adobe.com/browse/MWPW-140623)

**Test URLs:**
N/A


mokimo (approved):
LGTM - we should set the proper context that the long term goal should be blocking main and only merging to stage, at which point this action will become obsolete.

But since we can't assess on how long this takes - this is pretty valuable 👍

chrischrischris (approved):

overmyheadandbody (approved):

PR #1663, opened 2023-12-15
MWPW-139915: Delivery of Spectrum Web Components

yesil:
This PR does not impact existing pages.
The goal is to allow commerce SPA apps like catalog, plans, TWP to consume Spectrum Web Components(SWC) in an on-demand fashion without loading all SWC components in bulk.

Milo discussion: https://github.com/orgs/adobecom/discussions/1618

Resolves: [MWPW-139915](https://jira.corp.adobe.com/browse/MWPW-139915)

**Test URLs:**
On CC:
https://mwpw-139915--cc--adobecom.hlx.live/drafts/ilyas/catalog?milolibs=mwpw-139915-catalog--milo--yesil&martech=off&georouting=off

PR #1676, opened 2023-12-20
MWPW-136866: merch-cards block

yesil:
This PR depends on https://github.com/adobecom/milo/pull/1663.

Merch-cards block is a merch card collection with sorting/filtering/deep-linking capabilities.
 

Resolves: [MWPW-136866](https://jira.corp.adobe.com/browse/MWPW-136866)

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.live/drafts/ilyas/catalog?martech=off&georouting=off
- After: https://mwpw-139915--cc--adobecom.hlx.live/drafts/ilyas/catalog?milolibs=mwpw-136866-stage--milo--yesil&martech=off&georouting=off


3ch023:
i see you created it against 'stage' branch, do we already have some process for it outlined? 
will the psi check run when we create a pr from stage to main?

libs/blocks/merch-cards/merch-cards.js
```diff
@@ -0,0 +1,257 @@
...
+      .textContent.split(',').map((s) => s.trim());
+
+    settingsEl.parentElement.remove();
+
+    if (filtered) {
+      attributes.filtered = filter;
+      preferences[filter] = parsePreferences(el.querySelectorAll('p'));
+    } else {
+      if (filter) {
+        attributes.filter = filter;
```
3ch023, line 169: find it a bit confusing that both `attributes.filtered` and `attributes.filter` are set to `filter`

`attributes.filtered` sound like a true/false flag but in reality its not?

3ch023: maybe we could have a short readme describing what is held in 'filtered' and what in 'filter'

VKniaz (approved):

libs/blocks/merch-cards/merch-cards.js
```diff
@@ -0,0 +1,257 @@
...
+            filters[key] = { order: index, size: cardSize };
+          }
+        });
+    });
+    merchCard.filters = filters;
+  });
+  return cardsRoot;
+}
+
+export default async function main(el) {
```
Axelcureno, line 126: Just curious, any reason why `main()` instead of `init`?

libs/blocks/merch-cards/merch-cards.js
```diff
@@ -0,0 +1,257 @@
...
+    return fail(el, 'Missing collection type');
+  }
+
+  const config = getConfig();
+  if (!config.queryIndexCardPath) {
+    return fail(el, 'Missing queryIndexCardPath config');
+  }
+
+  const { miloLibs } = getConfig();
+  loadStyle(`${miloLibs}/blocks/merch/merch.css`);
```
Axelcureno, line 137: Why this is needed? My understanding is, If there is a `merch` or `merch-card` block present in the page then Milo should pull its CSS automatically, isn't it?

Axelcureno (approved):
LGTM, just left 2 small questions for clarification.

libs/blocks/merch-cards/merch-cards.js
yesil: @Axelcureno, just oversight, changing it 👍

libs/blocks/merch-cards/merch-cards.js
yesil: this is to speed up the loading of all deps.
If we wait until cards are fetched from the query index, processed with the merch auto block inside, it will be delayed and the bandwith will be used for loading other less important things.

Since `merch-cards` block is special as it loads author blocks dynamically this is a perf optimisation I decided to add after many tests.

afmicka:
overall looks good to me.

PR #1677, opened 2023-12-20
MWPW-140627 use placeholder keys in analytics to keep them in English

vgoodric:
The recent addition of the martech metadata block allows analytics to stay in English when the page is localized into other languages.

The placeholder feature can be used to translate common phrases.  Currently, the only way to keep the analytics in English is to add every placeholder used to the martech attributes table.  This is unnecessary because the placeholder keys are always in English and prone to error.

This PR automates using the placeholder key in analytics instead of its translated text.

I also realized I could centralize the logic a little more and not import martech/attributes.js in as many places.

The example page has 3 strings being used in the analytics:
1. the header (from a manifest's placeholder tab)
2. hollow button (from the geo folder's placeholder spreadsheet)
3. solid button (just German text but there is martech metadata table with the translation)

If you inspect the buttons and look at the daa-ll attributes, you'll notice only item 3 is English.  The other 2 strings are in German.  In the after version, all 3 strings are in English.

Resolves: [MWPW-140627](https://jira.corp.adobe.com/browse/MWPW-140627)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/de/drafts/vgoodrich/fragments/placeholder-analytics/?martech=off
- After: https://placeholder-analytics--milo--adobecom.hlx.page/de/drafts/vgoodrich/fragments/placeholder-analytics/?martech=off

### DIFF
--- a/libs/features/placeholders.js
+++ b/libs/features/placeholders.js
@@ -16,8 +16,10 @@ const fetchPlaceholders = (config, sheet) => {
       const json = resp.ok ? await resp.json() : { data: [] };
       if (json.data.length === 0) { resolve({}); return; }
       const placeholders = {};
+      if (!config.analyticLocalization) config.analyticLocalization = {};
       json.data.forEach((item) => {
         placeholders[item.key] = item.value;
+        config.analyticLocalization[item.value.trim()] = item.key;
       });
       resolve(placeholders);
     });
--- a/libs/blocks/martech-metadata/martech-metadata.js
+++ b/libs/blocks/martech-metadata/martech-metadata.js
@@ -1,10 +1,9 @@
 import { getConfig } from '../../utils/utils.js';
-import { processTrackingLabels } from '../../martech/attributes.js';
 
-export const getMetadata = (el, config) => [...el.childNodes].reduce((rdx, row) => {
+export const getMetadata = (el) => [...el.childNodes].reduce((rdx, row) => {
   if (row.children?.length > 1) {
-    const key = processTrackingLabels(row.children[0].textContent, config);
-    const value = processTrackingLabels(row.children[1].textContent, config);
+    const key = row.children[0].textContent.trim();
+    const value = row.children[1].textContent.trim();
     if (key && value) rdx[key] = value;
   }
   return rdx;
@@ -16,7 +15,7 @@ export default function init(el) {
   if (ietf !== 'en-US') {
     config.analyticLocalization = {
       ...analyticLocalization,
-      ...getMetadata(el, config),
+      ...getMetadata(el),
     };
   }
   el.remove();
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -285,7 +285,7 @@ export function parseConfig(data) {
 }
 
 /* c8 ignore start */
-function parsePlaceholders(placeholders, config, selectedVariantName = '') {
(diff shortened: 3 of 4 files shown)


libs/features/placeholders.js
```diff
@@ -16,8 +16,10 @@ const fetchPlaceholders = (config, sheet) => {
       const json = resp.ok ? await resp.json() : { data: [] };
       if (json.data.length === 0) { resolve({}); return; }
       const placeholders = {};
+      if (!config.analyticLocalization) config.analyticLocalization = {};
```
auniverseaway, line 19: What is this property? Who approved this property going into the config object?

I fear we are making the config object a mess for no reason... and it's not just this one, but it's others as well. We should be shipping "really good defaults" and less optional stuff people must be aware of when they spin up a new project.

We are already littering our DOM with stuff almost no one cares about like `nopzn` and I fear this is more of that type of stuff.

@chrischrischris @rgclayton @narcis-radu

auniverseaway: I don't think "analytics localization" is anything placeholders should need to be aware of. This makes a tight coupling here and I think this is a bad idea. Most features should be portable and free of these types of dependencies.

auniverseaway (changes requested):
1. I don't think placeholders should need any knowledge of analytics to function.
2. I think this sets a bad precedent for the use of placeholders in places placeholders should not be used (blue cta buttons)
3. The new config object seems like a bad idea.

The moment a placeholder is a dependency to render a block (text in a cta), the entire placeholder file must be downloaded which puts more data in LCP. We should not be encouraging the use of placeholders in these scenarios.

Also, the whole point of localization is so that you can localize text inside the page without external resources being downloaded. That's why we went out of our way to ask for the martech-metadata block to be built in the first place... so we _didn't_ have to resort to placeholders for these types of ephemeral pieces of text.

If we move forward with this, I think we have different behaviors to account for...

Blue button 1 - Buy Now
Blue button 2 - {{buy-now}}

If I understand the PR correctly, the idea is that the key is used for `martech-metadata`. The question would be, what about the first one? Even if they're not on the same page, you're effectively getting a different outcome across pages. So you would end up putting "Buy Now" in the `value` for the 2nd blue button? Otherwise you're left with fetching the US placeholders so you can convert `buy-now` to the correct en-US value.

libs/features/placeholders.js
vgoodric: `analyticsLocalization` is just the same config property the martech metadata block is using.  So it is adding to the object that the block has already created.  Line 19 is there to prevent an error in case placeholders was used but the martech metadata block was not.

vgoodric:
@auniverseaway
1. I agree that placeholders should generally be avoided and my assumption is that they mostly will be.  I hoped this would be the least painful way to fill the gap IF the feature is used.
2. MEP manifests have a way to include placeholders.  These placeholders are NOT part of the permanent placeholder files, but live and die with the manifest.  This allows much faster turn aroud time for short term promos.
3. This is not adding an object to the config file.  It is adding to the same object `martech-metadata` blocks use
4. I think 2 different pages having a different value for "Buy now" is a worthwhile tradeoff to not download another placeholder file.

That being said, I could look at moving this code into the MEP code so that pages not using MEP don't have any additional code.  This would allow for keeping MEP placeholders in English without encouraging placeholders in "permanent" content.

PR #1684, opened 2023-12-22
MWPW-140726 adding FF support for clipboard copy

npeltier:
Resolves: [MWPW-140726](https://jira.corp.adobe.com/browse/MWPW-140726)

**Test URLs:**
- Before: https://milo.adobe.com/tools/ost
- After: https://mwpw-140726--milo--npeltier.hlx.page/tools/ost


npeltier:
closing as i don't think we need to support FF :p (incoming traffic coming exclusively from chrome thanks to sidekick)

PR #1690, opened 2024-01-03
Integrate Universal Navigation

narcis-radu:
Integrate the Universal Navigation via metadata property.
**Note:** This is part of the Universal Navigation integration initiative. Additional PRs will be open against the `gnav` branch until full functionality is implemented.
VPN is needed to test

Resolves: [MWPW-140227](https://jira.corp.adobe.com/browse/MWPW-140227)

**Test URLs:**
- Before: https://gnav--milo--adobecom.hlx.page/drafts/narcis/default?martech=off
- After: https://utilnav--milo--narcis-radu.hlx.page/drafts/narcis/default?martech=off

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.css
+++ b/libs/blocks/global-navigation/global-navigation.css
@@ -373,6 +373,12 @@ header.global-navigation {
 .feds-utilities {
   display: flex;
   align-items: center;
+  padding-right: 8px;
+}
+
+[dir = "rtl"] .feds-utilities {
+  padding-right: 0;
+  padding-left: 8px;
 }
 
 /* Profile styles - to be removed after IMS takes over the profile */
@@ -631,6 +637,14 @@ header.global-navigation {
 
   .feds-search-trigger:hover svg path {
     fill: var(--feds-color-link--hover--light);
+  }
+
+  .feds-utilities {
+    padding: 0;
+  }
+
+  [dir = "rtl"] .feds-utilities {
+    padding: 0;
   }
 
   /* Breadcrumbs */
--- a/libs/blocks/global-navigation/global-navigation.css
+++ b/libs/blocks/global-navigation/global-navigation.css
@@ -373,12 +373,7 @@ header.global-navigation {
 .feds-utilities {
   display: flex;
   align-items: center;
-  padding-right: 8px;
-}
-
-[dir = "rtl"] .feds-utilities {
-  padding-right: 0;
-  padding-left: 8px;
+  padding: 0 var(--feds-gutter);
 }
 
 /* Profile styles - to be removed after IMS takes over the profile */
@@ -640,10 +635,6 @@ header.global-navigation {
   }
 
   .feds-utilities {
-    padding: 0;
-  }
-
-  [dir = "rtl"] .feds-utilities {
     padding: 0;
   }
 
--- a/libs/blocks/global-navigation/global-navigation.css
+++ b/libs/blocks/global-navigation/global-navigation.css
@@ -370,6 +370,11 @@ header.global-navigation {
   content: '/';
 }
 
+.feds-utilities {
+  display: flex;
+  align-items: center;
+}
+
 /* Profile styles - to be removed after IMS takes over the profile */
 .feds-profile {
   display: flex;
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -4,6 +4,8 @@ import {
   getMetadata,
   loadIms,
   decorateLinks,
+  loadScript,
+  loadStyle,
 } from '../../utils/utils.js';
 import {
   toFragment,
@@ -163,13 +165,23 @@ const closeOnClickOutside = (e) => {
 class Gnav {
   constructor(body, el) {
     this.blocks = {
-      profile: {
-        rawElem: body.querySelector('.profile'),
-        decoratedElem: toFragment`<div class="feds-profile"></div>`,
-      },
       search: { config: { icon: CONFIG.icons.search } },
       breadcrumbs: { wrapper: '' },
     };
+
+    // TODO: metadata + project config
+    // TODO: enhance universal-nav metadata to allow other than "on"
+    this.useUniversalNav = getMetadata('universal-nav')?.toLowerCase() === 'on';
+    // TODO: remove this once implementation is done
+    this.useUniversalNav = true;
+    if (this.useUniversalNav) {
+      this.blocks.universalNav = toFragment`<div class="feds-utilities"></div>`;
+    } else {
+      this.blocks.profile = {
+        rawElem: body.querySelector('.profile'),
+        decoratedElem: toFragment`<div class="feds-profile"></div>`,
+      };
+    }
 
     this.el = el;
     this.body = body;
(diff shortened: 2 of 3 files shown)


libs/blocks/global-navigation/global-navigation.js
```diff
@@ -163,14 +207,23 @@ const closeOnClickOutside = (e) => {
...
-      profile: {
-        rawElem: body.querySelector('.profile'),
-        decoratedElem: toFragment`<div class="feds-profile"></div>`,
-      },
       search: { config: { icon: CONFIG.icons.search } },
       breadcrumbs: { wrapper: '' },
     };
 
+    // TODO: metadata + project config
+    this.universalNavComponents = getMetadata('universal-nav')?.toLowerCase()?.split(',').map((option) => option.trim())
```
mokimo, line 215: Do we need anything else in terms of the authoring for this? Or is a list of the components in the metadata (+ the on toggle) enough?

mokimo (approved):

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -373,6 +431,81 @@ class Gnav {
...
+      const agent = navigator.userAgent;
+      if (agent.includes('Mac')) return 'macOS';
+      if (agent.includes('Win')) return 'windows';
+      if (agent.includes('Linux')) return 'linux';
+      if (agent.includes('CrOS')) return 'chromeOS';
+      if (agent.includes('Android')) return 'android';
+      if (agent.includes('iPad')) return 'iPadOS';
+      if (agent.includes('iPhone')) return 'iOS';
+
+      return 'linux';
```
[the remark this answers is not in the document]
narcis-radu: We need one of these values as default and linux was the most general one; using `undefined` will not render the universal navigation, one of the predefined values is needed.

libs/blocks/global-navigation/global-navigation.js
narcis-radu: the list of components should be good enough, but I'm thinking about enhancing the components, e.g. providing custom fields in Profile. If this does not turn out to be a requirement I will remove the todo.

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -628,6 +639,14 @@ header.global-navigation {
     fill: var(--feds-color-link--hover--light);
   }
 
+  .feds-utilities {
+    padding: 0;
+  }
+
+  [dir = "rtl"] .feds-utilities {
+    padding: 0;
+  }
```
overmyheadandbody, line 648: You could simply group these together
```
.feds-utilities,
[dir = "rtl"] .feds-utilities {
  padding: 0;
}
```

LE: We should probably have some padding, though, as the focus state of the right side logo overlaps the utilities:
Screenshot 2024-01-05 at 12 17 51

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -370,6 +370,17 @@ header.global-navigation {
   content: '/';
 }
 
+.feds-utilities {
+  display: flex;
+  align-items: center;
+  padding-right: 8px;
```
overmyheadandbody, line 376: There is a `--feds-gutter` variable you could use instead

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -373,6 +431,81 @@ class Gnav {
...
+          platform: 'Web',
+          device: getDevice(),
+          os_version: navigator.platform,
+        },
+        event: { visitor_guid: visitorGuid },
+      },
+      children: getChildren(),
+    });
+
+    window.UniversalNav(getConfiguration());
```
overmyheadandbody, line 502: Are we absolutely positive that `window.UniversalNav` will exist? Should we add some optional chaining, just in case?

overmyheadandbody (commented):
Good job on this first PR, I know this required a ton of back and forth with a bunch of other teams!

Experience notes:
* you should remove the search from your experience, as it's no longer used on production. This will limit CLS issues, as the utility nav will push the search icon to the left. And without the search, does it even make sense to set asymmetric paddings for the utilities section? Could we not just have it at the `--feds-gutter` value for both right and left and then disregard RTL styles altogether?

Utilnav notes
* the App Launcher dropdown has a cursor pointer even when not hovering over an app. I know this is not anything we should handle, but wondering if it's a known implementation/issue;
* the help dropdown has a dark theme, while the App Launcher has a light one, is this expected?
* after hovering over the App Launcher icon for a couple of seconds, a border appears underneath a few elements, is this the desired behaviour?
* right and left keys don't move to the next/previous element in the App Launcher dropdown, should this be brought up with accessibility champions?
* pressing ESC while navigating through the App Launcher items should ideally bring back focus to the waffle icon, has this been brought up with them?
* on RTL, it is likely that the App Launcher dropdown goes off screen. Is RTL officially supported? Should it be hidden on RTL if not?
* inside the help dropdown, TAB never focuses on the first element. Up/Down arrows do focus it, it's likely a bug on their side that needs fixing
* the Profile dropdown has a lot of padding, is using a dark theme and has an odd-looking white border around it, is this the final implementation from their end?
* hover/active background colours between the Profile and the other utilities are quite different, is anyone pursuing some consistency for them?
Screenshot 2024-01-05 at 12 47 27
Screenshot 2024-01-05 at 12 47 42

libs/blocks/global-navigation/global-navigation.js
narcis-radu: yes

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -373,12 +373,7 @@ header.global-navigation {
 .feds-utilities {
   display: flex;
   align-items: center;
-  padding-right: 8px;
-}
-
-[dir = "rtl"] .feds-utilities {
-  padding-right: 0;
-  padding-left: 8px;
+  padding: var(--feds-gutter);
```
overmyheadandbody, line 376: Should we do `0 var(--feds-gutter)`? Is there a benefit to having up/down padding as well?

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -481,6 +483,11 @@ class Gnav {
       return children;
     };
 
+    const onAnalyticsEvent = (data) => {
+      // TODO: update based on requirements
+      console.log('analytics data', data);
```
overmyheadandbody, line 488: I had already pressed the comment button for the `console` message, but then I saw the code comment 😅 Since this is going to the `gnav` branch, it should be fine as-is for now

libs/blocks/global-navigation/global-navigation.js
overmyheadandbody: Even if the scripts don't load due to network issues? Do we stop execution early in that case?

overmyheadandbody (approved):

libs/blocks/global-navigation/global-navigation.css
narcis-radu: you're correct, we need `0 var(--feds-gutter)`

SilviuLCF:
PR can be merged to MAIN. with following comments: 

Analytics not available
Some UNAV bugs , listed in the https://jira.corp.adobe.com/browse/MWPW-140227 TEST comment

From integration point of view -- PASS

PR #1696, opened 2024-01-03
MWPW-137523 - Allow dark quiet marquees (redux)

elan-tbx:
Note: The previous PR for this feature was reverted because, in the process of cleaning up stylelint errors, a rule targeting rtl pages was removed and never re-inserted. That being said, in the spirit of keeping things as simple as possible, this PR no longer addresses those stylelint errors -- I think doing so should be its own separate PR in the future, to reduce noise around all included changes.

The following features are still included:

* Authors would like the option of having a dark-themed quiet marquee. Previously, a `.dark` class would still be applied to quiet marquees by default, but CSS would override this and force a light theme on quiet marquees at all times. This PR applies a bit more sanity to the process, only adding `.dark` when explicitly declared on quiet marquees, and allowing the normal CSS cascade to style it from there.

Resolves: [MWPW-137523](https://jira.corp.adobe.com/browse/MWPW-137523)

**Test URLs:**

Milo:
- Before: https://main--milo--adobecom.hlx.page/drafts/ebartholomew/quiet-marquee?martech=off
- After: https://ebartholomew-quiet-marquee-updates4--milo--adobecom.hlx.page/drafts/ebartholomew/quiet-marquee?martech=off

Bacom:
- Before: https://main--bacom--adobecom.hlx.page/products/experience-manager/sites/testing-optimization
- After: https://main--bacom--adobecom.hlx.page/products/experience-manager/sites/testing-optimization?milolibs=ebartholomew-quiet-marquee-updates4

RTL Homepage (both pages should look the same):
- Before: https://main--homepage--adobecom.hlx.page/ae_ar/homepage/index-loggedout
- After: https://main--homepage--adobecom.hlx.page/ae_ar/homepage/index-loggedout?milolibs=ebartholomew-quiet-marquee-updates4


JasonHowellSlavin (approved):
Looks good.

Do you foresee that there will be additions to the disallowed class list for adding the dark class? Might be worth instantiating the list into a variable if so.

overmyheadandbody (commented):
It's not immediately obvious to me how the code relates to the story requirement of setting a width to `800px`. Is the correct story referenced? Have the requirements changed?

elan-tbx:
@JasonHowellSlavin 

> Do you foresee that there will be additions to the disallowed class list for adding the dark class? Might be worth instantiating the list into a variable if so.

That's a good point, I'll make the change.

@overmyheadandbody 

> It's not immediately obvious to me how the code relates to the story requirement of setting a width to 800px. Is the correct story referenced? Have the requirements changed?

That was actually taken care of in #1512. I made that PR after the other updates were reverted, in order to at least get that feature merged. Sorry for the confusion... it's been a surprisingly bumpy road for this one.

libs/blocks/marquee/marquee.css
```diff
@@ -8,7 +8,7 @@
   min-height: 560px;
 }
 
-.marquee.small, .marquee.quiet, .marquee.inline, .marquee.large.compact {
+.marquee.small, .marquee.inline, .marquee.large.compact {
```
rgclayton, line 11: What is this update about? Doesn't seem to relate to the dark/light theming.

elan-tbx: This is related to the min-height adjustments requested in [the linked Jira ticket](https://jira.corp.adobe.com/browse/MWPW-137523) in the description

rgclayton: Got it. Thanks! That ticket is a mess... Makes quality control very difficult.

rgclayton: With that said, that makes the default `quiet` variant taller (560px) on any page it's used. I assume to get back to the smaller height (360px) the `small` variation will need to be added, correct?

You can see this in play on the bacom sample page provided.

rgclayton (changes requested):
Blocking for the time being. The default height change will impact consumers negatively. See Jira ticket for comments from Rikio.

rgclayton (approved):
Thanks for the changes Elan.

overmyheadandbody (approved):

ryanmparrish (approved):
Looks good. I'd suggest you update the desc to remove the mentions of 800 height etc as well as the Jira ticket title/desc just to clarify/narrow the ask. Probs not a big deal, but incase there is a ref to this PR in any context it'd be good to keep the deets sanitized.

Sartxi (approved):

PR #1698, opened 2024-01-04
[#1641] add parent option to createTag

npeltier:
- add options parameter to createTag signature, with only one parent option for now, as discussed in https://github.com/orgs/adobecom/discussions/1641,
- add unit test for create Tag

Resolves: https://github.com/orgs/adobecom/discussions/1641

After: https://discussion_1641_utils--milo--npeltier.hlx.page/?martech=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -239,7 +239,7 @@ export function isInTextNode(node) {
   return node.parentElement.firstChild.nodeType === Node.TEXT_NODE;
 }
 
-export function createTag(tag, attributes, html) {
+export function createTag(tag, attributes, html, options = {}) {
   const el = document.createElement(tag);
   if (html) {
     if (html instanceof HTMLElement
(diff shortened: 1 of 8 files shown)


mokimo (approved):

libs/utils/utils.js
```diff
@@ -257,6 +257,9 @@ export function createTag(tag, attributes, html) {
       el.setAttribute(key, val);
     });
   }
+  if (options.parent) {
+    options.parent.append(el);
+  }
```
overmyheadandbody, line 262: I don't remember if we discussed this in the initial PR, but couldn't we go with `options.parent?.append(el);` and save a couple of bits?

npeltier: :) we sure can

npeltier:
done @overmyheadandbody

overmyheadandbody (approved):

libs/utils/utils.js
hparra: I'm on the fence but coverage will still recognize the branch, so that's good.

hparra (approved):
One unrelated test is failing, but that's the way it is... Please remember to add a test page url(s) so that check passes.

PR #1699, opened 2024-01-04
Fix broken gnav test

mokimo:
### Description
Fixes a broken test for the old gnav, that was failing due to some changes in the core utils where `isSignedInUser` might not be a function for _tests_. In production this should not be an issue since we specifically wait for IMS to properly load, thus this is only a mocking/test issue.

**Test URLs:**
- After: https://fix-failing-gnav-test--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/test/blocks/gnav/gnav.test.js
+++ b/test/blocks/gnav/gnav.test.js
@@ -25,7 +25,7 @@ describe('Gnav', () => {
       client_id: 'milo',
       scope: 'gnav',
     };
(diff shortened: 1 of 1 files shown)


robert-bogos (approved):
Looks good 👍

overmyheadandbody (approved):
I think someone had opened a PR for this during the RCP, but since it's still an issue, we can merge this one

PR #1706, opened 2024-01-08
[#1641][commerce] using parent option in createTag, as well as innerHtml sometimes

npeltier:
**Test URLs:**
- After: https://discussion_1641_commerce--milo--npeltier.hlx.page/docs/library/kitchen-sink/merch-card?martech=off

### DIFF
--- a/libs/blocks/merch-card/merch-offer-select.js
+++ b/libs/blocks/merch-card/merch-offer-select.js
@@ -3,14 +3,12 @@ import { decorateButtons } from '../../utils/decorate.js';
 import '../../deps/merch-offer-select.js';
 
 function createDynamicSlots(el, bodySlot) {
-  const price = createTag('h5', { class: 'merch-card-price' });
-  price.append(createTag('span', { slot: 'price', is: 'inline-price' }));
-  bodySlot.append(price);
+  const price = createTag('h5', { class: 'merch-card-price' }, null, { parent: bodySlot });
+  createTag('span', { slot: 'price', is: 'inline-price' }, null, { parent: price });
 
-  const p = createTag('p', { class: 'action-area' });
-  p.append(createTag('a', { slot: 'cta', is: 'checkout-link' }));
   const footer = el.querySelector('div[slot="footer"]');
-  footer.append(p);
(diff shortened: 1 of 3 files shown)


libs/blocks/merch-card/merch-offer-select.js
```diff
@@ -3,14 +3,12 @@ import { decorateButtons } from '../../utils/decorate.js';
...
-  bodySlot.append(price);
+  const price = createTag('h5', { class: 'merch-card-price' }, null, { parent: bodySlot });
+  createTag('span', { slot: 'price', is: 'inline-price' }, null, { parent: price });
 
-  const p = createTag('p', { class: 'action-area' });
-  p.append(createTag('a', { slot: 'cta', is: 'checkout-link' }));
   const footer = el.querySelector('div[slot="footer"]');
-  footer.append(p);
+  const p = createTag('p', { class: 'action-area' }, null, { parent: footer });
+  createTag('a', { slot: 'cta', is: 'checkout-link' }, null, { parent: p });
```
overmyheadandbody, line 11: Not sure how the render flow is or if it matters for your use-case, but you're appending a child to the footer, then appending another grandchild, this could lead to some minor CLS or other such issues. As in: footer is initially just `<footer></footer>` and already is an element added to the DOM. L10 makes it `<footer><p class="action-area"></p></footer>`, adds it to the DOM, causes some paints/reflows depending on styling. Then L11 makes it `<footer><p class="action-area"><a slot="cta" is="checkout-link"></a></p></footer>`, adds it to the DOM, causes another round of paint/reflow. It might be worth building the final element before appending it to the DOM. Same comment might be valid for L6-7

npeltier: i thought about that but as no "long" action is done was not believing such CLS could happen, is it really the case?

yesil: +1 to @overmyheadandbody points.
We should attach the DOM elements at the end of the construction of their trees.

honstar:
@npeltier , any update on this PR?

PR #1710, opened 2024-01-09
MWPW-140191: Fixes Merch Card custom <hr> color, inline heading variant

Axelcureno:
This PR fixes the following: 
* Custom `<hr>` color or gradient when present.
* Cleans up EDS-generated DOM in Merch Card whenever an `<hr>` is present.
* Updates merch-card dependency
* Fixes `inline-heading` and `product` XS-heading positioning ([MWPW-140282](https://jira.corp.adobe.com/browse/MWPW-140282))
Screenshot 2024-01-08 at 6 37 40 PM

Resolves: [MWPW-140191](https://jira.corp.adobe.com/browse/MWPW-140191)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/docs/library/kitchen-sink/merch-card
- After: https://mwpw-140191--milo--adobecom.hlx.page/docs/library/kitchen-sink/merch-card

### DIFF
--- a/libs/blocks/merch-card/merch-card.js
+++ b/libs/blocks/merch-card/merch-card.js
@@ -219,6 +219,15 @@ const addStock = (merchCard, styles) => {
   }
 };
 
+const simplifyHrs = (el) => {
+  const hrs = el.querySelectorAll('hr');
+  hrs.forEach((hr) => {
+    if (hr.parentElement.tagName === 'P') {
+      hr.parentElement.replaceWith(hr);
+    }
+  });
+};
+
 const init = async (el) => {
   const styles = [...el.classList];
   const lastClass = styles[styles.length - 1];
@@ -347,6 +356,7 @@ const init = async (el) => {
     initOfferSelection(merchCard, offerSelection);
   }
   decorateBlockHrs(merchCard);
+  simplifyHrs(merchCard);
   if (merchCard.classList.contains('has-divider')) {
     merchCard.setAttribute('custom-hr', true);
   }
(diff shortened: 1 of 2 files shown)


libs/blocks/merch-card/merch-card.js
```diff
@@ -219,6 +219,15 @@ const addStock = (merchCard, styles) => {
   }
 };
 
+const simplifyHrs = (el) => {
+  const hrs = el.querySelectorAll('hr');
+  hrs.forEach((hr) => {
+    if (hr.parentElement.tagName === 'P') {
+      hr.parentElement.replaceWith(hr);
```
overmyheadandbody, line 226: Is there a use-case where the `hr` has some siblings? Say `<p><hr /> some text</p>`. This logic would leave just a `<hr />`, the text or other nodes would be removed.

yesil: do you think this is possible with EDS?
after hr, I would expect EDS to render a `p` for the next text node.

yesil (approved):

libs/blocks/merch-card/merch-card.js
overmyheadandbody: EDS would likely handle this case, yes. But I see a lot of other parsing going on before this method is called, it might be that some of that decoration adds this type of markup

Axelcureno: This use case is not possible because for an `hr` with custom color to render it should have its own line where the color or gradient is defined. If the hr (which in authoring starts with `---`) is in the same line with text, the decorator function would not recognize it as a valid hex color or gradient. It would also look very weird in Word for authors to have everything in the same line.
Example: 
Screenshot 2024-01-09 at 9 09 50 AM

JackySun9:
I tested at https://main--cc--adobecom.hlx.live/cc-shared/uar/plan-recommender/quiz-results?primary=lr_cct,pr_cct,id_cct&quizKey=cc-quiz-en-US&milolibs=mwpw-140191
it is working for UAR

JackySun9 (approved):

PR #1714, opened 2024-01-09
GNAV & Footer content centralisation

mokimo:
### Description
Adds the possibility to centralize globalnav content such as large menu's by supplying a link from the project for centralized milo content such as https://main--federal--adobecom.hlx.live/federal/globalnav/drafts/osahin/gnav - which can be reused across any milo consumer.

On (www.|business.|blog.|milo.)adobe.com this would simply be mapped to `/federal/globalnav/drafts/osahin/gnav` and does not require any extra TLS handshake.

**There is no extra setup needed for any milo consumer** - subdomains or domains without any extra setup will simply default fetching central content from `www.adobe.com/federal/...` 

### Solution design
The overarching idea of this is that `https://main--federal--adobecom.hlx.live` serves as a content hub for anything that is shared across consumers. In the (near) future 404s can be moved there, commerce related content might make sense in there and I'm sure a bunch of use cases will pop up.

### Features added
While we added support for a bunch of (globalnav) use cases, for now only strictly _shared_ global navigation content should live in the new content repository. The normal case will still be that each consumer can maintain their own global navigation or footer.

- Ability to fetch the global footer&navigation, menu's & base-breadcrumbs from a central location
- Fetches globalnav/footer placeholders from `/federal/globalnav/placeholders` instead of `milo.adobe.com/placeholders?sheet=feds` which leads to an unnecessary TLS handshake
- Ability to fragmentize large globalnav menus or footer sections

Resolves: [MWPW-138979](https://jira.corp.adobe.com/browse/MWPW-138979)
[MWPW-136659](https://jira.corp.adobe.com/browse/MWPW-136659)
Based on [this discussion](https://github.com/orgs/adobecom/discussions/1388)

**Test URLs:**
Further use-cases can be found in `SP>milo>ch_de>drafts>osahin>federalisation` & `SP>milo>drafts>osahin>federalisation`

Ensure to be on VPN when viewing those links. The `hlx.page` content is guarded
- Before: https://main--milo--adobecom.hlx.page/ch_de/drafts/osahin/federalisation/nav-and-footer
- After: https://mwpw-138979--milo--adobecom.hlx.page/ch_de/drafts/osahin/federalisation/nav-and-footer

- Before: https://main--milo--adobecom.hlx.page/drafts/osahin/federalisation/nav-and-footer
- After: https://mwpw-138979--milo--adobecom.hlx.page/drafts/osahin/federalisation/nav-and-footer

**Regression URLs**

**Acrobat:**
- After: https://www.stage.adobe.com/acrobat/online/sign-pdf.html?martech=off&milolibs=mwpw-138979--milo--adobecom

**BACOM:**
- After: https://business.stage.adobe.com/fr/customer-success-stories.html?martech=off&milolibs=mwpw-138979--milo--adobecom

**CC:**
- After: https://main--cc--adobecom.hlx.live/?martech=off&milolibs=mwpw-138979--milo--adobecom

**Homepage:**
- After: https://main--homepage--adobecom.hlx.page/homepage/index-loggedout?martech=off&milolibs=mwpw-138979--milo--adobecom

**Blog:**
- After: https://main--blog--adobecom.hlx.page/?milolibs=mwpw-138979--milo--adobecom

**CC**
- Before: https://main--cc--adobecom.hlx.page/drafts/jbrown/firefly/firefly?martech=off
- After: https://main--cc--adobecom.hlx.page/drafts/jbrown/firefly/firefly?martech=off&milolibs=mwpw-138979--milo--adobecom

### DIFF
--- a/libs/blocks/global-footer/global-footer.js
+++ b/libs/blocks/global-footer/global-footer.js
@@ -74,7 +74,7 @@ class Footer {
       .catch((e) => lanaLog({
         message: `Error fetching footer content ${url}`,
         e,
-        tags: 'errorType=error,module=gnav',
+        tags: 'errorType=error,module=global-footer',
       }));
 
     if (!this.body) return;
--- a/libs/blocks/global-footer/global-footer.js
+++ b/libs/blocks/global-footer/global-footer.js
@@ -1,22 +1,24 @@
 /* eslint-disable no-async-promise-executor */
 import {
   decorateAutoBlock,
-  getConfig,
-  getMetadata,
-  loadBlock,
   decorateLinks,
+  getMetadata,
+  getConfig,
+  loadBlock,
 } from '../../utils/utils.js';
 
 import {
-  toFragment,
-  getExperienceName,
-  loadDecorateMenu,
   getFedsPlaceholderConfig,
+  federatePictureSources,
+  getExperienceName,
   getAnalyticsValue,
+  loadDecorateMenu,
+  getFederatedUrl,
   loadBaseStyles,
   yieldToMain,
   lanaLog,
   logErrorFor,
+  toFragment,
 } from '../global-navigation/utilities/utilities.js';
 
 import { replaceKey, replaceText } from '../../features/placeholders.js';
@@ -30,11 +32,11 @@ const CONFIG = {
 };
 
 class Footer {
-  constructor(footerEl, contentUrl) {
-    this.footerEl = footerEl;
+  constructor({ contentUrl, block, useFederatedContent } = {}) {
+    this.useFederatedContent = useFederatedContent;
     this.contentUrl = contentUrl;
+    this.block = block;
     this.elements = {};
-
     this.init();
   }
 
@@ -56,7 +58,7 @@ class Footer {
       }
     }, intersectionOptions);
 
-    observer.observe(this.footerEl);
+    observer.observe(this.block);
 
     // Set timeout after which we load the footer automatically
     decorationTimeout = setTimeout(() => {
@@ -101,9 +103,9 @@ class Footer {
     }
 
     const mepMartech = mep?.martech || '';
-    this.footerEl.setAttribute('daa-lh', `gnav|${getExperienceName()}|footer${mepMartech}`);
+    this.block.setAttribute('daa-lh', `gnav|${getExperienceName()}|footer${mepMartech}`);
 
-    this.footerEl.append(this.elements.footer);
+    this.block.append(this.elements.footer);
   }, 'Failed to decorate footer content', 'errorType=error,module=global-footer');
 
   fetchContent = async () => {
@@ -120,7 +122,7 @@ class Footer {
 
     if (!html) return null;
 
-    const parsedHTML = await replaceText(html, getFedsPlaceholderConfig(), undefined, 'feds');
+    const parsedHTML = await replaceText(html, getFedsPlaceholderConfig());
 
     try {
       return new DOMParser().parseFromString(parsedHTML, 'text/html').body;
@@ -167,7 +169,7 @@ class Footer {
 
     const content = await file.text();
     const elem = toFragment`<div class="feds-footer-icons">${content}</div>`;
-    this.footerEl.append(elem);
+    this.block.append(elem);
   };
 
   decorateProducts = async () => {
@@ -181,7 +183,7 @@ class Footer {
     this.elements.featuredProducts = toFragment`<div class="feds-featuredProducts"></div>`;
 
     const [placeholder] = await Promise.all([
-      replaceKey('featured-products', getFedsPlaceholderConfig(), 'feds'),
+      replaceKey('featured-products', getFedsPlaceholderConfig()),
       this.loadMenuLogic(),
     ]);
 
@@ -208,10 +210,8 @@ class Footer {
       url = new URL(regionSelector.href);
     } catch (e) {
       lanaLog({ message: `Could not create URL for region picker; href: ${regionSelector.href}`, tags: 'errorType=error,module=global-footer' });
-      throw e;
+      return this.elements.regionPicker;
     }
-
-    if (!url) return this.elements.regionPicker;
 
     const regionPickerClass = 'feds-regionPicker';
     const regionPickerTextElem = toFragment`<span class="feds-regionPicker-text">${regionSelector.textContent}</span>`;
@@ -357,12 +357,22 @@ class Footer {
         </div>
       </div>`;
 
+    if (this.useFederatedContent) federatePictureSources(this.elements.footer);
     return this.elements.footer;
   };
 }
 
 export default function init(block) {
-  const url = getMetadata('footer-source') || `${locale.contentRoot}/footer`;
-  const footer = new Footer(block, url);
-  return footer;
+  try {
+    const contentUrl = getFederatedUrl(getMetadata('footer-source') || `${locale.contentRoot}/footer`);
+    const footer = new Footer({
+      block,
+      contentUrl,
+      useFederatedContent: contentUrl.includes('/federal/'),
+    });
+    return footer;
+  } catch (e) {
+    lanaLog({ message: 'Could not create footer', e });
+    return null;
+  }
 }
--- a/libs/blocks/global-footer/global-footer.js
+++ b/libs/blocks/global-footer/global-footer.js
@@ -13,7 +13,7 @@ import {
   getExperienceName,
   getAnalyticsValue,
   loadDecorateMenu,
-  getFederatedUrl,
+  fetchAndProcess,
   loadBaseStyles,
   yieldToMain,
   lanaLog,
@@ -21,7 +21,7 @@ import {
   toFragment,
 } from '../global-navigation/utilities/utilities.js';
 
-import { replaceKey, replaceText } from '../../features/placeholders.js';
+import { replaceKey } from '../../features/placeholders.js';
 
 const { miloLibs, codeRoot, locale, mep } = getConfig();
 const base = miloLibs || codeRoot;
@@ -32,9 +32,7 @@ const CONFIG = {
 };
 
 class Footer {
-  constructor({ contentUrl, block, useFederatedContent } = {}) {
-    this.useFederatedContent = useFederatedContent;
-    this.contentUrl = contentUrl;
+  constructor({ block } = {}) {
     this.block = block;
     this.elements = {};
     this.init();
@@ -69,7 +67,13 @@ class Footer {
 
   decorateContent = () => logErrorFor(async () => {
     // Fetch footer content
-    this.body = await this.fetchContent();
+    const url = getMetadata('footer-source') || `${locale.contentRoot}/footer`;
+    this.useFederatedContent = url.includes('/federal/');
+    this.body = await fetchAndProcess({
+      url,
+      message: 'Error fetching footer content',
+      shouldDecorateLinks: false,
+    });
 
     if (!this.body) return;
 
@@ -107,30 +111,6 @@ class Footer {
 
     this.block.append(this.elements.footer);
   }, 'Failed to decorate footer content', 'errorType=error,module=global-footer');
-
-  fetchContent = async () => {
-    const resp = await fetch(`${this.contentUrl}.plain.html`);
-
-    if (!resp.ok) {
-      lanaLog({
-        message: `Failed to fetch footer content; content url: ${this.contentUrl}, status: ${resp.statusText}`,
-        tags: 'errorType=warn,module=global-footer',
-      });
-    }
-
-    const html = await resp.text();
-
-    if (!html) return null;
-
-    const parsedHTML = await replaceText(html, getFedsPlaceholderConfig());
-
-    try {
-      return new DOMParser().parseFromString(parsedHTML, 'text/html').body;
-    } catch (e) {
-      lanaLog({ message: 'Footer could not be instantiated', tags: 'errorType=error,module=global-footer' });
-      return null;
-    }
-  };
 
   loadMenuLogic = async () => {
     this.menuLogic = this.menuLogic || new Promise(async (resolve) => {
@@ -364,12 +344,7 @@ class Footer {
 
 export default function init(block) {
   try {
-    const contentUrl = getFederatedUrl(getMetadata('footer-source') || `${locale.contentRoot}/footer`);
-    const footer = new Footer({
-      block,
-      contentUrl,
-      useFederatedContent: contentUrl.includes('/federal/'),
-    });
+    const footer = new Footer({ block });
     return footer;
   } catch (e) {
     lanaLog({ message: 'Could not create footer', e });
--- a/libs/blocks/global-footer/global-footer.js
+++ b/libs/blocks/global-footer/global-footer.js
@@ -9,11 +9,10 @@ import {
 
 import {
   getFedsPlaceholderConfig,
-  federatePictureSources,
   getExperienceName,
   getAnalyticsValue,
   loadDecorateMenu,
-  fetchAndProcess,
+  fetchAndProcessPlainHtml,
   loadBaseStyles,
   yieldToMain,
   lanaLog,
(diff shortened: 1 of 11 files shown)


libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -36,21 +54,65 @@ export function toFragment(htmlStrings, ...values) {
...
+  try {
+    const { pathname, search, hash } = new URL(url);
+    return `${getFederatedContentRoot()}${pathname}${search}${hash}`;
+  } catch (e) {
+    lanaLog({ message: `getFederatedUrl errored parsing the URL: ${url}`, e, tags: 'errorType=warn,module=global-footer' });
   }
+  return url;
+};
+
+export const federatePictureSources = (section) => {
```
mokimo, line 92: PS: It makes sense to eventually move those into the milo utils and integrate this capability with milo core, but for now the gnav is the guinea pig and we'll go from there.

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -243,3 +288,46 @@ export function processMartechAttributeMetadata(html) {
...
+      const { default: loadInlineFrags } = await import('../../fragment/fragment.js');
+
+      const fragPromises = inlineFrags.map((link) => {
+        // Replacing paragraphs should happen in the fragment module
+        // https://jira.corp.adobe.com/browse/MWPW-141039
+        if (link.parentElement && link.parentElement.nodeName === 'P') {
+          const div = document.createElement('div');
+          link.parentElement.replaceWith(div);
+          div.appendChild(link);
+        }
```
mokimo, line 310: PS: I'll integrate this change in the fragments directly AFTER this initial PR.

libs/blocks/global-footer/global-footer.js
```diff
@@ -356,12 +336,17 @@ class Footer {
         </div>
       </div>`;
 
+    if (this.useFederatedContent) federatePictureSources(this.elements.footer);
```
mokimo, line 339: self-review: As the pictureSources are already being federated at the time of fetching, this should not be needed anymore.

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -236,7 +238,8 @@ class Gnav {
         ${breadcrumbs}
       </div>`;
 
-    this.el.append(this.elements.curtain, this.elements.aside, this.elements.topnavWrapper);
+    if (this.useFederatedContent) federatePictureSources(this.elements.topnavWrapper);
```
mokimo, line 241: self-review: As the pictureSources are already being federated right after fetching, this should not be needed anymore.

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -243,3 +288,46 @@ export function processMartechAttributeMetadata(html) {
   });
   return null;
 }
+
+export async function fetchAndProcess({ url, message, shouldDecorateLinks = true } = {}) {
```
overmyheadandbody, line 292: Could we make the method name a bit more specific? It's not immediately clear what is fetched and processed, could be images, some data, fragments, etc.

Also, I'd make the `message` argument name more specific, it's an error message.

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -243,3 +288,46 @@ export function processMartechAttributeMetadata(html) {
...
+    blocks.forEach((block) => {
+      import('../../martech-metadata/martech-metadata.js')
+        .then(({ default: decorate }) => decorate(block));
+    });
+
+    body.innerHTML = await replaceText(body.innerHTML, getFedsPlaceholderConfig());
+    return body;
+  } catch (e) {
+    lanaLog({ message: `${message} url: ${url}`, e, tags: 'errorType=error,module=utilities' });
+    return null;
```
overmyheadandbody, line 331: Is `null` the right default here? Could it be added to the DOM if things fail? If so, we might want to go with `''` to not expose `null` to the end user.

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -243,3 +288,46 @@ export function processMartechAttributeMetadata(html) {
...
+
+    if (path.includes('/federal/')) federatePictureSources(body);
+
+    if (shouldDecorateLinks) decorateLinks(body);
+
+    const blocks = body.querySelectorAll('.martech-metadata');
+    blocks.forEach((block) => {
+      import('../../martech-metadata/martech-metadata.js')
+        .then(({ default: decorate }) => decorate(block));
+    });
```
overmyheadandbody, line 325: Could we not check for the `.martech-metadata` class and if there are any matches, import the module, which, when resolved, decorates all the blocks? The resource will likely be cached, but I think it's an easy enough change to avoid eventual multiple requests altogether.

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -243,3 +288,46 @@ export function processMartechAttributeMetadata(html) {
...
+
+    body.innerHTML = await replaceText(body.innerHTML, getFedsPlaceholderConfig());
+    return body;
+  } catch (e) {
+    lanaLog({ message: `${message} url: ${url}`, e, tags: 'errorType=error,module=utilities' });
```
overmyheadandbody, line 330: I can't say I'm entirely sold on having the message here. Could we not tie it to where we actually call the method and avoid passing it in as a param?

libs/blocks/global-footer/global-footer.js
overmyheadandbody: In which case we wouldn't even need `this.useFederatedContent`, right?

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -710,21 +714,23 @@ class Gnav {
...
-  processMartechAttributeMetadata(parsedHTML);
-
+export default async function init(block) {
   try {
-    const gnav = new Gnav(new DOMParser().parseFromString(parsedHTML, 'text/html').body, header);
+    const { locale } = getConfig();
+    const url = getMetadata('gnav-source') || `${locale.contentRoot}/gnav`;
+    const content = await fetchAndProcess({
+      url,
+      e: 'Error fetching gnav content',
```
overmyheadandbody, line 723: I think `e` should be `message`? I added some thoughts on passing this in the utilities file

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -710,21 +714,23 @@ class Gnav {
...
+    });
+    if (!content) return null;
+    const gnav = new Gnav({
+      content,
+      block,
+      useFederatedContent: url.includes('/federal/'),
```
overmyheadandbody, line 729: It's not obvious if we actually have a need for `useFederatedContent` anymore, does it serve any purpose?

overmyheadandbody (commented):
Looking good overall, I had a look over the code, but will do some actual testing later on

libs/blocks/global-navigation/utilities/utilities.js
mokimo: What do you think about: `fetchAndProcessPlainHtml` ?
This should be useful for any place we fetch anything `.plain.html` and enrich it with all the neccesary things (localization, placeholders, federated pictures etc...)

libs/blocks/global-footer/global-footer.js
mokimo: Yes, though I'll need to check this in detail

libs/blocks/global-navigation/utilities/utilities.js
mokimo: I'm a bit cautious as the method carries a lot of logic and can error.  So we'd always need a try/catch upstream of this method if we removed the try/catch here. If we'd just remove the error message itself and keep the try catch, we'd lose context of _what_ errored.

libs/blocks/global-navigation/global-navigation.js
mokimo: I'm unsure about one menu use-case, but I'll check. We might be able to completely get rid of this now as I've restructured the code around the central `fetchAndProcess(PlainHtml)` method

libs/blocks/global-navigation/utilities/utilities.js
overmyheadandbody: I'l all for keeping the `try/catch` here, just maybe throw an error and handle it in the module itself, so that the error message is closer to the logic itself; might also make it easier to investigate LANA logs

libs/blocks/global-navigation/utilities/utilities.js
overmyheadandbody: Sounds reasonable, sure!

libs/blocks/global-navigation/global-navigation.js
mokimo: Yeah, we should be good to get rid of it completely - since we federatePictureSources right as we fetch anything

libs/blocks/global-navigation/global-navigation.js
mokimo: Slightly refactored this part

libs/blocks/global-navigation/utilities/utilities.js
mokimo: Slightly changed the logic to bring the lana logs closer to where they should be

libs/blocks/global-navigation/utilities/utilities.js
mokimo: Copy/pasted this from the base work Vivian had done, I refactored this slightly

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -36,21 +54,65 @@ export function toFragment(htmlStrings, ...values) {
...
 
-export const getFedsPlaceholderConfig = () => {
-  const { locale } = getConfig();
-  let libOrigin = 'https://milo.adobe.com';
+let federatedContentRoot;
+export const getFederatedContentRoot = () => {
+  if (federatedContentRoot) return federatedContentRoot;
+
   const { origin } = window.location;
+  const allowedOrigins = [
```
narcis-radu, line 62: I would move the values at the top of the file for easy update.
Also, is the any problems with allowing `*.adobe.com` ? I know we would need additional steps in Akamai, but might be easier to allow everything here.

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -36,21 +54,65 @@ export function toFragment(htmlStrings, ...values) {
...
+    const type = source.hasAttribute('src') ? 'src' : 'srcset';
+    const value = source.getAttribute(type);
+    if (!value) return;
+    source.setAttribute(type, value.replace(/^\.?\//, `${getFederatedContentRoot()}/`));
+  });
+};
+
+let fedsPlaceholderConfig;
+export const getFedsPlaceholderConfig = ({ useCache = true } = {}) => {
+  if (useCache && fedsPlaceholderConfig) return fedsPlaceholderConfig;
```
narcis-radu, line 103: any usage for this? I've seen it used without cache only in tests

libs/blocks/global-navigation/utilities/utilities.js
narcis-radu: We can have a TODO

narcis-radu:
Great work, Okan! I checked multiple use-cases, both new and existing functionality are working perfectly. I would like to add some TODOs to move methods in utils and allow other blocks to use federated content. We should also discuss a bit more about the allowed origins, maybe we agree on a pattern to avoid adding a new origin every time a new project requires it.

overmyheadandbody (approved):
Looks good! We'll need to look out for the Akamai changes and update based on that. I see this is a branch on the main repo, so we can make those eventual updates easily while you're out.

libs/blocks/global-navigation/utilities/utilities.js
mokimo: @overmyheadandbody introduced this initially I think - either way, since placeholders get cached at module level on import - we couldn't otherwise test different configs.
So we need it for the unit tests

libs/blocks/global-navigation/utilities/utilities.js
mokimo: I'm not sure if we forgot any subdomain, so I'd like to default to www.adobe.com in those cases and take the small performance hit (extra TLS handshake), instead of potentially breaking any projects // end up without placeholders.

libs/blocks/global-footer/global-footer.js
```diff
@@ -67,7 +66,16 @@ class Footer {
...
-    this.body = await this.fetchContent();
+    const url = getMetadata('footer-source') || `${locale.contentRoot}/footer`;
+    this.body = await fetchAndProcessPlainHtml({
+      url,
+      shouldDecorateLinks: false,
+    })
+      .catch((e) => lanaLog({
+        message: `Error fetching footer content ${url}`,
+        e,
+        tags: 'errorType=error,module=gnav',
```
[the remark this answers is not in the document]
mokimo: emission on my end, should be global-footer - fixed

libs/blocks/global-navigation/utilities/utilities.js
overmyheadandbody: Indeed, I couldn't find a better way for unit tests, this was the lesser evil

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -1,7 +1,14 @@
...
 import { processTrackingLabels } from '../../../martech/attributes.js';
+import { replaceText } from '../../../features/placeholders.js';
 
 loadLana();
+const allowedOrigins = [
+  'https://www.adobe.com',
+  'https://business.adobe.com',
+  'https://blog.adobe.com',
+  'https://milo.adobe.com',
+];
```
chrischrischris, line 11: Should this be defined in the config level instead?  Possibly hard code these in milo but allow consumers to add their site to the list?

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -36,21 +60,60 @@ export function toFragment(htmlStrings, ...values) {
...
-    libOrigin = `https://main--milo--adobecom.hlx.${origin.includes('hlx.live') ? 'live' : 'page'}`;
+    federatedContentRoot = `https://main--federal--adobecom.hlx.${origin.includes('hlx.live') ? 'live' : 'page'}`;
   }
 
-  return {
+  return federatedContentRoot;
+};
+
+export const getFederatedUrl = (url = '') => {
+  if (typeof url !== 'string' || !url.includes('/federal/')) return url;
```
chrischrischris, line 82: `/federal/` can be anywhere in the url?  Should we only be matching when it's the first path value?

libs/blocks/global-navigation/utilities/utilities.js
mokimo: Yes, that could make sense, I added a TODO to revisit this when porting this closer to milo core and out of the GNAV.

libs/blocks/global-navigation/utilities/utilities.js
mokimo: It's always `/federal` or `/locale/federal/` and defined the same way in Akamai, it would make sense to match the Akamai locale aware patterns.

I added a TODO once we port this out of the gnav and make this more of a milo core feature.

test/blocks/global-navigation/global-navigation.test.js
```diff
@@ -25,6 +27,52 @@ describe('global navigation', () => {
...
+    beforeEach(() => {
+      fetchStub = sinon.stub(window, 'fetch');
+      sinon.stub(window.lana, 'log');
+      setConfig({ locale: { ietf: 'en-US', prefix: '' } });
+    });
+
+    afterEach(() => {
+      fetchStub = null;
+      window.fetch.restore();
+      window.lana.log.restore();
```
robert-bogos, line 43: A possible optimisation, that reduces the number of lines, would be to use `sinon.restore()` instead of individually restoring stubs.

robert-bogos (approved):
Everything looks fine 👍. Just added a comment on a small optimisation opportunity.

narcis-radu (approved):

PR #1719, opened 2024-01-10
MWPW-135163 added quantity selection functionality

VKniaz:
Title: merch-quantity-select component added.

Description:

This PR includes the new merch-quantity-select component and updates to the merch-card component.

1. **Merch Card Updates**: The `merch-offer-select.js` file has been updated to work with merhc-quantity-select. 
2. **Merch-quantity-select**: The `Merch-quantity-select.js` file has been added to include new functionality for quantity selection. 
3. **New pattern introduced**: The new patter for authoring was introduced in this PR: 
Screenshot 2024-01-08 at 8 52 34 AM
where: 

- Quantity it's markup.
- Select a quantity as a title.
- 1,3,1 in order 1 - as the initial value, 3 as the last value, 1 step. 
With this config we will have component like:
 
Screenshot 2024-01-09 at 5 11 50 PM

Resolves: [MWPW-135163](https://jira.corp.adobe.com/browse/MWPW-135163)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/vkniaziev/document6?martech=off
- After: https://mwpw-135163--milo--vkniaz.hlx.page/drafts/vkniaziev/document6?martech=off


libs/blocks/merch-card/merch-card.js
```diff
@@ -4,6 +4,7 @@ import { getMetadata } from '../section-metadata/section-metadata.js';
 import { processTrackingLabels } from '../../martech/attributes.js';
 import { replaceKey } from '../../features/placeholders.js';
 import '../../deps/merch-card.js';
+import '../../deps/merch-quantity-select.js';
```
yesil, line 7: this should be imported only if there is a quantity selector.

libs/blocks/merch-card/merch-card.js
```diff
@@ -228,6 +229,23 @@ const simplifyHrs = (el) => {
   });
 };
 
+function createQuantitySelect(el) {
+  const quantitySelectConfig = el.querySelector('ul');
+  if (!quantitySelectConfig) return null;
+  const configMarkup = quantitySelectConfig.querySelector('li');
+  if (!configMarkup || !configMarkup.textContent.includes('Quantity')) return null;
+  const config = configMarkup.querySelector('ul').querySelectorAll('li');
```
yesil, line 237: ` const config = configMarkup.querySelectorAll('ul li');`

libs/blocks/merch-card/merch-card.js
```diff
@@ -228,6 +229,23 @@ const simplifyHrs = (el) => {
...
+  if (config.length !== 2) return null;
+  const attributes = {};
+  attributes.title = config[0].textContent.trim();
+  const quantityValues = config[1].textContent.split(',').map((value) => value.trim());
+  if (quantityValues.length !== 3) return null;
+  [attributes.min, attributes.max, attributes.step] = quantityValues.map(Number);
```
yesil, line 243: please make sure that `quantityValues.map(Number);` is safe against to invalid values.
e.g:
`['', 'a', "0"].map(Number)`

the minimal value for any of the above is 1.

libs/blocks/merch-card/merch-card.js
```diff
@@ -228,6 +228,24 @@ const simplifyHrs = (el) => {
...
+function createQuantitySelect(el) {
+  const quantitySelectConfig = el.querySelector('ul');
+  if (!quantitySelectConfig) return null;
+  const configMarkup = quantitySelectConfig.querySelector('li');
+  if (!configMarkup || !configMarkup.textContent.includes('Quantity')) return null;
+  const config = configMarkup.querySelector('ul').querySelectorAll('li');
+  if (config.length !== 2) return null;
+  const attributes = {};
+  attributes.title = config[0].textContent.trim();
+  const quantityValues = config[1].textContent.split(',').map((value) => value.trim()).filter((value) => /^\d+$/.test(value));
```
Axelcureno, line 240: 100 max char per line please

libs/blocks/merch-card/merch-card.js
```diff
@@ -228,6 +228,24 @@ const simplifyHrs = (el) => {
   });
 };
 
+function createQuantitySelect(el) {
+  const quantitySelectConfig = el.querySelector('ul');
```
Axelcureno, line 232: I don't know if it's a good idea to hijack all `<ul>` tags of the entire block for this. I have seen other merch card variants coming up soon with unordered list text. Can we instead be more specific to detect quantity select? How about creating a new row that starts with `<ul>` for this?

Thoughts? @3ch023 @yesil @npeltier @honstar

libs/deps/merch-card.js
```diff
@@ -38,6 +38,7 @@ import{html as n,LitElement as T}from"/libs/deps/lit-all.min.js";import{css as u
     .top-section {
         display: flex;
         justify-content: flex-start;
+        flex-flow: wrap;
```
Axelcureno, line 41: Please pull the latest from tacocat.js and rebuild the dependency again. This change would reopen a ticket already closed.

libs/blocks/merch-card/merch-card.js
yesil: Yes, the selector should be more specific.
e.g:

get the latest `li` of a `ul`, and  check that it starts with a numerical value, if yes, assume that its parent `ul` is the one you are looking for.

VKniaz: In my case, I added the check for card-type plans, and in step 2, I'm checking if it contains 'Quantity' inside, so it's already pretty specific. 
Screenshot 2024-01-16 at 4 33 44 PM
Screenshot 2024-01-16 at 4 34 28 PM

libs/blocks/merch-card/merch-card.js
Axelcureno: As per Milo guidelines the limit is 100 characters. It would be good if we can keep new lines under that limit and later we can clean up the rest.

VKniaz: Changed.

libs/blocks/merch-card/merch-card.js
```diff
@@ -350,11 +369,20 @@ const init = async (el) => {
...
   }
   merchCard.appendChild(footer);
-  const offerSelection = cardType === 'plans' ? el.querySelector('ul') : null;
-  if (offerSelection) {
-    const { initOfferSelection } = await import('./merch-offer-select.js');
-    initOfferSelection(merchCard, offerSelection);
+
+  if (cardType === 'plans') {
+    const quantitySelect = createQuantitySelect(el);
+    const offerSelection = el.querySelector('ul');
```
Axelcureno, line 375: Here you are again looking for any `ul` and if a card has text content with this tag it will cause unexpected behavior. 
I suggest to refactor `createQuantitySelect(el)` to initialize respectively without introducing variant-specific code inside `init` and avoid running the same `querySelector()` more than once. I feel like everything quantity select-related could be inside one function.

libs/blocks/merch-card/merch-card.js
```diff
@@ -350,11 +369,20 @@ const init = async (el) => {
     footer.append(ctas);
   }
   merchCard.appendChild(footer);
-  const offerSelection = cardType === 'plans' ? el.querySelector('ul') : null;
-  if (offerSelection) {
-    const { initOfferSelection } = await import('./merch-offer-select.js');
-    initOfferSelection(merchCard, offerSelection);
+
+  if (cardType === 'plans') {
```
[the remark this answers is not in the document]
VKniaz: It will not work.
initOfferSelection(merchCard, offerSelection, quantitySelect); can be without quantitySelect.

libs/blocks/merch-card/merch-card.js
VKniaz: This is a different component, so that is why  I need to query it again.

yesil (approved):

PR #1723, opened 2024-01-10
MWPW-140456 - [Aside] fix notification static links color

Sartxi:
When an author uses the `section.static-links` style with an `aside.notification.extra-small`, the link color is not applied because of a style that targets trailing links that aren't styled as buttons. I am adding an additional style that sets color to `inherit` when aside is nested in a section with static links style.

Note: I also removed a duplicate selector that was creating a styleLint error

Resolves: [MWPW-140456](https://jira.corp.adobe.com/browse/MWPW-140456)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/sarchibeque/aside-links?martech=off
- After: https://sarchibeque-mwpw-140456-aside-links--milo--adobecom.hlx.page/drafts/sarchibeque/aside-links?martech=off

### DIFF
--- a/libs/blocks/aside/aside.css
+++ b/libs/blocks/aside/aside.css
@@ -72,6 +72,7 @@
 
 .aside .foreground.container .image {
   position: relative;
+  display: flex;
 }
 
 .aside .foreground.container > div {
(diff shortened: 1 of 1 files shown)


meganthecoder (approved):

meganthecoder:
@Sartxi Is this fixing a bug on production? If so, it seems like it should be tested rather than being labeled as "trivial"

elan-tbx (approved):
LGTM

ryanmparrish (approved):

sukamat (approved):

PR #1726, opened 2024-01-11
MWPW-141139 Parameter PZN tags don't match if user uses uppercase characters

vgoodric:
MEP PZN "param-" tags fail if the URL uses uppercase letters in the parameter you are matching against.  Parameter matches shoul

Resolves: [MWPW-141139](https://jira.corp.adobe.com/browse/MWPW-141139)

**Test URLs:**
- Before:
https://main--milo--adobecom.hlx.page/drafts/vgoodrich/fragments/mep-param-case-insensitive/?ttid=UPPERCASE&martech=off
https://main--milo--adobecom.hlx.page/drafts/vgoodrich/fragments/mep-param-case-insensitive/?ttid=LOWERCASE&martech=off
https://main--milo--adobecom.hlx.page/drafts/vgoodrich/fragments/mep-param-case-insensitive/?TTID=alluppercase&martech=off
- After: 
https://mep-param-case-insensitive--milo--adobecom.hlx.page/drafts/vgoodrich/fragments/mep-param-case-insensitive/?ttid=UPPERCASE&martech=off
https://mep-param-case-insensitive--milo--adobecom.hlx.page/drafts/vgoodrich/fragments/mep-param-case-insensitive/?ttid=LOWERCASE&martech=off
https://mep-param-case-insensitive--milo--adobecom.hlx.page/drafts/vgoodrich/fragments/mep-param-case-insensitive/?TTID=alluppercase&martech=off


libs/features/personalization/personalization.js
```diff
@@ -311,7 +311,7 @@ function parsePlaceholders(placeholders, config, selectedVariantName = '') {
 const checkForParamMatch = (paramStr) => {
   const [name, val] = paramStr.split('param-')[1].split('=');
   if (!name) return false;
-  const searchParamVal = PAGE_URL.searchParams.get(name);
+  const searchParamVal = PAGE_URL.searchParams.get(name).toLowerCase();
```
mokimo, line 314: Since searchParamVal can be `null`, let's safeguard the chained method.
```suggestion
  const searchParamVal = PAGE_URL.searchParams.get(name)?.toLowerCase();
```

overmyheadandbody: Agree with @mokimo, optional chaining should be used

chrischrischris (changes requested):
We should also have the parameter name be case-insensitive.  Using:

https://mep-param-case-insensitive--milo--adobecom.hlx.page/drafts/vgoodrich/fragments/mep-param-case-insensitive/?TTid=UPPERCASE&martech=off

does not work.

libs/features/personalization/personalization.js
vgoodric: little embarrased I missed that.  Fixed.

vgoodric:
> We should also have the parameter name be case-insensitive. Using:
> 
> https://mep-param-case-insensitive--milo--adobecom.hlx.page/drafts/vgoodrich/fragments/mep-param-case-insensitive/?TTid=UPPERCASE&martech=off
> 
> does not work.

@chrischrischris Updated to work.  Also added an example in the manifest where the param name is capitalized to make sure that works.

https://mep-param-case-insensitive--milo--adobecom.hlx.page/drafts/vgoodrich/fragments/mep-param-case-insensitive/?TTID=alluppercase&martech=off

libs/features/personalization/personalization.js
```diff
@@ -318,7 +318,11 @@ function parsePlaceholders(placeholders, config, selectedVariantName = '') {
 const checkForParamMatch = (paramStr) => {
   const [name, val] = paramStr.split('param-')[1].split('=');
   if (!name) return false;
-  const searchParamVal = PAGE_URL.searchParams.get(name);
+  const params = new URLSearchParams(PAGE_URL.search);
+  const newParams = new URLSearchParams(
+    Array.from(params, ([key, value]) => [key.toLowerCase(), value]),
+  );
+  const searchParamVal = newParams.get(name.toLowerCase())?.toLowerCase();
```
chrischrischris, line 325: nit: small refactor
```suggestion
  const params = new URLSearchParams(
    Array.from(PAGE_URL.searchParams, ([key, value]) => [key.toLowerCase(), value?.toLowerCase()]),
  );
  const searchParamVal = params.get(name.toLowerCase());
```

chrischrischris (approved):
Approved with small tweak

seanchoi-dev (approved):

PR #1729, opened 2024-01-11
MWPW-136866: merch-cards block

yesil:
`merch-cards` block requires a custom helix-query definition(e.g, https://github.com/adobecom/cc/blob/main/helix-query.yaml#L557) therefore can be only tested on the consumer projects.

Resolves:
[MWPW-136866](https://jira.corp.adobe.com/browse/MWPW-136866)

**Test URLs:**
- After: https://commercev3--cc--adobecom.hlx.live/drafts/ilyas/catalog?milolibs=mwpw-136866-main--milo--yesil&martech=off&georouting=off


libs/blocks/merch-cards/merch-cards.js
```diff
@@ -0,0 +1,257 @@
...
+  'searchResultMobileText',
+  'searchResultsMobileText',
+  'noSearchResultsText',
+  'noSearchResultsMobileText',
+  'showMoreText',
+];
+
+// allows improve TBT by returning control to the main thread.
+// eslint-disable-next-line no-promise-executor-return
+const makePause = async () => new Promise((resolve) => setTimeout(resolve, 0));
```
honstar, line 31: Can you share some data around this TBT optimization? It'd be interesting to understand the impact of interrupting processing in the `initMerchCards` function further down.

libs/blocks/merch-cards/merch-cards.js
```diff
@@ -0,0 +1,257 @@
...
+      err = res.statusText || res.status;
+    }
+  } catch (error) {
+    err = error.message;
+  }
+  if (!cardsData) {
+    fail(el, err);
+  }
+
+  // TODO add aditional parameters.
```
honstar, line 84: Is this still relevant and an outstanding task? If yes, is there a follow-up ticket for this?

yesil: I think it is not relevant, removing it.

libs/blocks/merch-cards/merch-cards.js
```diff
@@ -0,0 +1,256 @@
...
+          // avoid layout shift due to size of first cards.
+          merchCards.requestUpdate();
+        }
+        await makePause();
+      }
+      merchCards.displayResult = true;
+    }
+    ));
+
+  const appContainer = document.querySelector('.merch.app');
```
honstar, line 242: What is this `.merch.app` container? Is this driven by authoring?

libs/blocks/merch-cards/merch-cards.js
npeltier: i think it would deserve some central discussion (and feature!) as it as been used by @narcis-radu and @mokimo iirc on feds

npeltier: (not saying we need feature/discussion before end of that PR)

honstar: > i think it would deserve some central discussion (and feature!) as it as been used by @narcis-radu and @mokimo iirc on feds

Right, a TBT optimization discussion sounds like a good idea 👍

yesil: @honstar we have around 93 cards(fragment, block) in the catalog. We don't need to initialise all of them at the same time to render. It is increasing TBT considerably.
Here by interrupting the processes and giving the UI thread chance to do something else, we keep the TBT under control.

libs/blocks/merch-cards/merch-cards.js
yesil: Yes,  basically if a merch block finds a `.merch.app` element in the DOM, it will go stick to it instead of being rendered in-place.

sidenav in cc do the same so that we can achieve the desired layout.

For the catalog page, we have a dumb block serving as a merch.app container.

libs/blocks/merch-cards/merch-cards.css
```diff
@@ -0,0 +1,7 @@
+.section[class*="-merch-cards"] .full-width {
+  grid-column: 1 / -1;
+}
+
+merch-cards {
+  min-height: 200px;
```
npeltier, line 6: where is that 200px come from?

npeltier (approved):

libs/blocks/merch-cards/merch-cards.css
yesil: @npeltier it is an arbitrary value to allocate some space for the cards.
in a second PR,, I'll adjust spacing after feedback from different stakeholders and more visual tests.

honstar (approved):

libs/blocks/merch-cards/merch-cards.js
yesil: the thing is, we don't know which card is the first card, and the card metadata must be initialised so that we know whether it must be displayed or not.
So we should still eagerly initialise them as early as possible, but with mini fake pauses :)

PR #1735, opened 2024-01-12
MWPW-137932 - Bullet-proof quiz and result metadata keys

colloyd:
* add a method to remove characters other than alpha-numeric
* standardized on no dashes or spaces in property names

Resolves: [MWPW-137932](https://jira.corp.adobe.com/browse/MWPW-137932)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/xiasun/quiz-2/?martech=off
- After: https://mwpw-137932-bulletproof-property-names--milo--colloyd.hlx.page/drafts/colloyd/quiz/?martech=off

Testing notes: To test, observe the following URL on milo main: https://main--milo--adobecom.hlx.page/drafts/colloyd/quiz/
Note the odd characters in the property names that are causing block load failure. This is the same page used for the "After" URL above.

The same has been done for two result pages, though you will have to submit a quiz to generate local storage data before observing them and it may be easier to observe the junk characters in the keys by editing the page with the sidekick
 * https://main--milo--adobecom.hlx.page/drafts/colloyd/quiz-results-block/quiz-results-tester?debug=quiz-results
 * https://main--milo--adobecom.hlx.page/drafts/colloyd/quiz-results-block/quiz-results-nested-tester?debug=quiz-results

### DIFF
--- a/libs/blocks/quiz/utils.js
+++ b/libs/blocks/quiz/utils.js
@@ -22,9 +22,9 @@ const initQuizKey = () => {
   return locale?.ietf ? `${quizKey}-${locale.ietf}` : quizKey;
 };
 
-const initAnalyticsType = () => metaData['analytics-type']?.text;
+const initAnalyticsType = () => metaData.analyticstype?.text;
 
-const initAnalyticsQuiz = () => metaData['analytics-quiz']?.text;
+const initAnalyticsQuiz = () => metaData.analyticsquiz?.text;
 
 async function fetchContentOfFile(path) {
   const response = await fetch(configPath(path));
@@ -32,8 +32,8 @@ async function fetchContentOfFile(path) {
 }
 
 export const initConfigPathGlob = (rootElement) => {
-  metaData = getMetadata(rootElement);
-  shortQuiz = metaData['short-quiz']?.text === 'true';
+  metaData = getNormalizedMetadata(rootElement);
+  shortQuiz = metaData.shortquiz?.text === 'true';
   configPath = initConfigPath(metaData);
   quizKey = initQuizKey(rootElement);
   analyticsType = initAnalyticsType();
@@ -237,6 +237,23 @@ const getNestedFragments = (resultResources, productCodes, fragKey) => {
     });
   });
   return fragArray;
+};
+
+/**
+ * Normalizes the values of the metadata keys in the metadata table, cleaning them up
+ * and removing all but alphanumeric characters in preparation for getMetadata();
+ */
+const normalizeKeys = (el) => {
+  const metadata = [...el.childNodes];
+  for (const row of metadata) {
+    if (row.children) {
+      let key = row.children[0].textContent;
+      if (key) {
+        key = key.match(/[a-zA-Z0-9]/g).join('');
+        row.children[0].innerHTML = key;
+      }
+    }
+  }
 };
 
 export const getRedirectUrl = (destinationPage) => {
@@ -465,3 +482,8 @@ export const getAnalyticsDataForLocalStorage = (config) => {
 };
 
 export const isValidUrl = (url) => VALID_URL_RE.test(url);
+
+export const getNormalizedMetadata = (el) => {
+  normalizeKeys(el);
+  return getMetadata(el);
+};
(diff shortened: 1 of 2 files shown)


libs/blocks/quiz/utils.js
```diff
@@ -239,6 +239,22 @@ const getNestedFragments = (resultResources, productCodes, fragKey) => {
...
 
+/**
+ * Normalizes the metadata keys in the metadata object,
+ * cleaning them up and removing all but alphanumeric characters
+ */
+const normalizeKeys = (data) => {
+  const keys = Object.keys(data);
+  for (const key of keys) {
+    const newKey = key.match(/[a-zA-Z0-9]/g).join('');
+    if (key !== newKey) {
```
fullcolorcoder, line 250: I think it would be safer to add them to a new empty object rather than edit the one we are iterating over.

fullcolorcoder (changes requested):

libs/blocks/quiz/utils.js
colloyd: Is there a specific danger or issue that you had in mind here? I am leaving keys alone that don't have issues with the match and only adding a new key and removing the old if the key has unsafe characters and needs to change. 

If I make a new object, the result is the same, though I would have to add keys that don't have issues to the new object. The returned data will be the new object and the original will just be discarded.

Would you feel less worried if I refrain from deleting any keys?

echampio-at-adobe: Here's what the machine gods of OpenAI say about it:

Modifying keys in an object while iterating through them can lead to unexpected behavior and errors in JavaScript. When you iterate over the keys of an object using a loop like for...in or Object.keys(), you are essentially creating a snapshot of the keys at the beginning of the iteration.

If you add or remove keys during the iteration, it can affect the loop's behavior. For example, adding a new key might cause the loop to skip some keys or repeat others. Removing a key may cause the loop to skip or miss keys as well.

Here's an example to illustrate the issue:

javascript
Copy code
let myObject = { key1: 'value1', key2: 'value2', key3: 'value3' };

for (let key in myObject) {
  console.log(key); // key1, key2, key3

  // Modifying the object during iteration
  if (key === 'key2') {
    delete myObject.key1; // Removing a key
    myObject.newKey = 'newValue'; // Adding a new key
  }
}

console.log(myObject); // { key2: 'value2', key3: 'value3', newKey: 'newValue' }
In this example, the loop does not iterate over the newly added key (newKey), and it skips the deleted key (key1). This behavior can lead to unintended consequences and bugs in your code.

To avoid such issues, it's recommended not to modify the structure of the object while iterating over its keys. If you need to make changes, consider creating a new object or using a different approach to achieve your desired result.

JackySun9: yes, this is from ChatGPT:
To improve the normalizeKeys function in JavaScript for better performance and clarity, consider the following modifications:

Use Object.entries and reduce: Instead of iterating with a for loop and modifying the original object, use Object.entries to get key-value pairs and reduce to build a new object. This approach avoids modifying the original object and is generally considered a more functional programming style.

Simplify Regular Expression: If the goal is to keep only alphanumeric characters in the keys, the regular expression is fine. However, ensure it meets your specific needs.

Check for Unnecessary Changes: Avoid creating a new key if it's identical to the original, which can save processing time.

Here's how the improved function might look:

javascript
Copy code
const normalizeKeys = (data) => {
  return Object.entries(data).reduce((newObj, [key, value]) => {
    const newKey = key.match(/[a-zA-Z0-9]/g)?.join('') || key;
    newObj[newKey] = value;
    return newObj;
  }, {});
};

colloyd: @echampio-at-adobe that makes sense, but I'm not actually iterating through the object here: I've taken a snapshot of the keys only using Object.keys, which gives an array of the keys, which is what I'm actually iterating through. I run a newKey through the match and if it's not the same as the base key, then update the object. 

Still think this is an issue, given that I'm looping through an array of the keys, rather than the object itself?

colloyd: I've adjusted this to just use a new object. If there's a difference, use the matched key in the new object, if there wasn't anything to change, just add it to the new object as is.

JackySun9 (approved):

echampio-at-adobe (approved):

fullcolorcoder (approved):

PR #1739, opened 2024-01-16
MWPW-136892: Entitlement aware checkout links

yesil:
enabled by default, can be disabled with &entitlements=0 in the OST url

**This feature will be merged as completely disabled for now, and will be enabled once GWP is ready with the download links**

Resolves: [MWPW-136892](https://jira.corp.adobe.com/browse/MWPW-136892)

Includes: 
* https://git.corp.adobe.com/wcms/tacocat.js/pull/513
* https://git.corp.adobe.com/wcms/tacocat.js/pull/517

To test, sign-in with a Photoshop single-app user on stage

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.live/drafts/ilyas/cta-entitlements?martech=off&georouting=off
- After: https://ims--milo--adobecom.hlx.live/drafts/ilyas/cta-entitlements?martech=off&georouting=off


3ch023 (approved):

libs/blocks/merch/merch.js
```diff
@@ -116,17 +124,67 @@ export async function getPriceContext(el, params) {
...
+  return [entitlements, entitlementsMetadata];
+}
+
+export async function handleEntitlements(cta, promise) {
+  try {
+    const [{ arrangement_codes: aCodes }, entitlementsMappings] = await promise;
+    if (aCodes !== undefined) {
+      await cta.onceSettled();
+      const { value: [{ productArrangementCode }] } = cta;
+      if (aCodes[productArrangementCode] === true) {
```
3ch023, line 148: right now you are comparing by arrangementCode, will it be the case also in future?
or we will switch to productFamily?

if we switch to product family you could already now get the family on an entitlement, without using the mapping. I think we need mapping only for the cta offer, but not for user entitlements.

```
const entitlements = await getUserEntitlements({ params: [{ name: 'include', value: 'OFFER.PRODUCT_ARRANGEMENT' }], format: 'raw' });

....

entitlements.map((offer) => offer.offer.product_arrangement.family)

```

libs/blocks/merch/merch.js
```diff
@@ -116,17 +124,67 @@ export async function getPriceContext(el, params) {
...
+         || cta.firstElementChild.innerHTML;
+        cta.href = mapping.Target;
+      }
+    }
+  } catch {
+    // ignore
+  } finally {
+    cta.classList.remove(LOADING_ENTITLEMENTS);
+  }
+}
```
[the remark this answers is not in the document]
yesil: @JasonHowellSlavin  we need to call the finally block. I didn't want to duplicate the line.
I'll se if I can flatten more.

libs/blocks/merch/merch.js
yesil: As soon as ODM-15772 is ready(pandora clientlibs must be updated as well), we can reduce it to a comparison of product families IMO.

But for now, I'll keep it as is.

Axelcureno (approved):

VKniaz (approved):

JasonHowellSlavin (approved):

libs/blocks/merch/merch.js
```diff
@@ -116,17 +124,62 @@ export async function getPriceContext(el, params) {
...
+export async function loadEntitlements() {
+  await loadIms();
+  if (!window.adobeIMS?.isSignedInUser?.()) return Promise.resolve([{}, {}]);
+  const [{ default: getUserEntitlements },
+    mappings] = await Promise.all([
+    import('../global-navigation/utilities/getUserEntitlements.js'),
+    fetch(entitlementsURL),
+  ]);
+  const entitlements = await getUserEntitlements();
+  const entitlementsMetadata = await mappings.json();
```
overmyheadandbody, line 138: Did you have a look whether grouping these into a `Promise.all` might have any benefit?

libs/blocks/merch/merch.js
```diff
@@ -116,17 +124,62 @@ export async function getPriceContext(el, params) {
...
+  return [entitlements, entitlementsMetadata];
+}
+
+export async function handleEntitlements(cta, promise) {
```
overmyheadandbody, line 142: Personal preference, but I usually steer away from using reserved keywords with different capitalisation, like `let Let`, this could cause confusion. Instead of just `promise`, maybe you could find a more specific name so it's immediately obvious what it represents

libs/blocks/merch/merch.js
```diff
@@ -116,17 +124,62 @@ export async function getPriceContext(el, params) {
...
+  if (aCodes === undefined) return;
+  await cta.onceSettled();
+  const { value: [{ productArrangementCode }] } = cta;
+  if (aCodes[productArrangementCode] === true) {
+    const mapping = entitlementsMappings.data
+      ?.find(({ [TITLE_PRODUCT_ARRANGEMENT_CODE]: code }) => code === productArrangementCode);
+    if (!mapping) return;
+    cta.firstElementChild.innerHTML = (await replaceKey(mapping.CTA, getConfig()))
+         || cta.firstElementChild.innerHTML;
+    cta.href = mapping.Target;
```
overmyheadandbody, line 153: Any chance `mapping.Target` might be `undefined`?

overmyheadandbody (approved):

libs/blocks/merch/merch.js
yesil: given that the fetch is already done, I don't expect big difference, but I can still do it.

libs/blocks/merch/merch.js
yesil: This is an excel file owned by GWP,  so it should not be.
This is for logged-in only use case.
Still I'll keep the original value as fallback.

libs/blocks/merch/merch.js
yesil: I find `.promise`  very self describing but open to suggestions.

yesil:
codecov/patch ratio will will be back to original once I restore disabled tests in a future PR once GWP work is done.

PR #1750, opened 2024-01-17
MWPW-141446: Remove the locale from the quiz key in the URL; continue to use it in the background; normalize other chars

JackySun9:
* Remove the locale from the quiz key in the URL
* continue to use quiz key with locale in the background
* normalize other chars and use lowercase key and value in URL

Resolves: [MWPW-141446](https://jira.corp.adobe.com/browse/MWPW-141446)

**Test URLs:**
- Before: https://uar-integration--milo--adobecom.hlx.page/drafts/quiz/quiz-2/?martech=off
- After: https://mwpw-141446--milo--jackysun9.hlx.page/drafts/quiz/quiz-2/?martech=off

### DIFF
--- a/libs/blocks/quiz-results/quiz-results.js
+++ b/libs/blocks/quiz-results/quiz-results.js
@@ -1,4 +1,4 @@
-import { createTag } from '../../utils/utils.js';
+import { createTag, getConfig } from '../../utils/utils.js';
 import { handleStyle } from '../section-metadata/section-metadata.js';
 import { getNormalizedMetadata } from '../quiz/utils.js';
 
@@ -54,7 +54,13 @@ export default async function init(el, debug = null, localStoreKey = null) {
 
   /* eslint-disable no-param-reassign */
   // handle these two query param values in this way to facilitate unit tests
-  localStoreKey ??= params.get('quizKey');
+  localStoreKey ??= params.get('quizkey');
+
+  const { locale } = getConfig();
+  if (locale?.ietf) {
+    localStoreKey = `${localStoreKey}-${locale.ietf}`;
+  }
+
   debug ??= params.get('debug');
 
   el.replaceChildren();
--- a/libs/blocks/quiz-results/quiz-results.js
+++ b/libs/blocks/quiz-results/quiz-results.js
@@ -57,9 +57,7 @@ export default async function init(el, debug = null, localStoreKey = null) {
   localStoreKey ??= params.get('quizkey');
 
   const { locale } = getConfig();
-  if (locale?.ietf) {
-    localStoreKey = `${localStoreKey}-${locale.ietf}`;
-  }
+  localStoreKey = locale?.ietf ? `${localStoreKey}-${locale.ietf}` : localStoreKey;
 
   debug ??= params.get('debug');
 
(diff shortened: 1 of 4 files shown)


libs/blocks/quiz-results/quiz-results.js
```diff
@@ -54,7 +54,13 @@ export default async function init(el, debug = null, localStoreKey = null) {
 
   /* eslint-disable no-param-reassign */
   // handle these two query param values in this way to facilitate unit tests
-  localStoreKey ??= params.get('quizKey');
+  localStoreKey ??= params.get('quizkey');
+
+  const { locale } = getConfig();
+  if (locale?.ietf) {
```
echampio-at-adobe, line 60: Rather than multi-line statements, folks seem to prefer single-line variants to be as obscure as possible. Would something like this variation work? 
localStoreKey = locale?.ietf ? `${localStoreKey}-${locale.ietf}` : localStoreKey;

libs/blocks/quiz/utils.js
```diff
@@ -150,7 +160,7 @@ export const storeResultInLocalStorage = (
     ),
     pageloadHash: getAnalyticsDataForLocalStorage(analyticsConfig),
   };
-  localStorage.setItem(quizKey, JSON.stringify(resultToDelegate));
+  localStorage.setItem(quizLocalKey, JSON.stringify(resultToDelegate));
```
echampio-at-adobe, line 163: Since quizLocalKey isn't used anywhere else, perhaps define and init it here, within the storeResultInLocalStorage method?

colloyd (approved):

fullcolorcoder:
@JackySun9 what is the best way to verify this with the test URLs?

JackySun9:
@fullcolorcoder 
go to this url https://mwpw-141446--milo--jackysun9.hlx.page/drafts/quiz/quiz-2/?martech=off and select any product and check quizkey in url and localstorage

libs/blocks/quiz/utils.js
```diff
@@ -35,7 +36,7 @@ export const initConfigPathGlob = (rootElement) => {
   metaData = getNormalizedMetadata(rootElement);
   shortQuiz = metaData.shortquiz?.text === 'true';
   configPath = initConfigPath(metaData);
-  quizKey = initQuizKey(rootElement);
+  quizKey = initQuizKey();
```
echampio-at-adobe, line 39: This is a complete nit-pick, but given that the init methods for quizKey, analyticsType and analyticsQuiz are so trivial, why not just do away with the methods and set them directly, like we do the shortQuiz value?

export const initConfigPathGlob = (rootElement) => {
  metaData = getNormalizedMetadata(rootElement);
  shortQuiz = metaData.shortquiz?.text === 'true';
  configPath = initConfigPath(metaData);
  quizKey = metaData.storage?.text.toLowerCase();
  analyticsType = metaData.analyticstype?.text;
  analyticsQuiz = metaData.analyticsquiz?.text;
  return { configPath, quizKey, analyticsType, analyticsQuiz, shortQuiz };
};

echampio-at-adobe:
> @fullcolorcoder go to this url https://mwpw-141446--milo--jackysun9.hlx.page/drafts/quiz/quiz-2/?martech=off and select any product and check quizkey in url and localstorage

I think the key is to watch local storage in the application tab between tests, nuking the value that we're using to be sure it's actually being set so that the results page still reads it. 

The real side-effect that I'm trying to avoid is that, today, with a key like cc-quiz-en-US in the URL, if other locales were live, and we used the region selector, we would wind up going to a results page in, say, France, and still loading English data. With this change, we would redirect to the French results page, which would append the French locale to the key, and then, finding nothing, would redirect to the French quiz, which would set its own French data.

echampio-at-adobe (approved):

PR #1751, opened 2024-01-18
MWPW-141053: Enable additional components for the MWEB

robert-bogos:
## Description
This PR enables the universal navigation on mobile.

## Related Issue
Resolves: [MWPW-141053](https://jira.corp.adobe.com/browse/MWPW-141053)

## Screenshots:
Screenshot 2024-01-17 at 17 53 21

Screenshot 2024-01-18 at 09 34 55

## Test URLs
**Milo:**
- Before:  https://gnav--milo--adobecom.hlx.page/drafts/rbogos/unav?martech=off
- After: https://mwpw-141053-mweb-additional-components--milo--robert-bogos.hlx.page/drafts/rbogos/unav?martech=off

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -466,17 +466,15 @@ class Gnav {
       // reset sign up value on change
       children[0].attributes.isSignUpRequired = false;
 
-      if (isDesktop.matches) {
-        this.universalNavComponents?.forEach((component) => {
-          if (component === 'profile') return;
-          if (component === 'signup') {
-            children[0].attributes.isSignUpRequired = true;
-            return;
-          }
+      this.universalNavComponents?.forEach((component) => {
+        if (component === 'profile') return;
+        if (component === 'signup') {
+          children[0].attributes.isSignUpRequired = true;
+          return;
+        }
 
-          children.push(CONFIG.universalNav.components[component]);
-        });
-      }
+        children.push(CONFIG.universalNav.components[component]);
+      });
 
       return children;
     };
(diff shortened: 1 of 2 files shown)


robert-bogos:
The conflict will be resolved after testing, we need the `imsClientId: 'fedsmilo'` to enable IMS.

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -466,17 +466,15 @@ class Gnav {
       // reset sign up value on change
       children[0].attributes.isSignUpRequired = false;
 
-      if (isDesktop.matches) {
```
overmyheadandbody, line 469: I remember there was a `change` event listener for `isDesktop`, L509 in this file. Is that still needed?

narcis-radu: @overmyheadandbody - I was thinking the same, but thought about the following flow: user has a device with width suitable for desktop, the universal nav is rendered; then user rotates the device and it goes into mobile width; without reinitialising the universal navigation I am concerned that some previous widths, positions will be preserved and the dropdowns might appear to be detached from the navigation. I think it's safer to reload the navigation when viewport changes.

narcis-radu (approved):
all good, we need to remove the custom client ID before merging.

libs/blocks/global-navigation/global-navigation.js
robert-bogos: Good catch! Thanks 👍 We won't need to reload the unav on switching between desktop and mobile viewport anymore as it should always be visible. I made the change.

overmyheadandbody (approved):

SilviuLCF:
PR can be merged to MAIN. with following comments:

Analytics not available
Some UNAV bugs , listed in the https://jira.corp.adobe.com/browse/MWPW-140227 TEST comment

From integration point of view -- PASS

PR #1758, opened 2024-01-19
MWPW-136727 MEP: Support for simplified selectors

chrischrischris:
Adds support for simplified selectors:
* Can now use `blockNameX` where `X` is the index of which block to select if there are multiple blocks of the same kind on the page.  E.g: `merchcard3` would select the 3rd merchcard on the page.
* Can now target specific table cells within a block, using `rowX colX` notation (row X, col X).  Note that for row, the header row is not counted.
  * E.g: `marquee1 row2 col2` would target the first marquees' table cell in the 2nd row, 2nd column.

Resolves: [MWPW-136727](https://jira.corp.adobe.com/browse/MWPW-136727)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/cpeyer/mep-cell/surfing-marquee?martech=off
- After: 
    https://mep-cell--milo--adobecom.hlx.page/drafts/cpeyer/mep-cell/surfing-marquee?martech=off
    https://mep-cell--milo--adobecom.hlx.page/drafts/cpeyer/mep-cell/surfing-marquee?martech=off&m-pic=shuttle
    https://mep-cell--milo--adobecom.hlx.page/drafts/cpeyer/mep-cell/surfing-marquee?martech=off&m-pic=octopus

### DIFF
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -239,11 +239,86 @@ function normalizeKeys(obj) {
   }, {});
 }
 
+const getDivInTargetedCell = (tableCell) => {
+  tableCell.replaceChildren();
+  const div = document.createElement('div');
+  tableCell.appendChild(div);
+  return div;
+};
+
+const querySelector = (el, selector, all = false) => {
+  try {
+    return all ? el.querySelectorAll(selector) : el.querySelector(selector);
+  } catch (e) {
+    /* eslint-disable-next-line no-console */
+    console.log('Invalid selector: ', selector);
+    return null;
+  }
+};
+
+function getTrailingNumber(s) {
+  const match = s.match(/\d+$/);
+  return match ? parseInt(match[0], 10) : null;
+}
+
+function getSection(rootEl, idx) {
+  return rootEl === document
+    ? document.querySelector(`body > main > div:nth-child(${idx})`)
+    : rootEl.querySelector(`:scope > div:nth-child(${idx})`);
+}
+
+function getSelectedEl(rootEl, selector) {
+  if (!selector) return null;
+
+  let selectedEl = querySelector(rootEl, selector);
+  if (selectedEl) return selectedEl;
+
+  const terms = selector.split(/\s+/);
+
+  let section = null;
+  if (terms[0]?.startsWith('section')) {
+    const sectionIdx = getTrailingNumber(terms[0]) || 1;
+    section = getSection(rootEl, sectionIdx);
+    terms.shift();
+  }
+
+  if (terms.length) {
+    const blockStr = terms.shift();
+    if (blockStr.includes('.')) {
+      selectedEl = querySelector(section || rootEl, blockStr);
+    } else {
+      const blockIndex = getTrailingNumber(blockStr) || 1;
+      const blockName = blockStr.replace(`${blockIndex}`, '');
+      if (blockName === 'block') {
+        if (!section) section = getSection(rootEl, 1);
+        selectedEl = querySelector(section, `:scope > div:nth-of-type(${blockIndex})`);
+      } else {
+        selectedEl = querySelector(section || rootEl, `.${blockName}`, true)?.[blockIndex - 1];
+      }
+    }
+  } else if (section) {
+    return section;
+  }
+
+  if (terms.length) {
+    // find targeted table cell in rowX colY format
+    const rowColMatch = /row(?<row>\d+)\s+col(?<col>\d+)/gm.exec(terms.join(' '));
+    if (rowColMatch) {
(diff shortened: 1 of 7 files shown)


libs/features/personalization/personalization.js
```diff
@@ -216,11 +216,67 @@ function normalizeKeys(obj) {
...
+};
+
+function getSelectedEl(rootEl, selector) {
+  if (!selector) return null;
+
+  let selectedEl = querySelector(rootEl, selector);
+  if (selectedEl) return selectedEl;
+
+  // TODO: do we support space delimeted selectors?
+  const [firstTerm, secondTerm] = selector.split(' ').filter((str) => str?.trim());
```
npeltier, line 246: may be we want to define a "space" as `\s+` ?

libs/features/personalization/personalization.js
```diff
@@ -216,11 +216,67 @@ function normalizeKeys(obj) {
...
+    /* eslint-disable-next-line no-console */
+    console.log('Invalid selector: ', selector);
+    return null;
+  }
+};
+
+function getSelectedEl(rootEl, selector) {
+  if (!selector) return null;
+
+  let selectedEl = querySelector(rootEl, selector);
```
npeltier, line 242: shouldn't we rather do before this string tests on selector for `<blockname>-<num>` or cell regexp? i guess those two cases are supposed to be seldom, but we never know where that goes, and i guess string tests should be quicker than empty DOM query.

npeltier (commented):
shouldn't we rather

vgoodric:
@chrischrischris 
1. This could just end up in a separate PR, but wondering if you should add support for sections here.  "section1" etc.  In particular combining them "section1 text-2" to get the second instance of a text block in that specific section.
2. Have you tested this with a multi word block name like merch-card?  Will it throw it to have a dash in the middle? Should we support them using spaces like "merch card"?
3. I think it is more prone to user error to need a dash after the block name, but only there.  I know it's easier to code for, but wondering if that's possible.
But the challenge is balancing common use cases versus bloating the code with a huge system.

libs/features/personalization/personalization.js
```diff
@@ -216,11 +216,67 @@ function normalizeKeys(obj) {
...
+  if (selectedEl) return selectedEl;
+
+  // TODO: do we support space delimeted selectors?
+  const [firstTerm, secondTerm] = selector.split(' ').filter((str) => str?.trim());
+  if (firstTerm.includes('-')) {
+    const [, num] = firstTerm.split('-');
```
overmyheadandbody, line 248: Or, if the colon approach doesn't sound appealing, maybe we could split by a regex: dash, number, end of string

chrischrischris:
> @chrischrischris still relevant or can this be closed?

@mokimo Yes still relevant.  PM is currently testing and I need a new round of reviews.

libs/features/personalization/personalization.js
```diff
@@ -216,11 +216,67 @@ function normalizeKeys(obj) {
...
+  return div;
+};
+
+const querySelector = (el, selector, all = false) => {
+  try {
+    if (all) {
+      return el.querySelectorAll(selector);
+    }
+    return el.querySelector(selector);
+  } catch (e) {
```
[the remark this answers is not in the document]
chrischrischris: I don't get one, but I do if there is no arg.

3ch023:
hi @chrischrischris is this ready to merge? labels check is missing, would you say it needs to be verified or trivial?

3ch023 (approved):

3ch023:
Verified on live CC and Milo, no impact for Promos:
CC: https://main--cc--adobecom.hlx.page/de/products/special-offers?milolibs=mep-cell 
Milo: https://mep-cell--milo--adobecom.hlx.live/drafts/mariia/pr/promo/promo?martech=off

libs/features/personalization/personalization.js
```diff
@@ -239,11 +239,89 @@ function normalizeKeys(obj) {
...
+  tableCell.appendChild(div);
+  return div;
+};
+
+const querySelector = (el, selector, all = false) => {
+  try {
+    return all ? el.querySelectorAll(selector) : el.querySelector(selector);
+  } catch (e) {
+    /* eslint-disable-next-line no-console */
+    console.log('Invalid selector: ', selector);
```
overmyheadandbody, line 254: Should this be left here for the final version? I see there's also a unit test for this, so I assume so. Should we at least consider logging only on non-prod environments? Is `log` the right choice, or would `warn` be more visible for authors?

overmyheadandbody (approved):

chrischrischris:
Merged as part of https://github.com/adobecom/milo/pull/1975

PR #1659, opened 2023-12-14
Mwpw 139968: Mnemonic list block 

Axelcureno:
- Introduces Mnemonic list block, which can be used in two ways: 
  - As an independent block.
  - Inside another block (using inline fragments and dynamic import).
- Implemented for Marquee block using the style `mnemonic-list` (Merch card implementation coming soon).
- Implementation inside another block happens using dynamic imports and inline fragments to avoid impact on LCP.

Resolves: [MWPW-139968](https://jira.corp.adobe.com/browse/MWPW-139968)
Xd: https://xd.adobe.com/view/432e3b75-3e56-4ea2-bc4b-b5e13c9ee2cd-167d/screen/83d63683-6649-47d4-951d-444ef03bff12/specs/
### Marquee
289697337-a919b31d-3be1-4f4c-a125-c0f8466f7d40

### Merch Card
image-2023-08-15-15-36-37-202

### Individual block 
Screenshot 2024-01-16 at 1 53 03 PM

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/axel/fragments/catalog-mnemonic-list?martech=off
- After: https://mwpw-139968--milo--axelcureno.hlx.page/drafts/axel/fragments/catalog-mnemonic-list?martech=off

PR #1759, opened 2024-01-19
Mwpw 139968: Mnemonic list block comments

Axelcureno:
Addresses review comments made on https://github.com/adobecom/milo/pull/1659 including:

- Removes explicit height of mnemonics.
- Simplifies DOM inside block.
- Removes unnecessary media query CSS.

Resolves: [MWPW-139968](https://jira.corp.adobe.com/browse/MWPW-139968)

**Test URLs:**
- Before: https://main--milo--axelcureno.hlx.page/drafts/axel/fragments/catalog-mnemonic-list?martech=off
- After: https://mwpw-139968--milo--axelcureno.hlx.page/drafts/axel/fragments/catalog-mnemonic-list?martech=off


libs/blocks/marquee/marquee.css
```diff
@@ -718,7 +718,7 @@
   color: inherit;
 }
 
-@media screen and (min-width: 768px) {
+@media screen and (min-width: 900px) {
```
ryanmparrish, line 721: Media queries should go in a natural order. aka 900 should come before 1200. 

@Axelcureno  - Where is the example of this in a marquee block. The `.marquee.mnemonic-list ` selector indicates this is a variant of the marquee? We shouldn't separate concerns and not have nested block styles where ever possible. Big picture, Is this even needed. What happens if your in a marquee.quiet w/ out a split content?

Axelcureno: I removed this piece of code since it's not required anymore (confirmed this with Consonant and product team).
The example of the Marquee using this block can be found here: https://main--cc--adobecom.hlx.page/drafts/axel/catalog-with-mnemonic

libs/blocks/mnemonic-list/mnemonic-list.js
```diff
@@ -17,7 +17,13 @@ export const decorateMnemonicList = (container) => {
     if (title) product.appendChild(title);
     productList.appendChild(product);
     paragraph.replaceWith(productList);
+    paragraph.remove();
```
overmyheadandbody, line 20: Paragraph is already replaced by `productList`, so it should no longer be attached to the DOM, what purpose does this extra statement serve?

libs/blocks/mnemonic-list/mnemonic-list.js
```diff
@@ -17,7 +17,13 @@ export const decorateMnemonicList = (container) => {
...
   });
+  if (targetElement) {
+    targetElement.prepend(productList);
+  }
```
overmyheadandbody, line 24: You could go for optional chaining and make it a one-liner, `targetElement?.prepend(productList);`

vladen (approved):

yesil (commented):
+1 for all the comments of @overmyheadandbody

ryanmparrish (approved):
Thanks for the quick adjustment. 
Maybe just remove that second bullet in the description since it doesn't look like your doing that? 
> Sets breakpoint to 900px as per Consonant

After [Rares](https://github.com/adobecom/milo/pull/1759/files#r1464846388) feedback, LGTM

libs/blocks/mnemonic-list/mnemonic-list.js
Axelcureno: good catch, removed.

libs/blocks/mnemonic-list/mnemonic-list.js
Axelcureno: Thanks! Fixed.

libs/blocks/mnemonic-list/mnemonic-list.js
```diff
@@ -17,7 +17,13 @@ export const decorateMnemonicList = (container) => {
...
+  const divs = targetElement.querySelectorAll('div:not([class])');
```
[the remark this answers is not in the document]
Axelcureno: Thanks for the suggestion.

overmyheadandbody (approved):

PR #1760, opened 2024-01-19
MWPW-135540 - Enable a new section-metadata style: reverse-mobile

ryanmparrish:
BACOM is looking to have an opt in option to display `two-up` section content to display reverse position on mobile views so they can easily author their changing content on landing pages w/ out having to move a bunch of copy around. 

This will only be available when using a style: two-up w/ an additional `reverse-mobile` 

Resolves: [MWPW-135540](https://jira.corp.adobe.com/browse/MWPW-135540)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rparrish/ups/two-up-row-reverse?martech=off
- After: https://rparrish-reverse-two-up--milo--adobecom.hlx.page/drafts/rparrish/ups/two-up-row-reverse?martech=off

Example impacted URLs:
https://main--bacom--adobecom.hlx.page/products/experience-manager/new
https://main--bacom--adobecom.hlx.page/products/experience-platform/launch


libs/blocks/section-metadata/section-metadata.css
```diff
@@ -203,6 +203,17 @@ main > .section[class*='-up'] > .content {
...
+@media (max-width: 600px) {
+  .section.reverse-mobile > div:nth-child(1) {
+    order: 2;
+  }
+
+  .section.reverse-mobile > div:nth-child(2) {
+    order: 1;
+  }
+}
+
```
rgclayton, line 216: Your description says this will only work for `two-up`. But according to this it will work with any of the ups. 

Do we want to scope it to  `.two-up` or leave it open like it is? 

I guess my main thinking is. What's stopping authoring from trying this on three-up, four-up, etc., then submitting bug tickets since it only orders the first two children.

ryanmparrish: Ah, yes, Good eye mate. I've addressed this to scope as described. 
I don't want this option available on any more `ups` then a `two-up`, otherwise it'll get real hard to manage orders.

libs/blocks/accordion/accordion.css
```diff
@@ -158,6 +150,18 @@ html[dir="rtl"] .accordion dt button::before {
...
+}
+
+html[dir="rtl"] .accordion dt button::before {
+  right: 0;
+}
+
+.section[class*='-up'] dl.accordion {
+  max-width: unset;
+}
+
```
rgclayton, line 164: I see this makes the accordion wider when in an up pattern. I didn't see any mention of this change in the ticket. We sure this change is ok? 

Thinking bigger picture. Technically the masonry grid can work with any blocks. `masonry` would need this update too. That would be a separate ticket, but wanted to note it.

ryanmparrish: Good call. I can leave this part out. I just noticed all the accordions in an up weren't using the full space available and still had a gap so I figured it should have been this way from the start. But agree, probs a question first.

ryanmparrish: @rgclayton  - I spoke w/ bacom and consonant and they are in agreement that this should actually stay as is so that the accordion is as wide as possible when in a 'ups' grid. I am going to leave this in for the time being and if we need to address/extend this on other grid offerings we can do so in a future PR. Good thoughts on this. Thanks.

rgclayton (approved):

elan-tbx (approved):

Sartxi (approved):

PR #1766, opened 2024-01-22
MWPW-141609: Universal navigation NO localization tag

robert-bogos:
## Description
This PR changes the IETF language tag, for Norway, that is sent as config to the universal navigation.

## Related Issue
Resolves: [MWPW-141609](https://jira.corp.adobe.com/browse/MWPW-141609)

## Screenshots:
**Before:**
before

**After:**
after

## Test URLs
**Milo:**
- Before:  https://gnav--milo--adobecom.hlx.page/no/drafts/sghimpos/gnav-refactor-1-1
- After: https://mwpw-141609-unav-no-localization--milo--robert-bogos.hlx.page/no/drafts/sghimpos/gnav-refactor-1-1

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -487,7 +487,7 @@ class Gnav {
     const getConfiguration = () => ({
       target: this.blocks.universalNav,
       env: environment,
-      locale,
+      locale: locale === 'no_NO' ? 'nb_NO' : locale,
       imsClientId: window.adobeid?.client_id,
       theme: 'light',
       analyticsContext: {
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -206,8 +206,7 @@ const closeOnClickOutside = (e) => {
 
 const getIetfLocale = (ietfLocale) => {
   const nonStandardLocaleMap = { no_NO: 'nb_NO' };
-  if (nonStandardLocaleMap[ietfLocale]) return nonStandardLocaleMap[ietfLocale];
-  return ietfLocale;
+  return nonStandardLocaleMap[ietfLocale] || ietfLocale;
 };
 
 class Gnav {
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -204,6 +204,12 @@ const closeOnClickOutside = (e) => {
   }
 };
 
+const getIetfLocale = (ietfLocale) => {
+  const nonStandardLocaleMap = { no_NO: 'nb_NO' };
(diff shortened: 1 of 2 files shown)


libs/blocks/global-navigation/global-navigation.js
```diff
@@ -487,7 +487,7 @@ class Gnav {
     const getConfiguration = () => ({
       target: this.blocks.universalNav,
       env: environment,
-      locale,
+      locale: locale === 'no_NO' ? 'nb_NO' : locale,
```
narcis-radu, line 490: I would rather have the condition outside of `getConfiguration` to make sure the value is consistent across code. With this current setup, `nb_NO` would be passed to Universal Navigation, but `no_NO` would get passed if we use `locale` in a different place.

narcis-radu (changes requested):

libs/blocks/global-navigation/global-navigation.js
robert-bogos: Made the change 👍

libs/scripts/scripts.js
```diff
@@ -123,7 +123,7 @@ const config = {
   geoRouting: 'on',
   fallbackRouting: 'on',
   links: 'on',
-  imsClientId: 'milo',
+  imsClientId: 'fedsmilo',
```
narcis-radu, line 126: I assume this will not go into the final code!

narcis-radu (approved):

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -204,6 +204,12 @@ const closeOnClickOutside = (e) => {
   }
 };
 
+const getIetfLocale = (ietfLocale) => {
+  const nonStandardLocaleMap = { no_NO: 'nb_NO' };
+  if (nonStandardLocaleMap[ietfLocale]) return nonStandardLocaleMap[ietfLocale];
+  return ietfLocale;
```
overmyheadandbody, line 210: Nit: could we not combine the two lines in `return nonStandardLocaleMap[ietfLocale] || ietfLocale `?

libs/scripts/scripts.js
overmyheadandbody: I see this is becoming a pattern used so that QA can perform testing. Should we have some logic that replaces the IMS client id based on a query parameter? Should we add some documentation on some steps to replace it via the browser developer tools? That way we wouldn't have to worry about this update reaching the `main` branch

overmyheadandbody (approved):
Conditional approval, the IMS client ID change should be reverted before merging

robert-bogos:
IMS client ID will be changed after QA approval

libs/blocks/global-navigation/global-navigation.js
robert-bogos: Good suggestion 👍 Made the change

PR #1768, opened 2024-01-22
Implemented option to add more links to qr-code media block variant, …

saragajic11:
Resolves: [MWPW-141580](https://jira.corp.adobe.com/browse/MWPW-141580)
Authoring team needs an option to have multiple links in media block's qr-code variant. 
In order to achieve this, my proposition is to set qr-code-links to be bold, while all other links keep font-weight normal.

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/sara/document4?martech=off
- After: https://mwpw-141580-media-block-qr-code-variant-fix--milo--saragajic11.hlx.page/drafts/sara/document4?martech=off


libs/blocks/media/media.js
```diff
@@ -75,15 +75,16 @@ export default function init(el) {
       if (imgQRCode) {
         imgQRCode.classList.add('qr-code-img');
       }
-      const qrCodeLinks = row.querySelectorAll('a');
+      const qrCodeLinks = row.querySelectorAll('.con-button');
+      qrCodeLinks.forEach(qrCodeLink => {
```
Blainegunn, line 79: fix the syntax please  
`qrCodeLinks.forEach((qrCodeLink) => {`

libs/blocks/media/media.js
```diff
@@ -75,15 +75,16 @@ export default function init(el) {
...
+        qrCodeLink.textContent = '';
+        qrCodeLink.classList.remove('con-button');
+        qrCodeLink.parentNode.classList.add('qr-button-container');
+      })
```
Blainegunn, line 83: missing semicolon

Blainegunn (changes requested):
other than2 minor lint issue this looks good to me! I'll let the Ryans determine if that's an acceptable solution for authors.

libs/blocks/media/media.js
saragajic11: Resolved: https://github.com/adobecom/milo/pull/1768/commits/9ce75840bca514bd3ba927660ba02f2bc914aade

Blainegunn (approved):

gabr1ela:
Dear Ryans @ryanmparrish @rgclayton do you mind reviewing please?

chrischrischris (approved):

ryanmparrish (changes requested):
@saragajic11  - I think we need to revisit the logic on this guy. Since we already use **bold** as a general button decorator, I worry that this will confuse authors. Also this breaks if/when the app links aren't bold, so it's not very author resilient. Im going to look at it some more and see if any suggestions come to mind, but this is my initial reaction. More to come...

rgclayton:
@saragajic11 Just to clarify what we are looking at and testing.
The description mentions the QR code view. I do know this block shows different content at desktop (the QR code) and then on mobile it hides the QR code and shows app store buttons. 

The description also mentions applying `bold` to links in the body. However, i'm only seeing `bold` applied to the app store links that appear in mobile. 

Any clarity is appreciated. 

cc: @Blainegunn

rgclayton:
Just to be clear. [This is what happens](https://mwpw-141580-media-block-qr-code-variant-fix--milo--saragajic11.hlx.page/drafts/klee/test-media-qr-code) to ALL existing pages if this code gets merged in.

image

rgclayton (changes requested):
This breaks any page that doesn't currently have the app store buttons bolded. 
We need a better solution.

ryanmparrish:
@saragajic11  - I took a look a bit at this logic and wanted to share another approach. 
https://rparrish-media-qr--milo--adobecom.hlx.page/drafts/rparrish/media/media-qr
Take a peek and let me know if this is helpful. Thanks

saragajic11:
Changed PR based on @ryanmparrish 's comment, thanks a lot! 
Please have another look: https://github.com/adobecom/milo/pull/1768/commits/738a3996f949ec50ad54e666ae69947028efa5f1
@rgclayton @Blainegunn

ryanmparrish (changes requested):
Thanks for the refactor on this. I think it'll be less fragile this way. We should probably add a note in the library docs that when using the `media.qr-code` variant the authoring expects to have the last content in a specific order. 
1) headline/copy/cta etc. 
2) qr-code image/svg - 3rd to last element
3) google-play link - 2nd to last element
4) app-store link - Last element

I am marking this as request changes because the related tests are failing. Can you look at this? Thanks

Screen Shot 2024-01-25 at 10 26 42 AM

saragajic11:
@ryanmparrish unit tests are updated: https://github.com/adobecom/milo/pull/1768/commits/261016e680ac699ab4a61bfdf34fd80bc4d9bab4
Regarding the library docs, I don't have access to it

PR #1773, opened 2024-01-23
MWPW-141413 - JP - wbr tags are causing too many gaps in Firefox

tamanyan:
* Add shouldExclude function
* Prevent the JP word wrap from being applied to the display flex/grid tags in Firefox

Resolves: [MWPW-141413](https://jira.corp.adobe.com/browse/MWPW-141413)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/jp/drafts/takyoshi/?martech=off
- After: https://main--milo--tamanyan.hlx.page/jp/drafts/takyoshi/?martech=off


libs/features/japanese-word-wrap.js
```diff
@@ -72,6 +72,19 @@ function updateParserModel(parser, pattern, score, markerSymbol = '#') {
   }
 }
 
+/**
+ * Determines whether an element should be excluded from Japanese word wrapping.
+ *
+ * @param {Element} element - The element to check.
+ * @returns {boolean} - True if the element should be excluded, false otherwise.
+ */
+function shouldExclude(element) {
```
sharg1, line 81: can this be further simplified?

 ```
 if (navigator.userAgent.includes('Firefox') && getComputedStyle(element).display === 'flex') {
    return true;
  }
  return false;
```

Comment (tamanyan) [reply libs/features/japanese-word-wrap.js:81]: @sharg1 How about this.

```javascript
  const elStyles = getComputedStyle(element);

  return navigator.userAgent.includes('Firefox')
    && (elStyles.display === 'flex' || elStyles.display === 'grid');
```

sharg1: > @sharg1 How about this?
> 
> ```js
>   const elStyles = getComputedStyle(element);
> 
>   return navigator.userAgent.includes('Firefox')
>     && (elStyles.display === 'flex' || elStyles.display === 'grid');
> ```

even better 💯 👍

tamanyan: @sharg1 Thank you. All task and tests have been finished. It can be approved and merged at any time.

libs/features/japanese-word-wrap.js
```diff
@@ -72,6 +72,19 @@ function updateParserModel(parser, pattern, score, markerSymbol = '#') {
...
+  const elStyles = getComputedStyle(element);
+
+  return navigator.userAgent.includes('Firefox')
+    && (elStyles.display === 'flex' || elStyles.display === 'grid');
```
overmyheadandbody, line 85: Does `flex-wrap` play any role in this? Flex elements might flow differently depending on its value, we should check whether it has any impact.

overmyheadandbody (commented):
Also, please make sure to add the appropriate labels to the PR - https://github.com/adobecom/milo/wiki/PR-Labels

tamanyan:
@overmyheadandbody Thank you.
> Also, please make sure to add the appropriate labels to the PR - https://github.com/adobecom/milo/wiki/PR-Labels
I have no permission to add PR labels, please give me permission or you can add them instead of me.

libs/features/japanese-word-wrap.js
tamanyan: Let me explain the background of the change. We use `<wbr>` for Japanese word wrap, which make Japanese sentences more readable. Unfortunately, <wbr> and text are recognized as Flex items in Firefox, which leads to the following issue.

image

tamanyan: @overmyheadandbody In the impact analysis, this function simply determines whether Japanese word wrap needs to be applied. **Even if a false positive occurs, it only means that the Japanese word wrap is not applied.** In other words, it will be displayed according to the browser's default behavior.

libs/features/japanese-word-wrap.js
```diff
@@ -72,6 +72,19 @@ function updateParserModel(parser, pattern, score, markerSymbol = '#') {
   }
 }
 
+/**
+ * Determines whether an element should be excluded from Japanese word wrapping.
+ *
+ * @param {Element} element - The element to check.
+ * @returns {boolean} - True if the element should be excluded, false otherwise.
+ */
```
overmyheadandbody, line 80: I'm a bit surprised to see JSDoc here, as to my knowledge, Milo doesn't build it. We also have a tight file size budget, so large comments are usually discouraged, unless complexity is very high or a special use-case that's not immediately obvious needs to be considered.

libs/features/japanese-word-wrap.js
overmyheadandbody, line 81: The method name seems misleading to me. On L162 & L171, there are multiple exclusion criteria, yet this is called `shouldExclude`. I'd either incorporate the others into a single `shouldExclude` method to have a single source of truth for element exclusion or I'd name this something closer to its purpose, such as `isFlexOrGrid`

libs/features/japanese-word-wrap.js
tamanyan: @overmyheadandbody Dose it mean that JSDoc or comment is not eliminated at the build stage?

tamanyan: It help us read function more easily on vscode.

Screenshot 2024-01-26 at 13 47 37

libs/features/japanese-word-wrap.js
tamanyan: @overmyheadandbody I got your point. I changed it to `hasFlexOrGrid`

libs/features/japanese-word-wrap.js
overmyheadandbody: Yes, Milo is buildless, anything you code will reach the production environment in its entirety, no minification, obfuscation, etc. You'll notice other files don't have JSDoc. We aim for well-named methods and variable names to make their purpose clear. When higher complexity is needed, some simple short comments should suffice

tamanyan: @overmyheadandbody it makes sense. I removed JSDoc comments from source files.

overmyheadandbody (commented):
Thanks for implementing the changes! One final comment from my end - I see the PSI score is quite low. It may not be related to the changes suggested in this PR, but it's definitely worth investigating. We aim for 100 LH points, but most PRs are upwards of 90. This PR's score is lower than 50

tamanyan:
@overmyheadandbody The score is low because of Japanese Adobe Clean Han font loading as far as I know. Please look at this message https://github.com/adobecom/milo/pull/405#issuecomment-1408161987

tamanyan:
@overmyheadandbody @sharg1 Hi, what should I do to proceed with this pull request?

overmyheadandbody (approved):

mokimo (approved):

mokimo:
@tamanyan Since you have no dedicated QA resource, would be great if you can give it one more round of testing and leave a quick comment once you are done. 

Since this only affects `ja-JP` we should be good to go ahead then and I'll merge this afterwards.

tamanyan:
@mokimo Thank you. I have tested on Chrome, Firefox and Safari and make sure no error occurring.

sharg1:
> @overmyheadandbody @sharg1 Hi, what should I do to proceed with this pull request?

PR LGTM.. I don't seem to have access to add labels or approve PRs yet. I'll lean on @overmyheadandbody @chrischrischris @narcis-radu @mokimo to help here with reviews, adding labels and approvals.

PR #1776, opened 2024-01-23
MWPW-141639 Quick follow to do an early return if mep=off

vgoodric:
Got a suggestion I liked right after merging. Switching to doing an early return if mep is disabled with mep=off flag

Resolves: [MWPW-141639](https://jira.corp.adobe.com/browse/MWPW-141639)

Test URLs:
Before: https://main--milo--adobecom.hlx.page/drafts/vgoodrich/fragments/mepmartechoff/?mep=off
After: https://mepoffearlyreturn--milo--adobecom.hlx.page/drafts/vgoodrich/fragments/mepmartechoff/?mep=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -829,12 +829,16 @@ async function loadMartech({ persEnabled = false, persManifests = [] } = {}) {
 }
 
 async function checkForPageMods() {
+  const search = new URLSearchParams(window.location.search);
+  const offFlag = (val) => search.get(val) === 'off';
+  if (offFlag('mep')) {
+    document.body.dataset.mep = 'nopzn|nopzn';
+    return;
+  }
   const persMd = getMetadata('personalization');
   const promoMd = getMetadata('manifestnames');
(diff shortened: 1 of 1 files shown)


chrischrischris (approved):

sukamat (approved):

libs/utils/utils.js
```diff
@@ -829,12 +829,16 @@ async function loadMartech({ persEnabled = false, persManifests = [] } = {}) {
 }
 
 async function checkForPageMods() {
+  const search = new URLSearchParams(window.location.search);
+  const offFlag = (val) => search.get(val) === 'off';
+  if (offFlag('mep')) {
+    document.body.dataset.mep = 'nopzn|nopzn';
```
overmyheadandbody, line 835: I see this is used again in an identical way later in the file. If another such use-case arises, it might be worth having a small method for it to be re-used everywhere

overmyheadandbody (approved):

3ch023 (approved):

PR #1781, opened 2024-01-24
MWPW-141638: Fix for commerce modals event listeners

mirafedas:
**How it works now:**
In the current implementation, when the commerce modal is open, we establish a message event listener to address two scenarios:
- If the page within the iFrame cannot detect its viewport width due to CORS restrictions, it may request the viewport width from the parent page.
- The page within the iFrame measures its content height and sends this information to adjust the modal height, eliminating any blank space at the bottom.

Upon closing the commerce modal, we remove the message event listener using the abortController.
Additionally, the dynamic import of the debounce function has been optimized by relocating it outside the function that is invoked within the event listener, ensuring that the import is executed only once. The debouncing time decreased from 50 to 10 to avoid the modal getting stuck when the resize is done quickly.

**How it worked before:**
Previously, the message event listener was set during page load, and the function called upon that event retained the context of the initially opened modal window. This led to an issue where, upon closing one modal and opening another, the height adjustment would fail because the event listener function attempted to adjust the height of the previously opened modal. The current enhancements address and resolve this limitation.

**Resolves:** [MWPW-141638](https://jira.corp.adobe.com/browse/MWPW-141638)

**Test URLs:**
- After: https://mwpw-141638-modal-height-multiple--milo--mirafedas.hlx.page/drafts/mirafedas/twp

Note for testers:
To verify the modals on my test page you would need to set the frame-ancestors in ModHeaders plugin in the CSP response modifiers to asterisk (see screenshot)
Screenshot 2024-02-06 at 18 45 05

### DIFF
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -9,9 +9,10 @@ const CLOSE_ICON = `<svg xmlns="http://www.w3.org/2000/svg" width="20" height="2
     <line x1="8" y1="8" transform="translate(10506 -3397)" fill="none" stroke="#fff" stroke-width="2"/>
   </g>
 </svg>`;
-let isInitialPageLoad = true;
 const MOBILE_MAX = 599;
 const TABLET_MAX = 1199;
+let messageAbortController;
+let resizeAbortController;
 
 export function findDetails(hash, el) {
   const id = hash.replace('#', '');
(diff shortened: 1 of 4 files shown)


libs/blocks/modal/modal.js
```diff
@@ -202,15 +199,17 @@ export async function getModal(details, custom) {
...
+    const { debounce } = await import('../../utils/action.js');
+    messageAbortController = new AbortController();
+    window.addEventListener('message', (messageInfo) => {
+      if (dialog.classList.contains('height-fit-content')) {
+        adjustModalHeight({ contentHeight: messageInfo?.data?.contentHeight, dialog });
+      }
+      /* If the page inside iFrame comes from another domain, it won't be able to retrieve
+      the viewport dimensions, so it sends a request to receive the viewport dimensions
+      from the parent window. */
+      sendViewportDimensionsOnRequest({ debounce, messageInfo });
```
vladen, line 211: If the user opens first modal, then closes it and opens another one, won't this result in multiple active `resize` listeners in the parent window?

yesil (approved):

vladen (approved):

Axelcureno (approved):

PR #1795, opened 2024-01-26
MWPW-141810: updated custom ietf locale logic for unav

robert-bogos:
## Description
This PR updates the custom ietf locale logic for the universal navigation to match the [mapping received from the Universal Navigation team.](https://git.corp.adobe.com/wcms/hawks/pull/3080/files)

## Related Issue
Resolves: [MWPW-141810](https://jira.corp.adobe.com/browse/MWPW-141810)

## Test URLs
**Milo:**
- Before:  https://gnav--milo--adobecom.hlx.page/drafts/rbogos/unav?martech=off
- After: https://mwpw-141810-update-locales-unav--milo--robert-bogos.hlx.page/drafts/rbogos/unav?martech=off

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -204,27 +204,29 @@ const closeOnClickOutside = (e) => {
   }
 };
 
-const getUniversalNavLocale = () => {
-  const { locale: { prefix: configPrefix = '', ietf = '' } = {} } = getConfig() || {};
-  const prefix = configPrefix.replace('/', '');
+const getUniversalNavLocale = (locale) => {
+  const prefix = locale.prefix.replace('/', '');
+  const { ietf } = locale;
   const DEFAULTLANG = {
-    en_US: ['ae_en', 'africa', 'au', 'be_en', 'ca', 'cis_en', 'cy_en', 'eg_en', 'gr_en', 'hk_en', 'id_en', 'ie', 'il_en', 'in', 'kw_en', 'langstore', 'lu_en', 'mena_en', 'mt', 'my_en', 'ng', 'nz', 'ph_en', 'qa_en', 'sa_en', 'sg', 'th_en', 'uk', 'vn_en', 'za'],
-    fr_FR: ['be_fr', 'ca_fr', 'ch_fr', 'lu_fr'],
     de_DE: ['at', 'ch_de', 'lu_de'],
+    en_US: ['ae_en', 'africa', 'au', 'be_en', 'ca', 'cis_en', 'cy_en', 'eg_en', 'gr_en', 'hk_en', 'id_en', 'ie', 'il_en', 'in', 'kw_en', 'langstore', 'lu_en', 'mena_en', 'mt', 'my_en', 'ng', 'nz', 'ph_en', 'qa_en', 'sa_en', 'sg', 'th_en', 'uk', 'vn_en', 'za'],
     es_ES: ['ar', 'ar_es', 'cl', 'co', 'cr', 'ec', 'gt', 'la', 'mx', 'pe', 'pr'],
+    fr_FR: ['be_fr', 'ca_fr', 'ch_fr', 'lu_fr'],
   };
   const CUSTOMLANG = {
     be_nl: 'nl_NL',
     ch_it: 'it_IT',
+    cis_ru: 'ru_RU',
     hk_zh: 'zh_TW',
     no: 'nb_NO',
-    cis_ru: 'ru_RU',
   };
   const prefixParts = prefix.split('_').reverse();
 
   return Object.keys(DEFAULTLANG).find((key) => DEFAULTLANG[key].includes(prefix))
          || CUSTOMLANG[prefix]
-         || (ietf.includes('-') ? ietf.replace('-', '_') : `${prefixParts[0].toLowerCase()}_${prefixParts[1].toUpperCase()}`)
+         || (ietf.includes('-')
+           ? ietf.replace('-', '_')
+           : prefixParts.length === 2 && `${prefixParts[0].toLowerCase()}_${prefixParts[1].toUpperCase()}`)
          || 'en_US';
 };
 
@@ -457,7 +459,7 @@ class Gnav {
 
   decorateUniversalNav = async () => {
     const config = getConfig();
-    const locale = getUniversalNavLocale();
+    const locale = getUniversalNavLocale(config.locale);
     const environment = config.env.name === 'prod' ? 'prod' : 'stage';
     const visitorGuid = window.alloy ? await window.alloy('getIdentity').then((data) => data?.identity?.ECID) : undefined;
     const getDevice = () => {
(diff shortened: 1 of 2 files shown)


overmyheadandbody (commented):
Even if we're merging this to a feature branch and the change is trivial, I still think it's useful to follow the PR guidelines, add test links for PSI purposes and labels. The PR checks should be green for all PRs

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -204,9 +204,28 @@ const closeOnClickOutside = (e) => {
   }
 };
 
-const getIetfLocale = (ietfLocale) => {
-  const nonStandardLocaleMap = { no_NO: 'nb_NO' };
-  return nonStandardLocaleMap[ietfLocale] || ietfLocale;
+const getUniversalNavLocale = () => {
+  const { locale: { prefix: configPrefix = '', ietf = '' } = {} } = getConfig() || {};
+  const prefix = configPrefix.replace('/', '');
```
overmyheadandbody, line 209: Are there consumers using slashes in their prefixes?

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -204,9 +204,28 @@ const closeOnClickOutside = (e) => {
...
+    fr_FR: ['be_fr', 'ca_fr', 'ch_fr', 'lu_fr'],
+    de_DE: ['at', 'ch_de', 'lu_de'],
+    es_ES: ['ar', 'ar_es', 'cl', 'co', 'cr', 'ec', 'gt', 'la', 'mx', 'pe', 'pr'],
+  };
+  const CUSTOMLANG = {
+    be_nl: 'nl_NL',
+    ch_it: 'it_IT',
+    hk_zh: 'zh_TW',
+    no: 'nb_NO',
+    cis_ru: 'ru_RU',
```
overmyheadandbody, line 221: I'd vote we keep the keys in alphabetical order

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -204,9 +204,28 @@ const closeOnClickOutside = (e) => {
...
+  };
+  const prefixParts = prefix.split('_').reverse();
+
+  return Object.keys(DEFAULTLANG).find((key) => DEFAULTLANG[key].includes(prefix))
+         || CUSTOMLANG[prefix]
+         || (ietf.includes('-') ? ietf.replace('-', '_') : `${prefixParts[0].toLowerCase()}_${prefixParts[1].toUpperCase()}`)
```
overmyheadandbody, line 227: Are we certain their are no prefixes without a separator? Such as `africa`, `sea`, etc.? In that case we'd have just one prefix part so we might get unexpected results.

libs/scripts/scripts.js
```diff
@@ -123,7 +123,7 @@ const config = {
   geoRouting: 'on',
   fallbackRouting: 'on',
   links: 'on',
-  imsClientId: 'milo',
+  imsClientId: 'fedsmilo',
```
overmyheadandbody, line 126: As usual, let's not forget to remove this before merging

libs/blocks/global-navigation/global-navigation.js
narcis-radu: no, but that's how the prefix is returned. similar code is already on this branch https://github.com/adobecom/milo/blob/gnav/libs/blocks/global-navigation/global-navigation.js#L446

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -204,9 +204,28 @@ const closeOnClickOutside = (e) => {
   }
 };
 
-const getIetfLocale = (ietfLocale) => {
-  const nonStandardLocaleMap = { no_NO: 'nb_NO' };
-  return nonStandardLocaleMap[ietfLocale] || ietfLocale;
+const getUniversalNavLocale = () => {
+  const { locale: { prefix: configPrefix = '', ietf = '' } = {} } = getConfig() || {};
```
narcis-radu, line 208: I don't think this is needed, we should be good with `{ locale } = getConfig();`. You can also use locale as an argument, it should help when calling the method since you already have the config object.

libs/blocks/global-navigation/global-navigation.js
narcis-radu: agree

libs/blocks/global-navigation/global-navigation.js
narcis-radu: `africa` is mapped to `en`, but for `sea` it would be a problem

narcis-radu (changes requested):
see comments

libs/blocks/global-navigation/global-navigation.js
robert-bogos: Agree. Made the change 👍

libs/blocks/global-navigation/global-navigation.js
robert-bogos: We don't have mapping for `sea`, but I added a fix to handle that situation 👍

libs/blocks/global-navigation/global-navigation.js
robert-bogos: Made the change 👍

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -204,9 +204,30 @@ const closeOnClickOutside = (e) => {
...
 
-const getIetfLocale = (ietfLocale) => {
-  const nonStandardLocaleMap = { no_NO: 'nb_NO' };
-  return nonStandardLocaleMap[ietfLocale] || ietfLocale;
+const getUniversalNavLocale = (locale) => {
+  const prefix = locale.prefix.replace('/', '');
+  const { ietf } = locale;
+  const DEFAULTLANG = {
+    de_DE: ['at', 'ch_de', 'lu_de'],
+    en_US: ['ae_en', 'africa', 'au', 'be_en', 'ca', 'cis_en', 'cy_en', 'eg_en', 'gr_en', 'hk_en', 'id_en', 'ie', 'il_en', 'in', 'kw_en', 'langstore', 'lu_en', 'mena_en', 'mt', 'my_en', 'ng', 'nz', 'ph_en', 'qa_en', 'sa_en', 'sg', 'th_en', 'uk', 'vn_en', 'za'],
```
overmyheadandbody, line 212: This looks like it might have a line length larger than 100, ESLint might complain about it

overmyheadandbody (approved):
Let's not forget to exclude `scripts.js` before merging so we leave the IMS config intact

libs/blocks/global-navigation/global-navigation.js
robert-bogos: Fixed, added multiple lines in there 👍

narcis-radu:
@robert-bogos - looks good to me, please update the client ID

robert-bogos:
IMS client ip updated 👍

libs/scripts/scripts.js
robert-bogos: Done 👍

PR #1796, opened 2024-01-26
MWPW-138544 [merch-card] storage options

npeltier:
- horizontal options when text is storage,
- support for 2 CTAs (at the moment depending on the number of CTAs authored),,
- if a price placeholder is there, it's used as "slot" by merch-offer-select,
- renamed createQuantitySelect to extractQuantitySelect as it actually can do nothing if no option, and it removes UL if there is,
- unit tests

Resolves: [MWPW-138544](https://jira.corp.adobe.com/browse/MWPW-138544)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/npeltier/photography?martech=off
- After: https://mwpw-138544--milo--npeltier.hlx.page/drafts/npeltier/photography?martech=off


libs/blocks/merch-card/merch-card.js
```diff
@@ -428,8 +428,8 @@ const init = async (el) => {
   }
   merchCard.appendChild(footer);
 
-  if (cardType === 'plans') {
-    const quantitySelect = createQuantitySelect(el);
+  if (['plans', 'product'].includes(cardType)) {
```
yesil, line 431: make of `['plans', 'product']` a constant?

libs/blocks/merch-card/merch-offer-select.js
```diff
@@ -3,12 +3,17 @@ import { decorateButtons } from '../../utils/decorate.js';
 import '../../deps/merch-offer-select.js';
 
 function createDynamicSlots(el, bodySlot) {
-  const price = createTag('h5', { class: 'merch-card-price' });
-  price.append(createTag('span', { slot: 'price', is: 'inline-price' }));
-  bodySlot.append(price);
-
+  const pricePlaceholder = el.querySelector("span[is='inline-price']");
+  if (pricePlaceholder) {
+    pricePlaceholder.parentNode.replaceChild(createTag('span', { slot: 'price', is: 'inline-price' }), pricePlaceholder);
```
yesil, line 8: I don't understand why this is needed? creating a `span[is='inline-price']` without merch block seems wrong to me.

libs/blocks/merch-card/merch-offer-select.js
```diff
@@ -3,12 +3,17 @@ import { decorateButtons } from '../../utils/decorate.js';
...
+  } else {
+    const priceSlot = createTag('h5', { class: 'merch-card-price' });
+    createTag('span', { slot: 'price', is: 'inline-price' }, null, { parent: priceSlot });
+    bodySlot.append(priceSlot);
+  }
   const p = createTag('p', { class: 'action-area' });
-  p.append(createTag('a', { slot: 'cta', is: 'checkout-link' }));
+  createTag('a', { slot: 'secondary-cta', is: 'checkout-link' }, null, { parent: p });
```
yesil, line 15: same here, `createTag('a', { slot: 'secondary-cta', is: 'checkout-link' }` gives the impression that you create a checkout link but it is not because it must be created via `merch` autoblock.

libs/blocks/merch-card/merch-offer-select.js
```diff
@@ -42,6 +49,18 @@ export const initOfferSelection = (merchCard, offerSelection, quantitySelector)
     merchOffers.append(createMerchOffer(option, quantitySelector));
   });
   merchOffers.querySelectorAll('a[is="checkout-link"]').forEach((link) => { link.setAttribute('slot', 'cta'); });
+  if (isHorizontal(merchOffers)) {
+    merchOffers.setAttribute('variant', 'horizontal');
+  }
+  merchOffers.querySelectorAll('merch-offer').forEach((offer) => {
+    const links = offer.querySelectorAll('a[is="checkout-link"]');
+    if (links.length > 1) {
+      links[0].setAttribute('slot', 'secondary-cta');
```
yesil, line 58: I thought we didn't need an explicit slot named `secondary-cta`

the problem I see is that, authors can add a merch link as primary or secondary CTA in the order they want, but here it is assumed that the first link is secondary, and the second links is primary.

I think we should be less opinionated and let authors stack CTAs in the order they want.

npeltier: can this happen?

libs/blocks/merch-card/merch-card.js
npeltier: OFFER_SELECT_CARD_TYPES ?

libs/blocks/merch-card/merch-offer-select.js
yesil: I think yes if my understanding of the above code is correct.

npeltier: well i guess i can indeed "just" dub the links in the order they came, how they came under the cta "slot"

npeltier: i agree with you, but for now it's complicated to change this as target placeholder/slot is supposed to be a link. As discussed offline we should quickly make this better, and have that solution in the meanwhile

npeltier: as discussed, we'll work on related technical debt (i'll attach it to the actual ticket) asap, but this is complicated to change now

libs/blocks/merch-card/merch-card.js
3ch023: or MULTI_OFFER_CARDS

libs/blocks/merch-card/merch-offer-select.js
3ch023: we can probably remove `is: 'inline-price' ` part. and only keep ` slot: 'price'`
and as discussed offline, maybe it shouldn't be called slot at all since its not used as one.. but that can be refactored later.

3ch023 (approved):

yesil (approved):

afmicka:
@npeltier when you toggle from 1TB back to 20GB, the Free Trial CTA does not get removed as per setup. 20 GB option has one CTA configured. 
Screenshot 2024-01-29 at 12 59 01
Screenshot 2024-01-29 at 12 59 07

npeltier:
@afmicka as discussed on ticket for now "hybrid" cases with one CTA / two options CTA is not supported, pending on technical debt ticket discussed above. If that's really needed now, i'll add a fix.

afmicka:
> @afmicka as discussed on ticket for now "hybrid" cases with one CTA / two options CTA is not supported, pending on technical debt ticket discussed above. If that's really needed now, i'll add a fix.

thanks @npeltier. The rest looks ok to me. Will verify PR. Could you please add 'run-nala' label so that more merch-card tests get executed?

libs/blocks/merch-card/merch-card.js
```diff
@@ -105,6 +105,8 @@ const TAG_PATTERN = /^[a-zA-Z0-9_-]+:[a-zA-Z0-9_-]+\/[a-zA-Z0-9_-].*$/;
 
 const CARD_TYPES = ['segment', 'special-offers', 'plans', 'catalog', 'product', 'inline-heading', 'image', 'mini-compare-chart'];
 
+const MULTI_OFFER_CARDS = ['plans', 'product'];
```
Axelcureno, line 108: we can add `MINI_COMPARE_CHART` here

PR #1797, opened 2024-01-26
MWPW-141691: Merch card: align footer rows

yesil:
Resolves: [MWPW-141691](https://jira.corp.adobe.com/browse/MWPW-141691)

Includes:
* https://git.corp.adobe.com/wcms/tacocat.js/pull/520

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/axel/mini-compare-chart?martech=off
- After: https://mwpw-141691--milo--yesil.hlx.page/drafts/axel/mini-compare-chart?martech=off

**CC**
https://main--cc--adobecom.hlx.page/drafts/axel/mini-compare-chart?milolibs=mwpw-141691--milo--yesil

### DIFF
--- a/libs/blocks/merch-card/merch-card.js
+++ b/libs/blocks/merch-card/merch-card.js
@@ -108,6 +108,14 @@ const CARD_TYPES = ['segment', 'special-offers', 'plans', 'catalog', 'product', 
 const MINI_COMPARE_CHART = 'mini-compare-chart';
 
 const MULTI_OFFER_CARDS = ['plans', 'product', MINI_COMPARE_CHART];
+// Force cards to refresh once they become visible so that the footer rows are properly aligned.
+const sectionObserver = new MutationObserver((mutations) => {
+  mutations.forEach((mutation) => {
+    if (mutation.type === 'attributes' && mutation.attributeName === 'data-status') {
+      [...mutation.target.querySelectorAll('merch-card')].forEach((card) => card.requestUpdate());
+    }
+  });
+});
 
 const textStyles = {
   H5: 'detail-m',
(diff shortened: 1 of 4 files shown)


Axelcureno (approved):

VKniaz (approved):

libs/blocks/merch-card/merch-card.js
```diff
@@ -107,6 +107,15 @@ const CARD_TYPES = ['segment', 'special-offers', 'plans', 'catalog', 'product',
 
 const MINI_COMPARE_CHART = 'mini-compare-chart';
 
+// Force cards to refresh once they become visible so that the footer rows are properly aligned.
+const sectionObserver = new MutationObserver((mutations) => {
+  mutations.forEach((mutation) => {
+    if (mutation.type === 'attributes' && mutation.attributeName === 'data-status') {
+      [...mutation.target.querySelectorAll('merch-card')].forEach((card) => card.requestUpdate());
+    }
```
overmyheadandbody, line 115: Should the MutationObserver be disconnected after executing relevant logic?

libs/blocks/merch-card/merch-card.js
```diff
@@ -353,6 +362,8 @@ const init = async (el) => {
   }
   let footerRows;
   if (cardType === MINI_COMPARE_CHART) {
+    const container = el.closest('[data-status="decorated"]');
```
overmyheadandbody, line 365: Any chance this doesn't exist? If so, it will cause errors on the next line, a check should be put in place

libs/blocks/merch-card/merch-card.js
yesil: @overmyheadandbody it is a shared instance for all the `merch-card`s.  
There is not a right moment to call disconnect method as it might break pending observartions.
Since we explicitly check against the `data-status` attribute at the line `113`,  I think it is safe to assume that the line 114 be invoked only once per merch-card.

PR #1802, opened 2024-01-29
Bulk Publish Tool v2

Sartxi:
- Using new [bulk api endpoints](https://www.aem.live/docs/admin.html) - now all page URLs are included in a single request  
- Interfaces with AEM Sidekick to provide users permissions for each respective process
- Job Resume - resumes unfinished jobs if window is closed and/or reopened
- UI/UX updates & tools for managing errors and malformed page URLs
- New styling and layout options for a fresh new look and ease of use
- Uses `LitElement` web components to manage component lifecycle and state

Note: I renamed the old bulk publish tool to `bulk-publish-v1` to keep that around as a fallback, but I can delete or rename them if that is preferable.

Resolves: [MWPW-140283](https://jira.corp.adobe.com/browse/MWPW-140283)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/tools/bulk-publish?martech=off
- After: https://sarchibeque-bulk-publish-v2--milo--adobecom.hlx.page/tools/bulk-publish?martech=off


libs/blocks/bulk-publish-v2/components/bulk-publisher.js
```diff
@@ -0,0 +1,419 @@
...
+        @click=${handleToggle}>
+        <span class="title">
+          ${count ? html`<strong>${count}</strong>` : ''}
+          Job Results
+        </span>
+        <div class="jobs-tools${showList}">
+          <div 
+            class="clear-jobs${showClear}"
+            @click=${this.clearJobs}>
+            <img src=${clearJobsIcon} alt="Clear Job List" />
```
ryanmparrish, line 288:
```suggestion
            <img src=${clearJobsIcon} alt="Clear list icon" title="Clear Job List" />
```
nit: accessibility - Title attribute for tooltip on hover.

rgclayton (approved):

vhargrave (approved):
This looks great. Gave it a small spin and am looking forward for when we get to use it in Express. Nice one! :)

vhargrave (changes requested):
darn it, one test is failing. Can you fix and then I'll give an approval ?

Sartxi:
> darn it, one test is failing. Can you fix and then I'll give an approval ?

Thanks for the review Victor! Yeah I was in the process of fixing that. Fixed and got full testing code cov!

ryanmparrish (approved):
Amazing work. Nice job on this! 
I don't see an entry in the `CODEOWNERS` file, so you might want to add a line... lookin great! 👍

vhargrave (approved):

vhargrave:
@Sartxi one more thing, you'll want to point this towards stage now. No longer at main since the stage change that Blaine has been working on.

libs/blocks/bulk-publish-v2/utils.js
```diff
@@ -0,0 +1,129 @@
...
+  }
+  return { code, text, color };
+};
+
+const displayDate = (newDate) => {
+  const date = new Date(newDate);
+  const today = new Date();
+  const isToday = date.getDate() === today.getDate()
+    && date.getMonth() === today.getMonth()
+    && date.getFullYear() === today.getFullYear();
```
elan-tbx, line 105: You could probably simplify this comparison using [`toDateString`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toDateString), something like this:

```
const isToday = date.toDateString() === today.toDateString();
```

test/blocks/bulk-publish-v2/bulk-publish-v2.test.js
```diff
@@ -0,0 +1,184 @@
...
+import { setConfig } from '../../../libs/utils/utils.js';
+import './mocks/authentication.js';
+
+const conf = { miloLibs: 'http://localhost:2000/libs' };
+setConfig(conf);
+
+document.body.innerHTML = await readFile({ path: './mocks/body.html' });
+const { default: init } = await import('../../../libs/blocks/bulk-publish-v2/bulk-publish-v2.js');
+
+const testPage = 'https://main--milo--adobecom.hlx.page/drafts/sarchibeque/bulk-publish-test';
```
elan-tbx, line 14: Is this the right place for this to live on SharePoint? Feels like it fits better in `/tools/`, but that's just my opinion

libs/blocks/bulk-publish-v2/components/job-process.css
```diff
@@ -0,0 +1,93 @@
...
+  animation: dotty steps(1, end) 1s infinite;
+  content: '';
+}
+
+@keyframes dotty {
+  0% { content: '' }
+  25% { content: '.' }
+  50% { content: '..' }
+  75% { content: '...' }
+  100% { content: '' }
```
elan-tbx, line 92: small nit: if the `0%` and `100%` values are exactly the same and the keyframes are only ever used in an infinitely-repeating animation, you can remove the `100%` rule as it does the exact same thing as the `0%` rule, at (perceptually) the exact same time.

elan-tbx (approved):
Left a couple comments on some small things, but this is looking great. Excellent work.

test/blocks/bulk-publish-v2/bulk-publish-v2.test.js
Sartxi: oh yeah good point, I will move it there

libs/blocks/bulk-publish-v2/utils.js
Sartxi: I will try this 👍🏻

mokimo:
@Sartxi is there anyone who can QA this and add the `verified` label? Otherwise you can also just add the trivial label, since this is a tool and not consumer facing, that should be fine IMO

PR #1803, opened 2024-01-30
MWPW-140229: Sync Universal Nav analytics from Dunamis to AA

robert-bogos:
## Description
This PR syncs the analytics from Dunamis to AA

## Related Issue
Resolves: [MWPW-140229](https://jira.corp.adobe.com/browse/MWPW-140229)

## Testing instructions
1. click on any element from the universal navigation;
2. go to the network tab in dev tools;
3. search for 'collect' and you'll find the sent tag inside the payload, as you can see in the screenshot below;
Screenshot 2024-01-30 at 14 28 51

The sent tags can be verified against https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=marketingtech&title=Universal+Nav+on+Adobe.com+-+Custom+Tracking

## Test URLs
**Milo:**
- Before: https://gnav--milo--adobecom.hlx.page/drafts/rbogos/unav?martech=off
- After: https://mwpw-140229-sync-unav-analytics--milo--robert-bogos.hlx.page/drafts/rbogos/unav?martech=off

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -524,7 +524,9 @@ class Gnav {
                 data.content?.name?.split('-').map((str) => str.charAt(0).toUpperCase() + str.slice(1)).join(' ')
               }`,
               footer: {
-                // two of these will be removed in the future, when the unav team updates the data model
+                // two of these will be removed in the future,
+                // when the unav team updates the data model for consistency
+                // between logged in and logged out states
                 'all-apps': 'AppLauncher.allapps',
                 'adobe-dot-com': 'AppLauncher.adobe.com',
                 'adobe-home': 'AppLauncher.adobe.com',
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -524,8 +524,8 @@ class Gnav {
                 data.content?.name?.split('-').map((str) => str.charAt(0).toUpperCase() + str.slice(1)).join(' ')
               }`,
               footer: {
-                'adobe-home': 'AppLauncher.adobe.com',
-                'all-apps': 'AppLauncher.allapps',
+                'adobe-dot-com': 'AppLauncher.adobe.com',
+                'see-all-apps': 'AppLauncher.allapps',
               },
             },
             render: { component: 'AppLauncher.appIconToggle' },
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -538,9 +538,6 @@ class Gnav {
       const interaction = getInteraction();
 
       if (!interaction) return;
-      // __ added for testing purposes, will be removed before merging
-      console.log('interaction -->', interaction);
-      // _____________________________________________________________
       // eslint-disable-next-line no-underscore-dangle
       window._satellite?.track('event', {
         xdm: {},
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -538,6 +538,9 @@ class Gnav {
       const interaction = getInteraction();
 
       if (!interaction) return;
+      // __ added for testing purposes, will be removed before merging
+      console.log('interaction -->', interaction);
+      // _____________________________________________________________
       // eslint-disable-next-line no-underscore-dangle
       window._satellite?.track('event', {
         xdm: {},
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -236,6 +236,11 @@ const getUniversalNavLocale = (locale) => {
          || 'en_US';
 };
 
+const convertToPascalCase = (str) => str
+  ?.split('-')
+  .map((s) => s.charAt(0).toUpperCase() + s.slice(1))
+  .join(' ');
+
 class Gnav {
   constructor(body, el) {
     this.blocks = {
@@ -468,6 +473,8 @@ class Gnav {
     const locale = getUniversalNavLocale(config.locale);
     const environment = config.env.name === 'prod' ? 'prod' : 'stage';
     const visitorGuid = window.alloy ? await window.alloy('getIdentity').then((data) => data?.identity?.ECID) : undefined;
+    const experienceName = getExperienceName();
+
     const getDevice = () => {
       const agent = navigator.userAgent;
       if (agent.includes('Mac')) return 'macOS';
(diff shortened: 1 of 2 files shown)


libs/blocks/global-navigation/global-navigation.js
```diff
@@ -505,8 +505,44 @@ class Gnav {
...
+                'adobe-home': 'AppLauncher.adobe.com',
+                'all-apps': 'AppLauncher.allaps',
+              },
+            },
+            render: { component: 'AppLauncher.appIconToggle' },
+          },
+        };
+        const interactionResult = interactionMapping[name]?.[type]?.[subtype];
+
+        return (interactionResult && interactionResult[data.content?.name]) || interactionResult;
```
narcis-radu, line 536: I think this is problematic - when I click on "see all apps", interactionResult is an object with
```
adobe-home : "AppLauncher.adobe.com"
all-apps: "AppLauncher.allaps"
```

narcis-radu (changes requested):

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -505,8 +505,44 @@ class Gnav {
...
+            render: { component: `Account|gnav|${getExperienceName()}` },
+          },
+          'app-switcher': {
+            click: {
+              app: `AppLauncher.appClick.${
+                data.content?.name?.split('-').map((str) => str.charAt(0).toUpperCase() + str.slice(1)).join(' ')
+              }`,
+              footer: {
+                'adobe-home': 'AppLauncher.adobe.com',
+                'all-apps': 'AppLauncher.allaps',
```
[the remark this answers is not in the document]
robert-bogos: fixed 👍

libs/blocks/global-navigation/global-navigation.js
robert-bogos: It looks like on my local setup I receive different values for `data.content?.name` which are used for the mapping. I was able to reproduce it on incognito and made the mapping adjustment. Now it should work properly 👍

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -505,8 +505,49 @@ class Gnav {
     };
 
     const onAnalyticsEvent = (data) => {
-      // TODO: update based on requirements
-      console.log('analytics data', data);
+      if (!data) return;
+
+      const getInteraction = () => {
+        const { event: { type, subtype } = {}, source: { name } = {} } = data;
+        const interactionMapping = {
```
mokimo, line 512: Since this is a mostly static Dunamis->AA mapping, I'd move it to the top of the file.

Since the app switcher click mapping has a bit more logic, you could do something such as:
```js
const interactionMapping = {
  ...analyticsMapping,
  'app-switcher': {
    ...analyticsMapping['app-switcher'],
    click: {
      app: getAppLauncherClickAnalyticsName(data)
    }
  }
}
```

If you define a `getAppLauncherClickAnalyticsName` function you could also export it and write a few unit tests for it to make sure the method works in all cases with various invalid input

narcis-radu: @mokimo - we have a dedicated story for unit testing, the whole implementation would need them.

narcis-radu (approved):

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -505,8 +505,46 @@ class Gnav {
...
+          case 'profile|render|component|':
+            return `Account|gnav|${getExperienceName()}`;
+          case 'profile|click|account|':
+            return `View Account|gnav|${getExperienceName()}`;
+          case 'profile|click|sign-out|':
+            return `Sign out|gnav|${getExperienceName()}|unav`;
+          case 'app-switcher|render|component|':
+            return 'AppLauncher.appIconToggle';
+          case `app-switcher|click|app|${contentName}`:
+            return `AppLauncher.appClick.${contentName?.split('-')
```
mokimo, line 525: I'd still put this in a separate function
```
function convertToPascalCase(str) {
  return str
    .split('-')
    .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
    .join(' ');
}
...
app: `AppLauncher.appClick.${convertToPascalCase(data.content?.name || '')}`,
```

mokimo (approved):

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -505,8 +505,46 @@ class Gnav {
     };
 
     const onAnalyticsEvent = (data) => {
-      // TODO: update based on requirements
-      console.log('analytics data', data);
+      if (!data) return;
+
+      const getInteraction = () => {
+        const { event: { type, subtype } = {}, source: { name } = {} } = data;
+        const contentName = data.content?.name;
```
overmyheadandbody, line 512: Could you not destructure `contentName` as well? Something like `{ event: { type, subtype } = {}, source: { name } = {}, content: { name: contentName } = {} } = data;`

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -505,8 +505,46 @@ class Gnav {
...
+        switch (`${name}|${type}|${subtype}|${contentName || ''}`) {
+          case 'profile|click|sign-in|':
+            return `Sign in|gnav|${getExperienceName()}|unav`;
+          case 'profile|render|component|':
+            return `Account|gnav|${getExperienceName()}`;
+          case 'profile|click|account|':
+            return `View Account|gnav|${getExperienceName()}`;
+          case 'profile|click|sign-out|':
+            return `Sign out|gnav|${getExperienceName()}|unav`;
```
overmyheadandbody, line 521: Maybe it's a micro-optimisation, but: could we cache `getExperienceName()` in a variable outside this method so it's not called for each analytics event? It shouldn't be an expensive operation, but it's easy enough to avoid multiple calls to it.

overmyheadandbody (approved):

libs/blocks/global-navigation/global-navigation.js
robert-bogos: Done 👍 Moved it to a separate function. Thanks for the suggestion 👍

libs/blocks/global-navigation/global-navigation.js
robert-bogos: Makes sense. Made the change 👍

libs/blocks/global-navigation/global-navigation.js
robert-bogos: Nice optimisation, made the change 👍 It seems so obvious after you pointed it out

PR #1806, opened 2024-01-30
MWPW-141594 Move georouting to post LCP

meganthecoder:
* Moves georouting to post-LCP event for performance improvements
* Meaningfully improves georouting pages with low lighthouse scores due to LCP
* Has little/no change on georouting pages with high lighthouse scores
* Has little/no change to pages without georouting
* May increase CLS in some cases on georouting pages but has overall positive effect on lighthouse score because of LCP improvement

Resolves: [MWPW-141594](https://jira.corp.adobe.com/browse/MWPW-141594)

**Bacom FR Homepage:**
Lighthouse score: 79 -> 94
LCP: 5.6s -> 3.0s
Screenshot 2024-01-29 at 5 07 37 PM
Screenshot 2024-01-29 at 5 07 46 PM

**CC Business Card Design Page:**
Lighthouse score: 75 -> 98
LCP: 1.9s -> 1.7s
Screenshot 2024-01-29 at 4 54 07 PM
Screenshot 2024-01-29 at 4 54 17 PM

**Test URLs:**
(Copied from Bacom AEM sites product page)
- Before: https://main--milo--adobecom.hlx.page/de/drafts/methomas/aem-sites?martech=off
- After: https://geo-move--milo--adobecom.hlx.page/de/drafts/methomas/aem-sites?martech=off

Lighthouse score: 90 -> 90
LCP: 1.3s -> 1.4s 
Screenshot 2024-01-30 at 9 58 45 AM
Screenshot 2024-01-30 at 9 59 25 AM

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -896,6 +896,11 @@ async function checkForPageMods() {
 }
 
 async function loadPostLCP(config) {
+  const georouting = getMetadata('georouting') || config.geoRouting;
+  if (georouting === 'on') {
+    const { default: loadGeoRouting } = await import('../features/georoutingv2/georoutingv2.js');
+    await loadGeoRouting(config, createTag, getMetadata, loadBlock, loadStyle);
+  }
   loadMartech();
   const header = document.querySelector('header');
   if (header) {
@@ -1005,13 +1010,6 @@ function decorateDocumentExtras() {
 }
 
 async function documentPostSectionLoading(config) {
-  const georouting = getMetadata('georouting') || config.geoRouting;
-  if (georouting === 'on') {
-    // eslint-disable-next-line import/no-cycle
-    const { default: loadGeoRouting } = await import('../features/georoutingv2/georoutingv2.js');
-    loadGeoRouting(config, createTag, getMetadata, loadBlock, loadStyle);
-  }
-
   decorateFooterPromo();
 
   const appendage = getMetadata('title-append');
@@ -1067,12 +1065,13 @@ async function processSection(section, config, isDoc) {
   // Only move on to the next section when all blocks are loaded.
   await Promise.all(loaded);
 
-  if (isDoc && section.el.dataset.idx === '0') {
-    loadPostLCP(config);
-  }
-
   // Show the section when all blocks inside are done.
   delete section.el.dataset.status;
+
+  if (isDoc && section.el.dataset.idx === '0') {
+    await loadPostLCP(config);
+  }
+
   delete section.el.dataset.idx;
 
   return section.blocks;


JasonHowellSlavin (approved):
LGTM

auniverseaway:
Thanks for taking the time to check this out and get it to the finish line.

Probably worth mentioning to the martech folks when this gets merged as it will slightly delay martech compared to what we do today. This is because we are now awaiting GeoRouting and putting it in front of martech.

mokimo (approved):

chrischrischris (changes requested):
Marking as "request-changes" to block merge as I need to do some additional perf testing.

libs/utils/utils.js
```diff
@@ -896,6 +896,11 @@ async function checkForPageMods() {
 }
 
 async function loadPostLCP(config) {
+  const georouting = getMetadata('georouting') || config.geoRouting;
+  if (georouting === 'on') {
+    const { default: loadGeoRouting } = await import('../features/georoutingv2/georoutingv2.js');
+    await loadGeoRouting(config, createTag, getMetadata, loadBlock, loadStyle);
```
vhargrave, line 902: I think this is great, but think it really only makes sense to `await` on mobile devices. On desktop we should be able to call martech a little faster. 

In general I think the best solution would actually be to do the following: 
If a user is on a mobile device & georoutingMetadata is active we call the logic to figure out if we'll show the georouting modal before even loading the first block. This call should be non-blocking so that we can load the first section still right away, but at the same time, the moment we get a response informing us that the georouting modal will be displayed we load the modal as quickly as we can and make it blocking. 

I think this makes sense to do because the georouting modal is displayed often and is more importantly often LCP on mobile. 
The price we pay is of course kicking off one more network request on mobile devices before LCP, so we'd have to do some testing on what this materially means
do we lose one point of LH on average but gain 5 points when georouting is displayed. Etc.

vhargrave (approved):
Looks great! 
I have some thoughts on how we could do this differently and am curious on what you think, but nothing that's worth blocking this from getting merged ASAP :)

libs/utils/utils.js
meganthecoder: @auniverseaway Thoughts on this?

meganthecoder:
> Marking as "request-changes" to block merge as I need to do some additional perf testing.

@chrischrischris Sounds good. Let me know if I can help at all. This is a high priority for us

Ruchika4 (approved):

chrischrischris (approved):

Brandon32 (approved):

libs/utils/utils.js
meganthecoder: Created a ticket to look into this approach https://jira.corp.adobe.com/browse/MWPW-142218

PR #1814, opened 2024-02-01
MWPW-142125: fix CLS issue on mnemonic list

yesil:
Resolves: [MWPW-142125](https://jira.corp.adobe.com/browse/MWPW-142125)

Pagespeed:
*Before: https://pagespeed.web.dev/analysis/https-main--milo--adobecom-hlx-live-fragments-ilyas-marquee-mnemonic/fuiich1ovb?form_factor=mobile. (**0.100 CLS**)

*After: https://pagespeed.web.dev/analysis/https-main--milo--yesil-hlx-live-fragments-ilyas-marquee-mnemonic/30kv3yr2jt?form_factor=mobile  (**0 CLS**)

**Test URLs:**
 - Before: https://main--milo--adobecom.hlx.live/fragments/ilyas/marquee-mnemonic?martech=off&georouting=off
 - After: https://mwpw-142125--milo--yesil.hlx.live/fragments/ilyas/marquee-mnemonic?martech=off&georouting=off

the margin-top on the marquee should not happen.

image


mokimo (approved):

libs/blocks/mnemonic-list/mnemonic-list.css
```diff
@@ -1,4 +1,4 @@
-.mnemonic-list {
+.marquee .mnemonic-list {
```
overmyheadandbody, line 1: This does fix the effect, but I'm not certain it's the cause. Looking at the markup, I see `.mnemonic-list` being applied to the `.marquee` element, but is also a standalone element nested inside. From the outside, I'd say that `.mnemonic-list` should not be added as a class to the marquee. Is there a reason behind it? I'd argue that we should keep class names relevant to their purpose and this feels counter-intuitive to me
Screenshot 2024-02-01 at 14 44 43

vladen (approved):

libs/blocks/mnemonic-list/mnemonic-list.css
yesil: I agree, I'll defer the refactoring to @Axelcureno , this is to fix the immediate CLS issue.

Axelcureno (approved):

libs/blocks/mnemonic-list/mnemonic-list.css
Axelcureno: This block is made to be used either standalone or inside another block and when using inside a block it requires minimal styling (this may not be standard across blocks). Since Marquee is the first block using it, we are simply introducing styles to match the XD specs of the usage. We can remove it for now and make the styles standard, but it's likely that we will need to re introduce this later. For example, Merch Card will use this block soon.

yesil:
fyi: @ryanmparrish thanks.

libs/blocks/marquee/marquee.js
```diff
@@ -64,16 +64,15 @@ const decorateImage = (media) => {
...
-
 export async function loadMnemonicList(foreground) {
   try {
-    await dynamicImport('../mnemonic-list/mnemonic-list.js')
-      .then((module) => {
-        module.decorateMnemonicList(foreground);
-      });
+    const { miloLibs, codeRoot } = getConfig();
+    const stylePromise = new Promise((resolve) => {
+      loadStyle(`${miloLibs || codeRoot}/blocks/mnemonic-list/mnemonic-list.css`, resolve);
```
auniverseaway, line 71: You can use `base` here.

auniverseaway (approved):
My hunch is that this is more complicated than it needs to be and we're trying to optimize something that doesn't need optimizing.

Also, why did something with such bad CLS ship to begin with? Who is approving these PRs? The best way to not have CLS issues to fix is to not ship them in the first place.

libs/blocks/marquee/marquee.js
yesil: It is trivial probably but i think since `base` is already there, I think there is no need to initialize an URL instance by parsing an url, thanks for the hint.

Sartxi (approved):

yesil:
@auniverseaway you are right CLS should not be merged at the first place. Going forward, I'll be carefully checking it in my reviews.

PR #1816, opened 2024-02-01
Sync stage from main

Ruchika4:
* An issue has been observed with merch card on stage.adobe.com on page https://www.stage.adobe.com/acrobat/pdf-reader.html
* This works fine with main branch
* Hence synching stage with main so that team can finish their testing on stage.adobe.com

**Test URLs:**
- After: https://main--milo--adobecom.hlx.page/?martech=off

DC:
- Before: https://stage--dc--adobecom.hlx.live/acrobat/pdf-reader?milolibs=stage
- After: https://stage--dc--adobecom.hlx.live/acrobat/pdf-reader

Slack Conversation: https://adobe-mwp.slack.com/archives/C0475EYRAKA/p1706761685855619


overmyheadandbody (approved):
We should really aim to never merge PRs with failing unit tests, as that reduces unit test credibility and reliability. I see the current failing tests are related to geo routing. Could someone have a look at those?

Approving since I know the root cause is not this PR, but a fast-follow story and implementation should be created.

mokimo (approved):

Ruchika4:
> We should really aim to never merge PRs with failing unit tests, as that reduces unit test credibility and reliability. I see the current failing tests are related to geo routing. Could someone have a look at those?
> 
> Approving since I know the root cause is not this PR, but a fast-follow story and implementation should be created.

All tests are passing now. It appears to be a flaky test that needs to be looked into. I'll try to look into it in a later PR.

PR #1818, opened 2024-02-01
MWPW-139296-Promo Aside Enhancement - Time Delay

Ruchika4:
* Add an option to author time delay in showing the aside promo bar popup
* If time entered is greater than 99 it'll be treated as milliseconds
* If time entered is less than 99 it'll be treated as seconds

Resolves: [MWPW-139296](https://jira.corp.adobe.com/browse/MWPW-139296)

**Test URLs:**
Time in milliseconds:
- Before: https://main--milo--adobecom.hlx.page/drafts/ruchika/aside-promo?martech=off
- After: https://time-delay--milo--ruchika4.hlx.page/drafts/ruchika/aside-promo?martech=off

Time in seconds:
- Before: https://main--milo--adobecom.hlx.page/drafts/ruchika/aside-promobar-seconds?martech=off
- After: https://time-delay--milo--ruchika4.hlx.page/drafts/ruchika/aside-promobar-seconds?martech=off

Note:
Sharing some before after urls for other use cases of promobar to make sure no impact on existing feature. Below scenarios should have same experience in before after urls
1.  no-delay sticky bottom popup promobar
- Before: https://main--milo--adobecom.hlx.page/drafts/ruchika/nodelay-promobar?martech=off
- After: https://time-delay--milo--ruchika4.hlx.page/drafts/ruchika/nodelay-promobar?martech=off

2. sticky bottom popup promobar that shows up after marquee only in mobile
- Before: https://main--milo--adobecom.hlx.page/drafts/tat28144/promobar/photoshop5
- After: https://time-delay--milo--ruchika4.hlx.page/drafts/tat28144/promobar/photoshop5


libs/blocks/section-metadata/section-metadata.js
```diff
@@ -15,13 +15,13 @@ function handleBackground(div, section) {
   }
 }
 
-export async function handleStyle(text, section) {
+export async function handleStyle(text, section, delay) {
```
ryanmparrish, line 18: Screen Shot 2024-02-05 at 12 58 42 PM
Should we make this arg optional so the other files using this function don't need it. 
```suggestion
export async function handleStyle(text, section, delay = null) {
```

ryanmparrish: Another thought, Rather than passing delay into `handleStyle()`, 
can we just get metadata.delay in the `handleStickySection()` below so we don't have to inject metadata.delay props into the metadata.style function?

Ruchika4: Incorporated the suggestion. Thanks for the review.

Ruchika4:
> Overall looks and works great. I had a few comments on trying to keep the section-metadata properties and the handleStyle function clean & testing against the files that import. LMK your thoughts
> 
> Not related to this enhancement, but we may want to consider when the sticky section is closed (el.classList.contains('close-sticky-section')), it would be beneficial to remove the resize event listener to avoid potential memory leaks. 'sticky-section.js:11'
> 
> ```
> if (el.classList.contains('close-sticky-section')) {
>   window.removeEventListener('resize', handleTopHeight);
>   observer.unobserve(entry.target);
>   return;
> }
> ```

I have added the removeEventLiistener and see nothing breaks. However tagging @aishwaryamathuria to also confirm if she sees any issue as she originally worked on it.

ryanmparrish (approved):

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -44,6 +44,12 @@ function handleLayout(text, section) {
   section.classList.add(layoutClass);
 }
 
+function handleDelay(time, section) {
+  if (!(time || section)) return;
+  section.classList.add('hide-sticky-section');
+  setTimeout(() => { section.classList.remove('hide-sticky-section'); }, time);
+}
```
rgclayton, line 51: Multiples of 1000 to get seconds is more of developer thinking. I think most people would put `delay | 5` and expect a five second delay.
What about adding some logic to check if it is authored with a single digit number (5), your code corrects for that because we know nobody would actually set the delay for five one thousands of a second. 

On the flip side if it is authored with 5000 that would delay for five seconds.

Thoughts?
cc: @ryanmparrish @chrischrischris

Patil31:
Verified the below two scenarios
 1. Promobar should load without delay
  https://time-delay--milo--ruchika4.hlx.page/drafts/shrinidhivp/promodelay/nodelay-promo

2. promobar should show up after marquee in mobile only 
https://time-delay--milo--ruchika4.hlx.page/drafts/shrinidhivp/promodelay/promo-photoshop

libs/blocks/section-metadata/section-metadata.js
Ruchika4: Thanks for the review @rgclayton Kyungwha and I discussed this as there is a block already in CC where we add time which is in milliseconds we agreed on having a common unit for time so that it's easier for authors. However I agree 5 seconds looks more readable, but we wanted to have parity across blocks to author time.

rgclayton: To base an authoring pattern off of what's done in a consumer doesn't always work well with what goes into Milo.

With what I'm proposing, either option would work and I think it's a more friendly authoring pattern. Less questions/confusion about what values `delay` would take.

ryanmparrish: I agree w @rgclayton - Using a simpler authoring seconds approach is likely preferred. 
The first thing I did testing this was hunt down what the delay value was. Seconds or milliseconds. 🤷  
I think the majority of authors would defer to simple seconds rather than mil. and having some logic that checks the length of the digit could be a nice way to have both.

Ruchika4: Thanks for the feedback @rgclayton and @ryanmparrish . I had a meeting with Ryan C and Kyung on Thursday to revisit the authoring pattern and I will get Ryan's suggestion implemented in this PR

rgclayton (approved):
Works with both seconds and milliseconds. 

Thanks Ruchika.

Patil31:
Reverified the delay with seconds and milliseconds
https://time-delay--milo--ruchika4.hlx.page/drafts/shrinidhivp/promodelay/new-promobar-delay?georouting=off
https://time-delay--milo--ruchika4.hlx.page/drafts/shrinidhivp/promodelay/promo-milliseconds?georouting=off

PR #1821, opened 2024-02-02
MWPW-141650 - [Icon Block] full-width alignment

Sartxi:
In all viewports the `icon-block.full-width` are always supposed to be centered. But currently the full width variation isn't styled in mobile so is inheriting a left alignment. Adding a style to target full-width icon blocks to match consonant design examples.

icon-block-full-width

Note: I also noticed that there were a lot of un-needed elements in icon block unit test, so I removed the unneeded mocks.

Resolves: [MWPW-141650](https://jira.corp.adobe.com/browse/MWPW-141650)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/sarchibeque/icon-block-full-width?martech=off
- After: https://sarchibeque-mwpw-141650-icon-block--milo--adobecom.hlx.page/drafts/sarchibeque/icon-block-full-width?martech=off


libs/blocks/icon-block/icon-block.css
```diff
@@ -83,6 +83,10 @@
   gap: 16px;
 }
 
+.five-up .icon-block.full-width .foreground .text-content {
+  text-align: center;
+}
+
```
rgclayton, line 89: Out of curiosity. Instead of adding this, why not remove the left align declaration[ from here](https://github.com/adobecom/milo/blob/0ba51c4aceeeca0dddc8d18acef150f92ab0b9ea/libs/blocks/icon-block/icon-block.css#L44)?  Then, the [center alignment here](https://github.com/adobecom/milo/blob/0ba51c4aceeeca0dddc8d18acef150f92ab0b9ea/libs/blocks/icon-block/icon-block.css#L33) kicks in.

Sartxi: Thanks @rgclayton I realized the left alignment wasn't needed anyway... I would also rather remove it 👍🏻

rgclayton (approved):

ryanmparrish (approved):

elan-tbx (approved):

PR #1846, opened 2024-02-08
MWPW-142433: Fixes various bugs in CaaS Marquee Metadata block

sanrai:
- Fixes various bugs in CaaS Marquee Metadata block

Resolves: [MWPW-142433](https://jira.corp.adobe.com/browse/MWPW-142433)

## Issue
If only one CTA is authored, the second CTA would return and try to send `undefined` values.

### Before
image
image

### After
image

**Milo Test URLs:**
- [Before](https://main--milo--adobecom.hlx.page/drafts/sanrai/marquee-metadata-block/only-one-cta-authored?martech=off)
- [After](https://MWPW-142433--milo--adobecom.hlx.page/drafts/sanrai/marquee-metadata-block/only-one-cta-authored?martech=off)

**Fix:**
This was fixed by forcing the for loop to always do two executions and in the case the second cta doen't exist, it smartly set `''` values.

```
  for (let i = 1; i <= MAX_NUM_CTAS; i += 1) { // MAX_NUM_CTAS = 2
```

Then updating `sendUtils.js` to not reject an empty string for `cta2`:

```
checkCtaUrl(s, options, 2)
```

## Issue
Instead of sending `target=_blank` in different attribute, append to end of URL to avoid sending additional attributes.

### Before
image
image

### After
image

**Milo Test URLs:**
- [Before](https://main--milo--adobecom.hlx.page/drafts/sanrai/marquee-metadata-block/default-from-block-library?martech=off)
- [After](https://MWPW-142433--milo--adobecom.hlx.page/drafts/sanrai/marquee-metadata-block/default-from-block-library)

**Fix:**
This was fixed by checking if a target exists, and if so, appending that value to the url.

```
function getUrl(ctaLink) {
  // ...
  const target = ctaLink.target ? `#${ctaLink.target}` : '';
  return `${ctaLink.href}${target}`;
}
```

## Issue
Description fields weren't sending authored links
image

**Fix:**
The `caas-marquee-metadata` block was correctly mapping the metadata field to have the anchor links, but it was being stripped by the `send-utils.js` before being sent. 

This line was stripping the html.
```
let val = n.children?.[1]?.textContent;
```

Added if condition to bypass this for marquee
```
if (marqueeMetadata) {
  val = n.children?.[1]?.innerHTML; 
} 
```
### Before
image

### After
image

**Milo Test URLs:**
- [Before](https://milo.adobe.com/drafts/sanrai/marquee-metadata-block/links-in-description?martech=off)
- [After](https://milo.adobe.com/drafts/sanrai/marquee-metadata-block/links-in-description)

Note: Needs to be merged before can test.

## Issue
Removes sending an empty target for modals as they are already handled by being appended to end of URL. 

### Before
image
image

### After
image

**Milo Test URLs:**
- [Before](https://main--milo--adobecom.hlx.page/drafts/sanrai/marquee-metadata-block/default-from-block-library?martech=off)
- [After](https://MWPW-142433--milo--adobecom.hlx.page/drafts/sanrai/marquee-metadata-block/default-from-block-library)

**Fix:**
This was fixed by just deleting the following fields

```
-     cta1target: `${metadata.cta1target}`,
-     cta2target: `${metadata.cta2target}`,
```


libs/blocks/caas-marquee-metadata/caas-marquee-metadata.js
```diff
@@ -14,7 +14,7 @@ function getUrl(ctaLink) {
   if (modalPath && modalHash) {
     return `${modalPath}${modalHash}`;
   }
-  return ctaLink.href;
+  return `${ctaLink.href}${ctaLink.target}`;
```
cmiqueo, line 17: should be just 
`return ctaLink.href;`
The target should be included on the link for the viewer top parse and decorate the link

arshadparwaiz (approved):

sanrai:
Will be merged by 02/16 at 2pm PST at the latest.

PR #1852, opened 2024-02-08
MWPW-138460 Update entitlements and add support for stage and consumer entitlements

chrischrischris:
Adds support for prod and stage entitlements.
Consumers can define their own entitlements via `config.entitlements`
Added a bunch of new entitlements

Resolves: [MWPW-138460](https://jira.corp.adobe.com/browse/MWPW-138460)

**Test URLs:**
- After: https://ccc-ent-update--milo--adobecom.hlx.page/


libs/features/personalization/stage-entitlements.js
```diff
@@ -0,0 +1,12 @@
+const STAGE_ENTITLEMENTS = {
+  '5a5fd14e-f4ca-49d2-9f87-835df5477e3c': 'photoshop-any',
```
honstar, line 2: Shouldn't we have parity between stage and prod entitlements? stage has fewer than prod.

chrischrischris: This is what has been given by the PM, I'll check with him but we can followup with another PR for that

honstar: No worries, was just thinking this might cause confusion when switching between environments.

libs/features/personalization/entitlements.js
```diff
@@ -1,15 +1,29 @@
...
 };
 
-const getEntitlements = (data) => {
-  const { entitlements = {} } = getConfig();
-  const entitlementMap = { ...entitlements, ...ENTITLEMENT_MAP };
+export const getEntitlementMap = async () => {
+  const { env, consumerEntitlements } = getConfig();
+  if (env.name === 'prod') return { ...consumerEntitlements, ...ENTITLEMENT_MAP };
+  const { default: STAGE_ENTITLEMENTS } = await import('./stage-entitlements.js');
+  return { ...consumerEntitlements, ...STAGE_ENTITLEMENTS };
```
honstar, line 22: Just to double-check, entitlement label collisions are not an issue since we are matching against the entitlement key/hash, right?

honstar (approved):

seanchoi-dev (approved):
Looks good except the unit test: https://github.com/adobecom/milo/actions/runs/7874267405/job/21483483978?pr=1852

Will you fix it before merge this?

libs/features/personalization/entitlements.js
```diff
@@ -1,15 +1,29 @@
...
+};
+
+const getEntitlements = async (data) => {
```
overmyheadandbody, line 25: Since this is now `async` and `init` returns it, should `init` be also marked as `async`? Someone who's not familiar with the logic might not realise they need to `await`. On the other hand, do we see anyone calling this directly or is it just used to expose the segments in `config.entitlements`?

test/features/personalization/personalization.test.js
```diff
@@ -20,12 +19,18 @@ const setFetchResponse = (data, type = 'json') => {
...
-ENTITLEMENT_MAP['11111111-aaaa-bbbb-6666-cccccccccccc'] = 'my-special-app';
-ENTITLEMENT_MAP['22222222-xxxx-bbbb-7777-cccccccccccc'] = 'fireflies';
-
 // Note that the manifestPath doesn't matter as we stub the fetch
 describe('Functional Test', () => {
+  before(() => {
+    // Add custom keys so tests doesn't rely on real data
+    const config = getConfig();
+    config.env = { name: 'prod' };
+    config.consumerEnts = {
```
overmyheadandbody, line 28: Shouldn't this be updated to `consumerEntitlements` as well?

test/features/personalization/personalization.test.js
```diff
@@ -231,15 +236,15 @@ describe('Functional Test', () => {
 
   it('should read and use entitlement data', async () => {
     setConfig(getConfig());
-    const { entitlements } = getConfig();
+    const config = getConfig();
+    config.consumerEnts = { 'consumer-defined-entitlement': 'fireflies' };
```
overmyheadandbody, line 240: Same naming question here

overmyheadandbody (approved):
Just a few quick notes

vgoodric (approved):

PR #1853, opened 2024-02-12
MWPW-141021: Project PEP Accessibility Requirements

robert-bogos:
## Description
This PR introduces some accessibility requirements for PEP

## Related Issue
Resolves: [MWPW-141021](https://jira.corp.adobe.com/browse/MWPW-141021)

## Testing instructions
- Initial focus should be on the X button;
- PEP should be dismissed when pressing Escape;
- Pressing Enter should dismiss PEP when the focus is on the cancel button or the X button;
- The focus on the X button should resemble the image below:
image
- When focused on the X button and pressing Tab, the focus should move to the cancel button;

## Test URLs
**Milo:**
- After: https://project-pep--milo--robert-bogos.hlx.page/drafts/rbogos/pep-prompt?martech=off&imsClientId=fedsmilo


libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -142,6 +143,10 @@ class AppPrompt {
     //   if (!e.target.closest(CONFIG.selectors.prompt)) AppPrompt.close();
     // });
 
+    document.addEventListener('keydown', (event) => {
+      if (['Tab', 'Escape', 'Enter'].includes(event.key)) this.close();
+    });
```
mokimo, line 148: We should IMO only attach/re-attach this when the modal is open.

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -123,7 +124,7 @@ class AppPrompt {
       <div class="appPrompt-icon">
         ${this.image}
       </div>
-      <div class="appPrompt-title">${this.title}</div>
+      <h2 class="appPrompt-title">${this.title}</h2>
```
overmyheadandbody, line 127: Did we get SEO requirements/approval for this? I remember they wanted to keep heading element numbers low

libs/features/webapp-prompt/webapp-prompt.js
overmyheadandbody: If the logic reaches this point, the modal should be open. However, we should remove the event listener when the modal is closed. If the user clicks on any close button, we don't have any need for this event listener, since there won't be any prompt to close

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -80,7 +80,8 @@ class AppPrompt {
     // TODO: we might need to also define a primary-like CTA
     const cancelText = content.querySelector('em > a');
     // TODO: add placeholder to sheet and document that consumers will need to add it too
-    // TODO: is there a better way to let authors define the CTA text? Should we just use a placeholder?
+    // TODO: is there a better way to let authors define the CTA text?
+    //       Should we just use a placeholder?
```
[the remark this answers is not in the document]
robert-bogos: Changed the indent 👍

libs/features/webapp-prompt/webapp-prompt.js
robert-bogos: Yes, it's stated in the figma specs

libs/features/webapp-prompt/webapp-prompt.js
robert-bogos: True, better remove the event listener when PEP is closed. Made the change 👍

mokimo (approved):

narcis-radu (approved):

libs/features/webapp-prompt/webapp-prompt.js
overmyheadandbody: The Figma specs are suggesting the look of the element, not necessarily the markup to use. business.adobe.com has 2 `h2` elements, adobe.com has just one, we should probably aim to keep these numbers low. We might end up in a situation where PEP content appears in search results, when it's actually a feature for a limited set of users.

We had heading elements in the header and footer navigations and the SEO team asked us to remove them because they were polluting search results.

mokimo: I agree with Rares here, it would be good to get some SEO input on this.

libs/features/webapp-prompt/webapp-prompt.css
```diff
@@ -133,6 +133,12 @@
     content: '✕';
   }
 
+  .appPrompt-close:focus {
+    background-color: #e9e9e9;
+    border-radius: 50%;
+    border: 3px solid white;
```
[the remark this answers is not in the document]
robert-bogos: This would make the border take the background color of the focused element, which is grey. Following the figma, it should be white. Am I missing something ? 🤔

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -142,10 +143,16 @@ class AppPrompt {
...
     // });
 
+    document.addEventListener('keydown', this.handleKeyDown);
+
     [this.elements.closeIcon, this.elements.cta]
       .forEach((elem) => elem.addEventListener('click', this.close));
   };
 
+  handleKeyDown = (event) => {
+    if ([' ', 'Escape'].includes(event.key)) this.close();
```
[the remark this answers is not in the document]
robert-bogos: Discussed with our PM and the prompt should not be focused, so we're good with the current behaviour 👍

libs/features/webapp-prompt/webapp-prompt.css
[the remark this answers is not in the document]
overmyheadandbody: I'm not entirely sure the white border is needed in the first place, to me it looks like a spec omission, as we don't have that sort of styling for any other elements. If someone from Studio requested it specifically, let's at least turn its color and the prompt background color into a variable so we can easily update it in the future.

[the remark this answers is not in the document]
robert-bogos: Done 👍 I introduced the css variables

libs/features/webapp-prompt/base.css
```diff
@@ -0,0 +1,4 @@
+:root{
+  --pep-background-prompt: #ffffff;
+  --pep-background-progress: #e9e9e9;
+}
```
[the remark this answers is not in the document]
robert-bogos: Done. I was trying to mimic the structure from the global navigation, but you're right, it doesn't make sense in here 👍

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -145,10 +145,16 @@ class AppPrompt {
...
     [this.elements.closeIcon, this.elements.cta]
       .forEach((elem) => elem.addEventListener('click', this.close));
   };
 
+  handleKeyDown = (event) => {
+    if (event.key === 'Escape') this.close();
+  };
+
   initRedirect = () => setTimeout(() => {
     // TODO: uncomment for actual redirect
```
honstar, line 159: Is this TODO still valid?

honstar (approved):

libs/features/webapp-prompt/webapp-prompt.js
overmyheadandbody: For the time being, yes. But this is being merged in a feature branch, so it won't affect anything.

After a certain amount of time, a user would get redirected to another page; we want that behavior disabled during development, as we're interested in UI, data, accessibility, dismissal logic, etc. Another option would be to set the redirect time to a larger value, say 30 minutes, but we have plenty more TODOs in the code, so we'll make sure to address them all before opening a PR against `main`

libs/features/webapp-prompt/webapp-prompt.css
```diff
@@ -1,3 +1,8 @@
+:root {
+  --pep-background-prompt: #ffffff;
+  --pep-background-progress: #e9e9e9;
+}
```
overmyheadandbody, line 4: A final nit - these values are specific to PEP. `:root` would never need them, so why not add the custom properties inside `.appPrompt`?

overmyheadandbody (approved):
One final note

libs/features/webapp-prompt/webapp-prompt.js
robert-bogos: Yes, we still need it for the moment

libs/features/webapp-prompt/webapp-prompt.css
robert-bogos: @overmyheadandbody Good point! Made the change 👍

SilviuLCF:
QA verified

PR #1854, opened 2024-02-12
Mwpw 142278: Thumbnail icons are not aligned vertically on iPad- fix

drashti1712:
* Fixed - Thumbnail icons are not aligned vertically on iPad while they are aligned on desktop.

Resolves: [MWPW-142278](https://jira.corp.adobe.com/browse/MWPW-142278)

**Test URLs:**
- Before: https://main--milo--drashti1712.hlx.page/drafts/drashti/bugs/features?martech=off
- After: https://mwpw-142278--milo--drashti1712.hlx.page/drafts/drashti/bugs/features?martech=off

Test pages with enabled icons:
- Before: https://main--cc--adobecom.hlx.live/creativecloud/business/enterprise/features
- After: https://main--cc--adobecom.hlx.live/creativecloud/business/enterprise/features?milolibs=mwpw-142278--milo--drashti1712


spadmasa:
Verified the flow in dekstop , iphone, ipad and android device , after the fix :
https://main--cc--adobecom.hlx.live/creativecloud/business/enterprise/features?milolibs=mwpw-142278--milo--drashti1712
image
image
image

In actual device
IMG_20240213_145231
IMG_20240213_145234
IMG_20240213_145531
IMG_20240213_145536
IMG_20240213_145540
IMG_20240213_145227
IMG_20240213_145205
IMG_20240213_145211
IMG_20240213_145216
IMG_20240213_145827
IMG_20240213_145830

In saucelab
image
image

mokimo (approved):

libs/blocks/icon-block/icon-block.css
```diff
@@ -187,8 +187,7 @@
 }
 
 .icon-block.inline .foreground .text-content .icon-area {
-  width: fit-content;
-  display: contents;
+  width: var(--icon-size-m);
```
mokimo, line 190: Is it fine to always have a fixed size, are there any use-cases where the icons should have a different size?

overmyheadandbody: I agree that this might not be the full answer. If an `l` size icon is used, there might be some unexpected results. I see a lot of rules that set `width`, `min-width`, `max-width` and their `height` equivalents for the `.icon-area` element. An `l` size icon would have a `max-width` of `234px`, a `min-width` of `56px` and a `width` of `40px`, these should probably be consolidated better.

drashti1712: > Is it fine to always have a fixed size (also on other use-cases/other consumers), are there any use-cases where the icons should have a different size?



> I agree that this might not be the full answer. If an `l` size icon is used, there might be some unexpected results. I see a lot of rules that set `width`, `min-width`, `max-width` and their `height` equivalents for the `.icon-area` element. An `l` size icon would have a `max-width` of `234px`, a `min-width` of `56px` and a `width` of `40px`, these should probably be consolidated better.

@mokimo, @overmyheadandbody I looked into this point and I realised that `width: min-content` could be used here if we don't want the height to be fixed. Do let me know if this works.

chrischrischris (approved):

Ruchika4 (approved):

overmyheadandbody (approved):
I think this is a better solution than explicit values, thanks for looking into the change! 👍

rgclayton (approved):

ryanmparrish:
@drashti1712  - Sorry, I'm late to the party. This PR appears to have broken the aligment of icon-block.inline w/ a single header. Do we want to revert or create a fix PR to address this? 
https://main--milo--adobecom.hlx.page/drafts/rparrish/icon/icon-block-inline-milo (edited) 
https://main--cc--adobecom.hlx.page/products/illustrator/free-trial-download

drashti1712:
> @drashti1712 - Sorry, I'm late to the party. This PR appears to have broken the alignment of icon-block.inline w/ a single header. Do we want to revert or create a fix PR to address this? https://main--milo--adobecom.hlx.page/drafts/rparrish/icon/icon-block-inline-milo (edited) https://main--cc--adobecom.hlx.page/products/illustrator/free-trial-download

@ryanmparrish We can fix this by adding (not removing) `display:contents` for icon-area.

PR #1882, opened 2024-02-15
MWPW-142617: Microsoft Edge & Chrome Win - progress bar not completely loaded

robert-bogos:
## Description
This PR fixes the PEP prompt progress bar bug, on Microsoft Edge 121 & Win, where the bar was not completely loaded at the end of the CSS animation.

## Related Issue
Resolves: [MWPW-142617](https://jira.corp.adobe.com/browse/MWPW-142617)

## Testing instructions
Open the test URL on Microsoft Edge 121 & Win

## Test URLs
**Milo:**
- After: https://mwpw-142617-project-pep--milo--robert-bogos.hlx.page/drafts/rbogos/pep-prompt?martech=off&imsClientId=fedsmilo

### DIFF
--- a/libs/features/webapp-prompt/webapp-prompt.css
+++ b/libs/features/webapp-prompt/webapp-prompt.css
@@ -156,9 +156,6 @@
   }
 
   @keyframes appPrompt-animate {
-    0% {
-        transform: scaleX(0);
-    }
     100% {
         transform: scaleX(1);
     }
--- a/libs/features/webapp-prompt/webapp-prompt.css
+++ b/libs/features/webapp-prompt/webapp-prompt.css
@@ -151,15 +151,14 @@
     width: 1%;
     height: 100%;
     animation: appPrompt-animate 5s linear forwards;
-    transform-origin: 0 0;
   }
 
   @keyframes appPrompt-animate {
     0% {
-        transform: scaleX(1);
+        width: 1%;
     }
     100% {
-        transform: scaleX(100);
+        width: 100%;
     }
   }
 }
--- a/libs/features/webapp-prompt/webapp-prompt.css
+++ b/libs/features/webapp-prompt/webapp-prompt.css
@@ -148,17 +148,19 @@
   }
 
   .appPrompt-progress {
-    width: 1%;
+    width: 100%;
     height: 100%;
     animation: appPrompt-animate 5s linear forwards;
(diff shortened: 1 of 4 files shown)


libs/features/webapp-prompt/webapp-prompt.css
```diff
@@ -151,15 +151,14 @@
...
   }
 
   @keyframes appPrompt-animate {
     0% {
-        transform: scaleX(1);
+        width: 1%;
     }
     100% {
-        transform: scaleX(100);
+        width: 100%;
```
overmyheadandbody, line 161: I think we should investigate a bit more before falling back to animating `width`. The issue with it is that it can cause performance issues, transforms rely on GPU and are less of a performance bottleneck

overmyheadandbody: A quick reference on that - https://web.dev/articles/animations-overview

robert-bogos: Sure 👍 Will investigate a bit more

robert-bogos: Done. Found a solution that puts the GPU to work 👍

libs/features/webapp-prompt/webapp-prompt.css
```diff
@@ -148,18 +148,19 @@
...
     height: 100%;
     animation: appPrompt-animate 5s linear forwards;
     transform-origin: 0 0;
+    transform: scaleX(0);
   }
 
   @keyframes appPrompt-animate {
     0% {
-        transform: scaleX(1);
+        transform: scaleX(0);
```
overmyheadandbody, line 160: Do we even need the `0%` in this case?

overmyheadandbody (approved):
Thanks for looking into this!

mokimo (approved):

mokimo (approved):
Would be good if you fix this unrelated eslint error!

libs/features/webapp-prompt/webapp-prompt.css
robert-bogos: Good point. The animation will start from the initial state anyway. Made the change 👍

robert-bogos:
The eslint error is fixed in another pending PR, but I fixed it in here too so we can have green on all tests.

PR #1897, opened 2024-02-19
MWPW-142622: [Project PEP] Close button focus 

robert-bogos:
## Description
This PR fixes the bug where the blue focus outline of the PEP X button does not show, when PEP is initiated after the region picker modal.

## Related Issue
Resolves: [MWPW-142622](https://jira.corp.adobe.com/browse/MWPW-142622)

## Testing instructions
1. Go to the test URL;
2. Sign in;
3. Wait for the region picker modal to show up and close it;
4. When PEP is initiated, the X button should be focused and have a blue outline;

## Screenshots (if appropriate):
image

## Test URLs
**Milo:**
- After: https://mwpw-142622-project-pep--milo--robert-bogos.hlx.page/drafts/rbogos/pep-prompt?martech=off&imsClientId=fedsmilo


libs/features/webapp-prompt/webapp-prompt.css
```diff
@@ -138,6 +138,10 @@
     left: 12px;
   }
 
+  .appPrompt-close:focus {
+    outline: 2px solid #157AF3;
```
narcis-radu, line 142: Looks good to me, however I'm wondering whether we should use variables for colours. Maybe @overmyheadandbody has additional feedback on this.
Also, one thing I noticed: the focus order is "cancel" button, then "X", even though the "X" mark is higher than the "cancel" button. Is this intended?

narcis-radu (approved):

libs/features/webapp-prompt/webapp-prompt.css
robert-bogos: The focus order should be indeed different, but that's addressed in [another pending PR](https://github.com/adobecom/milo/pull/1853)

SilviuLCF (approved):
QA verified , PR can be merged

libs/features/webapp-prompt/webapp-prompt.css
overmyheadandbody: We don't set any other focus colors on any of the Gnav elements. This is intentional, as we want all page elements to be consistent in their focused state. Since a lot of other elements don't have an explicit value, it means we ultimately rely on the browser to set this color, through its default `:focus-visible` value. Setting an explicit value might break this consistency for this particular element. Is there no other way of handling this issue, but maintain the user agent defaults?

overmyheadandbody (commented):
This comes with the risk of some consistency issues, maybe we could find another way of handling this

libs/features/webapp-prompt/webapp-prompt.css
robert-bogos: @overmyheadandbody I'll investigate further and see if there's a way without setting the focus color 👍

robert-bogos: I found a solution that preserves the user agent focus color, it's a bit off on safari, but it's better than what we had before

PR #1907, opened 2024-02-21
MWPW-142830-Play video only on hoverplay and not on page load

Ruchika4:
* Play video only on hoverplay and not on page load

Resolves: [MWPW-142830](https://jira.corp.adobe.com/browse/MWPW-142830)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ruchika/explore-ai-features?martech=off
- After: https://hover--milo--ruchika4.hlx.live/drafts/ruchika/explore-ai-features?martech=off

Note: Above page has first video with #_autoplay#_hoverplay all the other videos are with #_hoverplay only
Authoring notes:
1. Please use a poster image with video when only using #_hoverplay. We'll be documenting the details about this in authoring document [here](https://milo.adobe.com/docs/authoring/videos)


elan-tbx (approved):

fullcolorcoder (approved):
Code looks good, first video plays automatically and then again if I hover like the ticket asks.

One thing to point out, there is a slight content shift when you hover that could be undesirable. I don't think it's being caused by anything here looking at the before test URL that behavior exists already. 

https://github.com/adobecom/milo/assets/7562106/e31b4938-c882-4a78-bb82-a59c42e940cf

It seems like there's also room for this feature to be misused from an accessibility perspective, might be worth an extra documentation note.

Ruchika4:
> Code looks good, first video plays automatically and then again if I hover like the ticket asks.
> 
> One thing to point out, there is a slight content shift when you hover that could be undesirable. I don't think it's being caused by anything here looking at the before test URL that behavior exists already.
> 
>  demo-hover-size.mov 
> It seems like there's also room for this feature to be misused from an accessibility perspective, might be worth an extra documentation note.

Thanks for detailed review, @fullcolorcoder . I'll make sure to add accessibility concerns with hover play videos in the documentation.

salonijain3 (approved):

robert-bogos (approved):

chrischrischris (approved):

PR #1908, opened 2024-02-21
MWPW-142318 Featured Article Image Loading

Brandon32:
* Add high priority and eager loading to improve LCP
* Using document fragment to reduce reflow

Resolves: [MWPW-142318](https://jira.corp.adobe.com/browse/MWPW-142318)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/bmarshal/featured-article?martech=off
- After: https://bmarshal-featured-image--milo--adobecom.hlx.page/drafts/bmarshal/featured-article?martech=off

**BACOM BLOG URLs:**
- Before: https://main--bacom-blog--adobecom.hlx.live/blog/?martech=off
- After: https://main--bacom-blog--adobecom.hlx.live/blog/?milolibs=bmarshal-featured-image&martech=off

### DIFF
--- a/libs/blocks/featured-article/featured-article.js
+++ b/libs/blocks/featured-article/featured-article.js
@@ -34,6 +34,7 @@ export default async function init(el) {
   const pic = doc.body.querySelector('picture');
   const img = pic.querySelector('img');
   img.removeAttribute('loading');
+  img.setAttribute('fetchpriority', 'high');
   const featuredImg = createTag('div', { class: 'featured-article-card-image' }, pic);
   const categoryEl = createTag('div', { class: 'featured-article-card-category' });
   img.addEventListener('load', () => {
@@ -49,5 +50,7 @@ export default async function init(el) {
   const date = createTag('p', { class: 'featured-article-card-date' }, dateMeta);
 
   body.append(categoryEl, title, description, date);
-  a.append(featuredImg, body);
+  const fragment = document.createDocumentFragment();
+  fragment.append(featuredImg, body);
+  a.append(fragment);
 }


libs/blocks/featured-article/featured-article.js
```diff
@@ -33,7 +33,8 @@ export default async function init(el) {
   const doc = parser.parseFromString(html, 'text/html');
   const pic = doc.body.querySelector('picture');
   const img = pic.querySelector('img');
-  img.removeAttribute('loading');
+  img.setAttribute('loading', 'eager');
```
rgclayton, line 36: Is this giving you a performance bump? The [default is already eager.](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/loading#value)

Brandon32: I'm not seeing a performance bump from explicitly setting eager loading

JasonHowellSlavin: I think the way the JIRA ticket was written may be the issue. I misdiagnosed the performance based on the results of the pagespeed.web.dev's output at the time of the ticket's writing regarding the LCP. That being said, Brandon pinpointed that performance issues were more likely due to the "featured article" being the LCP and technically a fragment. 

In addition, I am getting good scores from pagespeed:
Screen Shot 2024-02-21 at 1 10 18 PM

Brandon32: I'll remove the explicitly eager loading, are the other performance changes useful for a PR? I was seeing some minor improvements in LCP Image Load Delay but could be margin of error.

rgclayton: Yeah, I think adding code to add `loading="eager"` isn't adding any benefit here. 

On blog.adobe.com we added some [code to target the featured-article block](https://github.com/adobecom/blog/blob/main/scripts/scripts.js#L186-L197) and wrap it in a section separate from the the article feed. The idea is that it's the first section and gets priority over the rest of the content - mainly the article feed.

Brandon32: @rgclayton That sounds good, I'll close this PR then. 
We don't have enough pages to using featured-article block to target it specifically in scripts.js but we will keep this approach in mind for the future.

PR #1916, opened 2024-02-22
MWPW-142746 Change mobile georouting background to svg

meganthecoder:
* Change mobile georouting background from png to inline svg
* Goal is better performance score, especially LCP
* Note: I updated only mobile because the desktop performance score was fine and it kept the file size down.

Resolves: [MWPW-142746](https://jira.corp.adobe.com/browse/MWPW-142746)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/de/drafts/methomas/georouting?martech=off
- After: https://methomas-geo-lcp-svg--milo--adobecom.hlx.page/de/drafts/methomas/georouting?martech=off

Bacom:
Regular:
- https://main--bacom--adobecom.hlx.live/fr/?milolibs=methomas-geo-lcp-svg
Tabs:
- https://main--bacom--adobecom.hlx.live/?milolibs=methomas-geo-lcp-svg&akamaiLocale=CA

Before:
geo-lcp-svg_baseline
After:
geo-lcp-svg


meganthecoder:
Forgot about tests. I'll reopen when those are done.

JasonHowellSlavin (approved):
LGTM!

Brandon32 (approved):

libs/features/georoutingv2/georoutingv2.css
```diff
@@ -1,15 +1,21 @@
 .dialog-modal.locale-modal-v2 .georouting-wrapper {
+  display: block !important;
```
honstar, line 2: Is this absolutely necessary?

meganthecoder: Yes, I'm setting `display: none;` within javascript so this is necessary to override the inline style.

libs/features/georoutingv2/georoutingv2.js
```diff
@@ -236,26 +236,30 @@ function buildContent(currentPage, locale, geoData, locales) {
...
-    const tabsContainer = createTabsContainer(sortedLocales.map((l) => l.language));
-    georoutingWrapper.appendChild(tabsContainer);
-
-    sortedLocales.forEach((locale) => {
-      const content = buildContent(currentPage, locale, geoData, sortedLocales);
-      const tab = createTab(content, locale.language);
-      georoutingWrapper.appendChild(tab);
-    });
+  if (!availableLocales.length) return null;
+  const { innerWidth } = window;
```
overmyheadandbody, line 240: `innerWidth` might not always take scrollbars into consideration, `matchMedia` could be a good alternative. You'd have to properly resize the window for unit testing, but it might be more consistent.

libs/features/georoutingv2/georoutingv2.js
```diff
@@ -236,26 +236,30 @@ function buildContent(currentPage, locale, geoData, locales) {
...
+  let svg = null;
+  if (innerWidth < 480) {
+    const { default: getMobileBg } = await import('./getMobileBg.js');
+    svg = getMobileBg();
+  }
+  const georoutingWrapper = createTag('div', { class: 'georouting-wrapper fragment', style: 'display:none;' }, svg);
```
overmyheadandbody, line 246: To address @honstar's comment - you could add a class that you could toggle, handling the `display` value from CSS to avoid the `!important`

vhargrave: agree 👍

libs/features/georoutingv2/georoutingv2.css
```diff
@@ -1,15 +1,21 @@
...
   padding: 56px 24px 40px;
 }
 
 .dialog-modal.locale-modal-v2 {
-  background-image: url('/libs/features/georoutingv2/img/GeoModal_BG_Map_Mobile.png');
```
vhargrave, line 6: the background image isn't positioning quite right with this solution. In addition, you could also get this to still work by just referencing an svg here instead of a png. That gets you the same awesome result. 🙌 
I would recommend reverting all this css code and just referencing an svg.

libs/features/georoutingv2/georoutingv2.js
```diff
@@ -236,26 +236,30 @@ function buildContent(currentPage, locale, geoData, locales) {
...
-    georoutingWrapper.appendChild(tabsContainer);
-
-    sortedLocales.forEach((locale) => {
-      const content = buildContent(currentPage, locale, geoData, sortedLocales);
-      const tab = createTab(content, locale.language);
-      georoutingWrapper.appendChild(tab);
-    });
+  if (!availableLocales.length) return null;
+  const { innerWidth } = window;
+  let svg = null;
```
vhargrave, line 241: you can get rid of all this too if you just reference an svg instead of a png

vhargrave (changes requested):
~~you can simplify this quite a bit by just referencing an svg . ~~
I love the approach of using an svg in general though. What a difference it makes! 

~~btw - I quickly tested out what I'm suggesting here - https://main--bacom--adobecom.hlx.live/?milolibs=vhargrave-test-lcp-svg&akamaiLocale=CA . Just in case you wanted to ensure page speed is still good. ~~

I'm wrong with the svg reference 🤔. I was getting a good score, but it's now swinging back into negative territory.

libs/features/georoutingv2/georoutingv2.js
meganthecoder: A class would defeat the purpose. Let me explain:

The modal content (including the svg) gets added to the DOM before the css rules have been processed, creating a split second where the modal is unstyled (picture on left).
Screenshot 2024-02-28 at 12 06 05 PM
This creates CLS. To clarify, there was CLS before my changes, but because the inline svg is so tall, it increased the amount of CLS.

To prevent this, I'm hiding the modal until the stylesheet is loaded. That's why I have the `display: none;` inline and the `display: block !important;` in the css file. This means that the modal content is only seen once the styles are in place, preventing CLS.

There are, of course, other possible solutions to make sure the CSS loads before the modal is added to the DOM, but this method seemed simpler, would change less code, and wouldn't block the modal rendering while waiting for the styles.

Additionally, the reason why using `!important` is bad practice is because highjacks the normal inheritance structure and makes it messy to override the property downstream. However, I don't think is really a problem in our case. The `georouting-wrapper` is a very specific element. Are we really going to be creating different versions of it that will need different `display` values?

This was my thought process. If you still object, I can look at preventing the CLS another way. Let me know, thanks

meganthecoder:
> **Edit** **I'm wrong with the svg reference 🤔. I was getting a good score, but it's now swinging back into negative territory because the georouting is still lcp. So the only thing to fix imo is the background image positioning**

Yes, that's consistent with the information I found here: https://web.dev/articles/lcp#what-elements-are-considered

They even say that full `<svg>` support may be added in the future, so it's something for us to keep an eye on.

libs/features/georoutingv2/georoutingv2.css
meganthecoder: What device or breakpoint are you are you looking at?

libs/features/georoutingv2/georoutingv2.js
meganthecoder: Good to know! Since `innerWidth` is several times faster than `matchMedia`, I think I'll leave it as is.

libs/features/georoutingv2/georoutingv2.css
vhargrave: The iPhone 12 Pro for example. You'll see the background image should scale to take up all the space in the background
So here's how it looks on main currently: 
image

And here's how it looks on the branch:
image

libs/features/georoutingv2/georoutingv2.js
vhargrave: That's really interesting . I can't even reproduce that CLS no matter how much I throttle on my machine 👍 
I would have probably just waited for the css to finish processing then in the showModal method like so:
```
let resolve;
const cssProcessed = new Promise((resolveFunc) => resolve = resolveFunc);
const promises = [
  tabs ? loadBlock(tabs) : null,
  tabs ? loadStyle(`${miloLibs || codeRoot}/blocks/section-metadata/section-metadata.css`) : null,
  loadStyle(`${miloLibs || codeRoot}/features/georoutingv2/georoutingv2.css`, () => resolve()),
  cssProcessed,
];
await Promise.all(promises);
```
I think this is slightly better since we're avoiding an !important and you're waiting for the precise thing that you really wanted, but I'll leave it in your court if you really want to use. I'm not even sure it properly works, since I couldn't reproduce the cls ! :)

libs/features/georoutingv2/georoutingv2.css
meganthecoder: Good catch! Updated

libs/features/georoutingv2/georoutingv2.js
meganthecoder: I tried waiting for it to resolve ([see commit](https://github.com/adobecom/milo/pull/1916/commits/52c9fec7416df0fe5c6a8a76f6b966be758d5049)) and it added 0.222 CLS, pushing it into yellow territory. Given that, I'd prefer to keep the code as is. Please let me know if you still consider this a blocking issue. Thanks!

meganthecoder:
FYI I am no longer getting the near-100 results as seen in the description on the homepage due to authoring changes. They added a fragment to the top of the page for an event promo. However, other pages like AEM sites that don't have a fragment can still match those results:
Screenshot 2024-02-29 at 5 52 58 PM

libs/features/georoutingv2/georoutingv2.js
vhargrave: ok, thanks for trying. 🙌 
All good from my side then.

vhargrave (approved):

vhargrave:
@meganthecoder you probably need to change the target branch to point to stage now btw.

echampio-at-adobe (approved):
Verified on https://www.stage.adobe.com/creativecloud/plan-recommender/quiz.html#langnav

PR #1922, opened 2024-02-23
MWPW-143199  Remove 

Blainegunn:
* Add your
* Specific
* Features or fixes

**Test URLs:**

### DIFF
--- a/libs/blocks/global-navigation/features/profile/dropdown.js
+++ b/libs/blocks/global-navigation/features/profile/dropdown.js
@@ -2,6 +2,9 @@ import { getConfig } from '../../../../utils/utils.js';
 import { toFragment, getFedsPlaceholderConfig, trigger, closeAllDropdowns, logErrorFor } from '../../utilities/utilities.js';
 import { replaceKeyArray } from '../../../../features/placeholders.js';
 
+const ACOM_SIGNED_IN_STATUS = 'acomsis';
+const isStage = document.location.host.includes('stage');
+
 const getLanguage = (ietfLocale) => {
   if (!ietfLocale.length) return 'en';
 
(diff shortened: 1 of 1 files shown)


Axelcureno (approved):

seanchoi-dev (approved):

mokimo (changes requested):
I don't think this PR belongs into milo core - Ilyas had a very good summary about it https://github.com/adobecom/milo/pull/1301#discussion_r1331095051

libs/utils/utils.js
```diff
@@ -794,8 +794,12 @@ export async function loadIms() {
     if (!window.adobeIMS?.isSignedInUser()) {
       getConfig().entitlements([]);
     }
+    // remove acomsis cookie that is used in Akamai, for SignOut event from ims.
+    window.adobeProfile?.addEventListener('adobeProfile:SignOut', () => {
```
narcis-radu, line 798: Hi @Blainegunn - Milo does not use the Profile Menu, the logic added here will never execute.

libs/blocks/global-navigation/features/profile/dropdown.js
```diff
@@ -162,6 +165,8 @@ class ProfileDropdown {
 
     signOutLink.addEventListener('click', (e) => {
       e.preventDefault();
+      document.cookie = `${ACOM_SIGNED_IN_STATUS}=;path=/;expires=${new Date(0).toUTCString()};`;
```
narcis-radu, line 168: I don't think this approach is correct, it will only work on MWP pages, but the users can sign out from different sources like behance.net, creativecloud.adobe.com, photoshop web, express, etc. Those surfaces don't use Milo or FEDS on AEM, they use the standalone Profile Menu or a different implementation which will not remove the cookie. Use-cases like the one below will break:
* go to adobe.com and sign in -> cookie is set
* go to a page which is not Milo / FEDS on AEM and sign out -> cookie is not removed
* go to adobe.com -> cookie is still set, issue

Additionally, we will roll out the Universal Navigation which will replace the current code you have changed; a new Profile Menu will be loaded from an external library and we will not be able to set the custom cookie logic.

narcis-radu (changes requested):

PR #1923, opened 2024-02-23
MWPW-136885: Accessibility in Merch cards

Axelcureno:
- Adds tabindex="0" to all links and buttons inside merch cards to make it accessible to the users.

Resolves: [MWPW-136885](https://jira.corp.adobe.com/browse/MWPW-136885)

**Test URLs:**
- Before: https://main--milo--axelcureno.hlx.page/docs/library/kitchen-sink/merch-card?martech=off
- After: https://mwpw-136885--milo--axelcureno.hlx.page/docs/library/kitchen-sink/merch-card?martech=off

### DIFF
--- a/libs/blocks/merch-card/merch-card.js
+++ b/libs/blocks/merch-card/merch-card.js
@@ -504,7 +504,6 @@ const init = async (el) => {
   }
   decorateFooterRows(merchCard, footerRows);
   el.replaceWith(merchCard);
-
   decorateMerchCardLinkAnalytics(merchCard);
   return merchCard;
 };
--- a/libs/blocks/merch-card/merch-card.js
+++ b/libs/blocks/merch-card/merch-card.js
@@ -505,7 +505,6 @@ const init = async (el) => {
   decorateFooterRows(merchCard, footerRows);
   el.replaceWith(merchCard);
 
-  merchCard.querySelectorAll('a, button').forEach((element) => element.setAttribute('tabindex', '0'));
   decorateMerchCardLinkAnalytics(merchCard);
   return merchCard;
 };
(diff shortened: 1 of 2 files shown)


Roycethan:
@Axelcureno In the kitchen sink example i see tabindex=0 exist for <merch-card ...</merch-card>(parent) also,
and this makes tab order/focus will  goes to each whole card which i feel not necessary for screen readers since its not actionable .. yes for buttons/links its needed as you did right in this PR. Plz review with accessibility experts once if possible. thanks

libs/blocks/merch-card/merch-card.js
```diff
@@ -504,6 +504,11 @@ const init = async (el) => {
   }
   decorateFooterRows(merchCard, footerRows);
   el.replaceWith(merchCard);
+
+  const linksAndButtons = merchCard.querySelectorAll('a, button');
+  linksAndButtons.forEach((element) => {
+    element.setAttribute('tabindex', '0');
+  });
```
robert-bogos, line 511: Nit: Don't need to cache the NodeList
```suggestion
  merchCard.querySelectorAll('a, button').forEach((element) => element.setAttribute('tabindex', '0'));
```

robert-bogos (approved):

overmyheadandbody (commented):
I second @Roycethan's concern, setting `tabindex` on all anchor and button elements might be detrimental to accessibility. Those elements should already be focusable, as long as the anchor has a `href` attribute and there isn't any other logic excluding them from being focused. Some insight from accessibility champions might be helpful here

mokimo (approved):
I see some comments with concerns, as long as this is checked with the accessibility team, this is good to go. 👍

PR #1924, opened 2024-02-25
MWPW-143460: Add support for "Mash-Up" marquee

sanrai:
New design of marquees where an author can author a center/left and right marquee.

image

Resolves: [[MWPW-NUMBER](https://jira.corp.adobe.com/browse/MWPW-NUMBER)](https://jira.corp.adobe.com/browse/MWPW-143460)

**Test URLs:**


libs/blocks/caas-marquee/caas-marquee.js
```diff
@@ -217,7 +239,7 @@ async function responseHandler(response, fnName) {
 
 async function getAllMarquees(promoId, origin) {
   const endPoint = isProd() ? API_CONFIGS.caas.prod : API_CONFIGS.caas.stage;
-  const payload = `originSelection=${origin}&promoId=${promoId}&language=en&country=US&perf=true`;
+  const payload = `originSelection=${origin}&promoId=${promoId}&language=en&country=US`;
```
cmiqueo, line 242: Why are we hard coding the language and locale?
we could use something like:
`const [language, country] = document.documentElement.lang.split('-');`

libs/blocks/caas-marquee/caas-marquee.js
```diff
@@ -112,6 +133,7 @@ function getMetadata(el) {
 function isProd() {
   const { host } = window.location;
   return !(host.includes('hlx.page')
+    || host.includes('hlx.live')
     || host.includes('localhost')
     || host.includes('stage.adobe')
     || host.includes('corp.adobe'));
```
cmiqueo, line 139: I would go with something more specific here:
```
|| host.endsWith('.stage.adobe.com')
|| host.endsWith('.corp.adobe.com')
```

libs/blocks/caas-marquee/caas-marquee.css
```diff
@@ -798,3 +798,782 @@
     max-width: 800px;
   }
 }
+
+/* marquee temp support. */
```
cmiqueo, line 802: Is this temporary?

cmiqueo (commented):
Some code suggestions

libs/blocks/caas-marquee/caas-marquee.js
sanrai: This only gets the milo locales. We would need to then map those into the ones for CaaS. 

```
const [language, country] = document.documentElement.lang.split('-');
```

It was easier to just hardcode this for now since we are just supporting en/us rather than dealing with the mappings/imports (which bring potential latency issues for LH scores).

libs/blocks/caas-marquee/caas-marquee.css
sanrai: This line of css was over copied from the reference implementation. I am not sure the context of that css comment.

libs/blocks/caas-marquee/caas-marquee.js
sanrai: Done 👍

sanrai: Actually changing this back as it's failing CodeQL:

image

cmiqueo (approved):

arshadparwaiz (approved):

PR #1925, opened 2024-02-26
fix(MWPW-142926):video modal does not close when esc button is pressed.

sharath-kannan:
The video modal does not close when the esc button is pressed because the video modal does not have any focusable elements other than the fragment url and hence we need to focus on close button inside the modal which will enable this feature.
* An anchor tag with the "hide-video" class can not be a part of the focusable elements, excluding it from the query selector resolves this isssue.

Resolves: [MWPW-142926](https://jira.corp.adobe.com/browse/MWPW-142926)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/fragments/sharath_kannan/pr-explore-carousel-mixed-media?martech=off
- After: https://mwpw-142926--milo--sharath-kannan.hlx.page/fragments/sharath_kannan/pr-explore-carousel-mixed-media?martech=off


libs/blocks/modal/modal.js
```diff
@@ -152,9 +152,11 @@ export async function getModal(details, custom) {
   const focusVisible = { focusVisible: true };
   const focusablesOnLoad = [...dialog.querySelectorAll(FOCUSABLES)];
   const titleOnLoad = dialog.querySelector('h1, h2, h3, h4, h5');
+  const isVideoFragment = dialog.querySelector('.fragment:first-child')?.dataset?.path?.includes("modals/videos") || false;
```
robert-bogos, line 155: No need for `|| false` in here. [includes() will return false if the string is not found](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes).

robert-bogos, line 155: eslint has some complain about the double quotes

robert-bogos (changes requested):
I have tested with the provided URL, but the model still does not close on pressing the Escape key

libs/blocks/modal/modal.js
sharath-kannan: Includes returns a boolean value but this has been added to handle the case where the querySelector does not find a fragment child, in such a scenario the value would be null and since null is not equal to false i have added "||false".

sharath-kannan: I have fixed this error now.

robert-bogos: Still no need for `|| false`. When using the [logical not operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_NOT) in `!isVideoFragment`, it will equal `true` if `isVideoFragment` has any falsy value, that includes the possible `null` from the querySelector not finding it's target.

sharath-kannan: yes, i could use this alternate syntax here.
since i have used the !isVideoFragment inside the "if condition" on line 159 i had added that to improve the understandability of the code.
now that i have stated my train of thought for adding  "||false" do you think its valid or is it understandable without it as well?

JasonHowellSlavin (commented):
Looks good. 

Is there another test we should add for this functionality, or a test to be modified?

libs/blocks/modal/modal.js
robert-bogos: I think it's readable without it as well. Better keep the code in the least verbose version IMO

sharath-kannan: okay, i have updated the code.

robert-bogos (dismissed):
I noticed that if you open the video and press Escape, it closes. However, if you open the video, pause it (or do anything else), and then press Escape, it does not close. 
**Is this the expected behavior?**
I think that as long as the modal is open, it should close when you press Escape, regardless of any subsequent actions (such as pausing the video or adjusting the volume).

sharath-kannan:
> I noticed that if you open the video and press Escape, it closes. However, if you open the video, pause it (or do anything else), and then press Escape, it does not close. **Is this the expected behavior?** I think that as long as the modal is open, it should close when you press Escape, regardless of any subsequent actions (such as pausing the video or adjusting the volume).

This is the expected functionality, as the contents within the modal are inside an iframe there is no way to set the event listeners on the focusable elements inside the iframe with out using post messages. I have also noticed the same behaviour on the non migrated pages as well.
kindly check this non migrated page.
https://www.adobe.com/creativecloud/photography/discover/underwater-text-effect.html

libs/blocks/modal/modal.js
```diff
@@ -152,9 +152,11 @@ export async function getModal(details, custom) {
   const focusVisible = { focusVisible: true };
   const focusablesOnLoad = [...dialog.querySelectorAll(FOCUSABLES)];
   const titleOnLoad = dialog.querySelector('h1, h2, h3, h4, h5');
+  const isVideoFragment = dialog.querySelector('.fragment:first-child')?.dataset?.path?.includes('modals/videos');
```
mokimo, line 155: Should a modal know about all the content it can have, e.g. a video fragment?

sharath-kannan: There are two scenarios for the focusing functionality inside a dialog.
1)dialog modal is aware of all the contents inside it and here we focus on the first focusable element. eg:geo routing pop up.
2)dialog is unaware of its contents during initialization, any dialog that contains an iframe inside it. eg:commerce modal, video modal.
The code would be much simpler if all dialog were of the first type, since we would be focusing on the first focusable element but there's a need for us to handle this specific case because the fragment url is being considered as a focusable element since it is an anchor tag in milo.This anchor is converted into an iframe with a video tag and leaving nothing for us to focus on.

mokimo (commented):
This breaks accessibility. Other elements are not focusable anymore - if you tab through the video, you can't access any of the video controls anymore with the keyboard

sharath-kannan:
> This breaks accessibility. Other elements are not focusable anymore - if you tab through the video, you can't access any of the video controls anymore with the keyboard

This does not break the accessibility of the modal, its just the browser's default behaviour as it doesn't have anywhere else to go after the close button,This can be observed in the "before" url as well when the tabbing cycle is focused on the close button pressing tab again doesn't lead to any other element.
I have seen the same behaviour even in the commerce modals.
https://main--cc--adobecom.hlx.page/creativecloud/animation/discover#animate

mokimo (approved):

Ruchika4 (approved):

Anupamasonnalagi:
Verified for the below URL's. Able to close the Video modal when clicked on Escape button.
https://mwpw-142926--milo--sharath-kannan.hlx.page/fragments/sharath_kannan/pr-explore-carousel-mixed-media?martech=off
https://stage--cc--adobecom.hlx.page/products/premiere/explore?milolibs=mwpw-142926--milo--sharath-kannan

PR #1928, opened 2024-02-26
MWPW-143135: mep page filter to auto match optional .html extension

3ch023:
In Milo, some pages have '.html' extensions on Live, but not on Production:
Live: https://www.adobe.com/products/special-offers.html 
Prod: https://main--cc--adobecom.hlx.live/products/special-offers 

When providing a page filter in the MEP manifest, Authors shouldn't be required to put in a regex to match both cases.
MEP can assume and accept the optional HTML extension.
 
Resolves: [MWPW-143135](https://jira.corp.adobe.com/browse/MWPW-143135)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.live/drafts/mariia/pr/promo/promo?martech=off
- After:  https://htmlextension--milo--adobecom.hlx.live/drafts/mariia/pr/promo/promo?martech=off

### DIFF
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -200,9 +200,9 @@ const consolidateObjects = (arr, prop) => arr.reduce((propMap, item) => {
   return propMap;
 }, {});
 
-const matchGlob = (searchStr, inputStr) => {
+export const matchGlob = (searchStr, inputStr) => {
   const pattern = searchStr.replace(/\*\*/g, '.*');
-  const reg = new RegExp(`^${pattern}$`, 'i'); // devtool bug needs this backtick: `
+  const reg = new RegExp(`^${pattern}(\\.html)?$`, 'i'); // devtool bug needs this backtick: `
   return reg.test(inputStr);
 };
 
--- a/test/features/personalization/personalization.test.js
+++ b/test/features/personalization/personalization.test.js
@@ -3,7 +3,7 @@ import { readFile } from '@web/test-runner-commands';
 import { stub } from 'sinon';
 import { getConfig, setConfig, loadBlock } from '../../../libs/utils/utils.js';
 import initFragments from '../../../libs/blocks/fragment/fragment.js';
-import { applyPers, normalizePath } from '../../../libs/features/personalization/personalization.js';
+import { applyPers, normalizePath, matchGlob } from '../../../libs/features/personalization/personalization.js';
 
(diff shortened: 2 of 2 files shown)


npeltier:
looks good to me but i don't dare to approve in restriction period 😬

mokimo:
@npeltier the branch should be locked now

vgoodric:
@3ch023 I believe this will error if the author has already added the regex you're adding on.  I think you should test if it's needed before appending it for retroactive support.

3ch023:
> @3ch023 I believe this will error if the author has already added the regex you're adding on. I think you should test if it's needed before appending it for retroactive support.

thx @vgoodric , good catch.. i will think how to prevent that

chrischrischris (approved):

3ch023:
> @3ch023 I believe this will error if the author has already added the regex you're adding on. I think you should test if it's needed before appending it for retroactive support.

hi @vgoodric , I've just double-checked and this is backward-compatible, since I add an optional `(.html)?`
so even if author configured `/drafts/mariia/PR/promo/promo.html` as a page filter, the end regex will be
`/drafts/mariia/PR/promo/promo.html(.html)?` which will match /drafts/mariia/PR/promo/promo.html page

I'd say we can leave it without an extra check, but pls let me know if you think it's better to add it.

overmyheadandbody (approved):

vgoodric:
@3ch023 I agree it will work with `/drafts/mariia/PR/promo/promo.html` but I'm worried about people who already did something like `/drafts/mariia/PR/promo/promo(.html)?`

vgoodric (approved):
@3ch023 I was struggling with how to test this and then I realized I could just run the code with a couple of values (removed `export` so I could just run in the console):
```
const matchGlob = (searchStr, inputStr) => {
  const pattern = searchStr.replace(/\*\*/g, '.*');
  const reg = new RegExp(`^${pattern}(\\.html)?$`, 'i'); // devtool bug needs this backtick: `
  return reg.test(inputStr);
};
console.log(matchGlob('/drafts/mariia/PR/promo/promo.html', '/drafts/mariia/PR/promo/promo.html'));
console.log(matchGlob('/drafts/mariia/PR/promo/promo', '/drafts/mariia/PR/promo/promo.html'));
console.log(matchGlob('/drafts/mariia/PR/promo/promo(.html)', '/drafts/mariia/PR/promo/promo.html'));
```

They all return as true without errors which is what we want.  So I agree we're good.

echampio-at-adobe (approved):

PR #1931, opened 2024-02-26
MWPW-142069 - Cache control via URL parameter

elan-tbx:
* Calls to fragments, json (placeholders, charts, etc.) are heavily cached client-side. While this is helpful for page load times, it can cause uncertainty in the authoring workflow over whether or not your latest changes have been published successfully. This feature puts control over usage of the client-side cache into explicit terms via a `cache` URL parameter.
* Under the hood, this feature simply applies caching rules based on [the Request interface](https://developer.mozilla.org/en-US/docs/Web/API/Request/cache). In order to force bypassing the local cache and requesting/storing a new version from the server, add `cache=off` to your URL parameters (for example: https://main--milo--adobecom.hlx.page/?martech=off&cache=off)
* Currently, this cache control is applied to requests for the following:
  * Fragments
  * Placeholders
  * Icons
  * Chart data
* _NOTE: This feature does NOT preview your document changes for you. As an author, you will still need to preview/publish your document changes in order for them to be force-retrieved (via this feature) from the server._

Resolves: [MWPW-142069](https://jira.corp.adobe.com/browse/MWPW-142069)

**Test URLs:**
- Before: https://stage--milo--adobecom.hlx.page/docs/library/kitchen-sink/icon-block?martech=off
- After: https://ebartholomew-cache-control--milo--adobecom.hlx.page/docs/library/kitchen-sink/icon-block?martech=off&cache=off


mokimo:
Stage was out of sync with main, so I force pushed stage - you'll need to rebase onto stage!

overmyheadandbody (commented):
Since the intended target audience would be GWP, is it worth bringing the complexity of allowing all `cache` values via the query parameter? Could we not follow the same `something=off` pattern that we use for other pieces of logic and set it to `reload` under the hood?

elan-tbx:
@overmyheadandbody Fair point, this has been updated accordingly

libs/blocks/chart/chart.js
```diff
@@ -122,7 +122,9 @@ export function processMarkData(series, xUnit) {
 }
 
 export async function fetchData(link) {
-  const resp = await fetch(link.href.toLowerCase());
+  const params = new URLSearchParams(window.location.search);
+  const cache = params.get('cache') === 'off' ? 'reload' : 'default';
+  const resp = await fetch(link.href.toLowerCase(), { cache });
```
Sartxi, line 127: Since the exact same functionality is being used across multiple files, I think we should make this a utility function that is shared by the blocks in need. Making that update also makes it extendable to future blocks, and simplifies testing a bit as well. (You will only have to mock the search param to get code cov, and add a utility test)

elan-tbx:
@JasonHowellSlavin The tests have been updated and are passing successfully 👍

Sartxi (commented):
Looks and works great, I left a comment concerning maintainability, let me know your thoughts.

libs/features/placeholders.js
```diff
@@ -1,3 +1,5 @@
+import { fetchMaybeCache } from '../utils/utils.js';
```
overmyheadandbody, line 1: Couldn't this lead to some circular dependency issues? `utils.js` imports `placeholders.js` (and `icons.js` actually), which then import a method from `utils.js`. I was assuming that's the reason for not having gone this single method path in the first place. Might be worth checking that there are no undesirable side-effects to this.

libs/utils/utils.js
```diff
@@ -1157,3 +1157,9 @@ export function loadLana(options = {}) {
 }
 
 export const reloadPage = () => window.location.reload();
+
+export async function fetchMaybeCache(path) {
```
overmyheadandbody, line 1161: Definitely a personal nit, but I can't say I'm crazy about the name 😅 How about `fetchWithCacheRules` or something of the likes to make the purpose clear?

elan-tbx: Totally fair haha. I was drawing a blank when I wrote it, I like your version better.

libs/features/placeholders.js
elan-tbx: oh wow, THANK YOU. This was the root of the issue I was having with unit tests. Did not realize there were dynamic imports for them buried in `utils.js`.

For now I've moved it to `/libs/utils/helpers.js`... hopefully that's fine 🤷

Sartxi (approved):

libs/blocks/chart/chart.js
```diff
@@ -122,7 +122,8 @@ export function processMarkData(series, xUnit) {
 }
 
 export async function fetchData(link) {
-  const resp = await fetch(link.href.toLowerCase());
+  const { fetchWithCacheRules } = await import('../../utils/helpers.js');
+  const resp = await fetchWithCacheRules(link.href.toLowerCase()).catch(() => ({}));
```
mokimo, line 126: I would propose a slightly different variant
```suggestion
  const { customFetch } = await import('../../utils/helpers.js');
  const resp = await customFetch({
    resource: link.href.toLowerCase(),
    withCacheRules: true
  }).catch(() => ({}));
```

libs/utils/helpers.js
```diff
@@ -19,3 +19,9 @@ export function updateLinkWithLangRoot(link) {
     return link;
   }
 }
+
+export async function fetchWithCacheRules(path) {
+  const params = new URLSearchParams(window.location.search);
+  const cache = params.get('cache') === 'off' ? 'reload' : 'default';
+  return fetch(path, { cache });
+}
```
mokimo, line 27: Given the [global nav already has a use-case](https://github.com/adobecom/milo/blob/stage/libs/blocks/global-navigation/utilities/utilities.js#L301) to do _something_ after fetching which will eventually be ported to a wider range of blocks that could to use this feature, it might make sense to make the fetch a little more generic and allow to plug a few more features as a great spot in the future. Wdyt?
```suggestion
// Makes this method extendable
export async function customFetch({resource, withCacheRules}) {
  const cache = 'default';
  if(withCacheRules) {
    const params = new URLSearchParams(window.location.search);
    params.get('cache') === 'off' ? 'reload' : 
  }
  // if(fetchCentralContent) {
  // apply custom logic
  // }
  return fetch(resource, { cache });
}
```

mokimo (commented):
Cool PR! I'd even expand the current helper a little bit to allow for future extensibility. Would love to hear your thoughts on that idea

JasonHowellSlavin (approved):

mokimo (approved):
Cool 👍

libs/utils/helpers.js
```diff
@@ -19,3 +19,12 @@ export function updateLinkWithLangRoot(link) {
     return link;
   }
 }
+
+export async function customFetch({ resource, withCacheRules }) {
```
overmyheadandbody, line 23: Judging by the use cases so far, I'd suggest we default `withCacheRules` to `true` and remove the property from the method calls

overmyheadandbody (approved):

PR #1938, opened 2024-02-28
Fix: Excluding modals from Active link check

bandana147:
* Excluding the active link check from modal element which was removing the modal element href attribute

Resolves: [[MWPW-NUMBER](https://jira.corp.adobe.com/browse/MWPW-NUMBER)](https://jira.corp.adobe.com/browse/MWPW-143213)

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/products/aftereffects/ae-test-page-localnav-twp
- After: https://main--cc--adobecom.hlx.page/products/aftereffects/ae-test-page-localnav-twp?milolibs=MWPW-143213--milo--bandana147&martech=off

### DIFF
--- a/libs/blocks/global-navigation/utilities/utilities.js
+++ b/libs/blocks/global-navigation/utilities/utilities.js
@@ -241,7 +241,7 @@ export const [hasActiveLink, setActiveLink, getActiveLink] = (() => {
     () => activeLinkFound,
     (val) => { activeLinkFound = !!val; },
     (area) => {
-      if (hasActiveLink() || !(area instanceof HTMLElement)) return null;
+      if (hasActiveLink() || !(area instanceof HTMLElement) || area.querySelector('a').dataset.modalHash) return null;
       const { origin, pathname } = window.location;
       let activeLink;
 


libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -241,7 +241,7 @@ export const [hasActiveLink, setActiveLink, getActiveLink] = (() => {
     () => activeLinkFound,
     (val) => { activeLinkFound = !!val; },
     (area) => {
-      if (hasActiveLink() || !(area instanceof HTMLElement)) return null;
+      if (hasActiveLink() || !(area instanceof HTMLElement) || area.querySelector('a').dataset.modalHash) return null;
```
overmyheadandbody, line 244: An area can contain multiple links. If the first link happens to open a modal, the others would be ignored. The logic should probably be added lower, to check that when a candidate for an active link is found, it doesn't open a modal.

overmyheadandbody (commented):
I don't think this is the right approach for all use-cases, the logic is used for both single links and groups of link, it should tailor to both cases.

The PR should probably also be raised against the `stage` branch, as we're currently in a RCP and starting next week all PRs will need to go through `stage`.

bandana147:
> I don't think this is the right approach for all use-cases, the logic is used for both single links and groups of link, it should tailor to both cases.
> 
> The PR should probably also be raised against the `stage` branch, as we're currently in a RCP and starting next week all PRs will need to go through `stage`.

@overmyheadandbody Noted! I will be closing this PR then and I will address the fix for multiple link use cases and raise a PR for `stage` branch instead.

PR #1947, opened 2024-03-01
[MWPW-143313] IMS scopes based on UNav enablement

overmyheadandbody:
This aims to adapt the IMS `scope` configuration property based on the Universal Navigation enablement on the page.

To make use of the Universal Navigation features, additional scopes need to be used. However, if those scopes are defined, but not configured in IMSS for the respective client ID, the IMS library will fail to load, thus users would no longer be able to sign in or see their profile.

The desire is to define those additional scopes only on pages using the Universal Navigation. This tailors to Milo consumers that have not set their own scopes and instead rely on some sensible defaults.

The change has been made in two places:
* _scripts.js_ - this is where the Milo-specific configs are defined, we want to have this change here to make sure the Universal Navigation feature works correctly on Milo pages, _main--milo--adobe.com/hlx.(page|live)_;
* _utils.js_ - since this is being used by all Milo consumers, we need to consider two cases:
    * the consumer already has scopes defined - in this case, their scopes are used;
    * no scopes have been defined by the consumer - in this case, the basic scopes would be defined, unless they're using the Universal Navigation, in which case the additional scopes get appended as well.

**NOTE**: even though all consumers call `setConfig` in their scripts, as suggested in [Milo College](https://github.com/adobecom/milo-college/blob/main/scripts/scripts.js#L61), there aren't any defaults being set for IMS, as can be seen in the `setConfig` [method](https://github.com/adobecom/milo/blob/main/libs/utils/utils.js#L209-L236). Thus, it seems reasonable to me to adapt both files where scopes are declared.

**Testing instructions**: on a test page, open the browser's developer tools and check the value for `adobeid.scope`.

Resolves: [MWPW-143313](https://jira.corp.adobe.com/browse/MWPW-143313)

**Test URLs:**
- Before (additional scopes are set regardless of the Universal Navigation enablement):  https://main--milo--adobecom.hlx.page/drafts/ramuntea/gnav-refactor?martech=off&georouting=off
- After (basic scopes are set): https://gnav--milo--adobecom.hlx.page/drafts/ramuntea/gnav-refactor?martech=off&georouting=off
- After (additional scopes are used on Universal Navigation enabled page): https://gnav--milo--adobecom.hlx.page/drafts/ramuntea/universal-nav?martech=off&georouting=off

### DIFF
--- a/libs/scripts/scripts.js
+++ b/libs/scripts/scripts.js
@@ -11,7 +11,6 @@
  */
 
 import {
-  getMetadata,
   loadArea,
   loadLana,
   setConfig,
@@ -120,14 +119,11 @@ const locales = {
   sea: { ietf: 'en', tk: 'hah7vzn.css' },
 };
 
-const unavMeta = getMetadata('universal-nav')?.trim();
-
 const config = {
   geoRouting: 'on',
   fallbackRouting: 'on',
   links: 'on',
   imsClientId: 'milo',
-  imsScope: `AdobeID,openid,gnav${unavMeta && unavMeta !== 'off' ? ',pps.read,firefly_api' : ''}`,
   codeRoot: '/libs',
   locales,
   prodDomains,
--- a/libs/scripts/scripts.js
+++ b/libs/scripts/scripts.js
@@ -11,6 +11,7 @@
  */
 
 import {
+  getMetadata,
   loadArea,
   loadLana,
   setConfig,
@@ -119,12 +120,14 @@ const locales = {
   sea: { ietf: 'en', tk: 'hah7vzn.css' },
 };
 
+const unavMeta = getMetadata('universal-nav')?.trim();
+
 const config = {
   geoRouting: 'on',
   fallbackRouting: 'on',
   links: 'on',
   imsClientId: 'milo',
-  imsScope: 'AdobeID,openid,gnav,pps.read,firefly_api',
+  imsScope: `AdobeID,openid,gnav${unavMeta && unavMeta !== 'off' ? ',pps.read,firefly_api' : ''}`,
   codeRoot: '/libs',
   locales,
   prodDomains,
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -775,10 +775,12 @@ export async function loadIms() {
       reject(new Error('Missing IMS Client ID'));
       return;
     }
+    const unavMeta = getMetadata('universal-nav')?.trim();
+    const defaultScope = `AdobeID,openid,gnav${unavMeta && unavMeta !== 'off' ? ',pps.read,firefly_api' : ''}`;
     const timeout = setTimeout(() => reject(new Error('IMS timeout')), 5000);
     window.adobeid = {
       client_id: imsClientId,
-      scope: imsScope || 'AdobeID,openid,gnav,pps.read,firefly_api',
+      scope: imsScope || defaultScope,
       locale: locale?.ietf?.replace('-', '_') || 'en_US',
       autoValidateToken: true,
       environment: env.ims,


libs/scripts/scripts.js
```diff
@@ -119,12 +120,14 @@ const locales = {
...
 };
 
+const unavMeta = getMetadata('universal-nav')?.trim();
+
 const config = {
   geoRouting: 'on',
   fallbackRouting: 'on',
   links: 'on',
   imsClientId: 'milo',
-  imsScope: 'AdobeID,openid,gnav,pps.read,firefly_api',
```
mokimo, line 127: For the milo token, this should be irrelevant since it has the correct scopes, no?

mokimo (approved):

robert-bogos (approved):

SilviuLCF (approved):
QA verified , approved for stage

Testing details https://jira.corp.adobe.com/browse/MWPW-143313

libs/scripts/scripts.js
overmyheadandbody: Indeed. The use-case I was thinking of is when a developer would overwrite the client ID to something that suits their testing purposes. That client ID might not be onboarded to these additional scopes and it might not be obvious why IMS is no longer working; not sure if scopes are the first culprit for everyone in such situations. So I decided to keep the logic here as well. However, you do have a point - when everyone will be using the Universal Nav, then basing scopes on the feature would no longer be relevant.

With this in mind, would you still just leave the static scopes or have the dynamically built one until the Universal Nav becomes standard?

narcis-radu: What if we do something like:
```js
const config = {
    // some existing config
    imsScope: 'AdobeID,openid,gnav',
    // other config
};

const unavMeta = getMetadata('universal-nav')?.trim();
if (unavMeta && unavMeta !== 'off') config.imsScope += ',pps.read,firefly_api';
```
I personally find this more readable and would allow folks to add more default scopes _before_ the ones needed for universal navigation.

libs/utils/utils.js
```diff
@@ -775,10 +775,12 @@ export async function loadIms() {
       reject(new Error('Missing IMS Client ID'));
       return;
     }
+    const unavMeta = getMetadata('universal-nav')?.trim();
+    const defaultScope = `AdobeID,openid,gnav${unavMeta && unavMeta !== 'off' ? ',pps.read,firefly_api' : ''}`;
     const timeout = setTimeout(() => reject(new Error('IMS timeout')), 5000);
     window.adobeid = {
       client_id: imsClientId,
-      scope: imsScope || 'AdobeID,openid,gnav,pps.read,firefly_api',
+      scope: imsScope || defaultScope,
```
narcis-radu, line 783: is this needed? we have two use-cases:
1. consumer does not override the scope: `getConfig()` will return the default Milo scope from https://github.com/adobecom/milo/pull/1947/files#diff-12979217e83810ba83417c3e7b030afd400eff7c83cd6390d9f9de3a337618ecR130
2. consumer sets a custom scope: `getConfig()` will return the scope set by the consumer.

@mokimo - do you remember why we added this fallback in the first place?

mokimo: The first use-case doesn't sound correct [this](https://github.com/adobecom/milo/pull/1947/files#diff-12979217e83810ba83417c3e7b030afd400eff7c83cd6390d9f9de3a337618ecR130) isn't the default scope - it's the milo specific scope. Consumers will never inherit anything in `scripts.js` nor load the milo core one at all.

So we do need the defaultScope in case a consumer doesn't supply one.

narcis-radu: @mokimo - thank you, that's correct! In that case it makes sense to keep the default scope, but maybe we can use the pattern described above - https://github.com/adobecom/milo/pull/1947#discussion_r1511032456.

libs/scripts/scripts.js
overmyheadandbody: The reason I didn't go for this is that I wanted the scopes to be grouped in a single location. Otherwise one would look and see just 3 scopes, while on a UNav-enabled page they see 5. They'd have to go back to the file and check additional logic, whereas my suggested change has everything in one line so the additional logic is clear from the begining.

libs/utils/utils.js
overmyheadandbody: Hey @narcis-radu, the note section in the PR body was based on a previous conversation we had:
> **NOTE**: even though all consumers call setConfig in their scripts, as suggested in [Milo College](https://github.com/adobecom/milo-college/blob/main/scripts/scripts.js#L61), there aren't any defaults being set for IMS, as can be seen in the setConfig [method](https://github.com/adobecom/milo/blob/main/libs/utils/utils.js#L209-L236). Thus, it seems reasonable to me to adapt both files where scopes are declared.

It's the same as @mokimo mentioned, that consumers don't use the Milo-specific configs under any circumstance..

I also added a comment on why I think it's best to keep the config as a one-liner, to make it instantly clear what the possible IMS scopes are and what influences them.

libs/scripts/scripts.js
mokimo: > With this in mind, would you still just leave the static scopes or have the dynamically built one until the Universal Nav becomes standard?

Should we not set any config at all and consume the standard config? That way we consume them in a dog-food manner as a milo college project would

overmyheadandbody: Had a look over the history of this line and saw that @narcis-radu first introduced it [here](https://github.com/adobecom/milo/pull/736/files#diff-12979217e83810ba83417c3e7b030afd400eff7c83cd6390d9f9de3a337618ec). So with the current way we load IMS, I'd say it's safe to remove the declaration from here altogether and just rely on the sensible default, as you're suggesting.

libs/utils/utils.js
narcis-radu: ok, we can go with that since the default `imsScope` does not change that often. we should also look into removing the default config value as you mentioned, but it can be a separate PR.

overmyheadandbody: Added a new commit to remove the config from scripts.js

mokimo (approved):

SilviuLCF (approved):
QA verified

narcis-radu:
@SilviuLCF - has this been tested for all Milo consumers as well ?
@overmyheadandbody

SilviuLCF:
> @SilviuLCF - has this been tested for all Milo consumers as well ? @overmyheadandbody

Yes, details here https://jira.corp.adobe.com/browse/MWPW-143313?

PR #1949, opened 2024-03-04
MWPW-143655

Blainegunn:
**CODEOWNERS for Main**
This only contains T1 Consumers / Sign-off teams on the `main` branch now. The Original `CODEOWNERS` will remain as is on `stage`.

Resolves: [MWPW-143655](MWPW-143655)
Linked to: https://github.com/adobecom/milo/pull/1950
**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/?martech=off
- After: https://mwpw-143655--milo--adobecom.hlx.page/?martech=off

PR #1950, opened 2024-03-04
MWPW-139416

Blainegunn:
Moving `CODEOWNERS` so it the changes can be ignored and not passed to `main`
Auto Raise PR workflow to help managing stage process.
Resolves: [MWPW-139416](https://jira.corp.adobe.com/browse/MWPW-139416)

Linked to: https://github.com/adobecom/milo/pull/1949

**Test URLs:**
- After: https://MWPW-139416--milo--adobecom.hlx.page/?martech=off


.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,18 @@
...
+      - stage
+
+jobs:
+  pull-request:
+    runs-on: ubuntu-latest
+    steps:
+      - name: pull-request
+        uses: actions/checkout@v3
+      - name: Create Pull Request
+        uses: diillson/auto-pull-request@v1.0.1
```
mokimo, line 13: Can we clone this action? Seems like the safer thing to do & we will likely need to adapt it anyway

mokimo (commented):
I don't understand on how this would work with [this PR](https://github.com/adobecom/milo/pull/1949)

Once you merge stage into main, you would always include the `.github/CODEOWNERS` file and run into merge conflicts

mokimo (changes requested):
Given the pending merge-conflict question & security risk assosciated with the action, this will need an iteration.

Blainegunn:
@mokimo 
I forgot to include the `.gitignore` update. It's here now and will insure there's no conflicts. Already tested on other projects.

As far as the security concerns go. 
>GitHub Actions can only read a secret if you explicitly include the secret in a workflow.

from, [Using secrets in GitHub Actions](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions)

.gitignore
```diff
@@ -6,3 +6,4 @@ node_modules/*
 .DS_Store
 .idea
 .iml
+.github/CODEOWNERS
```
mokimo, line 9: So in this case, what is the model of updating the codeowners, since changes are ignored? 🤔

mokimo: I am fairly sure tracked files are no longer ignored and this will lead to merge conflicts, I don’t mind testing this out though - if u previously tested this on a fork and didn’t encounter any issues.

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,18 @@
...
+        with:
+          destination_branch: "main"
+          pr_title: "Ready for Main"
+          pr_body: "Merge to main, commit: ${{ github.sha }}"
```
mokimo, line 17: This will fail certain checks, such as the PSI tests.
Can we actually get a meaningful PR template that also includes links to all the relevant T1 stage pages & a publicly accessible milo one for PSI? It would always be main—milo—adobecom & stage—milo—adobecom

.gitignore
mokimo: Tested this, this leads to merge conflicts. https://github.com/mokimo/milo/pull/3

Blainegunn: @mokimo the ignore [file](https://github.com/mokimo/milo/blob/main/.gitignore) isn't merged yet

mokimo: It is on stage, but even merging it on main is irrelevant - I've included it on main as well and still has the same conflicts.

.github/workflows/pull-request.yml
mokimo: I would not say the PR description does not matter. Anyone can look into those PRs, the descriptions serve an important role from a git/history perspective - also the SOTs & also devs should have a few links handy to click and go through each project.

mokimo: An engineer that looks into stage->main sync, should also always be able to know what changes are about to go live and block them if any issues are encountered. We want to improve the process, not shift every responsibility away from an engineer to the SOTs

honstar: I agree with @mokimo here: having meaningful PR descriptions will definitely be helpful for reviewers and approvers to understand the context of changes planned to be merged into `main`.

Since checks are run regardless of the branch, we should make sure they are green no matter what.

.gitignore
honstar: Looks like this setup needs a bit of rework to make this work.

PR #1951, opened 2024-03-04
MWPW-141020: [Project PEP] Implement Analytics Enablement

robert-bogos:
## Description
This PR enables analytics for PEP

## Related Issue
Resolves: [MWPW-141020](https://jira.corp.adobe.com/browse/MWPW-141020)

## Testing instructions
1. open the `after test URL` and wait for the PEP modal to show up
2. open dev tools and watch the network tab, searching for calls to `https://sstats.adobe.com/ee/irl1/v1/interact`
3. click on the X button
4. the request payload for searched api call should have `events[0].xdm.web.webInteraction.name` equal to `"Close Modal|PEP Modal_Photoshop"` ( see screenshots )
5. delete the `dismissedAppPrompts` cookie so that the modal will show up again on refresh
6. click the "Stay on this page" button
7. the request payload for searched api call should have `events[0].xdm.web.webInteraction.name` equal to `"Stay on this page|PEP Modal_Photoshop"` ( see screenshots )

## Screenshots
Screenshot 2024-03-04 at 11 06 14
Screenshot 2024-03-04 at 11 04 07

## Test URLs
**Milo:**
- Before: https://project-pep--milo--adobecom.hlx.page/drafts/rbogos/pep-prompt?imsClientId=fedsmilo
- After: https://mwpw-141020-pep-analytics--milo--robert-bogos.hlx.page/drafts/rbogos/pep-prompt?imsClientId=fedsmilo

### DIFF
--- a/libs/features/webapp-prompt/webapp-prompt.js
+++ b/libs/features/webapp-prompt/webapp-prompt.js
@@ -11,8 +11,6 @@ const CONFIG = {
   icon: '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 133.5 118.1" title="Adobe, Inc."><defs><style>.cls-1 {fill: #eb1000;}</style></defs><g><g><polygon class="cls-1" points="84.1 0 133.5 0 133.5 118.1 84.1 0"/><polygon class="cls-1" points="49.4 0 0 0 0 118.1 49.4 0"/><polygon class="cls-1" points="66.7 43.5 98.2 118.1 77.6 118.1 68.2 94.4 45.2 94.4 66.7 43.5"/></g></g></svg>',
 };
 
-const capitalize = (str) => str.charAt(0).toUpperCase() + str.slice(1);
-
 const getElemText = (elem) => elem?.textContent?.trim().toLowerCase();
 
 // Note: this might be used outside of Milo,
@@ -124,7 +122,7 @@ class AppPrompt {
       </div>`
       : '';
 
-    return toFragment`<div daa-state="true" daa-im="true" daa-lh="PEP Modal_${capitalize(this.options['product-name'])}" class="appPrompt">
+    return toFragment`<div daa-state="true" daa-im="true" daa-lh="PEP Modal_${this.options['product-name']}" class="appPrompt">
       ${this.elements.closeIcon}
       <div class="appPrompt-icon">
         ${this.image}
--- a/libs/features/webapp-prompt/webapp-prompt.js
+++ b/libs/features/webapp-prompt/webapp-prompt.js
@@ -45,7 +45,7 @@ class AppPrompt {
     await this.getData(content);
     // TODO: we might need to set this at the page level through metadata
     // so that the same prompt can lead to different apps
-    if (!this.options['redirect-url']) return;
+    if (!this.options['redirect-url'] || !this.options['product-name']) return;
 
     this.template = this.decorate();
 
--- a/libs/features/webapp-prompt/webapp-prompt.js
+++ b/libs/features/webapp-prompt/webapp-prompt.js
@@ -11,7 +11,7 @@ const CONFIG = {
   icon: '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 133.5 118.1" title="Adobe, Inc."><defs><style>.cls-1 {fill: #eb1000;}</style></defs><g><g><polygon class="cls-1" points="84.1 0 133.5 0 133.5 118.1 84.1 0"/><polygon class="cls-1" points="49.4 0 0 0 0 118.1 49.4 0"/><polygon class="cls-1" points="66.7 43.5 98.2 118.1 77.6 118.1 68.2 94.4 45.2 94.4 66.7 43.5"/></g></g></svg>',
 };
 
-const capitalize = (str) => str && (str.charAt(0).toUpperCase() + str.slice(1));
+const capitalize = (str) => str.charAt(0).toUpperCase() + str.slice(1);
 
 const getElemText = (elem) => elem?.textContent?.trim().toLowerCase();
 
--- a/libs/features/webapp-prompt/webapp-prompt.js
+++ b/libs/features/webapp-prompt/webapp-prompt.js
@@ -11,7 +11,7 @@ const CONFIG = {
   icon: '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 133.5 118.1" title="Adobe, Inc."><defs><style>.cls-1 {fill: #eb1000;}</style></defs><g><g><polygon class="cls-1" points="84.1 0 133.5 0 133.5 118.1 84.1 0"/><polygon class="cls-1" points="49.4 0 0 0 0 118.1 49.4 0"/><polygon class="cls-1" points="66.7 43.5 98.2 118.1 77.6 118.1 68.2 94.4 45.2 94.4 66.7 43.5"/></g></g></svg>',
 };
 
-const capitalize = (str) => str.charAt(0).toUpperCase() + str.slice(1);
+const capitalize = (str) => str && (str.charAt(0).toUpperCase() + str.slice(1));
 
 const getElemText = (elem) => elem?.textContent?.trim().toLowerCase();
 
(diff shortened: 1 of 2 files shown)


libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -108,8 +111,8 @@ class AppPrompt {
 
   // TODO: should we allow for app icon to be set as SVG?
   decorate = () => {
-    this.elements.closeIcon = toFragment`<button class="appPrompt-close"></button>`;
-    this.elements.cta = toFragment`<button class="appPrompt-cta appPrompt-cta--close">${this.cancelText}</button>`;
+    this.elements.closeIcon = toFragment`<button daa-ll="Close Modal" class="appPrompt-close"></button>`;
+    this.elements.cta = toFragment`<button daa-ll="${this.cancelText}" class="appPrompt-cta appPrompt-cta--close">${this.cancelText}</button>`;
```
mokimo, line 115: What is special about those attributes, we usually used the this [utility](https://github.com/adobecom/milo/blob/stage/libs/blocks/global-navigation/utilities/utilities.js#L124) to process any analytics values

mokimo: We already use `toFragment` - also the microbundle would just use that specific function and tree-shake the rest, so it's safe to use

overmyheadandbody, line 115: Is analytics looking for the translated text or the English one? Usually it's the latter they want, so placeholders might not be the right approach here.

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -25,12 +27,13 @@ const getMetadata = (el) => [...el.childNodes].reduce((acc, row) => {
...
 class AppPrompt {
-  constructor({ promptPath, id, profileApi } = {}) {
+  constructor({ promptPath, id, profileApi, entName } = {}) {
     this.promptPath = promptPath;
     this.id = id;
     this.profileApi = profileApi;
     this.elements = {};
-
+    // TODO - make sure this is still valid after we have the entitlements mapping
+    this.productName = capitalize(entName);
```
overmyheadandbody, line 36: If a product name is needed for analytics purposes, we need to go back to a previous requirement, as the entitlement name would not be sufficiently clear. An entitlement name would be something like 'isEntitledAndHasUsedPhotoshop', whereas we just want 'Photoshop'.

We need to add a new field to the section metadata of the document to allow authors to define the product name that should be used here.

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -122,7 +125,7 @@ class AppPrompt {
       </div>`
       : '';
 
-    return toFragment`<div class="appPrompt">
+    return toFragment`<div daa-state="true" daa-im="true" daa-lh="PEP Modal_${this.productName}" class="appPrompt">
```
overmyheadandbody, line 128: Same product name comment as opposed to using entitlement

libs/scripts/delayed.js
```diff
@@ -97,7 +97,7 @@ export const loadAppPrompt = async (getConfig, getMetadata, loadStyle) => {
     loadStyle(`${base}/features/webapp-prompt/webapp-prompt.css`),
   ]);
 
-  webappPrompt.default({ promptPath, id, profileApi });
+  webappPrompt.default({ promptPath, id, profileApi, entName });
```
overmyheadandbody, line 100: Same comment regarding product name, entitlement is probably not needed here.

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -11,6 +11,8 @@ const CONFIG = {
   icon: '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 133.5 118.1" title="Adobe, Inc."><defs><style>.cls-1 {fill: #eb1000;}</style></defs><g><g><polygon class="cls-1" points="84.1 0 133.5 0 133.5 118.1 84.1 0"/><polygon class="cls-1" points="49.4 0 0 0 0 118.1 49.4 0"/><polygon class="cls-1" points="66.7 43.5 98.2 118.1 77.6 118.1 68.2 94.4 45.2 94.4 66.7 43.5"/></g></g></svg>',
 };
 
+const capitalize = (str) => str.charAt(0).toUpperCase() + str.slice(1);
+
```
overmyheadandbody, line 15: If we go with authored product name, this should no longer be required

overmyheadandbody (changes requested):
Using entitlements would not lead to the expected labels, we should have a new product name field instead.

libs/features/webapp-prompt/webapp-prompt.js
narcis-radu: Agreed!

libs/features/webapp-prompt/webapp-prompt.js
narcis-radu, line 115: I assume `this.cancelText` will be a static value, otherwise we would get multiple analytics for the same action.

robert-bogos: Changed to static values so I don't think we need that utility function anymore. From what I saw, the purpose of that was to ensure no invalid characters are present, which could be the case for dynamic values, but not for static.

robert-bogos: There was no specific requirement for translated text, so I changed it to the English version, which makes more sense 👍. I initially opted for the translated version because I observed this approach being used in some places in the global navigation.

libs/features/webapp-prompt/webapp-prompt.js
robert-bogos: Done 👍

libs/scripts/delayed.js
robert-bogos: Done 👍

libs/features/webapp-prompt/webapp-prompt.js
robert-bogos: I think this would still be required as the metadata value is not case sensitive I noticed. Am I missing something ?

libs/features/webapp-prompt/webapp-prompt.js
robert-bogos: Changed to a static value, so we don't have `this.cancelText` in there anymore 👍

libs/features/webapp-prompt/webapp-prompt.js
overmyheadandbody: In my view, this shouldn't be a deal breaker for analysts. We can ask authors to capitalize the product name, but ultimately, as long as the analytics values are consistent for the same prompt (which they should be, since we're using a fragment-like pattern), analysts would be able to get the necessary data. So TL;DR, I vote for this being removed.

robert-bogos: Asking the authors to capitalize won't actually help because the values are lowercased in [here](https://github.com/adobecom/milo/blob/project-pep/libs/features/webapp-prompt/webapp-prompt.js#L14). I can leave it lowercase, which will be consistent for the analytics, but won't be according to the excel requirements file, which is capitalized.
What do you think ?

mokimo (approved):

libs/features/webapp-prompt/webapp-prompt.js
robert-bogos: Talked with analytics and they are ok with lowercase. So I removed that function 👍

SilviuLCF (approved):

SilviuLCF:
Ready for STAGE 
QA verified - https://jira.corp.adobe.com/browse/MWPW-141020

honstar:
As per JIRA ticket, this PR is `Ready for Stage`, hence marking respectively.

overmyheadandbody:
Hey @honstar! This was actually opened against a feature branch, the code for the whole functionality is not ready for stage or production, as we're waiting for additional details. I've added this detail to the stage PR already.

It might make sense to bookmark a filter with stories ready to be merged on one of the Slack channels, such as https://github.com/adobecom/milo/pulls?q=is%3Aopen+is%3Apr+label%3A%22Ready+for+Stage%22+base%3Astage

As a personal opinion, I think it's the developer's job to make sure they follow all the steps to have their PR looked at by the MPoC, MPoCs should not have to sort out PRs. We should probably also come up with a rule that if something is open for more than X weeks, it automatically gets closed.

CC: @SilviuLCF

PR #1952, opened 2024-03-04
MWPW-142130: Update the source of merch-card 

yesil:
Resolves: [MWPW-142130](https://jira.corp.adobe.com/browse/MWPW-142130)

Includes: https://git.corp.adobe.com/wcms/tacocat.js/pull/536

As per the [discussion](https://github.com/orgs/adobecom/discussions/1847),  the source of card name has been moved to the tags section. Card name and card title can be used as a sort key for sorting cards in different filters in `merch-cards` block.

**Test URLs:**
- After: https://mwpw-142130-stage--milo--yesil.hlx.page/docs/library/kitchen-sink/merch-card?martech=off

CC Test URLs: 
Catalog page:
https://main--cc--adobecom.hlx.live/drafts/ilyas/catalog?milolibs=mwpw-142130-stage--milo--yesil

Test page: 
https://main--cc--adobecom.hlx.live/drafts/ilyas/merch-cards-sorting?milolibs=mwpw-142130-stage--milo--yesil


libs/blocks/merch-card/merch-card.js
```diff
@@ -371,11 +273,15 @@ const init = async (el) => {
   if (el.dataset.removedManifestId) {
     merchCard.dataset.removedManifestId = el.dataset.removedManifestId;
   }
-  if (name) {
-    merchCard.setAttribute('name', name);
-  }
   let tags = {};
-  if (el.lastElementChild) {
+  if (el.lastElementChild?.previousElementSibling?.querySelector('h3,h4,h5,h6')) {
```
npeltier, line 277: am i right to assume you aim at heading in the last section of the card that is the one for the tags? doing this we enforce that:
- only one heading should be in that section,
- no other section that could hold headings should be added after the tag section
right?

yesil: If a heading is present in the first section of the block we ignore it.
>only one heading should be in that section:

we pick the first heading in the last section. But the others are ignored but do not break the feature.

>no other section that could hold headings should be added after the tag section
right?

yes.

VKniaz (approved):

mokimo (approved):

yesil:
@Blainegunn the kitchensink page is not optimized for performance and ~75 is its average score.
It is good to merge.

PR #1956, opened 2024-03-04
MWPW-139939 - Support for quiz state in local storage

colloyd:
* Renders the quiz at any point as captured in local storage

Resolves: [MWPW-139939](https://jira.corp.adobe.com/browse/MWPW-139939)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/colloyd/quiz/?martech=off
- After: https://mwpw-139939-load-userflow-from-local-storage--milo--colloyd.hlx.page/drafts/colloyd/quiz/?martech=off

Testing Notes: To test this code, you will need to load up the URLs above and set the following in local storage for each
- key: stored-quiz-state
- value: {"userFlow":["q-rather","q-photo","q-video","q-design"],"userSelection":[{"selectedQuestion":{"questions":"q-category","max-selections":"3","min-selections":"1"},"selectedCards":{"photo":true,"video":true,"design":true}}]}

In the before page, the quiz will load the first question as normal. In the after page, the quiz will load up on the second question with the first question already answered with Photo, Video and Design options pre-chosen You can verify this by choosing "Take the time to control every detail" on the second question. You should then see questions corresponding to each of the 3 pre-chosen options from question one.

### DIFF
--- a/libs/blocks/quiz/quiz.js
+++ b/libs/blocks/quiz/quiz.js
@@ -202,6 +202,8 @@ const App = ({
     const nextQuizViewsLen = nextQuizViews.length;
     const [firstQuizView] = nextQuizViews;
 
+    localStorage.removeItem('stored-quiz-state');
+
     if (nextQuizViewsLen === 1 && isValidUrl(firstQuizView)) {
       window.location.href = firstQuizView;
       return;
--- a/libs/blocks/quiz/quiz.js
+++ b/libs/blocks/quiz/quiz.js
@@ -21,7 +21,7 @@ export async function loadFragments(fragmentURL) {
 const App = ({
   initialIsDataLoaded = false,
   preQuestions = {}, initialStrings = {}, shortQuiz: isShortQuiz = false,
-  preselections = [], nextQuizViewsExist: preNextQuizViewsExist = true, storedQuizState = {},
+  preselections = [], nextQuizViewsExist: preNextQuizViewsExist = true, storedQuizState = true,
 }) => {
   const [btnAnalytics, setBtnAnalytics] = useState(null);
   const [countSelectedCards, setCountOfSelectedCards] = useState(0);
@@ -339,7 +339,7 @@ export default async function init(
 ) {
   const configData = initConfigPathGlob(el);
   const updatedShortQuiz = shortQuiz || configData.shortQuiz;
-  let storedQuizState = localStorage.getItem('stored-quiz-state');
+  let storedQuizState = localStorage.getItem('stored-quiz-state') || {};
 
   try {
     storedQuizState = JSON.parse(storedQuizState);
--- a/libs/blocks/quiz/quiz.js
+++ b/libs/blocks/quiz/quiz.js
@@ -21,7 +21,7 @@ export async function loadFragments(fragmentURL) {
 const App = ({
   initialIsDataLoaded = false,
   preQuestions = {}, initialStrings = {}, shortQuiz: isShortQuiz = false,
-  preselections = [], nextQuizViewsExist: preNextQuizViewsExist = true,
+  preselections = [], nextQuizViewsExist: preNextQuizViewsExist = true, storedQuizState = {},
 }) => {
   const [btnAnalytics, setBtnAnalytics] = useState(null);
   const [countSelectedCards, setCountOfSelectedCards] = useState(0);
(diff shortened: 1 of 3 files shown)


echampio-at-adobe:
Neat!. In the hackathon prototype:

https://mwpw-139939-load-userflow-from-local-storage--milo--colloyd.hlx.page/drafts/quiz/quiz-ai/?martech=off

I was able to hack in ficode values into either q-category or into the q-ficode question that only exists in the results workbook. 

localStorage.setItem('stored-quiz-state', '{"userFlow":["q-rather","q-customer"],"userSelection":[{"selectedQuestion":{"questions":"q-category","max-selections":"1","min-selections":"1"},"selectedCards":{"photoshop_cc":true}}]}');

localStorage.setItem('stored-quiz-state', '{"userFlow":["q-rather","q-customer"],"userSelection":[{"selectedQuestion":{"questions":"q-ficode","max-selections":"1","min-selections":"1"},"selectedCards":{"illustrator_cc":true}}]}');

Even though that isolated q-ficode approach might work using this pathway, I still think the approach we all discussed this morning, of having the api return codes listed under q-category will work best. 

Your code seemed to happily ignore the invalid stored data I threw at it too. 

localStorage.setItem('stored-quiz-state', '{"userFlow":["q-rather","q-customer"],"userSelection":[{"selectedQuestion":{"questions":"q-ficode","max-selections":"1","min-selections":"1"},"selectedCards":{}]}');
localStorage.setItem('stored-quiz-state', '{"userFlow":[],"userSelection":[{"selectedQuestion":{"questions":"q-ficode","max-selections":"1","min-selections":"1"},"selectedCards":{"illustrator_cc":true}}]}');
localStorage.setItem('stored-quiz-state', '{"userFlow":{},"userSelection":{}}');

libs/blocks/quiz/quiz.js
```diff
@@ -56,7 +56,14 @@ const App = ({
         strMap[question.q] = question;
       });
 
-      setUserFlow([questions.questions.data[0].questions]);
+      if (!!Object.keys(storedQuizState).length
+        && !!storedQuizState?.userFlow.length
+        && !!storedQuizState?.userSelection.length) {
+        setUserFlow(storedQuizState.userFlow);
+        updateUserSelection(storedQuizState.userSelection);
```
fullcolorcoder, line 63: Missing test coverage

fullcolorcoder (changes requested):
This is working great so far. The only gotcha I can find is there is no destruction of the storedQuizState yet, which actually is subtly a new feature. If you hit refresh, close, or navigate away and come back, the quiz is still on question 2.

I think it would be awesome to see it get destroyed on the next button click, and then actually re-created right away with the new selections/state so if they do close accidentally then we can put them right where they were easily. 

Perhaps some scope creep for another task, but some destruction of the state once were in would be good here.

libs/blocks/quiz/quiz.js
echampio-at-adobe: Thanks, @fullcolorcoder, for noting that the state wasn't cleared. I logged a story for that update. https://jira.corp.adobe.com/browse/MWPW-143761

echampio-at-adobe: And on reflection, I moved my story to a sub-task of the current story. We may as well make the changes before merging this one.

JackySun9 (approved):
This is cool, small change, make big impact, thanks!

libs/blocks/quiz/quiz.js
colloyd: @fullcolorcoder This should now be cleared as soon as handleOnNextClick() is called - when the next button is clicked for the first time after loading a stored quiz state

colloyd: Also, regarding test coverage, parts of this will change when we bring in the ML field. I want to have a complete picture before I add the tests.

echampio-at-adobe (approved):
The latest change to clear the stored quiz state when hitting the next button seems to be working well.

fullcolorcoder (approved):

libs/blocks/quiz/quiz.js
```diff
@@ -332,6 +341,14 @@ export default async function init(
 ) {
   const configData = initConfigPathGlob(el);
   const updatedShortQuiz = shortQuiz || configData.shortQuiz;
+  let storedQuizState = localStorage.getItem('stored-quiz-state') || {};
```
mokimo, line 344: I don't think you need the default, as you'd set it in the catch block if JSON.parse would fail

libs/blocks/quiz/quiz.js
```diff
@@ -56,7 +56,14 @@ const App = ({
         strMap[question.q] = question;
       });
 
-      setUserFlow([questions.questions.data[0].questions]);
+      if (!!Object.keys(storedQuizState).length
```
mokimo, line 59: I'm fairly sure you don't need to convert those to bools

mokimo (approved):
A unit test for this would be cool

JackySun9 (approved):

fullcolorcoder (approved):

PR #1957, opened 2024-03-05
MWPW-143648 - [Action Item] Fix float icon position

Sartxi:
In the Action Item float icon variant the icon position is inconsistent in fire fox browser when compared to chrome. This PR adds the proper css to make the icons position match across the browsers.

**_Please verify this fixes issues in FireFox_**

Resolves: [MWPW-143648](https://jira.corp.adobe.com/browse/MWPW-143648)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/sarchibeque/action-item-cross-browser?martech=off
- After: https://sarchibeque-mwpw-143648-action-item--milo--adobecom.hlx.page/drafts/sarchibeque/action-item-cross-browser?martech=off


ryanmparrish (approved):
Looks good. I wasn't able to see the issue initially on the link in the description. 
Here are the one's that I was able to see it on though. 

Before
https://main--milo--adobecom.hlx.page/docs/library/kitchen-sink/action-items
After
https://sarchibeque-mwpw-143648-action-item--milo--adobecom.hlx.page/docs/library/kitchen-sink/action-items

mokimo (approved):

libs/blocks/action-item/action-item.css
```diff
@@ -133,8 +133,12 @@
   transform: scale(1.05);
 }
 
+.action-item.float-icon .main-image {
+  display: block;
+}
+
```
rgclayton, line 139: Is there a reason why you don't want `.action-item .main-image` to be a block level element? 

I noticed `.main-image` is added to elements that are inline by default (`<strong>` and `<p>`). If you declare those to be block level it doesn't look like you need to change `.action-item.float-icon .floated-icon` to be `inline-flex`. One less change, and one less thing altered. :). Maybe i'm missing something?

Sartxi: @rgclayton You make a solid point. I was really just trying to avoid any conflicts with the other variants by only making this change for the float icon variant.

rgclayton (approved):

libs/blocks/action-item/action-item.css
Sartxi: @rgclayton okay I checked it out and reverted my `inline-flex`. Thanks for looking out!

auniverseaway (approved):

overmyheadandbody (approved):

rgclayton:
Verified with the kitchen sink documentation that Ryan P shared.

https://sarchibeque-mwpw-143648-action-item--milo--adobecom.hlx.page/docs/library/kitchen-sink/action-items

PR #1966, opened 2024-03-06
[MILO] Allow svg or media files to be located in a fragments folder

vgoodric:
As an author or test developer, I want to be able to add an svg to a folder path containing "fragments" so that they can be stored next to their related fragments.

Resolves: [MWPW-143781](https://jira.corp.adobe.com/browse/MWPW-143781)

**Test URLs**
 - Before: https://stage--milo--adobecom.hlx.page/drafts/vgoodrich/fragments/2024/q2/svginfragments/?martech=off
 - After: https://svginfragments--milo--adobecom.hlx.page/drafts/vgoodrich/fragments/2024/q2/svginfragments/?martech=off


mokimo (commented):
I'm not certain about this change but don't have full context.
Wasn't the going opinion that svgs should always live in an assets folder? I would not start mixing things and introducing extra IFs for something that is author education

vgoodric:
> I'm not certain about this change but don't have full context. Wasn't the going opinion that svgs should always live in an assets folder? I would not start mixing things and introducing extra IFs for something that is author education

@mokimo My bad, I should have given more context.  So we try to keep all test-related assets together.  Makes it easier to know if something is no longer needed.  And on the homepage repo, our test folder is in the fragments folder.  A dev asked for assistance when his page did not work.  I realized it was because the code did not allow for an file like an SVG to be in a fragments folder.  I asked about it and at first @auniverseaway recommended we just not keep them in that folder. Part of his reasoning was that he didn't want to have to carve out a bunch of file extensions (my original code just carved out SVGs).

He then Slacked me and said that he had thought about it some more and realized that a fragment would never have any extension.  So if we just checked that there was no . (dot) we could be certain it was a fragment.  So he recommended I adjust and submit a PR.

Sartxi (approved):

auniverseaway:
@mokimo I think there's two paths to take:

1. Just train the authors to put assets in the asset folder.
2. Things should "just work" without extra training.

@vgoodric had a pretty good use case to group test content together that makes a lot of sense. I can also see our non-technical folks (blog, etc.) not really caring about our conventions.

Given the small amount of code, I would ship, but that's just me. From a utils payload perspective, I think this is value add and if you wanted to, you could pick off some of the functions that shouldn't be in utils... the iObserver stuff, etc.

3ch023 (approved):

libs/utils/utils.js
```diff
@@ -562,7 +562,9 @@ export function decorateAutoBlock(a) {
       return false;
     }
 
-    if (key === 'fragment') {
+    const nameSplit = a.href.split('/').pop().split('.');
+    const mp4Match = 'media_.*.mp4';
+    if (key === 'fragment' && (nameSplit.length <= 1 || a.textContent.match(mp4Match))) {
```
overmyheadandbody, line 567: Regular expressions can be costly, so I'd avoid calling `match` two times, once here and once on L584. Could we cache the result once and use it later as well?

vgoodric: @overmyheadandbody That's a good point.  Tweaked to save results and only run the match once.

overmyheadandbody (approved):

libs/utils/utils.js
overmyheadandbody: Thanks for looking into this!

PR #1969, opened 2024-03-06
MWPW-143202 firefox gnav not appearing bug

vhargrave:
Start using a normal fetch request to get the geo2 location data of the user since this request is timing out in latest versions of firefox on sonoma. 

To test this, if you're in the US you'll need to join a VPN in Europe for example and not use the akamaiLocale=whatever param that you usually would.
Additionally, this problem can only be tested on firefox > 118 and on Mac Sonoma. Browserstack can help you test in that constellation. 

Resolves: [MWPW-143202](https://jira.corp.adobe.com/browse/MWPW-143202)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/vhargrave/georouting?akamaiLocale=CA
- After: https://vhargrave-firefix-georouting-bug--milo--adobecom.hlx.page/drafts/vhargrave/georouting?milolibs=&akamaiLocale=CA

- Before: https://main--homepage--adobecom.hlx.live/homepage/index-loggedout
- After: https://main--homepage--adobecom.hlx.live/homepage/index-loggedout?milolibs=vhargrave-firefix-georouting-bug

- Before: https://main--bacom--adobecom.hlx.live
- After: https://main--bacom--adobecom.hlx.live/?milolibs=vhargrave-firefix-georouting-bug


mokimo (approved):

libs/features/georoutingv2/georoutingv2.js
```diff
@@ -35,35 +35,25 @@ export const getCookie = (name) => document.cookie
...
   const akamaiLocale = urlParams.get('akamaiLocale') || sessionStorage.getItem('akamai');
   if (akamaiLocale !== null) {
     resolve(akamaiLocale.toLowerCase());
   } else {
     /* c8 ignore next 5 */
-    geo2jsonp((data) => {
-      const code = data.country.toLowerCase();
-      sessionStorage.setItem('akamai', code);
-      resolve(code);
+    fetch('https://geo2.adobe.com/json/').then((resp) => {
```
chrischrischris, line 45: Why not keep the `geo2jsonp` function instead of inlining it?

I thought that the randomized callback name is also for cache busting?

Roycethan:
@vhargrave Observations for https://main--homepage--adobecom.hlx.live/homepage/index-loggedout?milolibs=vhargrave-firefix-georouting-bug#. plz review thanks
#1) 404 for utils.js:
image
#2) Signing in with user takes me to an infinite loop - may be its related some other issue but mentioning here as well

afmicka:
I have the same findings as @Roycethan. Given that i get into infinite loop on main hlx page as well (https://main--homepage--adobecom.hlx.live/homepage/index-loggedout) in all browsers, does not seem to be related to this PR. @vhargrave if there is a way to check the cause, please do. Otherwise, we could verify this PR. 

btw, i could not reproduce the issue on real device (I have Macbook Air with Sonoma and latest Firefox, changed the VPN, used regular and incognito window, etc.). Only on browserstack.

vhargrave:
perfect, thank you for testing! @afmicka 
I'm not sure why we're having that issue with the utils, but it definitely doesn't have to do with this PR as you pointed out, since we're having the issue on main as well. 
did you test with ?akamaiLocale in your url btw? If you do that then you can't reproduce the issue. It only happens with a completely clear cache and when you test directly with https://main--homepage--adobecom.hlx.live/homepage/index-loggedout?milolibs=vhargrave-firefix-georouting-bug , but in country so that the georouting does appear. 

Whatever the case, it sounds like we're good to give this a verified label. Please do so at your discretion. :) @afmicka & @Roycethan

overmyheadandbody (approved):

Blainegunn:
PSI test fails, seems the test page isn't set up to be tested

PR #1970, opened 2024-03-07
daa-ll not present when using footer-promo-tag in metadata.xlsx

vgoodric:
DAA-LL value not present when using the footer-promo-tag metadata property to insert CTA's on BACOM pages.

Steps to replicate:
1. go to https://business.adobe.com/products/experience-manager/sites/testing-optimization.html
2. scroll to red blade at bottom of the page “Let’s talk about what Adobe Experience Manager Sites can do for your business.” 
3. Inspect “Get started” button.

Note, no daa-ll is showing. 
 
However, if you inspect the fragment where the button is authored, https://main--bacom--adobecom.hlx.live/fragments/footer-promos/products/aem-sites, you will see the daa-ll.

Because the footer promo is triggered by the gnav, it happens asynchronously after the rest of the analytics are added.

Resolves: [MWPW-143827](https://jira.corp.adobe.com/browse/MWPW-143827)

**Test URLs:**
- Before: https://main--bacom--adobecom.hlx.page/products/experience-manager/sites/testing-optimization?martech=off
- After: https://main--bacom--adobecom.hlx.page/products/experience-manager/sites/testing-optimization?milolibs=footerpromotaganalytics&martech=off
- URL for testing:

- https://footerpromotaganalytics--milo--adobecom.hlx.page/


libs/features/footer-promo.js
```diff
@@ -42,4 +44,6 @@ export default async function initFooterPromo(footerPromoTag, footerPromoType) {
   document.querySelector('main > div:last-of-type').insertAdjacentElement('afterend', section);
   await loadFragment(a);
   section.classList.add('section');
+  const sections = document.querySelectorAll('main > div');
```
mokimo, line 47: Could you reuse the selector from L44 or otherwise just use 
```suggestion
  const lastSection = document.querySelectorAll('main > div:last-of-type');
```

mokimo (approved):
Should we add any unit test for this, since it already caused trouble?

libs/features/footer-promo.js
vgoodric: @mokimo I am only selecting to get the number of sections. I use the length of the results in the function call on the next line.  I don't think the suggested update would still get me this info.

libs/features/footer-promo.js
```diff
@@ -25,7 +26,8 @@ async function getPromoFromTaxonomy(contentRoot) {
 }
 
 export default async function initFooterPromo(footerPromoTag, footerPromoType) {
-  const { locale: { contentRoot } } = getConfig();
+  const config = getConfig();
+  const { locale: { contentRoot } } = config;
```
chrischrischris, line 30: super nit:
```suggestion
  const { locale: { contentRoot } } = getConfig();
```

libs/features/footer-promo.js
```diff
@@ -42,4 +44,6 @@ export default async function initFooterPromo(footerPromoTag, footerPromoType) {
   document.querySelector('main > div:last-of-type').insertAdjacentElement('afterend', section);
   await loadFragment(a);
   section.classList.add('section');
+  const sections = document.querySelectorAll('main > div');
+  decorateSectionAnalytics(section, sections.length - 1, config);
```
chrischrischris, line 48: Looking at `decorateSectionAnalytics` the 2nd param (`idx`) seems to indicate the section index, not the total amount of sections?

vgoodric: where it's used normally, it's looping through all sections and passing in the array index from a querySelector all.  I don't want to rerun the other sections, but I am trying to determine what it's index would have been if it existed in time to be part of the initial loop.  And that number is 1 less than the length.

PR #1976, opened 2024-03-08
fix(MWPW-142461):fixed Inconsistent font seen i-phone safari

sharath-kannan:
* Adjusted text size to 100%.

Resolves: [MWPW-142461](https://jira.corp.adobe.com/browse/MWPW-142461)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/sharathkannan/buying-guide?martech=off
- After: https://mwpw-142461--milo--sharath-kannan.hlx.page/drafts/sharathkannan/buying-guide?martech=off


mokimo (commented):
Not blocking the PR in any way, but still want to pose this question:
Do we really want to overwrite the default browser behavior? This might impact  accessibility and prevent users from adjusting text sizes. On top of that, this is an [experimental property](https://developer.mozilla.org/en-US/docs/Web/CSS/text-size-adjust?retiredLocale=de#browser_compatibility)

sharath-kannan:
> I am not blocking the PR in any way, but I still want to pose this question: Do we want to overwrite the default browser behavior? This might impact accessibility and prevent users from adjusting text sizes. On top of that, this is an [experimental property](https://developer.mozilla.org/en-US/docs/Web/CSS/text-size-adjust?retiredLocale=de#browser_compatibility)

This property is ignored in most browsers other than i-phone's safari. hence this might be a way to fix this issue and wouldn't cause an issue elsewhere.

overmyheadandbody (commented):
One thing that caught my eye reading the docs is
> as web pages designed with small screens in mind do not need it

So the question is - why does this particular element need it and none of the others do? Additionally, it would be helpful to attach some screenshots so the change is understood and adapt the 'After' link, since it's invalid (two `?`)

fullcolorcoder (commented):
Based on the latest comment in https://jira.corp.adobe.com/browse/MWPW-142461
is this still needed?

sharath-kannan:
> Based on the latest comment in https://jira.corp.adobe.com/browse/MWPW-142461 is this still needed?

yes,this pr is still required as QA has mentioned that the issue is fixed with this PR.

sharath-kannan:
> One thing that caught my eye reading the docs is
> 
> > as web pages designed with small screens in mind do not need it
> 
> So the question is - why does this particular element need it and none of the others do? Additionally, it would be helpful to attach some screenshots so the change is understood and adapt the 'After' link, since it's invalid (two `?`)



> One thing that caught my eye reading the docs is
> 
> > as web pages designed with small screens in mind do not need it
> 
> So the question is - why does this particular element need it and none of the others do? Additionally, it would be helpful to attach some screenshots so the change is understood and adapt the 'After' link, since it's invalid (two `?`)

Updated the 'After' link.
I have noticed this issue in other pages as well, the font size is inconsistent if the accordion label exceeds a single line.
Here are few sample documents that have the same issue.
[faq.docx](https://adobe.sharepoint.com/:w:/r/sites/adobecom/CC/www/drafts/sharathkannan/MWPW-137714/faq.docx?d=w60175440dc4f45cd95c4a42480091bd8&csf=1&web=1&e=dqRloF)
[faq_indesign](https://adobe.sharepoint.com/:w:/r/sites/adobecom/CC/www/drafts/sharathkannan/MWPW-137714/faq_indesign.docx?d=w4af5107ead874805b69cea236e432493&csf=1&web=1&e=stAgqx)

fullcolorcoder (approved):

Anupamasonnalagi:
Verified the issue. It's fine now.
https://main--cc--adobecom.hlx.live/products/indesignserver/buying-guide?milolibs=mwpw-142461--milo--sharath-kannan
IMG_3044
https://adobe.com/creativecloud.html?milolibs=mwpw-142461--milo--sharath-kannan
IMG_3046
https://www.adobe.com/products/illustrator.html?milolibs=mwpw-142461--milo--sharath-kannan
IMG_3045

mokimo (approved):
> I have noticed this issue in other pages as well, the font size is inconsistent if the accordion label exceeds a single line.

That sounds like the browser wants to do this for accessibility reasons? But if we don't want this - since it just affects webkit, should be fine and lead to more cross browser consistency.

mokimo:
Since this is verified, apply the "Ready for Stage" label if you feel like this is ready to go to stage and then prod

PR #1982, opened 2024-03-11
[MWPW-141050] Fetch profile for PEP

overmyheadandbody:
## Description
This adapts the current PEP logic to use the new method of fetching profile data, rather than relying on IMS and cc-collab. Two new utility methods have been created, `setUserProfile` and `getUserProfile` so that the profile data can be used in other parts of the logic, if needed. `getUserProfile` will return a `Promise` that resolves as soon as data becomes available. I've decided to simply resolve with an empty object if errors are encountered or 5 seconds have passed; this is due to the number of edge cases where uncaught errors could be thrown to the console.

PEP-specific segments [1] have also been added and the logic slightly adapted to decide whether the prompt needs to be shown. Because we don't have a clear way of testing this on Milo pages, I've added some test functionality so that a `extraPepEnts` query parameter can be used to define segment names for which the user qualifies for. This, of course, will need to be removed before the final merge.

This PR is also a first step towards loading the prompt based on Universal Nav status.

[1] The segment names are `lightroom-web-usage`, `photoshop-web-usage`, `firefly-web-usage` and `acrobat-web-usage` and are defined for both production and stage environments.

## Related Issue
Resolves: [MWPW-141050](https://jira.corp.adobe.com/browse/MWPW-141050) & [MWPW-141050](https://jira.corp.adobe.com/browse/MWPW-141050)

## Testing instructions
* use VPN, load test page and log in;
* append a `extraPepEnts` query parameter to the URL, specifying the additional segments the user qualifies for, as defined in the PEP prompt content document (for example, `?extraPepEnts=photoshop,firefly-web-usage`);
* refresh the page;
* PEP prompt should load and show the user's profile;
* no network requests should be made for cc-collab.

## Test URLs
**Milo:**
- Before: https://project-pep--milo--adobecom.hlx.page/drafts/ramuntea/universal-nav?imsClientId=fedsmilo&martech=off&georouting=off
- After: https://project-pep--milo--overmyheadandbody.hlx.page/drafts/ramuntea/universal-nav?imsClientId=fedsmilo&martech=off&georouting=off

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -259,8 +259,6 @@ const convertToPascalCase = (str) => str
   .map((s) => s.charAt(0).toUpperCase() + s.slice(1))
   .join(' ');
 
-// TODO: debate whether this should be here or in a separate file,
-// since it's not time-sensitive and we want to potentially bundle it
 // TODO: position the prompt relative to the App Switcher
 const loadAppPrompt = async () => {
   const state = getMetadata('app-prompt')?.toLowerCase();
@@ -273,20 +271,6 @@ const loadAppPrompt = async () => {
     || !entName?.length
     || !promptPath?.length) return;
 
-  const id = promptPath.split('/').pop();
-  const isDismissed = JSON.parse(document.cookie
-    .split(';')
-    .find((item) => item.trim().startsWith('dismissedAppPrompts='))
-    ?.split('=')[1] || '[]')
-    .includes(id);
-
-  if (isDismissed) return;
-
-  // const entitlements = await getConfig().entitlements();
-  const entitlements = ['photoshop']; // TODO: replace this line with the one above
-
-  if (!entitlements?.length || !entitlements.includes(entName)) return;
-
   const { base } = getConfig();
   const [
     webappPrompt,
@@ -295,7 +279,7 @@ const loadAppPrompt = async () => {
     loadStyle(`${base}/features/webapp-prompt/webapp-prompt.css`),
   ]);
 
-  webappPrompt.default({ promptPath, id });
+  webappPrompt.default({ promptPath, entName });
 };
 
 class Gnav {
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -8,29 +8,30 @@ import {
   loadStyle,
 } from '../../utils/utils.js';
 import {
-  toFragment,
-  getFedsPlaceholderConfig,
-  getAnalyticsValue,
-  decorateCta,
-  getExperienceName,
-  loadDecorateMenu,
-  loadBlock,
-  loadStyles,
-  trigger,
-  setActiveDropdown,
   closeAllDropdowns,
-  loadBaseStyles,
-  yieldToMain,
+  decorateCta,
+  fetchAndProcessPlainHtml,
+  getActiveLink,
+  getAnalyticsValue,
+  getExperienceName,
+  getFedsPlaceholderConfig,
+  hasActiveLink,
   isDesktop,
   isTangentToViewport,
-  setCurtainState,
-  hasActiveLink,
-  setActiveLink,
-  getActiveLink,
-  selectors,
-  logErrorFor,
   lanaLog,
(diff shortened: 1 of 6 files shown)


libs/blocks/global-navigation/global-navigation.js
```diff
@@ -251,6 +259,45 @@ const convertToPascalCase = (str) => str
   .map((s) => s.charAt(0).toUpperCase() + s.slice(1))
   .join(' ');
 
+// TODO: debate whether this should be here or in a separate file,
+// since it's not time-sensitive and we want to potentially bundle it
+// TODO: position the prompt relative to the App Switcher
+const loadAppPrompt = async () => {
```
narcis-radu, line 265: Should this be available on all pages? I was thinking it will be something that gets loaded via metadata and then the logic is executed. I am all for moving this to a separate file since it will only apply to a few pages.

mokimo: Yes, agree with Narcis. I also wonder whether this should be in the `webapp-prompt.js` directly and you pass a few more (potentially optional) properties?
It seems like base-setup for this to work, so any consumer would also need to retrieve this data, correct?

overmyheadandbody: The thing I was trying to avoid is loading a module on a lot of pages, when the feature would actually be used on a very small subset of pages. The checks here are 35 lines long and ensure that we're loading `webapp-prompt.js` only if actually required. Moving them to the module will import some 200 lines, although they might not be used.

As a compromise, I'm thinking of keeping the metadata/signed-in/device checks here and moving the dismissal state and entitlement checks to the `webapp-prompt.js` file. Do you think that's a good compromise?

SilviuLCF (commented):
Profile is not always loaded on PEP

Test details can be found on https://jira.corp.adobe.com/browse/MWPW-141050

overmyheadandbody:
What @SilviuLCF is bringing up is a limitation of the Universal Navigation ecosystem. We depend on the Profile Menu to load, part of the Universal Navigation. First loads can be quite slow, so the profile data might not be available in under 5 seconds. My current decision was to show the prompt without the profile data to not wait for too long; I'm preferring behavior over UI. Of course, it's not my decision to make, so I've brought this up with our PM for input in https://jira.corp.adobe.com/browse/MWPW-141050#comment-40668605

libs/blocks/global-navigation/global-navigation.js
mokimo: Definitely a tradeoff, yeah lets go with your suggestion for now @overmyheadandbody

overmyheadandbody: I've split the logic between `global-navigation.js` and `webapp-prompt.js` so that the footprint in the former is as small as possible, while still avoiding to load the PEP logic if not really needed.

overmyheadandbody:
Hi @SilviuLCF! In the comments section from https://jira.corp.adobe.com/browse/MWPW-141050 you'll see that the stakeholders are aware of this limitation and are fine with the current implementation, as we can't have a meaningful impact over 3rd party library load times right now. We can resume testing when there's bandwidth.

mokimo (approved):
My feeling would be the statement weather to load the pep module is a bit too long - so your suggestion of a middle ground would be great 👍 
Otherwise the PR looks good!

SilviuLCF (approved):

narcis-radu (approved):

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -63,6 +64,13 @@ const CONFIG = {
             config: {
               enableLocalSection: true,
               miniAppContext: {
+                onMessage: (name, payload) => {
+                  if (name === 'System' && payload.subType === 'AppInitiated') {
+                    window.adobeProfile?.getUserProfile()
```
honstar, line 69: Just curious, what is this `adobeProfile` object on `window` and where is that set? I'm curious what exact profile is fetched from where. The PR description says that IMS and cc-collab are not being used anymore.

honstar: Thanks @overmyheadandbody for the additional details 👍

PR #1992, opened 2024-03-12
[MILO][MEP] MEP parameter spoof breaks if chosen experience has a comma

vgoodric:
Detailed Description: the MEP parameter supports multiple manifests by using a comma to separate the manifest/experience pairs. This delimiter was chosen before the MEP manifests supported commas in the column names. These 2 features collide when trying to spoof a column containing a comma.

URL: https://main--cc--adobecom.hlx.page/products/premiere?mep=%2Fproducts%2Fpremiere.json--cc-all-apps-any%2C+premiere-any%2C%2Fcc-shared%2Ffragments%2Fpromos%2F2024%2Famericas%2Fcci-all-apps-q1%2Fcci-all-apps-q1.json--all
Steps to Reproduce:
1. Go to the link above
2. Use the MEP button to preview an experience with a comma
3. Check the MEP button panel again.

Expected Results: The experience you previously selected will be chosen
Actual Results: The experience you previously selected will not be chosen

Solution, switch from using a comma to separate manifest/experience pairs and use 3 dashes instead.

Resolves: [MWPW-143067](https://jira.corp.adobe.com/browse/MWPW-143067)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/vgoodrich/fragments/2024/q2/meppreviewdelimiter/?martech=off
- After: https://meppreviewdelimiter--milo--adobecom.hlx.page/drafts/vgoodrich/fragments/2024/q2/meppreviewdelimiter/?martech=off

### DIFF
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -417,7 +417,7 @@ async function getPersonalizationVariant(manifestPath, variantNames = [], varian
   if (config.mep?.override) {
     let manifest;
     /* c8 ignore start */
-    config.mep?.override?.split(',').some((item) => {
+    config.mep?.override?.split('---').some((item) => {
       const pair = item.trim().split('--');
       if (pair[0] === manifestPath && pair.length > 1) {
         [, manifest] = pair;
@@ -606,6 +606,7 @@ function compareExecutionOrder(a, b) {
 export function cleanAndSortManifestList(manifests) {
   const manifestObj = {};
   manifests.forEach((manifest) => {
+    if (!manifest) return;
     manifest.manifestPath = normalizePath(manifest.manifestUrl || manifest.manifest);
     if (manifest.manifestPath in manifestObj) {
       let fullManifest = manifestObj[manifest.manifestPath];
--- a/libs/features/personalization/preview.js
+++ b/libs/features/personalization/preview.js
@@ -203,7 +203,7 @@ function createPreviewPill(manifests) {
   const personalizationOn = getMetadata('personalization');
   const personalizationOnText = personalizationOn && personalizationOn !== '' ? 'on' : 'off';
   const simulateHref = new URL(window.location.href);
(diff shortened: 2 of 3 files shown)


mokimo (approved):
I assume those parameters are authored? Is there any backwards compatibility that needs to be considered?

mokimo:
The PSI check is failing, would be great to get that green. There's some [guidance here](https://github.com/orgs/adobecom/discussions/997) (aka just restart tests or use published `hlx.live` pages)

chrischrischris (approved):

libs/features/personalization/personalization.js
```diff
@@ -606,6 +606,7 @@ function compareExecutionOrder(a, b) {
 export function cleanAndSortManifestList(manifests) {
   const manifestObj = {};
   manifests.forEach((manifest) => {
+    if (!manifest) return;
```
overmyheadandbody, line 609: What's the use-case here? Can `manifests` contain falsy values?

overmyheadandbody (approved):

vgoodric:
> I assume those parameters are authored? Is there any backwards compatibility that needs to be considered?

@mokimo the preview link feature is considered internal only.  I can't think of a way to make this backwards compatible if a person has an "old link". But the preview button makes it very easy to generate a new link, so this should not be much of a hardship.

libs/features/personalization/personalization.js
vgoodric: yes, my sample page was producing an array with a length of 3 (2 manifests and an "empty" manifest).  Hadn't seen it before, but decided to add this line to protect from other pages having the same issue.

vgoodric:
> The PSI check is failing, would be great to get that green. There's some [guidance here](https://github.com/orgs/adobecom/discussions/997) (aka just restart tests or use published `hlx.live` pages)

Because almost all of my PR's involve MEP, I can't ever seem to pass the psi-check.

PR #1997, opened 2024-03-13
Add support for bold as header if no header

vgoodric:
* If an author is asked to use bold instead of a header, bolded text can be used in the analytic. But if a header is present, bolded text is not used

**Real page example:**
Before: https://main--cc--adobecom.hlx.page/products/firefly/discover/ai-art-use-cases
After: https://main--cc--adobecom.hlx.page/products/firefly/discover/ai-art-use-cases?milolibs=analyticsusebold

Resolves: [MWPW-143995](https://jira.corp.adobe.com/browse/MWPW-143995)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/vgoodrich/fragments/2024/q2/analyticsusebold/?martech=off
- After: https://analyticsusebold--milo--adobecom.hlx.page/drafts/vgoodrich/fragments/2024/q2/analyticsusebold/?martech=off


libs/martech/attributes.js
```diff
@@ -19,7 +19,14 @@ export function decorateDefaultLinkAnalytics(block, config) {
     && block.nodeName === 'DIV') {
     let header = '';
     let linkCount = 1;
-    block.querySelectorAll('h1, h2, h3, h4, h5, h6, a:not(.video.link-block), button, .tracking-header').forEach((item) => {
+
+    let headerSelector = 'h1, h2, h3, h4, h5, h6, .tracking-header';
+    const headers = block.querySelectorAll(headerSelector);
+    if (!headers.length) headerSelector = `${headerSelector}, b, strong`;
+    block.querySelectorAll(`${headerSelector}, a:not(.video.link-block), button`).forEach((item) => {
+      if (item.nodeName === 'STRONG' || item.nodeName === 'B') {
```
chrischrischris, line 27: nit: it's a bit hard to see that if this condition is true then the else condition on line 53 is  what's next.  Maybe instead something like:
```
  if (item.nodeName === 'STRONG' || item.nodeName === 'B') {
    item.classList.add('tracking-header');
  }

  if (item.nodeName !== 'A' && item.nodeName !== 'BUTTON') {
    header = processTrackingLabels(item.textContent, config, 20);
    return;
  }

  if (item.classList.contains('tracking-header')) {
  ...
```

chrischrischris (approved):
Approved with a nit

libs/martech/attributes.js
vgoodric: @chrischrischris I think the issue is that the strong or bold check really only needs to happen if it's not a link or button. but if it's not, I want to add also add that class for the bold/strong tag.  So I moved the if.

rgclayton (approved):

PR #1999, opened 2024-03-13
MWPW-135821 add ability to replace merch-card in mep

npeltier:
- mep decorates whatever selected element with override path specified in manifest,
- merch-cards fetches specified override paths and replace them if matching name properties

Resolves: [MWPW-135821](https://jira.corp.adobe.com/browse/MWPW-135821)

with following manifest added to the page

| action | selector | page filter (optional) | target-challenger name |
|---|---|---|---|
| replaceMerchCard | .merch-cards |   | /drafts/npeltier/fragments/merch/products/catalog/photoshop/promo |

before & after personalisation (see photoshop card)

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/drafts/npeltier/promocard-test?milolibs=mwpw-135821--milo--npeltier
- After: https://main--cc--adobecom.hlx.page/drafts/npeltier/promocard-test?milolibs=mwpw-135821--milo--npeltier&mep=%2Fdrafts%2Fnpeltier%2Fpromo-test.json--target-challenger+name

### DIFF
--- a/libs/blocks/merch-cards/merch-cards.js
+++ b/libs/blocks/merch-cards/merch-cards.js
@@ -5,6 +5,7 @@ import '../merch/merch.js';
 import '../merch-card/merch-card.js';
 import { createTag, decorateLinks, getConfig, loadBlock, loadStyle } from '../../utils/utils.js';
 import { replaceText } from '../../features/placeholders.js';
+import { OVERRIDE_PATHS } from '../../features/personalization/personalization.js';
 
 const DIGITS_ONLY = /^\d+$/;
 
@@ -65,25 +66,24 @@ export function parsePreferences(elements) {
   });
 }
 
-async function initMerchCards(config, type, filtered, el, preferences) {
-  let cardsData;
-  let err;
+const localizedPath = (path, config) => `${config?.locale?.prefix ?? ''}${path}`;
 
+const fetchOverrideCard = (path, config) => new Promise((resolve, reject) => {
   try {
-    const res = await fetch(`${config?.locale?.prefix ?? ''}${config.queryIndexCardPath}.json?sheet=${type}`);
-    if (res.ok) {
-      cardsData = await res.json();
-    } else {
-      err = res.statusText || res.status;
-    }
+    fetch(`${localizedPath(path, config)}.plain.html`).then((res) => {
+      if (res.ok) {
+        res.text().then((cardContent) => {
+          resolve({ path, cardContent: /^<div>(.*)<\/div>$/.exec(cardContent.replaceAll('\n', ''))[1] });
+        });
+      }
+    });
   } catch (error) {
-    err = error.message;
+    reject(error);
   }
-  if (!cardsData) {
-    fail(el, err);
-  }
+});
 
-  const cards = `<div>${cardsData.data.map(({ cardContent }) => cardContent).join('\n')}</div>`;
+async function generateCards(html, config) {
+  const cards = `<div>${html}</div>`;
   const fragment = document.createRange().createContextualFragment(cards);
   const cardsRoot = fragment.firstElementChild;
   await makePause();
@@ -93,6 +93,7 @@ async function initMerchCards(config, type, filtered, el, preferences) {
   const autoBlocks = await decorateLinks(cardsRoot).map(loadBlock);
   await Promise.all(autoBlocks);
   await makePause();
+
   const blocks = [...cardsRoot.querySelectorAll(':scope > div')].map(loadBlock);
 
   // process merch card blocks in batches of 3.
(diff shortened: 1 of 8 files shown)


libs/blocks/merch-cards/merch-cards.js
```diff
@@ -65,25 +66,24 @@ export function parsePreferences(elements) {
...
-    const res = await fetch(`${config?.locale?.prefix ?? ''}${config.queryIndexCardPath}.json?sheet=${type}`);
-    if (res.ok) {
-      cardsData = await res.json();
-    } else {
-      err = res.statusText || res.status;
-    }
+    fetch(`${localizedPath(path, config)}.plain.html`).then((res) => {
+      if (res.ok) {
+        res.text().then((cardContent) => {
+          resolve({ path, cardContent: /^<div>(.*)<\/div>$/.exec(cardContent.replaceAll('\n', ''))[1] });
```
yesil, line 76: we should make sure that `/^<div>(.*)<\/div>$/.exec(cardContent.replaceAll('\n', ''))[1]` never breaks the base flow.

libs/blocks/merch-cards/merch-cards.js
```diff
@@ -65,25 +66,24 @@ export function parsePreferences(elements) {
...
+        });
+      }
+    });
   } catch (error) {
-    err = error.message;
-  }
-  if (!cardsData) {
-    fail(el, err);
+    reject(error);
```
yesil, line 81: if the promo fails, we should still give a change to the default page to render and not reject here.

libs/features/personalization/personalization.js
```diff
@@ -142,6 +144,13 @@ const COMMANDS = {
     el.insertAdjacentElement('beforebegin', createFrag(el, target, manifestId));
     el.classList.add(CLASS_EL_DELETE, CLASS_EL_REPLACE);
   },
+  replacemerchcard: (el, cardPath) => {
```
yesil, line 147: I would leave MEP neutral to blocks and just expose the custom verbs/config so that blocks can read and execute.

libs/blocks/merch-cards/merch-cards.js
npeltier: we _are_ in a try catch, see l80

libs/features/personalization/personalization.js
npeltier: you mean an action agnostic of what block it's amending? do you have an action name in head (same question for @chrischrischris )

yesil: action name should `replaceMerchCard` but without any implementation.
we can have an array of pull/consume only actions where MEP would simply expose it but do nothing.
:)

libs/blocks/merch-cards/merch-cards.js
yesil: sorry I confused `reject` with `fail` helper.

yesil (approved):
I verified manually and works fine.
Added my comments w.r.t MEP integration, but it is my opinion and is trivial.

libs/features/personalization/personalization.js
npeltier: well, yes, different debate then, (and different JIRA / PR) but i would rather in that case use some "registry" pattern where in that case merch-cards would register its action in mep (cc @chrischrischris)

libs/blocks/merch-cards/merch-cards.js
```diff
@@ -5,6 +5,7 @@ import '../merch/merch.js';
 import '../merch-card/merch-card.js';
 import { createTag, decorateLinks, getConfig, loadBlock, loadStyle } from '../../utils/utils.js';
 import { replaceText } from '../../features/placeholders.js';
+import { OVERRIDE_PATHS } from '../../features/personalization/personalization.js';
```
chrischrischris, line 8: Not ideal for performance: personalization.js is always loaded when merch cards are, even if nothing else is used.

libs/features/personalization/personalization.js
```diff
@@ -142,6 +144,13 @@ const COMMANDS = {
...
+    const paths = el.dataset[OVERRIDE_PATHS]?.split(',') || [];
+    if (!paths.includes(cardPath)) {
```
chrischrischris, line 149: `paths` could be undefined
```suggestion
    if (paths && !paths.includes(cardPath)) {
```

libs/features/personalization/personalization.js
```diff
@@ -142,6 +144,13 @@ const COMMANDS = {
...
+      paths.push(cardPath);
+    }
+    el.dataset[OVERRIDE_PATHS] = paths.join(',');
```
chrischrischris, line 152: Do you always want the data attribute defined even if empty?

chrischrischris (changes requested):
@npeltier After discussing with @vivgoodrich we realized there's a cleaner implementation that doesn't require the use of a new `replacemerchcard` command.

There is an open PR that has to be merged first, and then what authors will be able to do is use a `replace` command, and if targetting merch-blocks then it will add the `overrides` data attribute.  I'll sync with you on this tmrw.

npeltier:
ok @chrischrischris then i guess i'll wait for that discussion before adressing your other comments :)

npeltier:
@chrischrischris @vivgoodrich can you point me to related PR? may be i could start build this one upon it

PR #2013, opened 2024-03-14
MWPW-144473 Reduce search time for contextual search

meganthecoder:
* Reduces the time it takes for results to appear for specific/uncommon searches in the contextual search feature.
* Note: This fix will slightly change the user experience. Before, all of the results displayed at once, and after, each result will show up progressively as it's processed.

Resolves: [MWPW-144473](https://jira.corp.adobe.com/browse/MWPW-144473)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/methomas/contextual-search
- After: https://methomas-search-timing--milo--adobecom.hlx.page/drafts/methomas/contextual-search?martech=off

Note: This is most noticeable on the brand blog rather than the bacom blog. It happens when searching for a query that doesn't have 12 results within the most recent 500 posts.

Brand Blog:
Try searching "Olivia" or "Preventing Edits to Bates Numbers applied in Acrobat"
- Before: https://main--blog--adobecom.hlx.live/en/uk
- After: https://main--blog--adobecom.hlx.live/en/uk?milolibs=methomas-search-timing

Bacom Blog:
Trying searching for "launches"
- Before: https://main--bacom-blog--adobecom.hlx.live/blog/
- After: https://main--bacom-blog--adobecom.hlx.live/blog/?milolibs=methomas-search-timing


Brandon32:
@meganthecoder Awesome improvement. Can we signal when a search is still in progress or is finished? 
With the "Olivia" search I might expect more than 2 results and would just be waiting for nothing to appear.

libs/blocks/gnav/gnav-contextual-search.js
```diff
@@ -53,25 +52,33 @@ function highlightTextElements(terms, elements) {
   });
 }
 
-async function fetchResults(signal, terms, config) {
-  let hits = [];
+async function getResults(signal, terms, config, resultsEl, dataSegment, prevHits) {
```
JasonHowellSlavin, line 55: I think this many params is a sign that this function can/should be broken up. Because of the number of checks based on input, the recursive pattern might not be the best. We could keep everything needed for triggering the abort signal enclosed within a function and adding a `while` loop I think. 

That being said, I am not sure fixing a bug is the right place to do a code overhaul like that. Do you think we ought to make a ticket for tech debt? Or do you like this pattern as is?

JasonHowellSlavin (commented):
Looks good. Are there tests to update?

meganthecoder:
> @meganthecoder Awesome improvement. Can we signal when a search is still in progress or is finished? With the "Olivia" search I might expect more than 2 results and would just be waiting for nothing to appear. The article feed has a spinner and that could be useful here.

@Brandon32 Good idea. I'll ask Malik about this in the ticket.

libs/blocks/gnav/gnav-contextual-search.js
meganthecoder: It seems like you're describing how it was originally written, but that was causing the duplication bug https://github.com/adobecom/milo/blob/063d5b5dd99c6691c36ac560b7c936017ccb44d6/libs/blocks/gnav/gnav-contextual-search.js

I went through several patterns and this made the most sense to me. I'm open to suggestions though. Do you have a code sample?

meganthecoder:
> Looks good. Are there tests to update?

@JasonHowellSlavin I don't believe we ever test the time it takes to perform an action. Is that the kind of test you're looking for?

libs/blocks/gnav/gnav-contextual-search.js
```diff
@@ -53,25 +52,33 @@ function highlightTextElements(terms, elements) {
...
     }
   }
 
-  const getMoreArticles = !articles.length || (hits.length !== LIMIT && !complete);
+  const noArticles = !articles.length;
+  const getMoreArticles = noArticles || (hits.length !== LIMIT && !complete);
 
   if (!signal.aborted && getMoreArticles) {
-    const index = await fetchBlogArticleIndex(config);
+    const fetchLimit = noArticles ? 500 : 10000;
```
Brandon32, line 76: Nit, 500 and 10,000 should be constants

Brandon32 (approved):

libs/blocks/gnav/gnav-contextual-search.js
JasonHowellSlavin: I think if you went through several and found this to make the most sense that is fine for me

JasonHowellSlavin (approved):

PR #1946, opened 2024-03-01
MWPW-142798 + MWPW-142872: Made the height-fit-content class to be set automatically

mirafedas:
When Trial with Purchase pages (e.g. https://www.adobe.com/mini-plans/photoshop.html?mid=ft&amp;web=1&amp;pint=Photoshop ) are rendered inside modals, we need to set up an automatic height adjustment to eliminate the blank space on the bottom of the page. For that the 'height-fit-content' class should be set on the modal dialog element. As adding this class manually would require a significant authoring effort, we now set it automatically in the code for all cases when the URL of the page inside iFrame includes the '/mini-plans/' part together with 'mid=ft' and 'web=1' params. 
This PR also includes the CSS changes from [this PR](https://github.com/adobecom/milo/pull/1911) as they are required for the modal styles to work properly.

Resolves: 
[MWPW-142798](https://jira.corp.adobe.com/browse/MWPW-142798)
[MWPW-142872](https://jira.corp.adobe.com/browse/MWPW-142872)

**Test URLs:**
- Before: https://www.adobe.com/products/illustrator/campaign/pricing.html#modal-illustrator
- After: https://mwpw-142798-autoset-height-fit-content--milo--mirafedas.hlx.live/drafts/mirafedas/twp#mini-plans-web-cta-creative-cloud-card

For testers, please test the following issues:
https://jira.corp.adobe.com/browse/MWPW-142872
https://jira.corp.adobe.com/browse/MWPW-142798
https://jira.corp.adobe.com/browse/MWPW-142399 (check for regression)

PR #2016, opened 2024-03-15
MWPW-144575: Modal-Iframe for apps-whats-included is broken for Desktop resolutions

mirafedas:
This is a fix for PR #1946 , where the wrong CSS selector caused the .section height in the modals with an iframe to be 0 instead of 100%, which resulted in no content shown on desktop resolution.

Resolves: [MWPW-144575](https://jira.corp.adobe.com/browse/MWPW-144575)

**Test URLs:**
- Before: https://www.adobe.com/products/illustrator/campaign/pricing.html?mboxDisable=1&adobe_authoring_enabled=true#modal-see-whats-included
- After: will add in a moment


yesil (approved):

3ch023 (approved):

yesil:
Added high prio because this will fix an issue with some modals in production.

PR #2017, opened 2024-03-15
MWPW-144588: decorate placeholders of inline fragments

yesil:
Resolves: [MWPW-144588](https://jira.corp.adobe.com/browse/MWPW-144588)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ilyas/segment-pods-page?martech=off
- After: https://mwpw-144588--milo--yesil.hlx.page/drafts/ilyas/segment-pods-page?martech=off


chrischrischris (approved):

3ch023 (approved):

mokimo (approved):

mokimo:
Can we add a unit test for this?

libs/utils/utils.js
```diff
@@ -1042,6 +1042,7 @@ async function processSection(section, config, isDoc) {
     const { default: loadInlineFrags } = await import('../blocks/fragment/fragment.js');
     const fragPromises = inlineFrags.map((link) => loadInlineFrags(link));
     await Promise.all(fragPromises);
+    await decoratePlaceholders(section.el, config);
```
overmyheadandbody, line 1045: Wouldn't we want to replace placeholder in fragments as a rule? Can't this be added as part of `fragments.js` so that any loaded fragment comes with placeholders already replaced?

mokimo: That is a very good catch actually - placeholders should also be replaced when the function is consumed standalone

yesil: I thought that was already the case. Didn't really check, I agree  all fragments should be processed with placeholders. I'll take a look.

yesil: @overmyheadandbody I think the title is misleading and this fix will also apply to fragments.

overmyheadandbody: It's not clear to me how placeholders are replaced in https://github.com/adobecom/milo/blob/stage/libs/blocks/fragment/fragment.js. Other pieces of logic might use `fragment.js` directly, so the logic here in `utils.js` will not apply. In my view, the fragment logic should handle placeholder replacement OOTB

yesil: @overmyheadandbody normal fragments are processed with `loadArea` and placeholders are already processed therein.

yesil:
@mokimo the existing tests already cover this line, I don't think an extra test case is needed.

PR #2028, opened 2024-03-18
MWPW-143314: lana logs for feds-regionpicker.js 404s

robert-bogos:
## Description
This PR adds LANA logs to further investigate region picker 404s.

## Related Issue
Resolves: [MWPW-143627](https://jira.corp.adobe.com/browse/MWPW-143627)

## Test URLs
**Acrobat:**
- After: https://www.stage.adobe.com/acrobat/online/sign-pdf.html?martech=off&milolibs=mwpw-143627-region-picker-404s--milo--robert-bogos

**BACOM:**
- Before: https://main--bacom--adobecom.hlx.page/products/real-time-customer-data-platform/rtcdp?martech=off
- After: https://main--bacom--adobecom.hlx.page/products/real-time-customer-data-platform/rtcdp?martech=off&milolibs=mwpw-143627-region-picker-404s--milo--robert-bogos

**CC:**
- Before: https://main--cc--adobecom.hlx.page/?martech=off
- After: https://main--cc--adobecom.hlx.page/?martech=off&milolibs=mwpw-143627-region-picker-404s--milo--robert-bogos

**Homepage:**
- After: https://main--homepage--adobecom.hlx.page/homepage/index-loggedout?martech=off&milolibs=mwpw-143627-region-picker-404s--milo--robert-bogos

**Blog:**
- After: https://main--blog--adobecom.hlx.page/?milolibs=mwpw-143627-region-picker-404s--milo--robert-bogos

**Milo:**
- Before: https://stage--milo--adobecom.hlx.page/drafts/rbogos/page-default?martech=off
- After: https://mwpw-143627-region-picker-404s--milo--robert-bogos.hlx.page/drafts/rbogos/page-default?martech=off


libs/utils/utils.js
```diff
@@ -461,6 +461,16 @@ export async function loadBlock(block) {
         await init(block);
       } catch (err) {
         console.log(`Failed loading ${name}`, err);
+
+        const { message, errorType, module } = error;
+        if (message && errorType && module) {
+          window.lana.log(`${message}; blockPath: ${blockPath} | href: ${window.location.href} | ${err.reason || err.error || err.message || err}`, {
+            clientId: 'feds-milo',
```
overmyheadandbody, line 468: This file is used by all Milo consumers, we should not add any logic here that is not relevant to everyone, like this FEDS-specific client id.

libs/utils/utils.js
```diff
@@ -461,6 +461,16 @@ export async function loadBlock(block) {
...
+            sampleRate: 1,
+            tags: `errorType=${errorType},module=${module}`,
+          });
+        }
```
overmyheadandbody, line 472: We should find another way to expose errors in the blocks themselves. utils could dispatch an error or something of the sorts, but handling should be done in the block.

overmyheadandbody (changes requested):
We want to keep the utils file as slim as possible, we'll actually want to review every line of code inside it in the near future, so I'd like to have some other alternatives to handle this.

robert-bogos:
@overmyheadandbody, I agree with keeping the utils file slim 👍. I've found another solution that does not involve modifying the utils. Essentially, the `loadBlock` function would only attempt to load a `feds-regionPicker` block (and fail) if the first class of the region picker is feds-regionPicker, or more generally, not `modal`. So, I tested for that and modified it to send a LANA log.

libs/blocks/global-footer/global-footer.js
```diff
@@ -226,6 +226,12 @@ class Footer {
     if (url.hash !== '') {
       // Hash -> region selector opens a modal
       decorateAutoBlock(regionPickerElem); // add modal-specific attributes
+      if (regionPickerElem.classList[0] !== 'modal') {
```
mokimo, line 229: Is this early enough to give us enough context? Or do we also need some logs post `loadBlock(regionPickerElem)` ?

mokimo (approved):
A lot better 👍

libs/blocks/global-footer/global-footer.js
mokimo, line 229: A TODO to remove those logs again, might make sense after MWPW-000000 is completed. (It will 100% be forgotten and stay here forever, but at least there might be a chance we remove this again 😁 )

overmyheadandbody (approved):
Yup, was thinking that this is our best bet for now as well. Let's add a TODO for the committers of the future to remove this if we manage to find the root cause

libs/blocks/global-footer/global-footer.js
overmyheadandbody: What I discussed with Robert is that the issues might be before that call to `loadBlock`, as that [method](https://github.com/adobecom/milo/blob/stage/libs/utils/utils.js#L442) seems to choose the first class. So we need to ensure the first class is not the region picker one

robert-bogos: Good idea 👍 I added a TODO

SilviuLCF (approved):
Regression pass 
https://jira.corp.adobe.com/browse/MWPW-143627

narcis-radu:
@robert-bogos - please take a look at the failing coverage

narcis-radu:
I have removed the `Ready for Stage` label until the checks pass. https://github.com/orgs/adobecom/discussions/997

Blainegunn:
@robert-bogos i see coverage failing but i also see @mokimo added the RFS label 
are you updating coverage or nah?

mokimo:
@Blainegunn for this particular case it's fine to skip on the coverage check, especially since it's throwaway/debug code. I'll leave it up to @robert-bogos though and remove the label for now. 

Also the tattva team should take this forward imo @robert-bogos wdyt?

robert-bogos:
@mokimo @Blainegunn The uncovered code is debug code, so I don't think it's relevant to spend time adding unit tests for it. Especially since the uncovered code will be removed in the future, which would result in failing unit tests. That means more work to figure out why they failed and remove them afterward.

Blainegunn:
I appreciate you @robert-bogos thank you

PR #2031, opened 2024-03-18
MWPW-143071 added new copy to OST dropdown

VKniaz:
* Add your
* Specific
* Features or fixes
Add "Upgrade now" as an option to the checkout dropdown.

Before:
Screenshot 2024-03-18 at 2 26 09 PM

After:
Screenshot 2024-03-18 at 1 37 58 PM

Resolves: [MWPW-143071](https://jira.corp.adobe.com/browse/MWPW-143071)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/vkniaziev/document?martech=off
- After: https://MWPW-143071--milo--VKniaz.hlx.page/drafts/vkniaziev/document?martech=off


mokimo (approved):

overmyheadandbody (approved):
The PSI check is failing because the page content is too short, thus the footer takes over as the LCP element. The footer is loaded with a delay or based on an intersection observer to keep the main thread free for other tasks, thus it loads later in the flow, causing the LCP issues.

Ideally, the page would be authored to mimic a real-life example and contain more content.

3ch023 (approved):

PR #2048, opened 2024-03-20
MWPW-144813: Add 'xl-size' class for dialog-modal

mirafedas:
This change adds a new `xl-size` class for the modal window, which allows the modal window to fill all available screen height on mobile and tablet, and to be wider on desktop.
We set this class to the modal window that renders an iframe with the following pages:
https://www.adobe.com/creativecloud/whats-included/mini-plans/cci-all-apps-whats-included.html
https://www.adobe.com/creativecloud/whats-included/mini-plans/edu-all-apps-whats-included.html
https://www.adobe.com/creativecloud/whats-included/mini-plans/cct-all-apps-whats-included.html
https://www.adobe.com/creativecloud/whats-included/plans/cci-all-apps-whats-included.html 
https://www.adobe.com/creativecloud/whats-included/plans/edu-all-apps-whats-included.html
https://www.adobe.com/creativecloud/whats-included/plans/cct-all-apps-whats-included.html 
We may use it later for other use cases as well.

Resolves: [MWPW-144813](https://jira.corp.adobe.com/browse/MWPW-144813)

**Test URLs:**
- Before:
https://main--milo--adobecom.hlx.live/drafts/mirafedas/all-modals#cci-all-apps-whats-included
- After:
https://mwpw-144813-whatsincluded-styles--milo--mirafedas.hlx.page/drafts/mirafedas/all-modals#see-whats-included-xl

Note:
Make sure the ModHeaders plugin has the frame-ancestors set to * in the CSP response modifiers, without this on the 'After' page the modal content will not render because of the CORS.

### DIFF
--- a/libs/blocks/modal/modal.css
+++ b/libs/blocks/modal/modal.css
@@ -93,7 +93,7 @@
   margin: 12px;
 }
 
-.locale-modal{
+.locale-modal {
   text-align: center;
   padding: 48px 32px 30px;
 }
@@ -203,6 +203,17 @@
   .dialog-modal.commerce-frame > .section,
   .dialog-modal.commerce-frame > .fragment > .section {
     height: 100vh;
+  }
+
+  .dialog-modal.xl-size,
+  .dialog-modal.xl-size > .fragment,
+  .dialog-modal.xl-size > .fragment > .section {
+    height: 100%;
+  }
+
+  .dialog-modal.xl-size .milo-iframe {
+    height: 100%;
+    padding-bottom: 0;
   }
 }
 


libs/blocks/modal/modal.css
```diff
@@ -204,6 +204,17 @@
...
+  .dialog-modal.xl,
+  .dialog-modal.xl > .fragment,
+  .dialog-modal.xl > .fragment > .section {
+    height: 100%;
+  }
+
+  .dialog-modal.xl .milo-iframe {
+    height: 100%;
+    padding-bottom: 0;
+  }
```
npeltier, line 217: what are implicit values for those rules otherwise?
i'm noticing same setup for `> 1200px` and  `.dialog-modal.commerce-frame .milo-iframe`

mirafedas: @npeltier 
Without these rules:
- for the .dialog-modal the height value is 'fit-content'.
- for .fragment and .section we don't set values, so the default value ('auto') applies.
- for .milo-iframe height is set to 0, and padding-bottom to 56.25%.

The class .commerce-frame should be used only on commerce related modal windows.
'What's Included' content is not related to commerce, so it makes sense to use a more general class name.

libs/blocks/modal/modal.js
```diff
@@ -171,6 +171,9 @@ export async function getModal(details, custom) {
       For modals without the 'commerce-frame' class height auto adjustment is not applicable */
       iframe.style.height = '100%';
     }
+    if (/\/creativecloud\/whats-included\//.test(iframe.src)) {
```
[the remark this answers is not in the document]
mirafedas: Got it, I'll remove this and request authors to set the 'xl' class on all What's Included modal windows

honstar (approved):

Axelcureno (approved):

Roycethan (commented):
@mirafedas  Plz review this url once , , i still see the issue here:
https://main--cc--adobecom.hlx.page/drafts/fredw/document?milolibs=mwpw-144813-whatsincluded-styles--milo--mirafedas#modal-buy-ind
image

mirafedas:
@Roycethan the modal you are checking has classes `dialog-modal commerce-frame`. We agreed that all non-commerce-related modals should not have the `commerce-frame` class, I guess the authors just didn't remove that class yet. To test the current change the modal should have classes `dialog-modal xl`, then it will be wider on desktop.
cc @afmicka

mokimo (approved):

overmyheadandbody (commented):
Just a few notes:
* the modal content doesn't load on the After link, it might be worth checking out;
* if I'm remembering correctly, we don't usually have just `.<size>` classes, they-re attached to some other meaning, say `.<size>-spacing`. Do you have another reference to such classes, are we using this pattern anywhere else? We want to keep things as consistent as possible for authors, so this might deserve some attention. An author could interpret a simple `xl` as a font choice for example, the purpose should be clear

mirafedas:
@overmyheadandbody 
good point, I renamed `xl` > `size-xl`.
The modal content does load, but with a delay (same as on prod), does it load for you if you wait for a couple of seconds?

overmyheadandbody:
Hey @mirafedas! Thanks for the changes! I just saw [another PR](https://github.com/adobecom/milo/pull/2067/files), suggesting `.size-s` as a class name, could you consolidate and use one or the other?

mirafedas:
@overmyheadandbody yes, doing that right now :)

PR #2059, opened 2024-03-21
MWPW-139842 - Block button styling - Support button.fill w/ <strong><em> authoring

ryanmparrish:
As an author in Milo, I want the ability to add a 'fill' button to a Milo page. This should be available in the same capacity as the rest of the buttons as shown here: https://milo.adobe.com/docs/authoring/button-mapping  

Screen Shot 2024-03-04 at 2 43 45 PM

Resolves: [MWPW-139842](https://jira.corp.adobe.com/browse/MWPW-139842)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rparrish/buttons?martech=off
- After: https://rparrish-button-fill--milo--adobecom.hlx.page/drafts/rparrish/buttons?martech=off

Additional live pages can be tested w/ the url param `?miloLibs=rparrrish-button-fill`
(No`fill` buttons should be present on any live content w/ the exception of bricks)
https://main--cc--adobecom.hlx.page/products/illustrator/free-trial-download?miloLibs=rparrrish-button-fill
https://main--cc--adobecom.hlx.page/products/illustrator?miloLibs=rparrrish-button-fill
https://main--cc--adobecom.hlx.page/products/premiere?miloLibs=rparrrish-button-fill
https://main--cc--adobecom.hlx.page/products/firefly?miloLibs=rparrrish-button-fill

### DIFF
--- a/libs/utils/decorate.js
+++ b/libs/utils/decorate.js
@@ -8,8 +8,8 @@ export function decorateButtons(el, size) {
     let buttonType = 'outline';
     if (parent.nodeName === 'EM' || parent.nodeName === 'STRONG') {
       buttonType = parent.nodeName === 'EM' ? 'outline' : 'blue';
-      if ((parent.parentElement?.nodeName === 'EM') ||
-          (parent.parentElement?.nodeName === 'STRONG')) {
+      if ((parent.parentElement?.nodeName === 'EM')
+        || (parent.parentElement?.nodeName === 'STRONG')) {
         buttonType = 'fill';
       }
     }
@@ -19,7 +19,7 @@ export function decorateButtons(el, size) {
     } else {
       button.classList.add('con-button', buttonType);
       if (size) button.classList.add(size);
-    }    
+    }
     const actionArea = button.closest('p, div');
     if (actionArea) {
       actionArea.classList.add('action-area');
@@ -29,7 +29,7 @@ export function decorateButtons(el, size) {
 
   for (const btn of buttons) {
     const parent = btn.parentElement;
-    if (parent.parentElement?.nodeName === 'EM' || parent.parentElement?.nodeName === 'STRONG') { 
+    if (parent.parentElement?.nodeName === 'EM' || parent.parentElement?.nodeName === 'STRONG') {
       const grandParentBtns = parent.parentElement.querySelectorAll('.con-button');
       parent.parentElement.replaceWith(...grandParentBtns);
     } else if (parent.nodeName === 'EM' || parent.nodeName === 'STRONG') {
--- a/libs/utils/decorate.js
+++ b/libs/utils/decorate.js
@@ -6,7 +6,8 @@ export function decorateButtons(el, size) {
   const buttonTypeMap = { STRONG: 'blue', EM: 'outline', A: 'blue' };
   buttons.forEach((button) => {
     const parent = button.parentElement;
-    const buttonType = buttonTypeMap[parent.nodeName] || 'outline';
+    let buttonType = buttonTypeMap[parent.nodeName] || 'outline';
+    if (parent.nodeName === 'EM' && parent.parentElement.nodeName === 'STRONG') buttonType = 'fill';
     if (button.nodeName === 'STRONG') {
       parent.classList.add('con-button', buttonType);
       if (size) parent.classList.add(size); /* button-l, button-xl */
--- a/libs/utils/decorate.js
+++ b/libs/utils/decorate.js
@@ -3,26 +3,40 @@ import { createTag } from './utils.js';
 export function decorateButtons(el, size) {
   const buttons = el.querySelectorAll('em a, strong a, p > a strong');
   if (buttons.length === 0) return;
-  const buttonTypeMap = { STRONG: 'blue', EM: 'outline', A: 'blue' };
   buttons.forEach((button) => {
     const parent = button.parentElement;
-    let buttonType = buttonTypeMap[parent.nodeName] || 'outline';
-    if (parent.nodeName === 'EM' && parent.parentElement.nodeName === 'STRONG') buttonType = 'fill';
+    let buttonType = 'outline';
+    if (parent.nodeName === 'EM' || parent.nodeName === 'STRONG') {
+      buttonType = parent.nodeName === 'EM' ? 'outline' : 'blue';
+      if ((parent.parentElement?.nodeName === 'EM') ||
+          (parent.parentElement?.nodeName === 'STRONG')) {
+        buttonType = 'fill';
+      }
+    }
     if (button.nodeName === 'STRONG') {
       parent.classList.add('con-button', buttonType);
-      if (size) parent.classList.add(size); /* button-l, button-xl */
+      if (size) parent.classList.add(size);
     } else {
       button.classList.add('con-button', buttonType);
-      if (size) button.classList.add(size); /* button-l, button-xl */
-      parent.insertAdjacentElement('afterend', button);
-      parent.remove();
-    }
+      if (size) button.classList.add(size);
+    }    
     const actionArea = button.closest('p, div');
     if (actionArea) {
       actionArea.classList.add('action-area');
       actionArea.nextElementSibling?.classList.add('supplemental-text', 'body-xl');
     }
   });
+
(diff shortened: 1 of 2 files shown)


Sartxi (approved):
LGTM

meganthecoder:
@ryanmparrish I thought we used to have this, but removed it because it was flaky. When authoring, sometimes adding bold and italic came out like `<strong><em>CTA</em></strong>` and sometimes it was `<em><strong>CTA</strong></em>`. Is this still a concern?

ryanmparrish:
@meganthecoder  - I remember that pattern in the past but I have not been able to reproduce it since. I will do some more testing on possible authoring scenarios that may cause this and report back. If this is found to be a potential pitfall, I can include both conditions to cover in case. If anyone has an example of this LMK. Thanks

overmyheadandbody:
I initially wanted to +1 @meganthecoder's comment, because I definitely remember that this was in fact an issue in the past. I created a quick [test page](https://main--milo--adobecom.hlx.page/drafts/ramuntea/strong-italics) with a few basic scenarios. Except for the 3rd use-case, the order suggested by @ryanmparrish seems to check out.

However, that 3rd use-case led me to the 6th - say an author replaces a section that already hat italics applied with this new fill button, to which they also apply bold styling. Then the markup is the other way around. I've seen 'leftover' styles in the past, this might justify Megan's suggestions of checking for both.

ryanmparrish:
@overmyheadandbody  - Thanks for the test page. I created another [test document](https://rparrish-button-fill--milo--adobecom.hlx.page/drafts/rparrish/links-strong-italics) and I am not seeing this authoring gotcha when dealing w/ different link and `<em><strong>` combos. (I do see some oddities when splitting the string w/ part bold and part italic but it just falls back on the logic for blue button) 
ex. `<a><em>label1</em><strong>Label2</strong></a>`

It looks like EDS must have fixed the link inconsistencies at some point because it's rendering very consistently on all my testing. @auniverseaway - Any word on the grapevine around this behavior that your aware of? 

W/ that said, I can include a few conditions to cover both cases if needed, but I haven't been able to get that alt #6 combo to appear within an <a> tag at all so I'm unable to validate that condition.

rgclayton:
I think the "flayness" has been cleared up by the Franklin team. 
@auniverseaway Where you able to confirm that the `strong`, `italic` pattern has been corrected?

rgclayton:
I am seeing an issue if I set an link to italic first, then bold. It renders in plain.html just like `<em><strong>my button</strong></em>`. When the DOM is decorated that pattern removes the button completely. 

Otherwise you get <strong><em>my button</em></strong> which renders the fill button correctly. 

If I do bold first, then italic the fill button appears to work.

[Example page](https://rparrish-button-fill--milo--adobecom.hlx.page/drafts/rclayton/test/button-fill)
[Example page.plain.html](https://rparrish-button-fill--milo--adobecom.hlx.page/drafts/rclayton/test/button-fill.plain.html)

See the last Text block. 

- Learn more our solution is set to outline (italic only). 
- Fill was set with italic first, then bold and it's not rendering. 

image

libs/utils/decorate.js
```diff
@@ -6,7 +6,8 @@ export function decorateButtons(el, size) {
   const buttonTypeMap = { STRONG: 'blue', EM: 'outline', A: 'blue' };
   buttons.forEach((button) => {
     const parent = button.parentElement;
-    const buttonType = buttonTypeMap[parent.nodeName] || 'outline';
+    let buttonType = buttonTypeMap[parent.nodeName] || 'outline';
+    if (parent.nodeName === 'EM' && parent.parentElement.nodeName === 'STRONG') buttonType = 'fill';
```
rgclayton, line 10: I think we'll have to check for the reverse of this too.

rgclayton: I'm also noticing all <em> references are completely removed from the DOM when nested with a <strong>

elan-tbx (approved):

libs/utils/decorate.js
ryanmparrish: @rgclayton Thanks for the finds on that. I have updated my PR w/ some adjustments.
 
1) I am checking for both fill authoring patterns. <em><strong> or <strong><em>
2) Your example should show as expected now w/ out those removals. .. Here's why and the fix - We've traditionally trimmed all the button DOM elements that are used to target them. <em><strong> etc.  So in the view it's just an <a> tag (we occasionally style this action-area wrapper) When we start to author w/ these different potential orders and wrappings it can make that button .parentElement scope differ. Since the `.full` is two deep, and sometime authoring will stuff a side by side button example in a wrapping element and only one of them is in another it can get a bit tricky to strip it clean. I have included some logic to target the different button authoring potentials and clean the related DOM (conditional replaceWith) anyhow. [Take a peek](https://rparrish-button-fill--milo--adobecom.hlx.page/drafts/rparrish/links-strong-italics), when you get a sec. Thanks

rgclayton: Looking good now!

rgclayton (approved):

libs/utils/decorate.js
```diff
@@ -3,25 +3,40 @@ import { createTag } from './utils.js';
...
+      if (size) button.classList.add(size);
     }
     const actionArea = button.closest('p, div');
     if (actionArea) {
       actionArea.classList.add('action-area');
       actionArea.nextElementSibling?.classList.add('supplemental-text', 'body-xl');
     }
   });
+
+  for (const btn of buttons) {
```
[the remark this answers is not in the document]
ryanmparrish: @megan - I originally had it in the first loop but there was some content disappearing due to some authoring caveats. See @rgclaytons [example](https://github.com/adobecom/milo/pull/2059#issuecomment-2019127887). This way with a second for loop  vs. forEach we can incrementally check the buttons outerElement structure and not .remove any content when we run into the following pattern.  
```
<em> 
<a>
<strong> <a>... </strong
</a>
</em>
```
Originally, and on `main` the nested <a> tag gets removed all together. 

Thats the reasoning for this second loop. Open to other ideas but it got kind of gnarly supporting any/all authoring types and still sanitizing the DOM of their target elements.

ryanmparrish:
@narcis-radu @vivgoodrich  @vhargrave - After looking at this [discussion](https://github.com/orgs/adobecom/discussions/2073#discussioncomment-9059593), I wonder if considering the `#_btn-fill` authoring pattern is better for this approach and then could be expanded on for future button styling needs. Curious your thoughts when you have a minute.

vhargrave:
> @narcis-radu @vivgoodrich @vhargrave - After looking at this [discussion](https://github.com/orgs/adobecom/discussions/2073#discussioncomment-9059593), I wonder if considering the `#_btn-fill` authoring pattern is better for this approach and then could be expanded on for future button styling needs. Curious your thoughts when you have a minute.

I would say the following. In general I support using the #_btn-fill or | btn_fill pattern . What I've heard from our authors in express though, is that if they can use a bold link or an italic link to create a button, they prefer the authoring simplicity of that approach over | btn_fill. 
That's why they've asked us to allow for bold or italic link authoring for our most used buttons and only to use the #_btn-fill or | btn-fill pattern for buttons that are rarer, because bold and/or italic links can't cover more than 3 button types. I'd therefore say for this case it probably still makes sense to combine bold & italic to create a button type, since I imagine authors will find that simpler, even if the code is a bit gnarly. 
The moment we need more button types though, then we don't really have a choice anymore but to switch to the | btn-fill pattern imo. I also think with tools like dark alley, it will be much simpler to author those types of buttons.

I tested out your buttons on this old test page of mine btw, and couldn't find any bugs. So very nice! 
https://rparrish-button-fill--milo--adobecom.hlx.page/drafts/vhargrave/buttons

ryanmparrish:
@vhargrave  - Thanks for the thoughts and I agree, This 3 button any block authoring pattern makes sense and using thats mentioned authoring notation for any other buttons that express or others need fills the bill for those use cases. So w/ that said, i'm happy to hear there is no issues on you test page and I feel good about releasing this PR as proposed. Thanks for the testing and considerations. Carry on. Cheers

ryanmparrish:
@Sartxi - I assigned this to you for a verified label. Thanks

PR #2062, opened 2024-03-22
MWPW-140112 Preflight link check

rgclayton:
**Enhancement to preflights link checking.**
* All links on the page are now checked.
* New API call to check link status. Broken links show up in a table under SEO tab.
* With the use of the new API off origin and local link URLs are checked.

**Broken/problem links table:**
* Displays problem URL.
* Notes where the problem link is located on the page
* Notes status of URL

**Broken/problem links on page (preflight modal closed):**
* Highlights link in red with status

Resolves: [MWPW-140112](https://jira.corp.adobe.com/browse/MWPW-140112)

**Notes**
Each consumer will need to have the API added to their ./milo/config
BACOM is currently set up.
Current version needs VPN running to hit the link check API. We are working to move the API so it does not need VPN connection. 

**Test URLs:**
- After: https://preflight-check-links--milo--adobecom.hlx.page/drafts/rclayton/test-template?martech=off

**BACOM Test URLs**
- Before: https://main--bacom--adobecom.hlx.page/?martech=off
- After: https://main--bacom--adobecom.hlx.page/?milolibs=preflight-check-links&martech=off


rgclayton:
I will circle back to check on unit tests. Should be good to start testing.

libs/blocks/preflight/panels/seo.js
```diff
@@ -133,25 +135,109 @@ async function checkLorem() {
...
+      const liverHref = link.dataset.liveHref;
+      return liverHref;
+    });
+    const opts = {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({ urls }),
+    };
+    try {
+      const resp = await fetch(`${spidy.url}/api/url-http-status`, opts);
```
chrischrischris, line 199: nit: move `resp` out of try block so all this code doesn't have to be in the try block:
```
let resp;
try {
  resp = await fetch(...);
  if (!resp.ok) return;
} catch  (e) {
  console.error(`There was a problem connecting to the link check API ${spidy.url}/api/url-http-status. ${e}`);
  return;
}
...
```

libs/blocks/preflight/panels/seo.js
```diff
@@ -133,25 +135,109 @@ async function checkLorem() {
...
+          group[linkResult.id].dataset.status = status;
+        }
+      });
+    } catch (e) {
+      console.error(`There was a problem connecting to the link check API ${spidy.url}/api/url-http-status. ${e}`);
+    }
   }
 
-  if (badLink) {
+  if (badLinks.value.length > 0) {
```
chrischrischris, line 230: super duper nit - I've always preferred `if (badLinks.value.length) {` but doesn't really matter.

Out of curiosity I checked milo codebase:
`.length > 0)` used 81 times
`.length)` used 414 times

chrischrischris (approved):
Nice!

libs/blocks/preflight/panels/seo.js
rgclayton: Updated. Thanks!

Ruchika4 (approved):
news.adobe.com is coming up as 403 although the link is accessible in test link. However as confirmed by @rgclayton this is api issue and is being looked by Dipanshu's team hence approving this PR.
This feature will be helpful for authors. Thanks for adding this in preflight

PR #2065, opened 2024-03-22
MWPW-144125 and MWPW-144536 [MILO][MEP] Request for New Personalization Tags

vgoodric:
Adding a few pzn tags to the prod and stage objects.

Resolves: [MWPW-144125](https://jira.corp.adobe.com/browse/MWPW-144125) and [MWPW-144536](https://jira.corp.adobe.com/browse/MWPW-144536)

**Test URLs:**
N/A but for PR check:
- Before: https://main--milo--adobecom.hlx.page?martech=off
- After: https://addmeppzntags--milo--adobecom.hlx.page?martech=off

Note: Psi check will fail since there are no test URL's.

This PR is considered a blocker, but the other team that needs this data is also in a code freeze.  So as long as we can deploy to prod by 4/1 we're fine.

### DIFF
--- a/libs/features/personalization/entitlements.js
+++ b/libs/features/personalization/entitlements.js
@@ -13,6 +13,10 @@ const ENTITLEMENT_MAP = {
   'ab713720-91a2-4e8e-b6d7-6f613e049566': 'any-cc-product-no-stock',
   'b0f65e1c-7737-4788-b3ae-0011c80bcbe1': 'any-cc-product-with-stock',
   '934fdc1d-7ba6-4644-908b-53e01e550086': 'any-dc-product',
+  '6dfcb769-324f-42e0-9e12-1fc4dc0ee85b': 'always-on-promo',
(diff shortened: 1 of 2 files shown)


libs/features/personalization/entitlements.js
```diff
@@ -13,6 +13,10 @@ const ENTITLEMENT_MAP = {
   'ab713720-91a2-4e8e-b6d7-6f613e049566': 'any-cc-product-no-stock',
   'b0f65e1c-7737-4788-b3ae-0011c80bcbe1': 'any-cc-product-with-stock',
   '934fdc1d-7ba6-4644-908b-53e01e550086': 'any-dc-product',
+  '6dfcb769-324f-42e0-9e12-1fc4dc0ee85b': 'always-on-promo',
+  '015c52cb-30b0-4ac9-b02e-f8716b39bfb6': 'not-q-always-on-promo',
```
overmyheadandbody, line 17: Are there no stage equivalents for these two?

overmyheadandbody (approved):

mokimo (approved):

PR #43, opened 2022-06-02
Accordion

fullcolorcoder:
- Introduces version 1 of Accordion component 

Resolves: MWPW-108557

Test URLs:
- Before: https://main--milo--adobecom.hlx.page/drafts/bradjohn/accordion-how-to
- After: https://faq-refactor--milo--adobecom.hlx.page/drafts/bradjohn/accordion-how-to

PR #2068, opened 2024-03-25
MWPW-145566 remove faq from MILO_BLOCKS array

vhargrave:
Currently express is migrating its faq block so that it works with milo code like any other milo college project.There used to be a block in milo named faq, but it was refactored to be called accordion with addtional functionality - https://github.com/adobecom/milo/pull/43. 
Our faq block in express is not working though because milo believes it still has a faq block.

The faq entry in the block array in utils.js in milo needs to have this entry removed, so that our faq block in express works correctly. For milo this is simply code cleanup - less bytes to all our clients  ;)

Resolves: [MWPW-145566](https://jira.corp.adobe.com/browse/MWPW-145566)

**Test URLs:**
- After: https://vhargrave-mwpw-145566-faq-utils-cleanup--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -27,7 +27,6 @@ const MILO_BLOCKS = [
   'chart',
   'columns',
   'faas',
-  'faq',
   'featured-article',
   'figure',
   'form',


overmyheadandbody (approved):
Looks good. I see there's another reference in https://github.com/adobecom/milo/blob/stage/libs/blocks/list.js#L6, does that need to remain there?

fullcolorcoder (approved):

vhargrave:
> Looks good. I see there's another reference in https://github.com/adobecom/milo/blob/stage/libs/blocks/list.js#L6, does that need to remain there?

@overmyheadandbody I saw that too 🙌 . I'm pretty sure we could remove it, but am not fully certain, which is why I didn't.

mokimo (approved):

overmyheadandbody:
@vhargrave - yeah, that file seems to be a forgotten one to me, I couldn't find any references to it inside the Milo repo, we may want to remove it entirely at some point.

PR #2070, opened 2024-03-25
MWPW-142590 - [Share] enhancement - ability to edit text above icons

ryanmparrish:
Authors would like to modify the 'Share this page' text that appears above the social icons in the `Share` block.

- Enabled author-able share titles
- Updated the twitter logo use new `X` version
Screen Shot 2024-03-25 at 1 47 09 PM

Note: The variant `.inline` does not support a title and strips out any authored text in the block content. 

Resolves: [MWPW-142590](https://jira.corp.adobe.com/browse/MWPW-142590)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rparrish/share?martech=off
- After: https://rparrish-share-update--milo--adobecom.hlx.page/drafts/rparrish/share?martech=off

[Authoring Docs](https://main--milo--adobecom.hlx.page/docs/authoring/blocks/share)

- Before: https://main--milo--adobecom.hlx.page/docs/authoring/examples/share-block?martech=off
- After: https://rparrish-share-update--milo--adobecom.hlx.page/docs/authoring/examples/share-block?martech=off

Testing notes: To author a custom title in the `Share` block, simply add the title in the blocks first content row. Works with or without the custom share url authoring options. 

Screen Shot 2024-03-25 at 2 00 16 PM


vgoodric (approved):

libs/blocks/share/share.js
```diff
@@ -37,35 +37,58 @@ export async function getSVGsfromFile(path, selectors) {
...
-    'linkedin',
-    'pinterest',
-    'reddit',
-  ];
-  block.innerHTML = '';
+  const platforms = getPlatforms(block);
+  const rows = block.querySelectorAll(':scope > div');
+  const childDiv = rows[0].querySelector(':scope > div');
+  const emptyRow = childDiv?.innerText.trim() === '';
+  /* eslint-disable  no-confusing-arrow,no-useless-escape */
```
elan-tbx, line 60: nit: I know there's a pre-existing comment like this on this file, but we should really only be disabling eslint rules for specific lines/blocks of code. As it is right now, this comment would disable those rules for the entire file... Probably best practice to use [`eslint-disable-next-line` or add an `eslint-enable`](https://eslint.org/docs/latest/use/configure/rules#disabling-rules) at the end of the block you're targeting.

libs/blocks/share/share.js
```diff
@@ -37,35 +37,58 @@ export async function getSVGsfromFile(path, selectors) {
...
+  const toSentenceCase = (str) => str && typeof str === 'string' ? str.toLowerCase().replace(/(^\s*\w|[\.\!\?]\s*\w)/g, (c) => c.toUpperCase()) : '';
```
elan-tbx, line 61: nit: If we're already disabling eslint for this line, do you think we could break it out into multiple lines for better readability?

elan-tbx (approved):
Couple nits but looking good to me otherwise!

libs/blocks/share/share.js
ryanmparrish: Ah yep. Good call. 
I added the `/* eslint-disable-next-line no-useless-escape */` part 
(I tested removing those 'useless' escapes and it didn't 'Sentence Case' right so I left that part as is) 
I adjusted the arg into a few lines for better readability and fixes 'no-confusing-arrow' clause. :) 
Thanks for the feedback

Sartxi (approved):

robert-bogos:
Hi @ryanmparrish ! Please do not forget to assign a QA engineer to verify the changes.

ryanmparrish:
@elan-tbx I assigned this to you for a verified label. Thanks

PR #2080, opened 2024-03-27
mwpw-145710: load ims via param from old domain

mokimo:
## Description
Load IMS from the old domain via query parameter `useAlternateImsDomain` to run some LH tests with both the old and new self hosted domain to measure the LH score impact.

## Related Issue
Resolves: [MWPW-145710](https://jira.corp.adobe.com/browse/MWPW-145710)

## Testing instructions
Scenario-1: Load a page, signing in should still work.
Secnario-2: Load a page with `&useAlternateImsDomain=true` and IMS should be loaded from the old domain

## Screenshots
Without query param
Screenshot 2024-03-27 at 14 51 33

With query param
Screenshot 2024-03-27 at 14 51 21

## Test URLs
**Milo using the old ims url**
- Before: https://main--milo--adobecom.hlx.live/drafts/osahin/gnav-1?martech=off&useAlternateImsDomain=true
- After: https://mwpw-145710-ims-self-hosting--milo--mokimo.hlx.live/drafts/osahin/gnav-1?martech=off&useAlternateImsDomain=true

**Milo using the new self hosted ims url**
- Before: https://main--milo--adobecom.hlx.live/drafts/osahin/gnav-1?martech=off&useAlternateImsDomain=true
- After: https://mwpw-145710-ims-self-hosting--milo--mokimo.hlx.live/drafts/osahin/gnav-1?martech=off&georouting=off&useAlternateImsDomain=true

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -136,6 +136,7 @@ export const MILO_EVENTS = { DEFERRED: 'milo:deferred' };
 
 const LANGSTORE = 'langstore';
 const PAGE_URL = new URL(window.location.href);
+
 function getEnv(conf) {
   const { host } = window.location;
   const query = PAGE_URL.searchParams.get('env');
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -793,7 +793,13 @@ export async function loadIms() {
       },
       onError: reject,
     };
-    loadScript(`${base}/deps/imslib.min.js`);
+    const params = new URLSearchParams(window.location.search);
+    const useLHControlGroup = params.get('useAlternateImsDomain');
+    if (useLHControlGroup) {
+      loadScript('https://auth.services.adobe.com/imslib/imslib.min.js');
+    } else {
+      loadScript(`${base}/deps/imslib.min.js`);
+    }
   }).then(() => {
     if (!window.adobeIMS?.isSignedInUser()) {
       getConfig().entitlements([]);
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -136,7 +136,6 @@ export const MILO_EVENTS = { DEFERRED: 'milo:deferred' };
 
 const LANGSTORE = 'langstore';
 const PAGE_URL = new URL(window.location.href);
-
 function getEnv(conf) {
   const { host } = window.location;
   const query = PAGE_URL.searchParams.get('env');
@@ -670,8 +669,7 @@ function decorateHeader() {
     header.remove();
     return;
   }
-  const headerQuery = new URLSearchParams(window.location.search).get('headerqa');
-  header.className = headerQuery || headerMeta || 'gnav';
+  header.className = headerMeta || 'gnav';
   const metadataConfig = getMetadata('breadcrumbs')?.toLowerCase()
   || getConfig().breadcrumbs;
   if (metadataConfig === 'off') return;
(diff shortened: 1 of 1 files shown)


libs/utils/utils.js
```diff
@@ -793,7 +793,13 @@ export async function loadIms() {
       },
       onError: reject,
     };
-    loadScript(`${base}/deps/imslib.min.js`);
+    const params = new URLSearchParams(window.location.search);
```
overmyheadandbody, line 796: I see query params are also fetcher [here](https://github.com/adobecom/milo/blob/6c13dded0e8dc1a7cd22504c684b6cad3477a473/libs/utils/utils.js#L833). Could we get them early on and then re-use wherever neccessary?

libs/utils/utils.js
```diff
@@ -793,7 +793,13 @@ export async function loadIms() {
...
+    const useLHControlGroup = params.get('useAlternateImsDomain');
+    if (useLHControlGroup) {
+      loadScript('https://auth.services.adobe.com/imslib/imslib.min.js');
+    } else {
+      loadScript(`${base}/deps/imslib.min.js`);
+    }
```
overmyheadandbody, line 802: Could we trim this down a bit?
```
const path = useLHControlGroup ? 'auth.services...' : 'deps';
loadScript(path);
```

libs/utils/utils.js
```diff
@@ -793,7 +793,13 @@ export async function loadIms() {
       },
       onError: reject,
     };
-    loadScript(`${base}/deps/imslib.min.js`);
+    const params = new URLSearchParams(window.location.search);
+    const useLHControlGroup = params.get('useAlternateImsDomain');
```
overmyheadandbody, line 797: If you have something against the query param caching idea, we could couple this line with the previous to save some bytes.

overmyheadandbody (commented):
This being utils, I think we can tighten the belt a bit on number of characters

libs/utils/utils.js
mokimo: We already have `PAGE_URL` cached, using that!

libs/utils/utils.js
```diff
@@ -793,7 +790,10 @@ export async function loadIms() {
       },
       onError: reject,
     };
-    loadScript(`${base}/deps/imslib.min.js`);
+    const path = PAGE_URL.searchParams.get('useAlternateImsDomain')
+      ? 'https://auth.services.adobe.com/imslib/imslib.min.js'
```
overmyheadandbody, line 794: I'm guessing there's no need to consider stage URL, right?

libs/utils/utils.js
```diff
@@ -136,7 +136,6 @@ export const MILO_EVENTS = { DEFERRED: 'milo:deferred' };
 
 const LANGSTORE = 'langstore';
 const PAGE_URL = new URL(window.location.href);
-
```
overmyheadandbody, line 139: I think we usually do leave an empty line before functions, though

overmyheadandbody (approved):
Just a final nit session, but otherwise looks great! 👍

libs/utils/utils.js
mokimo: Don't think so 🤔 We didn't have anything before at least

narcis-radu: We did have an use-case for loading STAGE, but I think we can skip it now.

narcis-radu (approved):

SilviuLCF (approved):
Testing details 
https://jira.corp.adobe.com/browse/MWPW-145710 

@mokimo is any reason why navs are not working on testing urls ?

mokimo:
@SilviuLCF I've updated the nav links

PR #2083, opened 2024-03-27
MWPW-145728 Increase timeout; add metadata support

chrischrischris:
Increase the default timeout to 3.5s.  Also add ability to modify the timeout 

Resolves: [MWPW-145728](https://jira.corp.adobe.com/browse/MWPW-145728)

**Test URLs:**
- After: https://ccc-target-timeout--milo--adobecom.hlx.page/?martech=off

As promised, here is the Jira ticket to create unit tests for martech.js https://jira.corp.adobe.com/browse/MWPW-145975

### DIFF
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -605,29 +605,6 @@ export async function getPersConfig(info, override = false) {
     config.executionOrder = '1-1';
   }
 
-  if (infoTab) {
-    const infoObj = infoTab?.reduce((acc, item) => {
-      acc[item.key] = item.value;
-      return acc;
-    }, {});
-    config.manifestOverrideName = infoObj?.['manifest-override-name']?.toLowerCase();
-    config.manifestType = infoObj?.['manifest-type']?.toLowerCase();
-    const executionOrder = {
-      'manifest-type': 1,
-      'manifest-execution-order': 1,
-    };
-    Object.keys(infoObj).forEach((key) => {
-      if (!infoKeyMap[key]) return;
-      const index = infoKeyMap[key].indexOf(infoObj[key]);
-      executionOrder[key] = index > -1 ? index : 1;
-    });
-    config.executionOrder = `${executionOrder['manifest-execution-order']}-${executionOrder['manifest-type']}`;
-  } else {
-    // eslint-disable-next-line prefer-destructuring
-    config.manifestType = infoKeyMap[1];
-    config.executionOrder = '1-1';
-  }
-
   config.manifestPath = normalizePath(manifestPath);
   const selectedVariantName = await getPersonalizationVariant(
     config.manifestPath,
--- a/libs/martech/martech.js
+++ b/libs/martech/martech.js
@@ -71,6 +71,23 @@ const handleAlloyResponse = (response) => {
     .filter(Boolean);
 };
 
+function sendAnalytics(val) {
+  window.alloy('sendEvent', {
+    documentUnloading: true,
+    xdm: {
+      eventType: 'web.webinteraction.linkClicks',
+      web: {
+        webInteraction: {
+          linkClicks: { value: 1 },
+          type: 'other',
+          name: val,
+        },
+      },
+    },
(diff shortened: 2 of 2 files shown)


vgoodric:
@chrischrischris Just updated with analytics.  Please review.

mokimo (approved):
Do we have any data on how this impacts LH scores when target is on?

libs/martech/martech.js
```diff
@@ -71,20 +71,48 @@ const handleAlloyResponse = (response) => {
...
 };
 
+function sendAnalytics(val) {
+  window.alloy('sendEvent', {
+    documentUnloading: true,
+    xdm: {
+      eventType: 'web.webinteraction.linkClicks',
+      web: {
+        webInteraction: {
+          linkClicks: { value: 1 },
```
overmyheadandbody, line 81: I'm guessing there's no other relevant property we can use other than `linkClicks`?

libs/martech/martech.js
```diff
@@ -1,7 +1,7 @@
-import { getConfig, loadIms, loadLink, loadScript } from '../utils/utils.js';
+import { getConfig, getMetadata, loadIms, loadLink, loadScript } from '../utils/utils.js';
 
 const ALLOY_SEND_EVENT = 'alloy_sendEvent';
-const TARGET_TIMEOUT_MS = 2000;
+const TARGET_TIMEOUT_MS = 3500;
```
overmyheadandbody, line 4: I share @mokimo's concerns regarding performance impact of increasing this value. Hopefully we can bring it back down with self-hosted IMS and martech scripts

overmyheadandbody (approved):

chrischrischris:
@mokimo @overmyheadandbody 

> Do we have any data on how this impacts LH scores when target is on?

It pretty much doesn't impact lighthouse scores because in the US the call returns in ~500ms, so the timeout is never hit - therefore max time doesn't matter.

What it may affect is the 30 day Core Web Vitals because the timeout is happening much more often in other countries.  However we don't have any data as to how much it happens, where it happens, and what kind of response times we're getting.  The analytics part of this PR is to get those measurements.

libs/features/personalization/personalization.js
```diff
@@ -605,29 +605,6 @@ export async function getPersConfig(info, override = false) {
     config.executionOrder = '1-1';
   }
 
-  if (infoTab) {
```
chrischrischris, line 608: fyi - this was a duplicate code block that was added in a merge by mistake

libs/martech/martech.js
vgoodric: @overmyheadandbody @mokimo 
Yeah, we are doing a few things at once:
1. increasing the timeout because we strongly suspect 2 seconds is too short.
2. adding an analytic event to get quickly a sense of how many more users we get for every .25 seconds longer we wait. 3. Also trying to ascertain our current fail rate.  The hope is that this is a temporary measure (couple of months)
4. creating a grafana dashboard to track LHS on the same URL's with the target-timeout parameter dictating different wait times.
5. add a permanent lana log metric so we can monitor spikes in Target response times

But we're trying to get data to find the "goldilocks" timeout.  Wait too little and too many people don't get the Target activity.  Too long and we're waiting for a few fringe users.

vgoodric (approved):

libs/martech/martech.js
vgoodric: consulted with some analytics experts.  Here was their answer:
We are using fields currently set up to capture data in certain ways. if we were doing something other then this temporary setup we would look to set these values part of other tags, but given this is firing it's own tag it makes more sense for it to be seen as a link click rather then a page view to collect correctly.

narcis-radu (approved):

libs/martech/martech.js
narcis-radu: It would be great to capture the connection type (where possible) through `navigator.connection.effectiveType` and see whether the timeout is hit for just 2g only.

narcis-radu:
The PR is failing because of code coverage, however unit tests will be added in a subsequent PR. See details in the JIRA ticket - https://jira.corp.adobe.com/browse/MWPW-145975

PR #2088, opened 2024-03-28
Mwpw-136871: TwP merch card

Axelcureno:
- TwP merch card 

Resolves: [MWPW-136871](https://jira.corp.adobe.com/browse/MWPW-136871)

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/drafts/axel/stories/mwpw-136871?martech=off
- After: https://main--cc--adobecom.hlx.page/drafts/axel/stories/mwpw-136871?milolibs=MWPW-136871--milo--axelcureno&martech=off


yesil:
@Axelcureno TWP should support multiple offers with possibly strikethrough prices. And only 1 offer (ABP, M2M, PUF) will be visible at a time.
I guess for that, we need to use `merch-offer-select` or whatever web component we have for that use case.

3ch023 (approved):

PR #2092, opened 2024-03-28
MWPW-145574 - Dynamic Nav

JasonHowellSlavin:
- Adds code to add the gnav source to sessionStorage to allow for the same nav experience in subsequent clicks
- `entry` pages will set gnavSource to sessionStorage is a gnav source meta tag is present
- `on` pages will check for a gnavSource in sessionStorage, and if one is found, either modify the gnav-source meta tag, or create one
- pages without the dynamic-nav metadata will do nothing or remove gnavSource from sessionStorage

Context:
The wiki in the ticket provides more context, but it was desired by our stakeholders that this only be a single tab experience.

Resolves: [MWPW-145574](https://jira.corp.adobe.com/browse/MWPW-145574)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/slavin/dynamic-nav/hugo-boss-case-study-entry?martech=off
- After: https://dynamic-nav--milo--jasonhowellslavin.hlx.page/drafts/slavin/dynamic-nav/hugo-boss-case-study-entry?martech=off
- Before: https://main--bacom--adobecom.hlx.page/drafts/slavin/dynamic-nav/aem-rtcdp-entry?martech=off
- After: https://dynamic-nav-config--bacom--adobecom.hlx.page/drafts/slavin/dynamic-nav/aem-rtcdp-entry?milolibs=dynamic-nav&martech=off
To test: 

1. Go to the second after link. On the Hugo Boss page
2. Observe the nav (pricing and resources)
3. Right click "watch now" and open in a new tab
4. Observe the nav is the default milo nav ("Explore")
5. From the Hugo Boss page left click "Watch Now" to open in the same tab
6. Observe the pricing and resources nav loads


narcis-radu:
@JasonHowellSlavin - I was under the impression that this would go into `bacom` and not Milo. If we want to add this feature in Milo we should support more use-cases but at this point there aren't any requirements.
The custom code that you've built can be added in in scripts.js just before `loadArea`, this way you ensure that all the logic is executed before the navigation is actually loaded.

JasonHowellSlavin:
> @JasonHowellSlavin - I was under the impression that this would go into `bacom` and not Milo. If we want to add this feature in Milo we should support more use-cases but at this point there aren't any requirements. The custom code that you've built can be added in in scripts.js just before `loadArea`, this way you ensure that all the logic is executed before the navigation is actually loaded.

I suppose no other teams replied to the [discussion thread ](https://github.com/orgs/adobecom/discussions/1914) so perhaps it is better to put this functionality in Bacom at the moment so bytes are not put in front of other consumers. Do you think it is preferable to move this code to bacom? 

I chose to put it in `checkForPageMods` because there is similar code there checking for `meta` tags and running code based on those tags and properties. This is called before `decorateDocumentExtras` so I thought it would be fine, but if you see that there may be an issue with execution timing, I am happy to move it to scripts.js.

Thank you for your help and insight!

meganthecoder:
@JasonHowellSlavin Can you fix the lighthouse error? It'll be interesting to see the performance impacts.

narcis-radu:
@JasonHowellSlavin - but if we use the current approach, wouldn't a bacom nav be displayed on cc or the other way around if the meta is on?

libs/utils/utils.js
```diff
@@ -827,11 +827,13 @@ async function checkForPageMods() {
   const persMd = getMetadata('personalization');
   const promoMd = getMetadata('manifestnames');
   const targetMd = getMetadata('target');
+  const dynamicNavMd = getMetadata('dynamic-nav');
   let persManifests = [];
   const persEnabled = persMd && persMd !== 'off' && !offFlag('personalization');
   const targetEnabled = targetMd && targetMd !== 'off' && !offFlag('target') && !offFlag('martech');
   const promoEnabled = promoMd && promoMd !== 'off' && !offFlag('promo');
   const mepEnabled = persEnabled || targetEnabled || promoEnabled;
+  const dynamicNavEnabled = dynamicNavMd && (dynamicNavMd === 'entry' || dynamicNavMd === 'on');
```
narcis-radu, line 836: why not `const dynamicNavEnabled = ['entry', 'on'].includes(getMetadata('dynamic-nav'))`

libs/utils/utils.js
```diff
@@ -853,6 +855,31 @@ async function checkForPageMods() {
     persManifests = persManifests.concat(getPromoManifests(promoMd));
   }
 
+  if (dynamicNavEnabled) {
+    const gns = getMetadata('gnav-source');
```
narcis-radu, line 859: we should use descriptive names

libs/utils/utils.js
```diff
@@ -853,6 +855,31 @@ async function checkForPageMods() {
...
+
+    if (dynamicNavMd === 'entry' && gns) {
+      window.sessionStorage.setItem('gnavSource', gns);
```
narcis-radu, line 862: i think we can stop the execution after this, right?

narcis-radu (commented):
I am concerned that the session storage will be passed on from project to another (e.g. bacom to cc) and the dynamic nav will be wrongfully inherited.

libs/utils/utils.js
JasonHowellSlavin: Ah, great catch, thank you, I meant to change the name.

JasonHowellSlavin:
> I am concerned that the session storage will be passed on from project to another (e.g. bacom to cc) and the dynamic nav will be wrongfully inherited.

Hmm. That is a good point. I suppose we were assuming that the User Journey would not lead them out of the current project's pages, but the chance is not none. 

Perhaps we could check the `config`. Not all configs have `coderoot` or `imsClientId` but all the ones I checked (cc/dc/bacom/milo/express/helpx) had one or both. We could use those values in sessionStorage to check against to ensure the same project. 

I am a bit weary of adding something to the config as an identifier or something that authoring needs to author, like an additional metadata tag. What do you think?

narcis-radu:
@JasonHowellSlavin - I don't think that ims client ID or codeRoot are enough, we might have requirements where different set of pages in a project get different dynamic navigations. I think we need to add something else to distinguish between configs if this feature is going into Milo and can be used by all consumers.

JasonHowellSlavin:
> @JasonHowellSlavin - I don't think that ims client ID or codeRoot are enough, we might have requirements where different set of pages in a project get different dynamic navigations. I think we need to add something else to distinguish between configs if this feature is going into Milo and can be used by all consumers.

I have now added a requirement for `dynamicNavKey` to be set in a project's config. This key can be the project's shorthand name, in our instance `bacom`. The dynamic nav will now check for that key, and if it does not find it, stop executing. I also moved the code to it's own file in order to keep things cleaner. Let me know what you think, and thank you.

chrischrischris (approved):

libs/features/dynamic-navigation.js
```diff
@@ -0,0 +1,29 @@
...
+
+  if (!dynamicNavKey) return;
+
+  if (metadataContent === 'entry' && gnavSourceContent) {
+    window.sessionStorage.setItem('gnavSource', gnavSourceContent);
+    window.sessionStorage.setItem('dynamicNavKey', dynamicNavKey);
+    return;
+  }
+
+  if (metadataContent === 'on') {
```
mokimo, line 14: If you stick to your early return ruleset, you can avoid the nested if/else
if (metadataContent !== 'on') return

libs/utils/utils.js
```diff
@@ -853,6 +854,13 @@ async function checkForPageMods() {
     persManifests = persManifests.concat(getPromoManifests(promoMd));
   }
 
+  if (dynamicNavMd) {
```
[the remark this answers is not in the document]
mokimo: I'd do something such as this in the gnav block:
```
 let url = getMetadata('gnav-source') || `${locale.contentRoot}/gnav`;
 if(dynamicNav) {
    const { default: dynamicNav } = await import('../features/dynamic-navigation.js');
   url = getDynamicNavUrl(); // just import getting the config / metadata in the module itself rather than passing them as params
 }
```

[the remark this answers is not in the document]
narcis-radu: I will add my feedback here since I cannot add it in the compare URL. @JasonHowellSlavin - the dynamicNav method should only be imported if the key is set, see @mokimo 's code example.

JasonHowellSlavin:
Closing this PR in favor of https://github.com/adobecom/milo/pull/2180.

@narcis-radu I updated using your last comment here on the above PR. Thank you.

PR #2094, opened 2024-03-29
MWPW-145779: TWP modal not working when open in new tab

mirafedas:
When multiple 'link-block modal' links share the same modalPath and modalHash, the corresponding modal is preloaded multiple times. For instance, when navigating to a [page with a corresponding hash](https://main--cc--adobecom.hlx.page/products/substance3d/discover/3d-art-maker#modal-substance-3d-collection), the DOM contains multiple instances of the modal, resulting in overlapping modals being rendered.
This PR introduces a filter that removes duplicate links.

Resolves: [MWPW-145779](https://jira.corp.adobe.com/browse/MWPW-145779)

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/products/substance3d/discover/3d-art-maker#modal-substance-3d-collection
- After: https://main--cc--adobecom.hlx.page/products/substance3d/discover/3d-art-maker?milolibs=mwpw-145779-twp-modal-in-new-tab--milo--mirafedas#modal-substance-3d-collection
https://mwpw-145779-twp-modal-in-new-tab--milo--mirafedas.hlx.page/drafts/mirafedas/all-modals#mini-plans-web-cta-creative-cloud-card-2
Note: on the After page there will be CORS issues, so you need to have ModHeaders enabled with Access-Control-Allow-Origin: * and frame-ancestors: *


libs/utils/utils.js
```diff
@@ -751,7 +769,7 @@ function decorateSection(section, idx) {
     blocks: [...links, ...blocks],
     el: section,
     idx,
-    preloadLinks: blockLinks.autoBlocks,
+    preloadLinks: removeDuplicatedLinkBlocks(blockLinks.autoBlocks),
```
narcis-radu, line 772: Hi @mirafedas - I'm thinking whether we should not add the modals in the first place instead of removing them. Maybe in https://github.com/adobecom/milo/blob/stage/libs/blocks/modal/modal.js#L149 we can check whether a modal with the same ID has been already added and if so, skip loading it.

libs/utils/utils.js
```diff
@@ -715,6 +715,24 @@ async function loadFooter() {
   await loadBlock(footer);
 }
 
+export function removeDuplicatedLinkBlocks(blocks) {
```
yesil, line 718: nit:  `filterDuplicateLinkBlocks`

yesil (approved):

honstar (approved):

honstar:
@mirafedas : can you please add a test page on `milo` so that the LH check can also run if possible? Thx

PR #2095, opened 2024-04-01
MWPW-144810: Quiz Entry - Add option cards and text to the block

JackySun9:
* Quiz Entry - Add option cards and text to the block

Resolves: [MWPW-144810](https://jira.corp.adobe.com/browse/MWPW-144810)

**Test URLs:**
- After: https://quiz-entry-block--milo--jackysun9.hlx.page/drafts/colloyd/quiz-entry/?martech=off

### DIFF
--- a/libs/blocks/quiz-entry/mlField.js
+++ b/libs/blocks/quiz-entry/mlField.js
@@ -31,6 +31,21 @@ export const getFiResults = async (endpoint, apiKey, input, numberOfItems, valid
   return result;
 };
 
+const handleInput = () => {
+  const mlFieldText = document.querySelector('#ml-field-input').value;
+  if (mlFieldText.length > 0) {
+    const quizoptions = document.querySelectorAll('.quiz-option');
+    quizoptions.forEach((option) => {
+      option.classList.add('disabled');
+    });
+  } else {
+    const quizoptions = document.querySelectorAll('.quiz-option');
+    quizoptions.forEach((option) => {
+      option.classList.remove('disabled');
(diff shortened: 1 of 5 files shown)


fullcolorcoder (approved):
Looks good on the page. I'll help clean up the CSS in the next bits. If you have time to cleanup the linting that'd be cool otherwise not standing in the way!

JackySun9:
> Looks good on the page. I'll help clean up the CSS in the next bits. If you have time to cleanup the linting that'd be cool otherwise not standing in the way!

We need them later, so I will leave them there

libs/blocks/quiz-entry/mlField.js
```diff
@@ -31,6 +31,21 @@ export const getFiResults = async (endpoint, apiKey, input, numberOfItems, valid
   return result;
 };
 
+const handleInput = () => {
```
colloyd, line 34: Ideally, we could just update the state of all of the cards and add or removed disabled from them, allowing the standard card machinery to take effect, rather than branching it to add classes as we are here. I'm not sure how to do this, but if we could pull that off, we wouldn't have two ways of disabling, which is confusing. Any ideas on how we can just update card state from the field @fullcolorcoder ?

One reason that this is problematic, is that we are missing the bit that adds the disabled attribute, which is what actually prevents tabbing on the card. If you enter text in the field, and tab through, it should go straight to the continue button, skipping the cards. However, since the cards are only getting a class, they look disabled, but aren't actually disabled.

PR #2109, opened 2024-04-04
[MWPW-142625] PEP relative to App Switcher

overmyheadandbody:
This makes the necessary changes to:
* position the PEP prompt relative to the App Switcher;
* focus the App Switcher if the PEP prompt is dismissed;
* cancel displaying the PEP prompt if the App Switcher is open.

This is safe to be merged in the feature branch, but we'll need to have a few things on our radar:
* we need to get confirmation from the UNav team on the 1.1 version release that contains the updates to interact with the App Switcher, currently this is available just on production; we do have safety mechanisms is place in case their methods aren't available;
* currently the Profile menu `adobeProfile.getUserProfile()` public method is throwing errors, thus we can't retrieve profile data, I've reported the issue on the appropriate Slack channel.

This PR also addresses a bunch of leftover TODOs from the code.

Resolves: [MWPW-142625](https://jira.corp.adobe.com/browse/MWPW-142625)

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/drafts/ramuntea/firefly?milolibs=project-pep--milo--adobecom
- After: https://main--cc--adobecom.hlx.page/drafts/ramuntea/firefly?milolibs=project-pep--milo--overmyheadandbody

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -16,6 +16,7 @@ import {
   getExperienceName,
   getFedsPlaceholderConfig,
   hasActiveLink,
+  icons,
   isDesktop,
   isTangentToViewport,
   lanaLog,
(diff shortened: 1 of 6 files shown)


mokimo (approved):

robert-bogos:
@overmyheadandbody I see that some universal navigation unit tests are failing

robert-bogos (approved):

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -536,8 +509,8 @@ class Gnav {
     };
 
     await Promise.all([
-      loadScript(`https://${environment}.adobeccstatic.com/unav/1.0/UniversalNav.js`),
-      loadStyle(`https://${environment}.adobeccstatic.com/unav/1.0/UniversalNav.css`),
+      loadScript(`https://${environment}.adobeccstatic.com/unav/1.1/UniversalNav.js`),
```
narcis-radu, line 512: I know this is a feature branch, but if we're planning to get this to STAGE soon, then we need to include @salonijain3 and her team for awareness, they have a similar request to update the universal navigation version.

narcis-radu (approved):

narcis-radu:
@overmyheadandbody - I see some tests are failing

overmyheadandbody:
The unit tests will be addressed by @robert-bogos as part of the overall unit tests work for PEP. Since this is open against a feature branch, I will merge for now.

libs/blocks/global-navigation/global-navigation.js
overmyheadandbody: Yes, this definitely needs to be considered before merging to `stage`, I've asked the UNav team when we can use this new version on production, but haven't seen a reply yet.

robert-bogos:
Unit tests are added through [this PR](https://github.com/adobecom/milo/pull/2136).

PR #2111, opened 2024-04-04
MWPW-144830 - [Loc v2] Cancel loc project

Sartxi:
**Users should  be able to cancel a Milo loc project.**

Once a project has been created (has project id) the actions area will include a `Cancel Project` button.
Clicking this button will result in the following:
- To avoid mistakes, UI will display a confirm dialog which includes a 'nevermind' option and proceed warnings.
- Will display success toast dialog when cancel job is done processing.
- Project goes into status 'cancelled' and all project actions are then unavailable.
- UI displays 'Project Cancelled' message with a note concerning SharePoint docs.
- Polling for status interval is then cleared when `miloc` status API returns 'cancelled' to avoid unnecessary requests.

Screenshots of pertinent UI changes:
Screenshot 2024-04-04 at 2 28 13 PM
Screenshot 2024-04-04 at 2 23 38 PM

**_Note: I also fixed a few typos in UI status and excel status calls._**

Resolves: [MWPW-144830](https://jira.corp.adobe.com/browse/MWPW-144830)

**Test URLs:**
Will need to pull down branch and serve locally with a localization project to do any testing


libs/blocks/locui/utils/miloc.js
```diff
@@ -86,24 +99,38 @@ export async function getProjectStatus() {
...
+export async function cancelProject() {
+  allowSyncToLangstore.value = false;
+  allowSendForLoc.value = false;
+  allowCancelProject.value = false;
+  let url = await getMilocUrl();
+  setStatus('service', 'info', 'Cancelling project');
+  const opts = { method: 'POST' };
+  url = `${url}cancel-project?project=${heading.value.projectId}`;
+  const resp = await fetch(url, opts);
+  if (resp.status === 200) setExcelStatus('Project cancelled', '');
```
sabyamon, line 129: Hi @Sartxi - How are we handling the error scenario?

1. When the project Id is missing (which is difficult to replicate through the UI i think) - HTTP 400 - we might not need to address this
2. When the project is already in cancelled state (again, sort of difficult to happen through the UI as project status call will send the status and u will be hiding all the CTAs in that case) - HTTP 500 - we might not need to address this
3. Error in cancelling a GLaaS project and eventually cancelling the project - HTTP 500 - possible use case to handle

sabyamon (commented):
Left a comment regarding the error handling. The rest looks good to me.

libs/blocks/locui/utils/miloc.js
Sartxi: @sabyamon I have added a handler for 500 errors that will display the error for the user in the status toast dialog

sabyamon (approved):

PR #2123, opened 2024-04-09
MWPW-146129 App Launcher overlaps the menu in Unav in the devices

sharmrj:
## Description

In non dekstop viewports, if the user clicks on one of the unav buttons while the menu is open, the popovers and the menu overlap. This PR brings the behavior in line with what happens on AEM.

## Related Issue
Resolves: [MWPW-146129](https://jira.corp.adobe.com/browse/MWPW-146129)

## Testing instructions

See the JIRA ticket

URL for testing: 

Before: https://main--milo--adobecom.hlx.live
After: https://unav-mobile--milo--sharmrj.hlx.live


libs/blocks/global-navigation/global-navigation.js
```diff
@@ -280,6 +280,9 @@ class Gnav {
     if (this.useUniversalNav) {
       delete this.blocks.profile;
       this.blocks.universalNav = toFragment`<div class="feds-utilities"></div>`;
+      this.blocks.universalNav.addEventListener('click', () => {
+        document.querySelector('.feds-toggle[aria-expanded=true]')?.click();
```
mokimo, line 284: There's a `closeAllDropdowns` method that you can call

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -280,6 +280,9 @@ class Gnav {
     if (this.useUniversalNav) {
       delete this.blocks.profile;
       this.blocks.universalNav = toFragment`<div class="feds-utilities"></div>`;
+      this.blocks.universalNav.addEventListener('click', () => {
```
mokimo, line 283:
```suggestion
      this.blocks.universalNav.addEventListener('click', () => closeAllDropdowns())
```

sharmrj: This seems not to work in adobe.com due to the following line (line 289 in utilities.js) in `closeAllDropDowns()`
```javascript
    if ('fedsPreventautoclose' in el.dataset) return;
```

mokimo: IMO emulating a click on the open element is **not** the right path. You should investigate why `fedsPreventautoclose` exists to begin with (might be related to the keyboard navigation using arrow keys, escape, etc.) and otherwise look into adapting the `closeAllDropdowns()` method

sharmrj: I don't. The test links don't load the unav because the request for the JS file `403`s. Since it's a one liner, I tested it by adding a breakpoint in the Gnav Constructor and adding in the line manually (you need to do cmd + s for it to work).

sharmrj: AFAICT it is related somewhat to the keyboard navigation files. It's difficult to say just from reading the code, but it seems that it makes somewhat of a distinction between specifically a dropdown and the menu as it opens in mobile and tablet viewports.
IMO `closeAllDropdowns` isn't what we're looking for here. What we're looking for is specifically inside the `decorateToggle` method:
```javascript
    const onToggleClick = async () => {
      const isExpanded = toggle.getAttribute('aria-expanded') === 'true';
      toggle.setAttribute('aria-expanded', !isExpanded);
      this.elements.navWrapper.classList.toggle('feds-nav-wrapper--expanded', !isExpanded);
      closeAllDropdowns();
      setCurtainState(!isExpanded);
      toggle.setAttribute('daa-ll', `hamburgermenu|${isExpanded ? 'open' : 'close'}`);

      if (this.blocks?.search?.instance) {
        this.blocks.search.instance.clearSearchForm();
      } else {
        await this.loadSearch();
      }

      if (isExpanded) setHamburgerPadding();
    };

```
The most expedient way to call this function without a fairly significant refactor is by clicking on the element it's attached to. Furthermore it ties the behavior we want to the behavior of the actual html elements the end user interacts with, and not to the abstract way it is done in the code, which is liable to churn. In other words clicking on the toggle element decouples the implementation of the behavior from the implementation of how exactly clicking on the toggle hides/shows the menu.

overmyheadandbody (changes requested):
I agree that this might not be the best approach. My hunch - this should actually work if it wasn't for the desktop viewport check [here](https://github.com/adobecom/milo/blob/stage/libs/blocks/global-navigation/global-navigation.js#L226). We want to keep the behaviors between desktop and mobile separate, though, so I'd try adding an if/else. If the viewport is desktop, execute the current logic. If the viewport is mobile, close the hamburger menu. There's likely already logic that closes it (see the `decorateToggle` method), you might have to expose it to be able to simply re-use is where needed.

sharmrj:
I'm not sure why unit tests are failing. The global-navigation tests seem to be passing (and global-navigation is the only modue I've touched).

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -295,6 +296,16 @@ export function closeAllDropdowns({ type } = {}) {
...
+export const toggleMenuMobile = (toggle) => {
+  if (!toggle) return;
+  const isExpanded = toggle.getAttribute('aria-expanded') === 'true';
+  toggle.setAttribute('aria-expanded', !isExpanded);
+  document.querySelector(selectors.navWrapper)?.classList?.toggle('feds-nav-wrapper--expanded', !isExpanded);
+  closeAllDropdowns();
+  setCurtainState(!isExpanded);
+  toggle.setAttribute('daa-ll', `hamburgermenu|${isExpanded ? 'open' : 'close'}`);
+};
+
```
overmyheadandbody, line 308: I'd argue that since this isn't used in multiple files, it can be left in the `global-navigation.js` one. You would then have direct access to `this.elements.navWrapper` and `toggle` and you'd keep the logic close to where it's being used.

sigadamvenkata:
Verified the changes and we done see overlap issue now. PF below screenshots
image

salonijain3 (approved):

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -615,6 +619,16 @@ class Gnav {
     });
   };
 
+  toggleMenuMobile = () => {
+    const toggle = this.elements.mobileToggle;
+    const isExpanded = toggle?.getAttribute('aria-expanded') === 'true';
```
overmyheadandbody, line 624: Same exact code is used on L284 & L656, might be worth turning it into a function
```
const isToggleExpanded = () => this.elements.mobileToggle?.getAttribute('aria-expanded') === 'true';
```

overmyheadandbody (commented):
Let's add some relevant pages to the PR body where the hamburger menu interactions can actually be tested.

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -638,12 +654,8 @@ class Gnav {
...
 
     const onToggleClick = async () => {
-      const isExpanded = toggle.getAttribute('aria-expanded') === 'true';
-      toggle.setAttribute('aria-expanded', !isExpanded);
-      this.elements.navWrapper.classList.toggle('feds-nav-wrapper--expanded', !isExpanded);
-      closeAllDropdowns();
-      setCurtainState(!isExpanded);
-      toggle.setAttribute('daa-ll', `hamburgermenu|${isExpanded ? 'open' : 'close'}`);
+      this.toggleMenuMobile();
+      const isExpanded = this.isToggleExpanded();
```
overmyheadandbody, line 658: You're just using the value of this variable once, you could simply use the function call on L666, where it's used.

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -280,6 +280,10 @@ class Gnav {
     if (this.useUniversalNav) {
       delete this.blocks.profile;
       this.blocks.universalNav = toFragment`<div class="feds-utilities"></div>`;
+      this.blocks.universalNav.addEventListener('click', () => {
+        const isExpanded = this.isToggleExpanded();
+        if (isExpanded) this.toggleMenuMobile();
```
overmyheadandbody, line 285: This could be just one line, since `isExpanded` is not used anywhere else, `if (this.isToggleExpanded()) this.toggleMenuMobile();`

overmyheadandbody (approved):
You could trim down a couple more lines, added comments, otherwise looks good.

Personally, I would have still integrated this into the `closeOnClickOutside` [logic](https://github.com/adobecom/milo/blob/stage/libs/blocks/global-navigation/global-navigation.js#L225), which would have also covered the use-case where a user clicks anywhere outside the hamburger menu area, but maybe that can happen in the future.

sharmrj:
> Personally, I would have still integrated this into the `closeOnClickOutside` [logic](https://github.com/adobecom/milo/blob/stage/libs/blocks/global-navigation/global-navigation.js#L225), which would have also covered the use-case where a user clicks anywhere outside the hamburger menu area, but maybe that can happen in the future.

The issue with that is that it changes the current behavior in mobile viewports where tapping in the thin area outside the menu (when it's open) does nothing. I'm not sure if the current behavior should change.

PR #2131, opened 2024-04-10
Fix: Adding gray bg and making header bold for group link on gnav

bandana147:
- Adds a gray background option for link header
- Makes gradient link group header bold 

Resolves: [MWPW-146166](https://jira.corp.adobe.com/browse/MWPW-146166)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.live/drafts/blaishram/index-loggedout-ace0687v4?martech=off
- After: https://mwpw-146166--milo--bandana147.hlx.live/drafts/blaishram/index-loggedout-ace0687v4?martech=off

### DIFF
--- a/libs/blocks/global-navigation/base.css
+++ b/libs/blocks/global-navigation/base.css
@@ -222,6 +222,7 @@
   .feds-navLink[class *= '-gradient'] {
     display: flex;
     border-radius: 4px;
+    font-weight: 600;
   }
 
   .feds-menu-column--group .feds-navLink[class *= '-gradient'] {
@@ -280,5 +281,9 @@
 
   .feds-navLink--ai-gradient {
     background: linear-gradient(90deg, #bce3ff, #ffe9d3, #f8d5e4);
+  }
+
+  .feds-navLink--gray-gradient {
+    background: #eaeaea;
   }
 }


libs/blocks/global-navigation/base.css
```diff
@@ -222,6 +222,7 @@
   .feds-navLink[class *= '-gradient'] {
     display: flex;
     border-radius: 4px;
+    font-weight: 600;
```
mokimo, line 225: Is it **always bolded**, or just in the gray gradient?

mokimo (approved):

libs/blocks/global-navigation/base.css
bandana147: I was informed that all gradients should be bolded.

overmyheadandbody: @mokimo - for context, we shouldn't be currently using any colorful gradient on production currently, it's still in a test phase. So I think it's safe to apply the rule to all such elements 

And come to think of it, could we confirm with the stakeholder whether they still see a need for all the other gradient classes? Maybe we could remove them already if that test failed.

overmyheadandbody (approved):
I'd suggest we discuss the need for all the other gradients with the stakeholder, it might be that we can remove them to keep the file clean.

bandana147:
The failing Lighthouse score could be attributed to the usage of a global navigation with multiple fragments including some from other sites, some of which result in 404 errors. Since this change involves only a CSS modification with background color and font weight adjustments, it should have no impact on performance.

robert-bogos (approved):

salonijain3 (approved):

salonijain3:
This is a high priority PR as this is needed for ACE0687 which needs be live for 5/7 and 28 days is needed for the process.

PR #2139, opened 2024-04-11
MWPW-146034 - Quiz entry block accessibility

colloyd:
* resolved accessibility concerns when the ml input is used - cards are now properly disabled with the attribute
* general code refinements for more clarity, specifically for getting string values

Resolves: [MWPW-146034](https://jira.corp.adobe.com/browse/MWPW-146034)

**Test URLs:**
- After: https://quiz-entry-accessibility--milo--colloyd.hlx.page/drafts/colloyd/quiz-entry/?martech=off


fullcolorcoder (approved):
Nice!

PR #2152, opened 2024-04-16
MWPW-135821 introduce mep custom action & use it for card collection

npeltier:
- and inaugurating first custom action for replacing cards in collection

Resolves: [MWPW-135821](https://jira.corp.adobe.com/browse/MWPW-135821)

with following manifest added to the page

| action | selector | page filter (optional) | all |
|---|---|---|---|
| replace | in-block:merch-card-collection |   | /drafts/npeltier/fragments/merch/products/catalog/photoshop/promo |

before & after personalisation (see photoshop card)
**Test URLs:**
- After: https://mwpw-135821--milo--npeltier.hlx.page/?martech=off

- Before: https://main--cc--adobecom.hlx.page/drafts/npeltier/promocard-test?milolibs=mwpw-135821--milo--npeltier
- After: https://main--cc--adobecom.hlx.page/drafts/npeltier/promocard-test?milolibs=mwpw-135821--milo--npeltier&mep=%2Fdrafts%2Fnpeltier%2Fpromo-test.json--target-challenger+name


libs/features/personalization/personalization.js
```diff
@@ -51,6 +51,10 @@ const DATA_TYPE = {
   TEXT: 'text',
 };
 
+export const MERCH_CARD_COLLECTION_SELECTOR = 'in-card-collection';
```
yesil, line 54: I think this constant should not be here, personalization.js should not know about blocks needing delayed MEP based personalization.

libs/features/personalization/personalization.js
```diff
@@ -51,6 +51,10 @@ const DATA_TYPE = {
...
+
+const CUSTOM_SELECTORS = [MERCH_CARD_COLLECTION_SELECTOR];
```
yesil, line 56: this too.

libs/features/personalization/personalization.js
```diff
@@ -289,6 +293,14 @@ function getSection(rootEl, idx) {
     : rootEl.querySelector(`:scope > div:nth-child(${idx})`);
 }
 
+function registerCustomAction(cmd) {
```
yesil, line 296: should we differentiate normal MEP rules from these ones?
if we expose all the MEP rules and have a flag when they are successfully executed at page load, then blocks such as `merch-card-collection` can go over the rules without that flag and find matching rules and execute them.

test/blocks/merch-card-collection/merch-card-collection.test.js
```diff
@@ -161,7 +161,23 @@ describe('Merch Cards', async () => {
 
   it('should override cards when asked to', async () => {
     const el = document.getElementById('multipleFilters');
-    el.dataset.overrides = '/override-photoshop,/override-express';
+    setConfig({
```
yesil, line 164: I suggest putting MEP rules in own config object.
it can be a module scoped config and be exposed from `personalization.js`.

libs/features/personalization/personalization.js
npeltier: this is the smallest footprint to have to get "custom" actions defined so mep knows it just needs to register them as such.

test/blocks/merch-card-collection/merch-card-collection.test.js
npeltier: this is what we do, in config.mep.custom

libs/features/personalization/personalization.js
npeltier: they are, all normal processing is stopped if identified action is custom, and then put in config.mep.custom

yesil: actually I meant, do we need to differentiate, sorry for the language.

test/blocks/merch-card-collection/merch-card-collection.test.js
yesil: I see 👍

test/blocks/merch-card-collection/merch-card-collection.test.js
```diff
@@ -161,7 +161,23 @@ describe('Merch Cards', async () => {
...
+      ...conf,
+      mep: {
+        custom: {
```
yesil, line 167: suggestion: `deferred` or `lazy` instead of `custom`?

libs/features/personalization/personalization.js
yesil: If selectors had a pattern or prefix, we wouldn't need to add custom selectors in this file.

yesil (approved):
In my opinion, personalization.js should not know about blocks needing to perform MEP during block initialisation.

Otherwise, the current code looks concise and simple.

test/blocks/merch-card-collection/merch-card-collection.test.js
npeltier: To me it's less talkative than custom. custom means in that context that it will not be handled by perso, which implies it's lazy/deferred, but also that something else should take care of it :)

npeltier:
> If selectors had a pattern or prefix, we wouldn't need to add custom selectors in this file.
i suggested "custom-" prefix for actions :) but @vgoodric was preferring custom selectors :)

libs/features/personalization/personalization.js
vgoodric: if we did that, you wouldn't need the constant.  You just extract the 2nd part of the string and in your block you just use your own block name.

npeltier: sure :) or custom:whatever but yes in that case in-block:merch-card-collection sounds right indeed :)

libs/blocks/merch-card-collection/merch-card-collection.js
```diff
@@ -291,8 +292,7 @@ export default async function init(el) {
   }
 
   const cardsRoot = await cardsRootPromise;
-  const overrides = el.dataset[OVERRIDE_PATHS];
-  const overridePromises = overrides?.split(',').map(fetchOverrideCard);
+  const overridePromises = mep?.custom?.[MERCH_CARD_COLLECTION_SELECTOR]?.map(fetchOverrideCard);
```
vgoodric, line 295: look in personalization.js for the handleFragmentCommand function.  In particular, this line:
`if (config.mep.preview) a.dataset.manifestId = manifestPath;`

This is what makes the highlight feature possible.  I recommend doing this as well so people can see what has been changed.

npeltier:
thanks for the great suggestion about prefixes @yesil @vivgoodrich, just updated

vgoodric:
> thanks for the great suggestion about prefixes @yesil @vivgoodrich, just updated

Can you also update your sample manifests? So I can see it all in action?

npeltier:
just did @vivgoodrich , will now tackle your 2nd suggestion

3ch023:
do we have a task to cover it with nala tests? 
cc @afmicka

afmicka:
> do we have a task to cover it with nala tests? cc @afmicka

@3ch023 not as jira. I am working on updating promo tests today and tomorrow (all actions on fragments turned out to be supported but not working) and had this in mind too. We can create the jira task for this as well. cc: @npeltier

vgoodric (approved):

PR #2171, opened 2024-04-22
MWPW-146342 Implement the base functionality for MEP on the Global Nav bar

sharmrj:
Minimal implementation for MEP to work with the global navigation component.

Tested using a sharepoint site (see 'After' url below. See [here](https://adobe.sharepoint.com/:x:/r/sites/milo-academy/_layouts/15/Doc.aspx?sourcedoc=%7B116dc9c6-7c21-400e-8f47-6f83b26f1160%7D&action=editnew) for the manifest used during testing)

Resolves: [MWPW-146342](https://jira.corp.adobe.com/browse/MWPW-146342)

**Test URLs:**

After: https://gnav-mep--milo--sharmrj.hlx.page/?martech=off

Before: https://main--moo--sharmrj.hlx.page/?mep=%2Fpersonalization.json--target-challenger+name&martech=off
After: https://main--moo--sharmrj.hlx.page/?mep=%2Fpersonalization.json--target-challenger+name&milolibs=gnav-mep--milo--sharmrj&martech=off


mokimo:
Is this PR so that any gnav "feature" can be personalized, just as it is currently possible with blocks?

Since the current limitation is that only complete navs can be swapped out, rather than individual elements

sharmrj:
> Is this PR so that any gnav "feature" (brand, menu, universal nav, etc.) can be personalized, just as it is currently possible with blocks?
> 
> Since the current limitation is that only complete navs can be swapped out, rather than individual elements
Yes any of the incoming html can be personalized with this implementation.

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -307,11 +308,19 @@ export function trigger({ element, event, type } = {}) {
...
 
 export async function fetchAndProcessPlainHtml({ url, shouldDecorateLinks = true } = {}) {
+  const { mep } = getConfig();
+
   const path = getFederatedUrl(url);
   const res = await fetch(path.replace(/(\.html$|$)/, '.plain.html'));
   const text = await res.text();
   const { body } = new DOMParser().parseFromString(text, 'text/html');
 
+  if (mep.globalnav.length) { // just in case
```
chrischrischris, line 318: nit: comment not needed

libs/features/personalization/personalization.js
```diff
@@ -154,7 +156,9 @@ const COMMANDS = {
 };
 
 function checkSelectorType(selector) {
-  return selector?.includes('/fragments/') ? 'fragment' : 'css';
+  if (selector?.includes('/fragments/')) return 'fragment';
+  if (selector?.startsWith(GLOBAL_NAV_SELECTOR)) return GLOBAL_NAV_SELECTOR;
```
chrischrischris, line 160: Should we account for possible whitespace before "global-navigation"?
```suggestion
  if (selector?.trim().startsWith(GLOBAL_NAV_SELECTOR)) return GLOBAL_NAV_SELECTOR;
```

libs/features/personalization/personalization.js
```diff
@@ -393,6 +397,10 @@ const getVariantInfo = (line, variantNames, variants) => {
         val: normalizePath(line[vn]),
         action,
       });
+    } else if ((action in COMMANDS || action in CREATE_CMDS)
+        && variantInfo.selectorType === GLOBAL_NAV_SELECTOR) {
```
chrischrischris, line 401: super nit: swap the && so that selectorType is first so it's a bit more obvious it's for gnav
```suggestion
    } else if (variantInfo.selectorType === GLOBAL_NAV_SELECTOR
        && (action in COMMANDS || action in CREATE_CMDS)) {
```

libs/features/personalization/personalization.js
```diff
@@ -393,6 +397,10 @@ const getVariantInfo = (line, variantNames, variants) => {
         val: normalizePath(line[vn]),
         action,
       });
+    } else if ((action in COMMANDS || action in CREATE_CMDS)
```
chrischrischris, line 400: I wonder if we want to gate it to `action` in COMMANDS / CREATE_CMDS - I could see gnav having its own set of specialized commands that wouldn't need to be defined in personalization.js.

libs/features/personalization/personalization.js
```diff
@@ -735,12 +744,18 @@ export function handleFragmentCommand(command, a) {
   return false;
 }
 
+export const handleGlobalNavCommands = (commands, manifestId, rootEl) => {
+  handleCommands(commands, manifestId, rootEl);
+  deleteMarkedEls(rootEl);
+};
```
chrischrischris, line 750: I think instead we should export `handleCommands` and `deleteMarkedEls`.  This will give globalnav code more flexibility to handle specialized cases.

libs/features/personalization/personalization.js
sharmrj: It's certainly possible that we could have those in the future. Would it be fine if we defined those somewhere in the global-navigation folder and import them here?

libs/features/personalization/personalization.js
sharmrj: I've exported `handleCommands` and `deletedMarkedEls`, and removed `handleGlobalNavCommands`.

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -307,11 +308,20 @@ export function trigger({ element, event, type } = {}) {
...
 
 export async function fetchAndProcessPlainHtml({ url, shouldDecorateLinks = true } = {}) {
+  const { mep } = getConfig();
+
   const path = getFederatedUrl(url);
   const res = await fetch(path.replace(/(\.html$|$)/, '.plain.html'));
   const text = await res.text();
   const { body } = new DOMParser().parseFromString(text, 'text/html');
 
+  if (mep.globalnav.length) {
```
mokimo, line 318: Will mep & mep.globalnav be always available or do we need to add a safety net? `mep?.globalnav?.length`

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -307,11 +308,20 @@ export function trigger({ element, event, type } = {}) {
...
+    mep.globalnav.forEach(({ manifestId, commands }) => {
```
mokimo, line 319: Why does the globalnav need to call this method, rather than MEP calling it? 
Is it because MEP does not know when the nav is rendered?

mokimo (approved):
LGTM, I just wonder if it's a good concept and idea to couple MEP & blocks.

With this PR the block(gnav) code is aware the MEP feature exists.

This makes it harder to maintain if we ever switch MEP out as it's a performance offender, as we'd not only have to refactor the feature - but also go through all blocks that use it.

libs/blocks/global-navigation/utilities/utilities.js
sharmrj: Good catch. They should be available AFAICT (I've not had any errors while testing), but I'm gonna add the optional chaining in anyway.

libs/blocks/global-navigation/utilities/utilities.js
sharmrj: Yes because MEP does not know when the nav is rendered but also because the html MEP operates on never actually enters the DOM. The information is extracted from it and inserted into the relevant html templates we make for the nav. 
In an earlier iteration I had a message passing implementation that used Custom Events/Broadcast Channel but that ended up introducing even more coupling.

vgoodric (changes requested):
As discussed in the meeting this morning, this PR is likely duplicate work.  Please hold on merging.

salonijain3:
@vgoodric Can you ping the other Milo PR on this? Is it already merged?

vgoodric:
> @vgoodric Can you ping the other Milo PR on this? Is it already merged?
> 
> cc: @narcis-radu @overmyheadandbody

@salonijain3 , just need to finish the unit tests, but no PR yet

vgoodric:
@sharmrj @salonijain3 @narcis-radu @overmyheadandbody : New PR ready https://github.com/adobecom/milo/pull/2252

PR #2180, opened 2024-04-22
MWPW-145574 - Dynamic Nav Gnav Implementation

JasonHowellSlavin:
Adds functionality to allow consumer projects to opt in to a dynamic nav experience. If a dynamicNavKey is registered in a consumer repo's config, this code adds logic to save the gnav source to sessionStorage allowing for the same nav experience in subsequent clicks.

- `entry` pages will set gnavSource to sessionStorage is a gnav source meta tag is present
- `on` pages will check for a gnavSource in sessionStorage, and if one is found, either modify the gnav-source meta tag, or create one
- pages without the dynamic-nav metadata will do nothing or remove gnavSource from sessionStorage
- projects without the key present, or where the key mismatches with the key saved in storage will use the default gnav experience

Context:
The wiki in the ticket provides more context, but it was desired by our stakeholders that this only be a single tab experience.

Based on comments in https://github.com/adobecom/milo/pull/2092 I am adding this PR as the approach and closing 2092. 

Resolves: [MWPW-145574](https://jira.corp.adobe.com/browse/MWPW-145574)

Test URLs:

Before: https://main--milo--adobecom.hlx.page/drafts/slavin/dynamic-nav/hugo-boss-case-study-entry?martech=off
After: https://dynamic-nav-gnav-impl--milo--jasonhowellslavin.hlx.page/drafts/slavin/dynamic-nav/hugo-boss-case-study-entry?martech=off
Before: https://main--bacom--adobecom.hlx.page/drafts/slavin/dynamic-nav/aem-rtcdp-entry?martech=off
After: https://dynamic-nav-gnav-impl-config--bacom--adobecom.hlx.page/drafts/slavin/dynamic-nav/aem-rtcdp-entry?milolibs=dynamic-nav&martech=off

To test:
Go to the second after link. On the Hugo Boss page
Observe the nav (pricing and resources)
Right click "watch now" and open in a new tab
Observe the nav is the default milo nav ("Explore")
From the Hugo Boss page left click "Watch Now" to open in the same tab
Observe the pricing and resources nav loads

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -947,14 +947,20 @@ class Gnav {
   };
 }
 
+const getSource = async () => {
+  const { locale, dynamicNavKey } = getConfig();
+  let url = getMetadata('gnav-source') || `${locale.contentRoot}/gnav`;
+  if (dynamicNavKey) {
+    const { default: dynamicNav } = await import('../../features/dynamic-navigation.js');
+    url = dynamicNav(url, dynamicNavKey);
+  }
+  return url;
+};
+
 export default async function init(block) {
   try {
-    const { locale, mep, dynamicNavKey } = getConfig();
-    let url = getMetadata('gnav-source') || `${locale.contentRoot}/gnav`;
-    if (dynamicNavKey) {
-      const { default: dynamicNav } = await import('../../features/dynamic-navigation.js');
-      url = dynamicNav(url, dynamicNavKey);
-    }
+    const { mep } = getConfig();
+    const url = await getSource();
     const content = await fetchAndProcessPlainHtml({ url })
       .catch((e) => lanaLog({
         message: `Error fetching gnav content url: ${url}`,
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -35,8 +35,6 @@ import {
 
 import { replaceKey, replaceKeyArray } from '../../features/placeholders.js';
 
-import dynamicNav from '../../features/dynamic-navigation.js';
-
 const CONFIG = {
   icons: {
     company: '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 133.5 118.1"><defs><style>.cls-1 {fill: #eb1000;}</style></defs><g><g><polygon class="cls-1" points="84.1 0 133.5 0 133.5 118.1 84.1 0"/><polygon class="cls-1" points="49.4 0 0 0 0 118.1 49.4 0"/><polygon class="cls-1" points="66.7 43.5 98.2 118.1 77.6 118.1 68.2 94.4 45.2 94.4 66.7 43.5"/></g></g></svg>',
@@ -954,6 +952,7 @@ export default async function init(block) {
     const { locale, mep, dynamicNavKey } = getConfig();
     let url = getMetadata('gnav-source') || `${locale.contentRoot}/gnav`;
     if (dynamicNavKey) {
+      const { default: dynamicNav } = await import('../../features/dynamic-navigation.js');
       url = dynamicNav(url, dynamicNavKey);
     }
     const content = await fetchAndProcessPlainHtml({ url })
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -35,6 +35,8 @@ import {
 
 import { replaceKey, replaceKeyArray } from '../../features/placeholders.js';
 
+import dynamicNav from '../../features/dynamic-navigation.js';
+
 const CONFIG = {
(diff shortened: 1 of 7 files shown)


libs/blocks/global-navigation/global-navigation.js
```diff
@@ -949,8 +949,12 @@ class Gnav {
...
 export default async function init(block) {
   try {
-    const { locale, mep } = getConfig();
-    const url = getMetadata('gnav-source') || `${locale.contentRoot}/gnav`;
+    const { locale, mep, dynamicNavKey } = getConfig();
+    let url = getMetadata('gnav-source') || `${locale.contentRoot}/gnav`;
+    if (dynamicNavKey) {
+      const { default: dynamicNav } = await import('../../features/dynamic-navigation.js');
+      url = dynamicNav(url, dynamicNavKey);
+    }
```
sharmrj, line 957: I think we should probably have a new function `getGnavSource` (or something along those lines) that produces the url and contains all the logic for generating the url within it.

bandana147: I agree with Raghav, we should extract it into a separate function that generates the URL.

libs/features/dynamic-navigation.js
```diff
@@ -0,0 +1,21 @@
...
+
+  if (metadataContent !== 'on') return url;
+
+  if (key !== window.sessionStorage.getItem('dynamicNavKey')) return url;
+
+  const source = window.sessionStorage.getItem('gnavSource');
+
+  if (!source) return url;
+
+  return source;
```
bandana147, line 20: Since lines 12 and 14 have the same return value if the check fails, we can combine them as below
`if (metadataContent !== 'on' || key !== window.sessionStorage.getItem('dynamicNavKey')) return url;
`

Additionally, the following three lines can be reduced to one line:
  `  return window.sessionStorage.getItem('gnavSource') || url;`

libs/blocks/global-navigation/global-navigation.js
JasonHowellSlavin: Hmm. I am not so sure that I agree.  The `dynamic nav` is an opt in feature requiring a consumer to update their `config` to have the `dynamicNavKey` in order to kick off the experience. 

Though it may make the `init` function a bit cleaner, a function combining the two would mix both the default behavior and an opt in feature, which I am not certain aids in readability. 

Interested to see what you think.

sharmrj: In the future we may have logic added or subtracted from our hypothetical `getSource` function. `init` is and should be agnostic about the details of getting the url and that is communicated well by a separate function.

The separation between default and opt in behaviour can be communicated well enough in the function itself by the if statement you'll have to add. And the notion of what constitutes default vs non default behavior now and in the future is malleable enough that it makes sense to cordon it off from our main `init` logic so that it's clear in the future where gnav url generation logic should go.

JasonHowellSlavin: I appreciate your thoughtful and full response! I see what you're saying. To be honest, I am still on the fence because it feels a bit like premature optimization and we'd be adding 3 lines of code by pulling these 5 out into a function.

Maybe @mokimo can weigh in since he recommended the approach.

mokimo: @sharmrj So you'd think something like this?

```js
const getSource = async () => {
    const { locale, dynamicNavKey } = getConfig();
    let url = getMetadata('gnav-source') || `${locale.contentRoot}/gnav`;
    if (dynamicNavKey) {
      const { default: dynamicNav } = await import('../../features/dynamic-navigation.js');
      url = dynamicNav(url, dynamicNavKey);
    }
}

  try {
    const { locale, mep } = getConfig();
    const url = await getSource()
    const content = await fetchAndProcessPlainHtml({ url })
      .catch((e) => lanaLog({
        message: `Error fetching gnav content url: ${url}`,
        e,
        tags: 'errorType=error,module=gnav',
      }));
....
```

@sharmrj and @bandana147 are the new gnav code owners 😁 So if that's how they'd prefer it, no objections

sharmrj: Yup. That's what I was envisioning.

JasonHowellSlavin: Sounds good. I'll make those changes you requested @sharmrj, and I'll be sure to update the unit tests to reflect the new function. Hopefully should have something for you to look at a bit later today. I'll be sure to update the docs as well @mokimo. 

Thank you both for your input!

sharmrj (approved):

Deva309 (approved):

bandana147 (approved):

mokimo:
@JasonHowellSlavin what's the timing on this? Make sure to assign a QA/QE and have them apply the ready for stage label to take this forward

PR #2185, opened 2024-04-23
MWPW-147127 Scrolling on the gnav in mobile portrait and landscape view is not working as expected ,navigation menu when opened is not seen rendered completely in MILO Pages

sharmrj:
The content in the gnav menu (in mobile/tablet viewports, when it's open) was being cut-off on the bottom due to an incorrect height value in the wrapper. Correcting the height value for the wrapper in non desktop viewports fixes the issue

Resolves: [MWPW-147127](https://jira.corp.adobe.com/browse/MWPW-147127)

**Test URLs:**

Before: https://main--milo--adobecom.hlx.page?martech=off
After: https://mobile-gnav--milo--sharmrj.hlx.page?martech=off
  
Before: https://main--cc--adobecom.hlx.page/creativecloud?martech=off
After: https://main--cc--adobecom.hlx.live/creativecloud?milolibs=mobile-gnav--milo--sharmrj&martech=off

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.css
+++ b/libs/blocks/global-navigation/global-navigation.css
@@ -62,7 +62,7 @@ header.global-navigation {
   right: 20px; /* hamburger menu gutter */
   display: none;
   flex-direction: column;
-  height: 100vh;
+  height: calc(100vh - var(--feds-height-nav));
   border-top: 1px solid var(--feds-borderColor--light);
   background-color: var(--feds-background-nav--light);
 }
--- a/libs/blocks/global-navigation/global-navigation.css
+++ b/libs/blocks/global-navigation/global-navigation.css
@@ -62,7 +62,7 @@ header.global-navigation {
   right: 20px; /* hamburger menu gutter */
   display: none;
   flex-direction: column;
-  height: calc(100vh - var(--feds-height-nav));
+  height: 100vh;
   border-top: 1px solid var(--feds-borderColor--light);
   background-color: var(--feds-background-nav--light);
 }
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -615,7 +615,7 @@ class Gnav {
 
     return this.loadDelayed().then(() => {
       this.blocks.search.instance = new this.Search(this.blocks.search.config);
-    });
+    }).catch(() => {});
   };
 
   isToggleExpanded = () => this.elements.mobileToggle?.getAttribute('aria-expanded') === 'true';


narcis-radu:
Please provide a testing URL

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -62,7 +62,7 @@ header.global-navigation {
   right: 20px; /* hamburger menu gutter */
   display: none;
   flex-direction: column;
-  height: 100vh;
+  height: calc(100vh - var(--feds-height-nav));
```
overmyheadandbody, line 65: There might be promos or other elements on top of the gnav, there's padding logic that should be handling this. Please add some screenshots and some valid test URLs so we can better understand what you're trying to fix.

overmyheadandbody (changes requested):
The issue is actually here - https://github.com/adobecom/milo/blob/stage/libs/blocks/global-navigation/global-navigation.js#L658-L664

The logic that adds padding is no longer being called, that should be investigated.

Just subtracting the height does not cover all the use cases. Breadcrumbs count towards height. Eventual Promos on top of the nav count towards height. That's why it needs to be calculated on the fly. Figuring out what happened to the logic I've referenced at the top should handle all these use-cases.

sharmrj:
I've tested on https://main--cc--adobecom.hlx.live/creativecloud?milolibs=mobile-gnav--milo--sharmrj&martech=off

You're right . I was under the impression that `setHamburgerPadding` was being called but was accounting for items above it (I misinterpreted the bottom). The problem was in `await loadSearch`, which was throwing an error that was caught elsewhere (and the error is lana logged so it doesn't appear on the console), so execution was never returned to our bit of code. I imagine we might need to log this error or should the catch I put in be empty as it is now? I think we don't use search anywhere currently (?) so is the error in loading search a problem?
The padding is set only after we close the hamburger for the first time, not when we open it for the first time (this is in the original logic, prior to #2184 being merged into stage), so I've changed that. Should I add some logic so we don't get needless repeat invocations of this method? The padding being removed in the desktop check is overridden by a `padding: 0 !important` in the min-width: 900px media query, do we want both?

mokimo:
@sharmrj The mention about search is a great one - it might be that we are able to get rid of it completely. That would deserve a due diligence story and checking if any consumers still use it.

Blog currently uses a search variation, but they also use the old gnav and an incompatible search

sharmrj:
I've added [MWPW-147270](https://jira.corp.adobe.com/browse/MWPW-147270) as a followup

spadmasa:
Verified the changes on below url :
https://main--cc--adobecom.hlx.live/creativecloud?milolibs=mobile-gnav--milo--sharmrj
image
image
image

PR #2190, opened 2024-04-23
MWPW-146036 - Rig up quiz entry button

colloyd:
* Support for ml filtering
* Debug support using ?debug=quiz-entry
* Support for defining a maximum number of questions in the block before it will redirect
* Support for pressing the enter key in the ml field

Resolves: [MWPW-146036](https://jira.corp.adobe.com/browse/MWPW-146036)

**Test URLs:**
- After: https://quiz-entry-continue-button--milo--colloyd.hlx.page/drafts/colloyd/quiz-entry/?martech=off

Testing notes: 
You can now specify the number of questions that it will go through before redirecting to the main quiz, though this may not be used immediately, it does provide some future proofing if they want to try adding a few more questions before redirecting
You can prevent the redirect and get some console logs for ml field entries and quizState by adding debug=quiz-entry in the query string, like this:
- https://quiz-entry-continue-button--milo--colloyd.hlx.page/drafts/colloyd/quiz-entry/?martech=off&debug=quiz-entry

### DIFF
--- a/libs/blocks/quiz-entry/mlField.js
+++ b/libs/blocks/quiz-entry/mlField.js
@@ -1,7 +1,7 @@
 import { html } from '../../deps/htm-preact.js';
 import { getConfig } from '../../utils/utils.js';
 
-export const getFiResults = async (endpoint, apiKey, input, numberOfItems, validFiCodes) => {
+export const getMLResults = async (endpoint, apiKey, threshold, input, count, validFiCodes) => {
   const { env } = getConfig();
   const subdomain = env === 'prod' ? 'cchome-dev' : 'cchome-dev';
   const apiUrl = `https://${subdomain}.adobe.io/int/v1/models`;
@@ -11,7 +11,7 @@ export const getFiResults = async (endpoint, apiKey, input, numberOfItems, valid
     payload: {
       data: {
         input,
-        num_items: numberOfItems || 10,
+        num_items: count || 10,
         given_prod_list: validFiCodes,
       },
     },
@@ -28,9 +28,20 @@ export const getFiResults = async (endpoint, apiKey, input, numberOfItems, valid
     .then((response) => response.json())
     .catch((error) => window.lana.log(`ERROR: Fetching fi codes ${error}`));
 
+  let highestProb = null;
+  result.filtered = result?.data?.filter((item) => {
+    let isValid = false;
+    if (!highestProb) {
+      highestProb = item.prob;
+      isValid = true;
+    } else if (item.prob / highestProb > threshold) {
+      isValid = true;
+    }
+    return isValid;
+  });
   return result;
 };
 
-export const mlField = ({ cardsUsed, onMLInput, placeholderText }) => html`<div class="ml-field-container">
-    <input id="ml-field-input" class="ml-input" type="textarea" placeholder="${placeholderText}" oninput="${onMLInput}" disabled="${cardsUsed}" />
+export const mlField = ({ cardsUsed, onMLInput, onMLEnter, placeholderText }) => html`<div class="ml-field-container">
+    <input id="ml-field-input" class="ml-input" type="textarea" placeholder="${placeholderText}" oninput="${onMLInput}" onkeypress="${onMLEnter}" disabled="${cardsUsed}" />
   </div>`;
--- a/libs/blocks/quiz-entry/quiz-entry.js
+++ b/libs/blocks/quiz-entry/quiz-entry.js
@@ -1,18 +1,19 @@
-import { render, html, useEffect, useState } from '../../deps/htm-preact.js';
-import { createTag } from '../../utils/utils.js';
-import { getQuizEntryData } from './utils.js';
-import { mlField, getFiResults } from './mlField.js';
+import { render, html, useEffect, useState, useRef } from '../../deps/htm-preact.js';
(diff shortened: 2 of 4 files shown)


libs/blocks/quiz-entry/quiz-entry.js
```diff
@@ -73,127 +116,150 @@ const App = ({
...
+
   return html`<div class="quiz-entry-container">
     <div class="quiz-entry-title">${quizLists.strings[selectedQuestion.questions].heading}</div>
     <div class="quiz-entry-subtitle">${quizLists.strings[selectedQuestion.questions]['sub-head']}</div>
-    ${hasMLData && html`<${mlField} cardsUsed="${cardsUsed}" onMLInput="${onMLInput}" placeholderText="${getOptionsValue('fi_code', 'title')}"/><div class="results-container"></div>`}
+    ${hasMLData && html`<${mlField} 
+      cardsUsed="${cardsUsed}" 
+      onMLInput="${onMLInput}"
+      onMLEnter="${onMLEnter}"
+      placeholderText="${getOptionsValue('fi_code', 'title')}"/><div class="results-container"></div>`}
```
JackySun9, line 227: We don't need results-container

JackySun9 (changes requested):

libs/blocks/quiz-entry/quiz-entry.js
```diff
@@ -73,127 +116,150 @@ const App = ({
...
-      if (!mlInputUsed) setMLInputUsed(true);
-    } else {
-      setMLInputUsed(false);
-    }
-  };
+      const currentQuizState = {
+        userFlow: nextFlow,
+        userSelection: nextSelections,
+      };
+      localStorage.setItem('stored-quiz-state', JSON.stringify(currentQuizState));
```
JackySun9, line 176: do we need to consider locale like quiz-key?

libs/blocks/quiz-entry/quiz-entry.js
```diff
@@ -73,127 +116,150 @@ const App = ({
...
       onOptionClick=${onOptionClick}
       getOptionsValue=${getOptionsValue}
       mlInputUsed=${mlInputUsed}/>`}
     <div class="quiz-button-container">
         <button 
+          disabled="${!!(!mlInputUsed && !cardsUsed)}"
           aria-label="${quizLists.strings[selectedQuestion.questions].btn}" 
           class="quiz-button" 
-          onClick=${() => { onQuizButton(); }}>
+          onClick=${() => { continueQuiz(); }}>
```
JackySun9, line 243: do you want to update this like onOptionClick?

JackySun9 (changes requested):

fullcolorcoder (approved):

libs/blocks/quiz-entry/mlField.js
```diff
@@ -1,7 +1,7 @@
 import { html } from '../../deps/htm-preact.js';
 import { getConfig } from '../../utils/utils.js';
 
-export const getFiResults = async (endpoint, apiKey, input, numberOfItems, validFiCodes) => {
+export const getMLResults = async (endpoint, apiKey, threshold, input, count, validFiCodes) => {
   const { env } = getConfig();
   const subdomain = env === 'prod' ? 'cchome-dev' : 'cchome-dev';
```
fullcolorcoder, line 6: Side question, is there a prod id to add yet?

JackySun9: no, not yet

JackySun9 (approved):

libs/blocks/quiz-entry/quiz-entry.js
colloyd: No. First because the quiz block is looking for this locale storage key specifically and it doesn't benefit from locale like quiz-key does, finally, this is not displayed to the user in any way - quiz-key is seen in the query string. Second, they don't have any intentions of localizing this unless it wins in their testing. If it does win, see point one - no need to add locale to this.

libs/blocks/quiz-entry/mlField.js
colloyd: There isn't as far as I'm aware. Given that this is all specified in the questions sheet anyway, I think we don't strictly need stage vs prod - they can just update their sheets to prod when their ready to go there.

libs/blocks/quiz-entry/quiz-entry.js
colloyd: Resolving this since you agreed to remove the extraneous markup in your analytics PR.

libs/blocks/quiz-entry/quiz-entry.js
JackySun9: I mean change it to onClick=${continueQuiz}

const continueQuiz = () => () => { code }

libs/blocks/quiz-entry/mlField.js
JackySun9: currently, we only have cchome-dev, this is not able to config in excel sheet now

libs/blocks/quiz-entry/quiz-entry.js
colloyd: ....that's what it is set to.

JackySun9: ok

PR #2191, opened 2024-04-23
Sidekick publish button updates

rgclayton:
* Moved the publish button to right side of the sidekick to help avoid accidentally clicking.
* Added a warning pop up message on hover.

Resolves: [MWPW-146126](https://jira.corp.adobe.com/browse/MWPW-146126)

**Test URLs:**
- After: https://sidekick-publish-alt--milo--adobecom.hlx.page/drafts/rclayton/test-template?martech=off

image

### DIFF
--- a/libs/utils/sidekick.js
+++ b/libs/utils/sidekick.js
@@ -12,8 +12,6 @@ function stylePublish(sk) {
 
       color-scheme: light dark;
       display: flex;
-      flex: 1 1 auto;
-      justify-content: end;
       order: 100;
     }
     .publish.plugin > button {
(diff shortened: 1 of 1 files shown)


elan-tbx (approved):

elan-tbx (commented):
There's about a 3-second delay before the styles get applied to the Publish button on my end... is that to be expected?

sukamat (commented):
I see the same behavior that @elan-tbx mentioned about the delay. I also did a shift+refresh after the first load and it went back to the original sidekick layout. I did this in incognito. Not sure if it is just me.

rgclayton:
> I see the same behavior that @elan-tbx mentioned about the delay. I also did a shift+refresh after the first load and it went back to the original sidekick layout. I did this in incognito. Not sure if it is just me.

We have to wait for the sidekick to load, so there can be a slight delay due to that.

There was a recent request to only move the publish button to sit just after the tools button. Doing some testing locally and this helps the appearance delay. 

cc: @elan-tbx @sukamat

Sartxi (approved):
I don't get the delay at all, but I do have it revert back to the old layout after a page refresh.

sukamat (approved):
Yea, it seems to be much faster today :shrug:
But the revert-to-old-layout after refresh issue still exists.

PR #2193, opened 2024-04-24
Fed nav footer event

seanchoi-dev:
- New Akamai update will remove /media_ rules and QA found out fed nav and footer's images are getting 404 since they are getting loaded after media path altering function. This PR is to detect feds' loads so we know the exact time to run the function at consumer side.

image

image

Resolves: https://workfront.slack.com/archives/C05J8364J0J/p1713837998307309?thread_ts=1713558036.129549&cid=C05J8364J0J

**Test URLs:**
- After: https://fed-nav-footer-event--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/libs/blocks/global-footer/global-footer.js
+++ b/libs/blocks/global-footer/global-footer.js
@@ -117,6 +117,7 @@ class Footer {
     this.block.setAttribute('daa-lh', `gnav|${getExperienceName()}|footer${mepMartech}`);
 
     this.block.append(this.elements.footer);
+    this.block.dispatchEvent(new Event('milo:footer:loaded'));
   }, 'Failed to decorate footer content', 'errorType=error,module=global-footer');
 
   loadMenuLogic = async () => {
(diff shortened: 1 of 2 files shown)


mokimo (commented):
I see two issues with this:
1. From the gnav/footer side, some things such as menu's load even later, or even only on click. Adding this event might not work in all cases
2. From a long term perspective, you pretty much can't tell when something is loaded - u can't add an event for every block that decides it wants to load later. (Click, intersection observer, etc.)

How can it be that the homepage runs into this, but cc, dc, bacom, etc. are not?

mokimo:
My suggestion:
First: check on why other consumers are _not_ facing the same issue. 
Then: You see if there is a generic consumer side solution, such as an intersection observer checking if anything gets added to the DOM after loading, which would work for any block

overmyheadandbody (commented):
100% agree with @mokimo that this is not a scalable solution, it's unmanageable to track the loading of all blocks. Can you provide some details as to what changed for media paths at an Akamai level, requiring this update?

seanchoi-dev:
@mokimo @overmyheadandbody Thanks for the review. I am going to try to fix from consumer side.

PR #2199, opened 2024-04-24
MWPW-146878 [LocUI] - Actions polling state management

Sartxi:
When polling for status updates and a user triggers an action, the status polling interval can return and re-render according to the status conditions before new updates from the action event are completed. This causes the UI to temporarily show the wrong action buttons until the request is finished and the status API returns a new status.

In this PR I am leveraging a polling state which is toggled on/off when an event is fired so we can avoid rendering updates from the status polling service until the actions event request is completed. I have also dialed in the polling functionality to avoid setting multiple polling intervals.

Resolves: [MWPW-146878](https://jira.corp.adobe.com/browse/MWPW-146878)

### DIFF
--- a/libs/blocks/locui/actions/index.js
+++ b/libs/blocks/locui/actions/index.js
@@ -163,6 +163,7 @@ export async function syncToLangstore() {
     }, 3000);
   } else {
     await startSync();
+    getServiceUpdates();
   }
 }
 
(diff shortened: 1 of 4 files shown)


auniverseaway (approved):
I wonder if we should add a new URL to Graph's allow domains and test this. It's a pretty substantial change and I'd hate to impact folks using loc. At the very least, lets wait until tomorrow morning to merge.

PR #2204, opened 2024-04-25
Quiz entry block

JackySun9:
1. Add Analytics for Quiz Entry Block
2. Integrate Auto-Complete for Open Text Field on UAR Exposed Front Door
3. remove "results-container" from previous feedback

Resolves: 
[MWPW-147031](https://jira.corp.adobe.com/browse/MWPW-147031)
[MWPW-146028](https://jira.corp.adobe.com/browse/MWPW-146028)
[MWPW-146080](https://jira.corp.adobe.com/browse/MWPW-146080)

**Test URLs:**
- After: https://quiz-entry-block--milo--jackysun9.hlx.page/drafts/colloyd/quiz-entry/?martech=off

How to test:
1. Analytics
a) use this url: https://quiz-entry-block--milo--jackysun9.hlx.page/drafts/colloyd/quiz-entry/?martech=off&debug=quiz-entry
click different cards and check daa-ll value on continue button
type in different words and click continue button and check daa-ll value on continue button
b) use this url: https://quiz-entry-block--milo--jackysun9.hlx.page/drafts/colloyd/quiz-entry/?martech=off
use cards or mlfield to get through to result page and check pageloadHash in localStorage quiz key

2. Autocomplete
use this url: https://quiz-entry-block--milo--jackysun9.hlx.page/drafts/colloyd/quiz-entry/?martech=off
type any words in mlField, you will see autocomplete

Feature work: need address some css issues in another story

### DIFF
--- a/libs/blocks/quiz-entry/quiz-entry.js
+++ b/libs/blocks/quiz-entry/quiz-entry.js
@@ -1,6 +1,6 @@
 import { render, html, useEffect, useState } from '../../deps/htm-preact.js';
 import { createTag } from '../../utils/utils.js';
-import { getQuizEntryData, getQuizJson } from './utils.js';
+import { getQuizEntryData } from './utils.js';
 import { mlField, getFiResults } from './mlField.js';
 import { GetQuizOption } from './quizoption.js';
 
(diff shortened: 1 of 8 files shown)


libs/blocks/quiz-entry/quiz-entry.js
```diff
@@ -13,7 +14,7 @@ const App = ({
   debug = false,
 }) => {
   const [dataLoaded, setDataLoaded] = useState(false);
-  const [quizState, setQuizState] = useState({ userFlow: [], userSelection: [] });
+  const [quizState, setQuizState] = useState({ userFlow: [], userSelection: [], isML: false });
```
colloyd, line 17: What is the purpose of isML here? Can this be inferred from userSelection by the presence of ficodes, rather than standard card responses intead? It looks like you're carrying this all the way over to handleResultFlow in quiz/utils to determine if the analytics should include the "interest". 

Or better, can we just add this isML bit to userSelection, so we're adding less code elsewhere? you could drop all of the added ML stuff by just adding it to userSelection instead if mlInputUsed is true. userSelection is already passed to handleResultFlow so it'd just be a matter of intercepting it there. 

userSelection is made up of two objects, selectedQuestion, and selectedCards. Adding an isML object to userSelection should make it easier to get the information to handleResultFlow, without the need to include it separately in quizState.

libs/blocks/quiz-entry/quiz-entry.js
```diff
@@ -199,18 +211,43 @@ const App = ({
...
+      setSuggestions([]);
+      setShowPopover(false);
     }
   };
 
   const onMLEnter = (event) => {
     if (event.keyCode === enterKeyCode) continueQuiz();
   };
 
+  const onSuggestionClick = (suggestion) => () => {
```
colloyd, line 237: Can we add a way to put focus back into the input field after a suggestion is selected? without restoring focus, onMLEnter doesn't work - you can't just select from autocomplete and press enter to continue

libs/blocks/quiz-entry/utils.js
```diff
@@ -1,5 +1,8 @@
 import { getNormalizedMetadata } from '../quiz/utils.js';
 
+let analyticsType;
```
colloyd, line 3: These values should be handled like all the rest of these, a const in the getQuizEntryData function that is returned. Local variables in utils is the primary reason why I couldn't use more from the quiz block. 

I have given preference to leaving functions that are using app properties (like getAnalyticsDataForBtn) inside of the app, rather than move them to utils where you would need to use local variables like this. Using "export const getAnalyticsDataForBtn" won't actually be useful in other blocks outside of quiz-entry, just the same as the issues that I ran into in trying to use more from quiz/utils.

colloyd (changes requested):

colloyd (approved):
Looks great now, good work!

PR #2187, opened 2024-04-23
MWPW-146001 parallelize literals call

npeltier:
Before
image
After
image

Resolves: [MWPW-146001](https://jira.corp.adobe.com/browse/MWPW-146001)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/?martech=off
- After: https://mwpw-146001--milo--npeltier.hlx.page/?martech=off

- Before: https://main--cc--adobecom.hlx.live/products/substance3d/apps/designer?martech=off
- After: https://main--cc--adobecom.hlx.live/products/substance3d/apps/designer?milolibs=mwpw-146001--milo--npeltier&martech=off

Page | LHS Before | LHS After | FCP | TBT
-- | -- | -- | -- | --
https://main–cc–adobecom.hlx.live/products/substance3d/apps/designer ?martech=off | 86 | 89 | -0,6 | -50
https://main--cc--adobecom.hlx.live/products/substance3d/apps/modeler?martech=off | 87 | 90 | -0,7 | 0
https://main--cc--adobecom.hlx.live/products/substance3d/apps/painter?martech=off | 87 | 93 | -0,5 | -10
https://main--cc--adobecom.hlx.live/products/substance3d/apps/sampler?martech=off | 88 | 94 | -0,6 | 30
https://main--cc--adobecom.hlx.live/products/substance3d/apps/stager?martech=off | 89 | 94 | -0,6 | 0
https://main--cc--adobecom.hlx.live/products/substance3d/3d-assets?martech=off | 84 | 86 | -0,2 | -90

https://git.corp.adobe.com/wcms/tacocat.js/pull/570
https://git.corp.adobe.com/wcms/tacocat.js/pull/573

PR #2212, opened 2024-04-26
MWPW-147239 use federal for checkout-links too

npeltier:
will send numbers & more details when related PR https://github.com/adobecom/milo/pull/2187 is merged 

Resolves: [MWPW-147239](https://jira.corp.adobe.com/browse/MWPW-147239)

**Test URLs:**
- After: https://MWPW-147239--milo--npeltier.hlx.live/?martech=off

### DIFF
--- a/libs/blocks/merch/merch.js
+++ b/libs/blocks/merch/merch.js
@@ -4,7 +4,7 @@ import {
 import { replaceKey } from '../../features/placeholders.js';
 
 export const PRICE_LITERALS_URL = 'https://milo.adobe.com/libs/commerce/price-literals.json';
-export const CHECKOUT_LINK_CONFIG_PATH = '/commerce/checkout-link.json'; // relative to libs.
+export const CHECKOUT_LINK_URL = 'https://www.adobe.com/federal/commerce/checkout-link.json'; // relative to libs.
 
 export const PRICE_TEMPLATE_DISCOUNT = 'discount';
 export const PRICE_TEMPLATE_OPTICAL = 'optical';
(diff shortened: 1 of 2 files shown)


mokimo (approved):
Coolio

npeltier:
well not so cool @mokimo :) 
i think it's better using configuration, but i can't do that until we fix https://github.com/orgs/adobecom/discussions/2220 
ideal would be here that default configuration points to federal and i let code exactly like this.

what is clear is that PR will not bring any perf gain (actually perf will be slightly worse on lower env due to DNS fetch).

libs/blocks/merch/merch.js
```diff
@@ -4,7 +4,7 @@ import {
 import { replaceKey } from '../../features/placeholders.js';
 
 export const PRICE_LITERALS_URL = 'https://milo.adobe.com/libs/commerce/price-literals.json';
-export const CHECKOUT_LINK_CONFIG_PATH = '/commerce/checkout-link.json'; // relative to libs.
+export const CHECKOUT_LINK_URL = 'https://www.adobe.com/federal/commerce/checkout-link.json'; // relative to libs.
```
3ch023, line 7: as GWP, how do I validate my change before publishing it live?
should we use https://www.stage.adobe.com/federal/commerce/checkout-link.json?
Does it point to hlx.page content?
cc @mokimo

3ch023 (changes requested):
the current change takes away any ability to test changes to the file.
imo before merging we should address this and make checkout-links changes testable by GWP content QA before they publish their changes

libs/blocks/merch/merch.js
mokimo: no, stage.adobe.com is not yet connected to stage/federal/hlx.page. I'd recommend to use stage--federal--adobecom.hlx.page/commerce/checkout-link.json
https://github.com/orgs/adobecom/discussions/2220#discussioncomment-9260878 

until we properly set up stage as part of the helix-5 migration

npeltier:
closing for now

PR #2225, opened 2024-04-29
MWPW-147455: Prevent MEP errors from breaking the page

3ch023:
Prevent MEP errors from breaking the page

Resolves: [MWPW-147455](https://jira.corp.adobe.com/browse/MWPW-147455)

Broken page: https://main--cc--adobecom.hlx.page/products/illustrator
Fixed page with a 'warn' message: https://main--cc--adobecom.hlx.page/products/illustrator?milolibs=mepmuteerror

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/products/illustrator
- After: https://main--cc--adobecom.hlx.page/products/illustrator?milolibs=mepmuteerror
- After: https://mepmuteerror--milo--adobecom.hlx.page/drafts/mariia/promo/replace/page

### DIFF
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -541,9 +541,10 @@ const createDefaultExperiment = (manifest) => ({
   disabled: manifest.disabled,
   event: manifest.event,
   manifest: manifest.manifestPath,
-  variantNames: ['all'],
-  selectedVariantName: 'default',
   selectedVariant: { commands: [], fragments: [] },
+  selectedVariantName: 'default',
+  variantNames: ['all'],
+  variants: {},
 });
 
 export async function getPersConfig(info, override = false) {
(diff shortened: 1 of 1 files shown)


mokimo (approved):
This will make it a little less obvious when MEP is misconfigured
But fine to me 👍

libs/features/personalization/personalization.js
```diff
@@ -751,43 +751,47 @@ export function handleFragmentCommand(command, a) {
...
+      const val = r.experiment.selectedVariantName.replace(TARGET_EXP_PREFIX, '').trim().slice(0, 15);
+      return val === 'default' ? 'nopzn' : val;
+    });
+    const pznManifests = pznList.map((r) => {
+      const val = r.experiment?.manifestOverrideName || r.experiment?.manifest;
+      return getFileName(val).replace('.json', '').trim().slice(0, 15);
+    });
+    config.mep.martech = `|${pznVariants.join('--')}|${pznManifests.join('--')}`;
+  } catch (e) {
+    console.warn(e);
```
mokimo, line 795: We should add a lana log in this case

overmyheadandbody (commented):
I'd definitely like @vgoodric's feedback here, she may be able to think of reasons why we want such failures to be more visible. And we should log such occurrences to LANA, yes.

3ch023:
> I'd definitely like @vgoodric's feedback here, she may be able to think of reasons why we want such failures to be more visible. And we should log such occurrences to LANA, yes.

yep, am also waiting for her feedback.
but to be honest i think there can't be a reason when we want to see the broken page. broken personalization or broken promo is of a lower severity than a blank Illustrator page. 

regarding LANA, I can add it, yes, but what bothers me is that we don't monitor those messages in splunk. 
for commerce, we have an alert setup, but not for mep afaik.

vgoodric:
so I'm in meetings, but this is my first stab at it: https://main--cc--adobecom.hlx.page/products/illustrator?milolibs=mepfreshmanifestbug
https://github.com/adobecom/milo/compare/mepfreshmanifestbug?expand=1

vgoodric:
ok, so it looks like my initial stab at it still has issues and Mariia's is working.  So since I have meetings non-stop meetings today, I think we should add this try catch around the issue and I'll try to dig in more tomorrow to see if there is a better way to address the root cause.

But the bug is blocking GWP.

vgoodric (approved):

PR #2227, opened 2024-04-29
MWPW-144022: Quiz entry block

fullcolorcoder:
Adds carousel functionality to quiz entry block

Resolves: [MWPW-144022](https://jira.corp.adobe.com/browse/MWPW-144022)

Todo: Style refinement and test coverage will follow this PR. 

**Test URLs:**
- Before: https://quiz-entry-block--milo--adobecom.hlx.page/drafts/colloyd/quiz-entry/
- After: [https://<branch>--milo--adobecom.hlx.page/?martech=off](https://quiz-entry-block--milo--fullcolorcoder.hlx.page/drafts/colloyd/quiz-entry/?martech=off)

### DIFF
--- a/libs/blocks/quiz-entry/quiz-entry.css
+++ b/libs/blocks/quiz-entry/quiz-entry.css
@@ -85,6 +85,10 @@ input:disabled {
   justify-content: center;
 }
 
+.carousel-slides {
+  display: flex;
+}
+
 .quiz-option {
   box-shadow: 10px 10px 6px 0 #00000029;
   flex-direction: column;
--- a/libs/blocks/quiz-entry/quiz-entry.css
+++ b/libs/blocks/quiz-entry/quiz-entry.css
@@ -390,3 +390,48 @@ html[dir="rtl"] .quiz-option-icon {
     width: 360px;
   }   
 }
+
+
+.carousel-arrow-next {
+  background-image: url('https://main--milo--adobecom.hlx.page/libs/blocks/carousel/img/arrow.svg');
+  background-size: 15px; 
+  background-position: 50% 50%; 
+  background-repeat: no-repeat;
+  border: none;
+  background-color: #fff;
+  height: 63px; 
+  width: 63px;  
+  border-radius: 50%;
+  box-shadow: 10px 10px 15px -5px rgba(0,0,0,0.23);
+  position: relative;
+  top: 90px;
+  left: -30px;
+  cursor: pointer;
+}
+
+.carousel-arrow-prev {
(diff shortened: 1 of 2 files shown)


JackySun9:
Fix eslint?

colloyd:
A couple of things I noticed:

- The tabbing feels strange when going through the cards. On desktop, the 3d slide is not visible (as expected) when tabbing through the cards, it moves from pdf to the next arrow, but then I am unable to get to 3d without shift tabbing or tabbing through the whole set of cards. I wonder if we need to have some kind of focus reset or something after arrows are activated with keyboard nav or something? something to make it easier to get to the newly revealed cards.
- I'm seeing some odd breaking of the arrows at certain device widths like this:

quiz-entry-block--milo--fullcolorcoder hlx page_drafts_colloyd_quiz-entry__martech=off

colloyd (changes requested):
See my previous comments about tabbing and visual breaking of carousel arrows.

fullcolorcoder:
@colloyd 

> A couple of things I noticed:
> 
> * The tabbing feels strange when going through the cards. On desktop, the 3d slide is not visible (as expected) when tabbing through the cards, it moves from pdf to the next arrow, but then I am unable to get to 3d without shift tabbing or tabbing through the whole set of cards. I wonder if we need to have some kind of focus reset or something after arrows are activated with keyboard nav or something? something to make it easier to get to the newly revealed cards. EDIT: It does feel reasonable in the tabbing if you navigate with the arrow keys to cards that aren't visible. I wonder if we ought to consider making it so the arrows aren't tab-able? Did MJ from the accessibility team have a look at the finished experience yet?

You can still access the right/left arrows in the situation you describe, and the tab order is generally not shabby. Its a bit more straightforward than the original regarding tabbing, and still remains consistent with the experience. Check out the tabbing experience on the last carousel of this page for example: https://main--milo--adobecom.hlx.page/docs/library/kitchen-sink/carousel 

Resetting the index seems like a risky move regarding user experience / accessibility standards.

MJ & team has not finished reviewing. I've let them know for now I'll go with the default Milo core interactions to keep experiences consistent across pages until everything is worked out for alignment. I think these are small things we can iterate on.

> * I'm seeing some odd breaking of the arrows at certain device widths like this:

This does need a bit more CSS follow up still as noted:

> Todo: Style refinement and test coverage will follow this PR.

I'll pick up on the testing bits next for the block.

JackySun9:
I tested it, it didn't move like a carousel, will this be fixed by css refinement?

colloyd (approved):

PR #2244, opened 2024-05-01
MWPW-147594: Consumers mappings for origin not being sent when using sidekick

sanrai:
If an author tries sending content to caas from a consumer of milo, the origin was always upper case so the mapping never matched and an uppercase origin would always be sent. 

Resolves: [MWPW-147594](https://jira.corp.adobe.com/browse/MWPW-147594)

**Test URLs:**
- Before:
https://mwpw-147594--milo--adobecom.hlx.page/drafts/sanrai/foo?martech=off

- After:


jedjedjedM (approved):

cmiqueo (approved):

gmirijan (approved):

mokimo (approved):

sanrai:
No unit tests exist for the caas bulk publisher. Fixing GH check would require writing unit tests for entire file.

mokimo:
As discussed, in the future please exclude codecov https://github.com/orgs/adobecom/discussions/2206#discussioncomment-9429454

PR #2248, opened 2024-05-01
MWPW-147599: CaaS Bulk Publisher not working after MS authentication change

sanrai:
After some sharepoint change, we are not able to fetch `.page` pages from the caas bulk publisher, parse and send them.

Instead of trying to authenticate, we will now just pull from the `.live` pages. If authors don't want that content to go out, they can have CDN rules to deny list certain paths until a given date.

This should be logically sound as the `uuids` will use the final prod urls to be generated so there should be no conflicts with current authoring.

Resolves: [MWPW-147599](https://jira.corp.adobe.com/browse/MWPW-147599)

**Test URLs:**
- Before:
https://MWPW-147599--milo--adobecom.hlx.page/drafts/sanrai/foo?martech=off

- After:


sheridansunier (approved):

sanrai:
For some reason the aem-psi-check is not updating to accept the test url put in the PR description.

PR #2251, opened 2024-05-02
MWPW-147472: auto adjust commerce iframe

yesil:
of the managed checkout-link modals.

Resolves: [MWPW-147472](https://jira.corp.adobe.com/browse/MWPW-147472)

To test, click on the TWP and D2P modal CTAs at the after URL:
image

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ilyas/checkout-links?martech=off&georouting=off
- After: https://mwpw-147472--milo--yesil.hlx.page/drafts/ilyas/checkout-links?martech=off&georouting=off

In order to test the page at the after URL, use the following Modheader config:

image

You can import the following JSON as profile as well:

```javascript
[
  {
    "alwaysOn": false,
    "cspHeaders": [
      {
        "appendMode": false,
        "enabled": true,
        "name": "frame-ancestors",
        "value": "'self' https://*.adobe.com https://*.hlx.page https://*.hlx.live http://localhost:3000"
      }
    ],
    "respHeaders": [
      {
        "appendMode": false,
        "enabled": true,
        "name": "Access-Control-Allow-Origin",
        "value": "*"
      },
      {
        "appendMode": false,
        "enabled": true,
        "name": "Access-Control-Allow-Headers",
        "value": "x-api-key, Authorization"
      }
    ],
    "shortTitle": "1",
    "title": "WCS",
    "urlFilters": [
      {
        "enabled": true,
        "urlRegex": "https://wcs.adobe.com/*"
      },
      {
        "enabled": true,
        "urlRegex": "https://www.stage.adobe.com/aos-api/users/*"
      },
      {
        "enabled": true,
        "methods": [
          "GET"
        ],
        "urlRegex": "https://www.adobe.com/"
      }
    ],
    "version": 2
  }
]
```


mirafedas (approved):

3ch023 (approved):

libs/blocks/merch/merch.js
```diff
@@ -152,8 +152,8 @@ export async function fetchLiterals(url) {
 
 export async function fetchCheckoutLinkConfigs(base = '') {
   fetchCheckoutLinkConfigs.promise = fetchCheckoutLinkConfigs.promise
-    ?? fetch(`${base}${CHECKOUT_LINK_CONFIG_PATH}`).catch(() => {
-      log?.error('Failed to fetch checkout link configs');
+    ?? fetch(`${base}${CHECKOUT_LINK_CONFIG_PATH}`).catch((e) => {
+      log?.error('Failed to fetch checkout link configs', e);
```
mokimo, line 156: Should you add a lana log here?

mokimo (approved):

libs/blocks/merch/merch.js
yesil: it does to lana automatically behind the scene :)

Roycethan:
@yesil cc @mirafedas  Few issues observed during the checks:
https://main--cc--adobecom.hlx.live/products/catalog?martech=off&milolibs=mwpw-147472--milo--yesil
1) In Mac-Safari the twp modal is rendering blank when clicked on (free-trial) CTA in Card collection (say photoshop)
image

2) I see in the sample test the Modal Hash id is not getting appended - is this intended change, if so one thing to note is while navigationg back from checkout or other pages the modal will not be in open position, due to absense of #modal_hash_id

yesil:
@Roycethan  

1. I wasn't able to reproduce it on Safari. 
2. This is expected. These are managed modals, and do not require to set a hash as they are programatically initialized by the merch logic and not the generic modal logic.
3. Indeed this is not something new.  This issue happens because as per https://web.dev/articles/bfcache#observe iframe are not compatible with the caching optimisation for the page transition (another reason to not to use iframes).
So I implemented a recommended solution, where when clicked on back, if there is a commerce modal with an iframe, we will reload the page.

Roycethan:
> @Roycethan
> 
> 1. I wasn't able to reproduce it on Safari.
> 2. This is expected. These are managed modals, and do not require to set a hash as they are programatically initialized by the merch logic and not the generic modal logic.
> 3. Indeed this is not something new.  This issue happens because as per https://web.dev/articles/bfcache#observe iframe are not compatible with the caching optimisation for the page transition (another reason to not to use iframes).
>    So I implemented a recommended solution, where when clicked on back, if there is a commerce modal with an iframe, we will reload the page.

@yesil  #1 is bit tricky to reproduce and not specific to safari, it happens when i moved to #page2 by say by clicking show more button or scrolling to bottom, it adds up "#page2" to url and in this situation when clicked on free trial CTA to launch modal , in desktop , the modal content is blocked

Repro:
Click on Free-trial CTA of any card in Desktop view to launch I-frame- Not specific to Safari it can be reproe'd in any browser
https://main--cc--adobecom.hlx.live/products/catalog?martech=off&milolibs=mwpw-147472--milo--yesil#page=2

image

yesil:
Thanks @Roycethan.  Actually this wasn't a new issue but was easier to hit on catalog page due to deeplinking used on the catalog page.
Nonetheless, I fixed the issue and added a test case.

Roycethan:
@yesil 
#1) Post fix, Still reproducible in Mac-Safari, although now on firefox,chrome works fine.
https://main--cc--adobecom.hlx.live/products/catalog?martech=off&milolibs=mwpw-147472--milo--yesil#page=2
image
Plz check thanks

#2) As you suggested its by design, modal hash won't be retained for re-visits

#3) Post fix the modal is not retained so no more freeze issue, it behaves like #2

yesil:
@Roycethan might this be due to a cache issue? I cannot reproduce it with Safari.
If you see a blank screen, please share what you see in the console.

Roycethan:
> https://main--cc--adobecom.hlx.live/products/catalog?martech=off&milolibs=mwpw-147472--milo--yesil#page=2

@yesil I didn't see any errors logged but looks like this causing only for TWP modal (Free-trial) and not for buy now modal on Mac-Safari, on digging more found out that this is casing when theres 'height-fit-content' style in below:
class="dialog-modal commerce-frame height-fit-content twp

yesil:
@Roycethan `height-fit-content` is automatically added by the code.
are you able to share step-by-step instructions to reproduce it?

Roycethan:
> https://main--cc--adobecom.hlx.live/products/catalog?martech=off&milolibs=mwpw-147472--milo--yesil#page=2

Ok @yesil

mirafedas:
@Roycethan thank you for discovering this. Actually, this issue is reproducible on the main branch as well, most likely it was introduced some time ago when I was implementing the automatic height adjustment for TwP modals, but it is intermittent so we didn't spot it then.
What happens: in the CORS situation we send the postMessage request from within the iframe to retrieve the viewport width, and in Safari it throws an error about the invalid string pattern. From what I see, this issue should only be present in lower environments (and in general only in cases when the page domain does not match the domain of the page inside the iframe), only in Safari, and only on the first page load (when the page is loaded from cache).
Could you open an issue in Jira for this and assign it to me? It would be helpful if you could check if this issue happens only in some versions of Safari, or in all, and if it works in other browsers.

As this issue is not caused by changes in this PR, I believe we can proceed with verification and merge it.
cc @yesil @afmicka

Roycethan:
> Repro steps:
> 
> 1. Launch https://main--cc--adobecom.hlx.live/products/catalog?martech=off&milolibs=mwpw-147472--milo--yesil on Mac-Safari browser
> 2. Click on Free-trial CTA on CC-ALL_APPS card and
> 3. Observe whether the content of i-frame loads
>    Observation: Content is not loading and blank - explained the issue and causing element in the video attached as well:

@mirafedas  I'm consistently able to reproduce on certain resolutins say you can reproduce in browser-stack for Mac Safari Sonoma 1512 * 751 , For the same resolution it was not reproduced on Main catalog branch, hence i did not marked verified earlier
Anyway captured this here https://jira.corp.adobe.com/browse/MWPW-148283 and marking this PR as verified as you requested.
cc @yesil  @afmicka

PR #2281, opened 2024-05-09
MWPW-147175 Support Extending window.adobeid

Blainegunn:
Support an optional property `adobeid` in CONFIG. It is used for overriding or extending `window.adobeid`.
 
We need these properties for some updates we're making over on [DC](https://github.com/adobecom/dc).
[IMS Lib Docs](https://git.corp.adobe.com/pages/IMS/imslib.js/adobeid.html)

Resolves: [MWPW-147175](https://jira.corp.adobe.com/browse/MWPW-147175)

**Test URLs:**
- After: https://mwpw-147175--milo--adobecom.hlx.live/?martech=off

DC
- Before: https://mwpw-147175--dc--adobecom.hlx.live/acrobat/online/word-to-pdf?martech=off
- After: https://mwpw-147175--dc--adobecom.hlx.live/acrobat/online/word-to-pdf?milolibs=mwpw-147175&martech=off


libs/utils/utils.js
```diff
@@ -810,6 +812,10 @@ export async function loadIms() {
         clearTimeout(timeout);
       },
       onError: reject,
+      api_parameters: apiParams,
+      onAccessTokenHasExpired: accessTokenExpired,
+      onReauthAccessToken: reauthAccessToken,
+      onAccessToken: accessToken,
```
overmyheadandbody, line 818: You mentioned these are needed for DC, so I'm not convinced everyone needs to have access to them through utils. Have you considered a sort of additional IMS config that a consumer could set? As in send in an object with extra fields, iterate trough them and attach to the universal IMS config.

overmyheadandbody:
Additionally, wouldn't hurt to have a more descriptive title to be easy to find in the future if need be :)

libs/utils/utils.js
mokimo: I agree with @overmyheadandbody, sending an object and passing everything on - also makes it easier in case any consumer wants some different configs, or adds more configs in the future.

TsayAdobe (approved):

libs/utils/utils.js
TsayAdobe: Updated to use an `adobeid` object

mokimo (approved):

mokimo:
@Blainegunn @TsayAdobe make sure to add a proper testable url for the PSI check to pass. This could just be the milo homepage, so that the automated process can merge this PR.

PR #2290, opened 2024-05-10
MWPW-148050: auto merge stage to main

mokimo:
### Description
Merge PRs from stage->main automatically given 4 sign-offs, as long as it's monday-thursday 8AM UTC to 8PM UTC

Resolves: [MWPW-148050](https://jira.corp.adobe.com/browse/MWPW-148050)

**Test URLs:**
- After: https://merge-process--milo--mokimo.hlx.live/?martech=off

### DIFF
--- a/.github/workflows/helpers.js
+++ b/.github/workflows/helpers.js
@@ -3,7 +3,7 @@
 const owner = process.env.REPO_OWNER || ''; // example owner: adobecom
 const repo = process.env.REPO_NAME || ''; // example repo name: milo
 const auth = process.env.GH_TOKEN || ''; // https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens
-
+const CURRENT_YEAR = 2024;
 const RCPDates = [
   {
     start: new Date('2024-05-26T00:00:00-07:00'),
@@ -45,9 +45,8 @@ const RCPDates = [
 
 const isWithinRCP = () => {
   const now = new Date();
-  let currentYear = 2024;
-  if (now.getFullYear() !== currentYear) {
-    console.log(`ADD NEW RCPs for ${currentYear + 1}`);
+  if (now.getFullYear() !== CURRENT_YEAR) {
+    console.log(`ADD NEW RCPs for ${CURRENT_YEAR + 1}`);
     return true;
   }
 
@@ -69,7 +68,12 @@ Then run: node --env-file=.env .github/workflows/update-ims.js`);
 
   const { Octokit } = require('@octokit/rest');
   return {
-    github: { rest: new Octokit({ auth }) },
+    github: {
+      rest: new Octokit({ auth }),
+      repos: {
+        createDispatchEvent: () => console.log('local mock createDispatch'),
+      },
+    },
     context: {
       repo: {
         owner,
--- a/.github/workflows/helpers.js
+++ b/.github/workflows/helpers.js
@@ -50,11 +50,9 @@ const isWithinRCP = () => {
     return true;
   }
 
-  for (const { start, end } of RCPDates) {
-    if (start <= now && now <= end) {
-      console.log('Current date is within a RCP. Stopping execution.');
-      return true;
-    }
+  if (RCPDates.some(({ start, end }) => start <= now && now <= end)) {
+    console.log('Current date is within a RCP. Stopping execution.');
+    return true;
   }
 
   return false;
--- a/.github/workflows/helpers.js
+++ b/.github/workflows/helpers.js
@@ -4,6 +4,63 @@ const owner = process.env.REPO_OWNER || ''; // example owner: adobecom
 const repo = process.env.REPO_NAME || ''; // example repo name: milo
 const auth = process.env.GH_TOKEN || ''; // https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens
 
+const RCPDates = [
+  {
+    start: new Date('2024-05-26T00:00:00-07:00'),
+    end: new Date('2024-06-01T00:00:00-07:00'),
+  },
+  {
+    start: new Date('2024-06-13T11:00:00-07:00'),
+    end: new Date('2024-06-13T14:00:00-07:00'),
+  },
+  {
+    start: new Date('2024-06-30T00:00:00-07:00'),
+    end: new Date('2024-07-06T00:00:00-07:00'),
+  },
+  {
+    start: new Date('2024-08-25T00:00:00-07:00'),
+    end: new Date('2024-08-31T00:00:00-07:00'),
+  },
+  {
+    start: new Date('2024-09-12T11:00:00-07:00'),
+    end: new Date('2024-09-12T14:00:00-07:00'),
+  },
+  {
+    start: new Date('2024-10-14T00:00:00-07:00'),
+    end: new Date('2024-11-18T17:00:00-08:00'),
+  },
+  {
+    start: new Date('2024-11-17T00:00:00-08:00'),
+    end: new Date('2024-11-30T00:00:00-08:00'),
+  },
+  {
+    start: new Date('2024-12-12T11:00:00-08:00'),
+    end: new Date('2024-12-12T14:00:00-08:00'),
+  },
+  {
+    start: new Date('2024-12-15T00:00:00-08:00'),
+    end: new Date('2025-01-02T00:00:00-08:00'),
+  },
+];
+
+const isWithinRCP = () => {
+  const now = new Date();
+  let currentYear = 2024;
+  if (now.getFullYear() !== currentYear) {
+    console.log(`ADD NEW RCPs for ${currentYear + 1}`);
+    return true;
+  }
+
+  for (const { start, end } of RCPDates) {
+    if (start <= now && now <= end) {
+      console.log('Current date is within a RCP. Stopping execution.');
+      return true;
+    }
+  }
+
+  return false;
+};
+
 const getLocalConfigs = () => {
   if (!owner || !repo || !auth) {
     throw new Error(`Create a .env file on the root of the project with credentials.
@@ -83,4 +140,5 @@ module.exports = {
   getLocalConfigs,
   slackNotification,
   pulls: { addLabels, addFiles, getChecks, getReviews },
+  isWithinRCP,
 };
--- a/.github/workflows/mark-stale-prs.yaml
+++ b/.github/workflows/mark-stale-prs.yaml
@@ -1,7 +1,7 @@
-name: "Close stale pull requests"
+name: 'Close stale pull requests'
 on:
   schedule:
-    - cron: "0 0 * * *"
+    - cron: '0 0 * * *'
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/mark-stale-prs.yaml
+++ b/.github/workflows/mark-stale-prs.yaml
@@ -1,4 +1,4 @@
-name: 'Close stale pull requests'
+name: Close stale pull requests
 on:
   schedule:
     - cron: '0 0 * * *'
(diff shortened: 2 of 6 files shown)


overmyheadandbody (commented):
Are we ready to merge this in? I was under the impression that we give the feature branch to stage merge a few weeks to figure out and address any issues and only then would we feel confident to do automatic merges to main. What timeline do you have in mind for this?

mokimo:
@overmyheadandbody No specific timeline, I figured I can open the PR at any point, so this doesn't get lost and gives people the chance to give input.

Since this is the 2nd week with the stage automation, unless we see any big issues this should be good to be ready next week. This would also nicely fall in place with the milo deep dives on the 29th of May, where we can present all the changes to the process as a whole.

.github/workflows/merge-to-main.js
```diff
@@ -0,0 +1,70 @@
...
+    .then((pr) => pr && addLabels({ pr, github, owner, repo }));
+
+const workingHours = () => {
+  const now = new Date();
+  const day = now.getUTCDay();
+  const hour = now.getUTCHours();
+  const isSunday = day === 0;
+  const isSaturday = day === 6;
+  const isFriday = day === 5;
+  return hour >= 8 && hour <= 20 && !isFriday && !isSaturday && !isSunday;
```
robert-bogos, line 28: I guess we are caching `isSunday`, `isSaturday` and `isFriday` just for readability. Otherwise we could do 
```suggestion
  return hour >= 8 && hour <= 20 && day !== 5 && day !== 6 && day !== 0;
```

robert-bogos (approved):

.github/workflows/merge-to-main.js
mokimo: I agree with the suggestion & had it like that before, but `0, 5, 6` is not immediately obvious compared to its not friday/saturday/sunday - hence you are spot on, it's a readability thing!

.github/workflows/merge-to-main.js
```diff
@@ -0,0 +1,70 @@
...
+        owner,
+        repo,
+        pull_number: stageToMainPR.number,
+        merge_method: 'merge',
+      });
+      await slackNotification(
+        `:rocket: Production release <${stageToMainPR.html_url}|${stageToMainPR.number}>`
+      );
+    }
+
```
mokimo, line 55: We need to trigger the stage->main PR workflow after a merge

.github/workflows/merge-to-main.js
```diff
@@ -0,0 +1,70 @@
+const {
```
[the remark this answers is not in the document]
mokimo: Sent out the communication, if you didn't see it, let me know - then we have a big gap in our communication strategy, I've cross posted it a lot.

.github/workflows/merge-to-main.js
```diff
@@ -0,0 +1,70 @@
...
+  owner = params.context.repo.owner;
+  repo = params.context.repo.repo;
+
+  if (isWithinRCP()) return console.log('Stopped, within RCP period.');
+  if (!workingHours()) return console.log('Stopped, outside working hours.');
+
+  try {
+    const stageToMainPR = await getStageToMainPR();
+    const signOffs = stageToMainPR?.labels.filter((l) => l.includes('SOT'));
+    console.log(`Currently ${signOffs.length} sign offs on the PR`);
```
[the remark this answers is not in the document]
mokimo: In case of CSOs that's precisely what we don't want. We can adjust this if it doesn't work in its current form IMO.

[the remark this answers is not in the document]
narcis-radu: @mokimo - can you elaborate on that? Generally we're looking for sign offs from SOTs, but in the case of CSOs we will mostly do manual merges to speed up the fix.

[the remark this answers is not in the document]
mokimo: > CSO
Revert stage batch. Go on the merged PRs and click the "revert" button. (Notify folks to re-open their PRs afterwards)
Create PR to merge into stage with the fix or production-PR-revert
Get 2 approvals on it (or an admin can insta merge after tests are green)
Try to get the fix tested, but in a CSO case it's fine to simply apply the 4 SOT labels and merge into main.

As per the opening post in https://github.com/orgs/adobecom/discussions/2206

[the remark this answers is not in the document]
mokimo: With the new process, after you get approvals on a revert PR - merged it to stage - and then just apply the 4 SOT labels yourself, that enables everybody to step in to get the fix merged.

.github/workflows/helpers.js
```diff
@@ -4,6 +4,63 @@ const owner = process.env.REPO_OWNER || ''; // example owner: adobecom
...
+  },
+  {
+    start: new Date('2024-12-15T00:00:00-08:00'),
+    end: new Date('2025-01-02T00:00:00-08:00'),
+  },
+];
+
+const isWithinRCP = () => {
+  const now = new Date();
+  let currentYear = 2024;
```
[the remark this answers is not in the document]
narcis-radu: @mokimo - what do you think about this? it's an easy change.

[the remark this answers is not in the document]
mokimo: Yup will change that!

.github/workflows/merge-to-main.js
```diff
@@ -0,0 +1,70 @@
...
+    if (signOffs.length >= 4) {
+      console.log('Stage to Main PR has all required labels. Merging...');
+      await github.rest.pulls.merge({
+        owner,
+        repo,
+        pull_number: stageToMainPR.number,
+        merge_method: 'merge',
+      });
```
mokimo, line 50: We will need to merge main into stage, after merging stage into main. Otherwise, our stage branch is always going to be outdated which might lead to issues trying to automatically merge a PR unless we force merge it. We don't want to force merge to avoid any potential issues.

Stage:
Screenshot 2024-05-17 at 13 44 40

Main:
Screenshot 2024-05-17 at 13 42 10

Main receives an extra commit `e6d5f7c` due to the merge-commit. We will need to include `e6d5f7c` in stage as well, right after merging into main to get the two histories aligned.

Now as we'd merge `e6d5f7c` into stage, that would also always create a new commit, but this means stage is ahead of main, which is the normal case - so we'd still end up with an unnecessary commit but it's the lesser evil.

.github/workflows/merge-to-main.js
mokimo: Done

.github/workflows/helpers.js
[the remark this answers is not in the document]
mokimo: Done

.github/workflows/merge-to-main.js
mokimo: Testing if that's really needed or an outdated branch doesn't stop the integration from merging.
I'd like to not back-merge if we can avoid it.

mokimo:
https://github.com/adobecom/milo/pull/2362
image

Permissions are working, SOT labels are accounted for, this should be good to go

.github/workflows/merge-to-main.js
mokimo: Not needed. We can live with an outdated branch.

.github/workflows/helpers.js
```diff
@@ -3,6 +3,62 @@
...
+  }
+
+  for (const { start, end } of RCPDates) {
+    if (start <= now && now <= end) {
+      console.log('Current date is within a RCP. Stopping execution.');
+      return true;
+    }
+  }
+
+  return false;
```
robert-bogos, line 60: Nit: this can be written shorter
```suggestion
  return RCPDates.some(({ start, end }) => start <= now && now <= end)
    ? (console.log('Current date is within a RCP. Stopping execution.'), true) : false;
```
Feel free to not make this change if you don't want to. It's less code, but might be less readable as it uses the less common [JS comma operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comma_operator)

robert-bogos (approved):
Nice work, looks good from my side 👍

.github/workflows/helpers.js
```diff
@@ -12,7 +68,12 @@ Then run: node --env-file=.env .github/workflows/update-ims.js`);
 
   const { Octokit } = require('@octokit/rest');
   return {
-    github: { rest: new Octokit({ auth }) },
+    github: {
+      rest: new Octokit({ auth }),
+      repos: {
+        createDispatchEvent: () => console.log('local mock createDispatch'),
```
overmyheadandbody, line 74: Is this still needed?

.github/workflows/mark-stale-prs.yaml
```diff
@@ -1,7 +1,7 @@
-name: "Close stale pull requests"
+name: 'Close stale pull requests'
```
overmyheadandbody, line 1: Not important, but I see we now have [names](https://github.com/adobecom/milo/blob/stage/.github/workflows/merge-to-stage.yaml#L1) without any quotes, [names](https://github.com/adobecom/milo/blob/stage/.github/workflows/mark-stale-prs.yaml#L1) with double quotes and now this. We should stick to a rule for consistency

.github/workflows/merge-to-main.js
```diff
@@ -0,0 +1,77 @@
...
+    .then((pr) => pr && addLabels({ pr, github, owner, repo }));
+
+const workingHours = () => {
+  const now = new Date();
+  const day = now.getUTCDay();
+  const hour = now.getUTCHours();
+  const isSunday = day === 0;
+  const isSaturday = day === 6;
+  const isFriday = day === 5;
+  return hour >= 8 && hour <= 20 && !isFriday && !isSaturday && !isSunday;
```
overmyheadandbody, line 28: This sounds like a good start for IST/Europe. Do you see this expanding to the US timezone as well at one point?

.github/workflows/merge-to-main.yaml
```diff
@@ -0,0 +1,37 @@
+name: Merge to main
+
+on:
+  pull_request:
+    types: [labeled]
```
overmyheadandbody, line 5: Judging by this and the JS file, we're looking just for SOT approvals and nothing else, right? Meaning: no more 'Ready for Main' label required. Is there any use-case where we would need to look out for the 'do not merge' label? I can't think of some particular use-cases right now, but I just saw the label on a [PR](https://github.com/adobecom/milo/pull/2362) of yours which gave me the idea.

overmyheadandbody (approved):

.github/workflows/helpers.js
mokimo: Yes

narcis-radu (approved):

.github/workflows/mark-stale-prs.yaml
mokimo: There is no easy was to decide on this, for longer text with special characters, using (single) quotes is the way to go.
This won't work with numbers though, as it would convert numbers to strings.
Then there are smaller lines such as
```
    runs-on: ubuntu-latest
```

...

For the sake of simplicity, I'll just remove the quotes from the name for now and keep it on the cron schedule

mokimo: If we don't like the inconsistency we should do _one_ PR and potentially look into configuring the linter to handle a standard config for us in the future

.github/workflows/merge-to-main.js
mokimo: 8PM UTC would be:
- 4PM east coast (New York)
- 2PM (Utah)
- 1PM (San Jose)

I tried to account for an overlap of all (US) time zones without stretching them out too aggressively to give us good coverage in terms of available engineers. 

If you have any better suggestions, I'd gladly implement them - otherwise I'd recommend we start with trying to merge "earlier" into the day

.github/workflows/merge-to-main.yaml
mokimo: Yes, just the SOT approvals & a PR needs to be mergeable.
- Conflicts = PR can't be merged
- Requesting changes = PR can't be merged
- only 3/5 SOT labels = PR will not be merged

I thought about the `do not merge` label, but the standard GitHub workflow is requesting changes on a PR, which should be good enough

mokimo: `Ready for Main` was used to quickly filter/indicate the Stage->Main PRs, but as we have #milo-changelog + a recurring title, that should be good enough to find the prod PRs IMO

PR #2291, opened 2024-05-10
MWPW-147634 - Add support for a custom Form block endpoint

elan-tbx:
* The forms service in Franklin is getting turned off, which means the Form block as it exists right now will become obsolete. This PR adds the ability to include a custom endpoint for the data, so this block can still be used (assuming the author has another forms service they can plug it in to)
* This is accomplished by simply splitting the cell that includes the _input_ URL, and adding the _output_ URL to the second cell.

288181950-8be89e53-a91f-42c5-b862-77b7426c9fa6

Resolves: [MWPW-147634](https://jira.corp.adobe.com/browse/MWPW-147634)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ebartholomew/elan-form3?martech=off
- After: https://ebartholomew-MWPW-147634-form-endpoints--milo--adobecom.hlx.page/drafts/ebartholomew/elan-form3?martech=off


mokimo:
Why not just use metadata to read the input & output url?

libs/blocks/form/form.js
```diff
@@ -290,8 +290,9 @@ async function createForm(formURL, thankYou, formData) {
 }
 
 export default async function decorate(block, formData = null) {
-  const form = block.querySelector('a[href$=".json"]');
+  const [form, endpoint] = block.querySelectorAll(':scope > div:first-child a');
   const thankYou = block.querySelector(':scope > div:last-of-type > div');
   thankYou.remove();
-  if (form) form.replaceWith(await createForm(form.href, thankYou, formData));
+  if (endpoint) endpoint.parentElement.remove();
+  if (form) form.replaceWith(await createForm(form.href, thankYou, formData, endpoint?.href));
```
overmyheadandbody, line 297: Nit: optional chaining instead of `if`s?

overmyheadandbody, line 297: 4 arguments might justify transforming the config into an object to make it easier and less error-prone to define and maintain.

overmyheadandbody:
@mokimo - I guess there could be multiple forms per page, maybe?

Sartxi (approved):

chrischrischris (approved):

libs/blocks/form/form.js
```diff
@@ -232,7 +233,7 @@ function lowercaseKeys(obj) {
   }, {});
 }
 
-async function createForm(formURL, thankYou, formData) {
+async function createForm({ formURL, thankYou, formData, endpoint }) {
```
overmyheadandbody, line 236: Could we add a default to avoid `cannot access property of undefined`? As in
```suggestion
async function createForm({ formURL, thankYou, formData, endpoint } = {}) {
```

overmyheadandbody (approved):

elan-tbx:
Closing for now while a proper E2E pipeline is figured out, will re-open in the future.

PR #2301, opened 2024-05-14
MWPW-148054: Match Banner content with production

mirafedas:
The div's width with class `.text` inside the marquee was set to 600px, while on production the width is set to 66.667%.
 In this PR we change the width to match the XD.

Resolves: [MWPW-148054](https://jira.corp.adobe.com/browse/MWPW-148054)

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.live/products/catalog
- After: https://mwpw-148054-marquee-text-width--milo--mirafedas.hlx.page/drafts/mirafedas/catalog?martech=off

### DIFF
--- a/libs/blocks/marquee/marquee.css
+++ b/libs/blocks/marquee/marquee.css
@@ -386,7 +386,7 @@
   }
 
   .marquee.quiet .foreground .text {
-    max-width: 600px;
+    max-width: 66.6667%;
   }
 
   .marquee.center .foreground .text,


mokimo (approved):

yesil (approved):

Roycethan:
@mirafedas Latest post fix looks like this: I see the title moving to next line.. Lets check with Lucy whether it has to be in same line or not... XD expects it to be in same line
Mnemonics looks good now
image

rgclayton (approved):

Sartxi (approved):

overmyheadandbody (commented):
I see in the story that some font changes are still being discussed. The stakeholder should be aware that we shouldn't have exceptions for a particular copy in English; the text will change in the future and other languages will definitely have different text lengths.

mirafedas:
@stereoke 
The mnemonics block is set to take up all available space of the wrapping container (marquee in our case). 
The reason why mnemonics wrap to the next line is that the marquee content width is set to 600px. 
On prod the marquee content width is set to be 66.67% of the whole screen width. 
So this PR changes the marquee width to be the same as on prod - 66.67%. 
As we make the marquee content wider, the marquee title gets more space as well and on some desktop screen sizes it wraps leaving only the last word in the second row.
Lucy wants the title to be in one line, and normally we could do that by either decreasing its font size or increasing the marquee width, but as I understand both solutions do not go along with XD.
If we just set the title to always be in one line and not wrap, this will create problems for locales, and for cases when in the future the text will be longer.
Could you please advice what should be the proper solution here?

mirafedas:
Closed because Consonant team does not approve this change, will open a PR with another solution soon.

PR #2304, opened 2024-05-14
MWPW-148252: auto label zero impact PRs

mokimo:
### Description
Automatically apply the zero impact label on PRs. By auto labeling the PRs, we can always include `zero-impact` PRs to stage at any point, even after testing has started. They are purely technical changes (workflows, tests) that don't need to stay open for a long time and need to wait for a special batch, as they don't impact production

### MEP as zero impact
As discussed with @vgoodric MEP follows their own guidelines and has their own sign off process and experimental blocks can be turned off in target within a second. Additionally, anything contributed within the MEP folder only affects a fraction of the traffic.

Resolves: [MWPW-148252](https://jira.corp.adobe.com/browse/MWPW-148252)

**Test URLs:**
- Before: https://main--milo--mokimo.hlx.live/?martech=off
- After: https://auto-label-zero-impact--milo--mokimo.hlx.live/?martech=off


.github/workflows/label-zero-impact.js
```diff
@@ -0,0 +1,68 @@
+const { getLocalConfigs } = require('./helpers.js');
+const zeroImpactDirs = ['.github', '.kodiak', '.vscode', '.test', 'libs/mep'];
```
mokimo, line 2: @vgoodric can you confirm that `libs/mep` is special and never touches prod code or basically can be included in stage/prod at any point in time during IST/EU/US times?

narcis-radu, line 2: What about changes to CODEOWNERS or similar files in the root? I think on a PR like this https://github.com/adobecom/milo/pull/2194, the `zero-impact` label would be removed.

mokimo: Added a few more files & we can always enhance the list as we see the need.
If we wrongfully remove the `zero-impact` label, your PR would still get merged regularly with the next batch.

But for developing workflows / making repo adjustments / mep / unit tests, this will come in handy to reduce the number of open PRs.

overmyheadandbody (approved):

robert-bogos (approved):

PR #2315, opened 2024-05-16
MWPW-146214 [MILO][MEP] Option to have gnav but not main wait for Target

vgoodric:
Note: I'm creating the PR now so I can get the review process started, but I still need to finish updating unit tests.

Currently, MEP Target integration can be turned "on" or "off" via the page's metadata.  If Target is on, the page build is paused until a response from Target can be received and processed.  With Gnav testing, we will need to enable Target on many pages that do not normally need it, thus hurting LHS and core web vitals.

We need to have more refined control of how Target is used with MEP.  This PR introduces a third option: "gnav". When a page's Target metadata is set to "gnav", Personalization and Promos are still applied, but LCP can proceed without Target.  Then the martech call in the loadPostLCP function uses await before the gnav is built.
* Add functionality to support Target metadata using "gnav"
* Still support personalization and promo manifests impacting main
* Not executing commands from a post LCP manifest that would alter the main's DOM (ie replace, remove etc)
* Re-organize mep object to better support the above functionality
* Also "clean up" config by moving 3 MEP related objects to be nested inside mep.config (expBlocks, expFragments, experiments)
* Update preview button to reflect Target is on, but only for Gnav

Resolves: [MWPW-146214](https://jira.corp.adobe.com/browse/MWPW-146214)

**Test URLs for psi-check:**
- After: https://meptargetgnavonly--milo--adobecom.hlx.page/?martech=off

**Note: I ran the URLs below in PageSpeed Insights, but there is statistical noise.**

**Test URLS:**
Target manifest changing gnav to say "Large menu swap". There should be no visual between these links:
Page with Target fully "on". There is no significant LHS difference between these links:
Before (LHS 88): https://main--cc--adobecom.hlx.live/drafts/mepdev/fragments/2024/q2/gnavinternal/index-target-on?gnav=block
After (LHS 90): https://main--cc--adobecom.hlx.live/drafts/mepdev/fragments/2024/q2/gnavinternal/index-target-on?milolibs=meptargetgnavonly&gnav=block

Same page and manifest but with Target for gnav only. LHS shows large improvement with new code:
Before (LHS 84): https://main--cc--adobecom.hlx.live/drafts/mepdev/fragments/2024/q2/gnavinternal/index-target-gnav?gnav=block
After (LHS 99): https://main--cc--adobecom.hlx.live/drafts/mepdev/fragments/2024/q2/gnavinternal/index-target-gnav?milolibs=meptargetgnavonly&gnav=block

********************************
2 Target manifests.  One through Target updates the menu to say "Complete Menu Swap" and inserts a blade saying "I’m a section inserted by the gnav manifest". 2nd manifest through pzn metadata that inserts a blade saying "I’m a section inserted by the pzn manifest"
Page with Target fully "on". Both before and after will execute all 3 actions. There is no significant LHS difference between these links:
Before (LHS 86): https://main--cc--adobecom.hlx.live/drafts/mepdev/fragments/2024/q2/gnavinternal/index-pzn-target-on?gnav=withinsert
After (LHS 78): https://main--cc--adobecom.hlx.live/drafts/mepdev/fragments/2024/q2/gnavinternal/index-pzn-target-on?milolibs=meptargetgnavonly&gnav=withinsert

Same page and manifest but with Target for gnav only. Before will execute all 3 actions, but after will not execute the new blade action in the gnav manifest. LHS shows large improvement with new code:
Before (LHS 82): https://main--cc--adobecom.hlx.live/drafts/mepdev/fragments/2024/q2/gnavinternal/index-pzn-target-gnav?gnav=withinsert
After (LHS 99): https://main--cc--adobecom.hlx.live/drafts/mepdev/fragments/2024/q2/gnavinternal/index-pzn-target-gnav?milolibs=meptargetgnavonly&gnav=withinsert

### DIFF
--- a/libs/blocks/quiz-entry/quiz-entry.js
+++ b/libs/blocks/quiz-entry/quiz-entry.js
@@ -30,7 +30,6 @@ const App = ({
   const [selectedCards, setSelectedCards] = useState({});
   const questionCount = useRef(0);
   const [btnAnalytics, setBtnAnalytics] = useState(null);
-  const [fiCodeResults, setFiCodeResults] = useState({});
   const [showPopover, setShowPopover] = useState(false);
   const [suggestions, setSuggestions] = useState([]);
 
@@ -53,6 +52,35 @@ const App = ({
       (item) => item.options === optionsType,
     );
     return optionItem && optionItem[prop] ? optionItem[prop] : '';
+  };
+
+  const sendMLFieldFiCodesAnalytics = (filtered) => {
+    let val = '';
+    if (filtered.length > 0) {
+      const fiCodes = filtered.map((result) => result.ficode);
+      val = `Filters|${analyticsType}|${selectedQuestion?.questions}/interest-${fiCodes.join('-')}`;
+    }
+    window.alloy('sendEvent', {
+      documentUnloading: true,
+      xdm: {
+        eventType: 'web.webinteraction.linkClicks',
+        web: {
+          webInteraction: {
+            linkClicks: { value: 1 },
+            type: 'other',
+            name: val,
+          },
+        },
+      },
+      data: {
+        _adobe_corpnew: {
+          digitalData: {
+            primaryEvent:
+            { eventInfo: { eventName: val } },
+          },
+        },
+      },
+    });
   };
 
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


sharmrj:
I think we may see an increase in CLS with this change, but I can't confirm whether it is specifically due to this change. I ran a couple of lighthouse checks on https://main--cc--adobecom.hlx.live/drafts/mepdev/fragments/2024/q2/gnavinternal/index-pzn-target-gnav?milolibs=meptargetgnavonly&gnav=withinsert
and I found a large CLS jump (0.266). It seems to be because of the georouting modal, but I'm unsure if this is related to these specific changes. It might have an outsized effect on our P75 CLS for whatever pages this is enabled for, but it is impossible to say without actually putting it out there and looking at our RUM measurements. Is there a way to look at RUM numbers (namely P75 CLS and LCP) for just the pages this will be enabled on in a test?
[Here's the lighthouse report I just ran on the page above (I had to zip it so I could upload it)](https://github.com/adobecom/milo/files/15332572/main--cc--adobecom.hlx.live_2024-05-16_14-43-21.html.zip)

libs/features/personalization/personalization.js
```diff
@@ -200,27 +200,44 @@ const getBlockProps = (fVal) => {
...
 
-const consolidateObjects = (arr, prop) => arr.reduce((propMap, item) => {
+const consolidateArray = (arr, prop, existing) => {
+  let results = [];
+  arr.forEach((i) => {
+    results = [...results, ...i[prop]];
+  });
+  if (!existing) return results;
+  return [...existing, ...results];
+};
```
chrischrischris, line 210: nit: use reduce
```suggestion
const consolidateArray = (arr, prop, existing = []) => arr
  .reduce((results, i) => [...results, ...i[prop]], existing);
```

chrischrischris (approved):

PR #2317, opened 2024-05-16
FIX: Carousel CLS

sivasadobe:
* Fixed Indicators & Control arrows CLS in Carousel
* Also fixes the FIREFLY - CORE WEB VITALS - Ai-painting-generator | CLS issue

Resolves: [MWPW-147593](https://jira.corp.adobe.com/browse/MWPW-147593)

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.live/drafts/sivas/firefly/features/ai-painting-generator?martech=off
- After: https://main--cc--adobecom.hlx.live/drafts/sivas/firefly/features/ai-painting-generator?milolibs=mwpw-147593--milo--sivasadobe&martech=off

- Before: https://main--milo--sivasadobe.hlx.live/docs/library/kitchen-sink/carousel?martech=off
- After: https://mwpw-147593--milo--sivasadobe.hlx.live/docs/library/kitchen-sink/carousel?martech=off

### DIFF
--- a/libs/blocks/carousel/carousel.css
+++ b/libs/blocks/carousel/carousel.css
@@ -1,5 +1,5 @@
 .d-none {
-  display: none;
+  display: none !important;
 }
 
 .carousel {
--- a/libs/blocks/carousel/carousel.css
+++ b/libs/blocks/carousel/carousel.css
@@ -1,3 +1,7 @@
+.d-none {
+  display: none;
+}
+
 .carousel {
   /* Scope variables to carousel */
   --carousel-slide-button-height: 40px;
--- a/libs/blocks/carousel/carousel.js
+++ b/libs/blocks/carousel/carousel.js
@@ -20,7 +20,7 @@ function decorateNextPreviousBtns() {
   const previousBtn = createTag(
     'button',
     {
-      class: 'carousel-button carousel-previous',
+      class: 'carousel-button carousel-previous d-none',
       'aria-label': 'Previous',
       'data-toggle': 'previous',
     },
@@ -30,7 +30,7 @@ function decorateNextPreviousBtns() {
   const nextBtn = createTag(
     'button',
     {
-      class: 'carousel-button carousel-next',
+      class: 'carousel-button carousel-next d-none',
       'aria-label': 'Next',
       'data-toggle': 'next',
     },
@@ -342,7 +342,7 @@ export default function init(el) {
   const fragment = new DocumentFragment();
   const nextPreviousBtns = decorateNextPreviousBtns();
   const slideIndicators = decorateSlideIndicators(slides);
-  const controlsContainer = createTag('div', { class: 'carousel-controls' });
+  const controlsContainer = createTag('div', { class: 'carousel-controls d-none' });
 
   fragment.append(...slides);
   const slideWrapper = createTag('div', { class: 'carousel-wrapper' });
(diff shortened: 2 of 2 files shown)


libs/blocks/carousel/carousel.css
```diff
@@ -1,3 +1,7 @@
+.d-none {
+  display: none !important;
```
mokimo, line 2: `!important` is heavily discourged.  It breaks the CSS specificity
& it is horrible for maintainability. `!important` is the absolute last solution given a huge list of justifications, not something you just use

libs/blocks/carousel/carousel.js
```diff
@@ -20,7 +20,7 @@ function decorateNextPreviousBtns() {
   const previousBtn = createTag(
     'button',
     {
-      class: 'carousel-button carousel-previous',
+      class: 'carousel-button carousel-previous d-none',
```
mokimo, line 23: Could we rename the class to something more fitting such as `delay-display`

libs/blocks/carousel/carousel.js
```diff
@@ -378,6 +378,16 @@ export default function init(el) {
...
 
+  function handleImageLoad() {
+    if (!controlsContainer.classList.contains('d-none') && nextPreviousBtns.findIndex((btnEl) => btnEl.classList.contains('d-none')) === -1) this.removeEventListener('load', handleImageLoad, false);
+    else {
+      if (controlsContainer.classList.contains('d-none')) controlsContainer.classList.remove('d-none');
+      nextPreviousBtns.forEach((btnEl) => {
+        if (btnEl.classList.contains('d-none')) btnEl.classList.remove('d-none');
+      });
+    }
+  }
```
mokimo, line 389: I'm not particularly happy with this solution, it adds a lot of code paths and makes maintainability worse.

mokimo (commented):
Thanks for the contribution! It's important to avoid any sort of CLS so it's great to see a contribution in the name of performance.

This PR still needs improvement and I'm sure there should be an easier way to determine "image loaded: display button" 

Would be great if you can re-iterate on your solution and see if you can come up with something easier.

As a last step, you also want to ensure there are unit tests covering the new code - you can also start on that once you have a re-iterated solution.

overmyheadandbody (changes requested):
I agree with @mokimo. We should avoid using `!important` and define classes that are not specific to the block they're affecting. If you need elements to be hidden on initial load, why not hide them from CSS, then when the first image finishes loading, add a modifier class to the carousel parent that in turn display the arrows and controls?

rgclayton (changes requested):
I also agree with the above statements. Thanks @mokimo and @overmyheadandbody.

Looking at the example page. I'm not convinced any CLS is coming from the carousel controls. LightHouse suggests there is CLS coming from the Sign in button in the header. 

Throttling the network speed (fast 3G) and scrolling to the carousel, there isn't any shift/jump happening related to it. 

It does look like there is a slight shift coming from the Marquee "start creating" button, which causes a shift to ALL content under it.

I want to make sure we aren't adding unnecessary code to over complicate things and still not fix the problem.

I've added a screen capture. I placed the cursor right below the button. If you watch closely it looks like the button size is changing, causing a shift in all content below it.

You can also see the shift happening with the reference page [pointed to your code fix branch.](https://main--cc--adobecom.hlx.live/drafts/sivas/firefly/features/ai-painting-generator?milolibs=mwpw-147593--milo--sivasadobe&martech=off)



https://github.com/adobecom/milo/assets/2539954/c59411bd-5194-42d3-9fed-eebdec442341

sivasadobe:
Working on another solution based on the discussions in this PR.

PR #2324, opened 2024-05-16
[MWPW-141785] Implement custom link actions

narcis-radu:
Allow authors to define links / CTAs that prompt users to sign in. This is a generic approach and can be reused for other custom actions. https://github.com/orgs/adobecom/discussions/2303

Resolves: [MWPW-141785](https://jira.corp.adobe.com/browse/MWPW-141785)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/narcis/signout?martech=off
- After: https://login--milo--narcis-radu.hlx.page/drafts/narcis/signout?martech=off


narcis-radu:
@VKniaz - I had to refactor the `addFooter` method in order to keep the custom event listeners added to the card buttons. Please check the code, cc @Axelcureno

libs/blocks/card/cardUtils.js
```diff
@@ -32,19 +32,18 @@ export const addFooter = (links, container, merch) => {
...
-  footer = linksArr.reduce(
-    (combined, link, index) => (
-      `${combined}<div class="consonant-CardFooter-cell consonant-CardFooter-cell--${(linksLeng === 2 && index === 0) ? 'left' : 'right'}">${link.outerHTML}</div>`),
-    footer,
-  );
-  footer += '</div></div>';
-
-  container.insertAdjacentHTML('beforeend', footer);
-  links.forEach((link) => {
+  const footer = createTag('div', { class: 'consonant-CardFooter' }, hrTag);
```
mokimo, line 35: This looks unrelated to the custom links actions & like a separate improvement

narcis-radu: @mokimo - see https://github.com/adobecom/milo/pull/2324#issuecomment-2115321401 and https://github.com/orgs/adobecom/discussions/2303#discussioncomment-9457318

libs/utils/utils.js
```diff
@@ -626,6 +626,15 @@ export function decorateLinks(el) {
         rdx.push(a);
       }
     }
+    // Custom action links
+    const loginEvent = '#_evt-login';
```
mokimo, line 630: Can we also have a unit test for this?

mokimo (approved):

mokimo (approved):
Nice!

vhargrave (approved):
Works very nicely! I quickly tested a card or two as well, just to double check nothing's broken
https://login--milo--narcis-radu.hlx.page/drafts/vhargrave/merch

The only thing I'm finding a bit strange atm is the login itself. When I login, i'm taken away from the page I was on. I'm just guessing that's because we're testing on non-productive pages.

overmyheadandbody (approved):
Are you aware of other future custom actions? If so, would it be worth having a different file for them? As in: scan for links containing `#_evt-`; if there are matches, import (async) a custom actions module that handles the action. Maybe one action doesn't justify the additional complexity, but it might be worth it in the long run.

kundanpa:
Verified. After clicking on the Sign-In button, it redirects to the Sign-In page where login email and password submission is successful. Upon logging in, the Sign-In button disappears from the page.

robert-bogos (approved):

PR #2327, opened 2024-05-16
MWPW-143562 - Add "radio" variant to Tabs block

elan-tbx:
* As an author, I want the ability to recreate complex pricing tables like [those that previously were created in Dexter](https://www.adobe.com/acrobat/pricing/business.html). This PR adds a `radio` variant to the Tabs block to accomplish this -- essentially allowing for tabs within tabs, with an alternate style applied to the inner block.
* As the Dexter version of this price table required a full page refresh when switching between Individual/Business tabs, I believe this variant will allow for a much improved user experience.

Resolves: [MWPW-143562](https://jira.corp.adobe.com/browse/MWPW-143562)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ebartholomew/pricing?martech=off
- After: https://ebartholomew-MWPW-143562-radio-button-group--milo--adobecom.hlx.page/drafts/ebartholomew/pricing?martech=off


libs/blocks/tabs/tabs.css
```diff
@@ -323,6 +344,61 @@
...
+}
+
+.tabs.radio div[role="tablist"] .tab-list-container {
+  border: 1px solid #d9d9d9;
+  border-radius: 24px;
+  padding: var(--spacing-xs) var(--spacing-s) var(--spacing-s);
+  font-size: var(--type-body-s-size);
+  color: var(--text-color);
+  flex-direction: column;
+  align-items: start;
```
robert-bogos, line 366: We can remove the `align-items: start;` from here. It will default to `align-items: stretch;` which will result in the same behaviour in this scenario.
```suggestion
  flex-direction: column;
```

libs/blocks/tabs/tabs.css
```diff
@@ -205,6 +229,17 @@
...
 
+.tabs.radio div[role="tablist"] button {
+  padding: 0 0 var(--spacing-s);
+  font-size: var(--type-body-s-size);
+  font-weight: normal;
+  color: var(--text-color);
+  display: flex;
+  align-items: center;
+  white-space: normal;
+  text-align: start;
```
robert-bogos, line 240: I think we can remove these 2 lines as they are the default values.

robert-bogos (approved):
Looks good! 👍 I just have those 2 suggestions

ryanmparrish (changes requested):
@elan-tbx  - I know your on the beat w/ these, but just putting them here for transparency. Sorry to play hard ball

1) Title and direction in JIRA - This is a variant of the 'tabs' block and not a new block - details in this should be updated after the 1st sub-task: vet authoring pattern
2) content alignment - Most all block are left by default. Since this is using the tabs block, the authoring should be the same. Left default, center/right are variants. 
3) Dark theme - This needs to be considered in the design and requirements - All blocks are subject to a dark/light theme scope. Anytime an existing block gets an enhancement, this needs to be a consideration.

libs/blocks/tabs/tabs.css
```diff
@@ -323,6 +343,66 @@
...
+  display: none;
+}
+
+.tabs.radio div[role="tablist"] {
+  margin: 0 auto;
+  width: var(--grid-container-width);
+}
+
+.tabs.radio div[role="tablist"] .tab-list-container {
+  border: 1px solid #d9d9d9;
```
ryanmparrish, line 364: for consistency, should this be the `--tabs-border-color: #d8d8d8;` instead, may also help inherit dark theming?

libs/blocks/tabs/tabs.css
```diff
@@ -323,6 +343,66 @@
...
+  border-radius: 24px;
+  padding: var(--spacing-xs) var(--spacing-s) var(--spacing-s);
+  font-size: var(--type-body-s-size);
+  color: var(--text-color);
```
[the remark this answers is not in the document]
elan-tbx: @ryanmparrish Not on the labels themselves, but I did add a hover style to the "radio" element for each item -- the outline turns blue on hover. Figma designs did not include grayed out text for inactive tabs, so I kept it the same text color for all.

ryanmparrish (approved):

libs/blocks/tabs/tabs.css
```diff
@@ -331,10 +415,50 @@
   background: transparent;
 }
 
+.tabs.radio.center div[role="tablist"] {
+  justify-content: center;
+}
+
+.tabs.radio.right div[role="tablist"] {
```
overmyheadandbody, line 422: I'm on fire with the RTL stuff today 😅 On RTL, _right_ would be left, are we using the same naming conventions in other blocks?

overmyheadandbody (commented):
A few RTL comments from my side. I know Ryan had a very good initial analysis of this suggested change, maybe he has some more insight 🙂

libs/blocks/tabs/tabs.css
```diff
@@ -323,6 +347,66 @@
...
+  padding: var(--spacing-xs) var(--spacing-s) var(--spacing-s);
+  font-size: var(--type-body-s-size);
+  color: var(--tabs-active-text-color);
+  flex-direction: column;
+  align-items: start;
+  width: 100%;
+}
+
+.tabs.radio div[role="tablist"] button:last-of-type {
+  padding-right: 0;
```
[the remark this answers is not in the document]
ryanmparrish: maybe use `padding-inline-start` instead to inherit `rtl` flow

libs/blocks/tabs/tabs.css
```diff
@@ -323,6 +347,66 @@
...
+  padding-bottom: 0;
+}
+
+.tabs.radio div[role="tablist"] button::before {
+  content: '';
+  border: 2px solid var(--tabs-radio-gray);
+  border-radius: 100%;
+  height: 14px;
+  min-width: 14px;
+  margin-right: 11px;
```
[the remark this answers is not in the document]
ryanmparrish: maybe use `margin-lnline-end` instead of `margin-right`

libs/blocks/tabs/tabs.css
elan-tbx: @overmyheadandbody I agree with you on this, but it's unfortunately a convention that has been baked in everywhere, even in the Text block:

https://github.com/adobecom/milo/blob/bbdd381f9c3d178b80b4987b3ea50ec9862c43ad/libs/blocks/text/text.css#L95

overmyheadandbody: Oh, I see, thanks for the clarification! Then yes, it makes sense to keep it as-is for now for consistency.

overmyheadandbody (approved):
Looks good, thanks for integrating all the changes!

elan-tbx:
Copying over feedback from comments in the Jira ticket, for added clarity. These have all been implemented.

Height should be 56px instead of 50px.
The font should be Body S Bold instead of font-weight of 700.
The gap between the title and the radio button should be XS instead of S.
The gap between the radio button and the text should be 12px instead of 11px.

PR #2339, opened 2024-05-21
MWPW-148609 [Project PEP] Optionally Bypass xLG for Testing

sharmrj:
* Added the ability to mock entitlements for PEP for testing purposes. Users on non-prod environments can use the url param `?mockPepEnts=ent1,ent2,...` to do so.
* Built on top of a code snippet @overmyheadandbody pointed out [here](https://github.com/adobecom/milo/pull/1982/files#diff-e93790efe8a7a100504209494b3df6dd38476dcd35bf8538eda4479ae3bfce9aR61-R68)

Currently content qa takes 4-5 days when it should take 1-2 since it's difficult to see the prompt as authored. This PR fixes that problem.

Resolves: [MWPW-148609](https://jira.corp.adobe.com/browse/MWPW-148609)

**Test URLs:**
- After: https://bypassXlg--milo--sharmrj.hlx.page/?martech=off


nishantka (approved):

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -84,7 +84,12 @@ class AppPrompt {
   };
 
   doesEntitlementMatch = async () => {
-    const entitlements = await getConfig().entitlements();
+    const config = getConfig();
+    const entitlements = await config.entitlements();
+    if (config?.env?.name !== 'prod') {
+      const extraEnts = new URLSearchParams(window.location.search).get('mockPepEnts');
+      extraEnts?.split(',').forEach((ent) => entitlements.push(ent.trim()));
+    }
```
mokimo, line 92: Is content QA not required on prod as well? From what I heard, it's always a pain to set this up and consequently should work on every environment. Since this just runs if the query parameter is available IMO there is no harm to also run this on prod.

Rather than trying to get the right entitlements, should we just make this a bool?

E.g. `skipPepEntitlements=true`

L53/54 in the init:
```js
    const skipEntitlements = new URLSearchParams(window.location.search).get('skipPepEntitlements');
    const entMatch = skipEntitlements || await this.doesEntitlementMatch();
    if (!entMatch) return;
```

mokimo (approved):
Consider the comments, otherwise looks good!

robert-bogos (approved):

libs/features/webapp-prompt/webapp-prompt.js
sharmrj: Enabling this on prod is fine by me. It might be worth having both the ability to add specific entitlements, and to skip the whole thing altogether.

I'll run these by MarkL and add them in.

bandana147 (approved):

sigadamvenkata:
Issue tested with PS/Firefly product pages where PEP authored,

https://stage--cc--adobecom.hlx.page/products/firefly?milolibs=bypassXlg&skipPepEntitlements=true
https://stage--cc--adobecom.hlx.page/products/photoshop/ai?milolibs=bypassXlg&skipPepEntitlements=true

PEP showing up > and navigated to respective product page as its authored.

image

libs/features/webapp-prompt/webapp-prompt.js
sharmrj: > Enabling this on prod is fine by me. It might be worth having both the ability to add specific entitlements, and to skip the whole thing altogether.
> 
> I'll run these by MarkL and add them in.

Mark is fine with the additions and I've added them in. So this should be good to merge.

PR #2343, opened 2024-05-21
MWPW-148439 - FIX - Text wraps under icon & No space between Icon and text in RTL Locale.

sivasadobe:
* Changed the icons margin to `inline-start|end` rather than `left|right`. When the RTL locale is enabled, it will resolve the space issue for all milo pages.
* Fixed text wrapping under the icon.

Resolves: [MWPW-148439](https://jira.corp.adobe.com/browse/MWPW-148439)

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/bg/products/illustrator/free-trial-download?martech=off
- After: https://main--cc--adobecom.hlx.page/bg/products/illustrator/free-trial-download?milolibs=mwpw-148439--milo--sivasadobe&martech=off

**Milo Icon-Block**
- Before: https://main--cc--adobecom.hlx.page/bg/products/illustrator/free-trial-download?martech=off
- After: https://main--cc--adobecom.hlx.page/bg/products/illustrator/free-trial-download?milolibs=mwpw-148439--milo--sivasadobe&martech=off

### Browser Capabilities (margin-inline-<start|end>)
- https://caniuse.com/?search=margin-inline-end

**Mobile**
- Chrome (Android & iOS): All versions
- Safari: ≥ 3.2
- Firefox (Android & iOS): All versions

**Desktop**
- Chrome: ≥ 4.0
- Safari: ≥ 3.1
- Firefox: ≥ 3.0
- MS Edge: ≥ 88

### DIFF
--- a/libs/blocks/icon-block/icon-block.css
+++ b/libs/blocks/icon-block/icon-block.css
@@ -177,6 +177,11 @@
   margin-top: var(--spacing-s);
 }
 
+.icon-block .foreground .second-column .title-row {
+  display: flex;
+  align-items: center;
+}
+
 .icon-block.full-width.small .foreground .icon-area,
 .icon-block.vertical.small .foreground .icon-area {
   margin-bottom: var(--spacing-s);
--- a/libs/blocks/icon-block/icon-block.js
+++ b/libs/blocks/icon-block/icon-block.js
@@ -70,7 +70,15 @@ function decorateContent(el) {
       const textContent = el.querySelectorAll('.text-content > :not(.icon-area)');
       const secondColumn = createTag('div', { class: 'second-column' });
       textContent.forEach((content) => {
-        secondColumn.append(content);
+        let nodeToInsert = content;
+        const firstIcon = content.querySelector('.icon:first-child');
+        if (firstIcon) {
+          const titleRowSpan = createTag('span', { class: 'title-row' });
+          titleRowSpan.append(firstIcon, content);
+          nodeToInsert = titleRowSpan;
+        }
+
+        secondColumn.append(nodeToInsert);
       });
       if (secondColumn.children.length === 1) el.classList.add('items-center');
       el.querySelector('.foreground .text-content').append(secondColumn);
--- a/libs/styles/styles.css
+++ b/libs/styles/styles.css
@@ -553,7 +553,17 @@ span.icon.margin-right { margin-right: 8px; }
 
 span.icon.margin-left { margin-left: 8px; }
 
-span.icon.margin-inline-end { margin-inline-end: 8px; }
+span.icon.margin-inline-end { 
+  -moz-margin-end: 8px;
+    -webkit-margin-end:8px;
+      margin-inline-end:8px;
+}
+
+span.icon.margin-inline-start { 
+  -webkit-margin-start:8px;
+    -moz-margin-start: 8px;
+      margin-inline-start:8px;
+}
 
 .button-l .con-button span.icon.margin-left,
 .con-button.button-l span.icon.margin-left { margin-left: 12px; }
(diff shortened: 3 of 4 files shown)


libs/styles/styles.css
```diff
@@ -553,7 +553,17 @@ span.icon.margin-right { margin-right: 8px; }
 
 span.icon.margin-left { margin-left: 8px; }
 
-span.icon.margin-inline-end { margin-inline-end: 8px; }
+span.icon.margin-inline-end { 
+  -moz-margin-end: 8px;
+    -webkit-margin-end:8px;
```
overmyheadandbody, line 558: Are the prefixes actually needed? Based on caniuse, the browser versions requiring them go back to 2018, we'd very likely have much bigger issues on those browsers than alignment :)

Screenshot 2024-05-21 at 17 53 40

libs/blocks/icon-block/icon-block.css
```diff
@@ -177,6 +177,11 @@
   margin-top: var(--spacing-s);
 }
 
+.icon-block .foreground .second-column .title-row {
+  display: flex;
+  align-items: center;
+}
```
overmyheadandbody, line 183: Since you're already using flex, couldn't a `column-gap` be used instead of the more complex margin logic?

sivasadobe: This code is primarily used to fix the text under the icon.
Yes, `column-gap` will solve this problem on pages written by **ICON BLOCK**. However, if RTL is enabled, it will not fix icon spacing on other pages.
The goal is to solve the **no space** problem on all milo pages. As a result, we changed the margin in **icon.js** to `margin-inline`, allowing the margin to be added based on direction.

libs/styles/styles.css
sivasadobe: @overmyheadandbody 
Make sense.
Prefix are removed.

libs/blocks/icon-block/icon-block.js
```diff
@@ -70,7 +70,15 @@ function decorateContent(el) {
       const textContent = el.querySelectorAll('.text-content > :not(.icon-area)');
       const secondColumn = createTag('div', { class: 'second-column' });
       textContent.forEach((content) => {
-        secondColumn.append(content);
+        let nodeToInsert = content;
+        const firstIcon = content.querySelector('.icon:first-child');
```
overmyheadandbody, line 74: Just to be sure: you're targeting the icon that is the first sibling within its parent, right? If it had a text before it (or any other element), this would no longer select it and apply the logic. Is that the expectation?

libs/styles/styles.css
```diff
@@ -553,7 +553,9 @@ span.icon.margin-right { margin-right: 8px; }
 
 span.icon.margin-left { margin-left: 8px; }
 
-span.icon.margin-inline-end { margin-inline-end: 8px; }
+span.icon.margin-inline-end { margin-inline-end:8px; }
+
+span.icon.margin-inline-start { margin-inline-start:8px; }
```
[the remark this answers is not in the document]
overmyheadandbody: Oh, I was just referring to the formatting of it:
```suggestion
span.icon.margin-inline-end { margin-inline-end: 8px; }

span.icon.margin-inline-start { margin-inline-start: 8px; }
```

overmyheadandbody (approved):

suhjainadobe (approved):

spadmasa:
Validation done on the below url
https://mwpw-148439--milo--sivasadobe.hlx.live/drafts/siva/illustrator/free-trial-download?martech=off
 : with rtl set
image
https://main--cc--adobecom.hlx.page/bg/products/illustrator/free-trial-download?milolibs=mwpw-148439--milo--sivasadobe&martech=off
image
https://main--cc--adobecom.hlx.page/il_he/products/illustrator/free-trial-download?milolibs=mwpw-148439--milo--sivasadobe&martech=off
image
image

https://main--milo--adobecom.hlx.live/docs/library/kitchen-sink/icon-block?milolibs=mwpw-148439--milo--sivasadobe&martech=off

PR #2345, opened 2024-05-21
mepUpdate

kandavadivelan:
Currently in Milo MEP, the Target propositionDisplay call is not happenning after the Target propositions are rendered on the page.

This change is to make the Target propositionDisplay call from the MEP Martech.js file once the Target propositions are applied.

This wiki explains the changes in detail:
https://wiki.corp.adobe.com/pages/viewpage.action?pageId=3169882054

### DIFF
--- a/libs/martech/martech.js
+++ b/libs/martech/martech.js
@@ -76,6 +76,18 @@ const handleAlloyResponse = (response) => {
     .filter(Boolean);
 };
 
+const getPropositionsInfo = (response) => {
+  const propositionsInfo = ((response.propositions?.length && response.propositions) || []).map((i) => {
+    return {
+      id : i.id,
+      scope : i.scope,
+      scopeDetails : i.scopeDetails,
+      renderAttempted :  i.renderAttempted
+    }
+  })
+  return propositionsInfo;
+};
+
 function roundToQuarter(num) {
   return Math.ceil(num / 250) / 4;
 }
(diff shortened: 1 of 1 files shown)


libs/martech/martech.js
```diff
@@ -123,17 +135,21 @@ const getTargetPersonalization = async () => {
     window.lana.log('target response time', responseTime);
   }, { once: true });
 
-  let manifests = [];
+  let manifests = [], propositionsInfo = [];
```
chrischrischris, line 138: Linter should have caught this.  One line per var declaration.

libs/martech/martech.js
```diff
@@ -76,6 +76,18 @@ const handleAlloyResponse = (response) => {
     .filter(Boolean);
 };
 
+const getPropositionsInfo = (response) => {
```
chrischrischris, line 79: Do you have to modify `response.propositions` since the only thing being changed is removal of the `items` array?  Can it be sent to `satellite.track` with `items`?

libs/martech/martech.js
```diff
@@ -76,6 +76,18 @@ const handleAlloyResponse = (response) => {
...
+  const propositionsInfo = ((response.propositions?.length && response.propositions) || []).map((i) => {
+    return {
+      id : i.id,
+      scope : i.scope,
+      scopeDetails : i.scopeDetails,
+      renderAttempted :  i.renderAttempted
+    }
+  })
+  return propositionsInfo;
+};
```
chrischrischris, line 89:
```suggestion
const getPropositionsInfo = (response) => {
  const propositionsInfo = response.propositions
    ?.map(({ id, scope, scopeDetails, renderAttempted }) => (
      { id, scope, scopeDetails, renderAttempted }
    ));
  return propositionsInfo || [];
};
```

PR #2346, opened 2024-05-21
High Impact Alert Workflow

overmyheadandbody:
This defines a new workflow that sends notifications to a Slack channel as soon as a PR with a `high-impact` label is opened. The previous logic from the `merge-to-stage.js` script has been removed, since we want such a notification to be sent out as soon as the PR is opened, not when it eventually reaches the `stage` branch, to ensure it gets visibility early on. Additionally, some workflows have been adapted to only run if the owner is `adobecom` to reduce workflow runs on forks.

Resolves: [MWPW-148268](https://jira.corp.adobe.com/browse/MWPW-148268)

**Test URLs:**
- After: https://high-impact-workflow--milo--overmyheadandbody.hlx.page/?martech=off

### DIFF
--- a/.github/workflows/high-impact-alert.js
+++ b/.github/workflows/high-impact-alert.js
@@ -1,0 +1,28 @@
+const {
+  slackNotification,
+  getLocalConfigs,
+} = require('./helpers.js');
+
+const main = async (params) => {
+  const { context } = params;
+
+  try {
+    if (context.payload.label.name === 'high-impact') {
+      const { html_url, number, title } = context.payload.pull_request;
+      console.log('High impact label detected, sending Slack notification');
+      slackNotification(`:alert: High Impact PR has been opened: <${html_url}|#${number}: ${title}>.` +
+         ` Please prioritize testing the proposed changes.`, process.env.SLACK_HIGH_IMPACT_PR_WEBHOOK);
+    } else {
+      console.log('No high impact label detected');
+    }
+  } catch (error) {
+    console.error(error);
+  }
+};
+
+if (process.env.LOCAL_RUN) {
+  const { context } = getLocalConfigs();
+  main({ context });
+}
+
+module.exports = main;
--- a/.github/workflows/high-impact-alert.js
+++ b/.github/workflows/high-impact-alert.js
@@ -7,14 +7,15 @@ const main = async (params) => {
   const { context } = params;
 
   try {
-    if (context.payload.label.name === 'high-impact') {
-      const { html_url, number, title } = context.payload.pull_request;
-      console.log('High impact label detected, sending Slack notification');
-      slackNotification(`:alert: High Impact PR has been opened: <${html_url}|#${number}: ${title}>.` +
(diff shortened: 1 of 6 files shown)


.github/workflows/high-impact-alert.js
```diff
@@ -0,0 +1,28 @@
...
+    }
+  } catch (error) {
+    console.error(error);
+  }
+};
+
+if (process.env.LOCAL_RUN) {
+  const { context } = getLocalConfigs();
+  main({ context });
+}
```
mokimo, line 26: I'm curious, were the helpers to run things locally helpful?
I very much rely on developing locally first and only switching to CI when it's becoming necessary

.github/workflows/high-impact-alert.js
```diff
@@ -0,0 +1,28 @@
...
+
+  try {
+    if (context.payload.label.name === 'high-impact') {
+      const { html_url, number, title } = context.payload.pull_request;
+      console.log('High impact label detected, sending Slack notification');
+      slackNotification(`:alert: High Impact PR has been opened: <${html_url}|#${number}: ${title}>.` +
+         ` Please prioritize testing the proposed changes.`, process.env.SLACK_HIGH_IMPACT_PR_WEBHOOK);
+    } else {
+      console.log('No high impact label detected');
+    }
```
mokimo, line 17: I prefer habitually using early returns, as if/else cases don't scale well once you add one-two more nested if/else statements.

The spacing is off, but I'd go with something such as this:
```suggestion
    if (context.payload.label.name !== 'high-impact') {
      console.log('No high impact label detected');
      return
    }
    const { html_url, number, title } = context.payload.pull_request;
    console.log('High impact label detected, sending Slack notification');
    slackNotification(`:alert: High Impact PR has been opened: <${html_url}|#${number}: ${title}>.` + 
    ` Please prioritize testing the proposed changes.`, process.env.SLACK_HIGH_IMPACT_PR_WEBHOOK);
```

mokimo (approved):
Cool! The only use-case I'm thinking off is adding/removing labels, which would send multiple messages.

We can't easily verify a message has/hasn't already been sent though, since actions are stateless and we can't access slack.

So we need a bit of contributor education to avoid 10 messages because of adding and removing labels.

.github/workflows/high-impact-alert.js
overmyheadandbody: Initially, yes, but because I don't have a whole lot of experience with workflows, I ended up testing things on my fork and just updating the code there. In the future local development will likely be a more viable option.

mokimo: Just FYI u can also run the workflow on pushing updates, so working on a fork and triggering the workflow with new commits is also quite okay

overmyheadandbody:
I think most non-stage PRs get just one batch of labels, so we'll have to test this in the wild and see if it pollutes the Slack channel. The workflow could add its own label, say `notification-triggered`, and a PR with such a label would no longer send a notification. Except if someone manually removes it 🙂

mokimo (approved):

robert-bogos (approved):

narcis-radu (approved):

PR #2361, opened 2024-05-23
[MWPW-146056] Links are not converted to their stage counterparts

robert-bogos:
## Description
This PR solves the issue where, in the context of milo pages on Stage, authored Production links were not being converted to their Stage counterparts.

## Related Issue
Resolves: [MWPW-146056](https://jira.corp.adobe.com/browse/MWPW-146056)

## Testing instructions
Go to the test url and check the `a` tags `href` attribute. It should point to the stage environment.

## Screenshots:
**Before:**
Screenshot 2024-05-28 at 07 35 21

**After:**
Screenshot 2024-05-28 at 07 36 38

## Test URLs
**Milo:**
- Before: https://main--milo--robert-bogos.hlx.page/drafts/rbogos/page-default?martech=off&env=stage
- After: https://mwpw-146056-stage-links--milo--robert-bogos.hlx.page/drafts/rbogos/page-default?martech=off&env=stage


mokimo (commented):
Make sure to add a unit test for this scenario

libs/utils/utils.js
```diff
@@ -614,6 +614,12 @@ export function decorateLinks(el) {
     appendHtmlToLink(a);
     a.href = localizeLink(a.href);
     decorateSVG(a);
+    if (window.location.hostname.includes('.stage.adobe.com')
```
mokimo, line 617: One of the checks seems redundant. This should work as well:
```
if (window.location.hostname.includes('.stage.adobe.com') && !a.hostname.includes('.stage.adobe.com')) {
  a.href = a.href.replace('.adobe.com', '.stage.adobe.com');
}
```

robert-bogos: That is true 👍 Before I was thinking that we have links pointing to stage already, but they won't pass the `!a.hostname.includes('.stage.adobe.com'))` check anyway

libs/utils/utils.js
```diff
@@ -614,6 +614,9 @@ export function decorateLinks(el) {
     appendHtmlToLink(a);
     a.href = localizeLink(a.href);
     decorateSVG(a);
+    if (window.location.hostname.includes('.stage.adobe.com') && !a.hostname.includes('.stage.adobe.com')) {
```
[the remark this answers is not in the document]
robert-bogos: Will do 👍

libs/utils/utils.js
```diff
@@ -614,6 +614,9 @@ export function decorateLinks(el) {
...
+      a.href = a.href.replace('.adobe.com', '.stage.adobe.com');
```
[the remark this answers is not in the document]
robert-bogos: I was not aware of the anti pattern consumers. Will reiterate 👍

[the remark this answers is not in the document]
overmyheadandbody: Could very well be, I was just trying to showcase this might need a bit more thought for all Milo consumers

libs/utils/utils.js
```diff
@@ -133,6 +133,13 @@ ENVS.local = {
   ...ENVS.stage,
   name: 'local',
 };
+const STAGE_DOMAINS_MAP = {
+  'www.adobe.com': 'www.stage.adobe.com',
+  'blog.adobe.com': 'blog.stage.adobe.com',
+  'business.adobe.com': 'business.stage.adobe.com',
+  'helpx.adobe.com': 'helpx.stage.adobe.com',
+  'news.adobe.com': 'news.stage.adobe.com',
+};
```
overmyheadandbody, line 142: I'm still not convinced Milo should be maintaining this list. My suggestion would be to ask consumers to add it the way they do `prodDomains`, like Milo [does](https://github.com/adobecom/milo/blob/stage/libs/scripts/scripts.js#L20), for example.

mokimo: Was just about to write the same, milo should be agnostic about who implements it.

robert-bogos: I was thinking of it as a default value that the consumers can overwrite [through the config they set](https://github.com/adobecom/milo/pull/2361/files#diff-28218f239d8d9c4d41e1924854440b50964619e56058fdd0ae243ff5ff6c616dR225-R227). Anyway, makes sense to keep this completely separate from milo 👍

mokimo: Having sensible defaults was one of my earlier suggestions as well, but we add precious pre LCP bytes to the utils for something that will _never_ be used on production

libs/scripts/scripts.js
```diff
@@ -19,6 +19,8 @@ import {
 // Production Domain
 const prodDomains = ['milo.adobe.com'];
 
+const stageDomainsMap = { 'www.adobe.com': 'www.stage.adobe.com' };
```
mokimo, line 22: IMO we should add all known envs here, a lot of the script-configs are simply copy pasted from milo from what I've seen.

test/utils/utils.test.js
```diff
@@ -414,6 +414,13 @@ describe('Utils', () => {
       expect(block).to.be.null;
       expect(document.querySelector('.quote.hide-block')).to.be.null;
     });
+
+    it('should convert prod links to stage links on stage env', async () => {
+      const links = document.body.querySelectorAll('a[href*="www.adobe.com"]');
+      utils.setConfig({ ...config, env: { name: 'stage' }, stageDomainsMap: { 'www.adobe.com': 'www.stage.adobe.com' } });
+      await utils.decorateLinks(document.body);
+      for (const link of links) expect(link.hostname === 'www.stage.adobe.com').to.be.true;
```
mokimo, line 422: Can we expand the test to check another consumer as well? e.g. bacom or something?

libs/scripts/scripts.js
robert-bogos: My reasoning was to keep it as short as possible because it won't be used on production. Anyway, if consumers copy paste this for their script it would make sense to have a more complete list. Will add more envs 👍

test/utils/utils.test.js
robert-bogos: Done 👍

overmyheadandbody (approved):
Looking much better! We should open a discussion to communicate these changes, as Milo consumers will need to define their own map for the feature to be enabled.

libs/utils/utils.js
```diff
@@ -608,12 +608,16 @@ export function decorateAutoBlock(a) {
...
 
 export function decorateLinks(el) {
+  const config = getConfig();
   decorateImageLinks(el);
   const anchors = el.getElementsByTagName('a');
   return [...anchors].reduce((rdx, a) => {
     appendHtmlToLink(a);
     a.href = localizeLink(a.href);
     decorateSVG(a);
+    if (config.env?.name === 'stage' && config.stageDomainsMap?.[a.hostname]) {
```
mokimo, line 618: The only part that I'm not happy with, is loading extra code for stage. But I don't think there is a (good) way around it.

mokimo (approved):

libs/utils/utils.js
robert-bogos: True, I also don't think there is a good way around it. At least the extra code is minimal.

libs/scripts/scripts.js
```diff
@@ -19,6 +19,14 @@ import {
 // Production Domain
 const prodDomains = ['milo.adobe.com'];
 
+const stageDomainsMap = {
```
narcis-radu, line 22: Would we need the same mapping to be configured by each consumer or is it enough to have it in Milo? Should consumers only add new values or maybe contribute the mapping to Milo directly ?

robert-bogos: The idea of the current solution is to be agnostic. Each consumer would need to add their own mapping and, in this way, will have complete control over which links get converted to their stage counterparts. Milo would not have any default mapping that the consumer can add to.

overmyheadandbody: This is in `scripts.js`, thus only specific to Milo to make it easy for people previewing things inside Milo's Sharepoint to test this. It will be up to individual consumers to define which hostnames they want converted to their stage counterparts. I'm not sure whether Milo should maintain a list of all its consumer's hostnames, that may be overhead we'd rather avoid.

I also get your concern that multiple consumers would have to maintain a similar map, but if we think of Milo as a sort of 3rd party, it would never have client-specific information within it, instead it would ask client to define setting through a config.

Maintaining such a list in a spreadsheet is also not ideal, since it would entail an extra request; granted, it would be just for Stage, but it would still affect performance in a critical point of the page's loading timeline.

Do you have any other ideas of how we could achieve this without overhead in either Milo or an extra network request?

narcis-radu: Great feedback, thank you. Indeed, my problem is with repetitive content (and more specific global navigation content where we have all sort of links), but it makes sense to ask consumers to define their own mappings if we're trying to be agnostic. There might be consumers that don't need the navigation or these mappings.
As long as we're able to define a full mapping in `federal`, I think most of the limitations should be automatically solved.

narcis-radu (approved):

SilviuLCF (approved):
Verified and ready to STAGE
TEsting details https://jira.corp.adobe.com/browse/MWPW-146056

PR #2373, opened 2024-05-27
Start the RCP for stage earlier

mokimo:
### Description
We want to keep `stage` clean in case of CSOs during RCPs to make it easier to contribute changes without pushing unwanted changes to production.

For this reason, we'd want to stop opening PRs to stage **before** the RCP so that nothing is open **during** the RCP.

**Test URLs:**
- After: https://keep-stage-clean--milo--mokimo.hlx.page/?martech=off

### DIFF
--- a/.github/workflows/merge-to-stage.js
+++ b/.github/workflows/merge-to-stage.js
@@ -179,7 +179,7 @@ const main = async (params) => {
   github = params.github;
   owner = params.context.repo.owner;
   repo = params.context.repo.repo;
-  if (isWithinRCP()) return console.log('Stopped, within RCP period.');
+  if (isWithinRCP(2)) return console.log('Stopped, within RCP period.');
 
   try {
     const stageToMainPR = await getStageToMainPR();
--- a/.github/workflows/helpers.js
+++ b/.github/workflows/helpers.js
@@ -43,15 +43,21 @@ const RCPDates = [
   },
 ];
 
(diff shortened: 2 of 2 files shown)


.github/workflows/merge-to-stage.js
```diff
@@ -224,7 +224,10 @@ const main = async (params) => {
     throw new Error('ADD NEW RCPs');
   }
   for (const { start, end } of RCPDates) {
-    if (start <= now && now <= end) {
+    const adjustedStart = new Date(start);
+    adjustedStart.setDate(adjustedStart.getDate() - 1);
```
overmyheadandbody, line 228: As discussed, let's do 2, one to get SOT approvals, one to actually merge.

Also, I think most RCPs start on Sundays, even this approach might not be enough and we should instead look for the Wednesday before an RCP, so that we can eventually merge it on Thursday. But that may complicate things even further 🙂

mokimo: The alternative is looking into the RCP dates and keeping a seperate list for stage and manually going through all the dates and editing them.

Contributions welcome 😁

overmyheadandbody (approved):
Agree, let's see how far this gets us and manually adjust dates if we run into issues.

robert-bogos (approved):
Looks good 👍

overmyheadandbody (approved):

PR #2374, opened 2024-05-27
[MWPW-148268] Adjust high impact workflow

overmyheadandbody:
This attempts to solve a workflow issue where the Slack Webhook secret was not accessible to forks, thus PRs open from forks would not be able to send Slack notifications.

Resolves: [MWPW-148268](https://jira.corp.adobe.com/browse/MWPW-148268)

**Test URLs:**
- Before: https://main--milo--overmyheadandbody.hlx.page/?martech=off
- After: https://adjust-high-impact-workflow--milo--overmyheadandbody.hlx.page/?martech=off

### DIFF
--- a/.github/workflows/high-impact-alert.yml
+++ b/.github/workflows/high-impact-alert.yml
@@ -1,7 +1,7 @@
 name: High Impact Alert
 
 on:
-  pull_request:
+  pull_request_target:
     types:
       - labeled
 
@@ -15,6 +15,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4.1.4
+        with:
+          ref: ${{ github.event.pull_request.base.ref }}
 
       - name: Send Slack message for high impact PRs
         uses: actions/github-script@v7.0.1


.github/workflows/high-impact-alert.yml
```diff
@@ -1,7 +1,7 @@
 name: High Impact Alert
 
 on:
-  pull_request:
+  pull_request_target:
```
mokimo, line 4: Just to double check, I know you tested this: The payload will stay the same? `const { html_url, number, title } = context.payload.pull_request;`

mokimo (approved):

.github/workflows/high-impact-alert.yml
mokimo: The webhook payload will be [pull_request](https://docs.github.com/en/webhooks/webhook-events-and-payloads#pull_request) so self-answer: Yes

overmyheadandbody: I'd say those stay the same, yes, they still use the correct PR data

Screenshot 2024-05-27 at 14 52 18

robert-bogos (approved):

PR #2375, opened 2024-05-28
Add news.adobe.com as known federal origin

mokimo:
### Description
As part of the news.adobe.com migration, we want to add news.adobe.com as known federal origin as it was set up for federal from an akamai perspective

Resolves: [MWPW-151236](https://jira.corp.adobe.com/browse/MWPW-151236)

### Visuals
Screenshot 2024-05-28 at 09 40 25

### Test instructions
Milolibs don't work on news.adobe.com as the consumer has not set this up yet. Using local overrides to test this change is the most convenient and quickest method to verify this PR works as intended on the consuming pages.
Test link from @Ruchika4 : https://news.stage.adobe.com/news/news-details/test-redirect

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.live/?martech=off
- After: https://mwpw-151236--milo--mokimo.hlx.live/?martech=off

PR #2380, opened 2024-05-29
MWPW-148278: Dynamic-Nav hide breadcrumbs

JasonHowellSlavin:
* Ensures that breadcrumbs do not show up on "on" pages when the dynamic nav is active

Resolves: [MWPW-148278](https://jira.corp.adobe.com/browse/MWPW-148278)

Context: 
Our PMs and SEO teams determined that within the journey where the dynamic nav follows the user, breadcrumbs that do not reflect the path the user has taken may be confusing. However, SEO still wanted the page's breadcrumbs to be visible to Google, so we are only stopping the creation of the feds-breadcrumbs and adjusting the classes to reflect the absence of breadcrumbs. 

Note: You will need to add a gnavSource to the session storage in order to verify that this works:
`window.sessionStorage.setItem('gnavSource', 'https://dynamic-nav-config--bacom--adobecom.hlx.page/gnav/localnav-aem-sites')`
In the normal workflow, a user would have visited an "entry page" where this would have been set. 

**Test URLs:**

- Before: https://main--milo--adobecom.hlx.page/drafts/slavin/dynamic-nav/bacom-sample-on?martech=off
- After: https://dynamic-nav-hide-breadcrumbs--milo--jasonhowellslavin.hlx.page/drafts/slavin/dynamic-nav/bacom-sample-on?martech=off

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -951,9 +951,6 @@ class Gnav {
   };
 
   decorateBreadcrumbs = async () => {
-    if (getMetadata('dynamic-nav') === 'on' && window.sessionStorage.getItem('gnavSource') !== null) {
-      return null;
-    }
     if (!this.block.classList.contains('has-breadcrumbs')) return null;
     if (this.elements.breadcrumbsWrapper) return this.elements.breadcrumbsWrapper;
     const breadcrumbsElem = this.block.querySelector('.breadcrumbs');
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -951,6 +951,10 @@ class Gnav {
   };
 
   decorateBreadcrumbs = async () => {
+    if (getMetadata('dynamic-nav') === 'on' && window.sessionStorage.getItem('gnavSource') !== null) {
+      document.querySelector('header').classList.remove('has-breadcrumbs');
+      return null;
+    }
     if (!this.block.classList.contains('has-breadcrumbs')) return null;
     if (this.elements.breadcrumbsWrapper) return this.elements.breadcrumbsWrapper;
     const breadcrumbsElem = this.block.querySelector('.breadcrumbs');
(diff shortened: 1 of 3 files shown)


libs/blocks/global-navigation/global-navigation.js
```diff
@@ -950,6 +950,10 @@ class Gnav {
   };
 
   decorateBreadcrumbs = async () => {
+    if (getMetadata('dynamic-nav') === 'on' && window.sessionStorage.getItem('gnavSource') !== null) {
+      document.querySelector('header').classList.remove('has-breadcrumbs');
```
overmyheadandbody, line 954: You would still have some breadcrumb elements in the DOM, right? https://github.com/adobecom/milo/blob/stage/libs/utils/utils.js#L683 Should those be removed as well? Also, wouldn't initially adding the class in utils and then removing it in the later loaded gnav have an effect on CLS, since the `has-breadcrumbs` class is defined in styles.css? https://github.com/adobecom/milo/blob/stage/libs/styles/styles.css#L714

JasonHowellSlavin: TBH, I was wondering where in the chain of things to put this, I think you are right about the CLS, I'll look into it. 

Thank you for the good catch @overmyheadandbody

JasonHowellSlavin: I made the change in decorate header, and removed the code from global nav. If `has-breadcrumbs` is never added then https://github.com/adobecom/milo/blob/stage/libs/blocks/global-navigation/global-navigation.js#L953 will return null as well. 

Does this seem a better solution @overmyheadandbody? Thank you again for your input.

overmyheadandbody (approved):
Hey @JasonHowellSlavin! Thanks for making the change! Yes, I'd say this is the better place to add this to address the issue at the root and avoid any flash of content that then gets removed and/or CLS issues.

sharmrj (approved):

mokimo (approved):

JasonHowellSlavin:
Nala is failing on video block tests, I am removing it from this PR since this does not touch that code and it has been verified.

mokimo:
@JasonHowellSlavin Good question, in theory we can contribute to nala as well and fix things ourselves. That has not happened yet - my [PR](https://github.com/adobecom/milo/pull/2375) also runs into failing tests and I've posted in [#milo-automation](https://adobe-mwp.slack.com/archives/C0442CQHJ69/p1717432939699159)

IMO in this case it might be fine to remove the label and have the PR passing if those are _really_ unrelated tests that are failing on multiple PRs

mokimo:
@JasonHowellSlavin so whenever this is ready, apply the 'ready for stage' label and this will get merged into prod!

PR #2391, opened 2024-05-31
MWPW-149124 Improve Focus Page for Performance Improvement Tiger Team 

TsayAdobe:
* Improve LCP for a marquee with a modal video
* This PR only improves loading of icons .css and .svg.
* A more general improvement for blocks is proposed in [this discussion](https://github.com/orgs/adobecom/discussions/2390). It can improve LCP more.

Resolves: [MWPW-149124](https://jira.corp.adobe.com/browse/MWPW-149124)

**Test URLs:**
- After: https://MWPW-149124--milo--adobecom.hlx.page/?martech=off

**DC Test URLs:**
- Before: https://www.stage.adobe.com/acrobat/how-to/pdf-editor-pdf-files.html
- After: https://www.stage.adobe.com/acrobat/how-to/pdf-editor-pdf-files.html?milolibs=MWPW-149124

Before
Before
After
After


libs/utils/utils.js
```diff
@@ -690,7 +690,8 @@ async function decorateIcons(area, config) {
   if (icons.length === 0) return;
   const { miloLibs, codeRoot } = config;
   const base = miloLibs || codeRoot;
-  await new Promise((resolve) => { loadStyle(`${base}/features/icons/icons.css`, resolve); });
+  loadStyle(`${base}/features/icons/icons.css`);
+  loadLink(`${base}/img/icons/icons.svg`, { rel: 'preload', as: 'fetch', crossorigin: 'anonymous' });
   const { default: loadIcons } = await import('../features/icons/icons.js');
```
mokimo, line 695: Since you don't wait for the css to finish loading at all, is there anything that stops `icons.js` from adding the icons to the page, an undecorated image showing up & only afterwards you add the CSS leading to CLS?

libs/utils/utils.js
```diff
@@ -690,7 +690,8 @@ async function decorateIcons(area, config) {
...
   await loadIcons(icons, config);
```
mokimo, line 696: I would suggest one of two things:

- Directly load the icons.svg if it's LCP relevant `await Promise.all([css, icons, js])`

OR

- move the css and svg loading into icons.js wait for both of those resources to be loaded in parallel then execute the logic.  In this case you could do this:
```js
async function decorateIcons(area, config) {
  const icons = area.querySelectorAll('span.icon');
  if (icons.length === 0) return null;
  return import('../features/icons/icons.js')
    .then(({ default: loadIcons }) => loadIcons(area, config));
}
```

or just load the utils from the icons.js

libs/utils/utils.js
TsayAdobe: The function name `loadStyle` is confusing. It just adds the CSS file to the HTML head.

mokimo: True, that's why you would need to promisify it, so that the callback within loadStyle/loadLink resolves the promise once the `onload` callback is called.

Otherwise, it's getting unpredictable when the CSS is ready

mokimo: The milo [loadBlock](https://github.com/adobecom/milo/blob/stage/libs/utils/utils.js#L454-L474) method loads the CSS and JS in parallel. However, it hides the section until everything is loaded.

Thanks @overmyheadandbody for finding those two critical lines:
https://github.com/adobecom/milo/blob/stage/libs/utils/utils.js#L766
https://github.com/adobecom/milo/blob/stage/libs/utils/utils.js#L1121

TsayAdobe: I don't get what you were trying to prove. The file `icons.css` is only 22 lines and it's for tooltip styles, which won't affect CLS at all.

libs/utils/utils.js
TsayAdobe: "Preloading is best suited for resources typically discovered late by the browser."
-- https://web.dev/articles/preload-critical-assets

Blainegunn (approved):

joaquinrivero (approved):

auniverseaway:
Can we get better test URLs?

stage.adobe.com cannot be tested and the Milo homepage is not a valid test as there are no icons in it.

auniverseaway:
Something to also note: your performance screenshots appear to be showing desktop experiences. It's really important when sharing performance related data that we are sharing mobile details and to be transparent about the ways in which we achieved certain numbers. I may test on 3G mobile, but you may test on 4G desktop, etc.

auniverseaway:
> Improve LCP for a marquee with a modal video

This is unfortunately not correct. This PR only impacts icons. If you link to a video modal with an icon button, sure... this will improve the LCP for that marquee. But we should not mis-represent what this PR is doing.

auniverseaway:
**I ended up using**

https://main--dc--adobecom.hlx.live/acrobat/how-to/pdf-editor-pdf-files
VS
https://main--dc--adobecom.hlx.live/acrobat/how-to/pdf-editor-pdf-files?milolibs=MWPW-149124

**Here are the deep PSI numbers for Mobile 4G:**

Screenshot 2024-06-25 at 3 39 27 PM

auniverseaway (changes requested):
I think this is not currently a meaningful enough change to warrant the complexity and  three extra lines in utils.

I do think you have a path to getting an improvement to icons, but will need some more tweaks in this PR.

1. 715: This works around a file called helpers.js that should be stripped out of icons altogether in favor of regular fetch. There is no need to have this dependency as it is only for cache busting content in specific situations for GWP. Icons does not fall into this.
    1. AI: remove the helpers.js preload
    2. AI: remove the use of helpers.js in icons.js
2. AI: Replace 711 & 712 with `const { base } = config;` on 711. This gains you a line.
3. 713 = Good.
4. 714 = probably needs to go. We don't want to preload icons if we can do something better. More on this below.

**Better:** we want to make icons as non-blocking as possible. Right now, everything is awaiting on icons. We await decorateIcons, we await loadIcons, we await the import.

icons.js unfortunately has critical parts to prevent CLS. there's styling and behavior in there that we do have to await for. However, **we don't have to wait for the actual SVG to load**. You'll notice at the beginning of the default function in icons.js, we _await_ for the SVG to be fetched. This is wrong. We do not have to await that file. We can still kick it off, but there's no reason for the function to wait for that. We don't really care about the contents and it's not a dependency for the other work.

What will happen in practice is that everything will be decorated, icons.js will not block, and the rest of the milo machine can continue while the svg is downloaded. When the svg is downloaded, we simply inject it into the slots determined through the rest of the function.

My guess is the improvement is still tiny (or non-existent, or worse.. who knows), but utils can stay lighter while we also move in a more sustainable direction of lazily loading icons. I'd rather us try this (I see it as a 4 hour task) to help us make an informed decision going forward.

I'm happy to jump on a call if you want to walk through this. I know digesting a wall of text can be a challenge. LMK.

CC @ryanmparrish

chrischrischris:
This is not directly related, but looking at [icon.css](https://github.com/adobecom/milo/blob/stage/libs/features/icons/icons.css) the css in there only applies to tooltips.  Thoughts on only loading icon.css when `icon-tooltip` is present?  The only issue I can see is if other icon things are added to the file then that condition would have to be updated or removed.  We could put a comment into icon.css as a warning to future devs.

chrischrischris:
One other possible improvement: Put a dummy element in the icon location that matches the icon size, then do not await any of the icon code and let it lazy load.  This may affect CLS but the tradeoff may be worth it for an improvement to LCP.

chrischrischris:
One thing that I've noticed is that increasing the amount of concurrent downloads is more advantageous on desktop vs mobile due to the limited bandwidth on mobile.  That may explain the numbers that @auniverseaway is seeing.  As long as there isn't a negative impact to perf on mobile then I think it's still worthwhile doing, but it may not be as much of a perf win than expected.

TsayAdobe:
> **I ended up using**
> 
> https://main--dc--adobecom.hlx.live/acrobat/how-to/pdf-editor-pdf-files VS https://main--dc--adobecom.hlx.live/acrobat/how-to/pdf-editor-pdf-files?milolibs=MWPW-149124
> 
> **Here are the deep PSI numbers for Mobile 4G:**
> 
> Screenshot 2024-06-25 at 3 39 27 PM

I run two tests to see the measurement error of the tool. 

Screenshot 2024-06-26 at 7 59 44 AM

Screenshot 2024-06-26 at 7 59 13 AM

A difference is considered significant only if it is larger than the possible error in measurement.

JingleH (approved):

PR #2392, opened 2024-05-31
MWPW-141022 [Project PEP] Prompt Dismissal + Tie-in with App Launcher UX

sharmrj:
* Dismissing the PEP prompt now points out the app switcher icon on the universal nav by playing an animation and displaying a tooltip for a limited time.
* Authors can now add four new fields to the section metadata of the PEP prompt to control the tooltip and the animation
    - `dismissal-animation-count`: The number of times the animation repeats
    - `dismissal-animation-duration`: How long each iteration of the animation lasts
    - `dismissal-tooltip-message`: Message to display on the tooltip
    - `dismissal-tooltip-duration`: How long to display the tooltip 
* Fixed an issue where urls were being made lowercase, breaking linking into some products (fixes [this issue](https://jira.corp.adobe.com/browse/MWPW-150562))

Resolves: [MWPW-141022](https://jira.corp.adobe.com/browse/MWPW-141022), [MWPW-150562](https://jira.corp.adobe.com/browse/MWPW-150562)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/raghavs/pep-prompt?martech=off&skipPepEntitlements=true
- After: https://dismiss-pep--milo--adobecom.hlx.page/drafts/raghavs/pep-prompt?martech=off&skipPepEntitlements=true
- Firefly: https://stage--cc--adobecom.hlx.live/products/firefly?milolibs=dismiss-pep&skipPepEntitlements=true

Testing notes:
- Make sure you delete the dismissedAppPrompt cookie after dismissing to make the prompt show up on subsequent reloads
- If the Universal Nav does not load, the PEP prompt will not load either.


libs/features/webapp-prompt/focus-animation.css
```diff
@@ -0,0 +1,63 @@
...
+	border-style: solid;
+	border-width: var(--coach-indicator-ring-border-size);
+	border-color: var(--coach-indicator-ring-default-color);
+
+	inline-size: var(--coach-indicator-ring-inline-size);
+	block-size: var(--coach-indicator-ring-block-size);
+	animation: pulse var(--animation-duration) linear;
+  animation-fill-mode: both;
+
+  border-radius: 0.313rem;
```
overmyheadandbody, line 44: I'd argue that such values with so many decimals make it counter-intuitive to understand and update. Also, with exception of some quiz styles, all other border radius styles are defined as pixel values, it might be a good idea to stick to that pattern, as related upcoming Spectrum tokens will also be defined in pixels.

libs/features/webapp-prompt/tooltip.css
```diff
@@ -0,0 +1,50 @@
+[data-pep-dismissal-tooltip]::after {
+  content: attr(data-pep-dismissal-tooltip);
+  position: absolute;
+  display: inline-flex;
+  z-index: 201;
```
overmyheadandbody, line 5: Is this high value actually needed? We usually go for the lowest value that would achieve the desired result. Is Unav setting some values to justify this?

libs/features/webapp-prompt/tooltip.css
```diff
@@ -0,0 +1,50 @@
...
+  left: calc(50% - 0.22rem);
+  top: 110%;
+  pointer-events: none;
+  
+  transform: rotate(45deg);
+
+  transition: opacity 0.5s;
+}
+
+@media (min-width: 1520px) {
```
overmyheadandbody, line 45: This is a non-standard media query value, not found anywhere else throughout the project. Is it really necessary? Could the `left` value not be defined in some other way?

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -35,6 +39,33 @@ const getIcon = (content) => {
...
 
+const showTooltip = (element, message, time = 5000) => {
+  element.setAttribute('data-pep-dismissal-tooltip', message);
+  setTimeout(() => {
+    element.removeAttribute('data-pep-dismissal-tooltip');
+  }, time);
+};
+
+const playFocusAnimation = (element, iterationCount = 2, animationDuration = 2500) => {
+  document.documentElement.style.setProperty('--animation-duration', `${animationDuration}ms`);
```
overmyheadandbody, line 50: As suggested in the CSS file, I'd avoid defining this globally. Could the value not be attached to the element itself?

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -35,6 +39,33 @@ const getIcon = (content) => {
...
+  const rings = [];
+  for (let i = 0; i < 3; i += 1) {
+    const ring = document.createElement('div');
+    ring.classList.add('coach-indicator-ring');
+    element.insertAdjacentElement('afterbegin', ring);
```
overmyheadandbody, line 55: Could `toFragment` not be used, since it's already imported in the module?

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -71,6 +102,11 @@ class AppPrompt {
       }));
     if (this.isAnchorExpanded) return;
 
+    const { base } = getConfig();
+
+    await loadStyle(`${base}/features/webapp-prompt/focus-animation.css`);
+    await loadStyle(`${base}/features/webapp-prompt/tooltip.css`);
```
overmyheadandbody, line 108: Was the performance impact of making two additional network requests compared to just concatenating these styles with the general PEP one?

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -209,14 +249,29 @@ class AppPrompt {
...
+  close = ({ saveDismissal = true, dismissalActions = true } = {}) => {
     const appPromptElem = document.querySelector(CONFIG.selectors.prompt);
     appPromptElem?.remove();
     clearTimeout(this.redirectFn);
     if (saveDismissal) this.setDismissedPrompt();
     document.removeEventListener('keydown', this.handleKeyDown);
     this.anchor?.focus();
     this.anchor?.removeEventListener('click', this.close);
+
+    const appSwitcher = document.querySelector('#unav-app-switcher');
```
overmyheadandbody, line 261: We likely don't want to maintain references to 3rd party code that might change at any time. The Unav already offers a way to fetch the element via `window.UniversalNav.getComponent?.('app-switcher')` which was already used for the PEP logic in another context.

sharmrj: Yeah I missed that. Fixed now.

libs/features/webapp-prompt/tooltip.css
sharmrj: Excepting cases where there's an icon to the right of the Unav (which, afaict, is exceedingly rare), we don't have enough space for the tooltip if it's centered unless the viewport is 1520px or greater. So the tooltip is off center when there's not enough space. See [here](https://www.figma.com/proto/xbe8zXEg1LWmTl0PCzhck4/PEP?page-id=0%3A1&node-id=1324-4136&t=R6HV7GiYLkDIwlvh-0&scaling=scale-down-width&starting-point-node-id=2%3A848)
The tooltip is 160px (10rem, including the padding -- from the designs), and we begin to get more space to the right at 1440px. So including half the waffle icon and the profile icon (16+32) and the gap between icons (4px/0.25rem) and the 8px feds gutter, we have 60px; 1440+60. The extra 20 px is for a little extra space to the right.
The reason we need to do this is because the tooltip has `absolute` positioning to position it relative to the waffle icon. So it can't know when it might be cut off by the viewport. So we need this non standard query to know when it's okay to center the tooltip.

libs/features/webapp-prompt/webapp-prompt.js
sharmrj: This is difficult to say. Networks have variance and are subject to long tails; singular or even tens of runs cannot adequately quantify which is faster because the question is really _how often_ is it faster. The answer to that question depends on the P75, P90, P99, (and so on) latency of our serverless environment. The answer to this question also depends on the user's device (and how hot it's running/how much battery it has, etc.) and the performance characteristics of brotli/gzip/zstandard decompression of files of different sizes.
A few cursory runs seem to indicate that the server should respond in under ~10ms, but this is subject to resource queueing delays in the browser, which are usually also under 10ms. But this is not data from the long tail.
If we want to optimize for the long tail (and I think we should), I would hazard a guess that we should probably put all the css into a single file -- the tradeoff is between project organization vs potential micro performance gain. 
I'd welcome more opinions on this.

Side note: More generally this question might be worth looking into with Milo more generally -- at what level of bundling vs splitting are we getting optimal performance (because we make many many blocking calls, and it takes only one P90 or worse latency to ruin the performance of the whole page; the probability of at least one P90 or worse latency in 30 network calls is 1 - 0.90^30 = ~96%. The probability that at least one will be P99 or worse is ~26%. 1 in 4 users would experience that)

libs/features/webapp-prompt/webapp-prompt.js
sharmrj: I've addressed this as part of the earlier comment

libs/features/webapp-prompt/tooltip.css
sharmrj: I've changed the z-indices to 3 and 2 so they remain higher than the z-index of the hover tooltip.

libs/features/webapp-prompt/focus-animation.css
sharmrj: You're right I've changed it to pixels. I've kept the `rem` units in `tooltip.css` that deal with text, though I converted the `border-radius` there to pixels as well.

libs/features/webapp-prompt/focus-animation.css
```diff
@@ -0,0 +1,63 @@
+:root {
+  --blue-700: rgba(56,146,243);
+  --coach-indicator-ring-default-color:var(--blue-700);
+  --coach-indicator-ring-diameter:1.25rem;
+  --coach-indicator-ring-default-color:var(--blue-700);
+	--coach-indicator-ring-border-size: 2px;
+	--coach-indicator-ring-inline-size: var(--coach-indicator-ring-diameter);
+	--coach-indicator-ring-block-size: var(--coach-indicator-ring-diameter);
+  --animation-duration: 3000ms;
```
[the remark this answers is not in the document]
sharmrj: See the reply to your previous comment

libs/features/webapp-prompt/focus-animation.css
```diff
@@ -0,0 +1,63 @@
+:root {
+  --blue-700: rgba(56,146,243);
```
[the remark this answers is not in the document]
sharmrj: The color is specific to the PEP prompt. I'll clean up the variables.

libs/features/webapp-prompt/tooltip.css
```diff
@@ -0,0 +1,50 @@
...
+  [data-pep-dismissal-tooltip]::after {
+    left: calc(50% - 5rem);
+    right: 0;
```
overmyheadandbody, line 48: Same RTL question

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -35,6 +39,50 @@ const getIcon = (content) => {
   return icons.company;
 };
 
+const showTooltip = (element, message, time = 5000) => {
```
overmyheadandbody, line 42: Shouldn't the default be set to `CONFIG.tooltipDuration `?

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -35,6 +39,50 @@ const getIcon = (content) => {
...
+  element.setAttribute('data-pep-dismissal-tooltip', message);
+  const cleanup = () => {
+    element.removeAttribute('data-pep-dismissal-tooltip');
+  };
+  const id = setTimeout(() => {
```
overmyheadandbody, line 47: What does `id` represent in this case? It doesn't seem to reflect its purpose at first sight

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -35,6 +39,50 @@ const getIcon = (content) => {
...
+    cleanup();
+  }, time);
+  element.addEventListener('click', () => {
+    cleanup();
+    clearTimeout(id);
+  }, { once: true });
+};
+
+const playFocusAnimation = (element, iterationCount = 2, animationDuration = 2500) => {
```
overmyheadandbody, line 56: Defaults should reflect the `CONFIG` values

libs/features/webapp-prompt/webapp-prompt.js
overmyheadandbody: Personally, I think this is a bit of an overhead. If we were to conditionally load the focus/tooltip styles based on user interaction, then it would've made sense to separate. But since we need to load everything anyway, it just adds additional requests. Sure, that may be a very cost, but it adds up with time. I'd personally just visually separate the three areas in the main CSS file.

Not saying that this is a blocker, since PEP is loaded later on the page anyway, but might be worth considering.

overmyheadandbody (commented):
Great progress, I still have a few more curiosities 🙂

sharmrj:
Regarding the RTL question, the rule is that if both left and right are specified then left wins in ltr and right wins in rtl (https://developer.mozilla.org/en-US/docs/Web/CSS/position). There seems to be only one place where this is the case and it seems to be a vestigial `right:0` in the media query of 1520px, which I'll remove.

libs/features/webapp-prompt/webapp-prompt.js
sharmrj: The `CONFIG` values are defaulted into `this.options`, but I think it doesn't hurt to add them here too.

libs/features/webapp-prompt/tooltip.css
sharmrj: see RTL comment below. I've also redone the absolute positioning to only use `top` and `left`

libs/features/webapp-prompt/webapp-prompt.js
sharmrj: It's the unique positive integer value returned by `setTimeout`. I've changed the name to be more expressive.

libs/features/webapp-prompt/tooltip.css
```diff
@@ -0,0 +1,50 @@
...
+
+[data-pep-dismissal-tooltip]::before { 
+  content: '';
+  position: absolute;
+  z-index: 2;
+  background-color: #3B63FB;
+  width: 0.44rem;
+  height: 0.44rem;
+  border-radius: 0.05rem;
+  left: calc(50% - 0.22rem);
```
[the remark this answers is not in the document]
sharmrj: see RTL comment below. I've also redone the absolute positioning to only use top and left

libs/features/webapp-prompt/tooltip.css
```diff
@@ -0,0 +1,50 @@
+[data-pep-dismissal-tooltip]::after {
+  content: attr(data-pep-dismissal-tooltip);
+  position: absolute;
+  display: inline-flex;
+  z-index: 3;
+  background-color: #3B63FB;
+  height: fit-content;
+  width: 8.875rem;
+  top: 120%; 
+  right: -75%;
```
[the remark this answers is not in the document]
sharmrj: See RTL comment below,

libs/features/webapp-prompt/focus-animation.css
```diff
@@ -0,0 +1,62 @@
...
+		transform: scale(2);
+		opacity: 0;
+	}
+}
+
+.coach-indicator .coach-indicator-ring {
+	display: block;
+	position: absolute;
+  top: 14%;
+  left: 13%;
```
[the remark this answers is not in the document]
sharmrj: I can see the argument for using pixels. UNav uses `rem` for the most part with the parent element having `font-size: medium`.

I use percentages in most other places since it allows me to reason about positioning using fractions. This turned out a bit wonky so I had to position it using a bit of trial and error, but I kept it percentage based for the consistency's sake.

libs/features/webapp-prompt/tooltip.css
```diff
@@ -0,0 +1,50 @@
+[data-pep-dismissal-tooltip]::after {
+  content: attr(data-pep-dismissal-tooltip);
+  position: absolute;
+  display: inline-flex;
+  z-index: 3;
+  background-color: #3B63FB;
```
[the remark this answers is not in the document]
sharmrj: I didn't make it a variable but I grouped together the common rules.

libs/features/webapp-prompt/webapp-prompt.css
```diff
@@ -188,3 +188,121 @@
...
+
+.coach-indicator .coach-indicator-ring:nth-child(3) {
+  animation-delay: calc(var(--animation-duration)*(0.66));
+}
+
+@media (prefers-reduced-motion: reduce) {
+	.coach-indicator .coach-indicator-ring {
+		animation: none;
+	}
+}
```
JasonHowellSlavin, line 308: Are the newlines on purpose on the lines: `205, 210, 211, 227, 244, 279, 283, 288`?

test/features/webapp-prompt/webapp-prompt.test.js
```diff
@@ -152,6 +179,26 @@ describe('PEP', () => {
     });
   });
 
+  describe('PEP dismissal tests', () => {
+    it('should add three rings to the app switcher and remove them after the alloted time', async () => {
```
JasonHowellSlavin, line 183: super nit: verbs to start test description statements: `adds three rings to the app switcher and removes them after the allotted time`. You can always add `should` to the start of test descriptions because our code always should do something.

JasonHowellSlavin (approved):
Good stuff. Some small nitpicky comments but looks good to me. Looks like a lot or Rares' comments were addressed. 

The CSS  file may do with some reorganization, since many of Milo's files have the media queries placed together going from smallest to largest display sizes.

sharmrj:
The Code Coverage check is failing since line 275 is not covered. This line wasn't covered earlier either as it exists inside a `setTimeout`, and due to quirks with how `useFakeTimers` works, the test doesn't seem to cover the lines inside the `setTimeout` callback.

salonijain3:
@mokimo The tests are not passing in this PR due to the comment above by @sharmrj. Can you take a look at this? 
cc: @overmyheadandbody

spadmasa:
https://main--cc--adobecom.hlx.page/de/creativecloud/animation/testdoc/peprollout?milolibs=dismiss-pep&skipPepEntitlements=true
image
image
image
image

image

image

image

narcis-radu:
@sharmrj - I am not able to see the profile / prompt on any of the links.
https://stage--cc--adobecom.hlx.live/products/firefly?milolibs=dismiss-pep&skipPepEntitlements=true

narcis-radu (dismissed):

libs/features/webapp-prompt/webapp-prompt.css
```diff
@@ -188,3 +188,119 @@
...
+  width: 8.875rem;
+  top: 125%;
+  left: -300%;
+  word-break: break-word;
+  border-radius: 7px;
+
+  padding-inline-start: 0.5625rem;
+  padding-inline-end: 0.5625rem;
+  padding-block-start: 0.25rem;
+  padding-block-end: 0.3125rem;
```
overmyheadandbody, line 209: Why not just define `padding-inline` and `padding-block`, since you're defining both values for each?

libs/features/webapp-prompt/webapp-prompt.css
```diff
@@ -188,3 +188,119 @@
...
+  font-family: 'Adobe Clean';
+  font-size: 0.75rem;
+  line-height: 0.9375rem;
+  color: white;
+}
+
+[data-pep-dismissal-tooltip]::before {
+  content: '';
+  z-index: 2;
+  background-color: #3B63FB;
```
overmyheadandbody, line 220: This is used in multiple places, it should be a scoped variable

libs/features/webapp-prompt/webapp-prompt.css
```diff
@@ -188,3 +188,119 @@
...
+  border-radius: 0.05rem;
+  left: calc(50% - 0.22rem);
+  top: 115%;
+
+  transform: rotate(45deg);
+}
+
+[data-pep-dismissal-tooltip]::before,
+[data-pep-dismissal-tooltip]::after {
+  background-color: #3B63FB;
```
overmyheadandbody, line 232: This is already defined in the individual `:after` / `:before` definitions, does it need to be defined here as well?

libs/features/webapp-prompt/webapp-prompt.css
```diff
@@ -188,3 +188,119 @@
...
+	inline-size: var(--coach-indicator-ring-inline-size);
+	block-size: var(--coach-indicator-ring-block-size);
+	animation: pulse var(--animation-duration) linear;
+  animation-fill-mode: both;
+
+  border-radius: 5px;
+}
+
+.coach-indicator .coach-indicator-ring:nth-child(1) {
+  animation-delay: calc(var(--animation-duration)*(0));
```
overmyheadandbody, line 291: Couldn't this be just `0`?

test/features/webapp-prompt/test-utilities.js
```diff
@@ -17,13 +17,19 @@ export const allSelectors = {
...
 
 export const defaultConfig = {
   color: '#b30b00',
   loaderDuration: 7500,
   redirectUrl: 'https://www.adobe.com/?pep=true',
   productName: 'photoshop',
+  animationCount: 5,
+  animationDuration: 2500,
+  tooltipMessage: 'Use the App Switcher to quickly find apps.',
+  tooltipDuration: 5000,
```
overmyheadandbody, line 32: Could these not be exported from the main logic file and imported here to avoid duplication?

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -212,7 +272,7 @@ class AppPrompt {
   };
 
   initRedirect = () => setTimeout(() => {
-    this.close({ saveDismissal: false });
+    this.close({ saveDismissal: false, dismissalActions: false });
```
overmyheadandbody, line 275: Could you test this by defining a hash as the redirect URL? That would avoid the page reload, but would trigger the UI changes.

sharmrj:
> MWPW-141022

Yeah the UNav `403`s if we're not connected to the vpn and sometimes even when we are (loading once without connecting to the vpn, connecting to the vpn, and then hard reloading, even after clearing the cache, doesn't fix the issue). 
I find that opening it in an incognito tab is helpful (provided we are already on the vpn when loading the page.

libs/features/webapp-prompt/webapp-prompt.js
sharmrj: Unfortunately this still results in this line ostensibly never being run. It's difficult to say why, but it may have something to do with how fakeTimers interacts with the event loop.n

libs/features/webapp-prompt/webapp-prompt.css
sharmrj: I'd kept them like this so I could fiddle with them. I've made it into a variable.

libs/features/webapp-prompt/webapp-prompt.css
sharmrj: I'd just missed deleting this one.

libs/features/webapp-prompt/webapp-prompt.css
sharmrj: I'd just missed deleting this one

libs/features/webapp-prompt/webapp-prompt.css
```diff
@@ -188,3 +188,119 @@
...
+  left: -300%;
+  word-break: break-word;
+  border-radius: 7px;
+
+  padding-inline-start: 0.5625rem;
+  padding-inline-end: 0.5625rem;
+  padding-block-start: 0.25rem;
+  padding-block-end: 0.3125rem;
+
+  font-family: 'Adobe Clean';
```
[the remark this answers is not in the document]
sharmrj: I've used `var(--body-font-family)` instead

libs/features/webapp-prompt/webapp-prompt.css
```diff
@@ -188,3 +188,119 @@
...
+[data-pep-dismissal-tooltip]::after {
+  content: attr(data-pep-dismissal-tooltip);
+  display: inline-flex;
+  z-index: 3;
+  background-color: #3B63FB;
+  height: fit-content;
+  width: 8.875rem;
+  top: 125%;
+  left: -300%;
+  word-break: break-word;
```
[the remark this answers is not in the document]
sharmrj: I just tested it with Japanese. It seems to be working fine. Are there other languages where this might be an issue?
Screenshot 2024-06-20 at 12 19 12 PM

narcis-radu:
@sharmrj - I am on VPN, I am in Incognito, I am signed in with my STAGE account but still cannot see the Profile Menu in https://stage--cc--adobecom.hlx.live/products/firefly?milolibs=dismiss-pep&skipPepEntitlements=true. I am however able to see the Profile Menu when I navigate to stage.adobe.com (which redirects to /home).

libs/features/webapp-prompt/webapp-prompt.css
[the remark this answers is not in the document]
narcis-radu: Visually things are good, yes, but JP has a different system where
In normal Japanese writing, no spaces are left between words, except if the writing is exclusively in [hiragana](https://en.wikipedia.org/wiki/Hiragana) or [katakana](https://en.wikipedia.org/wiki/Katakana) (or with very little [kanji](https://en.wikipedia.org/wiki/Kanji)), in which case spaces may be required to avoid confusion. See https://en.wikipedia.org/wiki/Japanese_punctuation

sharmrj:
> @sharmrj - I am on VPN, I am in Incognito, I am signed in with my STAGE account but still cannot see the Profile Menu in https://stage--cc--adobecom.hlx.live/products/firefly?milolibs=dismiss-pep&skipPepEntitlements=true. I am however able to see the Profile Menu when I navigate to stage.adobe.com (which redirects to /home).

This is very odd. I intermittently get the same problem, but incognito usually fixes it for me (or it kinda just resolves itself, not sure why). Is it the same on other browsers as well? Or maybe if you try deleting your auth cookie (for adobe.com and for the MS SSO), then the thing might show up?

libs/features/webapp-prompt/webapp-prompt.css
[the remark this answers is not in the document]
sharmrj: [According to MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break), `break-word` "Has the same effect as overflow-wrap: anywhere combined with word-break: normal, regardless of the actual value of the [overflow-wrap](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) property.". And `overflow-wrap: anywhere` has the [following description](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap): "To prevent overflow, an otherwise unbreakable string of characters — like a long word or URL — may be broken at any point if there are no otherwise-acceptable break points in the line. No hyphenation character is inserted at the break point. Soft wrap opportunities introduced by the word break are considered when calculating min-content intrinsic sizes."

So I think even in cases where there are no spaces, the line will break when it's necessary (and this is the behavior we want).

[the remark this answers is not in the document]
narcis-radu: I'm not arguing with that, I just think we need some validation to make sure things are good. I know we have a special library to deal with word wrap on JP - libs/features/japanese-word-wrap.js but not sure how your code will interact with that.

spadmasa:
Validation done  on double byte locale  , please find the below url with details 
https://main--cc--adobecom.hlx.page/jp/creativecloud/animation/testdoc/peprollout?milolibs=dismiss-pep&skipPepEntitlements=true
image

image
image
image
CC : @narcis-radu   @overmyheadandbody  @sharmrj @salonijain3 @sigadamvenkata

libs/features/webapp-prompt/webapp-prompt.css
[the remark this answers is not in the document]
salonijain3: @narcis-radu  This has been verified and working as expected.

salonijain3 (approved):

libs/features/webapp-prompt/webapp-prompt.css
chrischrischris: Also there is a mix of tabs and spaces.  All should be spaces.

sharmrj: I've fixed both the issues (some of the newlines were on purposes; I replaced each tab with the appropriate number of spaces)

narcis-radu (approved):

narcis-radu:
I wasn't able to test on Firefox / Safari - I have the same issue where the user does not appear logged-in in the URLs. It's working as expected in Chrome after several attempts.

PR #2399, opened 2024-05-31
MWPW-147741 [Geo-Routing] Need to load modal after bootstrap in order to ensure Analytics fires

vgoodric:
I've been asked by my boss' boss to drop all work to get this out Monday.  Marking high priority.

Geo modal analytics are being sent when the user closes the modal or chooses to stay on the current page.  Analytics are NOT being sent when the modal loads or they choose to change countries.

image

Issues solved:
1. Load events were being fired but it wasn't working because it relied on something done by martech.js.  So if the geomodal is needed, we await martech before creating the modal (note this is postLCP).
2. It sometimes said gb and sometimes said uk. This is a common issue with just this geo, so added a small fix to force it to always be uk.
3. daa-ll is preferable for capturing user events.  The close button was already using daa-ll.  So daa-ll was added to switch and stay, and the event to send analytics was removed from stay's click event. 

Resolves: [MWPW-147741](https://jira.corp.adobe.com/browse/MWPW-147741)

**Test URLs for psi-check:**
- After: https://geomodalanalytics--milo--adobecom.hlx.page/?martech=off

This is a tricky one to test.  This is the least painful way:
1. Be sure all incognito windows are closed.
2. Open a fresh incognito window
3. Open your network tab in dev tools to capture traffic
4. In the network tab, filter for the word `collect`
5. Go to https://main--bacom--adobecom.hlx.live/?akamaiLocale=GB&countryCode=GB&mboxOverride.browserIp=5.10.159.255&milolibs=geomodalanalytics
6. click on the first collect in the list. Click payload and expand the Request Payload.  You should see an entry called `Load:us-uk|Geo_Routing_Modal`  This is event 1 of 4 we need to capture
7. In your console, type `_satellite.setDebug(true);`
8. Click the close button
9. You'll see something saying `[Core] Rules using the direct call event type with identifier "event" have been triggered with additional detail` with an object.  Expand the object and look for `localeModal:modalClose:buttonClose`.  This is event 2 of 4 we need to capture.
10. Reload the page
11. Clear your console because it will have been spammed
12. Click the United States link.  Once again you'll see a message about the event with the object. Expand and you'll see 
`Stay:us-uk|Geo_Routing_Modal`. This is event 3 of 4 we need to capture.
13. Delete your "international" cookie and reload.  If you don't know how to clear cookies, you can close the window, open the URL in a new incognito window, and re-enter `_satellite.setDebug(true);` in your console.
14. If your console is spammed, you can clear it.
15. Hold down your command key (control key for Windows) and click the `United Kingdom` button. Once again you'll see a message about the event with the object. Expand and you'll see 
`Switch:us-uk|Geo_Routing_Modal`. This is event 4 of 4 we need to capture.

### DIFF
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -19,7 +19,7 @@ export function findDetails(hash, el) {
   return { id, path, isHash: hash === window.location.hash };
 }
 
-export function sendAnalytics(event) {
+function fireAnalyticsEvent(event) {
   // eslint-disable-next-line no-underscore-dangle
   window._satellite?.track('event', {
     xdm: {},
@@ -28,6 +28,17 @@ export function sendAnalytics(event) {
       _adobe_corpnew: { digitalData: event?.data },
     },
   });
+}
+
+export function sendAnalytics(event) {
+  // eslint-disable-next-line no-underscore-dangle
+  if (window._satellite?.track) {
+    fireAnalyticsEvent(event);
+  } else {
+    window.addEventListener('alloy_sendEvent', () => {
+      fireAnalyticsEvent(event);
+    }, { once: true });
+  }
 }
 
 export function closeModal(modal) {
--- a/libs/features/georoutingv2/georoutingv2.js
+++ b/libs/features/georoutingv2/georoutingv2.js
@@ -103,19 +103,18 @@ function getGeoroutingOverride() {
 }
 
 function decorateForOnLinkClick(link, urlPrefix, localePrefix) {
+  let eventType = 'Switch';
+  if (localePrefix !== undefined) eventType = 'Stay';
+  const modCurrPrefix = localePrefix || 'us';
+  const modPrefix = urlPrefix || 'us';
+  const eventName = `${eventType}:${modPrefix.split('_')[0]}-${modCurrPrefix.split('_')[0]}|Geo_Routing_Modal`;
+  link.setAttribute('daa-ll', eventName);
   link.addEventListener('click', () => {
-    const modPrefix = urlPrefix || 'us';
     // set cookie so legacy code on adobecom still works properly.
     const domain = window.location.host === 'adobe.com'
       || window.location.host.endsWith('.adobe.com') ? 'domain=adobe.com' : '';
     document.cookie = `international=${modPrefix};path=/;${domain}`;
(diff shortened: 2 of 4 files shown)


markpadbe (approved):

vhargrave (changes requested):
The daa-ll description for the Switch URLs seems to always set the locale of the page the user is switching away from to US, which is wrong
You can see that here - https://geomodalanalytics--milo--adobecom.hlx.page/ch_de/drafts/vhargrave/georouting?akamaiLocale=om
I'd expect the daa-ll description of switch URL here to be mena-ch for example, but instead it's mena-us 
 
I've written a bit of extra code which fixes the issue. If you want you can review & copy paste it here - https://github.com/adobecom/milo/pull/2404/files#diff-5353f6a7bd38256954f04062d9effabd975b061e55049968316ce24db587e768

vhargrave:
> The daa-ll description for the Switch URLs seems to always set the locale of the page the user is switching away from to US, which is wrong You can see that here - https://geomodalanalytics--milo--adobecom.hlx.page/ch_de/drafts/vhargrave/georouting?akamaiLocale=om I'd expect the daa-ll description of switch URL here to be mena-ch for example, but instead it's mena-us
> 
> I've written a bit of extra code which fixes the issue. If you want you can review & copy paste it here - https://github.com/adobecom/milo/pull/2404/files#diff-5353f6a7bd38256954f04062d9effabd975b061e55049968316ce24db587e768

@vivgoodrich maybe one more thing that just came to mind. Why are we having to await martech specifically for the georouting and not for other things? I'm a bit confused by that.

vgoodric:
> @vivgoodrich maybe one more thing that just came to mind. Why are we having to await martech specifically for the georouting and not for other things? I'm a bit confused by that.

@vhargrave It doesn't matter for setting daa-ll attributes, but the load event is sending a custom JS event.  Modal load calls `sendAnalyticsFunc` which is just `sendAnalytics` from modal.js.   That fires `window._satellite?.track` which is created by loading martech.  So the method can't be called until after martech.

vhargrave:
@vivgoodrich I've tested and am struggling a bit with step 4. I am not seeing any event firing if I click on the link and don't open it in a new tab. If I hold command and open the uk link in a separate tab then it seems to work, but the event should probably be firing in both cases.
Can you reproduce that or is it just me?

vhargrave (approved):

PR #2314, opened 2024-05-16
MWPW-138032: added tags to LANA logs

robert-bogos:
### Description
This PR is adding tags to all the existing LANA logs in milo.

Tags legend: 
`info` - network issues
`warn` - authoring related misconfigs or similar
`error` - actual error ( ex. cannot read Y of undefind )
`<module>`- extra tag to help centralise all logs related to a specific module

### Related Issue
Resolves: [MWPW-138032](https://jira.corp.adobe.com/browse/MWPW-138032)

## Test URLs
**Milo:**
- Before: https://main--milo--robert-bogos.hlx.page/drafts/rbogos/default?martech=off
- After: https://mwpw-138032-lana-logs-tags--milo--robert-bogos.hlx.page/drafts/rbogos/default?martech=off

PR #2402, opened 2024-06-03
MWPW-140523: Update existing LANA logs to capture potential client errors

Deva309:
## Description
This PR contains changes for adding lana logs for any client errors with `errorType=error` that will disrupt real user. 

Resolves: [MWPW-140523](https://jira.corp.adobe.com/browse/MWPW-140523)

## Testing
Change the index value in the Splunk search to `lana_nonprod` and verify the LANA logs.

## Screenshot
Screenshot 2024-06-03 at 10 43 49 AM

## Test URLs:
**MILO:**
- Before: https://main--milo--adobecom.hlx.live/drafts/devashish/lana-logs?martech=off&lanadebug
- After: https://mwpw-140523-lana-logs--milo--deva309.hlx.live/drafts/devashish/lana-logs?martech=off&lanadebug


libs/blocks/global-footer/global-footer.js
```diff
@@ -153,7 +153,9 @@ class Footer {
 
   loadIcons = async () => {
     const file = await fetch(`${base}/blocks/global-footer/icons.svg`);
-
+    if (!file.ok) {
+      throw new Error(`${file.statusText} url: ${file.url}`);
```
mokimo, line 157: I don't think throwing an error here is correct. Due to a network failure this can happen at any point and then we'd destroy the UX instead of gracefully failing (just not showing the icons)

### Before, load fails
Social icons are missing, but the footer looks "okay" from a consumer perspective
Screenshot 2024-06-03 at 10 29 00

### After
Footer is non-existent // only an empty box
Screenshot 2024-06-03 at 10 30 15

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -465,7 +465,7 @@ class Gnav {
     const profileData = await fetch(`https://${env.adobeIO}/profile`, { headers });
 
     if (profileData.status !== 200) {
-      return;
+      throw new Error(`${profileData.statusText} url: ${profileData.url}`);
```
mokimo, line 468: If you think that we should log the error, I'd rather suggest to add a lana log than stopping the execution.

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -512,9 +512,20 @@ class Gnav {
...
     };
 
+    const scriptUrl = `https://${environment}.adobeccstatic.com/unav/1.1/UniversalNav.js`;
+    const styleUrl = `https://${environment}.adobeccstatic.com/unav/1.1/UniversalNav.css`;
+
     await Promise.all([
-      loadScript(`https://${environment}.adobeccstatic.com/unav/1.1/UniversalNav.js`),
-      loadStyle(`https://${environment}.adobeccstatic.com/unav/1.1/UniversalNav.css`),
+      loadScript(scriptUrl),
+      loadStyle(styleUrl, (e) => {
```
mokimo, line 520: Would the error be part of the callback and not simply propagate? 🤔 
Why did you add the lana log here, but nowhere else for loadStyle/loadScript in this case?

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -330,6 +337,9 @@ export async function fetchAndProcessPlainHtml({ url, shouldDecorateLinks = true
     path = mepFragment.target;
   }
   const res = await fetch(path.replace(/(\.html$|$)/, '.plain.html'));
+  if (res.status !== 200) {
+    throw new Error(`${res.statusText} url: ${res.url}`);
```
mokimo, line 341: Again, not sure if throwing errors is the right idea, especially if the PR resolves around lana logs as the title indicates

mokimo (changes requested):
The story and ticket resolve around understanding current lana logs and fixing issues where needed.

I feel like the PR tackles error handling instead and leads to some less graceful failing scenarios, which is fine to tackle as well, but in this case this should have a different ticket and more dilligent testing links/scenarios

libs/blocks/global-footer/global-footer.js
Deva309: Understood, yes, we can refrain from including these types of Lana logs.

libs/blocks/global-navigation/global-navigation.js
Deva309: This code is part of `decorateUniversalNav` and is again used in `imsReady` as task with the existing lana logs as showd in previous comment. While an error for loadScript is already captured [here](https://github.com/adobecom/milo/blob/e6731831f7ecf2ceff64eb66b59a001f7c12b110/libs/utils/utils.js#L404), there isn't one for loadStyle. Currently, if there's a network issue while loading the CSS, the universalNav will display incorrectly like this:
Screenshot 2024-06-03 at 4 19 20 PM

Which is leading to a poor user experience. Therefore, the existing LANA log in `imsReady` doesn't capture this, which is why the log is now added for loadStyle.

@mokimo do you think this should be covered [here](https://github.com/adobecom/milo/blob/e6731831f7ecf2ceff64eb66b59a001f7c12b110/libs/utils/utils.js#L317) by thorwing an error?

libs/blocks/global-navigation/utilities/utilities.js
Deva309: This code is part of fetchAndProcessPlainHtml. Although we already have LANA logs for this, they don't work if the HTML assets fail to load. 
Refer to these code sections:
[Code Snippet 1](https://github.com/adobecom/milo/blob/stage/libs/blocks/global-navigation/global-navigation.js#L1008-L1014)
[Code Snippet 2](https://github.com/adobecom/milo/blob/stage/libs/blocks/global-footer/global-footer.js#L72-L82)

To make it work with the existing logs, an error throw has been added. Previously, if there was no content or if HTML assets failed to load, the code execution would also stop, as shown in Code Snippet 1 [here](https://github.com/adobecom/milo/blob/stage/libs/blocks/global-navigation/global-navigation.js#L1014) and similarly in Code Snippet 2.

Deva309:
> The story and ticket resolve around understanding current lana logs and fixing issues where needed.
> 
> I feel like the PR tackles error handling instead and leads to some less graceful failing scenarios, which is fine to tackle as well, but in this case this should have a different ticket and more dilligent testing links/scenarios

This PR addresses issues related to client errors which are there but somehow missed under the existing lana log. Currently, we call `logErrorFor` at the topmost call, which should cover all major client errors that disrupt real users.

For example, in the global-footer, we already log errors for [init](https://github.com/adobecom/milo/blob/e6731831f7ecf2ceff64eb66b59a001f7c12b110/libs/blocks/global-footer/global-footer.js#L42) and [decorateContent](https://github.com/adobecom/milo/blob/e6731831f7ecf2ceff64eb66b59a001f7c12b110/libs/blocks/global-footer/global-footer.js#L69C3-L69C18). Since `decorateContent` is part of init, wrapping it in `logErrorFor` is unnecessary. Therefore, to cover all legitimate client errors, this PR ensures that real errors are checked as part of the existing LANA logs only.

In a similar manner, wherever we currently have Lana logs but they fail to cover any client errors are addressed in this PR.
@mokimo WDYT?
Please let me know if my understanding is wrong with this.

libs/blocks/global-navigation/global-navigation.js
mokimo: You're right, the error would be caught - however I'd still suggest to stay consistent with the [fetch API](https://developer.mozilla.org/en-US/docs/Web/API/fetch)

> A fetch() promise only rejects when the request fails, for example, because of a badly-formed request URL or a network error. A fetch() promise does not reject if the server responds with HTTP status codes that indicate errors (404, 504, etc.). Instead, a then() handler must check the [Response.ok](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok) and/or [Response.status](https://developer.mozilla.org/en-US/docs/Web/API/Response/status) properties.

If a fetch fails because of a 404, we shouldn't throw an error in the method unless there is a legit error.
Also I'd assume that this could pollute the error logs even further, for something that sporadically happens.

My suggestion, and this is totally up to you / the tattva team:
```
    if (profileData.status !== 200) {
      lanaLog({ message: 'decorateProfile has failed to fetch profile data, e, tags: 'errorType=warn,module=gnav' });
      return
```

mokimo: IMO making the function fail, to trigger the method that calls `this.decorateProfile` to catch the error and hence log the lana log, is very error prone in case this behavior is ever refactored.

If you want to log something, you just log it - instead of throwing an error.

libs/blocks/global-footer/global-footer.js
mokimo: It's okay to log this if you see value in it for debugging purposes, I'd just not make the method fail inorder to log something.

libs/blocks/global-navigation/global-navigation.js
mokimo: I don't see a lot of value adding this to the utils unless there is a good use-case for tracking this and have people act based on the logs.

If this you see this enough in the gnav, having a gnav-util for this would be more appropriate. If you want to track when/how this happens just in the universalNav case, this is fine 👍

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -183,8 +183,15 @@ export function getExperienceName() {
 
 export function loadStyles(path) {
```
mokimo, line 184: Should you use the `loadStyles` method instead for the [unav?](https://github.com/adobecom/milo/pull/2402/files#diff-628bf5a25903978c5b36da4943ec813da171346d24bd64f80907f012a840b098R520)

Otherwise you duplicate the error handling behavior

mokimo:
@Deva309 I don't fully agree with logging networking errors because they can't be avoided or fixed. So even if they are in the logs, there's not much to do about them - other than gracefully failing or having a plan B.

However, if you see value in them - I'm 👍 on adding logs for those areas. The PR title now also makes more sense with what the PR changes now, appreciate you changing that!

libs/blocks/global-footer/global-footer.js
Deva309: Added a log as error type warning instead of method fail by throwing error.

libs/blocks/global-navigation/global-navigation.js
Deva309: Yes agree, throwing an error and making method fail will not be a good practice.
Added log near to the failure is the better way to handle this.

libs/blocks/global-navigation/global-navigation.js
Deva309: Instead of `loadStyle` from `utils`, I have used `loadStyles` from `utilities` of global navigation.

libs/blocks/global-navigation/utilities/utilities.js
Deva309: Same here as well.
To log it near to the failure will be a good approach and easy for debugging the error as well.

libs/blocks/global-navigation/utilities/utilities.js
Deva309: Done.
Using `loadStyles` instead of `loadStyle` to add log for the static assets like css.

Deva309:
> @Deva309 I don't fully agree with logging networking errors because they can't be avoided or fixed. So even if they are in the logs, there's not much to do about them - other than gracefully failing or having a plan B.
> 
> However, if you see value in them - I'm 👍 on adding logs for those areas. The PR title now makes more sense with what the PR changes. Appreciate you changing that!

I've reviewed the implementation of the `global-navigation` and `global-footer` components. We've wrapped root method calls with `logErrorFor` to capture most error scenarios, excluding issues related to some loading static assets. This approach ensures comprehensive error logging for current functionality and future additions.

Currently, I haven't identified any potential client-side errors beyond those covered by `logErrorFor`. In other words, any code-related errors (excluding network issues for static assets) should be logged.

@mokimo However, I'd appreciate it if you could clarify if there are any specific error scenarios logs you're concerned about, aside from network problems.

libs/blocks/global-footer/global-footer.js
```diff
@@ -72,12 +72,7 @@ class Footer {
     this.body = await fetchAndProcessPlainHtml({
       url,
       shouldDecorateLinks: false,
-    })
-      .catch((e) => lanaLog({
-        message: `Error fetching footer content ${url}`,
-        e,
-        tags: 'errorType=error,module=global-footer',
-      }));
+    });
```
mokimo, line 75: What happens if the `fetchAndProcessPlainHtml` has a JS error - is the catch not needed?

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -357,8 +376,8 @@ export async function fetchAndProcessPlainHtml({ url, shouldDecorateLinks = true
 
   const blocks = body.querySelectorAll('.martech-metadata');
   if (blocks.length) {
-    import('../../martech-metadata/martech-metadata.js')
-      .then(({ default: decorate }) => blocks.forEach((block) => decorate(block)));
+    const { default: decorate } = await import('../../martech-metadata/martech-metadata.js');
```
mokimo, line 379: Does this need to be blocking if before it wasnt?

mokimo (approved):

libs/blocks/global-footer/global-footer.js
Deva309: if there is any JS error in `fetchAndProcessPlainHtml` it will get logged with `logErrorFor` used with `decorateContent` in this case. Similar way it will get logged at other places as well.

libs/blocks/global-navigation/utilities/utilities.js
Deva309: With this non blocking behaviour it is printing error for clientid 'milo'.
Screenshot 2024-06-07 at 10 14 31 AM

But we need this log to be logged with clientId as 'feds-milo'. So fixing it by adding log with `catch` as errorType `warn`. Like shown below
 
Screenshot 2024-06-07 at 10 17 01 AM

Deva309:
@mokimo Fixed review comments.
Please review

mokimo (approved):

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -465,6 +465,11 @@ class Gnav {
     const profileData = await fetch(`https://${env.adobeIO}/profile`, { headers });
 
     if (profileData.status !== 200) {
+      lanaLog({
+        message: 'GNAV: decorateProfile has failed to fetch profile data',
+        e: `${profileData.statusText} url: ${profileData.url}`,
+        tags: 'errorType=warn,module=gnav',
```
robert-bogos, line 471: I think the `errorType=info` would be more appropriate in here. You can see the tags legend in [here](https://github.com/adobecom/milo/pull/2314). 
```suggestion
        tags: 'errorType=info,module=gnav',
```

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -330,6 +342,13 @@ export async function fetchAndProcessPlainHtml({ url, shouldDecorateLinks = true
     path = mepFragment.target;
   }
   const res = await fetch(path.replace(/(\.html$|$)/, '.plain.html'));
+  if (res.status !== 200) {
+    lanaLog({
+      message: 'Error in fetchAndProcessPlainHtml',
+      e: `${res.statusText} url: ${res.url}`,
+      tags: 'errorType=error,module=utilities',
```
robert-bogos, line 349: This also seems to fall under network issues, so I would go for:
```suggestion
      tags: 'errorType=info,module=utilities',
```

libs/blocks/global-navigation/global-navigation.js
Deva309: @robert-bogos 
Done.

libs/blocks/global-navigation/utilities/utilities.js
Deva309: @robert-bogos 
Done.

robert-bogos (commented):
Thank you for implementing the feedback! 👍

Can you add some unit tests for the changes? I'm asking because the `global-footer` and `global-navigation` have good code coverage and it would be beneficial to maintain this.

Deva309:
> Thank you for implementing the feedback! 👍
> 
> Can you add some unit tests for the changes? I'm asking because the `global-footer` and `global-navigation` have good code coverage and it would be beneficial to maintain this.

@robert-bogos 
I have added all the unit tests for the changes but for one change its not possible to add those changes as we re using `import` and cannot be mocked with `sinon`. I think this is exception and we can ignore covering this change. Please let me know if any other way we can achieve it.
Change not covered: 

> Thank you for implementing the feedback! 👍
> 
> Can you add some unit tests for the changes? I'm asking because the `global-footer` and `global-navigation` have good code coverage and it would be beneficial to maintain this.

@robert-bogos 
I have covered all the unit tests for my changes except for one scenario, specifically the `catch` method for `import`. Unfortunately, we cannot create a mock for an `import` using `Sinon`. If you have any suggestions on how to include this, please let me know. However, it appears this is an exception, and we might need to ignore coverage for this particular change as mentioned below:
Not covered for these lines: [libs/blocks/global-navigation/utilities/utilities.js#L384-L388](https://github.com/Deva309/milo/blob/758cda3cde62d48591c11202d76691c497dee8e1/libs/blocks/global-navigation/utilities/utilities.js#L384-L388)

salonijain3:
@robert-bogos  can you approve if the code reviews are resolved?

narcis-radu (approved):

narcis-radu:
@Deva309 - I see the patch coverage is 90%, you might want to add tests for the missing lines.

Deva309:
> > Thank you for implementing the feedback! 👍
> > Can you add some unit tests for the changes? I'm asking because the `global-footer` and `global-navigation` have good code coverage and it would be beneficial to maintain this.
> 
> @robert-bogos I have covered all the unit tests for my changes except for one scenario, specifically the `catch` method for `import`. Unfortunately, we cannot create a mock for an `import` using `Sinon`. If you have any suggestions on how to include this, please let me know. However, it appears this is an exception, and we might need to ignore coverage for this particular change as mentioned below: Not covered for these lines: [libs/blocks/global-navigation/utilities/utilities.js#L384-L388](https://github.com/Deva309/milo/blob/758cda3cde62d48591c11202d76691c497dee8e1/libs/blocks/global-navigation/utilities/utilities.js#L384-L388)

@narcis-radu 
This is the reason why I am not able to cover the patch coverage for this PR.

narcis-radu (approved):

robert-bogos (approved):

spadmasa:
Verified on the below url :
https://mwpw-140523-lana-logs--milo--deva309.hlx.page/drafts/devashish/lana-logs?martech=off&lanadebug
image
verified in https://splunk-us.corp.adobe.com/en-GB/app/app_log_always_never_assume/search
image

PR #2403, opened 2024-06-03
MWPW-142935 [Project PEP] PEP <> Geo Routing Modal Interactions

sharmrj:
* Both Georouting v1 and v2 now return a boolean value depending on whether or not they show the modal (the `showModal` function)
* The boolean value is stored in the config for PEP to use
* If the Georouting modal was shown, PEP will check to see if the modal has been dismissed every 200ms (this can be changed) in a **non blocking** manner (`setTimeout` will add the rechecking function to the event queue). Once the modal is no longer in the DOM, PEP's init function will fire
* If the Georouting modal was not shown for any reason, PEP will load normally

Resolves: [MWPW-142935](https://jira.corp.adobe.com/browse/MWPW-142935 )

**Test URLs:**
- After: https://pep-georouting--milo--adobecom.hlx.page/?martech=off
- QA: https://stage--cc--adobecom.hlx.page/products/photoshop/ai?milolibs=pep-georouting&skipPepEntitlements=true


libs/utils/utils.js
```diff
@@ -940,7 +940,7 @@ async function loadPostLCP(config) {
   const georouting = getMetadata('georouting') || config.geoRouting;
   if (georouting === 'on') {
     const { default: loadGeoRouting } = await import('../features/georoutingv2/georoutingv2.js');
-    await loadGeoRouting(config, createTag, getMetadata, loadBlock, loadStyle);
+    config.grmActive = await loadGeoRouting(config, createTag, getMetadata, loadBlock, loadStyle);
```
mokimo, line 943: Maybe `geoRoutingActive` ?

mokimo (approved):
Would be great if you can add some unit tests for this

bandana147 (approved):

spadmasa:
Verified the flow in the below url :

when the GRM is loaded the PEP modal is not seen
https://stage--cc--adobecom.hlx.live/products/firefly?milolibs=pep-georouting&skipPepEntitlements=true
image
after the modal close 
image

https://stage--cc--adobecom.hlx.page/products/photoshop/ai?milolibs=pep-georouting&skipPepEntitlements=true
image
image


after the geo routing cookie is set , PEP modal is seen loaded on page load only
image

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -35,6 +35,16 @@ const getIcon = (content) => {
   return icons.company;
 };
 
+const waitForClosedGRMThen = (loadPEP) => {
```
vhargrave, line 38: instead of polling the DOM, you should listen for milo:modal:closed on the window object. 
Whent that event is triggered you could then check if .locale-modal-v2 is no longer present.

sharmrj: The problem is that the global-navigation.js/webapp-prompt.js might not have loaded by the time the user gets rid of the GRM, especially on slow connections, so the event would be missed and we would have to poll anyway unless we set a global flag. I'm not sure this would be all that much more readable than what we have now. Is there a particular compunction we should have against polling for the GRM?

libs/features/georouting/georouting.js
```diff
@@ -137,27 +137,34 @@ export default async function loadGeoRouting(config, createTag, getMetadata) {
   const storedLocale = storedInter === 'us' ? '' : storedInter;
 
   const resp = await fetch(`${config.contentRoot ?? ''}/georouting.json`);
-  if (!resp.ok) return;
+  if (!resp.ok) return false;
```
vhargrave, line 140: all these changes are unnecessary

libs/utils/utils.js
```diff
@@ -940,7 +940,13 @@ async function loadPostLCP(config) {
...
     const { default: loadGeoRouting } = await import('../features/georoutingv2/georoutingv2.js');
-    await loadGeoRouting(config, createTag, getMetadata, loadBlock, loadStyle);
+    config.geoRoutingActive = await loadGeoRouting(
+      config,
+      createTag,
+      getMetadata,
+      loadBlock,
+      loadStyle,
+    );
   }
```
vhargrave, line 950: you don't need to make all these changes to the georouting files. 
You're already awaiting a response here . 
Just check if locale-modal-v2 has been added to the DOM right here
document.querySelector('.locale-modal-v2')

vhargrave (changes requested):

libs/features/webapp-prompt/webapp-prompt.js
vhargrave: you can check to see if .locale-modal-v2 is present in the DOM first and if it's available there, then add an event listening for it to close.

vhargrave: in general polling is worse for performance than just listening to the right event.

libs/utils/utils.js
vhargrave: actually in retrospect I don't even think you need to have any code here. Just add that listener I described in your webapp-prompt file, no?

vhargrave (changes requested):
looking much better thx! 
don't forget to stop polling though and listen to the right event.

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -43,8 +50,8 @@ class AppPrompt {
     this.getAnchorState = getAnchorState;
     this.id = this.promptPath.split('/').pop();
     this.elements = {};
-
-    this.init();
+    if (geoRoutingActive()) waitForClosedGRMThen(this.init);
```
vhargrave, line 53:
```suggestion
    if (geoRoutingActive()) document.addEventListener('milo:modal:closed', loadPEP);
```

vhargrave (approved):
Now we're cooking! 
That's looking pretty good. I think you could simplify your code a little more to be sending less bytes to the client if you wanted, but now it's just nitpicking. 
Very nice :)

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -35,6 +35,13 @@ const getIcon = (content) => {
   return icons.company;
 };
 
+const geoRoutingActive = () => !!document.querySelector('.locale-modal-v2')
+          || !!document.querySelector('.locale-modal');
+
+const waitForClosedGRMThen = (loadPEP) => {
+  window.addEventListener('milo:modal:closed', loadPEP);
```
vhargrave, line 42: sorry for doing this, but I just realized that it's possible for two dialogs to be open.
see this link - https://stage--cc--adobecom.hlx.page/products/photoshop/ai?milolibs=pep-georouting&skipPepEntitlements=true#langnav
we need to add a check here that geoRoutingActive() is now false.

vhargrave (changes requested):

libs/features/webapp-prompt/webapp-prompt.js
sharmrj: For some reason the `#langnav` short circuits my check and the pep dialog shows up anyway. Even while the GRM is not yet dismissed.

sharmrj: Also since the #langnav shows up late, it's now an open question whether the pep modal should show up anyway or wait until it's gone.

sharmrj: Also, I've discovered that when the event is emitted it won't necessarily trigger PEP's `init` method, since it's possible for the DOM to still exist at that point. I'm looking into why this might be.

vhargrave: completely agree. 
In my opinion PEP should not be appearing if any modal is open, but since that makes your task more complicated I didn't mention it.  
Maybe you can mention that to your PdM and if they deem it necessary then tackle that in a separate task.

vhargrave: oh and btw, you want to clean up your listener once you've triggered it once. 
https://dev.to/js_bits_bill/addeventlistener-once-js-bits-565d

sharmrj:
@vhargrave After some experimentation I believe it would be less bad to go with a pull (polling) approach over a push approach (events), simply because it allows us to reduce the surface area of the change. As described by my earlier comment and the video, the modal is not closed when the `milo:modal:closed` event is fired and its event listener callback is  called. While in many situations polling can cause a performance impact (e.g. polling multiple services for relatively rare changes), in other situations it's the preferred option (e.g. keyboard interrupts). In this particular situation:

- The polling is for a relatively short amount of time, and it ends after it finds what it's looking for.
- it happens only every 200ms and does not block the main thread
- execution time of the function checking the DOM isn't the bottleneck (even if it's done many times)
- The GRM (or another modal) will be open during the polling, so we should see minimal impact (if at all) on users even on the long tail of devices (the state of the page in the background is less relevant)

libs/features/webapp-prompt/webapp-prompt.js
vhargrave: @sharmrj it seems to me that the right solution would be to change the modal.js so that the close event is sent after removing the modal.  
@JasonHowellSlavin can you weigh in if you see issues with moving the window.dispatchEvent of milo:modal:closed after removing the modal?

JasonHowellSlavin: Hi @vhargrave, I do think both you and @sharmrj have identified a key issue with the `milo:modal:closed` dispatch being placed too high in the function and not reflecting what is actually occurring. I don't see an issue with moving it down, but since there are a few other files that rely on it, it might be good to raise that as an issue and specifically test for those implementations. 

It also looks like @sharmrj has done some experimentation and found the performance impact of the current code to be negligible.

vhargrave: @JasonHowellSlavin ok, thanks for confirming 👍

vhargrave:
> @vhargrave After some experimentation I believe it would be less bad to go with a pull (polling) approach over a push approach (events), simply because it allows us to reduce the surface area of the change. As described by my earlier comment and the video, the modal is not closed when the `milo:modal:closed` event is fired and its event listener callback is called.
> 
> While in many situations polling can cause a performance impact (e.g. polling multiple services for relatively rare changes), in this particular situation:
> 
> * The polling is for a relatively short amount of time, and it ends after it finds what it's looking for.
> * it happens only every 200ms and does not block the main thread
> * execution time of the function checking the DOM isn't the bottleneck (even if it's done many times)
> * The GRM (or another modal) will be open during the polling, so we should see minimal impact (if at all) on users even on the long tail of devices (the state of the page in the background is less relevant)

I do see the point regarding performance, but for me it's also about architecture. If there's an event that one can listen to then it's better to listen for an event than to poll. Even kicking off polling after the event happens is better than polling from the get go.

sharmrj:
> > @vhargrave After some experimentation I believe it would be less bad to go with a pull (polling) approach over a push approach (events), simply because it allows us to reduce the surface area of the change. As described by my earlier comment and the video, the modal is not closed when the `milo:modal:closed` event is fired and its event listener callback is called.
> > While in many situations polling can cause a performance impact (e.g. polling multiple services for relatively rare changes), in this particular situation:
> > 
> > * The polling is for a relatively short amount of time, and it ends after it finds what it's looking for.
> > * it happens only every 200ms and does not block the main thread
> > * execution time of the function checking the DOM isn't the bottleneck (even if it's done many times)
> > * The GRM (or another modal) will be open during the polling, so we should see minimal impact (if at all) on users even on the long tail of devices (the state of the page in the background is less relevant)
> 
> I do see the point regarding performance, but for me it's also about architecture. If there's an event that one can listen to then it's better to listen for an event than to poll. Even kicking off polling after the event happens is better than polling from the get go.

If that's acceptable I can change the implementation to start polling only after the event is emitted.

vhargrave:
> > > @vhargrave After some experimentation I believe it would be less bad to go with a pull (polling) approach over a push approach (events), simply because it allows us to reduce the surface area of the change. As described by my earlier comment and the video, the modal is not closed when the `milo:modal:closed` event is fired and its event listener callback is called.
> > > While in many situations polling can cause a performance impact (e.g. polling multiple services for relatively rare changes), in this particular situation:
> > > 
> > > * The polling is for a relatively short amount of time, and it ends after it finds what it's looking for.
> > > * it happens only every 200ms and does not block the main thread
> > > * execution time of the function checking the DOM isn't the bottleneck (even if it's done many times)
> > > * The GRM (or another modal) will be open during the polling, so we should see minimal impact (if at all) on users even on the long tail of devices (the state of the page in the background is less relevant)
> > 
> > 
> > I do see the point regarding performance, but for me it's also about architecture. If there's an event that one can listen to then it's better to listen for an event than to poll. Even kicking off polling after the event happens is better than polling from the get go.
> 
> If that's acceptable I can change the implementation to start polling only after the event is emitted.

@sharmrj for me it's not perfect, but it's good enough yeah. And it means that you don't need to make any changes to the event emit that would require thorough testing. 👍

sharmrj:
> > > > @vhargrave After some experimentation I believe it would be less bad to go with a pull (polling) approach over a push approach (events), simply because it allows us to reduce the surface area of the change. As described by my earlier comment and the video, the modal is not closed when the `milo:modal:closed` event is fired and its event listener callback is called.
> > > > While in many situations polling can cause a performance impact (e.g. polling multiple services for relatively rare changes), in this particular situation:
> > > > 
> > > > * The polling is for a relatively short amount of time, and it ends after it finds what it's looking for.
> > > > * it happens only every 200ms and does not block the main thread
> > > > * execution time of the function checking the DOM isn't the bottleneck (even if it's done many times)
> > > > * The GRM (or another modal) will be open during the polling, so we should see minimal impact (if at all) on users even on the long tail of devices (the state of the page in the background is less relevant)
> > > 
> > > 
> > > I do see the point regarding performance, but for me it's also about architecture. If there's an event that one can listen to then it's better to listen for an event than to poll. Even kicking off polling after the event happens is better than polling from the get go.
> > 
> > 
> > If that's acceptable I can change the implementation to start polling only after the event is emitted.
> 
> @sharmrj for me it's not perfect, but it's good enough yeah. And it means that you don't need to make any changes to the event emit that would require thorough testing. 👍

@vhargrave Cool thanks for all the great feedback. Could you take one more look at the code?

vhargrave (approved):
looks good!

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -43,8 +53,13 @@ class AppPrompt {
     this.getAnchorState = getAnchorState;
     this.id = this.promptPath.split('/').pop();
     this.elements = {};
-
-    this.init();
+    if (modalsActive()) {
+      window.addEventListener(
+        'milo:modal:closed',
+        () => waitForClosedModalsThen(this.init),
```
overmyheadandbody, line 59: Could we not add `waitForClosedModalsThen` as a class method and avoid passing the `init` method?

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -35,6 +35,16 @@ const getIcon = (content) => {
   return icons.company;
 };
 
+const modalsActive = () => !!document.querySelector('.dialog-modal');
+
+const waitForClosedModalsThen = (loadPEP) => {
+  if (modalsActive()) {
+    setTimeout(() => waitForClosedModalsThen(loadPEP), 200);
```
overmyheadandbody, line 42: Could we not use a `setInterval` to avoid the recursive calls?

robert-bogos (approved):
Looks good from my side 👍

overmyheadandbody (commented):
Is the reverse a concern as well? As in: PEP prompt is open, but the georouting modal pops up. The user wouldn't be able to dismiss the PEP prompt and would get redirected, potentially causing confusion. It might be valuable to pause the PEP timer or find some alternative solutions.

libs/features/webapp-prompt/webapp-prompt.js
sharmrj: I like to be explicit about what I'm passing to functions, and I like the CPS-ish semantics of writing "do an action and then do this - " by explicitly passing the function (so long as we aren't doing long chains of it, of course). Though if you think the `init` method shouldn't be passed I can change it (my reasoning for why it's ok is that it's not a private method; anyone can call it).

libs/features/webapp-prompt/webapp-prompt.js
sharmrj: I'd prefer to not use setInterval since we'd have to keep track of the `setInterval` ID and clean it up after with `clearInterval`. And `waitForClosedModalsThen` isn't recursive in the sense that it's growing the callstack indefinitely, it, like `setInterval`, adds a function to the task queue, and in that sense is (to us, at least) indistinguishable in behavior from `setInterval`. The question is of whether we prefer the extra code to clean up the interval or the extra code for adding to the event queue.

sharmrj:
> Is the reverse a concern as well? As in: PEP prompt is open, but the georouting modal pops up. The user wouldn't be able to dismiss the PEP prompt and would get redirected, potentially causing confusion. It might be valuable to pause the PEP timer or find some alternative solutions.

We don't need to worry about the PEP modal being loaded before the GRM, because we have an `await loadGeoRouting(...)` before we even begin loading the global header (in `loadPostLCP`). In general, modals that open after PEP has already begun its countdown won't stop the countdown. If that needs to change we'll revisit it later.

spadmasa:
Validation done on the pr in below url :
https://stage--cc--adobecom.hlx.page/products/photoshop/ai?milolibs=pep-georouting&skipPepEntitlements=true

spadmasa:
Validation done on below feature url : 
https://main--cc--adobecom.hlx.page/creativecloud/animation/testdoc/peprollout?milolibs=pep-georouting&skipPepEntitlements=true
image
until the geo popup is closed the prompt is not seen
image

PR #2406, opened 2024-06-03
MWPW-151513: Search results vanished when user clicks on Marquee CTA:Start free trail

mirafedas:
When we search for some cards on the catalog page, for example, 'acrobat', the hash `#search=acrobat` gets added to the URL, and analogically for other filter params. When we open a modal in the marquee, we overwrite the hash value, and when we close the modal, the hash gets removed, so we lose the search results. To avoid this, before opening the modal, we remember the previous hash and restore it after it gets closed.

Related tacocat PR: https://git.corp.adobe.com/wcms/tacocat.js/pull/615 (the change from it is already included in /deps in this PR).

Resolves: [MWPW-151513](https://jira.corp.adobe.com/browse/MWPW-151513)

**Test URLs:**
CC:
- Before: https://main--cc--adobecom.hlx.live/products/catalog?martech=off#search=acrobat
- After: https://main--cc--adobecom.hlx.live/products/catalog?milolibs=MWPW-151513-search-results--milo--mirafedas&martech=off#search=acrobat

Milo:
- Before: https://main--milo--mirafedas.hlx.page/drafts/mirafedas/all-modals?martech=off#samsung
- After: https://mwpw-151513-search-results--milo--mirafedas.hlx.page/drafts/mirafedas/all-modals?martech=off#samsung


yesil:
@mirafedas thanks for the fix, can you add a unit test?

libs/blocks/modal/modal.js
```diff
@@ -245,5 +250,8 @@ window.addEventListener('hashchange', (e) => {
   } else {
     const details = findDetails(window.location.hash, null);
     if (details) getModal(details);
+    if (e.oldURL?.includes('#search=')) {
```
yesil, line 253: Milo should not be opinionated about the deep link values. So I don't expect to see `search=`.

Also, when I change the deeplink values by clicking on the category filters, they are not captured.

yesil (changes requested):

mirafedas:
@yesil I added tacocat change in this PR: https://git.corp.adobe.com/wcms/tacocat.js/pull/615 , and included it in this PR

yesil (changes requested):
asked for a change in the related tacocat PR.

yesil:
@mirafedas the tacocat PR for merch-card-collection has been merged, please rebuild it from the develop branch and update here.

yesil:
@mirafedas  you also need to ship sidenav in CC so that the search text doesn't vanish when you click the modal link.

mirafedas:
@yesil I have a pending CC PR for updating sidenav, it's in a scope of different task, but it will ship the latest sidenav to CC: https://github.com/adobecom/cc/pull/335

npeltier (approved):

libs/blocks/modal/modal.js
```diff
@@ -65,6 +66,10 @@ export function closeModal(modal) {
   const hashId = window.location.hash.replace('#', '');
   if (hashId === modal.id) window.history.pushState('', document.title, `${window.location.pathname}${window.location.search}`);
   isDelayedModal = false;
+  if (prevHash) {
```
Axelcureno, line 69: How does this work? If _prevHash_ is defined as empty string in L15, wouldn't this _if_ always pass as `true`?

mirafedas: We set the `prevHash` value on line 263 in the event listener which gets executed on the hash change. So when the `closeModal` function gets called, the `prevHash` may have a value

yesil (approved):

PR #2412, opened 2024-06-03
MWPW-144801 - [Hero-Marquee] 🆕 block

ryanmparrish:
As an author in Milo, I want the ability to add a `Hero Marquee` to a Milo page. 

The hero marquee is designed to have flexibility over content order, content types and viewport layout options. I have a well documented draft below that outlines the block features, authoring patterns and variants being used. 
Here's a brief on these offerings. 

- Author content types and ordering via [con-rows](https://rparrish-marquee-hero--milo--adobecom.hlx.page/drafts/rparrish/hero/hero#con-block-row-types?martech=off) block model 
-- `['list', 'qrcode', 'lockup', 'text', 'bgcolor', 'supplemental']`
- Ability to have background position options - `bg-top-mobile, bg-bottom-tablet, etc`
- Support for `breakpoint-themes.css` - This allows light as default and tablet-dark for theme context changing per viewport. (I made this a separate css file for sharing w/ other blocks) 
- Moved the `handleObjectFit()` func. into a shared `decorate.js` file so this and the [bricks] block can both use it. 

Resolves: [MWPW-144801](https://jira.corp.adobe.com/browse/MWPW-144801)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rparrish/hero/hero?martech=off
- After: https://rparrish-marquee-hero--milo--adobecom.hlx.page/drafts/rparrish/hero/hero?martech=off

Additional testing pages
- Hero Kit (Testing bg-top-* and breakpoint-theme offerings)
https://rparrish-marquee-hero--milo--adobecom.hlx.page/drafts/rparrish/hero/hero-kit?martech=off
- Hero Milestone (real world examples)
https://rparrish-marquee-hero--milo--adobecom.hlx.page/drafts/rparrish/hero/hero-milestone?martech=off

- Smoke testing. Thanks Kyung
https://rparrish-marquee-hero--milo--adobecom.hlx.page/drafts/klee/test-hero?martech=off

Document vs Frontend

image

---

🐘 in the room?
 - Localization - This `[ key | value ]` document authoring pattern is a newish paradigm, similar to quiz-marquee, I have the rows 1st cell textContent looking for a contains `con-block-row-` in the blocks .js. My hope is that we can include a similar pattern to omit these keys cells from any localization scripts, or is there a pattern in place that would already include these that I could adapt this document model to? Similar to `#_dnt` @janaki-r-bhagwath @sukamat @jenssingler  - Do you have any thoughts?

- MEP - I'd love to see if this block has any usage/conflicts or issues if this was going to be considered for MEP integrations? @vivgoodrich  - LMK any initial gut reactions.


SilviuLCF:
@ryanmparrish Added a comment related to a JS error on https://jira.corp.adobe.com/browse/MWPW-144801 

Also  codecov/project checks are not passing

libs/blocks/brick/brick.js
```diff
@@ -1,11 +1,13 @@
-import { decorateTextOverrides, decorateBlockText, decorateBlockBg, decorateIconStack, decorateButtons } from '../../utils/decorate.js';
+import {
+  // eslint-disable-next-line max-len
```
overmyheadandbody, line 2: I'd say it's better to simply add returns after each method than add 35 characters to disable a rule. Would make things more readable, would not go against any formatting rules and would not impact the size 🤷

libs/blocks/hero-marquee/hero-marquee.js
```diff
@@ -0,0 +1,208 @@
...
+    if (!elTaps.length) return;
+    const parts = elClass.split('-');
+    el.classList.add(`${parts[1]}-${parts[0]}`);
+    el.classList.remove(elClass);
+  });
+}
+
+function decorateList(el, classes) {
+  el.classList.add('body-l');
+  const listItems = el.querySelectorAll('ul li', 'ol li');
```
overmyheadandbody, line 24: Do we have `li` elements outside of `ol` or `ul`? Could we not just look for `li`?

libs/blocks/hero-marquee/hero-marquee.js
```diff
@@ -0,0 +1,208 @@
...
+}
+
+function loadContentType(el, key, classes) {
+  if (classes !== undefined && classes.length) el.classList.add(...classes);
+  if (key === 'bgcolor') decorateBg(el);
+  if (key === 'list') decorateList(el, classes);
+  if (key === 'lockup') decorateLockupRow(el);
+  if (key === 'qrcode') decorateQr(el);
+  if (key === 'supplemental') decorateSup(el, classes);
+  if (key === 'text') decorateText(el, classes);
```
overmyheadandbody, line 106: Was `switch` avoided intentionally? Looks like a good use-case for it

libs/features/icons/icons.js
```diff
@@ -63,12 +63,12 @@ export default async function loadIcons(icons, config) {
...
       if (parent.lastChild === icon) {
-        icon.classList.add('margin-left');
+        icon.classList.add('margin-inline-start');
       } else if (parent.firstChild === icon) {
-        icon.classList.add('margin-right');
+        icon.classList.add('margin-inline-end');
         if (parent.parentElement.tagName === 'LI') parent.parentElement.classList.add('icon-list-item');
       } else {
-        icon.classList.add('margin-left', 'margin-right');
+        icon.classList.add('margin-inline-start', 'margin-inline-end');
```
overmyheadandbody, line 71: I think these have already been updated, maybe rebase would eliminate this from the changes https://github.com/adobecom/milo/blob/stage/libs/features/icons/icons.js#L66-L71

libs/styles/breakpoint-theme.css
```diff
@@ -0,0 +1,193 @@
...
+.con-block.static-links[class*="light-"] a:not(.con-button):hover,
+.static-links .con-block[class*="light-"] a:not(.con-button):hover {
+  color: inherit;
+}
+
+/* mobile ONLY */
+@media (max-width: 600px) {
+  /* mobile-only theme */
+  .con-block.light-mobile-only { color: var(--text-color); }
+  .con-block.dark-mobile-only {  color: var(--color-white); }
```
overmyheadandbody, line 19: Nit of the year: double spacing before `color` 😅 Same for tablet and desktop
```suggestion
  .con-block.dark-mobile-only { color: var(--color-white); }
```

libs/styles/breakpoint-theme.css
```diff
@@ -0,0 +1,193 @@
...
+
+  .con-block.light-mobile-only a:not(.con-button) { color: var(--link-color); }
+  .con-block.dark-mobile-only a:not(.con-button) { color: var(--link-color-dark); }
+
+  .con-block.light-mobile-only a:not(.con-button):hover,
+  .con-block.light-mobile-only a:not(.con-button):focus,
+  .con-block.light-mobile-only a:not(.con-button):active { color: var(--link-hover-color); }
```
overmyheadandbody, line 26: Could we trim these down with `:is`?
```suggestion
  .con-block.light-mobile-only a:not(.con-button):is(:hover, :focus, :active) { color: var(--link-hover-color); }
```

libs/styles/breakpoint-theme.css
```diff
@@ -0,0 +1,193 @@
...
+.static-links .con-block[class*="light-"] a:not(.con-button),
+.con-block.static-links[class*="dark-"] a:not(.con-button):hover,
+.static-links .con-block[class*="dark-"] a:not(.con-button):hover,
+.con-block.static-links[class*="light-"] a:not(.con-button):hover,
+.static-links .con-block[class*="light-"] a:not(.con-button):hover {
+  color: inherit;
+}
+
+/* mobile ONLY */
+@media (max-width: 600px) {
```
overmyheadandbody, line 16: Since tablet applies starting 600, shouldn't this be `599px`?

libs/styles/breakpoint-theme.css
```diff
@@ -0,0 +1,193 @@
...
+  .con-block.light-tablet a:not(.con-button) { color: var(--link-color); }
+  .con-block.dark-tablet a:not(.con-button) { color: var(--link-color-dark); }
+
+  .con-block.light-tablet a:not(.con-button):hover,
+  .con-block.light-tablet a:not(.con-button):focus,
+  .con-block.light-tablet a:not(.con-button):active { color: var(--link-hover-color); }
+  
+  .con-block.dark-tablet a:not(.con-button):hover,
+  .con-block.dark-tablet a:not(.con-button):focus,
+  .con-block.dark-tablet a:not(.con-button):active  { color: var(--link-hover-color-dark); }
```
overmyheadandbody, line 120:
```suggestion
  .con-block.dark-tablet a:not(.con-button):active { color: var(--link-hover-color-dark); }
```

libs/styles/styles.css
```diff
@@ -126,6 +126,9 @@
   --icon-size-s: 32px;
   --icon-size-xs: 24px;
   --icon-size-xxs: 16px;
+
+  /* z-index */
+  --above-all: 9000; /* Used for page tools that overlay page content */
```
overmyheadandbody, line 131: I'm not sold on this idea. What if there are multiple elements competing? I usually like to be between the `1-10` range, except when 3rd parties declare very high values that can't be overwritten otherwise.

libs/styles/styles.css
```diff
@@ -550,6 +553,8 @@ span.icon.margin-right { margin-right: 8px; }
 
 span.icon.margin-left { margin-left: 8px; }
 
+span.icon.margin-inline-start { margin-inline-start: 8px; }
```
overmyheadandbody, line 556: This may have already been contributed in https://github.com/adobecom/milo/blob/stage/libs/styles/styles.css#L562

libs/blocks/hero-marquee/hero-marquee.css
```diff
@@ -0,0 +1,581 @@
+:root {
+  --s-min-height: 248px; /* 360px */
+  --m-min-height: 448px; /* 560px */
+  --l-min-height: 588px; /* 700px */
```
overmyheadandbody, line 4: Is this universally true from a Consonant perspective? If not, they might be better scoped to the `.hero-marquee` parent element.

libs/blocks/hero-marquee/hero-marquee.css
```diff
@@ -0,0 +1,581 @@
...
+  gap: var(--spacing-m);
+  width: 100%;
+  overflow: hidden;
+  min-height: var(--m-min-height);
+  justify-content: center;
+  align-items: center;
+}
+
+.hero-marquee.s-min-height { min-height: var(--s-min-height);}
+.hero-marquee.l-min-height { min-height: var(--l-min-height);}
```
overmyheadandbody, line 23: Linter might complain
```suggestion
.hero-marquee.s-min-height { min-height: var(--s-min-height); }
.hero-marquee.l-min-height { min-height: var(--l-min-height); }
```

libs/blocks/hero-marquee/hero-marquee.css
```diff
@@ -0,0 +1,581 @@
...
+}
+
+.hero-marquee .background-split picture img {
+  object-fit: cover;
+  height: 100%;
+  width: 100%;
+}
+
+.hero-marquee.fg-media-cover picture {
+  line-height: 0em;
```
overmyheadandbody, line 56: Is the unit even needed in this case?

libs/blocks/hero-marquee/hero-marquee.css
```diff
@@ -0,0 +1,581 @@
...
+.hero-marquee .lockup-area { 
+  font-weight: 700;
+  display: flex;
+  align-items: center;
+  gap: var(--spacing-xs);
+  margin: 0 0 var(--spacing-xxs); 
+  font-size: var(--type-body-m-size);
+  line-height: var(--type-body-m-lh);
+  text-transform: initial;
+  white-space: nowrap
```
overmyheadandbody, line 77:
```suggestion
  white-space: nowrap;
```

libs/blocks/hero-marquee/hero-marquee.css
```diff
@@ -0,0 +1,581 @@
...
+} 
+
+.hero-marquee .lockup-area picture { 
+  line-height: 0;
+  display: block;
+}
+
+.hero-marquee .lockup-area picture img {
+  width: auto;
```
overmyheadandbody, line 86: Is the image limited to its container somewhere else?

libs/blocks/hero-marquee/hero-marquee.css
```diff
@@ -0,0 +1,581 @@
...
+.hero-marquee .l-icon .lockup-area,
+.hero-marquee.l-icon .lockup-area {
+  font-size: var(--type-body-m-size);
+  line-height: var(--type-body-m-lh);
+}
+
+.hero-marquee .lockup-area picture img,
+.hero-marquee .l-icon .lockup-area picture img,
+.hero-marquee.l-icon .lockup-area picture img {
+  min-width:var(--icon-size-l);
```
overmyheadandbody, line 122:
```suggestion
  min-width: var(--icon-size-l);
```

libs/blocks/hero-marquee/hero-marquee.css
```diff
@@ -0,0 +1,581 @@
...
+
+.hero-marquee .m-icon .lockup-area,
+.hero-marquee.m-icon .lockup-area {
+  font-size: var(--type-body-s-size);
+  line-height: var(--type-body-s-lh);
+}
+
+.hero-marquee .m-icon .lockup-area picture img, 
+.hero-marquee.m-icon .lockup-area picture img {
+  min-width:var(--icon-size-m);
```
overmyheadandbody, line 134:
```suggestion
  min-width: var(--icon-size-m);
```

libs/blocks/hero-marquee/hero-marquee.css
```diff
@@ -0,0 +1,581 @@
...
+
+.hero-marquee .xl-icon .lockup-area,
+.hero-marquee.xl-icon .lockup-area {
+  font-size: var(--type-body-xl-size);
+  line-height: var(--type-body-xl-lh);
+}
+
+.hero-marquee .xl-icon .lockup-area picture img,
+.hero-marquee.xl-icon .lockup-area picture img {
+  min-width:var(--icon-size-xl);
```
overmyheadandbody, line 146:
```suggestion
  min-width: var(--icon-size-xl);
```

libs/blocks/hero-marquee/hero-marquee.css
```diff
@@ -0,0 +1,581 @@
...
+
+.hero-marquee.center .action-area,
+.hero-marquee.center .lockup-area { 
+  justify-content: center; 
+}
+
+.hero-marquee .norm p { margin: var(--spacing-xs) 0; }
+.hero-marquee .main-copy p { margin: 0; }
+.hero-marquee .main-copy p:only-child { margin: 0; }
+.hero-marquee .main-copy [class^="heading"] { margin: 0; }
```
overmyheadandbody, line 163: Would it be worth grouping these 3, since they share the same style?

libs/blocks/hero-marquee/hero-marquee.css
```diff
@@ -0,0 +1,581 @@
...
+.hero-marquee .row-qrcode .qr-code-img {
+  display: none;
+}
+
+.hero-marquee .row-qrcode .google-play a,
+.hero-marquee .row-qrcode .app-store a {
+  width: 135px;
+  height: 40px;
+  display: inline-flex;
+  text-indent:-999em;
```
overmyheadandbody, line 264: I have mixed feeling about this strategy of hiding the text. Could we set the color to transparent or remove the text in JS?
```suggestion
  text-indent: -999em;
```

libs/blocks/hero-marquee/hero-marquee.css
```diff
@@ -0,0 +1,581 @@
...
+}
+
+.hero-marquee li.icon-item span.icon {
+  position: absolute;
+  left: -28px;
+}
+
+html[dir="rtl"] .hero-marquee li.icon-item span.icon {
+  left: unset;
+  right: -38px;
```
overmyheadandbody, line 308: Is this intended to be larger than LTR? I originally thought it may be an icon quirk, but the spacing different is noticeable.

libs/blocks/hero-marquee/hero-marquee.css
```diff
@@ -0,0 +1,581 @@
...
+  display: none;
+}
+
+.hero-marquee.asset-left > .foreground.cols-2 > .asset {
+  order: 2;
+}
+
+
+/* CLS */
+
```
overmyheadandbody, line 326: Is this still useful?

libs/blocks/hero-marquee/hero-marquee.css
```diff
@@ -0,0 +1,581 @@
...
+    padding-bottom: 0;
+  }
+
+  /* con-vars support */
+  .hero-marquee.bg-stack-mobile .background,
+  .hero-marquee.bg-stack-top-mobile .background,
+  .hero-marquee.bg-stack-bottom-mobile .background,
+  .hero-marquee.bg-stack-mobile-tablet .background,
+  .hero-marquee.bg-stack-top-mobile-tablet .background,
+  .hero-marquee.bg-stack-bottom-mobile-tablet .background   {
```
overmyheadandbody, line 355: Maybe `:is` can shorten this one too

libs/blocks/hero-marquee/hero-marquee.css
```diff
@@ -0,0 +1,581 @@
...
+
+  .hero-marquee.bg-stack-bottom-mobile .background {
+    order: 2;
+  }
+  
+  .hero-marquee.fg-media-hidden-mobile .foreground .asset,
+  .hero-marquee.fg-media-hidden-mobile-tablet .foreground .asset, 
+  .hero-marquee.fg-media-hidden-mobile .foreground-media,
+  .hero-marquee.fg-media-hidden-mobile-tablet .foreground-media { display: none; }
+
```
overmyheadandbody, line 391:
```suggestion
```

overmyheadandbody (commented):
Mostly nits, some more important than others 🙂 

Great job on this! I'll reserve another session for some visual testing, reviewing this PR is a handful already 😅

libs/blocks/brick/brick.js
ryanmparrish: Agree, simple return is better. I've updated this in bricks.js and hero-marquee.js

libs/styles/breakpoint-theme.css
ryanmparrish: Ill check my linting, This should have been flagged. These are resolved now.

libs/blocks/hero-marquee/hero-marquee.js
ryanmparrish: No other `li` elements should be considered, a a simple `li` works fine. Updated

libs/blocks/hero-marquee/hero-marquee.css
ryanmparrish: 🤷 I think it's just a typo, I've made em the same `28` and verified the visual alignment Thanks

libs/blocks/hero-marquee/hero-marquee.css
ryanmparrish: Good thought, I'm a little weary of this strategy as well. 
Since this is also used for `daa-ll` attributes, I've just set the `color: transparent`.

libs/blocks/hero-marquee/hero-marquee.js
ryanmparrish: No, not intention, it just ended up being more concise this way. 
I've updated this to use a `switch` instead which is likely more performant, but does increase line count.

libs/blocks/hero-marquee/hero-marquee.css
ryanmparrish: True, Could also use a [class^="bg-stack-"]. I've updated to use :is and shortened it up a bit. There may be some other css selectors that could be optimized, ill keep an eye out.

libs/features/icons/icons.js
ryanmparrish: That did it 👍

libs/blocks/hero-marquee/hero-marquee.css
ryanmparrish: Yes, there is an inherited LH with the `picture` element that can cause extra element whitespace depending on position. (usually a slight extra margin below element) this just clears that out so img dimensions are accurate.

libs/blocks/hero-marquee/hero-marquee.css
ryanmparrish: No, this selector does nothing and has been removed.

libs/styles/breakpoint-theme.css
ryanmparrish: Looks like this is set on a number of files as is, so i'm leaving this alone for now.

libs/styles/breakpoint-theme.css
ryanmparrish: Certainly, I've gone thru this css and reduced some bloat using `:is` Tested and checks out.

libs/styles/styles.css
ryanmparrish: oops yep, your right. Thanks

libs/styles/styles.css
ryanmparrish: N/A: This was from another contributor. I've rebased and not introducing style.css changes now

ryanmparrish:
@overmyheadandbody I've address everything you brought up and resolved / left comments on any questions. LMK if you see anything else. 
@SilviuLCF - I believe I have addressed the error you reported. Have another quick look if you could and LMK. 

Thanks for the feedback

libs/blocks/hero-marquee/hero-marquee.css
```diff
@@ -0,0 +1,554 @@
...
+/* mobile ONLY */
+@media (max-width: 600px) {
+  .hero-marquee .con-button {
+    display: block;
+    text-align: center;
+  }
+  
+  .hero-marquee.fg-media-stack-top-mobile,
+  .hero-marquee.fg-media-stack-top-mobile-tablet {
```
rgclayton, line 312: The discussion going on [here](https://github.com/orgs/adobecom/discussions/2420#discussioncomment-9692041) made me think of this.

If there is a need to differentiate between background and foreground images, I think we should default to one and call out the other.

- `media`
- `bg-media`

Seeing `fg` threw my brain for a loop. I had no idea what that was. Seeing `bg` is more recognizable as background. Maybe that's just me? Any other opinions?

I do think adding `stack` is overly verbose. I think the position name says it all. 

- `media-top-mobile`
-  `bg-media-top-mobile` 

 tell me what I need to know about those options.
 
 I also think `mobile-tablet` in `fg-media-stack-top-mobile-tablet` reads odd. 
 Is this an effort to combine the two? 
 
 I think the more common pattern in use is to declare both. `media-top-mobile` AND `media-top-tablet`

rgclayton (commented):
I think we need to take a step back and rethink these class names

- `fg-media-stack-top-mobile` 
- `fg-media-stack-top-mobile-tablet`
-  `bg-media-stack-top-mobile`
- etc.

I'm assuming some variation of these names show up as variant names in authoring?

libs/blocks/hero-marquee/hero-marquee.css
elan-tbx: +1 for `bg` over `fg`. We're already using `bg` in [styles.css](https://github.com/adobecom/milo/blob/cde7824f4dc0e07b041dfebc03efa33f5e6523cc/libs/styles/styles.css#L453)

elan-tbx: ...eating my words because apparently `fg` is also used in [home.css](https://github.com/adobecom/milo/blob/cde7824f4dc0e07b041dfebc03efa33f5e6523cc/libs/blocks/home/home.css#L45). Still though, bg does seem like much more common of a convention than fg.

rgclayton: Wait, [home.css](https://github.com/adobecom/milo/blob/cde7824f4dc0e07b041dfebc03efa33f5e6523cc/libs/blocks/home/home.css#L45).? What does the home block in Milo do? Not sure that should be an guiding example.

elan-tbx: I too have this question, haha. Maybe some analytics could tell us if it's even being used anywhere?

ryanmparrish: @rgclayton && @elan-tbx Thanks for the thoughts on these things. After thinking thru these patterns, I am going to drop all prefixed naming conventions of `fg-` (foreground) and it's implied that media is targeting this content type. I am also going to omit the `stack-` part of these variants as `top` or `bottom` is enough to distinguish. 
I am going to keep the `bg-` prefix as thats pretty universally known to represent background. I'll comment on this task once that change is made and reflected in this feature branch and supporting drafts.

ryanmparrish:
FYI, I have updated this branch and referenced docs to address the naming convention things that were mentioned. 
 - removed variant names `fg-` and `stack-`
 - removed shorthand `mobile-tablet` vp selectors
 
 @rgclayton @elan-tbx Thanks for the thoughts and cross thread discussion. 
 Please LMK if you have any other concerns or suggestions.

libs/blocks/hero-marquee/hero-marquee.css
ryanmparrish: ✅ This branch and ref. docs have been updated.

rgclayton (approved):

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-144801

meganthecoder:
@ryanmparrish A lot of really good stuff here! Thanks for showing me some of the features.

I'm curious to know, what does GWP think of the `key | value` system?  It seems like a big shift in authoring, and I'm wondering what their take is. It looks quite complex and not as human-readable compared to how we usually do things.

Also, what's the performance like?

ryanmparrish:
@meganthecoder  - Thanks for the look. I did a pretty basic performance review comparing the [marquee](https://main--milo--adobecom.hlx.page/drafts/rparrish/hero/marquee-performance) to the [hero-marquee](https://rparrish-marquee-hero--milo--adobecom.hlx.page/drafts/rparrish/hero/hero-performance) and I only see marginal differences. 

Screen Shot 2024-06-14 at 1 08 48 PM

I have also communicated this authoring pattern to GWP/#design-publishing channel around as well as raised this up to a group dealing w/ LOC (@mentions in the desc) 
I'll share any details in those conversations. Thanks

ryanmparrish:
@meganthecoder Just a quick update. I spoke w/ some of the loc folks and confirmed we can adopt this table model into their workflow. I've also shared this block w/ design-publishing to get some more eyes from authoring and they so far seem pretty on board w/ these new patterns. I'll post up if there's any feedback. Also I re-ran the PSI check as it was failing, but it is passing now.

PR #2413, opened 2024-06-03
MWPW-147594: Consumers mappings for origin not being sent when using sidekick

sanrai:
If an author tries sending content to caas from a consumer of milo, the origin was always upper case so the mapping never matched and an uppercase origin would always be sent. 

This was originally thought to be fixed in this PR: https://github.com/adobecom/milo/pull/2244 but it ended up not fixing the issue.

Resolves: [MWPW-147594](https://jira.corp.adobe.com/browse/MWPW-147594)

**Test URLs:**
- Before:
https://mwpw-147594-2--milo--adobecom.hlx.page/drafts/sanrai/foo?martech=off

- After:


tools/send-to-caas/send-utils.js
```diff
@@ -193,9 +193,8 @@ const getOrigin = (fgColor) => {
     cc: 'hawks',
     dc: 'doccloud',
   };
-  const originLC = (mappings[origin] || origin).toLowerCase();
-  if (originLC) {
-    return originLC;
+  if (mappings[origin.toLowerCase()] || origin) {
+    return mappings[origin.toLowerCase()] || origin;
```
narcis-radu, line 197: you might want to use a constant here
```suggestion
    const originLC = mappings[origin.toLowerCase()] || origin;
    if (originLC) return originLC;
```

mokimo (approved):

test/blocks/caas/send-to-caas.test.js
```diff
@@ -1,7 +1,8 @@
 import { readFile } from '@web/test-runner-commands';
 import { expect } from '@esm-bundle/chai';
 
-import { checkUrl } from '../../../tools/send-to-caas/send-utils.js';
+import { checkUrl, getOrigin } from '../../../tools/send-to-caas/send-utils.js';
+import { setConfig } from '../../../tools/send-to-caas/send-utils.js';
```
overmyheadandbody: Can't this be grouped with the import on top of it?

jedjedjedM (approved):

cmiqueo (approved):

PR #2428, opened 2024-06-05
MWPW-150665: [POC] Spectrum 2 Styles and Spectrum CSS Tokens Integration Script

saashapina:
This PR serves as a POC for ingesting Spectrum 2 Tokens into Milo. It introduces a script designed to extract and format Spectrum CSS Token values and integrate them into a single Spectrum 2 Milo CSS file.

* **Introduces Spectrum 2 Milo CSS styles:** This file includes all the core required Spectrum 2 custom properties for Milo. It consists of Milos specfic S2 custom properties sourced from [Spectrum CSS Tokens*](https://github.com/adobe/spectrum-css/tree/main/tokens) and "Narrative" values that only exist in Milo.
    * `--spectrum` prefixed custom properties are properties sourced from Spectrum CSS Tokens
    * `--sn` prefixed custom properties are "Narrative" properties specific to Milo

* **Automated Spectrum 2 Token Script:** A script is added to streamline the process of updating Spectrum 2 Milo CSS styles with S2 Token values. This script extracts and formats S2 values from Spectrum CSS Tokens, and updates the corresponding values in the Spectrum 2 Milo CSS file.

**Demonstrated Concept:**

This POC demonstrates a viable option for integrating Spectrum Tokens into Milo. It formats and extracts only the essential custom properties from the original 25kB gzipped Spectrum Tokens file resulting in a leaner 5.5kB Milo CSS file gzipped that's more in alignment with Milo's performance standards.

Resolves: [MWPW-150665](https://jira.corp.adobe.com/browse/MWPW-15066)

**Validation**
- Verify custom properties in `s2-styles.css` align with [Milo Spectrum 2 Values from design](https://docs.google.com/spreadsheets/d/18SdT8J2xkP4gVqWLJFXE2koPwRCHMNrUJ5NNsRmfH9k/edit#gid=2131325156)
- Verify working build by changing the values of `--s2` prefixed custom properties. 
    - Run `npm run build:spectrum2-styles` 
    - Verify the build automatically updates the custom property to the correct value as specified in the document in the previous step.

*Spectrum 2 CSS Tokens are currently sourced from Spectrum CSS tokens `v14.0.0-next.7` beta branch.

URL for testing:
- https://saasha-mwpw-150665-spectrum2-poc--milo--adobecom.hlx.page/


overmyheadandbody:
Hi @saashapina! Thanks for the contribution! I'd suggest we don't merge this to `stage` directly, I've created a new `spectrum-narrative` one from it, it might be a better fit and would allow for some easier testing.

I also have a quick question - would it be possible to make some adjustments so we can run the script similar to some other ones, like [these](https://github.com/adobecom/milo/blob/stage/package.json#L16-L18)? I'd just like to follow a pattern so we don't introduce any exceptions.

overmyheadandbody:
And one more question - there are a lot of `--spectrum-...` property names, is there anything speaking against replacing that with `--s2-...`? I think the reduction might add up to a noticeable difference in file size.

libs/features/spectrum2-styles/s2-styles.css
```diff
@@ -0,0 +1,931 @@
+/* stylelint-disable selector-class-pattern */
```
mokimo, line 1: I would suggest to put the output of the file directly into `libs/deps/spectrum2-styles.css`, but for the poc here is completely fine 👍

saashapina:
> And one more question - there are a lot of `--spectrum-...` property names, is there anything speaking against replacing that with `--s2-...`? I think the reduction might add up to a noticeable difference in file size.

@overmyheadandbody Yes we should be able to do this. The `--spectrum` prefix is directly from Spectrum CSS sources I should be able to add in a conversion to an `--s2` prefix for the Milo file. This could reduce the file size by a couple kb
Screenshot 2024-06-05 at 12 04 34 PM
Screenshot 2024-06-05 at 12 04 55 PM

mokimo:
@saashapina how did you create the KB output? If we can read that, it might be valuable to log changes of the size BEFORE/AFTER so that if the size changes considerably between versions, we have a quick comparison

saashapina:
> @saashapina how did you create the KB output? If we can read that, it might be valuable to log changes of the size BEFORE/AFTER within the script so that if the size changes considerably between versions, we have a quick comparison

Hey @mokimo the kB output above is the raw local file size from a VScode extension called filesize. But looking into it, we could probably get the outputted file size from the `statSync` method from the fs module and log that. I can throw up a commit for it.

libs/features/spectrum2-styles/build.mjs
```diff
@@ -0,0 +1,84 @@
+import fs from 'fs';
+import path from 'path';
+import postcss from 'postcss';
+
+const spectrumCSSPath = path.resolve('node_modules/@spectrum-css/tokens/dist/index.css');
+const miloCSSPath = path.resolve('../../deps/spectrum2-styles.css');
```
mokimo, line 6: I see that we are reading the CSS too - have a separate question about that.

Depending on your answer, I'd suggest to rename this:
```suggestion
const outputPath = path.resolve('../../deps/spectrum2-styles.css');
```

libs/features/spectrum2-styles/build.mjs
```diff
@@ -0,0 +1,84 @@
...
+
+const spectrumCSS = fs.readFileSync(spectrumCSSPath, 'utf8');
+const miloCSS = fs.readFileSync(miloCSSPath, 'utf8');
```
mokimo, line 15: I don't understand on why we read the `miloCSS`, could you elaborate on that?

In my mind: read spectrum styles -> transform properties -> output styles

libs/features/spectrum2-styles/build.mjs
```diff
@@ -0,0 +1,84 @@
...
+      });
+    }
+  });
+
+  return root.toString();
+};
+
+const updatedCustomCSS = updateCustomPropertiesInMiloCSS(miloCSS);
+
+fs.writeFileSync(miloCSSPath, updatedCustomCSS, 'utf8');
```
mokimo, line 91: Given we are in the context of the `spectrum2-styles/build.mjs` - IMO we can simplify the variable names
```suggestion
const styles = buildStyles(miloCSS);
fs.writeFileSync(outputPath, styles, 'utf8');
```

libs/features/spectrum2-styles/build.mjs
```diff
@@ -0,0 +1,84 @@
...
+    rule.walkDecls((decl) => {
+      let propName = decl.prop;
+      if (propName.startsWith(prefix)) propName = `--spectrum${propName.slice(prefix.length)}`;
+      customProperties[selector][propName] = decl.value;
+    });
+  });
+  return customProperties;
+};
+
+const spectrumCustomProperties = extractAndTransformCustomProperties(spectrumCSS, '--spectrum');
```
mokimo, line 65: The names are a little too long and make it hard to read and understand IMO.
```suggestion
const spectrumProps = transformProperties(spectrumCSS, '--spectrum')
```

libs/features/spectrum2-styles/build.mjs
```diff
@@ -0,0 +1,84 @@
...
+
+    if (prop.endsWith('-opacity')) {
+      const baseName = prop.replace('-opacity', '');
+      properties[baseName] = properties[baseName] ? properties[baseName].replace(')', ` / ${parseFloat(value) * 100}%)`) : `rgb(0 0 0 / ${value})`;
+      decl.remove();
+    }
+  });
+
+  Object.keys(properties).forEach((prop) => rule.append({ prop, value: properties[prop] }));
+};
```
mokimo, line 44: I can't put my finger on it, but I think this can be simplified a bit. Since you write the rgb values & then re-write them for the opacity, it seems like a bit of redundant work.

mokimo (commented):
A few nits, I like the approach. 

IMO the names can be simplified given that we already in the context of being within the "spectrum2-styles" build script. I had a few suggestions here and there, but would be great if you can just give it another spin and see if you can make the names clearer and shorter.

Great to see the POC coming in so fast though!

mokimo:
> Hey @mokimo the kB output above is the raw local file size from a VScode extension called filesize. But looking into it, we could probably get the outputted before/after file size from the statSync method from the fs module and log that. I can throw up a commit for it.
> But, I don't see this file size ever changing drastically after a build since we aren't retrieving new custom properties in this script. We are only receiving values for our already established custom properties. If there are requests for additional custom properties they would need to be added on our end first.

@saashapina Yeah no worries, if you can get something running quick that's cool - otherwise it's not critical enough to spend a lot of time/resource on it!

libs/features/spectrum2-styles/build.mjs
```diff
@@ -0,0 +1,84 @@
+import fs from 'fs';
+import path from 'path';
+import postcss from 'postcss';
+
+const spectrumCSSPath = path.resolve('node_modules/@spectrum-css/tokens/dist/index.css');
+const miloCSSPath = path.resolve('../../deps/spectrum2-styles.css');
+
+const spectrumCSS = fs.readFileSync(spectrumCSSPath, 'utf8');
```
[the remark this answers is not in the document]
saashapina: Yea this is just to avoid having to install an additional bundler to import the css as a module.

libs/features/spectrum2-styles/build.mjs
saashapina: So, we read the miloCSS because this is where we are _establishing_ all the custom property names that we want to extract from spectrum styles. 

Spectrum CSS is 3,600 lines of custom properties. We're establishing only the necessary properties given to us from design, in the initial `miloCSS` file, so its not necessarily an output. This file is just being updated with Spectrum CSS values, the property names themselves are unchanging.

We read the Milo files `--s2` prefixed properties -> use them locate and retrieve the 'values' we need from Spectrum CSS -> transform the values -> update the `miloCSS` file custom properties with the new corresponding Spectrum CSS values.

libs/features/spectrum2-styles/build.mjs
saashapina: Tried to clarify this a bit below, but we do read the miloCSSPath to define the custom properties we'll need from spectrum css.

libs/features/spectrum2-styles/build.mjs
saashapina: sounds good to me!

saashapina: Sure! I think its important to note this function is actually where the extracting/filtering properties is happening along with the transforming/formatting, if you think transform suffices and is still clear that works for me.

saashapina: sure! I think its important to note this function is extracting as well as formatting/transforming the properties, but I think `transform` should be able to suffice if you think this is still clear to you.

libs/features/spectrum2-styles/build.mjs
saashapina: hmm, maybe you can explain this a little more? I'm not sure what you mean by 're-writing' the rgb values for opacity.
So to clarify a little bit here, we are receiving values like this from Spectrum CSS Tokens:
```
  --spectrum-transparent-white-900-rgb: 255, 255, 255;
  --spectrum-transparent-white-900-opacity: 0.94;
```

and we're transforming and combining these into a singe property like this, to format and reduce file size.

```
--s2-transparent-white-900: rgb(255 255 255 / 94%);
```
This code identifies `-rgb` appended css properties, transforms the value from the `255, 255, 255` format into `rgb( 255 255 255)` and sets it to the new `--s2-transparent-white-900` (un-appended base name) property in the properties obj.

It then identifies the `-opacity` appended css property, checks if there is an existing associated base name in the obj, if one exists it will add in the formatted opacity value ` / 94% )` to the formatted rgb value. So `--s2-transparent-white-900` now becomes `rgb(255 255 255 / 94%)`. It should fallback to `rgb(0 0 0 / ${value})` if there is no associated rgb value.

Let me know if this clarifies that further? Or if you think this can be improved in some way.

libs/features/spectrum2-styles/build.mjs
saashapina: I think it would be important to make the distinction that this is the updated Milo styles to distinguish them from the original Milo CSS styles and other Spectrum styles that are also in this file, but if this is still remains clear to you we can simplify these

libs/features/spectrum2-styles/build.mjs
mokimo: If SpectrumCSS has 3.6k lines, and miloCSS has whatever that has - what's the process of adding variables in there?

If I get it correctly:
1. miloCSS already exists and knows what properties we want
2. We compare SpectrumCSS with miloCSS update miloCSS based on the SpectrumCSS

overmyheadandbody: I was going to ask the same yesterday, but forgot. It's much clearer now that we don't need to maintain 2 files, this totally makes sense.

My other curiosity is how you manage to keep track of "static" properties/values when updating. For example `--s2-corner-radius-900: 16px; /* S2 missing */` or `--s2-font-size-1300: 60px; /* should be 58px */`. If Spectrum added that value, it would automatically update here as well and we'd just remove the comment? Or is this something better defined as Spectrum Narrative? Does Spectrum 2 plan to have those values in the final release?

overmyheadandbody (approved):
I don't have a good picture just yet as to how we'd want to use the classes for buttons, links, details, titles, headings, etc. We don't use those classes in our blocks, nor is there any flexibility to add them through authoring. Furthermore, no build process means we can't "import" class definitions inside already existing class definitions. We might want to consolidate this output with [styles.css](https://github.com/adobecom/milo/blob/stage/libs/styles/styles.css#L197).

But that can be another discussion outside the scope of the POC, I think this serves our purposes to prove that this mechanism could work. Great job on this!

libs/features/spectrum2-styles/build.mjs
mokimo: I was thinking about something like this:

```js
rule.walkDecls((decl) => {
  const { prop, value } = decl;

  decl.value = replaceZeroPx(value);

  let baseName;
  let isRgb = prop.endsWith('-rgb');
  let isOpacity = prop.endsWith('-opacity');

  if (isRgb || isOpacity) {
    baseName = prop.replace(isRgb ? '-rgb' : '-opacity', '');
    properties[baseName] = isRgb ? `rgb(${value.replace(/,\s*/g, ' ')})` : 
      properties[baseName] ? properties[baseName].replace(')', ` / ${parseFloat(value) * 100}%)`) : `rgb(0 0 0 / ${value})`;
    decl.remove();
  }
});
```

libs/features/spectrum2-styles/build.mjs
saashapina: @overmyheadandbody Yes, that's correct. If it is prefixed with --s2 and has comments like `missing`, `should be xx`, it means these should be in SpectrumCSS S2 Tokens but are currently missing or incorrect. This is their beta release of S2 tokens, so it's expected. These should be included in their final release. When Spectrum updates these, they will automatically update here, and we can remove these comments. All the Spectrum Narrative (milo) static specific properties are prefixed with --sn and we manually manage these.

saashapina: @mokimo yes thats correct. The properties that milo needs currently in MiloCSS is specified by design. These should be unchanging but if there is additional properties/tokens wanted by design, yes they would need to be added in manually initially here with the correct token(property) name to be sourced from Spectrum. Yea, I can add this this somewhere in the documentation

libs/features/spectrum2-styles/build.mjs
saashapina: Nice! 👏🏼 .  I think milo's eslint has some opinions about nested ternary's I can adjust but I like this consolidation.

saashapina:
> I don't have a good picture just yet as to how we'd want to use the classes for buttons, links, details, titles, headings, etc. We don't use those classes in our blocks, nor is there any flexibility to add them through authoring. Furthermore, no build process means we can't "import" class definitions inside already existing class definitions. We might want to consolidate this output with [styles.css](https://github.com/adobecom/milo/blob/stage/libs/styles/styles.css#L197).
> 
> But that can be another discussion outside the scope of the POC, I think this serves our purposes to prove that this mechanism could work. Great job on this!

Yea, I don't have clear picture of this just yet either. I know we aren't just yet to the point of propagating Spectrum 2 throughout milo, would we still want to consolidate this with the main style.css file? At least for our upcoming showcase page? I seen a few places throughout Milo utilizing a `loadStyle()` utility, is something we could utilize?

overmyheadandbody:
`loadStyle` would do it in a JS context, to fetch styles for a particular block. Since the Spectrum ones should tailor to all our styles, I'd say they should be loaded OOTB. In the short run, for the purposes of the showcase page, `styles.css` and the Spectrum 2 ones could probably co-exist. In the long run, they should be consolidated into a single file without much redundancy.

PR #2438, opened 2024-06-06
MWPW-147206 making promo schedule geo aware

npeltier:
mostly to allow authors enabling promos in separate smaller files, schedule directive has now a locales setting to narrow its activation to those countries

promoEnabled parameter passed to getPromoManifests in promo utils is now an object that has entries for each region promo name that are fetched at very early time (no geo resolution), default being kept for backward compatibility reasons with key `manifestnames`. getPromoManifests then resolves current locale and serves merged manifests between regional specific ones (if locale matches), and 'global' ones.

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.live/be_fr/drafts/npeltier/geopromos/promo
- After: https://mwpw-147206--milo--npeltier.hlx.live/be_fr/drafts/npeltier/geopromos/promo

- Before: https://main--milo--adobecom.hlx.live/fr/drafts/npeltier/geopromos/promo
- After: https://mwpw-147206--milo--npeltier.hlx.live/fr/drafts/npeltier/geopromos/promo

with following new metadatas, in [promotions-emea.xlsx](https://adobe.sharepoint.com/:x:/r/sites/adobecom/Shared%20Documents/milo/promos/promotions-emea.xlsx?d=w92e31fc14e0f47a4851cbc314dac588d&csf=1&web=1&e=YknAEg). Note also [promotions-jp.xlsx](https://adobe.sharepoint.com/:x:/r/sites/adobecom/Shared%20Documents/milo/promos/promotions-jp.xlsx?d=w0af3704a08de455d8caff38a205362a8&csf=1&web=1&e=jDvkQz) have matching metadatas that will be displayed in the page meta but ignored by mep

| URL |	emea_manifestNames  | emea_schedule |
|----  | -------------------------  | ---------------- |
| /**/geopromos/promo | q2-promo-roc-fr | q2-promo-roc-fr &124; 2024-05-22T14:30:00 &124; 2024-07-17T04:00:00 &124; https://main--milo--adobecom.hlx.page/promos/2024/emea/nico-promo/nicopromo-fr.json &124; fr;ch_fr;be_fr |


3ch023 (changes requested):
some nala use cases are not working anymore

libs/features/personalization/promo-utils.js
```diff
@@ -1,36 +1,70 @@
...
         const disabled = isDisabled(event, searchParams);
         return { manifestPath, disabled, event };
       }
       return null;
     })
     .filter((manifest) => manifest != null);
+};
+
+export default function getPromoManifests(manifestNames, searchParams) {
+  const promoManifests = getRegionalPromoManifests(
```
3ch023, line 59: maybe we want to add here `if (region !== null)` to avoid calling getRegionalPromoManifests twice with null value if geo is not listed in the regions map yet

libs/features/personalization/promo-utils.js
```diff
@@ -1,36 +1,70 @@
...
-      const [name, start, end, manifestPath] = manifest.trim().split('|').map((s) => s.trim());
+      const [name, start, end, manifestPath, locales] = manifest.trim().split('|').map((s) => s.trim());
       if (attachedManifests.includes(name)) {
         const event = {
           name,
           start: GMTStringToLocalDate(start),
           end: GMTStringToLocalDate(end),
         };
+        if (locales) {
+          event.locales = locales?.split(';').map((locale) => locale.trim());
```
3ch023, line 48: i think we should check here if current page is not in target locales.
and if its not, it should not have an event.
currently you create a disabled event, which means it would show on all unaffected GEOs in the MEP Preview.
instead we could just have:

```
const localesList = locales?.split(';').map((locale) => locale.trim().toLowerCase());
if (!localesList.includes(localeCode)) return true;
```
(I added toLowerCase, it doesn't hurt)
I created https://mwpw-147206--milo--npeltier.hlx.page/de/npeltier/geopromos/promo to demonstrate what i mean, but it doesn't show preview because of another problem
[jp promotions](https://adobe.sharepoint.com/:x:/r/sites/adobecom/_layouts/15/Doc.aspx?sourcedoc=%7B0AF3704A-08DE-455D-8CAF-F38A205362A8%7D&file=promotions-jp.xlsx&action=default&mobileredirect=true) file have wrong column names  (emea) and rewrite metadata for emea

libs/features/personalization/promo-utils.js
```diff
@@ -1,36 +1,70 @@
-import { getMetadata } from '../../utils/utils.js';
+import { getMetadata, getConfig } from '../../utils/utils.js';
 
 const GMTStringToLocalDate = (gmtString) => new Date(`${gmtString}+00:00`);
 
+const APAC = ['au', 'cn', 'hk_en', 'hk_zh', 'id_en', 'id_id', 'in', 'in_hi', 'kr', 'my_en', 'my_ms', 'nz', 'ph_en', 'ph_fil', 'sg', 'th_en', 'th_th', 'tw', 'vn_en', 'vn_vi'];
```
narcis-radu, line 5: Would it make sense to get these mappings from a json? I'm thinking about geo-expansion and the need to update the list of supported languages.

libs/utils/utils.js
```diff
@@ -873,12 +875,42 @@ const getMepValue = (val) => {
...
+    'apac_manifestnames',
+    'emea_manifestnames',
+    'americas_manifestnames',
+    'jp_manifestnames',
+    'manifestnames',
+  ];
+  const mdObject = mds.reduce((obj, key) => {
+    const val = getMdValue(key);
+    if (val) {
+      obj[key] = getMdValue(key);
```
narcis-radu, line 897:
```suggestion
      obj[key] = val;
```

libs/utils/utils.js
```diff
@@ -873,12 +875,42 @@ const getMepValue = (val) => {
...
+    }
+    return obj;
+  }, {});
+  if (Object.keys(mdObject).length > 0) {
```
narcis-radu, line 901:
```suggestion
  if (Object.keys(mdObject).length) {
```

libs/utils/utils.js
```diff
@@ -873,12 +875,42 @@ const getMepValue = (val) => {
...
+  return false;
+};
+
 export const getMepEnablement = (mdKey, paramKey = false) => {
   const paramValue = PAGE_URL.searchParams.get(paramKey || mdKey);
   if (paramValue) return getMepValue(paramValue);
-  const mdValue = getMetadata(mdKey);
-  if (!mdValue) return false;
-  return getMepValue(mdValue);
+  if (PROMO_PARAM === paramKey) {
```
narcis-radu, line 910:
```suggestion
  if (PROMO_PARAM === paramKey) return getPromoMepEnablement();
```

libs/features/personalization/promo-utils.js
npeltier: i think i do agree with you. Fix the file you created and this is weird to see on de page
image
 i fixed it, please have another look @3ch023

libs/features/personalization/promo-utils.js
npeltier: i understand but what is the added value to separate it beside clarity for developers? Those lists are region mapping for promotion country, for now not used elsewhere.

3ch023: >and the need to update the list of supported languages.

for this use case its probably better to keep it in code similar to the list of locales in scripts.js.
hopefully geoexpansion PR will include changes to both lists at the same time, worst case we can always update it later.
json will not make it much easier

libs/features/personalization/promo-utils.js
```diff
@@ -1,27 +1,51 @@
...
   return Boolean(event.start && event.end
     && (currentDate < event.start || currentDate > event.end));
 };
 
-export default function getPromoManifests(manifestNames, searchParams) {
+const isManifestWithinLocale = (locales) => {
+  if (locales) {
+    return locales.split(';').map((locale) => locale.trim()).includes(localeCode);
+  }
+  return true;
```
3ch023, line 33: this means if the excel locales cell is empty if will apply to all geos, right?
this is probably ok but needs to be documented.
same as the fact that the only can use
/**/... once

3ch023 (approved):

libs/features/personalization/promo-utils.js
npeltier: yes, typically we won't ask locales in "default" manifest.
/**/ once per region!

vgoodric (approved):

libs/utils/utils.js
```diff
@@ -873,12 +875,40 @@ const getMepValue = (val) => {
...
+    if (val) {
+      obj[key] = val;
+    }
+    return obj;
+  }, {});
+  if (Object.keys(mdObject).length) {
+    return mdObject;
+  }
+  return false;
+};
```
overmyheadandbody, line 905: Is this logic needed in multiple places? Is it needed pre-LCP? Is there any other place is can exist? We plan on cleaning up utils in the near future to trim it down to the bare minimum.

afmicka:
@npeltier something wrong is happening with the manifest file when i add stuff to that. The second line seems ok and on the page i can see the manifest is served but not displaying (that page also has a promo from the default (old) file. The third row does not pickup the manifest schedule, therefore nothing is applied to that page. 
https://adobe.sharepoint.com/:x:/r/sites/adobecom/_layouts/15/Doc.aspx?sourcedoc=%7B92E31FC1-4E0F-47A4-851C-BC314DAC588D%7D&file=promotions-emea.xlsx&action=default&mobileredirect=true
Screenshot 2024-06-11 at 11 40 55

https://milo.adobe.com/promos/promotions-emea.json
Screenshot 2024-06-11 at 11 42 03

DE page: https://mwpw-147206--milo--npeltier.hlx.page/de/drafts/nala/features/promotions/promo-page-filter
(there should be text component with EMEA text inserted)
https://mwpw-147206--milo--npeltier.hlx.page/de/drafts/nala/features/promotions/promo-page-filter
Screenshot 2024-06-11 at 11 43 42

CH_DE page: https://mwpw-147206--milo--npeltier.hlx.page/ch_de/drafts/nala/features/promotions/promo-emea
(no manifest at all)

Please get in touch on a call when available to untangle this.

libs/utils/utils.js
npeltier: it is indeed needed pre-lcp like all mep related stuff in that file, orientating future rendering of the page based on metadata & query parameters. There is a dedicated mep utils test file. May be we should have a mep-prelcp one, but yeah, that would be an additional pre-lcp call.

afmicka:
@nateekar @Dli3 @JackySun9 caas and howto tests fail on every fork branch from our team. Could you please take a look into this? 

I am approving the PR. These failures are not related to the changes, rather test instabilities.

PR #2442, opened 2024-06-07
MWPW-149504 [MILO][MEP] Move entitlements object to the same JSON file used by the library

vgoodric:
We are being asked to add entitlement segments to MEP more often, and often with great urgency.  This update:
- Allows updates by modifying and publishing XLS to JSON instead of a PR
- By using the same spreadsheet the library will use, we will automatically have the library stay synced with new segments
- Makes it more clear which segments have only one of the 2 ids

Resolves: [MWPW-149504](https://jira.corp.adobe.com/browse/MWPW-149504)

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q3/mepentitlementsjson/
- After: https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q3/mepentitlementsjson/?milolibs=mepentitlementsjson

Psi-check: https://mepentitlementsjson--milo--adobecom.hlx.page/drafts/mepdev/psicheck/?martech=off


libs/features/personalization/entitlements.js
```diff
@@ -1,38 +1,18 @@
...
-};
+import { fetchData, DATA_TYPE } from './personalization.js';
 
 export const getEntitlementMap = async () => {
-  const { env, consumerEntitlements } = getConfig();
-  if (env?.name === 'prod') return { ...consumerEntitlements, ...ENTITLEMENT_MAP };
-  const { default: STAGE_ENTITLEMENTS } = await import('./stage-entitlements.js');
-  return { ...consumerEntitlements, ...STAGE_ENTITLEMENTS };
+  const { env, consumerEntitlements, base } = getConfig();
+  const dataFile = `${base.replace('libs', 'docs')}/library/entitlement-tags.json`;
```
mokimo, line 6: We don't want to maintain this in the milo Sharepoint, the odds of this getting changed involuntarily are pretty high. 
The [federal](https://adobe.sharepoint.com/sites/adobecom/Shared%20Documents/Forms/AllItems.aspx?id=%2Fsites%2Fadobecom%2FShared%20Documents%2Ffederal%2Ffederal&viewid=d776cf70%2D9b7e%2D4ab7%2Db9da%2D9e0f8e03a7d2) Sharepoint would be better

mokimo (commented):
Does this have any performance impact if we wait for fetching the entitlements first?

libs/features/personalization/entitlements.js
```diff
@@ -1,38 +1,18 @@
...
+  const fetchedData = await fetchData(dataFile, DATA_TYPE.JSON);
+  if (!fetchedData) return consumerEntitlements;
+  const entitlements = {};
+  const { data } = fetchedData;
+  data.forEach((ent) => {
```
narcis-radu, line 11:
```suggestion
  fetchedData?.data?.forEach((ent) => {
```

overmyheadandbody (commented):
I also wonder whether the performance impact justifies going down this road. Since you're doing this for MEP, could these not reside in a new sheet from the manifest? That would avoid 2 separate calls. Then again, entitlements are needed outside of MEP too, so we'd still have to make a call 🤔 

The reason you're invoking is around urgency. We discussed that certain MEP features can be fast-tracked, could we not reach an agreement on having a separate entitlement files with some similar exemptions?

We recently delivered a feature to transform links to their stage counterpart on non-prod environments and we debated whether an additional call would actually be justified, even if just for stage. We decided performance shouldn't be impacted. The end user (getting worse performance) and authors (who now _could_ be asked to maintain entitlement spreadsheets) should not have to "suffer" because we can't agree on a better contribution model.

libs/features/personalization/entitlements.js
vgoodric: @mokimo 2 things:
1. I plan to restrict access to this spreadsheet to help with that (similar to how metadata spreadsheet is restricted)
2. My hope was to use the same file used by the library to ensure it stays in sync

vgoodric: @mokimo I moved the entitlements data file to the federal folder.  I realized that the library can pull in from the feds folder if I updated the block code processing entitlements for the library.  And this puts the JSON output onto the same domain instead of milo.adobe.com@mokimo I moved the entitlements data file to the federal folder.  I realized that the library can pull in from the feds folder if I updated the block code processing entitlements for the library.  And this puts the JSON output onto the same domain instead of milo.adobe.com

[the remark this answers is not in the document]
vgoodric: @overmyheadandbody see above comment.  moved to feds folder.

libs/features/personalization/entitlements.js
```diff
@@ -1,38 +1,18 @@
...
+    const { id, stageid, tagname } = ent;
+    entitlements[env?.name === 'prod' ? id : stageid] = tagname;
```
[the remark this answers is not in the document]
vgoodric: updated

vgoodric:
delaying

PR #2443, opened 2024-06-07
MWPW-148002: Adjust Strike-through price font size for: merch-card (all variations)

mirafedas:
This change sets the strikethrough price font size to 14px and font weight to normal in the merch-card block.
See XD: https://xd.adobe.com/view/fdec345a-503e-436f-af76-f15d375609b6-a8bf/screen/09ce5290-f27e-4b42-8654-8da85bea840a/specs/

Resolves: [MWPW-148002](https://jira.corp.adobe.com/browse/MWPW-148002)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/docs/authoring/commerce/merch-card-product?martech=off
- After: https://MWPW-148002-price-font-size--milo--mirafedas.hlx.page/docs/authoring/commerce/merch-card-product?martech=off


mokimo (approved):

libs/blocks/merch/merch.css
```diff
@@ -8,6 +8,8 @@ span[data-wcs-osi] {
 }
 
 span.placeholder-resolved[data-template="strikethrough"], span.price.price-strikethrough {
```
npeltier, line 10: isn't `span.price.price-strikethrough` enough? or is to keep layout room before price is loaded?

mirafedas: @npeltier I did not change this line, and not sure I want to change it now, because I also don't know the reason why .placeholder-resolved is there, maybe it's needed :)

npeltier: ho true, sorry, ok :)

npeltier (approved):
lgtm

libs/blocks/merch/merch.css
```diff
@@ -8,6 +8,8 @@ span[data-wcs-osi] {
...
+  font-size: 14px;
```
overmyheadandbody, line 11: Could we not use `--type-body-xs-size`?

overmyheadandbody (approved):

libs/blocks/merch/merch.css
mirafedas: Yes, thank you, added

mirafedas:
@afmicka I see it's a problem with the branch itself, cannot tell what's wrong with it (I'm sure it's not related to the CSS changes I made, because most of the cards don't even have strikethrough prices), but it was way faster to open a new PR with the same change: https://github.com/adobecom/milo/pull/2459 . I will close this one

yesil:
@mirafedas this PR looks similar to https://github.com/adobecom/milo/pull/2459/files and doesn't contain any change as of now, closing it. Please re-open if needed.

PR #2448, opened 2024-06-10
Increase 404 rum tracking to 10%

mokimo:
### Description
To get a more accurate signal for 404 pages in the RUM data, we want to increase tracking to 10% rather than 1% for 404s.

Resolves: [MWPW-152132](https://jira.corp.adobe.com/browse/MWPW-152132)

**Test URLs:**
- After: https://mwpw-152132--milo--mokimo.hlx.page/?martech=off

### DIFF
--- a/libs/utils/samplerum.js
+++ b/libs/utils/samplerum.js
@@ -47,8 +47,11 @@ export function sampleRUM(checkpoint, data = {}) {
       window.hlx.rum = { weight, id, random, isSelected, firstReadTime, sampleRUM, sanitizeURL: urlSanitizers[window.hlx.RUM_MASK_URL || 'path'], rumSessionStorage };
     }
 
-    const { weight, id, firstReadTime } = window.hlx.rum;
-    if (window.hlx && window.hlx.rum && window.hlx.rum.isSelected) {
+    const { id, firstReadTime } = window.hlx.rum;
+    const weight404 = 10;
+    const isSelected404 = checkpoint === '404' && (Math.random() * weight404 < 1);
+    const weight = isSelected404 ? weight404 : window.hlx.rum;
+    if (window.hlx && window.hlx.rum && (window.hlx.rum.isSelected || isSelected404)) {
       const knownProperties = ['weight', 'id', 'referer', 'checkpoint', 't', 'source', 'target', 'cwv', 'CLS', 'FID', 'LCP', 'INP', 'TTFB'];
       const sendPing = (pdata = data) => {
         // eslint-disable-next-line object-curly-newline, max-len, no-use-before-define


libs/utils/samplerum.js
```diff
@@ -47,8 +47,11 @@ export function sampleRUM(checkpoint, data = {}) {
       window.hlx.rum = { weight, id, random, isSelected, firstReadTime, sampleRUM, sanitizeURL: urlSanitizers[window.hlx.RUM_MASK_URL || 'path'], rumSessionStorage };
     }
 
-    const { weight, id, firstReadTime } = window.hlx.rum;
-    if (window.hlx && window.hlx.rum && window.hlx.rum.isSelected) {
+    const { id, firstReadTime } = window.hlx.rum;
+    const weight404 = 10;
+    const isSelected404 = checkpoint === '404' && (Math.random() * weight404 < 1);
+    const weight = isSelected404 ? weight404 : window.hlx.rum;
```
overmyheadandbody, line 53: Shouldn't this be `window.hlx.rum.weight`?

libs/utils/samplerum.js
```diff
@@ -47,8 +47,11 @@ export function sampleRUM(checkpoint, data = {}) {
...
+    if (window.hlx && window.hlx.rum && (window.hlx.rum.isSelected || isSelected404)) {
```
overmyheadandbody, line 54: On L50, you're directly using `window.hlx.rum`, but here you're making a check for it. Shouldn't we fail faster if it doesn't exist?

mokimo: Since the data is always set on L47, we don't need the safeguard at all.

overmyheadandbody (approved):

overmyheadandbody:
Weren't the dependencies already pushed in https://github.com/adobecom/milo/pull/2451? We should probably keep them separate.

mokimo:
A potential 404 page: https://mwpw-152132--milo--mokimo.hlx.live/404test?martech=off

mokimo:
After discussion with @chicharr - I'm switching to the approach used in https://github.com/adobe/helix-rum-js/pull/159

libs/scripts/scripts.js
```diff
@@ -161,6 +161,9 @@ const eagerLoad = (img) => {
 }());
 
 (async function loadPage() {
+  if (document.head.querySelector('meta[name="template"]')?.content === '404') {
```
chicharr, line 164: I'm not familiar with the load sequence on milo, but it is important that this happens before the first call to `sampleRUM`

mokimo: Yes this is the first thing that's called, only downside is it needs to be implemented by every consumer. 
Which might be a good thing, since we only need to increase 404 tracking for BACOM for now - so that particular project can implement it!

chicharr: Then, PR looks good to me

libs/utils/samplerum.js
```diff
@@ -28,8 +28,10 @@ export function sampleRUM(checkpoint, data = {}) {
   try {
     window.hlx = window.hlx || {};
     if (!window.hlx.rum) {
-      const usp = new URLSearchParams(window.location.search);
-      const weight = (usp.get('rum') === 'on') ? 1 : 100; // with parameter, weight is 1. Defaults to 100.
+      const weight = (window.SAMPLE_PAGEVIEWS_AT_RATE === 'high' && 10)
+      || (window.SAMPLE_PAGEVIEWS_AT_RATE === 'low' && 1000)
+      || (new URLSearchParams(window.location.search).get('rum') === 'on' && 1)
+      || 100;
       const id = Array.from({ length: 75 }, (_, i) => String.fromCharCode(48 + i)).filter((a) => /\d|[A-Z]/i.test(a)).filter(() => Math.random() * 75 > 70).join('');
```
chicharr, line 35: we have switched to a simpler and less prone to collisions, logic to generate ids.
you may want to change it here as well https://github.com/adobe/helix-rum-js/blob/80a9e5e041ad1f935926c888da9dcccce4abb250/src/index.js#L20

mokimo: Added 👍

robert-bogos (approved):
Looks good 👍

libs/scripts/scripts.js
overmyheadandbody: This file already includes methods from `utils.js`, could we not use `getMetadata` for consistency?

On another note, we might need to help/reach out to create this PR on BACOM's side as well.

overmyheadandbody (approved):

libs/scripts/scripts.js
mokimo: Yup, will open a PR on the BACOM side of things!

SilviuLCF (approved):
Verified , ready for STAGE

Testing details https://jira.corp.adobe.com/browse/MWPW-152132

ramboz (approved):

PR #2454, opened 2024-06-10
MWPW-151790[MILO][MEP] Organize getPersConfig function in personalization.js

markpadbe:
The code refactors personalization.js in the following ways:
Renames getPersConfig and parseConfig functions to getManifestConfig and parseManifestVariants for clarity. Within the getManifestConfig function, renames the config variable which is assigned the value returned by parseManifestVariants to manifestConfig and renames of globalConfig to config, to match usage throughout the codebase. 

Resolves: [MWPW-MWPW-151790](https://jira.corp.adobe.com/browse/MWPW-MWPW-151790)

**Test URLs:**
--Before: https://main--cc--adobecom.hlx.page/products/photoshop
--After: https://main--cc--adobecom.hlx.page/products/photoshop?milolibs=orggetpersconfigfrommepplaceholderfix

**Test URLs for psi check:**
- After: https://orggetpersconfigfrommepplaceholderfix--milo--adobecom.hlx.page/?martech=off


libs/features/personalization/personalization.js
```diff
@@ -786,6 +786,7 @@ export function cleanAndSortManifestList(manifests) {
         freshManifest.name = fullManifest.name;
         freshManifest.selectedVariantName = fullManifest.selectedVariantName;
         freshManifest.selectedVariant = freshManifest.variants[freshManifest.selectedVariantName];
+        delete freshManifest.variants;
```
denlight, line 789: looks like this one is the only business logic change. Everything else is a name swap. Is this prop dele intentional ?

markpadbe: yes, this is intentional. I just updated the comment to mention it. Thanks.

denlight (approved):
lgtm!

libs/features/personalization/personalization.js
vgoodric: @markpadbe I hate to say it, but let's not remove.  I forgot about this use of the object.  We're essentially trying to take the decision from the Target instance and merge it with the fresh data of the imported instance (when there's a duplicate).  So I worry if we delete the variants we create a small fringe case of deleting it and then needing it later in the loop.

vgoodric (approved):

PR #2458, opened 2024-06-11
MWPW-140906 Graybox Misc UI Fixes

chrischrischris:
Fixes a few small UI bugs found during testing.

Resolves: [MWPW-140906](https://jira.corp.adobe.com/browse/MWPW-140906)

Marking as `zero-impact` as the graybox block is not yet in use (and will only be internal when used).

**Test URLs:**
- After: https://gb-border-fix--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/libs/blocks/graybox/graybox.css
+++ b/libs/blocks/graybox/graybox.css
@@ -39,13 +39,15 @@
   position: fixed;
 }
 
-.gb-no-click::before {
+.gb-no-click::after {
   content: "";
   height: 100%;
   pointer-events: all;
   position: absolute;
   width: 100%;
   z-index: var(--above-all);
+  left: 0;
+  top: 0;
 }
 
 .gb-no-change {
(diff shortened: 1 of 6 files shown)


arshadparwaiz (approved):

chrischrischris:
Adding Verified label as it was verified by @amauch-adobe before he left for PTO

overmyheadandbody:
Hey @chrischrischris! With the recent automations, the `verified` label is no longer required. 'Ready for Stage' assumes that the PR has been tested and is in good shape.

Will Graybox always be a non-prod-impact thing? If so, it probably belongs in this list - https://github.com/adobecom/milo/blob/stage/.github/workflows/label-zero-impact.js#L3-L13. Then it would automatically be labeled as 'zero-impact' and get fast-tracked.

chrischrischris:
@overmyheadandbody Yes graybox is an internal only tool, but I'm ok with leaving off that list.  The reason right now it's zero-impact is that it's still in development and so is not being used internally yet.

PR #2463, opened 2024-06-11
MWPW-151885 - [LocUI] messaging for auth redirect

Sartxi:
When a user launches their localization project without having already authenticated, the user is redirected to login and upon return the project details passed as query parameters are then automatically cleared and the user gets the missing block message. This PR adds a pseudo locui block that simply displays a message to the user to re-run their project. (This block would be destroyed if we ever merge the locui branch to main)

Screenshot 2024-06-11 at 3 27 18 PM

Resolves: [MWPW-151885](https://jira.corp.adobe.com/browse/MWPW-151885)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/tools/loc?martech=off
- After: https://sartxi-locui-message--milo--adobecom.hlx.page/tools/loc?martech=off

### DIFF
--- a/libs/blocks/locui/locui.css
+++ b/libs/blocks/locui/locui.css
@@ -1,0 +1,9 @@
+.locui.container {
+  margin: 0 auto;
+  margin-top: 10vh;
+  max-width: 700px;
+  border-radius: 10px;
+  border: 10px solid #eee;
+  padding: 10px 30px 30px;
+  text-align: center;
+}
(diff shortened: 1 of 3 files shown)


auniverseaway:
Gonna back seat drive the messaging a little bit...

```
The project details were removed after you logged in. To resolve this:

1. Close this window or tab.
2. Open your project Excel file.
3. Click "Localize..." in Sidekick again.
```

The idea is to keep it short and list-y.

auniverseaway (approved):
It would probably be better to have the content driven from Word, but then we would have to change every consuming page and keep them in sync.

Also, I was very confused at first why this is here and being merged into stage, but now I get it. With no query params, locui branch doesn't get loaded off Milo either.

rgclayton (approved):

ryanmparrish (approved):

PR #2465, opened 2024-06-12
MWPW-148256 add lh to merch card in collections

npeltier:
based on name, then title of the card

Resolves: [MWPW-148256](https://jira.corp.adobe.com/browse/MWPW-148256)

**Test URLs:**
- After: https://mwpw-148256--milo--npeltier.hlx.live/?martech=off

- Before: https://main--cc--adobecom.hlx.live/products/catalog?martech=off
- After: https://main--cc--adobecom.hlx.live/products/catalog?martech=off&milolibs=mwpw-148256--milo--npeltier

### DIFF
--- a/libs/blocks/merch-card-collection/merch-card-collection.js
+++ b/libs/blocks/merch-card-collection/merch-card-collection.js
@@ -158,6 +158,7 @@ async function initMerchCards(filtered, preferences, cardsRoot) {
   // re-order cards, update card filters
   [...cardsRoot.children].filter((card) => card.tagName === 'MERCH-CARD').forEach((merchCard) => {
     merchCard.style.display = 'none';
+    merchCard.setAttribute('daa-lh', `card-${merchCard.name || merchCard.title}`);
     const filters = { ...merchCard.filters };
     Object.keys(filters).forEach((key) => {
       const preference = preferences[key];


Axelcureno (approved):

libs/blocks/merch-card-collection/merch-card-collection.js
```diff
@@ -158,6 +158,7 @@ async function initMerchCards(filtered, preferences, cardsRoot) {
   // re-order cards, update card filters
   [...cardsRoot.children].filter((card) => card.tagName === 'MERCH-CARD').forEach((merchCard) => {
     merchCard.style.display = 'none';
+    merchCard.setAttribute('daa-lh', `card-${merchCard.name || merchCard.title}`);
```
yesil, line 161: as per https://github.com/adobecom/milo/blob/stage/libs/martech/attributes.md,

> do not add daa-lh to any element on or inside the blocks
if you add daa-ll to an element, you must have exactly 2 "levels".

I recommend using `decorateSectionAnalytics` on the cards root.

yesil (changes requested):

libs/blocks/merch-card-collection/merch-card-collection.js
npeltier: actually we shouldn't even use this as when merch-card-collection is in a "normal" section, the quoted method *is* used (see this attribute rewritten with "normal" analytics string without catalog container) https://main--cc--adobecom.hlx.page/drafts/npeltier/promocard-test?milolibs=MWPW-148256--milo--npeltier

if you look at related ticket requirement for catalog is a much smaller string, but thinking it twice i don't see any reason setting it up in milo. Will close that PR, and raise another one specific to catalog

yesil: do you know why we have daa-lh in your promocard PR but not at https://main--cc--adobecom.hlx.page/products/catalog?

PR #2467, opened 2024-06-12
[MWPW-151172][Kodiak] Package updates

overmyheadandbody:
This is a dependency update to address Kodiak stories, more details can be found in JIRA.

Resolves: [MWPW-151172](https://jira.corp.adobe.com/browse/MWPW-151172)

**Test URLs:**
- After: https://mwpw-151172-kodiak--milo--overmyheadandbody.hlx.page/?martech=off

### DIFF
--- a/.github/workflows/label-zero-impact.js
+++ b/.github/workflows/label-zero-impact.js
@@ -11,6 +11,7 @@ const zeroImpactDirs = [
   'LICENSE',
   'codecov.yaml',
   '.gitignore',
+  'package-lock.json',
 ];
 const zeroImpactLabel = 'zero-impact';
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "@adobecom/milo",
   "version": "1.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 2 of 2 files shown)


.github/workflows/label-zero-impact.js
```diff
@@ -11,6 +11,7 @@ const zeroImpactDirs = [
   'LICENSE',
   'codecov.yaml',
   '.gitignore',
+  'package-lock.json',
```
mokimo, line 14: @chrischrischris mentioned graybox is also zero impact, however we currently only allow full directories rather than paths to be checked.
Wonder if we should change that?

mokimo (commented):
How come there's only updates to the package-lock.json and not the package.json?

overmyheadandbody:
My understanding is that `package-lock.json` has the outdated versions, leading to Kodiak alerts. Simply running `npm install` and `npm update` updates `package-lock.json` to use the stable versions of the packages we're interested in, without updating anything in `package.json`.

.github/workflows/label-zero-impact.js
overmyheadandbody: I asked for some details on that. If it's confirmed that Graybox needs to be part of the exception list, we can open a quick story to make the change.

mokimo (approved):

SilviuLCF (approved):
Verified, testing details https://jira.corp.adobe.com/browse/MWPW-151172

PR #2473, opened 2024-06-14
MWPW-139487 - Fix for Gnav scroll issue with Landscape Orientation on devices

Deva309:
## Description
On the landscape orientation for mobile, the user cannot access (scroll to) the lower part of the mega menu.

Resolves: [MWPW-139487](https://jira.corp.adobe.com/browse/MWPW-139487)

## Screenshots:
Before:
Screenshot 2024-06-14 at 2 00 42 PM

After:
Screenshot 2024-06-14 at 2 00 34 PM

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/devashish/wam?martech=off
- After: https://mwpw-139487--milo--deva309.hlx.page/drafts/devashish/wam?martech=off

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.css
+++ b/libs/blocks/global-navigation/global-navigation.css
@@ -551,7 +551,7 @@ header.global-navigation {
   }
 
   .feds-menu-container {
-    max-width: 1200px;
+    max-width: var(--feds-minWidth-menu);
     margin: 0 auto;
     padding: 0 12px 40px;
   }
--- a/libs/blocks/global-navigation/global-navigation.css
+++ b/libs/blocks/global-navigation/global-navigation.css
@@ -551,7 +551,8 @@ header.global-navigation {
   }
 
   .feds-menu-container {
-    min-width: 100%;
+    max-width: 1200px;
+    margin: 0 auto;
     padding: 0 12px 40px;
   }
 
@@ -568,6 +569,8 @@ header.global-navigation {
     right: 0;
     justify-content: center;
     padding: 40px 0 0;
+    max-height: calc(100vh - 100%);
+    overflow: auto;
   }
 
   [dir = "rtl"] .feds-navItem--megaMenu .feds-popup {
--- a/libs/blocks/global-navigation/global-navigation.css
+++ b/libs/blocks/global-navigation/global-navigation.css
@@ -551,8 +551,7 @@ header.global-navigation {
   }
 
   .feds-menu-container {
-    max-width: var(--feds-minWidth-menu);
-    margin: 0 auto;
+    min-width: 100%;
     padding: 0 12px 40px;
   }
 
--- a/libs/blocks/global-navigation/global-navigation.css
+++ b/libs/blocks/global-navigation/global-navigation.css
@@ -567,7 +567,6 @@ header.global-navigation {
 
   .feds-navItem--megaMenu .feds-popup {
     right: 0;
-    justify-content: center;
     padding: 40px 0 0;
     max-height: calc(100vh - 100%);
     overflow: auto;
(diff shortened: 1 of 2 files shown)


libs/blocks/global-navigation/utilities/menu/menu.css
```diff
@@ -5,8 +5,7 @@
 
 .feds-navLink--hoverCaret[aria-expanded = "true"] + .feds-popup,
 .feds-menu-headline[aria-expanded = "true"] + .feds-menu-items {
-  display: flex;
-  flex-direction: column;
+  display: inline-block;
```
overmyheadandbody, line 8: This will cause issues in the footer

Screenshot 2024-06-17 at 12 09 57

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -551,7 +551,8 @@ header.global-navigation {
   }
 
   .feds-menu-container {
-    min-width: 100%;
+    max-width: 1200px;
```
overmyheadandbody, line 554: There's already a variable for this value, `--feds-minWidth-menu`

overmyheadandbody (changes requested):
I'm not sure this is the right direction. The aim is to adapt the height, yet width and layout properties are affected instead, leading to additional issues. I added the highlighted properties from the screenshot to the `feds-popup` styles starting at `900px` and things seem to work as expected. Will need a lot more testing, but this just impacts the height of the relevant element.

Screenshot 2024-06-17 at 12 27 39

libs/blocks/global-navigation/utilities/menu/menu.css
Deva309: Fixed.
Replaced with `display: block;`.

libs/blocks/global-navigation/global-navigation.css
Deva309: Added `--feds-minWidth-menu` variable for width.

Deva309:
> I'm not sure this is the right direction. The aim is to adapt the height, yet width and layout properties are affected instead, leading to additional issues. I added the highlighted properties from the screenshot to the `feds-popup` styles starting at `900px` and things seem to work as expected. Will need a lot more testing, but this just impacts the height of the relevant element.
> 
> Screenshot 2024-06-17 at 12 27 39

@overmyheadandbody 
`flex-direction` does not have any impact now because we updated `display` property from `flex` to `block`.
I have verified all the viewports with this change.Also, height impact only on `feds-popup` as we are calculating based on viewport height and this will only applied when viewport height is less than `mega-menu` height.
I have verified all the viewports with this change. But yeah this change needs to be tested properly.

overmyheadandbody (commented):
I still don't agree with this approach. Again, you want to adapt behavior on the Y axis, yet this PR makes changes to the X axis. The screenshot I provided yesterday with another way of achieving scroll inside popups only should be reconsidered. That only affects the Y axis and makes changes to a single element.

Deva309:
> I still don't agree with this approach. Again, you want to adapt behavior on the Y axis, yet this PR makes changes to the X axis. The screenshot I provided yesterday with another way of achieving scroll inside popups only should be reconsidered. That only affects the Y axis and makes changes to a single element.

@overmyheadandbody 
I understand your point about the unintended problems caused by changing the width and layout. This affected other sections, as you pointed out in footer. So, I've undone those changes and now I'm only fixing the height of the `megamenu` and adding scrolling just for that element. However, we still need horizontal scrolling (scroll to X axis) because we have
promo that get cut off on the right side, similar to what you might see on a Photoshop page below:

Screenshot 2024-06-18 at 3 40 02 PM

That's why I included `overflow: auto` in this scenario.

overmyheadandbody (approved):
Thanks for updating the approach!

salonijain3 (approved):

prativas22:
https://main--cc--adobecom.hlx.live/creativecloud/file-types/image?milolibs=mwpw-139487--milo--deva309
https://main--cc--adobecom.hlx.live/creativecloud?milolibs=mwpw-139487--milo--deva309
https://main--cc--adobecom.hlx.live/creativecloud/file-types/image?milolibs=mwpw-139487--milo--deva309&akamaiLocale=tr
vertical scrolling to the lower part of Mega menu
image
image
horizontal scrolling to the right part of Mega menu
image
image

PR #2395, opened 2024-05-31
Mepupdate

kandavadivelan:
* Add your
* Specific
* Features or fixes

Resolves: [MWPW-NUMBER](https://jira.corp.adobe.com/browse/MWPW-NUMBER)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/?martech=off
- After: https://<branch>--milo--adobecom.hlx.page/?martech=off

PR #2474, opened 2024-06-14
mwpw-152624: improve stage PR communication

mokimo:
### Description
- If PRs are not merged (no approvals, failing tests, overlap in files) the process should an actionable message on the PR in question.
- If the stage batch is created, we also want to post to #milo-stage so SOTs know that they can start testing

Resolves: [MWPW-152624](https://jira.corp.adobe.com/browse/MWPW-152624)

**Test URLs:**
- After: https://mwpw-152624--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/.github/workflows/merge-to-stage.js
+++ b/.github/workflows/merge-to-stage.js
@@ -55,6 +55,28 @@ const hasFailingChecks = (checks) =>
       name !== 'merge-to-stage' && conclusion === 'failure'
   );
 
+const commentOnPR = async (comment, prNumber) => {
+  console.log(comment); // Logs for debugging the action.
+  const { data: comments } = await github.rest.issues.listComments({
+    owner,
+    repo,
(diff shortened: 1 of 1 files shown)


robert-bogos (approved):
Looks good 👍

overmyheadandbody (approved):
Since workflows add so many comments, I wonder whether it still makes sense to have the stale PR workflow, it may be that they would add more noise.

mokimo:
@overmyheadandbody To be fair, this only adds comments on PRs that are `Ready for Stage` - I think it still catches a small amount of stale PRs such as this one: https://github.com/adobecom/milo/pull/2395

But yes, it doesn't play together nicely, we should keep a mental note of that workflow // sunsetting it.

SilviuLCF (approved):
Zero Impact PR, no QA needed

PR #2477, opened 2024-06-14
MWPW-152627 [MILO][MARTECH] add Target propositionDisplay call

vgoodric:
Currently in Milo MEP, the Target propositionDisplay call is not happening after the Target propositions are rendered on the page.

This update makes the Target propositionDisplay call from the MEP Martech.js file once the Target propositions are applied.

This wiki explains the changes in detail (Testing instructions are available in this wiki):
https://wiki.corp.adobe.com/pages/viewpage.action?pageId=3169882054

Resolves: [MWPW-152627](https://jira.corp.adobe.com/browse/MWPW-152627)

**Test URLs:**
- After: https://main--homepage--adobecom.hlx.page/homepage/index-loggedout?milolibs=targetpropdisplay

Note: check the network tab for an interact call to sstats.  Should be listed as eventType in events property in Payload tab

Psi-check: https://targetpropdisplay--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/libs/martech/martech.js
+++ b/libs/martech/martech.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-underscore-dangle */
 import { getConfig, getMetadata, loadIms, loadLink, loadScript } from '../utils/utils.js';
 
 const ALLOY_SEND_EVENT = 'alloy_sendEvent';
@@ -220,7 +221,6 @@ const loadMartechFiles = async (config, url, edgeConfigId) => {
     const env = ['stage', 'local'].includes(config.env.name) ? '.qa' : '';
     const martechPath = `martech.main.standard${env}.min.js`;
     await loadScript(`${config.miloLibs || config.codeRoot}/deps/${martechPath}`);
-    // eslint-disable-next-line no-underscore-dangle
     window._satellite.track('pageload');
   };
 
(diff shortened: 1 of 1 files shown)


libs/martech/martech.js
```diff
@@ -143,9 +145,13 @@ const getTargetPersonalization = async () => {
   } else {
     sendTargetResponseAnalytics(false, responseStart, timeout);
     manifests = handleAlloyResponse(response.result);
+    propositions = response.result && response.result.propositions;
```
markpadbe, line 148: In the case that something unexpected is returned for response.result.propositions like undefined or null, is it better that window._satellite.track on lined 255 is not called, or that it's called with an empty array? if it's the latter, small suggestion:
`propositions = response.result && response.result.propositions || [];`

markpadbe (approved):

libs/martech/martech.js
narcis-radu, line 148: We define propositions as an empty array, but this can override the value with `undefined` if the `response.result` object is not available. Maybe we want to set the `propositions` value only if `response.result.propositions` is defined, otherwise fallback to default value (empty array).

vgoodric: @narcis-radu , I added an ` || []`.
Please let me know if that satisfies your concern.

PR #2478, opened 2024-06-14
MWPW-152659 [MILO][MODAL][ANALYTICS] Analytic on modal load is missing

vgoodric:
Note: this PR is flagged as high priority because we're getting pushed from above to resolve this ASAP.

Detailed Description: when using _satellite.track, most of the data that will be sent is already known.  You only need to update specific data.  Attempting to override data with undefined will destroy needed data.   So we need to check if the value is defined before adding it to the object we send to the function.
URL: https://business.adobe.com/uk/ or any geo you are not in
Steps to Reproduce:
1. Open network tab and filter for the sstats
2. Check collect entry for info on modal load

Expected Results: Call sends _adobe_corpnew info
Actual Results: Call does not send _adobe_corpnew info

Cause, event.data is undefined, and destroys the object.  By first checking if event.data is defined, we can not overwrite the data that will automatically be sent.

Resolves: [MWPW-152659](https://jira.corp.adobe.com/browse/MWPW-152659)

**Test URLs:**
https://business.adobe.com/uk/

Unfortunately, this cannot be tested with preview URLs.  You must override content for the modal.js file.

psi check: https://loadmodaleventfix--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-underscore-dangle */
 /* eslint-disable import/no-cycle */
 import { createTag, getMetadata, localizeLink, loadStyle, getConfig } from '../../utils/utils.js';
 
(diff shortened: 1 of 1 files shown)


markpadbe (approved):

denlight (approved):

libs/blocks/modal/modal.js
```diff
@@ -20,18 +21,15 @@ export function findDetails(hash, el) {
...
+  const data = {
     xdm: {},
-    data: {
-      web: { webInteraction: { name: event?.type } },
-      _adobe_corpnew: { digitalData: event?.data },
-    },
-  });
+    data: { web: { webInteraction: { name: event?.type } } },
+  };
+  if (event?.data) data.data._adobe_corpnew.digitalData = event.data;
```
overmyheadandbody, line 28: Wouldn't this throw an error? On L26 you define `data.data`, which in turn doesn't have the `_adobe_corpnew` property, so we'd get a `Cannot set properties of undefined (setting 'digitalData')` error. Am I missing something?

overmyheadandbody (changes requested):
I think this can lead to issues, if I'm missing something, feel free to re-request my review so it's no longer blocked.
Screenshot 2024-06-18 at 11 02 54

libs/blocks/modal/modal.js
vgoodric: @overmyheadandbody That's a good point. Updated.

overmyheadandbody (approved):
Thanks for addressing this!

narcis-radu (approved):

PR #2484, opened 2024-06-18
MWPW-150666: Spectrum 2 Tokens Integration into Milo

saashapina:
This work presents 3 potential options for integrating Spectrum 2 Tokens into Milo's main stylesheet.

<s>**Option 1:** Slimmed Spectrum 2 Tokens Integration into Milo `styles.css` - **48.6kB - (7.95kB gzipped) Raw**
   - This option presents an integration of a slimmed down version of Spectrum 2 Tokens, consisting only of what's currently needed. Additional properties would be added as needed. We will need to further assess and determine the immediate tokens required for Milo.</s>

**Option 2:** Entire Spectrum 2 Tokens Integration into Milo `styles2.css` - **54.5kb - (9.4KB gzipped)** 
   - This option presents an integration of the entire Spectrum 2 Tokens properties required by Milo from design.

<s>**Option 3:** Spectrum 2 Token Variables Integration into Milo `styles2-var.css`-  **68kb - (10.46kb gzipped) Raw**
  - This option explores a Spectrum 2 integration into Milo css using variables imported from `spectrum2-styles.css`</s>
  
  ### Changes
- Combines Spectrum 2 Tokens into Milo css file in the 3 scenarios above.
- Creates new build that reads Spectrum Tokens and replaces values directly from and to main Milo style.css (`build-main.mjs`)
-  Refactors original script to accommodate the restructuring of architecture in `spectrum2-styles.css`
- `spectrum2-styles.css` size reduction to **3.18kb gzipped RAW** after changes.

Resolves: [MWPW-150666](https://jira.corp.adobe.com/browse/MWPW-150666)

**Test URLs:**
- After: https://saasha-mwpw-150666-spectrum2-milo-integration--milo--adobecom.hlx.page/


libs/styles/styles2.css
```diff
@@ -0,0 +1,1606 @@
...
+  /* aspect ratios (milo original) */
+  --aspect-ratio-square: 1/1;
+  --aspect-ratio-standard: 4/3;
+  --aspect-ratio-wide: 16/9;
+
+
+  /**** Style Specs ****/
+
+  /* S2 Spacing */
+  --spacing-xxs: var(--s2-spacing-75, 8px); /* 4px */
```
overmyheadandbody, line 368: What's the purpose of having default values for these? Do they reflect what is currently being used in Milo?

overmyheadandbody (commented):
This will require a few more review sessions, since it's quite large. I have some initial quick feedback. Option-wise, I'm currently leaning towards the second one. First one would need maintenance as soon as something new is required and consumers might not be aware of this, slowing down the process. Third one adds too much overhead. So the second one feels like a good middle ground.

libs/styles/styles2.css
```diff
@@ -0,0 +1,1606 @@
...
+ :root {
+  --global-height-nav: 64px;
+  --global-height-breadcrumbs: 33px;
+  /* stylelint-disable-next-line custom-property-pattern */
+  --global-height-navPromo: 65px;
+  --feds-totalheight-nav: calc(var(--feds-height-nav, --global-height-nav) + var(--feds-height-breadcrumbs, --global-height-breadcrumbs));
+
+  /* Fonts */
+  --body-font-family: 'Adobe Clean', adobe-clean, 'Trebuchet MS', sans-serif;
+  --headline-font-family: 'Adobe Headline', adobe-headline, 'Trebuchet MS', sans-serif; /* new */
```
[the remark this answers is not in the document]
saashapina: This new headline font is a part of Spectrum 2. Maybe we can bring up some these concerns in our next Narrative engineering meeting? Or reach out in slack.

libs/styles/styles2.css
```diff
@@ -0,0 +1,1606 @@
...
+  --fixed-font-family: 'Roboto Mono', menlo, consolas, 'Liberation Mono', monospace;
+
+  /* Milo Color Specs */
+  --color-white: var(--s2-white); 
+  --color-black: var(--s2-black); 
+  --color-gray-100: var(--s2-gray-100,#F8F8F8);
+  --color-gray-200: var(#E8E8E8); /* milo only */
```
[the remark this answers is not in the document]
saashapina: Good catch, this is a typo.

libs/styles/styles2.css
```diff
@@ -0,0 +1,1606 @@
...
+  /* Focus Indicator Aliases */
+  --s2-focus-indicator-color: var(--s2-blue-800);
+  --s2-static-white-focus-indicator-color: var(--s2-white);
+  --s2-static-black-focus-indicator-color: var(--s2-black);
+
+  /*** S2 Elevation (Drop Shadow) ***/
+  --s2-drop-shadow-x-100: 0;
+  --s2-drop-shadow-x-200: 0;
+  --sn-drop-shadow-x-300: 0; /* NARRATIVE */
+  --sn-drop-shadow-x-400: 0; /* NARRATIVE */
```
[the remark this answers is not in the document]
saashapina: Agreed! We could possibly just create `--sn-drop-shadow-x: 0` and use it where needed. These are currently the final values in spectrum 2.

libs/styles/styles2.css
saashapina: Yes these default values reflect whats currently being used in Milo to demonstrate the difference in sizes between S2 and Milo. These can be removed when ready.

libs/styles/styles2.css
```diff
@@ -0,0 +1,1606 @@
...
+  --spacing-xl-static: 56px;
+  --spacing-xxl-static: 80px;
+  --spacing-xxxl-static: 104px;
+
+  /* S2 Headings */
+  --type-heading-xl-weight: var(--s2-black-font-weight); 
+  --type-heading-all-weight: var(--s2-extra-bold-font-weight);
+  --type-heading-lh: 1.25; 
+  --type-heading-xxxl-size: calc(var(--font-size-multiplier, 1) * 44px); /* original not in S2 */
+  --type-heading-xxxl-lh: 55px;
```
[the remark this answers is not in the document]
saashapina: Yes the ones defined in px values is the original from Milo and not apart of S2. But I can update these for consistency.

libs/styles/styles2.css
```diff
@@ -0,0 +1,1606 @@
...
+  --type-heading-xxl-size: calc(var(--font-size-multiplier, 1) * var(--s2-font-size-1200, 36)); 
+  --type-heading-xxl-lh: var(--type-heading-lh, 45);
```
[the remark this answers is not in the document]
saashapina: Yes these should be `px`, typo as well. These are also default values of the original Milo values to keep track of them and should be able to be removed when ready.

saashapina:
> This will require a few more review sessions, since it's quite large. I have some initial quick feedback. Option-wise, I'm currently leaning towards the second one. First one would need maintenance as soon as something new is required and consumers might not be aware of this, slowing down the process. Third one adds too much overhead. So the second one feels like a good middle ground.

Agreed! Removed the other options to cut down the files.

libs/styles/styles.css
```diff
@@ -284,166 +742,282 @@
...
+  min-width: 126px;
 }
 
-.light .con-button {
-  border-color: var(--text-color);
-  color: var(--text-color);
-}
+.xxl-button-mobile .con-button, /* S2 only */
+.con-button.button-xxl { 
+  --xxl-button-height: 56;
```
overmyheadandbody, line 839: Same missing unit question. And looking at it, they (L827 & L837) seem to share values, could we consolidate them?

libs/styles/styles.css
```diff
@@ -284,166 +742,282 @@
...
+
+.con-button.primary-white-button[aria-disabled="true"] {
+  background-color: var(--s2-transparent-white-100);
+}
+
+/* Secondary Buttons Black */
+.con-button.secondary-black-button {
+  background-color: transparent;
+  border: 2px solid var(--s2-gray-800);
+  color: var(--s2-transparent-black-400)
```
overmyheadandbody, line 929: Nit: missing semicolon

libs/styles/styles.css
```diff
@@ -284,166 +742,282 @@
...
+.con-button.secondary-white-button[aria-disabled="true"]:hover {
+  cursor: default;
+}
+
+span.icon.margin-right { margin-right: 8px; }
+span.icon.margin-left { margin-left: 8px; }
+span.icon.margin-inline-end { margin-inline-end: 8px; }
+
+.button-l .con-button span.icon.margin-left,
+.con-button.button-l span.icon.margin-left { margin-left: 12px; }
```
overmyheadandbody, line 1015: Could we use a variable such as `--s2-spacing-200` for these spacing values?

libs/styles/styles.css
```diff
@@ -812,15 +1347,15 @@ a.static:hover {
...
+@media (max-width: 768px) {
   .con-button.button-justified-mobile {
     display: block;
     text-align: center;
   }
 }
 
 /* Tablet up */
-@media screen and (min-width: 600px) {
+@media screen and (min-width: 768px) {
```
overmyheadandbody, line 1358: When dealing with both min and max width media queries, it's usually a good idea to use a 1px difference, otherwise there might be issues at exactly 768px viewport width. However, I see that wasn't considered before either, so maybe we don't change it right now.

libs/features/spectrum2-styles/build.mjs
```diff
@@ -68,7 +68,28 @@ const extractAndTransform = (css, prefix) => {
   return customProperties;
 };
 
+const mergeProperties = (baseProps, additionalProps) => {
```
overmyheadandbody, line 71: This is also defined in `build-main.mjs`, could we import it to avoid duplicates?

saashapina: There are actually a few duplicate functions that I can pull out and import. But considering we're going in the direction of all the s2 tokens living in the main styles.css file anyway, we could probably get rid of this original `build.mjs` that is building `spectrum2-styles.css` and get rid of `spectrum2-styles.css` too.

libs/styles/styles.css
saashapina: Good catch! This was actually an incorrect value here.

saashapina:
> Looking better! I have a few more observations. It's still a complex PR, so it's difficult to spot everything in one go, I may need a few more sessions on this one.

Thanks for hanging in there! This was a lot of manual data input so I appreciate the second pair of eyes.

libs/styles/styles.css
```diff
@@ -284,166 +742,282 @@
...
+
+.con-button.premium-button[aria-disabled="true"] {
+  background-color: var(--s2-gray-100);
+  outline: none;
+}
+
+/* Disabled Button */
+.con-button[aria-disabled="true"]:hover,
+.con-button.blue[aria-disabled="true"]:hover,
+.con-buttton.accent-button[aria-disabled="true"]:hover,
```
robert-bogos, line 1001: I think there is a typo in here.
```suggestion
.con-button.accent-button[aria-disabled="true"]:hover,
```

libs/styles/styles.css
```diff
@@ -1,134 +1,587 @@
...
   /* stylelint-disable-next-line custom-property-pattern */
   --global-height-navPromo: 65px;
   --feds-totalheight-nav: calc(var(--feds-height-nav, --global-height-nav) + var(--feds-height-breadcrumbs, --global-height-breadcrumbs));
 
-  /* Colors */
-  --link-color: #035fe6;
-  --link-hover-color: #136ff6;
+  /* Fonts */
+  --body-font-family: 'Adobe Clean', adobe-clean, 'Trebuchet MS', sans-serif;
+  --headline-font-family: 'Adobe Headline', adobe-headline, 'Trebuchet MS', sans-serif; /* new */
```
narcis-radu, line 14: wondering whether we should fallback to `Adobe Clean` first ?!

libs/styles/styles.css
```diff
@@ -1,134 +1,587 @@
...
+  --s2-negative-color-1100: var(--s2-red-1100);
+  --s2-notice-color-700: var(--s2-orange-700);
+  --s2-positive-color-700: var(--s2-green-700);
+  --s2-positive-color-900: var(--s2-green-900);
+  --s2-positive-color-1000: var(--s2-green-1000);
+  --s2-accent-color-900: var(--s2-blue-900);
+  --s2-accent-color-1000: var(--s2-blue-1000);
+
+  /*** Milo Semantic Colors ***/
+  --link-color: var(--color-blue-900, #035fe6);
```
narcis-radu, line 111: We should take the opportunity to make all color declarations consistent, either uppercase or lowercase.

libs/styles/styles.css
```diff
@@ -1,134 +1,587 @@
...
-  --type-body-xs-lh: 21px;
-  --type-body-xxs-size: calc(var(--font-size-multiplier, 1) * 12px);
-  --type-body-xxs-lh: 18px;
-
-  /* Detail */
-  --type-detail-all-weight: 700;
+  /* S2 Titles (new) */
+  --type-title-xl-weight: var(--s2-extra-bold-font-weight);
+  --type-title-all-weight: var(--s2-bold-font-weight);
+  --type-title-lh: 130%;
```
narcis-radu, line 490: I see the line height is used as percentage (e.g. `--type-title-lh`), fixed value (e.g. `--type-heading-base-lh`) and px value (e.g. `--type-heading-xxxl-lh`). Should we go the extra mile and make them consistent?

libs/styles/styles.css
```diff
@@ -1,134 +1,587 @@
...
+    --type-title-xxxl-size: var(--s2-font-size-700);
+    --type-title-xxl-size: var(--s2-font-size-600);
+    --type-title-xl-size: var(--s2-font-size-500);
+    --type-title-l-size: var(--s2-font-size-400);
+    --type-title-m-size: var(--s2-font-size-300);
+    --type-title-s-size: var(--s2-font-size-200);
+    --type-title-xs-size: var(--s2-font-size-100);
+  }
+  
+  @media (min-width: 768px) { /* Small desktop */
```
narcis-radu, line 509: Other media queries mention 'small desktop' at `min-width: 1280px`.

libs/styles/styles.css
```diff
@@ -284,164 +794,305 @@
...
 }
 
-.detail-xs {
-  font-size: var(--type-body-xs-size);
-  line-height: var(--type-body-xs-lh);
-  font-weight: 700;
+/* Button Sizes */ 
+.con-button { /* Size M - Secondary Static Black (Default) */
+  --m-button-height: 32px;
+
```
narcis-radu, line 820: I assume the extra line is intentional

libs/features/spectrum2-styles/build.mjs
mokimo: I don't have the context, but if that is (still) the direction - then your suggestion of removing the unnecessary files sounds good 👍

libs/features/spectrum2-styles/build.mjs
```diff
@@ -81,10 +105,15 @@ const updateMiloCSS = (css) => {
         const { prop } = decl;
         if (prop.startsWith('--s2')) {
           const spectrumProp = `--spectrum${prop.slice(4)}`;
-          if (spectrumProperties[selector][spectrumProp] !== undefined) {
-            decl.value = spectrumProperties[selector][spectrumProp];
+          const spectrumSelector = selector === ':root' ? '.spectrum' : selector === '.light' ? '.spectrum--light' : selector === '.dark' ? '.spectrum--dark' : null;
```
mokimo, line 108: Nested ternaries are hard to read, error-prone & discouraged by most linting configurations. Hence I'd suggest to switch to an if/else:
```suggestion
let spectrumSelector;
if (selector === ':root') {
  spectrumSelector = '.spectrum';
} else if (selector === '.light') {
  spectrumSelector = '.spectrum--light';
} else if (selector === '.dark') {
  spectrumSelector = '.spectrum--dark';
} else {
  spectrumSelector = null;
}
```

libs/styles/styles.css
saashapina: I believe they are currently all lowercase, let me know if I missed anything?

libs/styles/styles.css
saashapina: Yea, this is because the new S2 properties are used in percentages and the Milo originals are px and fixed values there. Wasn't sure if we should change the Milo originals but I can update these for consistency.

libs/styles/styles.css
saashapina: Good catch!

libs/styles/styles.css
saashapina: Yes there is a linting opinion expecting an empty line before variable declarations but we can remove and ignore the linter opinion if necessary.

libs/styles/styles.css
narcis-radu: I noticed some upper case values
https://github.com/adobecom/milo/pull/2484/files#diff-c685e2a5191e8c907b9958b7f267d04cc48fdce1491da2c38461955f5d2bd913R19
https://github.com/adobecom/milo/pull/2484/files#diff-c685e2a5191e8c907b9958b7f267d04cc48fdce1491da2c38461955f5d2bd913R20-R23

libs/styles/styles.css
overmyheadandbody: Definitely makes sense to have it before `Trebuchet`

libs/styles/styles.css
overmyheadandbody: I'd also suggest to have these as factors, not percentages, `1.3` as opposed to `130%`

libs/styles/styles.css
overmyheadandbody: I agree with the Linter's opinion, variables should be clearly separated.

libs/styles/styles.css
```diff
@@ -1,134 +1,586 @@
...
   /* stylelint-disable-next-line custom-property-pattern */
   --global-height-navPromo: 65px;
   --feds-totalheight-nav: calc(var(--feds-height-nav, --global-height-nav) + var(--feds-height-breadcrumbs, --global-height-breadcrumbs));
 
-  /* Colors */
-  --link-color: #035fe6;
-  --link-hover-color: #136ff6;
+  /* Fonts */
+  --body-font-family: 'Adobe Clean', adobe-clean, 'Trebuchet MS', sans-serif;
+  --headline-font-family: 'Adobe Headline', adobe-headline, 'Adobe Clean', sans-serif; /* new */
```
overmyheadandbody, line 13: We should probably mimic the value from `--body-font-family` and just prefix it with the headline variants
```suggestion
  --headline-font-family: 'Adobe Headline', adobe-headline, 'Adobe Clean', adobe-clean, 'Trebuchet MS', sans-serif; /* new */
```

libs/styles/styles.css
```diff
@@ -1,134 +1,586 @@
...
+  --fixed-font-family: 'Roboto Mono', menlo, consolas, 'Liberation Mono', monospace;
+
+  /* Milo Color Specs */
+  --color-white: var(--s2-white); 
+  --color-black: var(--s2-black); 
+  --color-gray-100: var(--s2-gray-100,#F8F8F8);
```
overmyheadandbody, line 19: I don't remember where we landed on the default values. Does it make sense to continue defining them? If we start using S2, we'd get all of the variables, so the default would never be used, right?

If, for some reason, we need them, nit: space after the comma, same for L21.

libs/styles/styles.css
```diff
@@ -1,134 +1,586 @@
...
+  /* S2 Grid sizes */
+  --grid-container-width: 100%; /* previously 83.4% */
+  --grid-column-width: calc(var(--grid-container-width) / 8);
+  --grid-margins-width: 16px; /* previously calc((100vw - var(--grid-container-width)) / 2) */
+  /* milo originals */
+  --grid-margins-width-10: calc((100vw - (var(--grid-column-width) * 10)) / 2);
+  --grid-margins-width-8: calc((100vw - (var(--grid-column-width) * 8)) / 2);
+  --grid-margins-width-6: calc((100vw - (var(--grid-column-width) * 6)) / 2);
+
+  @media (min-width: 768px) { /* Tablet */
```
overmyheadandbody, line 393: Where are we getting the `768` value from? Is it from somewhere in Spectrum? Currently, we consider `900` to be the shift from mobile to desktop, we should get alignment on the value. On resolutions between 768-900, we would get mobile styles from the blocks, but tablet styles from this file. Similar question for 1280, 1680.

libs/styles/styles.css
```diff
@@ -1,134 +1,586 @@
...
-  --type-heading-s-size: calc(var(--font-size-multiplier, 1) * 18px);
-  --type-heading-s-lh: 22.5px;
-  --type-heading-xs-size: calc(var(--font-size-multiplier, 1) * 16px);
-  --type-heading-xs-lh: 20px;
-  --type-heading-xxs-size: calc(var(--font-size-multiplier, 1) * 14px);
-  --type-heading-xxs-lh: 18px;
+  /* S2 Headings */
+  --type-heading-xl-weight: var(--s2-black-font-weight); 
+  --type-heading-all-weight: var(--s2-extra-bold-font-weight);
+  --type-heading-lh: 125%;
```
overmyheadandbody, line 441: Had a comment on this in the PR main page, but we could go for `1.25` directly.

libs/styles/styles.css
```diff
@@ -1,134 +1,586 @@
...
-  --type-heading-xs-size: calc(var(--font-size-multiplier, 1) * 16px);
-  --type-heading-xs-lh: 20px;
-  --type-heading-xxs-size: calc(var(--font-size-multiplier, 1) * 14px);
-  --type-heading-xxs-lh: 18px;
+  /* S2 Headings */
+  --type-heading-xl-weight: var(--s2-black-font-weight); 
+  --type-heading-all-weight: var(--s2-extra-bold-font-weight);
+  --type-heading-lh: 125%; 
+  --type-heading-xxxl-size: calc(var(--font-size-multiplier, 1) * 44px); /* original not in S2 */
+  --type-heading-xxxl-lh: 125%;
```
overmyheadandbody, line 443: Others use the variable, this one should probably use it too.

libs/styles/styles.css
```diff
@@ -1,134 +1,586 @@
...
-  --type-body-xxs-lh: 18px;
-
-  /* Detail */
-  --type-detail-all-weight: 700;
+  --type-heading-base-lh: 120%;
+
+  /* S2 Titles (new) */
+  --type-title-xl-weight: var(--s2-extra-bold-font-weight);
+  --type-title-all-weight: var(--s2-bold-font-weight);
+  --type-title-lh: 130%;
```
overmyheadandbody, line 489:
```suggestion
  --type-title-lh: 1.3;
```

libs/styles/styles.css
```diff
@@ -1,134 +1,586 @@
...
+    --type-title-l-size: var(--s2-font-size-500);
+    --type-title-m-size: var(--s2-font-size-400);
+    --type-title-s-size: var(--s2-font-size-300);
+    --type-title-xs-size: var(--s2-font-size-200);
+  }
+
+  /* S2 Detail */
+  --sn-type-detail-xl-weight: var(--s2-bold-font-weight);
+  --type-detail-all-weight: var(--s2-bold-font-weight);
+  --type-detail-lh: 125%;
```
overmyheadandbody, line 531:
```suggestion
  --type-detail-lh: 1.25;
```

libs/styles/styles.css
```diff
@@ -1,134 +1,586 @@
...
+  @media (min-width: 1680px) { /* Large desktop */
+    --sn-type-detail-l-size: var(--s2-font-size-500);
+    --sn-type-detail-m-size: var(--s2-font-size-300);
+    --sn-type-detail-s-size: var(--s2-font-size-100);
+  }
 
-  /* z-index */
-  --above-all: 9000; /* Used for page tools that overlay page content */
+  /* S2 Body */
+  --type-body-lh: 150%;
```
overmyheadandbody, line 561:
```suggestion
  --type-body-lh: 1.5;
```

libs/styles/styles.css
```diff
@@ -160,95 +612,152 @@
...
+.detail-s {
+  font-size: var(--type-detail-s-size);
+  font-weight: var(--type-detail-all-weight);
+  line-height: var(--type-detail-s-lh);
+}
+
+.detail-xs {
+  font-size: var(--type-body-xs-size);
+  line-height: var(--type-body-xs-lh);
+  font-weight: 700;
```
overmyheadandbody, line 758: Could we use `--s2-bold-font-weight`?

libs/styles/styles.css
```diff
@@ -284,164 +793,305 @@
...
-  border-color: var(--text-color);
+.con-button.blue:focus,
+.con-button.accent-button:focus {
+  outline: 2px solid var(--s2-blue-800);
+}
+
+.con-button.blue[aria-disabled="true"], 
+.con-button.accent-button[aria-disabled="true"] {
+  background-color: var(--color-gray-100);
+  color: var(--color-gray-400)
```
overmyheadandbody, line 928:
```suggestion
  color: var(--color-gray-400);
```

libs/styles/styles.css
```diff
@@ -284,164 +793,305 @@
...
-.dark .con-button.fill {
-  background: var(--color-white);
-  border-color: var(--color-white);
-  color: var(--text-color);
+.con-button.fill:focus,
+.con-button.primary-black-button:focus {
+  outline: 2px solid var(--s2-static-black-focus-indicator-color);
+}
+
+
```
overmyheadandbody, line 952:
```suggestion
```

libs/styles/styles.css
```diff
@@ -284,164 +793,305 @@
...
+}
+
+.con-button.secondary-black-button:focus {
+  outline: 2px solid var(--s2-blue-800);
+}
+
+.con-button.secondary-black-button[aria-disabled="true"] {
+  background-color: transparent;
+  outline: 2px solid var(--s2-gray-300);
+  color: var(--s2-transparent-black-400)
```
overmyheadandbody, line 1000:
```suggestion
  color: var(--s2-transparent-black-400);
```

libs/styles/styles.css
```diff
@@ -284,164 +793,305 @@
...
+}
+
+/* Secondary Buttons White */
+.con-button.secondary-white-button {
+  background-color: var(--s2-transparent-white-25);
+  border: 2px solid var(--s2-gray-800);
+  color: var(--s2-transparent-white-800)
```
overmyheadandbody, line 1007:
```suggestion
  color: var(--s2-transparent-white-800);
```

libs/styles/styles.css
```diff
@@ -284,164 +793,305 @@
...
+
+/* Disabled Button */
+.con-button[aria-disabled="true"]:hover,
+.con-button.blue[aria-disabled="true"]:hover,
+.con-button.accent-button[aria-disabled="true"]:hover,
+.con-button.fill[aria-disabled="true"]:hover,
+.con-button.primary-black-button[aria-disabled="true"]:hover,
+.con-button.primary-white-button[aria-disabled="true"]:hover,
+.con-button.secondary-black-button[aria-disabled="true"]:hover,
+.con-button.secondary-white-button[aria-disabled="true"]:hover {
```
overmyheadandbody, line 1060: This seems a bit excessive 🤔 I don't see other `cursor` declarations, so would `.con-button[aria-disabled="true"]:hover` not suffice?

libs/styles/styles.css
```diff
@@ -767,40 +1363,61 @@ h6 {
...
-/* Links Static */
-a.static,
-a.static:hover {
-  color: var(--text-color);
+a.primary-link:hover, a.primary-link:active, a.primary-link:focus  {
+  color: var(--s2-blue-1000);
+}
+
+/* Links (Secondary) */
+a.static,a.secondary-link {
```
overmyheadandbody, line 1406:
```suggestion
a.static, a.secondary-link {
```

overmyheadandbody (commented):
Mostly nits; however, I would like to discuss the new media queries, as that would cause some discrepancies between block styles and these ones.

libs/styles/styles.css
saashapina: I think we can get rid of the default values at this point. These were only here to keep note of for reference purposes.

libs/styles/styles.css
saashapina: Yes these are the new media queries specified by S2. They can be found in these specs [here. ](https://www.figma.com/design/9R2PrV8kBaICcz4xLX1rZQ/Spectrum-Narrative-Exploration?node-id=1379-37572&m=dev)

libs/features/spectrum2-styles/build.mjs
saashapina: I realized this ternary actually turned out to be redundant so its been removed after looking into this a bit more.

libs/features/spectrum2-styles/build.mjs
```diff
@@ -68,7 +65,34 @@ const extractAndTransform = (css, prefix) => {
...
+  } else if (selector === '.light') {
+    mergeProperties(properties, spectrumProperties['.spectrum--light'] || {});
+  } else if (selector === '.dark') {
+    mergeProperties(properties, spectrumProperties['.spectrum--dark'] || {});
+  } else {
+    mergeProperties(properties, spectrumProperties[selector] || {});
+  }
+  return properties;
+};
+
```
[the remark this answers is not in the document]
saashapina: Nice!

libs/styles/styles.css
saashapina: Good catch, I think it would.

libs/styles/styles.css
saashapina: Ahh just realized you were referencing the hex values, while I was looking for color variable names. These should be updated now👍

overmyheadandbody (approved):
Aside from the media queries discussion, this can probably be merged into the feature branch to not stall any longer. Additional feedback could be addressed in subsequent PRs.

narcis-radu (approved):

robert-bogos (approved):

PR #2486, opened 2024-06-18
MWPW-151814 - [LocUI] URL sanitization

Sartxi:
Sometimes `--` in our url hostnames are autocorrected to be an EN Dash. That special character breaks the URL constructor and the URL is then is flagged as an invalid URL. This PR adds sanitization to correct this problem in the front end.

Resolves: [MWPW-151814](https://jira.corp.adobe.com/browse/MWPW-151814)

**Test URLs:**
https://sean-loc--milo--adobecom.hlx.page/tools/loc?ref=main&repo=milo&owner=adobecom&host=milo.adobe.com&project=Milo&referrer=https%3A%2F%2Fadobe.sharepoint.com%2F%3Ax%3A%2Fr%2Fsites%2Fadobecom%2F_layouts%2F15%2FDoc.aspx%3Fsourcedoc%3D%257B25CBF365-16EF-4190-B278-35614F0ACD35%257D%26file%3DUrls%2520404s%2520sync%25201.xlsx%26action%3Ddefault%26mobileredirect%3Dtrue


libs/blocks/locui/loc/index.js
```diff
@@ -97,12 +97,20 @@ async function loadProjectSettings(projSettings) {
   }
 }
 
+function sanitize(url) {
+  const hlxUrl = /^https?:\/\/[^/?#]+/g;
+  const [hostname] = url.match(hlxUrl);
```
overmyheadandbody, line 102: Is there high certainty that the URL is never invalid? Otherwise this will return `null`, which will throw an error on the next line.

libs/blocks/locui/loc/index.js
```diff
@@ -97,12 +97,20 @@ async function loadProjectSettings(projSettings) {
...
+  return new URL(newURL);
+}
+
 async function loadDetails() {
   setStatus('details', 'info', 'Loading languages and URLs.');
   try {
     const resp = await fetch(previewPath);
     const json = await resp.json();
-    const jsonUrls = json.urls.data.map((item) => new URL(item.URL));
+    const jsonUrls = json.urls.data.map(({ URL }) => sanitize(URL));
```
overmyheadandbody, line 113: Has the data structure also changed? Previously `item.URL` was being sent, now it's just `URL`, which corresponds to the previous `item`.

libs/blocks/locui/loc/index.js
```diff
@@ -97,12 +97,20 @@ async function loadProjectSettings(projSettings) {
   }
 }
 
+function sanitize(url) {
+  const hlxUrl = /^https?:\/\/[^/?#]+/g;
```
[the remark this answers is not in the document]
Sartxi: @overmyheadandbody This regex matches only the hostname of a valid URL. I updated the var name to make more sense semantically. But only wanting to check the hostnames for the EN Dash special char.

libs/blocks/locui/loc/index.js
Sartxi: There is a high certainty that this will always be a valid URL but I added a check to make sure. Thanks for spotting this!

libs/blocks/locui/loc/index.js
Sartxi: I simply thought spreading this to get the property name would look better, but dont want to introduce a new pattern so I have reverted this.

overmyheadandbody: I 100% missed the extra curly braces, my bad 😅

rgclayton (approved):

libs/blocks/locui/loc/index.js
```diff
@@ -97,12 +97,23 @@ async function loadProjectSettings(projSettings) {
...
 
+function sanitize(url) {
+  let newURL = url;
+  const matchHost = /^https?:\/\/[^/?#]+/g;
+  const [hostname] = url.match(matchHost);
+  if (hostname) {
+    const enDash = hostname.replaceAll('–', '--');
+    newURL = url.replace(matchHost, enDash);
+  }
+  return new URL(newURL);
```
overmyheadandbody, line 108: Wouldn't a lack of a hostname indicate a faulty URL that might yield errors when using `new URL`? Comparing it to the previous logic, this is ok, since this was happening before as well. But from a contained `sanitize` method perspective, it may be counter-intuitive to not check the validity of the URL before returning it.

PR #2493, opened 2024-06-19
MWPW-152286 Optimize marquee loading

TsayAdobe:
* Use a block preload map for known imports in a block
* A section doesn't need to wait for loading of modal links. Marquee and modal are loaded at the same time.
* More performance gain on top of #2391 

Resolves: [MWPW-152286](https://jira.corp.adobe.com/browse/mwpw-152286)

**Test URLs:**

- After: https://mwpw-152286--milo--adobecom.hlx.page/?martech=off

**DC Test URLs:**

- Before: https://www.stage.adobe.com/acrobat/how-to/pdf-editor-pdf-files.html
- After: https://www.stage.adobe.com/acrobat/how-to/pdf-editor-pdf-files.html?milolibs=mwpw-152286

Before:
before1

After:
after1

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -1150,7 +1150,7 @@ async function documentPostSectionLoading(config) {
   document.body.appendChild(createTag('div', { id: 'page-load-ok-milo', style: 'display: none;' }));
 }
 
-function partition(arr, fn) {
+export function partition(arr, fn) {
   return arr.reduce(
     (acc, val, i, ar) => {
       acc[fn(val, i, ar) ? 0 : 1].push(val);
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -87,6 +87,9 @@ const MILO_BLOCKS = [
   'share',
   'reading-time',
 ];
+
+const MILO_BLOCKS_PRELOADS = { marquee: ['/utils/decorate.js'] };
+
 const AUTO_BLOCKS = [
   { adobetv: 'tv.adobe.com' },
   { gist: 'https://gist.github.com' },
@@ -451,6 +454,8 @@ export async function loadBlock(block) {
 
   const base = miloLibs && MILO_BLOCKS.includes(name) ? miloLibs : codeRoot;
   let path = `${base}/blocks/${name}`;
+
+  MILO_BLOCKS_PRELOADS[name]?.forEach((preload) => loadLink(`${base}${preload}`, { rel: 'preload', as: 'script', crossorigin: 'anonymous' }));
 
   if (mep?.blocks?.[name]) path = mep.blocks[name];
 
@@ -1145,6 +1150,16 @@ async function documentPostSectionLoading(config) {
   document.body.appendChild(createTag('div', { id: 'page-load-ok-milo', style: 'display: none;' }));
 }
 
+function partition(arr, fn) {
+  return arr.reduce(
+    (acc, val, i, ar) => {
+      acc[fn(val, i, ar) ? 0 : 1].push(val);
+      return acc;
+    },
+    [[], []],
+  );
+}
+
 async function processSection(section, config, isDoc) {
   const inlineFrags = [...section.el.querySelectorAll('a[href*="#_inline"]')];
   if (inlineFrags.length) {
@@ -1158,8 +1173,10 @@ async function processSection(section, config, isDoc) {
   }
 
   if (section.preloadLinks.length) {
-    const preloads = section.preloadLinks.map((block) => loadBlock(block));
+    const [modals, nonModals] = partition(section.preloadLinks, (block) => block.classList.contains('modal'));
+    const preloads = nonModals.map((block) => loadBlock(block));
     await Promise.all(preloads);
+    modals.map((block) => loadBlock(block));
   }
 
   const loaded = section.blocks.map((block) => loadBlock(block));
--- a/test/utils/utils-partition.test.js
+++ b/test/utils/utils-partition.test.js
@@ -1,0 +1,27 @@
+import { expect } from '@esm-bundle/chai';
+import { partition } from '../../libs/utils/utils.js';
+
+describe('Utils Partition', () => {
+  it('partition array', async () => {
+    const arr = [
+      { name: 'John', age: 23 },
+      { name: 'James', age: 40 },
+      { name: 'Mary', age: 31 },
+    ];
+    const result = partition(arr, (x) => x.age > 30);
+    expect(result[0]).to.have.deep.members(
+      [
+        { name: 'Mary', age: 31 },
+        { name: 'James', age: 40 },
+      ],
+    );
+    expect(result[1]).to.have.deep.members(
+      [{ name: 'John', age: 23 }],
+    );
+  });
+  it('empty array', async () => {
+    const arr = [];
+    const result = partition(arr, (x) => x.age > 30);
+    expect(result).to.eql([[], []]);
+  });
+});


Blainegunn (approved):
fair enough, i like it @TsayAdobe

libs/utils/utils.js
```diff
@@ -1158,8 +1173,10 @@ async function processSection(section, config, isDoc) {
   }
 
   if (section.preloadLinks.length) {
-    const preloads = section.preloadLinks.map((block) => loadBlock(block));
+    const [modals, nonModals] = partition(section.preloadLinks, (block) => block.classList.contains('modal'));
+    const preloads = nonModals.map((block) => loadBlock(block));
     await Promise.all(preloads);
+    modals.map((block) => loadBlock(block));
```
auniverseaway, line 1179: **Nits:**

1. I think this particular part is more complicated than necessary. I find the partition function hard to read.
2. I think marquee block preload should not be mixed in the same PR as whatever we're trying to do here with modals. These are different problem spaces.

libs/utils/utils.js
```diff
@@ -452,6 +455,8 @@ export async function loadBlock(block) {
   const base = miloLibs && MILO_BLOCKS.includes(name) ? miloLibs : codeRoot;
   let path = `${base}/blocks/${name}`;
 
+  MILO_BLOCKS_PRELOADS[name]?.forEach((preload) => loadLink(`${base}${preload}`, { rel: 'preload', as: 'script', crossorigin: 'anonymous' }));
```
chrischrischris, line 458: If a block is used in the preload array, then the block.css will still be awaited and will negate the perf gain.  We should check for js vs css and set the `as` param accordingly

libs/utils/utils.js
chrischrischris: Since the arrays are very short and the `fn` is simple, `partition` could use `filter`.  This does make it so the array is parsed twice, but there should be a negligible change in js perf.

```
function partitionArray(arr, fn) {
  const matchedItems = arr.filter(fn);
  const unmatchedItems = arr.filter(item => !fn(item));
  return [matchedItems, unmatchedItems];
}
```

TsayAdobe: Thank you for your suggestion. The current function works as expected and there are tests to guard any changes in the future. It's a single purpose function which unlikely needs any more modification. I will leave it as is.

libs/utils/utils.js
TsayAdobe: The current solution is to preload imports in a block. decorate.js is always imported by marquee. It's not dealing with CSS.
```
const MILO_BLOCKS_PRELOADS = { marquee: ['/utils/decorate.js'] };
```

libs/utils/utils.js
JingleH: non-blocking nit: semantically maybe a forEach makes more sense than map for modals? It could save a tiny bit of memory too as a mapping array won't be created, but really negligible even if the engine doesn't optimize it away

JingleH (approved):

libs/utils/utils.js
```diff
@@ -87,6 +87,9 @@ const MILO_BLOCKS = [
   'share',
   'reading-time',
 ];
+
+const MILO_BLOCKS_PRELOADS = { marquee: ['/utils/decorate.js'] };
```
mokimo, line 91: Do we want to maintain a manual list of preloads for blocks? While this works as of right now, dependencies change, get added - removed - refactored and we might hurt performance eventually if we start preloading assets that might never get used.

I'm not convinced this is a great long term solution, but a quick band-aid that will lead to issues down the road.

mokimo:
@TsayAdobe I'm not blocking this, but there seem to be a few open concerns, so it would be great if you can give them some attention.

Additionally, a QA/QE should be testing this, rather than the PR initiator just applying the 'ready for stage' label. Especially since this touches `utils.js` which is consumed on every single milo consumer on every page. 

See the [contribution](https://github.com/adobecom/milo/wiki/Milo-Contribution-Process#getting-your-code-to-production) wiki
> QE: apply "Ready for Stage" label after testing

libs/utils/utils.js
narcis-radu: It is not dealing with CSS right now, but if we're looking for a long term solution we should consider this as well.

mokimo (changes requested):
Sadly the discussion never fully rounded up, but [I'd agree with](https://github.com/orgs/adobecom/discussions/2390#discussioncomment-9616557) @npeltier that we lose a lot of flexibility with this and start hardcoding for the sake of "quick" performance gains.

Could we trial this on a consumer and get some real world numbers before taking this up in milo core?
This also currently only considers the marquee and is very opinionated, while milo tries to stay agnostic of the usage of certain blocks, patterns, LCPs etc.

JingleH:
I personally find that in certain scenarios, hardcoding logic on the client side (without build process/bundler) for performance gains is worth it. After all, this is not the only scenario where we have decided to trade maintainability for performance. In Express, we have a metadata-controlled behavior that preloads several preload scripts/assets/js/css for modal SUSI-Light, and it saves tangible hundreds of ms when interacting and seconds on slow 3g devices (https://github.com/adobecom/express/pull/974/files). Difference with this is we preload in the Delay phase for responsiveness where in this PR it's marquee Eager LCP. As a future Milo customer, I would love to have this feature (for both Milo blocks and consumer blocks) and seems like it will be mostly remembering to validate the mapping whenever a block is updated. But I totally agree that it will encourage much more frequent edits to utils.js and violate certain principles like Close for Modification. This kind of churns is acceptable in some repos but might be too much for Milo as a consumed repo

libs/utils/utils.js
TsayAdobe: Yes, this is a temporary fix. Implementing a comprehensive solution will require a larger effort, potentially an epic. This approach aligns with the performance tiger team's process. Try out various solutions to identify what effectively improves performance.

CC: @jedjedjedM, @auniverseaway

TsayAdobe:
I agree with @JingleH . Currently, Milo lacks a build process, which means we rely on the browser to discover dependencies from scripts and content, resulting in prolonged loading times. To achieve a performance gain, we may need to consider trading off some maintainability.

mokimo:
@TsayAdobe I see the value in the performance gains, I'm just not sure it belongs into the milo utils (yet).

I'd love to see this implemented on a consumer first (pick 1-2 high traffic ones) in their scripts so that we can evaluate the real-world gains and track the improvement over a wide range of pages, this could be done using RUM data.

@overmyheadandbody @narcis-radu @robert-bogos anything to add?

TsayAdobe:
@mokimo 
* The consumer app relies on Milo's code to discover the dependency. Once the marquee block is discovered and loaded, we want to preload the scripts needed by the marquee. Having the consumer app redo the discovery doesn't make sense.
* Preloading is unlikely to break the app. It will either help or hurt performance, which is being monitored by the performance team.
* At this point, we are focused solely on improving the marquee. There are no maintainability issues.

robert-bogos (commented):
@TsayAdobe I think we should test this on some high-traffic consumers first and get LH numbers. If the improvement is not significant, it wouldn’t make sense to add something new to maintain in Milo.

mokimo:
> The consumer app relies on Milo's code to discover the dependency. Once the marquee block is discovered and loaded, we want to preload the scripts needed by the marquee. Having the consumer app redo the discovery doesn't make sense.

Most consumers already detect marquee's on their own already - so that would be a suitable place to trial this code. Let's take DC as an example: https://github.com/adobecom/dc/blob/stage/acrobat/scripts/scripts.js#L344-L353

I'd love to quantify the impact whether this is a 1LH improvement, 5LH points and then we can elevate this to milo core based on the gains.

mokimo:
As discussed in the performance meeting:
- Marquee preloading will be trialed on a consumer to quantify the LH score gains
- This PR will focus on the marquee & a new PR will be opened for the modal partitioning, to be reviewed separately.

I'll close this PR for now, as the marquee preloading trials are outstanding and would love to see [discussion](https://github.com/orgs/adobecom/discussions/2390#discussioncomment-9616557) updated once we have some numbers.
For the modal optimisations, I'm curious on seeing a new PR to see the gains of that!

PR #2496, opened 2024-06-20
[MWPW-145375][Security Fix] - Moved hardcoded secrets to milo config

Deva309:
## Description
Moved hardcoded secrets detected by kodiak in ganv-search JS files to `.milo` configs.

## Resolves: 
[MWPW-145375](https://jira.corp.adobe.com/browse/MWPW-145375)
[MWPW-145419](https://jira.corp.adobe.com/browse/MWPW-145419)
[MWPW-144311](https://jira.corp.adobe.com/browse/MWPW-144311)

**Test URLs:**
- After: https://mwpw-145375--milo--deva309.hlx.page/drafts/devashish/wam?martech=off

### DIFF
--- a/libs/blocks/global-navigation/features/search/gnav-search.js
+++ b/libs/blocks/global-navigation/features/search/gnav-search.js
@@ -10,12 +10,9 @@ import {
 import { replaceKeyArray } from '../../../../features/placeholders.js';
 import { getConfig } from '../../../../utils/utils.js';
 import { debounce } from '../../../../utils/action.js';
+import getServiceConfig from '../../../../utils/service-config.js';
 
 const CONFIG = {
-  suggestions: {
-    scope: 'adobecom',
-    apiKey: 'adobedotcom2',
-  },
   selectors: {
     hasResults: 'has-results',
     inputIsPopulated: 'feds-search-input--isPopulated',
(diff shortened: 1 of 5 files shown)


libs/blocks/global-navigation/features/search/gnav-search.js
```diff
@@ -122,12 +119,13 @@ class Search {
     });
   }
 
-  getSuggestions(query = this.query) {
+  async getSuggestions(query = this.query) {
     const { env } = getConfig();
+    const { search } = await getServiceConfig(window.location.origin);
```
narcis-radu, line 124: would this have any impact on performance?

libs/blocks/global-navigation/features/search/gnav-search.js
```diff
@@ -10,12 +10,9 @@ import {
 import { replaceKeyArray } from '../../../../features/placeholders.js';
 import { getConfig } from '../../../../utils/utils.js';
 import { debounce } from '../../../../utils/action.js';
+import getServiceConfig from '../../../../utils/service-config.js';
 
 const CONFIG = {
-  suggestions: {
-    scope: 'adobecom',
-    apiKey: 'adobedotcom2',
```
narcis-radu, line 17: these are not really secrets, I think we can suppress the kodiak alerts cc @salonijain3

sharmrj: Ideally I still don't think we should be exposing them. It makes it easy for some third party to send requests to that endpoint and causing issues.

libs/blocks/global-navigation/features/search/gnav-search.js
Deva309: Not really, because `getServiceConfig` uses cached data, and the `origin` is accessed through the `window` object. So, overall, I think there is no impact on performance.

Deva309:
Closing this PR.
As these are not secrets and marking these kind of issues as false positive.

PR #2501, opened 2024-06-22
Exposing the UNav Config for consumers to modify and reload the UNav if needed

amol-anand:
Exposing the UNav config to be consumed and updated by any consumer like below:

If a Milo Global Navigation consumer wants to change the signInClick callback or listen to messages from the Profile Menu sub-navigation (which is the case for us in Firefly), then we can get the `CONFIG.universalNav.universalNavConfig` object, update anything we need to update, and then call `window.UniversalNav.reload(updatedConfig)`

**Test URLs:**
- After: https://stage--milo--amol-anand.hlx.page/?martech=off

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -621,6 +621,7 @@ class Gnav {
       children: getChildren(),
     });
 
+    CONFIG.universalNav.universalNavConfig = getConfiguration();
     window.UniversalNav(getConfiguration());
 
     isDesktop.addEventListener('change', () => {
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -621,7 +621,7 @@ class Gnav {
       children: getChildren(),
     });
 
-    // exposing unav config for consumers to modify certain values/functions and reload UNAV if they need to
+    // Exposing UNAV config for consumers to modify certain values/functions and reload UNAV if needed
     CONFIG.universalNav.universalNavConfig = getConfiguration();
     window.UniversalNav(getConfiguration());
 
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -621,6 +621,7 @@ class Gnav {
       children: getChildren(),
     });
 
+    // exposing unav config for consumers to modify certain values/functions and reload UNAV if they need to
     CONFIG.universalNav.universalNavConfig = getConfiguration();
     window.UniversalNav(getConfiguration());
 
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -621,12 +621,12 @@ class Gnav {
       children: getChildren(),
     });
 
-    // Exposing UNAV config for consumers to modify certain values/functions and reload UNAV if needed
+    // Exposing UNAV config for consumers
     CONFIG.universalNav.universalNavConfig = getConfiguration();
-    window.UniversalNav(getConfiguration());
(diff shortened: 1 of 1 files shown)


libs/blocks/global-navigation/global-navigation.js
```diff
@@ -621,6 +621,8 @@ class Gnav {
       children: getChildren(),
     });
 
+    // Exposing UNAV config for consumers to modify certain values/functions and reload UNAV if needed
+    CONFIG.universalNav.universalNavConfig = getConfiguration();
     window.UniversalNav(getConfiguration());
 
     isDesktop.addEventListener('change', () => {
```
sharmrj, line 628: Would you want to change the `reload` inside this event listener callback to use `CONFIG.universalNav.universalNavConfig`? The user changing the viewport might overwrite your config during the unav reload.

amol-anand: that would still call getConfiguration() as it does today (same as window.UniversalNav(getConfiguration()). the only point of this is to add it to the CONFIG object so that it can be exported and consumed by others as needed.

sharmrj: My understanding is that you'll be making changes to the object referenced by `CONFIG.universalNav.universalNavConfig` but `getConfiguration` will generate a new object with it's own reference that won't contain any changes you made to `CONFIG.universalNav.universalNavConfig`, so this new object could "overwrite" the previous config you reloaded unav with.

sharmrj (approved):

salonijain3 (approved):

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -621,10 +621,12 @@ class Gnav {
...
     });
 
-    window.UniversalNav(getConfiguration());
+    // Exposing UNAV config for consumers
+    CONFIG.universalNav.universalNavConfig = getConfiguration();
+    window.UniversalNav(CONFIG.universalNav.universalNavConfig);
 
     isDesktop.addEventListener('change', () => {
-      window.UniversalNav.reload(getConfiguration());
+      window.UniversalNav.reload(CONFIG.universalNav.universalNavConfig);
```
overmyheadandbody, line 629: I'm not very familiar with this logic, but judging by its structure, it seems to indicate that the config could be different between viewports. This approach would just re-use the initial config. You could dispatch some new events when the config changes or define a new method in this file that handles all of the reload complexities. Consumers would then use that method instead of the regular unav one.

sharmrj: @overmyheadandbody It seems as though `getConfiguration` returns the same thing every time:
```javascript
    const getConfiguration = () => ({
      target: this.blocks.universalNav,
      env: environment,
      locale,
      imsClientId: window.adobeid?.client_id,
      theme: 'light',
      onReady: () => {
        this.decorateAppPrompt({ getAnchorState: () => window.UniversalNav.getComponent?.('app-switcher') });
      },
      analyticsContext: {
        consumer: {
          name: 'adobecom',
          version: '1.0',
          platform: 'Web',
          device: getDevice(),
          os_version: navigator.platform,
        },
        event: { visitor_guid: visitorGuid },
        onAnalyticsEvent,
      },
      children: getChildren(),
    });
```
It seems that we simply call `getConfiguration` again because the unav's reload method requires it. But if modifications have been made to the original configurations, we don't want those to be lost if the viewport is changed.

spadmasa:
regression done on unav  
https://main--cc--adobecom.hlx.page/creativecloud?milolibs=stage--milo--amol-anand
image
https://main--cc--adobecom.hlx.page/creativecloud?milolibs=stage--milo--amol-anand&unavVersion=1.2
image
logged in
https://main--cc--adobecom.hlx.page/products/photoshop/ai?milolibs=stage--milo--amol-anand
image
image
image

PR #2508, opened 2024-06-25
MWPW-147270 Fix the error caused by trying to load Search in Global Navigation

sharmrj:
This bug was filed in response to this comment: https://github.com/adobecom/milo/pull/2185#pullrequestreview-2019744236

- In mobile viewports, search is now only loaded upon clicking the trigger if the `.search` element is present in the global-navigation content

Resolves: [MWPW-147270](https://jira.corp.adobe.com/browse/MWPW-147270)

**Test URLs:**
- After: https://fix-search-error--milo--sharmrj.hlx.page/?martech=off

QA: https://stage--cc--adobecom.hlx.page/products/photoshop/ai?milolibs=fix-search-error--milo--sharmrj

Note: No behavior is changed by this PR.

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -650,7 +650,9 @@ class Gnav {
   };
 
   loadSearch = () => {
-    if (this.blocks?.search?.instance) return null;
+    const instanceAlreadyExists = !!this.blocks?.search?.instance;
+    const searchNotInContent = !this.searchPresent();
+    if (instanceAlreadyExists || searchNotInContent) return null;
 
     return this.loadDelayed().then(() => {
       this.blocks.search.instance = new this.Search(this.blocks.search.config);
@@ -966,10 +968,10 @@ class Gnav {
     return this.elements.breadcrumbsWrapper;
   };
 
-  decorateSearch = () => {
-    const searchBlock = this.content.querySelector('.search');
+  searchPresent = () => !!this.content.querySelector('.search');
 
-    if (!searchBlock) return null;
+  decorateSearch = () => {
+    if (!this.searchPresent()) return null;
 
     this.blocks.search.config.trigger = toFragment`
       <button class="feds-search-trigger" aria-label="Search" aria-expanded="false" aria-controls="feds-search-bar" daa-ll="Search">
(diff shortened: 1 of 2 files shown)


libs/blocks/global-navigation/global-navigation.js
```diff
@@ -650,7 +650,9 @@ class Gnav {
   };
 
   loadSearch = () => {
-    if (this.blocks?.search?.instance) return null;
+    const instanceAlreadyExists = !!this.blocks?.search?.instance;
+    const searchNotInContent = !this.searchPresent();
+    if (instanceAlreadyExists || searchNotInContent) return null;
```
nishantka, line 655: minor: can directly use `!this.searchPresent()` in `if` condition instead of creating variable

overmyheadandbody (commented):
I had the same comment for another search-related PR: is there any data supporting any investment in it? It was decommissioned across most adobe.com sites. Blog and Bacom-blog use the previous Gnav version. I'd suggest some logs are added to check out its usage and eventually delete it. Block would probably still need to be in place to avoid issues for any old experiences that still have it.

nishantka (approved):

Deva309 (approved):

sharmrj:
> I had the same comment for another search-related PR: is there any data supporting any investment in it? It was decommissioned across most adobe.com sites. Blog and Bacom-blog use the previous Gnav version. I'd suggest some logs are added to check out its usage and eventually delete it. Block would probably still need to be in place to avoid issues for any old experiences that still have it.

I agree that we should move in favor of eventually removing search in it's entirety from global-navigation, and I'd argue that this PR is a step in that direction. Currently on mobile viewports we always try to load search after the user clicks on the trigger, regardless of whether the author has configured search to load or not; with this change search will only load if the author has explicitly configured it.

mokimo (approved):
I'd agree with @overmyheadandbody and see if search is still used // if not decomission//remove it completely.

spadmasa:
Validation done on product pages with given repo , no console error is seen , verified in device and desktop
https://stage--cc--adobecom.hlx.page/products/photoshop/ai?milolibs=fix-search-error--milo--sharmrj
https://main--cc--adobecom.hlx.page/products/photoshop?milolibs=fix-search-error--milo--sharmrj
https://main--cc--adobecom.hlx.page/creativecloud?milolibs=fix-search-error--milo--sharmrj
https://main--cc--adobecom.hlx.page/products/illustrator?milolibs=fix-search-error--milo--sharmrj

PR #2522, opened 2024-06-26
Adding support for federated urls in modals

bandana147:
* Adds support for federated URLs in modals to centralize content for the region selector.

Resolves: https://jira.corp.adobe.com/browse/MWPW-130787

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/blaishram/document?martech=off#langnav
- After: https://MWPW-130787--milo--bandana147.hlx.page/drafts/blaishram/document?martech=off#langnav

### DIFF
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -1,5 +1,6 @@
 /* eslint-disable import/no-cycle */
 import { createTag, getMetadata, localizeLink, loadStyle, getConfig } from '../../utils/utils.js';
+import { getFederatedUrl } from '../global-navigation/utilities/utilities.js';
 
 const FOCUSABLES = 'a:not(.hide-video), button, input, textarea, select, details, [tabindex]:not([tabindex="-1"]';
 const CLOSE_ICON = `<svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 20 20">
@@ -82,7 +83,7 @@ function getCustomModal(custom, dialog) {
 }
 
 async function getPathModal(path, dialog) {
-  const block = createTag('a', { href: path });
+  const block = createTag('a', { href: getFederatedUrl(path) });
   dialog.append(block);
 
   // eslint-disable-next-line import/no-cycle


sharmrj:
I'm not super familiar with the details of this ticket, so lmk if any of my implicit assumptions are incorrect.

Is it wise to federate all urls going into `getPathModal`? Should there be some choice given to the caller of the function, or is there some way to defer the decision of whether a given fragment url should be federated or not further downstream (so that any url given to `getPathModal` has already been federated should it need to be)?

bandana147:
> I'm not super familiar with the details of this ticket, so lmk if any of my implicit assumptions are incorrect.
> 
> Is it wise to federate all urls going into `getPathModal`? Should there be some choice given to the caller of the function, or is there some way to defer the decision of whether a given fragment url should be federated or not further upstream (so that any url given to `getPathModal` has already been federated should it need to be)?

Currently, modals work only with same-domain URLs. With content centralization, we need the ability to load URLs from the federal folder. What getFederatedContentRoot does is check if the provided path contains "federal". If it does, then only it adds the content root of "federal", which would fetch the fragment from the federal repository. 
The same approach has been used for rendering fragments. - [here](https://github.com/adobecom/milo/blob/8b46dea80d3db2eb5ef78d1ee3876200db4f605a/libs/blocks/global-navigation/utilities/utilities.js#L367)

sharmrj (approved):

libs/blocks/modal/modal.js
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable import/no-cycle */
 import { createTag, getMetadata, localizeLink, loadStyle, getConfig } from '../../utils/utils.js';
+import { getFederatedUrl } from '../global-navigation/utilities/utilities.js';
```
narcis-radu, line 3: We might want to think about exporting the `getFederatedUrl` method to a more generic utilities and decouple it from the Global Navigation.

bandana147: Is it okay if i move this out after this [PR](https://github.com/adobecom/milo/pull/2523) to avoid conflict, as it also touches the utils.js files?

narcis-radu: Thanks, @bandana147! Is there a JIRA ticket already in your backlog?

bandana147: Created one - https://jira.corp.adobe.com/browse/MWPW-153670

bandana147:
Closing this as this is failing a test whose fix would require the merge of this [PR](https://github.com/adobecom/milo/pull/2523). Will open another PR after it is merged.

PR #2523, opened 2024-06-26
Centralize Geo Routing Modals

bandana147:
* Picks georouting json from feds folder if georouting and georoutingv2 file is not found in consumers site's folder.
* Addressed 2 todos:
     - Moved _getFederatedContentRoot_ to utils.js
     - Allows consumers to add their own allowed origins

Resolves: https://jira.corp.adobe.com/browse/MWPW-151542

**Test URLs:**
- Before: https://main--genuine--adobecom.hlx.page/georouting?martech=off
- After: https://main--genuine--adobecom.hlx.page/georouting?milolibs=georouting-central--milo--bandana147&martech=off

Since the before is not reproducible in folders in which georouting already exists, so we can't reproduce the before url in milo repo. So adding below urls for psi checks to pass:
- Before: https://main--milo--adobecom.hlx.page/georouting?martech=off
- After: https://georouting-central--milo--bandana147.hlx.page/drafts/blaishram/document1&martech=off

### DIFF
--- a/libs/features/georoutingv2/georoutingv2.js
+++ b/libs/features/georoutingv2/georoutingv2.js
@@ -287,7 +287,7 @@ export default async function loadGeoRouting(
   if (!resp.ok) {
     resp = await fetch(`${config.contentRoot ?? ''}/georouting.json`);
     if (!resp.ok) {
-      resp = await fetch(`${getFederatedContentRoot()}/georouting/georoutingv2.json`);
+      resp = await fetch(`${getFederatedContentRoot()}/federal/georouting/georoutingv2.json`);
       if (!resp.ok) return;
     } else {
       const json = await resp.json();
--- a/libs/features/georoutingv2/georoutingv2.js
+++ b/libs/features/georoutingv2/georoutingv2.js
@@ -1,4 +1,4 @@
-import { getFederatedContentRoot } from '../../blocks/global-navigation/utilities/utilities.js';
+import { getFederatedContentRoot } from '../../utils/utils.js';
 
 let config;
 let createTag;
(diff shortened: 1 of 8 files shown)


libs/features/georoutingv2/georoutingv2.js
```diff
@@ -280,12 +282,19 @@ export default async function loadGeoRouting(
...
+    if (!resp.ok) {
+      resp = await fetch(`${getFederatedContentRoot()}/federal/georouting/georoutingv2.json`);
+      if (!resp.ok) return;
+    } else {
+      const json = await resp.json();
+      // eslint-disable-next-line import/no-cycle
+      const { default: loadGeoRoutingOld } = await import('../georouting/georouting.js');
+      loadGeoRoutingOld(config, createTag, getMetadata, json);
+      return;
+    }
```
sharmrj, line 297: Should the failover order be `georoutingv2 -> geouroutingv1 -> georoutingv2Federated` or `georoutingv2 -> georoutingv2Federated -> georoutingv1`? It's not obvious to me which one it should be from reading the ticket.

If it were to be the latter you could do something like this:
```javascript
let json;
try {
  json = await loadGeoRoutingJson(...args); // loads either non-federal, federal, or throws error
} catch (e) {
  // maybe some logging in case resp.json() errors out
  loadGeoRoutingOld(...args);
  return;
}
```
This would put the fetching of individual urls back in the hands of the respective functions (which means you won't have to change georouting.js at all), it would clean up the above logic, and it would handle `await resp.json()` potentially erroring out without being handled.

Overall I think a structure like this would make the code both safer (as in able to fail gracefully should something unexpected happen in a network request, say in a slow network, that causes either the fetch or the `resp.json()` to fail) and easier to read.

bandana147: The failover would be this - georoutingv2 -> geouroutingv1 -> georoutingv2Federated. This is why we need to check both versions first, and if neither is found, then fetch from the federal repository.

bandana147:
Closing this and opening a new.

PR #2524, opened 2024-06-26
[MWPW-152278] Avoid empty CSS requests

overmyheadandbody:
It has been noticed that requests are made for empty CSS files. For performance purposes, we want to avoid such requests, since they only add overhead and aren't contributing any advantages.

The initial scope of the story was focused on the `fragments.css` file, but a few more auto blocks follow the same pattern, so it made sense to update all of them in one go. Here are the files in question:
Screenshot 2024-06-26 at 15 58 07

Note that `event-rich-results` follows a different pattern and it should not be an issue at this point.

**BEFORE**
Filtering network calls by `/tiktok|slideshare|fragment|marquee/`

Screenshot 2024-06-26 at 16 21 28

**AFTER**
Filtering network calls by `/tiktok|slideshare|fragment|marquee/`

Screenshot 2024-06-26 at 16 21 58

Resolves: [MWPW-152278](https://jira.corp.adobe.com/browse/MWPW-152278)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ramuntea/empty-css-showcase?martech=off&georouting=off
- After: https://no-empty-css-load--milo--overmyheadandbody.hlx.page/drafts/ramuntea/empty-css-showcase?martech=off&georouting=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -92,16 +92,16 @@ const AUTO_BLOCKS = [
   { gist: 'https://gist.github.com' },
   { caas: '/tools/caas' },
   { faas: '/tools/faas' },
-  { fragment: '/fragments/' },
+  { fragment: '/fragments/', voidStyles: true },
   { instagram: 'https://www.instagram.com' },
-  { slideshare: 'https://www.slideshare.net' },
-  { tiktok: 'https://www.tiktok.com' },
+  { slideshare: 'https://www.slideshare.net', voidStyles: true },
+  { tiktok: 'https://www.tiktok.com', voidStyles: true },
   { twitter: 'https://twitter.com' },
   { vimeo: 'https://vimeo.com' },
   { vimeo: 'https://player.vimeo.com' },
   { youtube: 'https://www.youtube.com' },
   { youtube: 'https://youtu.be' },
-  { 'pdf-viewer': '.pdf' },
+  { 'pdf-viewer': '.pdf', voidStyles: true },
   { video: '.mp4' },
   { merch: '/tools/ost?' },
 ];
@@ -448,6 +448,7 @@ export async function loadBlock(block) {
   }
 
   const name = block.classList[0];
+  const voidStyles = block.classList.contains('void-styles');
   const { miloLibs, codeRoot, mep } = getConfig();
 
   const base = miloLibs && MILO_BLOCKS.includes(name) ? miloLibs : codeRoot;
@@ -457,9 +458,9 @@ export async function loadBlock(block) {
 
   const blockPath = `${path}/${name}`;
 
-  const styleLoaded = new Promise((resolve) => {
+  const styleLoaded = !voidStyles ? new Promise((resolve) => {
     loadStyle(`${blockPath}.css`, resolve);
-  });
+  }) : '';
 
   const scriptLoaded = new Promise((resolve) => {
     (async () => {
@@ -477,6 +478,7 @@ export async function loadBlock(block) {
       resolve();
     })();
   });
+
   await Promise.all([styleLoaded, scriptLoaded]);
   return block;
 }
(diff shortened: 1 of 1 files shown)


libs/utils/utils.js
```diff
@@ -92,16 +92,16 @@ const AUTO_BLOCKS = [
   { gist: 'https://gist.github.com' },
   { caas: '/tools/caas' },
   { faas: '/tools/faas' },
-  { fragment: '/fragments/' },
+  { fragment: '/fragments/', voidStyles: true },
```
auniverseaway, line 95: To save bytes, I would advocate for `styles: false` or `stylesOff`. Void would be new vocabulary for engineering and GWP.

overmyheadandbody: Sure, thanks, sounds good, I was actually mulling over this as well

libs/utils/utils.js
```diff
@@ -448,6 +448,7 @@ export async function loadBlock(block) {
   }
 
   const name = block.classList[0];
+  const voidStyles = block.classList.contains('void-styles');
```
auniverseaway, line 451: Do you have a scenario in the real world where this would be used?

overmyheadandbody: Not by GWP, no. The blocks not requiring styles are auto blocks, so they get an additional class that hints that styles shouldn't be loaded. Another way to do this is have a check here on whether the current block is an auto block and then see if styles are needed. Do you prefer the latter or have an even better idea?

libs/utils/utils.js
```diff
@@ -457,9 +458,9 @@ export async function loadBlock(block) {
 
   const blockPath = `${path}/${name}`;
 
-  const styleLoaded = new Promise((resolve) => {
+  const styleLoaded = !voidStyles ? new Promise((resolve) => {
     loadStyle(`${blockPath}.css`, resolve);
-  });
+  }) : '';
```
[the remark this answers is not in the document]
overmyheadandbody: I did initially go on a similar path, but eventually settled for this less invasive one, I think this whole method needs a revamp, but I'm not sure now is the right time to do it.

overmyheadandbody:
I slightly updated the approach. I'm not convinced of having to iterate through `AUTO_BLOCKS` each time we're loading a block; I did consider defining another array with blocks that don't need styles, but the performance gains are not truly significant and I chose to avoid duplication. However, one advantage to this approach would be that that array would not consider just `AUTO_BLOCKS`, we can define any block in there. It's not an issue right now, so I'd still avoid the duplication.
Screenshot 2024-06-27 at 12 43 58

The current approach also means that if consumers apply the `styles` property to their own `AUTO_BLOCKS`, the logic would not take it into account. This is not a showstopper right now, since only Bacom define a couple of additional `AUTO_BLOCKS` [here](https://github.com/adobecom/bacom/blob/69e3c0b5ce567dcb6213cd2f821d79b5446ed698/scripts/scripts.js#L132).

Wdyt?

narcis-radu (approved):

libs/utils/utils.js
```diff
@@ -448,6 +448,7 @@ export async function loadBlock(block) {
   }
 
   const name = block.classList[0];
+  const hasStyles = AUTO_BLOCKS.find((ab) => Object.keys(ab).includes(name))?.styles ?? true;
```
robert-bogos, line 451: I think we can make this shorter
```suggestion
  const hasStyles = !!AUTO_BLOCKS.find((ab) => Object.keys(ab).includes(name))?.styles;
```

robert-bogos (approved):
Looks good 👍 I left just a small suggestion

libs/utils/utils.js
robert-bogos: Or you can just leave it like this: 
`const hasStyles = AUTO_BLOCKS.find((ab) => Object.keys(ab).includes(name))?.styles`
The logic will still work

overmyheadandbody: We don't define a `styles` property for auto blocks that _do_ need them. So `styles` would be `undefined` and `!!undefined` will be `false`, when in fact we do want styles to be loaded unless explicitly stated otherwise.

robert-bogos: Got it 👍

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-152278

mokimo (approved):

PR #2530, opened 2024-06-28
MWPW-146962 [milo] Text link behaving like button in FAQ section

bozojovicic:
When checkout URL contains `#_tcl` it should be transformed to ordinary HTML link and not to CTA. At the moment some not desirable styles are applied to that link e.g. `width:100%;text-decoration:none` etc. which I fixed with this change.
TCL is an acronym for Text Checkout Link.

**BEFORE** :
Screenshot 2024-06-28 at 10 32 48

**AFTER** :
Screenshot 2024-06-28 at 10 33 07

Resolves: [MWPW-146962](https://jira.corp.adobe.com/browse/MWPW-146962)

**Test URLs:**
- After: https://mwpw146962--milo--bozojovicic.hlx.page/drafts/bozo/price-cta?martech=off


libs/blocks/merch/merch.css
```diff
@@ -15,7 +15,7 @@ span.placeholder-resolved[data-template="strikethrough"], span.price.price-strik
   visibility: inherit;
 }
 
-a[is='checkout-link'] > span {
+a[is='checkout-link']:not(.text-link) > span {
```
yesil, line 18: I think you can use `.con-button` here, since it is already added to CTA like co links.

libs/blocks/merch/merch.js
```diff
@@ -452,6 +452,7 @@ export async function buildCta(el, params) {
   const cta = service.createCheckoutLink(context, text);
   if (el.href.includes('#_tcl')) {
     el.href = el.href.replace('#_tcl', '');
+    cta.classList.add('text-link');
```
yesil, line 455: this should not be necessary with the above.

libs/blocks/merch/merch.css
bozojovicic: Right! Fixed.

libs/blocks/merch/merch.js
bozojovicic: Indeed

mirafedas (approved):

yesil (approved):

Axelcureno (approved):

overmyheadandbody:
@bozojovicic - NALA tests are failing for this PR, thus it will not get merged to Stage. You can try to rebase your branch with the current `stage` one, which usually does the trick. If not, you'll have to reach out to NALA to see where the issue is.

afmicka:
@overmyheadandbody @skumar09 @Dli3 the Nala failures are not related to the changes in this PR but are blocking it from being merged.

Table tests are failing due to the updated classes in chrome for the highlight heading, the daily run is failing with the same error. I guess the tests need an update. Could you please check? 

HowTo tests are failing because of the expected class for the image. Sometime on forks, it is resolved as 'how-to-image' instead of expected 'how-to-media'. Not sure how it comes to that, but have seen it before. 

Could you please confirm here and unblock this PR (and similar ones) for merging?

PR #2531, opened 2024-06-28
Central georouting support

bandana147:
* Picks georouting json from feds folder if georouting and georoutingv2 file is not found in consumers site's folder.
* Addressed 2 todos:
     - Moved _getFederatedContentRoot_ to utils.js
     - Allows consumers to add their own allowed origins

Resolves: https://jira.corp.adobe.com/browse/MWPW-151542

**Test URLs:**

- After: https://central-georouting--milo--bandana147.hlx.page/?martech=off

QA: https://main--genuine--adobecom.hlx.page/georouting?milolibs=central-georouting--milo--bandana147&martech=off


sharmrj (approved):

Deva309 (approved):

nishantka (approved):

libs/utils/utils.js
```diff
@@ -1270,4 +1270,32 @@ export function loadLana(options = {}) {
   window.addEventListener('unhandledrejection', lanaError);
 }
 
+const defaultAllowedOrigins = [
```
narcis-radu, line 1273: The allowed origins are strictly related to federal content, so we either rename the constant or we move it within the method.

libs/features/georoutingv2/georoutingv2.js
```diff
@@ -281,12 +283,19 @@ export default async function loadGeoRouting(
...
   loadStyle = loadStyleFunc;
 
-  const resp = await fetch(`${config.contentRoot ?? ''}/georoutingv2.json`);
+  let resp = await fetch(`${config.contentRoot ?? ''}/georoutingv2.json`);
   if (!resp.ok) {
-    // eslint-disable-next-line import/no-cycle
-    const { default: loadGeoRoutingOld } = await import('../georouting/georouting.js');
-    loadGeoRoutingOld(config, createTag, getMetadata);
-    return;
+    resp = await fetch(`${config.contentRoot ?? ''}/georouting.json`);
```
narcis-radu, line 288: I think we're mixing scope here and this will eventually be hard to remove when we fully deprecate the old geo-routing feature.

libs/features/georouting/georouting.js
```diff
@@ -129,24 +129,20 @@ async function showModal(details) {
...
   const urlLocale = locale.prefix.replace('/', '');
   const storedInter = sessionStorage.getItem('international') || getCookie('international');
   const storedLocale = storedInter === 'us' ? '' : storedInter;
 
-  const resp = await fetch(`${config.contentRoot ?? ''}/georouting.json`);
-  if (!resp.ok) return;
-  const json = await resp.json();
-
-  const urlGeoData = json.data.find((d) => d.prefix === urlLocale);
+  const urlGeoData = geoDetails.data.find((d) => d.prefix === urlLocale);
```
narcis-radu, line 139: would be great to use optional chaining to make sure it doesn't throw an error here

narcis-radu (commented):
Nice work, @bandana147. I have added some concerns.

libs/utils/utils.js
```diff
@@ -1270,4 +1270,32 @@ export function loadLana(options = {}) {
...
+  if (origin.includes('localhost') || origin.includes('.hlx.')) {
+    // Akamai as proxy to avoid 401s, given AEM-EDS MS auth cross project limitations
+    federatedContentRoot = origin.includes('.hlx.live')
+      ? 'https://main--federal--adobecom.hlx.live'
+      : 'https://www.stage.adobe.com';
+  }
+
+  return federatedContentRoot;
+};
+
```
mokimo, line 1300: IMO we should have a dynamic import for the whole federated content or move it to its own separate module.

I would guess that most pages do **not** use federal content and thus shouldn't load it at all. The blocks that do, should load it themselves

mokimo (changes requested):
We are already overloading the utils and should avoid adding into it.

The federated helpers are only used by a handful of blocks and the blocks that use this feature should just import a separate shared helper file that has all those methods.

Please move the code out of the utils and into it's own feature/helper file that can be imported

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -84,27 +74,6 @@ export function toFragment(htmlStrings, ...values) {
   return fragment;
 }
 
-// TODO we might eventually want to move this to the milo core utilities
```
mokimo, line 87: The TODO is a bit misleading, the utils isn't the best place for this. I've commented below on a better alternative

mokimo (commented):
Thanks for tackling this!! This has been a long standing task on my personal backlog, love to see finally finding a better home for the code to allow for better consumption of the feature outside of the global navigation.

libs/blocks/global-navigation/utilities/utilities.js
bandana147: So, you are suggesting we move it to a new file, something like `libs/blocks/global-navigation/features/helpers.js` file or `libs//features/helpers.js` ?

libs/features/georoutingv2/georoutingv2.js
bandana147: To provide backward compatibility, we still need to ensure that fetch(${config.contentRoot ?? ''}/georoutingv2.json) doesn't exist before we go ahead and load the federal georoutingv2 file.
I am not seeing any other way to handle this case without introducing additional configuration from consumer. However, I am open to suggestions.

libs/utils/utils.js
```diff
@@ -1270,4 +1270,32 @@ export function loadLana(options = {}) {
   window.addEventListener('unhandledrejection', lanaError);
 }
 
+const defaultAllowedOrigins = [
+  'https://www.adobe.com',
+  'https://business.adobe.com',
+  'https://blog.adobe.com',
```
auniverseaway, line 1276: Please remove blog from this list. Blog does not use GeoRouting.

libs/blocks/global-navigation/utilities/utilities.js
mokimo: Since the federal feature should be used outside of the global-navigation, I'd suggest `libs/utils/federalUtils.js` (or a better name)

libs/features/georoutingv2/georoutingv2.js
```diff
@@ -281,12 +283,19 @@ export default async function loadGeoRouting(
...
+    if (!resp.ok) {
+      resp = await fetch(`${getFederatedContentRoot()}/federal/georouting/georoutingv2.json`);
+      if (!resp.ok) return;
+    } else {
+      const json = await resp.json();
+      // eslint-disable-next-line import/no-cycle
+      const { default: loadGeoRoutingOld } = await import('../georouting/georouting.js');
+      loadGeoRoutingOld(config, createTag, getMetadata, json);
+      return;
+    }
```
mokimo, line 298: I don't have the full context of why the old georouting is required, but here a suggestion that would provide a bit more clarity on the load order, avoiding a nested if/else structure:

```js
const urls = [
  `${config.contentRoot ?? ''}/georoutingv2.json`,
  `${config.contentRoot ?? ''}/georouting.json`,
  `${getFederatedContentRoot()}/federal/georouting/georoutingv2.json`
];

for (const url of urls) {
  let resp = await fetch(url);
  if (resp.ok) {
    if (url.includes('georouting.json')) {
      // eslint-disable-next-line import/no-cycle
      const { default: loadGeoRoutingOld } = await import('../georouting/georouting.js');
      loadGeoRoutingOld(config, createTag, getMetadata, json);
    }
    return;
  }
}
```

The way I understand it, we want to load georoutingv2, if that's not available the legacy georouting and then the federal georouting as backup?

mokimo (approved):
Looking better now - a small suggestion to improve clarity on where/which georouting is loaded.

Can georouting be an LCP element? In this case, it might have some LH score impact. Would be cool to try to do some testing around that.

libs/utils/fedsHelpers.js
```diff
@@ -0,0 +1,29 @@
+import { getConfig } from './utils.js';
+
+let federatedContentRoot;
+/* eslint-disable import/prefer-default-export */
+export const getFederatedContentRoot = () => {
+  const defaultAllowedOrigins = [
+    'https://www.adobe.com',
+    'https://business.adobe.com',
+    'https://milo.adobe.com',
+    'https://news.adobe.com',
```
mokimo, line 10: I think you can re-add blog here, [Chris' comment](https://github.com/adobecom/milo/pull/2531#discussion_r1669375066) resolved around blog not having the georouting feature.
From an akamai perspective, it's set up to support federal features though to improve loading times.

Good that we moved this into a separate file, since it already caused confusion 😁

libs/features/georoutingv2/georoutingv2.js
bandana147: Yes, correct. Agree look cleaner and readable, will update it.

libs/features/georoutingv2/georoutingv2.js
narcis-radu: We could always configure a project and say pick up the georouting from federal via metadata.

libs/utils/fedsHelpers.js
```diff
@@ -0,0 +1,30 @@
+import { getConfig } from './utils.js';
```
narcis-radu, line 1: @bandana147 - would it be possible to rename the file to `federated` or something else that's more appropriate? Thanks!

bandana147: Done

PR #2538, opened 2024-07-04
[MWPW-151517] - Remove condition for promobar hidden on mobile from gnav

Deva309:
## Description
Removed the condition that hides the gnav promobar on mobile. We now have the ability to set separate copies for each viewport. The acceptance criteria to on and off the promobar only for mobile comes with a CLS issue, so it is not part of this PR, as discussed with the PDMs.
For example:
Screenshot 2024-07-16 at 10 30 54 AM

To avoid CLS impact, we need to adhere to the character count limits specified in the design [specs](https://www.figma.com/proto/yKDQl6NMZYM1pTwcQcwwRD/Feds---Promos?page-id=17%3A1839&node-id=18-2566&t=yLSphbWkVj0UKM6k-0&scaling=scale-down-width&content-scaling=fixed&starting-point-node-id=18%3A2566) for the feds promo bar, also mentioned in the ticket. Additionally, this information has been included in the authoring documentation for the aside promo bar.
Screenshot 2024-07-04 at 10 29 25 AM

Resolves: 
- [MWPW-151517](https://jira.corp.adobe.com/browse/MWPW-151517)
- [MWPW-153148](https://jira.corp.adobe.com/browse/MWPW-153148)
- [MWPW-154090](https://jira.corp.adobe.com/browse/MWPW-154090) (Bug Fix)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.live/drafts/devashish/promo-bar/feds-promobar?martech=off
- After: https://mwpw-151517-promobar--milo--deva309.hlx.live/drafts/devashish/promo-bar/feds-promobar?martech=off


libs/styles/styles.css
```diff
@@ -705,7 +705,6 @@ header:not(.global-navigation) ~ main {
   all of the ':not(.global-navigation)' rules can be entirely removed.
  */
 header.global-navigation {
-  height: var(--global-height-nav);
```
overmyheadandbody, line 708: I doubt this is what is desired, as removing this will increase CLS for all adobe.com pages. This change alone adds `0.03` CLS. Maybe that sounds negligible, but, per https://web.dev/articles/cls:
> To provide a good user experience, sites should strive to have a CLS score of 0.1 or less

So this change would essentially eat up 30% of the total CLS allowance on **all** our pages.

You likely don't want to tamper with this height. If the promo height should be different between viewports, you'd want to redefine `--global-height-navPromo` in https://github.com/adobecom/milo/blob/429001e956cdd9e4a36d4e8f6426c9e432ee5700/libs/styles/styles.css#L7 and adapt it based on media queries, to avoid CLS issues.

overmyheadandbody (changes requested):

libs/styles/styles.css
Deva309: @overmyheadandbody 
Added `global-navigation` height back to resolve that CLS issue.
Adjusted height `--global-height-navPromo` for feds promobar.

libs/styles/styles.css
```diff
@@ -4,7 +4,7 @@
   --global-height-nav: 64px;
   --global-height-breadcrumbs: 33px;
   /* stylelint-disable-next-line custom-property-pattern */
-  --global-height-navPromo: 65px;
+  --global-height-navPromo: 72px;
```
mokimo, line 7: Have there been any considerations regarding the two-line, 80px use-case utilizing a min-height here?

test/blocks/global-navigation/global-navigation.test.js
```diff
@@ -227,15 +227,6 @@ describe('global navigation', () => {
       });
       promoMeta.remove();
     });
-
-    it('doesn\'t exist on mobile', async () => {
```
mokimo, line 231: Instead of removing it, how about we write a test that the promo exists on mobile?

Deva309: Added unit tests for both cases: when it exists and when it does not exist on mobile.

test/blocks/global-navigation/global-navigation.test.js
```diff
@@ -229,12 +229,27 @@ describe('global navigation', () => {
...
+      expect(nav.block.classList.contains('has-promo')).to.be.true;
       const asideElem = nav.block.querySelector('.aside.promobar');
-      expect(asideElem).to.not.exist;
+      expect(asideElem).to.be.exist;
+      const prmobarNotOnMobile = asideElem.querySelector('.mobile-up').classList.contains('hide-block');
+      expect(prmobarNotOnMobile).to.be.true;
+      promoMeta.remove();
+    });
+
+    it('Exist on mobile', async () => {
```
sharmrj, line 243: I think we could have more descriptive test names that tell the reader when the promobar should or should not exist on mobile, i.e. when we have content for mobile

sharmrj (approved):

bandana147 (approved):

mokimo:
Let's wait for the consonant/design sign-off on the ticket before merging this. From what I've seen, once you got a branch ready - you send that over on the Jira, have them sign off and then raise the PR. @ryanmparrish would that be the correct flow involving any (design) changes for blocks?

prativas22:
Validation done on the below url  for promobar issue on mobile viewport after fix
https://mwpw-151517-promobar--milo--deva309.hlx.page/drafts/prativa/feds-promobar
image
when the mobile copy is empty, the gnav promobar is hidden on mobile viewports.
https://mwpw-151517-promobar--milo--deva309.hlx.page/drafts/prativa/test-feds-promobar
image

prativas22:
Validated the promobar functionality in Devices and observed that after enabling the promobar for the device, the Gnav does not show up on both the iPhone and the iPad. For Gnav, the iPhone does not display the Hamberg symbol.

Note: Android device is working as expected

URL:
https://mwpw-151517-promobar--milo--deva309.hlx.page/drafts/prativa/aside-promo-bar

iphone device: (both safari and chrome browser have same result)
Safari browser:
image
landscape orientation: no Gnav 
image
iPad Device:
Safari
image

In iPad landscape orientation, the gnav displays as expected:
image

A JIRA ticket has been created to track the details. Refer to https://jira.corp.adobe.com/browse/MWPW-154090 for further details. Validation would be done after the bug fix.

@salonijain3 @sigadamvenkata

mokimo:
@Deva309 there seems to be design feedback on the Ticket

Deva309:
> @Deva309 there seems to be design feedback on the Ticket

We will not incorporate any design feedback till we have design sign-off for aside block promobar variants as these fixes are part aside block.
Additionally, to avoid the CLS issue when for the promobar on and off for mobile vieport, we have agreed to either display the promobar across all viewports or not at all. Therefore, the mobile condition has been removed with this change, and the height has been adjusted for the gnav accordingly starting from mobile viewport.

@mokimo Please review this PR

prativas22:
The promobar functionality on devices is verified after the turning on and off mobile criteria were removed.
URL:
https://mwpw-151517-promobar--milo--deva309.hlx.page/drafts/prativa/feds-promobar-one (one cell)
https://mwpw-151517-promobar--milo--deva309.hlx.page/drafts/prativa/aside-promo-bar (3 cells)
iphone device: 
Safari browser:
image
image

ipad:
image
image

mokimo (approved):

narcis-radu:
>To avoid CLS impact, we need to adhere to the character count limits specified in the design [specs](https://www.figma.com/proto/yKDQl6NMZYM1pTwcQcwwRD/Feds---Promos?page-id=17%3A1839&node-id=18-2566&t=yLSphbWkVj0UKM6k-0&scaling=scale-down-width&content-scaling=fixed&starting-point-node-id=18%3A2566) for the feds promo bar, also mentioned in the ticket

@Deva309 - I see that between 1200px and 1270px the promo bar content is displayed on two rows
Screenshot 2024-07-19 at 11 25 42

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -571,6 +570,11 @@ header.global-navigation {
     overflow: auto;
   }
 
+  .global-navigation.has-promo .feds-navItem--megaMenu .feds-popup {
+    max-height: calc(100vh - 100% - var(--global-height-navPromo));
+    box-sizing: border-box;
```
narcis-radu, line 575: Most of the elements in the global navigation are `context-box` and this might generate some confusion when setting padding

libs/blocks/mobile-app-banner/mobile-app-banner.js
```diff
@@ -59,7 +59,7 @@ function branchInit(key) {
 /* eslint-enable */
 export default async function init(el) {
   const header = document.querySelector('.global-navigation');
-  if (!header) return;
+  if (!header || header.classList.contains('has-promo')) return;
```
narcis-radu, line 62:
```suggestion
if (!header?.classList.contains('has-promo')) return;
```
if header is not available, the execution will stop

libs/styles/styles.css
narcis-radu: I noticed it spawns on two lines between 1200 and 1270px

prativas22:
The promobar functionality  is verified after the resizing and scrolling issue was fixed.
URL:
https://mwpw-151517-promobar--milo--deva309.hlx.page/drafts/devashish/promo-bar/feds-promobar
image
image
image

https://mwpw-151517-promobar--milo--deva309.hlx.page/drafts/prativa/feds-promobar-one 
https://mwpw-151517-promobar--milo--deva309.hlx.page/drafts/prativa/aside-promo-bar
image
Scrolling 
image
image
image

without promobar:
image

libs/blocks/mobile-app-banner/mobile-app-banner.js
Deva309: If header is not available we want to stop the execution.
```if (!header?.classList.contains('has-promo')) return;```
this condition fail in the case if we have `has-promo` class available.
So basically we are checking 
if header is not available then return
if header is available the check for `has-promo` class if it is present then again return.

File: libs/blocks/global-navigation/global-navigation.css (hunk 5 above)
Comment (Deva309) [reply libs/blocks/global-navigation/global-navigation.css]: @narcis-radu Yes, but this case involves the megamenu, where scrolling is necessary if its height and width exceed the screen dimensions. To address this, we calculate the height for the megamenu and add a scroll. However, due to `content-box`, it only sets the scroll for the content area, but we need scrolling for both the content and the surrounding spacing. I've now moved it closer to `overflow:auto`, which is only required for the scroll functionality.

**Comment by Deva309** [comment]
> > To avoid CLS impact, we need to adhere to the character count limits specified in the design [specs](https://www.figma.com/proto/yKDQl6NMZYM1pTwcQcwwRD/Feds---Promos?page-id=17%3A1839&node-id=18-2566&t=yLSphbWkVj0UKM6k-0&scaling=scale-down-width&content-scaling=fixed&starting-point-node-id=18%3A2566) for the feds promo bar, also mentioned in the ticket
> 
> @Deva309 - I see that between 1200px and 1270px the promo bar content is displayed on two rows Screenshot 2024-07-19 at 11 25 42

@narcis-radu 
Yes, I missed this range. I will check with the PMs to see if the range of 1200px to 1286px, where the content for the feds promobar wraps to two lines, applies to any widely used devices. If it does, we will need to update the character count limit.

File: libs/blocks/mobile-app-banner/mobile-app-banner.js (hunk 6 above)
Comment (narcis-radu) [reply libs/blocks/mobile-app-banner/mobile-app-banner.js:62]: you are correct, my bad

File: libs/blocks/global-navigation/global-navigation.css (hunk 5 above)
Comment (narcis-radu) [reply libs/blocks/global-navigation/global-navigation.css]: We already have scrolling functionality on Production and the box-sizing is `content-box`, don't we ?

File: libs/blocks/global-navigation/global-navigation.css [hunk 7]
```diff
@@ -702,5 +706,6 @@ header.global-navigation {
 
   .feds-navItem--megaMenu .feds-popup {
     align-items: center;
+    flex-wrap: wrap;
```
overmyheadandbody, line 709: Why is this needed? Is it desirable to have columns span on multiple rows in certain scenarios? Doesn't this interfere with the `feds-crossCloudMenu-wrapper` element, since it's also inside the popup? This is how this tweak affects the current Photoshop page:
Screenshot 2024-07-23 at 15 39 27

Since this doesn't seem to be related to the promo, it might be good to remove this.

libs/styles/styles.css
overmyheadandbody: @narcis-radu is right, the Promo spans two rows between 1200 and 1287px. Either the copy needs to be adapted or some styles.

libs/blocks/global-navigation/global-navigation.css
```diff
@@ -569,6 +568,11 @@ header.global-navigation {
     padding: 40px 0 0;
     max-height: calc(100vh - 100%);
     overflow: auto;
+    box-sizing: border-box;
```
overmyheadandbody, line 571: I don't see any box-size model properties adapted to justify this change. Why is it needed?

overmyheadandbody (changes requested):
The `flex-wrap` declaration should be revisited, as it has the potential to affect a large number of production pages.

libs/styles/styles.css
nishantka: Yes. Based on discussions with _Joel Blytt_ the character count limit in design specs will be updated so that promobar displays on single line for this range also.

libs/blocks/global-navigation/global-navigation.css
nishantka: This was added to fix one of the issues where the left side of the large menu gets cut from left side (1200-1300px range). I'll remove this due to above issue and as its unrelated to promobar change. We will log a separate bug for this.

Screenshot 2024-07-24 at 12 47 44 PM

libs/blocks/global-navigation/global-navigation.css
nishantka: I think this was needed to compensate the 40px padding which was not included in height and the bottom-most link in the large menu was getting cut off from the scrolling as shown.

Screenshot 2024-07-24 at 1 29 19 PM

libs/blocks/global-navigation/global-navigation.css
nishantka: I think this was needed to compensate the 40px padding which was not included in height and the bottom-most link in the large menu was getting cut off from the scrolling.
image

libs/blocks/global-navigation/global-navigation.css
overmyheadandbody: 5 columns have never been supported in the dropdown. The initial Consonant guidelines (which I can't locate currently due to how many changes they had on their platform 😅 ) stated that a maximum of 4 columns should be added in a dropdown menu.

libs/styles/styles.css
```diff
@@ -706,16 +706,14 @@ header:not(.global-navigation) ~ main {
  */
 header.global-navigation {
   height: var(--global-height-nav);
-  visibility: hidden;
```
overmyheadandbody, line 709: Is this actually desired? There seems to be a conflict between this and https://github.com/adobecom/milo/pull/2603/files

nishantka: This is not required. We will revert this change.

overmyheadandbody (approved):
Thanks for making the changes!

prativas22:
Validation done for the promobar
https://mwpw-151517-promobar--milo--deva309.hlx.page/drafts/prativa/aside-promo-bar
https://mwpw-151517-promobar--milo--deva309.hlx.page/drafts/prativa/test-feds-promobar-with-app-banner
https://mwpw-151517-promobar--milo--deva309.hlx.page/drafts/prativa/feds-promobar-char-limit
https://mwpw-151517-promobar--milo--deva309.hlx.page/drafts/prativa/test-feds-promobar
iphone:
image
image
samsung:
image
Resizing screen
image

PR #2544, opened 2024-07-08
MWPW-153962: Introduce maslibs query parameter

yesil:
Similar to milolibs query parameter, when present, it will load commerce.js from the external M@S repository.

Resolves: [MWPW-153962](https://jira.corp.adobe.com/browse/MWPW-153962)

**Test URLs:**
- After: https://mwpw-153962--milo--yesil.hlx.page/?martech=off

Some test page URLs:
- https://main--cc--adobecom.hlx.live/products/catalog?milolibs=mwpw-153962&maslibs=main--mas--adobecom


libs/blocks/merch/merch.js
```diff
@@ -349,7 +377,13 @@ export async function initService(force = false) {
   const { env, commerce = {}, locale } = getConfig();
   commerce.priceLiteralsPromise = fetchLiterals(PRICE_LITERALS_URL);
   initService.promise = initService.promise ?? polyfills().then(async () => {
-    const commerceLib = await import('../../deps/commerce.js');
+    const { hostname, searchParams } = new URL(window.location.href);
+    const maslibs = searchParams.get('maslibs');
+    let commerceLibPath = '../../deps/commerce.js';
+    if (maslibs) {
+      commerceLibPath = `${getMasBase(hostname, maslibs)}/lib/commerce.js`;
+    }
```
overmyheadandbody, line 388:
```suggestion
    const commerceLibPath = `${!maslibs ? '../../deps' : `${getMasBase(hostname, maslibs)}/lib`}/commerce.js`;
```

head.html
```diff
@@ -1,4 +1,4 @@
 <meta name="viewport" content="width=device-width, initial-scale=1"/>
-<script src="/libs/scripts/scripts.js" type="module" crossorigin="use-credentials"></script>
+<script src="/libs/scripts/scripts.js" type="module"></script>
```
overmyheadandbody, line 2: I don't remember the details, but this file was previously heavily cached and making changes to it lead to various issues. Are you certain this should be removed? Why do Milo's scripts interfere with Merch logic?

libs/blocks/merch/merch.js
```diff
@@ -349,7 +377,13 @@ export async function initService(force = false) {
   const { env, commerce = {}, locale } = getConfig();
   commerce.priceLiteralsPromise = fetchLiterals(PRICE_LITERALS_URL);
   initService.promise = initService.promise ?? polyfills().then(async () => {
-    const commerceLib = await import('../../deps/commerce.js');
+    const { hostname, searchParams } = new URL(window.location.href);
```
overmyheadandbody, line 383: Couldn't the method itself get the `hostname` and `searchParams`? What's the value of sending these from here?

head.html
npeltier: is the use case to render a milo.adobe.com page with different commerce source, @yesil ? may be we could make that change a separate one as i feel like it's not going to be an easy one and we can still play around with most commerce pages on CC & DC

yesil: With `crossorigin="use-credentials"` all the subsequent js module loaded by `/libs/scripts/scripts.js` will be subject to CORS restrictions. Since `maslibs` param will allow to load commerce.js from a different domain instead of libs/deps folder, this will fail.  With `crossorigin="use-credentials"` we cannot use `Access-Control-Allow-Origin:*` on commerce.js neither, it must be an explicit domain, without wildcards.

I don't know changing this file will causes an immediate cache flush, but we don't need it immediately. We can start using it once CDN caches are naturally refreshed.

libs/blocks/merch/merch.js
```diff
@@ -122,6 +122,34 @@ const LOADING_ENTITLEMENTS = 'loading-entitlements';
...
+ */
+export function getMasBase(hostname, maslibs) {
+  let { baseUrl } = getMasBase;
+  if (baseUrl) return baseUrl;
+  baseUrl = 'https://www.adobe.com/mas';
+  if (!maslibs) return baseUrl;
+  if (maslibs === 'stage') {
+    baseUrl = 'https://www.stage.adobe.com/mas';
+  } else if (maslibs === 'local') {
+    baseUrl = 'http://localhost:8000';
```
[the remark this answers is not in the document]
yesil: 6456 is the Milo port to use with `milolibs=local`, and it is hard coded in Milo consumer apps (e.g: https://github.com/adobecom/cc/blob/stage/creativecloud/scripts/utils.js#L39)

I proposed `http://localhost:8000` for using MAS locally with `maslibs=local`, but we can use a different port if 8000 is a reserved port.

libs/blocks/merch/merch.js
yesil: I thought about it but I find it harder to read.  Most of the time, `maslibs` will be `false` and the whole expression will be skipped.

libs/blocks/merch/merch.js
yesil: I kept `getMasBase` as a pure function so that it is easier to test it.
Since we need to check whether `maslibs` param is present here and `getMasBase` only needs `hostname ` and `maslibs` parameters, I limited it to those params.

head.html
npeltier: we discussed this offline and agreed we can move this to a second one line PR

yesil: Moved this change in a separate PR: https://github.com/adobecom/milo/pull/2549

libs/blocks/merch/merch.js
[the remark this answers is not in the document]
npeltier: didn't find any obvious tool using 8000 port so i'm good with it :)

npeltier (approved):
lgtm

libs/blocks/merch/merch.js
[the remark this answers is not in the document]
narcis-radu: most of the quick local dev servers (e.g. python http.server) use 8000 as the default port and we might have some conflicts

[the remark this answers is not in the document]
yesil: for tacocat we used to use ports 9001-9008 for different modules.
I'll pick 9001 :)

libs/blocks/merch/merch.js
```diff
@@ -122,6 +122,34 @@ const LOADING_ENTITLEMENTS = 'loading-entitlements';
...
+  let { baseUrl } = getMasBase;
+  if (baseUrl) return baseUrl;
+  baseUrl = 'https://www.adobe.com/mas';
+  if (!maslibs) return baseUrl;
+  if (maslibs === 'stage') {
+    baseUrl = 'https://www.stage.adobe.com/mas';
+  } else if (maslibs === 'local') {
+    baseUrl = 'http://localhost:9001';
+  } else {
+    const extension = /.live$/.test(hostname) ? 'live' : 'page';
```
narcis-radu, line 142: we need to revisit this

libs/blocks/merch/merch.js
```diff
@@ -122,6 +122,34 @@ const LOADING_ENTITLEMENTS = 'loading-entitlements';
...
+    if (/--/.test(maslibs)) {
+      baseUrl = `https://${maslibs}.hlx.${extension}`;
+    } else {
+      baseUrl = `https://${maslibs}--mas--adobecom.hlx.${extension}`;
```
3ch023, line 146: how about we make 'maslibs' work the same way as 'milolibs' do?

`?maslibs=mwpw-153962--mas--yesil`
branch-fork-owner
Then you can save precious bytes and remove stage and prod cases, since the link could be given as
`?maslibs=main--mas--adobecom` or `?maslibs=stage--mas--adobecom` (when/if we have stage) instead of `?maslibs=stage`

I know the latter is shorter but it involves an additional learning curve for devs and IMO is not worth the space it takes.

libs/blocks/merch/merch.js
yesil: why?

libs/blocks/merch/merch.js
```diff
@@ -122,6 +122,34 @@ const LOADING_ENTITLEMENTS = 'loading-entitlements';
...
 let upgradeOffer = null;
 
+/**
+ * Given a url, calculates the hostname of MAS platform.
+ * Supports, www prod, stage, local and feature branches.
+ * @param {string} hostname
+ * @param {string} maslibs
+ * @returns base url for mas platform
+ */
+export function getMasBase(hostname, maslibs) {
```
3ch023, line 132: we will have to move it out of merch since it will be used in other blocks as well (merch-card, etc)
but for now i am fine to keep it here, just a heads up

libs/blocks/merch/merch.js
```diff
@@ -122,6 +122,34 @@ const LOADING_ENTITLEMENTS = 'loading-entitlements';
...
+  let { baseUrl } = getMasBase;
```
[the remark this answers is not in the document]
yesil: We used this pattern in the other functions in merch block. Thus, we can cache the calculated value, and erase it when needed (mostly in tests). Using a module scope variable makes testing harder. You need to provide a utility method to reset the module scope variables.

libs/blocks/merch/merch.js
```diff
@@ -376,8 +413,9 @@ export async function getCommerceContext(el, params) {
 }
 
 /**
- * Checkout parameter can be set Merch link, code config (scripts.js) or be a default from tacocat.
- * To get the default, 'undefinded' should be passed, empty string will trigger an error!
+ * Checkout parameter can be set on the merch link,
+ * code config (scripts.js) or be a default from tacocat.
+ * To get the default, 'undefined' should be passed, empty string will trigger an error!
```
3ch023, line 418: is this a lint change? (fine with me, just clarifying)

libs/blocks/merch/merch.js
yesil: During stage testing, I think  just using `&maslibs=stage` instead of `&maslibs=stage--mas--adobecom` will save a lot of time to humans. 
But saving bytes is also interesting. @npeltier @narcis-radu @3ch023 @overmyheadandbody 
👍 or 👎  for saving bytes?

libs/blocks/merch/merch.js
yesil: since merch.js will be already loaded regardless, it makes sense to keep it here.
And it is exported, and the value is cached, merch-card.js doesn't need to pass the hostname and maslibs.

libs/blocks/merch/merch.js
yesil: yes

yesil: also typo: undefinded :D

libs/blocks/merch/merch.js
narcis-radu: We will have to discuss offline

libs/blocks/merch/merch.js
npeltier: i think i prefer mariia/milolibs's approach, yes :)

libs/blocks/merch/merch.js
```diff
@@ -349,7 +380,15 @@ export async function initService(force = false) {
   const { env, commerce = {}, locale } = getConfig();
   commerce.priceLiteralsPromise = fetchLiterals(PRICE_LITERALS_URL);
   initService.promise = initService.promise ?? polyfills().then(async () => {
-    const commerceLib = await import('../../deps/commerce.js');
+    const { hostname, searchParams } = new URL(window.location.href);
+    let commerceLibPath = '../../deps/commerce.js';
+    if (/hlx\.(page|live)$|localhost$|www\.stage\.adobe\.com$/.test(hostname)) {
```
yesil, line 385: image

fyi: @narcis-radu

libs/blocks/merch/merch.js
yesil: fine

yesil: updated

libs/blocks/merch/merch.js
3ch023: ahhh true true, nevermind my comment

libs/blocks/merch/merch.js
3ch023: hehe

Axelcureno (approved):

3ch023 (approved):

libs/blocks/merch/merch.js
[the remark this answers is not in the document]
narcis-radu: But still, on the first pass the baseUrl value would be empty and I don't see a reason why you would need to call the method multiple times.

libs/blocks/merch/merch.js
narcis-radu: Should we consider other STAGE environments like BACOM or Firefly ?

yesil: Afaik BACOM doesn't have prices, and Firefly is not merch enabled yet.
This can be done later as part of merch enablement on Firefly.

libs/blocks/merch/merch.js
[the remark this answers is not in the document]
yesil: @narcis-radu I removed the caching part as it is not going to be called to many times but several times definitely.
later after commerce.js, we will start adding maslibs feature to merch-card, merch-card-collection and other merch blocks as well.
merch being auto block it will be the first one to call this method and have the base url initialised, then the other blocks can use the value.
@3ch023 I'm tempted to add merch-card in this PR, thus we can give a better shape to the code, wdyt? cc: @npeltier

libs/blocks/merch/merch.js
npeltier: even with that, for now it's mostly for testing, and most of our pages are on cc/dc so i think we are good

libs/blocks/merch/merch.js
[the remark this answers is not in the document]
narcis-radu: I am still not happy with the `baseUrl` approach since it doesn't follow the pattern used in other Milo blocks, but I don't think this is a blocker or reason to request changes. Thanks for addressing the important changes 👍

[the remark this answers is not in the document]
yesil: Actually,   something like `getMasLibs` `setMasLibs` would be better. I'll give it a try tomorrow my morning.

afmicka:
@yesil changes on this branch are breaking filter functionality on catalog page. If you change to alphabetical and back to Popularity, it does not take effect.

Also, on bacom page, gnav is not loaded with these changes. It works ok on other fork [branches](https://main--bacom--adobecom.hlx.page/products/adobe-experience-cloud-products?milolibs=mwpw-151513-search-results--milo--mirafedas)  
https://main--bacom--adobecom.hlx.page/products/adobe-experience-cloud-products?milolibs=mwpw-153962--milo--yesil&maslibs=main--mas--adobecom

Could you please take a look into these?

libs/blocks/merch/merch.js
yesil: we will address this in a different PR.

PR #2548, opened 2024-07-09
MWPW-153600 [PEP] loader bar on PEP prompt is seen loaded Left to right in RTL locale

sharmrj:
* Changed the transform origin of `.appPrompt-progress` to start on the right instead of at the origin (top left)

Resolves: [MWPW-153600](https://jira.corp.adobe.com/browse/MWPW-153600)

**Test URLs:**
- After: https://pep-rtl--milo--adobecom.hlx.page/?martech=off
- qa url: https://main--cc--adobecom.hlx.page/il_he/creativecloud/animation/testdoc/pepphsp1?milolibs=pep-rtl&skipPepEntitlements=true

### DIFF
--- a/libs/features/webapp-prompt/webapp-prompt.css
+++ b/libs/features/webapp-prompt/webapp-prompt.css
@@ -182,6 +182,10 @@
     transform: scaleX(0) translateZ(0);
   }
 
+  [dir = "rtl"] .appPrompt-progress {
+    transform-origin: top right;
+  }
+
   @keyframes appPrompt-animate {
     100% {
       transform: scaleX(1);


libs/features/webapp-prompt/webapp-prompt.css
```diff
@@ -182,6 +182,10 @@
     transform: scaleX(0) translateZ(0);
   }
 
+  [dir = "rtl"] .appPrompt-progress {
+    transform-origin: top right;
```
mokimo, line 186: Anything we need to keep in mind in regards to animations with this?

mokimo (approved):

libs/features/webapp-prompt/webapp-prompt.css
sharmrj: I don't think so. By increasing `scaleX` from `0` to `1` the loader expands to fill the space. By starting the animation on the right instead of the left, it has nowhere to expand but to the left. 
AFAICT this doesn't affect any other animations and it applies only in rtl domains.
Lmk if I've misunderstood your question.

narcis-radu:
@sharmrj - Your test URL has `milolibs=local` which does not work. I think the correct URL should be https://main--cc--adobecom.hlx.page/il_he/creativecloud/animation/testdoc/pepphsp1?milolibs=pep-rtl&skipPepEntitlements=true

robert-bogos (approved):
Code-wise, it looks good 👍. Please check the message from Narcis.

spadmasa:
Validation done on the below url in cc :
https://main--cc--adobecom.hlx.page/il_he/creativecloud/animation/testdoc/pepphsp1?milolibs=pep-rtl&skipPepEntitlements=true
image
image

https://main--cc--adobecom.hlx.page/kw_ar/creativecloud/animation/testdoc/pepphsp1?milolibs=pep-rtl&skipPepEntitlements=true
image

narcis-radu (approved):

PR #2551, opened 2024-07-09
MWPW-153573 - [LocUI] Transcreate UI

Sartxi:
Adds support for the new Transcreate action.

Resolves: [MWPW-153573](https://jira.corp.adobe.com/browse/MWPW-153573)

### DIFF
--- a/libs/blocks/locui/loc/index.js
+++ b/libs/blocks/locui/loc/index.js
@@ -17,7 +17,7 @@ import login from '../../../tools/sharepoint/login.js';
 import { getServiceUpdates } from '../utils/miloc.js';
 import { connectSK } from '../../../utils/sidekick.js';
 
-const LANG_ACTIONS = ['Translate', 'English Copy', 'Rollout'];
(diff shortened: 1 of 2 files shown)


janaki-r-bhagwath (approved):

libs/blocks/locui/loc/index.js
```diff
@@ -17,7 +17,7 @@ import login from '../../../tools/sharepoint/login.js';
 import { getServiceUpdates } from '../utils/miloc.js';
 import { connectSK } from '../../../utils/sidekick.js';
 
-const LANG_ACTIONS = ['Translate', 'English Copy', 'Rollout'];
+const LANG_ACTIONS = ['Translate', 'English Copy', 'Rollout', 'Transcreate'];
```
auniverseaway, line 20: This is a bummer these are tightly coupled and we cannot get this data from the excel file easily.

auniverseaway (approved):

libs/blocks/locui/loc/index.js
Sartxi: agreed

PR #2555, opened 2024-07-10
[MWPW-153224] Promote from the excel file

robert-bogos:
## Description
This PR enables promote from an Excel file, instead of the JSON file.

## Related Issue
Resolves: [MWPW-153224](https://jira.corp.adobe.com/browse/MWPW-153224)

## Testing instructions
This is part of the Garybox promote feature and all testing regarding it will be done using [this ticket.](https://jira.corp.adobe.com/browse/MWPW-153227)

## Test URLs
Before: https://grayboxui--milo--adobecom.hlx.page/drafts/rbogos/page-default?martech=off
After: https://grayboxui--milo--robert-bogos.hlx.page/drafts/rbogos/page-default?martech=off

### DIFF
--- a/libs/blocks/graybox-promote/components/graybox-promote-lit.js
+++ b/libs/blocks/graybox-promote/components/graybox-promote-lit.js
@@ -220,7 +220,7 @@ class GrayboxPromote extends LitElement {
           const { root, gbRoot, driveId } = spData;
           const mainRepo = repo.replace('-graybox', '');
           const params = {
-            adminPageUri: `https://milo.adobe.com/tools/graybox?ref=${ref}&repo=${mainRepo}&owner=${owner}&host=business.adobe.com&project=${mainRepo.toUpperCase()}&referrer=MOCK_REF`,
+            adminPageUri: `https://milo.adobe.com/tools/graybox-promote?ref=${ref}&repo=${mainRepo}&owner=${owner}&host=business.adobe.com&project=${mainRepo.toUpperCase()}&referrer=MOCK_REF`,
             projectExcelPath: getProjectExcelPath(url),
             rootFolder: root,
             gbRootFolder: gbRoot,
--- a/libs/blocks/graybox-promote/components/graybox-promote-lit.js
+++ b/libs/blocks/graybox-promote/components/graybox-promote-lit.js
@@ -114,17 +114,24 @@ const getSharepointData = async (url) => {
 };
 
 const getFilePath = async ({ ref, repo, owner, excelRef } = {}) => {
-  const status = await fetch(`${ADMIN}/status/${owner}/${repo}/${ref}?editUrl=${excelRef}`);
-  const statusJson = await status.json();
-  return (new URL(statusJson?.preview?.url)).pathname;
+  try {
+    const status = await fetch(`${ADMIN}/status/${owner}/${repo}/${ref}?editUrl=${excelRef}`);
+    if (!status.ok) throw new Error('Failed to fetch file path');
+    return (new URL((await status.json())?.preview?.url)).pathname;
+  } catch (e) {
+    throw new Error(`Error getting file path: ${e.message}`);
+  }
 };
 
-const getPreviewUrl = ({ owner, repo, ref, filePath } = {}) => fetch(
-  `${ADMIN}/preview/${owner}/${repo}/${ref}${filePath}`,
-  { method: 'POST' },
-)
-  .then((res) => res.json())
-  .then((res) => new URL(res.preview.url));
+const getPreviewUrl = async ({ owner, repo, ref, filePath } = {}) => {
+  try {
+    const res = await fetch(`${ADMIN}/preview/${owner}/${repo}/${ref}${filePath}`, { method: 'POST' });
+    if (!res.ok) throw new Error('Failed to fetch preview URL');
+    return new URL((await res.json())?.preview?.url);
+  } catch (e) {
+    throw new Error(`Error getting preview URL: ${e.message}`);
+  }
+};
 
 class GrayboxPromote extends LitElement {
   spToken = accessToken.value || accessTokenExtra.value;
@@ -137,7 +144,7 @@ class GrayboxPromote extends LitElement {
   constructor() {
     super();
 
-    this.loginToSharePoint = async () => {
+    this.loginToSharepoint = async () => {
       // TODO - delete below
       this.spToken = '1234';
       return null;
@@ -204,7 +211,7 @@ class GrayboxPromote extends LitElement {
       task: async () => {
         const { ref, repo, owner } = getAemInfo();
         return new Promise((resolve, reject) => {
-          this.loginToSharePoint(ref, repo, owner)
+          this.loginToSharepoint(ref, repo, owner)
             .then(() => {
               this.setupTask.run();
               resolve();
--- a/libs/blocks/graybox-promote/components/graybox-promote-lit.js
+++ b/libs/blocks/graybox-promote/components/graybox-promote-lit.js
@@ -1,7 +1,7 @@
 import { getSheet } from '../../../../tools/utils/utils.js';
 import { LitElement, html } from '../../../deps/lit-all.min.js';
 import { Task } from '../../../deps/lit-task.min.js';
-import login from '../../../tools/sharepoint/login.js';
+// import login from '../../../tools/sharepoint/login.js';
 import {
   accessToken,
   accessTokenExtra,
@@ -12,7 +12,7 @@ import { getConfig } from '../../../utils/utils.js';
 const { miloLibs, codeRoot } = getConfig();
 const base = miloLibs || codeRoot;
 const styleSheet = await getSheet(
-  `${base}/blocks/graybox-promote/graybox-promote.css`
+  `${base}/blocks/graybox-promote/graybox-promote.css`,
 );
 
 const KEYS = {
@@ -32,7 +32,7 @@ const KEYS = {
   },
 };
 
-const TELEMETRY = { application: { appName: 'Adobe Graybox Promote' } };
+// const TELEMETRY = { application: { appName: 'Adobe Graybox Promote' } };
 
 const getJson = async (url, errMsg = `Failed to fetch ${url}`) => {
   let res;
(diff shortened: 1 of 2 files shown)


mokimo:
PSI Check fails, if you don't have a testable link, dummy links works to get the aem-psi-check green

robert-bogos:
No testable links for now. I added dummy urls.

libs/blocks/graybox-promote/components/graybox-promote-lit.js
```diff
@@ -114,7 +113,19 @@ const getSharepointData = async (url) => {
   };
 };
 
-const getProjectExcelPath = (url) => url.pathname.replace('.json', '.xlsx');
+const getFilePath = async (ref, repo, owner, excelRef) => {
```
mokimo, line 116: I'm a fan of passing in an object if there's more than one or two function parameters.
```suggestion
const getFilePath = async ({ref, repo, owner, excelRef} = {}) => {
```

libs/blocks/graybox-promote/components/graybox-promote-lit.js
```diff
@@ -114,7 +113,19 @@ const getSharepointData = async (url) => {
...
+  return (new URL(statusJson?.preview?.url)).pathname;
+};
+
+const preview = async (owner, repo, ref, path) => {
+  const previewResp = await fetch(
+    `${ADMIN}/preview/${owner}/${repo}/${ref}${path}`,
+    { method: 'POST' },
+  );
+  return previewResp.json();
+};
```
mokimo, line 128: I'd suggest using the implied return of the arrow function and simply return the promise which is awaited anyway.
```suggestion
const preview = ({owner, repo, ref, path}) => fetch(
    `${ADMIN}/preview/${owner}/${repo}/${ref}${path}`,
    { method: 'POST' },
  ).then(res => res.json())
```

Or one step further
```suggestion
const previewUrl = ({owner, repo, ref, path}) => fetch(
    `${ADMIN}/preview/${owner}/${repo}/${ref}${path}`,
    { method: 'POST' },
  )
  .then(res => res.json())
  .then(preview => new URL(preview.url))
```

mokimo (approved):

libs/blocks/graybox-promote/components/graybox-promote-lit.js
robert-bogos: Agree 👍 It's safer to pass an object. Made the change

libs/blocks/graybox-promote/components/graybox-promote-lit.js
robert-bogos: The desired output is the preview url, so I'll take your suggested shorter version 👍

libs/blocks/graybox-promote/components/graybox-promote-lit.js
```diff
@@ -114,7 +113,18 @@ const getSharepointData = async (url) => {
   };
 };
 
-const getProjectExcelPath = (url) => url.pathname.replace('.json', '.xlsx');
+const getFilePath = async ({ ref, repo, owner, excelRef } = {}) => {
+  const status = await fetch(`${ADMIN}/status/${owner}/${repo}/${ref}?editUrl=${excelRef}`);
```
narcis-radu, line 117: thinking we should catch potential errors

libs/blocks/graybox-promote/components/graybox-promote-lit.js
```diff
@@ -114,7 +113,18 @@ const getSharepointData = async (url) => {
...
+  const statusJson = await status.json();
+  return (new URL(statusJson?.preview?.url)).pathname;
+};
+
+const getPreviewUrl = ({ owner, repo, ref, filePath } = {}) => fetch(
+  `${ADMIN}/preview/${owner}/${repo}/${ref}${filePath}`,
+  { method: 'POST' },
+)
+  .then((res) => res.json())
+  .then((res) => new URL(res.preview.url));
```
narcis-radu, line 127: we should catch potential errors

libs/blocks/graybox-promote/components/graybox-promote-lit.js
```diff
@@ -127,7 +137,7 @@ class GrayboxPromote extends LitElement {
   constructor() {
     super();
 
-    this.spLogin = async () => {
+    this.loginToSharePoint = async () => {
```
narcis-radu, line 140: Sharepoint

robert-bogos: Made the change 👍

libs/blocks/graybox-promote/components/graybox-promote-lit.js
robert-bogos: True, we might have unexpected errors that we'll want to catch. Made the change 👍

libs/blocks/graybox-promote/components/graybox-promote-lit.js
robert-bogos: True, we might have unexpected errors that we'll want to catch. Made the change 👍

overmyheadandbody (approved):

libs/blocks/graybox-promote/components/graybox-promote-lit.js
```diff
@@ -114,7 +113,25 @@ const getSharepointData = async (url) => {
...
-const getProjectExcelPath = (url) => url.pathname.replace('.json', '.xlsx');
+const getFilePath = async ({ ref, repo, owner, excelRef } = {}) => {
+  try {
+    const status = await fetch(`${ADMIN}/status/${owner}/${repo}/${ref}?editUrl=${excelRef}`);
+    if (!status.ok) throw new Error('Failed to fetch file path');
+    return (new URL((await status.json())?.preview?.url)).pathname;
+  } catch (e) {
+    throw new Error(`Error getting file path: ${e.message}`);
+  }
+};
```
mokimo, line 124: You already throw an error for the status, so it can be moved out of the try/catch
```suggestion
const getFilePath = async ({ ref, repo, owner, excelRef } = {}) => {
  const status = await fetch(`${ADMIN}/status/${owner}/${repo}/${ref}?editUrl=${excelRef}`);
  if (!status.ok) throw new Error('Failed to fetch file path');
  try {
    return (new URL((await status.json())?.preview?.url)).pathname;
  } catch (e) {
    throw new Error(`Error getting file path: ${e.message}`);
  }
};
```

mokimo, line 124: I'm a huge fan of top level catching for async/await functions. The only advantage that you get by adding "manual" error handling on the function level here, is improved error messages - especially since you re-throw the error.

That IMO isn't worth obscuring the function. I'd go with a clean version of the function and handle the error a level above:

```suggestion
const getFilePath = async ({ ref, repo, owner, excelRef } = {}) => fetch(`${ADMIN}/status/${owner}/${repo}/${ref}?editUrl=${excelRef}`)
.then(res => res.json())
.then(json => new URL(json.preview.url).pathname)
```

libs/blocks/graybox-promote/components/graybox-promote-lit.js
```diff
@@ -143,18 +160,19 @@ class GrayboxPromote extends LitElement {
...
-    this.getValuesTask = new Task(this, {
+    this.setupTask = new Task(this, {
       task: async () => {
-        const { ref, repo, owner, referrer } = getAemInfo();
-        const url = new URL(referrer);
-        const { experienceName, grayboxIoEnv } = await getProjectInfo(referrer);
+        const { ref, repo, owner, excelRef } = getAemInfo();
+        const filePath = await getFilePath({ ref, repo, owner, excelRef });
+        const previewUrl = await getPreviewUrl({ owner, repo, ref, filePath });
+        const { experienceName, grayboxIoEnv } = await getProjectInfo(previewUrl);
```
mokimo, line 168: [Task is already catching the errors ](https://github.com/adobecom/milo/pull/2555/files#diff-7214629f44f7a1dbc0bdfc18fb6309965da48000fe961e4aaeed56046e9aa06fR219) from what I can tell, so unless there is value in having specific error messages - just having the error bubble up would be enough IMO 🤔

mokimo (approved):
Might be a bit conflicting with the feedback from Rares/Narcis, some more food for thought 😁

libs/blocks/graybox-promote/components/graybox-promote-lit.js
```diff
@@ -213,15 +231,15 @@ class GrayboxPromote extends LitElement {
...
-        url,
         spData,
+        filePath,
       }) => {
         try {
           const { root, gbRoot, driveId } = spData;
           const mainRepo = repo.replace('-graybox', '');
           const params = {
             adminPageUri: `https://milo.adobe.com/tools/graybox-promote?ref=${ref}&repo=${mainRepo}&owner=${owner}&host=business.adobe.com&project=${mainRepo.toUpperCase()}&referrer=MOCK_REF`,
-            projectExcelPath: getProjectExcelPath(url),
```
sukamat, line 224: @arshadparwaiz `host` and `referrer` query param should get removed once the promote API change is made. That is not done yet, correct?

sukamat (approved):

libs/blocks/graybox-promote/components/graybox-promote-lit.js
robert-bogos: I took it one step further and removed the try/catch altogether. [The Task will catch any error](https://github.com/adobecom/milo/pull/2555/files#diff-7214629f44f7a1dbc0bdfc18fb6309965da48000fe961e4aaeed56046e9aa06fR290) and show the message in the ui.

robert-bogos: Got your point 👍 I simplified the function by removing the try catch [as the task will catch the errors](https://github.com/adobecom/milo/pull/2555/files#diff-7214629f44f7a1dbc0bdfc18fb6309965da48000fe961e4aaeed56046e9aa06fR290), but I would keep the `status.ok` check at least until this feature gets more stable. In this way we can debug easier in this incipient state

libs/blocks/graybox-promote/components/graybox-promote-lit.js
robert-bogos: Correct, the task is catching the errors [in here](https://github.com/adobecom/milo/pull/2555/files#diff-7214629f44f7a1dbc0bdfc18fb6309965da48000fe961e4aaeed56046e9aa06fR290). I removed some of the unnecessary try/catch, but left some manual throwing of errors for debugging purposes.

PR #2557, opened 2024-07-10
MWPW-151863 - 🆕 Notification Block

elan-tbx:
In an effort to overhaul the Aside block, certain variants of the Aside have been broken off into their own block: The Notification Block. These map to:

```
aside (notification) -> notification
aside (promo-bar) -> notification (ribbon)
aside (promo-bar, popup) -> notification (pill)
```

Furthermore, these new variants improve on the old ones in certain ways, such as the ability to remove the close button on non-sticky notifications, and fine-grained control over font sizes (l-heading, s-body, etc). Click the "After" links below to see full examples of each variant of the block.

Screenshot 2024-07-10 at 5 07 05 PM

Screenshot 2024-07-10 at 5 07 22 PM

Screenshot 2024-07-10 at 5 06 48 PM

Resolves: [MWPW-151863](https://jira.corp.adobe.com/browse/MWPW-151863), [MWPW-151832](https://jira.corp.adobe.com/browse/MWPW-151832), and [MWPW-151868](https://jira.corp.adobe.com/browse/MWPW-151868)

**Test URLs:**

Banner Notifications:
- Before: https://main--milo--adobecom.hlx.page/drafts/ebartholomew/notifications?martech=off
- After: https://ebartholomew-notification-block--milo--adobecom.hlx.page/drafts/ebartholomew/notifications?martech=off

Ribbon Notifications:
Before: https://main--milo--adobecom.hlx.page/drafts/ebartholomew/ribbon-notification?martech=off
After: https://ebartholomew-notification-block--milo--adobecom.hlx.page/drafts/ebartholomew/ribbon-notification?martech=off

Pill Notifications:
- Before: https://main--milo--adobecom.hlx.page/drafts/ebartholomew/pill-notification?martech=off
- After: https://ebartholomew-notification-block--milo--adobecom.hlx.page/drafts/ebartholomew/pill-notification?martech=off

### DIFF
--- a/libs/blocks/notification/notification.js
+++ b/libs/blocks/notification/notification.js
@@ -102,13 +102,11 @@ function decorateLayout(el) {
   foreground?.classList.add('foreground', 'container');
   const text = foreground?.querySelector('h1, h2, h3, h4, h5, h6, p')?.closest('div');
   text?.classList.add('text');
-  const picture = text?.querySelector('p picture');
-  const iconArea = picture?.closest('p');
+  const iconArea = text?.querySelector('p picture')?.closest('p');
   iconArea?.classList.add('icon-area');
   const fgMedia = foreground?.querySelector(':scope > div:not(.text) :is(img, video, a[href*=".mp4"])')?.closest('div');
   const bgMedia = el.querySelector(':scope > div:not(.foreground) :is(img, video, a[href*=".mp4"])')?.closest('div');
   const media = fgMedia ?? bgMedia;
-  el.classList.toggle('no-media', !media);
   media?.classList.toggle('image', media && !media.classList.contains('text'));
   foreground?.classList.toggle('no-image', !media && !iconArea);
   if (el.matches(`:is(.${pill}, .${ribbon}):not(.no-closure)`)) decorateClose(el);
--- a/libs/blocks/notification/notification.js
+++ b/libs/blocks/notification/notification.js
@@ -1,0 +1,138 @@
+/*
+ * Copyright 2024 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+/*
+* Notification - v1.0
+*/
+
+import { decorateBlockText, decorateBlockBg, decorateTextOverrides } from '../../utils/decorate.js';
+import { createTag } from '../../utils/utils.js';
+
+const variants = ['banner', 'ribbon', 'pill'];
+const sizes = ['small', 'medium', 'large'];
+const [banner, ribbon, pill] = variants;
+const [small, medium, large] = sizes;
+const defaultSize = medium;
+const defaultVariant = banner;
+const blockConfig = {
+  [banner]: {
+    [small]: ['s', 's', 's', 'm'],
+    [medium]: ['m', 'm', 'm', 'm'],
+    [large]: ['l', 'l', 'l', 'l'],
+  },
+  [ribbon]: {
+    [small]: ['s', 's', 's', 'm'],
+    [medium]: ['m', 'm', 'm', 'l'],
+    [large]: ['l', 'l', 'l', 'l'],
+  },
+  [pill]: {
+    [small]: ['s', 's', 's', 'm'],
+    [medium]: ['m', 'm', 'm', 'l'],
+    [large]: ['l', 'm', 'm', 'l'],
+  },
+};
+
+const closeSvg = `<svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 20 20" fill="none">
+  <g clip-path="url(#clip0_699_20329)">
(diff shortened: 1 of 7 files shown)


narcis-radu:
Hey @elan-tbx - I noticed a thing and wanted to confirm with you: the default notification has 800px width on desktop (> 1200px), but it goes up to 999px on viewports with < 1200px. Shouldn't we use the space on large desktops to stretch the notification to more than 800px ?

narcis-radu:
I think we need to handle notifications with text that spawns across multiple rows. I checked a pill notification and it looks like this:
Screenshot 2024-07-11 at 16 45 22

elan-tbx:
> Hey @elan-tbx - I noticed a thing and wanted to confirm with you: the default notification has 800px width on desktop (> 1200px), but it goes up to 999px on viewports with < 1200px. Shouldn't we use the space on large desktops to stretch the notification to more than 800px?

Hey @narcis-radu, having a hard limit of 800px was actually by design, but I can see how it's strange that it jumps up below the desktop breakpoint... I'll make that the max across the board 👍

elan-tbx:
> I think we need to handle notifications with text that spawns across multiple rows. I checked a pill notification and it looks like this:

@narcis-radu good catch, I'll push up a fix

test/blocks/notification/notification.test.js
```diff
@@ -0,0 +1,72 @@
...
+const locales = { '': { ietf: 'en-US', tk: 'hah7vzn.css' } };
+const conf = { locales };
+setConfig(conf);
+
+document.body.innerHTML = await readFile({ path: './mocks/body.html' });
+const { default: init } = await import('../../../libs/blocks/notification/notification.js');
+
+describe('notification', () => {
+  const notifs = document.querySelectorAll('.notification');
+  notifs.forEach((notif) => init(notif));
```
mokimo, line 14: This looks prone to building up test state. It would be better to adding the notifications to the body in a `beforeEach` block and clearing the DOM In a `afterEach` block.

Tests should be isolated, debugging and changing tests that already depend on shared test state is very tedious

libs/blocks/notification/notification.js
```diff
@@ -0,0 +1,138 @@
...
+
+const variants = ['banner', 'ribbon', 'pill'];
+const sizes = ['small', 'medium', 'large'];
+const [banner, ribbon, pill] = variants;
+const [small, medium, large] = sizes;
+const defaultSize = medium;
+const defaultVariant = banner;
+const blockConfig = {
+  [banner]: {
+    [small]: ['s', 's', 's', 'm'],
```
mokimo, line 28: Could you share some context on why the small banner could be `s` or `m`? I don't understand the need for the mapping

libs/blocks/notification/notification.js
```diff
@@ -0,0 +1,138 @@
...
+  </g>
+  <defs>
+    <clipPath id="clip0_699_20329">
+      <rect width="20" height="20" fill="white"/>
+    </clipPath>
+  </defs>
+</svg>`;
+
+function getOpts(el) {
+  const optRows = [...el.querySelectorAll(':scope > div:nth-of-type(n+3)')];
```
mokimo, line 56: Shouldn't these options row be section metadata or simply a class? I don't think we want to give full freedom on the choice of classes to authors. That's what we had in dexter and wanted to move away from in favor of standardization and consonant approved designs that usually are a couple and not an infinite way of possibilities.

libs/blocks/notification/notification.js
```diff
@@ -0,0 +1,138 @@
...
+  if (el.matches(`.${pill}.flexible`)) decorateFlexible(el);
+  return foreground;
+}
+
+export default function init(el) {
+  el.classList.add('con-block');
+  const { fontSizes, options } = getBlockData(el);
+  const blockText = decorateLayout(el);
+  decorateBlockText(blockText, fontSizes);
+  if (options.borderBottom) decorateBorder(el, options);
```
mokimo, line 134: Not convinced authors should have the freedom of setting their own styles. From a code perspective, the helper function is overkill imo
```suggestion
  if (options.borderBottom) {
    el.append(createTag('div', { style: `background: ${options.borderBottom};`, class: 'border' }))
  }
```

libs/blocks/notification/notification.js
```diff
@@ -0,0 +1,138 @@
...
+  decorateTextOverrides(el);
+  decorateStaticLinks(el);
```
mokimo, line 136: I'd also get rid of this helper function, given it's a simply one liner
```suggestion
  el.querySelectorAll('a:not([class])').forEach((staticLink) => staticLink.classList.add('static'));
```

libs/blocks/notification/notification.js
```diff
@@ -0,0 +1,138 @@
...
+  const fgMedia = foreground.querySelector(':scope > div:not(.text) :is(img, video, a[href*=".mp4"])')?.closest('div');
+  const bgMedia = el.querySelector(':scope > div:not(.foreground) :is(img, video, a[href*=".mp4"])')?.closest('div');
+  const media = fgMedia ?? bgMedia;
+  el.classList.toggle('no-media', !media);
+  media?.classList.toggle('image', (media && !media.classList.contains('text')));
+  foreground?.classList.toggle('no-image', (!media && !iconArea));
+  if (el.matches(`:is(.${pill}, .${ribbon}):not(.no-closure)`)) decorateClose(el);
+  if (el.matches(`.${pill}.flexible`)) decorateFlexible(el);
+  return foreground;
+}
```
mokimo, line 127: This could be a bit more declarative using destructuring to assign names to variables
```suggestion
function decorateLayout(el) {
  const [background, ...rest] = el.querySelectorAll(':scope > div');
  const foreground = rest.pop();

  if (background) decorateBlockBg(el, background);

  foreground?.classList.add('foreground', 'container');

  const text = foreground?.querySelector('h1, h2, h3, h4, h5, h6, p')?.closest('div');
  text?.classList.add('text');

  const picture = text?.querySelector('p picture');
  const iconArea = picture?.closest('p');
  iconArea?.classList.add('icon-area');

  const fgMedia = foreground?.querySelector(':scope > div:not(.text) :is(img, video, a[href*=".mp4"])')?.closest('div');
  const bgMedia = el.querySelector(':scope > div:not(.foreground) :is(img, video, a[href*=".mp4"])')?.closest('div');
  const media = fgMedia ?? bgMedia;

  el.classList.toggle('no-media', !media);
  media?.classList.toggle('image', media && !media.classList.contains('text'));
  foreground?.classList.toggle('no-image', !media && !iconArea);

  if (el.matches(`:is(.${pill}, .${ribbon}):not(.no-closure)`)) decorateClose(el);
  if (el.matches(`.${pill}.flexible`)) decorateFlexible(el);

  return foreground;
}
```

mokimo (commented):
A few comments & suggestions for the first round of reviews

libs/blocks/notification/notification.js
elan-tbx: The mapping like this is mostly due to the difference in button sizes (the final value in the array) across each size variant. If it weren't for that, we could just assume that a `small` notification had `s` for all four values, but there's not such thing as a small button in styles.css. But either way, `decorateBlockText` expects an array for a config, and the default (`['m', 's', 'm']`) doesn't work for this block.

libs/blocks/notification/notification.js
elan-tbx: The ability to author the border background value was a specific request I got from Consonant, so we'll have to include a way to do it somehow that doesn't involve simple classes... section metadata could be a home for it, but it does feel strange coupling this block to section metadata just for the sake of entering a bg color/gradient value

narcis-radu: I think I've seen this practice being used in [Editorial Cards](https://adobe.sharepoint.com/:w:/r/sites/adobecom/_layouts/15/Doc.aspx?sourcedoc=%7BD2753A55-82B2-4174-A941-A98D19C07BF2%7D&file=Editorial%20Cards.docx&action=default&mobileredirect=true) as well as some other cards related components. I do see @mokimo's point of using section metadata, but at the same time I understand the flexibility.
My only concern is around authoring consistency, I want to make sure authors are very clear on how they get these features to work.

libs/blocks/notification/notification.js
```diff
@@ -0,0 +1,130 @@
...
+  const foreground = rest.pop();
+  if (background) decorateBlockBg(el, background);
+  foreground?.classList.add('foreground', 'container');
+  const text = foreground?.querySelector('h1, h2, h3, h4, h5, h6, p')?.closest('div');
+  text?.classList.add('text');
+  const picture = text?.querySelector('p picture');
+  const iconArea = picture?.closest('p');
+  iconArea?.classList.add('icon-area');
+  const fgMedia = foreground?.querySelector(':scope > div:not(.text) :is(img, video, a[href*=".mp4"])')?.closest('div');
+  const bgMedia = el.querySelector(':scope > div:not(.foreground) :is(img, video, a[href*=".mp4"])')?.closest('div');
```
narcis-radu, line 109: How would we handle video / mp4? Do we have guidelines for mobile (where videos might not need to get loaded)?

libs/blocks/notification/notification.js
```diff
@@ -0,0 +1,130 @@
...
+}
+
+function decorateLayout(el) {
+  const [background, ...rest] = el.querySelectorAll(':scope > div');
+  const foreground = rest.pop();
+  if (background) decorateBlockBg(el, background);
+  foreground?.classList.add('foreground', 'container');
+  const text = foreground?.querySelector('h1, h2, h3, h4, h5, h6, p')?.closest('div');
+  text?.classList.add('text');
+  const picture = text?.querySelector('p picture');
```
narcis-radu, line 105: I don't know whether `picture` is needed, we could do
```js
  const iconArea = text?.querySelector('p picture')?.closest('p');
```

libs/blocks/notification/notification.js
elan-tbx: That's actually all left entirely to auto-blocks and utility functions like `decorateBlockBg`... so video gets loaded like it normally would everywhere else. The Figma file linked in the Jira ticket in the description here has some guidelines on a lot of the authoring use cases, I'll make sure to get them included in the documentation for this.

narcis-radu (approved):

mokimo (approved):

NadiiaSokolova (approved):
After the last round of validations, no issues were found. Ready for Stage.

PR #2560, opened 2024-07-11
Bootstrapper script for milo feds blocks

bandana147:
Adding a bootstrapper script which would initiate fed's global footer milo block which can be consumed by non milo client.

JIRA - https://jira.corp.adobe.com/browse/MWPW-152861

Resolves: [MWPW-152861](https://jira.corp.adobe.com/browse/MWPW-152861)

**Test URLs:**
After: https://feds--milo--adobecom.hlx.page/?martech=off

**QA**:  [sample-page](https://bootstrapper-index--milo--bandana147.hlx.page/libs/navigation/index.html)

[Solution design](https://wiki.corp.adobe.com/pages/viewpage.action?pageId=3214883871#GlobalnavfromMiloasacomponent-3.SolutionDesign2:)

@sonawanesnehal3


mokimo (approved):
Cool. Do I understand it right that the consuming project would need to add the appropriate `<meta>` attributes themselves before loading the boostrap function?

So for the global-navigation / global-footer nothing changes, but the consuming project add the `<meta>` attributes to the header?

bandana147:
> Cool. 🚀 Do I understand it right that the consuming project would need to add the appropriate `<meta>` attributes themselves before loading the boostrap function?
> 
> So for the global-navigation / global-footer nothing changes, but the consuming project add the `<meta>` attributes to the header, for fetching a specific footer/nav?

Integration flow would look something like:

1. The consumer creates a folder under federal and creates their footer.docx file.
2. Adds contentRoot (e.g., /federal/foldername) and locale to window.fedsGlobalConfig.
3. Imports the scripts 

Adding it to the meta tag in their project also works.

libs/navigation/footer.js
```diff
@@ -0,0 +1,25 @@
+const blockConfig = {
+  name: 'global-footer',
+  targetEl: 'footer',
+  appendType: 'appendChild',
+};
+
+async function loadBlock() {
+  const { fedsGlobalConfig = {} } = window;
+  const miloLibs = fedsGlobalConfig.miloLibs || 'https://stage--milo--adobecom.hlx.page';
```
narcis-radu, line 9: do we want STAGE hlx to be the default value?

libs/navigation/footer.js
```diff
@@ -0,0 +1,25 @@
...
+    contentRoot: fedsGlobalConfig.contentRoot,
+    miloLibs: `${miloLibs}/libs`,
+    pathname: `/${fedsGlobalConfig.locale || ''}`,
+    locales: fedsGlobalConfig.locales || locales,
+  };
+
+  bootstrapBlock(clientConfig, blockConfig);
+}
+
+loadBlock();
```
mokimo, line 25: One thing that I wonder about, is if a consuming project should just import the module and pass params, rather than writing the configs on the window
```js
const {loadFooter} = import("https://adobe.com/libs/navigation/footer.js")
loadFooter({...params})
```

libs/utils/locales.js
```diff
@@ -0,0 +1,99 @@
+export default {
```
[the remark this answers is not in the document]
bandana147: I tested multiple times on this branch and every time it gives almost same score with main branch, sometimes a small drop. 

Screenshot 2024-07-11 at 19 25 28
Screenshot 2024-07-11 at 19 25 45

libs/navigation/footer.js
bandana147: I am not against it, but do we want our consumers to explicitly call the function to use the block?
Thoughts? @salonijain3 @narcis-radu @amitbikram

libs/navigation/footer.js
bandana147: So, this will be main branch actually `https://main--milo--adobecom.hlx.page` and we had a plan to maintain another branch call feds in which the url would be `https://feds-milo--adobecom.hlx.page` which would kind of act like a stage environment for the consumers. 
Since Milo's stage branch gets updated frequently, consuming client wouldn't have enough time to test whenever there is any change in the code.
cc: @salonijain3 @amitbikram

narcis-radu: I think the default branch should be a stable one, also it would need to be something like www.adobe.com to avoid asking teams to allowlist adobecom.hlx.page / adobecom.hlx.live

libs/utils/locales.js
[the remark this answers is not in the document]
narcis-radu: I'm more interested in the TBT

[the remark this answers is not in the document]
bandana147: Screenshot 2024-07-15 at 10 21 34
Screenshot 2024-07-15 at 10 21 44

libs/navigation/footer.js
amitbikram: This is a very good point. I think we should point to a default stable branch and most likely www.adobe.com

bandana147: Updated to `https://www.adobe.com`
@narcis-radu @amitbikram

libs/utils/locales.js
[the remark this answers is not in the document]
bandana147: @narcis-radu TBT was also almost similar (in fact, the Feds branch shows slightly lower TBT).

libs/navigation/footer.js
mokimo: That's also the fallback we have for normal federal content, adobe.com should always be accessible

nishantka (approved):

libs/navigation/navigation.css
```diff
@@ -0,0 +1,21 @@
+ /* Extracting the essential styles required for rendering the component independently */
```
narcis-radu, line 1: I don't think these would be enough though, but we can adjust later

narcis-radu (approved):

sharmrj (approved):

Deva309 (approved):

libs/navigation/footer.js
```diff
@@ -11,20 +11,21 @@ const envMap = {
...
+  const { locale, authoringPath, env = 'prod', privacyId, privacyLoadDelay = 3000 } = configs;
   const branch = new URLSearchParams(window.location.search).get('navbranch');
   const miloLibs = branch ? `https://${branch}--milo--adobecom.hlx.page` : envMap[env];
 
   // Relative path can't be used, as the script will run on consumer's app
   const { default: bootstrapBlock } = await import(`${miloLibs}/libs/navigation/bootstrapper.js`);
   const { default: locales } = await import(`${miloLibs}/libs/utils/locales.js`);
   const clientConfig = {
-    contentRoot,
+    privacyId,
```
narcis-radu, line 22: wouldn't we need all the privacy configs ?

bandana147: This is the only required field from configs, rest is already set here.
Screenshot 2024-07-18 at 13 15 22

narcis-radu (approved):

PR #2561, opened 2024-07-11
Adding support for federated urls in modals

bandana147:
* Adds support for federated URLs in modals to centralize content for the region selector.
* Moved out getFederatedUrl function from utilities file to federated.js file

Resolves: [MWPW-130787](https://jira.corp.adobe.com/browse/MWPW-130787) and [MWPW-153670](https://jira.corp.adobe.com/browse/MWPW-153670)

**Test URLs:**
- Before:  https://main--milo--adobecom.hlx.page/?martech=off
- After:  https://central-regionpicker--milo--bandana147.hlx.page/?martech=off
- Qa url: https://central-regionpicker--milo--bandana147.hlx.page/drafts/blaishram/document?martech=off

### DIFF
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -1,6 +1,7 @@
 /* eslint-disable no-underscore-dangle */
 /* eslint-disable import/no-cycle */
 import { createTag, getMetadata, localizeLink, loadStyle, getConfig } from '../../utils/utils.js';
+import { getFederatedUrl } from '../../utils/federated.js';
 
 const FOCUSABLES = 'a:not(.hide-video), button, input, textarea, select, details, [tabindex]:not([tabindex="-1"]';
 const CLOSE_ICON = `<svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 20 20">
@@ -91,7 +92,7 @@ function getCustomModal(custom, dialog) {
 }
 
 async function getPathModal(path, dialog) {
-  const block = createTag('a', { href: path });
+  const block = createTag('a', { href: getFederatedUrl(path) });
   dialog.append(block);
 
   // eslint-disable-next-line import/no-cycle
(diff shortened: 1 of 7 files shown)


libs/blocks/modal/modal.js
```diff
@@ -1,6 +1,7 @@
 /* eslint-disable no-underscore-dangle */
 /* eslint-disable import/no-cycle */
 import { createTag, getMetadata, localizeLink, loadStyle, getConfig } from '../../utils/utils.js';
+import { getFederatedUrl } from '../../utils/federated.js';
```
mokimo, line 4: I'd conditionally import this just before ` const block = createTag('a', { href: getFederatedUrl(path) });` only IF an URL includes 'federal' - I imagine it's a quite niche use-case for the modal, thus we don't want to have this import for every single modal use-case

mokimo (commented):
A small improvement before this is ready from a code perspective.

The region picker is a tricky element, as every consumer supports different languages. That's something that I see as potentially tricky and might need some more thoughts. It's also why it was initially left on each consumer after introducing the content-centralisation (federal) feature

libs/utils/federated.js
```diff
@@ -28,3 +28,17 @@ export const getFederatedContentRoot = () => {
 
   return federatedContentRoot;
 };
+
+// TODO we should match the akamai patterns /locale/federal/ at the start of the url
+// and make the check more strict.
+export const getFederatedUrl = (url = '') => {
```
mokimo, line 34: It would be great to start a unit test file for `federated.js` covering those functions. They should already be covered within the `utilities.test.js` within the navigation context, but given that we extract them to be consumed anywhere - it makes sense to have it's own test file

libs/blocks/modal/modal.js
bandana147: Added the conditional import.

mokimo (approved):
Starting the unit test file would be great, but otherwise looks good to me! 👍

libs/utils/federated.js
narcis-radu: Thank you, @bandana147 !

narcis-radu (approved):

Deva309 (approved):

sigadamvenkata:
Now the getFederatedContentRoot() & getFederatedUrl() methods reside in same /libs/utils/federated.js. PF below
image

JackySun9:
@bandana147 could you add unit test for new code in modals?

PR #2563, opened 2024-07-11
MWPW-152270 - Performance updates and small changes to marketingtech.

bbythewa:
* Conditionally load assets from domain proxies to allow for better performance.
* Change default launch url for dev.

Resolves: [MWPW-152270](https://jira.corp.adobe.com/browse/MWPW-152270)

**Test URLs:**
- After: https://marketingtech-updates--milo--adobecom.hlx.page/?martech=off


mokimo (approved):
Seems like one test is failing, apart of that this looks good.
```js
    it('have target be set to gnav and save in config', async () => {
      window.fetch = sinon.stub().returns(htmlResponse());
      document.head.innerHTML = await readFile({ path: './mocks/mep/head-target-gnav.html' });
      await utils.loadArea();
      const resultConfig = utils.getConfig();
      expect(resultConfig.mep.targetEnabled).to.equal('gnav');
    });
```

We need to make sure this doesn't get merged before the akamai changes are live though.

libs/martech/martech.js
```diff
@@ -190,8 +182,23 @@ const setupEntitlementCallback = () => {
...
+    || hostname === 'www.stage.adobe.com'
+    || hostname === 'milo.adobe.com'
+    || hostname === 'milo.stage.adobe.com'
+    || hostname === 'business.adobe.com'
+    || hostname === 'business.stage.adobe.com'
+    || hostname === 'blog.adobe.com'
+    || hostname === 'blog.stage.adobe.com'
+  );
+}
+
```
robert-bogos, line 199: This could be shorter
```suggestion
const isProxied = () => [
    'www.adobe.com', 'www.stage.adobe.com', 'milo.adobe.com',
    'milo.stage.adobe.com', 'business.adobe.com', 
    'business.stage.adobe.com', 'blog.adobe.com', 'blog.stage.adobe.com'
  ].includes(window.location.hostname)
```

robert-bogos (approved):
Looks good 👍 Just please make sure to check that failing unit test.
I left one small suggestion

libs/martech/martech.js
```diff
@@ -190,8 +182,23 @@ const setupEntitlementCallback = () => {
   );
 };
 
+function isProxied() {
+  // return window.location.hostname.endsWith('adobe.com');
+  const { hostname } = window.location;
+  return (
```
narcis-radu, line 188: I think this would work better
```suggestion
  return /^(www|milo|business|blog)(\.stage)?\.adobe.com$/.test(hostname)
```

npeltier: nit:  `return /^(www|milo|business|blog)(\.stage)?\.adobe\.com$/.test(hostname)
`

libs/martech/martech.js
```diff
@@ -190,8 +182,23 @@ const setupEntitlementCallback = () => {
...
+  return /^(www|milo|business|blog)(\.stage)?\.adobe.com$/.test(hostname)
+    hostname === 'www.adobe.com'
+    || hostname === 'www.stage.adobe.com'
+    || hostname === 'milo.adobe.com'
+    || hostname === 'milo.stage.adobe.com'
+    || hostname === 'business.adobe.com'
+    || hostname === 'business.stage.adobe.com'
+    || hostname === 'blog.adobe.com'
+    || hostname === 'blog.stage.adobe.com'
+  );
```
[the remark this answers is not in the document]
bbythewa: fixed

bbythewa:
With the help of @chrischrischris , the tests are all passing now.  Thanks for the help!

All of the proxies are available on stage right now, so we can push this to stage, but we should NOT merge this to production until we have successfully tested all aspects of this as well as the proxies have been created. :) Thanks!

mokimo:
@bbythewa on milo we don't have a great distinction of prod / stage. Stage is essentially "pre-prod" and just exists for testing reasons just before going live.
So, IMO the best course of action would be just testing if the proxies work on stage / prod.
Once akamai changes are on prod we merge this into stage, test and then it will automatically be promoted to prod by our automated process once testing finished.

mokimo:
@bbythewa once you add the `Ready for Stage` label the process would merge this into stage on the next possible occasion automatically. Once in stage, you typically only have a day-ish with this residing on stage for regression testing before this goes onto prod.

The general advice is: Test on the feature branch, use [overrides](https://developer.chrome.com/docs/devtools/overrides) for any sort of testing on prod if you need and once testing has finished a QA should apply the `Ready for Stage` label which takes the code to (stage and) prod

PR #2570, opened 2024-07-15
MWPW-153703: UAR Milo Localization | JP - Page jumps, enlarges the cards and duplicates copy

JackySun9:
* Bug fix for UAR Localization: JP - Page jumps, enlarges the cards and duplicates copy

Resolves: [MWPW-153703](https://jira.corp.adobe.com/browse/MWPW-153703)

**Test URLs:**
- Before: https://stage--cc--adobecom.hlx.page/jp/creativecloud/plan-recommender/quiz2?martech=off
- After: https://stage--milo--jackysun9.hlx.page/drafts/quiz/plan-recommender-var/quiz?martech=off

How to test?
1. Test JP page
https://stage--cc--adobecom.hlx.page/jp/creativecloud/plan-recommender/quiz2?martech=off&milolibs=stage--milo--JackySun9

2. Test EN Page
https://stage--cc--adobecom.hlx.page/creativecloud/plan-recommender/quiz?martech=off&milolibs=stage--milo--JackySun9

3. Test DE Page
https://stage--cc--adobecom.hlx.page/de/creativecloud/plan-recommender/quiz2?martech=off&milolibs=stage--milo--JackySun9


mokimo (approved):

colloyd (approved):
Looks like our assumptions about the issue are correct. While this fix does resolve the issue, it also leaves our JP quizzes without support for the JP word wrap feature. 

It would be a good idea to do some follow up, get with the dev who authored jp word wrap and see if there's a way in which we can support the feature, but do so in a way that is friendly to the quiz Preact app - likely we would need to import and implement it ourselves in the quiz block manually. 

Or maybe he would look at the nature of quizzes and the cards options and determine that it's not a use-case that the feature makes sense for?

JackySun9:
> Looks like our assumptions about the issue are correct. While this fix does resolve the issue, it also leaves our JP quizzes without support for the JP word wrap feature.
> 
> It would be a good idea to do some follow up, get with the dev who authored jp word wrap and see if there's a way in which we can support the feature, but do so in a way that is friendly to the quiz Preact app - likely we would need to import and implement it ourselves in the quiz block manually.
> 
> Or maybe he would look at the nature of quizzes and the cards options and determine that it's not a use-case that the feature makes sense for?

I checked the quiz page which is good, and we still keep the function for quiz-result page, we should be good

PR #2576, opened 2024-07-16
MWPW-154480 Split global-navigation.test.js into multiple files

chrischrischris:
Brings total time to run tests from 40s -> 19s on my machine.

There are no code changes in this PR, only moving of code from global-navigation.test.js into multiple test files.

Unfortunately it does not speed up the running of the unit tests on the PR runs.

Resolves: [MWPW-154480](https://jira.corp.adobe.com/browse/MWPW-154480)

**Test URLs:**
- After: https://split-gnav-tests--milo--adobecom.hlx.page/?martech=off


sukamat (approved):

hparra (approved):
There was an increase in coverage 🧐

Screenshot 2024-07-16 at 14 41 30

chrischrischris:
@hparra I have no idea how/why that would happen.  Not going to complain though 😆

colloyd (approved):

mokimo (approved):
Makes sense, as the tests can't run in parallel otherwise, thanks for tackling this! The global-navigation tests always lagged behind all the other completions locally

PR #2583, opened 2024-07-17
Added events

zagi25:
Feature:
* Added events to track navigation rendering

Hi Milo Team, on Partner Portals we want to implement feature that will allow us to show some user data in navigation, or to remove some elements from navigation based on the user data. User data is stored in the cookie that we set when user logs in.

Because navigation is rendered asynchronously, currently we don't have a way of knowing when elements will appear in the DOM. So we would like to introduce 3 events that will give us this information: 

- `feds:navWrapper:decorated` for when main navigation elements are added to the DOM
- `feds:menu:decorated` for when main navigation dropdowns are added to the DOM
- `feds:profileDropdown:decorated` for when profile dropdown is added to the DOM

Resolves: [MWPW-152981](https://jira.corp.adobe.com/browse/MWPW-152981)

**Test URLs:**
- After: https://gnav-events--milo--zagi25.hlx.page/?martech=off

### DIFF
--- a/libs/blocks/global-navigation/features/profile/dropdown.js
+++ b/libs/blocks/global-navigation/features/profile/dropdown.js
@@ -66,6 +66,7 @@ class ProfileDropdown {
     if (this.openOnInit) trigger({ element: this.buttonElem });
 
     this.decoratedElem.append(this.dropdown);
+    window.dispatchEvent(new CustomEvent('feds:profileDropdown:decorated'));
   }
 
   async getData() {
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -354,6 +354,7 @@ class Gnav {
       </div>`;
 
     this.block.append(this.elements.curtain, this.elements.aside, this.elements.topnavWrapper);
+    window.dispatchEvent(new CustomEvent('feds:navWrapper:decorated'));
   };
 
   addChangeEventListeners = () => {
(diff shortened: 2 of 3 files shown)


sharmrj:
I believe the correct way to do this would be through MEP and target. MEP is already modifying the global-navigation markup based on a set of rules. You also might end up conflicting with what MEP is trying to do if you try and modify the markup on your own.

zagi25:
@sharmrj Thanks for the reply. We investigated MEP, correct me if I am wrong, but MEP will replace specific part of the page (or whole page etc.) based on target value. This can be useful for us in some cases but the additional challenge that we face is that we want to show user data in profile dropdown (eg. company name), this information is user specific and my understanding is that MEP doesn't handle this use case. If MEP can handle this, can you please leave the link to the documentation where this is explained so I can investigate further. Thanks.

sharmrj:
> @sharmrj Thanks for the reply. We investigated MEP, correct me if I am wrong, but MEP will replace specific part of the page (or whole page etc.) based on target value. This can be useful for us in some cases but the additional challenge that we face is that we want to show user data in profile dropdown (eg. company name), this information is user specific and my understanding is that MEP doesn't handle this use case. If MEP can handle this, can you please leave the link to the documentation where this is explained so I can investigate further. Thanks.

From looking at the XD linked on your JIRA ticket, if you want to change the content of the profile, you'll want to look into how the Universal Nav, which in turn consumes the [account menu mini app](https://wiki.corp.adobe.com/display/IdentityACCM/Account+Menu+Mini-App). You might want to look into that to see if they can fulfill your profile needs. You can look into this wiki as well: https://wiki.corp.adobe.com/display/IdentityACCM/Account+Identity+Personalization+Guide

mokimo:
I'll close this PR as there seem to be better solutions. @zagi25 feel free to re-open if you disagree

PR #2590, opened 2024-07-18
Upgrade web test runner and switch to playwright from puppeteer

mokimo:
Trigger a test run


web-test-runner.config.mjs
```diff
@@ -17,6 +18,10 @@ function customReporter() {
   };
 }
 export default {
+  playwright: true,
+  browsers: [
+    playwrightLauncher({ product: 'chromium', launchOptions: { headless: true } }),
```
mokimo, line 23: Added benefit, in the future we could run the unit tests on more products. With the big caveat that currently most of the unit tests fail on both firefox and webkit.
```
    playwrightLauncher({ product: 'chromium' }),
    playwrightLauncher({ product: 'firefox' }),
    playwrightLauncher({ product: 'webkit' }),
```

narcis-radu (approved):

robert-bogos (approved):

SilviuLCF (commented):
Huge improvement from 10 test failing on MAIN 

Still there is one failing 

❌ Bulk Publish Tool > can toggle job timing flyout
      AssertionError: expected true to be false
      + expected - actual
      
      -true
      +false
      
      at n.<anonymous> (test/blocks/bulk-publish-v2/bulk-publish-v2.test.js:178:63)

PR #2593, opened 2024-07-18
MWPW-146528[MILO][MEP][ANALYTICS] Add attribute to content changed by Target for analytics and MWPW-152274

markpadbe:
In Dexter, there are scripts that add analytics to portions of a page that have been updated by Target. These scripts key off an attribute added by the testing library the T&O team uses. As pages migrate to milo, these scripts need the attribute once again. 

- add data-adobe-target-testid to elements updated by mep, but only when chosen experience is a Target experience.
- moved some mep-related code from fragment.js, utils.js and martech.js to personalization.js
- added an init function to personalization.js to handle some of the logic that used to be in utils/utils.js
- updated unit tests to use the init function
- added unit tests around new functionality
- removed getExpFromParam function since it was preliminary version of the mep parameter and is no longer used 

Resolves: [MWPW-146528](https://jira.corp.adobe.com/browse/MWPW-146528)
Also resolves: 
[MWPW-152274 Move entitlements.js request up to happen while waiting on sstats / etc.](https://jira.corp.adobe.com/browse/MWPW-152274)
[MWPW-152275 Move manifest.json request up while waiting on sstats / etc.](https://jira.corp.adobe.com/browse/MWPW-152275)
[MWPW-154998 [MEP][MILO] Manifests do not execute in the right order when there is a disabled manifest](https://jira.corp.adobe.com/browse/MWPW-154998)

**Instructions for testing:**
Use the MEP button to switch between experiences.  For each, copy and paste the following into the console: `document.querySelectorAll('[data-adobe-target-testid]')`

In the before link, no experience will have elements with the attributes.

**Test URLs:**
--Before: https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q3/addanalyticsattribute/addanalyticsattribute
--After: https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q3/addanalyticsattribute/addanalyticsattribute?milolibs=mepmovefromutils

In the after links, expected value for any found elements should be "myoverride" and the amount of found elements should match the numbers listed below.

Number of expected found elements by experience:
target-replacepage: 1
target-otheractions: 8
all others: 0

**Test URLs:**
--Before: https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q3/addanalyticsattribute/inblockgnavversion
--After: https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q3/addanalyticsattribute/inblockgnavversion?milolibs=mepmovefromutils

In the after links, expected value for any found elements should be "apples" and the amount of found elements should match the numbers listed below.

Number of expected found elements by experience:
target-block: 1
target-gnavfragment: 1
target-gnavchainfragments: 1
target-gnavwithinset: 2
all others: 0

**Test URL for psi check:**
https://mepmovefromutils--milo--adobecom.hlx.page/?martech=off


libs/blocks/fragment/fragment.js
```diff
@@ -117,15 +113,9 @@ export default async function init(a) {
...
   updateFragMap(fragment, a, relHref);
-
-  if (a.dataset.manifestId) {
-    if (inline) {
-      setManifestIdOnChildren(sections, a.dataset.manifestId);
-    } else {
-      fragment.dataset.manifestId = a.dataset.manifestId;
-    }
+  if (a.dataset.manifestId || a.dataset.adobeTargetTestid) {
+    updateFragDataProps(a, inline, sections, fragment);
```
narcis-radu, line 117: we should dynamically import personalization.js and use updateFragDataProps only if needed
```suggestion
     const { updateFragDataProps } = await import('../../features/personalization/personalization.js');
     updateFragDataProps(a, inline, sections, fragment);
```

markpadbe: pushed fix for this. Thanks.

vgoodric (approved):

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -47,13 +47,12 @@ export const logErrorFor = async (fn, message, tags) => {
...
-export function addMepHighlight(el, source) {
-  let { manifestId } = source.dataset;
-  if (!manifestId) {
-    const closestManifestId = source?.closest('[data-manifest-id]');
-    if (closestManifestId) manifestId = closestManifestId.dataset.manifestId;
-  }
+export function addMepHighlightAndTargetId(el, source) {
+  let { manifestId, targetManifestId } = source.dataset;
+  manifestId = manifestId ?? source?.closest('[data-manifest-id]')?.dataset?.manifestId;
+  targetManifestId = targetManifestId ?? source?.closest('[data-adobe-target-testid]')?.dataset?.adobeTargetTestid;
```
chrischrischris, line 53: super nit: use `??==` instead

chrischrischris (approved):
Nice!

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -47,13 +47,12 @@ export const logErrorFor = async (fn, message, tags) => {
...
 
-export function addMepHighlight(el, source) {
-  let { manifestId } = source.dataset;
-  if (!manifestId) {
-    const closestManifestId = source?.closest('[data-manifest-id]');
-    if (closestManifestId) manifestId = closestManifestId.dataset.manifestId;
-  }
+export function addMepHighlightAndTargetId(el, source) {
+  let { manifestId, targetManifestId } = source.dataset;
+  manifestId ??= source?.closest('[data-manifest-id]')?.dataset?.manifestId;
```
denlight, line 52: ??= is new for me. Never seen it used before =). Maybe consider that a lot of people are like me and also haven't seen it used and may be confused ;) I dunno, up to you

libs/features/personalization/personalization.js
```diff
@@ -424,24 +403,39 @@ const addHash = (url, newHash) => {
   }
 };
 
+const setDataIdOnChildren = (sections, id, value) => {
+  [...sections[0].children].forEach(
```
denlight, line 407: do we know that sections[0] will always exist ?

denlight (approved):

libs/features/personalization/personalization.js
vgoodric, line 407:
```suggestion
  if (!sections.length) return;
  [...sections[0].children].forEach(
```

markpadbe: @denlight, @vivgoodrich, setDataIdOnChildren is only called within the updateFragDataProps function which is only called in the fragment.js init function. There is already a check for sections.length before the sections variable is passed into updateFragDataProps (fragment.js, line 102) so the current code is safe. I can come back to this in a future ticket, but I suggest we leave it as is tonight to avoid last second changes. Thanks.

markpadbe: @vivgoodrich, please see comment above.

libs/blocks/global-navigation/utilities/utilities.js
markpadbe: @denlight, I take your point about the Nullish coalescing assignment operator (I had to look up the name up again today when Chris suggested it ;)), but it is nice and concise and we've already used in a couple of other places in personalization.js so let's keep it. Thanks.

PR #2594, opened 2024-07-18
[Modal] Dedup Ongoing Dialogs for Same Fragment

JingleH:
There exists a race condition that when the same modal hashes appear multiple times on the page, the same fragment could be loaded more than once. More specifically, after the hash change, there is a period when getModal() needs to fetch its content but the dialog-modal hasn't been appended to document yet. If another modal on the page gets its init() called (via section preloadLinks) during this period, all the early-exit safeguard in init() we currently have will fail. This PR aims to cover this condition by keeping track of what dialogs are being loaded but not yet appended.

How to reproduce:
Locally or using overrides, in modal.js, you can fake a longer getPathModal() call by making it into something like `await Promise.all([getPathModal(details.path, dialog), new Promise((r) => setTimeout(r, 4000))]);`. Then to delay the second modal, either throw in more content, or just manually make it slower by making init() async and adding `if (document.querySelector('a.modal') !== el) { await new Promise((r) => setTimeout(r, 2000))}` to the beginning. Then clicking the primary CTA should make the FaaS fragment start overwriting itself and acting weird.

Resolves: https://jira.corp.adobe.com/browse/MWPW-147304

**Test URLs:**
- Before: https://stage--milo--adobecom.hlx.page/drafts/jinglhua/modal?martech=off
- After: https://modal-dedup--milo--adobecom.hlx.page/drafts/jinglhua/modal?martech=off

### DIFF
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -249,7 +249,7 @@ export function delayedModal(el) {
 export default function init(el) {
   const { modalHash } = el.dataset;
   if (delayedModal(el) || window.location.hash !== modalHash || document.querySelector(`div.dialog-modal${modalHash}`)) return null;
-  if (dialogLoadingSet.has(modalHash?.replace('#', ''))) return null;
+  if (dialogLoadingSet.has(modalHash?.replace('#', ''))) return null; // prevent duplicate modal loading
   const details = findDetails(window.location.hash, el);
   return details ? getModal(details) : null;
 }
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -13,6 +13,7 @@ const CLOSE_ICON = `<svg xmlns="http://www.w3.org/2000/svg" width="20" height="2
 
 let isDelayedModal = false;
 let prevHash = '';
+const dialogLoadingSet = new Set();
 
 export function findDetails(hash, el) {
   const id = hash.replace('#', '');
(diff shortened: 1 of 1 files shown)


libs/blocks/modal/modal.js
```diff
@@ -246,6 +249,7 @@ export function delayedModal(el) {
 export default function init(el) {
   const { modalHash } = el.dataset;
   if (delayedModal(el) || window.location.hash !== modalHash || document.querySelector(`div.dialog-modal${modalHash}`)) return null;
+  if (dialogLoadingSet.has(modalHash?.replace('#', ''))) return null;
```
mokimo, line 252: I'd suggest either a comment, unit test or both. Since it's not super obvious on why this line of code exists
```suggestion
  if (dialogLoadingSet.has(modalHash?.replace('#', ''))) return null; // prevent duplicate modal loading
```

mokimo (approved):
Not sure why this is a draft PR, but looks good to me. I'd love to see a unit test for this case

JingleH:
> Not sure why this is a draft PR, but looks good to me. I'd love to see a unit test for this case

Something else came up yesterday while I was reading the PR description 🥲 so I didn't get to finish it... I planned to record a video to reproduce the issue, but if @mokimo agrees that it's trivial enough, I am happy to not do that 🥺 I am going to open it up and ask for more approvals. Thanks, Okan!

libs/blocks/modal/modal.js
JingleH: Thanks @mokimo ! I personally feel that a comment would be enough, because otherwise it will bloat up modal's unit tests by quite some boilerplate stuff given how modal.test.js is currently structured and we'll need to mock the fetch again. Just from the code level, this change should be trivial enough for us to just feel confident 😆

chrischrischris (approved):

vhargrave (approved):
looks good!

PR #2597, opened 2024-07-19
[MWPW-152674] [Gray Box] Desktop gnav not hidden when device view is open

robert-bogos:
### Description
This PR solves the issue where the desktop global navigation is not hidden on mobile & tablet view.

### Related Issue
Resolves: [MWPW-152674](https://jira.corp.adobe.com/browse/MWPW-152674)

### Screenshots

**Before:**

image

**After:**

image

### Test URLs:
- Before: https://main--bacom-graybox--adobecom.hlx.page/pocone/customer-success-stories/graybox-changed
- After: https://main--bacom-graybox--adobecom.hlx.page/pocone/customer-success-stories/graybox-changed?milolibs=mwpw-152674-graybox-gnav--milo--robert-bogos

### URL for PSI check: (irrelevant for testing the changes)
https://mwpw-152674-graybox-gnav--milo--robert-bogos.hlx.page/drafts/rbogos/page-default?martech=off

### DIFF
--- a/libs/blocks/graybox/graybox.css
+++ b/libs/blocks/graybox/graybox.css
@@ -90,7 +90,7 @@
   position: fixed;
   right: 0;
   top: 15%;
-  z-index: calc(var(--above-all) + 1);
+  z-index: calc(var(--above-all) + 2);
 }
 
 .graybox-container .gb-toggle {
@@ -190,7 +190,7 @@
 }
 
 .dialog-modal.graybox-modal {
-  z-index: var(--above-all);
+  z-index: calc(var(--above-all) + 1);
 }
 
 .dialog-modal.graybox-modal.mobile > div {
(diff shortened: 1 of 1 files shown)


mokimo (approved):

libs/blocks/graybox/graybox.css
```diff
@@ -190,7 +190,7 @@
 }
 
 .dialog-modal.graybox-modal {
-  z-index: var(--above-all);
+  z-index: calc(var(--above-all) + 1);
```
mokimo, line 193: Shouldn't the above all value be updated instead?

robert-bogos: That value is also used in the desktop view. If we only update the above-all, the bug will persist.

robert-bogos: That will cause a different bug in the desktop view, where the highlighted sections will cover the gnav when scrolling.

SilviuLCF (approved):
Verified , testing details 
https://jira.corp.adobe.com/browse/MWPW-152674

narcis-radu (approved):

PR #2598, opened 2024-07-19
Combining martech.js into utils.js to reduce scripts in critical path leading to interact call.

bbythewa:
* Merges martech.js contents into utlis.js to reduce the number of network calls needed to get to the /interact call. 
* Makes performance around 80(+-80)ms better. 

**Test URLs:**
- After: https://marketingtech-combine-files--milo--adobecom.hlx.page/?martech=off


bbythewa:
@chrischrischris , I'm completely creeping into your area, so please feel free to reject this purely based on that. :)

Also, if you have some other reason why you want to keep martech.js separate, by all means, reject this PR.  

My main thought process was simply that while I was doing the last PR, I noticed that we still have to wait anywhere from 50ms to 500ms just for the martech.js file to load...It was only around 500 lines of code, so I figured why not just combine it with utils.js and save us some time in the critical path to the /interact or target call. 

I also added a test to increase test coverage...

Let me know your thoughts and please feel free to be critical. :) 

CC: @vgoodric

mokimo (changes requested):
Hey @bbythewa in theory this is a great idea, however for the most use-cases - this will hurt performance.

We only load Martech "pre-LCP" when target is enabled where it is crucial to have Martech to render a page.

In all other cases, we can wait until the user sees the first LCP element which in turn is one of the most important and hardest to optimize LH & CWV metric.

So, the conclusion to move this into utils is the correct one, however we already optimized the loading path of Martech in order to be even more efficient.

npeltier:
> Hey @bbythewa in theory this is a great idea, however for the most use-cases - this will hurt performance.
> 
> We only load Martech "pre-LCP" when target is enabled where it is crucial to have Martech to render a page.
> 
> In all other cases, we can wait until the user sees the first LCP element which in turn is one of the most important and hardest to optimize LH & CWV metric.
> 
> So, the conclusion to move this into utils is the correct one, however we already optimized the loading path of Martech in order to be even more efficient for a wider variety of use-cases.

well @mokimo that great explanation of yours drills down the decision making point to:
do we want to optimize pages with target enabled or pages with target disabled. And i would tend to think this decision does not belong to us, but marketing, no?

bbythewa:
> Hey @bbythewa in theory this is a great idea, however for the most use-cases - this will hurt performance.
> 
> We only load Martech "pre-LCP" when target is enabled where it is crucial to have Martech to render a page.
> 
> In all other cases, we can wait until the user sees the first LCP element which in turn is one of the most important and hardest to optimize LH & CWV metric.
> 
> So, the conclusion to move this into utils is the correct one, however we already optimized the loading path of Martech in order to be even more efficient for a wider variety of use-cases.

Thanks for your explanation @mokimo.  

My POV is that it would not hurt even the cases where we delay LCP, but would dramatically help with target use cases.  Here is my rationale: 
- Since we are already waiting to load martech after LCP when we don't use target regardless of whether the logic exists in another file or not, any non-target use cases wouldn't be impacted.
- The only thing we are doing is adding a few extra bytes to utils.js...this in my view would not hurt LCP for non-target use cases.
- The use case it does help is what you described here where we load martech pre-LCP...and in those cases it helps by reducing LCP by quite a bit (comparatively).

I may be wrong. Happy to chat more about it.

mokimo:
> And i would tend to think this decision does not belong to us, but marketing, no?

Yeah, however it should be a mostly data driven decision IMO. I'd wonder on how much traffic goes to pages with target enabled vs no target. Even though @vivgoodrich argued that the most important pages do have it.

There are a couple more negative benefits to this though, on top of my mind:

- The `utils.js` file is growing and will keep growing and removing is a lot harder than adding. There's a maintenance aspect to this
- Separation of concerns, the martech code is currently nicely separated which makes changes easier as they are scoped to martech. Again maintenance

mokimo:
@bbythewa do we have any traffic data for milo with target pages vs no target? (If this is not public, feel free to share on #milo-dev or ping me on slack)

chrischrischris (commented):
I agree with @mokimo that we should keep the files separate.  It bloats utils.js for non-target pages and makes maintenance / code separation messier.  However I do believe that you're right about not wanting to wait for it to load.  I did some initial experimentation and I think what we can do is have the consumer scripts.js add this before calling `loadArea`:

```
if (getMetadata('target') === 'on') {
  loadLink('/libs/martech/martech.js', { as: 'script', rel: 'modulepreload' });
}
```
We could also instead load in parallel with utils.js by adding this code to head.html instead (along with a preload for utils.js).

bbythewa:
@chrischrischris , 

Last thoughts before I give up on this PR 😄 :

- I may be wrong on the purpose of the code, but it still feels like there is a good portion of martech and personalization code/logic in utils.js (https://github.com/adobecom/milo/blob/stage/libs/utils/utils.js#L857-L1001), so martech/personalization/mep are already split across two files right now with half of the logic residing in both.  I personally like having that code in utils.js for the other reason of simply being easier to follow the logic of target/personalization/etc.  That's just personal preference.
- I agree that it does add extra code to utils.js, but what is our threshold? In my mind, adding 250 lines isn't a lot, but "a lot" depends on who you ask. 😆  Like if I were to simplify the logic to say only an additional 100 lines, then would it be okay?

I only bring up those points in case they feel relevant and in case that makes a difference before we discard the PR...

Having said all of that, I love your idea of preloading martech.js somewhere!  I think that does accomplish the same goal from a performance standpoint.  I personally would prefer to not rely on consumers to load this themselves simply because it takes time and we end up having to talk to every consumer about it and convince them, but I will defer to you on best way to implement these kinds of things.

chrischrischris:
@bbythewa 
Yes there is def a decent amount of martech code in utils, with most of it required to determine if when martech/mep should be loaded.  As it stands utils might already be too bloated, but it's a hard thing to tame.

I've been doing some testing with prefetching and for a test milo page have reduced LCP from [5.7s](https://www.webpagetest.org/result/240718_AiDc1R_EZ7/1/details/) to [4.1s](https://www.webpagetest.org/result/240722_BiDc7N_EEM/1/details/).  Only requires a change in headers.xlsx.  It would require consumers to update their excel files, but since there is no code change involved it can be done quickly (and reverted just as quickly).  It does require enabling it only on pages where target is running, but those are most of the high impact pages.

I filed https://jira.corp.adobe.com/browse/MWPW-154478 for bacom, but other consumers can use the exact same headers.

bbythewa:
Awesome.  Thanks @chrischrischris . That's a huge improvement!  I'll close this PR for now in favor of the headers.xlsx change you are suggesting. 👍

mokimo:
> I agree that it does add extra code to utils.js, but what is our threshold? In my mind, adding 250 lines isn't a lot, but "a lot" depends on who you ask. 😆 Like if I were to simplify the logic to say only an additional 100 lines, then would it be okay?

@bbythewa to give you some context and don't let this unanswered: we started to fight for every single line of code added to utils. In some cases it's unavoidable, but we gotten _very_ protective of it

vgoodric:
I know this PR is closed, but I wanted to mention a few things ([PR mentioned below](https://github.com/adobecom/milo/pull/2593)):
1. we're already working on improving on what Chris described: only keep logic in utils if it's needed to decide if we need more
2. we moved some of the MEP code out of utils and I have ideas of some more we things we can do
3. same PR has moved some MEP code out of martech and into personalization. So if we do pull in the martech code, there would be less to pull in


Now that martech.js is down to 250 lines (8kb), maybe we should consider pulling the remaining martech code into personalization.js.  So when martech impacts LCP, it will already be with the other MEP code and not require a separate download.  But if it's not needed for MEP, it's loaded post LCP.  Downside is a bunch of pages will download non needed MEP code.  Upside is that won't be until post LCP.

Thoughts?

bbythewa:
> I know this PR is closed, but I wanted to mention a few things ([PR mentioned below](https://github.com/adobecom/milo/pull/2593)):
> 
> 1. we're already working on improving on what Chris described: only keep logic in utils if it's needed to decide if we need more
> 2. we moved some of the MEP code out of utils and I have ideas of some more we things we can do
> 3. same PR has moved some MEP code out of martech and into personalization. So if we do pull in the martech code, there would be less to pull in
> 
> Now that martech.js is down to 250 lines (8kb), maybe we should consider pulling the remaining martech code into personalization.js. So when martech impacts LCP, it will already be with the other MEP code and not require a separate download. But if it's not needed for MEP, it's loaded post LCP. Downside is a bunch of pages will download non needed MEP code. Upside is that won't be until post LCP.
> 
> Thoughts?

I'm all for combining files together and loading files in parallel as much as possible.  :) I also did notice that personalization.js and martech.js sometimes get loaded one after the other in sequence...combining them together would solve this for sure. :)

PR #2603, opened 2024-07-22
Mwpw 154887: Making gnav visible in base.css 

bandana147:
On a slow network, the global navigation (gnav) is visible before the complete styling is applied, as the CSS to make it visible is in global-navigation.css. 
To resolve this, the CSS is moved to base.css.

Resolves: [MWPW-154887](https://jira.corp.adobe.com/browse/MWPW-154887)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/snehal/unav11?martech=off
- After: https://mwpw-154887--milo--bandana147.hlx.page/drafts/snehal/unav11?martech=off

### DIFF
--- a/libs/blocks/global-navigation/base.css
+++ b/libs/blocks/global-navigation/base.css
@@ -112,13 +112,7 @@
 }
 
 header.global-navigation {
-  position: sticky;
-  top: 0;
-  z-index: 10;
-  background-color: var(--feds-background-nav--light);
   visibility: visible;
-  box-sizing: content-box;
-  overflow-x: clip;
 }
 
 /* Desktop styles */
--- a/libs/blocks/global-navigation/base.css
+++ b/libs/blocks/global-navigation/base.css
@@ -111,6 +111,16 @@
   align-items: center;
 }
 
+header.global-navigation {
+  position: sticky;
+  top: 0;
+  z-index: 10;
+  background-color: var(--feds-background-nav--light);
+  visibility: visible;
+  box-sizing: content-box;
+  overflow-x: clip;
+}
+
 /* Desktop styles */
 @media (min-width: 900px) {
   .feds-navLink,
--- a/libs/blocks/global-navigation/global-navigation.css
+++ b/libs/blocks/global-navigation/global-navigation.css
@@ -28,6 +28,15 @@
 }
 
 /* General */
+header.global-navigation {
+  position: sticky;
+  top: 0;
+  z-index: 10;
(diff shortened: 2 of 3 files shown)


sharmrj (approved):

mokimo (approved):

robert-bogos (approved):

narcis-radu (approved):
@spadmasa - we need to test on all consuming projects (homepage, CC, DC, BACOM)

libs/blocks/global-navigation/base.css
```diff
@@ -111,6 +111,10 @@
   align-items: center;
 }
 
+header.global-navigation {
```
overmyheadandbody, line 114: Based on https://github.com/adobecom/milo/blob/stage/libs/blocks/global-navigation/global-navigation.js#L302-L305, it could still be that there is unstyled markup until the decoration methods are done. It might be worth investigating how things look in that scenario and possibly load the base styles a bit later.

Another thing to keep in mind is that the footer also loads the base styles - https://github.com/adobecom/milo/blob/stage/libs/blocks/global-footer/global-footer.js#L97. In the (relatively unlikely) case that there are some errors in loading the top navigation that would prevent it from being visible, the footer logic would also try to make it visible. This is because the base styles were intended to be a common ground for both the header and the footer, while this change brings in specific header styles.

overmyheadandbody (commented):
There might be a few more things to consider before merging this, but overall it should still be an incremental step forward. As @narcis-radu mentioned, we should thoroughly test this, also on consumers; and also check whether this change can have any collisions with some MEP logic that might be hiding something.

spadmasa:
Validation done in below cloud urls  on slow network ,global navigation is visible with complete styling applied 
https://main--cc--adobecom.hlx.page/creativecloud?milolibs=mwpw-154887--milo--bandana147
image
https://main--dc--adobecom.hlx.page/acrobat/online/sign-pdf?milolibs=mwpw-154887--milo--bandana147
https://mwpw-154887--milo--bandana147.hlx.page/drafts/snehal/unav11?martech=off
https://main--bacom--adobecom.hlx.page/?milolibs=mwpw-154887--milo--bandana147
https://main--homepage--adobecom.hlx.page/homepage/index-loggedout?milolibs=mwpw-154887--milo--bandana147
https://main--blog--adobe.hlx.page/?milolibs=mwpw-154887--milo--bandana147

overmyheadandbody (approved):
Thanks for integrating the changes, @bandana147!

PR #2607, opened 2024-07-23
MWPW-154898 No wait for modal fragment

TsayAdobe:
- A section doesn't need to wait for loading of modal fragments. Marquee and modal can be loaded at the same time.
- This PR is one part of #2493, which has been closed. The other part will be addressed in [MWPW-154899](https://jira.corp.adobe.com/browse/MWPW-154899)

Resolve: [MWPW-154898](https://jira.corp.adobe.com/browse/MWPW-154898)

**Test URLs:**

Before: https://main--milo--adobecom.hlx.live/docs/authoring/blocks/marquee-anchors?martech=off
After: https://mwpw-154898--milo--adobecom.hlx.live/docs/authoring/blocks/marquee-anchors?martech=off

**DC Test URLs:**

Before: https://www.stage.adobe.com/acrobat/how-to/pdf-editor-pdf-files.html
Marquee and icons are loaded after modal.

BeforeNoWait

After: https://www.stage.adobe.com/acrobat/how-to/pdf-editor-pdf-files.html?milolibs=mwpw-154898
Marquee, icons, and modal are loaded at the same time. 

AfterNoWait

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -1150,6 +1150,16 @@ async function documentPostSectionLoading(config) {
   document.body.appendChild(createTag('div', { id: 'page-load-ok-milo', style: 'display: none;' }));
 }
 
+export function partition(arr, fn) {
+  return arr.reduce(
+    (acc, val, i, ar) => {
+      acc[fn(val, i, ar) ? 0 : 1].push(val);
+      return acc;
+    },
+    [[], []],
+  );
+}
+
 async function processSection(section, config, isDoc) {
   const inlineFrags = [...section.el.querySelectorAll('a[href*="#_inline"]')];
   if (inlineFrags.length) {
@@ -1163,8 +1173,10 @@ async function processSection(section, config, isDoc) {
   }
 
   if (section.preloadLinks.length) {
-    const preloads = section.preloadLinks.map((block) => loadBlock(block));
+    const [modals, nonModals] = partition(section.preloadLinks, (block) => block.classList.contains('modal'));
+    const preloads = nonModals.map((block) => loadBlock(block));
     await Promise.all(preloads);
+    modals.forEach((block) => loadBlock(block));
   }
 
   const loaded = section.blocks.map((block) => loadBlock(block));
--- a/test/utils/utils-partition.test.js
+++ b/test/utils/utils-partition.test.js
@@ -1,0 +1,27 @@
+import { expect } from '@esm-bundle/chai';
+import { partition } from '../../libs/utils/utils.js';
+
+describe('Utils Partition', () => {
+  it('partition array', async () => {
+    const arr = [
+      { name: 'John', age: 23 },
+      { name: 'James', age: 40 },
+      { name: 'Mary', age: 31 },
+    ];
+    const result = partition(arr, (x) => x.age > 30);
+    expect(result[0]).to.have.deep.members(
+      [
+        { name: 'Mary', age: 31 },
+        { name: 'James', age: 40 },
+      ],
+    );
+    expect(result[1]).to.have.deep.members(
+      [{ name: 'John', age: 23 }],
+    );
+  });
+  it('empty array', async () => {
+    const arr = [];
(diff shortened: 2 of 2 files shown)


libs/utils/utils.js
```diff
@@ -1163,8 +1173,10 @@ async function processSection(section, config, isDoc) {
   }
 
   if (section.preloadLinks.length) {
-    const preloads = section.preloadLinks.map((block) => loadBlock(block));
+    const [modals, nonModals] = partition(section.preloadLinks, (block) => block.classList.contains('modal'));
+    const preloads = nonModals.map((block) => loadBlock(block));
     await Promise.all(preloads);
+    modals.forEach((block) => loadBlock(block));
```
mokimo, line 1179: @mirafedas can you confirm https://github.com/adobecom/milo/pull/2094 still works?

libs/utils/utils.js
```diff
@@ -1163,8 +1173,10 @@ async function processSection(section, config, isDoc) {
   }
 
   if (section.preloadLinks.length) {
-    const preloads = section.preloadLinks.map((block) => loadBlock(block));
+    const [modals, nonModals] = partition(section.preloadLinks, (block) => block.classList.contains('modal'));
+    const preloads = nonModals.map((block) => loadBlock(block));
```
mokimo, line 1177: If we don't ever need to wait for modals, should we just filter them before they even land in the preload section and let them load normally later on?
So in my mind the fix is filtering modals completely from here: https://github.com/adobecom/milo/blob/stage/libs/utils/utils.js#L798

mokimo (approved):
LGTM, however I don't have the most context on this portion of the utils.

@chrischrischris Could you have a look? We might get away with just filtering the modals out

joaquinrivero (approved):

chrischrischris (commented):
I don't think it will affect it, but has this been tested with delayed modals?

libs/utils/utils.js
```diff
@@ -1150,6 +1150,16 @@ async function documentPostSectionLoading(config) {
...
 
+export function partition(arr, fn) {
+  return arr.reduce(
+    (acc, val, i, ar) => {
+      acc[fn(val, i, ar) ? 0 : 1].push(val);
+      return acc;
+    },
+    [[], []],
+  );
+}
```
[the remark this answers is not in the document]
TsayAdobe: It's easier for unit tests with `export`. Testability is important.

Blainegunn (approved):

TsayAdobe:
> I don't think it will affect it, but has this been tested with delayed modals?

There is a unit test. It passed.
https://github.com/adobecom/milo/blob/c00c7777499c8b1ba594016ea735818a54a7b64e/test/blocks/modals/modals.test.js#L201-L217

libs/utils/utils.js
chrischrischris, line 1161: I know that this is probably the most efficient implementation of `partition`, but I think that considering the arrays being partitioned will always be small, I think readability should take precedence.  Also I think any testing of it can be part of general unit tests given it's simplicity.

cc @mokimo 

```suggestion
function partition(arr, fn) {
  return [
    arr.filter(fn),
    arr.filter((el, i) => !fn(el, i)),
  ];
}
```

mokimo: +1ing @chrischrischris opinion

sharmrj: There could be a middle ground to make it more readable [similar to the haskell implementation](https://hackage.haskell.org/package/ghc-internal-9.1001.0/docs/src/GHC.Internal.Data.OldList.html#partition)

```javascript

function partition(arr, predicate) {
    const select = ([ts, fs], x, ...args) => {
      if (predicate(x, ...args)) {
        ts.push(x);
      } else {
        fs.push(x);
      }
      return [ts, fs];
    };
    
  return arr.reduce(select, [[], []]);
}
```

The `select` tells the reader that we're folding over the array and 'selecting' the values matching the predicate, so the really only need to read the last line to understand the function.

PR #2599, opened 2024-07-22
[MWPW-154795] Style Feds Global-footer region picker drop-up variant (without hash)

sonawanesnehal3:
Add styling for region picker drop-up variant (without hash)

Resolves: [MWPW-154795](https://jira.corp.adobe.com/browse/MWPW-154795)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/snehal/unav11
- After: https://dropup--milo--adobecom.hlx.page/drafts/snehal/unav11

Before styles applied : 
image

After styles applied :
image

QA : [Sample Page](https://dropup--milo--adobecom.hlx.page/drafts/snehal/unav11)

PR #2611, opened 2024-07-23
Revert "[MWPW-154795] Style Feds Global-footer region picker drop-up variant (without hash)"

mokimo:
Reverts adobecom/milo#2599 based on https://github.com/adobecom/milo/pull/2599#issuecomment-2245428349

### DIFF
--- a/libs/blocks/global-footer/global-footer.css
+++ b/libs/blocks/global-footer/global-footer.css
@@ -142,13 +142,12 @@
   left: 0;
   display: none;
   padding: 20px;
-  min-width: 165px;
+  min-width: 130px;
   max-height: 300px;
   overflow-y: auto;
   background: var(--feds-background-nav--light);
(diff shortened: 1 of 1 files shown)


narcis-radu (approved):

overmyheadandbody (commented):
I'd like to get some context from the committers. The example page I used to showcase a potential issue is a demo page, so is the example provided in their PR. I checked against the Blog implementation and the new contribution doesn't seem to bring in anything new. I'd like to avoid reverting something due to a misunderstanding. I've tried isolating some other pages that use this pattern, but fell short so far.

overmyheadandbody (approved):
Approving for now to remove the risk of this getting to production without considering all the use-cases. The story connected to this enhancement is not a blocker or critical one, thus it could potentially wait until everything is clear.

PR #2622, opened 2024-07-24
[MWPW-153611] [Gray Box] environment aware links

robert-bogos:
## Description
Links on Graybox pages need to be environment aware.

**Examples of desired links conversions:** 
When on _**https://pocone.graybox.adobe.com/some-path**_:
`https://www.adobe.com/` -> `https://pocone.graybox.adobe.com/`
When on **_https://max24.business-graybox.adobe.com/some-path_**
`https://business.adobe.com/` -> `https://max24.business-graybox.adobe.com/` 

where 
`pocone` / `max` are graybox environment names. These can be anything.
`graybox` / `business-graybox` are the names of the graybox environments sharepoint container folder. Each supported Milo consumer folder will have a `–graybox` variant.

## Related Issue
Resolves: [MWPW-153611](https://jira.corp.adobe.com/browse/MWPW-153611)

## Testing instructions
`1.` go to the test URL
`2.` using the developer tools, add a breakpoint in `scripts.js`, where `LIBS` is defined.

Screenshot 2024-07-24 at 20 37 23

`3.` refresh the page, when the execution hits the break point. Overwrite, using the console, the `setLibs` function with `() => 'https://mwpw-153611-convert-graybox-urls--milo--robert-bogos.hlx.live/libs'`

image

`4.` resume the execution
`5.` scroll down on the page and inspect the `Link to be converted` link. It should send to `https://pocone.business-graybox.adobe.com/...`

Screenshot 2024-07-24 at 20 45 27

if you don't overwrite the `LIBS`, the `Link to be converted` link will send to `https://business.adobe.com/...`

## Test URL
https://pocone.business-graybox.adobe.com/customer-success-stories/graybox-links-conversion.html

## URL for PSI check: (irrelevant for testing the changes)
https://mwpw-153611-convert-graybox-urls--milo--robert-bogos.hlx.page/drafts/rbogos/page-default?martech=off


libs/blocks/graybox/graybox.js
```diff
@@ -264,7 +264,16 @@ const setupChangedEls = (globalNoClick) => {
   }
 };
 
+const transformLinks = () => {
+  const grayboxHostname = window.location.hostname;
```
overmyheadandbody, line 268: There isn't a need for local development where ports would also be involved, right?

libs/blocks/graybox/graybox.js
```diff
@@ -264,7 +264,16 @@ const setupChangedEls = (globalNoClick) => {
...
+  const consumerHostname = grayboxHostname.replace(/[^.]+\.(?:([^.]+)-)?graybox\./, (_, sub) => `${sub || 'www'}.`);
+  document.querySelectorAll(`a[href*="${consumerHostname}"]`).forEach((el) => {
+    el.href = el.href.replace(consumerHostname, grayboxHostname);
+  });
+};
+
 const grayboxThePage = (grayboxEl, grayboxMenuOff) => () => {
+  transformLinks();
```
overmyheadandbody, line 276: What happens to links loaded later on, such as the ones in the footer, for example? Does this logic apply to them as well, since those links are loaded with a delay?

overmyheadandbody (approved):
I'd guess my questions are outside the scope of this story, but async sections might become a topic if not already considered.

libs/blocks/graybox/graybox.js
robert-bogos: Correct, no ports are involved for this. So, I think we’re good with `.hostname`

libs/blocks/graybox/graybox.js
robert-bogos: Good catch 👍 The footer links will not be converted because of the loading delay. This is not an issue currently since navigations are grayed out by default.
I will create a new ticket for handling async sections, as this has more implications than just the links conversion, so it falls out of the scope of this ticket.

mokimo (approved):

sukamat (approved):

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-153611

PR #2624, opened 2024-07-24
MWPW-154901 Set debug=true on categories container

cmiqueo:
Sets the flag **_debug=true_** based on container type for Categories, so that the category pills above the card collection and filters work as expected

Resolves: [MWPW-154901](https://jira.corp.adobe.com/browse/MWPW-154901)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/tools/caas
- After: https://mwpw-154901--milo--adobecom.hlx.page/tools/caas


overmyheadandbody (commented):
There are some linting/PSI issues with this PR which need to get addressed before moving forward with this. Otherwise I don't have other concerns

arshadparwaiz (approved):

mokimo (approved):

PR #2638, opened 2024-07-26
MWPW-155314 [MEP] Enable Entitlements + Operator

JingleH:
This PR aims to address this scenario: when an experience column is authored to like `desktop & cc-all-apps-any` in MEP Manifest, right now, unless there's coincidently another column like `cc-all-apps-any`, the variant `desktop & cc-all-apps-any` will never get selected. Reason is that current logic in `getPersonalizationVariant()` for splitting variant name doesn't include `&`, so `hasEntitlementTag` will stay false, leading to no entitlement resolution. The code change tries to split by `&` when building out `variantInfo.allNames`. It is used to check if any entitlement tags are used. `variantInfo[name]` is used for variant mapping and thus unmodified.

This bug might again be trickier to reproduce, due to Milo by default doesn't do entitlements when logged out, and IMS only working on very specific branch links. ~~So on the feature branch, the page is in a deadlock situation, with MEP waiting for entitlements that will never be resolved.~~ (added a catch for IMS error so entitlements will still resolve). But you can create a MEP experience like https://gnav--milo--adobecom.hlx.page/drafts/jinglhua/desktop-entitlements and https://main--milo--adobecom.hlx.page/drafts/jinglhua/desktop-ent.json and then use logs or breakpoints around `variantInfo` and confirm that entitlements not getting resolved currently. To verify my fix, do the same, and you should see that at least it tries to resolve (will be resolved to an empty array due to IMS).

Resolves: https://jira.corp.adobe.com/browse/MWPW-155314

**Test URLs:**
- Before: https://stage--milo--adobecom.hlx.page/drafts/jinglhua/desktop-entitlements
- After: https://mep-entitlements-operator--milo--adobecom.hlx.page/drafts/jinglhua/desktop-entitlements

### DIFF
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -576,20 +576,26 @@ const checkForParamMatch = (paramStr) => {
   return false;
 };
 
+function trimNames(arr) {
+  return arr.map((v) => v.trim()).filter(Boolean);
+}
+export function buildVariantInfo(variantNames) {
+  return variantNames.reduce((acc, name) => {
+    let nameArr = [name];
+    if (!name.startsWith(TARGET_EXP_PREFIX)) nameArr = name.split(',');
+    acc[name] = trimNames(nameArr);
+    acc.allNames = [...acc.allNames, ...trimNames(name.split(/[,&]|\bnot\b/))];
+    return acc;
+  }, { allNames: [] });
+}
+
 async function getPersonalizationVariant(manifestPath, variantNames = [], variantLabel = null) {
   const config = getConfig();
   if (config.mep?.variantOverride?.[manifestPath]) {
     return config.mep.variantOverride[manifestPath];
   }
 
-  const variantInfo = variantNames.reduce((acc, name) => {
-    let nameArr = [name];
-    if (!name.startsWith(TARGET_EXP_PREFIX)) nameArr = name.split(',');
-    const vNames = nameArr.map((v) => v.trim()).filter(Boolean);
-    acc[name] = vNames;
-    acc.allNames = [...acc.allNames, ...vNames];
-    return acc;
-  }, { allNames: [] });
+  const variantInfo = buildVariantInfo(variantNames);
 
   const entitlementKeys = Object.values(await getEntitlementMap());
   const hasEntitlementTag = entitlementKeys.some((tag) => variantInfo.allNames.includes(tag));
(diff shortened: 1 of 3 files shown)


chrischrischris (commented):
If possible add a unit test case

libs/features/personalization/personalization.js
```diff
@@ -571,13 +571,12 @@ async function getPersonalizationVariant(manifestPath, variantNames = [], varian
...
-
+  const trimNames = (arr) => arr.map((v) => v.trim()).filter(Boolean);
   const variantInfo = variantNames.reduce((acc, name) => {
     let nameArr = [name];
     if (!name.startsWith(TARGET_EXP_PREFIX)) nameArr = name.split(',');
-    const vNames = nameArr.map((v) => v.trim()).filter(Boolean);
-    acc[name] = vNames;
-    acc.allNames = [...acc.allNames, ...vNames];
+    acc[name] = trimNames(nameArr);
+    acc.allNames = [...acc.allNames, ...trimNames(name.split(/[,&]/))];
```
chrischrischris, line 579: also need to account for the `not` keyword
see https://milo.adobe.com/docs/authoring/features/personalization-manifests-experiences-tab

JingleH: ahh... good catch sir!

JingleH:
i'll be adding UT on Monday

vhargrave (commented):
looks great to me asides from the points Chris made 👍 .
I'll approve once those are done. Nice one mate!

vgoodric:
Looking good so far to me as well. Will check back after "not" support and unit test updates are ready

libs/features/personalization/personalization.js
JingleH: Hi @chrischrischris , I updated the regex and unit tests. We should be all covered!

JingleH:
Hello friends @chrischrischris  @vivgoodrich @vhargrave, all comments are taken care of and changed lines are 100% covered by unit tests. Let me know if you find anything else!

JingleH:
> und `variantInfo` and confirm that entitlements not getting resolved currently. To verify my fix, do the same, and you should see that at le

done the rebase!

vgoodric:
I think I'm missing something.
are you reversing the value if there is a "not"?
like if they are in "cc-all-app-any" then "not cc-all-app-any" should return false
and I can't tell if you're doing that

mokimo (approved):

vgoodric (approved):

vgoodric:
Discussed the issue with Jingle and am no longer concerned about the handling of the "not" operator.

vhargrave (approved):
very nice!

PR #2639, opened 2024-07-26
MWPW-155316 - Fixing Pill Notification edge cases

elan-tbx:
* When using a Pill Notification without a heading, "undefined" would be injected into the markup.
* Furthermore, when using a hex color background with the "flexible" and "no-close" variants, "null" would be injected into the markup.
* This PR fixes the above issues.

Resolves: [MWPW-155316](https://jira.corp.adobe.com/browse/MWPW-155316)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ebartholomew/pill-testing?martech=off
- After: https://ebartholomew-pill-bugs--milo--adobecom.hlx.page/drafts/ebartholomew/pill-testing?martech=off

### DIFF
--- a/libs/blocks/notification/notification.css
+++ b/libs/blocks/notification/notification.css
@@ -2,6 +2,7 @@
   --min-block-size: 160px;
   --min-block-size-ribbon: 50px;
   --min-block-size-pill: 72px;
+  --margin-inline-pill-desktop: 80px;
   --margin-inline-ribbon: 30px;
   --max-inline-size-image: 75px;
   --max-inline-size-icon: 231px;
@@ -521,6 +522,10 @@
     padding-block: var(--spacing-xs);
     padding-inline: var(--spacing-m);
     margin: 0;
+  }
+
+  .notification.pill:not(.flexible) .foreground.container {
+    inline-size: calc(100% - var(--margin-inline-pill-desktop) * 2);
   }
 
   .notification.pill.flexible .foreground.container {
--- a/libs/blocks/notification/notification.css
+++ b/libs/blocks/notification/notification.css
@@ -364,10 +364,6 @@
     inline-size: auto;
   }
 
-  .notification.ribbon .foreground.container .icon-area {
-    flex-shrink: 1;
-  }
-
   .notification.ribbon .copy-wrap {
     margin-inline-end: var(--spacing-s);
   }
--- a/libs/blocks/notification/notification.css
+++ b/libs/blocks/notification/notification.css
@@ -47,7 +47,8 @@
   font-weight: unset;
 }
 
-.notification [class*="heading-"] + p {
+.notification [class*="heading-"] + p,
+.notification.pill p {
   margin-block-end: var(--spacing-s);
 }
 
@@ -553,6 +554,10 @@
     flex-wrap: wrap;
     align-items: baseline;
     text-align: start;
+  }
+
+  .notification.pill .copy-wrap > * {
+    max-inline-size: 100%;
   }
 
   .notification.ribbon.space-between .foreground.container .icon-area {
--- a/libs/blocks/notification/notification.css
+++ b/libs/blocks/notification/notification.css
@@ -452,6 +452,7 @@
   .notification.max-width-10-desktop .foreground.container {
     margin-inline: var(--grid-margins-width-10);
     gap: var(--spacing-l);
+    max-inline-size: calc(var(--grid-column-width) * 10);
   }
 
   .notification.full-width .foreground.container {
--- a/libs/blocks/notification/notification.css
+++ b/libs/blocks/notification/notification.css
@@ -364,6 +364,10 @@
     inline-size: auto;
   }
 
+  .notification.ribbon .foreground.container .icon-area {
+    flex-shrink: 1;
+  }
+
   .notification.ribbon .copy-wrap {
     margin-inline-end: var(--spacing-s);
   }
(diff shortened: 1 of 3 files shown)


libs/blocks/notification/notification.js
```diff
@@ -90,8 +90,15 @@ function decorateFlexible(el) {
     el.querySelector('.background'),
     el.querySelector('.foreground'),
     el.querySelector('.close'),
-  ];
+  ].reduce((acc, curr) => {
+    if (curr) return [...acc, curr];
+    return acc;
+  }, []);
```
mokimo, line 96: Why do you need the reduce here? Does innards get mutated somehow within `inner` after it's added to the DOM?

libs/blocks/notification/notification.js
```diff
@@ -72,7 +72,7 @@ function getBlockData(el) {
 function wrapCopy(foreground) {
   const text = foreground.querySelector('.text');
   if (!text) return;
-  const heading = text?.querySelector('h1, h2, h3, h4, h5, h6');
+  const heading = text?.querySelector('h1, h2, h3, h4, h5, h6') || text?.querySelector('p:not(.icon-area, .action-area)');
```
overmyheadandbody, line 75: Wouldn't this achieve the same result?
```suggestion
  const heading = text?.querySelector('h1, h2, h3, h4, h5, h6, p:not(.icon-area, .action-area)');
```

libs/blocks/notification/notification.js
```diff
@@ -90,8 +90,15 @@ function decorateFlexible(el) {
...
     el.querySelector('.close'),
-  ];
+  ].reduce((acc, curr) => {
+    if (curr) return [...acc, curr];
+    return acc;
+  }, []);
   const inner = createTag('div', { class: 'flexible-inner' }, innards);
+  if (el.style.background) {
+    inner.style.background = el.style.background;
+    el.style.background = '';
```
overmyheadandbody, line 100: Nit: is the empty value of use anywhere moving forward? Otherwise we could just `delete` it.

libs/blocks/notification/notification.js
overmyheadandbody: I'm also a bit confused on this. Would mapping the options so that any element that doesn't exist gets turned to a simple `''`? Or filter it to only include elements that exist?

chrischrischris: As @overmyheadandbody suggested, filter should work for this.

```suggestion
  ].filter(Boolean);
```

elan-tbx: It's accounting for situations where `el.querySelector('.background')` and/or `el.querySelector('.close')` return null -- otherwise "null" gets used as content in the `createTag` call below

libs/blocks/notification/notification.js
elan-tbx: The `CSSStyleDeclaration` interface doesn't recognize the `delete` keyword, but it does have a `removeProperty` method that achieves the same result -- I'll use that 👍 

https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/removeProperty

elan-tbx:
@mokimo @overmyheadandbody would this be preferable?

```js
const innards = [
    el.querySelector('.background') || '',
    el.querySelector('.foreground'),
    el.querySelector('.close') || '',
  ];
```

mokimo:
@elan-tbx either the empty string as you suggested or simply Bool filter as @chrischrischris neatly demonstrated

libs/blocks/notification/notification.js
elan-tbx: @chrischrischris Sorry -- for some reason my chrome tab of this PR didn't update to include your comment. Will do! 👍

libs/blocks/notification/notification.js
```diff
@@ -90,8 +90,15 @@ function decorateFlexible(el) {
     el.querySelector('.background'),
     el.querySelector('.foreground'),
     el.querySelector('.close'),
-  ];
+  ].reduce((acc, curr) => {
+    if (curr) return [...acc, curr];
+    return acc;
+  }, []).filter(Boolean);
```
overmyheadandbody, line 96: I'm still not clear on this 😅 Do you still need the reduce since you're using the filter?

elan-tbx: Oh man... I need more coffee. Fixing now 😅

overmyheadandbody (approved):
Thanks for the updates!

mokimo (approved):

ryanmparrish (approved):

elan-tbx:
Removing `Ready for Stage` label as a related bug was found, will re-assign to @NadiiaSokolova after next push

libs/blocks/notification/notification.css
```diff
@@ -362,10 +364,6 @@
     inline-size: auto;
   }
 
-  .notification.ribbon .foreground.container .icon-area {
-    flex-shrink: 1;
-  }
```
overmyheadandbody, line 367: Can a notification be both a pill AND a ribbon at the same time? I'm just wondering because the removed style related to the latter, while the PR is trying to address the former and want to make sure we're not removing something that might have value for other use-cases.

elan-tbx: Good point. This came up after another authoring audit, so the hope was to combine that work with this work. I'll keep them separate and work with the reporter to get a ticket made and make a separate PR 👍

Sartxi (approved):

NadiiaSokolova:
Verified, Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-155316

PR #2647, opened 2024-07-30
MWPW-153810 media image is seen loaded twice in product page urls in desktop

suhjainadobe:
Media image is seen loaded twice in product page urls in desktop
This fixes the replacing of whole innerHTML if there is any placeholder on the page.  It only the specific placeholder text.

Resolves: [MWPW-153810](https://jira.corp.adobe.com/browse/MWPW-NUMBER)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/suhjain/performance/marquee-light
- After: https://placeholderreplacement--milo--suhjainadobe.hlx.live/drafts/suhjain/performance/marquee-light


mokimo (commented):
Love the concept as already mentioned on slack 🚀 

Seems like the after link, linter & UTs are still failing:
Screenshot 2024-07-30 at 10 01 20

libs/utils/utils.js
```diff
@@ -724,10 +724,25 @@ async function decorateIcons(area, config) {
...
+  );
+  const nodes = [];
+  let node;
+  while (node = walker.nextNode()) {
+    nodes.push(node);
+  }
+  if (!nodes.length) return;
   const { replaceText } = await import('../features/placeholders.js');
-  el.innerHTML = await replaceText(el.innerHTML, config, regex);
+  for (const textNode of nodes) {
```
vhargrave, line 743: I'd put all these promises in a promise.all and await that, just so this for loop can truly run in parallel.

libs/utils/utils.js
```diff
@@ -724,10 +724,25 @@ async function decorateIcons(area, config) {
...
-  const found = regex.test(el.innerHTML);
-  if (!found) return;
+  const walker = document.createTreeWalker(
+    el,
+    NodeFilter.SHOW_TEXT,
+    {
+      acceptNode: (node) => {
+        regex.test(node.nodeValue) ? NodeFilter.FILTER_ACCEPT : NodeFilter.FILTER_REJECT; 
+      },
+    }
```
overmyheadandbody, line 734: This might also lead to some ESLint complaints, but it's a tiny bit smaller and has a more explicit `return`
```suggestion
    {
      acceptNode: (node) => regex.test(node.nodeValue)
        ? NodeFilter.FILTER_ACCEPT : NodeFilter.FILTER_REJECT,
    },
```

libs/utils/utils.js
```diff
@@ -724,10 +724,25 @@ async function decorateIcons(area, config) {
...
+  );
+  const nodes = [];
+  let node;
+  while (node = walker.nextNode()) {
```
overmyheadandbody, line 738: I'm a bit confused - why are we using `nextNode` and not `currentNode` for the actual node we're interested in?

overmyheadandbody (commented):
Great idea! ESLint has a few comments that need to be addressed on top of what the reviewers suggested so far. Since this is `utils.js`, it might make sense to have another look at places where we can optimize code footprint.

mokimo:
Code looking good, there are still some nala tests failing https://github.com/adobecom/milo/actions/runs/10192039682/job/28194263694?pr=2647

aishwaryamathuria (approved):

libs/utils/utils.js
```diff
@@ -724,10 +724,27 @@ async function decorateIcons(area, config) {
 async function decoratePlaceholders(area, config) {
   const el = area.querySelector('main') || area;
   const regex = /{{(.*?)}}|%7B%7B(.*?)%7D%7D/g;
-  const found = regex.test(el.innerHTML);
-  if (!found) return;
```
overmyheadandbody, line 728: Did you by any chance check whether this is quicker for areas that don't use placeholders compared to defining the tree walker? Some data on that might be interesting. Regular expressions are slow, but one is used anyway in the tree walker, so I'm curious to compare areas with and without placeholders from a performance perspective with this approach.

mokimo (approved):

libs/utils/utils.js
suhjainadobe: > Did you by any chance check whether this is quicker for areas that don't use placeholders compared to defining the tree walker? Some data on that might be interesting. Regular expressions are slow, but one is used anyway in the tree walker, so I'm curious to compare areas with and without placeholders from a performance perspective with this approach.

Ran tests on https://placeholderreplacement--milo--suhjainadobe.hlx.live/drafts/suhjain/performance/marquee-light
Treewalker function took : 0.19999998807907104 milliseconds
async function decoratePlaceholders(area, config) { const el = area.querySelector('main') || area; const regex = /{{(.*?)}}|%7B%7B(.*?)%7D%7D/g; const walker = document.createTreeWalker( el, NodeFilter.SHOW_TEXT, { acceptNode(node) { return regex.test(node.nodeValue) ? NodeFilter.FILTER_ACCEPT : NodeFilter.FILTER_REJECT; }, }, ); const nodes = []; // Start timing const startTime = performance.now(); let node = walker.nextNode(); while (node !== null) { nodes.push(node); node = walker.nextNode(); } // End timing const endTime = performance.now(); console.log(TreeWalker took ${endTime - startTime} milliseconds to find the nodes); if (!nodes.length) return; const { replaceText } = await import('../features/placeholders.js'); const replaceNodes = nodes.map(async (textNode) => { textNode.nodeValue = await replaceText(textNode.nodeValue, config, regex); }); await Promise.all(replaceNodes); }

Regrex function took: 0.30000001192092896 milliseconds
`async function decoratePlaceholders(area, config) {
const el = area.querySelector('main') || area;
const regex = /{{(.?)}}|%7B%7B(.?)%7D%7D/g;

// Start timing
const startTime = performance.now();
const found = regex.test(el.innerHTML);
// End timing
const endTime = performance.now();

console.log(Regex test took ${endTime - startTime} milliseconds);

if (!found) return;

const { replaceText } = await import('../features/placeholders.js');
el.innerHTML = await replaceText(el.innerHTML, config, regex);
}`

@overmyheadandbody : Please let me know if this is the data that you wanted to know?

suhjainadobe: > Did you by any chance check whether this is quicker for areas that don't use placeholders compared to defining the tree walker? Some data on that might be interesting. Regular expressions are slow, but one is used anyway in the tree walker, so I'm curious to compare areas with and without placeholders from a performance perspective with this approach.

test

spadmasa:
Validation done on the pr , details are updated in 
https://jira.corp.adobe.com/browse/MWPW-153810
https://main--cc--adobecom.hlx.page/products/premiere?milolibs=placeholderreplacement--milo--suhjainadobe&mep=off
image

PR #2664, opened 2024-07-31
Query Parms update & sendTargetResponse function update

theankit-sinha:
Query Parms update & sendTargetResponse function update

- sendTargetResponse: Added Check for window._satellite and changed direct alloy call to _satellite call 
- Added additional parameter as marketingtech to disable analytics calls

Resolves: [ENB-6995](https://jira.corp.adobe.com/browse/ENB-6995)

**Test URLs:**

**Before**: https://www.adobe.com/products/photoshop.html?marketingtech=off
**After**: https://www.adobe.com/products/photoshop.html?marketingtech=off

**Dummy Test url**: https://marketingtech--milo--theankit-sinha.hlx.page?marketingtech=off

**Note**
Currently when passing  marketingtech=off page is not loading, Expected behavior is martech bootstrap should not load and page should load as it is  

**More Information**
There are 2 changes

1. In Martech.js which currently resides in [milo/libs/martech/](https://github.com/adobecom/milo/blob/main/libs/martech/martech.js) folder, we noticed there is a function named by sendTargetResponseAnalytics which has sendEvent call using window.alloy now for the requirement to disable analytics tracking we have added certain conditions in launch due to which we are able to control data collection for most of the calls, but for this function as it is directly calling window.alloy data is being sent, we would like to modify this function.

2. In Utils.js which resides in [milo/libs/utils/](https://github.com/adobecom/milo/blob/main/libs/utils/utils.js) folder, there is another function named by loadMartech where there are certain condition to check if  "martech=off" exists in query params or not if it exists we do not load martech.js library, we would like to add another optional parameter as "marketingtech=off" , as our main bootstrap script has this as query prams to disable analytics library completely, currently when we pass "marketingtech=off" in any adotcom pages, page seems to crash / fail to load.


theankit-sinha:
Hey @vivgoodrich & @markpadbe 
Please find the attached PR for the request and refer to the description added

libs/martech/martech.js
```diff
@@ -100,20 +100,22 @@ function sendTargetResponseAnalytics(failure, responseStart, timeout, message) {
...
-    documentUnloading: true,
-    xdm: {
-      eventType: 'web.webinteraction.linkClicks',
-      web: {
-        webInteraction: {
-          linkClicks: { value: 1 },
-          type: 'other',
-          name: val,
+  if(window._satellite && window._satellite.track){
+    window._satellite.track('event', {
```
overmyheadandbody, line 104:
```suggestion
  window._satellite?.track?.('event', {
```

overmyheadandbody (commented):
Adding a few more people to have a look, as moving back to direct `_satellite` calls seems a bit unintuitive to me. CC: @chrischrischris, @bbythewa

libs/martech/martech.js
theankit-sinha: @overmyheadandbody -makes sense, will update the request Thanks :)

vgoodric (approved):

vgoodric:
@overmyheadandbody, @sharg1 I have recreated this PR with a branch I made to help get a QA link. @theankit-sinha was having trouble finishing the setup.

New PR: https://github.com/adobecom/milo/pull/2682

PR #2535, opened 2024-07-02
MWPW-147172: Support for callout texts in merch cards

rohitsahu:
Added support to add callout text which appear as grey background. To add any text as a "callout text", it needs to be added inside h6 and em tags.
AC: Optional feature to callelation copy within our merch cards
- Ability to include copy under the price
- Ability to include a "learn more" or "see terms" link (if needed) within the copy
- Ability to include a tool tip 
- Ability to include copy on rollover/click on mobile for the tooltip
- The Word styling solution should be the same across all merch cards 
Screenshot 2024-07-08 at 8 14 51 PM

Resolves: [MWPW-147172](https://jira.corp.adobe.com/browse/MWPW-147172)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/Priyanka/test-mwpw-147172-doc2
- After: https://mwpw-147172--milo--rohitsahu.hlx.page/drafts/Priyanka/test-mwpw-147172-doc2

PR #2663, opened 2024-07-31
MWPW-154191: Ability to have green promo text on merch-card to bottom of the card (under description)

mirafedas:
Added a promo-text slot in the listed merch card variants:
- merch-card (product)
- merch-card (mini-compare)
- merch-card (segment)
- merch-card (catalog)
- merch-card (plans)
- merch-card (image)

Resolves: [MWPW-154191](https://jira.corp.adobe.com/browse/MWPW-154191)

**Test URLs:**
Before: 
- Product: https://main--milo--adobecom.hlx.live/fragments/mirafedas/merch/merch-product?martech=off
- Mini-compare: https://main--milo--adobecom.hlx.live/fragments/mirafedas/merch/merch-mini-comparison?martech=off
- Segment: https://main--milo--adobecom.hlx.live/fragments/mirafedas/merch/merch-segment?martech=off
- Catalog: https://main--milo--adobecom.hlx.live/fragments/mirafedas/merch/merch-catalog?martech=off
- Plans: https://main--milo--adobecom.hlx.live/fragments/mirafedas/merch/merch-plans?martech=off
- Image: https://main--milo--adobecom.hlx.live/fragments/mirafedas/merch/merch-image?martech=off
After: 
- Product: https://mwpw-154191-promo-text-bottom--milo--mirafedas.hlx.live/fragments/mirafedas/merch/merch-product?martech=off
- Mini-compare: https://mwpw-154191-promo-text-bottom--milo--mirafedas.hlx.live/fragments/mirafedas/merch/merch-mini-comparison?martech=off
- Segment: https://mwpw-154191-promo-text-bottom--milo--mirafedas.hlx.live/fragments/mirafedas/merch/merch-segment?martech=off
- Catalog: https://mwpw-154191-promo-text-bottom--milo--mirafedas.hlx.live/fragments/mirafedas/merch/merch-catalog?martech=off
- Plans: https://mwpw-154191-promo-text-bottom--milo--mirafedas.hlx.live/fragments/mirafedas/merch/merch-plans?martech=off
- Image: https://mwpw-154191-promo-text-bottom--milo--mirafedas.hlx.live/fragments/mirafedas/merch/merch-image?martech=off

PR #2668, opened 2024-07-31
MWPW-152792: Adjust CSS in Catalog card for promo treatment

mirafedas:
- Added a slot 'payment-details' located inside the 'heading-m' slot;
- Restricted the catalog badge max width to 140px;

Changes related to the callout text were added in [this PR](https://github.com/adobecom/milo/pull/2535), I only moved them into a separate function here. Adding this info here as looks like it was not documented anywhere: 
to add a tooltip for the callout text, place the link to the tooltip icon with the text '#icon' and add the tooltip text in the link after the hash, example: `https://main--milo --adobecom.hlx.page/drafts/rosahu/info-icon.svg#this%20is%20a%20dummy%20tooltip%20text`

Updated authoring documentation: https://main--  milo--mirafedas.hlx.live/docs/authoring/commerce/merch-card-catalog

Resolves: [MWPW-152792](https://jira.corp.adobe.com/browse/MWPW-152792)

**Test URLs:**
- Before: https://main--cc --adobecom.hlx.page/es/cc-shared/fragments/merch/products/catalog/merch-card/cc/uat/promo-test/ai-promo-test?martech=off&georouting=off
- After: https://mwpw-152792-badge-vat--milo--mirafedas.hlx.page/drafts/mirafedas/do-not-modify/catalog-card-vat-label?martech=off&georouting=off


libs/blocks/merch-card/merch-card.js
```diff
@@ -210,36 +210,43 @@ const parseContent = async (el, merchCard) => {
...
-          });
-        } else {
-          const clone = child.cloneNode(true);
-          fragment.appendChild(clone);
-        }
-      });
+    if (tagName === 'H6') {
+      if (element.firstChild.nodeType === Node.TEXT_NODE) {
+        const paymentDetails = createTag('div', { slot: 'payment-details' });
+        paymentDetails.innerHTML = element.innerHTML;
```
overmyheadandbody, line 216: This might trigger some Kodiak issues, maybe a spread or recursive append would be better suited. This could also be troublesome if some event listeners have already been added to `element`. Also, could `element` be used after this piece of logic? If not, maybe it could be deleted.

3ch023 (changes requested):
image

it seems to still overlap with the drop down

libs/blocks/merch-card/merch-card.js
Axelcureno, line 216:
```suggestion
        const paymentDetails = createTag('div', { slot: 'payment-details' }, element.innerHTML);
```

libs/blocks/merch-card/merch-card.js
```diff
@@ -210,36 +210,43 @@ const parseContent = async (el, merchCard) => {
...
+        const headingM = merchCard.querySelector('h4[slot="heading-m"]');
+        headingM?.append(paymentDetails);
+      } else if (element.firstElementChild?.tagName === 'EM') {
+        const calloutContentWrapper = createTag('div');
```
Axelcureno, line 220: This `parseContent()` function is getting bulky and convoluted, why not split some logic into separate functions for easier reading and modularization, we can start by moving the callOut logic into its separate function.

Axelcureno (approved):
Approving with 2 small suggestions which are not blockers.

libs/features/mas/web-components/src/merch-card.css.js
```diff
@@ -140,6 +140,7 @@ export const styles = css`
         height: fit-content;
         flex-direction: column;
         width: fit-content;
+        max-width: 140px;
```
Axelcureno, line 143: Try adding `box-sizing: border-box;` to enforce `max-width`, otherwise it will still overlap.
Screenshot 2024-08-05 at 11 02 47 AM

Axelcureno (changes requested):
As @3ch023 stated, content is still overlapping. I have made one suggestion to fix it.

libs/blocks/merch-card/merch-card.js
```diff
@@ -248,12 +266,6 @@ const parseContent = async (el, merchCard) => {
     if (mnemonicList) bodySlot.append(mnemonicList);
   });
 
-  if (calloutContainer.children.length > 0) {
-    const calloutSlot = createTag('div', { slot: 'callout-content' });
-    calloutSlot.appendChild(calloutContainer);
-    merchCard.appendChild(calloutSlot);
```
rahulgupta999, line 254: @rohitsahu does the order of adding the callout slot matter

Axelcureno:
Conflicts must be resolved please

Axelcureno (approved):

libs/blocks/merch-card/merch-card.js
```diff
@@ -154,6 +154,52 @@ const parseTwpContent = async (el, merchCard) => {
   }
 };
 
+const appendPaymentDetails = (element, merchCard) => {
+  if (element.firstChild.nodeType !== Node.TEXT_NODE) return;
+  const paymentDetails = createTag('div', { slot: 'payment-details' }, element.innerHTML);
```
yesil, line 159: can you use `element` instead of `element.innerHTML` ?

libs/blocks/merch-card/merch-card.js
```diff
@@ -154,6 +154,52 @@ const parseTwpContent = async (el, merchCard) => {
...
+  if (element.firstElementChild?.tagName !== 'EM') return;
+  const calloutContainer = createTag('div');
+  const calloutContentWrapper = createTag('div');
+  const calloutContent = createTag('div');
+  const emElement = element.firstElementChild;
+  let imgElement = null;
+  const fragment = document.createDocumentFragment();
+
+  emElement.childNodes.forEach((child) => {
+    if (child.nodeType === Node.ELEMENT_NODE && child.tagName === 'A' && child.innerText.trim().toLowerCase() === '#icon') {
```
yesil, line 174: Can you describe what does ` '#icon'` in the PR description?

libs/features/mas/web-components/src/merch-card-collection.js
```diff
@@ -283,7 +283,7 @@ export class MerchCardCollection extends LitElement {
             <sp-action-menu
                 id="sortButton"
                 size="m"
-                @change="${this.sortChanged}"
+                onchange="${this.sortChanged}"
```
yesil, line 286: why this change?  this seems wrong to me.

yesil (changes requested):

libs/blocks/merch-card/merch-card.js
mirafedas: I'm passing `element.innerHTML` because I want to strip off the `<h6>` tag - semantically this text is not a heading, and I think there is no need to wrap this text in an additional tag. Is there any reason why we should change it to `element`?

libs/features/mas/web-components/src/merch-card-collection.js
mirafedas: This change got there by mistake, thanks for noticing!

libs/blocks/merch-card/merch-card.js
yesil: if you know that `element` is never an element with an even listener attached, that should be ok, if not, this will cause the listeners to be removed from the element.

libs/deps/mas/merch-card-collection.js
```diff
@@ -128,7 +128,7 @@ import{html as a,LitElement as A}from"/libs/deps/lit-all.min.js";var m=class{con
             <sp-action-menu
                 id="sortButton"
                 size="m"
-                @change="${this.sortChanged}"
+                onchange="${this.sortChanged}"
```
yesil, line 131: I thought I already asked about this change? what is the reason?

mirafedas: Yes, sorry, forgot to build mas. It should be good now

mirafedas:
@yesil According to designs `paymentDetails` is just a text and it's doubtful there will be any event listeners on it, so for the sake of DOM simplicity I will leave it as it is, but thanks for the heads-up :)

libs/features/mas/web-components/src/merch-card.js
```diff
@@ -432,7 +432,9 @@ export class MerchCard extends LitElement {
                     >${this.actionMenuContent}</slot
                 >
                 <slot name="heading-xs"></slot>
-                <slot name="heading-m"></slot>
+                <slot name="heading-m">
+                    <slot name="payment-details"></slot>
```
yesil, line 436: A `slot` inside another `slot` doesn't seem to be allowed as per chatgpt but the specs do not mention it.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot

technically this should be equal to:


```html
<slot name="payment-details"></slot>
<slot name="heading-m"></slot>
...
</slot>
``` 

what is the reason for nesting?

mirafedas: @yesil 
The merch card body has a flex gap between all elements (see screenshot).
If I place the `payment-details` slot outside the `heading-m` slot, there will be a gap between the two elements and it will not look as on the XD.

Screenshot 2024-08-13 at 10 52 54

libs/blocks/merch-card/merch-card.js
mirafedas: Thanks guys for bringing my attention to this, I rewrote the code, not all works as it should

libs/features/mas/web-components/src/merch-card.js
yesil: @mirafedas in this case, why it can't be a part of `heading-m` ?

mirafedas: @yesil we need a way to distinguish which text goes into `heading-m` and which into `payment-details` because both texts should be on different lines and have different styles

yesil: @mirafedas we need to find a way to do it without nested slots though.

mirafedas: I replaced the slot with a div, now should be good 👍

.github/codeql/codeql-config.yml
```diff
@@ -5,3 +5,4 @@ paths-ignore:
   - 'tools/floodgate/**'
   - 'tools/translation/**'
   - node_modules
+  - libs/deps/mas
```
3ch023, line 8: cool

3ch023 (approved):

yesil (approved):

Axelcureno (approved):

Roycethan:
@mirafedas Can you plz check these:
1) VAT label style is not yet fixed.having larger and non-italized font... 
per this ticket - "The VAT label style is also the same size as the price, and needs a smaller italicized text style"
https://main--cc--adobecom.hlx.page/es/cc-shared/fragments/merch/products/catalog/merch-card/cc/uat/promo-test/ai-promo-test?milolibs=mwpw-152792-badge-vat--milo--mirafedas
2) Per merged PR https://github.com/adobecom/milo/pull/2663 the call out text is moved to the bottom, but in the below example its not:
https://mwpw-152792-badge-vat--milo--mirafedas.hlx.page/drafts/mirafedas/do-not-modify/catalog-card-vat-label?martech=off&georouting=off

mirafedas:
@Roycethan 
1. I fixed it by changing the `{{small-tax-incl-label}}` to H6.
2. To place the promo on the bottom, authors must set the `promo bottom` parameter on the card. I added a second card with this parameter, now we have both cases covered.

PR #2682, opened 2024-08-01
ENB-6995 Query Params update & sendTargetResponse function update Resubmit for #2664

vgoodric:
- sendTargetResponse: Added Check for window._satellite and changed direct alloy call to _satellite call
- Added additional parameter as marketingtech to disable martech.  The martech=off also impacts launch.  So they have requested a new parameter that does the same thing in Milo as martech=off but does not impact their code.

Resolves: [ENB-6995](https://jira.corp.adobe.com/browse/ENB-6995)

Test URLs: https://www.adobe.com/products/photoshop.html?marketingtech=off

Psi-check: [https://martechteamupdates--milo--adobecom.hlx.page?marketingtech=off](https://martechteamupdates--milo--adobecom.hlx.page?marketingtech=off)

Note
Currently when passing marketingtech=off page is not loading, Expected behavior is martech bootstrap should not load and page should load as it is

More Information
There are 2 changes

In Martech.js which currently resides in [milo/libs/martech/](https://github.com/adobecom/milo/blob/main/libs/martech/martech.js) folder, we noticed there is a function named by sendTargetResponseAnalytics which has sendEvent call using window.alloy now for the requirement to disable analytics tracking we have added certain conditions in launch due to which we are able to control data collection for most of the calls, but for this function as it is directly calling window.alloy data is being sent, we would like to modify this function.

In Utils.js which resides in [milo/libs/utils/](https://github.com/adobecom/milo/blob/main/libs/utils/utils.js) folder, there is another function named by loadMartech where there are certain condition to check if "martech=off" exists in query params or not if it exists we do not load martech.js library, we would like to add another optional parameter as "marketingtech=off" , as our main bootstrap script has this as query prams to disable analytics library completely, currently when we pass "marketingtech=off" in any adotcom pages, page seems to crash / fail to load.

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -865,8 +865,9 @@ export async function loadMartech({
     return true;
   }
 
-  const query = PAGE_URL.searchParams.get('martech');
-  if (query === 'off' || getMetadata('martech') === 'off') {
+  const martechQuery = PAGE_URL.searchParams.get('martech');
+  const marketingQuery = PAGE_URL.searchParams.get('marketingtech');
+  if (martechQuery === 'off' || marketingQuery === 'off' || getMetadata('martech') === 'off') {
     return false;
   }
 
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -865,9 +865,9 @@ export async function loadMartech({
     return true;
   }
 
-  const martechQuery = PAGE_URL.searchParams.get('martech');
-  const marketingQuery = PAGE_URL.searchParams.get('marketingtech');
-  if (martechQuery === 'off' || marketingQuery === 'off' || getMetadata('martech') === 'off') {
+  if (PAGE_URL.searchParams.get('martech') === 'off'
+    || PAGE_URL.searchParams.get('marketingtech') === 'off'
+    || getMetadata('martech') === 'off') {
     return false;
   }
 
--- a/libs/martech/martech.js
+++ b/libs/martech/martech.js
@@ -89,20 +89,21 @@ function sendTargetResponseAnalytics(failure, responseStart, timeout, message) {
   const timeoutTime = roundToQuarter(timeout);
   let val = `target response time ${responseTime}:timed out ${failure}:timeout ${timeoutTime}`;
   if (message) val += `:${message}`;
-    window._satellite?.track?.('event', {
-      documentUnloading: true,
-      xdm: {
-        eventType: 'web.webinteraction.linkClicks',
-        web: {
-          webInteraction: {
-            linkClicks: { value: 1 },
-            type: 'other',
-            name: val,
-          },
+  // eslint-disable-next-line no-underscore-dangle
+  window._satellite?.track?.('event', {
+    documentUnloading: true,
+    xdm: {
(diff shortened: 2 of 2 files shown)


markpadbe (approved):

denlight (approved):

libs/utils/utils.js
```diff
@@ -865,8 +865,9 @@ export async function loadMartech({
     return true;
   }
 
-  const query = PAGE_URL.searchParams.get('martech');
-  if (query === 'off' || getMetadata('martech') === 'off') {
+  const martechQuery = PAGE_URL.searchParams.get('martech');
+  const marketingQuery = PAGE_URL.searchParams.get('marketingtech');
+  if (martechQuery === 'off' || marketingQuery === 'off' || getMetadata('martech') === 'off') {
```
robert-bogos, line 870: Nit: you can simplify this by not caching
```suggestion
  if (PAGE_URL.searchParams.get('martech') === 'off' 
  || PAGE_URL.searchParams.get('marketingtech') === 'off' 
  || getMetadata('martech') === 'off') {
```

robert-bogos (approved):

libs/utils/utils.js
vgoodric: sounds good.  added your suggestion

overmyheadandbody (commented):
It still feels a bit weird to me that we're moving back to direct `_satellite` calls after we spent time porting a lot of things to `alloy` ones. But I'm no expert on this, so I'd just like to understand the reasoning behind it for future reference.

Who is the audience for the new query parameter? `martech` and `marketingtech` likely sound awfully alike to most people, I fear this might cause some confusion. I don't have a better naming suggestion right now, but I think this deserves a bit more thought before we commit to a new query parameter that can be used by a non-technical audience.

theankit-sinha:
> It still feels a bit weird to me that we're moving back to direct `_satellite` calls after we spent time porting a lot of things to `alloy` ones. But I'm no expert on this, so I'd just like to understand the reasoning behind it for future reference.
> 
> Who is the audience for the new query parameter? `martech` and `marketingtech` likely sound awfully alike to most people, I fear this might cause some confusion. I don't have a better naming suggestion right now, but I think this deserves a bit more thought before we commit to a new query parameter that can be used by a non-technical audience.

Hey, @overmyheadandbody 
Let me just give some context - why both of these changes are requested
- For  _satellite calls - we have received a new request where requester wants to do automated testing and when doing so it should not collect any data but martech and other analytical libraries and event firring should work as it is. so we have configured a dummy datatream, when this flag is passed it will just switch and send all data to dummy datastream instead of production datastream, this change is controlled via launch setup now in this code it is a direct alloy event call which we cannot control, for more information regarding this please head to this ticket [https://jira.corp.adobe.com/browse/ENB-6995]
- I understand this - Further exploring we saw that in our bootstrap code this is already implemented as marketingtech=off and its being used, to make it consistent we are adding another query parms, currently if you pass marketingtech=off in any page it seems to crash / not load

narcis-radu (approved):

mokimo:
I'll merge this despite overlaps as there are a bunch of other files changing utils.js and it should be safe to include.

PR #2683, opened 2024-08-02
[MWPW-155566] [MEP] Allow relevant actions in MEP to use text and HTML instead of just fragments

denlight:
- Expanded the functionality of the several actions to allow text or HTML instead of just fragments: (replace, insertbefore, insertafter, prependToSection and appendToSection) It is intended to update innerText of the targeted element.(ex: update a column/row combination text to change the index of a selected tab)
- now that prependToSection and appendToSection can text and HTML, also lifted the restriction they be used on sections and are renamed append and prepend
- Added new "helper" selectors: **primary-cta**, **secondary-cta**, **action-area** with an ability to pair those with the *#_href* flag to update the *href* attributes where applicable.
- Added ability to combine simplified and "pseudo"(simplified) selectors.

Resolves: 
[MWPW-155566](https://jira.corp.adobe.com/browse/MWPW-155566)

Also resolves:
[MWPW-148285 [MILO][MEP] Add simplified selector to select the paragraph around a block's links](https://jira.corp.adobe.com/browse/MWPW-148285) - one of the best use cases for the update action, so was added to this PR
[MWPW-148284 [MILO][MEP] Allow selectors to have a mix of CSS and simplified selectors](https://jira.corp.adobe.com/browse/MWPW-148284) - update for 148285 also ended up solving this

**URL for testing:**
- Before: https://main--cc--adobecom.hlx.live/drafts/dfedotov/fragments/test-no-update?martech=off
- After: https://main--cc--adobecom.hlx.live/drafts/dfedotov/fragments/test-no-update?martech=off&milolibs=MWPW-148284

Psi-check: https://mwpw-148284--milo--adobecom.hlx.page/?martech=off 

Recent updates to this PR:
updated the getSelectedElement function to have a different approach:
1. no longer adding a new action called update but instead allowing most actions to accept html or text
2. because we allow html and text for most actions, changed appendToSection and prependToSection to just be called append and prepend (retro support for old names)
3.updated the getSelectedElement function to have a different approach:
- split and join the selector to add spaces around > and , for consistency
- split the selector by whitespace
- loop through each "term" in the selector to figure out if it starts with a simplified selector and/or ends in a child number

benefits to this approach:
1. can recognize consumer blocks
2. can enforce sections are direct children of main but still handle if main is in the selector from manifest
3. can add number when there's a dot. Example .text4


libs/features/personalization/personalization.js
```diff
@@ -131,20 +131,45 @@ const CREATE_CMDS = {
   appendtosection: 'beforeend',
 };
 
+const COMMANDS_KEYS = {
```
vgoodric, line 134: I don't feel this array is necessary, since it's a 1 to 1 with the keys in COMMANDS.

libs/features/personalization/personalization.js
```diff
@@ -131,20 +131,45 @@ const CREATE_CMDS = {
...
+  remove: 'remove',
+  replace: 'replace',
+  update: 'update',
+};
+
+const ATTRIBUTES_TO_UPDATE = {
```
vgoodric, line 140: Can we rename this updateFlags?  Also think it makes sense to move this inside of the update command declaration in 20 lines to help make the scope of where this is used clear.  The use of all caps and placing it at the top is more for constants to you use in multiple functions.

libs/features/personalization/personalization.js
```diff
@@ -427,12 +454,11 @@ export function handleCommands(commands, rootEl = document, forceInline = false)
       registerInBlockActions(cmd, manifestId, targetManifestId);
       return;
     }
-
+    const el = getSelectedElement({ selector, rootEl });
     if (action in COMMANDS) {
-      const el = getSelectedElement(selector, action, rootEl);
-      COMMANDS[action](el, target, manifestId, targetManifestId);
+      // eslint-disable-next-line max-len
```
vgoodric, line 459: instead of disabling max length, I believe the preferred method is to put the elements of the object you are passing on their own lines.

libs/features/personalization/personalization.js
```diff
@@ -427,12 +454,11 @@ export function handleCommands(commands, rootEl = document, forceInline = false)
       registerInBlockActions(cmd, manifestId, targetManifestId);
       return;
     }
-
+    const el = getSelectedElement({ selector, rootEl });
```
vgoodric, line 457: Good call on pulling this out beforehand.  Makes me realize we should also do a return command.
```suggestion
    const el = getSelectedElement({ selector, rootEl });
    if (!el || (!(action in COMMANDS) && !(action in CREATE_CMDS)) return;
```
if we do that we can get rid of the else clause at the end.
could even do a return inside the first if clase and remove the else if wrap around the other commands

narcis-radu (changes requested):
Hi @denlight - we usually aim for one PR per JIRA ticket, this is really helpful for managing code that gets merged. I see you've mentioned four JIRA tickets with different  requirements, it will be harder to test and also harder to revert in case things go crazy.
I would ask you to raise separate PRs for the features. Thanks!

vgoodric:
@narcis-radu completely agree.  This was kind of a special situation.  His real ticket was to add the update action.  This required updating the simplified selectors as well to achieve this.  When I looked through our backlog, I realized he had inadvertently also solved a couple of other tickets.

libs/features/personalization/personalization.js
```diff
@@ -329,70 +332,84 @@ function registerInBlockActions(cmd, manifestId, targetManifestId) {
...
 }
 
-function getSelectedElement(selector, action, rootEl) {
+function getSelectedElement({ selector, rootEl }) {
   if (!selector) return null;
-  if ((action.includes('appendtosection') || action.includes('prependtosection'))) {
-    if (!selector.includes('section')) return null;
-    const section = selector.trim().replace('section', '');
-    if (section !== '' && Number.isNaN(section)) return null;
-  }
```
vgoodric, line 338: I don't think we can get rid of this completely.  This is trying to enforce that they don't prepend or append to anything but a section.  But I think we can improve this.  See below for suggestion.

libs/features/personalization/personalization.js
```diff
@@ -329,70 +332,84 @@ function registerInBlockActions(cmd, manifestId, targetManifestId) {
...
-  if (terms.length) {
-    // find targeted table cell in rowX colY format
-    const rowColMatch = /row(?<row>\d+)\s+col(?<col>\d+)/gm.exec(terms.join(' '));
-    if (rowColMatch) {
-      const { row, col } = rowColMatch.groups;
-      const tableCell = querySelector(selectedEl, `:nth-child(${row}) > :nth-child(${col})`);
-      if (!tableCell) return null;
-      return getDivInTargetedCell(tableCell);
-    }
+    return element;
```
vgoodric, line 410: This assumes you send action into the function again.
```suggestion
    if (action.includes('pendtosection') && element?.parentNode?.nodeName) return;
    return element;
```

denlight: not sure what this one means . If I add this code the tests for append/prepend section breaks, because `getSelectedElement` will just return nothing

libs/features/personalization/personalization.js
denlight: i missed this one. Good catch. Adding it back

libs/features/personalization/personalization.js
vgoodric: After discussing it, I see the value more.  No longer concerned.

narcis-radu (approved):

libs/features/personalization/personalization.js
```diff
@@ -327,70 +324,70 @@ function registerInBlockActions(cmd, manifestId, targetManifestId) {
...
+      const regex = new RegExp(`(\\s|^)(${block})\\.?(\\d+)?(\\s|$)`, 'g');
+      const match = regex.exec(modifiedSelector);
+      if (match?.length) {
+        const simplifiedSelector = match[0].replace(/\s+/g, '');
+        const n = simplifiedSelector.match(/\d+/g) || '1';
+        const cleanClassSelector = match[2];
+        const cssOptimizedSelector = ` .${cleanClassSelector}:nth-child(${n} of .${cleanClassSelector})`;
+        modifiedSelector = modifiedSelector.replace(simplifiedSelector, cssOptimizedSelector);
+      }
+    });
```
chrischrischris, line 350: Instead of checking for every milo block, we should assume that any selector that starts with an alpha character, and is not an html element, can be assumed to be a block and  have a `.` prepended.  

Something like this (not tested):
```
const areaSelectors = ['section', 'row', 'col'];
const actionSelectors = {
	  'primary-cta': 'p strong a',
	  'secondary-cta': 'p em a',
	  'action-area': 'p:has(em a, strong a)'
	};
const htmlEls = ['div', 'p', 'strong', 'em', 'h1', 'h2', ...];
const notBlocks = [...areaSelectors, ...Object.keys(actionSelectors), ...htmlEls];
const terms = modifiedSelector.split(/\s+/);
modifiedSelector = terms.reduce((selStr, term) => {
  const { blockName, blockIndex } = getBlockName(term);
  if (notBlocks.includes(blockName)) {
    return `${selStr} ${term}`;
  }
  return `${selStr} .${blockName}:nth-child(${blockIndex} of .${blockName})`; 
}, '');
```

chrischrischris: NOTE: Looks like using `.block:nth-child(4 of .block)` does not work in all circumstances.   On the example page, try using `.text4` and it will not find the element

chrischrischris (changes requested):

libs/features/personalization/personalization.js
```diff
@@ -327,70 +344,73 @@ function registerInBlockActions(cmd, manifestId, targetManifestId) {
...
+        terms[i] = '';
+        if (next === '>') terms[i + 1] = '';
+      }
+    } else {
+      terms[i] = modifySelectorTerm(term);
+    }
+  });
+  modifiedSelector = terms.join(' ');
+  return modifiedSelector;
+}
```
chrischrischris, line 393: Seems like this is to remove `main sectionX` or `main > sectionX`?  This might be a case where regexp is better?

```suggestion
export function modifyNonFragmentSelector(selector) {
  return selector
    .split('>').join(' > ')
    .split(',').join(' , ')
    .replace(/^main( > section\d*| section\d*)/, '')
    .split(/\s+/)
    .map(modifySelectorTerm)
    .join(' ');
}
```

libs/features/personalization/personalization.js
```diff
@@ -327,70 +344,73 @@ function registerInBlockActions(cmd, manifestId, targetManifestId) {
...
+
+function getSelectedElement({ selector, rootEl }) {
+  let modifiedSelector = selector.trim();
```
chrischrischris, line 396:
```suggestion
function getSelectedElement({ selector: sel, rootEl }) {
  let selector = sel.trim();
```

chrischrischris (commented):
Awesome functionality improvements.  Only thing is refactoring some of the code to be a bit cleaner.

chrischrischris (commented):
Awesome functionality improvements.  Only thing is refactoring some of the code to be a bit cleaner.

libs/features/personalization/personalization.js
```diff
@@ -327,70 +324,70 @@ function registerInBlockActions(cmd, manifestId, targetManifestId) {
...
+          case simplifiedSelector.includes('secondary-cta'):
+            modifiedSelector = modifiedSelector.replace(simplifiedSelector, 'p em a');
+            break;
+          case simplifiedSelector.includes('action-area'):
+            modifiedSelector = modifiedSelector.replace(simplifiedSelector, 'p:has(em a, strong a)');
+            break;
+          default: break;
+        }
+      });
+    });
```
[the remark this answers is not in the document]
vgoodric: spoke on Slack, this is outdated

libs/features/personalization/personalization.js
vgoodric: spoke on Slack, this is outdated

vgoodric: spoke on Slack, this is outdated

libs/features/personalization/personalization.js
```diff
@@ -93,64 +93,101 @@ const isInLcpSection = (el) => {
...
+  if (targetManifestId) el.dataset.adobeTargetTestid = targetManifestId;
+}
+
+function checkSelectorType(selector) {
+  const sel = selector.toLowerCase().trim();
+  if (sel.includes('/fragments/') && (sel.startsWith('/') || sel.startsWith('http'))) return 'fragment';
+  return 'other';
+}
+
+export const createContent = (el, content, manifestId, targetManifestId, action, modifiers) => {
```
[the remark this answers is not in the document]
vgoodric: I like the direction of this, but decided to move more of the code out and call it createFrag. Also with the code in updateHref now in a function, changed it use return

libs/features/personalization/personalization.js
```diff
@@ -327,70 +344,73 @@ function registerInBlockActions(cmd, manifestId, targetManifestId) {
...
 
-function getSelectedElement(selector, action, rootEl) {
-  if (!selector) return null;
-  if ((action.includes('appendtosection') || action.includes('prependtosection'))) {
-    if (!selector.includes('section')) return null;
-    const section = selector.trim().replace('section', '');
-    if (section !== '' && Number.isNaN(section)) return null;
-  }
-  if (checkSelectorType(selector) === 'fragment') {
+function modifySelectorTerm(term) {
```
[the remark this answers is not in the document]
vgoodric: these changes look good to me.  adopted

libs/features/personalization/personalization.js
vgoodric: yeah, this is better

libs/features/personalization/personalization.js
vgoodric: adopted

markpadbe (approved):

PR #2684, opened 2024-08-02
MWPW-151456: Spectrum 2 Token Integration Demo

saashapina:
- Creates Spectrum 2 Integration Demo Page. This work includes token integration and related style adjustments for:
    - `Marquee` 
    - `Accordion`
    - `Aside`
       - adds `mobile-row-reverse` option: applies reverse order on mobile blocks  
    - `Editorial card`
        - adds `max-width` option: applies 376px max-width to editorial card
    - `Icon block`
       - adds black background to dark mode
    - `Footer`
      
screencapture-localhost-3000-drafts-ramuntea-spectrum-showcase-2024-08-02-14_47_27 (1)

Resolves: [MWPW-151456](https://jira.corp.adobe.com/browse/MWPW-151456)

**Test URLs:**
- After: https://saasha-mwpw-151456-spectrum2-demo--milo--adobecom.hlx.page/drafts/ramuntea/spectrum-showcase

### DIFF
--- a/libs/blocks/accordion/accordion.css
+++ b/libs/blocks/accordion/accordion.css
@@ -21,8 +21,8 @@ dl.accordion {
 .accordion dt button {
   align-items: center;
   border: none;
-  border-radius: var(--sn-corner-radius-800, 12px);
-  background: var(--s2-gray-50, #F8F8F8);
+  border-radius: var(--sn-corner-radius-800);
+  background: var(--s2-gray-50);
   margin-bottom: var(--spacing-xs);
   color: var(--text-color);
   display: flex;
--- a/libs/blocks/accordion/accordion.css
+++ b/libs/blocks/accordion/accordion.css
@@ -190,6 +190,10 @@ div.media-p {
 }
 
 @media screen and (min-width: 1280px) {
+  .accordion.foreground {
+    max-width: 1200px;
+  }
+
   .editorial {
     display: flex;
     gap: 54px;
--- a/libs/blocks/accordion/accordion.css
+++ b/libs/blocks/accordion/accordion.css
@@ -1,5 +1,5 @@
 .accordion-container {
-  padding: var(--spacing-m) 0;
+  padding: var(--spacing-l) var(--grid-margins-width) 0;
 }
 
 div.accordion {
@@ -9,28 +9,28 @@ div.accordion {
 dl.accordion {
   margin: 0 auto;
   max-width: var(--grid-container-width);
-  border-bottom: 1px solid var(--color-gray-500);
 }
 
 .accordion dd {
   margin: 0;
-  padding: var(--spacing-xs);
+  padding: var(--spacing-xl);
   font-size: var(--type-body-s-size);
   line-height: var(--type-body-s-lh);
 }
 
 .accordion dt button {
   align-items: center;
-  background: none;
-  border: 1px solid var(--color-gray-500);
-  border-width: 1px 0 0;
+  border: none;
+  border-radius: var(--sn-corner-radius-800, 12px);
+  background: var(--s2-gray-50, #F8F8F8);
+  margin-bottom: var(--spacing-xs);
   color: var(--text-color);
   display: flex;
   font-family: var(--body-font-family);
-  font-size: var(--type-heading-xs-size);
+  font-size: var(--type-detail-l-size);
   font-weight: 700;
-  line-height: var(--type-heading-s-lh);
-  padding: var(--spacing-s) var(--spacing-m) var(--spacing-s) var(--spacing-xs);
+  line-height: var(--type-title-lh);
+  padding: var(--spacing-l) var(--spacing-xxxl) var(--spacing-l) var(--spacing-l);
   position: relative;
   text-align: left;
   width: 100%;
@@ -48,7 +48,7 @@ dl.accordion {
 
 .accordion-icon {
   position: absolute;
-  right: var(--spacing-xs);
+  right: var(--spacing-xxl);
   top: 50%;
   margin-top: -6px;
   width: 12px;
@@ -83,28 +83,15 @@ html[dir="rtl"] .accordion-icon {
   height: 2px;
 }
 
-.accordion dt button::before {
-  content: "";
-  display: block;
-  position: absolute;
-  width: 2px;
-  height: 100%;
-  left: 0;
-  top: 0;
-  background-color: transparent;
-}
-
-.accordion dt button[aria-expanded="true"]::before {
-  background-color: var(--link-color-dark);
-}
-
-.quiet .accordion dt button[aria-expanded="true"]::before {
-  display: none;
+.accordion dt button[aria-expanded="true"]{
+  background: var(--color-black);
+  color: var(--color-white);
 }
 
 .accordion dt button[aria-expanded="true"] .accordion-icon::before,
 .accordion dt button[aria-expanded="true"] .accordion-icon::after {
   transform: rotate(90deg);
+  background: var(--color-white);
 }
 
 .accordion dt button[aria-expanded="true"] .accordion-icon::after {
@@ -113,13 +100,14 @@ html[dir="rtl"] .accordion-icon {
 
 .accordion dt button:hover .accordion-icon::before,
 .accordion dt button:hover .accordion-icon::after {
-  background: var(--color-black);
+  background: var(--color-white);
 }
 
 .accordion dt button:focus,
 .accordion dt button:hover,
 .accordion dt button[aria-expanded="true"] {
-  background: #00000005;
+  background: var(--color-black);
+  color: var(--color-white);
 }
 
 .accordion dd p:first-child {
@@ -201,7 +189,7 @@ div.media-p {
   padding: 0;
 }
 
-@media screen and (min-width: 1200px) {
+@media screen and (min-width: 1280px) {
   .editorial {
     display: flex;
     gap: 54px;
@@ -227,5 +215,11 @@ div.media-p {
     align-items: center;
     overflow: hidden;
     margin: 0;
+  }
+}
+
+@media screen and (min-width: 1680px) {
+  .accordion-container {
+    padding: var(--spacing-l) 0 0;
   }
 }
--- a/libs/blocks/aside/aside.css
+++ b/libs/blocks/aside/aside.css
@@ -39,19 +39,16 @@
   justify-content: center;
   flex-direction: column;
   gap: var(--spacing-l);
-}
-
-.aside:not(.notification):not(.inline) .foreground > :first-child {
-  padding: var(--spacing-xxxl) 0 0;
-}
-
-.aside:not(.notification):not(.inline) .foreground > :last-child {
-  margin-bottom: var(--spacing-xxxl);
+  padding: var(--spacing-xl) var(--grid-margins-width);
 }
 
 .aside .foreground.container .text {
   display: flex;
   flex-wrap: wrap;
+}
+
+.aside.mobile-row-reverse .foreground.container {
+  flex-direction: column-reverse;
 }
 
 .aside.split .foreground.container .text {
@@ -111,12 +108,16 @@
 }
 
 .aside .foreground.container .text .heading-xl {
-  margin: 0 0 var(--spacing-xs);
+  margin: 0 0 var(--spacing-m);
 }
 
 .aside .foreground.container .text .body-m,
 .aside .foreground.container .text .body-s {
-  margin-bottom: var(--spacing-s);
+  margin-bottom: var(--spacing-m);
+}
+
+.aside .foreground.container .text .body-xl{
+  margin-bottom: var(--spacing-l);
 }
 
 .aside .foreground.container .text .supplemental-text {
@@ -311,7 +312,7 @@
   min-height: 0;
   display: flex;
   flex-direction: column;
-  padding: 0;
+  padding: 0 var(--grid-margins-width);;
   margin: 0;
 }
 
@@ -611,7 +612,7 @@
   stroke: var(--color-white);
 }
 
-@media screen and (max-width: 600px) {
+@media screen and (max-width: 768px) {
   .aside.no-media:not(.notification) .foreground.container {
     gap: 0;
   }
@@ -625,7 +626,7 @@
   }
 }
 
-@media screen and (min-width: 600px) {
+@media screen and (min-width: 768px) {
   .aside.small {
     min-height: 420px;
   }
@@ -634,16 +635,16 @@
     min-height: 560px;
   }
 
-  .aside.large {
-    min-height: 700px;
-  }
 
   .aside .foreground.container {
     align-items: center;
     flex-direction: row;
     margin: 0 auto;
-    padding: var(--spacing-m) 0;
     gap: var(--spacing-xl);
+  }
+
+  .aside.mobile-row-reverse .foreground.container {
+    flex-direction: unset;
   }
 
   .aside:not(.notification):not(.inline) .foreground.container > :first-child {
@@ -656,6 +657,7 @@
 
   .aside .foreground.container .image {
     margin: 0;
+    max-height: 640px;
   }
 
   .aside .foreground.container .text.image {
@@ -922,13 +924,13 @@
   }
 }
 
-@media (min-width: 600px) and (max-width: 1200px) {
+@media (min-width: 768px) and (max-width: 1280px) {
   .aside.promobar.popup .tablet-up.hide-block ~ .promo-close {
     display: none;
   }
 }
 
-@media screen and (min-width: 1200px) {
+@media screen and (min-width: 1280px) {
   .aside .foreground.container {
     min-height: 420px;
     width: var(--grid-container-width);
@@ -945,10 +947,6 @@
 
   .aside .foreground.container .icon-area {
     max-width: 400px;
-  }
-
-  .aside .foreground.container .text .heading-xl {
-    margin: 0 0 var(--spacing-xs);
   }
 
   .aside.inline .foreground.container .text {
@@ -1138,4 +1136,15 @@
   .aside.promobar.popup .promo-text .text-area {
     margin-top: 0;
   }
+}
+
+@media screen and (min-width: 1680px) {
+  .aside .foreground.container {
+    padding: var(--spacing-xl) 0;
+  }
+
+  .aside.promobar .foreground.container {
+    padding: 0;
+  }
+  
 }
--- a/libs/blocks/aside/aside.css
+++ b/libs/blocks/aside/aside.css
@@ -130,6 +130,10 @@
   gap: var(--spacing-xs);
 }
 
+.aside.promobar strong {
+  font-weight: inherit;
+}
+
 .aside.promobar .promo-text .action-area,
 .aside .foreground.container .text .action-area {
   margin-bottom: 0;
@@ -635,6 +639,9 @@
     min-height: 560px;
   }
 
+  .aside.large {
+    min-height: 640px;
+  }
 
   .aside .foreground.container {
     align-items: center;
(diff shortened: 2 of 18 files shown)


libs/blocks/editorial-card/editorial-card.css
```diff
@@ -55,6 +55,11 @@
   min-width: initial;
 }
 
+.editorial-card.max-width {
+  max-width: 376px;
+  margin: auto
```
overmyheadandbody, line 60:
```suggestion
  margin: auto;
```

libs/blocks/editorial-card/editorial-card.css
```diff
@@ -55,6 +55,11 @@
   min-width: initial;
 }
 
+.editorial-card.max-width {
```
overmyheadandbody, line 58: I'm guessing this comes from a variation added through Sharepoint? The naming might be a bit misleading, but it should be acceptable for the purposes of the showcase page. Could you add a comment that the variant name might need to be updated?

libs/blocks/footer/footer.css
```diff
@@ -304,9 +304,6 @@ footer .footer-region-button.inline-dialog-active svg.icon-chevron-down {
 }
 
 /* PRIVACY INFO */
-.footer-privacy {
-  font-size: var(--type-detail-s-size);
-}
```
overmyheadandbody, line 309: This file is used just for older pages, are you sure it achieves your desired result? It might be that we need to make some metadata changes to the page if that's the case. Same comment for L15

libs/blocks/icon-block/icon-block.css
```diff
@@ -6,11 +6,20 @@
 
 .dark .icon-block,
 .icon-block.dark {
-  color: #fff;
+  color: var(--color-white);
+  background-color: var(--color-black);
+  width: 90%;
+  margin: 0 auto var(--spacing-xl);
+  border-radius: var(--s2-corner-radius-900)
```
overmyheadandbody, line 13:
```suggestion
  border-radius: var(--s2-corner-radius-900);
```

libs/blocks/icon-block/icon-block.css
```diff
@@ -6,11 +6,20 @@
 
 .dark .icon-block,
 .icon-block.dark {
-  color: #fff;
+  color: var(--color-white);
+  background-color: var(--color-black);
+  width: 90%;
```
overmyheadandbody, line 11: Why is this reduction needed for `.dark` variants?

libs/blocks/icon-block/icon-block.css
```diff
@@ -109,6 +118,12 @@
   padding-top: 24px;
 }
 
+.icon-block.m-icon .icon-area,
+.icon-block.m-icon .icon-area picture img {
+  min-width:var(--icon-size-xs);
```
overmyheadandbody, line 123:
```suggestion
  min-width: var(--icon-size-xs);
```

libs/blocks/marquee/marquee.css
```diff
@@ -190,7 +190,8 @@
 }
 
 .marquee.large {
-  display: block;
+  flex-direction: column-reverse;
+  background: var(--s2-background-layer-1-color)
```
overmyheadandbody, line 194:
```suggestion
  background: var(--s2-background-layer-1-color);
```

libs/blocks/marquee/marquee.css
```diff
@@ -321,12 +326,12 @@
...
 
 .marquee.large .text .action-area {
-  order: 2;
+  flex-direction: column-reverse;
   margin-bottom: var(--spacing-s);
 }
 
 .marquee.large .text .body-xl {
-  order: 3;
+  margin-bottom:var(--spacing-l);
```
overmyheadandbody, line 334:
```suggestion
  margin-bottom: var(--spacing-l);
```

libs/blocks/marquee/marquee.css
```diff
@@ -321,12 +326,12 @@
 }
 
 .marquee.large .text .action-area {
-  order: 2;
+  flex-direction: column-reverse;
   margin-bottom: var(--spacing-s);
 }
 
 .marquee.large .text .body-xl {
-  order: 3;
```
overmyheadandbody, line 329: Do we no longer need the `order` at all?

libs/blocks/marquee/marquee.css
```diff
@@ -548,12 +558,20 @@
...
+    background: unset;
+  }
+
   .marquee.large.desktop-dark {
     background-color: black;
   }
 
   .marquee.large.desktop-light {
-    background-color: initial;
+    background-color: blue;
```
overmyheadandbody, line 574: Is this intentional or forgotten after some testing? Just asking because we usually have variables for colors.

libs/styles/styles.css
```diff
@@ -19,12 +19,12 @@
   /* Milo Color Specs */
   --color-white: var(--s2-white);
   --color-black: var(--s2-black);
-  --color-gray-100: var(--s2-gray-100);
-  --color-gray-200: #e8e8e8; /* milo only */
+  --color-gray-100: #f8f8f8;
```
overmyheadandbody, line 22: Is the color no longer available from Spectrum?

libs/styles/styles.css
```diff
@@ -403,8 +403,8 @@
   }
 
   @media (min-width: 1680px) { /* Large Desktop */
-    --grid-container-width: 1600px;
-    --grid-margins-width: 0;
+    --grid-container-width: 1600px;
```
overmyheadandbody, line 406:
```suggestion
    --grid-container-width: 1600px;
```

libs/styles/styles.css
```diff
@@ -1066,26 +1068,24 @@ span.icon.margin-inline-end { margin-inline-end: var(--s2-spacing-100); }
...
 }
 
-.light .con-button {
-  border-color: var(--text-color);
-  color: var(--text-color);
-}
+.dark .con-button:hover {
+  background-color: var(--s2-transparent-white-100);
+  border-color: var(--s2-gray-900);
+}
```
overmyheadandbody, line 1079:
```suggestion
}
```

libs/styles/styles.css
```diff
@@ -1066,26 +1068,24 @@ span.icon.margin-inline-end { margin-inline-end: var(--s2-spacing-100); }
...
 
 .dark .con-button {
-  border-color: var(--color-white);
-  color: var(--color-white);
+  background-color: var(--s2-transparent-white-25);
+  color: var(--s2-transparent-white-900);
+  border-color: var(--s2-gray-800);
 }
 
-.light .con-button {
```
overmyheadandbody, line 1073: Why do we need to completely remove the `.light` styles? It might be that they are still valuable for backwards compatibility across our pages.

overmyheadandbody (commented):
A few nits and questions, but overall this looks good, great job!

I'll also dive into a visual inspection soon. The first thing that instantly popped up was the footer where the links seem a bit too close for comfort. It looks like it has to do with some `line-height` definitions
Screenshot 2024-08-05 at 14 39 53

overmyheadandbody:
As for the quick visual inspection:

### Marquee
* something is off with the Adobe Photoshop logo and its relationship with the Marquee title. It might be the logo size or some spacing between the two

### Aside
* the title seems to be `heading-xl`, but the Figma file seems to indicate size L should be used. The detail is M-sized, while Figma suggests L

### Editorial Cards
* these might need a bit more attention - the cell padding, the title size and the link color are different compared to the Figma file

| Demo Page | Figma |
| Screenshot 2024-08-05 at 15 10 40 | Screenshot 2024-08-05 at 15 11 08 |

### FAQ
* the Figma file seems to suggest a wider container of 1200px, while currently it's set at 1000px
* I don't remember where we landed on link colors 😅 Figma seems to suggest blue links, while we're using ones of the same color as the text. Am I remembering correctly that we said this is a current limitation?

### Icon Block
* I'm still not convinced what the intention with its width is, we should ask this question

### Footer
* I already brought this up, but I see your screenshot has the correct styles. I also tried the link in incognito, but get the same compact style. We can sync on this if needed to see what the issue might be

libs/blocks/aside/aside.css
```diff
@@ -634,18 +635,18 @@
     min-height: 560px;
   }
 
-  .aside.large {
-    min-height: 700px;
-  }
```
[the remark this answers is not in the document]
saashapina: Yea that's a good point. The reason this was removed was because the spectrum mockup specifies these new aside large blocks to be 640px in height instead of 700px. Removing this allows the height to fall to the height of the blocks content. However, we could consider explicitly setting a min-height of 640px here instead, which might help mitigate the impact on existing production use cases.

libs/blocks/editorial-card/editorial-card.css
saashapina: Yes it is. Yea I wasn't sure what the best variation name should be for this but I'm open to suggestions to make this more clear. I'll add in that comment for now.

libs/blocks/footer/footer.css
saashapina: Good catch! It seems I was experiencing a rendering issue locally, where this original file for the footer was being displayed instead of the global footer. That would explain why I wasn't seeing the line height issue I thought I fixed 🙃. Restarting my server looked like it fixed the issue.

libs/blocks/marquee/marquee.css
saashapina: Good catch, yes this was from a test.

libs/blocks/marquee/marquee.css
saashapina: No, I don't believe we need this property any more. It was previously used to position the action CTA buttons above the body text on mobile devices but the new mockups are indicating that the buttons should now fall in their original order.

libs/styles/styles.css
saashapina: I can add these styles back in for backwards compatibility and remove the light properties from the blocks here instead. I initially had these removed because the spectrum designs were created to default to the `light` styles automatically, so it's only needed to toggle to `dark` when needed.

libs/styles/styles.css
saashapina: Yes it is, there was a conflict in the designs I noticed when I was seeing the old footer background color using the `gray-100` s2 token color. It was a darker gray then specified in our mockups and the mockups called for this original color gray to still be used. I kept both this milo gray-100 and the s2 gray-100 in temporarily to see if it needed to be discussed with design, but that issue is actually fixed with correct usage of the global footer using the `--feds-background-footer--light`

saashapina:
> A few nits and questions, but overall this looks good, great job!
> 
> I'll also dive into a visual inspection soon. The first thing that instantly popped up was the footer where the links seem a bit too close for comfort. It looks like it has to do with some `line-height` definitions Screenshot 2024-08-05 at 14 39 53

Figured out the issue with the footer was because there was some issue causing the old footer to render for me locally instead of the global. This issue should be fixed now that the changes are added to the intended area.

libs/blocks/icon-block/icon-block.css
saashapina: This width reduction is needed to accommodate the black background on the `.dark` variant that is "full-width" but not entirely full-width per the mockup.
Screenshot 2024-08-06 at 12 08 10 AM

libs/blocks/global-footer/global-footer.css
```diff
@@ -228,6 +222,7 @@
   column-gap: 5px;
   flex-wrap: wrap;
   margin: 0;
+  font-size: 11px;
```
narcis-radu, line 225: Why is the font size different for the copyright section?

narcis-radu (commented):
I noticed a couple of small differences and I'm not sure whether it's because of the design change or just some inconsistencies.
1. The editorial cards are not really aligned with the text above anymore
Screenshot 2024-08-06 at 15 15 50

2. FAQ links have the same color as text and there's no style update on hover
Screenshot 2024-08-06 at 15 17 03

3. Region selector links have the same color as normal text and hovering does not produce a color change
Screenshot 2024-08-06 at 15 17 59

libs/blocks/aside/aside.css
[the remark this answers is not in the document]
overmyheadandbody: Yep, setting it to `640` is likely the better definition to ensure current marquees aren't impacted too much.

libs/blocks/accordion/accordion.css
```diff
@@ -9,28 +9,28 @@ div.accordion {
...
 }
 
 .accordion dt button {
   align-items: center;
-  background: none;
-  border: 1px solid var(--color-gray-500);
-  border-width: 1px 0 0;
+  border: none;
+  border-radius: var(--sn-corner-radius-800, 12px);
+  background: var(--s2-gray-50, #F8F8F8);
```
overmyheadandbody, line 25: Do we have a need for the fallbacks at this point?

overmyheadandbody (commented):
Thanks for integrating all the feedback! I see @narcis-radu also has a few questions worth following up on.

Looking at the design side of things, I still see a few gaps:
* Aside heading size is still off, it now has both `l` and `xl` heading classes, we might have to look into that;
* Editorial cards are looking better, but I feel the title doesn't match the Figma specs. I've tried to replace the current `xs` size with an `s`, but then it gets too large, have all heading font-sizes been updated?
* for the FAQ section, I just wonder about the inner links' color

libs/blocks/global-footer/global-footer.css
saashapina: This was a change specified in the mockups, but we can revert this if its not needed.
Screenshot 2024-08-06 at 10 12 36 AM

narcis-radu: @saashapina - is the other content in the footer 11px as well?

saashapina:
> Thanks for integrating all the feedback! I see @narcis-radu also has a few questions worth following up on.
> 
> Looking at the design side of things, I still see a few gaps:
> 
> * Aside heading size is still off, it now has both `l` and `xl` heading classes, we might have to look into that;
> * Editorial cards are looking better, but I feel the title doesn't match the Figma specs. I've tried to replace the current `xs` size with an `s`, but then it gets too large, have all heading font-sizes been updated?
> * for the FAQ section, I just wonder about the inner links' color

Thanks for catching those details.

- The link color issue in the FAQ section should now be addressed. Looked like there was a typo in the variables causing this.
- Regarding the editorial card heading, yes, all heading font sizes have been updated. However, the mockups specify using the new `Title XL` font size instead of the standard heading sizes for the editorial cards, which is likely why the xs or s heading sizes didn't work for those. It has now been corrected to use the title styles.
- The aside heading issue has also been addressed.

Another quick note regarding the editorial cards is the multi-line truncated paragraph with ellipses. CSS currently lacks a native property for this effect, and while we could use JS, it might have performance implications. It could be worth discussing this with the design team. The best solution might be to create a shortened version of the description and add a hard-coded ellipsis when necessary.

libs/blocks/global-footer/global-footer.css
saashapina: No, it looks like the other content in the footer remains the same as the original at 12px. It was only the details in this privacy section that was updated to the 11px.

saashapina:
> As for the quick visual inspection:
> 
> ### Marquee
> * something is off with the Adobe Photoshop logo and its relationship with the Marquee title. It might be the logo size or some spacing between the two
> 


Regarding the Marquee Adobe logo, it was challenging to match the exact specifications since the mockup uses text while we were provided an image. The current spacing is based on the variables specified in the mockup. We may need to adjust the logo size further, but this is the closest match I could achieve using the available Spectrum icon variables. Let me know if you have any other suggestions for this.

overmyheadandbody:
Re: editorial cards ellipsis, you should be able to do
```
display: -webkit-box;
-webkit-line-clamp: 2;
-webkit-box-orient: vertical;
overflow: hidden;
text-overflow: ellipsis;
```
and this is the result:
Screenshot 2024-08-07 at 13 21 40

libs/styles/styles.css
```diff
@@ -1379,7 +1399,7 @@ a.xl-link {
 
 a {
   color: var(--link-color);
-  text-decoration: none;
```
overmyheadandbody, line 1382: I'm not very convinced with removing this, it adds underline for links in the bottom part of the footer and the region modal, maybe we can adapt this closer to where we actually need it, like the FAQ / Editorial Card sections.

overmyheadandbody (commented):
Just one more nit. And I see the top navigation is now missing, it looks like a metadata issue, I'll look into this tomorrow and fix it.

Otherwise I think we're close to sending this to QA for another round of visual inspection and then we can bring it up to the stakeholders. Great job!

libs/styles/styles.css
saashapina: Yea the reason I removed this is because the Spectrum design system specified that the links should be underlined by default. It seems like the footer, bottom footer, and the region modal are special cases where it shouldn't be applied according to the mockups. It made sense to handle these specific cases individually rather than overriding the default styling globally. What do you think about this approach? Or do you think we should only add the underline to the FAQ/ Editorial cards for the purposes of the demo page?
Screenshot 2024-08-07 at 11 29 56 PM

libs/blocks/global-footer/global-footer.css
narcis-radu: I did some investigations and looks like the copyright text was 11px in the AEM Global Navigation, meaning the change is correct. I will mark this as resolved, no further action needed. Thanks!

libs/styles/styles.css
overmyheadandbody: If the new design decision is to underline, then I think what you're suggesting is better for now, remove text decoration where it's not needed.

libs/blocks/editorial-card/editorial-card.js
```diff
@@ -51,6 +51,10 @@ const decorateForeground = async (el, rows) => {
       row.classList.add('extra-row');
     }
     decorateBlockText(row, ['m', 'm', 'm']); // heading, body, detail
+    const headings = el.querySelectorAll('h1, h2, h3, h4, h5, h6');
+    const heading = headings[headings.length - 1];
+    heading?.classList.add('title-xl');
+    heading?.classList.remove('heading-m');
```
overmyheadandbody, line 57: Could we add a `// TODO` comment with some details to make it clear that this needs to be revisited?

overmyheadandbody (commented):
I was now looking over the mobile design as well and I have a few notes:
* Marquee CTAs have higher height in the Figma file compared to the implementation;
* for the editorial cards section we need to communicate we can't use a slider for the cards right now; additionally, I see the CTA drops below the cards, that's also something we can't achieve OOTB;
* in the icon block, the CTAs seem to stack and are wider and taller in the Figma file compared to the implementation;

Also a few notes for tablets:
* same CTA height note as for mobile;
* there seems to be more space between asides than expected:
Screenshot 2024-08-08 at 15 37 19

* same editorial cards comment, we'll have to communicate we can't achieve that layout right now without additional work;

I'm hoping this is the last round of feedback 😅  It's a big PR with a lot of moving parts, sorry for all the back and forth!

overmyheadandbody:
Also, I had a look over the Gnav issue, it seems it's related to your branch, the `global-navigation/base.css` file is not updated on your branch, things should be working as expected as soon as we merge this into the feature branch. Alternatively, you could try to rebase the feature branch with the current `stage` one, then rebase your changes to the feature branch after pulling it.

overmyheadandbody (approved):
Thanks for integrating all the feedback! I think this is ready for a round of QA and then we can provide details for the stakeholders on how to access this page for showcase purposes.

Pending a second approval, I'd like to merge this in the feature branch so QA can do a sort of E2E testing.

mokimo (approved):

PR #2686, opened 2024-08-02
MWPW-154969 Log LCP and key user attributes

chrischrischris:
This adds the ability to log performance data back to Lana.

To enable the feature, metadata `pageperf = ‘on’ ` must be set for the page.  This feature only works in Chrome browsers.

Additionally `pageperf-rate` can be set to determine what percentage of pages will send data.  E.g `pageperf-rate=15` would have 15% of the pages send perf data. The default rate is `50`%.

The sending of the data to Lana is delayed so as to not affect any parts of the page load.

Data sent:
* Chrome Version
* CLS Score
* Country (only if georouting is enabled via geo2)
* Downlink: Mbps - Chrome caps value at `10`
* LCP Score
* LCP Element Type
* LCP URL
  * If there is no identifying url, the element html will be used.
* Adobe IMS Logged In Status
* OS
* URL of page
* Window Width & Height
* MEP Manifest Data
  * Manifest(s) Used (only the ones that have been applied to the page)
  * Selected Experience in the manifest

Resolves: [MWPW-154969](https://jira.corp.adobe.com/browse/MWPW-154969)

**Test URLs:**
- After: https://154969-log-lcp--milo--adobecom.hlx.page/?martech=off


libs/utils/logWebVitals.js
```diff
@@ -0,0 +1,97 @@
+const LANA_CLIENT_ID = 'pageperf';
+
+let lanaSent = false;
+
+function sendToLana(lanaData) {
+  if (lanaSent) return;
+  const ua = window.navigator.userAgent;
+
+  Object.assign(lanaData, {
+    chromeVer: (ua.match(/Chrome\/(\d+\.\d+\.\d+\.\d+)/) || [])[1] || '',
```
robert-bogos, line 10: Nit: This can be shorter with optional chaining.
```suggestion
    chromeVer: ua.match(/Chrome\/(\d+\.\d+\.\d+\.\d+)/)?.[1] || '',
```

libs/utils/logWebVitals.js
```diff
@@ -0,0 +1,97 @@
...
+    country: sessionStorage.getItem('akamai') || '',
+    // eslint-disable-next-line compat/compat
+    downlink: window.navigator?.connection?.downlink || '',
+    loggedIn: window.adobeIMS?.isSignedInUser() || false,
+    os: (ua.match(/Windows/) && 'win')
+      || (ua.match(/Mac/) && 'mac')
+      || (ua.match(/Android/) && 'android')
+      || (ua.match(/Linux/) && 'linux')
+      || '',
```
robert-bogos, line 19: Nit: this can be less repetitive.
```suggestion
    os: [[/Windows/, 'win'], 
          [/Mac/, 'mac'], 
          [/Android/, 'android'], 
          [/Linux/, 'linux']].find(i => ua.match(i[0]))?.[1] || '';
```

libs/utils/utils.js
```diff
@@ -998,6 +998,21 @@ export function scrollToHashedElement(hash) {
   });
 }
 
+function logPagePerf() {
+  const isChrome = () => {
+    const nav = window.navigator;
+    return nav.userAgent.includes('Chrome') && nav.vendor.includes('Google');
+  };
+
+  if (getMetadata('pageperf') === 'on' && isChrome()) {
```
overmyheadandbody, line 1007: Minor nit: I'd guess the feature would be disabled more often than not, right? In which case maybe we can do an early return based on the `pageperf` metadata.
```suggestion
  if (getMetadata('pageperf') !== 'on') return;

  const isChrome = () => {
    const nav = window.navigator;
    return nav.userAgent.includes('Chrome') && nav.vendor.includes('Google');
  };

  if (!isChrome()) return;
```

libs/utils/logWebVitals.js
```diff
@@ -0,0 +1,97 @@
...
+}
+
+function observeCLS(lanaData) {
+  let cls = 0;
+  /* c8 ignore start */
+  new PerformanceObserver((entryList) => {
+    for (const entry of entryList.getEntries()) {
+      if (!entry.hadRecentInput) {
+        cls += entry.value;
+      }
```
robert-bogos, line 46: Nit: this could be an one liner.
```suggestion
      if (!entry.hadRecentInput) cls += entry.value;
```

libs/utils/logWebVitals.js
```diff
@@ -0,0 +1,97 @@
...
+}
+
+function getElementInfo(el) {
+  const elSrc = el.src || el.currentSrc || el.href || el.poster;
+  if (elSrc) {
+    try {
+      const srcUrl = new URL(elSrc);
+      return srcUrl.origin === window.location.origin ? srcUrl.pathname : srcUrl.href;
+    } catch {
+      // fall through
```
robert-bogos, line 59: Any reason for not logging the error in here?

robert-bogos (approved):
Looks good 👍 Just a few suggestions.

overmyheadandbody (approved):

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-154969

mokimo:
@SilviuLCF @robert-bogos @overmyheadandbody implemented the PR feedback as Chris is out, could you have another look?

libs/utils/logWebVitals.js
mokimo: I felt Chris version is a little easier to understand on the first read, left it at that

overmyheadandbody (approved):

PR #2700, opened 2024-08-07
Fix: Load gnav search files only if it is authored

bandana147:
* Removing unnecessary loading of gnav search JS and CSS files.

Resolves: [MWPW-154608](https://jira.corp.adobe.com/browse/MWPW-154608)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.live/drafts/blaishram/gnav-search-page?martech=off
- After: https://gnav-search--milo--bandana147.hlx.live/drafts/blaishram/gnav-search-page?martech=off

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -405,13 +405,15 @@ class Gnav {
       try {
         this.block.removeEventListener('click', this.loadDelayed);
         this.block.removeEventListener('keydown', this.loadDelayed);
-        const [
-          Search,
-        ] = await Promise.all([
-          loadBlock('../features/search/gnav-search.js'),
-          loadStyles(rootPath('features/search/gnav-search.css')),
-        ]);
-        this.Search = Search;
+        if (this.searchPresent()) {
+          const [
+            Search,
+          ] = await Promise.all([
+            loadBlock('../features/search/gnav-search.js'),
+            loadStyles(rootPath('features/search/gnav-search.css')),
+          ]);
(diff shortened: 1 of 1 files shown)


sharmrj (approved):

salonijain3 (approved):

Deva309 (approved):

nishantka (approved):

sigadamvenkata:
Issue verfied with given branch libraries. if search is not authored in Gnav then respective 'gnav-search.js' / 'gnav-search.css' files doesn't load.

without fix : https://stage--cc--adobecom.hlx.live/products/firefly 
After Fix : https://stage--cc--adobecom.hlx.live/products/firefly?milolibs=gnav-search--milo--bandana147
image
@bandana147 @salonijain3

overmyheadandbody (approved):

mokimo (approved):

mokimo (approved):

mokimo:
Is search even still used anywhere? You might be able to remove all the search related logic. 
You could also add lana logs to find out _where/if_ search is still used

overmyheadandbody:
@mokimo - I had the same question in the past, but didn't get a reply. I'm assuming that this PR comes out of a need from somewhere where search is used, but it would be great to understand what that use-case is. CC: @bandana147

bandana147:
Yes, that is our next step. We will remove it once we confirm that it is not required anywhere else.

sonawanesnehal3 (approved):

sigadamvenkata:
regression on stage
image

PR #2702, opened 2024-08-07
MWPW-152539 Strikethrough price is not displayed for legacy promo offers

bozojovicic:
This change is approved 2 times before ([#29](https://github.com/adobecom/mas/pull/29)) but PR is closed because we were moving the code from Tacocat to MAS, and then again from MAS to Milo. So let's try it again. 
At the moment we don't display the old price for legacy promo prices, no matter if the `old` query parameter is there and which value it has in the merch link. Example [page](https://main--milo--adobecom.hlx.page/drafts/bozo/old-price-drew-copy)
With this change the old price is displayed by default or when `old=true`. Example [page](https://mwpw152539legacypromo--milo--bozojovicic.hlx.page/drafts/bozo/old-price-drew-copy)
I'm not sure how many of such prices we still have. Drew said maybe even none but he asked for this to be fixed.

Resolves: [MWPW-152539](https://jira.corp.adobe.com/browse/MWPW-152539)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/bozo/price-cta?martech=off&georouting=off
- After: https://mwpw152539legacypromo--milo--bozojovicic.hlx.page/drafts/bozo/price-cta?martech=off&georouting=off


bozojovicic:
There are 615 legacy promo prices without "old" parameter with the old price manually added next to the price. With this code change the old price will be displayed twice. The content needs to be fixed before this is merged. Closing the PR.

PR #2703, opened 2024-08-07
MWPW-155943 Use correct image source for video when using poster Image

suhjainadobe:
It has been observed that, when poster Image is authored for video on the marquee, the image src is incorrect for the video. It is taking the mobile src of the image instead of the desktop src, which is making it download it again on the marquee affecting the marquee.

Resolves: [MWPW-155943](https://jira.corp.adobe.com/browse/MWPW-NUMBER)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.live/drafts/suhjain/performance/heromarquee
- After: https://video-poster--milo--suhjainadobe.hlx.live/drafts/suhjain/performance/heromarquee


overmyheadandbody (commented):
I know @mokimo recently looked into this from a performance perspective, he may have some good insights. Personally, I didn't look deep into this, but I want to avoid adding all this additional logic to `utils.js`. It seems that the target is the marquee, that's where the changes should ideally happen. You're also defining some seemingly global device sizes, when in reality they're specific to the marquee, strengthening my previous suggestion.

mokimo (commented):
I agree with @overmyheadandbody the utils is likely the wrong spot for this, as they should be reserved for code that's needed 100% of the time. (In the best case)

Can this go into the `video.js` or `getVideoAttrs` within decorate?`

suhjainadobe:
> I know @mokimo recently looked into this from a performance perspective, he may have some good insights. Personally, I didn't look deep into this, but I want to avoid adding all this additional logic to `utils.js`. It seems that the target is the marquee, that's where the changes should ideally happen. You're also defining some seemingly global device sizes, when in reality they're specific to the marquee, strengthening my previous suggestion.

I agree that the target is marquee, and that is why the device sizing is defined. But this is for all the poster-images in general. And I'll work on that, taking your suggestion.
CC: @mokimo

suhjainadobe:
@mokimo @overmyheadandbody : As this is a general approach for video-poster, thus changed the approach where it will detect for mobile and desktop devices in general. If, I shift this approach to some other file, there will be a dependency to call that file and will to the network call. 
CC: @aishwaryamathuria @salonijain3

mokimo:
@suhjainadobe `decorate.js` is already used for videos, can't the logic live in there? https://github.com/adobecom/milo/blob/stage/libs/utils/decorate.js

libs/utils/utils.js
```diff
@@ -531,7 +547,7 @@ export function decorateImageLinks(el) {
       const pic = img.closest('picture');
       const picParent = pic.parentElement;
       if (href.includes('.mp4')) {
-        const a = createTag('a', { href: url, 'data-video-poster': img.src });
+        const a = createTag('a', { href: url, 'data-video-poster': getImgSrc(picParent) });
```
mokimo, line 550: I would shift the device detection 
```suggestion
        const a = createTag('a', { 
          href: url, 
          'data-video-poster-mobile': picParent.querySelector('source[type="image/webp"]:not([media])'),
          'data-video-poster-desktop': picParent.querySelector('source[type="image/webp"][media]')
        });
```

Then in decorate.js https://github.com/adobecom/milo/blob/stage/libs/utils/decorate.js#L179

```js
const getVideoPoster = (dataset) => {
  if(!videoPosterMobile || !videoPosterDesktop) return
  if(window.innerWidth <= 600) return `poster='${videoPosterMobile}'`
  return `poster='${videoPosterDesktop}'`
}
const poster = getVideoPoster(dataset) || ''
```

overmyheadandbody: I like this option! Going a step further, could we not just send the element that contains all the sources and let `decorate.js` decide?

On another note, I'm still not convinced about the `600px` value, even marquees start mobile styles at `768px`, could we have another look?

suhjainadobe: I have made changes and set the video-poster dataset to the whole picture element and handled the logic in decorate.js. @overmyheadandbody : I observed that the styles for mobile start from 600px for marquee.

libs/utils/decorate.js
```diff
@@ -171,12 +171,29 @@ export function decorateTextOverrides(el, options = ['-heading', '-body', '-deta
...
+  return `poster='${source.srcset}'`;
+}
+
 export function getVideoAttrs(hash, dataset) {
   const isAutoplay = hash?.includes('autoplay');
   const isAutoplayOnce = hash?.includes('autoplay1');
   const playOnHover = hash?.includes('hoverplay');
   const playInViewport = hash?.includes('viewportplay');
-  const poster = dataset?.videoPoster ? `poster='${dataset.videoPoster}'` : '';
+  const poster = dataset?.videoPoster ? getImgSrc(dataset.videoPoster) : '';
```
mokimo, line 196:
```suggestion
  const poster = getImgSrc(dataset.videoPoster)
```

libs/utils/decorate.js
```diff
@@ -171,12 +171,29 @@ export function decorateTextOverrides(el, options = ['-heading', '-body', '-deta
...
+  return 'desktop';
+}
+
+export function getImgSrc(pic) {
+  let source = '';
+  const parser = new DOMParser();
+  const doc = parser.parseFromString(pic, 'text/html');
+  if (defineDeviceByScreenSize() === 'mobile') source = doc.querySelector('source[type="image/webp"]:not([media])');
+  else source = doc.querySelector('source[type="image/webp"][media]');
+  return `poster='${source.srcset}'`;
```
mokimo, line 188:
```suggestion
  return source.srcset ? `poster='${source.srcset}'` : ''
```

mokimo (commented):
Two suggestions, to ensure a poster is always added with an attribute, rather than `poster=undefined`

libs/utils/decorate.js
suhjainadobe: done

libs/utils/decorate.js
suhjainadobe: done

salonijain3 (approved):

spadmasa:
Validation done on the pr : 
https://main--cc--adobecom.hlx.live/drafts/performance/photoshop-original?milolibs=video-poster--milo--suhjainadobe
media image loaded is seen the same
https://main--cc--adobecom.hlx.live/drafts/performance/media_1af47aa74ac13e84f82ff198accfafb315ebdcdea.jpeg?width=2000&format=webply&optimize=medium
image
image
request url 

https://main--cc--adobecom.hlx.live/drafts/performance/media_1af47aa74ac13e84f82ff198accfafb315ebdcdea.jpeg?width=2000&format=webply&optimize=medium

Performance  
https://pagespeed.web.dev/analysis/https-main--cc--adobecom-hlx-live-drafts-performance-photoshop-original/ajd3lv4hn4?form_factor=mobile
image
https://pagespeed.web.dev/analysis/https-main--cc--adobecom-hlx-live-drafts-performance-photoshop-original/ajd3lv4hn4?form_factor=desktop
image
https://main--cc--adobecom.hlx.page/products/photoshop?milolibs=video-poster--milo--suhjainadobe&mep=off
image

PR #2705, opened 2024-08-07
[MWPW-153665] promo utils fastload

seanchoi-dev:
Let personalization.js loads entitlements.js and promo-utils.js simultaneously so promo-utils.js doesn't block to load entitlements.js.

Before:
Screenshot 2024-08-07 at 3 28 52 PM

After:
Screenshot 2024-08-07 at 3 30 41 PM

Related: [MWPW-153665](https://jira.corp.adobe.com/browse/MWPW-153665)

**Test URLs:**
- After: https://MWPW-153665-promo-utils-fastload--milo--adobecom.hlx.page/?martech=off


Ruchika4 (approved):

mokimo:
Wouldn't this mean that we now load extra bytes when promos are not used?

seanchoi-dev:
> Wouldn't this mean that we now load extra bytes when promos are not used, making performance slightly worse in those cases?

@mokimo  Yes, it will load extra bytes like 1.6kb and I think it is too small to affect performance even for those cases.
I think there is more benefit to load in parallel.

narcis-radu (changes requested):
@seanchoi-dev - I don't think we should load code that's not needed just to satisfy a bunch of use-cases. I tested on a random page with a logged-out visitor, no promo enabled and the file `promo-utils.js` is still loaded. I saw your explanation
> it will load extra bytes like 1.6kb and I think it is too small to affect performance

and I think that even though the file is currently small it can grow in the future and folks will forget that it is being loaded no matter what.
We should find other ways to tweak this.

npeltier:
to avoid systematic load, @seanchoi-dev could you add the load of promoUtils in utils in case promo enablement is defined? somewhere in `getPromoMepEnablement` or `checkForPageMods`

PR #2706, opened 2024-08-07
[MWPW-154970] allow authoring hash value delay=0 for promo modals

Ruchika4:
* Allow authoring hash value delay=0 for modals authored using promos.
* This is for the performance gain we see when these modals are authored with delay=0 compared to delay=1

Resolves: [MWPW-154970](https://jira.corp.adobe.com/browse/MWPW-154970)

**Test URLs:**
Milo:

- Before: https://stage--milo--adobecom.hlx.page/drafts/ruchika/document?martech=off
- After: https://MWPW-154970--milo--adobecom.hlx.page/drafts/ruchika/document?martech=off

CC
- Before: https://main--cc--adobecom.hlx.page/drafts/ruchika/tiger/promo/illustrator-no-delay?martech=off

- After: https://main--cc--adobecom.hlx.page/drafts/ruchika/tiger/promo/illustrator-no-delay?milolibs=MWPW-154970&martech=off

### DIFF
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -222,7 +222,7 @@ export function getHashParams(hashStr) {
       params.hash = part;
     } else {
       const [key, val] = part.split('=');
-      if (key === 'delay' && parseInt(val, 10) > 0) {
+      if (key === 'delay') {
         params.delay = parseInt(val, 10) * 1000;
       }
     }
@@ -232,7 +232,7 @@ export function getHashParams(hashStr) {
 
 export function delayedModal(el) {
   const { hash, delay } = getHashParams(el?.dataset.modalHash);
-  if (!delay || !hash) return false;
+  if (delay === undefined || !hash) return false;
   isDelayedModal = true;
   const modalOpenEvent = new Event(`${hash}:modalOpen`);
   const pagesModalWasShownOn = window.sessionStorage.getItem(`shown:${hash}`);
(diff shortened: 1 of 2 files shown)


mokimo (approved):

mokimo (approved):
if there is a performance gain, should this be the standard rather than authorable? Or can we programmatically decide when it makes sense?

mirafedas:
@Ruchika4 could you please update the documentation as well: https://milo.adobe.com/docs/authoring/promotions/delayed-modal ?

mirafedas (approved):

Ruchika4:
> if there is a performance gain, should this be the standard rather than authorable? Or can we programmatically decide when it makes sense?

I think we didn't wanted to let go of the whole delay feature. Current code supports modal load using promo on page load only if it has delay query param. Hence added an option to author delay=0 to improve performance. If not delay query param we'll still need some indication from authoring to know if that promo needs to be loaded on page load.
I also felt that we should keep the delay query param author-able as there might be a business use case in future where we'll need to define a delay value other than 0.

spadmasa:
Validation done on the below url and promo modal is seen loaded with hash value delay=0 
https://main--cc--adobecom.hlx.page/drafts/ruchika/tiger/promo/illustrator-no-delay?milolibs=MWPW-154970&martech=off&mep=%2Fproducts%2Fillustrator.json--default---%2Fcc-shared%2Ffragments%2Ftests%2F2024%2Fq2%2Face0875%2Face0875.json--default---%2Fdrafts%2Fpromos%2F2024%2Fglobal%2Ftest-promotion.json
image
regression on modal is done and details are updated in https://jira.corp.adobe.com/browse/MWPW-154970

PR #2712, opened 2024-08-08
MWPW-156220: Support pa code and product code for

yesil:
checkout link configuration in addition to product family.

Resolves: [MWPW-156220](https://jira.corp.adobe.com/browse/MWPW-156220)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ilyas/checkout-link-config?martech=off
- After: https://mwpw-156220--milo--adobecom.hlx.page/drafts/ilyas/checkout-link-config?martech=off

**CC**
https://main--cc--adobecom.hlx.page/products/catalog?milolibs=mwpw-156220#search=acrobat

### DIFF
--- a/libs/blocks/merch/merch.js
+++ b/libs/blocks/merch/merch.js
@@ -198,7 +198,13 @@ export async function fetchCheckoutLinkConfigs(base = '') {
   return fetchCheckoutLinkConfigs.promise;
 }
 
-export async function getCheckoutLinkConfig(productFamily) {
+/**
+ * @param {*} productFamily product family code
+ * @param {*} productCode product code, can be undefined
+ * @param {*} paCode product arrangement code, can be undefined
+ * @returns checkout link config for the given product family and code
+ */
+export async function getCheckoutLinkConfig(productFamily, productCode, paCode) {
   let { base } = getConfig();
   if (/\.page$/.test(document.location.origin)) {
     /* c8 ignore next 2 */
@@ -206,14 +212,27 @@ export async function getCheckoutLinkConfig(productFamily) {
   }
   const checkoutLinkConfigs = await fetchCheckoutLinkConfigs(base);
   const { locale: { region } } = getConfig();
-  const productFamilyConfigs = checkoutLinkConfigs.data?.filter(
-    ({ [NAME_PRODUCT_FAMILY]: mappingProductFamily }) => mappingProductFamily === productFamily,
-  );
-  if (productFamilyConfigs.length === 0) return undefined;
-  const checkoutLinkConfig = productFamilyConfigs.find(
+  let productCheckoutLinkConfigs = [];
+  if (paCode) {
+    productCheckoutLinkConfigs = checkoutLinkConfigs.data?.filter(
+      ({ [NAME_PRODUCT_FAMILY]: code }) => code === paCode,
+    );
+  }
+  if (productCode && productCheckoutLinkConfigs.length === 0) {
+    productCheckoutLinkConfigs = checkoutLinkConfigs.data?.filter(
+      ({ [NAME_PRODUCT_FAMILY]: code }) => code === productCode,
+    );
+  }
+  if (productCheckoutLinkConfigs.length === 0) {
+    productCheckoutLinkConfigs = checkoutLinkConfigs.data?.filter(
+      ({ [NAME_PRODUCT_FAMILY]: code }) => code === productFamily,
+    );
+  }
+  if (productCheckoutLinkConfigs.length === 0) return undefined;
+  const checkoutLinkConfig = productCheckoutLinkConfigs.find(
     ({ [NAME_LOCALE]: locale }) => locale === '',
   );
-  const checkoutLinkConfigOverride = productFamilyConfigs.find(
+  const checkoutLinkConfigOverride = productCheckoutLinkConfigs.find(
     ({ [NAME_LOCALE]: locale }) => locale === region,
   ) ?? {};
   const overrides = Object.fromEntries(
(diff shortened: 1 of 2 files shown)


libs/blocks/merch/merch.js
```diff
@@ -198,22 +198,41 @@ export async function fetchCheckoutLinkConfigs(base = '') {
...
   const checkoutLinkConfigs = await fetchCheckoutLinkConfigs(base);
   const { locale: { region } } = getConfig();
-  const productFamilyConfigs = checkoutLinkConfigs.data?.filter(
-    ({ [NAME_PRODUCT_FAMILY]: mappingProductFamily }) => mappingProductFamily === productFamily,
-  );
-  if (productFamilyConfigs.length === 0) return undefined;
-  const checkoutLinkConfig = productFamilyConfigs.find(
+  let productCheckoutLinkConfigs = [];
+  if (paCode) {
+    productCheckoutLinkConfigs = checkoutLinkConfigs.data?.filter(
```
npeltier, line 217: i would rather only use the filter function once, with 3 tests in there, no? only one traversal of the array, with quick test looks quicker to me

yesil: no, in that case, it will likely find first the entry with the product family rather than pa code  unless we require GWP to author product code and pa codes entries first.

test/blocks/merch/merch.test.js
```diff
@@ -499,6 +507,20 @@ describe('Merch Block', () => {
...
+      checkoutLinkConfig = await getCheckoutLinkConfig('', '', 'testPaCode');
+      expect(checkoutLinkConfig.DOWNLOAD_TEXT).to.equal('paCode');
+    });
+
+    it('getCheckoutLinkConfig: finds using productCode', async () => {
+      let checkoutLinkConfig = await getCheckoutLinkConfig(undefined, 'testProductCode', undefined);
+      expect(checkoutLinkConfig.DOWNLOAD_TEXT).to.equal('productCode');
+      checkoutLinkConfig = await getCheckoutLinkConfig('', 'testProductCode', '');
+      expect(checkoutLinkConfig.DOWNLOAD_TEXT).to.equal('productCode');
+    });
```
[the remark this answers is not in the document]
yesil: > twice the same family code 
Not sure if I understand it.  Too specific test cases will make test unnecessarily complex and verbose.
Having said that, I have verified the use case manually, see https://mwpw-156220--milo--adobecom.hlx.page/drafts/ilyas/checkout-link-config?martech=off

Clicking on Buy Now opens the AI buy now modal (the iframe doesn't open due to CSP but you can verify the correctness of TWP url in the DOM)

libs/blocks/merch/merch.js
npeltier: yes, but typical case with a family code line match implies traversal and test of 100 entries (l217) (no match), then another traversal of 100 entries (l221) (no match), then finally traversal of 100 entries to filter the right ones in l227.
my proposal was to do one traversal instead

yesil: I think I can do it with a reducer logic, let me give it a try.

3ch023: I noticed that the column value is either family or paCode or productCode, is it safe?
shouldn't it be 
family_productCode_productArrangementCode?
where productCode and paCode are optional?

yesil: @3ch023 I'll update the column name and the code after https://jira.corp.adobe.com/browse/ODM-19213 is done.

Roycethan:
@yesil Regression covered. Need 2 approvals for for making it ready for stage. thanks

Axelcureno (approved):

yesil:
@Roycethan the reported discrepancies are not real discrepancies. They are reproducible on main as well.
We need to address the entitlement related logic in a separate ticket/PR.

npeltier (approved):
i guess we _could_ make reduce bit slightly smarter and manage only one config per dimension + locale directly, but not sure it will make better readabilty

yesil:
@npeltier feel free to suggest a code change directly here.

We can give a second touch once product code is enabled, and we can disable the paCode.

3ch023 (approved):

libs/blocks/merch/merch.js
```diff
@@ -198,22 +198,47 @@ export async function fetchCheckoutLinkConfigs(base = '') {
   return fetchCheckoutLinkConfigs.promise;
 }
 
-export async function getCheckoutLinkConfig(productFamily) {
+/**
+ * @param {*} productFamily product family code
```
narcis-radu, line 202: Do we needs the docs?

yesil: not really. I guess you want to save bytes :)

libs/blocks/merch/merch.js
```diff
@@ -198,22 +198,47 @@ export async function fetchCheckoutLinkConfigs(base = '') {
...
-    ({ [NAME_PRODUCT_FAMILY]: mappingProductFamily }) => mappingProductFamily === productFamily,
-  );
-  if (productFamilyConfigs.length === 0) return undefined;
-  const checkoutLinkConfig = productFamilyConfigs.find(
+
+  const {
+    paCodeConfigs,
+    productCodeConfigs,
+    productFamilyConfigs,
+  } = checkoutLinkConfigs.data.reduce((acc, config) => {
```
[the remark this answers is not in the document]
yesil: The suggested change causing a strange eslint warning.
Since it was `checkoutLinkConfigs.data` before with no error, I'll keep the line as is. 
Instead, I initialize it as `checkoutLinkConfigs = { data: [] }` in case of error.

Roycethan:
> Expectation ( As per observed behaviour in con-a with same user credentials:
> User will see "Buy now" with CRM modal pointed

Yes ok tracked here: https://jira.corp.adobe.com/browse/MWPW-156362

PR #2714, opened 2024-08-08
[MWPW-154969] Performance consent

mokimo:
Require performance consent before tracking performance related data. 

Resolves: [MWPW-154969](https://jira.corp.adobe.com/browse/MWPW-154969)

**Test URLs:**
- After: https://performance-consent--milo--mokimo.hlx.page/?martech=off

### DIFF
--- a/libs/utils/logWebVitals.js
+++ b/libs/utils/logWebVitals.js
@@ -93,14 +93,17 @@ export default function webVitals(mep, { delay = 1000, sampleRate = 50 } = {}) {
     return nav.userAgent.includes('Chrome') && nav.vendor.includes('Google');
   };
   if (!isChrome() || Math.random() * 100 > sampleRate) return;
-
+  const getConsent = () => window.adobePrivacy?.activeCookieGroups().indexOf('C0002') !== -1;
   function handleEvent() {
-    const performanceConsent = window.adobePrivacy.activeCookieGroups().indexOf('C0002') !== -1;
-    if (!performanceConsent) return;
+    if (!getConsent()) return;
     const lanaData = {};
     logMepExperiments(lanaData, mep);
     observeCLS(lanaData);
     observeLCP(lanaData, delay);
+  }
+  if (getConsent()) {
+    handleEvent();
+    return;
   }
   window.addEventListener('adobePrivacy:PrivacyConsent', handleEvent);
   window.addEventListener('adobePrivacy:PrivacyCustom', handleEvent);
--- a/libs/utils/logWebVitals.js
+++ b/libs/utils/logWebVitals.js
@@ -87,9 +87,21 @@ function logMepExperiments(lanaData, mep) {
   });
 }
 
-export default function webVitals(mep, { delay = 1000 } = {}) {
-  const lanaData = {};
-  logMepExperiments(lanaData, mep);
-  observeCLS(lanaData);
-  observeLCP(lanaData, delay);
+export default function webVitals(mep, { delay = 1000, sampleRate = 50 } = {}) {
+  const isChrome = () => {
+    const nav = window.navigator;
+    return nav.userAgent.includes('Chrome') && nav.vendor.includes('Google');
+  };
+  if (!isChrome() || Math.random() * 100 > sampleRate) return;
+
+  function handleEvent() {
+    const performanceConsent = window.adobePrivacy.activeCookieGroups().indexOf('C0002') !== -1;
+    if (!performanceConsent) return;
+    const lanaData = {};
+    logMepExperiments(lanaData, mep);
+    observeCLS(lanaData);
+    observeLCP(lanaData, delay);
+  }
+  window.addEventListener('adobePrivacy:PrivacyConsent', handleEvent);
+  window.addEventListener('adobePrivacy:PrivacyCustom', handleEvent);
 }
--- a/libs/utils/logWebVitals.js
+++ b/libs/utils/logWebVitals.js
@@ -105,6 +105,6 @@ export default function webVitals(mep, { delay = 1000, sampleRate = 50 } = {}) {
     handleEvent();
     return;
   }
-  window.addEventListener('adobePrivacy:PrivacyConsent', handleEvent);
-  window.addEventListener('adobePrivacy:PrivacyCustom', handleEvent);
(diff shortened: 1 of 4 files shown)


libs/utils/utils.js
```diff
@@ -1000,7 +1000,9 @@ export function scrollToHashedElement(hash) {
 }
 
 function logPagePerf() {
-  if (getMetadata('pageperf') !== 'on') return;
+  const consent = document.cookie.split('OptanonConsent')[1];
```
narcis-radu, line 1003:
```suggestion
  const consent = document.cookie.split('OptanonConsent=')[1];
```

libs/utils/utils.js
```diff
@@ -1000,7 +1000,9 @@ export function scrollToHashedElement(hash) {
...
+  const performanceConsent = consent?.includes(encodeURIComponent('C0002:1')) || consent?.includes('C0002:1');
```
narcis-radu, line 1004:
```suggestion
  const performanceGroup = 'C0002:1';
  const performanceConsent = consent?.includes(encodeURIComponent(performanceGroup) || consent?.includes(performanceGroup);
```

libs/utils/utils.js
```diff
@@ -1000,7 +1000,10 @@ export function scrollToHashedElement(hash) {
 }
 
 function logPagePerf() {
-  if (getMetadata('pageperf') !== 'on') return;
```
overmyheadandbody, line 1003: I still wonder what's the ratio of pages where this will be set vs not. If we have more pages where we're not using it, I'd advocate we return ASAP without any sort of additional check.

libs/utils/utils.js
```diff
@@ -1000,7 +1000,10 @@ export function scrollToHashedElement(hash) {
...
+  const consent = document.cookie.split('OptanonConsent=')[1];
+  const performanceGroup = 'C0002:1';
+  const performanceConsent = consent?.includes(encodeURIComponent(performanceGroup)) || consent?.includes(performanceGroup);
```
overmyheadandbody:
```suggestion
  const performanceConsent = consent?.includes(encodeURIComponent(performanceGroup))
    || consent?.includes(performanceGroup);
```

libs/utils/utils.js
```diff
@@ -1000,7 +1000,11 @@ export function scrollToHashedElement(hash) {
 }
 
 function logPagePerf() {
-  if (getMetadata('pageperf') !== 'on') return;
+  const consent = document.cookie.split('OptanonConsent=')[1];
+  const performanceGroup = 'C0002:1';
+  const performanceConsent = consent?.includes(encodeURIComponent(performanceGroup))
+  || consent?.includes(performanceGroup);
```
overmyheadandbody, line 1006:
```suggestion
    || consent?.includes(performanceGroup);
```

narcis-radu (approved):

overmyheadandbody (approved):

robert-bogos (approved):

libs/utils/logWebVitals.js
```diff
@@ -87,9 +87,21 @@ function logMepExperiments(lanaData, mep) {
...
+export default function webVitals(mep, { delay = 1000, sampleRate = 50 } = {}) {
+  const isChrome = () => {
+    const nav = window.navigator;
+    return nav.userAgent.includes('Chrome') && nav.vendor.includes('Google');
+  };
+  if (!isChrome() || Math.random() * 100 > sampleRate) return;
+
+  function handleEvent() {
+    const performanceConsent = window.adobePrivacy.activeCookieGroups().indexOf('C0002') !== -1;
+    if (!performanceConsent) return;
```
overmyheadandbody, line 99: Similar comment to my previous one, why define functions and event listeners when they might not be used? I'd suggest returning early
```suggestion
  const performanceConsent = window.adobePrivacy.activeCookieGroups().indexOf('C0002') !== -1;
  if (!performanceConsent) return;

  function handleEvent() {
    const lanaData = {};
    // ...
```

libs/utils/utils.js
```diff
@@ -1039,7 +1027,13 @@ export async function loadDeferred(area, blocks, config) {
     sampleRUM.observe(area.querySelectorAll('picture > img'));
   });
 
-  logPagePerf();
+  if (getMetadata('pageperf') === 'on') {
+    import('./logWebVitals.js')
+      .then((mod) => mod.default(getConfig().mep, {
+        delay: getMetadata('pageperf-delay') || 1000,
+        sampleRate: parseInt(getMetadata('pageperf-rate'), 10) || 50,
```
overmyheadandbody, line 1034: Nit: we already have reasonable defaults in the method itself that match these ones, is there any worth in duplicating them here as well?

overmyheadandbody (approved):

libs/utils/logWebVitals.js
```diff
@@ -87,9 +87,21 @@ function logMepExperiments(lanaData, mep) {
...
-  observeLCP(lanaData, delay);
+export default function webVitals(mep, { delay = 1000, sampleRate = 50 } = {}) {
+  const isChrome = () => {
+    const nav = window.navigator;
+    return nav.userAgent.includes('Chrome') && nav.vendor.includes('Google');
+  };
+  if (!isChrome() || Math.random() * 100 > sampleRate) return;
+
+  function handleEvent() {
+    const performanceConsent = window.adobePrivacy.activeCookieGroups().indexOf('C0002') !== -1;
```
narcis-radu, line 98:
```suggestion
    const performanceConsent = window.adobePrivacy?.activeCookieGroups().indexOf('C0002') !== -1;
```
we should ensure the code is not failing when privacy is blocked

libs/utils/logWebVitals.js
narcis-radu: `adobePrivacy` object will be made available later in the waterfall, while the needs to execute sooner. I think the approach is correct.

mokimo: I did notice we load this quite late, so @overmyheadandbody has a point in the sense of - maybe privacy is already available in which case we would have missed the events.
I added support for this case as well

overmyheadandbody (approved):

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-156343

PR #2721, opened 2024-08-09
Table review

overmyheadandbody:
This addresses multiple table issues in an effort to bring it closer to the Consonant specs. The epic mentioned below contains all the stories relevant to these changes, as well as a few outstanding QA tasks. Here's a very short summary of a few things that have been addressed:

- backgrounds are now supported;
- border radii are more consistent across viewports;
- pricing aligns correctly at the bottom if defined as such;
- content flows correctly inside the cells;
- tooltips are no longer cut off on mobile devices;
- left and center alignment definition is now respected throughout the cells;
- highlights with additional styling (bold/italics) are no longer stacking on mobile;
- section expansion on mobile works without having to switch to a different column;
- sticky functionality works on mobile after switching to a different column;
- tables now expand to their grid container that can limit it to 6/8/10 columns;
- checkmarks have been updated;
- and many more.

A [Figma file](https://www.figma.com/design/l9pPVFggw1wkvmJzBOQwkB/Tables?node-id=526-19058&t=0dWt3ydQPkT8uOgX-0) is also available, detailing some of the required changes.

Any page containing tables can be tested using this feature branch. Simply:

- on a Milo preview URL (.hlx.page), replace `main` with `table-review`. `https://main--milo--adobecom/path/to/my/table` becomes  `https://table-review--milo--adobecom/path/to/my/table`;
- on a Milo consumer preview URL (.hlx.page), append a `milolibs=table-review` query parameter. `https://main--cc--adobecom/path/to/my/table` becomes `https://main--cc--adobecom/path/to/my/table?milolibs=table-review`.

| Before | After |
| Screenshot 2024-08-09 at 17 20 19 | Screenshot 2024-08-09 at 17 20 09 |

Epic: [MWPW-153313](https://jira.corp.adobe.com/browse/MWPW-153313)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ramuntea/table-tiger-team-mwpw-152305?martech=off
- After: https://table-review--milo--adobecom.hlx.page/drafts/ramuntea/table-tiger-team-mwpw-152305?martech=off


libs/blocks/table/table.css
```diff
@@ -545,20 +575,56 @@ header.global-navigation {
...
+  [dir="rtl"] .section-row-title .milo-tooltip:is(.top, .bottom, .left, .right)::before {
+    left: 100%;
+    margin: 0 0 0 12px;
+    right: initial;
+  }
+
+  [dir="rtl"] .section-row-title .milo-tooltip:is(.top, .bottom, .left, .right)::after {
+    border-color: transparent #0469E3 transparent transparent;
+    left: 100%;
+    margin: 0 0 0 -4px;
```
mokimo, line 604: -8 & -4 px seem a little weird

libs/blocks/table/table.js
```diff
@@ -533,9 +509,20 @@ export default function init(el) {
...
     });
+
+    isDecorated = true;
   };
 
   window.addEventListener(MILO_EVENTS.DEFERRED, () => {
     handleTable();
   }, true);
+
+  const observer = new window.IntersectionObserver((entries) => {
```
mokimo, line 520: Utils has an intersection observer wrapper

overmyheadandbody: Hey @mokimo! Thanks for the note. A few more details on this: this particular intersection observer is indeed replaceable with the one from `utils.js`. However, this file has another one [here](https://github.com/adobecom/milo/blob/b048123a0207264b5af083a55f74ed13a0df1737/libs/blocks/table/table.js#L306-L309). That one, unfortunately, can't be replaced with the one from `utils.js`; that is because it depends on being able to tell whether the entry is currently intersecting or not to toggle a class. The callback for `createIntersectionObserver` [here](https://github.com/adobecom/milo/blob/b048123a0207264b5af083a55f74ed13a0df1737/libs/utils/utils.js#L1233-L1235) is only called if the entry is intersecting, so the previously-mentioned toggle will never occur.

With that in mind, what do you choose between:
1. importing the `createIntersectionObserver` method from `utils.js` and using it in one place, while the other place keeps using a regular intersection observer? It could be argued that we're already importing `utils.js`, so there's no overhead in using the method it supplies where we can;
2. we keep things as they are now keep it consistent in the `table.js` file.

mokimo: Let's keep things consistent and the way they are right now

libs/blocks/table/table.css
```diff
@@ -1,29 +1,54 @@
...
 
-.table,
-.table.merch {
   max-width: 1200px;
   margin: auto;
   padding: 20px 0;
   border-color: var(--border-color);
 }
 
+.section[class *= "grid-width-"] .table {
```
robert-bogos, line 15: Any reason for the whitespace before and after the *= 🤔
```suggestion
.section[class*="grid-width-"] .table {
```

libs/blocks/table/table.js
```diff
@@ -316,16 +296,16 @@ function handleScrollEffect(table) {
   } else {
     headingRow.classList.add('top-border-transparent');
   }
-  headingRow.style.top = `${gnavHeight + (highlightRow ? highlightRow.offsetHeight : 0)}px`;
+  const topOffset = gnavHeight + (highlightRow ? highlightRow.offsetHeight : 0);
```
robert-bogos, line 299: nit: I think this could be shorter
```suggestion
  const topOffset = gnavHeight + (highlightRow?.offsetHeight || 0);
```

robert-bogos (approved):

SilviuLCF (approved):
Ready to stage

Testing details https://table-review--milo--adobecom.hlx.page/drafts/ramuntea/table-tiger-team-mwpw-152305?martech=off

PR #2723, opened 2024-08-12
MWPW-156672: refactor commerce & optimize pandora dependencies

3ch023:
This PR is a follow up of https://github.com/adobecom/milo/pull/2725 and:

- replaces pandora dependencies with a fetch implementation in wcs.js
- fetch use credentials:omit option to skip cookies, they are not needed for WCS
- set https://www.adobe.com/web_commerce_artifact as a default endpoint for all milo envs
- allows to access https://www.stage.adobe.com/web_commerce_artifact_stage with ?commerce.env=stage parameter (only on lover envs)
- removes extra commerce.js file
- always trigger tests on commerce before building 

Resolves: https://jira.corp.adobe.com/browse/MWPW-156672
Original discussion: https://git.corp.adobe.com/orgs/wcms/discussions/38

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.live/drafts/mariia/pr/price-cta?martech=off
- After: https://killwcs--milo--adobecom.hlx.live/drafts/mariia/pr/price-cta?martech=off


mokimo (approved):
fancy, is it also planned to support subdomains? E.g. same domain requests from business.adobe.com

libs/features/mas/commerce/src/wcs.js
```diff
@@ -63,29 +43,35 @@ export function Wcs({ settings }) {
...
+          url.searchParams.set('landscape', env === Env.STAGE ? 'ALL' : settings.landscape);
+          url.searchParams.set('api_key', apiKey);
+          if (options.promotionCode) {
+            url.searchParams.set('promotion_code', options.promotionCode);
+          }
+          if (options.currency) {
+            url.searchParams.set('currency', options.currency);
+          }
+          
+          const response = await fetch(url.toString());
```
yesil, line 69: we need to check this line against network issues.

3ch023:
todo: test on adobe-students

3ch023:
> fancy, is it also planned to support subdomains? E.g. same domain requests from business.adobe.com

from business.adobe.com most likely browser will initiate a preflight, since its a different domain, but without testing I can't say for sure.
Iirc right now bacom is not using commerce?

libs/features/mas/commerce/src/settings.js
```diff
@@ -135,15 +133,18 @@ function getSettings(config = {}) {
...
-                  getParameter(PARAM_ENV, commerce, { metadata: false }),
-                  HostEnv,
-                  HostEnv.PROD,
-              );
-    const env = hostEnv === HostEnv.STAGE ? Env.STAGE : Env.PRODUCTION;
+    const lowHostEnv = ['local', 'stage'].includes(config.env?.name);
+    let env = Env.PRODUCTION;
+    let wcsURL = 'https://www.adobe.com/web_commerce_artifact';
+    if (lowHostEnv) {
+      wcsURL = 'https://www.stage.adobe.com/web_commerce_artifact';
```
npeltier, line 140: is this because authority of lower env is stage? that would be a good reason to do this.

3ch023: not really no, it works also with `www.adobe.com`
i think our authority theory is wrong.

but I did use www.stage.adobe.com for lower envs just to be consistent with how rest of milo works, i think it would be more confusing for Devs to see 'prod' commerce URL on their local while they see stage one on stage.

overmyheadandbody, line 140: Nit: maybe string repetition could be avoided by determining the environment and then using a template literal to form the URL? Or maybe even a simple `replace`?

yesil (commented):
these are good changes that are needed, but I find it a bit risky to combine it with CDN change.

overmyheadandbody (commented):
No local issues sounds nice :) I'll let the MAS people approve this, as I don't have sufficient context.

npeltier (approved):
looks good to me, let's still discuss with both @yesil & @afmicka probably around here wether we split this in 2 or keep it unified as one big change for WCS

3ch023:
FYI tested adobe-students domain and it will keep working fine with CDN change.
image

However I agree with Ilyas, it feels also too risky to me and we can't afford even a small CSO with this.
I am going to split the PR in two.
First will be CDN change, 2 lines only, we will give it some time and then this PR will follow after its all good.

3ch023:
smaller PR: https://github.com/adobecom/milo/pull/2725
this PR should be a follow up of the one above.

libs/features/mas/commerce/src/wcs.js
3ch023: in case of a failed request it falls into a catch clause like before, could you specify what do you mean?

libs/features/mas/commerce/src/settings.js
3ch023: refactored a bit

libs/features/mas/commerce/src/wcs.js
```diff
@@ -99,25 +85,29 @@ export function Wcs({ settings }) {
...
-            }
+                ),
+            );
+          } else {
+            const error = await response.text();
+            log.error(`WCS request failed. Status: ${response.status}, message: ${error}`, options);
+            message = ERROR_MESSAGE_BAD_REQUEST;
+          }
+        } catch (e) {
+          message = 'Failed to fetch WCS offer.';
```
yesil, line 109: maybe use a constant as the line 106 above ?

libs/features/mas/commerce/src/wcs.js
yesil: When I block the wcs endpoint (screenshot 1), in the before URL, it logs with ERROR_MESSAGE_BAD_REQUEST whereas in the after URL, it doesn't fall in that clause, but is captured by the global catch you added.
So, I wonder what makes it follow a different path in your PR.

Screenshot 1
image
Screenshot 2
image
Screenshot 3
image

3ch023:
@yesil 
>whereas in the after URL, it doesn't fall in that clause, but is captured by the global catch you added.
i didn't add a global catch, its the same catch as before.
but before with the wcs client we couldn't make a difference of completely failed network request (e.g. blocked in devtools) and a 'successfull' network request returning response other than 2xx - example here https://killwcs--milo--adobecom.hlx.page/drafts/mariia/cta-commerce (I used unexisting OSI to simulate it)

I did change the log message, which I agree maybe is not the best idea right now. Not without a discussion at least, so I will revert it back to usual and in both cases will log 'Bad WCS request' as before.
When we decide that we want to separate these 2 cases we can do a separate PR.

yesil:
@3ch023 thanks for the clarification. In this case, let's keep the error message as you changed, it is better.

yesil (dismissed):

libs/features/mas/commerce/src/settings.js
3ch023: fixed to be www.adobe.com

libs/features/mas/commerce/src/wcs.js
3ch023: addressed

libs/features/mas/commerce/src/wcs.js
3ch023: addressed

3ch023:
I found an issue so removing the 'Ready for Stage' label

afmicka:
> I found an issue so removing the 'Ready for Stage' label

@3ch023 test case covering the issue you have found is now merged. Nala will be failing on this branch until it is resolved

3ch023:
added some changes on this PR so will need a re-review:

- there was an issue with UK - if the language is undefined we shouldn't add this request param (MULT/EN)
- added back few packages because of checkout builder depending on them.. unfortunately we need them for now to avoid corp artifactory in package-lock
- added credentials:omit to not send cookies over to WCS
- added back special handing for [www.stage.adobe.com](http://www.stage.adobe.com/) because it does need same domain to not trigger OPTIONS request, see OPTIONS here: https://www.stage.adobe.com/products/catalog.html

mirafedas (approved):

libs/features/mas/package-lock.json
```diff
@@ -69,17 +69,12 @@
...
         "@pandora/commerce-checkout-url-builder": "file:./internal/commerce-checkout-url-builder-1.6.0.tgz",
         "@pandora/data-models-odm": "file:./internal/data-models-odm-0.5.4.tgz",
         "@pandora/data-source-utils": "file:./internal/data-source-utils-0.3.1.tgz",
-        "@pandora/data-source-wcs": "file:./internal/data-source-wcs-0.2.8.tgz",
         "@pandora/fetch": "file:./internal/fetch-1.3.4.tgz",
-        "@pandora/logger": "file:./internal/logger-1.3.0.tgz",
-        "@pandora/react-auth-provider": "file:./internal/react-auth-provider-1.2.1.tgz",
-        "@pandora/react-env-provider": "file:./internal/react-env-provider-1.2.2.tgz",
-        "@pandora/react-fetch-provider": "file:./internal/react-fetch-provider-1.2.2.tgz"
+        "@pandora/logger": "file:./internal/logger-1.3.0.tgz"
```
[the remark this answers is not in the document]
3ch023: yes, the checkout builder is pulling the data source package, which pulls logger etc

PR #2487, opened 2024-06-19
MWPW-147468 - Delayed Activation of Blade Animations on Ps Product Page

sivasadobe:
* This feature ensures that videos automatically play when they come into view within the viewport and pause when they leave the viewport.
* Added logic to handle video play/pause based on the viewport intersection ratio( ≥0.8 ).

Resolves: [MWPW-147468](https://jira.corp.adobe.com/browse/MWPW-147468)

**Steps for Authors:**

- Edit the video URL to include `#viewportplay`.

### Test URLs

**MILO**
- Before: https://main--milo--sivasadobe.hlx.page/drafts/siva/products/videoplay-onviewport?martech=off
- After: https://mwpw-147468--milo--sivasadobe.hlx.page/drafts/siva/products/videoplay-onviewport?martech=off

**CC**
- Before: https://main--cc--adobecom.hlx.page/drafts/siva/products/photoshop?martech=off
- After: https://main--cc--adobecom.hlx.page/drafts/sivas/products/photoshop?martech=off&milolibs=mwpw-147468--milo--sivasadobe

PR #1106, opened 2023-08-11
Play video on hover

Ruchika4:
* Play video (franklin bot uploaded video) when cursor in and stop when cursor out
* hover play can only be used with hash value autoplay1 (play video once when in viewport. If used alone or with autoplay it'll not add any event
* To add hover play to the video append #_hoverplay to the video url

Resolves: [MWPW-129910](https://jira.corp.adobe.com/browse/MWPW-129910)

**Test URLs:**
- Before: https://main--milo--ruchika4.hlx.page/drafts/ruchika/videoplay-onhover?martech=off
- After: https://video-hoverplay--milo--ruchika4.hlx.page/drafts/ruchika/videoplay-onhover?martech=off

PR #2724, opened 2024-08-12
MWPW-155412: Fix video CLS netting +10 lighthouse points

mokimo:
### Description
Fixes 0.2 marquee CLS (+10LH points) for videos. The CLS was caused by only adding the video element & source to the page AFTER text has already been rendered next to it, as the intersection observer can only render things in the event cycle. We now only add the source of a video within the scope of the intersection observer, to keep the benefits of not loading videos that aren't displayed, however also dont cause any CLS.

### Test instructions
Few related PRs/features that should still work afterwards...
- https://github.com/adobecom/milo/pull/1106
- https://github.com/adobecom/milo/pull/2487
- https://github.com/adobecom/milo/pull/1343
And the CLS should be gone. 

Resolves: [MWPW-155412](https://jira.corp.adobe.com/browse/MWPW-155412)

**Test URLs:**
Onhover play
- Before: https://main--milo--mokimo.hlx.page/drafts/ruchika/videoplay-onhover?martech=off
- After: https://cls-marquee--milo--mokimo.hlx.page/drafts/ruchika/videoplay-onhover?martech=off

Onviewport play
- Before: https://main--milo--mokimo.hlx.page/drafts/siva/products/videoplay-onviewport?martech=off
- After: https://cls-marquee--milo--mokimo.hlx.page/drafts/siva/products/videoplay-onviewport?martech=off

Only desktop (mobile should not load the video for perf reasons)
- Before: https://main--milo--mokimo.hlx.page/drafts/seanchoi/marquee-video?martech=off
- After: https://cls-marquee--milo--mokimo.hlx.page/drafts/seanchoi/marquee-video?martech=off

CLS
- Before: https://main--milo--adobecom.aem.page/drafts/osahin/magento-commerce-1
- After: https://cls-marquee--milo--mokimo.aem.page/drafts/osahin/magento-commerce-1

### DIFF
--- a/libs/blocks/adobetv/adobetv.js
+++ b/libs/blocks/adobetv/adobetv.js
@@ -1,6 +1,7 @@
 import { turnAnchorIntoVideo } from '../../utils/decorate.js';
 
-const loadAdobeTv = (a) => {
+export default function init(a) {
+  a.classList.add('hide-video');
   const bgBlocks = ['aside', 'marquee', 'hero-marquee'];
   if (a.href.includes('.mp4') && bgBlocks.some((b) => a.closest(`.${b}`))) {
     a.classList.add('hide');
@@ -18,9 +19,4 @@ const loadAdobeTv = (a) => {
     a.insertAdjacentHTML('afterend', embed);
     a.remove();
   }
-};
-
-export default function init(a) {
-  a.classList.add('hide-video');
-  loadAdobeTv(a);
 }
--- a/libs/blocks/adobetv/adobetv.js
+++ b/libs/blocks/adobetv/adobetv.js
@@ -7,11 +7,11 @@ const loadAdobeTv = (a) => {
   const bgBlocks = ['aside', 'marquee', 'hero-marquee'];
   if (a.href.includes('.mp4') && bgBlocks.some((b) => a.closest(`.${b}`))) {
     a.classList.add('hide');
+    const parentElement = a.parentNode;
+    if (!parentElement) return;
     const { href, hash, dataset } = a;
     const attrs = getVideoAttrs(hash || 'autoplay', dataset);
     const video = `<video ${attrs}></video>`;
-    const parentElement = a.parentNode;
-    if (!parentElement) return;
     a.insertAdjacentHTML('afterend', video);
     createIntersectionObserver({
       el: parentElement,
--- a/libs/blocks/adobetv/adobetv.js
+++ b/libs/blocks/adobetv/adobetv.js
@@ -1,4 +1,4 @@
-import { createIntersectionObserver } from '../../utils/utils.js';
+import { createIntersectionObserver, createTag } from '../../utils/utils.js';
 import { applyHoverPlay, getVideoAttrs } from '../../utils/decorate.js';
 
 const ROOT_MARGIN = 1000;
@@ -9,11 +9,19 @@ const loadAdobeTv = (a) => {
     a.classList.add('hide');
     const { href, hash, dataset } = a;
     const attrs = getVideoAttrs(hash || 'autoplay', dataset);
-    const video = `<video ${attrs}>
-          <source src="${href}" type="video/mp4" />
-        </video>`;
-    if (!a.parentNode) return;
+    const video = `<video ${attrs}></video>`;
+    const parentElement = a.parentNode;
+    if (!parentElement) return;
     a.insertAdjacentHTML('afterend', video);
+    createIntersectionObserver({
+      el: parentElement,
+      options: { rootMargin: `${ROOT_MARGIN}px` },
+      callback: () => {
+        parentElement
+          .querySelector('video')
+          .appendChild(createTag('source', { src: href, type: 'video/mp4' }));
+      },
+    });
     const videoElem = document.body.querySelector(`source[src="${href}"]`)?.parentElement;
     applyHoverPlay(videoElem);
     a.remove();
@@ -29,13 +37,5 @@ const loadAdobeTv = (a) => {
 
 export default function init(a) {
   a.classList.add('hide-video');
-  if (a.textContent.includes('no-lazy')) {
-    loadAdobeTv(a);
-  } else {
-    createIntersectionObserver({
-      el: a,
-      options: { rootMargin: `${ROOT_MARGIN}px` },
-      callback: loadAdobeTv,
-    });
-  }
+  loadAdobeTv(a);
 }
--- a/libs/blocks/adobetv/adobetv.js
+++ b/libs/blocks/adobetv/adobetv.js
@@ -1,30 +1,15 @@
-import { createIntersectionObserver, createTag } from '../../utils/utils.js';
-import { applyHoverPlay, getVideoAttrs } from '../../utils/decorate.js';
-
-const ROOT_MARGIN = 1000;
+import { turnAnchorIntoVideo } from '../../utils/decorate.js';
 
 const loadAdobeTv = (a) => {
   const bgBlocks = ['aside', 'marquee', 'hero-marquee'];
   if (a.href.includes('.mp4') && bgBlocks.some((b) => a.closest(`.${b}`))) {
     a.classList.add('hide');
-    const parentElement = a.parentNode;
-    if (!parentElement) return;
-    const { href, hash, dataset } = a;
-    const attrs = getVideoAttrs(hash || 'autoplay', dataset);
-    const video = `<video ${attrs}></video>`;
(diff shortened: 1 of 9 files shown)


mokimo:
You can easily see the CLS using the [web vitals extension](https://chromewebstore.google.com/detail/web-vitals/ahfhijdlegdabablpippeagghigmibma?hl=en)
Screenshot 2024-08-12 at 17 09 08
Screenshot 2024-08-12 at 17 08 33

Also, here's more context on the specific CLS that was happening
https://github.com/user-attachments/assets/cd85b046-1420-4700-aa83-b9ac787702e9

libs/blocks/adobetv/adobetv.js
```diff
@@ -9,11 +9,19 @@ const loadAdobeTv = (a) => {
     a.classList.add('hide');
     const { href, hash, dataset } = a;
     const attrs = getVideoAttrs(hash || 'autoplay', dataset);
-    const video = `<video ${attrs}>
-          <source src="${href}" type="video/mp4" />
-        </video>`;
-    if (!a.parentNode) return;
+    const video = `<video ${attrs}></video>`;
+    const parentElement = a.parentNode;
+    if (!parentElement) return;
```
overmyheadandbody, line 14: Why define the video if it's possible we'll need to return early anyway? Maybe even earlier than this suggestion
```suggestion
    const parentElement = a.parentNode;
    if (!parentElement) return;
    const video = `<video ${attrs}></video>`;
```

libs/blocks/adobetv/adobetv.js
```diff
@@ -29,13 +37,5 @@ const loadAdobeTv = (a) => {
 
 export default function init(a) {
   a.classList.add('hide-video');
-  if (a.textContent.includes('no-lazy')) {
```
overmyheadandbody, line 32: Is there some documentation in need of updating once there's no longer a need for `no-lazy`?

overmyheadandbody (commented):
Looking at the adapted methods, they look awfully alike. Is there any way we could use `decorate.js` to avoid repetition?

PS: Good PR title, sounds like a tabloid, "You won't believe how much CLS this one change can reduce" 😅

libs/blocks/adobetv/adobetv.js
mokimo: There never was any docs about it. @seanchoi-dev do you know if this was exclusively used for unit tests? That's what it looked like to me

libs/blocks/adobetv/adobetv.js
```diff
@@ -7,13 +7,21 @@ const loadAdobeTv = (a) => {
...
+    if (!parentElement) return;
     const { href, hash, dataset } = a;
     const attrs = getVideoAttrs(hash || 'autoplay', dataset);
-    const video = `<video ${attrs}>
-          <source src="${href}" type="video/mp4" />
-        </video>`;
-    if (!a.parentNode) return;
+    const video = `<video ${attrs}></video>`;
     a.insertAdjacentHTML('afterend', video);
+    createIntersectionObserver({
```
mokimo, line 16: > Is there any way we could use decorate.js to avoid repetition?

@overmyheadandbody While we could move it into decorate, it's used by more than just the video blocks. IMO it already contains methods it should not and it's "growing" instead of `utils.js` ...

I don't want to shift the utils.js problem into another file, so IMO the better call is to live with the duplication. Both of the blocks seem similar enough to me, so that they could be the same thing 🤷 But that's too late to change

overmyheadandbody: If the video files are the only ones using then, would it be then worth having a video utils file that has the common methods and then take them out of `decorate.js`? I agree that we don't want to needlessly bloat common files, but I also think that having to duplicate functionality is not maintainable. If your tests show that loading yet another video-utils.js file or whatever comes with some performance penalties, then we can just go with duplication, but I'd like to get some insights firts.

overmyheadandbody (commented):
There are a few checks failing, could you check them out before we do another round of reviews, please?

libs/blocks/adobetv/adobetv.js
mokimo: I couldn't really split the files, however after testing things around, it's a lot cleaner moving the logic in decorate. Those differences are hard to maintain and already lead to a bug in the past by forgetting to set sth up correctly.

libs/utils/decorate.js
```diff
@@ -1,4 +1,6 @@
-import { createTag } from './utils.js';
+import { createTag, createIntersectionObserver } from './utils.js';
+
+const ROOT_MARGIN = 1000;
```
narcis-radu, line 3: I am thinking whether we should have this constant here or in the `turnAnchorIntoVideo` method.

libs/utils/decorate.js
```diff
@@ -260,3 +262,25 @@ export function applyInViewPortPlay(video) {
...
+  const attrs = getVideoAttrs(hash, dataset);
+  const video = `<video ${attrs}></video>`;
+  anchorTag.insertAdjacentHTML('afterend', video);
+  createIntersectionObserver({
+    el: parentElement,
+    options: { rootMargin: `${ROOT_MARGIN}px` },
+    callback: () => {
+      parentElement
+        .querySelector('video')
+        .appendChild(
```
narcis-radu, line 277: I would use `?` for chaining.

libs/utils/decorate.js
```diff
@@ -260,3 +262,25 @@ export function applyInViewPortPlay(video) {
...
 }
+
+export function turnAnchorIntoVideo({ hash, src, anchorTag }) {
+  const { dataset, parentElement } = anchorTag;
+  const attrs = getVideoAttrs(hash, dataset);
+  const video = `<video ${attrs}></video>`;
+  anchorTag.insertAdjacentHTML('afterend', video);
+  createIntersectionObserver({
+    el: parentElement,
+    options: { rootMargin: `${ROOT_MARGIN}px` },
```
overmyheadandbody: Seems `ROOT_MARGIN` is only used here, does it make sense to keep the value decoupled from the logic, since it's never re-used?

libs/utils/decorate.js
```diff
@@ -260,3 +262,25 @@ export function applyInViewPortPlay(video) {
...
+    callback: () => {
+      parentElement
+        .querySelector('video')
+        .appendChild(
+          createTag('source', { src, type: 'video/mp4' }),
+        );
+    },
+  });
+  const videoEl = parentElement.querySelector('video');
```
overmyheadandbody, line 280: Could we move this above the intersection observer and use `videoEl` on L275 instead of querying twice?

libs/utils/decorate.js
mokimo: Hmm we do set the video element ourselves, so it should always exist. However given this is within an intersection observer, an outside function might manipulate the DOM and we'd have errors.

Added 👍

libs/utils/decorate.js
mokimo: Moved into the fn directly

libs/utils/decorate.js
mokimo: Moved into the fn directly

libs/utils/decorate.js
mokimo: Good idea, done 👍

libs/utils/decorate.js
```diff
@@ -260,3 +260,21 @@ export function applyInViewPortPlay(video) {
...
+
+export function turnAnchorIntoVideo({ hash, src, anchorTag }) {
+  const { dataset, parentElement } = anchorTag;
+  const attrs = getVideoAttrs(hash, dataset);
+  const video = `<video ${attrs}></video>`;
+  anchorTag.insertAdjacentHTML('afterend', video);
+  const videoEl = parentElement.querySelector('video');
+  createIntersectionObserver({
+    el: parentElement,
+    options: { rootMargin: '1000px' },
```
overmyheadandbody, line 272: Just one last curiosity: I know this value was here to begin with, but are we sure it's the right one for our performance needs? Could there be situations where we prematurely load videos? Could the value be lower?

overmyheadandbody (approved):

libs/utils/decorate.js
mokimo: If you have a better suggestion, I'm happy to adapt it 😁 It was initially introduced in [this PR](https://github.com/adobecom/milo/pull/1343) by @seanchoi-dev 

I don't see any context // investigation on how that value got to be. Should we use sth similar to the footer -> 300px?

SilviuLCF (approved):
Verified

PR #2726, opened 2024-08-12
MWPW-155529 - Notification design parity

elan-tbx:
* This PR accounts for various edge-cases in Ribbon and Banner Notifications.
* Changed default button sizes and labels to align with other Milo blocks.
* Added support for product lockups in Notifications.
* Added a guardrail to `decorateBlockText` so "detail" text styles are not applied to product lockups.

Resolves: [MWPW-155529](https://jira.corp.adobe.com/browse/MWPW-155529)

**Test URLs:**

Test Page from ticket reporter:
- Before: https://main--milo--adobecom.hlx.page/drafts/vmorel/red-team/notification/mwpw-155529?martech=off
- After: https://ebartholomew-general-notification-fixes--milo--adobecom.hlx.page/drafts/vmorel/red-team/notification/mwpw-155529?martech=off

Banners
- Before: https://main--milo--adobecom.hlx.page/drafts/ebartholomew/notifications-pr?martech=off
- After: https://ebartholomew-mwpw-155529-notification-parity--milo--adobecom.hlx.page/drafts/ebartholomew/notifications-pr?martech=off

Ribbons
- Before: https://main--milo--adobecom.hlx.page/drafts/ebartholomew/ribbon-notification-pr?martech=off
- After: https://ebartholomew-mwpw-155529-notification-parity--milo--adobecom.hlx.page/drafts/ebartholomew/ribbon-notification-pr?martech=off

### DIFF
--- a/libs/blocks/notification/notification.css
+++ b/libs/blocks/notification/notification.css
@@ -329,7 +329,7 @@
     gap: var(--spacing-s);
   }
 
-  .notification:is(.full-width, .ribbon) .foreground {
+  .notification:is(.max-width-12-desktop, .ribbon) .foreground {
     max-inline-size: var(--full-width);
     margin-inline: var(--grid-margins-width);
   }
@@ -453,11 +453,11 @@
     gap: var(--spacing-m);
   }
 
-  .notification:is(.full-width, .max-width-10-desktop) .foreground {
+  .notification:is(.max-width-12-desktop, .max-width-10-desktop) .foreground {
     inline-size: unset;
   }
 
-  .notification.full-width .foreground {
+  .notification.max-width-12-desktop .foreground {
     gap: var(--spacing-xl);
     margin-inline: var(--grid-margins-width);
   }
@@ -499,7 +499,7 @@
     max-inline-size: var(--max-inline-size-image-10);
   }
 
-  .notification.full-width .foreground .image {
+  .notification.max-width-12-desktop .foreground .image {
     inline-size: var(--inline-size-image-full);
     max-inline-size: var(--max-inline-size-image-full);
   }
--- a/libs/blocks/notification/notification.css
+++ b/libs/blocks/notification/notification.css
@@ -10,10 +10,10 @@
   --inline-size-pill: 85%;
   --border-block-size: 10px;
   --close-size: 20px;
-  --icon-size: 56px;
-  --icon-size-m: 40px;
-  --icon-size-s: 32px;
   --icon-size-xs: 24px;
+  --icon-size-s: 32px;
+  --icon-size-m: 40px;
+  --icon-size: 56px;
   --icon-size-xl: 64px;
   --pill-radius: 16px;
 
--- a/libs/blocks/notification/notification.css
+++ b/libs/blocks/notification/notification.css
@@ -453,22 +453,23 @@
     gap: var(--spacing-m);
   }
 
-  .notification:is(.full-width, .ribbon) .foreground.container {
-    gap: var(--spacing-xl);
-  }
-
   .notification:is(.full-width, .max-width-10-desktop) .foreground.container {
     inline-size: unset;
+  }
+
+  .notification.full-width .foreground.container {
+    gap: var(--spacing-xl);
+    margin-inline: var(--grid-margins-width);
+  }
+
+  .notification.ribbon .foreground.container {
+    gap: var(--spacing-s);
   }
 
   .notification.max-width-10-desktop .foreground.container {
     margin-inline: var(--grid-margins-width-10);
     gap: var(--spacing-l);
     max-inline-size: calc(var(--grid-column-width) * 10);
-  }
-
-  .notification.full-width .foreground.container {
-    margin-inline: var(--grid-margins-width);
   }
 
   .notification .foreground.container > div {
--- a/libs/blocks/notification/notification.css
+++ b/libs/blocks/notification/notification.css
@@ -57,7 +57,7 @@
   margin-block-end: 0;
 }
 
-.notification .foreground.container {
+.notification .foreground {
   display: flex;
   position: relative;
   align-items: flex-start;
@@ -68,32 +68,32 @@
   justify-content: flex-start;
 }
 
-.notification.ribbon .foreground.container {
+.notification.ribbon .foreground {
   inline-size: 100%;
   margin-inline: var(--margin-inline-ribbon);
   margin-block: 0;
   padding-block: var(--spacing-s);
 }
 
-.notification .foreground.container [data-align=center],
-.notification.center .foreground.container,
-.notification.center .foreground.container > * {
+.notification .foreground [data-align=center],
+.notification.center .foreground,
+.notification.center .foreground > * {
   text-align: center;
   justify-content: center;
 }
 
-.notification.pill .foreground.container {
+.notification.pill .foreground {
   padding-inline: var(--spacing-xs) var(--spacing-xxs);
   padding-block: var(--spacing-xs) var(--spacing-xxs);
   margin: 0;
   inline-size: 100%;
 }
 
-.notification.ribbon.xxs-padding .foreground.container {
+.notification.ribbon.xxs-padding .foreground {
   padding-block: var(--spacing-xxs);
 }
 
-.notification.ribbon.xs-padding .foreground.container {
+.notification.ribbon.xs-padding .foreground {
   padding-block: var(--spacing-xs);
 }
 
@@ -136,7 +136,7 @@
   min-block-size: unset;
 }
 
-.notification .foreground.container .text {
+.notification .foreground .text {
   display: flex;
   flex-wrap: wrap;
   max-inline-size: none;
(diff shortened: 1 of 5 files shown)


libs/blocks/notification/notification.css
```diff
@@ -11,9 +11,10 @@
   --border-block-size: 10px;
   --close-size: 20px;
   --icon-size: 56px;
-  --icon-size-s: 40px;
+  --icon-size-m: 40px;
+  --icon-size-s: 32px;
   --icon-size-xs: 24px;
-  --icon-size-l: 64px;
+  --icon-size-xl: 64px;
```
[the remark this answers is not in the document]
elan-tbx: @JasonHowellSlavin really? I've got them scoped to the `.notification` class...

Screenshot 2024-08-12 at 12 53 50 PM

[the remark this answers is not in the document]
JasonHowellSlavin: Ah nice, my bad.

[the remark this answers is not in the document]
JasonHowellSlavin: Ah, nice,  my bad.

overmyheadandbody, line 17: OCD nit: could we have these in order by size to make it easy to follow?

libs/blocks/notification/notification.css
```diff
@@ -48,8 +49,7 @@
   font-weight: unset;
 }
 
-.notification [class*="heading-"] + p,
-.notification.pill p {
+.notification .text p:not(.icon-area, .action-area) {
```
overmyheadandbody, line 52: Such selectors usually increase specificity, making it a bit more difficult to overwrite if needed. Is there any other way to apply the styles with less specific selectors?

libs/blocks/notification/notification.js
```diff
@@ -65,7 +67,8 @@ function getBlockData(el) {
   const variant = variants.find((varClass) => el.classList.contains(varClass)) || defaultVariant;
   const size = sizes.find((sizeClass) => el.classList.contains(sizeClass)) || defaultSize;
   const fontSizes = [...blockConfig[variant][size]];
-  if (el.classList.contains('s-button')) fontSizes.splice(3, 1, 'm');
+  const buttonSize = el.className.match(/([xsml])+-button/);
+  if (buttonSize) fontSizes.splice(3, 1, buttonSize[1]);
```
overmyheadandbody, line 71: Wouldn't this use `s` for `xs-button` and `l` for `xl-button`? Is that the intention?

libs/blocks/notification/notification.js
```diff
@@ -118,15 +142,17 @@ function decorateLayout(el) {
   return foreground;
 }
 
-export default function init(el) {
+const init = async (el) => {
```
overmyheadandbody, line 145: We do use `export default async function` in the codebase, is there anything preventing us from using that pattern here as well?

libs/blocks/notification/notification.js
elan-tbx: D'oh, right you are. I put my `+` in the wrong place 😅

libs/blocks/notification/notification.js
elan-tbx: Not at all, I just picked a random block (believe it was [the Editorial Card](https://github.com/adobecom/milo/blob/38e330d70c9dcf657d59f609c958946b41dc49cb/libs/blocks/editorial-card/editorial-card.js#L122)) and went with the pattern I saw there. Turns out I happened to choose one of the few that uses this pattern by chance 😅

libs/blocks/notification/notification.css
```diff
@@ -442,12 +453,17 @@
     gap: var(--spacing-m);
   }
 
-  .notification:is(.full-width, .ribbon) .foreground.container {
+  .notification:is(.full-width, .max-width-10-desktop) .foreground.container {
```
ryanmparrish, line 456: Nit: Is there support for `max-width-6` or `-8` etc in this block? 

Also is there a reason for `.foreground.container` instead of just `.foreground` the appended classes can cause confusion unless theres a clear B.E.M treatment being used. Just curious if that was necessary?

libs/utils/decorate.js
```diff
@@ -56,7 +56,8 @@ export function decorateBlockText(el, config = ['m', 's', 'm'], type = null) {
     if (headings) {
       headings.forEach((h) => h.classList.add(`heading-${config[0]}`));
       if (config[2]) {
-        headings[0]?.previousElementSibling?.classList.add(`detail-${config[2]}`);
+        const prevSib = headings[0]?.previousElementSibling;
+        prevSib?.classList.toggle(`detail-${config[2]}`, !prevSib.querySelector('picture'));
```
ryanmparrish, line 60: Am I reading this right?  Toggle detail class if header.prevSib is not a picture? 
It's a little confusing and since this a shared func(), has this been tested across the rest of the blocks?

libs/blocks/notification/notification.css
```diff
@@ -442,12 +453,17 @@
...
+    inline-size: unset;
+  }
+
+  .notification.full-width .foreground.container {
```
ryanmparrish, line 460: On the topic of `.full-width`
Was it discussed to rename this variant? 
It just has other global implications that we should consider.
Masonry and ups both have this as a `1fr` backdoor to author content full col-span.

libs/utils/decorate.js
elan-tbx: Nope, it checks if the previous sibling _contains_ a <picture> (what will eventually be given the class of `icon-area`), and only applies `detail` styles if it does not.

This is going to be incredibly important as we move forward with "product lockups", as there are many use-cases where the lockup will be the header's previous sibling, and should absolutely not have detail styles applied to it.

In fact, this is already happening across the board -- it's just gone unnoticed because until now, icon-areas did not contain any text.

I've tested using a copy of your `kitchen-sink` doc, but am happy to do more if need be :)

libs/blocks/notification/notification.css
elan-tbx: 8 is the default for Banners (the default style of notification), 6 is not supported.

I am sorry about the `.container` addition, though -- in all honesty, this came from a time-saving measure as we were scrambling to get this done before shutdown. This all originally came from `aside.css`, and cherry-picking styles from that file made sense to me at the time. I can go through and remove it, but testing will take time.

libs/blocks/notification/notification.css
elan-tbx: Oh interesting -- no, "full-width" was the intention, but I was unaware we were trying to move away from that. Would `.max-width-12` make the most sense in that case, since we're already using similar elsewhere?

ryanmparrish: Yes `max-width-12` would make more sense IMO especially since it's has the auto of `max-width-8` as default foreground container width. If you could decouple .full-width completely and let that only live in the grid functionality, that would be good. Related: I have a future task to support this in the text block but it would need a new variant like `full-width-grid` etc to support regression, but in your case since this is essentially net new, we should nip this in the bud if possible.

libs/blocks/notification/notification.js
```diff
@@ -99,7 +102,27 @@ function decorateFlexible(el) {
...
+  await loadIconography();
+  const icon = lockupArea.querySelector('picture');
+  const content = icon.nextElementSibling || icon.nextSibling;
+  const label = createTag('span', { class: 'lockup-label' }, content.nodeValue || content);
+  if (content.nodeType === 3) {
+    lockupArea.replaceChild(label, content);
+  } else {
+    lockupArea.appendChild(label);
+  }
+  lockupArea.classList.add('lockup-area');
```
ryanmparrish, line 119: Do you have an example of this being used? `lockup-area`. I'm wondering how the iconography.css is scoped to this block as I don't see it being used in the examples? maybe i'm blind. :) 
| Added support for product lockups in Notifications

libs/utils/decorate.js
ryanmparrish: ok, that makes sense, I haven't seen many layouts having the lockup as a header sibling but It does seem a likely scenario. There may be some alignment on the `icon-area` vs 'lockup-area` definitions. 

Initially the `icon-area` was there to support `:icon-name: Label` text... for flex-gap. But some of that logic was also scoped as such for lockup and other icon size scenarios. Any-hoo the hope is moving forward for consistency, we can identify a given blocks true `lockup-area` and the global iconography.css will mostly take care of the rest. This block looks to be a little tricky w/ diff patterns and defaults. It would be good to see those examples w/ 'lockups' here if they're being included. 

I looked at a few test pages and looks good but do post up the link of the mentioned sink if you have one. Thanks

elan-tbx: Sure thing, I just added a bunch of stuff (non-notifications) to make sure the before and after looked the same:

[Before](https://main--milo--adobecom.hlx.page/drafts/ebartholomew/kitchen-sink)

[After](https://ebartholomew-mwpw-155529-notification-parity--milo--adobecom.hlx.page/drafts/ebartholomew/kitchen-sink)

libs/blocks/notification/notification.js
elan-tbx: D'oh, sorry, you're right -- I added it to the description but here's the page Vicky was looking at to test all the new things: https://ebartholomew-general-notification-fixes--milo--adobecom.hlx.page/drafts/vmorel/red-team/notification/mwpw-155529

libs/blocks/notification/notification.css
elan-tbx: Wow, actually, I'm happy to report that just doing a find and replace for `.container` and removing all instances produced zero changes across the board. I triple checked against the exhaustive list of possible variants on the pages below on all breakpoints, but yeah honestly it looks like that worked just fine!

[Banners](http://ebartholomew/MWPW-155529-notification-parity--milo--adobecom.hlx.page/drafts/ebartholomew/notifications?martech=off)

[Ribbons](http://ebartholomew/MWPW-155529-notification-parity--milo--adobecom.hlx.page/drafts/ebartholomew/ribbon-notification?martech=off)

[Pills](http://ebartholomew/MWPW-155529-notification-parity--milo--adobecom.hlx.page/drafts/ebartholomew/pill-notifications?martech=off)

ryanmparrish (approved):
Thanks for addressing those things, looks good.

overmyheadandbody (approved):

Sartxi (approved):

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-155529

PR #2728, opened 2024-08-12
Add support for npm start

hparra:
Test:

- https://mwpw-devex-1--milo--hparra.hlx.page/

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "version": "1.0.0",
   "description": "Website foundation technology.",
   "scripts": {
+    "start": "aem up",
     "test": "npx playwright install && wtr --config ./web-test-runner.config.mjs \"./test/**/*.test.(js|html)\" --node-resolve --port=2000 --coverage --concurrent-browsers 4",
     "test:watch": "npx playwright install && npm test -- --watch",
     "test:file": "npx playwright install && wtr --config ./web-test-runner.config.mjs --node-resolve --port=2000 --coverage",


hparra:
We have some flakey tests on stage. I had to rerun this.

overmyheadandbody (commented):
I don't think it hurts to have this, but at the same time it's quicker to write `aem up` than `npm run start`. With `npm run libs` I get it, one needs to remember how to define a port and which port to use. Also, all Milo projects specify running `aem up` in their README and I'd probably stick to that to avoid any confusion. It can also be that a new developer uses `npm run start` here and will expect the same to work on CC, which doesn't have this enabled, also potentially causing confusion.

If others see value in having this command, then we can probably have it, but we need to ensure all Milo projects are updated so we keep things consistent.

As some final nits - even if trivial, it's useful to have a story for any PR so we can check context in the future. We don't usually implement things unless they're needed and we define those needs in stories. Lastly, just as an FYI, the `zero-impact` label is applied automatically based on certain criteria, I'm wondering whether we should add this file to the exceptions, wdyt @mokimo?

mokimo:
@overmyheadandbody yeah, we should definitely add it

mokimo (commented):
I personally mapped it in my zshconfig and boot milo using https & tls certificates. That way I can just run `milo-s` in the terminal from anywhere and any location.
`alias milo-s="cd ~/repos/milo && aem up --tls-cert ~/repos/milo/.hlx/server.crt --tls-key ~/repos/milo/.hlx/server.key --port 3000"`

I agree with @overmyheadandbody that there should be some knowledge about what aem & the aem-cli is and how it works, so that developers know what they use across projects.

hparra:
Start is special, like test: npm start. Historically, start and test are always defined in node projects. See https://docs.npmjs.com/cli/v6/commands/npm-start

Feel free to close.

mokimo:
> Start is special, like test: npm start. Historically, start and test are always defined in node projects. See https://docs.npmjs.com/cli/v6/commands/npm-start

Yup, I agree @overmyheadandbody in the terms of consistency 'works milo, doesn't work on milo-cc', we'd need to update it for consumers and the boilerplate... hence I'll close this

PR #2732, opened 2024-08-13
MWPW-156447: NVDA not announced check mark on the Product Page

sivasadobe:
* Added an `aria-label` attribute to make SVG icons more accessible

Resolves: [MWPW-156447](https://jira.corp.adobe.com/browse/MWPW-156447)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/siva/156447/icons-example?martech=off
- After: https://mwpw-156447--milo--sivasadobe.hlx.page/drafts/siva/156447/icons-example?martech=off

### DIFF
--- a/libs/features/icons/icons.js
+++ b/libs/features/icons/icons.js
@@ -70,5 +70,7 @@ export default async function loadIcons(icons, config) {
       }
     }
     icon.insertAdjacentHTML('afterbegin', iconSVGs[iconName].outerHTML);
+    const svgTag = icon.querySelector('svg');
+    svgTag.setAttribute('aria-label', iconName.replaceAll('-', ' '));
   });
 }


mokimo (approved):

overmyheadandbody (changes requested):
I don't think this solves the requirement. As shown in https://www.w3.org/TR/SVG11/struct.html#DescriptionAndTitleElements, you would need a `title` element inside the SVG tag to ensure the label is announced, not an `aria-label` attribute.

Specifically for the checkmark, after discussions with the Consonant team, we landed on adding a title, you can have a look over this [commit](https://github.com/adobecom/milo/blob/b048123a0207264b5af083a55f74ed13a0df1737/libs/img/icons/icons.svg?short_path=9169c98#L28). This feature branch is currently being tested and is on the path to stage and then production.

I also don't really agree with doing this via JS, when the icons file itself could be updated with better labels. It's likely that "chevron down" is not that valuable to a user, a better label should be provided. Same goes for `checkmark` and `checkmark-circle`, a user is probably not interested in those details, all they need to know is that it's a checkmark.

sivasadobe:
> I don't think this solves the requirement. As shown in https://www.w3.org/TR/SVG11/struct.html#DescriptionAndTitleElements, you would need a `title` element inside the SVG tag to ensure the label is announced, not an `aria-label` attribute.
> 
> Specifically for the checkmark, after discussions with the Consonant team, we landed on adding a title, you can have a look over this [commit](https://github.com/adobecom/milo/blob/b048123a0207264b5af083a55f74ed13a0df1737/libs/img/icons/icons.svg?short_path=9169c98#L28). This feature branch is currently being tested and is on the path to stage and then production.
> 
> I also don't really agree with doing this via JS, when the icons file itself could be updated with better labels. It's likely that "chevron down" is not that valuable to a user, a better label should be provided. Same goes for `checkmark` and `checkmark-circle`, a user is probably not interested in those details, all they need to know is that it's a checkmark.

Adding `title` tag in SVG file that makes sense. 

 I will close this PR and will update the ticket with this information. Thanks.

PR #2737, opened 2024-08-13
MWPW-155687

Blainegunn:
Resolves: [MWPW-155687](https://jira.corp.adobe.com/browse/MWPW-155687)
Adding placeholders to the window in an object called `mph` (MILO Place Holders).
This will when building blocks that have error states or status messages that need be localized and would be too burdensome to ask Web Publisher to maintain on a page to page basis.

**Test URLs:**
- After: https://mwpw-155687--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/libs/features/placeholders.js
+++ b/libs/features/placeholders.js
@@ -1,4 +1,5 @@
 const fetchedPlaceholders = {};
+window.mph = {};
 
 const getPlaceholdersPath = (config, sheet) => {
   const path = `${config.locale.contentRoot}/placeholders.json`;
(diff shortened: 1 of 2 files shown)


TsayAdobe (approved):

Ruchika4 (approved):

libs/features/placeholders.js
```diff
@@ -20,6 +21,7 @@ const fetchPlaceholders = async (config, sheet) => {
       const placeholders = {};
       json.data.forEach((item) => {
         placeholders[item.key] = item.value;
+        window.mph[item.key] = item.value;
```
mokimo, line 24: why replicate the placeholders, rather than just exposing `fetchedPlaceholders` ?

PR #2739, opened 2024-08-13
MWPW-155720: Dynamic nav ignore values

JasonHowellSlavin:
* Allows for an additional metadata field `dynamic-nav-disable` which will search through metadata on the page and if it's value matches a key value pair put in by authoring, then it will disable the dynamic nav experience.

Resolves: [MWPW-155720](https://jira.corp.adobe.com/browse/MWPW-155720)

**Test URLs:**

- Before: https://main--milo--adobecom.hlx.page/drafts/slavin/dynamic-nav/bacom-sample-on?martech=off
- After: https://dynamic-nav-ignore-values--milo--adobecom.hlx.page/drafts/slavin/dynamic-nav/bacom-sample-on?martech=off


libs/features/dynamic-navigation.js
```diff
@@ -1,6 +1,21 @@
...
+  const dynamicNavDisableValues = getMetadata('dynamic-nav-disable');
+  if (!dynamicNavDisableValues) return false;
+
+  const metadataPairsMap = dynamicNavDisableValues.split(',').map((pair) => pair.split(';'));
+  return metadataPairsMap.reduce((rdx, pair) => {
+    const [metadataKey, metadataContent] = pair;
+    const metaTagContent = getMetadata(metadataKey.toLowerCase());
+    if (!metaTagContent
+      || metaTagContent.toLowerCase() !== metadataContent.toLowerCase()) return rdx;
+    return !rdx;
```
sharmrj, line 13: I'm not certain why there's a reduce here. You seem to be flipping the accumulator every time `metaTagContent` is equal to `metadataContent`.  So if there are two cases where the `metaTagContent` is equal to `metadataContent` the function will produce `False` since it will flip twice. 
Usually `reduce` is used when we need to 'fold' (or 'reduce') a list of values (of some type) into another value (of the same or some other type). This seems less of a reduction than an imperative loop (which imo would probably make more sense to read, even though all loops have a recursive form and all recursive functions can be expressed as a fold ([source](https://www.cs.nott.ac.uk/~pszgmh/fold.pdf)))
Perhaps you were looking, instead, to do `return (!metaTagContent 
      || metaTagContent.toLowerCase() !== metadataContent.toLowerCase()) || rdx`. But in this case you have the built in `Array.prototype.some()` so you can just write 
```javascript
metadataPairsMap.some(([metadataKey, metadataContent]) => {
  const metaTagContent = getMetadata(metadataKey.toLowerCase());
  return (!metaTagContent
      || metaTagContent.toLowerCase() !== metadataContent.toLowerCase())
})
```

test/features/dynamic-nav/dynamicNav.test.js
```diff
@@ -39,10 +39,34 @@ describe('Dynamic nav', () => {
...
+    document.head.innerHTML = await readFile({ path: './mocks/on-ignore-misspelled.html' });
+    const url = dynamicNav('gnav/aem-sites', 'bacom');
+    expect(url).to.equal('some-source-string');
+  });
+
+  it('Returns the sessionStorage url when dynamic nav ignore metadata content is empty', async () => {
+    document.head.innerHTML = await readFile({ path: './mocks/on-ignore-no-content.html' });
+    const url = dynamicNav('gnav/aem-sites', 'bacom');
+    expect(url).to.equal('some-source-string');
+  });
```
sharmrj, line 71: There should probably be a test case for if there are multiple metadata tags that match the data in `dynamic-nav-disable` (see comment above for context)

sharmrj (changes requested):
I'm not sure precisely what the intent here is. I think the function name `checkForIgnoreValues` should also be reconsidered, or a short comment should be added that explains the intent.

libs/features/dynamic-navigation.js
JasonHowellSlavin: Thank you. This makes perfect sense. I knew it was off, but wanted another set of eyes. I had refactored and used reduce, but `some` is much more appropriate. I appreciate it.

test/features/dynamic-nav/dynamicNav.test.js
JasonHowellSlavin: Great point. Added.

JasonHowellSlavin:
Hi @sharmrj, I've made changes you requested/suggested. Please take another look when you get a second.

sharmrj (approved):

Blainegunn (approved):

Brandon32 (approved):

libs/features/dynamic-navigation.js
```diff
@@ -1,6 +1,19 @@
...
 
+function isDynamicNavDisabled() {
+  const dynamicNavDisableValues = getMetadata('dynamic-nav-disable');
+  if (!dynamicNavDisableValues) return false;
+
+  const metadataPairsMap = dynamicNavDisableValues.split(',').map((pair) => pair.split(';'));
+  return metadataPairsMap.some(([metadataKey, metadataContent]) => {
+    const metaTagContent = getMetadata(metadataKey.toLowerCase());
+    return (metaTagContent
+        && metaTagContent.toLowerCase() === metadataContent.toLowerCase());
```
overmyheadandbody, line 11: Nit:
```suggestion
    return metaTagContent?.toLowerCase() === metadataContent.toLowerCase());
```

PR #2741, opened 2024-08-14
Making unav configurable for consuming client

bandana147:
* Making unav version configurable for consuming client's config

Resolves: [MWPW-156207](https://jira.corp.adobe.com/browse/MWPW-156207)

**Test URLs:**
- After: https://gnav--milo--adobecom.hlx.page/?martech=off

**QA**:
- https://adobecom.github.io/nav-consumer/navigation.html?env=stage&navbranch=gnav&unav-version=1.2

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -519,7 +519,7 @@ class Gnav {
       return 'linux';
     };
 
-    const unavVersion = new URLSearchParams(window.location.search).get('unavVersion') || '1.1';
+    const unavVersion = new URLSearchParams(window.location.search).get('unavVersion') || config.unavVersion || '1.1';
     await Promise.all([
       loadScript(`https://${environment}.adobeccstatic.com/unav/${unavVersion}/UniversalNav.js`),
       loadStyles(`https://${environment}.adobeccstatic.com/unav/${unavVersion}/UniversalNav.css`),
--- a/libs/navigation/navigation.js
+++ b/libs/navigation/navigation.js
@@ -55,6 +55,7 @@ export default async function loadBlock(configs, customLib) {
     pathname: `/${locale}`,
     locales: configs.locales || locales,
     contentRoot: authoringPath || footer.authoringPath,
+    unavVersion: header?.unavVersion,
     ...paramConfigs,
   };
   setConfig(clientConfig);


libs/navigation/navigation.js
```diff
@@ -55,6 +55,7 @@ export default async function loadBlock(configs, customLib) {
     pathname: `/${locale}`,
     locales: configs.locales || locales,
     contentRoot: authoringPath || footer.authoringPath,
+    unavVersion: header?.unavVersion,
```
sharmrj, line 58:
```suggestion
    unavVersion: header?.unavVersion ?? '1.1',
```

sharmrj (approved):

libs/navigation/navigation.js
sharmrj: You may want to do that depending on the error handling downstream.

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -519,7 +519,7 @@ class Gnav {
       return 'linux';
     };
 
-    const unavVersion = new URLSearchParams(window.location.search).get('unavVersion') || '1.1';
+    const unavVersion = new URLSearchParams(window.location.search).get('unavVersion') || config.unavVersion || '1.1';
```
mokimo, line 522: What would be the reason for allowing this to be configurable?

bandana147: This was a request from the client in case they needed time to verify their own navigation before upgrading the Unav version. Since we don't have versioning in our standalone code, we provide whatever is in our main branch, which means they don't have an option to opt out of the upgrade.

mokimo: Does that consumer have any special overwrites on top of the base nav? E.g. changes styles that aren't offered by you?

Ideally testing should happen centrally (by tattva, verifying that all consumers work) - and there shouldn't be any cause for concern that they need to do validations themselves, unless they really hacked into the gnav OR the unav has frequent breaking changes, which otherwise can't be rolled out.

If the unav does have a lot of breaking changes, that's a discussion to be had with them to discourage them from that or provide a way of opting into the breaking changes.

mokimo: I don't see a lot of contexts of the request, but I'd further enquire the consumer about his reasoning and trying to make central improvements, such as adding that consumer to a list of consumers to take extra care on for testing.

I'm not blocking the PR in any way, I would just further question on why it's needed, before opening this box-of-pandora - if you expect unav v1.5 for example, and consumers are on v1.3 you might have compatibility issues in the future on your end, when shipping new features

Deva309 (approved):

narcis-radu (changes requested):
All adobe.com and consumers should be on the same version of the universal navigation. As long as they don't have custom overrides (and they shouldn't), then all compatibility should be already tested in Milo.
I don't think we should allow such scenarios, this will be a nightmare to support and will create inconsistencies in the user journey.

bandana147:
Closing this for now as we need more discussion. Thanks @narcis-radu and @mokimo for your feedback.

PR #2690, opened 2024-08-05
MWPW-151392: Sorting button/menu issues on Mobile

mirafedas:
Changed the SWC version to 0.45.0, which includes the fix for the action menu component.

Resolves: [MWPW-151392](https://jira.corp.adobe.com/browse/MWPW-151392)

**Test URLs:**
- Before: https://main--  cc--adobecom.hlx.live/products/catalog/?martech=off&mep=off
- After: https://main--  cc--adobecom.hlx.page/products/catalog?milolibs=mwpw-151392-swc-version--milo--mirafedas&martech=off&mep=off
https://mwpw-151392-swc-version--milo--mirafedas.hlx.page/ - this is for passing the PSI check. As we don't have sidenav block in Milo, only in CC, it's impossible to create a test content page at the moment

PR #2743, opened 2024-08-14
MWPW-155619: Cards overlapping sorting section- IOS (Legacy versions)

mirafedas:
On iOS version 15 and lower there is a different way of managing the vertical stacking context, which puts the menu below the merch card. Adding z-index on the `sp-action-menu` resolves this problem.

Resolves: [MWPW-155619](https://jira.corp.adobe.com/browse/MWPW-155619)

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/products/catalog?martech=off&mep=off
- After: https://main--cc--adobecom.hlx.page/products/catalog?milolibs=mwpw-155619-cards-overlapping-sorting--milo--mirafedas&martech=off&mep=off
Dummy page for the PSI check: https://mwpw-155619-cards-overlapping-sorting--milo--mirafedas.hlx.page/fragments/mirafedas/catalog/default (there is no merch sidenav, so no 'filter' button, and icons are broken, please ignore these issues)


yesil (approved):

libs/deps/mas/merch-card-all.js
yesil, line 1: IMO you don't need to commit this.

libs/deps/mas/merch-card.js
yesil, line 1: IMO you don't need to commit this.

Roycethan:
@mirafedas Fix works on legacy versions. Need one more approval and psi-check pass , before going to stage.

.github/codeql/codeql-config.yml
```diff
@@ -5,3 +5,4 @@ paths-ignore:
   - 'tools/floodgate/**'
   - 'tools/translation/**'
   - node_modules
+  - libs/deps/mas
```
overmyheadandbody, line 8: Why is this ignored? Is it enabled on the MAS side? I'd argue that anything that goes into Milo should follow its standards. If the larger community doesn't agree, then we should consider disabling it for all dependencies, not just this one.

mokimo: MAS had some issues regarding lit creating code that's being flagged, which can't easily be changed / updated

mokimo: see here https://github.com/adobecom/milo/pull/2690#discussion_r1711189410 @overmyheadandbody

3ch023 (approved):

3ch023:
@mirafedas psi check is failing

PR #2744, opened 2024-08-14
fix(MWPW-148801):Added two New sticky header variants.

sharath-kannan:
* `sticky-tablet-up` which applies the sticky functionality for tablet and desktop.
* `sticky-desktop-up` which applies the sticky functionality only for desktops.

Resolves: [MWPW-148801](https://jira.corp.adobe.com/browse/MWPW-148801)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/sharathkannan/plans?martech=off
- After: https://mwpw-148801--milosh--sharath-kannan.hlx.page/drafts/sharathkannan/plans?martech=off


libs/blocks/table/table.css
```diff
@@ -607,4 +607,8 @@ header.global-navigation {
     flex-direction: column;
     align-items: initial;
   }
+  
+  .table .action-area .con-button.button-l {
+    max-width: 70%;
```
overmyheadandbody, line 612: PR description states:
> Added min-width for table Cta button
which doesn't seem to match this declaration.

Furthermore, `70%` seemed a bit arbitrary to me, so I had a look at a few use-cases and I'm not sure this is the way to go. I limited the CTA's parent, `.action-area` to `max-width: 100%` and applied an `overflow: hidden` to the CTA. Here are a few use-cases comparing this approach (left column) to the one suggested in this PR (right column):

| 360px (70% still overflows) | 768px (70% spans multiple rows) | 1200px (70% spans multiple rows) |
| ------------- | ------------- | ------------- |
| Screenshot 2024-08-21 at 14 35 04 | Screenshot 2024-08-21 at 14 36 30 | Screenshot 2024-08-21 at 14 37 41 |

overmyheadandbody (changes requested):
Blocking this since it triggered some controversy on the authoring side as to whether this is the default behavior we want.

overmyheadandbody (approved):
Make sure to rebase your changes to the latest `stage` branch, as some updates have been pushed for the `table` block. That would be helpful for QA purposes.

It would also be helpful to add some more examples to the test page with variations of all 3 possibilities.

drashti1712 (approved):

overmyheadandbody:
Hey @sharath-kannan! I see this has the required approvals, is there a QE that can look into getting this tested?

Anupamasonnalagi:
Verified the PR. Button in the table is proper in iPhone device and content is visible.
Completed the regression. Details are there in JIRA: [MWPW-157739](https://jira.corp.adobe.com/browse/MWPW-157739)

PR #2745, opened 2024-08-14
MWPW-147759 - quiz option contrast fixes

colloyd:
* slightly adjusts the blue text color of selected option cards so they have enough accessible color contrast

Resolves: [MWPW-147759](https://jira.corp.adobe.com/browse/MWPW-147759)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/colloyd/quiz/?martech=off
- After: https://quiz-color-contrast-fixes--milo--colloyd.hlx.page/drafts/colloyd/quiz/?martech=off

### DIFF
--- a/libs/blocks/quiz/quiz.css
+++ b/libs/blocks/quiz/quiz.css
@@ -1,3 +1,7 @@
+:root {
+  --quiz-option-text-selected: #0b56af;
+}
+
 .quiz-page * {
   box-sizing: border-box;
 }
@@ -150,11 +154,11 @@ html[dir="rtl"] .quiz-option-icon {
 }
 
 .quiz-option.selected .quiz-option-title {
-  color: var(--link-hover-color-dark);
+  color: var(--quiz-option-text-selected);
 }
 
 .quiz-option.selected .quiz-option-text {
-  color: var(--link-hover-color-dark);
+  color: var(--quiz-option-text-selected);
   font-weight: 700;
 }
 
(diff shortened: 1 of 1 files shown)


libs/blocks/quiz/quiz.css
```diff
@@ -1,3 +1,7 @@
+:root {
+  --quiz-option-text-selected: #0b56af;
```
overmyheadandbody, line 2: Does this need to be in the root, though? Could it not be scoped to `.quiz-page` or something specific to quiz? Naming-wise, why `selected`, since it doesn't seem to be specific to text that has been selected/focused/etc. by the user?

colloyd: @overmyheadandbody The variable is for the blue color of the selected state of the cards (and the active state of the button) - so that's why it has that naming. I could scope it to  something else, but I haven't seen that pattern elsewhere in milo, so I figured I'd just follow suit. Since it's unique to quiz, it'll only be present on quiz pages where this css is loaded anyway. 

happy to adjust if you still feel like it would be better to scope it some other way.

overmyheadandbody: I've started doing it for the Tables TT work, but indeed, it's not a common pattern just yet. I'd argue it should be, though, it would make theming a lot easier and the values would be more handy.

Nevertheless, I agree that we can go with this in this scenario, since it has a clear prefix.

overmyheadandbody (approved):

JackySun9 (approved):

PR #2752, opened 2024-08-16
Parallelize Placeholder loading for an LCP improvement

mokimo:
### Description

- Load placeholders for each section and not for the whole document. No placeholders in LCP? No placeholders loading
- Parallelize loading placeholders together with their JS
- Move `customFetch` into `utils.js` for faster LCP loading times for certain blocks & placeholders

There are substantial gains in this PR for high ping (mobile) & a bit less for low-ping (desktop) - but in both scenarios we should save over 100ms on the LCP

On my desktop I saw a 400 MS+ LCP improvement on an uncached hlx.page
Before: `195(placeholders.js) + 200 (helpers.js) + 300 (placeholders.json)` = 695 MS request time
After: 300ms (placeholders.js + placeholders.json)

**1. Sequential load**
Currently we load placeholders.js, then we load helpers.js to use customFetch (fetch without cache headers) then we load placeholders.json .... On a fast 3G network aka "high ping environment" that's 3x400ms per request
1  sequential load

**2. Parallelising the loading**
If we shift the loading sequence to load `placeholders.js` + `placeholders.json` together, we can save one 400ms request
2  parallel load -30-600ms

**3. Removing the customFetch dependency**
We are waiting another 400ms from another file for 7 lines of code. We can move those into `utils.js` to save on this precious pre-LCP time on mobile.
3  remove helpers dependency another -30-600ms

Resolves: [MWPW-156840](https://jira.corp.adobe.com/browse/MWPW-156840)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/osahin/gnav-1?martech=off
- After: https://placeholder-lcp-improvements--milo--mokimo.hlx.page/drafts/osahin/gnav-1?martech=off


mokimo:
Still resolving a bug in CC that fetches placeholders a few too many times... Still want to open this already (mainly to not lose the precious description I already done)

libs/utils/utils.js
```diff
@@ -727,33 +727,24 @@ async function decorateIcons(area, config) {
...
-    textNode.nodeValue = textNode.nodeValue.replace(/&nbsp;/g, '\u00A0');
-  });
-  await Promise.all(replaceNodes);
+  return fetch(resource, options);
+}
+
+async function decoratePlaceholders(area, config) {
+  const placeholderPath = `${config?.locale?.contentRoot}/placeholders.json`;
+  let resolve;
+  window.tempPlaceholders = { placeholderPath: new Promise((r) => { resolve = r; }) };
```
mokimo, line 742: TODO: Investigate and scrap this demo code.

overmyheadandbody (commented):
Sounds great overall, I'll give it a better look next week.

mokimo:
I went a step further to ensure we build the placeholders on a per-section-basis AND only if we have them.
If section 5 has placeholders, there is no reason we eagerly load them before the LCP.
image

mokimo:
CC test links
- Before: https://main--cc--adobecom.hlx.live/de/products/photoshop?martech=off
- After: https://main--cc--adobecom.hlx.live/de/products/photoshop?milolibs=placeholder-lcp-improvements--milo--mokimo&martech=off

mokimo:
@afmicka can you retest? @SilviuLCF this is ready now

afmicka:
@mokimo Looks good now. Nala would also fail if it wasn't :D . Thanks

libs/features/placeholders.js
```diff
@@ -1,20 +1,20 @@
...
+const fetchPlaceholders = async ({ config, sheet, placeholderRequest, placeholderPath }) => {
+  const path = placeholderPath || getPlaceholdersPath(config, sheet);
+  fetchedPlaceholders[path] = fetchedPlaceholders[path]
     // eslint-disable-next-line no-async-promise-executor
     || new Promise(async (resolve) => {
-      const resp = await customFetch({ resource: placeholdersPath, withCacheRules: true })
-        .catch(() => ({}));
+      const resp = await placeholderRequest
+        || await customFetch({ resource: path, withCacheRules: true })
+          .catch(() => ({}));
```
robert-bogos, line 17: Nit: we can remove some parentheses from here 
```suggestion
        || await customFetch({ resource: path, withCacheRules: true })
          .catch(() => {});
```

test/utils/utils.test.js
```diff
@@ -31,6 +30,13 @@ describe('Utils', () => {
     delete window.hlx;
   });
 
+  it('fetches with cache param', async () => {
+    window.fetch = mockFetch({ payload: true });
+    const mockUrl = './mocks/taxonomy.json';
+    const resp = await customFetch({ resource: mockUrl, withCacheRules: true });
```
robert-bogos, line 36: Nit: I think we can give the string value directly to the `resource` key. That would still be within the linter chars limit for 1 line.
```suggestion
    const resp = await customFetch({ resource: './mocks/taxonomy.json', withCacheRules: true });
```

robert-bogos (approved):
Nice 👌

libs/features/placeholders.js
mokimo: `({})` returns an empty object vs `{}` would implicity return undefined.

As we check `const json = resp.ok ? await resp.json() : { data: [] };` within the placeholders, removing the parentheses would lead to a JS error

libs/features/placeholders.js
```diff
@@ -1,20 +1,20 @@
+import { customFetch, getConfig } from '../utils/utils.js';
```
overmyheadandbody, line 1: As a pattern, we tend to avoid circular dependencies like these. Are we fine to keep them here?

libs/features/placeholders.js
overmyheadandbody, line 17: Having 5 indents is usually a sign that things might need to be simplified. I don't have a better suggestion right now, but this piece of logic seems to be more complicated than needed to me.

libs/utils/utils.js
```diff
@@ -983,6 +1000,7 @@ async function checkForPageMods() {
 }
 
 async function loadPostLCP(config) {
+  await decoratePlaceholders(document.body.querySelector('header'), config);
```
overmyheadandbody, line 1003: Why is this needed all of a sudden?

libs/utils/utils.js
```diff
@@ -1211,8 +1229,6 @@ export async function loadArea(area = document) {
   }
   const config = getConfig();
 
-  await decoratePlaceholders(area, config);
```
overmyheadandbody, line 1214: `loadArea` is used by a bunch of other [logic](https://github.com/search?q=org%3Aadobecom+loadarea&type=code). Are we sure this won't lead to placeholders not being decorated on other projects making use of the `loadArea` method?

libs/features/placeholders.js
```diff
@@ -100,18 +99,37 @@ export async function replaceKeyArray(keys, config, sheet = 'default') {
...
   // The .shift method is very slow, thus using normal iterator
   let i = 0;
   // eslint-disable-next-line no-plusplus
   const finalText = text.replaceAll(regex, () => placeholders[i++]);
   return finalText;
 }
+
+export async function decorateArea({ placeholderPath, placeholderRequest, nodes }) {
+  const config = getConfig();
+  if (!nodes.length) return;
```
overmyheadandbody, line 125: Nit: we even save config to memory if it's possible it's not needed, should `nodes` be `undefined`?

libs/features/placeholders.js
mokimo: We had a circular dependency before already.
`Utils.js -> placeholders.js -> helpers.js (customFetch) -> utils.js`

I remember we discussed with @auniverseaway a while back that the only reason we avoided them was linting rules, which we changed.

I haven't seen it leading to any issues yet 🤷

libs/utils/utils.js
mokimo: `loadArea` still calls `processSection` which in turn calls `decoratePlaceholders` 

So the call moved into an embedded function of `loadArea` - Hence, I don't think this will lead to issues

libs/utils/utils.js
mokimo: This is an interesting one. Before we had
```js
  ...
  await decoratePlaceholders(area, config);
```

which changed to
```js
async function processSection(section, config, isDoc) {
  ...
  await decoratePlaceholders(section.el, config);
```

I've only seen that the header (breadcrumbs) doesn't get decorated via `processSection` ... we might need the same call for the footer 🤔 or whole document to be sure.

I wanted to see if we can get away with having the more specific call first, as there is a lot more code to process for the whole document rather than just the header

libs/features/placeholders.js
mokimo: I've moved this part around quite a bit, also trying to expose the placeholderRequest // `fetchedPlaceholders` on the window obj, that wasn't much clearer unfortunately.

However rather than passing the request around via params we can populate `window.mph` https://github.com/adobecom/milo/pull/2737

Which option would you prefer, keeping things as is - or moving `window.mph` to the utils & adding the placeholderRequest into `window.mph` ?

overmyheadandbody: Maybe something like:
```
const fetchPlaceholders = async ({ config, sheet, placeholderRequest, placeholderPath }) => {
  const path = placeholderPath || getPlaceholdersPath(config, sheet);

  fetchedPlaceholders[path] = fetchedPlaceholders[path] || new Promise(async (resolve) => {
    try {
      const resp = await (placeholderRequest || customFetch({ resource: path, withCacheRules: true }));
      const json = resp.ok ? await resp.json() : { data: [] };

      if (json.data.length === 0) { resolve({}); return; }

      const placeholders = json.data.reduce((plc, item) => {
        plc[item.key] = item.value;
        return plc;
      }, {});

      resolve(placeholders);
    } catch (error) {
      resolve({});
    }
  });

  return fetchedPlaceholders[path];
};
```
Maybe I was even a bit too cautious with the `try/catch` and it can be removed as well, reducing one more neccessary indent.

libs/utils/utils.js
mokimo: We move the beadcrumbs to the header via JS(`decorateDocumentExtras()`>`decorateHeader()`), hence they are part of the header.

I wonder if they _should_ ? Metadata can be rolled out to all documents within a locale already, hence I'm not sure if we'd need to mix placeholders+metadata

libs/features/placeholders.js
mokimo: Ah I see, done as far as the linting rules allow me to 👍

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-156840

PR #2754, opened 2024-08-17
MWPW-156898 [MEP] Add setting to use XLG segments in MEP while logged out

vgoodric:
Although XLG segments can be sent to any user, MEP currently limits using them to logged in users.
Similar to the "target" setting for MEP, we will add an "xlg" setting that allows the use of XLG segments for logged out users.

* add if statement to loadIms to not set entitlements to empty when logged out if setting for xlg is set to loggedout
* move getMepEnablement and it's related functions higher so they are defined for use before loadIms
* call setupEntitlementCallback in martech.js if set to logged out, regardless of logged in status
* this also seems to fix a bug that causes a blank screen if: user is logged out, target enablement is off, and manifest uses xlg segments. we don't have any pages RIGHT NOW that use xlg without Target, but this will be good to fix for that future possibility.

Resolves: [MWPW-156898](https://jira.corp.adobe.com/browse/MWPW-156898)

QA instructions:
I've set the xlg metadata to loggedout for the page below.  I've also created a simple manifest that removes the header in the marquee so you can see if it's working.

How to check if/which segments are being sent:
1. Go to your network tab and filter for the string "interact?"
2. Look in your first entry at the response tab and see if you have anything that says "alias": "XLG Custom Personalization", with a list of segments.

QA steps:
1. Go to the https://www.adobe.com/products/acrobat-pro-cc.html
2. If you're logged out, check your xlg segments
3. If you're not getting any segments, you need to log in, wait 15 min, log out and check again
4. Once you confirm you're getting segments, you need to content override the 2 altered files in this PR
5. If you're not in the cc-free segment, also content override libs/features/personalization/entitlements.js and add any segment you are into the list with the name "test". This is also in the manifest.
6. Go to https://www.adobe.com/products/acrobat-pro-cc.html?mep=/cc-shared/fragments/tests/2024/q3/mepxlgloggedout/manifest.json&mepHighlight=true and see if the marquee headline is removed

Note, you can test the above on stage as well.

Psi-check: https://mepxlgloggedout--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/libs/features/personalization/entitlements.js
+++ b/libs/features/personalization/entitlements.js
@@ -30,6 +30,10 @@ const ENTITLEMENT_MAP = {
   '5c6a4bb8-a2f3-4202-8cca-f5e918b969dc': 'firefly-signup-source',
   '20106303-e88c-4b15-93e5-f6a1c3215a12': 'firefly-web-usage',
   '3df0b0b0-d06e-4fcc-986e-cc97f54d04d8': 'acrobat-web-usage',
+
+  // Express segments
+  '2a537e84-b35f-4158-8935-170c22b8ae87': 'express-entitled',
+  'eb0dcb78-3e56-4b10-89f9-51831f2cc37f': 'express-pep',
 };
 
 export const getEntitlementMap = async () => {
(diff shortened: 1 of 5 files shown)


vhargrave (approved):
this was harder than I thought it would be to test and understand, but looks good! 
If you can easily add a unit test to cover this use case, I'd do that.

JingleH (approved):

PR #2763, opened 2024-08-19
Revert "MWPW-151936 Aside Tiger Team Enhancements"

elan-tbx:
Reverts adobecom/milo#2644

Unfortunately an issue was found with this PR that affects older versions of the Aside block still using the deprecated `notification` variant. A fix will need to be implemented before it is ready to go.

Test URLs:
- Before: https://stage--milo--adobecom.hlx.page/docs/library/blocks/aside?martech=off
- After: https://revert-2644-methomas-aside-tiger--milo--adobecom.hlx.page/docs/library/blocks/aside?martech=off

Bacom:
- Before: https://stage--bacom--adobecom.hlx.page/?milolibs=stage
- After: https://stage--bacom--adobecom.hlx.page/?milolibs=revert-2644-methomas-aside-tiger


elan-tbx:
Additionally, to whoever sees this, I cannot for the life of me get the `aem-psi-check` to recognize my test URLs here 🤦🏻‍♂️ But that should not make a difference considering this is a revert PR

PR #2765, opened 2024-08-20
MWPW-156793 [MILO][Analytics] Change char limit for headers based on MD

markpadbe:
Note: this PR is high priority because it is needed in time to QA before MAX

Events team has a unique need for analytics: They don't need the extra characters reserved for MEP pzn, but they do need more for the full title. Since they will want this for all of their pages, we don't want a custom block, rather metadata so it can be set for entire folders.

* checks for analytics-header-limit metadata value
* Acceptable values off (no limit), a number (value is set to that number), and anything else including empty (uses same as default). 

Resolves: [MWPW-156793](https://jira.corp.adobe.com/browse/MWPW-156793)
QA method 1 steps:
1. inspect the only button each page
2. check the length of the daa-ll value.  The allowed length of the string after the "--" should change

QA method 2 steps:
Go to the following link, and set break points at the 3 places in the attributes file where the analytics header value is set:
https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q3/analyticscharlimit/limit15?milolibs=analyticscharlimit

Screenshot 2024-08-19 at 5 22 54 PM

step over and check the length of the header text it should be limited to the first 15 characters of the header text.

you can do the same for following 4 links and the header text length should be 50 characters, unlimited, default(20 characters), and default(20 characters), respectively:
https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q3/analyticscharlimit/limit50?milolibs=analyticscharlimit

https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q3/analyticscharlimit/limitoff?milolibs=analyticscharlimit

https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q3/analyticscharlimit/withwrongtext?milolibs=analyticscharlimit

https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q3/analyticscharlimit/nometadataentry?milolibs=analyticscharlimit 

**Test URLs:**
- After: https://analyticscharlimit--milo--adobecom.hlx.page/?martech=off


libs/martech/attributes.js
```diff
@@ -11,6 +13,13 @@ export function processTrackingLabels(text, config, charLimit) {
   return analyticsValue;
 }
 
+function getHeaderCharLimit(str) {
+  const defaultLimit = 20;
+  if (str === null || str === undefined) return defaultLimit;
+  if (str === 'off') return 0;
+  if (!Number.isNaN(Number(str))) return Number(str);
```
vgoodric, line 20:
```suggestion
  if (!isNaN(str)) return parseInt(str);
```

libs/martech/attributes.js
```diff
@@ -11,6 +13,13 @@ export function processTrackingLabels(text, config, charLimit) {
   return analyticsValue;
 }
 
+function getHeaderCharLimit(str) {
+  const defaultLimit = 20;
+  if (str === null || str === undefined) return defaultLimit;
+  if (str === 'off') return 0;
```
vgoodric, line 19: passing in 0 as the second value will return an empty string.  So you're turning off the value, but we want to turn off the limit.
```suggestion
  if (str === 'off') return false;
```

vgoodric (approved):

libs/martech/attributes.js
markpadbe: I kept the Number.isNan in the condition because it throws a linting error without it and I've updated to use parseInt(str, 10) to insure integer value. Thanks.

libs/martech/attributes.js
```diff
@@ -11,6 +13,13 @@ export function processTrackingLabels(text, config, charLimit) {
   return analyticsValue;
 }
 
+function getHeaderCharLimit(str) {
+  const defaultLimit = 20;
+  if (str === null || str === undefined) return defaultLimit;
```
robert-bogos, line 18: Nit: I think this can be shorter
```suggestion
  if (!str) return defaultLimit;
```

robert-bogos (commented):
Is it possible to add some unit tests for these changes?

ivanvatadobe (approved):

denlight (approved):
Agree with Robert's comments

robert-bogos (approved):

qiyundai (approved):
LGTM!

PR #2776, opened 2024-08-21
MWPW-153372 - Add ability to further customize buttons

elan-tbx:
* After some discussion, a pattern has been agreed upon for extending our button styles beyond the base 3 (text link, outlined button, filled button). For now, the main variant that this PR enables is the "black-filled" style, which already had preexisting CSS written for it in styles.css, but it opens the door for additional customizations in the future.
* The authoring pattern for this is simply adding `#_button-{style}` to the end of the button's URL. In the case of the black-filled button, for example, it would be `#_button-fill`. Whatever is in the `style` part of the tag will be applied as a CSS class to the button, via the `decorateButtons()` function.
* Note that this should not interfere with other link customizations such as #_dnt or modal links -- multiple tags can be added to a single link, and there is documented precedent for doing this already in Milo (see the "Link Features" tab of this [authoring documentation](https://main--milo--adobecom.hlx.page/docs/authoring/links))

Resolves: [MWPW-153372](https://jira.corp.adobe.com/browse/MWPW-153372)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ebartholomew/buttons?martech=off
- After: https://ebartholomew-MWPW-153372-custom-btns--milo--adobecom.hlx.page/drafts/ebartholomew/buttons?martech=off


libs/utils/decorate.js
```diff
@@ -5,17 +5,24 @@ export function decorateButtons(el, size) {
...
+      target = parent;
     } else {
-      button.classList.add('con-button', buttonType);
-      if (size) button.classList.add(size); /* button-l, button-xl */
       parent.insertAdjacentElement('afterend', button);
       parent.remove();
     }
+    target.classList.add('con-button', buttonType);
+    if (size) target.classList.add(size); /* button-l, button-xl */
+    const customClasses = [...target.href.matchAll(/#_button-([a-zA-Z-]+)/g)];
```
robert-bogos, line 19: Is there any scenario in which the `href` attribute would not be available? 
If there is, I think we should add a safety check to avoid errors for trying to iterate a not iterable value.
```suggestion
    const customClasses = target.href && [...target.href.matchAll(/#_button-([a-zA-Z-]+)/g)];
```

ryanmparrish (approved):
Amen 🙏

SilviuLCF (approved):
Verified , ready to STAGE 
Testing details: https://jira.corp.adobe.com/browse/MWPW-153372

PR #2778, opened 2024-08-22
MWPW-157117 Not changing the path for the media image in fragment if the image is same in main document to reduce the network call

suhjainadobe:
Not changing the path for the media image in fragment if the image is same in main document to reduce the network call

Resolves: [MWPW-157117](https://jira.corp.adobe.com/browse/MWPW-NUMBER)

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.live/drafts/suhjain/photoshop-new/photoshop
- After: https://main--cc--adobecom.hlx.live/drafts/suhjain/photoshop-new/photoshop?milolibs=media-path

### DIFF
--- a/libs/blocks/fragment/fragment.js
+++ b/libs/blocks/fragment/fragment.js
@@ -99,7 +99,7 @@ export default async function init(a) {
 
   const html = await resp.text();
   const doc = new DOMParser().parseFromString(html, 'text/html');
-  replaceDotMedia(a.href, doc);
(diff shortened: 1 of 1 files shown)


mokimo:
There seem to be a lot more calls now

### Before
Screenshot 2024-08-22 at 08 44 06

### After
Screenshot 2024-08-22 at 08 44 02

suhjainadobe:
Keeping this on hold

PR #2780, opened 2024-08-22
MWPW-156866 [MILO][MEP] Create martech metadata table if placeholders are used in non EN page

AdobeLinhart:
When we use MEP placeholders on a non EN page, the analytics don't stay in English. To combat this, we will add to the martech-metadata.

In libs/features/personalization/personalization.js parsePlaceholders function, we decide the column of values we will use.

We should check the value of config.locale.ietf. If it equals en-US we don't need to generate a table.

If it doesn't, we need to see if there is a good column for us to use for the original English text. Column names we want to use in order of preference: en-us, us, en.

I think you can make an array with those values, then reuse the code we use to build the placeholder object to build a translation object. Move it to a reusable function maybe.

Then add to config.mep and reference your new data in martech/attributes.js

Resolves: [MWPW-156866](https://jira.corp.adobe.com/browse/MWPW-156866)

QA instructions
In the before and after links, inspect the buttons in the marquee. The analytics should now show en-us version of the placeholder instead of the fr.

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/fr/products/illustrator?mep=%2Fproducts%2Fillustrator.json--default---%2Fdrafts%2Fdlinhart%2Ffilter-test%2Ffilter-test.json--non-geo-test
- After: https://main--cc--adobecom.hlx.page/fr/products/illustrator?mep=%2Fproducts%2Fillustrator.json--default---%2Fdrafts%2Fdlinhart%2Ffilter-test%2Ffilter-test.json--non-geo-test&milolibs=geometadata
- Psi-check: https://geometadata--milo--adobecom.hlx.page/?martech=off


mokimo (changes requested):
Is this really necessary? Can't we re-use the existing placeholder entries?
They've recently been exposed https://github.com/adobecom/milo/pull/2737/files

vgoodric:
> Is this really necessary? Can't we re-use the existing placeholder entries? They've recently been exposed https://github.com/adobecom/milo/pull/2737/files

@mokimo yes, because this is for MEP specific placeholders that are being introduced by the manifests. They are not part of the retreived placeholders.

libs/features/personalization/personalization.js
```diff
@@ -540,6 +540,27 @@ export function parseManifestVariants(data, manifestPath, targetId) {
...
+  const usCol = firstRow['en-us'] || firstRow.us || firstRow.en || firstRow.key;
+
+  if (!usCol) return;
+
+  Object.values(config.placeholders).forEach((item) => {
+    const row = createTag('div', undefined, `<div>${item}</div><div>${usCol}</div>`);
+    metaEl.appendChild(row);
+  });
+
+  document.querySelector('main').appendChild(section);
```
mokimo, line 561: Is there any reason to store the data in the html? I'd expose it on `window.X` rather than adding markup. I feel like it's somewhat of an anti-pattern to store things within the HTML unless there's a good reason to do so

mokimo:
There are some unit tests / linting failing, other than that IMO this looks good and can be approved after that's fixed 👍

vgoodric (approved):

vgoodric:
@mokimo can you review and lift request change if you are satisfied?

ivanvatadobe (approved):

libs/features/personalization/personalization.js
```diff
@@ -540,6 +540,25 @@ export function parseManifestVariants(data, manifestPath, targetId) {
   return null;
 }
 
+export async function createMartechMetadata(placeholders, config, column) {
+  if (config.locale.ietf === 'en-US') return;
+
+  await import('../../martech/attributes.js').then(({ processTrackingLabels }) => {
```
mokimo, line 546: I'd be a bit careful with those imports as they block the pipeline and we can't load things in parallel.

mokimo (approved):

libs/features/personalization/personalization.js
mokimo: Here's a bit of a case-study I did on project unity on this: https://github.com/adobecom/unity/pull/102

PR #2781, opened 2024-08-23
MWPW-154098: Retain Modal param while traversing back from Commerce to Page

mirafedas:
When users open a modal by clicking a CTA on a merch card, navigate away, and then return, the modal needs to reopen automatically. However, this wasn't working because our catalog page uses custom modals without updating the URL hash. As a result, the URL doesn't retain information about the previously open modal.

To address this, this PR introduces URL hash management. We now add a hash to the URL whenever a modal is opened and remove it when the modal is closed. On page load, each CTA is assigned a unique hash. If the browser's URL contains the hash of a previously opened modal, we automatically trigger a click on the corresponding CTA to reopen the modal.

Resolves: [MWPW-154098](https://jira.corp.adobe.com/browse/MWPW-154098)

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.live/products/catalog?martech=off&georouting=off&mep=off
- After: https://main--cc--adobecom.hlx.page/products/catalog?milolibs=mwpw-154098-catalog-modal-hash--milo--mirafedas&martech=off&georouting=off&mep=off

For PSI check: https://mwpw-154098-catalog-modal-hash--milo--mirafedas.hlx.live/docs/authoring/promotions/delayed-modal?martech=off&georouting=off&mep=off


libs/features/mas/commerce/src/checkout-link.js
```diff
@@ -185,8 +186,18 @@ export class HTMLCheckoutAnchorElement extends HTMLAnchorElement {
             if (text) this.firstElementChild.innerHTML = text;
             if (className) this.classList.add(...className.split(' '));
             if (handler) {
-                this.setAttribute('href', '#');
+              this.setAttribute('href', '#');
+              if (this.getAttribute('data-modal') === 'true') {
+                const hash = `${offers.flat()[0].offerType === "TRIAL" ? 'try' : 'buy'}-${this.closest('merch-card')?.getAttribute('name')}`;
+                this.setAttribute('hash', hash);
+                this.#checkoutActionHandler = handler.bind(this);
+                if (window.location.hash.replace('#', '') === hash && !this.#modalWasOpened) {
```
yesil, line 194: it feels look like this logic belongs to [openModal](https://github.com/adobecom/milo/blob/stage/libs/blocks/merch/merch.js#L360) function defined in merch.js.

checkout-link custom element cannot be too opinionated how things should be, since it can be used anywhere outside Milo.

libs/blocks/merch/merch.js
```diff
@@ -357,13 +357,20 @@ async function openExternalModal(url, getModal) {
...
 
-export async function openModal(e, url, offerType) {
+export async function openModal(e, url, offerType, hash) {
   e.preventDefault();
   e.stopImmediatePropagation();
   const { getModal } = await import('../modal/modal.js');
   await import('../modal/modal.merch.js');
   const offerTypeClass = offerType === OFFER_TYPE_TRIAL ? 'twp' : 'crm';
   let modal;
+  if (hash) {
```
yesil, line 367: I think we should require gwp to author hashes in the checkout-link.xlsx, generated hashes might not be good w.r.t SEO, when linked from other pages.

If the hash is present, we can use it to handle the modal, if not, we can keep the current behaviour.

GWP can progressively add it and publish the file.

libs/features/mas/commerce/src/checkout-link.js
mirafedas: @yesil It works this way: when a user opens a modal, goes to a different page, then goes back - on page load we need to initiate a click on the CTA with a respective hash.
The `openModal` method is not called at page load.
Alternatively, I can set up an on-page load event listener and on each page load check if there is a CTA with the hash that matches the hash in the page URL, and then click on it, but I think having this check on each page load would be redundant, or should I do it?

yesil: I think this logic could be in the merch.js, once initialized, if there is a hash, then we can perform this check.
It should be executed only on back navigation.

libs/blocks/merch/merch.js
```diff
@@ -375,7 +384,16 @@ export async function openModal(e, url, offerType) {
   }
 }
 
-export async function getModalAction(offers, options) {
+export function setCtaHash(el, checkoutLinkConfig, offerType) {
+  if (!(el && checkoutLinkConfig && offerType)) return undefined;
+  const authoredHash = checkoutLinkConfig[`${(offerType === OFFER_TYPE_TRIAL) ? FREE_TRIAL_HASH : BUY_NOW_HASH}`];
+  const autogeneratedHash = `${offerType === 'TRIAL' ? 'try' : 'buy'}-${el.closest('merch-card')?.getAttribute('name')}`;
```
yesil, line 390: I would like us to avoid supporting `authogenerated` hash values. Is it a hard requirement?

libs/blocks/merch/merch.js
```diff
@@ -375,7 +384,16 @@ export async function openModal(e, url, offerType) {
...
+  const hash = authoredHash ?? autogeneratedHash;
+  el.setAttribute('hash', hash);
```
yesil, line 392: `hash` it not a standard attribute, can you use `data-modal-hash`?

mirafedas: I changed to `data-modal-hash` to not mix it with other buttons that use `data-modal-hash`

yesil (approved):
Thanks for addressing the comments.

3ch023 (approved):

PR #2783, opened 2024-08-23
MWPW-155621 - Adding functionality to the Pill Notification

elan-tbx:
* This PR adds per-viewport content swapping and "popup" behavior capabilities to the Pill variant of the new Notification block, making it a viable replacement for the recently deprecated Aside Promobar.
* Additionally, some more minor cosmetic bugs were addressed:
  1. Apply max-width to block at 1600px, 85% below until 1200px
  2. Add the box shadow from Figma to the block (0 3px 6px #707070)
  3. Add support for Product Lockups to the block
  4. Update the button size class names so that l-button=40px height, m-button=30px height
  5. Increase the spacing between product tile and copy on desktop to 24px
  6. Style the border radius of the block to be completely rounded, even when copy flows to two lines
* Note that this is marked as `high-priority` because Consonant believes the lack of these features blocks the Notification block from release.

Resolves: [MWPW-155621](https://jira.corp.adobe.com/browse/MWPW-155621)

**Test URLs:**

Popup behavior comparison:
- Before: https://main--milo--adobecom.hlx.page/drafts/ebartholomew/popup-behavior-old-promobar-version?martech=off
- After: https://ebartholomew-mwpw-155621-pill-updates--milo--adobecom.hlx.page/drafts/ebartholomew/popup-behavior?martech=off

Regression test of last pill version:
- Before: https://main--milo--adobecom.hlx.page/drafts/ebartholomew/pill-notification?martech=off
- After: https://ebartholomew-mwpw-155621-pill-updates--milo--adobecom.hlx.page/drafts/ebartholomew/pill-notification?martech=off


rgclayton (approved):
Looks good. 
No more potential duplicate `<h>` tags with that new `decorateMultiViewPort()`

libs/blocks/notification/notification.css
```diff
@@ -298,6 +299,11 @@
 
 .notification .flexible-inner {
   inline-size: 100%;
+  box-shadow: 0 3px 6px #707070;
```
overmyheadandbody, line 302: Since these values are identical, would it be worth having them in a variable?

libs/blocks/notification/notification.js
```diff
@@ -122,21 +127,29 @@ async function decorateLockup(lockupArea, el) {
   if (pre && pre[2] === 'icon') el.classList.replace(pre[0], `${pre[1]}-lockup`);
 }
 
+async function decorateForegroundText(el, container) {
+  const text = container?.querySelector('h1, h2, h3, h4, h5, h6, p')?.closest('div');
+  text?.classList.add('text');
+  const iconArea = text?.querySelector('p picture')?.closest('p');
```
overmyheadandbody, line 133: This feels a bit weird with targeting a sibling to then target its parent. Could we get away with `text?.querySelector('p:has(picture)')`?

libs/utils/decorate.js
```diff
@@ -278,3 +278,23 @@ export function applyInViewPortPlay(video) {
...
+  ];
+  const foreground = el.querySelector('.foreground');
+  if (foreground.childElementCount === 2 || foreground.childElementCount === 3) {
+    [...foreground.children].forEach((child, index) => {
+      const mq = window.matchMedia(viewports[index]);
+      const setContent = ({ matches }) => {
+        if (matches) foreground.replaceChildren(child);
+      };
+      setContent(mq);
+      mq.addEventListener('change', setContent);
```
overmyheadandbody, line 296: Since this is nicely grouped together, do you need to pass the arguments around? On L293, could we not just simply check for `if (mq.matches)`?

overmyheadandbody (approved):
Should any of the examples from your test links be reflected in the kitchen sink? https://main--milo--adobecom.hlx.page/docs/library/kitchen-sink/aside-notification

elan-tbx:
> Should any of the examples from your test links be reflected in the kitchen sink? https://main--milo--adobecom.hlx.page/docs/library/kitchen-sink/aside-notification

@overmyheadandbody They won't be for this PR, since this is all related to the new (separate) Notification block. I _had_ been using that page to check against #2782 though, to make sure _NOTHING CHANGES_ since that PR is only meant to address standard and split Asides. Granted, using that page as a source of truth for all examples of the old Notification Asides is kind of what got us in trouble to begin with -- because while it appeared nothing was affected on that page, it did not cover all use-cases (and the PR had to be reverted)

libs/blocks/notification/notification.js
elan-tbx: Ah, so we can now! Safari had been holding us back from using `:has()` since it wasn't supported until 15.4, but looks like our [browser support matrix](https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=WP4&title=Test+Automation+Framework#TestAutomationFramework-Browser&PlatformSupportMatrix) has been updated 🥳

PR #2785, opened 2024-08-26
MWPW-156126 - quiz-entry contrast and accessibility

colloyd:
* adjustment to colors for better contrast
* aria-label for carousel arrows
* css clean up

Resolves: [MWPW-156126](https://jira.corp.adobe.com/browse/MWPW-156126)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/colloyd/quiz-entry/?martech=off
- After: https://quiz-entry-contrast-fixes--milo--colloyd.hlx.page/drafts/colloyd/quiz-entry/?martech=off

### DIFF
--- a/libs/blocks/quiz-entry/quiz-entry.css
+++ b/libs/blocks/quiz-entry/quiz-entry.css
@@ -1,3 +1,8 @@
+.quiz-entry {
+  --quiz-button-disabled-bg: #757575;
+  --quiz-button-disabled-text: #FFF6F6;
+}
+
 .quiz-container {
   align-items: center;
   color: var(--color-black);
@@ -17,15 +22,15 @@
 }
 
 .quiz-title {
-  font-size: 28px;
-  line-height: 36px;
+  font-size: var(--type-heading-xl-size);
+  line-height: var(--type-heading-xl-lh);
   font-weight: var(--type-heading-all-weight);
   margin-bottom: 8px;
 }
 
 .quiz-subtitle {
-  font-size: 20px;
-  line-height: 30px;
+  font-size: var(--type-heading-l-size);
+  line-height: var(--type-heading-l-lh);
 }
 
 .quiz-question-container {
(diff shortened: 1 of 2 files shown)


libs/blocks/quiz-entry/quizoption.js
```diff
@@ -144,7 +144,7 @@ export const GetQuizOption = ({
 
   return html`
   <div class="quiz-options-container" role="group" aria-labelledby="question" tabindex="0" onkeydown=${handleKey}>
-  ${index > 0 && html`<button onClick=${prev} class="carousel-arrow arrow-prev ${isRTL ? 'rtl' : ''}"></button>`}
+  ${index > 0 && html`<button onClick=${prev} class="carousel-arrow arrow-prev ${isRTL ? 'rtl' : ''}" aria-label="Previous"></button>`}
```
mokimo, line 147: Ideally, this should be translated as well for other locales

mokimo (approved):

libs/blocks/quiz-entry/quizoption.js
colloyd: Great suggestion! Of course we can work out translation like this, but it would be more involved because we would need to incorporate it into the strings spreadsheet. I'll have to write up a followup ticket for that, but there's a release deadline for US and they need these basic issues to be corrected first.

JackySun9 (approved):

libs/blocks/quiz-entry/quiz-entry.css
```diff
@@ -1,3 +1,8 @@
+.quiz-entry {
+  --quiz-button-disabled-bg: #757575;
+  --quiz-button-disabled-text: #FFF6F6;
+}
```
robert-bogos, line 4: I don't think we need to declare variables as they're not used in multiple places

colloyd: The point here isn't so much that it would be used in more than one place as it is to make it easier when they are updated. An update to these colors is imminent pending a consonant review - and these may very well be used in multiple places depending on what they decide to do.

robert-bogos (approved):

PR #2715, opened 2024-08-08
MWPW-153810 media image is seen loaded twice in product page urls in desktop

suhjainadobe:
Media image is seen loaded twice in product page urls in desktop
This fixes the replacing of whole innerHTML if there is any placeholder on the page. It only the specific placeholder text.

Resolves: [MWPW-153810](https://jira.corp.adobe.com/browse/MWPW-NUMBER)

Test URLs:

Before: https://main--milo--adobecom.hlx.page/drafts/suhjain/performance/marquee-light
After: https://placeholderReplace--milo--adobecom.hlx.live/drafts/suhjain/performance/marquee-light

PR #2787, opened 2024-08-26
Revert "MWPW-153810 media image is seen loaded twice in product page urls in desktop"

salonijain3:
Reverts adobecom/milo#2715


mokimo:
@salonijain3 do we want to revert this, or just ship an improvement? We do have a week to fix the issue rather than reverting.

mokimo:
Closing this in discussion with @suhjainadobe as we are in an RCP and can ship a fix rather than reverting.
This is a production issue, but currently not deemed critical//CSO worthy.

PR #2788, opened 2024-08-26
MWPW-155694 Dispatch sticky close event for dynamic chat

Brandon32:
* Dispatch sticky close event for dynamic chat
* Increase z-index to place above chat (Dynamic chat has 2 second loading delay)

Resolves: [MWPW-155694](https://jira.corp.adobe.com/browse/MWPW-155694)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/bmarshal/sticky/aside?martech=off
- After: https://bmarshal-chat-promo--milo--adobecom.hlx.page/drafts/bmarshal/sticky/aside?martech=off
- Before: https://main--milo--adobecom.hlx.page/drafts/bmarshal/sticky/notification?martech=off
- After: https://bmarshal-chat-promo--milo--adobecom.hlx.page/drafts/bmarshal/sticky/notification?martech=off

**Bacom URLs:**
- Before: https://business.stage.adobe.com/resources/digital-economy-index.html?martech=on
- After: https://business.stage.adobe.com/resources/digital-economy-index.html?martech=on&milolibs=bmarshal-chat-promo

### DIFF
--- a/libs/blocks/section-metadata/section-metadata.css
+++ b/libs/blocks/section-metadata/section-metadata.css
@@ -201,7 +201,7 @@
 
 .section.sticky-bottom.promo-sticky-section {
   background: none;
-  z-index: 4;
+  z-index: 40000;
 }
 
 .section.sticky-bottom.popup {
--- a/libs/blocks/aside/aside.js
+++ b/libs/blocks/aside/aside.js
@@ -96,6 +96,7 @@ function addCloseButton(el) {
   el.querySelector('.foreground').appendChild(closeBtn);
   closeBtn.addEventListener('click', (e) => {
     e.target.closest('.section').classList.add('close-sticky-section');
+    document.dispatchEvent(new CustomEvent('milo:sticky:closed'));
   });
 }
 
--- a/libs/blocks/notification/notification.js
+++ b/libs/blocks/notification/notification.js
@@ -84,7 +84,12 @@ function wrapCopy(foreground) {
 
 function decorateClose(el) {
   const btn = createTag('button', { 'aria-label': 'close', class: 'close' }, closeSvg);
-  btn.addEventListener('click', () => (el.style.display = 'none'));
+  btn.addEventListener('click', () => {
+    el.style.display = 'none';
+    el.closest('.section')?.classList.add('close-sticky-section');
+    document.dispatchEvent(new CustomEvent('milo:sticky:closed'));
+  });
+
   el.appendChild(btn);
 }
 


libs/blocks/section-metadata/section-metadata.css
```diff
@@ -201,7 +201,7 @@
 
 .section.sticky-bottom.promo-sticky-section {
   background: none;
-  z-index: 4;
+  z-index: 40000;
```
JasonHowellSlavin, line 204: I am guessing the other martech element has a high z-index like this as well?

JasonHowellSlavin (approved):

libs/blocks/section-metadata/section-metadata.css
Brandon32: Yes, the dynamic chat is at 10k.

meganthecoder (commented):
I think this is fine given what currently exists, but @ryanmparrish @elan-tbx we might want to rethink some of the sticky functionality. Didn't our tiger team say that any sticky section should be closable? It seems like maybe the close button/functionality should go with the section metadata code rather than the block code, if possible.

elan-tbx:
@meganthecoder Looks like that's already a part of (Asides)[https://github.com/adobecom/milo/blob/721645234c38ae69cf110e5b822ce4cc29ea821a/libs/blocks/aside/aside.js#L97], I can create a due-diligence ticket for this to be added to Notifications as well.

It's just tricky, because the actual DOM element of a close button would need to live within each block itself, so having the section inject a close button into the block feels wrong.

Would make sense to me if we dispatched the same event onClick across those blocks and let section-metadata handle the rest, though.

meganthecoder:
> @meganthecoder Looks like that's already a part of [Asides](https://github.com/adobecom/milo/blob/721645234c38ae69cf110e5b822ce4cc29ea821a/libs/blocks/aside/aside.js#L97), I can create a due-diligence ticket for this to be added to Notifications as well.
> 
> It's just tricky, because the actual DOM element of a close button would need to live within each block itself, so having the section inject a close button into the block feels wrong.
> 
> Would make sense to me if we dispatched the same event onClick across those blocks and let section-metadata handle the rest, though.

I don't think there's a need for another ticket. Brandon covered notifications as well as asides in this PR. I was more thinking about the fact that other blocks can use the sticky section metadata functionality, and we'd need to add this code to every one. However, if you think that the close button needs to be in the block code then that sounds good.

Thanks!

meganthecoder (approved):

PR #2796, opened 2024-08-28
MWPW-157292: CC and DC: Phone number links wrong

suhjainadobe:
The phone number on the following pages is appearing correctly in the text but showing random characters in the link. Previously, using the placeholder for both the display text and link text displayed the number correctly. 

Resolves: [[MWPW-157292]](https://jira.corp.adobe.com/browse/MWPW-NUMBER)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.live/drafts/suhjain/performance/marquee-light
- After:https://placeorigin--milo--suhjainadobe.hlx.live/drafts/suhjain/performance/marquee-light

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -752,11 +752,11 @@ async function decoratePlaceholders(area, config) {
   const replaceNodes = nodes.map(async (nodeEl) => {
     if (nodeEl.nodeType === Node.TEXT_NODE) {
       const newValue = await replaceText(nodeEl.nodeValue, config, regex);
-      nodeEl.nodeValue = newValue.replace(/&nbsp;/g, '\u00A0');
+      nodeEl.nodeValue = newValue;
     } else if (nodeEl.nodeType === Node.ELEMENT_NODE) {
       const hrefValue = nodeEl.getAttribute('href');
       const newValue = await replaceText(hrefValue, config, regex);
-      nodeEl.setAttribute('href', newValue.replace(/&nbsp;/g, '\u00A0'));
+      nodeEl.setAttribute('href', newValue);
     }
   });
   await Promise.all(replaceNodes);
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -749,14 +749,14 @@ async function decoratePlaceholders(area, config) {
   }
   if (!nodes.length) return;
   const { replaceText } = await import('../features/placeholders.js');
-  const replaceNodes = nodes.map(async (node) => {
-    if (node.nodeType === Node.TEXT_NODE) {
-      let newValue = await replaceText(node.nodeValue, config, regex);
-      node.nodeValue = newValue.replace(/&nbsp;/g, '\u00A0');
-    } else if (node.nodeType === Node.ELEMENT_NODE) {
-      let hrefValue = node.getAttribute('href');
-      let newValue = await replaceText(hrefValue, config, regex);
-      node.setAttribute('href', newValue.replace(/&nbsp;/g, '\u00A0'));
+  const replaceNodes = nodes.map(async (el) => {
+    if (el.nodeType === Node.TEXT_NODE) {
+      const newValue = await replaceText(el.nodeValue, config, regex);
+      el.nodeValue = newValue.replace(/&nbsp;/g, '\u00A0');
+    } else if (el.nodeType === Node.ELEMENT_NODE) {
+      const hrefValue = el.getAttribute('href');
+      const newValue = await replaceText(hrefValue, config, regex);
+      el.setAttribute('href', newValue.replace(/&nbsp;/g, '\u00A0'));
     }
   });
   await Promise.all(replaceNodes);
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -730,28 +730,34 @@ async function decorateIcons(area, config) {
 async function decoratePlaceholders(area, config) {
   const el = area.querySelector('main') || area;
   const regex = /{{(.*?)}}|%7B%7B(.*?)%7D%7D/g;
-  const walker = document.createTreeWalker(
-    el,
-    NodeFilter.SHOW_TEXT,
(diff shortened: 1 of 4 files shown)


mokimo (commented):
Can we add another unit test for this scenario?

salonijain3:
@suhjainadobe the test are not passing on this PR

libs/utils/utils.js
```diff
@@ -730,28 +730,34 @@ async function decorateIcons(area, config) {
...
-    textNode.nodeValue = await replaceText(textNode.nodeValue, config, regex);
-    textNode.nodeValue = textNode.nodeValue.replace(/&nbsp;/g, '\u00A0');
+  const replaceNodes = nodes.map(async (nodeEl) => {
+    if (nodeEl.nodeType === Node.TEXT_NODE) {
+      const newValue = await replaceText(nodeEl.nodeValue, config, regex);
+      nodeEl.nodeValue = newValue.replace(/&nbsp;/g, '\u00A0');
+    } else if (nodeEl.nodeType === Node.ELEMENT_NODE) {
+      const hrefValue = nodeEl.getAttribute('href');
+      const newValue = await replaceText(hrefValue, config, regex);
+      nodeEl.setAttribute('href', newValue.replace(/&nbsp;/g, '\u00A0'));
```
mokimo, line 759: I wonder, is this the only special case? I'm a bit reluctant, thinking we'll have to treat more and more special cases.

libs/utils/utils.js
```diff
@@ -730,28 +730,34 @@ async function decorateIcons(area, config) {
...
   }
   if (!nodes.length) return;
   const { replaceText } = await import('../features/placeholders.js');
-  const replaceNodes = nodes.map(async (textNode) => {
-    textNode.nodeValue = await replaceText(textNode.nodeValue, config, regex);
-    textNode.nodeValue = textNode.nodeValue.replace(/&nbsp;/g, '\u00A0');
+  const replaceNodes = nodes.map(async (nodeEl) => {
+    if (nodeEl.nodeType === Node.TEXT_NODE) {
+      const newValue = await replaceText(nodeEl.nodeValue, config, regex);
+      nodeEl.nodeValue = newValue.replace(/&nbsp;/g, '\u00A0');
```
mokimo, line 755: This should be part of the replaceText method and also recieve a unit test.

libs/utils/utils.js
suhjainadobe: The reason I specified the href is because, if we start checking all the attributes, the computational time will increase. Considering, our DOM will be a large. One thing we could do is to run this section by section, but then again checking all the attributes will be heavy

libs/utils/utils.js
suhjainadobe: done, unit tests are already covered for this

mokimo (approved):
I think it's a good start, esp. as it fixes a prod issue - I wonder if we need to check for {{}} within other attributes as well, but so far my assumption is that the `href` is a bit special in that regard.

We can handle more cases when they come up or think about a better strategy.

libs/utils/utils.js
suhjainadobe: got it

spadmasa:
Verified the fix with the branch and able to see it working in cc
https://main--cc--adobecom.hlx.page/jp/products/substance3d/apps/sampler?milolibs=placeorigin--milo--suhjainadobe

image


when verified the same in dc , not able to see the image icon display and phone number is seen changed after the milo libs append
please find the before and after url in dc
https://main--dc--adobecom.hlx.page/acrobat/business/departments
image
after the fix
https://main--dc--adobecom.hlx.page/acrobat/business/departments?milolibs=placeorigin--milo--suhjainadobe
image

mobile image is not seen and number is seen changed with functionality seen fine


same behaviour is seen in rever pr as well
https://stage--dc--adobecom.hlx.page/acrobat/business/integrations?milolibs=revert-2715-placeholderReplace
image

@JFernandezAdobe  @Blainegunn   can you please confirm on the behaviour in dc

CC : @suhjainadobe  , @sigadamvenkata  , @salonijain3

libs/utils/utils.js
```diff
@@ -730,28 +730,34 @@ async function decorateIcons(area, config) {
...
-    NodeFilter.SHOW_TEXT,
-    {
-      acceptNode(node) {
-        const a = regex.test(node.nodeValue) ? NodeFilter.FILTER_ACCEPT : NodeFilter.FILTER_REJECT;
-        regex.lastIndex = 0;
-        return a;
-      },
-    },
-  );
+  const walker = document.createTreeWalker(el, NodeFilter.SHOW_ALL, null, false);
```
overmyheadandbody, line 733: I'm a bit confused by the last 2 arguments. According to https://developer.mozilla.org/en-US/docs/Web/API/Document/createTreeWalker, the filter is optional, so it may be omitted, while `false` is a 4th argument that is not mentioned in their documentation. What is its purpose?

overmyheadandbody (approved):
Just a small comment, but otherwise looks ok

libs/utils/utils.js
mokimo: I wondered the same. It's `entityReferenceExpansion` https://www.w3.org/TR/WD-DOM-Level-2/traversal.html

mokimo: From what I've seen, the proposal didn't make it into the final draft: https://dom.spec.whatwg.org/#dom-document-createtreewalker

hence I'd say we can omit the 4th argument.

overmyheadandbody: And does it bring any advantages? From a quick search I could only find that it was useful for IE11 🤷

mokimo: Given the living standard, it seems that it isn't part of the specification at all anymore `createTreeWalker(root, whatToShow, filter)`

overmyheadandbody:
This should be rebased before it gets merged, @suhjainadobe

PR #2801, opened 2024-08-28
MWPW-156865 - Hero marquee supports heading element on detail text. 

ryanmparrish:
DX authors want the ability to author a heading tag in the hero-marquee detail copy. 

Screen Shot 2024-09-04 at 11 46 42 AM

 -  added a parameter `hasDetailHeading` on the shared decorateBlockText() function and invoked this on the [hero-marquee]

Resolves: [MWPW-156865](https://jira.corp.adobe.com/browse/MWPW-156865)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rparrish/hero/hero-w-detail-header?martech=off
- After: https://rparrish-hero-detail--milo--adobecom.hlx.page/drafts/rparrish/hero/hero-w-detail-header?martech=off

**Bacom Test**
- Before: https://main--bacom--adobecom.hlx.page/test/kyung/test-hero-marquee?martech=off
- After: https://main--bacom--adobecom.hlx.page/test/kyung/test-hero-marquee?milolibs=rparrish-hero-detail&martech=off


robert-bogos (approved):
Looks good 👍

libs/blocks/hero-marquee/hero-marquee.js
```diff
@@ -175,7 +175,7 @@ export default async function init(el) {
     : null;
   if (assetUnknown) assetUnknown.classList.add('asset-unknown');
 
-  decorateBlockText(copy, ['xxl', 'm', 'l']); // heading, body, detail
+  decorateBlockText(copy, ['xxl', 'm', 'l'], 'allowDetailHeading'); // heading, body, detail
```
overmyheadandbody, line 178: Naming nit: the function parameter of `decorateBlockText ` is `type`, while `allowDetailHeading` sounds more like an option/configuration to me. Maybe `hasDetailHeding` or a verb-based name is better suited for a type.

libs/utils/decorate.js
```diff
@@ -54,9 +54,16 @@ export function decorateBlockText(el, config = ['m', 's', 'm'], type = null) {
   const headings = el.querySelectorAll('h1, h2, h3, h4, h5, h6');
   if (!el.classList.contains('default')) {
     if (headings) {
-      headings.forEach((h) => h.classList.add(`heading-${config[0]}`));
+      const lastHeading = headings[headings.length - 1];
+      if (type === 'allowDetailHeading') {
+        lastHeading?.classList.add(`heading-${config[0]}`);
+      } else {
+        headings.forEach((h) => h.classList.add(`heading-${config[0]}`));
+      }
```
overmyheadandbody, line 62: Maybe I'm missing something, but here's how I see this: before, all headings would get a `heading-{size}` class. With this change, a block using `allowDetailHeading` would just apply that class to the last heading, leaving all others without class (pun intended?). On the test pages, I noticed the xxxl headline is no longer styled as such:

| Before | After |
| Screenshot 2024-09-04 at 12 40 02 |  Screenshot 2024-09-04 at 12 40 14 |

I think the approach should be switched to "all headings except the first get the `heading-{size}` class" or whatever you deem correct to cover all the use-cases.

overmyheadandbody (changes requested):
I think this might affect some current use-cases. It could be that that's the intention, although it's not what I gathered from the story.

overmyheadandbody:
There's some Git quirkiness here, as there now seem to be no changes to the code.

libs/utils/decorate.js
ryanmparrish: Thanks for catching that. Makes more sense to switch the approach like you said and less type checking in the code. 👍

ryanmparrish:
@overmyheadandbody  - Thanks for the review and good feedback. I stashed the changes and merged stage to get current, sorry there that threw you for a wobbly a sec. I've updated everything now when you get another look

overmyheadandbody (approved):
Thanks for integrating the changes, @ryanmparrish! It's looking more streamlined

PR #2802, opened 2024-08-29
Checking if Ims is already initialized before loading it

bandana147:
* Check if IMS is already initialized before loading it from Gnav

Resolves: [MWPW-157425](https://jira.corp.adobe.com/browse/MWPW-157425)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/nishantkaush/gnav/custom/dark?martech=off
- After: https://gnav--milo--adobecom.hlx.page/drafts/nishantkaush/gnav/custom/dark/?martech=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -882,7 +882,7 @@ export const getMepEnablement = (mdKey, paramKey = false) => {
   return getMdValue(mdKey);
 };
 
-let imsLoaded;
+let imsLoaded = window.adobeIMS?.initialized;
 export async function loadIms() {
   imsLoaded = imsLoaded || new Promise((resolve, reject) => {
     const {


sharmrj (approved):

mokimo (approved):

spadmasa:
Validation done on below url :
https://gnav--milo--adobecom.hlx.page/drafts/blaishram/unav-test?martech=off
image
verified the login  flow
https://main--cc--adobecom.hlx.page/creativecloud?milolibs=gnav
image
image
image

libs/utils/utils.js
```diff
@@ -882,7 +882,7 @@ export const getMepEnablement = (mdKey, paramKey = false) => {
   return getMdValue(mdKey);
 };
 
-let imsLoaded;
+let imsLoaded = window.adobeIMS?.initialized;
```
narcis-radu, line 885: I'm not fully convinced this will work every time, but we can adapt if needed. Here's an use-case: user has `adobeid` defined and is loading imslib script (window.adobeIMS?.initialized would be false) and at the same time we're executing `loadIms` method and we override the `adobeid` object and load imslib.min.js from www.adobe.com. The result is that ims would get configured incorrectly or might not work properly since we're loading two instances of imslib.min.js

mokimo (changes requested):
I just realized `imsLoaded = window.adobeIMS?.initialized;` will be a boolean rather than promise, if someone calls `loadIms.then()` we'd return a boolean and have JS errors.

Would be great to change the return value to a promise that resolves and write a unit test for this scenario.

libs/utils/utils.js
bandana147: If we expose a key in the config that can be set by the consuming client to indicate whether they want to use their own IMS or Gnav's IMS, we can modify the code as follows:

 ```
const { useClientIms } = getConfig();
    if (useClientIms) return;

    return loadIms()
      .then(() => this.imsReady())
      .catch((e) => {
        if (e?.message === 'IMS timeout') {
          window.addEventListener('onImsLibInstance', () => this.imsReady());
          return;
        }
        lanaLog({ message: 'GNAV: Error with IMS', e, tags: 'errorType=info,module=gnav' });
      });
```
      
What are your thoughts on this approach? @narcis-radu @mokimo @salonijain3

mokimo: Let's take a step back first:

What is the exact use-case for this? Where/how did the requirement come from, and which issue is this supposed to solve?

bandana147: With standalone navigation, consumers like Adobe Home want to load their own IMS. They use IMS before rendering global navigation, as it's apparently needed for other checks as well.

bandana147: [Wiki link ](https://wiki.corp.adobe.com/pages/viewpage.action?pageId=3285416619#MilobasedGNav&FooterIntegrationinAdobeHome-QueriesforMiloteam)to their requirement request.

mokimo: So, the case is:
- The page (non-milo Adobe Home) already loads IMS
- The page embeds the standalone GNAV
- GNAV should not initiate loading IMS

I don't think we should make it a general milo pattern that consumers can define their own IMS loading mechanism. This should be steered by milo which should provide the core logic for consuming projects.

The GNAV works as standalone version and hence should add the logic separately IMO. I'd suggest adapting the [GNAV logic to check whether IMS](https://github.com/adobecom/milo/blob/stage/libs/blocks/global-navigation/global-navigation.js#L325) has already loaded

mokimo: e: I saw this is basically the same thing you proposed, i somehow was still in the mindset of the milo utils 😆

mokimo: For the adobe scenario you can be sure IMS already finished loading, hence for now I'd just move the check to the GNAV
`ims = async () => window.adobeIMS?.initialized ? this.imsReady() : loadIms()`

bandana147: Yes, This sounds good. I will update the code as so.

bandana147:
> I just realized `imsLoaded = window.adobeIMS?.initialized;` will be a boolean rather than promise, if someone calls `loadIms.then()` we'd return a boolean and have JS errors trying to access `true.then`
> 
> Would be great to change the return value to a promise that resolves and write a unit test for this scenario.

Updated the code.

mokimo (approved):
Cool, this looks simple, clean & full fills the adobe home request. Love it 🚀

overmyheadandbody (approved):

spadmasa:
Validation done on below url
https://gnav--milo--adobecom.hlx.page/drafts/blaishram/unav-test?martech=off
image

image

check sign in functionality
https://main--cc--adobecom.hlx.page/creativecloud?milolibs=gnav--milo--adobecom
image

PR #2804, opened 2024-08-29
[MWPW-157017] Send a Link Feature - Implementation

drashti1712:
* Implementation of send-a-link feature (**mobile-only**)
* The solution is to append a `#_evt-copy` parameter to the link/button, which shows a share pop-up on clicking and enables the users to copy and send the link to themselves.
* GitHub discussion link: https://github.com/orgs/adobecom/discussions/2790

Resolves: [MWPW-157017](https://jira.corp.adobe.com/browse/MWPW-157017)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.live/drafts/drashti/share-links?martech=off
- After: https://share-link-updated--milo--drashti1712.hlx.live/drafts/drashti/share-links?martech=off


libs/utils/utils.js
```diff
@@ -656,6 +656,21 @@ export function decorateLinks(el) {
         window.adobeIMS?.signIn();
       });
     }
+    const copyEvent = '#_evt-copy-link';
+    if (a.href.includes(copyEvent)) {
+      const link = a.href.split('#_evt-copy-link')[0];
```
mokimo, line 661:
```suggestion
      const link = a.href.split(copyEvent)[0];
```

libs/utils/utils.js
```diff
@@ -656,6 +656,21 @@ export function decorateLinks(el) {
...
+      const isConButton = ['EM', 'STRONG'].includes(a.parentElement.nodeName) || a.classList.contains('con-button');
+      if (!isConButton) a.classList.add('static', 'copy-link');
+      a.href = '';
+      if (navigator.share) {
+        a.addEventListener('click', async (e) => {
+          e.preventDefault();
+          await navigator.share({ title: link, url: link });
+        });
```
mokimo, line 669: Do we not break the links on firefox? It seems it does not have any support for it at all.

mokimo (commented):
A unit test should also be added.

libs/utils/utils.js
drashti1712: Yes, this is not supported by some desktop browsers. However, this is a mobile-only feature and the API is supported for all the major mobile browsers.

drashti1712:
> A unit test should also be added.

I am currently working on the unit tests. Will them them by EOD.
Thanks.

drashti1712:
> > A unit test should also be added.
> 
> I am currently working on the unit tests. Will add them by EOD. Thanks.

Update: I have added the required unit tests now. This PR is ready to be reviewed.

robert-bogos (approved):
Looks good 👍

mokimo (commented):
I'm getting some weird behavior using this on a browser where this is not suppoted.
(Latest chrome/macOS)

https://github.com/user-attachments/assets/b0f9ebea-53a9-4587-aa92-ba8a4992f191

libs/utils/utils.js
mokimo: See https://github.com/adobecom/milo/pull/2804#pullrequestreview-2272059085 - I think the default fallback behavior is wrong, an empty href leads to a refresh.

libs/utils/utils.js
```diff
@@ -656,6 +656,21 @@ export function decorateLinks(el) {
         window.adobeIMS?.signIn();
       });
     }
+    const copyEvent = '#_evt-copy';
```
mokimo, line 659: How often is this feature used? It seems like a lot of code for something that might only be used on a couple of pages... I wonder if it would make sense to have a post-LCP method that attaches event listeners / does sth special with links?

I'm not certain that's really solvable in this PR, but I see the `decorateLinks` method growing and doubt it will stop anytime soon with the special cases.

libs/utils/utils.js
```diff
@@ -656,6 +656,21 @@ export function decorateLinks(el) {
...
+    if (a.href.includes(copyEvent)) {
+      const link = a.href.split(copyEvent)[0];
+      const isConButton = ['EM', 'STRONG'].includes(a.parentElement.nodeName) || a.classList.contains('con-button');
+      if (!isConButton) a.classList.add('static', 'copy-link');
```
mokimo, line 663: Another thought:
How about if we have those events, we just emit events, and the consumer side code can add their logic accordingly? 

The only remark I would have: Since you change classes within the scope of this event, this won't work or introduce CLS. Can't the style part of this be authored otherwise? 🤔

libs/utils/utils.js
drashti1712: Ideally this issue should not be seen because the mobile-specific case would be handled by mep and hence, we should not see the block ( with `#_evt-copy`) on browsers where the API is not supported. Nonetheless, I do agree that we need to have a proper fallback and had asked this question on jira.
To prevent the refresh, I can simply add the if check inside `addEventListener` after `e.preventDefault()`. This would prevent the refresh, but the click behavior of the link won't function.
@mokimo What do you think about this? Or do you suggest to redirect to the original link wherever this is not supported?

mokimo: Yes, I think we should handle this case. Even though _you_ intend that this feature is used on mobile through MEP, it exists and can be authored. 
We should ensure this works in any case or remove the element in cases when it suddenly appears (desktop) when it should not, guarding authors from using this feature in an unintended way.

libs/utils/utils.js
drashti1712: This feature is intended for use on the Features pages for Ps, Pr, Lr, Ai, and Id specifically for MAX, with After Effects to follow post-MAX:
[https://www.adobe.com/products/photoshop/features.html](https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fwww.adobe.com%2Fproducts%2Fphotoshop%2Ffeatures.html&data=05%7C02%7Cdmodasara%40adobe.com%7C183a6b2ec9b54f5953f008dcc5f9be13%7Cfa7b1b5a7b34438794aed2c178decee1%7C0%7C0%7C638602923624957894%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C0%7C%7C%7C&sdata=YQ4APsbTtnC2N7fvPxg09t0%2F7DjNkpbr4ya2Pb%2FdiLI%3D&reserved=0)
[https://www.adobe.com/products/photoshop-lightroom/features.html](https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fwww.adobe.com%2Fproducts%2Fphotoshop-lightroom%2Ffeatures.html&data=05%7C02%7Cdmodasara%40adobe.com%7C183a6b2ec9b54f5953f008dcc5f9be13%7Cfa7b1b5a7b34438794aed2c178decee1%7C0%7C0%7C638602923624974691%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C0%7C%7C%7C&sdata=dkdJqhDrztdzYbBQyiRPljCb3Xh3xvDCkeENOF86usg%3D&reserved=0)
[https://www.adobe.com/products/illustrator/features.html](https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fwww.adobe.com%2Fproducts%2Fillustrator%2Ffeatures.html&data=05%7C02%7Cdmodasara%40adobe.com%7C183a6b2ec9b54f5953f008dcc5f9be13%7Cfa7b1b5a7b34438794aed2c178decee1%7C0%7C0%7C638602923624988916%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C0%7C%7C%7C&sdata=Hx0R9AAsP2hOj5SI38ZVVM7YxFqxL7MARlQZxLd9y4k%3D&reserved=0)
[https://www.adobe.com/products/premiere/features.html](https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fwww.adobe.com%2Fproducts%2Fpremiere%2Ffeatures.html&data=05%7C02%7Cdmodasara%40adobe.com%7C183a6b2ec9b54f5953f008dcc5f9be13%7Cfa7b1b5a7b34438794aed2c178decee1%7C0%7C0%7C638602923625001187%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C0%7C%7C%7C&sdata=6e9aS6q4MoEjMUg%2Bw34geqy3fvjf2BDnb1pjKzTC6Y0%3D&reserved=0)
[https://www.adobe.com/products/indesign/features.html](https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fwww.adobe.com%2Fproducts%2Findesign%2Ffeatures.html&data=05%7C02%7Cdmodasara%40adobe.com%7C183a6b2ec9b54f5953f008dcc5f9be13%7Cfa7b1b5a7b34438794aed2c178decee1%7C0%7C0%7C638602923625012513%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C0%7C%7C%7C&sdata=4XD01AttLL8KZI8hncyqWQN%2FYPpWbsAiruuAkDvEz88%3D&reserved=0)

I am not sure if we have to time to explore alternative approaches and implement them, given that this is an urgent requirement for MAX and we need to merge this PR by 9/5. However, I do agree that the growing complexity of the `decorateLinks` method is something we should address in the future.

libs/utils/utils.js
drashti1712: I think there is some misunderstanding around this as I am simply adding a bold style to the link if it is a not a button. According to the design, the link was kind of a bold static link. Since, I am not changing/replacing any class, I am not sure how could this cause CLS? Could you please elaborate more or let me know if I have understood this correctly?

libs/utils/utils.js
drashti1712: Sure, let me remove the element in case `navigator.share()` is not supported. Thank you for your input on this.

libs/utils/utils.js
mokimo: Yup, just thinking out loud 😁 
As I said, in the scope of this PR fixing the use-case at hand will be fine 👍

libs/utils/utils.js
mokimo: Nono, it's connected to https://github.com/adobecom/milo/pull/2804#discussion_r1738503680 - thinking about a possible future path on how we can handle this more generically. Maybe the discussion shouldn't happen on this PR, but it surfaces the issue 😁

libs/utils/utils.js
```diff
@@ -656,6 +656,23 @@ export function decorateLinks(el) {
...
+      if (isMobile) {
+        a.addEventListener('click', async (e) => {
+          e.preventDefault();
+          if (navigator.share) await navigator.share({ title: link, url: link });
+        });
+      } else {
+        a.remove();
+        console.log('Copy link is not supported for this device');
+      }
+    }
```
mokimo, line 675: I'd probably restructure this a little, untested demo code:
```suggestion
    const copyEvent = '#_evt-copy';
    const decorateCopyEvt = () => {
      const isMobile = /android|iphone|ipod|blackberry|iemobile|opera mini|mobile/.test(navigator.userAgent.toLowerCase());
      if (!isMobile) {
        a.remove();
        console.log('Copy link is not supported for this device');
        return;
      }
      const link = a.href.split(copyEvent)[0];
      const isConButton = ['EM', 'STRONG'].includes(a.parentElement.nodeName) || a.classList.contains('con-button');
      if (!isConButton) a.classList.add('static', 'copy-link');
      a.href = '';
      a.addEventListener('click', async (e) => {
        e.preventDefault();
        if (navigator.share) await navigator.share({ title: link, url: link });
      });
    };
```

drashti1712:
Since this is a mobile-only feature, I have added a check for mobile devices which would remove the element if failed. This serves as a fallback behavior if this is mistakenly authored for desktop/tablet.

libs/utils/utils.js
```diff
@@ -620,6 +620,24 @@ export function decorateAutoBlock(a) {
   });
 }
 
+const decorateCopyLink = (a, evt) => {
+  const userAgent = navigator.userAgent.toLowerCase();
+  const isMobile = /android|iphone|ipod|blackberry|iemobile|opera mini|mobile/.test(userAgent) && !/ipad/.test(userAgent);
+  if (!isMobile) {
```
overmyheadandbody, line 626: Should we also return early and remove the element if `navigator.share` is not supported?

libs/utils/utils.js
```diff
@@ -620,6 +620,24 @@ export function decorateAutoBlock(a) {
...
+    a.remove();
+    console.log('Copy link is not supported for this device');
+    return;
+  }
+  const link = a.href.split(evt)[0];
```
overmyheadandbody, line 631: I was thinking whether we should simply `replace` the `evt` with `''`. Maybe the link is something like `https://www.adobe.com#_evt-copy?promoId=MAX`, case in which the `promoId` could be lost. It shouldn't be a regular occurrence, but it could lead to issues if an someone was not paying attention to the URL structure.

drashti1712: Hi @overmyheadandbody 
I agree that `promoId` would be lost with the approach. In the above example, do we want the `promoId` to be a part of the URL when the user wants to copy and share it? The URL shared would be `https://www.adobe.com?promoId=MAX` in that case. If this is expected, then I can go ahead and replace the param.

libs/utils/utils.js
```diff
@@ -620,6 +620,24 @@ export function decorateAutoBlock(a) {
   });
 }
 
+const decorateCopyLink = (a, evt) => {
+  const userAgent = navigator.userAgent.toLowerCase();
+  const isMobile = /android|iphone|ipod|blackberry|iemobile|opera mini|mobile/.test(userAgent) && !/ipad/.test(userAgent);
+  if (!isMobile) {
+    a.remove();
+    console.log('Copy link is not supported for this device');
```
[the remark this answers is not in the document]
drashti1712: Sure, fixed this. Thanks!

libs/utils/utils.js
drashti1712: Sure, made these changes. Thanks!

libs/utils/utils.js
overmyheadandbody: That should be confirmed with the feature's stakeholder. In my view we should keep query params. They might include some referrers, so I don't see why we would automatically remove them 🤷

libs/styles/styles.css
```diff
@@ -808,6 +808,10 @@ a.static:hover {
   text-decoration: underline;
 }
 
+.copy-link {
+  font-weight: bold;
```
narcis-radu, line 812: we should use a numeric value

libs/utils/utils.js
drashti1712: @overmyheadandbody As confirmed by the stakeholders, the query params should be reserved. Hence, I have made the required change. Thank you for your feedback.

mokimo (approved):

libs/utils/utils.js
```diff
@@ -620,6 +620,23 @@ export function decorateAutoBlock(a) {
   });
 }
 
+const decorateCopyLink = (a, evt) => {
+  const userAgent = navigator.userAgent.toLowerCase();
+  const isMobile = /android|iphone|ipod|blackberry|iemobile|opera mini|mobile/.test(userAgent) && !/ipad/.test(userAgent);
```
chrischrischris, line 625: I really don't think we need to worry about `blackberry`, `iemobile` and `opera mini`

@overmyheadandbody @narcis-radu Thoughts?

libs/utils/utils.js
```diff
@@ -620,6 +620,23 @@ export function decorateAutoBlock(a) {
...
+  if (!isMobile || !navigator.share) {
+    a.remove();
+    return;
+  }
+  const link = a.href.replace(evt, '');
+  const isConButton = ['EM', 'STRONG'].includes(a.parentElement.nodeName) || a.classList.contains('con-button');
```
chrischrischris, line 631: nit: line length should be < 100

libs/utils/utils.js
overmyheadandbody: I agree, good observation! I'd also argue that `ipod` is not needed. Per [Wikipedia](https://en.wikipedia.org/wiki/IPod), iPods have been discontinued on May 10, 2022. Looking at our RUM data, I couldn't find users navigating from iPods.

overmyheadandbody (approved):
Thanks for all the updates, @drashti1712!

spadmasa:
Validation has been done on the below url in different devices - and on below browser in device (safari, chrome , samsung, firefox )
https://main--cc--adobecom.hlx.page/drafts/souj/sendlink/features1?milolibs=share-link-updated--milo--drashti1712
on the desktop and ipad link will not be shown
image
image

In mobile (iphone and android) in landscape and portrait mode
https://main--cc--adobecom.hlx.page/drafts/souj/sendlink/sendlink?milolibs=share-link-updated--milo--drashti1712
image
image
image

Validations details are updated in https://jira.corp.adobe.com/browse/MWPW-157557

PR #2806, opened 2024-08-30
MWPW-157540: fix external modals

3ch023:
Go live 4th September blocker.
Fix external modals on catalog.
checkoutLinkConfig[columnName] is an external CRM modal `https://www.adobe.com/plans-fragments/modals/individual/modals-content-rich/all-apps/master.modal.html `
that gets localized and becomes '/plans-fragments/modals/individual/modals-content-rich/all-apps/master.modal.html'
that doesn't pass regex later on on openModal
`} else if (/^https?:/.test(url)) {`

Resolves: https://jira.corp.adobe.com/browse/MWPW-157540

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.live/drafts/mariia/pr/price-cta?martech=off
- After: https://MWPW-157540--milo--adobecom.hlx.live/drafts/mariia/pr/price-cta?martech=off

### DIFF
--- a/libs/blocks/merch/merch.js
+++ b/libs/blocks/merch/merch.js
@@ -374,7 +374,7 @@ export async function openModal(e, url, offerType) {
   if (/\/fragments\//.test(url)) {
     const fragmentPath = url.split(/hlx.(page|live)/).pop();
     modal = await openFragmentModal(fragmentPath, getModal);
-  } else if (/^https?:/.test(url)) {
+  } else {
     modal = await openExternalModal(url, getModal);
   }
   if (modal) {


yesil:
https://www.adobe.com/plans-fragments/modals/individual/modals-content-rich/all-apps/master.modal.html is not a valid url for the feature.  It was only developed for TWP and D2P modals and tested against.

yesil (approved):
Discussed with @3ch023  and looks good to me.

robert-bogos (approved):

libs/blocks/merch/merch.js
```diff
@@ -374,7 +374,7 @@ export async function openModal(e, url, offerType) {
   if (/\/fragments\//.test(url)) {
     const fragmentPath = url.split(/hlx.(page|live)/).pop();
     modal = await openFragmentModal(fragmentPath, getModal);
-  } else if (/^https?:/.test(url)) {
```
npeltier, line 377: can we rather try `/^(https)?\//` ? this way it's either https (original requirement) or relative path?

npeltier (changes requested):

libs/blocks/merch/merch.js
npeltier: ok if we neither wants https://... URLs i'm fine thought relative paths were an option

npeltier (approved):
lgtm (thought originally there was something to forbid http:// & relative paths)

afmicka:
Looks good to me too. Have checked it also together with @3ch023 and added the verified label.
@Roycethan  could you please also have a separate look and mark it Ready for stage if all good on your end as well?

Roycethan:
@3ch023 @afmicka i think you might have simulated the working fix, have covered regressions and marking to stage, lets check once its in Main, or if possible on stage.

PR #2807, opened 2024-08-30
MWPW-156357 Price-literals - fail gracefully

bozojovicic:
Use default price literals in english if the call for `price-literals.json` fails.

Resolves: [MWPW-156357](https://jira.corp.adobe.com/browse/MWPW-156357)

**Test URLs:**
- After: https://mwpw156357literals--milo--bozojovicic.hlx.page/drafts/bozo/price-cta?martech=off


bozojovicic:
In Dexter we use English as default in this case so I applied the same solution here. However you decide. Where would I save price-literals.json then?

3ch023 (changes requested):
Looks good! (and thank you for taking this ticket)
could you also add a catch block for the cases when price-literals file is not accessible at all?
similar to https://github.com/adobecom/milo/blob/stage/libs/blocks/merch/merch.js#L192 
with the same fallback to defaults.

when requests fails fetch will throw an error causing the blank page, so if (response.ok) is not enough to make us fail-safe.
I test it by blocking request in devtools
image


p.s. also code coverage needs some love ^^

yesil:
@bozojovicic  libs/features/mas/commerce, I think this would look good :)

3ch023:
>Instead of falling back to default that is english, I propose to copy price-literals.json to Milo and fall back to it.
We can update it from time to time.

if we copy them all to code then I vote for removing the authoring excel completely. - network call.
if we keep the pricing-literals.json request, then IMO en default should be enough, just to also take size of file into consideration. small thing after another small thing and it adds up, no? I am not sure I see benefit of storing multiple translations in our code that will never be used/updated.

bozojovicic:
New PR opened https://github.com/adobecom/milo/pull/2815. Closing this one.

PR #2812, opened 2024-09-02
[MWPW-156244] Reduce TBT Youtube/Vimeo Videos

robert-bogos:
## Description
With this PR, we will load facades instead of actual videos for YouTube and Vimeo, reducing TBT. The facade will be replaced with the real video only when clicked by the user.

## Related Issue
Resolves: [MWPW-156244](https://jira.corp.adobe.com/browse/MWPW-156244)

## Screenshots:
**Before**

image

**After**

image

## Test URLs
- Before: https://main--milo--robert-bogos.hlx.page/drafts/rbogos/youtube-vimeo?martech=off
- After: https://mwpw-156244-reduce-tbt--milo--robert-bogos.hlx.page/drafts/rbogos/youtube-vimeo?martech=off


libs/blocks/vimeo/vimeo.js
```diff
@@ -1,25 +1,81 @@
...
+    this.addEventListener('click', this.addIframe);
+  }
+
+  static warmConnections() {
+    if (LiteVimeo.preconnected) return;
+    LiteVimeo.preconnected = true;
+    addPrefetch('preconnect', 'https://player.vimeo.com');
+    addPrefetch('preconnect', 'https://i.vimeocdn.com');
+    addPrefetch('preconnect', 'https://f.vimeocdn.com');
+    addPrefetch('preconnect', 'https://fresnel.vimeocdn.com');
```
mokimo, line 28: We should be able to reuse the loadLink method similar to what Chris did here: https://github.com/adobecom/cc/pull/383/files#diff-b1f0360c1bbc27f3a78b118f8f7ff551a12fc06e54a5606e668b4846b26c1ac0R179

libs/blocks/vimeo/vimeo.js
```diff
@@ -1,25 +1,81 @@
...
-      src,
-      style: 'border: 0; top: 0; left: 0; width: 100%; height: 100%; position: absolute;',
+function addPrefetch(kind, url, as) {
+  const linkElem = createTag('link', { rel: kind, href: url });
+  if (as) linkElem.as = as;
+  linkElem.crossorigin = true;
+  document.head.append(linkElem);
+}
+
+class LiteVimeo extends HTMLElement {
```
mokimo, line 10: Whats the extends `HTMLElement` used for?

libs/blocks/vimeo/vimeo.js
```diff
@@ -1,25 +1,81 @@
...
+  setupThumbnail() {
+    const { width, height } = this.getBoundingClientRect();
+    const roundedWidth = Math.min(Math.ceil(width / 100) * 100, 1920);
+    const roundedHeight = Math.round((roundedWidth / width) * height);
+
+    fetch(`https://vimeo.com/api/v2/video/${this.videoId}.json`)
+      .then((response) => response.json())
+      .then((data) => {
+        const thumbnailUrl = data[0].thumbnail_large?.replace(/-d_[\dx]+$/i, `-d_${roundedWidth}x${roundedHeight}`);
+        this.style.backgroundImage = `url("${thumbnailUrl}")`;
```
mokimo, line 40: Is there a source for this, or was this just copied from express? Looks a little funky

libs/blocks/vimeo/vimeo.js
```diff
@@ -1,25 +1,81 @@
...
+}
+
+if (!customElements.get('lite-vimeo')) customElements.define('lite-vimeo', LiteVimeo);
 
+export default async function init(a) {
+  if (isInTextNode(a)) return;
+
+  const embedVimeo = () => {
+    const url = new URL(a.href);
+    const videoid = url.pathname.split('/')[url.hostname === 'player.vimeo.com' ? 2 : 1];
```
mokimo, line 76: Also here, any source for this?

libs/blocks/youtube/youtube.js
```diff
@@ -1,36 +1,111 @@
-import { createIntersectionObserver, isInTextNode } from '../../utils/utils.js';
+import { createIntersectionObserver, createTag, isInTextNode } from '../../utils/utils.js';
```
mokimo, line 1: Same questions about a source in this file, some things look a little different than how we'd usually do them or very specific (e.g. copied from a LiteYTEmbed example)

Maybe this comes from express though, and their coding style is a little different 🤔

mokimo (approved):
Generally looks good and if this helps to reduce TBT that's a great performance win 🚀

robert-bogos:
Code coverage will be improved in [this ticket](https://jira.corp.adobe.com/browse/MWPW-157600).
Currently we stand at 83% for Youtube and 96% for Vimeo.

mokimo:
On a side note, I wonder if we can optimize the loading order? It seems that it only starts to load the image after the json & iframe css. Maybe we can shift things around a little to start loading things earlier in case YT/Vimeo are the LCP?

image

libs/blocks/vimeo/vimeo.js
robert-bogos: Is for creating the custom `lite-youtube` and `lite-vimeo` html elements. 
Part of the code is an optimized version of [lite-youtube-embed](https://github.com/paulirish/lite-youtube-embed) and [lite-vimeo-embed](https://github.com/luwes/lite-vimeo-embed), which are also used in Express.

libs/blocks/vimeo/vimeo.js
```diff
@@ -1,25 +1,81 @@
...
+        const thumbnailUrl = data[0].thumbnail_large?.replace(/-d_[\dx]+$/i, `-d_${roundedWidth}x${roundedHeight}`);
+        this.style.backgroundImage = `url("${thumbnailUrl}")`;
+      })
+      .catch((e) => {
+        window.lana.log(`Error fetching Vimeo thumbnail: ${e}`, { tags: 'errorType=info,module=vimeo' });
+      });
+  }
+
+  setupPlayButton() {
+    const playBtnEl = createTag('button', { type: 'button', 'aria-label': 'Play video', class: 'ltv-playbtn' });
```
vhargrave, line 48: Ideally this aria-label would be translated for accessability. I think we're not even passing a label in Express though, which is also not great.

vhargrave (approved):
Wow, what a huge difference to TBT! 
Found a small things for accessability, but the rest looks good to me!

libs/blocks/vimeo/vimeo.js
vhargrave: tested it here as well btw - https://main--bacom--adobecom.hlx.live/customer-success-stories/coca-cola-case-study#watch-now

mokimo: > Ideally this aria-label would be translated for accessability. I think we're not even passing a label in Express though, which is also not great.

I've seen it a lot that this currently doesn't happen 😞

libs/blocks/vimeo/vimeo.js
```diff
@@ -1,25 +1,81 @@
...
+
+  setupThumbnail() {
+    const { width, height } = this.getBoundingClientRect();
+    const roundedWidth = Math.min(Math.ceil(width / 100) * 100, 1920);
+    const roundedHeight = Math.round((roundedWidth / width) * height);
+
+    fetch(`https://vimeo.com/api/v2/video/${this.videoId}.json`)
+      .then((response) => response.json())
+      .then((data) => {
+        const thumbnailUrl = data[0].thumbnail_large?.replace(/-d_[\dx]+$/i, `-d_${roundedWidth}x${roundedHeight}`);
```
overmyheadandbody, line 39: Let's go the extra mile with `data?.[0]`

libs/blocks/vimeo/vimeo.js
```diff
@@ -1,25 +1,81 @@
...
+  setupPlayButton() {
+    const playBtnEl = createTag('button', { type: 'button', 'aria-label': 'Play video', class: 'ltv-playbtn' });
+    this.append(playBtnEl);
+  }
+
+  addIframe() {
+    if (this.classList.contains('ltv-activated')) return;
+    this.classList.add('ltv-activated');
+    const iframeEl = createTag('iframe', {
+      style: 'border: 0; top: 0; left: 0; width: 100%; height: 100%; position: absolute; background-color: #000;',
```
overmyheadandbody, line 56: This is quite similar to https://github.com/adobecom/milo/pull/2812/files#diff-61fb0bef520820f126cd9fda15bcc79e246f1db0a89715f9a5da21a1ee832a58R20-R27, do we need the redundancy?

libs/blocks/vimeo/vimeo.js
```diff
@@ -1,25 +1,81 @@
...
+      src: `https://player.vimeo.com/video/${encodeURIComponent(this.videoId)}?autoplay=1`,
     });
-    const wrapper = createTag('div', { class: 'embed-vimeo' }, iframe);
+    this.insertAdjacentElement('afterend', iframeEl);
+    this.remove();
+    iframeEl.addEventListener('load', () => iframeEl.focus(), { once: true });
+  }
+}
+
+if (!customElements.get('lite-vimeo')) customElements.define('lite-vimeo', LiteVimeo);
```
overmyheadandbody, line 69: Does this need to be here or can we move it inside the `init`, after the early return?

libs/blocks/youtube/youtube.js
```diff
@@ -1,36 +1,111 @@
...
+    document.head.append(linkEl);
+  }
+
+  static warmConnections() {
+    if (LiteYTEmbed.preconnected) return;
+    LiteYTEmbed.addPrefetch('preconnect', 'https://www.youtube-nocookie.com');
+    LiteYTEmbed.addPrefetch('preconnect', 'https://www.google.com');
+    LiteYTEmbed.addPrefetch('preconnect', 'https://googleads.g.doubleclick.net');
+    LiteYTEmbed.addPrefetch('preconnect', 'https://static.doubleclick.net');
+    LiteYTEmbed.preconnected = true;
```
overmyheadandbody, line 32: OCD nit: for Vimeo you declare this before adding the preconnects. None is better or worse, but consistency is always nicer on the eye 🙂

overmyheadandbody (commented):
Is the Vimeo poster actually being used? The image seems to be downloaded, but having a quick look at the timeline, I just see a black rectangle.

Screenshot 2024-09-02 at 12 31 31

vhargrave:
> On a side note, I wonder if we can optimize the loading order? It seems that it only starts to load the image after the json & iframe css. Maybe we can shift things around a little to start loading things earlier in case YT/Vimeo are the LCP?
> 
> image

I love this idea! 🚀

libs/blocks/vimeo/vimeo.js
robert-bogos: Part of the code is an optimized version of [lite-youtube-embed](https://github.com/paulirish/lite-youtube-embed) and [lite-vimeo-embed](https://github.com/luwes/lite-vimeo-embed), that's where this is from.

libs/blocks/vimeo/vimeo.js
robert-bogos: Part of the code is an optimized version of [lite-youtube-embed](https://github.com/paulirish/lite-youtube-embed) and [lite-vimeo-embed](https://github.com/luwes/lite-vimeo-embed), that's where this is from.

libs/blocks/youtube/youtube.js
robert-bogos: Part of the code is an optimized version of [lite-youtube-embed](https://github.com/paulirish/lite-youtube-embed) and [lite-vimeo-embed](https://github.com/luwes/lite-vimeo-embed), that's why it looks a bit different than how we usually do things.

mokimo: I wonder if we should link to the two projects within the files. It's not obvious from the get-go.

I'm personally fine with having the PR as history, but we could consider a comment with the source

libs/blocks/vimeo/vimeo.js
robert-bogos: I've added placeholders for that in my feedback commit 👍

libs/blocks/youtube/youtube.js
robert-bogos: That's a good idea. I've added the links in my feedback commit 👍

libs/blocks/vimeo/vimeo.js
robert-bogos: I don't want to contribute to that so I've added a placeholder 👍

libs/blocks/vimeo/vimeo.css
```diff
@@ -5,3 +5,68 @@
...
+  left: 0;
+  border: 0;
+}
+
+lite-vimeo > .ltv-playbtn {
+  font-size: 10px;
+  padding: 0;
+  width: 6.5em;
+  height: 4em;
+  background: rgba(23, 35, 34, .75);
```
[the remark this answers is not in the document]
robert-bogos: Made the change 👍

libs/blocks/vimeo/vimeo.css
```diff
@@ -5,3 +5,68 @@
...
+
+lite-vimeo:hover > .ltv-playbtn {
+  background-color: rgb(0, 173, 239);
+  opacity: 1;
+}
+
+lite-vimeo > .ltv-playbtn::before {
+  content: '';
+  border-style: solid;
+  border-width: 10px 0 10px 20px;
```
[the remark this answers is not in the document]
robert-bogos: Not the case in here because that css is for the play icon, which I don't think we want to reverse on RTL.

libs/blocks/vimeo/vimeo.js
robert-bogos: Good catch! 🦅👁️ That [CSS](https://github.com/adobecom/milo/pull/2812/files#diff-61fb0bef520820f126cd9fda15bcc79e246f1db0a89715f9a5da21a1ee832a58R20-R27) is not needed as I remove the lite-vimeo element after adding the iframe. Removed the CSS 👍

libs/blocks/vimeo/vimeo.js
robert-bogos: Doesn't need to be there, moved it inside the `init` for both blocks 👍

libs/blocks/youtube/youtube.js
robert-bogos: Now that you've pointed it out I cannot unsee it and my OCD kicked in 😂 I've made the change 👍

robert-bogos:
> Is the Vimeo poster actually being used? The image seems to be downloaded, but having a quick look at the timeline, I just see a black rectangle.
@overmyheadandbody Yes, it is used, you can see is changing in the timeline. That black is from the background color of the `lite-vimeo`, and we need that for a smooth transition when the `lite-vimeo` is replaced by the iframe.

robert-bogos:
> On a side note, I wonder if we can optimize the loading order? It seems that it only starts to load the image after the json & iframe css. Maybe we can shift things around a little to start loading things earlier in case YT/Vimeo are the LCP?
@mokimo I don't think we can do much about those because the image url comes from the JSON, so the json needs to be first and the JSON > image is the first thing we load when creating the custom yt/vimeo elements so is as close to iframe css as possible. I don't know if I'm missing something 🤔

libs/blocks/vimeo/vimeo.js
```diff
@@ -1,25 +1,80 @@
...
-    }
-    const iframe = createTag('iframe', {
-      src,
-      style: 'border: 0; top: 0; left: 0; width: 100%; height: 100%; position: absolute;',
+class LiteVimeo extends HTMLElement {
+  static preconnected = false;
+
+  connectedCallback() {
+    this.videoId = this.getAttribute('videoid');
+    this.language = this.getAttribute('language') || '';
```
narcis-radu, line 10: Is this property used anywhere, I cannot find it ?!

libs/blocks/youtube/youtube.js
```diff
@@ -1,36 +1,107 @@
...
+    ['www.youtube-nocookie.com',
+      'www.google.com',
+      'googleads.g.doubleclick.net',
+      'static.doubleclick.net',
+    ].forEach((url) => loadLink(`https://${url}`, { rel: 'preconnect' }));
+  }
+
+  static loadYouTubeAPI() {
+    return new Promise((resolve) => {
+      if (window.YT && window.YT.Player) {
```
narcis-radu, line 33: nit: you can use `window.YT?.Player`

libs/blocks/youtube/youtube.js
```diff
@@ -1,36 +1,107 @@
...
+        resolve();
+        return;
+      }
+      const tag = createTag('script', { src: 'https://www.youtube.com/iframe_api' });
+      window.onYouTubeIframeAPIReady = resolve;
```
narcis-radu, line 38: I will assume this is used by the iframe_api script

narcis-radu (approved):

libs/blocks/vimeo/vimeo.js
robert-bogos: No, it got left behind. I removed it 👍

libs/blocks/youtube/youtube.js
robert-bogos: Thanks 👍

libs/blocks/youtube/youtube.js
robert-bogos: Yes, it is

overmyheadandbody (approved):
Thanks for all the updates!

SilviuLCF:
@robert-bogos On Mobile Iphone (safari and Chrome) , clicking on facade play will not start the video, will load the video iframe on pause

robert-bogos:
> @robert-bogos On Mobile Iphone (safari and Chrome) , clicking on facade play will not start the video, will load the video iframe on pause

@SilviuLCF Fixed 👍

SilviuLCF (approved):
Verified , testing details can be found here
https://jira.corp.adobe.com/browse/MWPW-156244

PR #2798, opened 2024-08-28
MWPW-155940: MAS AEM client 

yesil:
aem.js in mas is a partial client for several of the Sites APIs (https://developer.adobe.com/experience-cloud/experience-manager-apis/api/stable/sites/) that will support merch-card coming from Odin, as well as authoring of merch-cards in MAS Studio.

Also, this PR adds a github workflow to integrate MAS feature(libs/features/mas) unit tests to be run on PRs and send codecov reports.

Resolves: 
- [MWPW-155941](https://jira.corp.adobe.com/browse/MWPW-155941)
- [MWPW-156983](https://jira.corp.adobe.com/browse/MWPW-156983)
- [MWPW-157404](https://jira.corp.adobe.com/browse/MWPW-157404)
- [MWPW-157405](https://jira.corp.adobe.com/browse/MWPW-157405)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ilyas/merch-card?martech=off
- After: https://mwpw-155940--milo--yesil.hlx.page/drafts/ilyas/merch-card?martech=off

Note: this PR replaces https://github.com/adobecom/milo/pull/2761 (closed)

PR #2815, opened 2024-09-02
MWPW-156357 Price-literals - fail gracefully

bozojovicic:
`commerce.js` and `price-literals.json` are called always together and always at the same time. This PR injects price literals into `commerce.js` so there will be no need for some fallback service if `price-literals.json` fails. 
This change increases `commerce.js` for 5KB but **eliminates** the call for `price-literals.json` and hence eliminates the possibility of its failure. 

- old size of commerce.js is 26.6 KB
- new size of commerce.js 32.2 KB

It also removes the possibility for authors to change this file in sharepoint now when we have it in source code `/libs/features/mas/commerce/price-literals.json` but this content is changed only once during the last year.
I removed tons of code that we had only for loading of price literals.

Test [page](https://mwpw156357literals2--milo--bozojovicic.hlx.page/be_fr/drafts/mili/promo-placeholders)

Resolves: [MWPW-156357](https://jira.corp.adobe.com/browse/MWPW-156357)

**Test URLs:**
- After: https://mwpw156357literals2--milo--bozojovicic.hlx.page/drafts/bozo/price-cta?martech=off


libs/blocks/merch/merch.js
```diff
@@ -179,10 +180,21 @@ export async function fetchEntitlements() {
...
     fetch(url)
-      .then((response) => response.json().then(({ data }) => resolve(data)));
+      .catch(() => loadDefaultPriceLiterals())
+      .then(async (response) => {
+        if (response.ok) {
+          return response.json().then(({ data }) => resolve(data));
+        }
+        const defaultPriceLiterals = await loadDefaultPriceLiterals();
+        return defaultPriceLiterals.json().then(({ data }) => resolve(data));
+      });
```
mokimo, line 197: We usually write the `.then` block first and afterwards the `.catch`
I'd also chain the promises
```suggestion
      .then((response) => {
        if (response.ok) return response.json();
        return loadDefaultPriceLiterals()
          .then((defaultPriceLiterals) => defaultPriceLiterals.json());
      })
      .then(({ data }) => resolve(data));
      .catch(() => loadDefaultPriceLiterals())
```

bozojovicic: @mokimo are you sure? :) In method `fetchCheckoutLinkConfigs()` we have `fetch().catch().then();`
I chained promises as you describe but `catch()` needs to stay where it is since I load default json in `catch()` which is then processed in the following `then()` methods. I tried to move `catch()` to the last place but then nothing works of course. It seems that the order of handlers matters with `fetch()`.

mokimo: Are you sure you run into the `.catch` block? It's very rare to get into that.

https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch
> A fetch() promise only rejects when the request fails, for example, because of a badly-formed request URL or a network error. A fetch() promise does not reject if the server responds with HTTP status codes that indicate errors (404, 504, etc.). Instead, a then() handler must check the [Response.ok](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok) and/or [Response.status](https://developer.mozilla.org/en-US/docs/Web/API/Response/status) properties.

mokimo (approved):
I still think it's very rare to run into the catch block and wouldn't rely on the fetch to fail to begin with, but it's not a blocker to me.

libs/blocks/merch/merch.js
```diff
@@ -179,10 +180,20 @@ export async function fetchEntitlements() {
...
   fetchLiterals.promise = fetchLiterals.promise ?? new Promise((resolve) => {
     fetch(url)
-      .then((response) => response.json().then(({ data }) => resolve(data)));
+      .catch(() => loadDefaultPriceLiterals())
+      .then(async (response) => {
+        if (response.ok) return response.json();
+        return loadDefaultPriceLiterals()
+          .then((defaultPriceLiterals) => defaultPriceLiterals.json());
+      })
+      .then(({ data }) => resolve(data));
```
mokimo, line 196: I would suggest to change the `loadDefaultPriceLiterals` a little to return the data and call it `getDefaultPriceLiterals`
```suggestion
      .then((response) => response.json().then(({ data }) => resolve(data)))
      .catch(() => defaultPriceLiterals())
```

libs/blocks/merch/merch.js
bozojovicic: Sorry @mokimo we are changing the context for this PR like crazy. The code we are talking about here is now removed. I updated the PR description with new ideas.

libs/blocks/merch/merch.js
bozojovicic: This whole code is now removed.

libs/features/mas/commerce/src/price-literals.js
yesil, line 1: wouldn't esbuild import the js file?
https://esbuild.github.io/content-types/#json

yesil: I vote for keeping price-literals.json.

libs/features/mas/commerce/src/service.js
```diff
@@ -43,7 +43,7 @@ async function activateService(config, dataProviders) {
     const settings = Object.freeze(getSettings(config));
     // Fetch price literals
     try {
-        literals.price = await fetchPriceLiterals(settings);
+        literals.price = fetchPriceLiterals(settings);
```
yesil, line 46: I would not call it fetch since it is not fetch anymore.

bozojovicic: Right. Renamed to `getPriceLiterals`.

libs/features/mas/commerce/src/price-literals.js
bozojovicic: We have price-literals.json now. In "production" code we will `import('../price-literals.json')` but if we run the code from tests these literals will be received as an argument (loaded from test code from the same json file).

```
export async function getPriceLiterals(settings, priceLiterals) {
    const { data } = priceLiterals ? priceLiterals : await import('../price-literals.json');
```

I could not find other solution since unit tests do not use esbuild pipeline.

yesil (dismissed):

libs/deps/mas/commerce.js
```diff
@@ -1,4 +1,4 @@
...
 //# sourceMappingURL=commerce.js.map
```
narcis-radu, line 4: I know there's a ticket to handle this already, but sourcemaps references have to go (if the sourcemap is not available).

libs/features/mas/commerce/src/literals.js
```diff
@@ -1,29 +1,17 @@
...
+ * @param priceLiterals
  * @returns {Promise<Record<string, string>>}
  */
-export async function fetchPriceLiterals(settings) {
-    const priceLiteralsPromise =
-        settings.priceLiteralsPromise || generateLiteralsPromise(settings);
+export async function getPriceLiterals(settings, priceLiterals) {
     //we are expecting an array of objects with lang and literals
-    const data = await priceLiteralsPromise;
+    const { data } = priceLiterals ? priceLiterals : await import('../price-literals.json');
```
narcis-radu, line 14: I am a bit confused here; the description says:
>This change increases commerce.js for 5KB **but eliminates the call for price-literals.json** and hence eliminates the possibility of its failure.

If we're still calling price-literals, why aren't we taking it from content?

libs/features/mas/commerce/price-literals.json
```diff
@@ -0,0 +1 @@
...lusiveLabel":"{taxTerm, select, GST {GST अतिरिक्त} VAT {VAT अतिरिक्त} TAX {कर अतिरिक्त} IVA {IVA अतिरिक्त} SST {SST अतिरिक्त} KDV {KDV अतिरिक्त} other {}}","taxInclusiveLabel":"{taxTerm, select, GST {GST सहित} VAT {VAT सहित} TAX {कर सहित} IVA {IVA सहित} SST {SST सहित} KDV {KDV सहित} other {}}","alternativePriceAriaLabel":"वैकल्पिक रूप से इस पर {alternativePrice}","strikethroughAriaLabel":"नियमित रूप से इस पर {strikethroughPrice}"},{"lang":"iw","recurrenceLabel":"{recurrenceTerm, select, MONTH {/חודש} YEAR {/שנה} other {}}","recurrenceAriaLabel":"{recurrenceTerm, select, MONTH {לחודש} YEAR {לשנה} other {}}","perUnitLabel":"{perUnit, select, LICENSE {לרישיון} other {}}","perUnitAriaLabel":"{perUnit, select, LICENSE {לרישיון} other {}}","freeLabel":"חינם","freeAriaLabel":"חינם","taxExclusiveLabel":"{taxTerm, select, GST {ללא GST} VAT {ללא מע\"מ} TAX {ללא מס} IVA {ללא IVA} SST {ללא SST} KDV {ללא KDV} other {}}","taxInclusiveLabel":"{taxTerm, select, GST {כולל GST} VAT {כולל מע\"מ} TAX {כולל מס} IVA {כולל IVA} SST {כולל SST} KDV {כולל KDV} other {}}","alternativePriceAriaLabel":"לחלופין ב-{alternativePrice}","strikethroughAriaLabel":"באופן קבוע ב-{strikethroughPrice}"}],":type":"sheet"}
```
narcis-radu, line 1: >It also removes the possibility for authors to change this file in sharepoint now when we have it in source code /libs/features/mas/commerce/price-literals.json but this content is changed only once during the last year.
I removed tons of code that we had only for loading of price literals.

We should also consider geo expansion use-cases and also the speed of making a change when needed (e.g. changing "gratuit" to "gratis" for Romanian). I am a bit worried about this, I don't think we should keep content in the repository, but I do understand the point that things don't change that often.
@npeltier , @3ch023 , @yesil - what's your take on this?

yesil: Currently literals are language based only, and we support the same literals as the checkout page.
We already discussed supporting locales in price-literals.json instead of just languages, but AFAIK it is not a formal requirement yet.(@3ch023  keep me honest?)

For updating the wrong literals, it would be good to drop a message to `milo-commerce` with the description of the issue. Then we can follow-up with GWP.

For context: with MAS v1,  engineering took over of price literals and on Milo GWP wasn't meant to update the price-literals file on sharepoint.(it happened a few times under eng supervision).

Since literals do not change often, and are very critical, we want to prevent any issue caused by authoring and also reduce the impact of literal on the page performance, hence this PR.

libs/features/mas/commerce/src/literals.js
bozojovicic: @narcis-radu  We don't fetch `price-literals.json` any more, this is import, not fetch. There is no call `price-literals.json` ay more.
We are injecting it from source code into `commerce.js` with `import('../price-literals.json')` if it's not already provided with `priceLiterals` argument, which will be the case only for unit tests. In unit tests we cannot use `import` so I had to read it from source with `readFile` and then send it as an argument to `getPriceLiterals()`.

libs/deps/mas/commerce.js
yesil: with https://github.com/adobecom/milo/pull/2798 this should be fixed.

libs/features/mas/commerce/price-literals.json
3ch023: >but AFAIK it is not a formal requirement yet.
correct, no need for that as of now.

Just supporting the Ilyases statement above, price literals for commerce were decided to be managed by Eng, not GWP. We clarified it yesterday in the Warriors demo call.
And getting rid of an additional fetch request is just a nice bonus.

3ch023 (approved):

Axelcureno (approved):

yesil (approved):

PR #1463, opened 2023-10-27
feat: added tags to lana logs

robert-bogos:
### Description
This PR addresses the requirement of enabling filtering capabilities for Splunk logs based on tags from LANA logs. 

Recommended tags:
`info` - network issues
`warn` - authoring related misconfigs or similar
`error` - actual error ( ex. cannot read Y of undefind )
`<module>`- extra tag to help centralise all logs related to a specific module

### Related Issue
Resolves: [MWPW-138031](https://jira.corp.adobe.com/browse/MWPW-138031)

### Testing instructions
1. Change the index value in the Splunk search to `lana_nonprod`
2. Add a `tags` key with an array of strings as value to the json sent as argument for the `lanaLog` method 
or if you're using the `logErrorFor` method add the array of strings as the third argument ( the strings in the array are the tags that will be associated with the log )
E.g.
`lanaLog({ message: 'Could not create global navigation.', e,  tags: 'errorType=error,module=gnav'  });`
`logErrorFor(onToggleClick, 'Toggle click failed', 'errorType=error,module=gnav'  ));`
3. All the logs sent from your local environment should show in Splunk 

## Screenshots:
Screenshot 2023-10-27 at 16 31 50

## Test URLs
**Acrobat:**
- Before: https://main--dc--adobecom.hlx.page/drafts/rbogos/default?martech=off
- After: https://main--dc--adobecom.hlx.page/drafts/rbogos/default?milolibs=mwpw-138031-lana-logs-tags--milo--rbogos&martech=off

**BACOM:**
- Before: https://main--bacom--adobecom.hlx.page/drafts/rbogos/default
- After: https://main--bacom--adobecom.hlx.page/drafts/rbogos/default?milolibs=mwpw-138031-lana-logs-tags--milo--rbogos&martech=off

**CC:**
- Before: https://main--cc--adobecom.hlx.page/drafts/rbogos/default?martech=off
- After: https://main--cc--adobecom.hlx.page/drafts/rbogos/default?milolibs=mwpw-138031-lana-logs-tags--milo--rbogos&martech=off

**Milo:**
- Before: https://main--milo--robert-bogos.hlx.page/drafts/rbogos/default?martech=off
- After: https://lana-logs-tags--milo--robert-bogos.hlx.page/drafts/rbogos/default?martech=off

PR #2818, opened 2024-09-03
fix(MWPW-155964):Lana log error type changed to info.

sharath-kannan:
* Type of Lana log changed to info.

Resolves: [MWPW-155964](https://jira.corp.adobe.com/browse/MWPW-155964)

**Test URLs:**
- After: https://mwpw-155964--milosh--sharath-kannan.hlx.page/?martech=off


libs/martech/martech.js
```diff
@@ -118,14 +118,14 @@ export const getTargetPersonalization = async () => {
   const responseStart = Date.now();
   window.addEventListener(ALLOY_SEND_EVENT, () => {
     const responseTime = calculateResponseTime(responseStart);
-    window.lana.log(`target response time: ${responseTime}`, { tags: 'errorType=info,module=martech' });
+    window.lana.log(`target response time: ${responseTime}`, { tags: 'module-martech', errorType: 'i' });
```
npeltier, line 121: is `module-martech` tag existing? wouldn't `martech` be clear enough, yet smaller?

libs/martech/martech.js
```diff
@@ -118,14 +118,14 @@ export const getTargetPersonalization = async () => {
...
   }, { once: true });
 
   let manifests = [];
   let propositions = [];
   const response = await waitForEventOrTimeout(ALLOY_SEND_EVENT, timeout);
   if (response.error) {
-    window.lana.log('target response time: ad blocker', { tags: 'errorType=info,module=martech' });
+    window.lana.log('target response time: ad blocker', { tags: 'errorType-info,module-martech' });
```
npeltier, line 128: isn't this not consistent with the above? shouldn't it be 
`{ tags: 'module-martech', errorType: 'i' }` too?

libs/martech/martech.js
mokimo, line 121: Where do you have this errorType from? The type is correct as far as I can see according to the initial PR where we introduced the feature https://github.com/adobecom/milo/pull/1463 
// cc @robert-bogos

libs/martech/martech.js
sharath-kannan: It should be, if the log was informational but since this is clearly inside the error block i have left the errorType to default which is "error".

libs/martech/martech.js
sharath-kannan: I have used the Lana documentation to set the errorType option to 'i'.
Fyr: https://github.com/adobecom/milo/blob/main/libs/utils/lana.md

sharath-kannan: yes, I will make this change.

mokimo: Theat file is little older than the linked PR and wasn't updated it seems. Mind changing that?

libs/martech/martech.js
npeltier: mm to me "error" is a problem on our side, the fact the browser uses adblocker does not really fall into that category, but i guess i'm fine :)

libs/martech/martech.js
sharath-kannan: How do i set the errorType of a log to 'info' according to the initial PR?

mokimo: There's a lot of examples in the PR 🤔 https://github.com/adobecom/milo/pull/1463/files

sharath-kannan: My understanding was setting the errorType within tags as "info" would just help to filter logs in Splunk and doesn't change the log type to "info", correct me if i'm wrong.

thedoc31 (changes requested):
~key-value pairs are not supported in the tags. Please refer to https://wiki.corp.adobe.com/display/WCMSOps/LANA+-+Log+Always+Never+Assume#LANALogAlwaysNeverAssume-Parameters for the correct parameters to pass to LANA.~

I was wrong, see comment below.

thedoc31 (approved):
This format should work, per the documentation. Please validate in Splunk that these messages now come in as `info` type.

libs/martech/martech.js
robert-bogos: Yes, the errorType purpose was to help filter logs in Splunk

robert-bogos: After more discussions on the matter, we decided that the source of truth will be this [lana.md](https://github.com/adobecom/milo/blob/main/libs/utils/lana.md). [My initial PR](https://github.com/adobecom/milo/pull/1463/files) for this is not relevant anymore.

robert-bogos (commented):
Hi @sharath-kannan ! I think the Jira ticket aimed to correct all the LANA logs not being used properly, not just the ones from martech.

sharath-kannan:
> Hi @sharath-kannan ! I think the Jira ticket aimed to correct all the LANA logs that are not being used properly, not just the ones from martech.

Okay,I will correct all the lana logs that are used incorrectly in milo and update my PR. Hope using "module-<moduleName>" will be fine for now?

thedoc31:
> > Hi @sharath-kannan ! I think the Jira ticket aimed to correct all the LANA logs that are not being used properly, not just the ones from martech.
> 
> Okay,I will correct all the lana logs that are used incorrectly in milo and update my PR. Hope using "module-" will be fine for now?

I would recommend making the tags as concise as possible, but a dash character is fine within a comma-delimited list. The restricted characters are documented on the wiki [1] in the note above the table.

[1]
https://wiki.corp.adobe.com/display/WCMSOps/LANA+-+Log+Always+Never+Assume#LANALogAlwaysNeverAssume-Parameters

skumar09:
@sharath-kannan : can you please update your branch with latest code base( I guess it doesn't have video block changes, as corresponding nala tests are failing)

hparra (approved):
It's not clear why we're doing this. Did the LANA interface change? At the time this was correct.

video-metadata is a deprecated component so there is no issue here.

libs/blocks/faas/utils.js
```diff
@@ -248,7 +248,7 @@ const beforeSubmitCallback = () => {
       }),
     })
       .catch((error) => {
-        window.lana.log(`AA Sandbox Error: ${error.reason || error.error || error.message || error}`, { tags: 'errorType=info,module=faas' });
+        window.lana.log(`AA Sandbox Error: ${error.reason || error.error || error.message || error}`, { tags: 'module=faas', errorType: 'i' });
```
thedoc31, line 251: `tags` here should just be `faas`, not `module=faas`

thedoc31 (changes requested):

thedoc31:
> It's not clear why we're doing this. Did the LANA interface change? At the time this was correct.
> 
> video-metadata is a deprecated component so there is no issue here.

Unfortunately, the information provided at the time was incorrect. Key-value pairs have never been supported in `tags:` elements. The documentation has been updated to make this very clear.

thedoc31 (approved):

spadmasa:
Validated the change done in splunk logs
https://main--cc--adobecom.hlx.page/products/photoshop?milolibs=mwpw-155964--milosh--sharath-kannan&lanadebug
image
image

PR #2822, opened 2024-09-03
MWPW-156365 - Text Block tiger team updates

elan-tbx:
Requirements as listed in the Jira ticket (linked below), all currently available:
* add (large) Title element (18px, bold, lowercase) -- will have expanded abilities with the addition of Spectrum 2
* add two buttons and a text link horizontally stacked
* need more author awareness to have two outlined buttons stacked horizontally -- is actually already possible if the CTAs don't have line-breaks in between them
* also authorable to have the buttons be solid black/white or outlines black/white -- now available everywhere since the release of #2811 🥳 
* show/hide (multi-viewport) options
* add round corners of media containers, 16px
* handles edge cases of lockups and multiple icons used in icon areas
* add accent bar to build -- available as a variant to the block (`accent-bar`). After discussion with Consonant, it has been decided that adding this element to a "block" will be accomplished via "layouts" within the Sidekick library (e.g. a section containing `text (center, s-detail, etc...)` + `text (accent-bar)`).
  * Consequently, there should be a fast-follow task associated with this ticket that removes the extra code in the Notification block which accomplishes this on a per-block basis. Since that specific block is brand new and not being used yet, we shouldn't run the risk of regression errors by removing this code, thanks to the diligent work of our documentation writers 🫡 stay tuned for that part.

Resolves: [MWPW-156365](https://jira.corp.adobe.com/browse/MWPW-156365)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ebartholomew/fragments/basic-blocks-text-pr?martech=off
- After: https://ebartholomew-mwpw-156365-basic-blocks--milo--adobecom.hlx.page/drafts/ebartholomew/fragments/basic-blocks-text-pr?martech=off


libs/blocks/text/text.css
```diff
@@ -12,9 +14,7 @@
 
 .text-block [class^="body-"]:only-child { margin: 0; }
 
-.text-block [class^="heading"] { margin: 0 0 var(--spacing-xs) 0; }
-
-.text-block [class^="detail"] { margin: 0 0 var(--spacing-xxs) 0; }
+.text-block :is([class^="heading"], [class^="detail"]) { margin: 0 0 var(--spacing-xs) 0; }
```
ryanmparrish, line 17: This is going to change all the `-detail` default margin. Is this right?

elan-tbx: @ryanmparrish Yep, according to Consonant that is what they desire.

libs/blocks/text/text.css
```diff
@@ -73,6 +73,25 @@
   margin: var(--spacing-m) 0;
 }
 
+.text-block .title-l {
```
ryanmparrish, line 76: Where is this class name coming from? If possible we should try to align w/ the existing `-body` or `-detail` existing typography classes and not introduce more associations.

libs/blocks/text/text.css
```diff
@@ -73,6 +73,25 @@
...
+  font-weight: bold;
+  text-transform: none;
+  margin-bottom: var(--spacing-xs);
+}
+
+.text-block :is(.image, .lockup-area) {
+  margin-block-end: var(--spacing-m);
+}
+
+.text-block .image em {
```
ryanmparrish, line 88: Is this supporting a specific authoring pattern? It seems kind of unlikely to expect an image in an <em> tag?

libs/blocks/text/text.css
elan-tbx: @ryanmparrish yeah I copied this over from @meganthecoder's work on the Aside block -- Apparently we will have an update in the future that implements Spectrum 2 and renders this obsolete. For now, this is a band-aid and will require a due-diligence in the future :)

libs/blocks/text/text.css
```diff
@@ -83,12 +102,63 @@
...
+.text-block .icon-area {
+  display: flex;
+  column-gap: var(--spacing-xs);
+}
+
+.text-block p.icon-area { /* NOT <a/> tags with icons in them */
+  margin-block-end: var(--spacing-m);
+}
+
+.text-block.xxs-icon .icon-area {
```
ryanmparrish, line 129: Shouldn't these just use the iconograpy.css styles and these wouldn't be needed at all?

libs/blocks/text/text.css
elan-tbx: @ryanmparrish Assuming you're referring to the `em` in this rule -- yes, this is actually how lockup text is being handled in other blocks currently. However, `.image` overrides that and expects an `<em>` tag to achieve an image caption as detailed in the [Figma designs](https://www.figma.com/design/eUM6feFsRokh4uxM3vZFL3/Basic-Blocks?node-id=450-12849&node-type=FRAME&t=w9DHQD809VEgLP67-0)

libs/blocks/text/text.css
elan-tbx: @ryanmparrish I totally agree, this was just me correcting a `style-lint` error. I'd be happy to remove it, but I believe that will require some extra regression testing.

libs/blocks/text/text.css
ryanmparrish: Where is the example for this? I don't see why we need this when it sounds like tech-debt

libs/blocks/text/text.css
elan-tbx: FWIW, the authoring pattern behind this (and those other lockup implementations) is just adding text after an image _without_ a line break. In this block, it'll only apply lockup styles if you add a `-lockup` variant to the block.

libs/blocks/text/text.css
ryanmparrish: Yes,  since you're introducing `iconography.css` in this PR. 
We should definitely trim the fat since this is redundant now.

libs/blocks/text/text.css
elan-tbx: I'll let @meganthecoder speak to the why of this, but dfaf39fed9f10cd14986290cd1f96de75ea495a1 is where this got originally implemented on the Aside block. If we can reasonably rely on Spectrum 2 now, I'll happily remove this, but I have not heard otherwise.

libs/blocks/text/text.css
ryanmparrish: Hmm, This sounds like a use for the [[figure](https://milo.adobe.com/docs/library/blocks/figure)] block on top of a text block rather? Was this a requirement for the text block or a layout? I'd like to be careful bloating the text block w/ too many features if possible.

libs/blocks/text/text.css
elan-tbx: okay, I just ask because `iconography.css` does not include any rules related to `.icon-area` whatsoever. I do not import that stylesheet unless a block specifically uses a `lockup` variant, because otherwise we would be adding an HTTP request to every single use of the Text block.

Are you _sure_ we should be removing this?

libs/blocks/text/text.css
ryanmparrish: I mean in your [after link](https://ebartholomew-mwpw-156365-basic-blocks--milo--adobecom.hlx.page/drafts/ebartholomew/fragments/basic-blocks-text?martech=off). I don't see an example of this need/usage?

elan-tbx: lol, sorry about that, I handed that page off to authors for testing and it looks like they removed some things. I'll update the description with stable links 👍

libs/blocks/text/text.css
ryanmparrish: Wouldn't we then just change that selector to `.lockup-area` in the decorator and test it still works as expected. Its essentially just doubling up the styles w/ diff selectors doing the same thing.

elan-tbx: I agree moving forward that is probably the right idea, but like I said, this was a pre-existing style that I just moved to fix a style-lint error. Taking this rule out altogether runs the risk of messing up old usages of this block.

libs/blocks/text/text.css
elan-tbx: Very good point! I'll bring that back to the stakeholders and see what they say 👍

libs/blocks/text/text.css
meganthecoder: @ryanmparrish Title is a new stylistic element (like body and detail) that's going to be added in Spectrum 2. However, CC needs it now and can't wait until Spectrum 2 is officially out. The `title-l` class is our stopgap measure. As you can see, the large title is bold, 18px, and normal case, so none of the body or detail styles would work to match the requirements.

Screenshot 2024-09-03 at 3 50 53 PM

From: https://main--milo--adobecom.hlx.page/drafts/methomas/consonant-tiger/aside-standard-large?martech=off#authorable-text-sizes

libs/blocks/text/text.css
ryanmparrish: Fair, there also might be a day they are different. Probs safer to just leave em be for now.

libs/blocks/text/text.js
```diff
@@ -36,8 +35,14 @@ function decorateMultiViewport(el) {
...
-  const headings = el.querySelectorAll('h1, h2, h3, h4, h5, h6');
+function decorateBlockIconArea(content, el) {
+  const headings = content.querySelectorAll('h1, h2, h3, h4, h5, h6');
+  const first = content.children[0];
+  const firstImg = first?.querySelector('img');
+  if (firstImg) {
+    const areaClass = el.className.match(/-(lockup|icon)/);
+    first.classList.add(areaClass ? `${areaClass[1]}-area` : 'image');
+  }
   if (!headings) return;
```
chrischrischris, line 46: Move this to right after `headings` is defined.

libs/blocks/text/text.css
elan-tbx: Worked this out over Slack with everyone involved, but leaving this here to document it:

Apparently we are looking at deprecating the Figure block, because it does not respect Milo defaults (alignment, dark mode support, etc). That is probably a discussion for another time, but for now (since it lacks the aforementioned features), the general direction for this work is that we support this in the Text Block so that it is a more globally supported option.

meganthecoder (approved):

libs/blocks/text/text.css
```diff
@@ -73,6 +73,25 @@
   margin: var(--spacing-m) 0;
 }
 
+.text-block .title-l {
+  font-size: var(--type-body-m-size);
+  line-height: var(--type-body-m-lh);
+  font-weight: bold;
```
narcis-radu, line 79: shouldn't we use a numeric value here?

libs/blocks/text/text.css
```diff
@@ -158,6 +223,18 @@
...
 
+.text-block .image picture {
+  display: flex;
+}
+
+.text-block.center .image picture {
+  justify-content: center;
+}
+
+.text-block.right .image picture {
```
narcis-radu, line 234: I was looking at this declaration and noticed the `.text-block.right` class doesn't position the text properly on RTL locales.

**RTL**
Screenshot 2024-09-05 at 14 24 49

**RTL with `right` class applied to text-block**
Screenshot 2024-09-05 at 14 24 59

chrischrischris (approved):

ryanmparrish (approved):

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-156365

PR #2834, opened 2024-09-05
 [MWPW-157347] Environment relative URLs

robert-bogos:
## Description
This PR allows Milo consumers to set a domain mapping that converts URLs in the stage environment, with the aim of facilitating E2E testing.

## Guidelines for configuring `stageDomainsMap`:
**1. Define the stage domains as keys**
- Each key in the stageDomainsMap represents **a stage domain or a pattern** for which the rules will apply.
From the example below, these would be `'business.stage.adobe.com'` and `'--bacom--adobecom.hlx.page'`.
**2. Map production domains to corresponding stage domains:**
- For each stage domain key, provide a nested object that maps production domains to their respective stage domains.
- If the production domain should dynamically match the current environment (e.g., branch-based URLs), use the keyword `'origin'` to indicate that the replacement should be calculated dynamically based on the current URL context.

**Example:**
```
const stageDomainsMap = {
  'business.stage.adobe.com': {
    'www.adobe.com': 'www.stage.adobe.com',
    'business.adobe.com': 'business.stage.adobe.com',
  },
  '--bacom--adobecom.hlx.page': { 'business.adobe.com': 'origin' }
};
```

## Related Issue
Resolves: [MWPW-157347](https://jira.corp.adobe.com/browse/MWPW-157347)

## Screenshots:
**Before**

Screenshot 2024-09-05 at 10 17 11

**After**

Screenshot 2024-09-05 at 10 18 30

## Test URLs
- Before: https://main--milo--robert-bogos.hlx.page/drafts/rbogos/env-relative-urls?martech=off
- After: https://mwpw-157347-env-relative-urls--milo--robert-bogos.hlx.page/drafts/rbogos/env-relative-urls?martech=off


npeltier:
nit: please be careful using the "staging" word, as it can resonate as akamai staging (one machine holding akamai changes before it's propagated to the whole CDN), i would reserved `staging` keyword for this to avoid confusions in related discussions, and rather use `stage`

libs/scripts/scripts.js
```diff
@@ -22,11 +22,20 @@ import locales from '../utils/locales.js';
...
-  'news.adobe.com': 'news.stage.adobe.com',
+  'www.stage.adobe.com': { 'www.adobe.com': 'www.stage.adobe.com' },
+  'business.stage.adobe.com': {
+    'www.adobe.com': 'www.stage.adobe.com',
+    'business.adobe.com': 'business.stage.adobe.com',
+    'blog.adobe.com': 'blog.stage.adobe.com',
+    'helpx.adobe.com': 'helpx.stage.adobe.com',
+    'news.adobe.com': 'news.stage.adobe.com',
+  },
+  '--milo--robert-bogos.hlx.page': {
```
overmyheadandbody, line 33: This seems a bit self-centered 😛 I'd guess this shouldn't be in the final code?

overmyheadandbody (commented):
I'll have a deeper look later on

libs/scripts/scripts.js
robert-bogos: I will remove it 😂, is just to make QA's work easier

libs/scripts/scripts.js
```diff
@@ -22,11 +22,20 @@ import locales from '../utils/locales.js';
 const prodDomains = ['milo.adobe.com'];
 
 const stageDomainsMap = {
-  'www.adobe.com': 'www.stage.adobe.com',
-  'blog.adobe.com': 'blog.stage.adobe.com',
-  'business.adobe.com': 'business.stage.adobe.com',
-  'helpx.adobe.com': 'helpx.stage.adobe.com',
-  'news.adobe.com': 'news.stage.adobe.com',
+  'www.stage.adobe.com': { 'www.adobe.com': 'www.stage.adobe.com' },
+  'business.stage.adobe.com': {
```
npeltier, line 26: minor: can't we make just an array of prod domain to look for and then default replace function is .adobe.com -> .stage.adobe.com ? (and generate that mapping automatically once)

npeltier (dismissed):

libs/scripts/scripts.js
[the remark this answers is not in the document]
robert-bogos: Is just to speed up testing for QE's. I will remove it before going to stage

libs/scripts/scripts.js
robert-bogos: I don't think this will work because we don't have an universal replacement for .adobe.com. It can be .stage.adobe.com or .hlx.page or .hlx.live.
Also Milo should be consumer agnostic so I would avoid having consumer specific rules, such as a replace function for .adobe.com -> .stage.adobe.com

libs/scripts/scripts.js
```diff
@@ -22,11 +22,20 @@ import locales from '../utils/locales.js';
...
+    'blog.adobe.com': 'blog.stage.adobe.com',
+    'helpx.adobe.com': 'helpx.stage.adobe.com',
+    'news.adobe.com': 'news.stage.adobe.com',
+  },
+  '--milo--robert-bogos.hlx.page': {
+    'milo.adobe.com': 'origin',
+    'business.adobe.com': 'main--bacom--adobecom.hlx.page',
+  },
+  '--bacom--adobecom.hlx.live': { 'business.adobe.com': 'origin' },
+  '.business-graybox.adobe.com': { 'business.adobe.com': 'origin' },
```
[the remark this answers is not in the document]
robert-bogos: Maybe this could be a future enhancement? after a due diligence to assess the potential performance impact?

overmyheadandbody (commented):
We're getting into some deep nesting territory with this addition for something that would only be used on non-prod. Could/should we separate this into another file and import it on stage? I'd assume this would slightly lower performance on non-prod pages, but might be worth checking out. If not, maybe we could separate this logic into a new method that would return early, should any condition not be met.

Not a show-stopper, though, more of a "cosmetic" concern.

libs/utils/utils.js
```diff
@@ -628,8 +628,22 @@ export function decorateLinks(el) {
     appendHtmlToLink(a);
     a.href = localizeLink(a.href);
     decorateSVG(a);
-    if (config.env?.name === 'stage' && config.stageDomainsMap?.[a.hostname]) {
-      a.href = a.href.replace(a.hostname, config.stageDomainsMap[a.hostname]);
+    if (config.env?.name === 'stage' && config.stageDomainsMap) {
```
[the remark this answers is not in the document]
robert-bogos: Yes, makes sense. Made the change 👍

libs/scripts/scripts.js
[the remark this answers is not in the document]
robert-bogos: I see an additional layer of complexity in using a query parameter for this. In E2E testing, the query parameter is lost when navigating to another page via a link. A solution would be to append the query parameter to all links, but this adds more complexity 🤔

robert-bogos:
> We're getting into some deep nesting territory with this addition for something that would only be used on non-prod. Could/should we separate this into another file and import it on stage? I'd assume this would slightly lower performance on non-prod pages, but might be worth checking out. If not, maybe we could separate this logic into a new method that would return early, should any condition not be met.
> 
> Not a show-stopper, though, more of a "cosmetic" concern.

I agree that an early return would be better. Made the change 👍

libs/utils/utils.js
```diff
@@ -638,17 +638,36 @@ const decorateCopyLink = (a, evt) => {
...
+      const [, domainsMap] = matchedRules;
+      const matchedDomain = Object.keys(domainsMap)
+        .find((domain) => a.href.includes(domain));
+
+      if (matchedDomain) {
+        a.href = a.href.replace(a.hostname, domainsMap[matchedDomain] === 'origin'
+          ? hostname
+          : domainsMap[matchedDomain]);
+      }
+    }
```
overmyheadandbody, line 658: Could we get even greedier with early returns?
```suggestion
    if (!matchedRules) return;
    const [, domainsMap] = matchedRules;
    const matchedDomain = Object.keys(domainsMap)
      .find((domain) => a.href.includes(domain));

    if (!matchedDomain) return;
    a.href = a.href.replace(a.hostname, domainsMap[matchedDomain] === 'origin'
      ? hostname
      : domainsMap[matchedDomain]);
```

libs/scripts/scripts.js
narcis-radu: @npeltier - we thought about this in the previous iteration, but realised it's not something that would work.

libs/utils/utils.js
```diff
@@ -638,17 +638,32 @@ const decorateCopyLink = (a, evt) => {
   });
 };
 
+function convertStageLinks(config, anchors) {
```
mokimo, line 641: If you export this, you can write proper unit tests for this. I'd even say it's worth passing the as origin for the unit tests, given that this is a complex feature in terms of use-cases it tackles.

We should test
- If each hlx.page|hlx.live|stage.adobe.com|adobe|consumer is transformed and also NOT transformed

npeltier (commented):
switching to comment, as in its state i would prefer as Rares commented it (better than me :)) that non prod content & processing are moved to separate file, with probably only left some test on current domain (which would be better than a query param too :) )

libs/utils/utils.js
```diff
@@ -638,17 +638,32 @@ const decorateCopyLink = (a, evt) => {
...
+      ? hostname
+      : domainsMap[matchedDomain]);
+  });
+}
+
 export function decorateLinks(el) {
   const config = getConfig();
   decorateImageLinks(el);
   const anchors = el.getElementsByTagName('a');
+  convertStageLinks(config, anchors);
```
overmyheadandbody, line 662: At one point I remember we intentionally moved everything inside the existing `forEach` on L663 to avoid re-iterating over the same list of `a` elements. Is anything preventing you from sticking to that patterns?

I'd also be curious of how `convertStageLinks` interacts with the `localizeLink` and whether the link localization is maintained in all use-cases. If I'm on stage, on a Bacom page in German and I want to go to the Photoshop one, would I be redirected to the German version OOTB? It might be that we need to call `localizeLink` first, but you should check that hunch.

robert-bogos: > At one point I remember we intentionally moved everything inside the existing forEach on L663 to avoid re-iterating over the same list of a elements. Is anything preventing you from sticking to that pattern?

robert-bogos: > At one point I remember we intentionally moved everything inside the existing forEach on L663 to avoid re-iterating over the same list of a elements. Is anything preventing you from sticking to that pattern?

IMO the choice between these is a trade-off.
The current implementation:
Advantage: better for production by avoiding redundant checks on each `a` element.
Disadvantage: less efficient on stage due to iterating over `a` elements twice.

The alternative approach, moving the logic inside the current `a`'s `reduce()`:
Advantage: more efficient on stage by doing just 1 iteration over `a` elements.
Disadvantage: slight overhead on production by performing checks for each anchor.

Anyway, in both cases, the performance impact is minimal at most, but I chose the current approach because it is better for production and separates concerns.

> I'd also be curious of how convertStageLinks interacts with the localizeLink and whether the link localization is maintained in all use-cases

Did some testing around this and it works OOTB. The only problem is that the `.html` extension persists and that works only for production. Anyway, this `.html` problem is out of the scope of this ticket.

libs/utils/utils.js
robert-bogos: Added the unit tests 👍

overmyheadandbody (approved):

mokimo (approved):

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-157347

PR #2837, opened 2024-09-05
MWPW-157210 STE Promo card not appearing

bozojovicic:
On https://www.stage.adobe.com/products/catalog.html call for 
https://main--cc--adobecom.hlx.page/cc-shared/fragments/promos/2024/global/test/catalog-promo-test/fragments/catalog-ste.plain.html  
fails with error 401.
On https://main--cc--adobecom.hlx.page/products/catalog this works since the hostname is identical.
So I changed the hostname in the card URL to load it from `merch-card-collection.js`
before `localizeLink()` is called.
I started with an idea to change the code of `localizeLink()` to do the same thing but this method is used in too many places, hence too risky.
Can be tested only on https://www.stage.adobe.com/products/catalog.html

Resolves: [MWPW-157210](https://jira.corp.adobe.com/browse/MWPW-157210)

**Test URLs:**
- After: https://mwpw157210catalogste--milo--bozojovicic.hlx.page/drafts/bozo/price-cta?martech=off


test/utils/helpers.test.js
```diff
@@ -0,0 +1,14 @@
+import { expect } from '@esm-bundle/chai';
+import { overrideUrlOrigin } from '../../libs/utils/helpers.js';
+
+describe('overrideUrlOrigin', () => {
+  it('Change origin to http://localhost:2000', async () => {
```
npeltier, line 5: nit: I would not use a non prod origin for that test

npeltier (approved):
Looks good to me beside that test origin.

mirafedas (approved):

Roycethan:
@bozojovicic Do you have anything pending here to work on open comments? otherwise have covered regressions and mark it ready for stage

test/utils/helpers.test.js
bozojovicic: Checking utils is the first thing I did and I could not find it. Regarding origin, I tried to modify location.href to prod origin, didn't go very well and then I realized that localhost:2000 in many tests, so I did the same.

bozojovicic:
@Roycethan I answered all comments. Nothing pending here. It can be merged to stage.

PR #2838, opened 2024-09-05
Adding support for getting env config through client env

bandana147:
* Make IMS environment configurable from config in standalone navigation

Resolves: [MWPW-157474](https://jira.corp.adobe.com/browse/MWPW-157474)

**Test URLs:**
- After: https://MWPW-157474--milo--adobecom.hlx.page/?martech=off

**QA**
- https://adobecom.github.io/nav-consumer/navigation.html?navbranch=mwpw-157474

### DIFF
--- a/libs/navigation/navigation.js
+++ b/libs/navigation/navigation.js
@@ -52,6 +52,7 @@ export default async function loadBlock(configs, customLib) {
 
   const paramConfigs = getParamsConfigs(configs, miloLibs);
   const clientConfig = {
+    clientEnv: env,
     origin: `https://main--federal--adobecom.hlx.${env === 'prod' ? 'live' : 'page'}`,
     miloLibs: `${miloLibs}/libs`,
     pathname: `/${locale}`,
--- a/libs/navigation/navigation.js
+++ b/libs/navigation/navigation.js
@@ -37,6 +37,7 @@ export default async function loadBlock(configs, customLib) {
   const {
     header, footer, authoringPath, env = 'prod', locale = '', theme,
   } = configs || {};
+  window.miloClientEnv = env;
   const branch = new URLSearchParams(window.location.search).get('navbranch');
   const miloLibs = branch ? `https://${branch}--milo--adobecom.hlx.page` : customLib || envMap[env];
   if (!header && !footer) {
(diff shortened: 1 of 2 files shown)


narcis-radu:
@bandana147 - what's the use-case for this enhancement?

bandana147:
> @bandana147 - what's the use-case for this enhancement?

This applies to standalone gnav clients. If their page is not listed in this condition:

`if (host.includes(`${SLD}.page`)
    || host.includes(`${SLD}.live`)
    || host.includes('stage.adobe')
    || host.includes('corp.adobe')
    || host.includes('graybox.adobe')) {
    return { ...ENVS.stage, consumer: conf.stage };
}`
it always defaults to production.
We did suggest using the env URL parameter, but it seems to get lost sometimes during redirection. 
They’ve requested the ability to pass the env parameter while configuring the navigation, to ensure they can consistently use the correct environment for their stage setup.

libs/utils/utils.js
```diff
@@ -150,8 +150,11 @@ const PROMO_PARAM = 'promo';
 function getEnv(conf) {
   const { host } = window.location;
   const query = PAGE_URL.searchParams.get('env');
-
   if (query) return { ...ENVS[query], consumer: conf[query] };
+
+  const { clientEnv } = conf;
```
mokimo, line 155: I wonder if we should more generically look for `window.miloClientEnv` ? 
That would work for blocks exported outside of the gnav as well

mokimo (approved):
new config, new unit test!

libs/utils/utils.js
bandana147: Yes, I  will update it to `window.miloClientEnv`

sonawanesnehal3 (approved):

sharmrj (approved):

salonijain3 (approved):

prativas22:
The PR is validated, and the test results are updated in the ticket's subtask.
https://jira.corp.adobe.com/browse/MWPW-157474

https://adobecom.github.io/nav-consumer/navigation.html?navbranch=mwpw-157474&authoringpath=/federal/dev/gnavtest/ccgnav
image
image

libs/navigation/navigation.js
```diff
@@ -37,6 +37,7 @@ export default async function loadBlock(configs, customLib) {
   const {
     header, footer, authoringPath, env = 'prod', locale = '', theme,
   } = configs || {};
+  window.miloClientEnv = env;
```
narcis-radu, line 40: we should not export a global variable with environment details

libs/utils/utils.js
bandana147: @narcis-radu Initially it was added as part of config. Should we keep it this way?
cc: @salonijain3

PR #2846, opened 2024-09-06
MWPW-154980 - Milo advanced page publishing

Sartxi:
As a web producer, I want to limit who can publish certain files so that files are not published prematurely before planned launches.

New Feature Includes:
- Ability to fetch an optional custom configuration file from path `.milo/publish-permissions-config` to apply advanced publishing rules on a per page basis.
- New functionality will disable the publish button in AEM sidekick if path matches an item (url pathname) in the configuration file and can also reference a group with optional `allow/deny` user list which then compares the list to the email of profile that is currently logged into sidekick.
- Bulk Publish v2 tool will also check paths in the job list for item configuration and refuse to publish urls that have been configured unless of course the currently logged in user has permissions.

Resolves: [MWPW-154980](https://jira.corp.adobe.com/browse/MWPW-154980)

**Testing Details:**
1. _Sidekick Publishing:_ Go to the test url page -> open sidekick -> verify publish button is disabled.
2. _Bulk Tool Publishing:_ Go to [BP Tool](https://sartxi-mwpw-154980-publishing--milo--adobecom.hlx.page/tools/bulk-publish) -> Publish test page URL -> see error thrown with message.

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/sarchibeque/video-test?martech=off
- After: https://sartxi-mwpw-154980-publishing--milo--adobecom.hlx.page/drafts/sarchibeque/video-test?martech=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -1298,35 +1298,31 @@ export function loadLana(options = {}) {
 
 export const reloadPage = () => window.location.reload();
 
-export const userCanPublishPage = async (detail) => {
+export const userCanPublishPage = async (detail, isBulk = true) => {
   if (!detail) return false;
   const { live, profile, webPath } = detail;
-  let canPublish = live?.permissions?.includes('write');
-  let message = canPublish === undefined
-    ? 'Please sign into sidekick to publish this page'
-    : 'Publishing is currently disabled for this page';
-  if (canPublish) {
-    const config = await getCustomConfig('/.milo/publish-permissions-config.json');
-    if (config) {
-      const item = config.urls?.data?.find(({ url }) => (
-        url.endsWith('**') ? webPath.includes(url.slice(0, -2)) : url === webPath
-      ));
-      if (item) {
-        canPublish = false;
-        if (item.message) message = item.message;
-        const group = config[item.group];
-        if (group && profile?.email) {
-          let isDeny;
-          const user = group.data?.find(({ allow, deny }) => {
-            if (deny) {
-              /* c8 ignore next 3 */
-              isDeny = true;
-              return deny === profile.email;
-            }
-            return allow === profile.email;
-          });
-          canPublish = isDeny ? !user : !!user;
-        }
+  let canPublish = isBulk ? live?.permissions?.includes('write') : true;
+  let message = 'Publishing is currently disabled for this page';
+  const config = await getCustomConfig('/.milo/publish-permissions-config.json');
+  if (config) {
+    const item = config.urls?.data?.find(({ url }) => (
+      url.endsWith('**') ? webPath.includes(url.slice(0, -2)) : url === webPath
+    ));
+    if (item) {
+      canPublish = false;
+      if (item.message) message = item.message;
+      const group = config[item.group];
+      if (group && profile?.email) {
+        let isDeny;
+        const user = group.data?.find(({ allow, deny }) => {
+          if (deny) {
+            /* c8 ignore next 3 */
+            isDeny = true;
+            return deny === profile.email;
+          }
+          return allow === profile.email;
+        });
+        canPublish = isDeny ? !user : !!user;
       }
     }
   }
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -1317,6 +1317,7 @@ export const userCanPublishPage = async (detail) => {
           let isDeny;
           const user = group.data?.find(({ allow, deny }) => {
             if (deny) {
+              /* c8 ignore next 3 */
               isDeny = true;
               return deny === profile.email;
             }
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -1298,15 +1298,15 @@ export function loadLana(options = {}) {
 
 export const reloadPage = () => window.location.reload();
 
-export const miloUserCanPublish = async (page) => {
-  if (!page) return false;
+export const userCanPublishPage = async (detail) => {
+  if (!detail) return false;
   // sidekick status detail
-  const { live, profile, webPath } = page;
+  const { live, profile, webPath } = detail;
   let canPublish = live?.permissions?.includes('write');
   // check custom publish permissions config
   const config = await getCustomConfig('/.milo/publish-permissions-config.json');
   if (canPublish && config) {
-    // supporting only one wildcard(**)
+    // supporting one wildcard (**)
     const item = config.urls?.data?.find(({ url }) => (
       url.endsWith('**') ? webPath.includes(url.slice(0, -2)) : url === webPath
     ));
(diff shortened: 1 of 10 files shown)


tools/utils/utils.js
```diff
@@ -25,4 +26,19 @@ const getSheet = async (url) => {
...
-export { getImsToken, getSheet };
+const getCustomConfig = async (path) => {
+  /* c8 ignore next 3 */
+  if (CONFIGS[path] !== undefined) {
+    return CONFIGS[path];
+  }
+  let config = null;
+  const resp = await fetch(path);
+  if (resp.ok) {
+    const json = await resp.json();
```
yesil, line 37: alternatively: `config = await resp.json();` ?

Sartxi: nice! Thanks

rgclayton:
I almost think if the publish button is disabled the default hover fly out changes what it says.
Maybe something like "Publishing is disabled for this page"?

Currently the message reads like I can publish the page, which is a bit confusing.

image

libs/utils/utils.js
```diff
@@ -1295,3 +1297,34 @@ export function loadLana(options = {}) {
...
 
 export const reloadPage = () => window.location.reload();
+
+export const userCanPublishPage = async (detail, isBulk = true) => {
+  if (!detail) return false;
+  const { live, profile, webPath } = detail;
+  let canPublish = isBulk ? live?.permissions?.includes('write') : true;
+  let message = 'Publishing is currently disabled for this page';
+  const config = await getCustomConfig('/.milo/publish-permissions-config.json');
+  if (config) {
```
elan-tbx, line 1307: I feel like you could further leverage optional chaining here and at least remove one level of `if` statement... maybe something like this?

```js
const item = config?.urls?.data?.find(({ url }) => (
  url.endsWith('**') ? webPath.includes(url.slice(0, -2)) : url === webPath
));
```
if you put that directly below `config`'s definition, `item`'s definition should safely short-circuit with `config?.` in the event that `config` is falsy, in which case `item` would also be falsy, and you could then accurately use the following `if (item)` statement.

elan-tbx (approved):
Just left a 'food for thought' comment, awesome work otherwise 🤘

libs/utils/utils.js
Sartxi: Thanks @elan-tbx I will take that recommendation!

libs/utils/utils.js
```diff
@@ -1295,3 +1297,32 @@ export function loadLana(options = {}) {
 }
 
 export const reloadPage = () => window.location.reload();
+
+export const userCanPublishPage = async (detail, isBulk = true) => {
```
mokimo, line 1301: I don't think we should have code concerning authoring tools within the milo core utils.
The utils should be reserved for code that is meant for end users, it will end up on adobe.com and our end users are not involved in the publishing process, but currently would still load this code.

mokimo (changes requested):
Let's ensure that the bulk publishing tools don't touch the core milo utils - those both touch two different areas of concern.

libs/utils/utils.js
narcis-radu: this might be a good candidate for https://github.com/adobecom/milo/tree/stage/libs/tools/sharepoint or similar

libs/utils/utils.js
```diff
@@ -1295,3 +1297,32 @@ export function loadLana(options = {}) {
...
+          isDeny = true;
+          return deny === profile.email;
+        }
+        return allow === profile.email;
+      });
+      canPublish = isDeny ? !user : !!user;
+    }
+  }
+  return { canPublish, message };
+};
```
overmyheadandbody, line 1328: Any chance this can be moved to the tools utils or some other file? Ability to publish things should only be relevant for authoring purposes, not for our users.

test/blocks/bulk-publish-v2/bulk-publish-v2.test.js
```diff
@@ -114,6 +114,16 @@ describe('Bulk Publish Tool', () => {
     await mouseEvent(rootEl.querySelector('.fix-btn'));
   });
 
+  it('can trigger cannot publish config', async () => {
+    await clock.runAllAsync();
```
robert-bogos, line 118: I think the `await clock.runAllAsync()` might be redundant since we do not trigger any async operation before it.

libs/utils/utils.js
Sartxi: Okay sounds good, I will move it there

test/blocks/bulk-publish-v2/bulk-publish-v2.test.js
Sartxi: Because Bulk Publish testing involves rendering multiple web components, these functions are serving as a timing delay in order for the components to finish processing and rendering in the tests prior assertions. Originally I had setup a timeout function to serve this purpose, but they were updated to use this `runAllAsync` function a few months ago. I did notice some test failures due to that update and fixed them in this PR which is why you can see some of the assertions are moved around. 

Not sure if there is a better solution, but for now this method is allowing the UI to finish its work before running new assertions and subsequently failing because the components have not finished rendering.

libs/utils/utils.js
Sartxi: I moved this to the `tools/utils` dir.

test/blocks/bulk-publish-v2/bulk-publish-v2.test.js
```diff
@@ -132,6 +142,17 @@ describe('Bulk Publish Tool', () => {
     await mouseEvent(rootEl.querySelector('.switch.half'));
   });
 
+  it('can toggle job timing flyout', async () => {
+    await clock.runAllAsync();
```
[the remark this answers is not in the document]
Sartxi: I addressed the reasoning for these in the prior conversation 👍🏻

tools/utils/utils.js
```diff
@@ -25,4 +26,18 @@ const getSheet = async (url) => {
...
+  if (CONFIGS[path] !== undefined) {
+    return CONFIGS[path];
+  }
+  let config = null;
+  const resp = await fetch(path);
+  if (resp.ok) {
+    config = await resp.json();
+  }
+  CONFIGS[path] = config;
+  return CONFIGS[path];
```
ryanmparrish, line 40:
```suggestion
  if (CONFIGS[path] !== undefined) {
    return CONFIGS[path];
  }
  const resp = await fetch(path);
  const config = resp.ok ? await resp.json() : null;
  CONFIGS[path] = config;
  return config;
```
nit: slightly smaller way to write this fwiw

ryanmparrish (approved):
Looks good. Left one minor suggestion, take it or leave it. 
Is there any documentation around managing the `publish-permissions-config.json` file for author admins?

Sartxi:
> Looks good. Left one minor suggestion, take it or leave it. Is there any documentation around managing the `publish-permissions-config.json` file for author admins?

I am working on authoring documentation today!

rgclayton (approved):

NadiiaSokolova (approved):

PR #2853, opened 2024-09-09
MWPW-157024 - Editorial card authoring refinement for the 'open' variant.

ryanmparrish:
There was some feedback from GWP that the editorial-card (open) variant has some authoring issues w/ how the document is structured. At the core, this card type should not support a background row. 
This block has the ability to author w/ minimal content or 4+ rows that represent varying contexts. ex. 

| row count | content types |
| -------- | -------- |
| 3 rows | background, asset, copy |
| 4 rows | background, asset, copy, footer |

This update changes the authoring pattern when using the `open` variant as follows. 
This will allow us to offer a clean .open card example in our library w/ out the need or confusion for an empty background row. 

| row count | content types |
| -------- | -------- |
| 3 rows | 0:bg, 1:media, 2:copy |    
| 3 rows.open | 0:media, 1:copy, 2:card-footer | 
| 4 rows | 0:bg, 1:media, 2:copy, 3:card-footer | 
| 4 rows.open | 0:bg(removed), 1:media, 2:copy, 3:card-footer | 

Resolves: [MWPW-157024](https://jira.corp.adobe.com/browse/MWPW-157024)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/klee/test-editorial-cards?martech=off
- After: https://rparrish-edy-aux--milo--adobecom.hlx.page/drafts/klee/test-editorial-cards?martech=off

**Live URLs:**
- Library Before: https://main--milo--adobecom.hlx.page/docs/library/blocks/editorial-card?martech=off
- Library After: https://rparrish-edy-aux--milo--adobecom.hlx.page/docs/library/blocks/editorial-card?martech=off

- Kitchen Before: https://main--milo--adobecom.hlx.page/docs/library/kitchen-sink/editorial-card?martech=off
- Kitchen After: https://rparrish-edy-aux--milo--adobecom.hlx.page/docs/library/kitchen-sink/editorial-card?martech=off

### DIFF
--- a/libs/blocks/editorial-card/editorial-card.js
+++ b/libs/blocks/editorial-card/editorial-card.js
@@ -57,13 +57,14 @@ const decorateForeground = async (el, rows) => {
   });
 };
 
-const decorateBgRow = (el, background) => {
-  decorateBlockBg(el, background);
+const decorateBgRow = (el, background, remove) => {
   const rows = background.querySelectorAll(':scope > div');
   const bgRowsEmpty = [...rows].every((row) => row.innerHTML.trim() === '');
-  if (bgRowsEmpty) {
+  if (bgRowsEmpty || remove) {
     el.classList.add('no-bg');
     background.remove();
+  } else {
+    decorateBlockBg(el, background);
   }
 };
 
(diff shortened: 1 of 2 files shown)


libs/blocks/editorial-card/editorial-card.js
```diff
@@ -57,13 +57,14 @@ const decorateForeground = async (el, rows) => {
...
   const rows = background.querySelectorAll(':scope > div');
   const bgRowsEmpty = [...rows].every((row) => row.innerHTML.trim() === '');
-  if (bgRowsEmpty) {
+  if (bgRowsEmpty || remove) {
     el.classList.add('no-bg');
     background.remove();
+  } else {
+    decorateBlockBg(el, background);
   }
 };
```
JasonHowellSlavin, line 69:
```suggestion
  if (bgRowsEmpty || remove) {
    el.classList.add('no-bg');
    background.remove();
    return;
  }
  
  decorateBlockBg(el, background);
};
```
Remove else?

libs/blocks/editorial-card/editorial-card.js
```diff
@@ -80,38 +81,50 @@ function handleClickableCard(el) {
...
   el.classList.add('con-block');
-  if (el.className.includes('open')) {
-    el.classList.add('no-border', 'l-rounded-corners-image', 'static-links-copy');
-  }
-  if (el.className.includes('rounded-corners')) {
-    loadStyle(`${base}/styles/rounded-corners.css`);
-  }
+  const hasOpenClass = el.className.includes('open');
+  if (hasOpenClass) el.classList.add('no-border', 'l-rounded-corners-image', 'static-links-copy');
+  if (el.className.includes('rounded-corners')) loadStyle(`${base}/styles/rounded-corners.css`);
```
JasonHowellSlavin, line 86: I was doing a performance review recently and notices that the `rounded-corners.css` file tends to get loaded post LCP files like `brick.css` and `text.css`, yet it is a relatively small amount of CSS. Is there a reason this CSS cannot be included elsewhere and load in with other files? Even `styles.css`?

libs/blocks/editorial-card/editorial-card.js
```diff
@@ -80,38 +81,50 @@ function handleClickableCard(el) {
...
       case 1:
         // 1 row  (0:copy)
-        rows = head;
-        await decorateForeground(el, [rows]);
+        rows = [head];
         el.classList.add('no-bg', 'no-media');
+        await decorateForeground(el, rows);
         break;
       default:
     }
```
JasonHowellSlavin, line 130: Is there a better way to do this? This seems like it could get out of hand at some point, since it is already being changed to add `four-plus` instead of `three-plus`. Is there something the authors could do to signal the need for the programmatic variation? 

May want to remove the comments as well?

libs/blocks/editorial-card/editorial-card.js
```diff
@@ -57,13 +57,14 @@ const decorateForeground = async (el, rows) => {
   });
 };
 
-const decorateBgRow = (el, background) => {
-  decorateBlockBg(el, background);
+const decorateBgRow = (el, background, remove) => {
```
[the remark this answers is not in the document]
ryanmparrish: Agreed. Thanks

libs/blocks/editorial-card/editorial-card.js
ryanmparrish: Yes. This is updated.

libs/blocks/editorial-card/editorial-card.js
ryanmparrish: Im open to making this a more global style etc. and loading it accordingly. Mostly im just following suit here w/ other blocks that use this as an opt in style, if it's authored. I think if we want to move these styles all together it should be a separate task since it impacts many blocks.

libs/blocks/editorial-card/editorial-card.js
ryanmparrish: Technically we are already supporting 4+ `3+ rows (0:bg, 1:media, 2:copy, ...3:static, last:card-footer)` so it's too much more abstract. This is just giving authors a little more guard rails when using the open variant so the can't accidentally author it outside of that variants design. It is a little long winded, but lmk if you have any ides on handling this diff. Im not a huge fan of this in general, but it's what consonant wanted so. For the comments. I think these ones are actually helpful because of this abstraction.

JasonHowellSlavin: I think I see now. It looks like the class list is getting too long in some circumstances as well? I suppose I would try to solve with a class, but I guess that puts the onus on the authors. Not sure. Would like to see what some other folks think.

overmyheadandbody (approved):

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-157024

robert-bogos (approved):

PR #2854, opened 2024-09-09
MWPW-158059 [MEP] PHONE_SIZE limit is too large to exclude smaller tablets

vgoodric:
Note: this is listed a high priority because it is needed for MAX.

Detailed Description: experience columns using tablet are not catching smaller tablets
URL: https://main--cc--adobecom.hlx.page/products/photoshop-lightroom/features?target=off&mep=
Steps to Reproduce:
1. Simulate a device that is android and tablet
2. Go to the link above (choosing any samsung tab) should work

Expected Results: MEP button indicates Android & Tablet is chosen
Actual Results: MEP button indicates mobile-device is chosen

Resolves: [MWPW-158059](https://jira.corp.adobe.com/browse/MWPW-158059)

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/products/photoshop-lightroom/features?target=off&martech=off
- After: https://main--cc--adobecom.hlx.page/products/photoshop-lightroom/features?milolibs=meptablet&target=off&martech=off
- Psi-check: https://meptablet--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -7,7 +7,7 @@ import {
 import { getEntitlementMap } from './entitlements.js';
 
 /* c8 ignore start */
-const PHONE_SIZE = window.screen.width < 600 || window.screen.height < 600;
+const PHONE_SIZE = window.screen.width < 550 || window.screen.height < 550;
 export const PERSONALIZATION_TAGS = {
   all: () => true,
   chrome: () => navigator.userAgent.includes('Chrome') && !navigator.userAgent.includes('Edg'),
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -7,7 +7,7 @@ import {
 import { getEntitlementMap } from './entitlements.js';
 
 /* c8 ignore start */
-const PHONE_SIZE = window.screen.width < 768 || window.screen.height < 768;
+const PHONE_SIZE = window.screen.width < 600 || window.screen.height < 600;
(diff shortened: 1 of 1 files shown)


ivanvatadobe (approved):

AdobeLinhart (approved):

denlight (approved):

libs/features/personalization/personalization.js
```diff
@@ -7,7 +7,7 @@ import {
 import { getEntitlementMap } from './entitlements.js';
 
 /* c8 ignore start */
-const PHONE_SIZE = window.screen.width < 768 || window.screen.height < 768;
+const PHONE_SIZE = window.screen.width < 600 || window.screen.height < 600;
```
Axelcureno, line 10: Using `600` as breakpoint would exclude some smartphones when on landscape mode. Typically the smallest tablet has `768` of width (iPad mini), why not use `767` instead?

vgoodric: That's why the line checks width and height. Either being smaller avoids miscategorizing a device in landscape.  Essentially, I trying to check that the SMALLEST dimension is less than our checkpoint.

Someone pointed me to a site that lists the viewports for a bunch of devices and I've updated to go down to 550 instead.  Trying to catch the smaller tablets.

vgoodric: Sorry, I should mention.  This came up in reference to a specific use case: Samsung Galaxy tablets.  They are smaller than iPads.

PR #2855, opened 2024-09-09
MWPW-154059 - Milo templates library

rgclayton:
* Allows `template-metadata` block to be pasted as `metadata` block in template
* Fixes initial listing order
* Adds search ability to templates

Resolves: [MWPW-154059](https://jira.corp.adobe.com/browse/MWPW-154059)

**Test URLs:**
- After: https://template-library-updates--milo--adobecom.hlx.page/tools/library?martech=off

**BACOM Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/tools/library?ref=main&repo=bacom&owner=adobecom&host=business.adobe.com&project=BACOM&martech=off
- After: https://template-library-updates--milo--adobecom.hlx.page/tools/library?ref=main&repo=bacom&owner=adobecom&host=business.adobe.com&project=BACOM&martech=off

**Testing notes:**
With the updated features, the template named "demo" has search tags added (Rc-special demo test) via of a `library-metadata` block. This block allows authors to add custom search tags to the template. Searching with any of those tags will pull up the "Demo" template. However, when the template is copied from the library and pasted to a Word doc `library-metadata` block should not be part of the copied blocks. 

Testing this using the `main` branch will copy the `library-metadata` block to the Word doc.

To include a `metadata` block in the template, to be pasted into the Word doc, the template must include a `template-metadata` block with all the needed metadata fields. Once, pasted in the page that block converted to `metadata` and displays as a metadata block. 

If a `library-metadata` block with the key value pair (`searchTags` and search words) is added to a template, that template will be searchable by the name of the template along with the search words added to `library-metadata`.

Block searching should continue to work as it's always worked. 

---
image

### DIFF
--- a/libs/blocks/library-config/library-config.js
+++ b/libs/blocks/library-config/library-config.js
@@ -2,12 +2,12 @@ import { createTag } from '../../utils/utils.js';
 
 const LIBRARY_PATH = '/docs/library/library.json';
 
-async function loadBlocks(content, list, query, type) {
+async function loadBlocks({ content, list, query, type }) {
   const { default: blocks } = await import('./lists/blocks.js');
   blocks(content, list, query, type);
 }
 
-async function loadTemplates(content, list, query, type) {
+async function loadTemplates({ content, list, query, type }) {
   const { default: templates } = await import('./lists/templates.js');
   templates(content, list, query, type);
 }
@@ -32,7 +32,7 @@ async function loadPersonalization(content, list) {
   personalization(content, list);
 }
 
-function addSearch(content, list, type) {
+function addSearch({ content, list, type }) {
   const skLibrary = list.closest('.sk-library');
   const header = skLibrary.querySelector('.sk-library-header');
   let search = skLibrary.querySelector('.sk-library-search');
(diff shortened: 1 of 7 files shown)


rgclayton:
Will look into unit tests.

chrischrischris (approved):

libs/blocks/library-config/library-config.js
```diff
@@ -2,14 +2,14 @@ import { createTag } from '../../utils/utils.js';
 
 const LIBRARY_PATH = '/docs/library/library.json';
 
-async function loadBlocks(content, list, query) {
+async function loadBlocks(content, list, query, type) {
```
robert-bogos, line 5: With this many parameters I think is better to switch to passing an object. It helps avoid parameters related errors.
```suggestion
async function loadBlocks({ content, list, query, type }) {
```

libs/blocks/library-config/library-config.js
```diff
@@ -32,27 +32,47 @@ async function loadPersonalization(content, list) {
   personalization(content, list);
 }
 
-function addSearch(content, list) {
+function addSearch(content, list, type) {
```
robert-bogos, line 35: Same as above. I think it would be good to pass an object.
```suggestion
function addSearch({ content, list, type }) {
```

libs/blocks/library-config/library-config.js
rgclayton: Done

libs/blocks/library-config/library-config.js
rgclayton: Done

PR #2867, opened 2024-09-11
MWPW-157965: Upgrade Universal Nav to Version 1.3

nishantka:
Upgrades the UNav to 1.3

Adding eventListener for reloading Unav once Unav promise resolved as with v1.3 - the window.UniversalNav() returns a promise post which we can use methods like window.UniversalNav().reload() - [ref](https://wiki.corp.adobe.com/display/CCHome/Universal+Nav+Release+Info#:~:text=window.UniversalNav(config))
Also, Unav suggested to not rely on onReady method for triggering PEP modal instead we can wait for promise returned by `window.UniversalNav()` to be resolved.

Resolves: [MWPW-157965](https://jira.corp.adobe.com/browse/MWPW-157965)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.live/drafts/nishantkaush/test/unav?martech=off
- After: https://feds--milo--adobecom.hlx.live/drafts/nishantkaush/test/unav?martech=off

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -521,7 +521,7 @@ class Gnav {
       return 'linux';
     };
 
-    const unavVersion = new URLSearchParams(window.location.search).get('unavVersion') || '1.1';
+    const unavVersion = new URLSearchParams(window.location.search).get('unavVersion') || '1.3';
     await Promise.all([
       loadScript(`https://${environment}.adobeccstatic.com/unav/${unavVersion}/UniversalNav.js`),
       loadStyles(`https://${environment}.adobeccstatic.com/unav/${unavVersion}/UniversalNav.css`),
@@ -608,9 +608,6 @@ class Gnav {
       locale,
       imsClientId: window.adobeid?.client_id,
       theme: isDarkMode() ? 'dark' : 'light',
-      onReady: () => {
-        this.decorateAppPrompt({ getAnchorState: () => window.UniversalNav.getComponent?.('app-switcher') });
-      },
       analyticsContext: {
         consumer: {
           name: 'adobecom',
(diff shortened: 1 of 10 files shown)


libs/blocks/global-navigation/global-navigation.js
```diff
@@ -608,9 +608,6 @@ class Gnav {
       locale,
       imsClientId: window.adobeid?.client_id,
       theme: isDarkMode() ? 'dark' : 'light',
-      onReady: () => {
-        this.decorateAppPrompt({ getAnchorState: () => window.UniversalNav.getComponent?.('app-switcher') });
-      },
```
bandana147, line 613: Is `onReady` callback removed in new version?

nishantka: As mentioned in PR description - Unav suggested to not rely on onReady

sharmrj: and this is true for 1.2 as well?

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -627,11 +624,13 @@ class Gnav {
 
     // Exposing UNAV config for consumers
     CONFIG.universalNav.universalNavConfig = getConfiguration();
-    window.UniversalNav(CONFIG.universalNav.universalNavConfig);
-
-    isDesktop.addEventListener('change', () => {
-      window.UniversalNav.reload(CONFIG.universalNav.universalNavConfig);
-    });
+    window.UniversalNav(CONFIG.universalNav.universalNavConfig)
+      .then(() => {
```
sharmrj, line 628: Super minor, but I wonder if this won't be clearer with an `await` instead of a `.then`. That way we we get less nesting.

sharmrj (approved):
LGTM.

libs/blocks/global-navigation/global-navigation.js
nishantka: i don't think it affects 1.2 as the window.UNav() promise came up with 1.3

libs/blocks/global-navigation/global-navigation.js
nishantka: sure. I've updated this to await

narcis-radu (approved):

Deva309 (approved):

sigadamvenkata:
Verified the Unav 1.3 files and libs are loading in the given branch URL. Complete regression will be done across milo locales pages once PR merge to STAGE.

{2552412C-6228-4C5E-91A3-DB2C977F091F}

sigadamvenkata (approved):

PR #2873, opened 2024-09-12
Create an onReady and onError callback for standalone gnav

bandana147:
* Adds an onReady callback that will be triggered when Gnav loads
* Adds an onError callback that will be triggered if any error occurs inside gnav

Resolves: [MWPW-158168](https://jira.corp.adobe.com/browse/MWPW-158168)

**Test URLs:**
- After: https://gnav-event--milo--adobecom.hlx.page/?martech=off

Qa: https://adobecom.github.io/nav-consumer/navigation.html?authoringpath=/federal/dev&env=stage&navbranch=gnav-event


libs/blocks/global-navigation/global-navigation.js
```diff
@@ -317,7 +317,9 @@ class Gnav {
       await yieldToMain();
       await task();
     }
-
+    // Dispatching an event when gnav loads and is visible
+    window.dispatchEvent(new CustomEvent('feds:nav.ready'));
```
mokimo, line 321:
```suggestion
    window.dispatchEvent(new CustomEvent('feds:nav:ready'));
```

mokimo: On another note, since it's a milo block, should we dispatch `milo:globalnav:ready` ? That follows the other eventlisteners such as `milo:modal:closed`

mokimo:
As a more general question, why is that needed and what is the use-case? The ticket and description don't mention anything

bandana147:
> As a more general question, why is that needed and what is the use-case? The ticket and description don't mention anything

This is currently required for the Adobe Home team, but we thought it would be a good use case to support as a standalone component. They want to display a loader state until the Gnav fully loads and be able to handle/track any errors that occur within their app.

libs/navigation/navigation.js
```diff
@@ -35,12 +35,20 @@ function getParamsConfigs(configs) {
...
   const {
-    header, footer, authoringPath, env = 'prod', locale = '', theme,
+    header,
+    footer,
+    authoringPath,
+    env = 'prod',
+    locale = '',
+    theme,
+    onReady,
+    onError,
```
sharmrj, line 45: Since this is an async function couldn't the consumers just do
```javascript
 try {
  // do stuff
  await loadBlock(...);
  // do more stuff
} catch (e) {
  // handle error
}
```
Or they could even do
```javascript
loadBlock(...).then(onReady).catch(onError)
```
of course you'd have to `await` the `initBlock` function inside bootstrapper.js for this to work correctly (https://github.com/adobecom/milo/blob/af281740d8e910d0e677e3fb556f11c7095aeb50/libs/navigation/bootstrapper.js#L31).

This would allow us to avoid the rigmarole of callbacks and events that add more complexity to our model of how the global-navigation works. We'd just have to throw irrecoverable errors for the consumer to deal with (you may have to modify `loadPostLCP` with a `try...catch` to make milo correctly 'consume' global-navigation).
I realize I might be missing some nuance related to what errors should be propagated vs which errors should be handled internally, but my understanding is that the global-navigation already does this adequately. 
Ultimately I think it better to deal with this problem by propagating exceptions and dealing with exceptions on the consuming end rather than muck around with events and callbacks. Some time back we refused to allow Partner Portals to emit events to signify when specific parts of the global-navigation were loaded #2583 , and I think the same concerns apply here.

bandana147: If we implement it this way, the consuming client will receive the onError callback only when the gnav content is not found. 
All other errors, such as IMS issues or errors loading fragments, would be logged internally. To achieve this we could add a console.error line [here](https://github.com/adobecom/milo/blob/6fa95584902286108cfea82572288559b3517bb0/libs/blocks/global-navigation/utilities/utilities.js#L44)
Thoughts? @narcis-radu @mokimo @salonijain3

sharmrj (approved):

bandana147:
@narcis-radu @mokimo Could you please review the updated changes?

Deva309 (approved):

nishantka (approved):

prativas22:
The following URL on the gnav-event nav branch has been validated; the details are added to the ticket for reference.
https://jira.corp.adobe.com/browse/MWPW-158897
onReady call check:
https://adobecom.github.io/nav-consumer/navigation.html?authoringpath=/federal/dev/gnavtest/ccgnav/photographyshortform&env=stage&navbranch=gnav-event
{550EFA2D-C1DF-4C0A-958B-0E1718FFF4B0}
onError call check :
When an invalid client ID is entered, gnav loads successfully and unav does not load.
https://adobecom.github.io/nav-consumer/navigation.html?env=stage&theme=dark&navbranch=gnav-event&authoringpath=/federal/dev/gnavtest/ccgnav/photographyshortform&unav=help,appswitcher,profile,notifications&imsclientid=adobedotcom001
image

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -1025,27 +1025,32 @@ const getSource = async () => {
...
+  const [url, hash = ''] = sourceUrl.split('#');
+  if (hash === '_noActiveItem') {
+    setDisableAEDState();
+  }
+  const content = await fetchAndProcessPlainHtml({ url });
+  if (!content) {
+    const error = new Error('Could not create global navigation. Content not found!');
+    error.tags = 'errorType=error,module=gnav';
+    error.url = url;
+    window.lana.log(`${error.message} | gnav-source: ${error.url} | href: ${window.location.href}`, {
```
narcis-radu, line 1039: Any reason for not using `lanaLog` ?

libs/blocks/global-navigation/utilities/utilities.js
```diff
@@ -42,6 +42,7 @@ export const lanaLog = ({ message, e = '', tags = 'errorType=default' }) => {
     sampleRate: 1,
     tags,
   });
+  console.error(message);
```
narcis-radu, line 45: Why would we show a console error ?

bandana147: Now that we have non-Milo consumers as well, it seems appropriate to display the error in the console

libs/blocks/global-navigation/global-navigation.js
bandana147: This was an oversight, fixing it now. Thanks for pointing it out.

prativas22:
The following URL on the gnav-event nav branch has been validated after the push; the details
onReady call check:
https://adobecom.github.io/nav-consumer/navigation.html?authoringpath=/federal/dev/gnavtest/ccgnav/photographyshortform&env=stage&navbranch=gnav-event
{2946F91F-1A2C-49E1-8D43-2CB1DFBDE123}

https://adobecom.github.io/nav-consumer/navigation.html?env=stage&theme=dark&navbranch=gnav-event&authoringpath=/federal/dev&imsclientid=001
{5107841C-2DEC-483A-B5D7-E7CAFDD634C2}

https://adobecom.github.io/nav-consumer/navigation.html?authoringpath=/federal/dev/gnavtest/ccgnav/photographyshortform123&env=stage&navbranch=gnav-event&imsclientid=adobedotcom001
{9EF67D36-83C7-4590-881B-B4D42593EB93}

PR #1282, opened 2023-09-18
[9/20] MWPW-136568: Table - Mobile issue with new milo events

seanchoi-dev:
Fixing bug for table mobile logic doesn't work at all with new milo eventListener.
This issue is easily happening specially when there are more than one table in a page.

Resolves: [MWPW-136568](https://jira.corp.adobe.com/browse/MWPW-136568)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/docs/library/blocks/table?martech=off
- After: https://mwpw-136568--milo--adobecom.hlx.page/docs/library/blocks/table?martech=off

**Issue page on production**
- Before: https://main--bacom--adobecom.hlx.page/products/marketo/pricing
- After: https://main--bacom--adobecom.hlx.page/products/marketo/pricing?milolibs=mwpw-136568

*note: this fix is needed for CC release on 9/20 according to Ruchika Sinha.

PR #2881, opened 2024-09-16
MWPW-158182 add callback of URL for google yolo 

vhargrave:
Resolves: [MWPW-158182](https://jira.corp.adobe.com/browse/MWPW-158182)

**Test URLs:**
- After: https://vhargrave-MWPW-142878-events--milo--adobecom.hlx.page/?martech=off

This branch was originally built to listen to the two events to load both additional analytics and then get the primary cta url for google yolo which we redirect to.
Now after discussion with Rareș & Okan we've decided to just add a callback which resolves the url when necessary for the google yolo login. 
- https://github.com/adobecom/express-milo/pull/56/files
- https://main--express-milo--adobecom.hlx.page/express/colors/blue?milolibs=vhargrave-MWPW-142878-events

### DIFF
--- a/libs/features/google-login.js
+++ b/libs/features/google-login.js
@@ -9,7 +9,7 @@ const onToken = async (getMetadata, data, getConfig) => {
   try {
     destination = typeof config.googleLoginURLCallback === 'function' ? new URL(await config.googleLoginURLCallback()) : new URL(getMetadata('google-login-redirect'))?.href;
   } catch {
-    // Do nothing at all
+    // Do nothing
   }
 
   await window.adobeIMS.socialHeadlessSignIn({
--- a/libs/features/google-login.js
+++ b/libs/features/google-login.js
@@ -7,7 +7,7 @@ const onToken = async (getMetadata, data, getConfig) => {
   let destination;
   const config = getConfig();
   try {
-    destination = typeof config.googleLoginURLCallback === 'function' ? new URL(config.googleLoginURLCallback()) : new URL(getMetadata('google-login-redirect'))?.href;
+    destination = typeof config.googleLoginURLCallback === 'function' ? new URL(await config.googleLoginURLCallback()) : new URL(getMetadata('google-login-redirect'))?.href;
   } catch {
     // Do nothing
   }
--- a/libs/features/google-login.js
+++ b/libs/features/google-login.js
@@ -3,10 +3,11 @@ const GOOGLE_ID = '530526366930-l874a90ipfkn26naa71r010u8epp39jt.apps.googleuser
 const PLACEHOLDER = 'feds-googleLogin';
 const WRAPPER = 'feds-profile';
 
-const onToken = async (getMetadata, data) => {
+const onToken = async (getMetadata, data, getConfig) => {
   let destination;
+  const config = getConfig();
   try {
-    destination = new URL(getMetadata('google-login-redirect'))?.href;
+    destination = typeof config.googleLoginURLCallback === 'function' ? new URL(config.googleLoginURLCallback()) : new URL(getMetadata('google-login-redirect'))?.href;
   } catch {
     // Do nothing
   }
@@ -30,7 +31,7 @@ const onToken = async (getMetadata, data) => {
   });
 };
 
-export default async function initGoogleLogin(loadIms, getMetadata, loadScript) {
+export default async function initGoogleLogin(loadIms, getMetadata, loadScript, getConfig) {
   try {
     await loadIms();
   } catch {
@@ -45,7 +46,7 @@ export default async function initGoogleLogin(loadIms, getMetadata, loadScript) 
 
   window.google?.accounts?.id?.initialize({
     client_id: GOOGLE_ID,
-    callback: (data) => onToken(getMetadata, data),
+    callback: (data) => onToken(getMetadata, data, getConfig),
     prompt_parent_id: PLACEHOLDER,
     cancel_on_tap_outside: false,
   });
--- a/libs/features/google-login.js
+++ b/libs/features/google-login.js
@@ -9,7 +9,7 @@ const onToken = async (getMetadata, data, getConfig) => {
   try {
     destination = typeof config.googleLoginURLCallback === 'function' ? new URL(await config.googleLoginURLCallback()) : new URL(getMetadata('google-login-redirect'))?.href;
   } catch {
-    // Do nothing
+    // Do nothing at all
   }
 
   await window.adobeIMS.socialHeadlessSignIn({
--- a/libs/features/google-login.js
+++ b/libs/features/google-login.js
@@ -7,7 +7,8 @@ const onToken = async (getMetadata, data, getConfig) => {
   let destination;
   const config = getConfig();
   try {
-    destination = new URL(typeof config.googleLoginURLCallback === 'function' ? await config.googleLoginURLCallback() : getMetadata('google-login-redirect'))?.href;
+    destination = new URL(typeof config.googleLoginURLCallback === 'function' ? await config.googleLoginURLCallback()
+      : getMetadata('google-login-redirect'))?.href;
   } catch {
     // Do nothing
   }
--- a/libs/features/google-login.js
+++ b/libs/features/google-login.js
@@ -7,7 +7,7 @@ const onToken = async (getMetadata, data, getConfig) => {
   let destination;
   const config = getConfig();
   try {
-    destination = typeof config.googleLoginURLCallback === 'function' ? new URL(await config.googleLoginURLCallback()) : new URL(getMetadata('google-login-redirect'))?.href;
+    destination = new URL(typeof config.googleLoginURLCallback === 'function' ? await config.googleLoginURLCallback() : getMetadata('google-login-redirect'))?.href;
   } catch {
     // Do nothing
   }
--- a/libs/scripts/delayed.js
+++ b/libs/scripts/delayed.js
@@ -49,7 +49,7 @@ export const loadPrivacy = async (getConfig, loadScript) => {
   });
 };
 
-export const loadGoogleLogin = async (getMetadata, loadIms, loadScript) => {
+export const loadGoogleLogin = async (getMetadata, loadIms, loadScript, getConfig) => {
   const googleLogin = getMetadata('google-login')?.toLowerCase();
   if (window.adobeIMS?.isSignedInUser() || !['mobile', 'desktop', 'on'].includes(googleLogin)) return;
   const desktopViewport = window.matchMedia('(min-width: 900px)').matches;
@@ -57,7 +57,7 @@ export const loadGoogleLogin = async (getMetadata, loadIms, loadScript) => {
   if (googleLogin === 'desktop' && !desktopViewport) return;
 
   const { default: initGoogleLogin } = await import('../features/google-login.js');
-  initGoogleLogin(loadIms, getMetadata, loadScript);
+  initGoogleLogin(loadIms, getMetadata, loadScript, getConfig);
 };
 
 /**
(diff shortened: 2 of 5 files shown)


libs/scripts/delayed.js
```diff
@@ -69,8 +69,11 @@ const loadDelayed = ([
   loadScript,
   loadStyle,
   loadIms,
+  MILO_EVENTS,
 ], DELAY = 3000) => new Promise((resolve) => {
   setTimeout(() => {
+    const event = new Event(MILO_EVENTS.DELAYED);
+    document.dispatchEvent(event);
```
mokimo, line 76: As you don't re-use the event
```suggestion
    document.dispatchEvent(new Event(MILO_EVENTS.DELAYED));
```

libs/utils/utils.js
```diff
@@ -1026,6 +1026,8 @@ async function checkForPageMods() {
 }
 
 async function loadPostLCP(config) {
+  const event = new Event(MILO_EVENTS.POSTLCP);
+  document.dispatchEvent(event);
```
mokimo, line 1030: As you don't re-use the event
```suggestion
    document.dispatchEvent(new Event(MILO_EVENTS.POSTLCP));
```

mokimo (commented):
I'm not particularly fond of the idea of consumers hooking into the milo rendering pipeline.
This event previously has been removed https://github.com/adobecom/milo/pull/1282/files#diff-28218f239d8d9c4d41e1924854440b50964619e56058fdd0ae243ff5ff6c616dL126 I can imagine with a tight coupling, this would have broken consumer code.

Can you not run your code after loadArea? https://github.com/adobecom/bacom/blob/stage/scripts/scripts.js#L189

vhargrave:
> I'm not particularly fond of the idea of consumers hooking into the milo rendering pipeline. This event previously has been removed https://github.com/adobecom/milo/pull/1282/files#diff-28218f239d8d9c4d41e1924854440b50964619e56058fdd0ae243ff5ff6c616dL126 I can imagine with a tight coupling, this would have broken consumer code.
> 
> Can you not run your code after loadArea? https://github.com/adobecom/bacom/blob/stage/scripts/scripts.js#L189
> 
> Also to add to this `postLCP` might not really be postLCP. This method might go away or get refactored - from a performance standpoint, we also have the 2nd section popping up as LCP or other elements, currently we have a _very_ late LCP due to congestion shortly after loading the first section. We can migrate the event, but it makes change harder if you have consuming code relying on something that comes from milo core.

Hey @mokimo , 
Especially regarding the load delayed event, no we cannot. We have to execute the code before the google-yolo code on your end runs, and this is the only way to hook ourselves in. 
To be honest I also disagree with what you're saying about clients being able to hook themselves in. I think this is actually very useful and allows important things to be done at the right time. Otherwise you're forcing clients as you said to run their code before or after loadArea is called. There are times though, when code must be executed during because it's a long wait in between. 
Yes, we can't remove without breaking things, but that's kind of the case with most things in utils no? 
I'd conclude by saying that we discussed this with the previous milo core team at the beginning of this migration. -https://wiki.corp.adobe.com/pages/viewpage.action?pageId=3091611394

libs/utils/utils.js
```diff
@@ -1026,6 +1026,7 @@ async function checkForPageMods() {
 }
 
 async function loadPostLCP(config) {
+  document.dispatchEvent(new Event(MILO_EVENTS.POSTLCP));
```
npeltier, line 1029: can we move this at the bottom of loadPostLCP? i would give milo loadPostLCP calls priority over consumer's

libs/utils/utils.js
```diff
@@ -1074,8 +1075,7 @@ export function scrollToHashedElement(hash) {
 }
 
 export async function loadDeferred(area, blocks, config) {
-  const event = new Event(MILO_EVENTS.DEFERRED);
-  area.dispatchEvent(event);
+  area.dispatchEvent(new Event(MILO_EVENTS.DEFERRED));
```
npeltier, line 1078: same here, can we move it at the bottom?

npeltier (commented):
@vhargrave @mokimo i think i stand in the middle :) i agree with Victor that giving hooks to consumer is great, but with Okan that running code before some pretty structural milo processes are run might lead to surprises.
another thing: is it something we could do in loadPostLCP to cut event dispatch from main thread with timeout?

libs/utils/utils.js
vhargrave: @npeltier makes sense - I can move this one yes 👍

libs/utils/utils.js
vhargrave: this one I really need to execute at this point in time. Otherwise my hook in code is too late in Express. I could add two events, one at the beginning and one at the end, but I suspect we'll never use the one at the end. 
Are you ok with me keeping it at the beginning anyway? It's not as critical as the lcp stuff as well imo.

npeltier: does it make any sense caring about main thread at this stage?

vhargrave: How do you mean @npeltier? All our javascript runs on the main thread and and I think the only way to move logic to another thread would be to use a web worker.

vhargrave:
For fresh eyes looking at the PR, I've talked with with Okah & Rares and agreed to remove the events and instead add a callback to the google login function due to the concern of keeping these events maintainable. 
Maybe we can can come back to the topic of hooks for consumers at some later point in time though. 
cc: @mokimo @narcis-radu @npeltier @overmyheadandbody @fullcolorcoder

libs/features/google-login.js
```diff
@@ -3,10 +3,11 @@ const GOOGLE_ID = '530526366930-l874a90ipfkn26naa71r010u8epp39jt.apps.googleuser
 const PLACEHOLDER = 'feds-googleLogin';
 const WRAPPER = 'feds-profile';
 
-const onToken = async (getMetadata, data) => {
+const onToken = async (getMetadata, data, getConfig) => {
   let destination;
+  const config = getConfig();
   try {
-    destination = new URL(getMetadata('google-login-redirect'))?.href;
+    destination = typeof config.googleLoginURLCallback === 'function' ? new URL(config.googleLoginURLCallback()) : new URL(getMetadata('google-login-redirect'))?.href;
```
narcis-radu, line 10: why do we need a new method that returns an url? can't the existing metadata be used and updated?

vhargrave: @narcis-radu that hook already existed before this PR. I just simplified it a little while originally adding two other hooks. 
As for the why we need a callback, sure. 
currently in Express when we display the google yolo popup, we redirect the user to the link that is in the main CTA or floating button in Express. This means that the redirect link is dynamic and is only "known", once the page has loaded. I'd like to pass a method, so that the method can resolve and figure out what the link should be. Setting metadata doesn't quite work, because I don't know when the page loads what the link will be. 
Does that make sense or do you need a few more details?

libs/utils/utils.js
mokimo: Is the deferred event still needed after our discussion? 
My impression was that we shift loading either pre `loadArea()` or after it.

libs/features/google-login.js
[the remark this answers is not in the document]
vhargrave: I wasn't planning on it tbh, but I think it's a great idea. 
Done! 🙌

narcis-radu: It makes sense, the only concern is that we're adding too much complexity - the URL can be specified as a metadata or as a callback. I was wondering maybe you can do the following: in your project, update the metadata as soon as you have the URL ready, this way you wouldn't need to set a callback in Milo. Would that work ?

vhargrave: I completley agree with you, but not really unfortunately. I only know what the URL is for certain once all the sections have been processed. So I'd either need an event or a hook which tells me that runs once all the sections have finished processing and sets the metadata then, or I need a callback like this which resolves at the precise point in time that I need it.

vhargrave: actually hang on, I might have spoken too soon. Let me check 👍

vhargrave: ok, no , it doesn't work. If it was just Express blocks I needed to handle your solution would work. I really like it actually. 
The problem is that for milo blocks that load in I can't set this metadata in between while they're loading (unless I had a hook or something), so this method has to resolve as said once the page has finished loading.

narcis-radu (approved):

libs/features/google-login.js
```diff
@@ -3,10 +3,11 @@ const GOOGLE_ID = '530526366930-l874a90ipfkn26naa71r010u8epp39jt.apps.googleuser
 const PLACEHOLDER = 'feds-googleLogin';
 const WRAPPER = 'feds-profile';
 
-const onToken = async (getMetadata, data) => {
+const onToken = async (getMetadata, data, getConfig) => {
   let destination;
+  const config = getConfig();
   try {
-    destination = new URL(getMetadata('google-login-redirect'))?.href;
+    destination = typeof config.googleLoginURLCallback === 'function' ? new URL(await config.googleLoginURLCallback()) : new URL(getMetadata('google-login-redirect'))?.href;
```
narcis-radu, line 10: might be a nit: you should return the `href` string and based on that info you can rewrite this line as:
```suggestion
    destination = new URL(typeof config.googleLoginURLCallback === 'function' ? await config.googleLoginURLCallback() : getMetadata('google-login-redirect'))?.href;
```

mokimo (approved):

libs/features/google-login.js
vhargrave: oh thanks, I think it was more than just a nit tbh. Done!

fullcolorcoder (approved):

test/scripts/delayed.test.js
```diff
@@ -42,20 +42,22 @@ describe('Delayed', () => {
...
     const clock = sinon.useFakeTimers({ toFake: ['setTimeout'] });
     document.querySelector('head')?.insertAdjacentHTML('beforeend', '<meta name="interlinks" content="on">');
-    loadDelayed([getConfig, getMetadata, loadScript, loadStyle, loadIms]).then((module) => {
-      expect(module).to.exist;
-      expect(typeof module === 'object').to.equal(true);
-    });
+    loadDelayed([getConfig, getMetadata, loadScript, loadStyle, loadIms])
+      .then((module) => {
+        expect(module).to.exist;
+        expect(typeof module === 'object').to.equal(true);
```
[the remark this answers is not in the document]
vhargrave: I agree. It's not my code, but I've corrected 👍

overmyheadandbody:
I've removed the Ready for Stage label, as I didn't see any QA performed for this. We usually get either @SilviuLCF or @NadiiaSokolova involved to test and apply the label. The story is on a different board, so if you have QEs of your own, they can definitely help out, but I want to avoid pushing untested code.

vhargrave:
> I've removed the Ready for Stage label, as I didn't see any QA performed for this. We usually get either @SilviuLCF or @NadiiaSokolova involved to test and apply the label. The story is on a different board, so if you have QEs of your own, they can definitely help out, but I want to avoid pushing untested code.

@overmyheadandbody hmm, ok. The express team does technically have a QA now, but they have no experience with milo . They will have to get started at some point though, so I'll talk today with the team and @dstrong23 about them starting to test these PRs.... and if not I will gratefully take other QA support 🙏 
cc: @mokimo @SilviuLCF & @NadiiaSokolova

libs/features/google-login.js
```diff
@@ -3,10 +3,12 @@ const GOOGLE_ID = '530526366930-l874a90ipfkn26naa71r010u8epp39jt.apps.googleuser
 const PLACEHOLDER = 'feds-googleLogin';
 const WRAPPER = 'feds-profile';
 
-const onToken = async (getMetadata, data) => {
+const onToken = async (getMetadata, data, getConfig) => {
   let destination;
+  const config = getConfig();
   try {
-    destination = new URL(getMetadata('google-login-redirect'))?.href;
+    destination = new URL(typeof config.googleLoginURLCallback === 'function' ? await config.googleLoginURLCallback()
```
fullcolorcoder, line 10: @vhargrave In testing this further with Haris, I noticed you wouldn't ever get to getMetaData() as a fallback as the googleLoginURLCallback doesn't handle that scenario. I think we need it to look for the metadata in the callback in the event of nothing to return.

fullcolorcoder (changes requested):

fullcolorcoder (approved):

hadobe:
Looks good. Approved to move to stage. I don't have permission to update the Label.

PR #2906, opened 2024-09-18
MWPW-157555: Remove useless logic and putting import commerce.js top.

seanchoi-dev:
Resolves: [MWPW-157555](https://jira.corp.adobe.com/browse/MWPW-157555)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/seanchoi/merch-test?martech=off&georouting=off
- After: https://MWPW-157555--milo--adobecom.hlx.page/drafts/seanchoi/merch-test?martech=off&georouting=off

**Consumer tests:**
- Before: https://main--cc--adobecom.hlx.live/products/photoshop?martech=off&georouting=off&skipBanner=true
- After: https://main--cc--adobecom.hlx.live/products/photoshop?martech=off&georouting=off&skipBanner=true&milolibs=mwpw-157555


npeltier (approved):
looks good to me :) may be some more talkative commit message would be great :)

libs/blocks/merch/merch.js
```diff
@@ -2,6 +2,7 @@ import {
   createTag, getConfig, loadArea, loadScript, loadStyle, localizeLink,
 } from '../../utils/utils.js';
 import { replaceKey } from '../../features/placeholders.js';
+import '../../deps/mas/commerce.js';
```
Axelcureno, line 5: Is this needed? This is also being imported in L427

seanchoi-dev: Yes, we want to start to load this ASAP when merch.js get loaded.
This is what task's description:
"Remove the related logic and move import of commerce.js to the top."

Axelcureno (approved):

libs/blocks/merch/merch.js
```diff
@@ -422,14 +423,7 @@ export async function initService(force = false) {
...
   initService.promise = initService.promise ?? polyfills().then(async () => {
-    const { hostname, searchParams } = new URL(window.location.href);
-    let commerceLibPath = '../../deps/mas/commerce.js';
-    if (/hlx\.(page|live)$|localhost$|www\.stage\.adobe\.com$/.test(hostname)) {
-      const maslibs = searchParams.get('maslibs');
-      if (maslibs) {
-        commerceLibPath = `${getMasBase(hostname, maslibs)}/libs/commerce.js`;
-      }
-    }
+    const commerceLibPath = '../../deps/mas/commerce.js';
```
yesil, line 426: I think line and next one is not needed anymore.

yesil:
@seanchoi-dev  the following test should be removed.

`Merch Block > M@S consumption > maslibs parameter > should load commerce.js via maslibs`, it is failing.

libs/blocks/merch/merch.js
seanchoi-dev: After remove those, how do you want to define `commerceLib.init()`  from next line?

seanchoi-dev:
> @seanchoi-dev the following test should be removed.
> 
> `Merch Block > M@S consumption > maslibs parameter > should load commerce.js via maslibs`, it is failing.

I removed it.

libs/blocks/merch/merch.js
yesil: moving it to line 5 should work no?

seanchoi-dev: Maybe I am not getting it but, my question is,

The function scope goes like initService() -> initService.promise.then() -> commerceLib.init().
And the commerceLib is defined relying on importing ('../../deps/mas/commerce.js')
I am not sure how we want to define
image

Should I just put `const commerceLib = await import('../../deps/mas/commerce.js');` on top instead `import '../../deps/mas/commerce.js';` ?

yesil: `import commerceLib from '../../deps/mas/commerce.js';` should work.

seanchoi-dev: I don't think `commerceLib` is available from commerce.js.
I tested and it returns an error
`Failed loading merch SyntaxError: The requested module '../../deps/mas/commerce.js' does not provide an export named 'default' (at merch.js:5:8)`

Instead,
`const commerceLib = await import('../../deps/mas/commerce.js');` on line 6.

Worked fine. (await is still needed).

yesil: sorry @seanchoi-dev  it should be

`import * as commerceLib from '../../deps/mas/commerce.js';`

I just tried and it works.
I would like to clean up the code and get rid of code that is not needed any more.

yesil (approved):
@seanchoi-dev can you provide valid before/after URLs for PSI checks?

PR #2914, opened 2024-09-19
MWPW-158071: Optimize LCP loading times

mokimo:
### Description
It's relatively hard to track performance beforehand, however we observed a +500MS (13%) gain on CC pages speeding up loading [placeholders](https://github.com/adobecom/milo/pull/2752). We've also seen good success with [preloading](https://github.com/orgs/adobecom/discussions/2719)

### Decorate.js
[This PR](https://github.com/adobecom/milo/pull/2493) initially introduced the idea of preloading `decorate.js` for a performance gain, but introduced _too much_ flexibility. The idea of preloading `decorate.js` for specifically the two common (hero-)marquees blocks is valuable and worth the added 3 lines of code, as neither is likely to change anytime soon.

### Impressions
This should provide good benefits on a wide arrange of blocks, consumers and pages by optimizing the load order and chains, however we will want to observe the CRUM/RUM/CRUX data after merging the changes in this PR.

Given the success we seen with preloading & parallelizing work in general, this harmonizes the timeline and ensures the browser loads everything as fast as it can, without any impact on TBT/CLS.

Before
Screenshot 2024-09-19 at 15 13 16
After
Screenshot 2024-09-19 at 15 13 24

Resolves: [MWPW-158071](https://jira.corp.adobe.com/browse/MWPW-158071)

**Test URLs:**
- After: https://optimize-lcp-loading-times--milo--mokimo.hlx.page/?martech=off

Bacom
- Before: https://stage--bacom--adobecom.hlx.live/products/real-time-customer-data-platform/rtcdp?martech=off
- After: https://stage--bacom--adobecom.hlx.live/products/real-time-customer-data-platform/rtcdp?martech=off&milolibs=optimize-lcp-loading-times--milo--mokimo

CC
- Before: https://main--cc--adobecom.hlx.live/de/products/photoshop?martech=off
- After: https://main--cc--adobecom.hlx.live/de/products/photoshop?martech=off&milolibs=optimize-lcp-loading-times--milo--mokimo

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -1256,7 +1256,7 @@ async function processSection(section, config, isDoc) {
   const loadBlocks = [...stylePromises];
   if (section.preloadLinks.length) {
     const [modals, blocks] = partition(section.preloadLinks, (block) => block.classList.contains('modal'));
-    blocks.forEach((block) => loadBlocks.push(loadBlock(block)));
+    await Promise.all(blocks.map((block) => loadBlock(block)));
     modals.forEach((block) => loadBlock(block));
   }
 
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -446,27 +446,27 @@ export async function loadTemplate() {
   await Promise.all([styleLoaded, scriptLoaded]);
 }
 
+function getBlockData(block) {
+  const name = block.classList[0];
+  const { miloLibs, codeRoot, mep } = getConfig();
+  const base = miloLibs && MILO_BLOCKS.includes(name) ? miloLibs : codeRoot;
+  let path = `${base}/blocks/${name}`;
+  if (mep?.blocks?.[name]) path = mep.blocks[name];
+  const blockPath = `${path}/${name}`;
+  const hasStyles = AUTO_BLOCKS.find((ab) => Object.keys(ab).includes(name))?.styles ?? true;
+
+  return { blockPath, name, hasStyles };
+}
+
 export async function loadBlock(block) {
   if (block.classList.contains('hide-block')) {
     block.remove();
     return null;
   }
-
-  const name = block.classList[0];
-  const hasStyles = AUTO_BLOCKS.find((ab) => Object.keys(ab).includes(name))?.styles ?? true;
-  const { miloLibs, codeRoot, mep } = getConfig();
-
-  const base = miloLibs && MILO_BLOCKS.includes(name) ? miloLibs : codeRoot;
-  let path = `${base}/blocks/${name}`;
-
-  if (mep?.blocks?.[name]) path = mep.blocks[name];
-
-  const blockPath = `${path}/${name}`;
-
+  const { name, blockPath, hasStyles } = getBlockData(block);
   const styleLoaded = hasStyles && new Promise((resolve) => {
     loadStyle(`${blockPath}.css`, resolve);
   });
-
   const scriptLoaded = new Promise((resolve) => {
     (async () => {
       try {
@@ -731,6 +731,8 @@ function decorateDefaults(el) {
 }
 
 function decorateHeader() {
+  const breadcrumbs = document.querySelector('.breadcrumbs');
+  breadcrumbs?.remove();
   const header = document.querySelector('header');
   if (!header) return;
   const headerMeta = getMetadata('header');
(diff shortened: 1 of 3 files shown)


libs/utils/utils.js
```diff
@@ -731,6 +731,8 @@ function decorateDefaults(el) {
 }
 
 function decorateHeader() {
+  const breadcrumbs = document.querySelector('.breadcrumbs');
```
mokimo, line 734: I noticed that if a breadcrumbs is authored as first block, we immediately load it. The global-navigation should take care of this and on L752 we move the breadcrumbs into the header `if (breadcrumbs) header.append(breadcrumbs);`

mokimo: // FYI @salonijain3 @sharmrj @bandana147

mokimo:
nala tests also failing on stage: https://github.com/adobecom/milo/pull/2910

chrischrischris (approved):

JasonHowellSlavin (commented):
Looks good. Are there any unit tests for utils that require updating?

robert-bogos (approved):
Looks good 👍 
Maybe we can get the psi check to also pass.

libs/utils/utils.js
vhargrave: @mokimo where are you seeing that we immediately load this? For me it looks like it was already working properly before any change here. 😅

mokimo: Here's a test experience @vhargrave https://main--milo--adobecom.aem.page/drafts/osahin/education

When you turn the header off, it treats the breadcrumbs as block, which 404s.

vhargrave: ah ok, i see thanks 👍

mokimo:
@JasonHowellSlavin I've added a unit test covering the new behavior, other than that the old ones should continue working as is

vhargrave (commented):
looking pretty good!

libs/utils/utils.js
```diff
@@ -1221,41 +1223,51 @@ export function partition(arr, fn) {
...
+}
 
+async function processSection(section, config, isDoc) {
+  await resolveInlineFrags(section);
+  const firstSection = section.el.dataset.idx === '0';
+  const stylePromises = firstSection ? preloadBlocks(section.blocks) : [];
+  preloadBlocks(section.preloadLinks);
+  await Promise.all([
+    decoratePlaceholders(section.el, config),
+    decorateIcons(section.el, config),
```
vhargrave, line 1254: @mokimo oh, one last little thing , you can't think of any cases where we'd have icons in an Autoblock right? I don't think we do so should be fine, but if yes this could break that experience.

libs/utils/utils.js
```diff
@@ -1221,41 +1223,51 @@ export function partition(arr, fn) {
...
+  ]);
+  const loadBlocks = [...stylePromises];
```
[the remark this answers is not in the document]
mokimo: Kind of. We _usually_ await them. `processSection` -> `loadBlock` -> `loadLink` 
```js
// loadLink
  let link = document.head.querySelector(`link[href="${href}"]`);
  if (link) { return callback('noop') } // <-- we get in this function, not awaiting anything as we already added a link to load the styles on the page
```

overmyheadandbody (approved):
Glad you managed to avoid calling the placeholder logic multiple times. This is worth trying out; in the future, we might want to get data on additional blocks that might be used as LCP and expand the current marquee list.

libs/utils/utils.js
mokimo: I had a quick look at the current autoblocks, and haven't seen any cases where the auto block would add icons in at least an obvious way. 🤞

vhargrave (approved):
looking good!
i think the loadLink method would be good to take a look at, but defnitely out of scope for your ticket.

libs/utils/utils.js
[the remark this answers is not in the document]
mokimo: I tested that, I don't think there's a way to get any notification after a link has loaded. E.g. if you listen to `link.onload = (e) => callback(e.type);` after it has already loaded, it's not called.

If I missed something, would be happen to adapt!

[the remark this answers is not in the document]
mokimo: I guess we could use this? https://developer.mozilla.org/en-US/docs/Web/API/HTMLLinkElement/sheet

```
    if (link.sheet) {
      callback('already-loaded');
    } else {
      link.onload = (e) => callback(e.type);
      link.onerror = (e) => callback(e.type);
    }
```

narcis-radu (approved):

SilviuLCF (approved):
Ready to stage , Regression PASS , testing details https://jira.corp.adobe.com/browse/MWPW-158913

PR #2919, opened 2024-09-19
MWPW-158142 [MEP] check for MEP placeholders as fragmentless content is inserted

AdobeLinhart:
Now that MEP supports fragmentless content, people are going to use MEP placeholders more. To prevent the performance hit of using placeholders in section 1, we should update content as we insert it.

Inside the function called "createContent" there is an if statement that triggers when when the new content is not a fragment. If you run getConfig() there is a node called placeholders if there were MEP specific placeholders. I want to check the content string for placeholders and if they exist check the MEP placholders to do a replace before replace or insertion.

Resolves: [MWPW-158142](https://jira.corp.adobe.com/browse/MWPW-158142)

QA Instructions: Load the before and after links and check the performance tab in dev tools for when placeholders.json is downloaded.  In the before link it is before LCP, in the after link it is after.

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q4/mepnestedplaceholders/?milolibs=mepnestedplaceholders&martech=off
- After: https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q4/mepnestedplaceholders/?milolibs=placeholderperformance&martech=off
- psicheck: https://placeholderperformance--milo--adobecom.hlx.page/?martech=off


JasonHowellSlavin (commented):
Looks good. Any unit tests to update?

AdobeLinhart:
Linting should be fixed.

@JasonHowellSlavin This utilizes a reusable function that's also part of one of Viv's branches. The unit test is the same (parseNestedPlaceholders.test).

vgoodric (approved):

markpadbe (approved):

PR #2923, opened 2024-09-19
MWPW-157888 - Hero marquee w/ no-min-height and no l/r padding when using a full-width variant in masonry(repost)

ryanmparrish:
FYI: This a repost from [reverted PR](https://github.com/adobecom/milo/pull/2905#event-14314262936) due to some issues seen w/ bg images not showing. Testing pages didn't have these examples initially so i've added em. Sorry for the bump. :) 

---

As and author in a milo project using a [hero-marquee]

I want to use this block with out a minimal height so the block has a natural height based on content.
I want to use this block in a masonry grid w/ the variant `full-width` and not have the extra left/right padding seen in this scenario.
Example screenshot - Notice the block height is huge, the l/r padding is causing the block to go outside its bounds.

image

Fixed

Add a no-min-height variant to the hero-marquee.
Address padding issue on hero-block. Erroneously comes from the has-bg class being added.
Resolves: [MWPW-157888](https://jira.corp.adobe.com/browse/MWPW-157888)

Test URLs:

Before: https://main--milo--adobecom.hlx.page/drafts/steenmeyer/hero-editorial-layout-pattern-examples/hero-editorial-no-min-height-masonry1?martech=off
After: https://rparrish-hero-bg-fix-2--milo--adobecom.hlx.page/drafts/steenmeyer/hero-editorial-layout-pattern-examples/hero-editorial-no-min-height-masonry1?martech=off

Hero in a section grid (masonry & -up)
Before: https://main--milo--adobecom.hlx.page/drafts/rparrish/grids/hero-grids?martech=off
After: https://rparrish-hero-bg-fix-2--milo--adobecom.hlx.page/drafts/rparrish/grids/hero-grids?martech=off

Hero Marquee (regression)
These should all look the same
- hero before: https://main--milo--adobecom.hlx.page/docs/library/blocks/hero-marquee?martech=off
- hero after: https://rparrish-hero-bg-fix-2--milo--adobecom.hlx.page/docs/library/blocks/hero-marquee?martech=off

- hero kitchen before: https://main--milo--adobecom.hlx.page/docs/library/kitchen-sink/hero-marquee?martech=off
- hero kitchen after: https://rparrish-hero-bg-fix-2--milo--adobecom.hlx.page/docs/library/kitchen-sink/hero-marquee?martech=off

### DIFF
--- a/libs/blocks/hero-marquee/hero-marquee.css
+++ b/libs/blocks/hero-marquee/hero-marquee.css
@@ -17,6 +17,7 @@
   align-items: center;
 }
 
+.hero-marquee.no-min-height { min-height: unset;}
 .hero-marquee.s-min-height { min-height: var(--s-min-height); }
 .hero-marquee.l-min-height { min-height: var(--l-min-height); }
 
@@ -148,6 +149,11 @@ html[dir="rtl"] .hero-marquee .foreground {
 
 .hero-marquee .foreground div:empty {
   display: none;
+}
+
+div[class*='-up'] .hero-marquee.has-bg.grid-full-width { 
+  padding-left: 0;
+  padding-right: 0;
 }
 
 /* Row Types */
--- a/libs/blocks/hero-marquee/hero-marquee.css
+++ b/libs/blocks/hero-marquee/hero-marquee.css
@@ -151,6 +151,7 @@ html[dir="rtl"] .hero-marquee .foreground {
   display: none;
 }
 
+div[class*='-up'] .hero-marquee.has-bg.full-width,
 div[class*='-up'] .hero-marquee.has-bg.grid-full-width { 
   padding-left: 0;
   padding-right: 0;
--- a/libs/blocks/hero-marquee/hero-marquee.js
+++ b/libs/blocks/hero-marquee/hero-marquee.js
@@ -122,6 +122,7 @@ function extendButtonsClass(copy) {
     button.classList.add('button-xl', 'button-justified-mobile');
   });
 }
+
 function parseKeyString(str) {
   const regex = /^(\w+)\s*\((.*)\)$/;
   const match = str.match(regex);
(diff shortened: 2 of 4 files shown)


libs/blocks/hero-marquee/hero-marquee.js
```diff
@@ -132,6 +133,21 @@ function parseKeyString(str) {
   return result;
 }
 
+function isElementEmpty(element) {
```
overmyheadandbody, line 136: It's very likely I'm missing something, but couldn't we simply check for `element.textContent.trim()`? I'd assume that if there's some text anywhere, the element is not empty.

robert-bogos (commented):
Can we add some unit tests to have the codecov check pass?

libs/blocks/hero-marquee/hero-marquee.js
ryanmparrish: @overmyheadandbody  - Most of the time this row has a <picture> or <video> element and no textContent. This also uses per-viewport background cols that have similar conditions. That's why i've done the overly verbose conditions here, but lmk if you see a slimmer way to handle that.

libs/blocks/hero-marquee/hero-marquee.css
```diff
@@ -150,6 +151,11 @@ html[dir="rtl"] .hero-marquee .foreground {
   display: none;
 }
 
+div[class*='-up'] .hero-marquee.has-bg.grid-full-width { 
+  padding-left: 0;
+  padding-right: 0;
+}
```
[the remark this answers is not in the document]
ryanmparrish: @overmyheadandbody There are some use cases where the hero is in a section grid and 'has-bg`, I have added an additional example in the description now to showcase this. This is more a preventative style just to cover that unlikely authoring scenario.

libs/blocks/hero-marquee/hero-marquee.js
overmyheadandbody: Since we know of these `<picture>` and `<video>` use-cases, couldn't we do something like
```suggestion
  return element.textContent.trim() && !element.querySelector('source, src');
```

Sorry if I'm a bit dense about this, this just seems more complicated than needed to me. If you have other reasons why this wouldn't work we can move forward with this.

overmyheadandbody (commented):
Still some concerns about `isElementEmpty`. If the suggestion still doesn't work, we can leave it as-is for now

libs/blocks/hero-marquee/hero-marquee.js
mokimo: Just an FYI depending on the timing of this, the 'source' is not available // but added via an intersection observer.  So it might be safer to check if it has a `video` element

ryanmparrish: Thanks for the suggestion. I was overthinking it and yea since we know these are either text, video, picture etc this makes sense to just query that. I've updated the logic on that and just removed that isElementEmpty() func. 
I tested all the pages and it checks out.

ryanmparrish:
@overmyheadandbody & @robert-bogos  - I've updated this w/ your suggestions simplifying that elementEmpty() func and full coverage. Thanks for the feedback.

mokimo (approved):

overmyheadandbody (approved):

libs/blocks/hero-marquee/hero-marquee.js
overmyheadandbody: Great, thanks! Glad this could be simplified!

PR #2924, opened 2024-09-19
MWPW-158407, 158404, 158406, 158408, 158461: Ignoring decorateMeta if the metadata is from promo.

seanchoi-dev:
Resolves: 
- [MWPW-158407](https://jira.corp.adobe.com/browse/MWPW-158407)
- [MWPW-158461](https://jira.corp.adobe.com/browse/MWPW-158461)
- [MWPW-158404](https://jira.corp.adobe.com/browse/MWPW-158404)
- [MWPW-158406](https://jira.corp.adobe.com/browse/MWPW-158406)
- [MWPW-158408](https://jira.corp.adobe.com/browse/MWPW-158408)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page?martech=off&georouting=off
- After: https://MWPW-158407--milo--adobecom.hlx.page?martech=off&georouting=off

**Consumer test w/ real case:**
- Before: https://main--homepage--adobecom.hlx.page/homepage/index-loggedout?lanadebug=1
- After: https://main--homepage--adobecom.hlx.page/homepage/index-loggedout?lanadebug=1&milolibs=MWPW-158407

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -1147,7 +1147,7 @@ function decorateMeta() {
   const { origin } = window.location;
   const contents = document.head.querySelectorAll(`[content*=".${SLD}."]`);
   contents.forEach((meta) => {
-    if (meta.getAttribute('property') === 'hlx:proxyUrl' || meta.getAttribute('name').endsWith('schedule')) return;
+    if (meta.getAttribute('property') === 'hlx:proxyUrl' || meta.getAttribute('name')?.endsWith('schedule')) return;
     try {
       const url = new URL(meta.content);
       const localizedLink = localizeLink(`${origin}${url.pathname}`);
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -1147,8 +1147,7 @@ function decorateMeta() {
   const { origin } = window.location;
   const contents = document.head.querySelectorAll(`[content*=".${SLD}."]`);
   contents.forEach((meta) => {
-    if (meta.getAttribute('property') === 'hlx:proxyUrl') return;
-    if (meta.getAttribute('name').endsWith('schedule')) return;
+    if (meta.getAttribute('property') === 'hlx:proxyUrl' || meta.getAttribute('name').endsWith('schedule')) return;
     try {
       const url = new URL(meta.content);
       const localizedLink = localizeLink(`${origin}${url.pathname}`);
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -1148,6 +1148,7 @@ function decorateMeta() {
   const contents = document.head.querySelectorAll(`[content*=".${SLD}."]`);
   contents.forEach((meta) => {
     if (meta.getAttribute('property') === 'hlx:proxyUrl') return;
+    if (meta.getAttribute('name').endsWith('schedule')) return;
     try {
       const url = new URL(meta.content);
       const localizedLink = localizeLink(`${origin}${url.pathname}`);


libs/utils/utils.js
```diff
@@ -1148,6 +1148,7 @@ function decorateMeta() {
   const contents = document.head.querySelectorAll(`[content*=".${SLD}."]`);
   contents.forEach((meta) => {
     if (meta.getAttribute('property') === 'hlx:proxyUrl') return;
+    if (meta.getAttribute('name').endsWith('schedule')) return;
```
Ruchika4, line 1151: Can we put this if logic with an OR in the if condition on line number 1150 like below

`if (meta.getAttribute('property') === 'hlx:proxyUrl' || meta.getAttribute('name').endsWith('schedule')) return;`

libs/utils/utils.js
```diff
@@ -1147,7 +1147,7 @@ function decorateMeta() {
   const { origin } = window.location;
   const contents = document.head.querySelectorAll(`[content*=".${SLD}."]`);
   contents.forEach((meta) => {
-    if (meta.getAttribute('property') === 'hlx:proxyUrl') return;
+    if (meta.getAttribute('property') === 'hlx:proxyUrl' || meta.getAttribute('name').endsWith('schedule')) return;
```
overmyheadandbody, line 1150: Are there any other meta names with this issue, other than "schedule"? We might be in a situation where someone else has another schedule referencing a URL and they will be forced to update the meta name. If this is just `<meta name="schedule">` we're interested in, could we:
```suggestion
    if (meta.getAttribute('property') === 'hlx:proxyUrl' || meta.getAttribute('name') === 'schedule') return;
```

seanchoi-dev: @overmyheadandbody  The metadata can be like `americas_schedule` or `apac_schedule` according to @3ch023 .

npeltier: yes it can be emea_schedule too :) endswith('schedule') seems good to me

npeltier (approved):

libs/utils/utils.js
chrischrischris: Should it possibly be `.endsWith('_schedule')` or make it more explicit with `.endsWith('_promo_schedule')`?

seanchoi-dev: I think the answer is NO since the issue case (US homepage) has `schedule`. So it can be just schedule w/o prefix.
image

3ch023: confirming that .endsWith('_schedule') will not be enough since we also need to ignore 'schedule' itself.
I updated the doc with all possible variations of this metadata: https://milo.adobe.com/docs/authoring/metadata (see end of table)

>We might be in a situation where someone else has another schedule referencing a URL and they will be forced to update the meta name. 

I understand, but as of now I am not able to find such a field (checked code and https://milo.adobe.com/docs/authoring/metadata). + It is rather unlikely that the field containing a 'pure' URL would end with a 'schedule' in the name. 
If we find such a field we could switch to the exact check of `===` instead of endsWith, but it will cost more bytes in code. 

From my side it's ok to leave it as .endsWith('schedule'), so approving the PR.

afmicka:
@3ch023 @seanchoi-dev replace fragment promo is breaking the page on this branch. There might be some stale blocks on the branch or there might be the real issue. The page loads blank:
https://mwpw-158407--milo--adobecom.hlx.page/drafts/nala/features/promotions/promo-with-fragments

Expected (main): https://main--milo--adobecom.hlx.page/drafts/nala/features/promotions/promo-with-fragments

seanchoi-dev:
> @3ch023 @seanchoi-dev replace fragment promo is breaking the page on this branch. There might be some stale blocks on the branch or there might be the real issue. The page loads blank: https://mwpw-158407--milo--adobecom.hlx.page/drafts/nala/features/promotions/promo-with-fragments
> 
> Expected (main): https://main--milo--adobecom.hlx.page/drafts/nala/features/promotions/promo-with-fragments
> 
> As far as i can see, the error reported comes from the changed file Screenshot 2024-10-01 at 15 30 22

I fixed the js error.

PR #2931, opened 2024-09-23
MWPW-155489 [PEP] Add device type as PEP criteria

sharmrj:
Due to our specific viewport breakpoints, PEP was being served to iPhone/iPad users browsing with their device held horizontally. This lead to a requirement for serving PEP only to desktop users. 

- We've used the same check MEP uses to determine if the user is on a mobile device or not
- We haven't imported the MEP check since, in pages where personalization is turned off, the browser would needlessly download and parse the personalization.js just for the User Agent check.
- Requirements for MEP and PEP won't necessarily move in lockstep, so it makes sense to maintain our own list of User Agents we'll check for in PEP.

Note: As [MDN recommends](https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent#avoiding_user_agent_detection), it's preferable to not do useragent sniffing and prefer feature detection instead. In our case this would mean, if we're redirecting to a web app, do the feature detection for that particular webapp before we decide to redirect the user to that webapp. But this is a more complicated problem to solve and it'll need to be looked into (if it's even worth the effort).

Resolves: [MWPW-155489](https://jira.corp.adobe.com/browse/MWPW-155489)

**Test URLs:**
- After: https://pep-ua--milo--sharmrj.hlx.page/?martech=off
- QA: https://main--cc--adobecom.hlx.page/products/photoshop-lightroom?milolibs=pep-ua--milo--sharmrj&skipPepEntitlements=true

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -637,9 +637,12 @@ class Gnav {
     const state = getMetadata('app-prompt')?.toLowerCase();
     const entName = getMetadata('app-prompt-entitlement')?.toLowerCase();
     const promptPath = getMetadata('app-prompt-path')?.toLowerCase();
+    const hasMobileUA = /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini|Touch/i.test(navigator.userAgent);
+
     if (state === 'off'
       || !window.adobeIMS?.isSignedInUser()
       || !isDesktop.matches
+      || hasMobileUA
       || !entName?.length
       || !promptPath?.length) return;
 


sharmrj:
Nala tests seem to be failing for a reason unrelated to my changes

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -637,9 +637,12 @@ class Gnav {
     const state = getMetadata('app-prompt')?.toLowerCase();
     const entName = getMetadata('app-prompt-entitlement')?.toLowerCase();
     const promptPath = getMetadata('app-prompt-path')?.toLowerCase();
+    const hasMobileUA = /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini|Touch/i.test(navigator.userAgent);
```
npeltier, line 640: nit: wouldn't isMobile be clearer?

npeltier (approved):

libs/blocks/global-navigation/global-navigation.js
sharmrj: We have checks like `isDesktop` that check viewport size, which is what we check by default. I'd like to keep this name to differentiate it from the viewport size checks.

nishantka (approved):

sonawanesnehal3 (approved):

bandana147 (approved):

libs/blocks/global-navigation/global-navigation.js
overmyheadandbody, line 640: Is there any analytics data pointing to all of these being used by our users? The iPod was discontinued more than 2 years ago. Blackberry, IEMobile and Touch don't seem to be all that popular either according to [this](https://gs.statcounter.com/browser-market-share/mobile/worldwide).

PR #2933, opened 2024-09-23
MWPW-158911: Unable to scroll page if the filters modal closed by click outside of the modal

mirafedas:
In merch-sidenav.js, when the filters modal is closed by clicking outside of it (rather than using the 'Close' button), the overlay emits a close event. On this event, we need to remove the merch-modal class; otherwise, the page will remain unscrollable.

Resolves: [MWPW-158911](https://jira.corp.adobe.com/browse/MWPW-158911)

**Test URLs:**
Before: https://main--cc--adobecom.hlx.live/products/catalog?martech=off
After: https://mwpw-158911-modal-scroll--cc--adobecom.hlx.live/products/catalog?martech=off


libs/deps/mas/mas.js
yesil, line 1: this should not be in this PR

yesil:
@npeltier I think `merch-sidenav` should be moved to Milo as well.

libs/features/mas/web-components/test/merch-card-collection.test.html.js
```diff
@@ -89,18 +89,30 @@ runTests(async () => {
...
+          });
+
+          it('removes the class for modal when closing the filters modal by clicking outside the modal', async () => {
             render();
             await delay(100);
             merchCards.shadowRoot.querySelector('#filtersButton').click();
             await delay(100);
-            expect(document.body.classList.contains('merch-modal')).to.be.true;
-            document.querySelector('merch-sidenav').shadowRoot.querySelector('#sidenav').querySelector('sp-link').click();
+            document.querySelector('merch-sidenav').shadowRoot.querySelector('sp-overlay').dispatchEvent(new CustomEvent('close'));
```
yesil, line 111: nit: can you simulate a click to a point outside the modal?

yesil (approved):

libs/features/mas/web-components/test/merch-card-collection.test.html.js
```diff
@@ -89,18 +89,30 @@ runTests(async () => {
               merchCards.shadowRoot.querySelector('#filtersButton').click();
               await delay(100);
               expect(document.body.classList.contains('merch-modal')).to.be.true;
+              document.querySelector('merch-sidenav').removeAttribute('modal');
```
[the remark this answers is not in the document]
mirafedas: This attribute does not get removed when running unit tests (however it does get removed when running the code in a normal browser)

libs/features/mas/web-components/test/merch-card-collection.test.html.js
mirafedas: I tried it, but it doesn't close the modal (only when running unit tests). I think we should cover this not in unit tests, but in international tests (Nala)

chrischrischris (approved):

Axelcureno (approved):

PR #2939, opened 2024-09-23
MWPW-146262 - [Bulk Publish v2] Additional job tools

Sartxi:
Bulk Publishing enhancements to job process component. Adding two new tools:
1. Share icon: when clicking icon the share link URL will be copied to the clipboard.
2. Cancel icon: when clicking icon the job is cancelled in AEM if still unfinished, and cleared from the job panel if complete.

Screenshot 2024-09-23 at 10 09 21 AM

Resolves: [MWPW-146262](https://jira.corp.adobe.com/browse/MWPW-146262)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/tools/bulk-publish?martech=off
- After: https://sartxi-mwpw-144750-job-tools--milo--adobecom.hlx.page/tools/bulk-publish?martech=off


libs/blocks/bulk-publish-v2/components/job-info.js
```diff
@@ -128,18 +143,30 @@ class JobInfo extends LitElement {
...
               class="job-id-link"
               @click=${this.copyInvocationId} />
           ` : nothing}
-          ${this.copiedInvocationId ? html`
-            <div class="job-id-copied">
-              Copied to clipboard!
-            </div>
+          ${invocationId ? html`
+            <img
+              title="Share this Job"
```
ryanmparrish, line 160: Should this be `Copy job link` instead. Since both icons are copying to clipboard the alt text reads a little funny when one says copy, and the other says share. Just a thought when i tested the UI.

libs/blocks/bulk-publish-v2/img/cancel.svg
ryanmparrish, line 1: Probs not a big deal.. should these referenced svg assets be the same inherent size? I know they scale in the view, but might be worth cleaning up the source so they are same/same default view size/remove comments/give meaningful ids. I don't know if there's any accessibility concerns here, likely not, but in other cases it's been suggested to add <title> tags for screen-readers in the svg elements.

libs/blocks/bulk-publish-v2/img/share.svg
```diff
@@ -0,0 +1 @@
+<?xml version="1.0" ?><svg viewBox="0 0 32 32" xmlns="http://www.w3.org/2000/svg"><defs><style>.cls-1{fill:#101820;}</style></defs><title/><g data-name="Layer 51" id="Layer_51"><path class="cls-1" d="M2,29A1.12,1.12,0,0,1,1.69,29,1,1,0,0,1,1,28V27A19,19,0,0,1,17,8.24V4a1,1,0,0,1,1.6-.8l12,9a1,1,0,0,1,0,1.6l-12,9A1,1,0,0,1,17,22V18.25A18.66,18.66,0,0,0,4.93,25.67L2.81,28.59A1,1,0,0,1,2,29ZM19,6V9.12a1,1,0,0,1-.89,1,17,17,0,0,0-15,14.6l.16-.21A20.68,20.68,0,0,1,17.9,16.11a1,1,0,0,1,.77.26,1,1,0,0,1,.33.74V20l9.33-7Z"/></g></svg>
```
ryanmparrish, line 1: maybe remove the style and use currentColor. Looks like these are using diff colors?

libs/blocks/bulk-publish-v2/components/job-info.js
Sartxi: Cool yeah will update that

libs/blocks/bulk-publish-v2/img/cancel.svg
Sartxi: I removed the sizing and color styles on the SVGs

elan-tbx (approved):

ryanmparrish (approved):
LGTM. Thanks for the updates and clarifying on the related Jira ticket. 👍

PR #2941, opened 2024-09-23
MWPW-158652 [MEP] Only output non fatal errors to internal users and skip rows missing an action value

markpadbe:
Currently, a MEP manifest action row with no value in the action column outputs a message in the console. Instead, the action should just be skipped.

This PR creates log and warn functions in personalization.js that only output messages when we are in preview mode, aka if we are going to show the MEP button. 

QA steps: The Before link always logs "Invalid action found" to the console.
The After link should not log "Invalid action found" because it is not in preview mode. However, if you set the mepButton parameter to "on"(i.e., set it to preview mode) the After link should also the log "Invalid action found" message.

Resolves: [MWPW-158652](https://jira.corp.adobe.com/browse/MWPW-158652)

Test URLs:

Before: https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q4/skipmissingaction/base?mepButton=off&martech=off
After: https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q4/skipmissingaction/base?milolibs=addpreviewlog&mepButton=off&martech=off
Psi-check: https://addpreviewlog--milo--adobecom.hlx.page/?martech=off


libs/features/personalization/personalization.js
```diff
@@ -513,6 +523,10 @@ const getVariantInfo = (line, variantNames, variants, manifestPath, fTargetId) =
   // retro support
   const action = line.action?.toLowerCase()
     .replace('content', '').replace('fragment', '').replace('tosection', '');
+  if (!action) {
+    log('Invalid action found: ', line);
```
vgoodric, line 527:
```suggestion
    log('Row found with empty action field: ', line);
```
The message after this one is for an "invalid" action.  I think this row only triggers when the action is empty.  So let's reword it to differentiate and be more clear.

denlight (approved):
Lgtm! Love reusability

vgoodric:
Nala is reporting an issue, but we checked and it is unrelated to this PR

PR #2956, opened 2024-09-25
[MWPW-153616] Update 404 links in region selector

overmyheadandbody:
This makes a few adjustments to the link interaction logic inside the region nav modal. Currently, there are a few scenarios where we might confuse users. Clicking on a link makes a HEAD fetch request to check whether the analog page exists on the newly chosen locale. If so, the user gets redirected to that page. Otherwise, they are being redirected to the root of the newly chosen locale. Most other interactions with region nav links can lead to unexpected outcomes, which this PR tries to minimize.

| **Scenario** | Current behavior | Updated behavior |
| ------------- | ------------- | ------------- |
| **User clicks the link** | gets redirected to the equivalent page or locale root, international cookie set | gets redirected to the equivalent page or locale root, international cookie set |
| **User CMD+clicks the link** | gets redirected to the equivalent page or locale root in the same tab, international cookie set | slow action (>>100ms) - new tab is opened with equivalent page or locale root, international cookie set; <br /><br />fast action (<100ms) - new tab is opened and FOCUSED with equivalent page or locale root, international cookie set |
| **User tabs+enters link** | gets redirected to the equivalent page or locale root, international cookie set | gets redirected to the equivalent page or locale root, international cookie set |
| **User right clicks to copy the link** | link to potentially 404 equivalent page is copied, NO international cookie set | slow action (>>100ms) - link for equivalent page or locale root is copied, NO international cookie set; <br /><br />fast action (<100ms) - link to potentially 404 equivalent page is copied, NO international cookie set |
| **User right clicks to open link in new tab** | new tab is opened with potentially 404 equivalent page, NO international cookie set | slow action (>>100ms) - new tab is opened with equivalent page or locale root, NO international cookie set; <br /><br />fast action (<100ms) - new tab is opened with potentially 404 equivalent page, NO international cookie set |

QA notes:
* please make sure the international cookie is appropriately set according to the table above.

Developer notes:
* I initially tried to make use of the `contextmenu` event, but the `href` does not get reflected when the user eventually clicks to open the link in a new tab or copy its URL;
* the `100ms` timeout is based on trial and error. We want the value to be low, so that the fetch requests are made before the user manages to right click on the link, but we also don't want to make too many such requests for links the user might not actually be interested in. I found `100ms` to be a sweet spot;
* this sort of approach using `mouseover` will not play well with tools checking for 404 links, so authors might still bring up this issue;
* I did consider setting the international cookie after a `contextmenu` interaction coupled with a `visibilitychange` one, but there's no clear indication of whether the interaction was intentional or not. The user may have moved to a completely different tab, may come back to the initial one realizing they clicked the wrong locale, refresh the page after opening the context menu, etc. There's no sure way to check these scenarios, so I preferred to not complicate things for now, as this should be a rare occurrence in the first place.

Resolves: [MWPW-153616](https://jira.corp.adobe.com/browse/MWPW-153616)

**Test URLs:**
- Before: https://main--milo--overmyheadandbody.hlx.page/drafts/ramuntea/brick?martech=off#langnav
- After: https://langnav-context--milo--overmyheadandbody.hlx.page/drafts/ramuntea/brick?martech=off#langnav

### DIFF
--- a/libs/blocks/region-nav/region-nav.js
+++ b/libs/blocks/region-nav/region-nav.js
@@ -1,21 +1,26 @@
 import { getConfig } from '../../utils/utils.js';
 
-/* c8 ignore next 11 */
-function handleEvent(prefix, link) {
+function setInternational(prefix) {
   const domain = window.location.host.endsWith('.adobe.com') ? 'domain=adobe.com' : '';
   const maxAge = 365 * 24 * 60 * 60; // max-age in seconds for 365 days
   document.cookie = `international=${prefix};max-age=${maxAge};path=/;${domain}`;
   sessionStorage.setItem('international', prefix);
+}
+
+function handleEvent({ prefix, link, callback } = {}) {
+  if (typeof callback !== 'function') return;
+
   fetch(link.href, { method: 'HEAD' }).then((resp) => {
     if (!resp.ok) throw new Error('request failed');
-    window.location.assign(link.href);
+    callback(link.href);
   }).catch(() => {
     const prefixUrl = prefix ? `/${prefix}` : '';
-    window.location.assign(`${prefixUrl}/`);
+    callback(`${prefixUrl}/`);
   });
 }
 
-function decorateLink(link, config, path) {
+function decorateLink(link, path) {
+  let hrefAdapted;
   let pathname = link.getAttribute('href');
   if (pathname.startsWith('http')) {
     try { pathname = new URL(pathname).pathname; } catch (e) { /* href does not contain domain */ }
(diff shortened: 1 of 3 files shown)


robert-bogos (commented):
I see a unit test is failing, but does not seem related 🤔

overmyheadandbody:
@robert-bogos - this is due to MAS Unit Tests. They have opened a [PR](https://github.com/adobecom/milo/pull/2957) to address it, but that one is failing NALA checks, so at the moment we're stuck.

libs/blocks/region-nav/region-nav.js
```diff
@@ -25,10 +30,33 @@ function decorateLink(link, config, path) {
   let { href } = link;
   if (href.endsWith('/')) href = href.slice(0, -1);
   link.href = `${href}${path}`;
+
+  link.addEventListener('mouseover', () => {
```
mokimo, line 34: While we don't usually add comments, should we briefly touch on what this does or link the MWPW/PR for context? This is one of the things someone will see in 1-2 years and think "????"

mokimo (approved):

libs/blocks/region-nav/region-nav.js
overmyheadandbody: I did have a similar thought, but I argued that Git Blame would point back to this PR containing all the details 🤷

robert-bogos (approved):

PR #2962, opened 2024-09-26
MWPW-159197 Invoke onReady of standalone gnav when it is rendered

sharmrj:
* We now do `await gnav.init()` so that `onReady` is called only when `gnav.init` is finished executing

Resolves: [MWPW-159197](https://jira.corp.adobe.com/browse/MWPW-159197)

**Test URLs:**
- After: https://await-gnav-init--milo--sharmrj.hlx.page/?martech=off

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -1045,7 +1045,7 @@ export default async function init(block) {
     content,
     block,
   });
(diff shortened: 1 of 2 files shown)


sharmrj:
In order to QA this PR I'll have to raise it from a branch off the main repo (as opposed to my fork)

sharmrj:
See #2963

PR #2970, opened 2024-09-27
MWPW-159113 [Personalization] Parallelize personalization network requests to be as efficient as possible

sharmrj:
* Network requests, entitlements.js, promo-utils.js (if needed), and the manifest files are now made concurrently.

This optimization won't immediately bear fruit since it reveals a different bottleneck
Before: 
image

After:
image

Once entitlements.js, promo-utils.js, and the manifests load concurrently, we find that launch becomes our bottleneck, so thi optimization needs to be combined with two subsequent optimizations in order to bear fruit. Note that photoshop.json takes less time to load in prod environments, so launch is more likely to be bottleneck in prod.

1. Launch.js needs to be split up/optimized. Related tickets: https://jira.corp.adobe.com/browse/MWPW-158739, https://jira.corp.adobe.com/browse/MWPW-158749
2. The request for the manifest files may need to be made earlier (this depends on how long it takes to load in prod with the prior optimizations, since manifest files take less time to laod on prod.). We'll have to wait for prior optimizations to be done to decide the next step.

Resolves: [MWPW-159113](https://jira.corp.adobe.com/browse/MWPW-159113)

**Test URLs:**
- After: https://par-pers--milo--sharmrj.hlx.page/?martech=off

https://main--cc--adobecom.hlx.page/products/photoshop?milolibs=par-pers--milo--sharmrj


libs/features/personalization/personalization.js
```diff
@@ -1058,14 +1075,30 @@ export async function init(enablements = {}) {
...
-      if (manifest.disabled) return;
-      const normalizedURL = normalizePath(manifest.manifestPath);
-      loadLink(normalizedURL, { as: 'fetch', crossorigin: 'anonymous', rel: 'preload' });
-    });
+
+    manifestPromises = manifestPromises
+      .concat(await combineMepSources(pzn, promo, promoUtilsPromise, mepParam));
   }
 
+  manifests = await Promise.all((await Promise.all(manifestPromises))
```
mokimo, line 1083: Is the double `Promise.all` there for any specific reason, or is that an omission?

```js
// Untested demo code
manifests = await Promise.all(
  manifestPromises.map(async (promise) => {
    try {
      const resp = await promise;
      if (!resp.ok) {
        if (resp.status === 404) {
          throw new Error('File not found');
        }
        throw new Error(`Invalid response: ${resp.status} ${resp.statusText}`);
      }
      const manifestData = await resp.json();
      return { manifestData, manifestPath: resp.url };
    } catch (e) {
      console.log(`Error loading content: ${resp.url}`, e.message || e);
      return null;
    }
  })
).then(results => results.filter(Boolean));
```

mokimo (commented):
Very cool to see this being tackled, with MEP/Personalization we should have room for improvements in terms of performance.

sharmrj:
@mokimo pointed out an issue where we're waiting a long time in between the interact call and loading the blocks.

Turns out, if launch finishes loading before the ims token network request completes, the page stalls for 4 extra seconds while we wait for the alloy event (even though the interact call goes through just fine). I've confirmed this by making measurements of `getTargetPersonalization` like so:
```javascript
 performance.mark('targetstart');
    const { targetManifests, targetPropositions } = await getTargetPersonalization();
    performance.mark('targetend');
    console.log(performance.measure('target', 'targetstart', 'targetend'));
```
If launch loads before the ims token comes back we always get a 4second timing on this function: 
```
PerformanceMeasure {detail: null, name: 'target', entryType: 'measure', startTime: 5821.900000095367, duration: 4000.899999976158}
```
Note that the timeout duration in martech.js is exactly 4000ms.
If the ims token comes back before launch we get something more along the lines of:
```
PerformanceMeasure {detail: null, name: 'target', entryType: 'measure', startTime: 8346.5, duration: 596.2999999523163}
```
This is possibly why some Nala tests are timing out.

AFAICT this sometimes happens out in the wild, it's just that this optimization makes it more likely to happen. I'll continue to investigate the root cause.

vgoodric (changes requested):
Please let me know when this is finalized so we can thoroughly QA it.  SOT approval often does not check for MEP.

libs/features/personalization/personalization.js
```diff
@@ -1003,32 +1005,40 @@ export async function applyPers(manifests, postLCP = false) {
...
       .map((manifestPath) => ({ manifestPath }));
+    initialPersManifestsPromises = initialPersManifests.map(({ manifestPath }) => {
+      const normalizedURL = normalizePath(manifestPath);
+      return fetch(normalizedURL, { mode: 'same-origin' });
+    });
   }
 
   if (promoEnabled) {
-    const { default: getPromoManifests } = await import('./promo-utils.js');
+    const { default: getPromoManifests } = await promoUtilsPromise;
```
3ch023, line 1025: we could also move code from promo-utils.js directly here, i think it was not a good decision in the first place to have a separate file.
promo-utils.js contents seem small enough

libs/features/personalization/personalization.js
```diff
@@ -1037,15 +1047,22 @@ export const combineMepSources = async (persEnabled, promoEnabled, mepParam) =>
       }
     });
   }
-  return persManifests;
+  return initialPersManifestsPromises.concat(persManifests
+    .filter((m) => !('disabled' in m) || !m.disabled)
```
3ch023, line 1051: if i remember correct, we shouldn't filter out disabled manifests.
those are 'promotions' manifests that are schedule is in the future/past, but Content Authors need to see them in the Page mep preview interface. However we don't load the .json itself (at least this is how it used to work) unless the manifest is active.

if you are still able to see a disabled manifest in the page then we are good

sharmrj:
For whatever reason I'm seeing the timeout issue again after merging the stage branch into my branch and fixing the merge conflicts. It seems IMS loaded after launch and that's what triggered the issue again:
image
image
I'll attach the trace here: 
[Trace-20241107T131821.json.zip](https://github.com/user-attachments/files/17657964/Trace-20241107T131821.json.zip)

libs/features/personalization/personalization.js
```diff
@@ -1071,32 +1071,40 @@ export async function applyPers(manifests) {
...
 
   if (persEnabled) {
-    persManifests = persEnabled.toLowerCase()
+    initialPersManifests = persEnabled.toLowerCase()
       .split(/,|(\s+)|(\\n)/g)
       .filter((path) => path?.trim())
       .map((manifestPath) => ({ manifestPath }));
+    initialPersManifestsPromises = initialPersManifests.map(({ manifestPath }) => {
+      const normalizedURL = normalizePath(manifestPath);
+      return fetch(normalizedURL, { mode: 'same-origin' });
```
chrischrischris, line 1086: nit: inline `normalizePath`

libs/features/personalization/personalization.js
```diff
@@ -1145,14 +1160,29 @@ export async function init(enablements = {}) {
...
+          }
+          throw new Error(`Invalid response: ${resp.status} ${resp.statusText}`);
+        }
+        const manifestData = await resp.json();
+        return { manifestData, manifestPath: resp.url };
+      } catch (e) {
+        /* c8 ignore next 3 */
+        console.log(`Error loading content: ${resp.url}`, e.message || e);
+      }
+      if (pzn) loadLink(getXLGListURL(config), { as: 'fetch', crossorigin: 'anonymous', rel: 'preload' });
```
chrischrischris, line 1181: Doesn't this line still need to run even when the manifest loading is successful?

sharmrj:
> For whatever reason I'm seeing the timeout issue again after merging the stage branch into my branch and fixing the merge conflicts. I was also unable to reproduce the issue earlier which is why I thought it had been resolved. It seems IMS loaded after launch and that's what triggered the issue again: image image
> 
> I'll attach the trace here: [Trace-20241107T131821.json.zip](https://github.com/user-attachments/files/17657964/Trace-20241107T131821.json.zip)

Turns out the same issue exists in helpx as well. In this case the timeout is 3 seconds instead of four, and it causes the global nav to load late.

Screenshot 2024-11-22 at 12 58 17 PM


It's easier to see in the trace:  [Trace-20241122T125841.json.zip](https://github.com/user-attachments/files/17866514/Trace-20241122T125841.json.zip)

The relevant gap is from `feds.js` to `helpx.adobe.com.html`

It's possible that the cause is that main.min.js loads before ims (rather than launch).

PR #2971, opened 2024-09-27
[MWPW-159181] - links conversion should handle extensions

robert-bogos:
## Description
There have been multiple requests for the link conversion feature to handle extensions.
The extensions will be removed by default for `.page` or `.live` urls.

## Related Issue
Resolves: [MWPW-159181](https://jira.corp.adobe.com/browse/MWPW-159181)

## Testing instructions
Using the **[test after URL](https://mwpw-159181-html-extension--milo--robert-bogos.hlx.page/drafts/rbogos/env-relative-urls)**, check if the `link authored to send to https://www.adobe.com/acrobat/free-trial-download.html` link is converted to point to `https://main--dc--adobecom.hlx.page/acrobat/free-trial-download` which has the `.html` extension removed.

## Test URLs
**Acrobat:**
- Before: https://www.stage.adobe.com/acrobat.html
- After: https://www.stage.adobe.com/acrobat.html?milolibs=mwpw-159181-html-extension--milo--robert-bogos

**BACOM:**
- Before: https://main--bacom--adobecom.hlx.page/products/real-time-customer-data-platform/rtcdp
- After: https://main--bacom--adobecom.hlx.page/products/real-time-customer-data-platform/rtcdp?milolibs=mwpw-159181-html-extension--milo--robert-bogos

**CC:**
- Before: https://main--cc--adobecom.hlx.page/drafts/rbogos/default?martech=off
- After: https://main--cc--adobecom.hlx.page/drafts/rbogos/default?martech=off&milolibs=mwpw-159181-html-extension--milo--robert-bogos

**Homepage:**
- After: https://main--homepage--adobecom.hlx.page/homepage/index-loggedout?milolibs=mwpw-159181-html-extension--milo--robert-bogos

**Blog:**
- After: https://main--blog--adobecom.hlx.page/?milolibs=mwpw-159181-html-extension--milo--robert-bogos

**Milo:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rbogos/env-relative-urls?martech=off
- After: https://mwpw-159181-html-extension--milo--robert-bogos.hlx.page/drafts/rbogos/env-relative-urls?martech=off


libs/scripts/scripts.js
```diff
@@ -25,6 +25,7 @@ const stageDomainsMap = {
   'www.stage.adobe.com': {
     'www.adobe.com': 'origin',
     'helpx.adobe.com': 'helpx.stage.adobe.com',
+    'business.adobe.com': { to: 'business.stage.adobe.com', removeExt: true },
```
mokimo, line 28: Shouldn't the extension always be removed on stage.adobe.com? Or are there special cases

robert-bogos: I'm not aware of any special cases, but I'm trying to keep the solution consumer agnostic in case there are

mokimo: Should we do the reverse in this case, so that this will work without changes to the configs?

`addHtmlExt` if we need it anywhere? Having the extension should be the special case

Brandon32: For BACOM we have the .html extension but for BACOM Blog we do not.

- https://business.stage.adobe.com/resources/main.html ->
  - https://main--bacom--adobecom.hlx.page/resources/main
- https://business.stage.adobe.com/blog/tags/the-latest ->
  - https://main--bacom-blog--adobecom.hlx.page/blog/tags/the-latest
  
We use `htmlExclude` to prevent appending .html to the blog pages. 
https://github.com/adobecom/bacom/blob/stage/scripts/scripts.js#L124-L126

Perhaps the `useDotHtml` config and `appendHtmlToLink` would fit these use cases?

robert-bogos: @mokimo I initially considered whether this should be an opt-in or opt-out type of config setting. The current approach felt more agnostic, but for our consumers, I see the value of enabling this by default, with the option to opt-out through the config. I’ve made the change 👍.

robert-bogos: @Brandon32 Another enhancement will come to also consider paths when determining which conversion to make. That will fix the problems with use-cases like `business.state.adobe.com` <-> `business.state.adobe.com/blog `

libs/utils/utils.js
```diff
@@ -650,8 +650,15 @@ export function convertStageLinks({ anchors, config, hostname }) {
       .find((domain) => a.href.includes(domain));
     if (!matchedDomain) return;
     a.href = a.href.replace(a.hostname, domainsMap[matchedDomain] === 'origin'
-      ? hostname
-      : domainsMap[matchedDomain]);
+      ? hostname.to || hostname
+      : domainsMap[matchedDomain].to || domainsMap[matchedDomain]);
```
narcis-radu, line 654: We should use a constant for `domainsMap[matchedDomain]` and also add the optional chaining operator to make sure things don't break.

robert-bogos: Added a constant for `domainsMap[matchedDomain]` 👍
The optional chaining operator is not necessary as we do an early return in [here](https://github.com/adobecom/milo/pull/2971/files#diff-28218f239d8d9c4d41e1924854440b50964619e56058fdd0ae243ff5ff6c616dR651) if `matchedDomain` does not exist as a key of `domainsMap`.

mokimo:
@robert-bogos it seems that the nala tests are failing

libs/utils/utils.js
```diff
@@ -649,9 +649,17 @@ export function convertStageLinks({ anchors, config, hostname }) {
...
+    a.href = a.href.replace(a.hostname, convTargetDomain === 'origin'
+      ? hostname.to || hostname
+      : convTargetDomain.to || convTargetDomain);
+    if (convTargetDomain.useExt) return;
+    const urlObject = new URL(a.href);
+    const path = urlObject.pathname;
+    const extension = path.split('.').pop();
+    if (!extension) return;
+    urlObject.pathname = path.replace(`.${extension}`, '');
+    a.href = urlObject.toString();
```
overmyheadandbody, line 662: This looks like added complexity that overlaps with `appendHtmlToLink`. Could we not convert stage links first and then let the logic run as before? That might also mean that we don't need to update the object with the domain maps on the consumer side.

Or are you trying to tackle the use-case where a link has been already authored with `.html`? In that case, wouldn't checking whether the page is a .page/.live one suffice as a check?

robert-bogos: I'm trying to tackle the use-case where a link has been already authored with `.html`.
I like the .page/.live check approach, this will also remove the need for `stageDomainsMap` updates.
I will steal that 😉

mokimo: Here's some more context to links being authored with [.html](https://adobe-mwp.slack.com/archives/C03B0BUA75E/p1727826864607949) on slack

test/utils/utils.test.js
```diff
@@ -537,15 +539,16 @@ describe('Utils', () => {
         });
 
         anchors.forEach((a, index) => {
-          const expectedDomain = Object.values(domainsMap)[index];
+          const expectedDomain = Object.values(domainsMap)[index]?.to
```
mokimo, line 542: Is the `.to` value is relevant given the changes?

robert-bogos: No, it got left behind. Thank you for pointing out 👍

narcis-radu (approved):
@robert-bogos - Please add test URLs for all consumers, e.g. https://main--cc--adobecom.hlx.page/creativecloud?milolibs=mwpw-159181-html-extension--milo--robert-bogos

overmyheadandbody (approved):
Thanks for integrating the changes!

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-159181

PR #2976, opened 2024-09-27
MWPW-146211 [MILO][MEP] Option to select all elements

vgoodric:
* Fragment selectors automatically work on all instances and nested instances for replace and remove, but not for insertBefore and insertAfter.  Makes behavior consistent by making the action work for the "insert" actions.
* For CSS/Simplified selectors, add the "#_all" flag so that it can act on all found instances
* For CSS/Simplified selectors, add the "#_include-fragments so that fragments are checked as they load to see if the selectors can be found within. Behavior added in 2 places: fragments and fragments using #_inline
Also
* changed the attribute holding the experience column value from being called `target` to being called `content`.  Target has such a specific meaning for MEP that it was confusing people
* changed several function calls to just be passed the entire command object versus specific elements.  This makes it less prone to error and updates in the future if we update/add to the object.

Resolves: [MWPW-146211](https://jira.corp.adobe.com/browse/MWPW-146211)
Also Resolves: [MWPW-146210](https://jira.corp.adobe.com/browse/MWPW-146210)
Features were very intertwined.  So just combined into 1 branch.

**Psi-check**
https://mepall--milo--adobecom.hlx.page/?martech=off

**QA instructions:**
This page's manifest uses fragment selectors: https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q3/mepall/fragment-selectors/?milolibs=mepall
The manifest removes, replaces, and inserts after nested fragments.   Cards are on the page twice and should result in this for both:
image

There are 3 other pages all using the same manifest and only use CSS/Simplified selectors.  The manifest has 4 columns:
1. no flags
2. all flag
3. include flag
4. both flags
And all 4 columns perform the same 4 actions with the same 4 selectors.  The only difference is which flags are used.

The 3 pages have the same content, but differ in how they render the content.
Page 1 (content is directly on the page): https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q3/mepall/css-selectors/asides?milolibs=mepall&mepHighlight=true
Expected results:
1. no flags - all 4 changes only on first block
2. all flag - all 4 changes twice on both blocks
3. include flag - all 4 changes only on first block
4. both flags - all 4 changes twice on both blocks

Page 2 (content is nested in a fragment): https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q3/mepall/css-selectors/nested-content?milolibs=mepall&mepHighlight=true
Expected results:
1. no flags - no changes
2. all flag - no changes
3. include flag - all 4 changes only on first block
4. both flags - all 4 changes twice on both blocks

Page 3 (content is nested in inline fragments): https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q3/mepall/css-selectors/inline-content?milolibs=mepall&mepHighlight=true
Expected results:
1. no flags - no changes
2. all flag - no changes
3. include flag - all 4 changes only on first block
4. both flags - all 4 changes twice on both blocks


ivanvatadobe (approved):

AdobeLinhart (approved):

jpratt2 (approved):
I have done a QA Review and I approve this PR to be merged to stage.
The record of the QA process is attached to https://jira.corp.adobe.com/browse/MWPW-159646

PR #2977, opened 2024-09-27
MWPW-158886: update the way we send arbitrary fields to caas

sheridansunier:
Arbitrary fields are not exposed properly in Milo.  We have an established standard for how arbitrary fields are sent into Caas from Dexter, and to avoid data quality issues we need the way arbitrary fields are sent from Milo to match that. 

Resolves: [MWPW-158886](https://jira.corp.adobe.com/browse/MWPW-158886)

How our arbitrary fields are meant to look/how they look when they are sent from dexter.
Screen Shot 2024-09-27 at 10 53 50 AM

How they were coming from Milo:
opensearch_milo_card_stage

This change standardizes the way our data is sent into caas :) 

Validated with local overrides:
Screen Shot 2024-09-27 at 2 21 07 PM

**Test URLs:**
- After: https://MWPW-158886--milo--sheridansunier.hlx.page/?martech=off


test/blocks/caas/mocks/body.html
```diff
@@ -56,6 +56,10 @@ <h2 id="caas-card-metadata-test">CaaS Card Metadata Test</h2>
           <div data-valign="middle">PlayURL</div>
           <div data-valign="middle"><a href="https://www.youtube.com/watch?v=9H5S4Cc5SgM">https://www.youtube.com/watch?v=9H5S4Cc5SgM</a></div>
         </div>
+        <div>
+          <div data-valign="middle"><strong>Arbitrary</strong></div>
+          <div data-valign="middle">mpcVideoId:mpcVideoIdGM,thumbnailUrl: thumbnailUrlGM,sessionId: 1640105575257001KRjv, videoDuration: 01:00:0</div>
```
sanrai, line 61: We may want to consider adding some form of validation before accepting the authored row if possible

It seems like lots of authored cases could cause issues when parsing and break either on our side or our BE if we send invalid data over. 

A few examples:

Case sensitivity:
- "KEY: value, key: VALUE"

Fields copied/pasted from somewhere else that give escape chars:
- "key1: value\nwith\nnewlines, key2: value\twith\ttabs"

Malformed key-value pairs:
- Missing colon: "key1 value1, key2: value2"
- Multiple colons: "key1: value1: extra, key2: value2"
- Empty key or value: ": value, key:, :, key: value"

And the list goes on and on..

At the very least, we should at least have some unit tests with these cases as we know our code can safely handle it.

sanrai (changes requested):

test/blocks/caas/mocks/body.html
sheridansunier: our validation is done using a regular expression in this file (search "isKeyValPair" on the file to find it) - I added unit tests to cover this behavior and give us peace of mind that we are correctly handling this input :)

sanrai (approved):

PR #2981, opened 2024-09-30
[MWPW-159265] AI callout icon alignment fix in rtl and mobile

rohitsahu:
AI callout icon margin from the text in rtl and on mobile was not proper. Fixed by using inline-start and inline-end instead of margin left and right.

Resolves: [MWPW-159265](https://jira.corp.adobe.com/browse/MWPW-159265)

**Test URLs:**
Before: https://main--milo--adobecom.hlx.page/ae_ar/drafts/mili/kitchen-sink/merch-card
After: https://mwpw-159265_2--milo--rohitsahu.hlx.page/ae_ar/drafts/mili/kitchen-sink/merch-card


npeltier (approved):
lgtm

PR #2985, opened 2024-09-30
MWPW-138989 [Perf] Preload Milo Blocks if MEP enabled

markpadbe:
When using mep with Target, an await happens that prevents any block-loading from happening until Target makes a decision. This PR preloads LCP blocks (any block that appears in section 1 on the base page) by importing the existing preloadBlockResources from the utils.js file. The PR also adds an additional parameter to preloadBlockResources to allow any block that appears in section 1 to be preloaded. 

Resolves: [MWPW-138989](https://jira.corp.adobe.com/browse/MWPW-138989)

Steps to QA:
For the before and after links, set a breakpoint around line 1076 of personalization.js (after preloadBlockResources is called).

Screenshot 2024-09-30 at 2 02 41 PM

Then open the Network tab, select either js or css and. For the After url, at this breakpoint, you should see the relevant marquee, aside and text files have loaded (along with the decorate.js file), and for the Before Url, the files should not be loaded at this point.

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q4/preloadblocks/testpage?martech=off
- After: https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q4/preloadblocks/testpage?milolibs=preloadblocks&martech=off

PSI-check: https://preloadblocks--milo--adobecom.hlx.page/?martech=off


libs/utils/utils.js
```diff
@@ -1225,10 +1225,10 @@ export function partition(arr, fn) {
   );
 }
 
-const preloadBlockResources = (blocks = []) => blocks.map((block) => {
+export const preloadBlockResources = (blocks = [], marqueesOnly = true) => blocks.map((block) => {
   if (block.classList.contains('hide-block')) return null;
   const { blockPath, hasStyles, name } = getBlockData(block);
-  if (['marquee', 'hero-marquee'].includes(name)) {
+  if (['marquee', 'hero-marquee'].includes(name) || !marqueesOnly) {
```
mokimo, line 1231: Why would you modify this function? You can't be sure the blocks you preload from within the MEP scripts really use decorate. 

The idea is to have the small tradeoff of 2-3 extra lines for a benefit in most cases (marquee/hero-marquee) which are very common LCP elements.

libs/features/personalization/personalization.js
```diff
@@ -1066,6 +1066,12 @@ export async function init(enablements = {}) {
     });
   }
 
+  if (target === true) {
+    const { preloadBlockResources } = await import('../../utils/utils.js');
+    const lcpBlocks = [...document.querySelectorAll('body > main > div:first-child > *[class]')];
+    preloadBlockResources(lcpBlocks, false);
```
mokimo, line 1072: Can you really know which blocks are relevant before target loads? What if the markup is changed and the blocks you are loading are throw-away work, because target changes the site?

chrischrischris:
@markpadbe Do you have any data on the how much this improves loading?  Ideally tested on homepage, some product pages (photoshop/illustrator) and a bacom site.

libs/utils/utils.js
```diff
@@ -1230,11 +1230,12 @@ export function partition(arr, fn) {
...
 }
 
-const preloadBlockResources = (blocks = []) => blocks.map((block) => {
+export const preloadBlockResources = (blocks = []) => blocks.map((block) => {
   if (block.classList.contains('hide-block')) return null;
   const { blockPath, hasStyles, name } = getBlockData(block);
-  if (['marquee', 'hero-marquee'].includes(name)) {
+  if (['marquee', 'hero-marquee', 'homepage-brick'].includes(name)) {
     loadLink(`${getConfig().base}/utils/decorate.js`, { rel: 'preload', as: 'script', crossorigin: 'anonymous' });
+    if (name === 'homepage-brick') loadLink(`${getConfig().base}/martech/attributes.js`, { rel: 'preload', as: 'script', crossorigin: 'anonymous' });
```
mokimo, line 1238: Is the homepgae-brick part of a test? I'd try not to add new code for perf reasons while a test is running. We should keep utils slim and optimize for the >90% cases

markpadbe:
> @markpadbe Do you have any data on the how much this improves loading? Ideally tested on homepage, some product pages (photoshop/illustrator) and a bacom site.

PR #2986, opened 2024-09-30
MWPW-140452 - Icon authoring in milo using the federal repo and individual SVG assets

ryanmparrish:
To streamline icon management across Milo and other consuming sites, a centralized repository, `federal`, along with a directory named `icons/svgs`, will be established. This will provide shared access to a unified set of icons. The transition will maintain the current authoring experience, while also introducing new opportunities for contributors to expand the icon set. The key points of this change are outlined below:

Key Points:

**Centralized Icon Repository:** A new [/assets/icons/svgs](https://adobe.sharepoint.com/sites/adobecom/Shared%20Documents/Forms/AllItems.aspx?id=%2Fsites%2Fadobecom%2FShared%20Documents%2Ffederal%2Ffederal%2Fassets%2Ficons%2Fsvgs&viewid=d776cf70%2D9b7e%2D4ab7%2Db9da%2D9e0f8e03a7d2) directory was added to the [/federal/](https://adobe.sharepoint.com/sites/adobecom/Shared%20Documents/Forms/AllItems.aspx?id=%2Fsites%2Fadobecom%2FShared%20Documents%2Ffederal&viewid=d776cf70%2D9b7e%2D4ab7%2Db9da%2D9e0f8e03a7d2) repo, allowing multiple sites (Milo and others) to use the same set of icons. All icons currently available are listed here [/icons/icons.json](https://main--federal--adobecom.hlx.page/federal/assets/icons/icons.json)

**Consistent Authoring Notation:** Authors will continue to use the current notation, such as `:icon-play:`, to insert icons into content. This ensures a seamless transition with no change in authoring experience.

**Source Change:** Icons will no longer be served from the Milo code-bus. Instead, they will be fetched from the federal repository via the content-bus.

**Author Contributions:** This new system enables authors to contribute and expand the icon set by adding new icons to the centralized repository, a feature that was previously unavailable.

**Subsequent Ticket:** The Sidekick plugin library will be updated in a subsequent ticket to improve authoring accessibility for that component

Resolves: [MWPW-140452](https://jira.corp.adobe.com/browse/MWPW-140452)

**Draft URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rparrish/icon/icons-federal?martech=off
- After: https://rparrish-fed-icons--milo--adobecom.hlx.page/drafts/rparrish/icon/icons-federal?martech=off

**Doc URLs:**
- Before: https://main--milo--adobecom.hlx.page/docs/library/kitchen-sink/?martech=off
- After: https://rparrish-fed-icons--milo--adobecom.hlx.page/docs/library/kitchen-sink/?martech=off

All consuming sites should be tested for this change. `?milolibs=rparrish-fed-icons`
https://main--bacom--adobecom.hlx.page/?milolibs=rparrish-fed-icons


Anupamasonnalagi:
@ryanmparrish Verified in CC and noticed one observation that appending the domain in CC below icons are not showing up.
ai-generate, lock, open-in and remove icons

Before: https://main--cc--adobecom.hlx.page/drafts/anupama/pr/icons-federal
image
image

After: https://main--cc--adobecom.hlx.page/drafts/anupama/pr/icons-federal?milolibs=rparrish-fed-icons
image
image

libs/utils/utils.js
```diff
@@ -764,8 +764,9 @@ async function decorateIcons(area, config) {
   const { base } = config;
```
mokimo, line 764: From a performance standpoint, I'd suggest we do a very small check if there are any federal icons to begin with and start preloading the federal script:

```js
if([...icons].some(icon => icon.href.includes('/federal/')) loadLink(`...federal`, { rel: 'preload', as: 'fetch', crossorigin: 'anonymous' });
```

libs/utils/utils.js
```diff
@@ -764,8 +764,9 @@ async function decorateIcons(area, config) {
...
   loadStyle(`${base}/features/icons/icons.css`);
   loadLink(`${base}/img/icons/icons.svg`, { rel: 'preload', as: 'fetch', crossorigin: 'anonymous' });
```
mokimo, line 766: Does this preload need to change, do we always load from a federal icon sheet now? If we know the icons are _always_ coming from there, we should change this

libs/features/icons/icons.js
```diff
@@ -49,26 +52,56 @@ function decorateToolTip(icon) {
...
+  icons.forEach((icon) => {
+    icon.classList.add('milo-icon');
+    const iconName = [...icon.classList].find((c) => c.startsWith('icon-'))?.substring(5);
+    if (icon.dataset.svgInjected || !iconName) return;
+    if (iconName === 'tooltip') {
+      decorateToolTip(icon);
+      return;
+    }
+    if (!federalIcons[iconName] && !iconsToFetch.has(iconName)) {
+      const url = `${fedRoot}/federal/assets/icons/svgs/${iconName}.svg`;
```
mokimo, line 81: Does this mean, we switch from having one general purpose file
`const iconSVGs = await fetchIcons(config);` - to one icon per file now?

ryanmparrish: @mokimo - That is correct. This way we have full author ability to add more icons w out a code change. I have this explained in the description, but please let me know if this is not clear.

libs/utils/utils.js
ryanmparrish: Since these are all individual icons now, I think we actually should remove this line since it's obsolete. We could pre-load all svg icons, but would have to do some type of loop or query to target them all. LKM if you have any thoughts. Thanks

mokimo: Unless we have an _easy_ way to know which icons to load it's probably best to just remove

ryanmparrish:
@Anupamasonnalagi  - It looks like there were a few recently added icons to the sprite sheet that weren't accounted for. I've added and published these in the [/svgs/ ](https://adobe.sharepoint.com/sites/adobecom/Shared%20Documents/Forms/AllItems.aspx?id=%2Fsites%2Fadobecom%2FShared%20Documents%2Ffederal%2Ffederal%2Fassets%2Ficons%2Fsvgs&viewid=d776cf70%2D9b7e%2D4ab7%2Db9da%2D9e0f8e03a7d2)directory and they are showing up now. Thanks for flagging that.

libs/utils/utils.js
ryanmparrish: I actually think we just load It since we need it on any span.icon and don't have the href yet anyhow. 
...I made an update to preload federated.js instead of the non used icons.svg on util decorateIcon()

```
 loadStyle(`${base}/features/icons/icons.css`);
 loadLink(`${base}/utils/federated.js`, { rel: 'preload', as: 'fetch', crossorigin: 'anonymous' });
```
LMK if see any issue w/ this. Thanks

Sartxi (approved):
Just a few nit/suggestion item

libs/utils/utils.js
mokimo: Nope that's good 👍

SilviuLCF:
@ryanmparrish page https://rparrish-fed-icons--milo--adobecom.hlx.page/drafts/sghimpos/icon/icons-federal 
have CLS : 0.04 Poor , can you take a look ?

libs/utils/utils.js
mokimo: It seems to load the file twice now though 🤔 
Screenshot 2024-10-02 at 11 07 23
Screenshot 2024-10-02 at 11 07 15

libs/features/icons/icons.js
mokimo: Nono that's fine, I'm looking at this from a performance lense... So far I've seen that requests add overhead and usually lead to degraded performance. We now potentially do *a lot* of requests.

Screenshot 2024-10-02 at 11 09 30

I'm a little concerned about our mobile users, maybe we need to go back to reserving space and loading icons async? That's something you've been working on previously.

mokimo:
Could you create a test page with a marquee with say 3 icons after, and the sheet before? I'd love to run a few performance tests on that

libs/utils/utils.js
```diff
@@ -763,9 +763,10 @@ async function decorateIcons(area, config) {
   if (icons.length === 0) return;
   const { base } = config;
   loadStyle(`${base}/features/icons/icons.css`);
-  loadLink(`${base}/img/icons/icons.svg`, { rel: 'preload', as: 'fetch', crossorigin: 'anonymous' });
-  const { default: loadIcons } = await import('../features/icons/icons.js');
-  await loadIcons(icons, config);
+  loadLink(`${base}/utils/federated.js`, { rel: 'preload', as: 'fetch', crossorigin: 'anonymous' });
```
mokimo, line 766: This currently has a different mime/type leading to two requests.

image

```suggestion
  loadLink(`${base}/utils/federated.js`, { rel: 'preload', as: 'script', crossorigin: 'anonymous' });
```

ryanmparrish: Ahh. I missed that part. I've updated this and it only gets called 1 time now. Thanks

libs/blocks/text/text.css
```diff
@@ -100,7 +100,7 @@
   position: relative;
 }
 
-.text-block .icon-list-item .icon.margin-right:not(.margin-left) { /* target first node only */
+.text-block .icon-list-item .icon.node-index-first:not(.node-index-last) { /* target first node only */
```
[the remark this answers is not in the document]
ryanmparrish: Yes, I am trying to support previous use case where we used margin-right, margin-left. etc and now this is handled a bit different, I don't think the not() part is necessary since there will never be both. I'm going to update and test.

test/features/icons/icons.test.js
```diff
@@ -2,21 +2,45 @@ import { readFile } from '@web/test-runner-commands';
...
+const svgEx = `<?xml version="1.0" encoding="UTF-8"?>
+<svg id="arrow-down" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 18 18">
+  <path fill="currentcolor" d="M12,10V1.5a.5.5,0,0,0-.5-.5h-5a.5.5,0,0,0-.5.5V10H2.5035a.25.25,0,0,0-.177.427L9,17.1l6.673-6.673A.25.25,0,0,0,15.4965,10Z"></path>
+</svg>`;
 
 describe('Icon Suppprt', () => {
   let paramsGetStub;
 
+  beforeEach(() => {
+    window.fetch = stub().callsFake(() => mockRes({}));
```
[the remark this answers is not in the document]
ryanmparrish: Good call. Thanks

test/features/icons/icons.test.js
```diff
@@ -2,21 +2,45 @@ import { readFile } from '@web/test-runner-commands';
...
+  });
+
+  afterEach(() => {
+    // Do not build up any test state - reset window.fetch to it's original state
+    window.fetch = originalFetch;
```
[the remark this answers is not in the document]
ryanmparrish: Eye eye capt. 👍 Thanks

libs/blocks/text/text.css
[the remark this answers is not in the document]
ryanmparrish: FYI, I've simplified this and not using :not

ryanmparrish:
@mokimo  - I have made a few extra draft pages to test performance. I wasn't totally sure what you meant by 'sheet', I assume an example w/ all the icons? Anyhow LMK if this is helpful or if I mis-interpreted your ask. Thanks

Marquee below icons: https://rparrish-fed-icons--milo--adobecom.hlx.page/drafts/rparrish/icon/icon-performance
Marquee above icons: https://rparrish-fed-icons--milo--adobecom.hlx.page/drafts/rparrish/icon/icon-performance-2

libs/features/icons/icons.js
```diff
@@ -49,26 +52,56 @@ function decorateToolTip(icon) {
   wrapper.parentElement.replaceChild(icon, wrapper);
 }
 
-export default async function loadIcons(icons, config) {
-  const iconSVGs = await fetchIcons(config);
-  if (!iconSVGs) return;
+export async function setNodeIndexClass(icons) {
```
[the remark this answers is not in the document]
ryanmparrish: I like it. Thanks

Anupamasonnalagi:
@ryanmparrish After appending the Domain on CC product pages noticed that icons are not showing up. 
Before: https://main--cc--adobecom.hlx.page/products/illustrator/plans
image
https://main--cc--adobecom.hlx.page/products/illustrator/plans?milolibs=rparrish-fed-icons
image

ryanmparrish:
@Anupamasonnalagi  - Thanks again for the testing and finding these edge cases. I found a race condition w/ the way these tooltip icons were being loaded. I have fixed this and verified your page renders those icons as expected now. Please let me know if you see any further issues.

spadmasa:
Hi @ryanmparrish  , issue is seen in results page of UAR as with milo libs appended , Please take a look on this as well
before :
https://main--cc--adobecom.hlx.live/cc-shared/fragments/uar/marquee/illustrator?debug=quiz-results
{68ED928E-DD83-4774-9367-5B5E31B3F6AE}

After https://main--cc--adobecom.hlx.live/cc-shared/fragments/uar/marquee/illustrator?debug=quiz-results&milolibs=rparrish-fed-icons
{F560AAD2-6B5A-48E9-96BC-5B1987C69DF8}

CC: @sigadamvenkata  , @salonijain3

ryanmparrish:
Hi @spadmasa. Thanks for the find on that. This is caused by the .live url as none of these have been published yet and only previewed. I have published that one icon and will be publishing the rest momentarily. I'm am seeing one issue w/ the bulk publish where they will preview, but not publish together. Looking into it. Thanks

meganthecoder:
Hi @ryanmparrish , In a small scale test, I measured an average 158ms increase in LCP on the bacom homepage with your code https://main--bacom--adobecom.hlx.live/?milolibs=rparrish-fed-icons . Likely due to the svg network request being made later than before (and blocking). Any thoughts here @JasonHowellSlavin ?

Screenshot 2024-10-04 at 3 30 09 PM

mokimo:
We also don't only load the icons relevant for a section (LCP), but ALL ICONS on the page in one go.
Screenshot 2024-10-07 at 10 32 52

> Marquee below icons: https://rparrish-fed-icons--milo--adobecom.hlx.page/drafts/rparrish/icon/icon-performance

libs/utils/utils.js
```diff
@@ -763,9 +763,9 @@ async function decorateIcons(area, config) {
   if (icons.length === 0) return;
```
overmyheadandbody, line 763: Should icon loading be moved from area-level to section or block-level to avoid loading icons that might be below the fold? Is that an option?

test/features/icons/icons.test.js
```diff
@@ -1,45 +1,51 @@
...
-    paramsGetStub = stub(URLSearchParams.prototype, 'get');
-    paramsGetStub.withArgs('cache').returns('off');
+  beforeEach(() => {
+    stub(window, 'fetch').callsFake(() => mockRes({}));
   });
 
-  after(() => {
-    paramsGetStub.restore();
+  afterEach(() => {
+    window.fetch.restore();
```
overmyheadandbody, line 30: You could even do `sinon.restore();` which should handle any stubs without having to explicitly name them.

libs/features/icons/icons.js
```diff
@@ -36,39 +39,74 @@ export const fetchIcons = (config) => new Promise(async (resolve) => {
...
-      } else {
-        icon.classList.add('margin-inline-start', 'margin-inline-end');
-      }
+    if (parent && parent.parentElement.tagName === 'LI') parent.parentElement.classList.add('icon-list-item');
+  });
+
+  await Promise.all(iconRequests);
+
+  icons.forEach((icon) => {
+    const iconName = [...icon.classList].find((c) => c.startsWith('icon-'))?.substring(5);
```
overmyheadandbody, line 105: This is used on L73 already, might be worth turning into its own method.

libs/features/icons/icons.js
```diff
@@ -36,39 +39,74 @@ export const fetchIcons = (config) => new Promise(async (resolve) => {
...
+    if (iconName && federalIcons[iconName] && !icon.dataset.svgInjected) {
+      const svgClone = federalIcons[iconName].cloneNode(true);
```
overmyheadandbody, line 107: Why do we need to clone the SVG yet again, couldn't we reference it directly at this point?

overmyheadandbody (commented):
I share the concerns for performance impact on this. It might be that we need to explore loading icons only when needed or come up with some alternative strategies to avoid the negative performance impact this might introduce.

I think the concept is great and expands our icon strategy, but we need to make sure it doesn't impact other areas.

libs/utils/utils.js
```diff
@@ -763,9 +763,9 @@ async function decorateIcons(area, config) {
...
   const { base } = config;
   loadStyle(`${base}/features/icons/icons.css`);
-  loadLink(`${base}/img/icons/icons.svg`, { rel: 'preload', as: 'fetch', crossorigin: 'anonymous' });
+  loadLink(`${base}/utils/federated.js`, { rel: 'preload', as: 'script', crossorigin: 'anonymous' });
   const { default: loadIcons } = await import('../features/icons/icons.js');
-  await loadIcons(icons, config);
+  await loadIcons(icons);
```
mokimo, line 768: I would have two suggestions two help the performance concerns on this:

- Preload any icon that's used in LCP//the first section (in utils.js)
- Load icons on a section to section basis rather than all within the document at once as @overmyheadandbody suggested as well.

ryanmparrish: @mokimo Thanks for the suggestions. 
I've implemented the first one, preloading icons in 1st section, which does help a hair, I'm still doing some testing, just and update, but looking at the second one. it is loading those per section, Im going to try loading out of view ones postLPC and will let you know. also, Instead of preloading 1st section, should we look at whats in viewport?

libs/utils/utils.js
```diff
@@ -459,6 +459,13 @@ function getBlockData(block) {
   return { blockPath, name, hasStyles };
 }
 
+function getIconData(icon) {
+  const fedRoot = 'https://main--federal--adobecom.hlx.page'; // getFederatedContentRoot()
+  const name = [...icon.classList].find((c) => c.startsWith('icon-'))?.substring(5);
+  const path = `${fedRoot}/federal/assets/icons/svgs/${name}.svg`;
+  return { path, name };
+}
```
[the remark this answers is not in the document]
ryanmparrish: @mokimo thanks for the review and suggestions. I have made an update to refactor some things w/ performance in mind and the results are looking good. No increase in LCP or CLS etc on my branch vs. main etc.  If you could have another look and LMK if you have any other suggestions etc. i'm all ears. Thanks
 
 - Pre loading all in view icon fetch requests 
 - defer `loadIcons` till after sectionLoading to there is no impact on LCP.

libs/features/icons/icons.js
ryanmparrish: I think we need this by I may be wrong.. I tried to ref it directly and duplicate icons on a page don't get decorated w/ out the cloned node

ryanmparrish:
FYI, I am seeing one issue w/ tooltips in tables that i'm looking into A.T.M. Ill update soon as this is resolved

ryanmparrish:
Ok, @SilviuLCF @mokimo @overmyheadandbody. 
I have resolved the issue mentioned w/ the [tables](https://main--cc--adobecom.hlx.page/products/illustrator/plans?milolibs=rparrish-fed-icons) and done some performance testing and looking better. Screenshot 2024-10-09 at 6 50 13 PM
I'm gonna come back to this w/ a fresh set of eyes so LMK if anything stands out but LMK if there are any questions.

Key takeaways. Im pre-fetching all icons that are in the initial browser view and then `loadIcons()` after the section-process. (I tried loading these per section, and had some weird on screen loading order when the view was split w/ sections, so I diverged from that suggestion but open to ideas there if you think theres something too it)

Thanks!

SilviuLCF:
@ryanmparrish changes looks good from my perspective LMK if you need to do any changes, before adding ready for stage label. Thank you

libs/utils/utils.js
```diff
@@ -1253,7 +1242,6 @@ async function processSection(section, config, isDoc) {
   preloadBlockResources(section.preloadLinks);
   await Promise.all([
     decoratePlaceholders(section.el, config),
```
mokimo, line 1244: This still doesn't look right, we seen good effects on LCP scores by parallelizing the loading efforts. You want to structure the code in a way that ideally we load all of those at the same time, that we load the blocks.
- Icon.js
- Icon.css
- federal.js
- Actual icons needed for LCP only

Before: 
Screenshot 2024-10-22 at 10 11 59

After: https://main--cc--adobecom.hlx.live/de/products/illustrator/features?milolibs=rparrish-fed-icons--milo--adobecom
Screenshot 2024-10-22 at 10 11 03

mokimo (changes requested):
I'll request changes as long as we still have the outstanding performance penalty. We need to be restructuring on how we load the icons, otherwise this looks great!

mokimo:
There's also some spacing differences in the marquee icon & it leads a small bug in the georouting modal
image

mokimo:
//cc @saugatmalla

libs/utils/utils.js
saugatmalla: @mokimo I think `icon.js` needs to wait for `federated.js` and then make the network call to fetch the `svg` icons. The older structure was that all the logic was within `icon.js` along with the `svg` icons. Let me know if I am wrong.

mokimo: The idea would be to structure the code a bit like this again:
```
async function decorateIcons(area, config) {
  // see if there is an icon within the LCP section
  const icons = area.querySelectorAll('span.icon');
  if (icons.length === 0) return;

  // if we have an icon, we want to load icon.js+federated+js+icon.css+the icons relevant for the LCP
  const { base } = config;
  loadStyle(`${base}/features/icons/icons.css`);
  loadLink(`${base}/img/icons/federal.js`, { rel: 'preload', as: 'script', crossorigin: 'anonymous' });
  // load or preload the icon rather than having to wait for the icons.js to discover which icon is actually needed
  const { default: loadIcons } = await import('../features/icons/icons.js');
  await loadIcons(icons, config);
}
```

mokimo: We'll want to hint what to load to the browser *before* waiting for the blocking chains, so that the browser can load everything ASAP rather than when discovered sometime one after eachother

libs/utils/utils.js
```diff
@@ -1253,7 +1242,6 @@ async function processSection(section, config, isDoc) {
...
-    decorateIcons(section.el, config),
```
mokimo, line 1256: Here was a great place for this to live in in order to load everything in parallel, for which we've seen good real world benefits on the CRuX LCP data 
image

libs/utils/utils.js
saugatmalla: @mokimo  The current code structure moves all this logic to `icon.js` file. I think, we can move back the loading of files `federated.js` and `icons.css` to load in parallel. But for the actual icon itself, it must wait for the `federated.js` to get the federal url and make the network call, is it not? 
Screenshot 2024-10-24 at 1 15 41 PM

mokimo: @saugatmalla that's a great point that we'd need to load federal to determine which icon to load.

Option 1: Take a small performance and keep icon loading logic within icon.js 
Option 2: Add federal to utils.js and some logic to preload/load the first icon, which should be slightly better from a performance standpoint.

I would tend to say Option-2 is slightly favorable, as we have more and more features relying on this centralized-content feature and thus its justified within `utils` directly. @overmyheadandbody @narcis-radu @robert-bogos any opinions that?

vhargrave: @mokimo here's my take: 
I'm pretty neutral to adding the federal stuff to utils.js. Maybe just ever so slightly against, because I'm not sure it's necessary for LCP, but I don't think adding it would have a strong negative impact either. 

Personally though, I think the real issue here is that the request to get icons is blocking. I think a better approach would be to load the icons like we're doing, but non-blocking, and reserve space for the icons until they've loaded in to prevent any CLS. 
This way everything else still loads quickly and we're not needlessly loading icons.svg unless there are actual icons from there that need to be displayed.

mokimo: We did have a draft PR for [loading icons post LCP](https://github.com/orgs/adobecom/discussions/2325) - but ran into some more complex CLS issues according to @ryanmparrish ... if we can make it happen without the CLS tradeoff, I'd love to have it.

mokimo: >  I think, we can move back the loading of files federated.js and icons.css to load in parallel. 

This is the most un-intrusive change for now, so let's go with this to unblock this. @saugatmalla would you mind making the changes as long as Ryan is out?

Please also ensure that we only load federal.js/icon.js if there are icons within the first LCP section and that we load icons on a section-to-section basis. (e.g. we don't need to load icons somewhere on the bottom of the page that aren't strictly needed for the LCP, if we are just about to load the first section.)

saugatmalla: I noticed that icons are rather being fetched after sections are processed, so that it doesn't block for LCP ([see this feedback](#issuecomment-2394761870)). So `icon.js`, `icons.css` and `federared.js` can just be loaded in the later stage. 

I moved the blocking code piece of `icon.js` to `utils.js`.

mokimo:
> test/features/icons/icons.test.js:

 ❌ Icon Suppprt > Sets icon index class
      TypeError: setIconsIndexClass is not a function
        at n.<anonymous> (test/features/icons/icons.test.js:50:5)

saugatmalla:
> and on CC the icon is missing https://main--cc--adobecom.hlx.live/de/products/photoshop/features https://main--cc--adobecom.hlx.live/de/products/photoshop/features?milolibs=rparrish-fed-icons--milo--adobecom

@mokimo Could you please provide more details? I couldn't reproduce the missing icon
Screenshot 2024-10-31 at 10 14 47 AM

mokimo:
@saugatmalla my bad, seems like my local overwrites were interfereing. Spotted another issue, however, we are loading `icon.js` twice

https://main--dc--adobecom.hlx.live/acrobat/how-to/pdf-to-word-doc-converter?milolibs=rparrish-fed-icons--milo--adobecom
Screenshot 2024-11-04 at 11 31 33

mokimo:
> @mokimo Could you please provide more details? I couldn't reproduce the missing icon
This seems to be happening sporadically and not consistently @saugatmalla 

It might be a timing issue, captured an image of the dom state here:
Screenshot 2024-11-04 at 11 38 46

saugatmalla:
@mokimo
The issue was that it skipped loading icons when icon elements weren’t detected in the viewport. Over time, the rendering flow for icons evolved due to updates in the PR, and I think this issue got unnoticed. 

And, I removed the duplicate imports, however, I don't believe that was causing Nala tests to fail.

mokimo (approved):

SilviuLCF (approved):
Verified, testing details 
https://jira.corp.adobe.com/browse/MWPW-140452

overmyheadandbody:
PR does have sufficient approvals, merging as another PR depends on this one being merged.

PR #2997, opened 2024-10-02
MWPW-159291 Fix fragment recursive check algo

chrischrischris:
Detecting recursive fragments had a very simplistic check, but would incorrectly mark a fragment as recursive and therefore not load it.  This updates the algo to explicitly look up the fragment loading tree and see if the fragment about to load will end up loading the current fragment - aka start an infinite fragment loading loop.

This was happening when a fragment that has a nested fragment is used in different sections of a page.  In this case  the photoshop card fragment was being used multiple times in different sections (each tab is in its own section).

Resolves: [MWPW-159291](https://jira.corp.adobe.com/browse/MWPW-159291)

---

Circular Fragment Verification:

- Before: https://main--milo--adobecom.hlx.page/drafts/cpeyer/load-circ-frag?martech=off
- After: https://159291-frag--milo--adobecom.hlx.page/drafts/cpeyer/load-circ-frag?martech=off

---

For the test urls below the hello world banner and picture of weather radar should be displayed twice.

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/cpeyer/nested-frag-test?martech=off
- After: https://159291-frag--milo--adobecom.hlx.page/drafts/cpeyer/nested-frag-test?martech=off

### DIFF
--- a/libs/blocks/fragment/fragment.js
+++ b/libs/blocks/fragment/fragment.js
@@ -2,7 +2,6 @@
 import { createTag, getConfig, loadArea, localizeLink, customFetch } from '../../utils/utils.js';
 
 const fragMap = {};
-window.fm = fragMap;
 
 const removeHash = (url) => {
   const urlNoHash = url.split('#')[0];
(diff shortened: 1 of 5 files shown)


libs/blocks/fragment/fragment.js
```diff
@@ -2,6 +2,7 @@
 import { createTag, getConfig, loadArea, localizeLink, customFetch } from '../../utils/utils.js';
 
 const fragMap = {};
+window.fm = fragMap;
```
overmyheadandbody, line 5: I'd assume this can be removed, I didn't see it used in this file, nor did I see it in unit tests.

overmyheadandbody (commented):
I didn't get a chance to have a deeper look at this, I'll get back to it :)

libs/blocks/fragment/fragment.js
chrischrischris: Yep leftover debug code.  Thanks!

mokimo (approved):

meganthecoder (approved):

Axelcureno (approved):

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-159291

PR #2999, opened 2024-10-03
[MWPW-146354] - Geo modal crop issue fix for iphone devices in landscape orientation

Deva309:
Crop issue fixed by adding scroll if popup height is more than window height.
But this fix is only added for landscape orientation.

Resolves: [MWPW-146354](https://jira.corp.adobe.com/browse/MWPW-146354)

**Test URLs:**
- After: https://mwpw-146354-geo-modal--milo--deva309.hlx.page/?martech=off

QA:
https://main--cc--adobecom.hlx.page/drafts/devashish/vh?milolibs=mwpw-146354-geo-modal--milo--deva309

### DIFF
--- a/libs/features/georoutingv2/georoutingv2.css
+++ b/libs/features/georoutingv2/georoutingv2.css
@@ -151,3 +151,9 @@
     max-width: 720px;
   }
 }
+
+@media only screen and (max-device-width: 956px) and (orientation: landscape) {
+  .dialog-modal.locale-modal-v2 {
+      overflow: auto;
+  }
+}


sharmrj (approved):

libs/features/georoutingv2/georoutingv2.css
```diff
@@ -151,3 +151,9 @@
     max-width: 720px;
   }
 }
+
+@media only screen and (max-device-width: 956px) and (orientation: landscape) {
```
overmyheadandbody, line 155: Is there a reason to have this limitation to begin with? Could we not have this style for every device below desktop? Accessibility-wise, we should get some input as well. I tried mimicking a situation where the georouting modal overflows and:

The CTAs are not visible in the default view:
main--cc--adobecom hlx page_drafts_devashish_vh (1)

The close button is not visible when scrolled down:
main--cc--adobecom hlx page_drafts_devashish_vh

Deva309: There is an exception issue specifically with iPhone devices, which is why this fix hasn’t been applied universally.

For more details, please refer to this discussion: 
https://github.com/orgs/adobecom/discussions/2895

This is the only solution I found to be viable for this issue.

overmyheadandbody: I had a quick read through the discussion and I agree that the additional dropdown should expand upwards.

But I still don't see the reason for not applying the `overflow: auto;` property universally. I understand that the issue is just on iPhone devices, but why create an exception with a media query value that most people will not understand refers to an iPhone if we could simply apply the fix universally? There might be arguments against that, but I'd like to ensure that option has been explored before going down this path.

spadmasa:
Hi @Deva309  , Verified the flow in ipad 9th gen noticing the issue in landscape mode, please take a look
{8AA67070-B9C6-4FFB-9340-88C992A428FA}

CC : @prativas22 , @salonijain3 , @sigadamvenkata

PR #3000, opened 2024-10-03
[MWPW-159129] - Add full width GNAV css for adobe home

Deva309:
Added a CSS class for full-width GNAV on Adobe home, applied from tablet size and above.

Resolves: [MWPW-159129](https://jira.corp.adobe.com/browse/MWPW-159129)

**Test URLs:**
- After: https://mwpw-159129--milo--deva309.hlx.page/?martech=off

QA:
https://adobecom.github.io/nav-consumer/navigation-full-width.html?env=stage&navbranch=mwpw-159129&authoringpath=/federal/test-website

### DIFF
--- a/libs/navigation/navigation.css
+++ b/libs/navigation/navigation.css
@@ -36,7 +36,7 @@ header.global-navigation, header.global-navigation.feds--dark {
   .feds-promo-link:hover {
     color: #136FF6;
   }
-  .feds--full-width .feds-topnav{
+  .feds--full-width .feds-topnav {
     max-width: none;
     padding: 0 10px 0 15px;
   }
--- a/libs/navigation/navigation.css
+++ b/libs/navigation/navigation.css
@@ -36,4 +36,8 @@ header.global-navigation, header.global-navigation.feds--dark {
   .feds-promo-link:hover {
     color: #136FF6;
   }
+  .feds--full-width .feds-topnav{
+    max-width: none;
+    padding: 0 10px 0 15px;
+  }
 }
--- a/libs/navigation/navigation.css
+++ b/libs/navigation/navigation.css
@@ -38,6 +38,6 @@ header.global-navigation, header.global-navigation.feds--dark {
   }
   .feds--full-width .feds-topnav {
     max-width: none;
-    padding: 0 10px 0 15px;
+    padding: 0 15px;
   }
 }
--- a/libs/navigation/navigation.css
+++ b/libs/navigation/navigation.css
@@ -36,6 +36,7 @@ header.global-navigation, header.global-navigation.feds--dark {
   .feds-promo-link:hover {
     color: #136FF6;
   }
+  
   .feds--full-width .feds-topnav {
     max-width: none;
     padding: 0 15px;


bandana147 (approved):

libs/navigation/navigation.css
```diff
@@ -36,4 +36,8 @@ header.global-navigation, header.global-navigation.feds--dark {
   .feds-promo-link:hover {
     color: #136FF6;
   }
+  .feds--full-width .feds-topnav{
```
narcis-radu, line 39: nit: space before `{`

libs/navigation/navigation.css
```diff
@@ -36,4 +36,8 @@ header.global-navigation, header.global-navigation.feds--dark {
...
+    max-width: none;
+    padding: 0 10px 0 15px;
```
narcis-radu, line 41: I see two issues here:
1. this doesn't work on RTL
2. we're adding different left / right padding in order to center the content, but as soon as the content is changed (e.g. a different logo or different usage of universal navigation) things will not be centered anymore. Shouldn't we handle these use-cases as well?

narcis-radu (changes requested):

libs/navigation/navigation.css
Deva309: Done

libs/navigation/navigation.css
Deva309: @narcis-radu Regarding the above points:

1. I only removed the center-alignment behavior and did not make any changes related to RTL functionality. It should function as it did before. Is there a particular scenario you’d like to point out?
2. The change was made to enable full-width behavior specifically for Adobe Home, as they use a full-width GNAV. That’s why additional spacing was added, which is same as for current Adobe Home already. We plan to request that they include `feds--full-width` in their markup. This change is not being applied across the entire Milo codebase, but only for the standalone GNAV.

overmyheadandbody: I agree with @narcis-radu that I don't see a point in not using the same value for right and left paddings. The logo has a left and right padding of `8px`. Utilities have a left and right padding of `8px`. Why would we not have the same value for the entire navigation? This should be either `0 10px` or `0 15px`, which will reflect nicely on RTL as well.

If you leave it as is, you will not actually have the navigation centered, it will lean towards the left on LTR or the right on RTL by `5px`.

Deva309: Agreed, I’ve updated it to 0 15px to prevent any misalignment in the future, whether for different logos or varying uses of the universal navigation

libs/navigation/navigation.css
```diff
@@ -36,4 +36,8 @@ header.global-navigation, header.global-navigation.feds--dark {
   .feds-promo-link:hover {
     color: #136FF6;
```
overmyheadandbody, line 37: I'm guessing you can't use Milo variables in the standalone version, right? Even so, I'd argue it's valuable to define variables in this file as well, so that when changes are needed, updates are required in a single place. This, for example, matches `--link-hover-color`. L34 matches `--link-color`. Depending on how this standalone navigation is architected, you might want to use the same variable names or adapted ones.

Deva309: We will cover it in separate as it is not related to this issue.
If that works.

prativas22:
verified the full-width GNAV that was applied from tablet size and up on Adobe Home.
https://adobecom.github.io/nav-consumer/navigation.html?env=stage&navbranch=stage&authoringpath=/federal/dev/gnavtest/locgnav
Before:
{312FCEF1-50D1-4205-9F2C-8EC9AD9C7128}

After:
{642420E5-B516-44B6-B4A3-A97CA8520565}
signed in
{5B8CDCE6-6B43-43DC-8474-39DD876323E3}
Device:
{6F10DB09-E714-4FC7-9B42-E3A5EF710397}
{F8CA8F39-A013-4FB3-9327-CA308BA9D477}

libs/navigation/navigation.css
```diff
@@ -36,4 +36,8 @@ header.global-navigation, header.global-navigation.feds--dark {
...
   }
+  .feds--full-width .feds-topnav {
```
narcis-radu, line 39: nit: we usually do a new line before a new declaration

Deva309: Done

PR #3007, opened 2024-10-03
MWPW-152455 added handling of "no placeholders" case for parsePlaceholders()

denlight:
* added handling of "no placeholders" case for parsePlaceholders()

Resolves: [MWPW-152455](https://jira.corp.adobe.com/browse/MWPW-152455)

QA Instructions: The before link has an error message the after link does not:
MEP Error parsing manifests: TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q4/placeholders/placeholders
- After: https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q4/placeholders/placeholders?milolibs=MWPW-152455

Psi-check: https://mwpw-152455--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -644,7 +644,7 @@ export async function createMartechMetadata(placeholders, config, column) {
 }
 
 /* c8 ignore start */
-function parsePlaceholders(placeholders, config, selectedVariantName = '') {
+export function parsePlaceholders(placeholders, config, selectedVariantName = '') {
   if (!placeholders?.length || selectedVariantName === 'default') return config;
   const valueNames = [
     selectedVariantName.toLowerCase(),
@@ -655,8 +655,9 @@ function parsePlaceholders(placeholders, config, selectedVariantName = '') {
     'value',
     'other',
   ];
-  const [val] = Object.entries(placeholders[0])
-    .find(([key]) => valueNames.includes(key.toLowerCase()));
+  const keys = placeholders && placeholders.length ? Object.entries(placeholders[0]) : [];
+  const [val] = keys.find(([key]) => valueNames.includes(key.toLowerCase()));
+
   if (val) {
     const results = placeholders.reduce((res, item) => {
       res[item.key] = item[val];
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -655,7 +655,7 @@ export function parsePlaceholders(placeholders, config, selectedVariantName = ''
     'value',
     'other',
   ];
-  const keys = placeholders && placeholders.length ? Object.entries(placeholders[0]) : [];
+  const keys = placeholders?.length ? Object.entries(placeholders[0]) : [];
   const keyVal = keys.find(([key]) => valueNames.includes(key.toLowerCase()));
   const key = keyVal?.[0];
 
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -656,17 +656,18 @@ export function parsePlaceholders(placeholders, config, selectedVariantName = ''
     'other',
   ];
   const keys = placeholders && placeholders.length ? Object.entries(placeholders[0]) : [];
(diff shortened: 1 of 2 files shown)


libs/features/personalization/personalization.js
```diff
@@ -655,8 +655,9 @@ function parsePlaceholders(placeholders, config, selectedVariantName = '') {
     'value',
     'other',
   ];
-  const [val] = Object.entries(placeholders[0])
-    .find(([key]) => valueNames.includes(key.toLowerCase()));
+  const keys = placeholders && placeholders.length ? Object.entries(placeholders[0]) : [];
```
mokimo, line 658: Can we long term try to harmonize this? This seems to rebuild the existing milo placeholder functionality, it'd be better if you can use the core placeholders methods and extend them to support MEP use-cases where needed.

We shouldn't have two different places of code that do the same thing.

vgoodric: @mokimo I agree this is something we should keep an eye on.  If the functionalityfor MEP in the existing placeholders feature becomes larger, then I do think we want will want to use export to share the relevant code.  But right now, there is only one line of code in placeholders to use MEP. 

In `libs\features\placeholders.js` in the `getPlaceholder` function, there is this line:
`if (config.placeholders?.[key]) return config.placeholders[key];`

And that's it.  So I don't think it makes sense to do it at this point.

vgoodric (approved):

libs/features/personalization/personalization.js
vgoodric, line 658:
```suggestion
  const keys = placeholders?.length ? Object.entries(placeholders[0]) : [];
```

test/features/personalization/parseNestedPlaceholders.test.js
```diff
@@ -53,3 +54,19 @@ describe('replacePlaceholders()', () => {
...
+    const response = parsePlaceholders([
+      {
+        'en-US': 'bar',
+        key: 'foo',
+      },
+    ], config);
+    expect(response.placeholders.foo).to.equal('bar');
+  });
+  it('should not break when there are no placeholders available', () => {
+    const response = parsePlaceholders(null, config);
```
vgoodric, line 69: const response = parsePlaceholders([
      {
        'fr': 'bar',
        key: 'foo',
      },
    ], config);

vgoodric, line 69:
```suggestion
    const response = parsePlaceholders([
      {
        'fr': 'bar',
        key: 'foo',
      },
    ], config);
```

test/features/personalization/parseNestedPlaceholders.test.js
```diff
@@ -53,3 +54,24 @@ describe('replacePlaceholders()', () => {
...
+  it('should parse placeholders', () => {
+    const response = parsePlaceholders([
+      {
+        'en-US': 'bar',
+        key: 'foo',
+      },
+    ], config);
+    expect(response.placeholders.foo).to.equal('bar');
+  });
+  it('should not break when there are no placeholders available', () => {
```
vgoodric, line 68:
```suggestion
  it('should not break when there are no applicable placeholders available', () => {
```

libs/features/personalization/personalization.js
```diff
@@ -655,17 +655,19 @@ function parsePlaceholders(placeholders, config, selectedVariantName = '') {
     'value',
     'other',
   ];
-  const [val] = Object.entries(placeholders[0])
```
denlight, line 658: it was breaking here, where it was trying to deconstruct an array that was found by `find` function. But nothing was found `const [val]` would break because undefined is treated like an array

mokimo (approved):
seems like this breaks a commerce nala placeholder test, otherwise lgtm!

libs/features/personalization/personalization.js
mokimo: Keeping an eye on it sounds good and I'm fine to keep this as they are for now 👍

PR #3010, opened 2024-10-03
MWPW-150610: implement partial load options in configurator

sanrai:
- Add "Partial Load Enabled" checkbox to Basics panel
- Add "Partial Load Count" input field when partial load is enabled
- Update config object to include partialLoadWithBackgroundFetch settings
- add tests for partial load settings
- add test for partial load initialization

This feature allows authors to optimize page load times for large card collections by initially loading a subset of cards and fetching the rest in the background. 

Authors can enable partial loading and specify the initial load count through the CaaS configurator tool.

Resolves: [MWPW-150610](https://jira.corp.adobe.com/browse/MWPW-150610)

URL for testing:
- Before: https://main--milo--adobecom.hlx.page/drafts/sanrai/foo?martech=off
- After: https://MWPW-150610--milo--adobecom.hlx.page/drafts/sanrai/foo?martech=off

### DIFF
--- a/libs/blocks/caas-config/caas-config.js
+++ b/libs/blocks/caas-config/caas-config.js
@@ -109,6 +109,8 @@ const defaultOptions = {
       'business.stage.adobe.com/chimera-api/collection',
     '14257-chimera.adobeioruntime.net/api/v1/web/chimera-0.0.1/collection':
       '14257-chimera.adobeioruntime.net/api/v1/web/chimera-0.0.1/collection',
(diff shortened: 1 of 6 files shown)


chrischrischris (approved):
Looks fine, but I wonder if we should shorten the property name `partialLoadWithBackgroundFetch` to be shorter since that has to be inserted into the encoded state config.  I think `partialLoad` should be enough?

sheridansunier (approved):

PR #3018, opened 2024-10-07
[MWPW-159140] Overlapping and alignment issues in product merch cards

rohitsahu:
Fixed max width of heading and badge if badge is present for product cards.

Resolves: [MWPW-159140](https://jira.corp.adobe.com/browse/MWPW-159140)

**Test URLs:**
- Before: https://main--dc--adobecom.hlx.page/drafts/steenmeyer/acrobat-pricing-samples/acrobat-pricing
- After: https://mwpw-155061--milo--rohitsahu.hlx.page/drafts/steenmeyer/acrobat-pricing-samples/acrobat-pricing?milolibs=local

### DIFF
--- a/libs/deps/mas/merch-card.js
+++ b/libs/deps/mas/merch-card.js
@@ -113,7 +113,7 @@ import{LitElement as Dt}from"../lit-all.min.js";import{LitElement as wt,html as 
         right: 0;
         font-size: var(--type-heading-xxs-size);
         font-weight: 500;
-        max-width: 180px;
+        max-width: 30%;
         line-height: 16px;
         text-align: center;
         padding: 8px 11px;
@@ -1550,6 +1550,10 @@ merch-card [slot='heading-xs'] {
     line-height: var(--consonant-merch-card-heading-xs-line-height);
     margin: 0;
     text-decoration: none;
+}
+
+merch-card .badge-heading {
+    max-width: 70%;
 }
 
 merch-card.dc-pricing [slot='heading-xs'] {
--- a/libs/blocks/merch-card/merch-card.js
+++ b/libs/blocks/merch-card/merch-card.js
@@ -230,9 +230,6 @@ const parseContent = async (el, merchCard) => {
       if (slotName) {
         if (['H2', 'H3', 'H4', 'H5'].includes(tagName)) {
           element.classList.add('card-heading');
-          if (merchCard.badgeText) {
-            element.closest('div[role="tabpanel"')?.classList.add('badge-merch-cards');
-          }
           if (HEADING_MAP[merchCard.variant]?.[tagName]) {
             tagName = HEADING_MAP[merchCard.variant][tagName];
           } else {
@@ -244,6 +241,12 @@ const parseContent = async (el, merchCard) => {
             }
             tagName = `H${headingSize}`;
             headingSize += 1;
+          }
+          if (merchCard.badgeText) {
+            element.closest('div[role="tabpanel"')?.classList.add('badge-merch-cards');
+            if (tagName === 'H3') {
+              element.classList.add('badge-heading');
+            }
           }
         }
         element.setAttribute('slot', slotName);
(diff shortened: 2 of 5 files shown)


mirafedas:
@rohitsahu on bigger screens when there are three cards in one row it doesn't look right, could you please check it out:
Screenshot 2024-10-07 at 12 55 02

libs/deps/mas/merch-card.js
```diff
@@ -113,7 +113,7 @@ import{LitElement as Dt}from"../lit-all.min.js";import{LitElement as wt,html as
         right: 0;
         font-size: var(--type-heading-xxs-size);
         font-weight: 500;
-        max-width: 180px;
+        max-width: 30%;
```
yesil, line 116: Sometimes a card can be super-wide spanning over 4 columns. With 30% it will grow more than it is supposed to be. I think the 180px is safer.
Also in some locales 30% might be lower than 180px, causing the labels flowing to next line.

libs/deps/mas/merch-card.js
```diff
@@ -1552,6 +1552,10 @@ merch-card [slot='heading-xs'] {
     text-decoration: none;
 }
 
+merch-card .badge-heading {
+    max-width: 70%;
```
yesil, line 1556: If you think this approach is safer/better, I recommend to remove `adjustTitleWidth` method in `libs/features/mas/web-components/src/variants/variant-layout.js` and solely use this approach.

libs/deps/mas/merch-card.js
rohitsahu: @yesil got it. Though fixing the max-width to 180 px has the disadvantage that if the label width becomes 180px in some case and card width is around 300px, then badge can again overlap with the title

yesil (commented):
@rohitsahu as an alternative approach, could you please try:https://github.com/adobecom/milo/blob/stage/libs/features/mas/web-components/src/variants/plans.js#L15

the `adjustTitleWidth` is supposed to address excactly this issue.

for ref: https://github.com/adobecom/milo/blob/stage/libs/features/mas/web-components/src/variants/variant-layout.js#L111

libs/features/mas/web-components/src/variants/product.js
```diff
@@ -56,6 +56,7 @@ export class Product extends VariantLayout {
     if (!isMobile()) {
       this.adjustProductBodySlots();
     }
+    this.adjustTitleWidth();
```
rahulgupta999, line 59: should this be done for all card variants?
@yesil , @npeltier

yesil: Since this has performance impact, I would limit it to variants we know this is needed.

libs/features/mas/web-components/src/variants/product.js
```diff
@@ -82,5 +83,9 @@ export class Product extends VariantLayout {
         min-height: var(--consonant-merch-card-product-callout-content-height);
         display: block;
     }
+      
+    :host([variant='product']) ::slotted([slot='heading-xs']) {
+      max-width: var(--consonant-merch-card-heading-xs-max-width, 100%);
```
yesil, line 88: is this still needed?

yesil (approved):

libs/features/mas/web-components/src/variants/product.js
rohitsahu: Yes, this limits the title max-width for title.   "adjustTitleWidth" sets the calue of "--consonant-merch-card-heading-xs-max-width" and we need to apply the css here to set it into effect

Roycethan:
> @rohitsahu Second issue related to Secure transaction label alignment on devices is not addressed here, plz take a look thanks https://main--dc--adobecom.hlx.page/drafts/steenmeyer/acrobat-pricing-samples/acrobat-pricing?milolibs=mwpw-159140--milo--rohitsahu image

@rohitsahu Lets discuss over the ticket and create different ticket for this once after confirmed, meanwhile marking Ready for stage

Axelcureno (approved):
lgtm

PR #3021, opened 2024-10-08
MWPW-160061: Clickable App and Playstore links

mokimo:
### Description
Make the app store and play store links clickable again which got lost as part of https://github.com/adobecom/milo/pull/1768

Resolves: [MWPW-160061](https://jira.corp.adobe.com/browse/MWPW-160061)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/sara/media?martech=off
- After: https://mwpw-160061-qr-code-media--milo--mokimo.hlx.page/drafts/sara/media?martech=off

### DIFF
--- a/libs/blocks/media/media.js
+++ b/libs/blocks/media/media.js
@@ -30,9 +30,10 @@ function decorateAvatar(el) {
 function decorateQr(el) {
   const text = el.querySelector('.text');
   if (!text) return;
-  const appStore = text.children[(text.children.length - 1)];
-  const googlePlay = text.children[(text.children.length - 2)];
   const qrImage = text.children[(text.children.length - 3)];
+  const qrCodeLinks = el.querySelectorAll('a');
+  const appStore = qrCodeLinks[0];
+  const googlePlay = qrCodeLinks[1];
   appStore.classList.add('app-store');
   appStore.textContent = '';
   googlePlay.classList.add('google-play');
--- a/libs/blocks/media/media.js
+++ b/libs/blocks/media/media.js
@@ -32,8 +32,8 @@ function decorateQr(el) {
   if (!text) return;
   const qrImage = text.children[(text.children.length - 3)];
   const qrCodeLinks = el.querySelectorAll('a');
-  const appStore = qrCodeLinks[0];
-  const googlePlay = qrCodeLinks[1];
+  const appStore = qrCodeLinks[1];
+  const googlePlay = qrCodeLinks[0];
   appStore.classList.add('app-store');
   appStore.textContent = '';
   googlePlay.classList.add('google-play');
--- a/libs/blocks/media/media.js
+++ b/libs/blocks/media/media.js
@@ -30,15 +30,15 @@ function decorateAvatar(el) {
 function decorateQr(el) {
   const text = el.querySelector('.text');
   if (!text) return;
+  const appStore = text.children[(text.children.length - 1)]?.querySelector('a');
+  const googlePlay = text.children[(text.children.length - 2)]?.querySelector('a');
   const qrImage = text.children[(text.children.length - 3)];
-  const qrCodeLinks = el.querySelectorAll('a');
-  const appStore = qrCodeLinks[1];
-  const googlePlay = qrCodeLinks[0];
+  if (!qrImage || !appStore || !googlePlay) return;
+  qrImage.classList.add('qr-code-img');
   appStore.classList.add('app-store');
   appStore.textContent = '';
   googlePlay.classList.add('google-play');
   googlePlay.textContent = '';
-  qrImage.classList.add('qr-code-img');
 }
 
 export default async function init(el) {


libs/blocks/media/media.js
```diff
@@ -30,9 +30,10 @@ function decorateAvatar(el) {
 function decorateQr(el) {
   const text = el.querySelector('.text');
   if (!text) return;
-  const appStore = text.children[(text.children.length - 1)];
-  const googlePlay = text.children[(text.children.length - 2)];
   const qrImage = text.children[(text.children.length - 3)];
+  const qrCodeLinks = el.querySelectorAll('a');
```
overmyheadandbody, line 34: Should we not stick to `text.querySelectorAll('a')`?

libs/blocks/media/media.js
```diff
@@ -30,9 +30,10 @@ function decorateAvatar(el) {
...
+  const appStore = qrCodeLinks[1];
+  const googlePlay = qrCodeLinks[0];
```
robert-bogos, line 36: Should we add a fail check?
```suggestion
  const qrCodeLinks = el.querySelectorAll('a');
  const appStore = qrCodeLinks[1];
  const googlePlay = qrCodeLinks[0];
  if(!appStore || !googlePlay) return;
```

robert-bogos (approved):

libs/blocks/media/media.js
```diff
@@ -30,9 +30,10 @@ function decorateAvatar(el) {
 function decorateQr(el) {
   const text = el.querySelector('.text');
   if (!text) return;
-  const appStore = text.children[(text.children.length - 1)];
-  const googlePlay = text.children[(text.children.length - 2)];
   const qrImage = text.children[(text.children.length - 3)];
```
overmyheadandbody, line 33: Is it certain that all these elements exist? Otherwise, it wouldn't hurt to use optional chaining when trying to add classes below.

libs/blocks/media/media.js
```diff
@@ -30,9 +30,10 @@ function decorateAvatar(el) {
 function decorateQr(el) {
   const text = el.querySelector('.text');
   if (!text) return;
-  const appStore = text.children[(text.children.length - 1)];
```
overmyheadandbody, line 33: Alternatively, we could add `.querySelector('a')`, since the structure should be set in stone.

overmyheadandbody (commented):
A few questions before we rush into this :)

libs/blocks/media/media.js
mokimo: Yeah this seems to be hardcoded, so can do that

libs/blocks/media/media.js
mokimo: Added a early return as @robert-bogos suggested

libs/blocks/media/media.js
mokimo: Implemented your suggestion with the querySelector below

libs/blocks/media/media.js
mokimo: Added

libs/blocks/media/media.js
```diff
@@ -30,14 +30,15 @@ function decorateAvatar(el) {
...
-  const googlePlay = text.children[(text.children.length - 2)];
+  const appStore = text.children[(text.children.length - 1)]?.querySelector('a');
+  const googlePlay = text.children[(text.children.length - 2)]?.querySelector('a');
   const qrImage = text.children[(text.children.length - 3)];
+  if (!qrImage || !appStore || !googlePlay) return;
```
overmyheadandbody, line 36: Could we have situations where we just have a QR code or just store apps or just one store app? Judging by the rigidity of the markup, I'd assume not, but this feels like it might come back to haunt us. I'd vote for `qrImage?.classList.add('qr-code-img');` and so on for the next lines, but maybe I'm overly cautious.

overmyheadandbody (approved):
Thanks for integrating the feedback!

libs/blocks/media/media.js
mokimo: I was wondering about the same thing! But the current code would throw an error and optional chaining assignments are currently disabled by our linter and according to MDN https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#invalid_optional_chaining

SilviuLCF (approved):
Verified, testing details https://jira.corp.adobe.com/browse/MWPW-160061

libs/blocks/media/media.js
mokimo: We can also do this, if you prefer:

```js
  if(appStore) {
    appStore.classList.add('app-store');
    appStore.textContent = '';
  }
  if(googlePlay) {
    googlePlay.classList.add('google-play');
    googlePlay.textContent = '';
  }
  qrImage?.classList.add('qr-code-img');
```

mokimo: According to the requirements in the ticket, the QR code is supposed to turn into the app store links, so I would assume we should always have all 3

ryanmparrish (approved):

sukamat (approved):

narcis-radu (approved):

PR #3028, opened 2024-10-09
MWPW-157999 - [Table] Add-on content

Sartxi:
As a Milo Engineer, I want to add functionality to Milo table block so I can add design features from this [design](https://www.figma.com/design/g3esK4XXSAE18F3snLXLwm/AI-Assistant-Grey-Tag-Specs?node-id=1-15391&node-type=frame&t=FAQ9azoKtgmlgDLr-0). 

Screenshot 2024-10-09 at 9 37 08 AM

### **New features Include:**

1. Rendering additional 'add-on' content in the table headers:
The table header decorator function requires that the content be written in a specific order. This PR introduces an additional method of inserting content into the header columns. Utilizing additional rows that are titled with an `ADDON` key. To use, add the position, order, and optional tag name separated by underscores to the row titles in these additional header rows and the table block will handle the rest. Example: `ADDON_BODY_AFTER_PROMO`.

2. Rendering label/badge styles in the table body:
To provide support for this design we needed to be able to create a custom element style in the table body. We did this by authoring the desired content with the `quote` style in word. When using add-on content the table block will style any `blockquote` content in the row titles to be styled like a label/badge.

Resolves: [MWPW-157999](https://jira.corp.adobe.com/browse/MWPW-157999)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/sarchibeque/table/add-on?martech=off
- After: https://sartxi-mwpw-157999-table-addon--milo--adobecom.hlx.page/drafts/sarchibeque/table/add-on?martech=off

- Before: https://main--milo--adobecom.hlx.page/drafts/sarchibeque/table/add-on-greentext?martech=off
- After: https://sartxi-mwpw-157999-table-addon--milo--adobecom.hlx.page/drafts/sarchibeque/table/add-on-greentext?martech=off

### DIFF
--- a/libs/blocks/table/table.css
+++ b/libs/blocks/table/table.css
@@ -12,7 +12,7 @@
   border-color: var(--border-color);
 }
 
-.section[class *="grid-width-"] .table {
+.section[class *= "grid-width-"] .table {
   width: 100%;
 }
 
@@ -20,7 +20,7 @@
   display: inline-block;
 }
 
-.table>.row {
+.table > .row {
   display: grid;
   grid-template-columns: repeat(auto-fit, minmax(50%, 1fr));
 }
@@ -43,7 +43,7 @@
   align-items: center;
 }
 
-.table:not(.left, .header-left) .col {
+.table:not(.left, .header-left) .col{
   justify-content: center;
 }
 
@@ -206,15 +206,15 @@
   justify-content: flex-end;
 }
 
-.table.merch.button-right .row-heading .col.col-heading .buttons-wrapper>* {
+.table.merch.button-right .row-heading .col.col-heading .buttons-wrapper > * {
   margin: 12px 0 12px 12px;
 }
 
-.table .row-heading .col.col-heading .buttons-wrapper>* {
+.table .row-heading .col.col-heading .buttons-wrapper > * {
   margin: 10px 6px;
 }
 
-.table.header-left .row-heading .col.col-heading .buttons-wrapper>* {
+.table.header-left .row-heading .col.col-heading .buttons-wrapper > * {
   margin-inline-start: 0;
 }
 
@@ -280,7 +280,7 @@
   cursor: pointer;
 }
 
-.table .section-head .section-head-title> :not(.icon) {
+.table .section-head .section-head-title > :not(.icon) {
   font-size: var(--type-body-m-size);
   line-height: var(--type-heading-s-lh);
   width: calc(100% - 20px);
@@ -300,11 +300,11 @@
   display: none;
 }
 
-.table:not(.merch) .section-head> :not(:first-child) {
+.table:not(.merch) .section-head > :not(:first-child) {
   border-left-color: transparent;
 }
 
-.table:not(.merch) .section-head> :not(:last-child) {
+.table:not(.merch) .section-head > :not(:last-child) {
   border-right-color: transparent;
 }
 
@@ -549,11 +549,11 @@ header.global-navigation {
     display: grid;
   }
 
-  .table>.row {
+  .table > .row {
     grid-template-columns: repeat(auto-fit, minmax(100px, 1fr));
   }
 
-  .table.merch>.row {
+  .table.merch > .row {
     grid-template-columns: repeat(auto-fit, minmax(200px, 1fr));
     gap: var(--spacing-m);
   }
@@ -566,12 +566,12 @@ header.global-navigation {
   }
 }
 
-.table.merch>.row {
+.table.merch > .row {
   grid-template-columns: repeat(auto-fit, minmax(100px, 2fr));
   gap: var(--spacing-xs);
 }
 
-.table.merch .section-head .col.section-head-title,
+.table.merch .section-head .col.section-head-title ,
 .table.merch .row-heading .col.col-heading {
   text-align: left;
 }
@@ -586,7 +586,7 @@ header.global-navigation {
   justify-content: flex-start;
 }
 
-.table.merch .row-heading .col.col-heading .buttons-wrapper>* {
+.table.merch .row-heading .col.col-heading .buttons-wrapper > * {
   margin: 12px 12px 12px 0;
 }
 
@@ -613,13 +613,12 @@ header.global-navigation {
 
 /* Start tablet styles */
 @media (max-width: 899px) {
-
   .table,
   .table.merch {
     margin: 0 30px;
   }
 
-  .section[class *="grid-width-"] .table {
+  .section[class *= "grid-width-"] .table {
     margin: 0;
   }
 
@@ -745,12 +744,12 @@ header.global-navigation {
     border-top-right-radius: 0;
   }
 
-  .table .col-heading.force-last+.col-heading {
+  .table .col-heading.force-last + .col-heading {
     border-top-right-radius: 0;
     border-top-left-radius: var(--border-radius);
   }
 
-  [dir="rtl"] .table .col-heading.force-last+.col-heading {
+  [dir="rtl"] .table .col-heading.force-last + .col-heading {
     border-top-left-radius: 0;
     border-top-right-radius: var(--border-radius);
   }
@@ -767,7 +766,6 @@ header.global-navigation {
 }
 
 @media (max-width: 480px) {
-
   .table,
   .table.merch {
     min-width: 100%;
@@ -779,7 +777,7 @@ header.global-navigation {
     align-items: initial;
   }
 
-  .table :is(.heading-button, .action-area) {
+  .table :is(.heading-button,.action-area) {
     max-width: 100%;
   }
 
--- a/libs/blocks/table/table.css
+++ b/libs/blocks/table/table.css
@@ -12,7 +12,7 @@
   border-color: var(--border-color);
 }
 
-.section[class *= "grid-width-"] .table {
+.section[class *="grid-width-"] .table {
   width: 100%;
 }
 
@@ -20,7 +20,7 @@
   display: inline-block;
 }
 
-.table > .row {
+.table>.row {
   display: grid;
   grid-template-columns: repeat(auto-fit, minmax(50%, 1fr));
 }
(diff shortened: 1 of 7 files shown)


libs/blocks/table/table.css
```diff
@@ -456,6 +456,87 @@
...
+
+.table.has-addon .label {
+  background-color: var(--color-gray-300);
+  font-size: var(--type-body-xs-size);
+  line-height: var(--type-body-xs-lh);
+  border-radius: 4px;
+  display: flex;
+  align-items: center;
+  flex-direction: row;
+  text-align: left;
```
overmyheadandbody, line 482: Maybe this to account for RTL OOTB?
```suggestion
  text-align: start;
```

libs/blocks/table/table.css
```diff
@@ -722,4 +803,4 @@ header.global-navigation {
   .table .action-area .con-button.button-l {
     overflow: hidden;
   }
-}
+}
```
overmyheadandbody, line 806:
```suggestion
}

```

overmyheadandbody (commented):
Just started looking over this, but I noticed an issue on mobile I wanted to bring up already:
sartxi-mwpw-157999-table-addon--milo--adobecom hlx page_drafts_sarchibeque_table_add-on-greentext_martech=off(iPhone 12 Pro)

libs/blocks/table/table.css
Sartxi: great find!

libs/blocks/table/table.css
```diff
@@ -456,6 +456,87 @@
   }
 }
 
+.table.has-addon .row-heading .col.col-heading {
+  display: flex;
```
overmyheadandbody, line 460: This seems to be the culprit of the issue from my previous screenshot, as it collides with `.table:not(.merch) .col-heading.col-1` which is hidden on mobile, some tweaks should be made to take this into account.

libs/blocks/table/table.js
```diff
@@ -97,6 +97,36 @@ function handleHeading(table, headingCols) {
...
+  if (addOns.length) {
+    table.classList.add('has-addon');
+    addOns.forEach((addOn) => {
+      const addOnRow = addOn.parentElement;
+      addOnRow.remove();
+      const [position, order, style] = addOn.innerText.split('_')
+        .filter((key) => key !== addOnKey).map((key) => key.toLowerCase());
+      if (position && order) {
+        [...table.querySelector('.row-heading').children].forEach((headCol) => {
+          const indexAttr = 'data-col-index';
```
overmyheadandbody, line 113: Nit: we can move this outside of the loop, since it doesn't get updated

libs/blocks/table/table.js
```diff
@@ -97,6 +97,36 @@ function handleHeading(table, headingCols) {
   });
 }
 
+function handleAddOnContent(table) {
+  const addOnKey = 'ADDON';
+  const addOns = [...table.querySelectorAll('.section-row-title')]
+    .filter((row) => row.innerText.includes(addOnKey));
+  if (addOns.length) {
```
overmyheadandbody, line 104: This is getting quite "indenty", could we switch to an early return?
```suggestion
  if (!addOns.length) return;
```
Same goes for L111 & L115.

libs/blocks/table/table.js
```diff
@@ -97,6 +97,36 @@ function handleHeading(table, headingCols) {
...
+      const [position, order, style] = addOn.innerText.split('_')
+        .filter((key) => key !== addOnKey).map((key) => key.toLowerCase());
+      if (position && order) {
+        [...table.querySelector('.row-heading').children].forEach((headCol) => {
+          const indexAttr = 'data-col-index';
+          const colIndex = headCol.getAttribute(indexAttr);
+          if (colIndex > 1) { // ignore the key column
+            const content = `${position}-${order}`;
+            const tag = createTag('div', { class: content }, [...addOnRow.children]
+              .find((col) => col.getAttribute(indexAttr) === colIndex)?.innerHTML);
```
overmyheadandbody, line 118: Wherever possible, we aim to avoid `.innerHTML`, since it can lead to Kodiak alerts. Could we use `.childNodes` instead?

libs/blocks/table/table.js
```diff
@@ -97,6 +97,36 @@ function handleHeading(table, headingCols) {
...
+            if (style) tag.classList.add(style);
```
overmyheadandbody, line 119: I worry a bit about the fact that these classes can be quite generic, like `label`, which might already have some styles attached from a more general stylesheet. Should we make them a bit more specific to the use-case at hand by programatically prepending a more specific string?

overmyheadandbody (commented):
We should make sure the options are also documented, as it wouldn't be obvious for authors what addons they can define.

libs/blocks/table/table.css
Sartxi: okay I blocked styling on the first column in the header to avoid this collision. Let me know if that works.

libs/blocks/table/table.js
Sartxi: Done

libs/blocks/table/table.js
Sartxi: Okay I added early returns for all.

libs/blocks/table/table.js
Sartxi: Okay great, done

libs/blocks/table/table.js
Sartxi: I am prepending `addon-` to the tag now!

ryanmparrish:
Looking good. Few super minor design things to touch on if you think there worth it. 
- The `addon-label` icon can get pretty close to the text, maybe give it a little margin-start. 
- The `table-title-text` is italic, though it is authored w/ an`<em>`, so maybe its just a doc mistake.
Screenshot 2024-10-10 at 12 04 42 PM
Should both this content use the same authoring style (quote or code, etc) since it looks similar? maybe it's diff enough, just a though.

Sartxi:
> Looking good. Few super minor design things to touch on if you think there worth it.
> 
> * The `addon-label` icon can get pretty close to the text, maybe give it a little margin-start.
> * The `table-title-text` is italic, though it is authored w/ an`<em>`, so maybe its just a doc mistake.
> 
> Screenshot 2024-10-10 at 12 04 42 PM Should both this content use the same authoring style (quote or code, etc) since it looks similar? maybe it's diff enough, just a though. Oop, the font color too on these looks a lil off.

Okay I have addressed your findings, thanks!

overmyheadandbody (approved):
Thanks for looking into all the feedback, looks good!

ryanmparrish (approved):
Thanks for design polish. Looks goood!

Few follow up items once this is merged which your already aware of. 
- update table documentation
- test loc workflow and add any exceptions in the htmlFlow for cell keys `ADDON` etc.

Sartxi:
> * test loc workflow and add any exceptions in the htmlFlow for cell keys `ADDON` etc

Thanks for the reminders, I will make sure get them done!
- [Sub-Task](https://jira.corp.adobe.com/browse/MWPW-160293) for localization workflow updates
- [Sub-Task](https://jira.corp.adobe.com/browse/MWPW-158002) for authoring documentation

libs/blocks/table/table.js
```diff
@@ -97,6 +97,34 @@ function handleHeading(table, headingCols) {
...
+function handleAddOnContent(table) {
+  const addOnKey = 'ADDON';
+  const addOns = [...table.querySelectorAll('.section-row-title')]
+    .filter((row) => row.innerText.includes(addOnKey));
+  if (!addOns.length) return;
+  table.classList.add('has-addon');
+  addOns.forEach((addOn) => {
+    const addOnRow = addOn.parentElement;
+    addOnRow.remove();
+    const [position, order, style] = addOn.innerText.split('_')
```
rgclayton, line 109: Any reason you are using `_` for authoring of the key? I don't think we should introduce a new pattern from what I can see. We've only used `-` or spaces between words. Hero-marquee uses `-` for it's row keys. 

Doing a quick test and replacing `_` with `-` and the addons are not added.
image

rgclayton: Should also sanitize the key to lowercase. Currently if it's not authored in all caps it breaks too.

[Example page](https://sartxi-mwpw-157999-table-addon--milo--adobecom.hlx.page/drafts/rclayton/test/table/add-on-greentext)

image

Sartxi: @rgclayton okay I agree, I have updated the authoring and sanitized the keys so you can use lowercase

rgclayton (approved):
I see there is a preexisting issue with CLS and the table. 

cc: @overmyheadandbody

libs/blocks/table/table.js
```diff
@@ -124,6 +141,8 @@ function handleAddOnContent(table) {
       el?.insertAdjacentElement(order === 'before' ? 'beforebegin' : 'afterend', tag);
     });
   });
+  window.addEventListener('resize', () => handleEqualHeight(table, '.row-heading'));
```
overmyheadandbody, line 144: We already have a `resize` event listener [here](https://github.com/adobecom/milo/blob/29a6056c98e75f4695b1e6b454e7ace7e7664df3/libs/blocks/table/table.js#L585). Could we consolidate the two?

I was also thinking whether it would be worth having this [debounced](https://github.com/adobecom/milo/blob/29a6056c98e75f4695b1e6b454e7ace7e7664df3/libs/utils/action.js#L1).

overmyheadandbody (approved):
For an eventual future iteration of the table block, we should look into using `grid` for the heading row. However, right now the table logic is a bit fragile, so this is an acceptable compromise to meet the requirements.

Thanks @Sartxi for finding a solution for this!

libs/blocks/table/table.js
Sartxi: Okay I consolidated resize events and added the debounce!

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-157999

PR #3037, opened 2024-10-14
MWPW-159605 Allow consumer apps set IMS timeout value

TsayAdobe:
* After DC frictionless pages are switched to using the IMS guest token, we observed a small percentage of users may get an IMS timeout error. In the past we didn't log the error. The investigation is in [MWPW-158350](https://jira.corp.adobe.com/browse/MWPW-158350).  
* This change allows a consumer app to adjust the IMS timeout value. The DC team can fine tune the value to reduce the error rate.

Resolves: [MWPW-159605](https://jira.corp.adobe.com/browse/MWPW-159605)

**Test URLs:**
- After: https://mwpw-159605--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -947,7 +947,7 @@ let imsLoaded;
 export async function loadIms() {
   imsLoaded = imsLoaded || new Promise((resolve, reject) => {
     const {
-      locale, imsClientId, imsScope, env, base, adobeid,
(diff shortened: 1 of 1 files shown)


mokimo (approved):
Could would love if we can contribute a more fine tuned value back after your experimentation

Blainegunn (commented):
Thank you @TsayAdobe

Blainegunn (approved):

PR #3039, opened 2024-10-14
MWPW-153998, 148129, 149504 [MILO][MEP] Change Target gnav setting to be useable for any post LCP updates

vgoodric:
Due to backlog from RCP, also contains 2 fully cleared PRs:
[#3064: MWPW-148129](https://github.com/adobecom/milo/pull/3064)
[#3047: MWPW-149504](https://github.com/adobecom/milo/pull/3047)

* Update target gnav setting to postLCP (there are no pages currently using gnav so we do not need to provide retroactive support)
* looks like "gnav" setting was no longer working due to changes in launch.  So now instead of calling martech post LCP, we call martech pre LCP,  but do not await the response until post LCP
* update MEP button to expect new setting

Resolves: [MWPW-153998](https://jira.corp.adobe.com/browse/MWPW-153998)

QA instructions: before and after link should have 2 updates: the marquee title and the chosen tab should be the first one, not the 2nd
the after link for post LCP should have 1 update: the first tab should be selected

Also, when my network is throttled to the recommended levels for performance checks, the after link with Target fully on is clocking LCP at around 4.8 - 6.5 seconds. With Target set to postlcp, it clocks in at around 3.4 seconds.

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q4/meppostlcp/simple-example
- After: https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q4/meppostlcp/simple-example?milolibs=meppostlcp
- After with Target set to postlcp: https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q4/meppostlcp/simple-example?milolibs=meppostlcp&target=postlcp
- Psi-check: https://meppostlcp--milo--adobecom.hlx.page/?martech=off


libs/features/personalization/personalization.js
```diff
@@ -1054,13 +1059,33 @@ export const combineMepSources = async (persEnabled, promoEnabled, mepParam) =>
...
+
 export async function init(enablements = {}) {
   let manifests = [];
   const {
     mepParam, mepHighlight, mepButton, pzn, promo, target, postLCP,
   } = enablements;
   const config = getConfig();
-  if (!postLCP) {
+  if (postLCP) {
+    IS_POST_LCP = true;
```
chrischrischris, line 1087: All caps infers a constant

libs/utils/utils.js
```diff
@@ -1049,7 +1049,7 @@ async function checkForPageMods() {
 
 async function loadPostLCP(config) {
   await decoratePlaceholders(document.body.querySelector('header'), config);
-  if (config.mep?.targetEnabled === 'gnav') {
+  if (config.mep?.targetEnabled === 'postlcp') {
```
chrischrischris, line 1052: Will this break existing pages that are using `gnav`?

vgoodric: we had one test where we used gnav and that test ended.  So we removed the setting from all pages.  This update should be in place long before the next test that needs this setting.

libs/features/personalization/personalization.js
vgoodric: changed it to isPostLCP

chrischrischris (approved):

denlight (approved):

markpadbe:
@vivgoodrich, there are some tests that aren't breaking, but still point to 'gnav.' Should these tests also be updated to use 'postlcp' along with this pr?
Screenshot 2024-10-15 at 3 01 40 PM

vgoodric:
> @vivgoodrich, there are some tests that aren't breaking, but still point to 'gnav.' Should these tests also be updated to use 'postlcp' along with this pr? 

Good catch.  Updated.

markpadbe (approved):

PR #3044, opened 2024-10-15
MWPW-159561 Add lcpSectionOne metric to web vitals

chrischrischris:
Logs if the LCP Element is in Section One of the page.

Resolves: [MWPW-159561](https://jira.corp.adobe.com/browse/MWPW-159561)

**Test URLs:**
- After: https://mwpw-159561--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/libs/utils/logWebVitals.js
+++ b/libs/utils/logWebVitals.js
@@ -76,7 +76,10 @@ function isFragmentFromMep(fragPath, mep) {
   });
 }
 
+const boolStr = (val) => (val ? 'true' : 'false');
+
 function observeLCP(lanaData, delay, mep) {
+  const sectionOne = document.querySelector('main > div');
   new PerformanceObserver((list) => {
     const entries = list.getEntries();
     const lastEntry = entries[entries.length - 1]; // Use the latest LCP candidate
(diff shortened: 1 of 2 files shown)


sukamat (approved):

rgclayton (approved):

SilviuLCF (approved):
Verified , ready to stage 
Testing details : https://jira.corp.adobe.com/browse/MWPW-159561

libs/utils/logWebVitals.js
```diff
@@ -76,18 +76,22 @@ function isFragmentFromMep(fragPath, mep) {
   });
 }
 
+const boolStr = (val) => (val ? 'true' : 'false');
```
overmyheadandbody, line 79: No need to adapt, as this does the trick, but for brevity we could've also used
```suggestion
const boolStr = (val) => `${!!val}`;
```

libs/utils/logWebVitals.js
```diff
@@ -76,18 +76,22 @@ function isFragmentFromMep(fragPath, mep) {
...
 function observeLCP(lanaData, delay, mep) {
+  const sectionOne = document.querySelector('main > div');
   new PerformanceObserver((list) => {
     const entries = list.getEntries();
     const lastEntry = entries[entries.length - 1]; // Use the latest LCP candidate
     lanaData.lcp = parseInt(lastEntry.startTime, 10);
     const lcpEl = lastEntry.element;
     lanaData.lcpElType = lcpEl.nodeName.toLowerCase();
     lanaData.lcpEl = getElementInfo(lcpEl);
+    lanaData.lcpSectionOne = boolStr(sectionOne.contains(lcpEl));
```
overmyheadandbody, line 90: I was wondering whether an empty document translates to a childless `main`, but it seems a `div` is always added inside. I was going to suggest some optional chaining, but it's not needed.

overmyheadandbody (approved):

mokimo (approved):

PR #3045, opened 2024-10-15
MWPW-159555 Add support for lana-sample query param

chrischrischris:
The `lana-sample` query param will override any sample rate defined elsewhere.  Setting it to "100" will force lana to send everything to splunk.

Resolves: [MWPW-159555](https://jira.corp.adobe.com/browse/MWPW-159555)

**Test URLs:**
- After: https://mwpw-159555--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/libs/utils/lana.js
+++ b/libs/utils/lana.js
@@ -1,3 +1,4 @@
+/* eslint-disable */
 (function () {
   const MSG_LIMIT = 2000;
 
--- a/libs/utils/lana.js
+++ b/libs/utils/lana.js
@@ -65,7 +65,8 @@
       return;
     }
 
-    const sampleRate = o.errorType === 'i' ? o.implicitSampleRate : o.sampleRate;
+    const sampleRateParam = parseInt(new URL(window.location).searchParams.get('lana-sample'), 10);
+    const sampleRate = sampleRateParam || (o.errorType === 'i' ? o.implicitSampleRate : o.sampleRate);
 
     if (!w.lana.debug && !w.lana.localhost && sampleRate <= Math.random() * 100) return;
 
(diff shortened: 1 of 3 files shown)


sukamat (approved):

rgclayton (approved):

SilviuLCF:
Hi @chrischrischris can you provide some guidance on how testing should be performed for this change? 

thank you

libs/utils/lana.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
overmyheadandbody, line 1: Is this needed for the whole file? I'd rather have ESLint disabled for specific purposes, otherwise we might miss a lot of errors.

overmyheadandbody (approved):

overmyheadandbody:
Seems the new unit test is failing - `❌ LANA > The lana-sample query param overrides existing sampleRate`

mokimo (approved):

thedoc31 (changes requested):
Assuming someone will always send in a valid integer between 0 and 100 is dangerous without any (apparent) input validation or error handling. What happens if I send in an invalid number or a value long enough to be a BigInt? Would lana know what to do if it got a NaN in this param?

chrischrischris:
@thedoc31 The `parseInt` handles the sanitization of the input.  If the `lana-sample` param is invalid it will be `NaN` and the next line will ignore it:

```
const sampleRateParam = parseInt(new URL(window.location).searchParams.get('lana-sample'), 10);
const sampleRate = sampleRateParam || (o.errorType === 'i' ? o.implicitSampleRate : o.sampleRate);
```

It's fine if the param is negative or a gigantic number.  Negative will behave like 0, anything over 100 behaves like 100.

SilviuLCF (approved):
Verified, ready to STAGE

PR #3053, opened 2024-10-16
Feat(MWPW-146367):Added accessibility player controls (NON MPC)

sharath-kannan:
* Added accessibility controls for all blocks supporting videos. 

Resolves: [MWPW-146367](https://jira.corp.adobe.com/browse/MWPW-146367)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ramuntea/accessibility/video-use-cases?martech=off
- After: https://mwpw-146367--milosh--sharath-kannan.hlx.page/drafts/ramuntea/accessibility/video-use-cases?martech=off


spadmasa:
Hi @sharath-kannan PR changes are breaking the product pages  , please take a look 
https://main--cc--adobecom.hlx.live/products/illustrator?milolibs=mwpw-146367--milosh--sharath-kannan
https://main--cc--adobecom.hlx.page/education?milolibs=mwpw-146367--milosh--sharath-kannan
https://main--cc--adobecom.hlx.page/products/acrobat-pro-cc?milolibs=mwpw-146367--milosh--sharath-kannan

{7839BD79-F02A-4798-939E-1DC8FE2D4898}
{3746EE98-1A6E-4257-8579-456983226E0D}


CC : @salonijain3 , @sigadamvenkata

libs/blocks/video/video.css
```diff
@@ -6,3 +6,98 @@ video {
...
+
+.editorial-card .video-container,
+.editorial-card:not(:has(.video-container)) .pause-play-wrapper,
+.hero-marquee:not(:has(.video-container)) .pause-play-wrapper,
+.marquee.split:not(:has(.video-container)) .pause-play-wrapper {
+  width: auto;
+  height: 100%;
+}
+
+.brick .brick-media:not(:has(.video-container)) .pause-play-wrapper {
```
overmyheadandbody, line 72: Couldn't this be included in the selector list above as well? Could we rewrite the whole list using some `:is` statements to make it more straightforward?

libs/blocks/video/video.css
```diff
@@ -6,3 +6,98 @@ video {
...
+
+.hero-marquee .background .video-container {
+  position: inherit;
+}
+
+.video-container .pause-play-wrapper img.accessibility-control {
+  width: auto;
+}
+
+.video-container .pause-play-wrapper img.hidden{
```
overmyheadandbody, line 89:
```suggestion
.video-container .pause-play-wrapper img.hidden {
```

libs/blocks/video/video.css
```diff
@@ -6,3 +6,98 @@ video {
...
+.marquee .background .video-container {
+  display: contents;
+}
+
+@media (min-width: 900px) {
+.media:not(.media-reverse-mobile, .media-reversed) .pause-play-wrapper,
+.aside:not(.split-left) .pause-play-wrapper,
+.hero-marquee.asset-left .pause-play-wrapper {
+  left: 2%;
+}
```
overmyheadandbody, line 102: Indentation should be applied.

overmyheadandbody (commented):
I only got through a few files, I'll continue tomorrow.

As discussed on Slack, we may want to update the way in which we disable the play/pause buttons by using the `#_` pattern, thus accounting for the video auto block as well.

libs/blocks/figure/figure.css
```diff
@@ -69,6 +69,7 @@
   .figure .figure {
     margin: 0 auto;
     max-width: 800px;
+    width: fit-content;
```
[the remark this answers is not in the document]
sharath-kannan: image

But i believe this can be changed to just figure tag inside a figure class.

libs/blocks/figure/figure.js
```diff
@@ -21,6 +21,7 @@ function decorateVideo(clone, figEl) {
   if (anchorTag && !anchorTag.hash) anchorTag.hash = '#autoplay';
   if (anchorTag) decorateAnchorVideo({ src: anchorTag.href, anchorTag });
   if (videoTag) {
+    const videoContainer = clone.querySelector('.video-container') || clone.querySelector('.pause-play-wrapper') || clone.querySelector('video');
```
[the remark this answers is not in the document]
sharath-kannan: The video-container would be play-pause wrapper for hover effect blocks. yes, one selector would suffice.

libs/blocks/hero-marquee/hero-marquee.js
```diff
@@ -183,7 +183,7 @@ export default async function init(el) {
   foreground.classList.add('foreground', `cols-${fRows.length}`);
   let copy = fRows[0];
   const anyTag = foreground.querySelector('p, h1, h2, h3, h4, h5, h6');
-  const asset = foreground.querySelector('div > picture, div > video, div > a[href*=".mp4"], div > a.image-link');
+  const asset = foreground.querySelector('div > picture,div > .video-container,div > .pause-play-wrapper, div > video, div > a[href*=".mp4"], div > a.image-link');
```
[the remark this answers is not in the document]
sharath-kannan: The `.pause-play` wrapper would be the container of the video for hover play videos and this has been designed in such way to make use of the same element while adding events.

libs/blocks/video/video.css
```diff
@@ -6,3 +6,98 @@ video {
   max-width: 100%;
   height: auto;
 }
+
+
+.pause-play-wrapper img {
+  min-height:40px !important;
```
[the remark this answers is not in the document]
sharath-kannan: The `min-height` is set to 150px in marquee.css Line 659  which requires me to set min-height 40px.
!important is required because this overrides my min-height.

libs/blocks/aside/aside.css
```diff
@@ -298,6 +298,7 @@
 }
 
 .aside.rounded-corners .foreground .image img,
+.aside.rounded-corners .foreground .image:not(:has(.video-container)) .pause-play-wrapper,
```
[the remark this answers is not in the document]
sharath-kannan: The video does not have a video-container for hover play videos instead it has an pause-play wrapper.

libs/blocks/video/video.css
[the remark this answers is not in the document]
overmyheadandbody: I'd still advise that we find a way to make the selector more specific or have a specific override in the marquee file, rather than use `!important`

libs/utils/decorate.js
```diff
@@ -325,11 +386,23 @@ export async function loadCDT(el, classList) {
...
 }
 
+export function isAccessible(anchorTag) {
+  return !anchorTag.hash.includes('hide-controls');
+}
+
 export function decorateAnchorVideo({ src = '', anchorTag }) {
   if (!src.length || !(anchorTag instanceof HTMLElement)) return;
+  const accessibilityEnabled = isAccessible(anchorTag);
+  anchorTag.hash = anchorTag.hash.replace('#hide-controls', '');
```
overmyheadandbody, line 396: Could we use the same `#_` pattern here as well? Since we have `#_autoplay`, `#_hoverplay` and so on, we should keep things consistent. It might also be a good idea to remove the anchor in the function itself so the logic stays grouped together.

libs/blocks/hero-marquee/hero-marquee.js
```diff
@@ -183,7 +183,7 @@ export default async function init(el) {
   foreground.classList.add('foreground', `cols-${fRows.length}`);
   let copy = fRows[0];
   const anyTag = foreground.querySelector('p, h1, h2, h3, h4, h5, h6');
-  const asset = foreground.querySelector('div > picture, div > video, div > a[href*=".mp4"], div > a.image-link');
+  const asset = foreground.querySelector('div > picture, div > .video-container, div > .pause-play-wrapper, div > video, div > a[href*=".mp4"], div > a.image-link');
```
overmyheadandbody, line 186: Could we use `:is`, since everything starts with `div > `?

libs/blocks/video/video.css
```diff
@@ -5,4 +5,109 @@ a[href*='.mp4'].hide-video {
...
   max-width: 100%;
   height: auto;
+  object-fit: cover;
+}
+
+:is(.marquee, .aside.split) .pause-play-wrapper img.accessibility-control {
+  min-height: 40px;
+}
+
+.video-container{
```
overmyheadandbody, line 15:
```suggestion
.video-container {
```

libs/blocks/video/video.css
```diff
@@ -5,4 +5,109 @@ a[href*='.mp4'].hide-video {
...
+.pause-play-wrapper {
+  display: flex;
+  width: fit-content;
+  margin: auto;
+}
+
+.video-container .pause-play-wrapper {
+  width: 40px !important;
+  height: 40px !important;
+  display: flex;
```
overmyheadandbody, line 42: This is already defined on L34

libs/utils/decorate.js
```diff
@@ -228,12 +228,73 @@ function getVideoAttrs(hash, dataset) {
...
 
+export function syncPausePlayIcon(video) {
+  const playIcon = video.nextElementSibling?.querySelector('.play-icon');
+  const pauseIcon = video.nextElementSibling?.querySelector('.pause-icon');
+  if (video.paused || video.ended) {
+    playIcon?.classList.remove('hidden');
+    pauseIcon?.classList.add('hidden');
+  } else {
+    pauseIcon?.classList.remove('hidden');
+    playIcon?.classList.add('hidden');
```
overmyheadandbody, line 239: Could we not apply a class to their parent and then show/hide the appropriate child from CSS based on it?

libs/utils/decorate.js
```diff
@@ -228,12 +228,73 @@ function getVideoAttrs(hash, dataset) {
...
+  }
+}
+
+export function addAccessibilityControl(videoString, videoAttributes, tabIndex = 0) {
+  if (!videoAttributes.includes('controls')) {
+    if (videoAttributes.includes('hoverplay')) {
+      return `<a class='pause-play-wrapper' tabindex=${tabIndex} alt='play/pause motion' aria-label='play/pause motion' >${videoString}
```
overmyheadandbody, line 246: Would an `alt` attribute achieve anything here? It's likely that the `aria-label` one is enough. We might also want to adapt this to use a placeholder so we cover all languages.

libs/utils/decorate.js
```diff
@@ -228,12 +228,73 @@ function getVideoAttrs(hash, dataset) {
...
+    </a>`;
+    }
+    return `<div class='video-container'>${videoString}
+  <a class='pause-play-wrapper' role='button' tabindex=${tabIndex} alt='play/pause motion' aria-label='play/pause motion'>
```
overmyheadandbody, line 250: Same `alt` and placeholder comment as above.

libs/utils/decorate.js
```diff
@@ -228,12 +228,73 @@ function getVideoAttrs(hash, dataset) {
...
+    <img class='accessibility-control pause-icon ${videoAttributes.includes('autoplay') ? '' : 'hidden'}' alt='pause icon' src='https://main--federal--adobecom.hlx.page/federal/assets/svgs/accessibility-pause.svg'/>
+    <img class='accessibility-control play-icon ${videoAttributes.includes('autoplay') ? 'hidden' : ''}' alt='play icon' src='https://main--federal--adobecom.hlx.page/federal/assets/svgs/accessibility-play.svg'/>
```
overmyheadandbody, line 252: The `alt` values might also be better suited for placeholders.

libs/utils/decorate.js
```diff
@@ -228,12 +228,73 @@ function getVideoAttrs(hash, dataset) {
...
+      video.setAttribute('data-mouseevent', true);
+    }
+  }
+}
+
+export function applyAccessibiltyEvents(videoEl) {
+  const pausePlayWrapper = videoEl.parentElement.querySelector('.pause-play-wrapper') || videoEl.closest('.pause-play-wrapper');
+  if (pausePlayWrapper?.querySelector('.accessibility-control')) {
+    pausePlayWrapper?.addEventListener('click', handlePause);
+    pausePlayWrapper?.addEventListener('keydown', handlePause);
```
overmyheadandbody, line 294: The `if` clause already checks for the existence of `pausePlayWrapper`, right? So optional chaining might not be needed here as well.

overmyheadandbody (commented):
Thanks for looking into all the changes! I have a new round of comments and a potential bug: when focusing the play/pause button, it's being pushed from its original position:

https://github.com/user-attachments/assets/53922720-e8b6-466f-9f55-dcee17d8cacc

I noticed that in the carousel block, maybe it's not the case everywhere, but it should be investigated.

libs/blocks/video/video.css
sharath-kannan: `display: flex` can be removed.

libs/utils/decorate.js
sharath-kannan: isn't it implement in the same way now?
I am showing/hiding the child that i get from the video's sibling based on css.
could you elaborate a bit more if i have misunderstood this?

overmyheadandbody: What I mean by this is - instead of toggling the children classes
```
<div class="wrapper">
  <img class="play" ... />
  <img class="pause hidden" ... />
</div>
```
you could toggle the parent class and then show/hide the appropriate element from CSS:
```
<div class="wrapper is-playing"> <!-- or 'is-paused' -->
  <img class="play" ... />
  <img class="pause" ... />
</div>
```
That would mean a single class update instead of 2.

libs/utils/decorate.js
```diff
@@ -228,12 +228,80 @@ function getVideoAttrs(hash, dataset) {
...
+  if (!videoAttributes.includes('controls')) {
+    if (videoAttributes.includes('hoverplay')) {
+      return `<a class='pause-play-wrapper video-holder' tabindex=${tabIndex} aria-label='${labels.accessibilityLabel}' >${videoString}
+    </a>`;
+    }
+    return `<div class='video-container video-holder'>${videoString}
+    <a class='pause-play-wrapper' role='button' tabindex=${tabIndex} aria-label=' ${labels.accessibilityLabel}'>
+      <div class='offset-filler'>  
+        <img class='accessibility-control pause-icon ${videoAttributes.includes('autoplay') ? '' : 'hidden'}' alt='${labels.pauseLabel}' src='https://main--federal--adobecom.hlx.page/federal/assets/svgs/accessibility-pause.svg'/>
+        <img class='accessibility-control play-icon ${videoAttributes.includes('autoplay') ? 'hidden' : ''}' alt='${labels.playLabel}' src='https://main--federal--adobecom.hlx.page/federal/assets/svgs/accessibility-play.svg'/>
```
overmyheadandbody, line 258: I think we might still need to use actual placeholders for this, defined in the placeholders sheet and have them translated, so they remain relevant on all locales.

libs/blocks/adobetv/adobetv.css
```diff
@@ -1,4 +1,5 @@
 @import url('../../styles/iframe.css');
+@import url('../video/video.css');
```
overmyheadandbody, line 2: What exactly does adobetv use from the video styles? Doesn't that come with its own style rules OOTB?

libs/blocks/how-to/how-to.js
```diff
@@ -47,7 +47,9 @@ const setJsonLd = (heading, description, mainImage, stepsLd) => {
 };
 
 const getImage = (el) => el.querySelector('picture') || el.querySelector('a[href$=".svg"');
-const getVideo = (el) => el.querySelector('video') || el.querySelector('.milo-video');
+const getVideo = (el) =>{
+  return el.querySelector('.video-container, .pause-play-wrapper, video, .milo-video');
+}
```
overmyheadandbody, line 52: This might cause a Lint error, since in such scenarios it prefers to return directly
```suggestion
const getVideo = (el) => el.querySelector('.video-container, .pause-play-wrapper, video, .milo-video');
```

libs/blocks/video/video.css
```diff
@@ -5,4 +5,132 @@ a[href*='.mp4'].hide-video {
...
+  cursor: pointer;
+}
+
+.video-container .pause-play-wrapper .offset-filler {  
+  display: inherit;
+  justify-content: center;
+  align-items: center;
+  width: 40px;
+  height: 40px;
+  border-radius:inherit;
```
overmyheadandbody, line 59:
```suggestion
  border-radius: inherit;
```

libs/blocks/video/video.css
```diff
@@ -5,4 +5,132 @@ a[href*='.mp4'].hide-video {
...
+}
+
+.video-container .pause-play-wrapper .offset-filler:hover {
+  background: #000;
+  outline: var(--color-accent-focus-ring) solid 2px;
+}
+
+.video-container .pause-play-wrapper:focus .offset-filler {
+  border: 2px solid white;
+  outline: 1px solid white;
```
overmyheadandbody, line 78: I'm not certain I understand the need for both border and outline, are both really needed?

libs/blocks/video/video.css
```diff
@@ -5,4 +5,132 @@ a[href*='.mp4'].hide-video {
...
+:is(.editorial-card, .hero-marquee, .marquee):not(:has(.video-container)) .pause-play-wrapper,
+.editorial-card .video-container {
+  width: auto;
+  height: 100%;
+}
+
+.brick .brick-media:not(:has(.video-container)) .pause-play-wrapper {
+  border-radius: 0;
+  border-top-right-radius: inherit;
+  border-bottom-right-radius: inherit;
```
overmyheadandbody, line 95: Could the video be on the other side? How does this behave on RTL? During my quick check, it would make it lose its border radius values altogether.

libs/utils/decorate.js
```diff
@@ -228,12 +228,80 @@ function getVideoAttrs(hash, dataset) {
...
+
+export function addAccessibilityControl(videoString, videoAttributes, tabIndex = 0) {
+  const labels = {
+    accessibilityLabel: 'play/pause motion',
+    playLabel: 'play icon',
+    pauseLabel: 'pause icon'
+  }
+  if (!videoAttributes.includes('controls')) {
+    if (videoAttributes.includes('hoverplay')) {
+      return `<a class='pause-play-wrapper video-holder' tabindex=${tabIndex} aria-label='${labels.accessibilityLabel}' >${videoString}
```
overmyheadandbody, line 251:
```suggestion
      return `<a class='pause-play-wrapper video-holder' tabindex=${tabIndex} aria-label='${labels.accessibilityLabel}'>${videoString}
```

libs/utils/decorate.js
```diff
@@ -325,11 +393,23 @@ export async function loadCDT(el, classList) {
...
 }
 
+export function isAccessible(anchorTag) {
+  return !anchorTag.hash.includes('_hide-controls');
+}
+
 export function decorateAnchorVideo({ src = '', anchorTag }) {
   if (!src.length || !(anchorTag instanceof HTMLElement)) return;
+  const accessibilityEnabled = isAccessible(anchorTag);
+  anchorTag.hash = anchorTag.hash.replace('#_hide-controls', '');
```
overmyheadandbody, line 403: Might be worth having `_hide-controls` in a variable.

overmyheadandbody (commented):
Thanks for integrating all the changes so far! I have another round of notes. It would also be good to write some unit tests for the changes, as that check is currently failing on the PR.

As discussed in JIRA, we'll likely get a couple more requirements to either update the `aria-label` attribute's value or make use of a new `aria-pressed` one.

libs/blocks/how-to/how-to.js
sharath-kannan: That's right, I used this syntax to debug while writing the functionality and forgot to change it back.

libs/blocks/video/video.css
sharath-kannan: do we have a preferred css formatter(vscode extension) that i can make use of?

libs/blocks/video/video.css
sharath-kannan: yes, this was done to match the figma design for both dark and light mode.
The light mode has a white border of 3px, which is visible on focus along with the blue focus ring.
The dark mode has a 2px white border along with a 1px black outline and then the focus ring.

libs/blocks/adobetv/adobetv.css
sharath-kannan: Adobe tv videos which make use of the `decorateAnchorVideo` function have the accessibility controls but fail to load the css required for it.

libs/blocks/video/video.css
overmyheadandbody: Yes, we're using https://marketplace.visualstudio.com/items?itemName=stylelint.vscode-stylelint

libs/blocks/video/video.css
sharath-kannan: yes, the video can be on the other side.
yes, it would lose the border-radius but would maintain the structure of the brick block.

libs/blocks/hero-marquee/hero-marquee.js
```diff
@@ -183,7 +183,7 @@ export default async function init(el) {
   foreground.classList.add('foreground', `cols-${fRows.length}`);
   let copy = fRows[0];
   const anyTag = foreground.querySelector('p, h1, h2, h3, h4, h5, h6');
-  const asset = foreground.querySelector('div > picture, div > video, div > a[href*=".mp4"], div > a.image-link');
+  const asset = foreground.querySelector('div > picture, :is(.video-container, .pause-play-wrapper), div > video, div > a[href*=".mp4"], div > a.image-link');
```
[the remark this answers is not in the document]
sharath-kannan: Tried this and it brings in some bizarre changes with foreground images/icons.

image

overmyheadandbody (commented):
Before getting into a deeper review, I noticed that there are issues when rendering videos in the `figure` block - https://mwpw-146367--milosh--sharath-kannan.hlx.page/drafts/ramuntea/accessibility/fragments/figure. This should be looked into.

I'd also question the values of the play/pause placeholders, these should just be `Play` and `Pause`, as is also used on https://www.accessibilityoz.com/ozplayer/.

Lastly, I'm not sure the video index works as expected. All videos have a `0` index, the index should be increased for each video added to the page.
@
Screenshot 2024-11-06 at 16 07 25

overmyheadandbody (commented):
As discussed in the meeting, I'll gather up some confirmations from accessibility around index values and whether order is important.

Until then, I did however find that the current strategy does increase the LCP value by quite a lot, we should wait for placeholders just for the play/pause button. So all the `async` statements should not be needed.

rgclayton:
Should the mute/unmute icon be part of this too? I think it makes sense to include both play/pause and mute/unmute controls. 

Similar to what was done on [Max pages](https://www.adobe.com/max/2024/community.html):
image

rgclayton:
What about having this as an opt in so authors can specify if they want to use these buttons instead of the default <video> controls? 

cc: @overmyheadandbody

overmyheadandbody:
To my knowledge, designs and requirements for the moment revolve just around play/pause controls. Let's focus on those and we can think of next steps after we release this MVP.

libs/utils/decorate.js
```diff
@@ -325,20 +393,77 @@ export async function loadCDT(el, classList) {
...
+  const accessibilityEnabled = isAccessible(anchorTag);
+  anchorTag.hash = anchorTag.hash.replace(`#${HIDE_CONTROLS}`, '');
   if (anchorTag.closest('.marquee, .aside, .hero-marquee, .quiz-marquee') && !anchorTag.hash) anchorTag.hash = '#autoplay';
   const { dataset, parentElement } = anchorTag;
-  const video = `<video ${getVideoAttrs(anchorTag.hash, dataset)} data-video-source=${src}></video>`;
+  const attrs = getVideoAttrs(anchorTag.hash, dataset);
+  const tabIndex = anchorTag.tabIndex || 0;
+  const videoIndex = (tabIndex === -1) ? 'tabindex=-1' : '';
+  let video = `<video ${attrs} data-video-source=${src} ${videoIndex}></video>`;
+  videoCounter += 1;
```
overmyheadandbody, line 448: Should we increase the counter only for videos that actually get the controls?

libs/utils/decorate.js
```diff
@@ -228,12 +233,75 @@ function getVideoAttrs(hash, dataset) {
   return `${globalAttrs} controls`;
 }
 
+export function syncPausePlayIcon(video) {
+  if (!video.getAttributeNames().join().includes('hoverplay')) {
+    const offsetFiller = video.closest('.video-holder').querySelector('.offset-filler');
```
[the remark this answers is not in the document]
sharath-kannan: video-holder will alway be there while syncing the play/pause icon.

libs/utils/decorate.js
```diff
@@ -228,12 +233,75 @@ function getVideoAttrs(hash, dataset) {
...
+    const anchorTag = video.closest('.video-holder').querySelector('a');
+    offsetFiller?.classList.toggle('is-playing');
+    const isPlaying = offsetFiller?.classList.contains('is-playing');
+    const indexOfVideo = (anchorTag.getAttribute('video-index') === '1' && videoCounter === 1) ? '' : anchorTag.getAttribute('video-index');
+    const ariaLabel = `${isPlaying ? labels?.pauseMotion : labels?.playMotion} ${indexOfVideo}`.trim();
```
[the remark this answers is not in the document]
sharath-kannan: Adding default values in english.

libs/utils/decorate.js
sharath-kannan: would really matter? since we just need a unique identifier for the video, should we add an extra condition?

vgoodric (approved):
Was asked to review the update to daa-ll values.  And those seem fine to me.

libs/blocks/video/video.css
```diff
@@ -5,4 +5,142 @@ a[href*='.mp4'].hide-video {
...
+  width: fit-content;
+  margin: auto;
+}
+
+.video-container .pause-play-wrapper {
+  width: 45px !important;
+  height: 45px !important;
+  position: absolute;
+  bottom: 2%;
+  right: 2%;
```
narcis-radu, line 44: Did we get any directions for RTL languages? Right now the position is unchanged (bottom right side)

libs/blocks/video/video.css
```diff
@@ -5,4 +5,142 @@ a[href*='.mp4'].hide-video {
...
+}
+
+.pause-play-wrapper {
+  display: flex;
+  width: fit-content;
+  margin: auto;
+}
+
+.video-container .pause-play-wrapper {
+  width: 45px !important;
```
narcis-radu, line 40: I don't think we need `!important` here, specificity is higher than https://github.com/adobecom/milo/pull/3053/files#diff-26ba17abcec57d8f26a6cc6dfb62e44e6357ae3d11e14154c49db7bd68da343bR35

libs/blocks/video/video.css
```diff
@@ -5,4 +5,142 @@ a[href*='.mp4'].hide-video {
...
+  height: 45px !important;
```
narcis-radu, line 41: don't need `!important`

libs/blocks/video/video.css
```diff
@@ -5,4 +5,142 @@ a[href*='.mp4'].hide-video {
...
+  justify-content: center;
+  align-items: center;
+  width: 40px;
+  height: 40px;
+  border-radius: inherit;
+  background: var(--color-gray-800);
+}
+
+:is(.marquee:not(.light), .dark) .video-container .pause-play-wrapper .offset-filler {
+  border: 2px solid white;
```
narcis-radu, line 64: let's use `#fff` instead of `white`

libs/blocks/video/video.css
```diff
@@ -5,4 +5,142 @@ a[href*='.mp4'].hide-video {
...
+  outline: 1px solid #000;
+}
+
+.video-container .pause-play-wrapper .offset-filler:hover {
+  background: #000;
+  outline: var(--color-accent-focus-ring) solid 2px;
+}
+
+.video-container .pause-play-wrapper:focus-visible .offset-filler {
+  border: 2px solid white;
```
narcis-radu, line 77: let's use hex colors

libs/blocks/video/video.css
```diff
@@ -5,4 +5,142 @@ a[href*='.mp4'].hide-video {
...
+  outline: 1px solid white;
```
narcis-radu, line 78: let's use hex please

libs/blocks/video/video.css
```diff
@@ -5,4 +5,142 @@ a[href*='.mp4'].hide-video {
...
+    left: 2%;
+  }
+
+  :is(.section[class*="-up"] .media .foreground .image:first-child, .aside .foreground .image:nth-last-child(1)) 
+  .video-container .pause-play-wrapper {
+    left: auto;
+    right: 2%;
+  }
+
+  [dir="rtl"] :is(.marquee.split:not(.row-reversed), .media:is(.media-reverse-mobile, .media-reversed),
```
narcis-radu, line 140: there's a bit of confusion: 
* split marquee with video on the left -> the play/ pause buttons are on the left
* split marquee with video on the right -> play/ pause buttons are on the right
RTL mode -> no change

Is this the desired layout or a miss?

libs/utils/decorate.js
```diff
@@ -234,12 +244,81 @@ function getVideoAttrs(hash, dataset) {
...
+    offsetFiller?.classList.toggle('is-playing');
+    const isPlaying = offsetFiller?.classList.contains('is-playing');
+    const indexOfVideo = (anchorTag.getAttribute('video-index') === '1' && videoCounter === 1) ? '' : anchorTag.getAttribute('video-index');
+    const changedLabel = `${isPlaying ? videoLabels?.pauseMotion : videoLabels?.playMotion}`;
+    const oldLabel = `${!isPlaying ? videoLabels?.pauseMotion : videoLabels?.playMotion}`;
+    const ariaLabel = `${changedLabel} ${indexOfVideo}`.trim();
+    anchorTag?.setAttribute('aria-label', `${ariaLabel} `);
+    anchorTag?.setAttribute('aria-pressed', isPlaying ? 'true' : 'false');
+    const daaLL = anchorTag.getAttribute('daa-ll');
+    if (daaLL) {
```
narcis-radu, line 260: nit: can be a one-liner

libs/utils/decorate.js
```diff
@@ -234,12 +244,81 @@ function getVideoAttrs(hash, dataset) {
...
+
+export function addAccessibilityControl(videoString, videoAttrs, indexOfVideo, tabIndex = 0) {
+  if (!videoAttrs.includes('controls')) {
+    if (videoAttrs.includes('hoverplay')) {
+      return `<a class='pause-play-wrapper video-holder' tabindex=${tabIndex}>${videoString}</a>`;
+    }
+    return `<div class='video-container video-holder'>${videoString}
+    <a class='pause-play-wrapper' role='button' tabindex=${tabIndex} aria-pressed=true video-index=${indexOfVideo}>
+    <div class='offset-filler ${videoAttrs.includes('autoplay') ? 'is-playing' : ''}'>  
+      <img class='accessibility-control pause-icon' src='https://main--federal--adobecom.hlx.page/federal/assets/svgs/accessibility-pause.svg'/>
```
narcis-radu, line 274: This is going to introduce problems, the URL will not be transformed to load assets from the same domain. See here - https://main--cc--adobecom.hlx.live/products/photoshop?milolibs=video-a11y - the play / pause SVGs are still loaded from hlx.page

libs/utils/decorate.js
```diff
@@ -331,20 +410,84 @@ export async function loadCDT(el, classList) {
...
 }
 
+export function isAccessible(anchorTag) {
+  return !anchorTag.hash.includes(HIDE_CONTROLS);
+}
+
+function updateFirstVideo() {
+  if (firstVideo != null && firstVideo?.controls === false && videoCounter > 1) {
+    let videoHolder = document.querySelector('[video-index="1"]') || firstVideo.closest('.video-holder');
+    if (videoHolder.nodeName !== 'A') {
```
narcis-radu, line 420: nit: one liner

narcis-radu (dismissed):

libs/blocks/figure/figure.js
```diff
@@ -1,4 +1,4 @@
-import { applyHoverPlay, decorateAnchorVideo } from '../../utils/decorate.js';
+import { applyHoverPlay, decorateAnchorVideo, applyAccessibiltyEvents, decoratePausePlayWrapper } from '../../utils/decorate.js';
```
mokimo, line 1: Why are the `applyAccessibiltyEvents, decoratePausePlayWrapper` exported within `decorate.js` when they are only ever used once in the figure block?

libs/utils/decorate.js
```diff
@@ -331,20 +410,84 @@ export async function loadCDT(el, classList) {
...
+    const indexOfVideo = pausePlayWrapper.getAttribute('video-index');
+    const ariaLabel = `${videoAttrs.includes('autoplay') ? videoLabels.pauseMotion : videoLabels.playMotion} ${indexOfVideo === '1' && videoCounter === 1 ? '' : indexOfVideo}`.trim();
+    pausePlayWrapper.setAttribute('aria-label', ariaLabel);
+    pauseIcon.setAttribute('alt', videoLabels.pauseMotion);
+    playIcon.setAttribute('alt', videoLabels.playMotion);
+    updateFirstVideo();
+  }
+}
+
+export function decoratePausePlayWrapper(videoEl, videoAttrs) {
```
mokimo, line 443: I don't think this should live here when it's only used in the figure block.

libs/utils/decorate.js
```diff
@@ -331,20 +410,84 @@ export async function loadCDT(el, classList) {
...
+  if (!videoLabels.hasFetched) {
+    Promise.all([import('../features/placeholders.js'), import('../blocks/global-navigation/utilities/utilities.js')]).then(([{ replaceKeyArray }, { getFedsPlaceholderConfig }]) => {
```
mokimo, line 445: Why do you import `getFedsPlaceholderConfig` from the global-navigation? Should that move into `federated.js` ?

A dependency from `decorate` onto the gnav utilities sounds wrong

libs/blocks/video/video.css
sharath-kannan: yes, The buttons are placed on the left and right based on which side the video is placed no matter the direction of the page.

libs/blocks/video/video.css
sharath-kannan: yes.

libs/utils/decorate.js
```diff
@@ -331,20 +410,84 @@ export async function loadCDT(el, classList) {
...
+  }
+}
+
+function updateAriaLabel(videoEl, videoAttrs) {
+  if (!videoEl.getAttributeNames().includes('data-hoverplay')) {
+    const pausePlayWrapper = videoEl.parentElement.querySelector('.pause-play-wrapper') || videoEl.closest('.pause-play-wrapper');
+    const pauseIcon = pausePlayWrapper.querySelector('.pause-icon');
+    const playIcon = pausePlayWrapper.querySelector('.play-icon');
+    const indexOfVideo = pausePlayWrapper.getAttribute('video-index');
+    const ariaLabel = `${videoAttrs.includes('autoplay') ? videoLabels.pauseMotion : videoLabels.playMotion} ${indexOfVideo === '1' && videoCounter === 1 ? '' : indexOfVideo}`.trim();
```
JasonHowellSlavin, line 435: Personally not a fan of nested ternary, Should set off the linter. Can we unpack this?

test/blocks/video/video.test.js
```diff
@@ -4,20 +4,45 @@ import sinon from 'sinon';
...
     document.body.innerHTML = '';
   });
 
+  it('aria-label should not have index when page has only one video', async () => {
+    const block = document.querySelector('.video.autoplay.single');
+    const block2 = document.querySelector('.video.autoplay.second');
+    const a = block.querySelector('a');
+    const a2 = block2.querySelector('a');
+    block.append(a);
+    block2.append(a2);
```
JasonHowellSlavin, line 33: looks like `document.querySelector('.video.autoplay.single')` is a single div with an `a` tag inside, why are you appending the item to the same block it is found in? 

Also, arguably, we should stick to semantic naming conventions even in our tests.

test/blocks/video/video.test.js
```diff
@@ -114,6 +139,88 @@ describe('video uploaded using franklin bot', () => {
...
+    });
+
+    const a = block.querySelector('a');
+    block.append(a);
+    init(a);
+    const video = block.querySelector('video');
+    decoratePausePlayWrapper(video, '');
+    const pausePlayWrapper = block.querySelector('.pause-play-wrapper');
+    pausePlayWrapper.click();
+    await new Promise((resolve) => { setTimeout(resolve, 1000); });
```
JasonHowellSlavin, line 177: A little worried for a timeout of a second, but I understand if it is necessary, just will add time to our entire test suite.

libs/blocks/aside/aside.css
```diff
@@ -1074,6 +1075,10 @@
     max-width: 1396px;
   }
 
+  .aside.split .video-container .pause-play-wrapper img.accessibility-control {
+    width: auto;
+  }
+
```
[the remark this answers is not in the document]
sharath-kannan: These are added with a more specific rule to override the css mentioned in a generic way else where.(in this case any of `img` rule specified in aside.css)

libs/blocks/brick/brick.css
```diff
@@ -342,6 +346,12 @@
     position: absolute;
   }
 
+  .brick.split.row .foreground .brick-media .video-container img,
+  .brick.split.row .foreground .brick-media .video-container video {
+    width: 100%;
+    height: 100%;
```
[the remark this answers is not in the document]
sharath-kannan: Same answer as above.

libs/blocks/figure/figure.js
sharath-kannan: They are used in decorate.js L:486 as well.

sharath-kannan: They are used within `decorate.js` as well L: 484 and L:486.

libs/utils/decorate.js
```diff
@@ -331,20 +408,83 @@ export async function loadCDT(el, classList) {
   }
 }
 
+export function isAccessible(anchorTag) {
```
overmyheadandbody, line 411: I still have a general naming issue with the methods in this file. The file itself tailors to multiple types of content, including videos. `isAccessible` is relevant just for videos, but it could very well suggest it can apply to any other blocks. I'd say we need to better separate video logic as such through some better namings.

overmyheadandbody (commented):
Thanks @sharath-kannan for integrating all the changes!

And thanks @narcis-radu, @robert-bogos, @mokimo, @JasonHowellSlavin for your reviews! To clean up the PR page a bit, I've tried to go through all the comments and mark them as resolved if I deemed them as such, however I did leave a few open that were not immediately obvious to me, it would be great if you could have a look over those. And if you can spare the time, one more round of reviews would also be extremely helpful 😀

libs/blocks/carousel/carousel.js
```diff
@@ -230,10 +230,10 @@ function moveSlides(event, carouselElements, jumpToIndex) {
       if (index < show) {
         tabIndex = 0;
       }
-      slide.querySelectorAll('a').forEach((focusableElement) => { focusableElement.setAttribute('tabindex', tabIndex); });
+      slide.querySelectorAll('a,:not(.video-container, .pause-play-wrapper) > video').forEach((focusableElement) => { focusableElement.setAttribute('tabindex', tabIndex); });
     });
   } else {
-    activeSlide.querySelectorAll('a').forEach((focusableElement) => { focusableElement.setAttribute('tabindex', 0); });
+    activeSlide.querySelectorAll('a,:not(.video-container, .pause-play-wrapper) > video').forEach((focusableElement) => { focusableElement.setAttribute('tabindex', 0); });
```
overmyheadandbody, line 236: L233 and L236 seem to exceed the 100 character limit, it might be a good idea to move the `.forEach` on a new line.

libs/utils/decorate.js
```diff
@@ -234,12 +245,78 @@ function getVideoAttrs(hash, dataset) {
...
+    return `<a class='pause-play-wrapper video-holder' tabindex=${tabIndex}>${videoString}</a>`;
+  }
+  return `<div class='video-container video-holder'>${videoString}
+    <a class='pause-play-wrapper' role='button' tabindex=${tabIndex} aria-pressed=true video-index=${indexOfVideo}>
+    <div class='offset-filler ${videoAttrs.includes('autoplay') ? 'is-playing' : ''}'>  
+      <img class='accessibility-control pause-icon' src='${fedRoot}/federal/assets/svgs/accessibility-pause.svg'/>
+      <img class='accessibility-control play-icon' src='${fedRoot}/federal/assets/svgs/accessibility-play.svg'/>
+    </div>
+  </a>
+</div>`;
```
overmyheadandbody, line 278: This might need an additional indent

overmyheadandbody (approved):
Approving, since we had a lot of feedback on this one and things are looking a lot better now. We're counting on QEs to isolate any additional issues; if none ar found, we should socialize this again, so it gets the required approvals for it to receive the "Ready for Stage" label.

spadmasa:
Validation has been done on the pr , results are added in the ticket subtask https://jira.corp.adobe.com/browse/MWPW-146367 
https://main--cc--adobecom.hlx.live/products/photoshop?milolibs=mwpw-146367--milosh--sharath-kannan
{01EB975A-3729-49B2-9A9D-2C677A58A54E}
{8457F12F-D17D-4B59-930F-A1FD695A68FE}
{89E5D633-3813-47D2-BEE8-12CE19AF4EE2}
{4264779E-B4D2-4490-B13F-E218D6A1710F}
{DB588E88-4DE0-46ED-86E4-3411D2BA6657}
{CD786672-BD90-4957-977D-0EA279BB7983}
https://main--cc--adobecom.hlx.page/drafts/souj/accessibility/marqueecontrols?milolibs=mwpw-146367--milosh--sharath-kannan
{EAD7D6D8-A084-4713-932E-D9D0A367FB54}

{F6F7B023-8EAD-45BC-9023-855F18847B89}

salonijain3:
@spadmasa I got an alert for gnav file changes on this, can you check if there is any impact from this PR?

cc: @sharmrj @bandana147

sharath-kannan:
> @spadmasa I got an alert for gnav file changes on this, can you check if there is any impact from this PR?
> 
> cc: @sharmrj @bandana147

> @spadmasa I got an alert for gnav file changes on this, can you check if there is any impact from this PR?
> 
> cc: @sharmrj @bandana147

@salonijain3 the gnav files were modified to move a commonly used function out of the gnav module.

spadmasa:
https://main--cc--adobecom.hlx.page/drafts/souj/accessibility/marqueecontrols?milolibs=mwpw-146367--milosh--sharath-kannan
{327ABDC3-2454-4A7B-9349-C5BFAD90B145}
{09E115FC-6BA8-4F0F-AE93-A6E5F88C2EAB}
image
{11034737-0254-40A1-9942-4E9CC611DB8F}
{4C1EAC0D-0AEE-4A78-B8A2-CA134A8C5AC3}
{0AE20BEF-0D55-4B3E-BFAE-D4BB06156800}
{59CB7354-46CF-4DC3-B99B-6338D0593FA7}

robert-bogos (approved):

mokimo (approved):

PR #3055, opened 2024-10-16
MWPW-160156 - Add option to define mode (async, defer) when loading script

Ruchika4:
- Add option to define mode (async, defer) when loading script

Resolves: [MWPW-160156](https://jira.corp.adobe.com/browse/MWPW-160156)

Test URLs:

Milo:
Before: https://stage--milo--adobecom.hlx.page/drafts/ruchika/document
After: https://MWPW-160156--milo--adobecom.hlx.page/drafts/ruchika/document

Unity:
Before: https://stage--dc--adobecom.hlx.page/acrobat/online/test/sign-pdf-projectunity?unitylibs=MWPW160156
After: https://stage--dc--adobecom.hlx.page/acrobat/online/test/sign-pdf-projectunity?unitylibs=MWPW160156&milolibs=MWPW-160156

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -395,7 +395,7 @@ export function appendHtmlToLink(link) {
   }
 }
 
-export const loadScript = (url, type) => new Promise((resolve, reject) => {
+export const loadScript = (url, type, mode) => new Promise((resolve, reject) => {
   let script = document.querySelector(`head > script[src="${url}"]`);
   if (!script) {
     const { head } = document;
@@ -403,6 +403,9 @@ export const loadScript = (url, type) => new Promise((resolve, reject) => {
     script.setAttribute('src', url);
     if (type) {
       script.setAttribute('type', type);
+    }
+    if (mode) {
+      script.setAttribute(mode, true);
     }
     head.append(script);
   }


libs/utils/utils.js
```diff
@@ -395,7 +395,7 @@ export function appendHtmlToLink(link) {
   }
 }
 
-export const loadScript = (url, type) => new Promise((resolve, reject) => {
+export const loadScript = (url, type, mode) => new Promise((resolve, reject) => {
```
mokimo, line 398: Would have been better from the start to use an object, but we can at least start using one going forward rather than having this method with 5 params in the future
```suggestion
export const loadScript = (url, type, { mode } = {}) => new Promise((resolve, reject) => {
```

mokimo (approved):
Consider the feedback, but lgtm

libs/utils/utils.js
```diff
@@ -404,6 +404,9 @@ export const loadScript = (url, type) => new Promise((resolve, reject) => {
     if (type) {
       script.setAttribute('type', type);
     }
+    if (mode) {
+      script.setAttribute(mode, true);
+    }
```
robert-bogos, line 409: Nit: This can be a 1 liner
```suggestion
    if (mode) script.setAttribute(mode, true);
```

robert-bogos (approved):

libs/utils/utils.js
```diff
@@ -404,6 +404,9 @@ export const loadScript = (url, type) => new Promise((resolve, reject) => {
     if (type) {
       script.setAttribute('type', type);
     }
+    if (mode) {
```
narcis-radu, line 407: I would like us to be more precise with the attributes we're setting here.
```suggestion
    if (['async', 'defer'].includes(mode)) script.setAttribute(mode, true);
```

sukamat (approved):

overmyheadandbody (approved):

PR #3056, opened 2024-10-16
Added performance tab to the Preflight tool (#2773)

mokimo:
### Description
A few todo's here and there but the features and flow should work and this should be ready for feedback and a first batch of testing.

- Adds a new performance preflight tab
- adds new&polished performance docs: https://main--milo--adobecom.aem.page/docs/authoring/performance/ (Needs a bit more polishing and hands-on-advice)

### Look & Feel 
Screenshot 2024-10-17 at 01 08 39

Resolves: [MWPW-157715](https://jira.corp.adobe.com/browse/MWPW-157715)

### Test instructions
Use the provided sample URLs, try to destroy the thing, close it, open it, open it on incognito, have pages with LCP, without, use production pages from bacom etc. to test this. I've encountered a lot of issues during development and I'm sure theres still many issues out there.

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/osahin/preflight-performance/marquee-fragment?martech=off
- After: https://preflight-performance-tab--milo--mokimo.hlx.page/drafts/osahin/preflight-performance/marquee-fragment?martech=off

### DIFF
--- a/libs/blocks/preflight/panels/performance.js
+++ b/libs/blocks/preflight/panels/performance.js
@@ -1,0 +1,272 @@
+import { html, signal, useEffect } from '../../../deps/htm-preact.js';
+import { getMetadata } from '../../../utils/utils.js';
+
+const icons = {
+  pass: 'green',
+  fail: 'red',
+  empty: 'empty',
+};
+// TODO MEP/Personalization
+// TODO mobile / tablet?
+// TODO create ticket for PSI API
+// TODO link to documentation directly within the sections
+const text = {
+  lcpEl: {
+    key: 'lcpEl',
+    title: 'Valid LCP',
+    passed: { description: 'Valid LCP in the first section detected.' },
+    failed: { description: 'No LCP image or video in the first section detected.' },
+  },
+  singleBlock: {
+    key: 'singleBlock',
+    title: 'Single Block',
+    passed: { description: 'First section has exactly one block.' },
+    failed: { description: 'First section has more than one block.' },
+  },
+  imageSize: {
+    key: 'imageSize',
+    title: 'Images size',
+    empty: { description: 'No image as LCP element.' },
+    passed: { description: 'LCP image is less than 100KB.' },
+    failed: { description: 'LCP image is over 100KB.' },
+  },
+  videoPoster: {
+    key: 'videoPoster',
+    title: 'Videos',
+    empty: { description: 'No video as LCP element.' },
+    passed: { description: 'LCP video has a poster attribute' },
+    failed: { description: 'LCP video has no poster attribute.' },
+  },
+  fragments: {
+    key: 'fragments',
+    title: 'Fragments',
+    passed: { description: 'No fragments used within the LCP section.' },
+    failed: { description: 'Fragments used within the LCP section.' },
+  },
+  personalization: {
+    key: 'personalization',
+    title: 'Personalization',
+    passed: { description: 'Personalization is currently not enabled.' },
+    failed: { description: 'MEP or Target enabled.' },
+  },
+  placeholders: {
+    key: 'placeholders',
+    title: 'Placeholders',
+    passed: { description: 'No placeholders found within the LCP section.' },
+    failed: { description: 'Placeholders found within the LCP section.' },
+  },
+  icons: {
+    key: 'icons',
+    title: 'Icons',
+    passed: { description: 'No icons found within the LCP section.' },
+    failed: { description: 'Icons found within the LCP section.' },
+  },
+};
+
+export const config = {
+  items: signal([
+    ...Object.values(text).map(({ key, title }) => ({
+      key,
+      title,
+      icon: icons.empty,
+      description: 'Loading...',
+    })),
+  ]),
+  lcp: null,
+  cls: 0,
+};
+
+export const findItem = (key) => config.items.value.find((item) => item.key === key);
+export const updateItem = ({ key, ...updates }) => {
+  const { items } = config;
+  items.value = items.value.map((item) => (item.key === key ? { ...item, ...updates } : item));
+};
+
+export function conditionalItemUpdate({ emptyWhen, failsWhen, key }) {
+  if (emptyWhen) {
+    updateItem({
+      key,
+      icon: icons.empty,
+      description: text[key].empty.description,
+    });
+    return;
+  }
+  if (failsWhen) {
+    updateItem({
+      key,
+      icon: icons.fail,
+      description: text[key].failed.description,
+    });
+    return;
+  }
+  updateItem({
+    key,
+    icon: icons.pass,
+    description: text[key].passed.description,
+  });
+}
+
+// TODO do we also want to check the content-length header?
+// https://www.w3.org/TR/largest-contentful-paint/#largest-contentful-paint-candidate-element
+// candidate’s element is a text node, or candidate’s request's response's content length
+// in bytes is >= candidate’s element's effective visual size * 0.004
+export async function checkImageSize() {
+  const { lcp } = config;
+  const hasValidImage = lcp?.url && !lcp.url.match('media_.*.mp4');
+  let blob;
+  let isSizeValid;
+  if (hasValidImage) {
+    blob = await fetch(lcp.url).then((res) => res.blob());
+    isSizeValid = blob.size / 1000 <= 100;
+  }
+
+  conditionalItemUpdate({
+    failsWhen: !isSizeValid,
+    emptyWhen: !hasValidImage,
+    key: text.imageSize.key,
+  });
+}
+
+export function checkLCP() {
+  const { lcp } = config;
+  const mainSection = document.querySelector('main > div.section');
+  const validLcp = lcp && lcp.element && lcp.url && mainSection?.contains(lcp.element);
+  conditionalItemUpdate({
+    failsWhen: !validLcp,
+    key: text.lcpEl.key,
+    icon: icons.pass,
+    description: text.lcpEl.passed.description,
+  });
+  return validLcp;
+}
+
+export const checkFragments = () => conditionalItemUpdate({
+  failsWhen: config.lcp.element.closest('.fragment') || config.lcp.element.closest('.section')?.querySelector('[data-path*="fragment"]'),
+  key: text.fragments.key,
+});
+
+export const checkPlaceholders = async () => conditionalItemUpdate({
+  failsWhen: config.lcp.element.closest('.section').dataset.placeholdersDecorated === 'true',
+  key: text.placeholders.key,
+});
+
+export const checkForPersonalization = async () => conditionalItemUpdate({
+  failsWhen: getMetadata('personalization') || getMetadata('target') === 'on',
+  key: text.personalization.key,
+});
+
+export const checkVideosWithoutPosterAttribute = async () => conditionalItemUpdate({
+  failsWhen: !config.lcp.element.poster,
+  emptyWhen: !config.lcp.url.match('media_.*.mp4'),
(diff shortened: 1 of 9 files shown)


mokimo:
A few more test links https://preflight-performance-tab--milo--mokimo.hlx.page/drafts/osahin/preflight-performance/icons
https://preflight-performance-tab--milo--mokimo.hlx.page/drafts/osahin/preflight-performance/marquee-fragment
https://preflight-performance-tab--milo--mokimo.hlx.page/drafts/osahin/preflight-performance/marquee-fragment-inline
https://preflight-performance-tab--milo--mokimo.hlx.page/drafts/osahin/preflight-performance/mep
https://preflight-performance-tab--milo--mokimo.hlx.page/drafts/osahin/preflight-performance/placeholders
https://preflight-performance-tab--milo--mokimo.hlx.page/drafts/osahin/preflight-performance/no-lcp

libs/blocks/preflight/panels/performance.js
```diff
@@ -0,0 +1,272 @@
...
+      config.lcp = lastEntry;
+    }
+    const hasValidLcp = checkLCP();
+    if (!hasValidLcp) return;
+    checkFragments();
+    checkForPersonalization();
+    checkVideosWithoutPosterAttribute();
+    checkIcons();
+    checkForSingleBlock();
+    Promise.all([checkImageSize(), checkPlaceholders()]);
```
mokimo, line 234: This is a bit of a funky case.
- The hover for the element will be obviously
- None of the other checks are relevant if we don't have a good LCP candidate to begin with
Screenshot 2024-10-17 at 01 24 12

libs/blocks/preflight/panels/performance.js
```diff
@@ -0,0 +1,277 @@
...
+      icon: icons.fail,
+      description: text[key].failed.description,
+    });
+    return;
+  }
+  updateItem({
+    key,
+    icon: icons.pass,
+    description: text[key].passed.description,
+  });
```
overmyheadandbody, line 106: This feels a bit repetitive. Could we go for something like
```suggestion
  const icon = (emptyWhen && icons.empty) || (failsWhen && icons.fail) || icons.pass;
  const descriptionKey = (emptyWhen && 'empty') || (failsWhen && 'failed') || 'passed';
  updateItem({
    key,
    icon,
    description: text[key][descriptionKey].description,
  });
```

Haven't tested this, some changes might be needed.

libs/blocks/preflight/panels/performance.js
```diff
@@ -0,0 +1,277 @@
...
+  conditionalItemUpdate({
+    failsWhen: !isSizeValid,
+    emptyWhen: !hasValidImage,
+    key: text.imageSize.key,
+  });
+}
+
+export function checkLCP() {
+  const { lcp } = config;
+  const mainSection = document.querySelector('main > div.section');
```
overmyheadandbody, line 132: Would `firstSection` describe this more accurately?

libs/blocks/preflight/panels/performance.js
```diff
@@ -0,0 +1,277 @@
...
+  const validLcp = lcp && lcp.element && lcp.url && mainSection?.contains(lcp.element);
```
overmyheadandbody, line 133: Nit:
```suggestion
  const validLcp = lcp?.element && lcp?.url && mainSection?.contains(lcp.element);
```

libs/blocks/preflight/panels/performance.js
```diff
@@ -0,0 +1,277 @@
...
+    failsWhen: !validLcp,
+    key: text.lcpEl.key,
+    icon: icons.pass,
+    description: text.lcpEl.passed.description,
+  });
+  return validLcp;
+}
+
+export const checkFragments = () => conditionalItemUpdate({
+  failsWhen: config.lcp.element.closest('.fragment') || config.lcp.element.closest('.section')?.querySelector('[data-path*="fragment"]'),
```
overmyheadandbody, line 144: Is this taking deeper nested fragments into account?

libs/blocks/preflight/panels/performance.js
```diff
@@ -0,0 +1,277 @@
...
+  key: text.fragments.key,
+});
+
+export const checkPlaceholders = async () => conditionalItemUpdate({
```
overmyheadandbody, line 148: Does this need to be `async`? Same question for the next two methods.

libs/blocks/preflight/panels/performance.js
```diff
@@ -0,0 +1,277 @@
...
+  const clsObserver = new PerformanceObserver((entryList) => {
+    const entries = entryList.getEntries();
+    entries.forEach((entry) => {
+      if (!entry.hadRecentInput) {
+        config.cls += entry.value;
+      }
+    });
+    if (config.cls > 0) {
+      // TODO - Lana log? We should not have any CLS.
+      // console.log('CLS:', cls);
```
overmyheadandbody, line 253: Should we remove this console log for now, since we have a to do?

libs/blocks/preflight/panels/performance.js
```diff
@@ -0,0 +1,277 @@
...
+  });
+  clsObserver.observe({ type: 'layout-shift', buffered: true });
+}
+
+export function Panel() {
+  useEffect(() => observePerfMetrics(), []);
+
+  return html`
+    <div class="seo-columns">
+      <div class="seo-column">${config.items.value.slice(0, 4).map((item) => createPerformanceItem(item))}</div>
```
overmyheadandbody, line 264: `seo-column` may need to be renamed, since its properties are now used for this panel as well.

libs/utils/utils.js
```diff
@@ -809,6 +809,7 @@ async function decoratePlaceholders(area, config) {
   if (!area) return;
   const nodes = findReplaceableNodes(area);
   if (!nodes.length) return;
+  area.dataset.placeholdersDecorated = 'true';
```
overmyheadandbody, line 812: This seems to be added before the placeholders are decorated. Someone from the outside might hook onto this thinking that placeholders are replaced once this class is added. I'd suggest we either change the name to show that placeholders are used in the area; or move the declaration after the placeholders have actually been replaced.

overmyheadandbody (commented):
I like the general direction of this, good job! I do, however, have a few notes 😅

libs/blocks/preflight/panels/performance.js
```diff
@@ -0,0 +1,277 @@
...
+
+const icons = {
+  pass: 'green',
+  fail: 'red',
+  empty: 'empty',
+};
+// TODO MEP/Personalization
+// TODO mobile / tablet?
+// TODO create ticket for PSI API
+// TODO link to documentation directly within the sections
```
mokimo, line 12: todo, we could also check for commerce within the marquee

libs/blocks/preflight/panels/performance.js
```diff
@@ -0,0 +1,277 @@
...
+
+const removeHighlight = () => { document.querySelector('.lcp-tooltip-modal').classList.remove('show'); };
+
+function observePerfMetrics() {
+  // Observe LCP
+  const lcpObserver = new PerformanceObserver((entryList) => {
+    const entries = entryList.getEntries();
+    const lastEntry = entries[entries.length - 1];
+    if (lastEntry) config.lcp = lastEntry;
+    const hasValidLcp = checkLCP();
```
narcis-radu, line 226: no need for the constant

libs/blocks/preflight/panels/performance.js
```diff
@@ -0,0 +1,277 @@
...
+    clonedLcpSection.style.height = `${lcpSection.offsetHeight}px`;
+    clonedLcpSection.style.transform = `scale(${scaleFactor})`;
+    clonedLcpSection.style.transformOrigin = 'top left';
+  }
+  if (!tooltip.children.length) tooltip.appendChild(clonedLcpSection);
+  const { top, left } = event.currentTarget.getBoundingClientRect();
+  tooltip.style.width = `${offsetWidth * scaleFactor}px`;
+  tooltip.style.height = `${offsetHeight * scaleFactor}px`;
+  tooltip.style.top = `${top + window.scrollY - offsetHeight * scaleFactor - 10}px`;
+  tooltip.style.left = `${left + window.scrollX}px`;
```
robert-bogos, line 214: I will echo Narcis and point out that we can use `Object.assign` in here too.
```suggestion
Object.assign(tooltip.style, {
  width: `${offsetWidth * scaleFactor}px`,
  height: `${offsetHeight * scaleFactor}px`,
  top: `${top + window.scrollY - offsetHeight * scaleFactor - 10}px`,
  left: `${left + window.scrollX}px`
});
```

libs/blocks/preflight/panels/performance.js
```diff
@@ -0,0 +1,277 @@
...
+// https://www.w3.org/TR/largest-contentful-paint/#largest-contentful-paint-candidate-element
+// candidate’s element is a text node, or candidate’s request's response's content length
+// in bytes is >= candidate’s element's effective visual size * 0.004
+export async function checkImageSize() {
+  const { lcp } = config;
+  const hasValidImage = lcp?.url && !lcp.url.match('media_.*.mp4');
+  let blob;
+  let isSizeValid;
+  if (hasValidImage) {
+    blob = await fetch(lcp.url).then((res) => res.blob());
```
robert-bogos, line 119: Should add a `try catch` to handle a possible fetching fail?

libs/blocks/preflight/panels/performance.js
```diff
@@ -0,0 +1,277 @@
...
+}
+
+export function checkLCP() {
+  const { lcp } = config;
+  const mainSection = document.querySelector('main > div.section');
+  const validLcp = lcp && lcp.element && lcp.url && mainSection?.contains(lcp.element);
+  conditionalItemUpdate({
+    failsWhen: !validLcp,
+    key: text.lcpEl.key,
+    icon: icons.pass,
```
robert-bogos, line 137: I see that the icon parameter is passed but is not used in the function’s logic

libs/blocks/preflight/panels/performance.js
mokimo: Cool suggestion, will add in the next commit 👍

libs/blocks/preflight/panels/performance.js
mokimo: Yes, but just in the sense of, it tells you that there is a fragment being used: https://preflight-performance-tab--milo--mokimo.hlx.page/drafts/osahin/preflight-performance/marquee-deep-nested-fragments

... we could also surface nested fragments, e.g. "amount of fragments within fragments, 5" 🤔 
Wdyt?

libs/blocks/preflight/panels/performance.js
mokimo: We only fetch when it has a valid image... but for the sake of simplicity, let's assume if it can't fetch an image - we do in fact *not* have a valid image.

libs/blocks/preflight/panels/performance.js
overmyheadandbody: I've heard of 4 levels deep nested fragments, so I think we should put this on the idea list, yes.

overmyheadandbody (approved):
Thanks for addressing all the feedback!

robert-bogos (approved):
Thank you for addressing the feedback! 🙏🏻

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-157715

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-157715

PR #3060, opened 2024-10-17
MWPW-160360 - Decorate standalone text with body default styles

Sartxi:
When standalone content is not decorated according to its format or a text style override, the decorate text function will add default body styling like it does with other empty elements.

This will apply default styles to text in consonant blocks that use the shared text decorate functionality.

Resolves: [MWPW-160360](https://jira.corp.adobe.com/browse/MWPW-160360)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/sarchibeque/format-text?martech=off
- After: https://sartxi-mwpw-160360-text--milo--adobecom.hlx.page/drafts/sarchibeque/format-text?martech=off

- Before: https://main--milo--adobecom.hlx.page/drafts/sarchibeque/standalone-text?martech=off
- After: https://sartxi-mwpw-160360-text--milo--adobecom.hlx.page/drafts/sarchibeque/standalone-text?martech=off


libs/utils/utils.js
```diff
@@ -260,6 +260,14 @@ export function isInTextNode(node) {
   return node.parentElement.firstChild.nodeType === Node.TEXT_NODE;
 }
 
+export function elContainsText(el) {
```
mokimo, line 263: If this method is only used in decorate and only a specific method within decorate, should it stay in there?

Sartxi: I guess I thought it was generic enough that could be used elsewhere in the future, I guess if that happens we can move it to utils then. For now I will move back to decorate. Thanks

libs/utils/decorate.js
```diff
@@ -77,13 +85,12 @@ export function decorateBlockText(el, config = ['m', 's', 'm'], type = null) {
         decorateIconArea(el);
       }
     }
-    const emptyEls = el.querySelectorAll('p:not([class]), ul:not([class]), ol:not([class])');
+    const bodyStyle = `body-${config[1]}`;
+    const emptyEls = el.querySelectorAll('p:not([class]), ul:not([class]), ol:not([class], div:not([class]))');
```
overmyheadandbody, line 89: Could we benefit from some `:is` magic?
```suggestion
    const emptyEls = el.querySelectorAll(':is(p, ul, ol, div):not([class])');
```
Anyway, I think there may be a typo here, as `ol:not([class]` doesn't get closed, but is chained to the `div` as well, might deserve another look.

libs/utils/decorate.js
```diff
@@ -65,6 +65,14 @@ export function decorateIconArea(el) {
   });
 }
 
+function elContainsText(el) {
+  return !![...el.childNodes].filter((childNode) => {
+    const { innerText, nodeType, textContent } = childNode;
+    if (nodeType === Node.ELEMENT_NODE) return innerText.trim() !== '';
+    return nodeType === Node.TEXT_NODE && textContent.trim() !== '';
+  }).length;
+}
```
mokimo, line 74: This can be simplified, something along the lines of:
```suggestion
const elContainsText = el => 
  [...el.childNodes].some((childNode) => 
    (childNode.nodeType === Node.ELEMENT_NODE && childNode.innerText.trim() !== '') ||
    (childNode.nodeType === Node.TEXT_NODE && childNode.textContent.trim() !== '')
  );
```

Sartxi: Thanks, yeah I like that. I have updated it to use the `some`

mokimo (approved):
there's some nala tests failing, other than that - lgtm

robert-bogos (commented):
I see units tests and also nala tests are failing.
Can you have a look at that @Sartxi ?

For the failing nala tests you might try [to purge the cached assets](https://adobe-mwp.slack.com/archives/C03HAU5SU2E/p1727941650168049).

Sartxi:
> I see units tests and also nala tests are failing. Can you have a look at that @Sartxi ?
> 
> For the failing nala tests you might try [to purge the cached assets](https://adobe-mwp.slack.com/archives/C03HAU5SU2E/p1727941650168049).

I tried purging the cache but it didn't seem to work :(

overmyheadandbody (approved):

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-160360

robert-bogos (approved):
I would still try to find the reason why the Nala tests are failing 🤔

PR #3062, opened 2024-10-17
Limit maximum number of PRs per batch

mokimo:
### Description
Limits the number of PRs that are getting merged into a batch. We see around ~100 PRs a month and if we get 20 batches out per month that would result in 160 merges, to account for "backlogs" as we currently have it. 

Of the currently 62 raised PRs, 23 are ready for stage and we would merge those within 3 days, which sounds reasonable. Additionally, you can "skip the queue" of older PRs, by providing the `high priority` label if your work is blocking a release.
 
Screenshot 2024-10-17 at 20 00 48

Resolves: [MWPW-157289](https://jira.corp.adobe.com/browse/MWPW-157289)

**Test URLs:**
- After: https://improve-rcp-handling--milo--mokimo.hlx.page/?martech=off


thedoc31 (approved):

.github/workflows/merge-to-stage.js
```diff
@@ -9,6 +9,8 @@ const {
 const PR_TITLE = '[Release] Stage to Main';
 const SEEN = {};
 const REQUIRED_APPROVALS = process.env.REQUIRED_APPROVALS || 2;
+const MAX_MERGES = process.env.MAX_PRS_PER_BATCH || 8;
+let existingPRCount = 0
```
narcis-radu, line 13: `;`

.github/workflows/merge-to-stage.js
```diff
@@ -130,6 +132,7 @@ const merge = async ({ prs, type }) => {
 
   for await (const { number, files, html_url, title } of prs) {
     try {
+      if (mergeLimitExceeded()) return
```
narcis-radu, line 135: `;`

.github/workflows/merge-to-stage.js
```diff
@@ -150,6 +153,7 @@ const merge = async ({ prs, type }) => {
           merge_method: 'squash',
         });
       }
+      existingPRCount++
```
narcis-radu, line 156: `;`

.github/workflows/merge-to-stage.js
```diff
@@ -231,6 +235,8 @@ const openStageToMainPR = async () => {
   }
 };
 
+const mergeLimitExceeded = () => (MAX_MERGES - existingPRCount) < 0
```
narcis-radu, line 238: `;`

.github/workflows/merge-to-stage.js
```diff
@@ -241,6 +247,11 @@ const main = async (params) => {
     const stageToMainPR = await getStageToMainPR();
     console.log('has Stage to Main PR:', !!stageToMainPR);
     if (stageToMainPR) body = stageToMainPR.body;
+    existingPRCount = body.match(/https:\/\/github\.com\/adobecom\/milo\/pull\/\d+/g).length
```
narcis-radu, line 250: `;`

.github/workflows/merge-to-stage.js
overmyheadandbody, line 135: Couldn't we exit even earlier in the `merge` function?

.github/workflows/merge-to-stage.js
overmyheadandbody, line 156: Should this be brought into the `if (!process.env.LOCAL_RUN) {` conditional?

.github/workflows/merge-to-stage.js
```diff
@@ -9,6 +9,8 @@ const {
 const PR_TITLE = '[Release] Stage to Main';
 const SEEN = {};
 const REQUIRED_APPROVALS = process.env.REQUIRED_APPROVALS || 2;
+const MAX_MERGES = process.env.MAX_PRS_PER_BATCH || 8;
```
overmyheadandbody, line 12: I don't quite remember how we landed on `8`. How do you suggest we update the value should we need to adapt it?

overmyheadandbody (commented):
A few notes on top of the ones from @narcis-radu

.github/workflows/merge-to-stage.js
mokimo: We did a reasonable estimate based on the average of PRs we see +2

To adapt the value, we can add a [secret](https://github.com/adobecom/milo/settings/secrets/actions) and can just update that, rather than having to make a PR

.github/workflows/merge-to-stage.js
mokimo: Ran a proper linter over the whole file

.github/workflows/merge-to-stage.js
mokimo: Ran a proper linter over the whole file

.github/workflows/merge-to-stage.js
mokimo: Ran a proper linter over the whole file

.github/workflows/merge-to-stage.js
mokimo: Ran a proper linter over the whole file

.github/workflows/merge-to-stage.js
mokimo: No this should be fine - since you want to emulate the process "merging" PRs and stopping execution once the max is reached on a local run. Just without actually merging anything.

.github/workflows/merge-to-stage.js
mokimo: I considered this but it needs to be precisely here for this case:
- We check a batch it has 6 PRs
- We have 3 PRs that are ready to be merged
- We merge 2 PRs within this loop
- We are at 8 now
- Last PR is skipped due to exceeding the limit

narcis-radu (approved):

overmyheadandbody (approved):

PR #3068, opened 2024-10-21
Fix the merge to stage process

mokimo:
### Description
Seems like this [workflow run](https://github.com/adobecom/milo/actions/runs/11436722238/job/31814762385) errors out `TypeError: Cannot read properties of null (reading 'length')` ... we always have a string for the body, so I'm not sure how this is happening, but let's safeguard this. Related to https://github.com/adobecom/milo/pull/3062

Resolves: [MWPW-157289](https://jira.corp.adobe.com/browse/MWPW-157289)

**Test URLs:**
- After: https://fix-stage-process--milo--mokimo.hlx.page/?martech=off

### DIFF
--- a/.github/workflows/merge-to-stage.js
+++ b/.github/workflows/merge-to-stage.js
@@ -236,7 +236,8 @@ const main = async (params) => {
     const stageToMainPR = await getStageToMainPR();
     console.log('has Stage to Main PR:', !!stageToMainPR);
     if (stageToMainPR) body = stageToMainPR.body;
-    existingPRCount = body.match(/https:\/\/github\.com\/adobecom\/milo\/pull\/\d+/g).length;
+    console.log('Current PR body:', body);
+    existingPRCount = body.match(/https:\/\/github\.com\/adobecom\/milo\/pull\/\d+/g)?.length || 0;
     console.log(`Number of PRs already in the batch: ${existingPRCount}`);
 
     if (mergeLimitExceeded()) return console.log(`Maximum number of '${MAX_MERGES}' PRs already merged. Stopping execution`);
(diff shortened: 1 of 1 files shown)


.github/workflows/merge-to-stage.js
```diff
@@ -236,7 +236,8 @@ const main = async (params) => {
     const stageToMainPR = await getStageToMainPR();
     console.log('has Stage to Main PR:', !!stageToMainPR);
     if (stageToMainPR) body = stageToMainPR.body;
-    existingPRCount = body.match(/https:\/\/github\.com\/adobecom\/milo\/pull\/\d+/g).length;
+    console.log('Current PR body:', body);
+    existingPRCount = body.match(/https:\/\/github\.com\/adobecom\/milo\/pull\/\d+/g)?.length || 0;
```
mokimo, line 240: Ah nvm, we get a null value. My omission! 
image

.github/workflows/merge-to-stage.js
```diff
@@ -236,7 +236,8 @@ const main = async (params) => {
     const stageToMainPR = await getStageToMainPR();
     console.log('has Stage to Main PR:', !!stageToMainPR);
     if (stageToMainPR) body = stageToMainPR.body;
-    existingPRCount = body.match(/https:\/\/github\.com\/adobecom\/milo\/pull\/\d+/g).length;
+    console.log('Current PR body:', body);
```
overmyheadandbody, line 239: Would this log actually be helpful?

overmyheadandbody (approved):

.github/workflows/merge-to-stage.js
mokimo: If this errors out in the future on an existing PR, we know which body was detected. Im fine to remove it though

mokimo: Less is more, removed it!
_I added it because I was unsure on how this reproduces, given that in the meantime I found out that string.match returns null rather than an empty array, the purpose of adding this has already been fulfilled. So gone with it!_

narcis-radu (approved):

PR #3074, opened 2024-10-22
MWPW-160834: Clickable App and Playstore links

mokimo:
### Description
Small addition to https://github.com/adobecom/milo/pull/3021 to resolve the buttons being on top of eachother.

Before:
Screenshot 2024-10-22 at 11 17 18

After:
Screenshot 2024-10-22 at 11 17 21

Resolves: [MWPW-160834](https://jira.corp.adobe.com/browse/MWPW-160834)

**Test URLs:**
Milo
- Before: https://main--milo--adobecom.hlx.page/drafts/sara/media?martech=off
- After: https://mwpw-160061-layout--milo--mokimo.hlx.page/drafts/sara/media?martech=off

CC
- Before: https://main--cc--adobecom.hlx.page/in/creativecloud/roc/mobile-apps?milolibs=stage
- After: https://main--cc--adobecom.hlx.page/in/creativecloud/roc/mobile-apps?milolibs=mwpw-160061-layout--milo--mokimo

### DIFF
--- a/libs/blocks/media/media.css
+++ b/libs/blocks/media/media.css
@@ -165,7 +165,7 @@ div[class*="-up"] .media .foreground > .media-row {
 }
 
 .media.qr-code .qr-button-container {
-  display: inline;
+  display: inline-block;
 }
 
 .media.qr-code .google-play,
@@ -178,7 +178,8 @@ div[class*="-up"] .media .foreground > .media-row {
   justify-content: center;
   padding: 0;
   display: inline-flex;
-  margin: var(--spacing-xs) var(--spacing-s) 0 0;
+  margin-top: var(--spacing-xs);
+  margin-inline-end: var(--spacing-s);
 }
 
 .media.qr-code .google-play {
--- a/libs/blocks/media/media.css
+++ b/libs/blocks/media/media.css
@@ -166,6 +166,7 @@ div[class*="-up"] .media .foreground > .media-row {
 
 .media.qr-code .qr-button-container {
   display: inline-block;
+  margin: 0;
 }
 
 .media.qr-code .google-play,
--- a/libs/blocks/media/media.css
+++ b/libs/blocks/media/media.css
@@ -166,7 +166,6 @@ div[class*="-up"] .media .foreground > .media-row {
 
 .media.qr-code .qr-button-container {
   display: inline;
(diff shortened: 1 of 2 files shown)


mokimo:
Milo
- Before: https://main--milo--adobecom.hlx.page/drafts/sara/media?martech=off
- After: https://mwpw-160061-layout--milo--mokimo.hlx.page/drafts/sara/media?martech=off

libs/blocks/media/media.js
```diff
@@ -34,6 +34,8 @@ function decorateQr(el) {
   const googlePlay = text.children[(text.children.length - 2)]?.querySelector('a');
   const qrImage = text.children[(text.children.length - 3)];
   if (!qrImage || !appStore || !googlePlay) return;
+  appStore.parentElement.classList.add('qr-button-container');
+  googlePlay.parentElement.classList.add('qr-button-container');
```
robert-bogos, line 38: Nit: more of a subjective change, but I would make it an one liner
```suggestion
[appStore, googlePlay].forEach(el => el.parentElement.classList.add('qr-button-container'));
```

robert-bogos (approved):

libs/blocks/media/media.js
mokimo: Added 👍

libs/blocks/media/media.css
```diff
@@ -166,7 +166,6 @@ div[class*="-up"] .media .foreground > .media-row {
 
 .media.qr-code .qr-button-container {
   display: inline;
```
overmyheadandbody, line 168: Wouldn't `inline-block` be better suited here? It would also apply the same margins applied to other elements in the section. Overall this looks like it can use a bit more attention, as on RTL the icon positioning is also wrong:
Screenshot 2024-10-22 at 14 28 11

overmyheadandbody (approved):
Although I don't think the mobile app icon section is completely solved for, I understand that we want to have them on the same line for now. @SilviuLCF, it might be worth opening another follow-up story to address the spacings properly and handle RTL. Accessibility-wise, this is also less than ideal, as a screen reader user would not understand the purpose of the links; they don't contain any text, nor do they have any `aria-label`. We should aim to fix all of these in a future PR.

PR #3090, opened 2024-10-24
MWPW-160239 [MEP] Support searching for placeholders in URLs

AdobeLinhart:
When we search for placeholders, we are looking for a pattern of text within double curly brackets.  However, when a placeholder is used inside a URL, the double curly brackets are URL encoded.

Example URL: https://main--cc--adobecom.hlx.page/cc-shared/fragments/trial-modals/product-labeledu#mini-plans-web-ctaproduct-label-card-sb

Becomes: /cc-shared/fragments/trial-modals/%7B%7Bproduct-label%7D%7D-edu#mini-plans-web-cta-%7B%7Bproduct-label%7D%7D-card-sb

I want to add support for these encoded brackets.

**QA instructions**
Check the first link for data-modal-path or data-modal-hash on the Free trial anchor. Placeholders will be unresolved. This should be fixed on the second link.

Resolves:  [MWPW-160239](https://jira.corp.adobe.com/browse/MWPW-160239)

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q4/urlplaceholders/aside
- After: https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q4/urlplaceholders/aside?milolibs=urlplaceholders
- Psi-check: https://urlplaceholders--milo--adobecom.hlx.page/?martech=off

### DIFF
--- a/libs/features/placeholders.js
+++ b/libs/features/placeholders.js
@@ -135,8 +135,10 @@ export async function decoratePlaceholderArea({
     if (nodeEl.nodeType === Node.TEXT_NODE) {
       nodeEl.nodeValue = await replaceText(nodeEl.nodeValue, config);
     } else if (nodeEl.nodeType === Node.ELEMENT_NODE) {
-      const hrefVal = await replaceText(nodeEl.getAttribute('href'), config);
-      nodeEl.setAttribute('href', hrefVal);
+      for (const attr of nodeEl.attributes) {
+        const attrVal = await replaceText(attr.value, config);
+        nodeEl.setAttribute(attr.name, attrVal);
+      }
     }
   });
   await Promise.all(replaceNodes);
--- a/libs/features/placeholders.js
+++ b/libs/features/placeholders.js
@@ -135,10 +135,14 @@ export async function decoratePlaceholderArea({
     if (nodeEl.nodeType === Node.TEXT_NODE) {
       nodeEl.nodeValue = await replaceText(nodeEl.nodeValue, config);
     } else if (nodeEl.nodeType === Node.ELEMENT_NODE) {
-      for (const attr of nodeEl.attributes) {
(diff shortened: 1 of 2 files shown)


libs/features/placeholders.js
```diff
@@ -135,8 +135,10 @@ export async function decoratePlaceholderArea({
     if (nodeEl.nodeType === Node.TEXT_NODE) {
       nodeEl.nodeValue = await replaceText(nodeEl.nodeValue, config);
     } else if (nodeEl.nodeType === Node.ELEMENT_NODE) {
-      const hrefVal = await replaceText(nodeEl.getAttribute('href'), config);
-      nodeEl.setAttribute('href', hrefVal);
+      for (const attr of nodeEl.attributes) {
+        const attrVal = await replaceText(attr.value, config);
+        nodeEl.setAttribute(attr.name, attrVal);
+      }
```
markpadbe, line 141: Since an element could end up having multiple attributes and multiple network calls, it might be better for performance to collect these promises and execute them concurrently

markpadbe, line 141:
```suggestion
     const attrPromises = [...nodeEl.attributes].map(async (attr) => {
        const attrVal = await replaceText(attr.value, config);
        return { name: attr.name, value: attrVal };
      });
      const results = await Promise.all(attrPromises);
      results.forEach(({ name, value }) => {
        nodeEl.setAttribute(name, value);
      });
```

AdobeLinhart: @vgoodric Easy enough change. Do you concur?

AdobeLinhart: @markpadbe This has been updated.

vgoodric (approved):

markpadbe (approved):

PR #3092, opened 2024-10-28
MWPW-161082 [Base Implementation] Mobile Global Nav Redesign Rollout

sharmrj:
* Initial commit for the mobile gnav redesign

Resolves: [MWPW-161082](https://jira.corp.adobe.com/browse/MWPW-161082)

**Test URLs:**
- After: https://mgnav--milo--sharmrj.hlx.page/?martech=off

https://main--cc--adobecom.hlx.page/products/photoshop?milolibs=mgnav--milo--sharmrj&newNav=true&georouting=off


sharmrj:
Unit Tests will be taken care of before we raise the PR to Stage

mokimo:
image
The mobile nav experience does not look correct

mokimo:
The keyboard navigation also does not work as expected anymore, it should be possible to cycle through items using tab, arrow keys, shift tab etc

sharmrj:
> image The mobile nav experience does not look correct

Try using this link: https://main--cc--adobecom.hlx.page/products/photoshop?milolibs=mgnav--milo--sharmrj&newNav=true&georouting=off

sharmrj:
> The keyboard navigation also does not work as expected anymore, it should be possible to cycle through items using tab, arrow keys, shift tab etc

We'll be addressing this in a future ticket

sharmrj:
You're correct @mokimo, there is some CLS caused by the localnav. It's a bit of an open question what approach we're gonna take for now. One idea was to show the localnav only when the user starts scrolling down, but that isn't the UX we want. Probably what we'll go with is having the authors indicate in the metadata whether this is a localnav or not, and reserving the space accordingly.

PR #3098, opened 2024-10-28
 MWPW-160058: Fixes tag matching case logic

cmiqueo:
**The problem:**

Right now, the milo _send-utils.js_ script compares the tags from adobe.com with the tags in the page.

But before comparing the tags it "lower-cases" ONLY the tags from adobe.com and not the ones from the page:

**Proposed solution:**

Lower case both set of tag values before comparing them.

Resolves: [MWPW-160058](https://jira.corp.adobe.com/browse/MWPW-160058)

**Test URLs:**
- After: https://MWPW-160058--milo--adobecom.hlx.page/?martech=off

@mokimo 
This is a (fairly large and complex) utilities file that was created long time ago and it does not have any test coverage. We are planing on creating a story to add a full code coverage to all the functionalities in the file, but we should not hold up any small incremental improvements.


mokimo (approved):

mokimo (approved):

arshadparwaiz (approved):

sheridansunier (approved):

cmiqueo:
> This pull request is not passing all required checks. Please see [this discussion](https://github.com/orgs/adobecom/discussions/997) for information on how to get all checks passing. Inconsistent checks can be manually retried. If a test absolutely can not pass for a good reason, please add a comment with an explanation to the PR.

@mokimo 
This is a (fairly large and complex) utilities file that was created long time ago and it does not have any test coverage. We are planing on creating a story to add a full code coverage to all the functionalities in the file, but we should not hold up any small incremental improvements.

JackySun9 (approved):

mokimo:
You could ignore the coverage for thsi line https://github.com/bcoe/c8/blob/main/README.md
Other than that, someone (likely me) would have to manually merge this, but I have to hit the time where a batch isn't tested so that's more un-predictable and takes a bit longer!

PR #3100, opened 2024-10-28
MWPW-161130 - [Table] Handle dynamic pricing in addon labels

Sartxi:
When adding dynamic prices in the table blocks new addon feature the dynamic price element is rendered in the flex grid instead of inline with the text content. Fixing this issue by separating out the icon element and grouping all other authored content in addon labels.

Resolves: [MWPW-161130](https://jira.corp.adobe.com/browse/MWPW-161130)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/sarchibeque/table/add-on?martech=off
- After: https://sartxi-mwpw-161130-table--milo--adobecom.hlx.page/drafts/sarchibeque/table/add-on?martech=off


robert-bogos (commented):
Hi @Sartxi ! Are the failing unit tests related to your changes?

libs/blocks/table/table.css
```diff
@@ -485,7 +485,6 @@
   text-align: start;
   font-style: normal;
   font-weight: 400;
-  gap: var(--spacing-xxs);
```
overmyheadandbody, line 488: Does this really need to be removed? It could mean that text isn't separated from the icon, right? Manually adding the space back doesn't seem to lead to any isses.

Sartxi:
> Hi @Sartxi ! Are the failing unit tests related to your changes?

No they aren't, not sure why but I am having the same failures in all my PRs.

libs/blocks/table/table.css
Sartxi: Screenshot 2024-10-30 at 9 03 15 AM

@overmyheadandbody I noticed that the icon actually gets styled to have margin on the left. So the gap was adding another extra space that was not in the design.

overmyheadandbody (approved):

robert-bogos (approved):
Looks good 👍

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-161130

PR #3108, opened 2024-10-29
MWPW-161191:  Checkout cli parameter does not get rendered correctly

Axelcureno:
Fixes CLI parameter in checkout links in milo consumer projects like DC, CC.

Resolves: [MWPW-161191](https://jira.corp.adobe.com/browse/MWPW-161191)

Actual test URL (needs to be tested from milo consumer project): https://main--dc--adobecom.hlx.live/acrobat/features?milolibs=MWPW-161191--milo--adobecom

**Test URLs:**
- After: https://MWPW-161191--milo--adobecom.hlx.page/docs/library/kitchen-sink/merch-card?martech=off

### DIFF
--- a/libs/blocks/merch/merch.js
+++ b/libs/blocks/merch/merch.js
@@ -560,6 +560,23 @@ export async function initService(force = false, attributes = {}) {
     fetchCheckoutLinkConfigs.promise = undefined;
   }
   const { commerce, env: miloEnv, locale: miloLocale } = getConfig();
+
+  const extraAttrs = [
+    'checkout-workflow-step',
+    'force-tax-exclusive',
+    'checkout-client-id',
+    'allow-override',
+  ];
+
+  extraAttrs.forEach((attr) => {
+    const camelCaseAttr = attr.replace(/-([a-z])/g, (g) => g[1].toUpperCase());
+    // eslint-disable-next-line no-prototype-builtins
+    if (commerce.hasOwnProperty(camelCaseAttr)) {
+      const value = commerce[camelCaseAttr];
+      delete commerce[camelCaseAttr];
+      commerce[attr] = value;
+    }
+  });
   initService.promise = initService.promise ?? polyfills().then(async () => {
     await import('../../deps/mas/commerce.js');
     const { language, locale } = getMiloLocaleSettings(miloLocale);
@@ -629,7 +646,7 @@ export async function getCheckoutContext(el, params) {
 
   return {
     ...context,
-    checkoutClientId,
+    'checkout-client-id': checkoutClientId,
     checkoutWorkflow,
     checkoutWorkflowStep,
     checkoutMarketSegment,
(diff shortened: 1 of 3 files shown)


Roycethan (changes requested):
@Axelcureno Its breaking Pricing and Checkout CTA's (label and commerce url) at CC pages:
https://main--cc--adobecom.hlx.live/creativecloud?milolibs=MWPW-161191--milo--adobecom
image

cc @yesil  @npeltier @afmicka @3ch023

libs/blocks/merch/merch.js
```diff
@@ -629,7 +646,7 @@ export async function getCheckoutContext(el, params) {
 
   return {
     ...context,
-    checkoutClientId,
+    'checkout-client-id': checkoutClientId,
```
npeltier, line 649: i'm not sure why we are doing this: mapping in & out from kebab case is done for having well formed attributes but before and after camel case is used iirc

Axelcureno: true, I have removed it.

Roycethan:
> @Axelcureno Its breaking Pricing and Checkout CTA's (label and commerce url) at CC pages: https://main--cc--adobecom.hlx.live/creativecloud?milolibs=MWPW-161191--milo--adobecom image
> 
> cc @yesil @npeltier @afmicka @3ch023

@Axelcureno  Fixed now with latest commit. Thanks

Roycethan (approved):

Roycethan:
@Axelcureno  Verified, Need 2 approvals for Ready for stage label, and if there any changes a smoke test. Thanks

libs/blocks/merch/merch.js
```diff
@@ -560,6 +560,23 @@ export async function initService(force = false, attributes = {}) {
...
+    'checkout-client-id',
+    'allow-override',
+  ];
+
+  extraAttrs.forEach((attr) => {
+    const camelCaseAttr = attr.replace(/-([a-z])/g, (g) => g[1].toUpperCase());
+    // eslint-disable-next-line no-prototype-builtins
+    if (commerce?.hasOwnProperty(camelCaseAttr)) {
+      const value = commerce[camelCaseAttr];
+      delete commerce[camelCaseAttr];
```
yesil, line 576: I think we should not alter `commerce` object.

libs/blocks/merch/merch.js
```diff
@@ -560,6 +560,23 @@ export async function initService(force = false, attributes = {}) {
     fetchCheckoutLinkConfigs.promise = undefined;
   }
   const { commerce, env: miloEnv, locale: miloLocale } = getConfig();
+
+  const extraAttrs = [
+    'checkout-workflow-step',
+    'force-tax-exclusive',
+    'checkout-client-id',
+    'allow-override',
```
yesil, line 568: `allow-override` is not controlled by the consumer but by merch.js, and it is set only for non prod domains.
we should remove it from the list.

libs/blocks/merch/merch.js
npeltier: those attributes could be added through attributes object

npeltier (approved):
i generally agree with @yesil feedbacks, but this can be treated afterward

yesil (approved):
my feedback can be addressed later.

PR #3113, opened 2024-10-30
MWPW-161273 Standalone Gnav needs a release cycle

sharmrj:
Milo "releases" (so called because techincally Milo/EDS do continuous delivery but we have a staging process) fairly quickly, too quick for most standalone gnav consumers. We need to have a release process for changes in standalone gnav. The ersatz solution is to maintain a separate branch from main, called feds-gnav, and every two weeks merge main into this separate branch. 

We'll have an Akamai rule that redirects any requests going to `www.adobe.com/feds-gnav/` to `feds-gnav-milo--adobecom.hlx.live/libs/` (by the same token, requests to `www.stage.adobe.com/feds-gnav/` go to `stage-feds-gnav--milo--adobecom.hlx.live/libs/`)

To support the above change we also need to change navigation.js to point to /feds-gnav/* instead of /libs/* in the miloLibs portion of the config.

Resolves: [MWPW-161273](https://jira.corp.adobe.com/browse/MWPW-161273)

**Test URLs:**
- After: https://standalone-gnav--milo--sharmrj.hlx.page/?martech=off


libs/navigation/navigation.js
```diff
@@ -57,24 +57,24 @@ export default async function loadBlock(configs, customLib) {
...
     return;
   }
   // Relative path can't be used, as the script will run on consumer's app
   const [{ default: bootstrapBlock }, { default: locales }, { setConfig }] = await Promise.all([
-    import(`${miloLibs}/libs/navigation/bootstrapper.js`),
-    import(`${miloLibs}/libs/utils/locales.js`),
-    import(`${miloLibs}/libs/utils/utils.js`),
+    import(`${miloLibs}/navigation/bootstrapper.js`),
+    import(`${miloLibs}/utils/locales.js`),
+    import(`${miloLibs}//utils/utils.js`),
```
mokimo, line 70:
```suggestion
    import(`${miloLibs}/utils/utils.js`),
```

sharmrj:
superceded by #3132

PR #3065, opened 2024-10-18
 MWPW-156940 AU ABM price display

bozojovicic:
ABM price on Australian pages must be displayed together with annual price in brackets. Few examples from Figma :

Screenshot 2024-10-18 at 13 53 00

Screenshot 2024-10-18 at 13 53 19

Test URLs :
https://main--cc--adobecom.hlx.live/au/products/photoshop?milolibs=mwpw156940auabm--milo--bozojovicic
https://main--cc--adobecom.hlx.live/au/products/illustrator?milolibs=mwpw156940auabm--milo--bozojovicic
https://main--cc--adobecom.hlx.live/au/products/photoshop/plans?milolibs=mwpw156940auabm--milo--bozojovicic
https://main--cc--adobecom.hlx.live/au/products/illustrator/plans?milolibs=mwpw156940auabm--milo--bozojovicic 

https://mwpw156940auabm--milo--bozojovicic.aem.page/au/docs/library/kitchen-sink/merch-card

Resolves: [MWPW-156940](https://jira.corp.adobe.com/browse/MWPW-156940)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/bozo/price-cta?martech=off
- After: https://mwpw156940auabm--milo--bozojovicic.hlx.page/drafts/bozo/price-cta?martech=off

PR #3115, opened 2024-10-30
MWPW-156940 AU price display

bozojovicic:
This is the cleaner version of PR https://github.com/adobecom/milo/pull/3065
ABM price on Australian pages must be displayed together with annual price in brackets. Few examples from Figma :

Screenshot 2024-10-18 at 13 53 00

Screenshot 2024-10-18 at 13 53 19

Test URLs :
https://main--cc--adobecom.hlx.live/au/products/photoshop?milolibs=mwpw156940auabm2--milo--bozojovicic
https://main--cc--adobecom.hlx.live/au/products/illustrator?milolibs=mwpw156940auabm2--milo--bozojovicic
https://main--cc--adobecom.hlx.live/au/products/photoshop/plans?milolibs=mwpw156940auabm2--milo--bozojovicic
https://main--cc--adobecom.hlx.live/au/products/illustrator/plans?milolibs=mwpw156940auabm2--milo--bozojovicic 

https://mwpw156940auabm2--milo--bozojovicic.aem.page/au/docs/library/kitchen-sink/merch-card

Resolves: [MWPW-156940](https://jira.corp.adobe.com/browse/MWPW-156940)

**Test URLs:**
- After: https://mwpw156940auabm2--milo--bozojovicic.hlx.page/drafts/bozo/price-cta?martech=off


npeltier (commented):
looks much better to me @bozojovicic, just one move proposal comment left! can you create AU drafts so we can see it in action?

libs/blocks/hero-marquee/hero-marquee.js
```diff
@@ -258,6 +258,10 @@ export default async function init(el) {
       firstCol.parentElement.classList.add(`row-${parsed.key}`, 'con-block');
       firstCol.remove();
       cols[1].classList.add('row-wrapper');
+      if (locale.prefix === '/au') {
```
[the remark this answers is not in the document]
npeltier: typically in merch's initService, you _know_ you will need merch, and you could eventually check for AU & hero?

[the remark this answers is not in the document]
bozojovicic: @npeltier I removed this whole change related with hero-marquee from this PR. The request is :

- if we have strikethru price next to the ABM then the annual price needs to go to new line
- if there is no strikethru price, then ABM and annual price need to be in the same line
- this counts for hero-marquee component

But it is just **too** complicated change for something that is asked for AU only. And now I see that we have some marquee component, similar to hero-marquee ... So, let's go without this change. If they don't like it, we can discuss it again. It is not well defined anyway.

Therefore, there will be no `au-merch.js` file. Only `au-merch.css`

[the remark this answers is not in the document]
bozojovicic: What I mean by "not well defined"? They just sent few example screenshots in figma file. They never wrote in Jira some specific requirements like : do changes A and B in component C on page D.

npeltier (approved):
looks good to me now, thanks @bozojovicic  :)

libs/blocks/merch/au-merch.css
```diff
@@ -0,0 +1,53 @@
...
+  font-weight: 400;
+  position: relative;
+  left: -3px;
+}
+
+merch-card [slot='heading-m'] [is="inline-price"] .price-annual-prefix:before {
+  content: '\A';
+  white-space: pre;
+}
+merch-card [slot='heading-m'] [is="inline-price"] .price-annual-prefix,
```
yesil, line 25: merch-card styles do not belong here. They should go in global.css.js in mas web-components.

yesil (changes requested):

libs/blocks/merch/au-merch.css
bozojovicic: @npeltier @yesil we need some decision here. We had this piece of CSS in libs/features/mas/web-components/src/global.css.js before. Nicolas proposed to move all such changes to au-merch.css which will be loaded on AU pages only. Do you think that this part should be an exception and that I should keep it in global.css.js ? Or it is better like this?

npeltier: @bozojovicic (cc @yesil ) the two informations are not necessary contradictory. 
i say i don't like having AU css (and logic back then) in all of the code,
ilyas says that he does not like having card WC styling in milo styles. 

@yesil do you think that styling is generic to all m@s consumers? i would rather keep it in milo for now

yesil: @npeltier  part of the CSS rules like the following look generic to me for annual price formatting.
anything that is for annual price formatting should stay in global.css.js IMO.

`merch-card [slot='heading-m'] [is="inline-price"] .price-annual-prefix + .price-annual`

bozojovicic: @yesil CSS rules that we have now in au-merch.css I initially had in

- libs/blocks/table/table.css
- libs/features/mas/web-components/src/global.css.js
- libs/features/mas/web-components/src/variants/mini-compare-chart.css.js

and then we decided to move all of that to au-merch.css and load that file for AU pages only.
If I move this part to global.css.js, then I need to do the same for the part I copied from mini-compare-chart.css.js
And then only few rules will stay in au-merch.css and then this files loses its purpose.

For me, both approaches do the job. I personally vote to keep it all in au-merch.css. But I just need some decision which part goes where.
CC @npeltier

yesil (approved):
Discussed with @npeltier , I'm fine with keeping these styles in Milo au specific file.

afmicka:
@bozojovicic  Have added the following comments to the jira as well

1. Even though it has been agreed by my understanding that the manually added tax label placeholders will be fixed after the Nov 6th launch, there is one page where the display issue is more prominent and visible. needs verification if acceptable for the launch
https://main--cc--adobecom.hlx.live/au/creativecloud?milolibs=mwpw156940auabm2--milo--bozojovicic
Screenshot 2024-11-01 at 15 29 56 
Screenshot 2024-11-01 at 15 30 02 
Screenshot 2024-11-01 at 15 30 08

2. The price gets outside the table area on mobile view (iPhone 12 Pro, real iPhone 14 Pro)
https://main--cc--adobecom.hlx.live/au/products/illustrator/plans?milolibs=mwpw156940auabm2--milo--bozojovicic

Screenshot 2024-11-01 at 15 35 53


The rest of regression looks ok to me. If GWP approves to go as is, we can add verify and Ready for stage labels.
cc @Roycethan (please have a second look if time permits)
Otherwise, the changes and regression look ok. If you accept this to go as is (especially the #1), we can add verify and Ready for stage label to the PR

bozojovicic:
@afmicka @Roycethan I added couple of fixes. 
1. The label "incl GST" is not displaced any more, now it naturally flows after the annual price. It almost looks ok now. ABM price also goes after the ST price, in the same line now. Of course content still needs to be fixed to have what is requested in figma.
2. Should also be fixed now. And will be even better once the content is fixed.

Roycethan:
@bozojovicic 
1)Prices should not break at line end, Annual price here should wrap in one line
https://main--cc--adobecom.hlx.live/au/products/special-offers?milolibs=mwpw156940auabm2--milo--bozojovicic#
image

cc @afmicka

PR #3119, opened 2024-10-30
MWPW-158127 MEP Add support for identifying iPad on Safari as a mobile device and as iOS

markpadbe:
Currently all iPads (except for iPad mini) are not registered as iOS or as a mobile device when using Safari. This fix adds a new safariIpad check to the exisiting checks in PERSONANLIZATION_TAGS. Please note this fix will work on actual devices (e.g., iPads accessed via browser stack) but it does not address the issue when using devtools/emulators. 
Resolves: [MWPW-158127](https://jira.corp.adobe.com/browse/MWPW-158127)

QA instructions: I've included two sets of Before and After test URLS below, which should be viewed on an actual iPad device(not iPad mini), using the safari browser (can access via browserstack).

_**Expected header text in marquee:**_
Before(breakpoints): _"I'm on a desktop"_
After(breakpoints): _"I'm on a tablet"_
Before(ios): _"I'm on a mac"_
After(ios): _"I'm on ios'_

**Test URLs:**
- Before(breakpoints): https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q4/tabletbug/breakpoints
- After(breakpoints): https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q4/tabletbug/breakpoints?milolibs=tabletbug
- Before(ios): https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q4/tabletbug/iosmac
- After(ios): https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2024/q4/tabletbug/iosmac?milolibs=tabletbug
- Psi-check: https://tabletbug--milo--adobecom.hlx.page/?martech=off


libs/features/personalization/personalization.js
```diff
@@ -6,17 +6,20 @@ import { getFederatedUrl } from '../../utils/federated.js';
 
 /* c8 ignore start */
 const PHONE_SIZE = window.screen.width < 550 || window.screen.height < 550;
+const safariIpad = navigator.userAgent.includes('Macintosh') && navigator.maxTouchPoints > 1;
 export const PERSONALIZATION_TAGS = {
   all: () => true,
   chrome: () => navigator.userAgent.includes('Chrome') && !navigator.userAgent.includes('Edg'),
```
ivanvatadobe, line 12: This may not affect any checks, but since we're using full names of browsers: possibly change 'Edg' to 'Edge'.

vgoodric (approved):

libs/features/personalization/personalization.js
vgoodric: @ivanvatadobe Fun Fact, it's because the user agent doesn't say edge.  example:
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/130.0.0.0 Safari/537.36 Edg/130.0.0.0

vgoodric (approved):

PR #3122, opened 2024-10-31
Helix 5 support for milo blocks and features

mokimo:
### Description
- There are a few special cases, such as loc, where I'm unsure about implications of moving certain links to aem.page, hence those simply got a "TODO" for now.
- There are other special cases, such as unit tests, where currently porting to `aem.page` does not make any sense, as our default is `hlx.page` unless detected otherwise.
- This should *not* change any production behavior and hopefully development behavior should stay the same.

### Testing instructions
both `https://main--milo--adobecom.aem.page/docs/library/kitchen-sink/` and `https://main--milo--adobecom.hlx.page/docs/library/kitchen-sink/` should work and have no differences.

Consuming pages should also not have any difference in experience: https://main--cc--adobecom.hlx.live/de/products/photoshop?milolibs=hlx5-upgrade--milo--mokimo

Lastly, all tools should also be checked (caas, preflight, loc) to ensure we didn't accidentally introduce any errors.

Resolves: [MWPW-157482](https://jira.corp.adobe.com/browse/MWPW-157482)

**Test URLs:**
- After: https://hlx5-upgrade--milo--mokimo.aem.page/?martech=off

### DIFF
--- a/libs/blocks/bulk-publish-v2/utils.js
+++ b/libs/blocks/bulk-publish-v2/utils.js
@@ -6,7 +6,7 @@ const checkedIcon = `${base}/blocks/bulk-publish-v2/img/checked.svg`;
 const crossedIcon = `${base}/blocks/bulk-publish-v2/img/crossed.svg`;
 const indicatorIcon = `${base}/blocks/bulk-publish-v2/img/indicator.svg`;
 
-const getHost = (host) => (host === 'localhost' ? 'main--milo--adobecom.hlx.page' : host);
+const getHost = (host) => (host === 'localhost' ? 'main--milo--adobecom.aem.page' : host);
 
 const getStatusIcon = (status) => {
   switch (status) {
--- a/libs/blocks/quiz-entry/quiz-entry.css
+++ b/libs/blocks/quiz-entry/quiz-entry.css
@@ -178,7 +178,7 @@
 
 .carousel-arrow {
   background-color: #fff;
-  background-image: url('https://main--milo--adobecom.hlx.page/libs/blocks/carousel/img/arrow.svg');
+  background-image: url('https://main--milo--adobecom.aem.page/libs/blocks/carousel/img/arrow.svg');
   background-size: 15px; 
   background-position: 50% 50%; 
   background-repeat: no-repeat;
--- a/libs/blocks/version-history/index.js
+++ b/libs/blocks/version-history/index.js
@@ -15,6 +15,7 @@ function getSiteOrigin() {
   const search = new URLSearchParams(window.location.search);
   const repo = search.get('repo');
   const owner = search.get('owner');
+  // TODO ADD HLX5 SUPPORT
   return repo && owner ? `https://main--${repo}--${owner}.hlx.live` : window.location.origin;
 }
 
--- a/libs/scripts/delayed.js
+++ b/libs/scripts/delayed.js
@@ -28,7 +28,9 @@ export const loadPrivacy = async (getConfig, loadScript) => {
   const acom = '7a5eb705-95ed-4cc4-a11d-0cc5760e93db';
   const ids = {
     'hlx.page': '3a6a37fe-9e07-4aa9-8640-8f358a623271-test',
+    'aem.page': '3a6a37fe-9e07-4aa9-8640-8f358a623271-test',
     'hlx.live': '926b16ce-cc88-4c6a-af45-21749f3167f3-test',
+    'aem.live': '926b16ce-cc88-4c6a-af45-21749f3167f3-test',
   };
 
   const otDomainId = ids?.[Object.keys(ids)
--- a/libs/scripts/scripts.js
+++ b/libs/scripts/scripts.js
@@ -41,6 +41,12 @@ const stageDomainsMap = {
     '^https://business.adobe.com/blog': 'https://main--bacom-blog--adobecom.hlx.page',
     '^https://www.adobe.com': 'origin',
   },
+  '^https://.*--milo--.*.aem.page': {
+    '^https://www.adobe.com/acrobat': 'https://main--dc--adobecom.hlx.page',
+    '^https://business.adobe.com(?!/blog)': 'https://business.stage.adobe.com',
+    '^https://business.adobe.com/blog': 'https://main--bacom-blog--adobecom.hlx.page',
+    '^https://www.adobe.com': 'origin',
+  },
 };
 
 const config = {
--- a/nala/utils/global.setup.js
+++ b/nala/utils/global.setup.js
@@ -3,7 +3,7 @@
 const { execSync } = require('child_process');
 const { isBranchURLValid } = require('../libs/baseurl.js');
 
-const MAIN_BRANCH_LIVE_URL = 'https://main--milo--adobecom.hlx.live';
+const MAIN_BRANCH_LIVE_URL = 'https://main--milo--adobecom.aem.live';
 const STAGE_BRANCH_URL = 'https://milo.stage.adobe.com';
 
 async function getGitHubPRBranchLiveUrl() {
@@ -29,7 +29,7 @@ async function getGitHubPRBranchLiveUrl() {
   const prFromOrg = process.env.prOrg || toRepoOrg;
   const prFromRepoName = process.env.prRepo || toRepoName;
 
-  const prBranchLiveUrl = `https://${prBranch}--${prFromRepoName}--${prFromOrg}.hlx.live`;
+  const prBranchLiveUrl = `https://${prBranch}--${prFromRepoName}--${prFromOrg}.aem.live`;
 
   try {
     if (await isBranchURLValid(prBranchLiveUrl)) {
--- a/nala/utils/pr.run.sh
+++ b/nala/utils/pr.run.sh
@@ -33,6 +33,7 @@ toRepoName=${repoParts[1]}
 prRepo=${prRepo:-$toRepoName}
 prOrg=${prOrg:-$toRepoOrg}
 
+# TODO ADD HLX5 SUPPORT
 PR_BRANCH_LIVE_URL_GH="https://$FEATURE_BRANCH--$prRepo--$prOrg.hlx.live"
 
 # set pr branch url as env
--- a/test/blocks/caas/mocks/utils.js
+++ b/test/blocks/caas/mocks/utils.js
@@ -25,6 +25,9 @@ export function createIntersectionObserver({ el, callback /* , once = true, opti
   callback(el, { target: el });
 }
 
+const PAGE_URL = new URL(window.location.href);
+export const SLD = PAGE_URL.hostname.includes('.aem.') ? 'aem' : 'hlx';
+
 export const parseEncodedConfig = stub().returns({
   analyticsTrackImpression: false,
   analyticsCollectionName: '',
--- a/test/blocks/global-navigation/utilities/utilities.test.js
+++ b/test/blocks/global-navigation/utilities/utilities.test.js
@@ -19,7 +19,7 @@ import { setConfig, getConfig } from '../../../../libs/utils/utils.js';
 import { createFullGlobalNavigation, config } from '../test-utilities.js';
 import mepInBlock from '../mocks/mep-config.js';
 
-const baseHost = 'https://main--federal--adobecom.hlx.page';
+const baseHost = 'https://main--federal--adobecom.aem.page';
 describe('global navigation utilities', () => {
   beforeEach(() => {
     document.body.innerHTML = '';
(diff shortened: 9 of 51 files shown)


JasonHowellSlavin (approved):
Looks good! Will consumers have to update their setLibs functions? https://github.com/adobecom/bacom/blob/01f2a07ee84801db563d15beba3b7c4c0aeb6068/scripts/scripts.js#L158

mokimo:
@JasonHowellSlavin Yeah, however that will happen as part of the consumers migrating to hlx5. Additionally, it will require some coordination consumer<->WebPS.

I'm in the process of preparing additional docs for this and then work on a plan so that each consumer can make the switch! So far we are not under pressure, but long term `aem.page` & hlx5 will become a requirement from the product perspective

hparra (approved):
Only reviewed change in `libs/features/title-append/README.md`

libs/scripts/scripts.js
```diff
@@ -41,6 +41,12 @@ const stageDomainsMap = {
     '^https://business.adobe.com/blog': 'https://main--bacom-blog--adobecom.hlx.page',
     '^https://www.adobe.com': 'origin',
   },
+  '^https://.*--milo--.*.aem.page': {
+    '^https://www.adobe.com/acrobat': 'https://main--dc--adobecom.hlx.page',
+    '^https://business.adobe.com(?!/blog)': 'https://business.stage.adobe.com',
+    '^https://business.adobe.com/blog': 'https://main--bacom-blog--adobecom.hlx.page',
+    '^https://www.adobe.com': 'origin',
+  },
```
overmyheadandbody, line 49: This might be tricky to keep track of, we're now redirecting .aem. traffic to .hlx. on consumers, sincer they're not on Helix 5 just yet, but we'll need to update these at one point. Shouldn't be an issue at the moment, though.

nala/utils/global.setup.js
```diff
@@ -29,7 +29,7 @@ async function getGitHubPRBranchLiveUrl() {
   const prFromOrg = process.env.prOrg || toRepoOrg;
   const prFromRepoName = process.env.prRepo || toRepoName;
 
-  const prBranchLiveUrl = `https://${prBranch}--${prFromRepoName}--${prFromOrg}.hlx.live`;
+  const prBranchLiveUrl = `https://${prBranch}--${prFromRepoName}--${prFromOrg}.aem.live`;
```
overmyheadandbody, line 32: Could the repo be something else other than Milo? This might cause issues, since most consumers are not using `.aem.`

test/blocks/marketo/mocks/marketo-utils.js
```diff
@@ -69,3 +69,6 @@ export const loadLink = stub().returns(new Promise((resolve) => {
 }));
 
 export const customFetch = stub();
+
+const PAGE_URL = new URL(window.location.href);
+export const SLD = PAGE_URL.hostname.includes('.aem.') ? 'aem' : 'hlx';
```
overmyheadandbody, line 74: Nit: does this actually need to be exported?

test/blocks/merch/mocks/embed-utils.js
```diff
@@ -23,6 +23,9 @@ export function createTag(tag, attributes, html) {
   return el;
 }
 
+const PAGE_URL = new URL(window.location.href);
+export const SLD = PAGE_URL.hostname.includes('.aem.') ? 'aem' : 'hlx';
```
overmyheadandbody, line 27: Nit: does this actually need to be exported?

test/blocks/caas/mocks/utils.js
```diff
@@ -25,6 +25,9 @@ export function createIntersectionObserver({ el, callback /* , once = true, opti
   callback(el, { target: el });
 }
 
+const PAGE_URL = new URL(window.location.href);
+export const SLD = PAGE_URL.hostname.includes('.aem.') ? 'aem' : 'hlx';
```
overmyheadandbody, line 29: Nit: does this actually need to be exported?

libs/scripts/delayed.js
```diff
@@ -28,7 +28,9 @@ export const loadPrivacy = async (getConfig, loadScript) => {
   const acom = '7a5eb705-95ed-4cc4-a11d-0cc5760e93db';
   const ids = {
     'hlx.page': '3a6a37fe-9e07-4aa9-8640-8f358a623271-test',
+    'aem.page': '3a6a37fe-9e07-4aa9-8640-8f358a623271-test',
     'hlx.live': '926b16ce-cc88-4c6a-af45-21749f3167f3-test',
+    'aem.live': '926b16ce-cc88-4c6a-af45-21749f3167f3-test',
```
overmyheadandbody, line 33: In my initial PR, I changed this a bit to not duplicate IDs, maybe it's worth considering - https://github.com/adobecom/milo/compare/helix-5?expand=1#diff-c21f6b181685a27721a54ce2b519823f6186157499b08ea290437d3fdff64473

test/blocks/global-navigation/utilities/utilities.test.js
```diff
@@ -19,7 +19,7 @@ import { setConfig, getConfig } from '../../../../libs/utils/utils.js';
 import { createFullGlobalNavigation, config } from '../test-utilities.js';
 import mepInBlock from '../mocks/mep-config.js';
 
-const baseHost = 'https://main--federal--adobecom.hlx.page';
+const baseHost = 'https://main--federal--adobecom.aem.page';
```
overmyheadandbody, line 22: Extreme nitting: L209 has a `.hlx.` reference 🙈 https://github.com/adobecom/milo/blob/965fdeed6a2b052a2ef473b838eaac9001520925/test/blocks/global-navigation/utilities/utilities.test.js#L209

tools/send-to-caas/send-utils.js
```diff
@@ -195,7 +195,7 @@ export const getOrigin = (fgColor) => {
     return originLC;
   }
 
-  if (window.location.hostname.endsWith('.hlx.page')) {
+  if (window.location.hostname.endsWith(`.${SLD}.page`)) {
```
overmyheadandbody, line 198: Do we really need to test for `SLD` here or would `endsWith('.page')` be enough? Same question for L277

overmyheadandbody (commented):
This looks good in general, but I do have a few questions. I also tried to compare some notes with my [original commit](https://github.com/adobecom/milo/compare/helix-5?expand=1) to figure out whether we're missing something, but things seem on par.

libs/blocks/bulk-publish-v2/utils.js
```diff
@@ -6,7 +6,7 @@ const checkedIcon = `${base}/blocks/bulk-publish-v2/img/checked.svg`;
 const crossedIcon = `${base}/blocks/bulk-publish-v2/img/crossed.svg`;
 const indicatorIcon = `${base}/blocks/bulk-publish-v2/img/indicator.svg`;
 
-const getHost = (host) => (host === 'localhost' ? 'main--milo--adobecom.hlx.page' : host);
+const getHost = (host) => (host === 'localhost' ? 'main--milo--adobecom.aem.page' : host);
```
narcis-radu, line 9: I see we're not using `SLD` here

libs/blocks/version-history/index.js
```diff
@@ -15,6 +15,7 @@ function getSiteOrigin() {
   const search = new URLSearchParams(window.location.search);
   const repo = search.get('repo');
   const owner = search.get('owner');
+  // TODO ADD HLX5 SUPPORT
```
narcis-radu, line 18: Shouldn't we add support now?

libs/scripts/delayed.js
```diff
@@ -28,7 +28,9 @@ export const loadPrivacy = async (getConfig, loadScript) => {
   const acom = '7a5eb705-95ed-4cc4-a11d-0cc5760e93db';
   const ids = {
     'hlx.page': '3a6a37fe-9e07-4aa9-8640-8f358a623271-test',
+    'aem.page': '3a6a37fe-9e07-4aa9-8640-8f358a623271-test',
```
narcis-radu, line 31: We will need help from @salonijain3, @bandana147 to get new IDs for Privacy, the `aem.page|live` need allowlisting.

nala/utils/pr.run.sh
```diff
@@ -33,6 +33,7 @@ toRepoName=${repoParts[1]}
 prRepo=${prRepo:-$toRepoName}
 prOrg=${prOrg:-$toRepoOrg}
 
+# TODO ADD HLX5 SUPPORT
```
narcis-radu, line 36: ?

nala/utils/global.setup.js
mokimo: I was wondering about it, but given that it works 🤔 
@skumar09 is this exported and used by consumers, or if the nala tests pass on milo, we are good?

test/blocks/merch/mocks/embed-utils.js
mokimo: If the tests import a module that imports SLD, it is needed yes. Since this is a replication of the utils, it should be exported similar to the utils

test/blocks/caas/mocks/utils.js
mokimo: If the tests import a module that imports SLD, it is needed yes. Since this is a replication of the utils, it should be exported similar to the utils

libs/scripts/scripts.js
mokimo: Yeah, we'll need to live with the mix for now 😢

libs/blocks/bulk-publish-v2/utils.js
mokimo: When we refer to the milo repo itself, we don't need a check since that has hlx5 enabled

nala/utils/pr.run.sh
mokimo: I'll need @skumar09 help here - to ensure we test on `aem.page/aem.live` for migrated consumers and `hlx.page/hlx.live` for non migrated ones.

libs/blocks/quiz-entry/quiz-entry.css
```diff
@@ -178,7 +178,7 @@
 
 .carousel-arrow {
   background-color: #fff;
-  background-image: url('https://main--milo--adobecom.hlx.page/libs/blocks/carousel/img/arrow.svg');
+  background-image: url('https://main--milo--adobecom.aem.page/libs/blocks/carousel/img/arrow.svg');
```
auniverseaway, line 181: Not sure where you want to capture this, but the original is wrong. We should never use Helix URLs on production.

@colloyd do you know where this might be used on production? I want to show that we are making requests to the wrong place.

auniverseaway:
@mokimo incredible work here. Very cool to see.

libs/blocks/quiz-entry/quiz-entry.css
colloyd: @auniverseaway This is live on this page: https://www.adobe.com/creativecloud.html

Brad handled all of this carousel stuff originally, so he could say more about where the urls for these arrows came from. 

All of that said, there is a redesign of this quiz-entry block going on in which they are pulling the carousel (primarily due to accessibility / ux concerns), so these icons are going to be going away soon.

fullcolorcoder, line 181: @auniverseaway good catch. That was unfortunately an oversight on my end. I've provided a fix here consistent with the other historical fixes to the same problem.

```suggestion
background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='10' height='16' viewBox='0 0 10 16'%3E%3Cdefs%3E%3Cstyle%3E.a%7Bfill%3A%23f0f%3Bopacity%3A0%3B%7D.b%7Bfill%3A%23767676%3B%7D%3C/style%3E%3C/defs%3E%3Cg transform='translate(10 16) rotate(180)'%3E%3Crect class='a' width='10' height='16'/%3E%3Cpath class='b' d='M9.115,13.853l0,0L3.051,8,9.117,2.15l0,0A1.246,1.246,0,1,0,7.386.353l0,0L7.367.366h0L.382,7.1l0,0a1.237,1.237,0,0,0,0,1.794l0,0,6.982,6.732,0,0,.015.014,0,0a1.246,1.246,0,1,0,1.73-1.794Z'/%3E%3C/g%3E%3C/svg%3E");
```

fullcolorcoder (approved):
I suggested a fix to make a css rule better that Chris surfaced, not fully related to your changes but we shouldn't have .hlx urls in css if you'd like to include that.

libs/scripts/scripts.js
robert-bogos: Agree, we'll need to update at one point, but we can go live with this for now.

robert-bogos (approved):
Nice work 👍

mokimo:
Closed by https://github.com/adobecom/milo/pull/3140 which lives within the adobecom org to simplify some processes

PR #3128, opened 2024-11-01
MWPW-161174 fixed library query param

cod17828:
Hi, our `.hlx.live` page is protected with an access.live.allow rule in .config. and because of this we are not able to add our blocks to the library in sidekick. You can find more information here: https://github.com/orgs/adobecom/discussions/3121 

We found that there is a possibility to fetch library blocks data from custom url by adding `library` query param to library URL in `config.json`, but we found the bug in `library-config.js` and because of this bug, custom library URL will be fetched two times, and milo library data will be skipped. 
We propose solution that when `library` query param is present, additional blocks data will  be fetched from custom url specified in this param and then added to existing milo blocks. 

Resolves: [MWPW-161174](https://jira.corp.adobe.com/browse/MWPW-161174)

**Test URLs:**
- After: https://library-sidekick--milo--cod17828.hlx.page/?martech=off
**NOTE** my branch link doesn't work i have some issue with setting up AEM sync code

### DIFF
--- a/libs/blocks/library-config/library-config.js
+++ b/libs/blocks/library-config/library-config.js
@@ -112,9 +112,7 @@ async function loadList(type, content, list) {
   }
 }
 
-async function fetchLibrary(domain) {
-  const { searchParams } = new URL(window.location.href);
-  const suppliedLibrary = searchParams.get('library');
+async function fetchLibrary(domain, suppliedLibrary) {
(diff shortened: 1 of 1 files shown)


mokimo (approved):

vhargrave (approved):
looks good!

PR #3132, opened 2024-11-04
MWPW-161273 Standalone Gnav needs a release cycle [Bundle]

sharmrj:
* In order to support versioning and a release cadence for the standalone gnav and footer, we've created a dist folder that contains the built files for the standalone gnav.
* This was done by making all dynamic imports of javascript explicit and relative
* ~~The region nav in the global footer was rrefactored to have it's own modal instead of depending on the milo modal implementation. This had to be done to avoid dynamic imports whose paths had to be generated at runtime, and so could not be picked up by the bundler. There are slight differences in how the css behaves with the new modal since it uses the native HTML element `dialog`.~~ The implementation was refactored to explicitly load everything in the region-nav.
* supercedes #3113

Related discussion: https://github.com/orgs/adobecom/discussions/3114

Resolves: [MWPW-161273](https://jira.corp.adobe.com/browse/MWPW-161273)

**Test URLs:**
- After: https://bundled-standalone-gnav--milo--adobecom.hlx.page/?martech=off

https://adobecom.github.io/nav-consumer/navigation.html?env=stage&navbranch=bundled-standalone-gnav&theme=dark&usebundle=true


libs/navigation/build.mjs
```diff
@@ -0,0 +1,20 @@
+import * as esbuild from 'esbuild'; // eslint-disable-line
```
mokimo, line 1: Just FYI, we also use `microbundle` for things such as creating the preact-htm bundle

libs/navigation/build.mjs
```diff
@@ -0,0 +1,20 @@
...
+import fs from 'node:fs';
+
+fs.rmSync('./dist/', { recursive: true, force: true });
+
+await esbuild.build({
+  entryPoints: ['navigation.js'],
+  bundle: true,
+  splitting: true,
+  format: 'esm',
+  minify: true,
```
mokimo, line 11: Do we want to minify the bundle, does it make any sense? Personally I love to debug in chrome directly using overwrites, that would make this a bit more of an entrance barrier for other folks to come and help out / troubleshoot.

mokimo (approved):
Cool, I would love if you also opened a discussion highlighting the whole approach of exporting a milo component as standalone one and the expected consumption model, alternatives, thoughts that went into it etc.

It would be very valuable to point to in the future!

libs/navigation/build.mjs
sharmrj: Yeah it can save us, overall, a fair few kilobytes (5-10kb), and we can look into generating sourcemaps to help with debugging, and so on. (Source maps would be the preferred way of doing this as well since the chunking will mess up our organization of the code anyway)

salonijain3:
@mokimo  @narcis-radu This needs to be merged to stage on 12/2 and to main to 12/5 to support the production deployment on 12/9. 
Release calendar and the cadence is also updated on : https://wiki.corp.adobe.com/pages/viewpage.action?pageId=3330887979

cc: @skambi

skumar09 (changes requested):
@sharmrj : Could you please also update the locators for Nala Feds tests, 
`features/feds/footer/footer.test`, as tests are failing 

cc: @sigadamvenkata @narcis-radu @mokimo

amitbikram (approved):

sharmrj:
There's a high security vulnerability alert on this PR but it's not in any code I've touched @mokimo

sharmrj:
We no longer need evergreen files in the `dist` folder, so they've been removed. If clients want to consume a specific version they can point to the link of that tarball, and if they want the evergreen code they can directly import and consume `libs/navigation/navigation.js`

libs/navigation/build.mjs
sharmrj: Since we're not committing the built files to the repo anymore. We won't need to worry about inflating the number of files with sourcemaps.

bandana147 (approved):

Deva309 (approved):

salonijain3:
This is high priority as this contains the bundling changes to be consumed by the Adobe Home team and follows this release calendar : https://wiki.corp.adobe.com/pages/viewpage.action?pageId=3330887979

prativas22:
The Standalone Gnav release cycle was confirmed, and the result looks good. The test result is added below for reference.
https://jira.corp.adobe.com/browse/MWPW-162426

PR #3140, opened 2024-11-05
Helix 5 support for milo blocks and features

mokimo:
Re-raise of https://github.com/adobecom/milo/pull/3122 from adobecom rather than my fork

### Description
- There are a few special cases, such as loc, where I'm unsure about implications of moving certain links to aem.page, hence those simply got a "TODO" for now.
- There are other special cases, such as unit tests, where currently porting to `aem.page` does not make any sense, as our default is `hlx.page` unless detected otherwise. Some other unit tests *needed* to be adapted.
- This should *not* change any production behavior and development behavior apart of potentially pulling content from `.aem.` (federal/milo) rather than `.hlx.`

### Testing instructions
both `https://main--milo--adobecom.aem.page/docs/library/kitchen-sink/` and `https://main--milo--adobecom.hlx.page/docs/library/kitchen-sink/` should work and have no differences.

Consuming pages should also not have any difference in experience: https://main--cc--adobecom.hlx.live/de/products/photoshop?milolibs=hlx5-upgrade--milo--adobecom

Lastly, all tools should also be checked (caas, preflight, loc) to ensure we didn't accidentally introduce any errors.

Resolves: [MWPW-157482](https://jira.corp.adobe.com/browse/MWPW-157482)

**Test URLs:**
- After: https://hlx5-upgrade--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/libs/scripts/delayed.js
+++ b/libs/scripts/delayed.js
@@ -27,10 +27,8 @@ export const loadJarvisChat = async (getConfig, getMetadata, loadScript, loadSty
 export const loadPrivacy = async (getConfig, loadScript) => {
   const acom = '7a5eb705-95ed-4cc4-a11d-0cc5760e93db';
   const ids = {
-    'hlx.page': '3a6a37fe-9e07-4aa9-8640-8f358a623271-test',
-    'aem.page': '3a6a37fe-9e07-4aa9-8640-8f358a623271-test',
-    'hlx.live': '926b16ce-cc88-4c6a-af45-21749f3167f3-test',
-    'aem.live': '926b16ce-cc88-4c6a-af45-21749f3167f3-test',
+    '.page': '3a6a37fe-9e07-4aa9-8640-8f358a623271-test',
+    '.live': '926b16ce-cc88-4c6a-af45-21749f3167f3-test',
   };
 
   const otDomainId = ids?.[Object.keys(ids)
--- a/libs/scripts/delayed.js
+++ b/libs/scripts/delayed.js
@@ -28,7 +28,9 @@ export const loadPrivacy = async (getConfig, loadScript) => {
   const acom = '7a5eb705-95ed-4cc4-a11d-0cc5760e93db';
   const ids = {
     'hlx.page': '3a6a37fe-9e07-4aa9-8640-8f358a623271-test',
+    'aem.page': '3a6a37fe-9e07-4aa9-8640-8f358a623271-test',
     'hlx.live': '926b16ce-cc88-4c6a-af45-21749f3167f3-test',
+    'aem.live': '926b16ce-cc88-4c6a-af45-21749f3167f3-test',
   };
 
   const otDomainId = ids?.[Object.keys(ids)
--- a/libs/scripts/delayed.js
+++ b/libs/scripts/delayed.js
@@ -27,8 +27,10 @@ export const loadJarvisChat = async (getConfig, getMetadata, loadScript, loadSty
 export const loadPrivacy = async (getConfig, loadScript) => {
   const acom = '7a5eb705-95ed-4cc4-a11d-0cc5760e93db';
   const ids = {
-    '.page': '3a6a37fe-9e07-4aa9-8640-8f358a623271-test',
-    '.live': '926b16ce-cc88-4c6a-af45-21749f3167f3-test',
+    'hlx.page': '3a6a37fe-9e07-4aa9-8640-8f358a623271-test',
+    'hlx.live': '926b16ce-cc88-4c6a-af45-21749f3167f3-test',
+    'aem.page': '01930689-3b6a-7d5f-9797-8df2c3901a05-test',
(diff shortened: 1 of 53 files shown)


JackySun9 (approved):

hparra (approved):

fullcolorcoder (approved):

skumar09 (approved):
@mokimo  : Nala update looks good 🥇

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-157482

libs/utils/utils.js
```diff
@@ -144,7 +144,7 @@ export const MILO_EVENTS = { DEFERRED: 'milo:deferred' };
 const LANGSTORE = 'langstore';
 const PREVIEW = 'target-preview';
 const PAGE_URL = new URL(window.location.href);
-const SLD = PAGE_URL.hostname.includes('.aem.') ? 'aem' : 'hlx';
+export const SLD = PAGE_URL.hostname.includes('.aem.') ? 'aem' : 'hlx';
```
chrischrischris, line 147: I know that `SLD` was already in use but SLD is kind of cryptic - maybe a more descriptive name?  `AEM_DOMAIN`?

libs/scripts/delayed.js
```diff
@@ -27,8 +27,8 @@ export const loadJarvisChat = async (getConfig, getMetadata, loadScript, loadSty
 export const loadPrivacy = async (getConfig, loadScript) => {
   const acom = '7a5eb705-95ed-4cc4-a11d-0cc5760e93db';
   const ids = {
-    'hlx.page': '3a6a37fe-9e07-4aa9-8640-8f358a623271-test',
-    'hlx.live': '926b16ce-cc88-4c6a-af45-21749f3167f3-test',
+    '.page': '3a6a37fe-9e07-4aa9-8640-8f358a623271-test',
+    '.live': '926b16ce-cc88-4c6a-af45-21749f3167f3-test',
```
chrischrischris, line 31: These don't require the sld anymore?

libs/scripts/scripts.js
```diff
@@ -41,6 +41,12 @@ const stageDomainsMap = {
     '^https://business.adobe.com/blog': 'https://main--bacom-blog--adobecom.hlx.page',
     '^https://www.adobe.com': 'origin',
   },
+  '^https://.*--milo--.*.aem.page': {
+    '^https://www.adobe.com/acrobat': 'https://main--dc--adobecom.hlx.page',
+    '^https://business.adobe.com(?!/blog)': 'https://business.stage.adobe.com',
+    '^https://business.adobe.com/blog': 'https://main--bacom-blog--adobecom.hlx.page',
```
chrischrischris, line 47: Why would stage not use .aem.?

libs/utils/utils.js
mokimo: "second level domain" 😁 (If I remember correct) I know @overmyheadandbody and myself had some discussions back when we introduced it and couldn't settle on anything great.

`AEM_DOMAIN` sounds a bit like it's a fully qualified domain 🤔 Theres quite a few references, I'm fine to change it - but we should settle on something on agreement.

I'm still fine with taking `AEM_DOMAIN` @overmyheadandbody opinion?

libs/scripts/delayed.js
mokimo: Using the same config on both `aem.page` and `hlx.page` and `.live` should be fine as long as its properly whitelisted.
@salonijain3 @bandana147 can confirm though as the FEDs owners

libs/scripts/scripts.js
mokimo: Copied it from the config above to adapt the regex!

robert-bogos: Hi Chris! This is just an example of how you can configure the `stageDomainsMap`. Consumers will decide if they want to convert to `.aem.` or `.hlx.` according to their specific use-cases.

libs/utils/utils.js
chrischrischris: Yeah not sure that `AEM_DOMAIN` is best, but I had to look up SLD acronym to figure out what that represented.  I think even `AEM_SLD` would be better.  

I guess with `.hlx.` being deprecated and [eventually end-of-lifed](https://www.aem.live/docs/deprecation#hlxlive-and-hlxpage), this is a temp const that will go away in the future so maybe it doesn't really matter.

libs/scripts/delayed.js
narcis-radu: @mokimo - I don't see that possible. We need separate IDs for aem.page and aem.live, I will try to get them for you.

narcis-radu:
```
aem.page - 01930689-3b6a-7d5f-9797-8df2c3901a05-test
aem.live - 01930691-c4e5-75ba-aa0e-721e1213c139-test
```

PR #3141, opened 2024-11-05
Temporary workflow to watch for any feds files being changed

sharmrj:
We need to be sure no files that affect the standalone gnav are merged on the 14th of November, 2024.  This workflow helps us identify PRs that touch standalone gnav related files.

**Test URLs:**
- After: https://feds-filecheck-workflow--milo--sharmrj.hlx.page/?martech=off

### DIFF
--- a/.github/workflows/check-feds-files.yaml
+++ b/.github/workflows/check-feds-files.yaml
@@ -2,7 +2,7 @@ name: Feds File Watch
 
 on:
   pull_request_target:
-    types: [opened, edited, reopened]
+    types: [labeled]
 
 env:
   FEDS_WATCH_HOOK: ${{ secrets.FEDS_WATCH_HOOK }}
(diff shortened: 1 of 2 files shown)


salonijain3 (approved):

.github/workflows/check-feds-files.yaml
```diff
@@ -0,0 +1,25 @@
+name: Feds File Watch
+
+on:
+  pull_request_target:
+    types: [opened, edited, reopened]
```
mokimo, line 5: Might get a little spammy if you always send a message on edited... maybe on open and merge, or when the ready for stage label is applied?

sharmrj: I changed it to run only on labeled and on open(without checking the label to avoid the back and forth in case I get the label name wrong).  That way we get a run if labels are added when the PR is raised and one guaranteed run when ready for stage is applied.

bandana147 (approved):

PR #3150, opened 2024-11-06
Setup project skeleton

Zweihand3r:
* Added basic project skeleton for locui-create

Resolves: [161642](https://jira.corp.adobe.com/browse/MWPW-161642)


Zweihand3r:
Test urls are not accessible. Will re-create PR

PR #3176, opened 2024-11-12
MWPW-160237: Caas Bulk Publisher v2 (Beta)

cmiqueo:
This is parallel beta version of the Caas BulkPublisher Tool

It provides a new and easier to use UI with predetermined Presets for the consumers.

Resolves: [MWPW-160237](https://jira.corp.adobe.com/browse/MWPW-160237) & [MWPW-161613](https://jira.corp.adobe.com/browse/MWPW-161613)

Preset Mode:
image

Advanced Mode
image

**Test URLs:**
- Old: https://bulkpublisher-beta--milo--adobecom.hlx.page/tools/send-to-caas/bulkpublisher.html
- New: https://bulkpublisher-beta--milo--adobecom.hlx.page/tools/send-to-caas/bulkpublisher.beta.html


sheridansunier (approved):

sanrai (approved):
Should be good as this is a beta version and not merged into main bulk publisher yet.

tools/send-to-caas/bulk-publish-to-caas.beta.js
```diff
@@ -0,0 +1,543 @@
...
+  publishToFloodgate.value = 'default';
+  /* eslint-enable no-undef */
+};
+
+/* eslint-disable no-nested-ternary */
+const useDarkTheme = localStorage.getItem('bp-theme') === 'dark'
+  ? true
+  : localStorage.getItem('bp-theme') === 'light'
+    ? false
+    : window.matchMedia('(prefers-color-scheme: dark)').matches;
```
chrischrischris, line 347: Better to avoid nested ternary.  Something like:
```suggestion
const theme = localStorage.getItem('bp-theme');
const useDarkTheme = theme === 'dark' ||
  (theme !== 'light' && window.matchMedia('(prefers-color-scheme: dark)').matches);
```

tools/send-to-caas/bulk-publish-to-caas.beta.js
```diff
@@ -0,0 +1,543 @@
...
+/* eslint-enable no-nested-ternary */
+
+if (useDarkTheme) {
+  document.querySelector('.bulk-publisher').classList.add('dark');
+  document.querySelector('#option-dark').checked = true;
+} else {
+  document.querySelector('#option-light').checked = true;
+}
```
chrischrischris, line 355: nit break out into a function called from `init`

chrischrischris (commented):
Looks great!

I'd prefer if all the root level code be moved out into it's own function(s).  Even if just a giant `setup` function called first thing from `init`.  Or if it needs to run before init, the `setup` function can be called from the root at bottom of the file near the init declaration.

Even better would be to break out each setup item into its own function so you get something like:
```
function setup() {
  getPresets();
  checkDarkTheme();
  setupListeners();
  etc...
}
```

PR #3202, opened 2024-11-14
Fed Icons - Table icon decoration and collapse icon wrapping

Sartxi:
In reference to the [new federal icons work that was reverted](https://github.com/adobecom/milo/pull/2986), icons within tables were not having their respective SVG assets injected or tooltips being decorated. This was because the table block moves the icons around. I was able to see the icons being decorated by reselecting all icons before passing to icon js to be decorated and loaded.

There is also an issue where the expand/collapse icons wrapped to the bottom because of the new node index margin that was added. Unsetting that for this individual icon in the table.css.

Resolves: [MWPW-140452](https://jira.corp.adobe.com/browse/MWPW-140452)

**Test URLs:**
- Before: https://rparrish-fed-icons--milo--adobecom.hlx.page/drafts/sarchibeque/table-icons?martech=off
- After: https://sartxi-fed-icons-tables--milo--adobecom.hlx.page/drafts/sarchibeque/table-icons?martech=off

**URLs where issue was uncovered:**
- Before: https://main--dc--adobecom.hlx.page/acrobat/pricing?milolibs=rparrish-fed-icons&martech=off
- After: https://main--dc--adobecom.hlx.page/acrobat/pricing?milolibs=sartxi-fed-icons-tables&martech=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -1322,8 +1322,9 @@ export async function loadArea(area = document) {
 
   if (allIcons.length) {
     const { default: loadIcons, decorateIcons } = await import('../features/icons/icons.js');
-    await decorateIcons(area, allIcons, config);
-    await loadIcons(allIcons);
+    const areaIcons = area.querySelectorAll('span.icon');
+    await decorateIcons(area, areaIcons, config);
+    await loadIcons(areaIcons);
   }
 
   const currentHash = window.location.hash;
--- a/libs/blocks/table/table.css
+++ b/libs/blocks/table/table.css
@@ -351,6 +351,7 @@
   width: 15px;
   height: 15px;
   cursor: pointer;
+  margin-inline: unset;
 }
 
 .table .section-head-title:hover .icon.expand {


libs/utils/utils.js
```diff
@@ -1322,8 +1322,9 @@ export async function loadArea(area = document) {
 
   if (allIcons.length) {
     const { default: loadIcons, decorateIcons } = await import('../features/icons/icons.js');
-    await decorateIcons(area, allIcons, config);
-    await loadIcons(allIcons);
+    const areaIcons = area.querySelectorAll('span.icon');
+    await decorateIcons(area, areaIcons, config);
+    await loadIcons(areaIcons);
```
overmyheadandbody, line 1327: Do we need to "duplicate" this or could we bring L1306-1309 in here and just run `querySelectorAll` once?

Sartxi: I believe the code in L1306-1309 is there to address CLS concerns.

Sartxi: Yep I just verified that when I move it there is CLS. Not sure if there is another way to address that.

overmyheadandbody (approved):

mokimo (approved):

ryanmparrish (approved):
Looks good. Heads up - the before/after URLs in the description have a malformed url param: `?` instead of `&` on the second param `martech=off`.

mokimo (approved):

PR #3210, opened 2024-11-18
MWPW-159924 [CAI] | Integrate SDK (PIN & Modal) into Homepage

sharmrj:
* CAI-enabled images are identified by looking for the `CAI` prefix on the img alt text (delimited from the actual alt text by a `|`)
* After each section loads, If any image in the section is CAI enabled, an event listener is registered on the img so that on hover the src of the img is stripped of all its query parameters. This makes it so that if someone right clicks on the image and saves it, the image they download has the correct C2PA metadata in it [0].
* CAI is enabled only on viewports larger than 900px.
* The cai block is loaded and initialized. An event listener is registered to start preloading metadata when the image is hovered over. This is required since we lazy load all the metadata on demand (to minimize impact to the user and their potentially limited bandwidth and computing power) and this may require network requests to remote manifests.
* initCAI  is imported from another file, and initializes 
    1. the c2pa sdk,
    2. the logic for extracting the required data from the img,
    3. the DOM for the popover,
    4. and event listeners for displaying the popover.
* In order to initialize the c2pa sdk we host the `cai-toolkit.wasm` file and the `cai-worker.min.js`  files in the deps folder in milo. Since most of the computation takes place inside a worker spawned by `cai-worker.min.js`, we needn't worry about spawning our own worker thread.
* After the sdk has extracted the metadata (if the extraction fails for any reason we will display an error state on the popover). We need five pieces of information to build the DOM for the popover:
    1. The issuer of the manifest certificate – obtained from the `signatureInfo.issuer`  field of the `activeManifest`.
    2. The time the certificate was issued – obtained from the `signatureInfo.time`  field of the `activeManifest`.
    3. Information about whether AI was used when making the image – this is obtained by parsing the output of the `selectGenerativeInfo(activeManifest)` function from the c2pa sdk.
        * if we find the type `compositeWithTrainedAlgorithmicMedia`, it supercedes all other types, and we give back the corresponding message as found in https://cv.iptc.org/newscodes/digitalsourcetype/compositeWithTrainedAlgorithmicMedia
        * Else if we find the type trainedAlgorithmicMedia , we give back the corresponding message as found in https://cv.iptc.org/newscodes/digitalsourcetype/trainedAlgorithmicMedia
        * Else by default we say that an AI tool was not used in creating this image.
    4. The app used to generate or update the manifest – obtained using the `selectFormattedGenerator(activeManifest)` function from the c2pa sdk.
    5. AI tools used – obtained by doing [...new Set(selectGenerativeInfo(activeManifest).map(({ softwareAgent }) => softwareAgent))].join(', ') || 'None'; 
* Currently only `en_US` is supported

[0] - The query parameters tell the CDN which optimized version of the image needs to be served. Unfortunately the CDN strips away all metadata from the image when optimizing it, so the c2pa metadata is lost. However, without the query parameters the unoptimized image with the correct metadata is pointed to. In order to minimize LCP impact while still allowing the user to download the image and inspect its metadata, we do the first load with the optimized image and swap out the src of the `img` tag. The `source` tags are left untouched since only the src of the `img` tag matters when the user right clicks and downloads the image. This has been tested at different viewport sizes where different source sets should apply. 

Analytics will be implemented in a future PR.

Resolves: [MWPW-159924](https://jira.corp.adobe.com/browse/MWPW-159924)

**Test URLs:**
- After: https://cai--milo--sharmrj.aem.page/?martech=off

URL for QA (homepage clone): https://main--homepage--adobecom.hlx.page/homepage/drafts/raghavs/index-loggedout?milolibs=cai--milo--sharmrj

### DIFF
--- a/libs/blocks/cai/cai.js
+++ b/libs/blocks/cai/cai.js
@@ -38,7 +38,9 @@ const extractMetadata = async (img) => {
   const parseGenerativeInfo = (xs) => {
     // the only two options at this point are compositeWithTrainedAlgorithmicMedia
     // and trainedAlgorithmic media, and the former overrides the latter
-    const aiTool = [...new Set(xs.map(({ softwareAgent }) => softwareAgent))].join(', ') || 'None';
+    const aiTool = [...new Set(xs.map(({ softwareAgent }) => softwareAgent))]
+      .filter(Boolean)
+      .join(', ') || 'None';
     const composite = xs.find(({ type }) => type === 'compositeWithTrainedAlgorithmicMedia');
     if (composite) {
       return {


libs/blocks/cai/cai.js
```diff
@@ -0,0 +1,110 @@
...
+      `${miloLibs}/deps/cai-toolkit.wasm`,
+  workerSrc:
+      `${miloLibs}/deps/cai-worker.min.js`,
+});
+
+const extractMetadata = async (img) => {
+  const generativeInfoTypes = {
+    compositeWithTrainedAlgorithmicMedia: 'This image has been augmented, corrected or enhanced using a Generative AI model, such as with inpainting or outpainting operations',
+    trainedAlgorithmicMedia: 'This image has been created algorithmically using an Artificial Intellgence model trained on captured content',
+    default: 'An AI tool was not used in creating this image',
```
jenssingler, line 35: I think even if it's US only for now, these should be placeholders. If the wording needs to change, it's an authoring task rather than a DEV task.

libs/blocks/cai/cai.js
```diff
@@ -0,0 +1,110 @@
...
+  const tooltip = createTag('div');
+  tooltip.classList.add('hide');
+  tooltip.classList.add('cai-tooltip');
+  container.appendChild(tooltip);
+  const brick = container.closest('.homepage-brick');
+  // since homepage bricks have an anchor sibling element
+  // that covers the whole image, we need some special logic to
+  // handle that
+  (brick ?? container).addEventListener('pointerover', () => {
+    metadataPromise = metadataPromise ?? extractMetadata(img); // preload on hover
```
jenssingler, line 94: Nice - saw it working on the test page. The manifest is fetched as soon as you hover over the image.

test/blocks/cai/mocks/CAIManifest.json
```diff
@@ -0,0 +1,282 @@
+{
```
jenssingler, line 1: I think you have way too much unnecessary information in there. The tests were passing for me with a minimal file like this:
```json
{
  "manifestStore": {
    "activeManifest": {
      "title": "3-cr.png",
      "format": "image/png",
      "claimGenerator": "Adobe Photoshop/25.12.0 adobe_c2pa/0.12.2 c2pa-rs/0.32.5",
      "signatureInfo": {
        "issuer": "Adobe Inc.",
        "cert_serial_number": "123",
        "time": "2024-09-30T03:09:40+00:00"
      }
    }
  }
}
```
Probably still contains unused information, but I didn't bother trying to trim more.

jenssingler (approved):
Looks good from a feature perspective and contains most changes that have been discussed.
One question - I haven't seen the images swap from the optimized WebP version to the CR containing PNGs until hover. Is this intentional? 
I might misremember, but I thought the agreed upon process was to wait with fetching the manifest until hover, but switch the images to the CR images as soon as possible.

I'm setting to "Approve", but could you please minimize the mocked manifest file ?

Also - the before/after links don't contain anything CAI related (at least I did not see it).

But I did a comparison with:
Before: https://main--homepage--adobecom.hlx.page/homepage/drafts/raghavs/index-loggedout?martech=off
After: https://main--homepage--adobecom.hlx.page/homepage/drafts/raghavs/index-loggedout?martech=off&milolibs=cai--milo--sharmrj

And at least the LH scores did not change.

libs/blocks/cai/cai.js
sharmrj: Yes I'll turn these into placeholders.

PR #3222, opened 2024-11-19
MWPW-162739: Add white-space pre-wrap style to CCD suggested variant

Axelcureno:
Adds CSS changes to allow inline whitespaces and text in `prices` slot (CCD suggested card).

Resolves: [MWPW-162739](https://jira.corp.adobe.com/browse/MWPW-162739)

**Actual URL to test:** https://main--mas--adobecom.hlx.page/studio.html?milolibs=mwpw-162739--milo--adobecom#path=nala

**Test URLs:**
- After: https://mwpw-162739--milo--adobecom.aem.page/libs/features/mas/docs/ccd.html?martech=off


npeltier (approved):

yesil (approved):

3ch023 (changes requested):
could you run 'npm i' in the libs/features/mas 
and rebuild?
i am not sure what commerce.js is in the changed files, this might remove it

libs/features/mas/web-components/src/variants/ccd-suggested.js
```diff
@@ -112,6 +112,7 @@ export class CCDSuggested extends VariantLayout {
       font-size: var(--consonant-merch-card-body-xs-font-size);
       line-height: var(--consonant-merch-card-body-xs-line-height);
       min-width: fit-content;
+      white-space: pre-wrap;
```
yesil, line 115: I recommend to set price slot as an inline-block or block instead of this.
image

yesil (changes requested):

3ch023 (approved):

3ch023 (changes requested):
@Axelcureno @yesil correct me if I am wrong, but we decided this should be fixed in Studio?
atm studio editor removes the space between prices that Author adds, this should be changed? Should we close this PR then?

Axelcureno:
> @Axelcureno @yesil correct me if I am wrong, but we decided this should be fixed in Studio? atm studio editor removes the space between prices that Author adds, this should be changed? Should we close this PR then?

That was the plan initially but then I discovered we were was using different DOM structure in prices slot which was causing this weird behavior. Fixing in studio will not be required with this CSS fix.

yesil:
@Axelcureno this has been addressed in MWPW-163479 at https://github.com/adobecom/milo/blob/MWPW-163479/libs/features/mas/src/variants/ccd-suggested.css.js#L14.

PR #3232, opened 2024-11-21
New modal option to support tall videos

rgclayton:
**New tall video modal option**
* Adds ability to display tall/portrait style videos in modal
* Opt in feature - add `modal-metadata` block to fragment and apply new `style` "`tall-video`"

This new feature is enabled by adding a `modal-metadata` block to a fragment that will be used as a modal. Adding a new `style` value of "`tall-video`" will enable the feature. 

This use case is for the Photoshop mobile launch.
Resolves: [MWPW-162482](https://jira.corp.adobe.com/browse/MWPW-162482)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rclayton/tall-video?martech=off
- After: https://modal-tall-video--milo--adobecom.hlx.page/drafts/rclayton/tall-video?martech=off

--- 

image

image

### DIFF
--- a/libs/blocks/modal/modal.css
+++ b/libs/blocks/modal/modal.css
@@ -1,5 +1,7 @@
 :root {
   --modal-focus-color: #109cde;
+  --spacing-xl: 56px;
+  --aspect-ratio-tall: 9/16;
 }
 
 .dialog-modal {
@@ -220,10 +222,31 @@
   overflow: hidden;
 }
 
+.dialog-modal.tall-video {
+  line-height: 0;
+  width: fit-content;
+}
+
+.dialog-modal.tall-video .milo-video {
+  aspect-ratio: var(--aspect-ratio-tall);
+  height: auto;
+  padding-bottom: 0;
+}
+
 @media (min-width: 600px) {
   .dialog-modal {
     max-width: 80vw;
     width: fit-content;
+  }
+
+  .dialog-modal.tall-video {
+    max-width: 45vw;
+    padding: var(--spacing-xl);
+  }
+
+  .dialog-modal.tall-video .milo-video {
+    min-width: 45vw;
+    aspect-ratio: var(--aspect-ratio-tall);
   }
 
   .region-selector .content {
@@ -314,6 +337,37 @@
     max-width: 1100px;
     overflow: hidden;
     width: 80%;
+  }
+
+  .dialog-modal.tall-video {
+    max-width: 35vw;
+  }
+
+  .dialog-modal.tall-video .milo-video {
+    min-width: 35vw;
+  }
+}
+
+@media (min-width: 1400px) {
+  .dialog-modal.tall-video {
+    max-width: 25vw;
+    max-height: 1200px;
+  }
+
+  .dialog-modal.tall-video .milo-video {
+    min-width: 25vw;
+    max-height: 1200px;
+    height: 100%;
+  }
+}
+
+@media (min-width: 2000px) {
+  .dialog-modal.tall-video {
+    max-width: 22vw;
+  }
+
+  .dialog-modal.tall-video .milo-video {
+    min-width: 22vw;
   }
 }
 
--- a/libs/blocks/modal/modal.css
+++ b/libs/blocks/modal/modal.css
@@ -1,7 +1,5 @@
 :root {
   --modal-focus-color: #109cde;
-  --spacing-xl: 56px;
-  --aspect-ratio-tall: 9/16;
 }
 
 .dialog-modal {
@@ -231,6 +229,7 @@
   aspect-ratio: var(--aspect-ratio-tall);
   height: auto;
   padding-bottom: 0;
+  min-width: 90vw;
 }
 
 @media (min-width: 600px) {
@@ -240,13 +239,16 @@
   }
 
   .dialog-modal.tall-video {
-    max-width: 45vw;
+    --modal-width-var: 50vw;
+
+    max-width: var(--modal-width-var);
     padding: var(--spacing-xl);
   }
 
   .dialog-modal.tall-video .milo-video {
-    min-width: 45vw;
-    aspect-ratio: var(--aspect-ratio-tall);
+    --modal-width-var: 50vw;
+
+    min-width: var(--modal-width-var);
   }
 
   .region-selector .content {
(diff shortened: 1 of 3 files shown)


salonijain3:
@spadmasa will this impact the footer modal at all?

spadmasa:
Hi @salonijain3  , No impact on footer modal
https://main--cc--adobecom.hlx.page/creativecloud?milolibs=modal-tall-video#langnav
{6642AFDC-31BD-4A71-B9E2-7744418F84B4}

libs/blocks/modal/modal.css
```diff
@@ -1,5 +1,7 @@
 :root {
   --modal-focus-color: #109cde;
+  --spacing-xl: 56px;
```
overmyheadandbody, line 3: Shouldn't we scope this to the modal element? Otherwise we'd essentially overwrite https://github.com/adobecom/milo/blob/stage/libs/styles/styles.css#L49 on pages using modals. For viewports wider than `1200px`, the variable's value will match yours anyway - https://github.com/adobecom/milo/blob/stage/libs/styles/styles.css#L171.

So: do we need this value to be different? If so, let's scope it to `.dialog-modal`.

libs/blocks/modal/modal.css
```diff
@@ -1,5 +1,7 @@
...
+  --aspect-ratio-tall: 9/16;
```
overmyheadandbody, line 4: Is this set in stone? Do we foresee use-cases for other values, justifying having this authorable as well?

libs/blocks/modal/modal.css
```diff
@@ -220,12 +222,33 @@
...
   }
 
+  .dialog-modal.tall-video {
+    max-width: 45vw;
+    padding: var(--spacing-xl);
+  }
+
+  .dialog-modal.tall-video .milo-video {
+    min-width: 45vw;
+    aspect-ratio: var(--aspect-ratio-tall);
```
overmyheadandbody, line 249: This should have no effect, as it's already declared for all viewports on L231.

libs/blocks/modal/modal.css
```diff
@@ -315,6 +338,37 @@
...
+}
+
+@media (min-width: 2000px) {
+  .dialog-modal.tall-video {
+    max-width: 22vw;
+  }
+
+  .dialog-modal.tall-video .milo-video {
+    min-width: 22vw;
+  }
```
overmyheadandbody, line 371: These feel a bit repetitive throughout the file. How about we just declare:
```
  .dialog-modal.tall-video {
    max-width: var(--modal-width-variable);
  }

  .dialog-modal.tall-video .milo-video {
    min-width: var(--modal-width-variable);
  }
```
and declare the variables themselves to adapt based on viewport?

overmyheadandbody (commented):
Can we adjust things a bit for tablet devices using landscape? At the moment the modal window gets cut off a bit:
modal-tall-video--milo--adobecom hlx page_drafts_rclayton_tall-video_martech=off(iPad Mini)

libs/blocks/modal/modal.css
rgclayton: >So: do we need this value to be different? If so, let's scope it to .dialog-modal.

No, I was just trying carry that variable over to this modal. I must have missed it being accessible already. I'll make the correction.

libs/blocks/modal/modal.css
rgclayton: This is the only use-cause I have at the moment. Specifically for tall videos, which do not display well in our current modals.

I'll update the spacing token as mentioned in above comment.

libs/blocks/modal/modal.css
rgclayton: Interesting. I'll give that a try.

libs/blocks/modal/modal.css
ryanmparrish: Rather then redeclaring this spacing token,  can you just use `--spacing-xl-static` in styles.css?

libs/blocks/modal/modal.css
ryanmparrish: Same with this, we already declare this aspect ratio in styles.css. Seem like we shouldn't need this var

libs/blocks/modal/modal.css
rgclayton: Updated.

rgclayton:
@spadmasa Updates pushed to correct tablet view.

Sartxi (approved):

libs/blocks/modal/modal.css
```diff
@@ -220,12 +220,37 @@
...
 
+  .dialog-modal.tall-video {
+    --modal-width-var: 50vw;
+
+    max-width: var(--modal-width-var);
+    padding: var(--spacing-xl);
+  }
+
+  .dialog-modal.tall-video .milo-video {
+    --modal-width-var: 50vw;
```
overmyheadandbody, line 249: Already declared on L242, shouldn't be required here as well.

overmyheadandbody (commented):
Could we get some confirmation that this is the desired look on tablets? Now it looks quite small
modal-tall-video--milo--adobecom hlx page_drafts_rclayton_tall-video_martech=off(iPad Mini) (1)

libs/blocks/modal/modal.css
rgclayton: We may need to circle back to this for a better plan. I don't have specs or answers for all of these questions right now and the feature is an opt in feature.

The main use case and priority right now is phone, along with this update being in production next week.

overmyheadandbody: Then let's just remove the duplicate `--modal-width-var` declaration and we can follow-up with any enhancements in the future.

rgclayton: @overmyheadandbody, updated. Thanks for the help.

overmyheadandbody (approved):
Thanks for all the updates, looks good!

spadmasa:
Validation done on the pr , details are added in subtask (https://jira.corp.adobe.com/browse/MWPW-162855)
https://main--cc--adobecom.hlx.page/drafts/souj/carousel/editorialcards?milolibs=modal-tall-video
{2C0B05C9-AA37-48D2-8855-49D773C9EF24}
{6F2C6704-032B-45AC-9BB0-71391EDAA9CE}
validation in devices
https://modal-tall-video--milo--adobecom.hlx.page/drafts/rclayton/tall-video?martech=off
{0C709824-A288-4FFB-AAF9-88ECDBF29ABB}
{7D1E6D52-12FF-4AAA-87E6-C3505D47D658}
{A88E0460-2B05-4C7D-B2DD-68397A83BF34}
{83DE202D-C8C6-4215-A592-6869DF205819}
{1A51F221-C478-499F-A191-DBB7C3576AE2}

PR #3239, opened 2024-11-21
MWPW-162905 [Analytics] Analytics for content in tab containers needs improvement

vgoodric:
Tabbed content analytics needs several improvements:

reports block name as "tab" versus the block in the tab or the correct block number
does not report what tab it's in (will use t prefix and number)
tab names will increase the analytic string size.  so martech-metadata needs to be useable in US so we can provide short, unique nicknames for tab names

Resolves: [MWPW-162905](https://jira.corp.adobe.com/browse/MWPW-162905)

QA notes: this page has nested tabs.  If you use the hawkeye addon and hover over the first pods buy button to see the title
Before: Buy now-2--Acrobat Standard|s2|b1|tabs|nopzn|pricing-devicet
After: Buy now-2--Acrobat Standard|s2|b1|t1ind--t1abm--merch-card|nopzn|pricing-devicet

**Test URLs:**
- Before: https://main--dc--adobecom.hlx.page/acrobat/pricing?mep=%2Facrobat%2Fpricing-devicetype.json--default---%2Fdrafts%2Fmepdev%2Ffragments%2F2024%2Ftabsanalyticsudpate%2Fmanifest.json--all
- After: https://main--dc--adobecom.hlx.page/acrobat/pricing?mep=%2Facrobat%2Fpricing-devicetype.json--default---%2Fdrafts%2Fmepdev%2Ffragments%2F2024%2Ftabsanalyticsudpate%2Fmanifest.json--all&milolibs=tabsanalyticsudpate
- Psi-check: https://tabsanalyticsudpate--milo--adobecom.aem.page/?martech=off


libs/blocks/tabs/tabs.js
```diff
@@ -295,6 +296,10 @@ const init = (block) => {
         assocTabItem = rootElem.querySelector(`#tab-panel-${id}-${val}`);
       }
       if (assocTabItem) {
+        const tabLabel = tabListItems[val - 1]?.innerText;
```
denlight, line 299: is val always available and is always a number ? The method that gets `val` is called `getStringKeyName`, is the returned value a string ?

libs/martech/attributes.js
```diff
@@ -95,7 +95,15 @@ export async function decorateSectionAnalytics(section, idx, config) {
       block.setAttribute('daa-lh', `${lhAtt}${mepMartech}`);
     } else {
       const blockName = block.classList[0] || '';
-      block.setAttribute('daa-lh', `b${blockIdx + 1}|${blockName.slice(0, 15)}${mepMartech}`);
+      let closest = block;
+      let nestedLH = '';
+      while (closest) {
```
denlight, line 100: this looks like a forever loop. If `closest` resolves once it will always resolve, and the loop will never end, right? Is this intended?

libs/blocks/tabs/tabs.js
vgoodric: so I think it's always a string version of a number. example: '1'. It's parsing the value in that table which is always the id and index number separated by a comma. example: 'acrobat, 1'.
Since this is nested in `if (assocTabItem) {`, it means it found the item on 296. So it is very unlikely to not be a number if they reach this code.  But line 300 tests it did not come back undefined.

libs/martech/attributes.js
vgoodric: line 101 is `closest = closest.parentNode?.closest('[data-nested-lh]');`, so it will always run the loop once, but it immediately redefines it.  Adding an alternative below.  Tell me what you think.

markpadbe: @denlight, on line 101, for each pass of the while loop, the variable closest gets reassigned to closest.parentNode?.closest('[data-nested-lh]') so closest should eventually get set to null.

markpadbe (approved):

markpadbe (approved):

libs/blocks/tabs/tabs.js
vgoodric, line 299:
```suggestion
        const tabLabel = isNaN(val) ? tabListItems[val]?.innerText  tabListItems[val - 1]?.innerText;
```

denlight (approved):

ivanvatadobe (approved):

PR #3243, opened 2024-11-22
Poster play button update

rgclayton:
Update to change the poster image play button when linking to a modal.

* Change the play button icon and color

Resolves: [MWPW-162410](https://jira.corp.adobe.com/browse/MWPW-162410)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rclayton/poster-play-button?martech=off
- After: https://poster-play-icon--milo--adobecom.hlx.page/drafts/rclayton/poster-play-button?martech=off
image

### DIFF
--- a/libs/styles/consonant-play-button.css
+++ b/libs/styles/consonant-play-button.css
@@ -1,19 +1,19 @@
 
 :root {
-  --play-circle-size-s: 40px;
+  --play-circle-size-s: 44px;
   --play-circle-size-m: 72px;
   --play-circle-size-l: 96px;
   --play-border-width-s: 2px;
   --play-border-width-m: 2.5px;
   --play-border-width-l: 3px;
-  --play-icon-size-s: calc(var(--play-circle-size-s) + 2 * var(--play-border-width-s));
-  --play-icon-size-m: calc(var(--play-circle-size-m) + 2 * var(--play-border-width-m));
-  --play-icon-size-l: calc(var(--play-circle-size-l) + 2 * var(--play-border-width-l));
-  --play-triangle-width-s: 13.5px;
+  --play-icon-size-s: var(--play-circle-size-s);
+  --play-icon-size-m: var(--play-circle-size-m);
+  --play-icon-size-l: var(--play-circle-size-l);
+  --play-triangle-width-s: 16px;
   --play-triangle-height-s: 18px;
-  --play-triangle-width-m: 24px;
+  --play-triangle-width-m: 30px;
   --play-triangle-height-m: 32px;
-  --play-triangle-width-l: 31.5px;
+  --play-triangle-width-l: 40px;
   --play-triangle-height-l: 42px;
 }
 
@@ -35,11 +35,11 @@
   align-items: center;
   justify-content: center;
   cursor: pointer;
-  background-color: var(--link-color-dark);
+  background-color: rgb(0 0 0 / 60%);
   box-sizing: border-box;
-  border: var(--play-border-width-m) solid var(--color-white);
   border-radius: 50%;
-  outline: none;
+  outline: 0 solid var(--color-black);
+  transition: border .2s ease, outline .1s ease;
 }
 
 .modal-img-link .consonant-play-btn.btn-small {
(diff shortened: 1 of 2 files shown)


libs/styles/consonant-play-button.css
```diff
@@ -51,14 +51,16 @@
...
 
 .modal-img-link .consonant-play-btn:hover {
-  background-color: var(--color-accent-down);
+  background-color: rgb(0 0 0 / 100%);
 }
 
 .modal-img-link .consonant-play-btn:focus,
 .modal-img-link .consonant-play-btn:focus-visible {
-  outline: var(--play-border-width-m) solid var(--link-color-dark);
+  outline: var(--play-border-width-l) solid var(--color-black);
```
overmyheadandbody, line 59: Since just the size is changing, could we not declare just `outline-width`?

libs/utils/image-video-link.js
```diff
@@ -1,7 +1,7 @@
 import { getConfig, loadStyle, createTag } from './utils.js';
 
-const PLAY_ICON_SVG = `<svg xmlns="http://www.w3.org/2000/svg" width="24" height="32" viewBox="0 0 24 32" fill="none" class="play-icon">
-                        <path d="M24 16.0005L0 32L1.39876e-06 0L24 16.0005Z" fill="white"/>
+const PLAY_ICON_SVG = `<svg xmlns="http://www.w3.org/2000/svg" width="40" height="42" viewBox="0 0 40 42" fill="none" class="play-icon">
```
overmyheadandbody, line 3: Since this has no `title`, is it decorative? If so, let's add a `aria-hidden="true"` attribute.

overmyheadandbody (approved):
Just a couple of small comments, otherwise looking good!

overmyheadandbody:
I'm also guessing the desire is to update the style of this button _everywhere_, not just tall videos, right?

libs/styles/consonant-play-button.css
rgclayton: The color changed too. From blue to black.

rgclayton:
> I'm also guessing the desire is to update the style of this button _everywhere_, not just tall videos, right?

Currently, just this play button that launches modal. It's direct support for the mobile photoshop launch marketing.

Sartxi (approved):

libs/styles/consonant-play-button.css
overmyheadandbody: What I mean by this is that we declare color and type on L41. This declaration just updates width. Could we not declare just the width here?

rgclayton: No. This black outline is added on the outside of the focus ring so the focus ring has much better accessibility contrast (this was suggested by Taryn). Removing the color from this line removes that black that creates this effect
image

overmyheadandbody: It's not all that important, but here's the change I was suggesting:
```suggestion
  outline-width: var(--play-border-width-l);
```
On L41, you already have `outline: 0 solid var(--color-black);`, so the style and color are set. You're only updating the width, so unless there are other overrides, you just need to re-declare the width here. Made those changes locally and the styles are maintained:
Screenshot 2024-11-25 at 12 22 36

overmyheadandbody (approved):

spadmasa:
Validation done on the pr on the below urls and seen rendered fine, detailed results are updated in ticket
https://jira.corp.adobe.com/browse/MWPW-162410
In cc
https://main--cc--adobecom.hlx.page/creativecloud/video/discover/j-cut-and-l-cut?milolibs=poster-play-icon
{606BF2A3-AC20-4C69-A34A-9788329A8FE0}
on hover
{95DE91FF-990E-41FF-9E6B-D07A90A34500}
on select
{FACC3D4E-E935-441E-8C07-6E1405DFF143}
In milo : https://poster-play-icon--milo--adobecom.hlx.page/drafts/anupama/pr/carousel/carousel-videos1
bacom : https://main--bacom--adobecom.hlx.page/products/adobe-analytics/digital-analytics?milolibs=poster-play-icon
{6B81E550-A454-4285-864F-ACDED6FE5899}
https://main--dc--adobecom.hlx.page/acrobat/generative-ai-pdf?milolibs=poster-play-icon

PR #3253, opened 2024-11-25
Handle CLS issue on localnav: mobile redesign

bandana147:
LocalNavs must include the string 'localnav' in their name; otherwise, they will encounter CLS issues with the mobile design for localnav.

* Adding a check for localnav string on gnav source and preserving space
* Adding Lana log if content is localnav but file name doesn't include localnav

Resolves: [MWPW-161998](https://jira.corp.adobe.com/browse/MWPW-161998)

**Test URLs:**
- After: https://mgnav-cls--milo--adobecom.aem.page/?martech=off

QA: 
- Without localnav string -  https://main--federal--adobecom.hlx.page/federal/dev/blaishram/page-with-localnav?milolibs=mgnav-cls
- With localnav string - https://main--federal--adobecom.hlx.page/federal/dev/blaishram/express/express?milolibs=mgnav-cls

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -761,7 +761,7 @@ export async function getGnavSource() {
   return url;
 };
 
-function decorateHeader() {
+async function decorateHeader() {
   const breadcrumbs = document.querySelector('.breadcrumbs');
   breadcrumbs?.remove();
   const header = document.querySelector('header');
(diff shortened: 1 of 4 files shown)


libs/utils/utils.js
```diff
@@ -772,7 +782,10 @@ function decorateHeader() {
   const dynamicNavActive = getMetadata('dynamic-nav') === 'on'
     && window.sessionStorage.getItem('gnavSource') !== null;
   if (!dynamicNavActive && (baseBreadcrumbs || breadcrumbs || autoBreadcrumbs)) header.classList.add('has-breadcrumbs');
-  if (getMetadata('is-localnav') === 'true' && getMetadata('mobile-gnav-v2') !== 'false') {
+  const gnavSource = await getGnavSource();
+  const gnavName = gnavSource.split('/').pop();
```
sharmrj, line 786: Do we really need to split? We could just match `\/localnav\/` in the regex on line 788 below

sharmrj (approved):

libs/utils/utils.js
bandana147: I would still prefer using split and pop since it is more readable. For allocation i can change it to 
`gnavSource.split('/').pop().match('localnav')`

sharmrj: `split` will allocate an array on the heap, but it should be negligible enough.

PR #3255, opened 2024-11-25
Allow setting of aria label text for links

overmyheadandbody:
This allows authors to set individual `aria-label` values for links in order to make them accessibility-compliant. A few things that were considered (also available on the test page):

- certain links may have non-text children, such as icons;
- some links may actually need to have a pipe in their text, so just the text after the last pipe is considered for the `aria-label` value;
- I've queried a few projects for the presence of `|` inside links and came short, meaning there should be no links currently on production that would get affected by this change;

Resolves: [MWPW-161396](https://jira.corp.adobe.com/browse/MWPW-161396)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ramuntea/accessibility/aria-label-with-icon?martech=off
- After: https://links-aria-label--milo--adobecom.hlx.page/drafts/ramuntea/accessibility/aria-label-with-icon?martech=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -714,12 +714,22 @@ export function decorateLinks(el) {
     }
 
     // Pattern: "Link Text | Aria Label"
-    if (/\s\|\s/.test(a.textContent)) {
-      const ogContent = a.textContent;
-      const split = ogContent.split(/\s\|\s/);
+    // If an icon is defined just before the pipe, there is no space before the pipe
+    const pipeRegex = /\|\s/;
+    if (pipeRegex.test(a.textContent)) {
+      // the anchor may already have elements inside, fetching the matching child nodes
+      const nodes = [...a.childNodes].filter((node) => pipeRegex.test(node.textContent));
+      // get the last matching text node with pipe character(s)
+      const node = nodes[nodes.length - 1];
+      // get its text content
+      const ogContent = node.textContent;
+      // get the last occurrence of the pipe character
+      const split = ogContent.split(pipeRegex);
       const ariaLabel = split[split.length - 1];
-      const text = ogContent.replace(` | ${ariaLabel}`, '');
-      a.textContent = text;
+      // Delete the aria label value from the original text
+      const text = ogContent.replace(new RegExp(`\\s?\\|\\s?${ariaLabel}`), '');
+      node.textContent = text;
+      // Set the aria label
       a.setAttribute('aria-label', ariaLabel);
     }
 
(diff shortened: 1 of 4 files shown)


mokimo (approved):

test/utils/mocks/body.html
```diff
@@ -39,6 +39,28 @@
...
     </p>
+    <!-- Aria labels appendment -->
+    <p>
+      <a href="https://www.adobe.com/" class="aria-label-none">Text</a>
+    </p>
+    <p>
+      <a href="https://www.adobe.com/" class="aria-label-simple">Text | Aria label</a>
+    </p>
+    <p>
+      <a href="https://www.adobe.com/" class="aria-label-piped">Text | Other text | Aria label</a>
```
narcis-radu, line 50: We should also test use-cases where there's no space between words separated by `|`.

libs/utils/utils.js
```diff
@@ -712,6 +712,17 @@ export function decorateLinks(el) {
     if (a.href.includes(copyEvent)) {
       decorateCopyLink(a, copyEvent);
     }
+    // Append aria-label
+    const pipeRegex = /\s?\|\s?/;
+    if (pipeRegex.test(a.textContent) && !/\.[a-z]+/i.test(a.textContent)) {
```
narcis-radu, line 717: How would this work for links with `.mp4`? Do we need to handle such an use-case?

overmyheadandbody: There's an example of that on the [test page](https://links-aria-label--milo--adobecom.hlx.page/drafts/ramuntea/accessibility/aria-label-with-icon?martech=off). If the link has an extension, the aria label logic will ignore it. The alternative would be to modify the auto-block logic to ignore anything that qualifies as an auto-block from having the aria-label logic applied as well

test/utils/mocks/body.html
overmyheadandbody: Added another unit test 👍

narcis-radu (approved):

libs/utils/utils.js
```diff
@@ -712,6 +712,17 @@ export function decorateLinks(el) {
...
+      const node = [...a.childNodes].reverse()
+        .find((child) => pipeRegex.test(child.textContent));
```
robert-bogos, line 719: The `Array.reverse()` in here seems redundant. Is there any reason behind it? 🤔

overmyheadandbody: Yes! We want to find the last occurrence that meets the pipe regex criteria. So instead of having more complicated logic of finding all matches and getting the last, I simply reverse the array and find the first match.

robert-bogos: Nice 👍

robert-bogos (approved):

PR #3258, opened 2024-11-25
[MWPW-158749] Enhance Interact call efficiency and defer non-essentia…

swamu:
- Enhance Interact call efficiency
- defer non-essential JavaScript to post LCP
- Introduce a tag personalization-v2, that currently enables this feature
- Move some common methods to mertech/helpers

Resolves: [MWPW-158749](https://jira.corp.adobe.com/browse/MWPW-158749)

**Test URLs:**

Before: https://main--milo--adobecom.aem.page/?martech=on
After: https://mwpw-158749-optimise-interact-call--milo--swamu.aem.page/?martech=on

### DIFF
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -469,7 +469,7 @@ function getSelectedElements(sel, rootEl, forceRootEl) {
   try {
     els = root.querySelectorAll(modifiedSelector);
   } catch (e) {
-  /* eslint-disable-next-line no-console */
+    /* eslint-disable-next-line no-console */
     log('Invalid selector: ', selector);
     return null;
   }
@@ -1117,12 +1117,7 @@ export const combineMepSources = async (persEnabled, promoEnabled, mepParam) => 
   return persManifests;
 };
 
-async function callMartech(config, targetInteractionPromise = null) {
-  const { getTargetPersonalization } = await import('../../martech/martech.js');
-  const {
-    targetManifests,
-    targetPropositions,
-  } = await getTargetPersonalization(targetInteractionPromise);
+function updateManifestsAndPropositions({ config, targetManifests, targetPropositions }) {
   config.mep.targetManifests = targetManifests;
   if (targetPropositions?.length && window._satellite) {
     window._satellite.track('propositionDisplay', targetPropositions);
@@ -1134,15 +1129,113 @@ async function callMartech(config, targetInteractionPromise = null) {
   return targetManifests;
 }
 
+function roundToQuarter(num) {
+  return Math.ceil(num / 250) / 4;
+}
+
+function calculateResponseTime(responseStart) {
+  const responseTime = Date.now() - responseStart;
+  return roundToQuarter(responseTime);
+}
+
+function sendTargetResponseAnalytics(failure, responseStart, timeoutLocal, message) {
+  // temporary solution until we can decide on a better timeout value
+  const responseTime = calculateResponseTime(responseStart);
+  const timeoutTime = roundToQuarter(timeoutLocal);
+  let val = `target response time ${responseTime}:timed out ${failure}:timeout ${timeoutTime}`;
+  if (message) val += `:${message}`;
+  // eslint-disable-next-line no-underscore-dangle
+  window._satellite?.track?.('event', {
+    documentUnloading: true,
+    xdm: {
+      eventType: 'web.webinteraction.linkClicks',
+      web: {
+        webInteraction: {
+          linkClicks: { value: 1 },
+          type: 'other',
+          name: val,
+        },
+      },
+    },
+    data: { _adobe_corpnew: { digitalData: { primaryEvent: { eventInfo: { eventName: val } } } } },
+  });
+}
+
+const handleAlloyResponse = (response) => {
+  const items = (
+    (response.propositions?.length && response.propositions)
+    || (response.decisions?.length && response.decisions)
+    || []
+  ).map((i) => i.items).flat();
+
+  if (!items?.length) return [];
+
+  return items
+    .map((item) => {
+      const content = item?.data?.content;
+      if (!content || !(content.manifestLocation || content.manifestContent)) return null;
+
+      return {
+        manifestPath: content.manifestLocation || content.manifestPath,
+        manifestUrl: content.manifestLocation,
+        manifestData: content.manifestContent?.experiences?.data || content.manifestContent?.data,
+        manifestPlaceholders: content.manifestContent?.placeholders?.data,
+        manifestInfo: content.manifestContent?.info.data,
+        name: item.meta['activity.name'],
+        variantLabel: item.meta['experience.name'] && `target-${item.meta['experience.name']}`,
+        meta: item.meta,
+      };
+    })
+    .filter(Boolean);
+};
+
+async function handleMartechTargetInteraction(
+  { config, targetInteractionPromise, calculatedTimeout },
+) {
+  let targetManifests = [];
+  let targetPropositions = [];
+  let responseStart;
+  if (enablePersonalizationV2() && targetInteractionPromise) {
+    try {
+      responseStart = Date.now();
+      const targetInteractionData = await targetInteractionPromise;
+      sendTargetResponseAnalytics(false, responseStart, calculatedTimeout);
+      const responseTime = calculateResponseTime(responseStart);
+      try {
+        window.lana.log(`target response time: ${responseTime}`, { tags: 'martech', errorType: 'i' });
+      } catch (e) {
+        // eslint-disable-next-line no-console
+        console.error('Error logging target response time:', e);
+      }
+      targetManifests = handleAlloyResponse(targetInteractionData.result);
+      targetPropositions = targetInteractionData.result?.propositions || [];
+    } catch (err) {
+      console.log('Oops!! Interact Call didnt go through', err);
+    }
+  }
+
+  return updateManifestsAndPropositions({ config, targetManifests, targetPropositions });
+}
+
+async function callMartech(config) {
+  const { getTargetPersonalization } = await import('../../martech/martech.js');
+  const {
+    targetManifests,
+    targetPropositions,
+  } = await getTargetPersonalization({ handleAlloyResponse, sendTargetResponseAnalytics });
+  return updateManifestsAndPropositions({ config, targetManifests, targetPropositions });
+}
+
 const awaitMartech = () => new Promise((resolve) => {
   const listener = (event) => resolve(event.detail);
   window.addEventListener(MARTECH_RETURNED_EVENT, listener, { once: true });
 });
 
-export async function init(enablements = {}, targetInteractionPromise = null) {
+export async function init(enablements = {}) {
   let manifests = [];
   const {
-    mepParam, mepHighlight, mepButton, pzn, promo, target, postLCP,
+    mepParam, mepHighlight, mepButton, pzn, promo,
+    target, targetInteractionPromise, calculatedTimeout, postLCP,
   } = enablements;
   const config = getConfig();
   if (postLCP) {
@@ -1167,26 +1260,17 @@ export async function init(enablements = {}, targetInteractionPromise = null) {
     if (pzn) loadLink(getXLGListURL(config), { as: 'fetch', crossorigin: 'anonymous', rel: 'preload' });
   }
 
-  if (target === true) {
-    let localManifests = [];
-    if (enablePersonalizationV2()) {
-      localManifests = await callMartech(config, targetInteractionPromise);
-    } else {
-      localManifests = await callMartech(config);
+  if (enablePersonalizationV2()) {
+    manifests = manifests.concat(await handleMartechTargetInteraction(
+      { config, targetInteractionPromise, calculatedTimeout },
+    ));
+  } else {
+    if (target === true) manifests = manifests.concat(await callMartech(config));
+    if (target === 'postlcp') callMartech(config);
+    if (postLCP) {
+      if (!config.mep.targetManifests) await awaitMartech();
+      manifests = config.mep.targetManifests;
     }
-    manifests = manifests.concat(localManifests);
-  }
-
-  if (target === 'postlcp') {
-    if (enablePersonalizationV2()) {
-      await callMartech(config, targetInteractionPromise);
-    } else {
-      callMartech(config);
-    }
-  }
-  if (postLCP) {
-    if (!config.mep.targetManifests) await awaitMartech();
-    manifests = config.mep.targetManifests;
   }
   if (!manifests || !manifests.length) return;
   try {
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -3,6 +3,7 @@
 
 import { createTag, getConfig, loadLink, loadScript, localizeLink } from '../../utils/utils.js';
 import { getFederatedUrl } from '../../utils/federated.js';
+import { enablePersonalizationV2 } from '../../martech/helpers.js';
 
 /* c8 ignore start */
 const PHONE_SIZE = window.screen.width < 550 || window.screen.height < 550;
@@ -773,7 +774,11 @@ async function getPersonalizationVariant(manifestPath, variantNames = [], varian
 
   let userEntitlements = [];
   if (hasEntitlementTag) {
-    userEntitlements = await config.entitlements();
+    if(enablePersonalizationV2()){
+      userEntitlements = [];
+    } else {
+      userEntitlements = await config.entitlements();
+    }
   }
 
   const hasMatch = (name) => {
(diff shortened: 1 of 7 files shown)


libs/martech/helpers.js
```diff
@@ -0,0 +1,494 @@
+const TARGET_TIMEOUT_MS = 4000;
```
mokimo, line 1: This will introduce a new network call and blocks the rendering pipeline pre LCP. It's likely the opposite of what you want by trying to make the pipeline more efficient

libs/utils/utils.js
```diff
@@ -1,5 +1,9 @@
 /* eslint-disable no-console */
 
+import { enablePersonalizationV2, loadAnalyticsAndInteractionData, getEnv, getMetadata } from '../martech/helpers.js';
+
+export { getMetadata };
```
mokimo, line 5: Milo utils are *always* loaded for every page on adobe.com. Martech, is loaded under certain conditions at a certain time. This introduces a blocking network call for every single page on adobe.com:

image

If we say the call really only takes 50ms as in this example (x10 this value for mobile with higher latency) and the current LCP is 500ms, this drops the performance by 10% - so there's a lot of gravity in introducing any blocking imports here.

mokimo (changes requested):
There's some fundamental performance penalties introduced with this, let's take this back to the drawing board!

libs/martech/helpers.js
swamu: @mokimo this file has the interact call which earlier was in launch and now is moved to milo, that enables a faster load of a few dependant scripts. Since all these to begin were pre lcp.

libs/utils/utils.js
```diff
@@ -1,5 +1,9 @@
 /* eslint-disable no-console */
 
+import { enablePersonalizationV2, loadAnalyticsAndInteractionData, getEnv, getMetadata } from '../martech/helpers.js';
```
amitbikram, line 3: Can we load helpers.js on demand?

amitbikram (changes requested):

libs/utils/utils.js
sharmrj: If we're willing to pollute the utils namespace with the code to make the interact call early, we maximize our performance benefits. We'll be trading organization for a performance improvement.

libs/utils/utils.js
```diff
@@ -1038,14 +990,26 @@ async function checkForPageMods() {
...
+  const enablePersV2 = enablePersonalizationV2();
+
   if (!(pzn || target || promo || mepParam
     || mepHighlight || mepButton || mepParam === '' || xlg)) return;
-  if (target || xlg) {
+
+  if (martech !== 'off' && (target || xlg || pzn) && enablePersV2) {
+    const { locale } = getConfig();
+    try{
+      targetInteractionData = await loadAnalyticsAndInteractionData({ locale });
```
sharmrj, line 1007: We don't need to `await` this promise here. We should fire it here, store the promise and pass it to `personalization.js`. That way we won't wait until this data is loaded to start loading `personalization.js`. they both can be fired almost in parallel.  Then we can await the promise right before we need it inside `personalization.js`

sharmrj: Granted you'd have to also move the error handling logic if you did that.

mokimo:
Is this really running on https://mwpw-158749-optimise-interact-call--milo--swamu.aem.page/?martech=on ?
Do you have a few pages where we expect the gains?

sharmrj:
> Is this really running on https://mwpw-158749-optimise-interact-call--milo--swamu.aem.page/?martech=on ? Do you have a few pages where we expect the gains?

@mokimo you can try https://main--cc--adobecom.hlx.live/products/photoshop?milolibs=mwpw-158749-optimise-interact-call--milo--swamu -- override utils.js andchange line 1064 to `return true` (inside the `enablePersonalizationV2` function).

sharmrj:
As a preliminary result using the .live photoshop page

### Before
Screenshot 2024-11-26 at 11 49 36 PM
**trace:**
[Trace-20241126T234951.json.zip](https://github.com/user-attachments/files/17924206/Trace-20241126T234951.json.zip)



### After:
Screenshot 2024-11-26 at 11 46 57 PM
**trace:** 
[Trace-20241126T234740.json.zip](https://github.com/user-attachments/files/17924186/Trace-20241126T234740.json.zip)

libs/utils/utils.js
```diff
@@ -1059,10 +1104,14 @@ async function checkForPageMods() {
   const { init } = await import('../features/personalization/personalization.js');
```
mokimo, line 1105: Let's parallelize those two calls 
image

By adding loadLink(`${getConfig().base}/utils/federated.js`, { rel: 'preload', as: 'script', crossorigin: 'anonymous' }); just before personalization on [L1058](https://github.com/adobecom/milo/pull/3258/files#diff-28218f239d8d9c4d41e1924854440b50964619e56058fdd0ae243ff5ff6c616dL1058)

Bonus points, if we actually only load federal if it's really needed within personalization. This might involve some refactoring and might be harder to do though.

libs/utils/utils.js
```diff
@@ -1023,29 +1025,72 @@ export async function loadMartech({
...
   const xlg = martech === 'off' ? false : getMepEnablement('xlg');
+
   if (!(pzn || target || promo || mepParam
     || mepHighlight || mepButton || mepParam === '' || xlg)) return;
-  if (target || xlg) {
+
+  const enablePersV2 = enablePersonalizationV2();
+  if (martech !== 'off' && (target || xlg || pzn) && enablePersV2) {
+    const { locale } = getConfig();
+    const { loadAnalyticsAndInteractionData } = await import('../martech/helpers.js');
```
mokimo, line 1089: Can we actually do this in a non blocking way as it was before?

### Before
image

### After
image

mokimo: You can create a Promise that will then import the helpers and gets the interaction call and only then resolves. This way you don't need to wait for the module to load before moving on to loading the personalization module

libs/utils/utils.js
```diff
@@ -1023,29 +1025,72 @@ export async function loadMartech({
...
+    targetInteractionPromise = loadAnalyticsAndInteractionData(
+      { locale, env: getEnv({})?.name, timeoutMeta: getMetadata('target-timeout') }
+    );
+    delayedMartech = true;
```
mokimo, line 1093: Should we already load or at least preload `martech.js` given the right conditions?

libs/utils/utils.js
```diff
@@ -1059,10 +1104,14 @@ async function checkForPageMods() {
   const { init } = await import('../features/personalization/personalization.js');
   await init({
     mepParam, mepHighlight, mepButton, pzn, promo, target,
-  });
+  }, targetInteractionPromise);
```
mokimo, line 1108: Why add an extra param if you already have a nice object-param you can add the targetInteractionPromise to?

mokimo (commented):
I noticed the XLG tags load twice 🤔 - this seems to be a preexisting issue - if that's trivial to fix might be nice to do that as well though given this PR focusses around martech/pers/mep/xlg etc `https://www.adobe.com/federal/assets/data/mep-xlg-tags.json?sheet=stage` 
`https://main--federal--adobecom.aem.live/federal/assets/data/mep-xlg-tags.json?sheet=stage`

libs/utils/utils.js
```diff
@@ -1059,20 +1105,25 @@ async function checkForPageMods() {
...
 }
 
 async function loadPostLCP(config) {
+  const enablePersV2 = enablePersonalizationV2();
+  if (enablePersV2 && delayedMartech) {
+    await loadMartech();
```
mokimo, line 1114: It seems personalization also loads martech (pre-LCP) are both really required or can we be smarter about when we load that?

libs/features/personalization/personalization.js
```diff
@@ -1157,8 +1167,23 @@ export async function init(enablements = {}) {
...
+    manifests = manifests.concat(localManifests);
+  }
+
+  if (target === 'postlcp') {
+    if (enablePersonalizationV2()) {
+      await callMartech(config, targetInteractionPromise);
+    } else {
+      callMartech(config);
+    }
+  }
```
mokimo, line 1186: This can be combined, untested demo code:
```suggestion
if (target === true || target === 'postlcp') {
  const callMartechPromise = enablePersonalizationV2() 
    ? callMartech(config, targetInteractionPromise) 
    : callMartech(config);

  if (target === true) {
    let localManifests = await callMartechPromise;
    manifests = manifests.concat(localManifests);
  } else if (target === 'postlcp') {
    await callMartechPromise;
  }
}
```

libs/utils/utils.js
sharmrj: #2970 does a refactoring similar to what you're talking about, though `federated.js` was added fairly recently. I was planning to redo my work on that PR around this. So for now we can do the preload and I can conditionally load `federal.js` in a future PR.

libs/utils/utils.js
```diff
@@ -1023,29 +1025,73 @@ export async function loadMartech({
...
+      perf.getEntriesByType('navigation')?.[0]?.serverTiming?.map?.(
+        ({ name, description }) => ([name, description]),
+      ) ?? [],
+    );
+  }
+
+  const isSignedOutOnStagingOrProd = serverTiming && serverTiming.sis === '0';
+
+  // Return true if it's a dev environment or signed out on staging/prod
+  return !Object.keys(serverTiming || {}).length || isSignedOutOnStagingOrProd;
```
sharmrj, line 1056: Would it be correct to refactor this to
```javascript
const entries = perf
  ?.getEntriesByType('navigation')?.[0]
  ?.serverTiming
  ?.map?.(({ name, description }) => ([name, description])) ?? [];
  return !entries.length || Object.fromEntries(entries)?.sis === '0';
```

libs/utils/utils.js
```diff
@@ -1023,29 +1025,73 @@ export async function loadMartech({
...
     || mepHighlight || mepButton || mepParam === '' || xlg)) return;
-  if (target || xlg) {
+
+  const enablePersV2 = enablePersonalizationV2();
+  if (martech !== 'off' && (target || xlg || pzn) && enablePersV2) {
+    const { locale } = getConfig();
+    const { loadAnalyticsAndInteractionData } = await import('../martech/helpers.js');
+    targetInteractionPromise = loadAnalyticsAndInteractionData(
+      { locale, env: getEnv({})?.name, timeoutMeta: getMetadata('target-timeout') },
+    );
```
mokimo, line 1092: Some untested code to illustrate what I mean:
```suggestion
    targetInteractionPromise = new Promise((res) => {
      import('../martech/helpers.js').then(({ loadAnalyticsAndInteractionData }) => {
        res(loadAnalyticsAndInteractionData(
          { locale, env: getEnv({})?.name, timeoutMeta: getMetadata('target-timeout') },
        ));
      });
    });
```

libs/utils/utils.js
swamu: Addressed

libs/features/personalization/personalization.js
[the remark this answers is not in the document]
swamu: Addressed

libs/utils/utils.js
swamu: Addressed

libs/utils/utils.js
swamu: Addressed

libs/utils/utils.js
```diff
@@ -1023,28 +1024,83 @@ export async function loadMartech({
...
+  if (martech !== 'off' && (target || xlg || pzn) && enablePersV2) {
+    const params = new URL(window.location.href).searchParams;
+    const calculatedTimeout = parseInt(params.get('target-timeout'), 10)
+  || parseInt(getMetadata('target-timeout'), 10)
+  || TARGET_TIMEOUT_MS;
+
+    const { locale } = getConfig();
+    targetInteractionPromise = new Promise((res) => {
+      import('../martech/helpers.js').then(({ loadAnalyticsAndInteractionData }) => {
+        res(loadAnalyticsAndInteractionData(
```
sharmrj, line 1086: `loadAnalyticsAndInteractionData` is an async function, which means it returns a promise, but since we're doing `resolve(loadAnalyticsAndInteractionData(...args))` the resulting type of `targetInteractionPromise` is not `Promise<T>`, it's `Promise<Promise<T>>` (Correct me if I'm wrong here.
One option would be to await `loadAnalyticsAndInteractionData` inside the `.then`, or you could do:
```javascript
const targetInteractionPromise = (async () => {
  const { loadAnalyticsAndInteractionData } = await import('../martech/helpers.js');
  const data = await loadAnalyticsAndInteractionData(...args);
  return data;
})();
```
So you get the sequencing and scheduling behvior of the `.then` snippet above with the semantics of async and await. And since we aren't `await`ing the IIFE, `targetInteractionPromise` has the correct type of `Promise<T>`.

libs/features/personalization/personalization.js
```diff
@@ -1124,6 +1128,104 @@ async function callMartech(config) {
...
+
+async function handleMartechTargetInteraction(
+  { config, targetInteractionPromise, calculatedTimeout },
+) {
+  let targetManifests = [];
+  let targetPropositions = [];
+  let responseStart;
+  if (enablePersonalizationV2() && targetInteractionPromise) {
+    try {
+      responseStart = Date.now();
```
sharmrj, line 1200: We should use `performance.mark` or `performance.now`.  `performance.now()` is a drop in replacement for `Date.now()` but `performance.mark` will generate an object containing a bit more information for you (like how long after the pageload did the measurement start).

libs/features/personalization/personalization.js
```diff
@@ -1124,6 +1128,104 @@ async function callMartech(config) {
...
+}
+
+const handleAlloyResponse = (response) => {
+  const items = (
+    (response.propositions?.length && response.propositions)
+    || (response.decisions?.length && response.decisions)
+    || []
+  ).map((i) => i.items).flat();
+
+  if (!items?.length) return [];
```
sharmrj, line 1171: This is unecessary. You can just continue the chain since `[].map(f) === (null)?.map(f) ?? [] === []`

```javascript

const handleAlloyResponse = (response) => ((response.propositions || response.decisions))
    ?.map((i) => i.items)
    ?.flat()
    ?.map(() => /*more stuff */)
    ?.filter(Boolean) ?? [];

```

sharmrj (approved):
Looks good to me. Just a couple of fairly minor things.
Also, you'll need to raise the code coverage before you can merge; essentially just write tests for the new parts of the control flow.

amitbikram (approved):

libs/utils/utils.js
```diff
@@ -1023,28 +1024,84 @@ export async function loadMartech({
...
 
   return true;
 }
 
+/**
+ * Checks if the user is signed out based on the server timing and navigation performance.
+ *
+ * @returns {boolean} True if the user is signed out, otherwise false.
+ */
+function isSignedOut() {
```
mokimo, line 1040: Isn't there another hint to detect the user signed in state? @bbythewa what was settled on, what info is coming from akamai in that regard? 🤔

libs/utils/utils.js
```diff
@@ -1023,28 +1024,84 @@ export async function loadMartech({
...
   if (mepParam === 'off') return;
   const pzn = getMepEnablement('personalization');
   const promo = getMepEnablement('manifestnames', PROMO_PARAM);
   const target = martech === 'off' ? false : getMepEnablement('target');
   const xlg = martech === 'off' ? false : getMepEnablement('xlg');
+
   if (!(pzn || target || promo || mepParam
     || mepHighlight || mepButton || mepParam === '' || xlg)) return;
+
+  const enablePersV2 = enablePersonalizationV2();
```
mokimo, line 1076: Long term we need a plan to harmonize this, I feel like there's too many different code paths and ways of personalization and martech being loaded.

mokimo (approved):
Gave it a high level review to spot where we can improve performance on this, that's looking good now and early results look promising. Will look forward to hearing more about this in the performance meeting given CRUM data!!

Make sure to sync with @bbythewa and Bhim to ensure this works for their team as well

libs/features/personalization/personalization.js
```diff
@@ -1124,6 +1128,102 @@ async function callMartech(config) {
...
+  return Math.ceil(num / 250) / 4;
+}
+
+function calculateResponseTime(responseStart) {
+  const responseTime = Date.now() - responseStart;
+  return roundToQuarter(responseTime);
+}
+
+function sendTargetResponseAnalytics(failure, responseStart, timeoutLocal, message) {
+  // temporary solution until we can decide on a better timeout value
```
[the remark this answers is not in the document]
swamu: True 😬, however I have only moved this code from Martech.js.

libs/utils/utils.js
swamu: @amitbikram FYI

mokimo: Ah no, I think that's fine. Heres the context https://jira.corp.adobe.com/browse/WPS-26192

PR #3271, opened 2024-11-27
MWPW-161645: lana logging for CCD

yesil:
with `<mas-commerce-service data-lana-tags="consumer=ccd" ...>`

lana will be enabled.

Resolves: [MWPW-161645](https://jira.corp.adobe.com/browse/MWPW-161645)

**Test URLs:**
- After: https://mwpw-161645--milo--adobecom.aem.live/?martech=off


libs/features/mas/src/lana.js
```diff
@@ -1,58 +1,60 @@
-const Defaults = {
+// Default configuration for logging system
+const config = {
     clientId: 'merch-at-scale',
     delimiter: '¶',
     ignoredProperties: ['analytics', 'literals'],
     serializableTypes: ['Array', 'Object'],
-    // Sample rate is set to 100 meaning each error will get logged in Splunk
     sampleRate: 30,
-    tags: 'consumer=milo/commerce',
```
3ch023, line 8: we could update sample rate to 1

3ch023 (changes requested):
Tested it today and the error message still say's 'Bad WCS request' without specifying the status code
image

libs/features/mas/docs/ccd.html
```diff
@@ -53,6 +53,8 @@
         const value = params.get(attribute);
         if (value) masCommerceService.setAttribute(attribute, value);
       });
+      masCommerceService.setAttribute('lana-tags', 'consumer,ccd,mas');
+      masCommerceService.dataset.sampleRate = '30';
```
3ch023, line 57: nit: let's make it 1%

libs/features/mas/src/lana.js
```diff
@@ -1,58 +1,60 @@
-const Defaults = {
+// Default configuration for logging system
+const config = {
     clientId: 'merch-at-scale',
     delimiter: '¶',
     ignoredProperties: ['analytics', 'literals'],
     serializableTypes: ['Array', 'Object'],
-    // Sample rate is set to 100 meaning each error will get logged in Splunk
     sampleRate: 30,
```
3ch023, line 7: could you update it to 1% pls? wanted to do this chance for a while now..

libs/features/mas/src/lana.js
```diff
@@ -61,26 +64,36 @@ const lanaAppender = {
...
-export { Defaults, lanaAppender };
+// Allow dynamic config updates
+function updateConfig(newConfig) {
+    Object.assign(
+        config,
+        Object.fromEntries(
+            Object.entries(newConfig).filter(
+                ([key, value]) =>
+                    key in config &&
+                    !['', NaN, null, undefined].includes(value), // 0 is a valid value.
```
Axelcureno, line 93: `NaN` values will not be filtered out here since they're not equal to any value, including itself. Instead we can use `Number.isNaN(value)` to correctly check for NaN.
```suggestion
                    function updateConfig(newConfig) {
    Object.assign(
        config,
        Object.fromEntries(
            Object.entries(newConfig).filter(
                ([key, value]) =>
                    key in config &&
                    value !== '' &&
                    value !== null &&
                    value !== undefined &&
                    !Number.isNaN(value), // Correctly exclude NaN
            ),
        ),
    );
}
```

libs/features/mas/src/lana.js
```diff
@@ -1,58 +1,61 @@
...
             return [message, status, originatingRequest]
-                .filter((v) => v)
+                .filter(Boolean)
                 .join(' ');
         }
-        const name =
-            value[Symbol.toStringTag] ??
+
+        const objectType =
+            value[Symbol.toStringTag] ||
```
Axelcureno, line 39: I think we should check against null objects also to prevent runtime errors:
```suggestion
const objectType =
    value[Symbol.toStringTag] ||
    (Object.getPrototypeOf(value)?.constructor?.name || '');
```

libs/features/mas/src/lana.js
```diff
@@ -61,26 +64,36 @@ const lanaAppender = {
...
         if (values.length) {
-            payload += `${delimiter}facts=`;
+            payload += `${config.delimiter}facts=`;
             payload += JSON.stringify(values, serializeParam);
         }
 
         if (!seenPayloads.has(payload)) {
             seenPayloads.add(payload);
-            window.lana?.log(payload, { sampleRate, tags, clientId });
+            window.lana?.log(payload, config);
```
Axelcureno, line 80: Nit: Check if NALA is not available (undefined) to prevent silent failures:
```suggestion
 if (window.lana && typeof window.lana.log === 'function') {
    window.lana.log(payload, config);
} else {
    console.warn('Lana logging is not available.');
    // Optionally, implement a fallback logging mechanism here
}
```

Axelcureno (changes requested):

3ch023 (approved):

libs/features/mas/src/lana.js
yesil: `[NaN].includes(NaN) = true` though. Still to be on the safe side, I prefer your version.

libs/features/mas/src/lana.js
yesil: actually I revert my change from `??` to `||` that was unintentional.

libs/features/mas/src/lana.js
yesil: I think it is not good idea to log `Lana logging is not available` continuously.

npeltier (approved):
lgtm

libs/features/mas/docs/ccd.html
```diff
@@ -53,6 +53,8 @@
         const value = params.get(attribute);
         if (value) masCommerceService.setAttribute(attribute, value);
       });
+      masCommerceService.dataset.lanaTags = "consumer=ccd/mas";
```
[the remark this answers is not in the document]
3ch023: not sure how this code got through, i removed it yesterday

libs/features/mas/src/price/template.js
```diff
@@ -184,7 +185,7 @@ const createPriceTemplate =
         }).forEach(([key, value]) => {
             if (value == null) {
```
npeltier, line 186:
```suggestion
            if ( !value ) {
```

libs/features/mas/src/price/template.js
```diff
@@ -184,7 +185,7 @@ const createPriceTemplate =
...
               /* c8 ignore next 2 */
-                throw new Error(`Argument "${key}" is missing`);
+                throw new Error(`Argument "${key}" is missing for osi ${offerSelectorIds?.toString()}`);
```
npeltier, line 188: country & language also as well?

afmicka:
regressions look ok

PR #3295, opened 2024-12-03
[LOC]MWPW-163380 - Missing project details locui create

saurabhsircar11:
* Added missing project details page for locui create as query params are lost after sucessfull login

Resolves: https://jira.corp.adobe.com/browse/MWPW-163380

**Test URLs:**
- Before: https://stage--milo--adobecom.hlx.page/drafts/sircar/locui-create?martech=off
- After: https://mwpw-163380-instructions-loc-create--milo--saurabhsircar11.hlx.page/drafts/sircar/locui-create?martech=off

Screenshot 2024-12-19 at 12 53 53 PM


nkthakur48 (approved):

libs/blocks/locui-create/locui-create.js
```diff
@@ -0,0 +1,15 @@
+import { createTag } from '../../utils/utils.js';
+
+export default function init(el) {
+  el.classList.add('missing-details');
+  const heading = createTag('h2', null, 'Missing project details');
+  const paragraph = createTag('p', null, 'The project details were removed after you logged in. To resolve this:');
+  const steps = createTag('ol', null);
+  const goTo = createTag('p', { class: 'goto-step' }, 'Go to ');
+  const goToAnchor = createTag('a', { href: 'https://milostudio.adobe.com', target: '_blank', rel: 'noopener noreferrer' }, 'MiloStudio');
```
narcis-radu, line 9: Should we consider lower environments as well?

libs/blocks/locui-create/locui-create.js
```diff
@@ -0,0 +1,15 @@
...
+  goTo.append(goToAnchor);
+  goTo.append('.');
+  const stepsList = ['Close this window or tab.', goTo, 'Select the specific tenant.', 'Click on Localization.', 'Click on "Add New Project".'];
+  stepsList.forEach((step) => steps.append(createTag('li', null, step)));
```
narcis-radu, line 13: How is this working for `goTo`?

libs/blocks/locui-create/locui-create.js
saurabhsircar11: Hi @narcis-radu ,
It will be to consider lower environments as there is no way to identify from which env the workflow is triggered as after successful msal login the query params get removed. So for placeholder we have kept for milostudio.adobe.com. If required we can remove the link and keep the text only.

libs/blocks/locui-create/locui-create.js
saurabhsircar11: Hi @narcis-radu ,
goto is just a paragraph having the milostudio prod anchor link.

libs/blocks/locui-create/locui-create.js
narcis-radu: If I start a workflow on milostudio.dev.adob.com and click on the anchor, I might end up on milostudio.adobe.com where the workflow is not visible. Is my understanding correct?

saurabhsircar11: Yes, your understanding is correct.

saurabhsircar11: Hi @narcis-radu , 

We have added lower environments urls also in the instructions so that the user can click on the respective url to navigate to milostudio. Please have a look and let us know if it is fine by you.

Thanks

narcis-radu (approved):

amitbikram (approved):

PR #3344, opened 2024-12-10
MWPW-159180 [MILO][MEP] MEP Manifest Manager

vgoodric:
* Refactor code in preview so that the portion needed to create the info panel can also be called by MMM
* Add manifest source to the MEP Panel
* Clearer labeling of values that can be used by placeholders
* Saves MEP info to database when user is qualified to see the MEP button
MMM block:
* Pull list of pages where MEP has been seen
* filter and search page list
* page details pulls additional db info to repopulate MEP button panel (with additional information like last seen)

Resolves: [MWPW-159180](https://jira.corp.adobe.com/browse/MWPW-159180)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/docs/authoring/features/mmm?martech=off
- After: https://mmm--milo--adobecom.aem.page/docs/authoring/features/mmm?martech=off
- Example of saving data: https://main--cc--adobecom.hlx.page/products/photoshop?milolibs=mmm (check network tab for aws)


AdobeLinhart (approved):

AdobeLinhart (approved):

denlight (approved):

libs/features/personalization/preview.js
```diff
@@ -296,10 +343,37 @@ function addHighlightData(manifests) {
...
+  data.page.target = getMetadata('target') || 'off';
+  delete data.page.highlight;
+  return fetch(API_URLS.save, {
+    method: 'POST',
+    headers: { 'Content-Type': 'application/json' },
+    body: JSON.stringify(data),
+  })
+    .then((response) => {
+      /* c8 ignore next 1 */
+      if (!response.ok) throw new Error('Network response was not ok');
```
denlight, line 370: let's do it like this 
`.then((response) => {
      /* c8 ignore next 1 */
      if (response.ok) {
        return response.json();
      } else {
        const errorMessage = response.text();
         throw new Error(errorMessage || 'Network response was not ok');
      }
    });
    `

denlight: that way we will pass the actual server text of the error

denlight (changes requested):

markpadbe (approved):
@vivgoodrich, linter is flagging the imported/declared, but unused, "isDisabled" variable in preview.js and Denys has a request which should be addressed, but otherwise, looks good to me.

ivanvatadobe (approved):

libs/features/personalization/preview.js
vgoodric: @denlight updated to below to comply with linting rules:
    .then(async (response) => {
      const res = await response.json();
      if (response.ok) return res;
      /* c8 ignore next 1 */
      throw new Error(res.message || 'Network response failed');
    });

denlight (approved):

PR #3362, opened 2024-12-11
[MWPW-164097] Allow special chars in aria-label definition

overmyheadandbody:
This adapts the logic that allows authors to define `aria-label` attribute values for links, in order to allow the use of special characters. We're now using capturing groups to identify the pattern match at the end of the link's text content, which simplifies the logic and also addresses the bug.

Resolves: [MWPW-164097](https://jira.corp.adobe.com/browse/MWPW-164097)

**Test URLs:**
- Before: https://main--milo--overmyheadandbody.aem.page/drafts/ramuntea/accessibility/aria-label-with-icon?martech=off
- After: https://aria-label-special-chars--milo--overmyheadandbody.aem.page/drafts/ramuntea/accessibility/aria-label-with-icon?martech=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -715,13 +715,11 @@ export function decorateLinks(el) {
       decorateCopyLink(a, copyEvent);
     }
     // Append aria-label
-    const pipeRegex = /\s?\|\s?/;
+    const pipeRegex = /\s?\|([^|]*)$/;
     if (pipeRegex.test(a.textContent) && !/\.[a-z]+/i.test(a.textContent)) {
-      const node = [...a.childNodes].reverse()
-        .find((child) => pipeRegex.test(child.textContent));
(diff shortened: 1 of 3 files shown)


rgclayton:
I want to make sure I understand this. Code looks great.

When I look at the main branch version of your sample page, I do see the link with a special character applying the `aria-label`. I was under the assumption this should not work until viewed with your new feature branch, with the updated code.

What am I missing?

image

mokimo (approved):

overmyheadandbody:
@rgclayton - yes, the correct `aria-label` does get applied, but the link's label also contains it, so we'd get duplicate content and show details that might affect the UI or might not be relevant for all users.

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-164097

PR #3364, opened 2024-12-12
MWPW-159180 & 164118 [MMM] MEP Manifest Manager Fast Follow

vgoodric:
Fixes for a few issues found during stage review
* page value still includes geo when sent
* page and URL values sent with double .html when seen on stage
* page is not saved if 0 manifests even if Target is on
* promo start and end times for MEP button not correct

Resolves: [MWPW-159180](https://jira.corp.adobe.com/browse/MWPW-159180)
[MWPW-164118](https://jira.corp.adobe.com/browse/MWPW-164118)

**Test URLs:**
- After: https://mmm--milo--adobecom.aem.page/?martech=off


libs/features/personalization/promo-utils.js
```diff
@@ -42,8 +42,6 @@ const getRegionalPromoManifests = (manifestNames, region, searchParams) => {
           name,
           start: GMTStringToLocalDate(start),
           end: GMTStringToLocalDate(end),
-          startUtc: start,
```
vgoodric, line 45: these are values I added because I thought we needed them, but we don't.  So removing

denlight (approved):

markpadbe (approved):

jpratt2:
I can confirm that this was fixed:
"promo start and end times for MEP button not correct" See https://jira.corp.adobe.com/browse/MWPW-164118

PR #3375, opened 2024-12-13
MWPW-164084 [Catalog] Page position not preserved after closing the modal

bozojovicic:
Catalog page scrolls to the top when some modal is closed. With `replaceState` that does not happen.

Resolves: [MWPW-164084](https://jira.corp.adobe.com/browse/MWPW-164084)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/bozo/price-cta?martech=off
- After: https://mwpw164084modal--milo--bozojovicic.aem.page/drafts/bozo/price-cta?martech=off


Axelcureno (approved):

libs/blocks/merch/merch.js
```diff
@@ -489,7 +489,7 @@ export async function openModal(e, url, offerType, hash, extraOptions) {
     const prevHash = window.location.hash.replace('#', '') === hash ? '' : window.location.hash;
     window.location.hash = hash;
     window.addEventListener('milo:modal:closed', () => {
-      window.location.hash = prevHash;
+      window.history.replaceState({}, document.title, `#${prevHash}`);
```
mirafedas, line 492: The `replaceState` does not add an entry to the browser's history stack, this means if users close the modal, and then click 'Back' the modal will not be reopened. I guess we should replace this with smth else, that would both add an entry to the browser's history stack and emit the `hashchange` event

mirafedas (changes requested):

libs/blocks/merch/merch.js
bozojovicic: In this PR I fixed what is reported. This what you described is a different issue. I can use here either 'pushState' or 'replaceState' to remove the hash from URL when modal is closed. 
With 'pushState' when you close the modal and click back you will see the has parameter in URL again but the modal will not open. Should it open? I don't know. It's not requested in this ticket. The same happens with `window.location.hash = prevHash;`
With 'replaceState' at least the hash will not be displayed in URL when you click back.
If we need to open the modal when we click back then, you are right, 'pushState' should be used, but it's not requested in this ticket.

mirafedas: @bozojovicic It is not explicitly requested in the ticket, but from what I remember, this is how it worked before: when the modal gets closed, clicking Back restores the hash, and the modal gets opened again

bozojovicic: OK @mirafedas, I didn't know that. I will ask in the ticket what to do in this case just to double check. We have enough time to fix it before the end of RCP.

bozojovicic: Finally done @mirafedas when we click back, if hash modal ID is in the URL the modal will open, if it's not there and modal is opened it will be closed. You can try it on 
https://main--cc--adobecom.hlx.page/products/catalog?milolibs=mwpw164084modal--milo--bozojovicic

libs/blocks/merch/merch.js
```diff
@@ -747,3 +747,19 @@ export default async function init(el) {
   log.warn('Failed to get context:', { el });
   return null;
 }
+
+export async function handleHashChange() {
```
mirafedas, line 751: I guess we could make this more concise (if you prefer):
```
export async function handleHashChange() {
  const modalDialog = document.querySelector('.dialog-modal');
  if (window.location.hash && !modalDialog) {
    const cta = document.querySelector(`.con-button[data-modal-id="${window.location.hash.replace('#', '')}"]`);
    reopenModal(cta);
  } else if (!window.location.hash && modalDialog) {
    const { closeModal } = await import('../modal/modal.js');
    closeModal(modalDialog);
  }
}
```

mirafedas (approved):

PR #3377, opened 2024-12-13
MWPW-163603 benchmark card rendering time in PR

npeltier:
benchmark page (only CCD for now)
250ms limit in:
/libs/features/mas/docs/benchmarks.html
adjusted to network timing in : 
/libs/features/mas/docs/benchmarks.html?adjust=true

There is still a risk that some perf degradation passes tests if they are run on slow networks, but that's all we can do atm.

**Test URLs:**
- After: https://MWPW-163603--milo--npeltier.aem.page/?martech=off


nala/features/mas/benchmark/benchmark.test.js
```diff
@@ -0,0 +1,50 @@
...
+  if (browserName === 'chromium') {
+    await page.setExtraHTTPHeaders({ 'sec-ch-ua': '"Chromium";v="123", "Not:A-Brand";v="8"' });
+  }
+  // disabling cache
+  await page.route('**', (route) => route.continue());
+});
+
+test.describe('Benchmark feature test suite', () => {
+  test(`${features[0].name},${features[0].tags}`, async ({ page, baseURL }) => {
+    const testPage = `${baseURL}${features[0].path}${miloLibs}`;
```
3ch023, line 18: I do have a concern including this in Milo nala tests.
If freya or wcs cdn will be causing slower render it will fail Milo nala tests and block developers until someone will comment this test out.
We could try it out and if we trues in Freya/WCS stability it would work, however I think we should notify people so that they are aware about this test having external dependencies (it's ok for Nala tests)
And if this test will be flaky we will have to take it out and maybe run it only manually/before release.
We should make sure we don't create stress for Milo community with this

libs/features/mas/docs/common.js
```diff
@@ -0,0 +1,40 @@
+import '../../../deps/custom-elements.js';
```
3ch023, line 1: maybe its better to place it in `src` folder
and keep docs folder for html build artifacts

libs/features/mas/docs/common.js
```diff
@@ -0,0 +1,40 @@
...
+  document.head.appendChild(theme);
+  // mas-commerce-service
+  const masCommerceService = document.createElement('mas-commerce-service');
+  ['locale','language','env'].forEach((attribute) => {
+    const value = params.get(attribute);
+    if (value) masCommerceService.setAttribute(attribute, value);
+  });
+  masCommerceService.setAttribute('host-env', 'prod');
+  masCommerceService.setAttribute('lana-tags', 'ccd');
+  masCommerceService.setAttribute('lana-sample-rate', '100'); // optional, default value is set to '1'
```
3ch023, line 37: including this on every page would pollute our prod logs
ccd page is not throwing error, but https://mwpw-163603--milo--npeltier.aem.page/libs/features/mas/docs/merch-card.html does

not sure its a big deal, but something worth mentioning

nala/features/mas/benchmark/benchmark.test.js
npeltier: I agree, this is why i added alignement with network capabilities. For the rest items should be cached, so this is CDN response time for constant time we are measuring and this should not change much.
I thought about not adding this systematically and/or raise acceptance bar, but i think this will defeat the purpose of us guaranteeing cards performances

libs/features/mas/docs/common.js
npeltier: (even if it's 9 views a week)

libs/features/mas/docs/common.js
```diff
@@ -0,0 +1,40 @@
+import '../../../deps/custom-elements.js';
+import '../../../features/spectrum-web-components/dist/theme.js';
+import '../../spectrum-web-components/dist/action-button.js';
+import '../../../features/spectrum-web-components/dist/button.js';
+const init = () => {
+  const ENVS = {
```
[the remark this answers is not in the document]
yesil: Oh, I also something similar but not common but will replace with this once before merge.

libs/features/mas/docs/benchmarks.html
```diff
@@ -0,0 +1,130 @@
...
+    <!-- Include any additional stylesheets -->
+    <link rel="stylesheet" href="styles.css">
+    <script type="module">
+      import { init } from './common.js';
+      init();
+    </script>
+</head>
+
+<script>
+  const measureCardPerformances = async (container, fragmentId) => {
```
yesil, line 30: we should probably incorporate performance api to `merch-card` and `aem-fragment` custom elements directly.

and use performance api entries while reporting the measurements.

npeltier:
i'll incorporate @yesil feedbacks + will try to post results to some place

3ch023 (approved):

libs/features/mas/docs/ccd.html
```diff
@@ -51,7 +51,20 @@
     </script>
 </head>
 
+<<<<<<< HEAD
```
3ch023, line 54: merge issue?
https://mwpw-163603--milo--npeltier.aem.page/libs/features/mas/docs/ccd.html is broken a bit

3ch023 (changes requested):
needs a small fix for ccd.html

3ch023 (changes requested):
could you try running npm install in /libs/features/mas?
i wonder if it will remove changes in merch-stock.js etc

libs/features/mas/docs/ccd.html
```diff
@@ -50,9 +50,17 @@
       document.head.appendChild(masCommerceService);
     </script>
 </head>
-
-<body class="spectrum spectrum--medium spectrum--light">
-  <main>
+  <body class="spectrum spectrum--medium spectrum--light">
+    <div class="sidenav">
```
yesil, line 54: nit: you could use `aside` for element

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside

yesil (approved):
LGTM 👍

libs/features/mas/docs/ccd.html
npeltier: sure :)

3ch023 (approved):

PR #3201, opened 2024-11-14
[MWPW-161870] Fixing the issue that sidenav can't be collapsed.

seanchoi-dev:
Resolves:
- [MWPW-161870](https://jira.corp.adobe.com/browse/MWPW-161870)
- [MWPW-162024](https://jira.corp.adobe.com/browse/MWPW-162024)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/?martech=off
- After: https://MWPW-161870--milo--adobecom.aem.page/?martech=off

**Test for consumer site:**
- Before: https://main--cc--adobecom.hlx.page/products/catalog
- After: https://main--cc--adobecom.hlx.page/products/catalog?milolibs=MWPW-161870#category=creativity-design

PR #3389, opened 2024-12-17
Setting aria-live to shadowRoot element instead to the slot.

seanchoi-dev:
This is a followup PR from https://github.com/adobecom/milo/pull/3201

Resolves:
- [MWPW-162024](https://jira.corp.adobe.com/browse/MWPW-162024)

**Test URLs:**
- After: https://MWPW-162024--milo--adobecom.aem.page/?martech=off

**Test for consumer site:**
- Before: https://main--cc--adobecom.hlx.page/products/catalog
- After: https://main--cc--adobecom.hlx.page/products/catalog?milolibs=MWPW-162024


vhargrave (approved):
I wonder if you could not add the aria-live attribute while directly adding that text field to your component, but doesn't matter too much
looks good!

libs/blocks/merch-card-collection/merch-card-collection.js
```diff
@@ -338,5 +335,7 @@ export default async function init(el) {
     'merch-card-collection-render',
     'merch-card-collection-render:start',
   );
+
+  merchCardCollection.shadowRoot.getElementById('resultText')?.setAttribute('aria-live', 'polite');
```
yesil, line 339: @vhargrave is right, we should not update `shadowRoot` from outside.
we should add it at libs/features/mas/src/merch-card-collection.js

yesil (changes requested):

Roycethan (approved):

Axelcureno (approved):

PR #3395, opened 2024-12-18
DA Floodgate Copy UI

colloyd:
* UI for the copy option

Resolves: [MWPW-161908](https://jira.corp.adobe.com/browse/MWPW-161908)

**Test URLs:**
- Before: https://da.live/app/adobecom/milo/tools/floodgate?ref=floodbox
- After: https://da.live/app/colloyd/milo/tools/floodgate?ref=da-floodgate-copy-ui

### DIFF
--- a/tools/floodbox/floodgate/floodgate.js
+++ b/tools/floodbox/floodgate/floodgate.js
@@ -579,6 +579,7 @@ export default class MiloFloodgate extends LitElement {
     return html`
       <div class="tab-step active" data-id="find">
         <h3>Find Files, Referenced Fragments and Assets</h3>
+        ${this._canCopyPaths ? html`<p>Press the "Copy" button above to copy these files.</p>` : nothing}
         <div class="detail-cards find-cards">
           ${floodbox.renderBadge('Pages', this._filesToCopy.length)}
           ${floodbox.renderBadge('Fragments', this._fragmentsAssets.size)}
@@ -628,7 +629,7 @@ export default class MiloFloodgate extends LitElement {
       <form>    
         ${this._selectedOption === 'fgCopy' ? html`
           <div class="input-row">
-            <textarea name="copyPaths" rows="10" placeholder="Enter paths to copy to the pink site, separated by line-break." @input=${this.validateCopyPaths}></textarea>
+            <textarea name="copyPaths" .disabled=${this._findingPaths} rows="10" placeholder="Enter paths to copy to the pink site, separated by line-break." @input=${this.validateCopyPaths} ></textarea>
             ${floodbox.renderClearButton()}
           </div>
         ` : nothing}
--- a/tools/floodbox/floodgate/floodgate.js
+++ b/tools/floodbox/floodgate/floodgate.js
@@ -44,10 +44,13 @@ export default class MiloFloodgate extends LitElement {
     this._startPreview = false;
     this._startPublish = false;
     this._filesToPromote = [];
+    this._validCopyPaths = {};
+    this._findingPaths = false;
+    this._copyReady = false;
     this._filesToCopy = [];
     this._copiedFiles = [];
     this._copiedFilesCount = 0;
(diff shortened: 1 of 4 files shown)


arshadparwaiz (approved):

sukamat (commented):
@colloyd Few observations:
1. There might be a need to provide some sort of message that the user has to click 'Copy' button after the 'Find' step is completed. It will not be intuitive for a new user that the button context has changed and they have to click the button again. Alternatively, a 'Continue' or 'Start Copy' button in the 'Find' tab-step view might help as well. 

2. Making the text area uneditable after the process starts will be better as the list is technically locked in place unless they refresh the page. As the process pauses after 'Find' step and they have to click 'Copy' again, having a editable text-area gives the notion that it can be further updated.

3. In the 'Find' tab-step, it shows 'Pages' and 'Fragments' card. It will be good to rename 'Fragments' to 'Fragments and Assets' to avoid any confusion as the list/count includes both.

Thoughts?

tools/floodbox/floodgate/floodgate.js
```diff
@@ -507,36 +539,52 @@ export default class MiloFloodgate extends LitElement {
...
   renderReferencedFragmentsAssetsInfo() {
     return html`
-      <div class="fragments-assets-info info-box">
-        <h3>Find Referenced Fragments and Assets</h3>
-        <p>Found ${this._fragmentsAssets.size} referenced fragments and assets.</p>
+      <div class="tab-step active" data-id="find">
+        <h3>Find Files, Referenced Fragments and Assets</h3>
+        <div class="detail-cards find-cards">
+          ${floodbox.renderBadge('Pages', this._filesToCopy.length)}
+          ${floodbox.renderBadge('Fragments', this._fragmentsAssets.size)}
```
sukamat, line 584:
```suggestion
          ${floodbox.renderBadge('Fragments & Assets', this._fragmentsAssets.size)}
```

colloyd: the & would be problematic as the name is converted as part of a classname elsewhere. My suggestion, given the details I mentioned in my slack to you would be to have a separate card for fragments and one for assets we would then need to setup the url object to include whether it was a fragment or an asset in addition to the other details I mentioned.

colloyd:
@sukamat, we can certainly move the button for copy into the tab. I thought I'd give a stab at trying to cut down on the number of buttons on screen at once. I added the "Finding..." text to the button while it's finding fragments, so there's some precedent for the text changing. I figured we'd see what users thought of this and if it's not clear, we would refine from there.

colloyd:
@sukamat Check it now. I disabled the text field when the user starts and I added messaging below the heading inside of the Find tab to indicate that they should press Copy to continue (after the find process is complete).

sukamat (changes requested):
@colloyd 
> the & would be problematic as the name is converted as part of a classname elsewhere. My suggestion, given the details I mentioned in my slack to you would be to have a separate card for fragments and one for assets we would then need to setup the url object to include whether it was a fragment or an asset in addition to the other details I mentioned.

The classname conversion code in `renderBadge()` can be updated to exclude `&` with a simple regex change, right?  Currently it seems to only handle names with space. Is there more to it than the regex change?

From what I see, separating fragments and assets will be lot more work without any specific benefits.

colloyd:
> @colloyd
> 
> > the & would be problematic as the name is converted as part of a classname elsewhere. My suggestion, given the details I mentioned in my slack to you would be to have a separate card for fragments and one for assets we would then need to setup the url object to include whether it was a fragment or an asset in addition to the other details I mentioned.
> 
> The classname conversion code in `renderBadge()` can be updated to exclude `&` with a simple regex change, right? Currently it seems to only handle names with space. Is there more to it than the regex change?
> 
> From what I see, separating fragments and assets in the response object and then handling them separately will be lot more work without any specific benefits.

Done.

sukamat (approved):

PR #3406, opened 2024-12-24
fix for georouting popup link is seen rendered out of the popup in mobile portrait view in arabic locale

patel-prince:
georouting popup link is seen rendered out of the popup in mobile portrait view in arabic locale

Resolves: [MWPW-164340](https://jira.corp.adobe.com/browse/MWPW-164340)

**Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/mena_ar/products/photoshop?martech=off
- After: https://main--cc--adobecom.hlx.page/mena_ar/products/photoshop?milolibs=mwpw-164340-georouting--milo--patel-prince&martech=off

### DIFF
--- a/libs/features/georoutingv2/georoutingv2.css
+++ b/libs/features/georoutingv2/georoutingv2.css
@@ -49,7 +49,7 @@
 }
 
 .dialog-modal.locale-modal-v2 a:last-child {
-  display: inline-block;
+  white-space: no-wrap;
 }
 
 .dialog-modal.locale-modal-v2 .picker li {
--- a/libs/features/georoutingv2/georoutingv2.css
+++ b/libs/features/georoutingv2/georoutingv2.css
@@ -49,7 +49,7 @@
 }
 
 .dialog-modal.locale-modal-v2 a:last-child {
-  white-space: nowrap;
+  display: inline-block;
 }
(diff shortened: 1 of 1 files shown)


libs/features/georoutingv2/georoutingv2.css
```diff
@@ -49,7 +49,7 @@
 }
 
 .dialog-modal.locale-modal-v2 a:last-child {
-  white-space: nowrap;
```
overmyheadandbody, line 52: @vhargrave may have more insight on this, but I'd assume that this declaration was intentional. I had a look on `mena_ar` and the suggested change would make it look like:


Although without this change it looks like:


I'd also question the device used to surface this issue. I had to go down to `280px` width to replicate the bug, although it's not a common occurrence according to [this](https://gs.statcounter.com/screen-resolution-stats/mobile/worldwide).

patel-prince: the bug has been raised based on this url, https://www.adobe.com/mena_ar/products/photoshop.html
image-2024-12-17-13-28-30-031
as per the bug this issue is happening below 375px screen size.

vhargrave (approved):

PR #3407, opened 2024-12-24
[MWPW-161923] tab nested focus

DKos95:
This resolves the issue where the buttons were nested in a div and not selectable with keyboard and screenreader, which is needed for accessability.

Resolves: [MWPW-161923](https://jira.corp.adobe.com/browse/MWPW-161923)

| Before     | After      |
|------------|------------|
| Screenshot 2024-12-24 at 13 09 00|  Screenshot 2024-12-24 at 13 07 49 |
| Screenshot 2024-12-24 at 13 16 01| Screenshot 2024-12-24 at 13 16 21 |

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/docs/library/kitchen-sink/tab-block
- After: https://tab-nested-focus--milo--adobecom.hlx.page/docs/library/kitchen-sink/tab-block

### DIFF
--- a/libs/blocks/tabs/tabs.css
+++ b/libs/blocks/tabs/tabs.css
@@ -186,7 +186,7 @@
   cursor: pointer;
   width: 32px;
   height: 48px;
-  top: -4.5px;
+  top: 0;
   background: var(--tabs-active-bg-color);
   border: 0;
   border-bottom: 1px solid var(--tabs-border-color);
--- a/libs/blocks/tabs/tabs.css
+++ b/libs/blocks/tabs/tabs.css
@@ -49,20 +49,40 @@
   padding: var(--spacing-xxl) 0;
 }
 
+.tabs.xxl-spacing .paddle {
+  top: var(--spacing-xxl);
+}
+
 .tabs.xl-spacing {
   padding: var(--spacing-xl) 0;
+}
+
+.tabs.xl-spacing .paddle {
+  top: var(--spacing-xl);
 }
 
 .tabs.l-spacing {
   padding: var(--spacing-l) 0;
 }
 
+.tabs.l-spacing .paddle {
+  top: var(--spacing-l);
+}
+
 .tabs.s-spacing {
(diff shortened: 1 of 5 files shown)


libs/blocks/tabs/tabs.css
```diff
@@ -100,7 +120,7 @@
   display: flex;
   width: var(--grid-container-width);
   margin: 0 auto;
-  box-sizing: content-box;
+    box-sizing: content-box;
```
narcis-radu, line 123: the extra space can be removed

libs/blocks/tabs/tabs.css
```diff
@@ -303,28 +323,31 @@
   height: 24px;
 }
 
-.tabs .tab-paddles {
-  position: relative;
+.tabs .paddle-left {
```
narcis-radu, line 326: I think we can do since both `paddle-left` and `paddle-right` have this class.
```css
.tabs .paddle {
  z-index: 3;
}
```

libs/blocks/tabs/tabs.css
```diff
@@ -358,7 +381,11 @@
   border: none;
 }
 
-.tabs.radio .tab-paddles {
+.tabs.radio .paddle-left {
```
narcis-radu, line 384: can't we do `.tabs.radio .paddle { display: none }` here? It would target both `paddle-left` and `paddle-right` elements

overmyheadandbody (commented):
Narcis' comments are all valid, I have another note as well

overmyheadandbody (commented):
Looking good now, but our a11y partners are hinting that we might have additional issues to solve. I've requested details and will follow up.

overmyheadandbody (approved):

mokimo (approved):

robert-bogos (approved):

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-161923

PR #3408, opened 2024-12-24
MWPW-161354: MAS: export variant metadata

yesil:
to allow MAS Studio to toggle the fields
in the editor.

Resolves: [MWPW-161354](https://jira.corp.adobe.com/browse/MWPW-161354)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.live/libs/features/mas/docs/mas.html
- After: https://mwpw-161354--milo--adobecom.aem.live/libs/features/mas/docs/mas.html


libs/features/mas/docs/styles.css
```diff
@@ -71,11 +71,14 @@ span.placeholder-failed::after {
 }
 
 merch-card {
+  display: flex;
```
3ch023, line 74: imo we shouldn't have styles from merch-card in ccd.html (it's ok to have it on rest of pages)
reason is we have less change to catch an issue since ccd.html deviates from ccd usage by introducing additional styles. Example, in december, we had a space bug between prices that was only discovered on CCD build because in gallery it was 'fixed' by /docs/styles.css

I think on all gallery pages we shouldn't include <link rel="stylesheet" href="styles.css"> at all.

3ch023 (changes requested):
looks good, just a small comment

libs/features/mas/docs/styles.css
yesil: Good point @3ch023, let me remember why I have done this and find an alternative.

3ch023 (approved):

afmicka:
@yesil commenting here but it may be related to mas PR instead of this. 
Checking https://mwpw-161354--mas--adobecom.hlx.live/studio.html?milolibs=mwpw-161354--milo--adobecom#path=drafts

The 'size' field for suggested card  - either it should be removed or it needs to have a default value instead of being empty. Earlier it had a text to enter the size but now is changed to drop down. Screenshots for comparison
Screenshot 2025-01-21 at 14 25 30
Screenshot 2025-01-21 at 14 27 07

afmicka:
@yesil the id attribute is lost from the merch-card element. Do we want to lose it? Screenshots for comparison
Screenshot 2025-01-22 at 15 53 02
Screenshot 2025-01-22 at 15 53 09

afmicka:
@yesil on your branch, one CTA disappears from the card. 
https://mwpw-161354--milo--adobecom.aem.live/docs/library/kitchen-sink/merch-card?martech=off
Screenshot 2025-01-22 at 16 02 36
Screenshot 2025-01-22 at 16 02 46

afmicka:
@yesil disregard the previous message, was comparing .page with .live ... the change is not published,

yesil:
@afmicka yes, the following line was adding the ID it is dangerous as it could override the ID set by the consumer.


https://github.com/adobecom/milo/pull/3408/files#diff-c0e31e177931011ae39cb9321fdddb5cc8356f5df49a6ecf52e2b499c7564cdbL276

PR #3419, opened 2025-01-06
MWPW-163916 [Mobile GNAV] | Flickering issue while scrolling when LNAV is opened

sharmrj:
* on iOS Safari `overflow: hidden` does not prevent the screen from scrolling. Scrolling can still happen under the following circumstances:
  * when the bottom toolbar is collapsed
  * When the keyboard is visible
  * When focusing on an input (which centers it it)
* So we do `position:fixed` to prevetn scrolling. But this jumps the scroll to the top of the page. So we offset the old scroll height using `top`. 
* Once the background becomes scrollable again, we retrieve our original scroll value from `body`'s style and set the window scroll to that. 

Resolves: [MWPW-163916](https://jira.corp.adobe.com/browse/MWPW-163916)

**Test URLs:**
- After: https://mobile-gnav-ios-scroll-flicker--milo--adobecom.aem.page/?martech=off
https://main--federal--adobecom.hlx.page/federal/dev/prats/mobileredesign/page?milolibs=mobile-gnav-ios-scroll-flicker

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.css
+++ b/libs/blocks/global-navigation/global-navigation.css
@@ -1182,10 +1182,8 @@ header.new-nav .feds-breadcrumbs li:first-child:not(:nth-last-child(-n+3)):after
   width: 100%;
   height: calc(101lvh - (var(--feds-height-nav) + var(--feds-localnav-height)));
   padding-bottom: env(safe-area-inset-bottom);
-  position: absolute;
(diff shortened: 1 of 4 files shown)


bandana147:
I think you would have to reset the styles when the viewport is changed. Please test it.

sharmrj:
> I think you would have to reset the styles when the viewport is changed. Please test it.

Yeah I've done that

PR #3421, opened 2025-01-07
Adding support for no megamenu localnav

bandana147:
If a gnav has no megamenu but some links are present then they are now treated as localnav. So, with this change
1. Hamburger will no longer appear for such navs
2. Local nav will appear

Resolves: [MWPW-164314](https://jira.corp.adobe.com/browse/MWPW-164314)

**Test URLs:**
- After: https://mgnav-no-megamenu--milo--adobecom.aem.page/?martech=off
Qa - https://main--dc--adobecom.hlx.page/acrobat/hub/10-benefits-of-digitizing-documents?milolibs=mgnav-no-megamenu&newNav=true

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.css
+++ b/libs/blocks/global-navigation/global-navigation.css
@@ -89,7 +89,6 @@ header.global-navigation {
 
 /* Hamburger toggle */
 .feds-toggle {
-  width: 60px;
   margin: 0;
   padding: 0;
   border: none;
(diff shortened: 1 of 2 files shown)


prativas22:
Verified if a gnav has no megamenu but some links are present, then Hamburger will not display for such nav and LNAV will appear
https://main--dc--adobecom.hlx.page/acrobat/hub/how-to-print-excel-on-one-page?milolibs=mgnav-no-megamenu&newNav=true
image
Mobile view:
without sign in
image
image

with sign in
image
image

PR #3424, opened 2025-01-07
[MWPW-164132] Implementing Branch Quick Links with ECID Integration

drashti1712:
* Implements Branch Quick Links with ECID for testing in production with and without a loader to collect data and understand whether it provides a good user experience.
* The loader can be turned on/off by adding metadata to the page. 
* Github discussion link: https://github.com/orgs/adobecom/discussions/3032

Resolves: [MWPW-164132](https://jira.corp.adobe.com/browse/MWPW-164132) and [MWPW-152850](https://jira.corp.adobe.com/browse/MWPW-152850)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.live/drafts/drashti/milo-banner/ql-progress-bar-loader?martech=off
- After: https://ecid-test--milo--drashti1712.aem.live/drafts/drashti/milo-banner/ql-progress-bar-loader?martech=off

Different scenarios:
1. Quick Link with ECID passed, with no loader and a delay of around 3s:
 https://ecid-test--milo--drashti1712.hlx.live/drafts/drashti/milo-banner/ql-noloader
2. Quick Link with ECID passed after privacy consent check, with a progress bar loader: 
 https://ecid-test--milo--drashti1712.hlx.live/drafts/drashti/milo-banner/ql-progress-bar-loader


libs/utils/utils.js
```diff
@@ -725,6 +772,42 @@ export function decorateLinks(el) {
...
+        if (loaderCheck === 'progress-circle') addCircleLoader(a);
+        else if (loaderCheck === 'progress-bar') pb = addBarLoader(a);
+        const hasConsent = await waitForConsent;
+        if (hasConsent) {
+          if (loaderCheck === 'progress-bar') removeBarLoader(pb, a);
+          else if (loaderCheck === 'progress-circle') removeCircleLoader(a);
+        }
+        decorateQuickLink(a, hasConsent);
+      });
+    }
```
mokimo, line 810: Is all of this new code required for every consumer on every page? If not I'd not place this code in here. There are a ton of opt-in features and that's good, otherwise these core utils would be 10000's of lines of code long and our performance would have long dropped by now.

**From a performance standpoint:**
We want to parallelize loading everything that's needed for LCP and _only_ what's needed for LCP.

**From a maintenance perspective:**
We want to put code close to where it's used. Is this a core feature that needs to be checked on every single link that's being authored, on every single page, for every single milo consumer? If yes it really belongs here. **If no, we need to find a better home outside of utils for this.**

libs/utils/utils.js
```diff
@@ -686,6 +686,53 @@ export function convertStageLinks({ anchors, config, hostname, href }) {
...
+  return container;
+}
+
+function removeBarLoader(elem, a) {
+  elem.replaceWith(a);
+}
+
+async function decorateQuickLink(a, hasConsent) {
+  if (!window.alloy) return;
+  const { getECID } = await import('../blocks/mobile-app-banner/mobile-app-banner.js');
```
mokimo, line 726: Importing the `mobile-app-banner` to get the ECID is an anti-pattern. You create a dependency on an encapsulated method, which also would lead to things such as the lana logs being wrong, which would hint the `mobile-app-banner` itself had an error, while it occurred in the utils.
```.catch((err) => window.lana.log(`Error fetching ECID: ${err}`, { tags: 'mobile-app-banner' }));```

Utils is consumed by everyone and thus changes should be done more carefully, blocks should just know about their own existence making refactoring isolated to testing the block alone and that's enough.

**Review by mokimo — CHANGES_REQUESTED** [review]
> "Implements Branch Quick Links with ECID for testing in production with and without a loader to collect data and understand whether it provides a good user experience."

"Testing in production" sounds like a MEP use-case, I only had a rather quick glance though and didn't gather the whole context. However utils does not seem to be the right place for this feature.

You'll need to find a better place for this code to live in and restructure some of the code.

File: libs/utils/utils.js (hunk 2 above)
Comment (drashti1712) [reply libs/utils/utils.js:726]: Would it be a better solution to add the `getECID` function in utils and then import it in mobile app banner?

File: libs/utils/utils.js (hunk 1 above)
Comment (drashti1712) [reply libs/utils/utils.js:812]: Maybe another way to achieve this in utils without adding so much of code would be to add all the processing in a separate file and immediately invoke the function if we need to process the link. Something like:

``` const branchQuickLink = 'app.link';
    const ecidCheck = getMetadata('quick-link-ecid');
 
    if (a.href.includes(branchQuickLink) && ecidCheck === 'on') {
      (async () => {
        const { default: processQL } = await import('../features/branch-quick-links/branch-quick-links.js');
        processQL(a);
      })();
    }
```
   
We are using an IIFE here as we cannot await the processQL function (`decorateLinks` is not async). Let me know if this is a better way.

cc: @mokimo

mokimo: You could also test making this an auto block, which would be properly awaited for. If PMs are fine with potentially losing a bit of accuracy doing the processing "at some point" without blocking the LCP is 100% preferred.

libs/utils/utils.js
```diff
@@ -725,6 +774,42 @@ export function decorateLinks(el) {
...
+        let pb;
+        if (loaderCheck === 'progress-circle') addCircleLoader(a);
+        else if (loaderCheck === 'progress-bar') pb = addBarLoader(a);
+        const hasConsent = await waitForConsent;
+        if (hasConsent) {
+          if (loaderCheck === 'progress-bar') removeBarLoader(pb, a);
+          else if (loaderCheck === 'progress-circle') removeCircleLoader(a);
+        }
+        decorateQuickLink(a, hasConsent);
+      });
```
JasonHowellSlavin, line 811: This is a bit hard to read without the standard brackets that we use for `if` `else if` etc. I think you may also be able to flatten with early returns.

test/utils/utils.test.js
```diff
@@ -212,6 +212,29 @@ describe('Utils', () => {
...
+      });
+      it('should add Progress Circle Loader', async () => {
+        const localHead = await readFile({ path: './mocks/head-quick-links.html' });
+        document.head.innerHTML = localHead;
+        const qL = document.querySelector('a[href*="app.link"]');
+        await utils.loadArea();
+        window.dispatchEvent(new CustomEvent('adobePrivacy:PrivacyConsent'));
+        await qL.click();
+        document.head.innerHTML = head;
+      });
```
JasonHowellSlavin, line 235: I don't see expect statements for these tests. Please add a few, maybe pre click and post click, as well as checks for the element?

libs/utils/utils.js
mokimo: @drashti1712 did you try it as an autoblock, why/why wouldn't that work?

libs/utils/utils.js
mokimo: Hmm 😬 That's a good question, we started guarding adding to utils quite a lot.

If it's used in 3-4+ blocks yes. If it's used twice so far - it might be banner to have the connection to branch for now. We should ensure to test performance though, so we don't add any render-blocking long import chains.

libs/utils/utils.js
drashti1712: @mokimo Could you please help me understand if the autoblock would be immediately invoked like the IIFE, or will it take some time to load? If yes, how does an autoblock help our requirements (compared to an iife)? Sorry for these questions, I am just trying to learn more about this 🥹

Adding more details w.r.t processing - 
This task requires us to immediately process the branch quick link on finding one. I do not think that doing the processing 'at some point' would be possible here because we need to append the ECID to the link as soon as possible to enable analytics tracking. Any delay in processing would lead to losing data.
The main purpose here is to test different approaches - with and without loaders, and get accurate data on prod pages so that we can proceed with the approach with the best numbers.

mokimo: I'd personally suggest to keeping the `IIFE` and try to come up with a "smarter" solution, if the signal/accuracy isn't strong enough.

drashti1712:
Hi @mokimo, @JasonHowellSlavin 
Thank you for your help in reviewing this PR. I have considered all the reviews and restructured the code as per the comments. Here are some of the changes I have pushed in the past 2-3 days:

- Moved the most code to a separate file `branch-quick-links.js` and used an IIFE to process the Quick Links in `utils.js`. I also got this approach reviewed by the team. The Autoblock approach was not preferred because of the performance hit it would cause.
- Removed the circle loader code as it was not necessary as per the requirement.
- For handling the delay in loading the `branch-quick-links.css` file, we are adding some css while creating the loader in `addLoader()` function. This helps to prevent any glitches/delay between button click and css load.
- Added a fallbackTimeout to resolve the Promise in case none of the privacy events are triggered.
- Made all the changes suggested by @JasonHowellSlavin as well.
- Modified the Unit Tests as per the above changes.

Please let me know if this looks good.

cc: @salonijain3 @aishwaryamathuria @suhjainadobe @amitbikram

libs/features/branch-quick-links/branch-quick-links.js
```diff
@@ -0,0 +1,81 @@
...
+        clearTimeout(fallbackTimeout);
+        resolve(getConsentStatus());
+      });
+    });
+  }
+
+  a.addEventListener('click', async (e) => {
+    e.preventDefault();
+    let pb;
+    if (isLoader === 'on') pb = addLoader(a, loaderText);
```
mokimo, line 76: What does `pb` stand for, is that the loader? Should it rather be named loader?

mokimo, line 76:
```suggestion
    if (getMetadata('quick-link-loader') === 'on') pb = addLoader(a);
```

libs/features/branch-quick-links/branch-quick-links.js
```diff
@@ -0,0 +1,81 @@
...
+  if (hasConsent && !a.href.includes('ecid')) {
+    a.href = a.href.concat(`?ecid=${ecid}`);
+  }
+  window.location.href = a.href;
+}
+
+export default function processQL(a) {
+  a.classList.add('quick-link');
+  const isLoader = getMetadata('quick-link-loader');
+  const loaderText = getMetadata('quick-link-loader-text');
```
mokimo, line 44: IMO adding those variables here just adds noise and more code without a good reason, you don't re-use them.

libs/features/branch-quick-links/branch-quick-links.js
```diff
@@ -0,0 +1,81 @@
+import { getMetadata, loadStyle, getConfig, createTag } from '../../utils/utils.js';
+
+function addLoader(elem, text) {
```
mokimo, line 3: You pass an `a` (anchorTag?) into this function, but then it turns into an `elem` ... I'd stick to naming this `a` as well or something more descriptive to what you actually pass in, unless it its _any_ element.

libs/features/branch-quick-links/branch-quick-links.js
```diff
@@ -0,0 +1,81 @@
...
+  const { base } = getConfig();
+  loadStyle(`${base}/features/branch-quick-links/branch-quick-links.css`);
+  const lbl = createTag('div', { class: 'pbar-label' });
```
mokimo, line 6: just naming this `label` would help with comprehension compared to `lbl` u save 2 characters, but its not immediately obvious further down in the function on what `lbl` is.

libs/features/branch-quick-links/branch-quick-links.js
```diff
@@ -0,0 +1,81 @@
...
+  const tr = createTag('div', { class: 'pbar-value' });
+  tr.style.display = 'block';
+  const pb = createTag('div', { class: 'pbar' }, tr);
+  pb.style.width = '100%';
+  pb.style.height = '10px';
+  pb.style.borderRadius = '5px';
+  pb.style.backgroundColor = '#D5D5D5';
+  pb.style.marginTop = '10px';
+  pb.style.overflow = 'hidden';
+  const container = createTag('div', { class: 'pbar-container' }, [lbl, pb]);
```
mokimo, line 17: When you just read this line, `[lbl, pb]` is getting quite confusing for someone who reads this for the first time and didn't look at a lot more within the function.

test/features/branch-quick-links/branch-quick-links.test.js
```diff
@@ -0,0 +1,41 @@
...
+
+  afterEach(() => {
+    sinon.restore();
+  });
+
+  it('should add a class to the quick link', async () => {
+    window.alloy = new Promise(() => {});
+    const alloyS = sinon.stub(window, 'alloy');
+    alloyS.resolves({ identity: { ECID: '123' } });
+    const qL = document.querySelector('a');
```
mokimo, line 24: suggestion `quickLink` vs qL

libs/features/branch-quick-links/branch-quick-links.js
```diff
@@ -0,0 +1,81 @@
...
+  } catch (e) {
+    window.lana.log(`Error fetching ECID: ${e}`, { tags: 'branch-quick-links' });
+  }
+  if (hasConsent && !a.href.includes('ecid')) {
+    a.href = a.href.concat(`?ecid=${ecid}`);
+  }
+  window.location.href = a.href;
+}
+
+export default function processQL(a) {
```
mokimo, line 41: `processQuickLink` in favor of `processQL`

libs/features/branch-quick-links/branch-quick-links.js
```diff
@@ -0,0 +1,81 @@
+import { getMetadata, loadStyle, getConfig, createTag } from '../../utils/utils.js';
+
+function addLoader(elem, text) {
+  const { base } = getConfig();
+  loadStyle(`${base}/features/branch-quick-links/branch-quick-links.css`);
+  const lbl = createTag('div', { class: 'pbar-label' });
+  lbl.textContent = text;
```
mokimo, line 7: I would move the metadata closer to where it's used
```suggestion
  lbl.textContent = getMetadata('quick-link-loader-text') || '';
```

mokimo (commented):
A bit of naming and code structure nitpicking. Generally (IMO) every line of code should be understood by being able to read it in isolation.

Let's compare two fictive examples to illustrate the effect this has:
1: `const container = createTag('div', { class: 'quicklink-container' }, [loaderLabel, loader]);`
2: `const c = createT('div', {class: 'ql-c'} , {lbl, pb}` 

You would instantly be able to tell "Ah, okay I pass in a label and a loader into a container for the quicklinks" ... vs the 2nd version, it's very hard to understand without further context.

libs/features/branch-quick-links/branch-quick-links.js
drashti1712: It stands for Progress Bar, I have renamed it now

libs/features/branch-quick-links/branch-quick-links.js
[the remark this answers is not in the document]
drashti1712: Yes, the styles for the basic outline of the loader are added as a part of the js because of the possibility of a delay in the loading of the branch-quick-links css file. This helps to prevent any glitches/delay between button click and css load by showing the basic loader (in case of delay) for few microseconds before the animation starts. This was discussed and reviewed by the team.

libs/features/branch-quick-links/branch-quick-links.js
```diff
@@ -0,0 +1,76 @@
...
+        clearTimeout(fallbackTimeout);
+        resolve(getConsentStatus());
+      });
+    });
+  }
+
+  a.addEventListener('click', async (e) => {
+    e.preventDefault();
+    let loader;
+    if (getMetadata('quick-link-loader') === 'on') loader = addLoader(a);
```
mokimo, line 71: I think this should also work
```suggestion
    let loader = (getMetadata('quick-link-loader') === 'on') && addLoader(a)
```

mokimo (approved):

salonijain3 (approved):

JasonHowellSlavin (approved):

spadmasa:
Validation done on the pr on the below urls :
Quick Link with ECID passed, with no loader and a delay of around 3s:
https://ecid-test--milo--drashti1712.hlx.live/drafts/drashti/milo-banner/ql-noloader
Quick Link with ECID passed after privacy consent check, with a progress bar loader:
https://ecid-test--milo--drashti1712.hlx.live/drafts/drashti/milo-banner/ql-progress-bar-loader

https://github.com/user-attachments/assets/ab05b75c-c8ff-4bde-875b-04fd3d76241c


https://github.com/user-attachments/assets/5dbd9532-baf4-4775-8c80-cb691f97abeb

image
image
image
image
results are added in subtask https://jira.corp.adobe.com/browse/MWPW-158154

PR #3438, opened 2025-01-08
MWPW-161867: Add http check to preflight

JasonHowellSlavin:
* Adds a data attribute to links if the link was authored with http instead of https
* Adds error messaging in preflight if the http data attribute is found

Resolves: [MWPW-161867](https://jira.corp.adobe.com/browse/MWPW-161867)

**Test URLs:**
- Before: https://main--milo--jasonhowellslavin.hlx.live/drafts/slavin/bacom-pages/rtcdp?martech=off
- After: https://add-http-check-to-preflight--milo--jasonhowellslavin.hlx.live/drafts/slavin/bacom-pages/rtcdp?martech=off

### DIFF
--- a/libs/blocks/preflight/panels/seo.js
+++ b/libs/blocks/preflight/panels/seo.js
@@ -227,8 +227,9 @@ async function checkLinks() {
         && !link.closest('.preflight') // Is not inside preflight
         && !knownBadUrls.some((url) => url === link.hostname) // Is not a known bad url
       ) {
-        link.liveHref = link.href.replace('hlx.page', 'hlx.live');
-        link.liveHref = link.href.replace('aem.page', 'aem.live');
+        link.liveHref = link.href;
+        if (link.href.includes('hlx.page')) link.liveHref = link.href.replace('hlx.page', 'hlx.live');
+        if (link.href.includes('aem.page')) link.liveHref = link.href.replace('aem.page', 'aem.live');
         return true;
       }
       return false;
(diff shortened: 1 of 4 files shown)


skumar09:
@JasonHowellSlavin : Nala tests are run by default on pr branches, so we no longer require `run-nala` label

mokimo (approved):

libs/utils/utils.js
```diff
@@ -693,6 +693,7 @@ export function decorateLinks(el) {
   const { hostname, href } = window.location;
   const links = [...anchors].reduce((rdx, a) => {
     appendHtmlToLink(a);
+    if (a.href.includes('http:')) a.setAttribute('data-http-link', 'true');
```
narcis-radu, line 696: Any particular reason to use the data attribute here? We can just do the detection in `seo.js`.

libs/blocks/preflight/panels/seo.js
```diff
@@ -237,6 +237,16 @@ async function checkLinks() {
   const baseOpts = { method: 'POST', headers: { 'Content-Type': 'application/json' } };
   const badResults = [];
 
+  links.forEach((link) => {
+    if (link.dataset?.httpLink) {
```
narcis-radu, line 241:
```suggestion
    if (link.href.includes('http:')) {
```

JasonHowellSlavin: I resolved the above conversation with a comment, please let me know if that is not satisfactory @narcis-radu. Hoping to resolve this one as well.

JasonHowellSlavin: Resolving this conversation. I believe the linked discussion in the above comment illustrates why we need to do this check in utils, i.e. before milo link transformation.

PR #3456, opened 2025-01-10
[MWPW-164362] Table icon focus

DKos95:
This resolves the issue where the icon wasn't focusable while tabbing through the website making it inaccessible for screen readers.

Resolves: [MWPW-[164362](https://jira.corp.adobe.com/browse/MWPW-164362)]

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/docs/library/kitchen-sink/table?martech=off
- After: https://table-icon-focus--milo--adobecom.hlx.page/docs/library/kitchen-sink/table?martech=off


libs/blocks/table/table.css
```diff
@@ -824,4 +823,4 @@
   .table .action-area .con-button.button-l {
     overflow: hidden;
   }
-}
+}
```
overmyheadandbody, line 826: We should likely leave the spacing in this file as-is.

libs/features/icons/icons.css
```diff
@@ -4,7 +4,7 @@
   border-bottom: none;
 }
 
-.milo-tooltip::before {
+.milo-tooltip::before {
```
overmyheadandbody, line 7: Additional space probably not needed.

libs/features/icons/icons.js
```diff
@@ -46,6 +46,10 @@ function decorateToolTip(icon) {
   // Position is the next to last part of a tooltip
   const place = conf.pop()?.trim().toLowerCase() || 'right';
   icon.className = `icon icon-info milo-tooltip ${place}`;
+  icon.setAttribute('tabindex', '0');
+  icon.setAttribute('role', 'button');
+  icon.setAttribute('aria-label', content);
+  icon.setAttribute('aria-describedby', content);
```
overmyheadandbody, line 52: I'm not sure this is needed. `aria-describedby` should reference an ID, documentation [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby). Setting the `aria-label` should probably be enough.

overmyheadandbody (commented):
Let's also remove `table.css` from the changes. In the meantime, I'll reach out to our a11y contact to check whether announcing the tooltip content without user interaction is acceptable. Usually, they would need to press Enter/Return/Space to expand the content, but I'm not sure whether that's a requirement.

libs/blocks/table/table.css
```diff
@@ -322,14 +322,16 @@
 }
 
 /* icons */
-.table:not(.merch) .row .col span.milo-tooltip {
-  margin-left: var(--spacing-xs);
-  margin-right: 0;
+[dir='ltr'] .table:not(.merch) .row .col span.milo-tooltip {
```
narcis-radu, line 325: I don't think we need `[dir='ltr']` here, we can just write as:
```cs
.table:not(.merch) .row .col span.milo-tooltip {
  margin-right: -4px;
}
```
and make sure that on `[dir='rtl']` we reset the `margin-right` value to `0`
```css
[dir='rtl'] .table:not(.merch) .row .col span.milo-tooltip {
  margin-right: 0;
  margin-left: -4px;
}
```

libs/features/icons/icons.css
```diff
@@ -2,6 +2,11 @@
   position: relative;
   text-decoration: none;
   border-bottom: none;
+  min-width: 24px;
```
narcis-radu, line 5: This will introduce additional space, you can see the difference between (new) https://table-icon-focus--milo--adobecom.hlx.page/docs/library/kitchen-sink/table?martech=off and (old) https://main--milo--adobecom.hlx.page/docs/library/kitchen-sink/table?martech=off. Check for the quotes - info icon - quotes.

Screenshot 2025-01-22 at 13 48 10

Is there any usage for the min size?

DKos95: This was an additional requirement for the tooltip to be 24x24 from Taryn after the meeting.

overmyheadandbody: I think for this one the issue is more around the `margin-inline-start` and `margin-inline-end` classes, which in this case lead to an odd effect. A11y guidelines do require us to have the clickable area of the tooltip be a minimum of 24x24px, but the additional margins lead to a strange effect. My 2 cents - we've added lateral padding for the icon to meet A11y criteria, so we could cancel out the margins for the tooltip icon, wdyt?

DKos95: Putting both margins at 3px for the .milo-tooltip instead of 8px makes it identical as on main from what I'm seeing testing it currently, which makes sense since before the tooltip was 14px width and now it is 24px.

overmyheadandbody (commented):
This looks good overall, but I just noticed another side-effect of increasing the hit area of the icon. There's an 8px area on each side where hovering over with a mouse does trigger the tooltip to be shown, but doesn't trigger the icon's hover state. Probably not something critical, but it would be great to consider this as well.

Screenshot 2025-01-23 at 18 16 37
Screenshot 2025-01-23 at 18 19 35

overmyheadandbody (approved):
Looks good!

I did notice that the top/bottom tooltip positioning options are not centered with the icon, but I see that's already a problem in the `main` branch as well. If it's an easy fix, it might be worth addressing now, otherwise we can open a follow-up.

narcis-radu (approved):

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-164362

PR #3457, opened 2025-01-10
[MWPW-164529] [MWPW-164530] Sticky section a11y issues

robert-bogos:
## Description
This PR removes the logic which makes the sticky section hide when the footer is in the viewport.

## Related Issue
Resolves: [MWPW-164529](https://jira.corp.adobe.com/browse/MWPW-164529)
Resolves: [MWPW-164530](https://jira.corp.adobe.com/browse/MWPW-164530)

## Testing instructions
**MWPW-164529**
1. Open the test URL.
2. Use `Tab` to navigate the page.
3. Should be able to reach the CTA in the sticky section.

**MWPW-164530**
1. Open the test URL.
2. Open VoiceOver.
3. Should hear the contents of the sticky section when scrolled to the bottom of the page.

## Screenshots:
**Before:**
Screenshot 2025-01-21 at 10 35 28

**After:**
Screenshot 2025-01-21 at 10 36 20

## Test URLs

**Notification default variant:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rbogos/sticky-section/notification-full-page?martech=off
- After: https://mwpw-164529-a11y--milo--robert-bogos.hlx.page/drafts/rbogos/sticky-section/notification-full-page?martech=off

**Notification no-hide variant:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rbogos/sticky-section/notification-nohide-full-page?martech=off
- After: https://mwpw-164529-a11y--milo--robert-bogos.hlx.page/drafts/rbogos/sticky-section/notification-nohide-full-page?martech=off

**Aside Promobar with delay:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rbogos/sticky-section/promobar-full-page?martech=off
- After: https://mwpw-164529-a11y--milo--robert-bogos.hlx.page/drafts/rbogos/sticky-section/promobar-full-page?martech=off

**Aside Promobar without delay:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rbogos/sticky-section/promobar-no-delay-full-page?martech=off
- After: https://mwpw-164529-a11y--milo--robert-bogos.hlx.page/drafts/rbogos/sticky-section/promobar-no-delay-full-page?martech=off


overmyheadandbody (commented):
No issues with the code, but I think this may need some design input. This looks a bit weird to me.
Screenshot 2025-01-13 at 12 43 42

I'll try to loop in someone from Consonant. In the meantime, could you do a quick assessment to check whether we can easily change the notification styles to cover the width of the page once we reach that stopping point? I see we do have edge-to-edge notifications on the example pages, I'm wondering whether it's feasible to switch styles or we can't go back on the pill notification decoration decisions.

robert-bogos:
FYI: The failing code coverage test seems to be flaky. Locally the lines are covered.
Screenshot 2025-01-21 at 11 29 57

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-164530, https://jira.corp.adobe.com/browse/MWPW-164529.

robert-bogos:
Failed code coverage test is not related to the current changes. All new lines are covered.

Screenshot 2025-01-21 at 16 16 08

overmyheadandbody (commented):
The behavior seems to have changed between the first before/after link group. The notification is now added to the top and then moved to be sticky. This also leads to a CLS increase.

https://github.com/user-attachments/assets/0fd079fc-e571-4691-9455-abb97f4b91b7

overmyheadandbody:
A similar situation occurs for the second pair of test links on mobile. Previously, the notification would only appear after scroll, now it is initially visible even at the top of the page.

test/blocks/section-metadata/section-meta.test.js
```diff
@@ -83,6 +83,9 @@ describe('Section Metdata', () => {
     const sm = sec.querySelector('.section-metadata');
     await init(sm);
     expect(main.lastElementChild.classList.contains('hide-sticky-section')).to.be.true;
+    window.scrollTo(0, document.body.scrollHeight);
+    await delay(500);
```
overmyheadandbody, line 87: Could we use `clock.runAllAsync();` or something similar to avoid the wait time?

overmyheadandbody (approved):
Looking good now, thanks for all the updates!

test/blocks/section-metadata/section-meta.test.js
mokimo, line 87: Yeah let's not use a delay, this introduces flakiness and is not reliable. ( //cc @skholkhojaev )

robert-bogos: In this case it would not work because clock.runAllAsync() does not trigger intersection observer callbacks, which is what we need.

mokimo: According to the docs it should break the event loop, is there any async code within the intersection observer that you might need to account for, AFTER it has run?

https://sinonjs.org/releases/latest/fake-timers/

> The runAllAsync() will also break the event loop, allowing any scheduled promise callbacks to execute before running the timers.

PR #3465, opened 2025-01-13
Local nav should be displayed when the first element is dropdown

bandana147:
* Added support for Localnav with first item as a dropdown element in  the new redesigned mobile gnav 

Resolves: [MWPW-164828](https://jira.corp.adobe.com/browse/MWPW-164828)

**Test URLs:**
- After: https://mgnav-dropdown--milo--adobecom.aem.page/?martech=off

QA: https://main--cc--adobecom.hlx.page/drafts/blaishram/document2?milolibs=mgnav-dropdown&newNav=true

### DIFF
--- a/libs/blocks/global-navigation/global-navigation.js
+++ b/libs/blocks/global-navigation/global-navigation.js
@@ -405,7 +405,7 @@ class Gnav {
   decorateLocalNav = async () => {
     if (!this.isLocalNav()) return;
     const localNavItems = this.elements.navWrapper.querySelector('.feds-nav').querySelectorAll('.feds-navItem:not(.feds-navItem--section, .feds-navItem--mobile-only)');
-    const firstElem = localNavItems[0]?.querySelector('a');
+    const firstElem = localNavItems[0]?.querySelector('a') || localNavItems[0]?.querySelector('button');
     if (!firstElem) {
       lanaLog({ message: 'GNAV: Incorrect authoring of localnav found.', tags: 'errorType=info,module=gnav' });
       return;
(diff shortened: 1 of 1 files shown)


libs/blocks/global-navigation/global-navigation.js
```diff
@@ -402,13 +402,19 @@ class Gnav {
...
 
+  removeLocalNav = () => {
+    lanaLog({ message: 'Gnav Localnav was removed, potential CLS' });
+    document.querySelector('.feds-localnav')?.remove();
+    return;
+  }
+
   decorateLocalNav = async () => {
-    if (!this.isLocalNav()) return;
+    if (this.isLocalNav()) this.removeLocalNav();
```
sharmrj, line 412: should it also return here?

libs/blocks/global-navigation/global-navigation.js
```diff
@@ -402,13 +402,19 @@ class Gnav {
...
     const localNavItems = this.elements.navWrapper.querySelector('.feds-nav').querySelectorAll('.feds-navItem:not(.feds-navItem--section, .feds-navItem--mobile-only)');
-    const firstElem = localNavItems[0]?.querySelector('a');
+    const firstElem = localNavItems[0]?.querySelector('a') || localNavItems[0]?.querySelector('button');
     if (!firstElem) {
       lanaLog({ message: 'GNAV: Incorrect authoring of localnav found.', tags: 'errorType=info,module=gnav' });
-      return;
+      this.removeLocalNav();
```
sharmrj, line 417: and return here as well? That way it'll be less confusing to read and modify the logic below.

sharmrj (approved):
Make sure to fix lint

libs/blocks/global-navigation/global-navigation.js
bandana147: It returns from `removeLocalNav` function

PR #3473, opened 2025-01-13
MWPW-162027 Catalog action menu A11Y collapsed expanded

bozojovicic:
The screen reader should announce the expand/collapse state when the keyboard focus moves to the Action Menu button.

Test page : https://main--cc--adobecom.hlx.live/products/catalog?milolibs=mwpw162027a11y--milo--bozojovicic

Resolves: [MWPW-162027](https://jira.corp.adobe.com/browse/MWPW-162027)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.live/drafts/bozo/price-cta?martech=off
- After: https://mwpw162027a11y--milo--bozojovicic.aem.live/drafts/bozo/price-cta?martech=off


jpratt2:
Regarding the failed nala tests, could you try this: pull the recent changes from stage into your branch and then re-push your branch
cc: @skumar09

bozojovicic:
15 times I tried to rebuild this PR and always some random test fails : unit tests, coverage, nala or psi-check
I give up for today

libs/features/mas/src/variants/catalog.js
```diff
@@ -80,14 +81,18 @@ export class Catalog extends VariantLayout {
     };
 
     toggleActionMenu = (e) => {
-      const actionMenuContentSlot = this.card.shadowRoot.querySelector(
+      const shadowRoot = this.card.shadowRoot;
+      const actionMenu = shadowRoot.querySelector('.action-menu');
+      const actionMenuContentSlot = shadowRoot.querySelector(
```
3ch023, line 86: both actionMenu and actionMenuContentSlot could have been getter?
to avoid duplication in line 84 and 116

3ch023 (changes requested):
not sure the fix will work as expected, maybe @Roycethan can do a pre-test on NVDA.
If it works we can merge as is.

But on mac voiceover it didn't work for me, maybe because it ticket request was to swith to actionable element like button before adding aria-expanded attribute. 
In https://main--cc--adobecom.hlx.page/products/catalog?milolibs=mwpw162027a11y--milo--bozojovicic however it is still a div

libs/features/mas/src/variants/catalog.js
bozojovicic: Fixed

```
get actionMenu() {
    return this.card.shadowRoot.querySelector('.action-menu');
}

get actionMenuContentSlot() {
    return this.card.shadowRoot.querySelector('slot[name="action-menu-content"]');
}
```

bozojovicic:
@3ch023 AC says 

> The screen reader should announce the expand/collapse state when the keyboard focus moves to the button.

And that's what I did. And it works when I try it with VoiceOver. It doesn't say that the screen reader should announce the expand/collapse state every time when user opens/closes it. Maybe that's what you tried to do.

Roycethan:
> not sure the fix will work as expected, maybe @Roycethan can do a pre-test on NVDA. If it works we can merge as is.
> 
> But on mac voiceover it didn't work for me, maybe because it ticket request was to swith to actionable element like button before adding aria-expanded attribute. In https://main--cc--adobecom.hlx.page/products/catalog?milolibs=mwpw162027a11y--milo--bozojovicic however it is still a div

@3ch023 Checked on NVDA - windows fix looks good ... screen reader announces collapsed/expanded based on action performed on the button. May be a Vocice over bug in Mac for Safari, but as i checked it also works for Mac+Chrome combination. Post 2 approvals will do regression around.

cc @bozojovicic

3ch023 (approved):

npeltier (approved):

Roycethan:
> Skipped 3473: "MWPW-162027 Catalog action menu A11Y collapsed expanded" due to file "libs/deps/mas/mas.js" overlap. Merging will be attempted in the next batch

@bozojovicic You might need to resolve the conflicts here or pull latest for merge to happen...

PR #3475, opened 2025-01-14
MWPW-159369 [cc][LANA] Error reported: target response time: 0.5

vgoodric:
* Set sample rate to be lower

Resolves: [MWPW-159369](https://jira.corp.adobe.com/browse/MWPW-159369)

**Test URLs:**
- After: https://mepreducelana--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -1204,7 +1204,7 @@ async function handleMartechTargetInteraction(
       try {
         window.lana.log(`target response time: ${roundedResponseTime}`, {
           tags: 'martech',
-          errorType: 'i',
+          errorType: 'e',
           sampleRate: 0.5,
         });
       } catch (e) {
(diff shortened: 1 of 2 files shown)


libs/features/personalization/personalization.js
```diff
@@ -1202,7 +1202,11 @@ async function handleMartechTargetInteraction(
       performance.clearMarks();
       performance.clearMeasures();
       try {
-        window.lana.log(`target response time: ${roundedResponseTime}`, { tags: 'martech', errorType: 'i' });
+        window.lana.log(`target response time: ${roundedResponseTime}`, {
+          tags: 'martech',
+          errorType: 'i',
```
jedjedjedM, line 1207: Can you change the error type to "e".  The documentation was incorrect.

_i = implicit - log that is not explicitly accounted for or defined, but is thrown by the browser / app

e = explicit - log of a very specific and defined case_

vgoodric: done

chrischrischris (approved):

denlight (approved):

PR #3485, opened 2025-01-15
[MWPW-157116] - Add feature for pause animation on hover for PEP

Deva309:
**Implemented the 'Pause On Hover' feature for Feds PEP content, allowing it to adopt the same functionality available for Express PEP.**

Authoring changes via section metadata in the PEP content to support the ability to author this feature:
Screenshot 2024-09-03 at 6 02 52 PM

Due deligence wiki: https://wiki.corp.adobe.com/display/WEBT/%5BDue+Diligence%5D+PEP+-+Federalizing+Express

Resolves: [MWPW-157116](https://jira.corp.adobe.com/browse/MWPW-157116)

**Test URLs:**
- After: https://mwpw-157116--milo--deva309.aem.page/?martech=off

QA:
- https://feds--milo--adobecom.hlx.page/drafts/devashish/pep/pep-prompt?skipPepEntitlements=true (without pause on hover)
- https://feds--milo--adobecom.hlx.page/drafts/devashish/pep/pause/pep-prompt-pause-animation?skipPepEntitlements=true (with pause on hover)

### DIFF
--- a/libs/features/webapp-prompt/webapp-prompt.css
+++ b/libs/features/webapp-prompt/webapp-prompt.css
@@ -184,7 +184,7 @@
     transform-origin: top right;
   }
 
-  .appPrompt:hover .appPrompt-progress {
+  .appPrompt:hover .appPrompt-progressPauseOnHover {
     animation-play-state: paused;
   }
 
--- a/libs/features/webapp-prompt/webapp-prompt.css
+++ b/libs/features/webapp-prompt/webapp-prompt.css
@@ -184,6 +184,10 @@
     transform-origin: top right;
   }
 
+  .appPrompt:hover .appPrompt-progress {
+    animation-play-state: paused;
+  }
+
   @keyframes appPrompt-animate {
     100% {
       transform: scaleX(1);
--- a/libs/features/webapp-prompt/webapp-prompt.js
+++ b/libs/features/webapp-prompt/webapp-prompt.js
@@ -231,7 +231,7 @@ export class AppPrompt {
   };
 
   decorate = () => {
-    const animationPlayState = this.options['pause-on-hover'] === 'on';
+    const animationPauseOnHover = this.options['pause-on-hover'] === 'on';
     this.elements.closeIcon = toFragment`<button daa-ll="Close Modal" aria-label="${this.cancel}" class="appPrompt-close"></button>`;
     this.elements.cta = toFragment`<button daa-ll="Stay on this page" class="appPrompt-cta appPrompt-cta--close">${this.cancel}</button>`;
     this.elements.profile = this.profile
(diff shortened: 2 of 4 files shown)


libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -278,10 +281,35 @@ export class AppPrompt {
...
+      startTime = Date.now();
+      timeoutId = setTimeout(() => {
+        this.close({ saveDismissal: false, dismissalActions: false });
+        AppPrompt.redirectTo(this.options['redirect-url']);
+      }, remainingTime);
+    };
+
+    const stopTimeout = () => {
+      clearTimeout(timeoutId);
+      remainingTime -= Date.now() - startTime;
```
sharmrj, line 299: Use `performance.now` instead. `Date` depends on the user's system clock which may or may not be accurate.

libs/features/webapp-prompt/webapp-prompt.css
```diff
@@ -184,6 +184,10 @@
     transform-origin: top right;
   }
 
+  .appPrompt:hover .appPrompt-progress {
+    animation-play-state: paused;
```
sharmrj, line 188: I may be mistaken but, since this class is always on the prompt, the animation will pause even if pause on hover is turned off, right?

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -257,7 +260,7 @@ export class AppPrompt {
         ${this.elements.cta}
       </div>
       <div class="appPrompt-progressWrapper">
-        <div class="appPrompt-progress" style="background-color: ${this.options['loader-color']}; animation-duration: ${this.options['loader-duration']}ms;"></div>
+        <div class="appPrompt-progress" style="background-color: ${this.options['loader-color']}; animation-duration: ${this.options['loader-duration']}ms; ${!animationPlayState && 'animation-play-state: running;'}"></div>
```
sharmrj, line 263: Shouldn't it have `animation-play-state: running` by default? In that case shouldn't this be in the css stylesheet rather than inline here; since you also have the `animation-play-state: paused` rule there as well it makes sense for them to be colocated?

libs/features/webapp-prompt/webapp-prompt.js
Deva309: Done.
Using `performance.now` instead of `Date.now`.

libs/features/webapp-prompt/webapp-prompt.css
Deva309: I have updated the implementation for this. Now, the class `appPrompt-progressPauseOnHover` is added when `pause-on-hover` is enabled. CSS has been applied specifically to this class, ensuring that the styles are applied only when `pause-on-hover` is enabled.

sharmrj (approved):

sonawanesnehal3 (approved):

PR #3492, opened 2025-01-15
MWPW-165302: Calculate promo prices with duration

yesil:
This PR introduces additional annual price logic calculation based on the promotion display details in the WCS response.

Resolves: [MWPW-165302](https://jira.corp.adobe.com/browse/MWPW-165302)

**Screenshots**
image

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/au/drafts/ilyas/inline-price?martech=off
- After: https://mwpw-165302--milo--adobecom.aem.page/au/drafts/ilyas/inline-price?martech=off


libs/features/mas/src/price/utilities.js
```diff
@@ -252,15 +309,67 @@ const formatRegularPrice = ({ commitment, term, ...data }) =>
...
+            return formatPrice(data, RecurrenceTerm.YEAR, getAnnualPrice);
+        }
+        const {
+            displaySummary: { outcomeType, duration, minProductQuantity = 1 } = {},
+        } = promotion;
+        switch (outcomeType) {
+            case 'PERCENTAGE_DISCOUNT': {
+                if (
+                    quantity >= minProductQuantity &&
+                    isPromotionActive(promotion, instant)
```
bozojovicic, line 340: @yesil  What if `outcomeType` is `'PERCENTAGE_DISCOUNT'` but this condition 
`quantity >= minProductQuantity && isPromotionActive(promotion, instant)` in IF is false?
Then I guess `formatAnnualPrice()` will return undefined, it will not go to `default:`, that might create problems since it is expected that this function will return some object but I don't know if that is possible in real world. Just asking.

yesil: @bozojovicic it goes to `default:` default because there is no `break;`.

these cases are covered by unit tests:

https://github.com/adobecom/milo/pull/3492/files#diff-66faad1f378a7658cebf33b4a9fa66d2ec4ee53039ec0e6fd5ce41930c92c13eR343-R374

bozojovicic (approved):

3ch023 (approved):

Roycethan:
@yesil  Verified , plz resolve the JS conflicts thanks

PR #3498, opened 2025-01-16
[ENB-7652] Refactor flags and methods

swamu:
Add pageview send event, fix window.satellite empty issue
https://main--cc--adobecom.hlx.live/drafts/suhjain/swati-pages/photoshop-v2?target=on&milolibs=ENB-7652-refactor-enablePerseV2--milo--swamu#

Resolves: [ENB-7652](https://jira.corp.adobe.com/browse/ENB-7652)

Test URLs:
After: https://ENB-7652-refactor-enablePerseV2--milo--swamu.aem.page/?martech=off


Bhim12 (approved):

Nadeemakramn (approved):
Verified all scenarios in QA environment. Hence Approved it.

chrischrischris (approved):

vgoodric:
@swamu I think it would be better to add the pzn v2 flag in config.mep and pull it back out of config when possible instead of  passing it to so many functions. if you look at the top of the init function in personalization.js you'll see we add several settings for later use related to MEP.  It feels like this will also make it easier to remove later if we fully roll the feature out.

Also wondering if we want to leverage the utils.js function `getMepEnablement` in `enablePersonalizationV2`. This would let us use parameter override to test the feature on pages before we have the file itself updated.

swamu:
@vgoodric , noted and updated with both the points

Nadeemakramn:
Retested all new commits, Good to Merge.

PR #3507, opened 2025-01-17
MWPW-156181: a11y merch-card-collection fix - cards reordering

3ch023:
Screenreader is building a tree based on the dom so we can't rely on on aria 'tabindex' attribute.
Unfortunately we need to rearrange dom elements for screenreader every time user clicks on filter in merch-card-collection.

Measurements before/after:
https://pagespeed.web.dev/analysis/https-main--cc--adobecom-hlx-live-products-catalog/0kueyog7x7?form_factor=desktop
https://pagespeed.web.dev/analysis/https-main--cc--adobecom-hlx-live-products-catalog/fghnmjw4pg?form_factor=desktop

Resolves: [MWPW-156181](https://jira.corp.adobe.com/browse/MWPW-156181)

Catalog page:
https://main--cc--adobecom.hlx.live/products/catalog?milolibs=MWPW-156181&martech=off

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/docs/library/kitchen-sink/merch-card?martech=off
- After: https://MWPW-156181--milo--adobecom.aem.page/docs/library/kitchen-sink/merch-card?martech=off


yesil (approved):

libs/features/mas/src/merch-card-collection.js
```diff
@@ -156,20 +156,19 @@ export class MerchCardCollection extends LitElement {
             this.hasMore = result.length > pageSize;
             result = result.filter(([, index]) => index < pageSize);
         }
-
-        let reduced = new Map(result);
+        let reduced = new Map(result.reverse());
+        for (const card of reduced.keys()) {
+          this.prepend(card);
+        }
```
npeltier, line 162: doesn't this work and would be simpler?
```suggestion
        let reduced = new Map(result());
        for (const card of reduced.keys()) {
          this.appendChild(card);
        }
```

npeltier (approved):

libs/features/mas/src/merch-card-collection.js
npeltier: doesn't look like it would work at all :-p

3ch023: yeah visible cards need to be in the beginning

PR #3513, opened 2025-01-20
[MWPW-165791] Table select aria-label and column cell role added

DKos95:
This resolves the issues where a table row was missing an adequate role and select on mobile was missing a label causing accessibility issues.

Resolves: [MWPW-165791](https://jira.corp.adobe.com/browse/MWPW-165791)

| **Before** | **After** |
|------------|-----------|
|Screenshot 2025-01-20 at 12 33 12|Screenshot 2025-01-20 at 12 33 18|

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/nala/blocks/table/table?martech=off
- After: https://table-select-cell--milo--adobecom.hlx.page/drafts/nala/blocks/table/table?martech=off


libs/blocks/table/table.js
```diff
@@ -508,6 +516,7 @@ function applyStylesBasedOnScreenSize(table, originTable) {
       table.parentElement.insertBefore(filters, table);
       table.parentElement.classList.add(`table-${table.classList.contains('merch') ? 'merch-' : ''}section`);
       filterChangeEvent();
+      setAriaLabelsForElements(colSelect0, colSelect1);
```
overmyheadandbody, line 519: I'd like to future-proof this a bit, in case we'll have another number of columns visible on different breakpoints. So let's send these in an Array and have a single argument for `setAriaLabelsForElements` that iterates over the array's elements to set the `aria-label`.

test/blocks/table/table.test.js
```diff
@@ -1,10 +1,13 @@
...
 import { expect } from 'chai';
-import { MILO_EVENTS } from '../../../libs/utils/utils.js';
+import { getConfig, MILO_EVENTS, setConfig } from '../../../libs/utils/utils.js';
 import { delay, waitForElement } from '../../helpers/waitfor.js';
 
 document.body.innerHTML = await readFile({ path: './mocks/body.html' });
 const { default: init } = await import('../../../libs/blocks/table/table.js');
+const config = getConfig();
+config.env = { locale: { contentRoot: `${window.location.origin}` } };
+setConfig(config);
```
overmyheadandbody, line 10: Wouldn't this end up actually making the placeholders call during the unit tests? We generally avoid network requests, so this should be mocked. I suspect the palceholders tests should have a way of doing this already.

test/blocks/table/table.test.js
```diff
@@ -111,5 +114,15 @@ describe('table and tablemetadata', () => {
       expect(tooltipHeading.childNodes.length).to.equal(2);
       expect(tooltipHeading.querySelector('.milo-tooltip, .icon-tooltip')).to.exist;
     });
+
+    it('should apply aria-label to all selects within .filters on mobile', async () => {
+      window.innerWidth = 375;
+      window.dispatchEvent(new Event('resize'));
+      await delay(500);
```
overmyheadandbody, line 121: We try to avoid any wait time in our unit tests as to not delay the whole suite. You can look up `await clock.runAllAsync();` in the codebase and that should provide a hint on how to wait for things to happen without impact on run time.

test/blocks/table/table.test.js
```diff
@@ -111,5 +114,15 @@ describe('table and tablemetadata', () => {
...
+      const selectElements = document.querySelectorAll('.filters select');
+      selectElements.forEach((selectElement) => {
+        expect(selectElement.getAttribute('aria-label')).to.equal('choose table column');
```
overmyheadandbody, line 124: Again, we should probably mock the placeholders logic/values, so we might need to adapt the expected value here as well.

libs/blocks/table/table.js
```diff
@@ -153,6 +153,15 @@ function handleAddOnContent(table) {
   table.addEventListener('mas:resolved', debounce(() => { handleEqualHeight(table, '.row-heading'); }));
 }
 
+async function setAriaLabelsForExpandableIcons() {
```
overmyheadandbody, line 156: Should we have a unit test for this as well?

libs/blocks/table/table.js
```diff
@@ -153,6 +153,15 @@ function handleAddOnContent(table) {
...
+  const config = getConfig();
+  const ariaLabel = await replaceKey('toggle-row', config);
+  const icons = document.querySelectorAll('.icon.expand[role="button"]');
```
overmyheadandbody, line 159: Although I agree that all expand icons should get the `aria-label` attribute, I'm not sure this is the place to do it. I'd either apply the behavior in a more general icon file or scope these changes to just the table we're decorating.

libs/blocks/table/table.js
```diff
@@ -448,6 +457,15 @@ function applyStylesBasedOnScreenSize(table, originTable) {
...
       || (isMerch && !table.querySelector('.col-2'))) return;
 
+    async function setAriaLabelsForElements(selectArray) {
+      const config = getConfig();
+      const ariaLabel = await replaceKey('choose-table-column', config);
+
+      selectArray.forEach((item) => {
+        item.setAttribute('aria-label', ariaLabel);
+      });
+    }
```
overmyheadandbody, line 467: This logic and the one from `setAriaLabelsForExpandableIcons`, are quite similar, I'd suggest we combine them into a more general method that we can call. Another tip: there is a `replaceKeyArray` with which you could get all the placeholders in one go, if that's helpful for the generic method.

test/blocks/table/table.test.js
```diff
@@ -1,10 +1,15 @@
...
+import { getConfig, MILO_EVENTS, setConfig } from '../../../libs/utils/utils.js';
 import { delay, waitForElement } from '../../helpers/waitfor.js';
+import { replaceKey } from '../../../libs/features/placeholders.js';
 
 document.body.innerHTML = await readFile({ path: './mocks/body.html' });
 const { default: init } = await import('../../../libs/blocks/table/table.js');
+const locales = { '': { ietf: 'en-US', tk: 'hah7vzn.css' } };
+const conf = { locales, contentRoot: '/test/blocks/table/mocks' };
+setConfig(conf);
+const config = getConfig();
```
overmyheadandbody, line 12: These seem to be relevant to a single test, should we scope them to that particular test? I'm not sure how we treat this in other unit tests, might be worth a look.

DKos95: This seems to not be possible since in table.test.js we are initializing table.js and the calls for aria label get executed immediately and not just for that test.

overmyheadandbody (approved):

mokimo (approved):

skumar09:
@DKos95: Thanks a lot for fixing the accessibility issue for Table block ( Nala test id: 0)

Could you please also add/enable the nala accessibility test for the table

**Reference:** 
**File**: `/nala/blocks/table/table.test.js`

`import { runAccessibilityTest } from '../../libs/accessibility.js';`

// Test 0 : Table block (default)

```
    await test.step('step-3: Verify the accessibility test on the quote (contained) block', async () => {
      await runAccessibilityTest({ page, testScope: table.table });
    });
```
Please also refer to other blocks' nala accessibility test step on how to enable/disable examples

cc: @overmyheadandbody  @SilviuLCF @narcis-radu @mokimo

skumar09 (changes requested):
Please update the nala accessibility step for Table ( Default ) test.

skumar09 (approved):

SilviuLCF (approved):
Verified 
Testing details https://jira.corp.adobe.com/browse/MWPW-165791

PR #3518, opened 2025-01-21
MWPW-162026 [Milo][Sev2][Catalog] Group of checkboxes is missing FIELDSET …

bozojovicic:
… element - filter checkboxes

A11Y fix - In `merch-sidenav-checkbox-group` For Type Desktop/Mobile/Web checkbox group ARIA group/labelledby technique applied to announce the grouping as a part of the checkboxes. This does not work if we have aria-labelledby="elementId" and #elementId is in shadow DOM so I had to create new hidden DIV, outside shadow DOM, which will have that ID and contain the text "Types".

Test page : /products/catalog?milolibs=mwpw162026a11y--milo--bozojovicic

Resolves: [MWPW-162026](https://jira.corp.adobe.com/browse/MWPW-162026)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.live/drafts/bozo/price-cta?martech=off
- After: https://mwpw162026a11y--milo--bozojovicic.aem.live/drafts/bozo/price-cta?martech=off


npeltier (approved):

libs/features/mas/src/sidenav/merch-sidenav-checkbox-group.js
```diff
@@ -69,10 +70,22 @@ export class MerchSidenavCheckboxGroup extends LitElement {
         pushStateFromComponent(this, this.selectedValues.join(','));
     }
 
+    addAccessibilityAttributes() {
+        const id = 'sidenav-checkbox-group-title';
+        const groupIdEl = createTag('div', { class: 'invisible-and-shrank', id });
+        groupIdEl.textContent = this.sidenavCheckboxTitle;
+        this.append(groupIdEl);
+        this.querySelectorAll('sp-checkbox').forEach((checkboxEl) => {
```
3ch023, line 78: i am not sure i like the idea of merch-sidenav-checkbox-group being bound to sp-checkbox.
can't we add this 'role' attribute when we create those 'sp-checkboxes'? https://github.com/adobecom/cc/blob/stage/creativecloud/blocks/sidenav/sidenav.js#L79
I understand it means we need to rely on some stable id, so not sure if its better or not. maybe a bit less confusing for devs

3ch023: and same with `groupIdEl` element, it could be probably created by sidenav block
@npeltier @yesil wdyt?

npeltier (changes requested):
thanks @3ch023 

looked at it again. I guess @bozojovicic it would be better to pull out the group headings of the shadow DOM to the light DOM and you should be good. 
We should not duplicate content and hack around this.

libs/features/mas/src/sidenav/merch-sidenav-checkbox-group.js
yesil: I know the JIRA recommends a few options, but can we try the OOTB SWC approach with `sp-field-group`
https://opensource.adobe.com/spectrum-web-components/components/field-group/

image

Before implementing this solution, the accessibility tool can be run against checkboxes in the SWC documentation 

image
if it doesn't complain we can follow.

yesil (changes requested):

libs/features/mas/src/sidenav/merch-sidenav-checkbox-group.js
bozojovicic: @yesil I tried your idea directly on catalog page
Screenshot 2025-01-27 at 12 48 10
it doesn't work. :( Screen reader ignores the word "Types" although it is there twice, once in element "label" and once in attribute "label"

@3ch023 we had this in CC code before but then it was requested to move to Milo

@npeltier I can move that "types" element from shadow to light dom but again I need to iterate through checkboxes to set attributes 'aria-labelledby', and that is where Mariia complains about

So, option A : I do it in CC code in sidenav.js, I move the H3 element from shadow to light dom
option B : I do the same thing in milo code
For both options one reviewer will not be happy.

@3ch023 I could also try to iterate through direct children of `merch-sidenav-checkbox-group`, without bounding to `sp-checkbox`, in that case it would be OK to do it in Milo code, right? I think that is the only solution where everyone is happy. Let me try.

3ch023: I am fine with any solution you pick, its just the current one looked 'hacked' to me. and i wonder if there is a better way.
I've read @npeltier new suggestion a few times and didn't understand it still

3ch023 (approved):
works for me.
if/when we move sidenav block to Milo we probably should move 'sidenav-checkbox-group-title' creation to sidenav block. and keep `merch-sidenav-checkbox-group` more simple

i think its about generat expectation that web component will create shadow dom, but an outer actor creates the light dom children

libs/features/mas/src/global.css.js
```diff
@@ -432,5 +432,17 @@ body.merch-modal {
...
+    font-size: 14px;
+    font-style: normal;
+    font-weight: 700;
+    height: 32px;
+    letter-spacing: 0px;
+    padding: 0px;
+    line-height: 18.2px;
+    color: var(--color-gray-600);
+    margin: 0px;
+}
```
npeltier, line 445:
```suggestion
merch-sidenav-checkbox-group h3 {
    font-size: 14px;
    height: 32px;
    letter-spacing: 0px;
    line-height: 18.2px;
    color: var(--color-gray-600);
    margin: 0px;
}
```

npeltier (approved):
thanks a lot @bozojovicic looks good to me!

yesil (approved):

PR #3532, opened 2025-01-23
Fix Flaky Keyboard Navigation Tests by Isolating Mobile and Desktop Viewport Settings

skholkhojaev:
### Discription
resolved an issue where the `setViewPort()` function in the beforeEach block of the keyboard navigation tests caused inconsistent test results. whenever the correct viewport for mobile was set, the tests would pass successfully. However, when the viewport failed to set correctly, the tests would fail, and the await `sendKey()` function would not be handled properly.

### Changes
- isolated mobile keyboard navigation tests from desktop tests to prevent viewport conflicts 

**Test URLs:**
- After: https://flaky-keyboard-test--milo--skholkhojaev.aem.page/?martech=off


test/blocks/global-navigation/keyboard/keyboard-mobile.test.js
```diff
@@ -0,0 +1,343 @@
...
+const isClosed = (element) => element.getAttribute('aria-expanded') === 'false'
+  && element.hasAttribute('daa-lh', 'header|Open');
+const getPopup = (element) => element.parentElement.querySelector(selectors.popup);
+const getNavLinks = (trigger) => [...getPopup(trigger).querySelectorAll(`${selectors.navLink}, ${selectors.promoLink}, ${selectors.imagePromo}`)];
+let mainNavItems;
+// eslint-disable-next-line no-unused-vars
+let otherNavItems;
+let keyboardNavigation;
+// eslint-disable-next-line no-unused-vars
+let allNavItems;
```
mokimo, line 20: If those variables are not used, why don't you delete them instead of disabling the lint rule?

test/blocks/global-navigation/keyboard/keyboard-mobile.test.js
```diff
@@ -0,0 +1,343 @@
+/* eslint-disable no-restricted-syntax */
```
mokimo, line 1: Is this something you can fix, rather than adding a new rule?

test/blocks/global-navigation/keyboard/keyboard-mobile.test.js
```diff
@@ -0,0 +1,343 @@
...
+import { expect } from '@esm-bundle/chai/esm/chai.js';
+// eslint-disable-next-line import/extensions,import/no-relative-packages
```
mokimo, line 3: Same here. Relative imports directly from the node modules is a big anti-pattern, the linter correctly highlights this

mokimo (changes requested):
Instead of adding rules to avoid the linting errors, let's fix the errors that the linter hints at.

mokimo (approved):

test/blocks/global-navigation/keyboard/keyboard-mobile.test.js
```diff
@@ -0,0 +1,248 @@
...
+      loadStyles('blocks/global-navigation/features/search/gnav-search.css'),
+      loadStyles('blocks/global-navigation/features/profile/dropdown.css'),
+      loadStyles('blocks/global-navigation/utilities/menu/menu.css'),
+    ]);
+  });
+  let navLinks;
+  let trigger;
+  let triggerTwo;
+  let firstPopupItem;
+  beforeEach(async () => {
```
overmyheadandbody, line 35: OCD speaking here, but we usually have some empty lines between variables/method calls. So before `beforeEach`, before a `describe`, before an `it`, etc. That was also done like that in the initial file. This helps with readability when you just want to quickly scan the code and isolate areas that you might be interested in. It's like a book without paragraphs or chapters, it makes it tiring to read 😀

overmyheadandbody (approved):

test/blocks/global-navigation/keyboard/keyboard-mobile.test.js
skholkhojaev: Oh im sorry 😅 i will make sure to pay attention to these things 🙏
thankyou for the review!

skholkhojaev:
@overmyheadandbody should i still create a Jira ticket for this?

mokimo:
@skholkhojaev if its not done yet, it should be within a ticket 😁 The board should always reflect our current _in progress_ work status

PR #3536, opened 2025-01-23
[MWPW-158773] : Aside notification CLS performance improvements 

sharmeeatadobe:
The aside notification component considers the width:auto for the lazy load image. This causes the text to shift right by the width of the loaded image and hence degrades CLS.

Resolves: [MWPW-158773](https://jira.corp.adobe.com/browse/MWPW-158773)

**Test URLs:**
- Before: https://main--bacom--adobecom.hlx.page/fragments/homepage/summit/summit-2025-product-page-promoid-5wyl7trb?martech=off
- After: https://main--bacom--adobecom.hlx.page/fragments/homepage/summit/summit-2025-product-page-promoid-5wyl7trb?milolibs=mwpw-158773--milo--sharmeeatadobe&martech=off

- Before: https://main--milo--adobecom.hlx.page/drafts/sharmeeb/aside-notification?martech=off
- After: https://mwpw-158773--milo--sharmeeatadobe.hlx.page/drafts/sharmeeb/aside-notification?martech=off


libs/blocks/aside/aside.js
```diff
@@ -194,6 +194,8 @@ function decorateLayout(el) {
     const iconClass = iconVariant ? `${iconVariant[1]}-area` : 'icon-area';
     if (iconVariant) loadIconography();
     iconArea.classList.add(iconClass);
+    const image = iconArea.querySelector('img');
+    iconArea.style.aspectRatio = image.width / image.height;
```
mokimo, line 198: Out of curiosity, why does adding an aspectRatio reserve space compared to not adding it? 
I'd understand setting an explicit height and width 🤔

mokimo (approved):

robert-bogos (approved):

libs/blocks/aside/aside.js
sharmeeatadobe: Good question! Think responsiveness. When we set a specific height and width, we don't achieve good responsiveness. Instead adding an aspect ratio is going to take care of it on its own.

robert-bogos:
For the failing NALA tests you could try to clear the cache with this:
`CURL -si -X POST https://admin.hlx.page/code/sharmeeatadobe/milo/mwpw-158773/*`

overmyheadandbody (changes requested):
This might need another look, I see that the image is now in a different position than before. I'm guessing this isn't intentional? The paragraph may have different spacing rules that could differ from / add to the image's ones
Screenshot 2025-01-27 at 15 38 24
Screenshot 2025-01-27 at 15 38 35

sharmeeatadobe:
> This might need another look, I see that the image is now in a different position than before. I'm guessing this isn't intentional? The paragraph may have different spacing rules that could differ from / add to the image's ones Screenshot 2025-01-27 at 15 38 24 Screenshot 2025-01-27 at 15 38 35

This is a specific issue due to the image asset itself because the ratio of dimensions of the default image within <picture> tag is not matching the dimensions of the image that gets satisfied.
Whereas I did not find this issue where the ratios are maintained. Tried with few examples here: Eg: https://main--milo--adobecom.hlx.page/drafts/sharmeeb/aside-notification vs https://mwpw-158773--milo--sharmeeatadobe.hlx.page/drafts/sharmeeb/aside-notification

sharmeeatadobe:
> For the failing NALA tests you could try to clear the cache with this: `CURL -si -X POST https://admin.hlx.page/code/sharmeeatadobe/milo/mwpw-158773/\*`

I have tried this a couple of times but cache refresh did not help overcome the failed NALA test. Can we please re-run this since the tests are failing for a an unrelated component wrt this PR

overmyheadandbody:
@sharmeeatadobe - we can't ask authors to start changing such images, especially since we don't have a way to figure out all the pages where this might occur. Backwards compatibility needs to be maintained, without affecting layout. The tablet view would be even more affected with this change:
Screenshot 2025-01-28 at 14 52 53

We don't want to introduce visual regressions in favor of minor performance enhancements, other solutions should be considered to address this.

CC: @SilviuLCF

sharmeeatadobe:
> @sharmeeatadobe - we can't ask authors to start changing such images, especially since we don't have a way to figure out all the pages where this might occur. Backwards compatibility needs to be maintained, without affecting layout. The tablet view would be even more affected with this change: Screenshot 2025-01-28 at 14 52 53
> 
> We don't want to introduce visual regressions in favor of minor performance enhancements, other solutions should be considered to address this.
> 
> CC: @SilviuLCF

Sure, I'll try with some alternative solution. Do you have any suggestions in mind?

sharmeeatadobe:
> > @sharmeeatadobe - we can't ask authors to start changing such images, especially since we don't have a way to figure out all the pages where this might occur. Backwards compatibility needs to be maintained, without affecting layout. The tablet view would be even more affected with this change: Screenshot 2025-01-28 at 14 52 53
> > We don't want to introduce visual regressions in favor of minor performance enhancements, other solutions should be considered to address this.
> > CC: @SilviuLCF
> 
> Sure, I'll try with some alternative solution. Do you have any suggestions in mind?

Have implemented to hide the banner until the image loads. Please review again @overmyheadandbody @mokimo @robert-bogos

libs/blocks/aside/aside.js
```diff
@@ -194,6 +194,17 @@ function decorateLayout(el) {
     const iconClass = iconVariant ? `${iconVariant[1]}-area` : 'icon-area';
     if (iconVariant) loadIconography();
     iconArea.classList.add(iconClass);
+    const image = iconArea.querySelector('img');
+    if (image) {
```
zagi25, line 198: `visible` is default value for `visibility` so there is no need for first nested `if`
```
    if (image && !image.complete) {
       el.style.visibility = 'hidden';
       image.addEventListener('load', () => {
       el.style.visibility = 'visible';
      });
    }
```

zagi25:
@sharmeeatadobe along with cache clearing command, merging stage into your branch can help with NALA.

JasonHowellSlavin (commented):
Is there a test that can be updated to reflect this change in the code?

libs/blocks/aside/aside.js
```diff
@@ -194,6 +194,17 @@ function decorateLayout(el) {
...
+      if (image.complete) {
+        el.style.visibility = 'visible';
+      } else {
+        el.style.visibility = 'hidden';
+        image.addEventListener('load', () => {
+          el.style.visibility = 'visible';
```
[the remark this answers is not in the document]
sharmeeatadobe: Actually it is not necessary to have visibility. It'll work with "display" as well. made the change in the latest commit. 
From the CLS aspect, your suggestion to hide the entire aside banner until the banner loads works because if the text was never rendered or added to the DOM, there is no possibility for a layout shift once the image also loads. Hence, no CLS.

sharmeeatadobe:
> Is there a test that can be updated to reflect this change in the code?

Added test cases in latest commit

mokimo (approved):

libs/blocks/aside/aside.js
```diff
@@ -170,6 +170,15 @@ function loadIconography() {
   return new Promise((resolve) => { loadStyle(`${base}/styles/iconography.css`, resolve); });
 }
 
+export function handleImageLoad(el, image) {
+  if (image && !image.complete) {
+    el.style.visibility = 'hidden';
+    image.addEventListener('load', () => {
```
zagi25, line 176: one thing I forgot, shouldn't there be a fallback if for some reason image is not loaded ? Are we okay with aside staying hidden if there is some problem with an image ?
https://developer.mozilla.org/en-US/docs/Web/API/Window/error_event

sharmeeatadobe: Thanks for identifying, have added the code changes now.

zagi25 (approved):

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-158773

JasonHowellSlavin (approved):

overmyheadandbody (approved):

PR #3549, opened 2025-01-27
Mwpw 155418 Remove dependency on gist.github.com for gist.test.js

skholkhojaev:
### Description

This PR removes the dependency on gist.github.com in `gist.test.js`. Previously the test was reliant on external servers, which could lead to flaky tests and slower execution times.
Now the test now mocks the script creation process, to simulate the gist data locally without doing network calls.

### Changes
- Added sinon for stubbing document.createElement to mock script elements.
- Added `beforeEach` and `afterEach` hooks to handle stubbing and restoring.

Resolves: [MWPW-155418](https://jira.corp.adobe.com/browse/MWPW-155418)

**Test URLs:**
- After: https://MWPW-155418-remove-dependency-gist-github--milo--skholkhojaev.aem.page/?martech=off


test/blocks/gist/gist.test.js
```diff
@@ -6,10 +7,43 @@ document.body.innerHTML = await readFile({ path: './mocks/body.html' });
...
 describe('adobetv autoblock', () => {
-  it('creates iframe video block', async () => {
+  let createElementStub;
+  let originalCreateElement;
+
+  beforeEach(() => {
+    originalCreateElement = document.createElement;
+
+    createElementStub = sinon.stub(document, 'createElement').callsFake((tagName) => {
+      if (tagName === 'script') {
```
mokimo, line 17: Earlier returns are preferred
```suggestion
      if (tagName !=='script') return
```

test/blocks/gist/gist.test.js
```diff
@@ -6,10 +7,43 @@ document.body.innerHTML = await readFile({ path: './mocks/body.html' });
...
+        const scriptMock = originalCreateElement.call(document, 'script');
+        Object.defineProperty(scriptMock, 'src', {
+          set(url) {
+            const callbackMatch = url.match(/callback=([^&]+)/);
+            if (callbackMatch) {
```
mokimo, line 22: Let's be more explicit about which callback we expect here, you could also match this based on the gistlink you put in
```suggestion
            const gistCb = url.match(/callback=([^&]+)/);
            if (!gistCb) return
```

test/blocks/gist/gist.test.js
```diff
@@ -6,10 +7,43 @@ document.body.innerHTML = await readFile({ path: './mocks/body.html' });
...
+              const callbackName = callbackMatch[1];
+              Promise.resolve().then(() => {
+                if (window[callbackName]) {
```
mokimo, line 25: Why is this executed within a Promise resolve? should you rather use `clock.runAllAsync` if you awiat for a cb to execute?

test/blocks/gist/gist.test.js
```diff
@@ -6,10 +7,46 @@ document.body.innerHTML = await readFile({ path: './mocks/body.html' });
...
+      if (tagName !== 'script') {
+        return originalCreateElement.call(document, tagName);
+      }
+      const scriptMock = originalCreateElement.call(document, 'script');
+      Object.defineProperty(scriptMock, 'src', {
+        set(url) {
+          const gistCb = url.match(/callback=([^&]+)/);
+          if (!gistCb) return;
+          const callbackName = gistCb[1];
+          setTimeout(() => {
```
mokimo, line 28: Similar comment compared to the Promise one
> Why is this executed within a Promise resolve? should you rather use clock.runAllAsync if you awiat for a cb to execute?

See https://sinonjs.org/releases/latest/fake-timers/

test/blocks/gist/gist.test.js
```diff
@@ -6,10 +7,46 @@ document.body.innerHTML = await readFile({ path: './mocks/body.html' });
...
+    createElementStub = sinon.stub(document, 'createElement').callsFake((tagName) => {
+      if (tagName !== 'script') {
+        return originalCreateElement.call(document, tagName);
+      }
+      const scriptMock = originalCreateElement.call(document, 'script');
+      Object.defineProperty(scriptMock, 'src', {
+        set(url) {
+          const gistCb = url.match(/callback=([^&]+)/);
+          if (!gistCb) return;
+          const callbackName = gistCb[1];
```
mokimo, line 27:
```suggestion
          const gistCb = url.match(/callback=([^&]+)/)?.[1];
          if (!gistCb) return;
```

mokimo (approved):

robert-bogos (approved):

narcis-radu (approved):

PR #3553, opened 2025-01-27
MWPW-164658: Fix aside icon button

zagi25:
Resolves: [MWPW-164658](https://jira.corp.adobe.com/browse/MWPW-164658)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ratko/mwpw-164658/aside?martech=off
- After: https://mwpw-164658-aside-button-icon--milo--adobecom.hlx.page/drafts/ratko/mwpw-164658/aside?martech=off
- Before: https://main--milo--adobecom.hlx.page/drafts/klee/test-aside-button-issue?martech=off
- After: https://mwpw-164658-aside-button-icon--milo--adobecom.hlx.page/drafts/klee/test-aside-button-issue?martech=off


DKos95 (approved):

mokimo (approved):

mokimo:
The PS icon is not aligned anymore after the changes
Screenshot 2025-01-30 at 11 06 57
Screenshot 2025-01-30 at 11 06 52

libs/blocks/aside/aside.css
```diff
@@ -1236,3 +1236,10 @@
     flex: 0 0 calc(500px - 10.5vw);
   }
 }
+
+.aside .foreground.container .action-area a.icon-area.con-button,
+.aside.split.bio .foreground.container .text .action-area a.icon-area.con-button {
+  display: inline-block;
+  height: auto;
+  margin-bottom: 0;
```
overmyheadandbody, line 1244: This seems to make the button taller in comparison
Screenshot 2025-01-30 at 11 41 32

Could we not just cancel the `margin-bottom` and the `gap`?

zagi25: I spent some time on this, height difference doesn't occur because of this changes but because of svg icon. 
https://main--milo--adobecom.hlx.page/drafts/ratko/mwpw-164658/test
There is a difference in height for smaller and bigger buttons, middle one seems fine. 
I tried to fix it: https://ratko-icon-button--milo--adobecom.hlx.page/drafts/ratko/mwpw-164658/test
https://github.com/adobecom/milo/compare/ratko-icon-button?expand=1 
This also happens with other svgs, I tried the `add` one. I guess this needs a separate ticket because all icon buttons need to be checked. 

I tried to reverse all problematic rules back to those that we get from `.con-button` class, `inline-block` also cancels `gap` but I can swap it for `gap: 0` if you prefer it better.

libs/blocks/aside/aside.css
```diff
@@ -171,7 +171,7 @@
   align-items: center;
 }
 
-.aside:not(.notification) .foreground.container .icon-area {
+.aside:not(.notification):not(.promobar) .foreground.container .icon-area:not(.con-button) {
```
overmyheadandbody, line 174: We can group classes to exclude and this also keeps the specificity the same, while having two `:not` pseudo-classes would increase specificity.
```suggestion
.aside:not(.notification, .promobar) .foreground.container .icon-area:not(.con-button) {
```

libs/blocks/aside/aside.css
```diff
@@ -1236,3 +1236,10 @@
     flex: 0 0 calc(500px - 10.5vw);
   }
 }
+
+.aside .foreground.container .action-area a.icon-area.con-button,
+.aside.split.bio .foreground.container .text .action-area a.icon-area.con-button {
+  display: inline-block;
```
overmyheadandbody, line 1242: With the changes from L174, I think the `display` declaration might not be needed anymore, did you find use-cases where it's still relevant?

libs/blocks/aside/aside.css
overmyheadandbody: I still see that before the button was 29px, now it's 31.5px in height, but I get your point, we should open a follow-up story to look into standardizing the button heights.

libs/blocks/aside/aside.css
zagi25: Screenshot 2025-02-03 at 06 29 59

There is still this rule that get's overridden by `display: inline-block` (5th block on the test page)

robert-bogos (commented):
You could try to clear the cache for the failing NALA tests:
`CURL -si -X POST https://admin.hlx.page/code/adobecom/milo/mwpw-164658-aside-button-icon/\*`

zagi25:
@robert-bogos it seems that merging the stage branch fixed the issue. Thank you !

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-164658

PR #3562, opened 2025-01-28
[MWPW-163470] Migrate Search in DA

maagrawal16:
* Added Milo Studio : Search build for DA.
* Now IMS token and tenant getting fetched from session storage.

Resolves: [MWPW-163470](https://jira.corp.adobe.com/browse/MWPW-163470)

**Test URLs:**
- After: https://da.live/apps (A tile for Milo Studio Advance Search will be added in DA using the build)


nkthakur48:
@maagrawal16 Please add entry point HTMLs (search.html) along with auth.js to facilitate communication with exec shell and bundled react app

tools/search/ms-search.js
```diff
@@ -0,0 +1,31 @@
...
+      <div style='display: flex; align-items: center; justify-content: center; height: 100vh; flex-direction: column;'>
+        <h2>Please sign in to continue</h2>
+      </div>
+    `;
+    document.body.style.visibility = 'visible';
+    return;
+  }
+
+  let fullURL = window.location.protocol + '//' + window.location.hostname + (window.location.port ? ':' + window.location.port : '') + window.location.pathname + '#/search';
+  window.history.pushState({}, null, fullURL);
```
nkthakur48, line 28: Simple syntax could have been `window.location.replace(fullURL);`

tools/search/ms-search.js
```diff
@@ -0,0 +1,31 @@
...
+    document.body.innerHTML = `
+      <div style='display: flex; align-items: center; justify-content: center; height: 100vh; flex-direction: column;'>
+        <h2>Please sign in to continue</h2>
+      </div>
+    `;
+    document.body.style.visibility = 'visible';
+    return;
+  }
+
+  let fullURL = window.location.protocol + '//' + window.location.hostname + (window.location.port ? ':' + window.location.port : '') + window.location.pathname + '#/search';
```
nkthakur48, line 27: I suppose if `fullUrl` is not getting updated, you could use `const`

nkthakur48 (changes requested):
@maagrawal16 Could you please have a look at review comments?

tools/search/ms-search.js
maagrawal16: That would not work since this reloads the page and sets the token and tenant in the session storage again which would then not exists since we are removing the query params for the same. I can persist it in the session storage but I'll have to change some logic in the authentication. Should I go ahead with that?

tools/search/ms-search.js
maagrawal16: That's right, I'll change that.

tools/search/ms-search.js
nkthakur48: That should be fine, you could skip this comment then!

saurabhsircar11 (approved):

nkthakur48 (approved):

PR #3566, opened 2025-01-28
[MWPW-166631] Tab color contrast

DKos95:
This resolves a11y contrast problems for tabs.

Resolves: [MWPW-166631](https://jira.corp.adobe.com/browse/MWPW-166631)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/docs/library/kitchen-sink/tab-block?martech=off
- After: https://tab-color-contrast--milo--adobecom.aem.page/docs/library/kitchen-sink/tab-block?martech=off

### DIFF
--- a/libs/blocks/tabs/tabs.css
+++ b/libs/blocks/tabs/tabs.css
@@ -1,16 +1,25 @@
 :root {
   /* Tab Colors */
   --tabs-active-color: #1473e6;
-  --tabs-border-color: #d8d8d8;
+  --tabs-border-color: #909090;
   --tabs-border-hover-color: #acacac;
-  --tabs-text-color: #7e7e7e;
+  --tabs-text-color: #686868;
   --tabs-active-text-color: #2C2C2C;
   --tabs-bg-color: #f1f1f1;
   --tabs-active-bg-color: #fff;
-  --tabs-list-bg-gradient: linear-gradient(rgb(255 255 255 / 0%) 60%, rgb(245 245 245 / 80%));
-  --tabs-pill-bg-color: #e1e1e1;
+  --tabs-list-bg-gradient: linear-gradient(
+    to bottom, 
+    rgba(136, 136, 136, 0) 0%,
+    rgba(136, 136, 136, 0) 60%,
+    rgba(136, 136, 136, 0.03) 80%,
+    rgba(136, 136, 136, 0.08) 100%
+  );
+  --tabs-pill-bg-color: #242424;
   --tabs-pill-bg-color-hover: #dadada;
   --tabs-pill-bg-color-active: #c6c6c6;
+  --tabs-pill-selected-text-color: #ffffff;
+  --tabs-mobile-stacked-bg-color: #242424;
+  --tabs-mobile-stacked-text-color: #ffffff;
   --tabs-pill-text-color: #131313;
   --tabs-paddle-bs-color: #0000001a;
   --tabs-radio-gray: #6D6D6D;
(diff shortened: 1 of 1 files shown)


libs/blocks/tabs/tabs.css
```diff
@@ -232,6 +271,14 @@
   margin-inline-start: 24px;
 }
 
+.tabs.pill .tabList button{
+  border: none;
+}
+
+.tabs.quiet .tabList {
+  padding-bottom: 1px;
```
robert-bogos, line 279: I see a lot of very specific values, are they based on some specs from Consonant?

DKos95: Yes, this specifically was a request to add bottom padding so the hover and selected underline would not overlap with the border lines. Also we received some feedback yesterday so there will be more changes here.

libs/blocks/tabs/tabs.css
```diff
@@ -20,16 +28,23 @@
...
 :root .dark {
-  --tabs-border-color: rgb(56 56 56);
-  --tabs-text-color: #cdcdcd;
+  --tabs-border-color: #909090;
+  --tabs-text-color: #B6B6B6;
   --tabs-active-text-color: #fff;
   --tabs-bg-color: #1a1a1a;
-  --tabs-active-bg-color: #111;
-  --tabs-list-bg-gradient: linear-gradient(rgb(0 0 0 / 0%) 60%, rgb(0 0 0 / 80%));
+  --tabs-active-bg-color: #1E1E1E;
```
overmyheadandbody, line 35: Small nit: I see most other colors are lowercase

libs/blocks/tabs/tabs.css
```diff
@@ -539,11 +595,13 @@
...
   .tabs.stacked-mobile div[role="tablist"] button[aria-selected="true"] {
-    background-color: #dadada;
+    background-color: var(--tabs-mobile-stacked-bg-color);
+    color: var(--tabs-mobile-stacked-text-color);
   }
   
   .tabs.stacked-mobile.dark div[role="tablist"] button[aria-selected="true"] {
-    background-color: #393939;
+    background-color: var(--tabs-mobile-stacked-bg-color);
+    color: var(--tabs-mobile-stacked-text-color);
```
overmyheadandbody, line 604: These now look the same to me, could they be consolidated?

libs/blocks/tabs/tabs.css
```diff
@@ -49,49 +64,68 @@
   padding: var(--spacing-xxl) 0;
 }
 
-.tabs.xxl-spacing .paddle {
+.tabs:not(.quiet).xxl-spacing .paddle {
   top: var(--spacing-xxl);
 }
 
+.tabs.quiet.xxl-spacing .paddle {
+  top: calc(var(--spacing-xxl) + 1px);
```
overmyheadandbody, line 72: Is there no way around the `+ 1px` repetition? Could we add a `1px` top margin to `.tabs.quiet`. Could we translate them by `1px`? I feel like we may be able to improve this a bit.

libs/blocks/tabs/tabs.css
```diff
@@ -49,39 +64,59 @@
   padding: var(--spacing-xxl) 0;
 }
 
-.tabs.xxl-spacing .paddle {
+.tabs:not(.quiet).xxl-spacing .paddle {
+  top: var(--spacing-xxl);
+}
+
+.tabs.quiet.xxl-spacing .paddle {
   top: var(--spacing-xxl);
```
overmyheadandbody, line 72: Do we need the repetition, since the same declaration applies to both quiet and non-quiet tabs? Same question applies to all spacing tokens. If there is a need for repetition, let's at least consolidate the selectors, since they have the same declaration.

libs/blocks/tabs/tabs.css
```diff
@@ -232,6 +273,14 @@
   margin-inline-start: 24px;
 }
 
+.tabs.pill .tabList button{
```
overmyheadandbody, line 276:
```suggestion
.tabs.pill .tabList button {
```

libs/blocks/tabs/tabs.css
```diff
@@ -538,12 +598,10 @@
     margin-bottom: 0;
   }
 
-  .tabs.stacked-mobile div[role="tablist"] button[aria-selected="true"] {
-    background-color: #dadada;
-  }
-  
+  .tabs.stacked-mobile div[role="tablist"] button[aria-selected="true"],
   .tabs.stacked-mobile.dark div[role="tablist"] button[aria-selected="true"] {
```
overmyheadandbody, line 602: Can't these 2 selectors be consolidated? Wouldn't the first one also apply to the `.dark` variant or are there specificity issues?

libs/blocks/tabs/tabs.css
```diff
@@ -331,6 +380,9 @@
   cursor: default;
   box-shadow: none;
   background: var(--tabs-list-bg-gradient);
+  background-size: 100% 16px;
+  background-position: bottom;
+  background-repeat: no-repeat;
```
overmyheadandbody, line 385: A similar declaration is present on L138-140, do we need to repeat it here?

DKos95: L138-140 is for tablist div, this is for paddles.

overmyheadandbody (approved):
Thanks for integrating the changes!

mokimo (approved):

zagi25 (approved):

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-166631

PR #3568, opened 2025-01-28
[MWPW-156760] Table caret position

DKos95:
This changes the caret for table from "Title +/-" to ">/⌄ Title"

Resolves: [MWPW-156760](https://jira.corp.adobe.com/browse/MWPW-156760)

| Before       | After        |
|--------------|--------------|
|Screenshot 2025-01-28 at 16 29 46|Screenshot 2025-01-28 at 16 29 39|

**Test URLs:**
- After: https://table-caret-position--milo--adobecom.aem.page/docs/library/kitchen-sink/table?martech=off

### DIFF
--- a/libs/blocks/table/table.css
+++ b/libs/blocks/table/table.css
@@ -292,7 +292,7 @@
 .table .section-head .section-head-title > :not(.icon) {
   font-size: var(--type-body-m-size);
   line-height: var(--type-heading-s-lh);
-  width: calc(100% - 20px);
+  width: calc(100% - 30px);
 }
 
 .table .section-head .section-head-title,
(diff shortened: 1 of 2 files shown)


libs/blocks/table/table.css
```diff
@@ -354,20 +354,26 @@
...
   border: 0;
-  background-image: url('../../ui/img/expand2.svg');
+  background-image: url('../../ui/img/chevron-wide-black.svg');
   background-repeat: no-repeat;
-  background-position: center right;
-  width: 15px;
-  height: 15px;
+  background-position: center;
+  width: 12px;
+  height: 17px;
```
robert-bogos, line 361: Where are these specific values coming from? 🤔

DKos95: We have a chevron icon in the project already and without uploading a new one I tried to make it look as close to the design as possible since the original is a bit too big. If your concern here is the focus box is not even we can get the height to be 12px also, but the icon when tabbing is not being focused but the cell in the table.

narcis-radu (approved):

SilviuLCF (approved):
Verified, testing details https://jira.corp.adobe.com/browse/MWPW-156760

PR #3591, opened 2025-01-30
Move federal code to utils

mokimo:
### Description
The due the growing usage of the federal project and thus the `federated.js` file - we should move the code to the milo core utils as it has became a core feature.

This should provide performance benefits as this can currently lead to being a render blocking request where certain key features such as `decorate.js` require to load `federated.js` before the LCP.

Screenshot 2025-01-30 at 12 26 42

### Implications
We seem to have one deep import [adobecom/dme-partners](https://github.com/adobecom/dme-partners) · [edsdme/blocks/partners-navigation/utilities/utilities.js](https://github.com/adobecom/dme-partners/blob/17bb35bfd126882f9254857a3e00f09c82fe4365/edsdme/blocks/partners-navigation/utilities/utilities.js#L10) based on a [quick search](https://github.com/search?q=org%3Aadobecom%20federated.js&type=code)...

We will need to keep the `federated.js` file around to avoid breaking changes.

Resolves: [MWPW-166847](https://jira.corp.adobe.com/browse/MWPW-166847)

**Test URLs:**
- After: https://move-federal-code--milo--mokimo.aem.page/?martech=off


libs/utils/utils.js
```diff
@@ -261,6 +261,64 @@ export const [setConfig, updateConfig, getConfig] = (() => {
   ];
 })();
 
+let federatedContentRoot;
```
mokimo, line 264: Since we are very protective of utils, here's the rationale:

`federated.js` is required for the LCP in a lot of cases due to being used in `decorate.js`

On top of this, it has wide usage across multiple blocks that are being used for LCP.

Anything that we absolutely need for the 90% LCP use-case and critical to rendering the page is justified to live in the utils in my opinion. On top of that, wide usage across blocks makes this a perfect candidate to live here.

libs/utils/federated.js
```diff
@@ -1,59 +1,3 @@
-import { getConfig, SLD } from './utils.js';
```
overmyheadandbody, line 1: Do we have knowledge of anyone importing this file?

libs/utils/utils.js
overmyheadandbody: This does sound reasonable, but do you have any data to support that this is the better approach performance-wise?

On another note, would we need to move the unit tests in the utils test file?

overmyheadandbody (approved):
A couple of curiosities

libs/utils/federated.js
mokimo: Yes :( [here](https://github.com/adobecom/dme-partners/blob/17bb35bfd126882f9254857a3e00f09c82fe4365/edsdme/blocks/partners-navigation/utilities/utilities.js#L10)

libs/utils/utils.js
mokimo: We saw a 400ms LCP improvement when we merged https://github.com/adobecom/milo/pull/2752#issue-2470375361 which was a combination of parallelize 2 requests & get rid of one. This should net us ~200ms for all devices.

For the unit tests, since we still have the original file, for now we can keep them where they are IMO 🤔 
The unit test-tests are full enough anyway so the separation is not bad

vhargrave (approved):
looks very good and clever execution on the deprecated `federated.js` file.
@mokimo Could you potentially create a ticket for the Yugo team though to stop importing this deprecated file so we can properly delete it? 
Here's one of their tickets that I found if you want to clone - https://jira.corp.adobe.com/browse/MWPW-165786

libs/utils/federated.js
```diff
@@ -1,59 +1,3 @@
...
-      ...locale,
-      contentRoot: `${libOrigin}${locale.prefix}/federal/globalnav`,
-    },
-    placeholders,
-  };
-
-  return fedsPlaceholderConfig;
-};
+import { getFederatedContentRoot, getFederatedUrl, getFedsPlaceholderConfig } from './utils.js';
+// DO NOT USE THIS FILE, THIS IS HERE FOR HISTORY PURPOSES AND AVOIDING BREAKING CHANGES.
```
sharmrj, line 2: It might be worth having a `console.warn()` here to let any stray users know that they should no longer be importing this file. Maybe we could put a date on removing it entirely.

sharmrj (approved):

robert-bogos (approved):

libs/utils/federated.js
mokimo: I thought about it, however we only have one consumer which I created a ticket for. I dislike console.warn since it would end up on the end-users device as well. Not much harm if its only when someone checks the dev-tools, but nothing that we should have on prod either though

PR #3593, opened 2025-01-30
[MWPW-160510] tooltip enabled on all icons

DKos95:
This solution allows use of tooltips for all icons.

Resolves: [MWPW-160510](https://jira.corp.adobe.com/browse/MWPW-160510)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/dusan/icon-tooltip-enhancement?martech=off
- After: https://icon-tooltip-enhancement--milo--adobecom.hlx.page/drafts/dusan/icon-tooltip-enhancement?martech=off

### DIFF
--- a/libs/features/icons/icons.js
+++ b/libs/features/icons/icons.js
@@ -81,7 +81,7 @@ export default async function loadIcons(icons, config) {
     const existingIcon = icon.querySelector('svg');
     if (!iconSVGs[iconName] || existingIcon) return;
     const parent = icon.parentElement;
-    if (parent.childNodes.length > 1) {
+    if (parent?.childNodes.length > 1) {
       if (parent.lastChild === icon) {
         icon.classList.add('margin-inline-start');
       } else if (parent.firstChild === icon) {
--- a/libs/features/icons/icons.js
+++ b/libs/features/icons/icons.js
@@ -34,7 +34,17 @@ export const fetchIcons = (config) => new Promise(async (resolve) => {
   resolve(fetchedIcons);
 });
 
-function decorateToolTip(icon) {
+function checkIfIconIsTooltip(icon) {
+  const wrapper = icon.closest('em');
+  if (!wrapper) return null;
+
+  const conf = wrapper.textContent.includes('|');
+  if (!conf) return null;
+
+  return [...icon.classList].some((cls) => cls.includes('tooltip'));
+}
+
+function decorateToolTip(icon, iconName) {
   const wrapper = icon.closest('em');
   wrapper.className = 'tooltip-wrapper';
   if (!wrapper) return;
@@ -45,7 +55,7 @@ function decorateToolTip(icon) {
   icon.dataset.tooltip = content;
   // Position is the next to last part of a tooltip
   const place = conf.pop()?.trim().toLowerCase() || 'right';
-  icon.className = `icon icon-info milo-tooltip ${place}`;
+  icon.className = `icon icon-${iconName} milo-tooltip ${place}`;
   icon.setAttribute('tabindex', '0');
   icon.setAttribute('aria-label', content);
   icon.setAttribute('role', 'button');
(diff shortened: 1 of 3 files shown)


libs/features/icons/icons.js
```diff
@@ -34,7 +34,17 @@ export const fetchIcons = (config) => new Promise(async (resolve) => {
...
-function decorateToolTip(icon) {
+function checkIfIconIsTooltip(icon) {
+  const wrapper = icon.closest('em');
+  if (!wrapper) return null;
+
+  const conf = wrapper.textContent.includes('|');
+  if (!conf) return null;
+
+  return [...icon.classList].some((cls) => cls.includes('tooltip'));
+}
```
mokimo, line 45: You can shorten this a bit, in fact if you shorten this, you can probably get rid of this helper function
```suggestion
function checkIfIconIsTooltip(icon) {
  const hasAuthoredTooltipText = icon.closest('em')?.textContent.includes('|');
  return hasAuthoredTooltipText && [...icon.classList].some((cls) => cls.includes('tooltip'));
}
```

libs/features/icons/icons.js
```diff
@@ -34,7 +34,17 @@ export const fetchIcons = (config) => new Promise(async (resolve) => {
...
+
+function decorateToolTip(icon, iconName) {
```
mokimo, line 47: If this is a one-liner that isn't re-used, I'd not make it a helper function and move it closer to where its actual used.

```suggestion
function decorateToolTip(icon, iconName) {
    const hasTooltip = icon.closest('em')?.textContent.includes('|') && [...icon.classList].some((cls) => cls.includes('tooltip'))
    if(!hasTooltip) return
```

libs/features/icons/icons.js
```diff
@@ -55,13 +65,22 @@ export default async function loadIcons(icons, config) {
...
+
+    if (iconNameInitial === 'tooltip') {
+      iconName = 'info';
+    } else if (iconNameInitial.includes('tooltip-')) {
+      iconName = iconNameInitial.replace(/tooltip-/, '');
+    } else {
+      iconName = iconNameInitial;
+    }
+
+    if (checkIfIconIsTooltip(icon)) decorateToolTip(icon, iconName);
```
mokimo, line 79: I'd put the `if` condition within the function where its used as that encapsulates the whole logic a bit better

robert-bogos (approved):
Looks good 👍

libs/features/icons/icons.js
```diff
@@ -55,13 +65,22 @@ export default async function loadIcons(icons, config) {
...
+    const iconNameInitial = icon.classList[1].replace('icon-', '');
+    let iconName;
+
+    if (iconNameInitial === 'tooltip') {
+      iconName = 'info';
+    } else if (iconNameInitial.includes('tooltip-')) {
+      iconName = iconNameInitial.replace(/tooltip-/, '');
+    } else {
+      iconName = iconNameInitial;
+    }
```
mokimo, line 77: I'm not a big fan of if else conditions tbh, I'd rather go for something such as this:
```suggestion
    let iconName = iconNameInitial === 'tooltip' ? 'info' : 'iconNameInitial
    if (iconNameInitial.includes('tooltip-') iconName = iconNameInitial.replace(/tooltip-/, '');
```

mokimo: Should we have 1-2 more unit tests to also capture the different branches? The icon name seems to be quite important, so we should write unit tests to ensure all those conditions do what we want them to

mokimo (commented):
Small remark

mokimo (approved):

mokimo:
To @SilviuLCF or @NadiiaSokolova let's ensure we also test those icons in a few blocks, to be sure it's all well and compatible - since icons can be used anyhwere.

NadiiaSokolova (approved):
Verified.
Testing details https://jira.corp.adobe.com/browse/MWPW-160510

PR #3594, opened 2025-01-30
[MWPW-166755] table tab navigation stuck fix

DKos95:
This resolves the issue where navigating through the table with Tab causes the navigation to be stuck when getting to cells that have a caret for toggling rows.

Resolves: [MWPW-166755](https://jira.corp.adobe.com/browse/MWPW-166755)

**Test URLs:**
- After: https://table-keyboard-nav-stuck--milo--adobecom.hlx.page/docs/library/kitchen-sink/table?martech=off

### DIFF
--- a/libs/blocks/table/table.js
+++ b/libs/blocks/table/table.js
@@ -427,9 +427,8 @@ function applyStylesBasedOnScreenSize(table, originTable) {
       icon.parentElement.addEventListener('click', () => handleExpand(icon));
       icon.parentElement.setAttribute('tabindex', 0);
       icon.parentElement.addEventListener('keydown', (e) => {
-        if (e.key !== 'Tab') {
-          e.preventDefault();
-        }
+        if (e.key !== 'Tab') e.preventDefault();
+
         if (e.key === 'Enter' || e.key === ' ') handleExpand(icon);
       });
     });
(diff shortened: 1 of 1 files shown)


mokimo (approved):

libs/blocks/table/table.js
```diff
@@ -427,7 +427,9 @@ function applyStylesBasedOnScreenSize(table, originTable) {
       icon.parentElement.addEventListener('click', () => handleExpand(icon));
       icon.parentElement.setAttribute('tabindex', 0);
       icon.parentElement.addEventListener('keydown', (e) => {
-        e.preventDefault();
+        if (e.key !== 'Tab') {
+          e.preventDefault();
+        }
```
robert-bogos, line 432: nit:
```suggestion
        if (e.key !== 'Tab') e.preventDefault();
```

robert-bogos (approved):

libs/blocks/table/table.js
```diff
@@ -427,7 +427,8 @@ function applyStylesBasedOnScreenSize(table, originTable) {
       icon.parentElement.addEventListener('click', () => handleExpand(icon));
       icon.parentElement.setAttribute('tabindex', 0);
       icon.parentElement.addEventListener('keydown', (e) => {
-        e.preventDefault();
+        if (e.key !== 'Tab') e.preventDefault();
```
overmyheadandbody, line 430: This looks fine, but I'm not sure I understand the initial intent. Why do we need to prevent default? Is is just because of the fact that a Space press could scroll the page down? In that case, wouldn't it be better to prevent the default only for that key?

overmyheadandbody (approved):
Great, thanks! This makes more sense to me 😀

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-166755

PR #3599, opened 2025-01-30
MWPW-149470 [MILO][MEP] Remove sendTargetResponseAnalytics function

vgoodric:
* Remove all of the AA tracking for how long Target response takes
* Add in lana log tracking for when Target took longer than the page's Target timeout setting

Resolves: [MWPW-149470](https://jira.corp.adobe.com/browse/MWPW-149470)

**Test URLs:**
- After: https://mepremovetracking--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -1163,38 +1163,6 @@ async function updateManifestsAndPropositions(
   return targetManifests;
 }
 
-function roundToQuarter(num) {
-  return Math.ceil(num / 250) / 4;
-}
-
-function calculateResponseTime(responseStart) {
-  const responseTime = Date.now() - responseStart;
-  return roundToQuarter(responseTime);
-}
-
-function sendTargetResponseAnalytics(failure, responseStart, timeoutLocal, message) {
-  // temporary solution until we can decide on a better timeout value
-  const responseTime = calculateResponseTime(responseStart);
-  const timeoutTime = roundToQuarter(timeoutLocal);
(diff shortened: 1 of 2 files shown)


denlight (approved):

libs/features/personalization/personalization.js
```diff
@@ -1216,19 +1184,17 @@ const handleAlloyResponse = (response) => ((response.propositions || response.de
...
 async function handleMartechTargetInteraction(
-  { config, targetInteractionPromise, calculatedTimeout },
+  { config, targetInteractionPromise },
 ) {
   let targetManifests = [];
   let targetPropositions = [];
   if (config?.mep?.enablePersV2 && targetInteractionPromise) {
-    const { targetInteractionData, respTime, respStartTime } = await targetInteractionPromise;
-    sendTargetResponseAnalytics(false, respStartTime, calculatedTimeout);
+    const { targetInteractionData, respTime } = await targetInteractionPromise;
```
swamu, line 1192: Can we remove the `respStartTime` from the return value of  targetInteractionPromise in utils?
Since it is used at only this page.

swamu (changes requested):

libs/features/personalization/personalization.js
vgoodric: Sorry, maybe I'm misunderstanding the comment. I am using it on line 1197. So 5 lines later

swamu: @vgoodric, the targetInteractionPromise is written in utils.js and returns the respStartTime.
 Since this value - `respStartTime` is no longer needed, we can remove it from the return value in utils.js too.

PR #3601, opened 2025-01-31
[MWPW-166884] - PEP Bug fix

Deva309:
Clear Timeout issue fix for the scenario when pause on hover feature is off.

Resolves: [MWPW-166884](https://jira.corp.adobe.com/browse/MWPW-166884)

**Test URLs:**
- After: https://feds--milo--adobecom.aem.page/?martech=off

QA:

- https://feds--milo--adobecom.hlx.page/drafts/devashish/pep/pep-prompt?skipPepEntitlements=true (without pause on hover)
- https://feds--milo--adobecom.hlx.page/drafts/devashish/pep/pause/pep-prompt-pause-animation?skipPepEntitlements=true (with pause on hover)

### DIFF
--- a/libs/features/webapp-prompt/webapp-prompt.js
+++ b/libs/features/webapp-prompt/webapp-prompt.js
@@ -108,7 +108,6 @@ export class AppPrompt {
     this.entName = entName;
     this.parent = parent;
     this.getAnchorState = getAnchorState;
-    this.timeoutId = null;
     this.id = this.promptPath.split('/').pop();
     this.elements = {};
     if (modalsActive()) {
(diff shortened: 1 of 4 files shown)


sonawanesnehal3 (approved):

bandana147 (approved):

spadmasa:
Validation done on the below url after the fix :
https://www.stage.adobe.com/acrobat.html?skipPepEntitlements=true&milolibs=feds user is not seen redirected after optout . Details are in https://jira.corp.adobe.com/browse/MWPW-166884

libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -282,20 +283,19 @@ export class AppPrompt {
...
-      timeoutId = setTimeout(() => {
+      this.timeoutId = setTimeout(() => {
         this.close({ saveDismissal: false, dismissalActions: false });
         AppPrompt.redirectTo(this.options['redirect-url']);
       }, remainingTime);
     };
 
     const stopTimeout = () => {
-      clearTimeout(timeoutId);
+      clearTimeout(this.timeoutId);
```
sharmrj, line 298: Instead of having some state for the timeout Id, let's have `initRedirect` return a function that cleans up both the timeout and removes the event listeners for `mouseenter` and `mouseleave`. Then we can simply store that in `this.cancelRedirect` (instead of the somewhat confusing `this.redirectFn` on line 159 or so) and call that in the `close` method. It'll make what's happening in `close` slightly more readable and do a better job of cleaning up.

sharmrj (commented):
Make sure you add a unit test case to check for this problem.

sharmrj (approved):

libs/features/webapp-prompt/webapp-prompt.js
Deva309: Done

PR #3608, opened 2025-01-31
quiz-entry block removal

colloyd:
* deletes the quiz-entry block

Resolves: [MWPW-166952](https://jira.corp.adobe.com/browse/MWPW-166952)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/colloyd/quiz-entry/?martech=off
- After: https://quiz-entry-removal--milo--colloyd.hlx.page/drafts/colloyd/quiz-entry/?martech=off


fullcolorcoder (approved):

narcis-radu (commented):
I would suggest removing the block from [CODEOWNERS](https://github.com/adobecom/milo/blob/stage/CODEOWNERS#L44) as well. Do we need to remove any documentation? @elaineskpt

JackySun9 (approved):

colloyd:
> I would suggest removing the block from [CODEOWNERS](https://github.com/adobecom/milo/blob/stage/CODEOWNERS#L44) as well. Do we need to remove any documentation? @elaineskpt

Good call on codeowners. As for docs, we shouldn't completely remove any work done - I've moved all of this code to a feature branch before submitting this PR. This block may make a return in the future, but we don't know when, so @elaineskpt if we could pull published docs, but still preserve the work, that would be ideal.

colloyd:
@narcis-radu This thing is failing checks due to nala tests that are completely unrelated (table block accessibility). I've removed the ready for stage label so it doesn't continue to try, but can we please get this manually merged? 

This sort of test failing on unrelated stuff is become a very regular experience. Obviously it'd be good to address the root cause, but I wonder if a "manual merge request" tag would be useful? something to indicate that auto merging is not working, can we please look at a manual merge?

colloyd:
@skumar09 Can you please look into what's going on with these nala test failures? They are in no way related to the actual changes on this PR (they seem to be regarding accessibility issues with the table component?), but they're keeping it from auto merging with Ready for Stage.

JackySun9:
I think we can manual merge this and ignore the failed tests

PR #3613, opened 2025-02-03
[MWPW-152453] Overriding description elements for accordion

robert-bogos:
## Description
This PR is replacing all the `<dl>`, `<dt> `and `<dd>` elements, within the accordion block, with `<div>`. 
This will help with consistency across Acom pages and is an a11y best practice.

## Related Issue
Resolves: [MWPW-152453](https://jira.corp.adobe.com/browse/MWPW-152453)

## Testing instructions
1. Open the test URL.
2. Check the accordion block html structure, it should not have any `<dl>`, `<dt> `and `<dd>`.

## Screenshots:

**Before:**
Screenshot 2025-02-03 at 16 12 29

**After:**
Screenshot 2025-02-03 at 16 13 56

## Test URLs
- Before: https://main--cc--adobecom.hlx.page/products/photoshop-lightroom/plans
- After: https://main--cc--adobecom.hlx.page/products/photoshop-lightroom/plans?milolibs=mwpw-152453-a11y--milo--robert-bogos

PSI check:  https://mwpw-152453-a11y--robert-bogos--adobecom.hlx.page/drafts/rbogos/page-default?martech=off


overmyheadandbody (commented):
I see some NALA tests are now failing, we should address those too, it's probably looking for some specific markup.

Otherwise this looks good, maybe we can attach a quick screenshot of Axe with Best Practices enables to show the issue has disappeared.

libs/blocks/accordion/accordion.css
```diff
@@ -2,28 +2,28 @@
...
-.accordion dd {
+.accordion .descr-details {
   margin: 0;
   padding: var(--spacing-xs);
   font-size: var(--type-body-s-size);
   line-height: var(--type-body-s-lh);
 }
 
-.accordion dt button {
+.accordion .descr-term button {
```
vhargrave, line 26: @robert-bogos if there's no hTag on line 90 of the js file, then the .descr-term class won't get added and this css won't take affect, You'll probably want to fix that.

vhargrave: looks like this - 
image

robert-bogos: Good catch! I fixed that 👍

vhargrave:
@robert-bogos I think you might need to update the unit tests as well 
image

robert-bogos:
> @robert-bogos I think you might need to update the unit tests as well image

Currently looking into fixing those. I should've run the unit tests locally before opening the PR 🤦‍♂️

libs/blocks/accordion/accordion.js
```diff
@@ -87,10 +87,10 @@ function createItem(accordion, id, heading, num, edit) {
 
   const para = panel?.querySelector('p');
   const text = para ? para.textContent : panel?.textContent;
-  const dtAttrs = hTag ? {} : { role: 'heading', 'aria-level': 3 };
+  const dtAttrs = hTag ? { class: 'descr-term' } : { role: 'heading', 'aria-level': 3, class: 'descr-term' };
   const dtHtml = hTag ? createTag(hTag.tagName, { class: 'accordion-heading' }, button) : button;
-  const dt = createTag('dt', dtAttrs, dtHtml);
-  const dd = createTag('dd', { 'aria-labelledby': triggerId, id: panelId, hidden: true }, panel);
+  const dt = createTag('div', dtAttrs, dtHtml);
+  const dd = createTag('div', { 'aria-labelledby': triggerId, id: panelId, hidden: true, class: 'descr-details' }, panel);
```
mokimo, line 93: The naming `dt&dd` are now confusing when those are divs

test/blocks/accordion/accordion.test.js
```diff
@@ -31,7 +31,7 @@ describe('Accordion', () => {
     expect(script).to.exist;
 
     // handleClick()
-    const firstAccordionButton = document.body.querySelector('dt button');
+    const firstAccordionButton = document.body.querySelector('.descr-term button');
```
mokimo, line 34: I've seen consumers rely on milo specific markup before :( 

This _might_ be a breaking change if any consumer relies on that

mokimo (approved):

libs/blocks/accordion/accordion.js
robert-bogos: I went with this to minimize the code changes

test/blocks/accordion/accordion.test.js
robert-bogos: If so, I think this will be caught by unit tests or QA on the consumer side.

vhargrave (approved):

zagi25 (approved):

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-152453

PR #3618, opened 2025-02-04
[MWPW-167031] top tooltip rtl fix

DKos95:
This resolves the issue where the top tooltip was not properly displayed for right-to-left languages.

Resolves: [MWPW-167031](https://jira.corp.adobe.com/browse/MWPW-167031)

**Test URLs:**
- After: https://tooltip-rtl-position--milo--adobecom.aem.page/docs/library/kitchen-sink/table?martech=off

### DIFF
--- a/libs/blocks/table/table.css
+++ b/libs/blocks/table/table.css
@@ -402,7 +402,7 @@
   margin-left: 16px;
 }
 
-[dir="rtl"] .section-row-title .milo-tooltip:is(.top, .bottom, .left, .right)::before {
+[dir="rtl"] .section-row-title .milo-tooltip:is(.bottom, .left, .right)::before {
   left: initial;
   margin: 0 8px 0 0;
   right: 100%;
@@ -417,20 +417,6 @@
   transform: translateY(-50%);
   border: 8px solid transparent;
   border-left-color: #0469E3;
-}
-
-[dir="rtl"] .section-row-title .milo-tooltip.top::before {
-  left: 50%;
-  top: -8px;
-  right: unset;
-  margin: 0;
-  transform: translateX(-50%) translateY(-100%);
-}
-
-[dir="rtl"] .section-row-title .milo-tooltip.top::after {
-  left: 8px;
-  top: -8px;
-  transform: rotate(90deg);
 }
 
 /* hover */
@@ -739,11 +725,6 @@
     left: 100%;
     margin: 0 0 0 -4px;
     right: initial;
-  }
-
-  [dir="rtl"] .section-row-title .milo-tooltip.top::after {
-    transform: unset;
-    top: 4px;
   }
 
   .table .row-heading .col-heading .pricing {
--- a/libs/blocks/table/table.css
+++ b/libs/blocks/table/table.css
@@ -420,10 +420,11 @@
 }
 
 [dir="rtl"] .section-row-title .milo-tooltip.top::before {
-  left: unset;
-  top: -30px;
+  left: 50%;
+  top: -8px;
   right: unset;
   margin: 0;
+  transform: translateX(-50%) translateY(-100%);
 }
 
 [dir="rtl"] .section-row-title .milo-tooltip.top::after {
@@ -730,6 +731,7 @@
 
   [dir="rtl"] .section-row-title .milo-tooltip.top::before {
     top: 50%;
+    transform: translateY(-50%);
   }
 
   [dir="rtl"] .section-row-title .milo-tooltip:is(.top, .bottom, .left, .right)::after {
--- a/libs/blocks/table/table.css
+++ b/libs/blocks/table/table.css
@@ -402,23 +402,6 @@
   margin-left: 16px;
 }
 
-[dir="rtl"] .section-row-title .milo-tooltip:is(.bottom, .left, .right)::before {
-  left: initial;
-  margin: 0 8px 0 0;
-  right: 100%;
-  top: 50%;
-  transform: translateY(-50%);
-}
-
-[dir="rtl"] .section-row-title .milo-tooltip:is(.top, .bottom, .left, .right)::after {
-  left: -4px;
-  margin: 0 4px 0 -4px;
-  top: 50%;
-  transform: translateY(-50%);
-  border: 8px solid transparent;
-  border-left-color: #0469E3;
-}
-
 /* hover */
 @media (min-width: 900px) {
   .table .col.hover {
@@ -690,41 +673,6 @@
     grid-row: 1;
     grid-column: 1 / x;
     background-color: var(--color-gray-100);
-  }
-
-  .section-row-title .milo-tooltip:is(.top, .bottom, .left, .right)::before {
-    left: initial;
-    margin: 0 8px 0 0;
-    right: 100%;
-    top: 50%;
-    transform: translateY(-50%);
-  }
-
-  .section-row-title .milo-tooltip:is(.top, .bottom, .left, .right)::after {
-    left: -4px;
-    margin: 0 4px 0 -4px;
-    top: 50%;
-    transform: translateY(-50%);
-    border: 8px solid #0469E3;
-    border-color: transparent transparent transparent #0469E3;
-  }
-
-  [dir="rtl"] .section-row-title .milo-tooltip:is(.top, .bottom, .left, .right)::before {
-    left: 100%;
-    margin: 0 0 0 12px;
-    right: initial;
-  }
-
-  [dir="rtl"] .section-row-title .milo-tooltip.top::before {
-    top: 50%;
-    transform: translateY(-50%);
-  }
-
-  [dir="rtl"] .section-row-title .milo-tooltip:is(.top, .bottom, .left, .right)::after {
-    border-color: transparent #0469E3 transparent transparent;
-    left: 100%;
-    margin: 0 0 0 -4px;
-    right: initial;
   }
 
   .table .row-heading .col-heading .pricing {
--- a/libs/blocks/table/table.css
+++ b/libs/blocks/table/table.css
@@ -419,6 +419,19 @@
   border-left-color: #0469E3;
 }
 
+[dir="rtl"] .section-row-title .milo-tooltip.top::before {
+  left: unset;
+  top: -30px;
+  right: unset;
+  margin: 0;
+}
+
+[dir="rtl"] .section-row-title .milo-tooltip.top::after {
+  left: 8px;
+  top: -8px;
+  transform: rotate(90deg);
+}
+
 /* hover */
 @media (min-width: 900px) {
   .table .col.hover {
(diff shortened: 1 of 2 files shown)


libs/blocks/table/table.css
```diff
@@ -423,6 +423,19 @@
   border-left-color: #0469E3;
 }
 
+[dir="rtl"] .section-row-title .milo-tooltip.top::before {
+  left: unset;
+  top: -30px;
```
zagi25, line 428: This doesn't work if tooltip text is longer 
Screenshot 2025-02-04 at 12 05 23 (2)
Screenshot 2025-02-04 at 12 05 13 (2)

vhargrave (approved):
looking good!

robert-bogos (approved):
Neat 👍

zagi25 (approved):

libs/blocks/table/table.js
```diff
@@ -153,6 +153,32 @@ function handleAddOnContent(table) {
...
 
+function setTooltipPosition(el) {
+  const tooltips = el.querySelectorAll('.milo-tooltip');
+
+  tooltips.forEach((tooltip) => {
+    const classesToCheck = ['top', 'bottom', 'left', 'right'];
+    const defaultClass = classesToCheck.find((className) => tooltip.classList.contains(className)) || 'right';
+    tooltip.setAttribute('data-tooltip-default-position', defaultClass);
+
+    tooltip.addEventListener('mouseenter', () => {
```
overmyheadandbody, line 164: What about `focus`, since the tooltip is shown for that event as well?

libs/blocks/table/table.js
```diff
@@ -153,6 +153,32 @@ function handleAddOnContent(table) {
...
+      const isTabletOrMobile = ['TABLET', 'MOBILE'].includes(defineDeviceByScreenSize());
+      const isRtl = document.documentElement.dir === 'rtl';
+
+      if (!((el.getBoundingClientRect().width === window.innerWidth) || isTabletOrMobile)) {
```
overmyheadandbody, line 168: Are we actually concerned with the width of the element? Or would it be its left/right boundaries and making sure they don't exceed the viewport? If it's the former, couldn't we use `el.clientWidth` or `el.scrollWidth`? `getBoundingClientRect` does create a bit of a performance overhead and if it's easy enough to avoid, we should. If it's the latter, we should update the logic to check for left/right boundaries.

libs/blocks/table/table.js
```diff
@@ -153,6 +153,32 @@ function handleAddOnContent(table) {
...
+        tooltip.classList.remove(...classesToCheck);
+        tooltip.classList.add(defaultClass);
+        return;
+      }
+
+      const newClass = isTabletOrMobile === isRtl ? 'right' : 'left';
```
overmyheadandbody, line 174: Wouldn't an `&&` be better than `===` here? They're almost the same except if both are `false`. So on desktop with RTL, we'd be moving the tooltip in the opposite direction.

libs/blocks/table/table.js
```diff
@@ -153,6 +153,32 @@ function handleAddOnContent(table) {
...
+
+      tooltip.classList.remove(...classesToCheck);
+      tooltip.classList.add(newClass);
+    });
```
overmyheadandbody, line 178: A small optimization idea: we're repeating `.remove` and `.add`. What if we define a constant holding the class we want added and then remove the classes and add the relevant one at the end. Would that work?

libs/blocks/table/table.js
```diff
@@ -153,6 +153,32 @@ function handleAddOnContent(table) {
   table.addEventListener('mas:resolved', debounce(() => { handleEqualHeight(table, '.row-heading'); }));
 }
 
+function setTooltipPosition(el) {
+  const tooltips = el.querySelectorAll('.milo-tooltip');
+
+  tooltips.forEach((tooltip) => {
+    const classesToCheck = ['top', 'bottom', 'left', 'right'];
+    const defaultClass = classesToCheck.find((className) => tooltip.classList.contains(className)) || 'right';
+    tooltip.setAttribute('data-tooltip-default-position', defaultClass);
```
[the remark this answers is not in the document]
DKos95: This was a part of a previous solution I was working on, will be removed

robert-bogos (commented):
Hi @DKos95 ! Please mark as resolved the conversations from Rares, which are implemented/resolved. It will be easier to review for others 👍

DKos95:
> Hi @DKos95 ! Please mark as resolved the conversations from Rares, which are implemented/resolved. It will be easier to review for others 👍

Thanks for the tip, will definitely do that for future PR's.

Regarding this PR, I think that [this ticket](https://jira.corp.adobe.com/browse/MWPW-155282) has made this kind of overengineering, since it seems that with the tooltip width now reduced for Desktop it never seems to overflow, so a solution would only be needed for Mobile and Tablet which would be simpler.

vhargrave (commented):
The tooltip looks much smaller with these changes. I think that some styles aren't getting applied properly anymore. 

**Before:**
image

**After**
image

I don't know the complexity of this task either , but I'd also say that a pure CSS solution seems like the right way to go for performance and simplicity, which is also what you originally did. So I'm mainly just curious why you've now opted for the JS approach?

vhargrave (changes requested):
see comment

DKos95:
> The tooltip looks much smaller with these changes. I think that some styles aren't getting applied properly anymore.
> 
> **Before:** image
> 
> **After** image
> 
> I don't know the complexity of this task either , but I'd also say that a pure CSS solution seems like the right way to go for performance and simplicity, which is also what you originally did. So I'm mainly just curious why you've now opted for the JS approach?

This is from another ticket already on stage: https://jira.corp.adobe.com/browse/MWPW-155282

vhargrave:
> > The tooltip looks much smaller with these changes. I think that some styles aren't getting applied properly anymore.
> > **Before:** image
> > **After** image
> > I don't know the complexity of this task either , but I'd also say that a pure CSS solution seems like the right way to go for performance and simplicity, which is also what you originally did. So I'm mainly just curious why you've now opted for the JS approach?
> 
> This is from another ticket already on stage: https://jira.corp.adobe.com/browse/MWPW-155282

Got it, thanks! 
@DKos95 one more thing that I'm noticing is that the default positioning of the tooltip on rtl pages is now covering the text. 
I imagine that shouldn't be the case right? 
image

DKos95:
> > > The tooltip looks much smaller with these changes. I think that some styles aren't getting applied properly anymore.
> > > **Before:** image
> > > **After** image
> > > I don't know the complexity of this task either , but I'd also say that a pure CSS solution seems like the right way to go for performance and simplicity, which is also what you originally did. So I'm mainly just curious why you've now opted for the JS approach?
> > 
> > 
> > This is from another ticket already on stage: https://jira.corp.adobe.com/browse/MWPW-155282
> 
> Got it, thanks! @DKos95 one more thing that I'm noticing is that the default positioning of the tooltip on rtl pages is now covering the text. I imagine that shouldn't be the case right? image

Hmm, from what I can see on main branch, the right position was applied also to this situation but because of the size of the tooltip it went to left because it had no space to go right, now that the width is smaller it has space so it goes right where the author wants it.

Screenshot 2025-02-17 at 10 54 42

vhargrave:
> > > > The tooltip looks much smaller with these changes. I think that some styles aren't getting applied properly anymore.
> > > > **Before:** image
> > > > **After** image
> > > > I don't know the complexity of this task either , but I'd also say that a pure CSS solution seems like the right way to go for performance and simplicity, which is also what you originally did. So I'm mainly just curious why you've now opted for the JS approach?
> > > 
> > > 
> > > This is from another ticket already on stage: https://jira.corp.adobe.com/browse/MWPW-155282
> > 
> > 
> > Got it, thanks! @DKos95 one more thing that I'm noticing is that the default positioning of the tooltip on rtl pages is now covering the text. I imagine that shouldn't be the case right? image
> 
> Hmm, from what I can see on main branch, the right position was applied also to this situation but because of the size of the tooltip it went to left because it had no space to go right, now that the width is smaller it has space so it goes right where the author wants it.
> 
> Screenshot 2025-02-17 at 10 54 42

@DKos95 Shouldn't it go left by default on rtl settings if no position has been set? 
I looked on stage and it seems to be right there.

DKos95:
> > > > > The tooltip looks much smaller with these changes. I think that some styles aren't getting applied properly anymore.
> > > > > **Before:** image
> > > > > **After** image
> > > > > I don't know the complexity of this task either , but I'd also say that a pure CSS solution seems like the right way to go for performance and simplicity, which is also what you originally did. So I'm mainly just curious why you've now opted for the JS approach?
> > > > 
> > > > 
> > > > This is from another ticket already on stage: https://jira.corp.adobe.com/browse/MWPW-155282
> > > 
> > > 
> > > Got it, thanks! @DKos95 one more thing that I'm noticing is that the default positioning of the tooltip on rtl pages is now covering the text. I imagine that shouldn't be the case right? image
> > 
> > 
> > Hmm, from what I can see on main branch, the right position was applied also to this situation but because of the size of the tooltip it went to left because it had no space to go right, now that the width is smaller it has space so it goes right where the author wants it.
> > Screenshot 2025-02-17 at 10 54 42
> 
> @DKos95 Shouldn't it go left by default on rtl settings if no position has been set? I looked on stage and it seems to be right there.

From what I can see it has the class right by default, but it goes left because of these old css rules for that resolution which were probably implemented because if it went right it would overflow, which now is not an issue, in any case I will be syncing these concerns with Rares tomorrow.

libs/blocks/table/table.js
```diff
@@ -153,6 +153,58 @@ function handleAddOnContent(table) {
   table.addEventListener('mas:resolved', debounce(() => { handleEqualHeight(table, '.row-heading'); }));
 }
 
+function setTooltipPosition(el) {
+  const tooltips = el.querySelectorAll('.milo-tooltip');
+
+  tooltips.forEach((tooltip) => {
+    const classesToCheck = ['top', 'bottom', 'left', 'right'];
+    const defaultClass = classesToCheck.find((className) => tooltip.classList.contains(className)) || 'right';
+    let isTabletOrMobile = ['TABLET', 'MOBILE'].includes(defineDeviceByScreenSize());
```
narcis-radu, line 162: I have a suggestion here to avoid updating the value in multiple places.
```js
const isTabletOrMobile = () => ['TABLET', 'MOBILE'].includes(defineDeviceByScreenSize());
```
Then you can use it everywhere as `isTabletOrMobile()`

overmyheadandbody (approved):
Great to see we no longer have to duplicate the tooltip styles from the icons file! I noticed that we can once again trigger the tooltip by hovering around the icon, but without triggering the focus state. Didn't we fix this at one point?
Screenshot 2025-02-19 at 11 59 50

Not a big issue and not related to the changes from this PR, but we might want to look into this in the future.

mokimo (approved):

robert-bogos (approved):

PR #3622, opened 2025-02-04
Servicenow cmr id update

amauch-adobe:
In order to help flakiness in the ServiceNow CMR Action Workflow, I have added a backoff function to the CMR ID Retrieval GET request that can sometimes return an unknown status while the queue for CMRs is being cleared out.  This backoff will continue to attempt the GET call in increasing intervals for either 2 minutes or up to 15 attempts, which ever comes first. 

This was per the recommendation from the iPAAS API team for Change Management requests to help with performance issues.

Resolves: [MWPW-167087](https://jira.corp.adobe.com/browse/MWPW-167087)

**Test URLs:**
- After: https://servicenow-cmr-id-update--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/.github/workflows/servicenow.py
+++ b/.github/workflows/servicenow.py
@@ -74,7 +74,7 @@ if __name__ == "__main__":
     print(response.text)
     sys.exit(1)
   else:
-    print("IMS token request was successful")
+    print("IMS token request was successful: ", response.status_code)
     token = jsonParse["access_token"]
 
   print("Create CMR in ServiceNow...")
@@ -114,7 +114,8 @@ if __name__ == "__main__":
     print(response.text)
     sys.exit(1)
   else:
-    print("CMR creation was successful")
+    print("CMR creation was successful: ", response.status_code)
+    print(response.text)
     transaction_id = jsonParse["id"]
 
   print("Waiting for Transaction from Queue to ServiceNow then Retrieve CMR ID...")
@@ -140,7 +141,8 @@ if __name__ == "__main__":
     print(response.text)
     sys.exit(1)
   else:
-    print("CMR ID retrieval was successful")
+    print("CMR ID retrieval was successful: ", response.status_code)
+    print(response.text)
     cmr_id = jsonParse["result"]["changeId"]
 
   print("Setting Actual Maintenance Time Windows for CMR...")
@@ -175,4 +177,5 @@ if __name__ == "__main__":
     print(response.text)
     sys.exit(1)
   else:
-    print("CMR closure was successful")
+    print("CMR closure was successful: ", response.status_code)
+    print(response.text)
--- a/.github/workflows/servicenow.py
+++ b/.github/workflows/servicenow.py
@@ -1,7 +1,7 @@
 import requests
 import time
 import datetime
-import timedelta
+import random
 import json
 import os
 import sys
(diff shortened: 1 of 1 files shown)


.github/workflows/servicenow.py
```diff
@@ -35,6 +35,45 @@ def find_string_in_json(json_data, target_string):
...
+        try:
+            return operation()  # Attempt the operation
+        except Exception as e:
+            attempts += 1
+            if attempts > max_retries or (time.time() - start_time) >= timeout:
+                raise  # Re-raise the exception if max retries or timeout is reached
+
+            delay = min(base_delay * (2 ** (attempts - 1)), max_delay) + random.uniform(0, 0.1 * base_delay)
+            time.sleep(delay)
+    raise TimeoutError("Operation timed out after {} seconds or {} retries, whatever came first.".format(timeout, max_retries))
```
mokimo, line 51: Should we add a print per try? So the action has better logs on re-trying

.github/workflows/servicenow.py
```diff
@@ -35,6 +35,45 @@ def find_string_in_json(json_data, target_string):
...
+
+# Example usage
```
mokimo, line 53: Not sure what this example usage comment is supposed to mean or hint at 🤔

mokimo (approved):
On the last CSO, we struggled a bit finding the CMR - should we add a link to the created CMR so that one can get from the action to the CMR in one click?

amauch-adobe:
> On the last CSO, we struggled a bit finding the CMR - should we add a link to the created CMR so that one can get from the action to the CMR in one click?

I can pull it from the payload.  I'll add it.

.github/workflows/servicenow.py
amauch-adobe: forgot to remove it.

.github/workflows/servicenow.py
amauch-adobe: I'll add the print in just before the operation() method execution.

amauch-adobe:
> > On the last CSO, we struggled a bit finding the CMR - should we add a link to the created CMR so that one can get from the action to the CMR in one click?
> 
> The GET call to retrieve a CMR only provides the iPAAS API url for the change queue (which is unaccessible anyways). What I will do is print out the Change ID since it is in the form of "CHG001199977" which you then can search for in the Change Launchpad UI in ServiceNow by putting in the Change ID in the search box field up at the top of the window as shown below:
> 
> Screenshot 2025-02-04 at 10 00 22 AM

I have added this line at the end of successful runs to be able to find a change management record from a release more easily:

Screenshot 2025-02-04 at 10 30 02 AM

thedoc31:
Are the nala failures of any importance? I'm not sure what to make of them.

amauch-adobe:
@mokimo, there are quite of Nala errors that from my standpoint when looking at them some are from tests that may need to be updated because the text they check for is no longer what they expect (my guess an authoring content change), others, and frankly most or all of them have nothing to do with the changes I have made here.  I marked this PR as trivial and zero-impact because none of my changes will affect any current live site.  Are we able to move forward with the PR or do I need to reach out to the milo-automation channel to look into these failures to figure out what has caused Nala to fail on so many tests?

cc @thedoc31 .

mokimo:
FYI I ran `CURL -si -X POST https://admin.hlx.page/code/amauch-adobe/milo/servicenow-cmr-id-update/*` to purge stale hlx assets, nala tests are now green as they should be! WFs might not be mergeable by the process due to github security limitations - if that's the case I'll merge it manually

mokimo:
> Error merging 3622: Servicenow cmr id update Resource not accessible by integration

Yeah I feared so, well I can still merge it 😁

PR #3639, opened 2025-02-06
MWPW-157733: Fix checkmark with bold text

zagi25:
Resolves: [MWPW-157733](https://jira.corp.adobe.com/browse/MWPW-157733)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ratko/mwpw-157733/ul-checkmark?martech=off
- After: https://mwpw-157733-ul-checkmark--milo--adobecom.hlx.page/drafts/ratko/mwpw-157733/ul-checkmark?martech=off

### DIFF
--- a/libs/blocks/text/text.css
+++ b/libs/blocks/text/text.css
@@ -120,7 +120,7 @@
   align-items: center;
 }
 
-.text-block .icon-area {
+.text-block .icon-area:not(:has(span.icon)) {
   display: flex;
   column-gap: var(--spacing-xs);
 }


robert-bogos (approved):

libs/blocks/text/text.css
```diff
@@ -120,7 +120,7 @@
   align-items: center;
 }
 
-.text-block .icon-area {
+.text-block .icon-area:not(:has(span.icon)) {
```
overmyheadandbody, line 123: This does seem to work, but to me it sounds a bit odd that we're targeting an icon area without icons. Can we track down the original reasoning for why this was needed? I don't have a use-case where this wouldn't work in mind, but I'd suggest caution when QA-ing this.

SilviuLCF (approved):
Verified, testing details. https://jira.corp.adobe.com/browse/MWPW-157733

PR #3648, opened 2025-02-07
[MWPW-161562] Auto pause videos

overmyheadandbody:
This ensures that Sharepoint and MPC videos that aren't authored to be played automatically are paused when they are no longer in the viewport.

A few dev notes:

- although we already have a method to define intersection observers, it is currently set up to execute a callback only on intersection. For this requirement we need the exact opposite. I've fiddled around to augment the method from `utils.js`, but ultimately decided against it, since it created impractical overhead. We'd also have to call it twice under certain circumstances, since the `rootMargin` value is different;
- I've noticed that the `figure` block uses node cloning to decorate the content, I've created https://jira.corp.adobe.com/browse/MWPW-167455 to update this behavior

Resolves: [MWPW-161562](https://jira.corp.adobe.com/browse/MWPW-161562)

**Test URLs:**
- Before: https://main--milo--overmyheadandbody.aem.page/drafts/anupama/pr/carousel/carousel-videos1?martech=off
- After: https://auto-pause-videos--milo--overmyheadandbody.hlx.page/drafts/anupama/pr/carousel/carousel-videos1?martech=off

### DIFF
--- a/libs/blocks/adobetv/adobetv.js
+++ b/libs/blocks/adobetv/adobetv.js
@@ -1,4 +1,5 @@
 import { decorateAnchorVideo } from '../../utils/decorate.js';
+import { createTag } from '../../utils/utils.js';
 
 export default function init(a) {
   a.classList.add('hide-video');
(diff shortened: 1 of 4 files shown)


mokimo (approved):

SilviuLCF (approved):
Verified, testing details  https://jira.corp.adobe.com/browse/MWPW-161562

PR #3651, opened 2025-02-07
MWPW-164093: Adobe Home 'Pricing Widget' merch card

Axelcureno:
This PR introduces Adobe Home 'Pricing Widget' merch card with 2 variations:
Screenshot 2025-02-07 at 11 24 32 AM
Screenshot 2025-02-07 at 11 24 38 AM
Screenshot 2025-02-07 at 11 24 46 AM

Resolves: [MWPW-164093](https://jira.corp.adobe.com/browse/MWPW-164093)

M@S Studio Test URL: https://mwpw-164093--mas--adobecom.hlx.page/studio.html?milolibs=MWPW-164093--milo--adobecom#path=adobe-home&query=pricing
Test page: https://mwpw-164093--milo--adobecom.hlx.live/libs/features/mas/docs/adobe-home.html?martech=off

Test URLs:

Before: https://main--milo--adobecom.hlx.live/docs/library/kitchen-sink/merch-card?martech=off
After: https://mwpw-164093--milo--adobecom.hlx.live/docs/library/kitchen-sink/merch-card?martech=off


libs/features/mas/docs/checkout-button.html
yesil, line 1: I think this is not in the navigation menu(lost in previous PRs) can you add it back?

libs/features/mas/src/hydrate.js
```diff
@@ -175,16 +285,40 @@ function createSpectrumSwcButton(cta, aemFragmentMapping, isOutline, variant) {
...
     );
 
-    spectrumCta.addEventListener('click', (e) => {
-        if (e.target !== cta) {
-            /* c8 ignore next 3 */
+    (async () => {
+        try {
+          await customElements.whenDefined('checkout-button');
+          const CheckoutButtonEl = customElements.get('checkout-button');
+          const checkoutBtn = CheckoutButtonEl.createCheckoutButton({}, cta.innerHTML);
```
yesil, line 295: I think this block causes runtime errors in ah gallery:

mas.js:2370 Failed to initialize checkout-button logic: TypeError: Cannot read properties of null (reading 'setAttribute')
    at mas.js:2370:30403

libs/features/mas/src/hydrate.js
```diff
@@ -175,16 +285,40 @@ function createSpectrumSwcButton(cta, aemFragmentMapping, isOutline, variant) {
...
+    
+          for (const attr of cta.attributes) {
+            checkoutBtn.setAttribute(attr.name, attr.value);
```
yesil, line 298: I wonder why this is needed?

libs/features/mas/src/hydrate.js
```diff
@@ -175,16 +285,40 @@ function createSpectrumSwcButton(cta, aemFragmentMapping, isOutline, variant) {
...
-        cta,
+        cta.innerHTML,
     );
 
-    spectrumCta.addEventListener('click', (e) => {
-        if (e.target !== cta) {
-            /* c8 ignore next 3 */
+    (async () => {
+        try {
+          await customElements.whenDefined('checkout-button');
```
yesil, line 293: I think it is safe to assume that `checkout-button` is already defined before we hit here, if not wee need to check how mas.js is bundled.
I would avoid adding checks for web components that we know are in the bundle.

libs/features/mas/src/hydrate.js
```diff
@@ -175,16 +285,40 @@ function createSpectrumSwcButton(cta, aemFragmentMapping, isOutline, variant) {
...
+          if (typeof checkoutBtn.connectedCallback === 'function') {
+            checkoutBtn.connectedCallback();
+          }
+          await checkoutBtn.render();
+          await checkoutBtn.onceSettled();
+    
+          spectrumCta.addEventListener('click', (e) => {
             e.stopPropagation();
-            cta.click();
+            if (typeof checkoutBtn.clickHandler === 'function') {
```
yesil, line 309: isn't `checkoutBtn.click()` not enough?

libs/features/mas/src/match-media-controller.js
```diff
@@ -0,0 +1,34 @@
+import { LitElement } from 'lit';
+
+export default class MatchMediaController {
```
yesil, line 3: is this different than https://opensource.adobe.com/spectrum-web-components/tools/match-media/ ?

can we use the lit one instead?

libs/features/mas/src/merch-stock.js
```diff
@@ -1,6 +1,6 @@
 import { LitElement, css, html } from 'lit';
 import { EVENT_MERCH_STOCK_CHANGE } from './constants.js';
-import { MatchMediaController } from '@spectrum-web-components/reactive-controllers/src/MatchMedia.js';
+import MatchMediaController from './match-media-controller.js';
```
yesil, line 3: I would like to understand the difference

libs/features/mas/src/variants/ah-try-buy-widget.js
```diff
@@ -0,0 +1,176 @@
...
+
+    :host([variant='ah-try-buy-widget']) .footer {
+      display: flex;
+      width: fit-content;
+      flex-wrap: wrap;
+      gap: 8px;
+      flex-direction: row;
+    }
+
+    :host([variant='ah-try-buy-widget']) .footer .spectrum-Button-label {
```
yesil, line 138: if we don't support Spectrum CSS anymore for AH, we should remove such rules.

libs/features/mas/src/variants/ah-try-buy-widget.js
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    :host([variant='ah-try-buy-widget']) ::slotted(.spectrum-Button--secondary) {
+        background-color: var(--spectrum-secondary-color, #E5E5E5);
+        color: black;
+    }
+  `;
+}
+
+customElements.define('ah-try-buy-widget', AHTryBuyWidget);
```
yesil, line 176: please cleanup Spectrum CSS related rules since we switched to SWC.

yesil (changes requested):

libs/features/mas/src/match-media-controller.js
st-angelo-adobe: We talked about it a few weeks ago. You said if `MatchMediaController` from SWC is small we could use it, otherwise we can implement the logic in js. And this is it, the implementation in js. 
However, it's pretty much the same as the one from SWC to be honest. Let me know if you want me to remove this and just use the one from SWC.

libs/features/mas/src/merch-stock.js
st-angelo-adobe: See above

libs/features/mas/src/checkout-mixin.js
```diff
@@ -100,7 +100,6 @@ export function CheckoutMixin(Base) {
         }
 
         async render(overrides = {}) {
-            if (!this.isConnected) return false;
```
3ch023, line 103: is it to fix some issue? do you have more details?

3ch023:
noticed some console error in mas.js 
image

libs/features/mas/src/hydrate.js
```diff
@@ -106,38 +125,129 @@ export function processBackgroundImage(
...
-            fields.description,
-        );
-        merchCard.append(body);
+            attributes,
+            description,
+        ));
     }
 }
 
+export function getTruncatedTextData(text, limit, withSuffix = true) {
```
3ch023, line 183: could we hide extra text with styles instead of truncating description? https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow

libs/features/mas/src/merch-card-collection.js
```diff
@@ -1,6 +1,5 @@
 import { html, LitElement } from 'lit';
-import { MatchMediaController } from '@spectrum-web-components/reactive-controllers/src/MatchMedia.js';
-
+import MatchMediaController from './match-media-controller.js';
```
3ch023, line 2: we don't use collection for AH yet, right?
trying to understand what is this change for

3ch023 (approved):
Looks good!
added a few comments to fix/discuss.
Also, could you please add a simple nala test, similar to masccd? but masah?

libs/features/mas/src/variants/ah-try-buy-widget.js
st-angelo-adobe: I removed, hopefully all of them

libs/features/mas/src/match-media-controller.js
st-angelo-adobe: Actually since `MatchMediaController` from SWC is so small we can just use that. I changed it back everywhere too.

libs/features/mas/src/merch-card-collection.js
st-angelo-adobe: I needed the `MatchMediaController`, or similar login for the try buy card. I ran it past Ilyas as well and decided to just implement the logic in a local controller. Because I did that, I changed it everywhere as well.

But the thing is, what I implemented is pretty much the same as the SWC implementation, so I just scrapped my local controller, and went back to using it from SWC. So this is no longer a thing

libs/features/mas/src/hydrate.js
st-angelo-adobe: As mentioned yesterday in the call, we can't really do it with css only.

libs/features/mas/src/match-media-controller.js
yesil: true true, I thought this portion was from Axel and I didn't remember our discussion.
if putting `MatchMediaController` doesn't explode our bundle size, let's use it as it.

libs/features/mas/src/checkout-mixin.js
yesil: @3ch023 I asked them.  Since `sp-button` doesn't have checkout-link capabilities, instead we will render a checkout-link but not in the DOM and just forward clicks to it. 
But the DOM element must be rendered even if it is not connected to DOM which is fine IMO.

3ch023: i am good if there is no impact on events we throw for CCD

3ch023 (changes requested):
@Axelcureno @st-angelo-adobe there seem to be a regression
noticed a difference between
https://main--milo--adobecom.hlx.live/libs/features/mas/docs/merch-card.html
and 
https://mwpw-164093--milo--adobecom.hlx.live/libs/features/mas/docs/merch-card.html
merch-card is no longer marked as failed if checkout url is not resolved
image

cc @yesil

3ch023:
@Axelcureno another thing i noticed is that mas.js still contains process.env?.DEBUG, which mean AH will have their webpack issue.
pls update tacocat-core-1.13.0.tgz package to the new one with your fix.

libs/features/mas/src/hydrate.js
```diff
@@ -61,11 +63,21 @@ export function processSize(fields, merchCard, sizeConfig) {
 
 export function processTitle(fields, merchCard, titleConfig) {
     if (fields.cardTitle && titleConfig) {
+        const attributes = { slot: titleConfig.slot };
+        let title = fields.cardTitle;
+        const { maxCount } = titleConfig;
+        if (maxCount) {
+            const [truncatedTitle, cleanTitle] = getTruncatedTextData(fields.cardTitle, maxCount);
```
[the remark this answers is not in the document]
Axelcureno: Yes, created: https://jira.corp.adobe.com/browse/MWPW-167766

libs/features/mas/src/hydrate.js
Axelcureno: yes, fixed

Axelcureno:
> noticed some console error in mas.js image

this is now fixed @3ch023

Axelcureno:
> @Axelcureno @st-angelo-adobe there seem to be a regression noticed a difference between https://main--milo--adobecom.hlx.live/libs/features/mas/docs/merch-card.html and https://mwpw-164093--milo--adobecom.hlx.live/libs/features/mas/docs/merch-card.html merch-card is no longer marked as failed if checkout url is not resolved image
> 
> cc @yesil

I discovered this is caused by @bozojovicic's PR that was recently merged: https://github.com/bozojovicic/milo/blob/08b47405e1659883d82d28f1144e42d81a09d418/libs/features/mas/src/merch-card.js#L379

I removed the updateComplete from Promise.race(), not sure why was added but basically this means that if the fragment finishes reloading first it returns undefined instead of false, bypassing offer resolution errors. You can check it works now (it doesn't reproduce in main because Bozo's PR is in stage only atm).

Axelcureno:
> @Axelcureno another thing i noticed is that mas.js still contains process.env?.DEBUG, which mean AH will have their webpack issue. pls update tacocat-core-1.13.0.tgz package to the new one with your fix.

Thanks! Updated taco, it's now removed.

libs/features/mas/src/variants/ah-try-buy-widget.js
```diff
@@ -0,0 +1,160 @@
...
+  prices: { tag: 'p', slot: 'price' },
+  ctas: { slot: 'cta', size: 'S' },
+  backgroundImage: { tag: 'div', slot: 'image' },
+  backgroundColor: { attribute: 'background-color' },
+  borderColor: { attribute: 'border-color' },
+  allowedColors: ['gray'],
+  size: ['single', 'double', 'triple'],
+};
+
+export class AHTryBuyWidget extends VariantLayout {
```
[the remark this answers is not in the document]
3ch023: i would be also with current name, might get some actual widget later to implement?

3ch023 (approved):

3ch023 (commented):
Hi @Axelcureno , approving, but either you or QA would need to add nala tests for this.
Could be a follow up PR though

libs/features/mas/test/hydrate.test.js
```diff
@@ -514,3 +550,127 @@ describe('processDescription', async () => {
         );
     });
 });
+
+describe('getTruncatedTextData', () => {
+  it('closes any open tags in truncated text', () => {
```
yesil, line 555: can you add a test case for graceful handling in case an error happens inside `getTruncatedTextData`

libs/features/mas/src/hydrate.js
```diff
@@ -106,36 +136,127 @@ export function processBackgroundImage(
...
-            fields.description,
-        );
-        merchCard.append(body);
+            attributes,
+            description,
+        ));
+    }
+}
+
+export function getTruncatedTextData(text, limit, withSuffix = true) {
```
yesil, line 194: This method looks a bit sensible, should we try/catch and fallback to default value?

libs/features/mas/src/hydrate.js
```diff
@@ -175,17 +296,31 @@ function createSpectrumSwcButton(cta, aemFragmentMapping, isOutline, variant) {
...
+                  console.warn(`Failed to copy attribute ${attr.name}`, e);
+              }
+          }
+          checkoutBtn.connectedCallback();
+          await checkoutBtn.render();
+          await checkoutBtn.onceSettled();
+          spectrumCta.addEventListener('click', (e) => {
             e.stopPropagation();
-            cta.click();
+            checkoutBtn.click(e);
```
yesil, line 318: `click` method does not take parameters.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click

libs/features/mas/src/merch-card.js
```diff
@@ -376,7 +376,7 @@ export class MerchCard extends LitElement {
         const timeoutPromise = new Promise((resolve) =>
             setTimeout(() => resolve(false), MERCH_CARD_LOAD_TIMEOUT),
         );
-        const success = await Promise.race([successPromise, timeoutPromise, this.aemFragment?.updateComplete]);
+        const success = await Promise.race([successPromise, timeoutPromise]);
```
yesil, line 379: @bozojovicic didn't you add `this.aemFragment?.updateComplete`? 
I think we need to keep it.

libs/features/mas/src/variants/ah-try-buy-widget.js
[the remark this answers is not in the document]
yesil: I'm for using self explanatory names. `AdobeHomeTryBuyWidget` is also fine to me.

libs/features/mas/src/merch-card.js
3ch023: not sure we can keep it if we want merch-card to fire an error when price/cta failed - and thats our contract for ccd at the moment

3ch023: this basically means we ignore successPromise

libs/features/mas/src/variants/ah-try-buy-widget.js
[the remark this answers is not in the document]
st-angelo-adobe: I like `AHTryBuyWidget` because it's in line with the CCD stuff. Plus I've renamed it 3 times already 😁  . I will do it again if everyone agrees on one name, we can maybe decide it in a dev call.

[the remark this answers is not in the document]
yesil: it is fine too. no need to change.

libs/features/mas/src/merch-card.js
yesil: @bozojovicic noticed in his autoblock PR, that when you call `merchCard.checkReady()` too early, it does not work (gives false positive) because you first need to wait for the fragment to be loaded.
if the fragment is not resolved, we cannot know which offers are to be loaded anyway.

libs/features/mas/src/hydrate.js
st-angelo-adobe: Actually I'm finding hard to find a scenario where it fails, other than if the text is `null`/`undefined`, which I've accounted for now with a new commit.

One issue that I'm spotting though is if `<` or `>` are inside the text, as actual characters, the result isn't quite what you'd want. Doesn't break, but isn't right either.

I assume we should account for this scenario, right?

Axelcureno: added try/catch

Roycethan:
@Axelcureno 
1) When used dark theme , the theme is not apllied to few cards in double and single plan types:
https://mwpw-164093--milo--adobecom.hlx.live/libs/features/mas/docs/adobe-home.html?theme=dark
image

st-angelo-adobe:
@Roycethan 
2, 3: Done.
4. I seriously doubt Accent buttons need to change to Primary color when hovered. I'll confirm with the AH team if this is an actual requirement, it may have been a misunderstanding.
5. Actually, the title is truncated at 37 characters, plus 3 for ellipses = 40 in total. If this is not the desired behavior we can change that.
6. From what I'm seeing this is actually fine. The editor is counting more than actual characters, possibly (maybe `\n` is counted as 2 characters, not sure, this needs to be investigated).
image
8. That was just a remnant, previously `Triple` was just `Normal`, then you had the option to pick `Single` or `Double`. I changed it to `Triple`.

For 1 and 7 we'll look into it. cc: @Axelcureno

Roycethan:
> @Axelcureno Can you plz fix following , Thanks
> 
> 1. When used dark theme , the theme is not apllied to few cards in double and single plan types:
>    https://mwpw-164093--milo--adobecom.hlx.live/libs/features/mas/docs/adobe-home.html?theme=dark
> 
> image
> 2)Plz update the first card in triple plan to have background - Grey
> 
> 3. On first card - 'See terms'' and Annual billed monthly' should be in separate rows and not wrapped - Better add pricing in between
> 
> image
> 4. ac: buttons should be responsive to hover state, e.g. S2/Accent colored button changes to S2/Primary when hovered over - I don't see hover and change color working in gallery , Can you plz check this requirement ?
> 5. ac: product titles should be truncated with ellipses (...) if they exceed 40 characters or 3 lines
>    Looks like its truncating title before 40 characters or 3 lines -
>    ex: Creative Cloud All Apps Long Test Exa...
> 
> image
> 6. max character count in description should be 200 characters - Its taking more than that
> 
> image
> 7. In description field , if i enter around more than 25 characters in 1 row , its not wrapped into next line automatically and remaining characters are not rendered.
> 
> image
> 8. why the 'size' feild is blank for this card ?
>    https://mwpw-164093--mas--adobecom.hlx.page/studio.html?milolibs=MWPW-164093--milo--adobecom#path=adobe-home&page=content&query=f07084e9-1963-4877-8ab0-f7d2982e62cb
> 
> image


@Axelcureno  cc @st-angelo-adobe 
1 - Need fix
2,3 -fixed
4- Still a Question to PDM may be since its mentioned as AC
5- Ok as explained
6 - Not working as expected
7 - Need investigation
8- fixed

Axelcureno:
@Roycethan we are not delivering "widgets", we are only delivering the cards. The page in milo is just for demonstration purposes, it does not represent the actual widget. The widget containing the cards is built in adobe home code.

Axelcureno:
@Roycethan item #9 is a mistake in the color of figma, we are not overriding colors in dark theme and even in figma it says "Dark mode follows the same S2 guidelines as Light mode."
I've left a comment here for correction: https://www.figma.com/design/7tUtNgFelfMjgPoJ5QcE1k?node-id=5160-32869&m=dev#1142597000

Axelcureno:
I've fixed the remaining items, please check @Roycethan

Roycethan:
> I've fixed the remaining items, please check @Roycethan

Thanks 1 is fixed, for 6 and 7 will create a separate followup ticket, approving this to proceed to stage

narcis-radu:
> Thanks 1 is fixed, for 6 and 7 will create a separate followup ticket, approving this to proceed to stage
@Roycethan , @Axelcureno , @st-angelo-adobe - Have you created follow-up tickets for 6 / 7 ?

Axelcureno:
> le for someone outside the Merch team to test the changes. I had to go through comments to find a URL where the elements are rendered. T

@Roycethan 6 and 7 are working as expected. I have validated on my end.

Roycethan:
> https://jira.corp.adobe.com/browse/MWPW-168322

Cool thanks @Axelcureno for fixing them, closing as fixed

PR #3670, opened 2025-02-11
[MWPW-166126] Add new "updateAttribute" action to MEP

AdobeLinhart:
Support a new action called "updateAttribute" that allows a MEP user to end a CSS selector with the name of an attribute they would like to update/add. AKA the last word in the CSS selector would be the attribute name.

Example selectors:
marquee primary-cta href
marquee primary-cta href_osi

Resolves: [MWPW-166126)](https://jira.corp.adobe.com/browse/MWPW-166126))

**Instructions**
Verify that the [manifest ](https://adobe.sharepoint.com/:x:/r/sites/adobecom/_layouts/15/Doc.aspx?sourcedoc=%7BE17FC7EB-6655-41B9-97D4-1C3D6BDF009B%7D&file=attribute.xlsx&action=default&mobileredirect=true) is properly adding/updating attributes.

**Test URLs:**
- Psi check: https://update-attribute-action--milo--adobecom.aem.page/?martech=off
- Before: https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2025/q1/update-attribute-action/marquee
- After: https://main--cc--adobecom.hlx.page/drafts/mepdev/fragments/2025/q1/update-attribute-action/marquee?milolibs=update-attribute-action


jpratt2 (approved):
Verfied: with the new updateAttribute command, it is possible to add or modify an attribute

libs/features/personalization/personalization.js
```diff
@@ -222,6 +223,40 @@ const COMMANDS = {
       createContent(el, cmd),
     );
   },
+  [COMMANDS_KEYS.updateAttribute]: (el, cmd) => {
+    const { manifestId, targetManifestId } = cmd;
+    if (!cmd.attribute || !cmd.content) return;
+    const [attribute, parameter] = cmd.attribute.split('_');
+
+    let value;
+    if (attribute === 'href') {
```
vgoodric, line 232: I believe it is frowned upon to do a nested if right after an if.  I believe the preference would be to do `if (parameter && attribute === 'href') {` and then an else if for just `attribute === 'href'`

vgoodric: but not even use you need the else if because href with no param can behave like all other attributes I think

libs/features/personalization/personalization.js
```diff
@@ -433,22 +468,34 @@ function getModifiers(selector) {
   }
   return { sel, modifiers };
 }
-export function modifyNonFragmentSelector(selector) {
+export function modifyNonFragmentSelector(selector, action) {
   const { sel, modifiers } = getModifiers(selector);
+
+  let newSelector = sel
```
vgoodric, line 474: we should call this modifiedSelector so it's easier to assign it to the object after

libs/features/personalization/personalization.js
```diff
@@ -475,8 +521,9 @@ function getSelectedElements(sel, rootEl, forceRootEl) {
   }
   if (modifiers.includes(FLAGS.all) || !els.length) return { els, modifiers };
```
vgoodric, line 522:
```suggestion
  if (modifiers.includes(FLAGS.all) || !els.length) return { els, modifiers, attribute };
```

vgoodric (approved):

PR #3673, opened 2025-02-12
MWPW-156978 adding a lana log for gnav search usage

salonijain3:
* Adding a log for gnav search so that we can track the usage of the component, we would want to remove this component based on the usage. 

Resolves: [MWPW-156978](https://jira.corp.adobe.com/browse/MWPW-156978)

**Test URLs:**
- After: https://main--cc--adobecom.hlx.page/drafts/souj/unav/unavtest?milolibs=mwpw-156978--milo--salonijain3&martech=off


bandana147 (approved):

sharmrj (approved):

mokimo:
I think on the global-navigation it's not used or pretty safe to remove.

FYI However blog relies on the `gnav` (old) nav's search. Similar to that, we have a small internal use-case that likes that functionality of the old gnav.

PR #3680, opened 2025-02-13
MWPW-167716 Remove await from both gnav and georouting

sharmrj:
This is the ~~first~~ second of ~~four~~ five PRs (the first is #3720) aimed at improving Global-Navigation performance. This first PR does not directly affect the global-navigation (or georouting), but is rather aimed at improving the load times of the sections after the first one.  For more information look at the discussion #3558

The first section is assumed to be LCP and is prioritized. After the first section we load a series of features/blocks, among them georouting and global-navigation. Since we `await` both of these blocks, the subsequent sections _cannot_ continue until they stop blocking the main thread.

This can be bad because the user cannot see the rest of the page, or even scroll, until the subsequent sections load. The affect of this poor UX is not something we measure but it is important nonetheless. And even though we assume LCP _should_ be in the first section, in practice it is not always.

The solution is then to remove the `await`s on the global-navigation and georouting. The effects of this change have been documented in the discussion #3558.

Resolves: [MWPW-167716](https://jira.corp.adobe.com/browse/MWPW-167716)

**Test URLs:**
- https://non-blocking-gnav-geourouting--milo--adobecom.hlx.page/?martech=off


libs/utils/utils.js
```diff
@@ -1192,13 +1192,15 @@ async function loadPostLCP(config) {
...
+    import('../features/georoutingv2/georoutingv2.js')
+      .then(({ default: loadGeoRouting }) => {
+        loadGeoRouting(config, createTag, getMetadata, loadBlock, loadStyle);
+      });
   }
   const header = document.querySelector('header');
   if (header) {
     header.classList.add('gnav-hide');
-    await loadBlock(header);
+    loadBlock(header);
```
mokimo, line 1203: Is there any influence on CLS if you do this? Also what happens in terms of personalization? What if that loads before the GNAV?

libs/utils/utils.js
```diff
@@ -1192,13 +1192,15 @@ async function loadPostLCP(config) {
 
   const georouting = getMetadata('georouting') || config.geoRouting;
   if (georouting === 'on') {
-    const { default: loadGeoRouting } = await import('../features/georoutingv2/georoutingv2.js');
-    await loadGeoRouting(config, createTag, getMetadata, loadBlock, loadStyle);
+    import('../features/georoutingv2/georoutingv2.js')
```
mokimo, line 1195: Given @vhargrave concerns in the discussion, I suppose this was brought up with the georouting PM?

mokimo (approved):

libs/utils/utils.js
sharmrj: No there's no CLS.
 AFAICT if personalization must load it will load before either gnav or georouting; either right before georouting in `loadPostLCP` or in `checkForPageMods`. In both cases it is `await`ed. There is some personalization that takes place inside the gnav, but simply removing the `await` does not affect it in any way.

sharmrj:
To give an idea of what this change accomplishes, it essentially schedules our network requests postLCP more efficiently. As a visual representation, consider the below screen shot of our page load currently.

Screenshot 2025-02-14 at 9 25 54 AM

We see that many network requests are chained, blocking further requests until they are received. Specifically there's a visible bottleneck in the middle there. That happens to be georouting and global-navigation. See the trace here:
[Trace-20250214T092251.json.zip](https://github.com/user-attachments/files/18793227/Trace-20250214T092251.json.zip)

Now with the change in this PR:

Screenshot 2025-02-14 at 9 25 19 AM
trace: 

[Trace-20250214T092719.json.zip](https://github.com/user-attachments/files/18793256/Trace-20250214T092719.json.zip)

We can see that the bottleneck is gone. By removing just these two `await`s we'll see fewer gaps in how our code and, subsequently, our network requests are scheduled (This is what I mean by scheduling more efficiently).

bandana147 (approved):

libs/utils/utils.js
vhargrave, line 1195: If you haven't done the performance enhancement we discussed for the georouting, then please confirm that this delay was tested and discussed with the georouting PM. 
Otherwise please at least do the enhancement that was discussed (kicking off the loading of the .json file here).

libs/utils/utils.js
```diff
@@ -1192,13 +1192,15 @@ async function loadPostLCP(config) {
...
+      });
   }
   const header = document.querySelector('header');
   if (header) {
     header.classList.add('gnav-hide');
-    await loadBlock(header);
+    loadBlock(header);
     header.classList.remove('gnav-hide');
   }
   loadTemplate();
```
vhargrave, line 1206: Not a must, but i bet you could kick the loading of the `fonts.js` at the beginning of the postLCP somewhere and only await it here.. That would probably also give you some great savings.

vhargrave (changes requested):
see comments

libs/utils/utils.js
sharmrj: We've tagged them on JIRA, and we'll continue the conversation there.
I looked into doing that optimization here, but I think it would fit well as its own PR. I'd prefer to keep the PRs as granular as we can. If you prefer I can make it the very next optimization we do.

vhargrave: I understand that, but I can't approve a change to the load time of the georouting that goes live if you haven't approved it with the georouting PM or mitigated it with the change we discussed.

sharmrj: Ok I've raised a separate PR to optimize georouting first and then we can merge this one. Here's the PR @vhargrave : #3720

vhargrave: @sharmrj perfect, thank you. 
I've left some feedback 🙌

vhargrave:
@sharmrj nala tests are failing for some reason. Can you take a look?

vhargrave (approved):
nice!

sigadamvenkata:
this was tested in given branch. complete testing would be done for integrations, devices etc on stage.

PR #3683, opened 2025-02-13
MWPW-167306 [Plans Milo] Quantity Selector & Badge

bozojovicic:
Add quantity selector and badge to plans cards. Quantity selector also had one bug: in text field you could set the value smaller than the minimal configured value.

Resolves: [MWPW-167306](https://jira.corp.adobe.com/browse/MWPW-167306)

**Test URLs:**
- After: https://MWPW-167306--milo--bozojovicic.aem.page/?martech=off


libs/features/mas/src/variants/plans.js
```diff
@@ -9,8 +9,10 @@ export const PLANS_AEM_FRAGMENT_MAPPING = {
   description: { tag: 'div', slot: 'body-xs' },
   mnemonics: { size: 'l' },
   callout: {tag: 'div', slot: 'callout-content'},
+  quantitySelect: { tag: 'div', slot: 'quantity-select' },
   stockOffer: true,
   secureLabel: true,
+  badge: true,
```
Axelcureno, line 15: why this is needed?

Axelcureno (changes requested):

libs/features/mas/src/hydrate.js
```diff
@@ -123,10 +123,21 @@ export function processPrices(fields, merchCard, mapping) {
   appendSlot('prices', fields, merchCard, mapping); 
 }
 
+function isEmpty(html) {
```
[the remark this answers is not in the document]
bozojovicic: Done

libs/features/mas/src/variants/plans.js
bozojovicic: Card editor reads this mapping to decide which authoring fields will be displayed for 'plans' cards. For card badge (the yellow thing in the upper right corner) we don't have slots but it needs to be mentioned here to display that Badge text field in card editor. The same as for stock checkbox.

libs/features/mas/src/hydrate.js
```diff
@@ -123,10 +123,25 @@ export function processPrices(fields, merchCard, mapping) {
   appendSlot('prices', fields, merchCard, mapping); 
 }
 
+function parseHtml(html) {
+  const parser = new DOMParser();
+  return parser.parseFromString(html, 'text/html');
+}
```
JasonHowellSlavin, line 129: I dont see this getting reused, and only called in `isHtmlEmpty`. Is there a reason not to instatiate the parser and parse within `isHtmlEmpty` itself, it will technical save 3 lines of code.

libs/features/mas/src/merch-quantity-select.js
```diff
@@ -93,10 +95,12 @@ export class MerchQuantitySelect extends LitElement {
             inputValue > 0 &&
             inputValue !== this.selectedValue
         ) {
-            const adjustedInputValue =
+            let adjustedInputValue =
                 this.maxInput && inputValue > this.maxInput
                     ? this.maxInput
                     : inputValue;
+            adjustedInputValue = this.min && adjustedInputValue < this.min
+                ? this.min : adjustedInputValue;
```
JasonHowellSlavin, line 103:
```suggestion
            let adjustedInputValue = inputValue;
            if (this.maxInput && inputValue > this.maxInput)
                    adjustedInputValue = this.maxInput;
           if (this.min && adjustedInputValue < this.min)
                adjustedInputValue = this.min;
```
I think this makes it a bit more readable, I found the reassignments and ternaries to be a bit hard to read with the compound conditionals ahead of them. Let me know what you think.

libs/features/mas/test/merch-quantity-select.test.html.js
```diff
@@ -189,7 +189,7 @@ runTests(async () => {
                 });
                 event.composedPath = () => [quantitySelect];
                 inputField.dispatchEvent(event);
-                await delay();
+                await delay(600);
```
JasonHowellSlavin, line 192: In regards to all of these, is 600 the minimum ms you can delay? If i am reading this correctly, this will add 2.4s to our testing suite.

3ch023 (changes requested):
Hi @bozojovicic ,
could you pls also add a card to https://mwpw-167306--milo--bozojovicic.aem.page/libs/features/mas/docs/plans.html? 
You need to clone an existing card in Nala folder in studio (important to be a Nala folder!) and then you can reference it in plans.md and run 'npm run build:docs' in /libs/mas/features - this will generate a new plans.html

and then could you pls add a test to masacom.test.js to verify quantity selector presense on that card and badge?

PR #3684, opened 2025-02-13
MWPW-164755:Add <br> when copying multiple blocks

zagi25:
## Description

- Add spacing when multiple blocks are copied from `Library` (Layout patterns)

## Related Issue
Resolves: [MWPW-164755](https://jira.corp.adobe.com/browse/MWPW-164755)

## Testing instructions
- Open **Blocks** > **Layout Patterns** > copy one of the options
- Open empty document and paste copied pattern
- Blocks should have spacing between them

## Test URLs
- After: https://mwpw-164755-spacing-pattern-blocks--milo--adobecom.hlx.page/tools/library?martech=off


robert-bogos (approved):

libs/blocks/library-config/lists/blocks.js
```diff
@@ -283,7 +283,7 @@ export default async function loadBlocks(blocks, list, query, type) {
         const containerHtml = getHtml(container, block.path);
         e.target.classList.add('copied');
         setTimeout(() => { e.target.classList.remove('copied'); }, 3000);
-        const blob = new Blob([`${BLOCK_SPACING}${containerHtml}${BLOCK_SPACING}`], { type: 'text/html' });
```
mokimo, line 286: Why do you only move once`BLOCK_SPACING` to `getHtml` shouldn't you move both up? 🤔

mokimo (approved):

libs/blocks/library-config/lists/blocks.js
zagi25: You are probably right

libs/blocks/library-config/lists/blocks.js
```diff
@@ -283,7 +284,7 @@ export default async function loadBlocks(blocks, list, query, type) {
...
+        const blob = new Blob([`${containerHtml}`], { type: 'text/html' });
```
mokimo, line 287: Does this still need to be stringified, or can this just be 
```suggestion
        const blob = new Blob([containerHtml], { type: 'text/html' });
```

mokimo (approved):

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-164755

PR #3685, opened 2025-02-13
MWPW-165992 Fix Chart Font Family

meganthecoder:
* Updates chart font family to match milo body font
* Fixes issue where incorrect font is sometimes seen on Safari

Resolves: [MWPW-165992](https://jira.corp.adobe.com/browse/MWPW-165992)

**Test URLs:**
- After: https://chart-font--milo--meganthecoder.aem.page/drafts/methomas/chart-two-up?martech=off

### DIFF
--- a/libs/blocks/chart/chartLightTheme.js
+++ b/libs/blocks/chart/chartLightTheme.js
@@ -23,7 +23,7 @@ export default (deviceSize) => {
   const axisColor = '#767676';
   const options = {
     textStyle: {
-      fontFamily: 'Adobe Clean',
+      fontFamily: '"Adobe Clean", adobe-clean, "Trebuchet MS", sans-serif',
       fontSize: 16,
       fontWeight: 700,
     },


mokimo (approved):

Brandon32 (approved):

libs/blocks/chart/chartLightTheme.js
```diff
@@ -23,7 +23,7 @@ export default (deviceSize) => {
   const axisColor = '#767676';
   const options = {
     textStyle: {
-      fontFamily: 'Adobe Clean',
+      fontFamily: '"Adobe Clean", adobe-clean, "Trebuchet MS", sans-serif',
```
narcis-radu, line 26: Do we really need to set a font family here? Ideally we want to inherit from body, which has the font family defined.

robert-bogos (commented):
For the failing NALA tests you could try to clear the cached assets with something like:
`CURL -si -X POST https://admin.hlx.page/code/robert-bogos/milo/mwpw-142622-project-pep/*`

libs/blocks/chart/chartLightTheme.js
meganthecoder: The chart library will add a font family if you don't set any so inheriting from body won't work.

narcis-radu: @meganthecoder - using `'inherit'` should do the trick
Screenshot 2025-02-20 at 17 02 56

meganthecoder: @narcis-radu Right, of course. Sorry, I didn't see this until after it got merged to stage. Do you want me to create a follow-up PR to change to `inherit`?

narcis-radu: @meganthecoder - only if it's not too much trouble. We don't change fonts that often, but if we do, it would be one less line to change if we use `inherit`. Thank you!

meganthecoder: PR https://github.com/adobecom/milo/pull/3742

PR #3694, opened 2025-02-17
[MWPW-159511] Create anchor (ID) elements

narcis-radu:
This is a new feature that allows authors to create anchor elements by using specific markup `:anchor-id-of-the-anchor:`. The element is transformed by EDS into `<span class="icon icon-anchor-id-of-the-anchor"></span>` and then transformed by custom code into `<span class="anchor anchor-id-of-the-anchor" id="id-of-the-anchor"></span>`.
The anchor can be referenced as `#id-of-the-anchor` in the URL or via href.

Resolves: [MWPW-159511](https://jira.corp.adobe.com/browse/MWPW-159511)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/narcis/anchors?martech=off
- After: https://anchor--milo--narcis-radu.aem.page/drafts/narcis/anchors?martech=off

### DIFF
--- a/libs/styles/styles.css
+++ b/libs/styles/styles.css
@@ -893,6 +893,10 @@ a.static:active  {
 }
 
 /* Anchors */
+span.anchor {
+  pointer-events: none;
+}
+
 span.anchor:target::before {
   content: "";
   display: block;
--- a/libs/styles/styles.css
+++ b/libs/styles/styles.css
@@ -892,6 +892,14 @@ a.static:active  {
   visibility: hidden;
 }
 
+/* Anchors */
+span.anchor:target::before {
+  content: "";
+  display: block;
+  height: 100px;
+  margin: -100px 0 0;
+}
+
 /* mobile only */
 @media (max-width: 600px) {
   .con-button.button-justified-mobile {
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -833,7 +833,7 @@ async function decorateHeader() {
   }
 }
 
-async function decorateAnchors(area) {
+function decorateAnchors(area) {
   const anchors = area.querySelectorAll('span[class*="icon-anchor-"]');
   if (anchors.length === 0) return;
   anchors.forEach((anchor) => {
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -835,7 +835,7 @@ async function decorateHeader() {
 
 function decorateAnchors(area) {
   const anchors = area.querySelectorAll('span[class*="icon-anchor-"]');
-  if (anchors.length === 0) return;
+  if (!anchors.length) return;
   anchors.forEach((anchor) => {
     anchor.id = anchor.classList[1].replace('icon-anchor-', '');
     anchor.classList.replace(anchor.classList[0], 'anchor');
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -833,6 +833,16 @@ async function decorateHeader() {
   }
 }
 
+async function decorateAnchors(area) {
+  const anchors = area.querySelectorAll('span[class*="icon-anchor-"]');
+  if (anchors.length === 0) return;
+  anchors.forEach((anchor) => {
+    anchor.id = anchor.classList[1].replace('icon-anchor-', '');
+    anchor.classList.replace(anchor.classList[0], 'anchor');
+    anchor.classList.replace(anchor.classList[1], `anchor-${anchor.id}`);
+  });
+}
+
 async function decorateIcons(area, config) {
   const icons = area.querySelectorAll('span.icon');
   if (icons.length === 0) return;
@@ -1380,6 +1390,7 @@ async function processSection(section, config, isDoc) {
   preloadBlockResources(section.preloadLinks);
   await Promise.all([
     decoratePlaceholders(section.el, config),
+    decorateAnchors(section.el),
     decorateIcons(section.el, config),
   ]);
   const loadBlocks = [...stylePromises];
--- a/libs/blocks/marquee-anchors/marquee-anchors.js
+++ b/libs/blocks/marquee-anchors/marquee-anchors.js
@@ -16,7 +16,7 @@ const blockTypeSizes = {
 };
 
 const fetchedIcons = {};
-function decorateAnchors(anchors) {
+function decorateMarqueeAnchors(anchors) {
   const linkGroup = createTag('div', { class: 'links-group' });
   anchors[0].insertAdjacentElement('beforebegin', linkGroup);
   const { miloLibs, codeRoot } = getConfig();
@@ -79,5 +79,5 @@ export default function init(el) {
   });
 
   const anchors = el.querySelectorAll('.anchor-link');
-  if (anchors.length) decorateAnchors(anchors);
+  if (anchors.length) decorateMarqueeAnchors(anchors);
 }
--- a/test/utils/mocks/body.html
+++ b/test/utils/mocks/body.html
@@ -84,6 +84,11 @@
   </div>
   <!-- Default content -->
   <div>
+    <p>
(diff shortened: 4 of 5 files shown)


libs/utils/utils.js
```diff
@@ -833,6 +833,16 @@ async function decorateHeader() {
   }
 }
 
+async function decorateAnchors(area) {
+  const anchors = area.querySelectorAll('span[class*="icon-anchor-"]');
```
mokimo, line 837: Why is it `icon-anchor` I thought this is a general anchor thing?

libs/utils/utils.js
```diff
@@ -833,6 +833,16 @@ async function decorateHeader() {
...
+  if (anchors.length === 0) return;
+  anchors.forEach((anchor) => {
+    anchor.id = anchor.classList[1].replace('icon-anchor-', '');
```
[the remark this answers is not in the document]
narcis-radu: EDS will format everything `:something:` as icons. I have explored other options, but they are either non semantic or the resulting element gets translated in the process. Had a chat with LOC team and they also suggested this approach.

robert-bogos (approved):
A bit hacky, but I don't see a better option. 👍

We need to test this on localized pages to make sure the anchor id does not get translated. @SilviuLCF @NadiiaSokolova

libs/styles/styles.css
```diff
@@ -892,6 +892,14 @@ a.static:active  {
   visibility: hidden;
 }
 
+/* Anchors */
+span.anchor:target::before {
+  content: "";
+  display: block;
+  height: 100px;
+  margin: -100px 0 0;
+}
```
overmyheadandbody, line 901: This should be revised. Currently, it makes the content around the anchor point unselectable, which is not ideal. "After Effects" and the eyebrow text can't be selected in the screenshot below.
Screenshot 2025-02-18 at 13 56 37

overmyheadandbody (commented):
I do have a bit of a concern regarding hiding content behind the anchor area.

I had a quick look whether `decorateAnchors` was being imported in other projects, since we're renaming it, but it doesn't seem like it, [here](https://github.com/search?q=org%3Aadobecom+decorateAnchors&type=code). Each project redefines the method.

libs/styles/styles.css
overmyheadandbody: A simple negative `z-index` value (and relative position) or a `pointer-events: none;` should do the trick.

libs/utils/utils.js
```diff
@@ -833,6 +833,16 @@ async function decorateHeader() {
   }
 }
 
+function decorateAnchors(area) {
+  const anchors = area.querySelectorAll('span[class*="icon-anchor-"]');
+  if (anchors.length === 0) return;
+  anchors.forEach((anchor) => {
+    anchor.id = anchor.classList[1].replace('icon-anchor-', '');
```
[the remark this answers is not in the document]
narcis-radu: EDS dictates the order and it is fixed. We use the same logic for icons / tooltips https://github.com/adobecom/milo/blob/stage/libs/features/icons/icons.js#L62

vhargrave:
@narcis-radu maybe just as another idea - Express allows for anchors to be added over section-metadata. This has worked really well for that project, and also solves the problem of translation. 
I'm not sure that anchor tags really have to be so specific to headers, so maybe that approach could also solve this solve. 
https://main--express-milo--adobecom.aem.page/drafts/vhargrave/milo/business-milo#column-1

libs/styles/styles.css
narcis-radu: Great finding, thank you! I will also fix the existing code for marquee-anchors.

narcis-radu:
@vhargrave - that would work, but we have content that's not part of a block (e.g. plain text) as well as blocks that have multiple anchors (e.g. tabs).

libs/utils/utils.js
```diff
@@ -833,6 +833,16 @@ async function decorateHeader() {
   }
 }
 
+function decorateAnchors(area) {
+  const anchors = area.querySelectorAll('span[class*="icon-anchor-"]');
+  if (anchors.length === 0) return;
```
overmyheadandbody, line 838: This being utils, maybe we strip it down to
```suggestion
  if (!anchors.length) return;
```

libs/utils/utils.js
```diff
@@ -833,6 +833,16 @@ async function decorateHeader() {
...
+  anchors.forEach((anchor) => {
+    anchor.id = anchor.classList[1].replace('icon-anchor-', '');
+    anchor.classList.replace(anchor.classList[0], 'anchor');
+    anchor.classList.replace(anchor.classList[1], `anchor-${anchor.id}`);
```
overmyheadandbody, line 842: Is this class really useful, since we have the ID? Could we not remove it entirely?

overmyheadandbody (commented):
A couple more notes, but looking good otherwise.

libs/utils/utils.js
narcis-radu: We might need some CSS changes and I thought it would be easier if the class is there. Also, it follows the exact same track as "icon icon-something" defined by EDS, so wanted to be consistent.

mokimo (approved):

libs/utils/utils.js
overmyheadandbody: Specific styles for each anchor? I can't really think of a use-case for that. Again, this is more of a "this is utils, let's only keep what we need right now" comment.

narcis-radu: ok, will remove

overmyheadandbody (dismissed):
I found another issue with the authoring of the anchors. Using icons in aside headings seems to lead to the text being laid out on just one line
Screenshot 2025-02-18 at 16 57 00

That also occurs outside of this branch (https://main--milo--adobecom.aem.page/drafts/narcis/anchors?martech=off), so some additional code might need to be adapted. Asides without icons in the heading work as expected - https://main--milo--adobecom.hlx.page/docs/library/kitchen-sink/aside

LE: seems like this could've been a hiccup, I'm able to see the issue on the `main` branch, but not on the feature branch 🤷 

Later LE: seems that only if icons are used with titles does this occur. I was thinking that maybe this could lead to CLS (title going from one line to multiple lines), I haven't seen this myself, but should be investigated by QA. Maybe have such an example as the first block on the page. We additionally should consider whether actual icons should be tolerated by the block for future use-cases.

narcis-radu:
@overmyheadandbody - this is happening on `main` because the `icon-anchor` is not transformed to `anchor anchor-something`. Authors should not add anchors _before_ the code is merged, I don't think this is something that could happen on Production. As soon as the code gets merged the problem will go away so there's no point in fixing it.
I am happy to fix the issues, but don't really see the benefit.

overmyheadandbody:
@narcis-radu - yes, I agree, I've updated my comment that we should consider any CLS impact during QA and that we should discuss whether to address the issue in another story, in case authors do add icons to the aside block.

overmyheadandbody (approved):

vhargrave (changes requested):
It seems like this approach has been done before and the EDS team was also not recommending it. I'm right now just reading through the following threads, but requesting changes just to pause any merge for a bit. 😓 

https://adobe-mwp.slack.com/archives/C05QU7MMRNF/p1728332018246679?thread_ts=1728143191.642259&cid=C05QU7MMRNF
https://github.com/adobe/helix-html-pipeline/issues/538

vhargrave:
> @vhargrave - that would work, but we have content that's not part of a block (e.g. plain text) as well as blocks that have multiple anchors (e.g. tabs).

@narcis-radu I'm gathering more data still, but it would be possible to have anchor tags with both plain text and tabs for example. See here 

- https://main--express-milo--adobecom.aem.page/drafts/vhargrave/anchors#anchor1
- https://main--express-milo--adobecom.aem.page/drafts/vhargrave/anchors#plaintext

PR #3703, opened 2025-02-19
[MWPW-162639] youtube block a tag fix

DKos95:
This resolves the issue where the youtube link was being processed into a block because the link was first in the paragraph and not text.

Resolves: [MWPW-162639](https://jira.corp.adobe.com/browse/MWPW-162639)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/cpeyer/temp/og-issue?martech=off
- After: https://inline-youtube-link--milo--adobecom.aem.page/drafts/cpeyer/temp/og-issue?martech=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -264,7 +264,9 @@ export const [setConfig, updateConfig, getConfig] = (() => {
 })();
 
 export function isInTextNode(node) {
-  return (node.parentElement.children.length > 1 && node.parentElement.firstChild.tagName === 'A') || node.parentElement.firstChild.nodeType === Node.TEXT_NODE;
+  console.log('node.parentElement.children.length', node.parentElement.children.length);
+  
+  return (node.parentElement.childNodes.length > 1 && node.parentElement.firstChild.tagName === 'A') || node.parentElement.firstChild.nodeType === Node.TEXT_NODE;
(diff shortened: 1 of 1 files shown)


libs/utils/utils.js
```diff
@@ -264,7 +264,7 @@ export const [setConfig, updateConfig, getConfig] = (() => {
 })();
 
 export function isInTextNode(node) {
-  return node.parentElement.firstChild.nodeType === Node.TEXT_NODE;
+  return (node.parentElement.children.length > 1 && node.parentElement.firstChild.tagName === 'A') || node.parentElement.firstChild.nodeType === Node.TEXT_NODE;
```
zagi25, line 267: This doesn't cover the case when we have only youtube link https://inline-youtube-link--milo--adobecom.aem.page/drafts/ratko/test-document . Maybe using `childNodes` instead of `children` can solve this https://developer.mozilla.org/en-US/docs/Web/API/Node/childNodes

robert-bogos (approved):

zagi25 (approved):

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-162639

PR #3714, opened 2025-02-20
[MWPW-159511] Create anchor (ID) elements

narcis-radu:
New feature that allows authors to set anchors using section metadata. The [previous PR](https://github.com/adobecom/milo/pull/3694) has been closed in favour of this approach.

Resolves: [MWPW-159511](https://jira.corp.adobe.com/browse/MWPW-159511)

**Test URLs:**
- After: https://anchor2--milo--narcis-radu.aem.page/?martech=off

Test URL - https://anchor2--milo--narcis-radu.aem.page/drafts/narcis/anchors/default

### DIFF
--- a/libs/blocks/section-metadata/section-metadata.js
+++ b/libs/blocks/section-metadata/section-metadata.js
@@ -55,6 +55,12 @@ function handleDelay(time, section) {
   setTimeout(() => { section.classList.remove('hide-sticky-section'); }, getDelayTime(time));
 }
 
+function handleAnchor(anchor, section) {
+  if (!(anchor || section)) return;
+  section.id = anchor;
+  section.classList.add('section-anchor');
+}
+
 export const getMetadata = (el) => [...el.childNodes].reduce((rdx, row) => {
   if (row.children) {
     const key = row.children[0].textContent.trim().toLowerCase();
@@ -73,4 +79,5 @@ export default async function init(el) {
   if (metadata.layout) handleLayout(metadata.layout.text, section);
   if (metadata.masonry) handleMasonry(metadata.masonry.text, section);
   if (metadata.delay) handleDelay(metadata.delay.text, section);
+  if (metadata.anchor) handleAnchor(metadata.anchor.text, section);
 }
--- a/libs/blocks/section-metadata/section-metadata.js
+++ b/libs/blocks/section-metadata/section-metadata.js
@@ -56,7 +56,7 @@ function handleDelay(time, section) {
 }
 
 function handleAnchor(anchor, section) {
-  if (!(anchor || section)) return;
+  if (!anchor || !section) return;
   section.id = anchor;
   section.classList.add('section-anchor');
 }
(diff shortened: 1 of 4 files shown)


mokimo (approved):
That's not as hacky definitely better!!

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -73,4 +79,5 @@ export default async function init(el) {
   if (metadata.layout) handleLayout(metadata.layout.text, section);
   if (metadata.masonry) handleMasonry(metadata.masonry.text, section);
   if (metadata.delay) handleDelay(metadata.delay.text, section);
+  if (metadata.anchor) handleAnchor(metadata.anchor.text, section);
```
overmyheadandbody, line 82: How does this work for long sections? Assuming I have a section with 3 long blocks and I want to link to the 3rd, wouldn't I be taken to the top f the section? Or are we suggesting that blocks that should have anchors be separated into their own sections?

vhargrave (approved):
Agree with Rares' comments using variables, but this is great

libs/blocks/section-metadata/section-metadata.js
overmyheadandbody: That may be so, but we need to make sure this implementation satisfies GWP's needs and doesn't cause re-authoring overhead.

vhargrave: You don't think it will or you'd just like to see confirmation from GWP folks that they're happy with it?

libs/styles/styles.css
```diff
@@ -892,6 +892,17 @@ a.static:active  {
   visibility: hidden;
 }
 
+/* Anchors */
+.section-anchor {
+  scroll-margin-top: 64px;
```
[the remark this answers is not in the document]
narcis-radu: yes, yes

libs/blocks/section-metadata/section-metadata.js
narcis-radu: I think there will be use-cases where authors will need to make some changes in regards to sections (e.g. place the block into a separate section), but this downside is quite small compared to the benefits of using a solution tested in Production (Express) and agreed by the EDS team.

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -55,6 +55,12 @@ function handleDelay(time, section) {
   setTimeout(() => { section.classList.remove('hide-sticky-section'); }, getDelayTime(time));
 }
 
+function handleAnchor(anchor, section) {
+  if (!(anchor || section)) return;
```
overmyheadandbody, line 59: One last nit: `anchor` is a string, right? If the string is empty, the conditional returns `false` and it ends up adding an empty ID. Wouldn't it be better to have it as `!anchor || !section` or `!(anchor && section)`?

narcis-radu: you're correct, I made the changes.

overmyheadandbody (approved):
Thanks for the changes, looks good now! 🚀

robert-bogos (approved):

SilviuLCF:
Added High priority since will help authoring NAB scoped pages before loc sync starts on 7th March.

PR #3721, opened 2025-02-21
[MWPW-167309] - Table h tags replace

DKos95:
This resolves the issue where h tags are being added to a tables header which is conflicting semantics, so we are replacing it with a p tag.

Resolves: [MWPW-167309](https://jira.corp.adobe.com/browse/MWPW-167309)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/dusan/table-heading-testing?martech=off
- After: https://table-h-tags-replace--milo--adobecom.hlx.page/drafts/dusan/table-heading-testing?martech=off

### DIFF
--- a/libs/blocks/table/table.css
+++ b/libs/blocks/table/table.css
@@ -78,22 +78,6 @@
   margin: 0;
 }
 
-.table p.h1-margins,
-.table p.h2-margins, 
-.table p.h3-margins, 
-.table p.h4-margins,
-.table p.h5-margins, 
-.table p.h6-margins {
-  font-weight: bold;
-}
-
-.table p.h1-margins { margin: 0.67em 0; }
-.table p.h2-margins { margin: 0.83em 0; }
-.table p.h3-margins { margin: 1em 0; }
-.table p.h4-margins { margin: 1.33em 0; }
-.table p.h5-margins { margin: 1.67em 0; }
-.table p.h6-margins { margin: 2.33em 0; }
-
 .table .col p picture {
   margin-right: 8px;
 }
--- a/libs/blocks/table/table.css
+++ b/libs/blocks/table/table.css
@@ -78,6 +78,10 @@
   margin: 0;
 }
 
+.table p.h1-margins, .table p.h2-margins, .table p.h3-margins, .table p.h4-margins .table p.h5-margins, .table p.h6-margins {
+  font-weight: bold;
+}
+
 .table p.h1-margins { margin: 0.67em 0; }
 .table p.h2-margins { margin: 0.83em 0; }
 .table p.h3-margins { margin: 1em 0; }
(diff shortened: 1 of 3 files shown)


libs/blocks/table/table.css
```diff
@@ -78,6 +78,22 @@
...
+.table p.h6-margins {
+  font-weight: bold;
+}
+
+.table p.h1-margins { margin: 0.67em 0; }
+.table p.h2-margins { margin: 0.83em 0; }
+.table p.h3-margins { margin: 1em 0; }
+.table p.h4-margins { margin: 1.33em 0; }
+.table p.h5-margins { margin: 1.67em 0; }
+.table p.h6-margins { margin: 2.33em 0; }
```
robert-bogos, line 95: Where did we get this specific values from?🤔

libs/blocks/table/table.js
```diff
@@ -94,7 +93,18 @@ function handleHeading(table, headingCols) {
       col.setAttribute('role', 'columnheader');
     }
 
-    nodeToApplyRoleScope.setAttribute('scope', 'col');
+    const headings = col.querySelectorAll('h1, h2, h3, h4, h5, h6');
+
+    headings.forEach((heading) => {
```
robert-bogos, line 98: nit: no need to cache and saves 1 line
```suggestion
    col.querySelectorAll('h1, h2, h3, h4, h5, h6').forEach((heading) => {
```

libs/blocks/table/table.css
DKos95: From inspecting default style of h tags
Screenshot 2025-02-21 at 10 16 46

libs/styles/styles.css
```diff
@@ -792,32 +792,32 @@ h1, h2, h3, h4, h5, h6 {
   scroll-margin-top: var(--feds-totalheight-nav);
 }
 
-h1 {
+h1, .table .col .h1-margins {
```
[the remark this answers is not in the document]
overmyheadandbody: I agree that the styles here shouldn't be tied to particular blocks. I also get that we want to avoid style duplication, so how about we make the heading replacement class name more generic, so they could potentially be used in any block in the future if needed? We should also scan this file for other heading-specific files and make sure everything is covered.

So instead of `.table .col .h1-margins`, something like `.h1-proxy` (better ideas welcome 😅 )

overmyheadandbody (changes requested):
Before looking at my comments, here's another idea: the intention is to hide these elements from the Web Rotor and not announce them as being headings. We can achieve this result just be attaching a `role="paragraph"` to the current headings in tables. This wouldn't require any style changes and would be a lot more straightforward. I tried it out quickly and it seems to work. Let's tag Taryn in the story and ask whether this is acceptable.

overmyheadandbody (approved):

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-167309

PR #3723, opened 2025-02-21
MWPW-166076 and MWPW-168450 business/education plans page URLs to point to CME-1 M7

bozojovicic:
Replace all `/creativecloud/business-plans.html` and `/creativecloud/education-plans.html` with `/store/segmentation` links. The meta tag
`<meta name="m7-pa-code" content="ccsn_direct_individual">`
is required for this redirection.
For education page the M7 URL will have `ms=EDU` parameter.

Resolves: [MWPW-166076](https://jira.corp.adobe.com/browse/MWPW-166076)
Resolves: [MWPW-168450](https://jira.corp.adobe.com/browse/MWPW-168450)

Test URL `mwpw166076m7--milo--bozojovicic.aem.page/drafts/bozo/demo?martech=off`

**Test URLs:**
- Before: https://main--milo--adobecom.aem.live?martech=off
- After: https://mwpw166076m7--milo--bozojovicic.aem.live?martech=off


libs/utils/utils.js
```diff
@@ -689,6 +690,31 @@ export function convertStageLinks({ anchors, config, hostname, href }) {
   });
 }
 
+async function getImsCountry() {
```
narcis-radu, line 693: I don't think `utils` is the best place for this method, as it would be loaded on every single project and every single visitor, regardless of their logged-in status. I would rather load this conditionally.

libs/utils/utils.js
```diff
@@ -689,6 +690,31 @@ export function convertStageLinks({ anchors, config, hostname, href }) {
...
+  if (window.adobeIMS?.isSignedInUser()) {
+    const profile = await window.adobeIMS.getProfile();
+    return profile.countryCode;
+  }
+
+  return null;
+}
+
+export async function generateM7Link() {
```
narcis-radu, line 702: This method should not be in `utils`.

libs/utils/utils.js
```diff
@@ -737,6 +763,13 @@ export function decorateLinks(el) {
         processQuickLink(a);
       })();
     }
+
+    if (a.href.includes(BUSINESS_PLANS_PATH)) {
```
narcis-radu, line 767: Have you explored using an autoblock for such scenario? I think it would be better instead of adding all the logic in utils.

narcis-radu (changes requested):

libs/utils/utils.js
narcis-radu: @bozojovicic - you can see `caas` or `faas` autoblocks for reference

bozojovicic: I created it `libs/blocks/m7business/m7business.js`
BTW, autoblock is awesome idea!

libs/utils/utils.js
bozojovicic: I moved all my changes to autoblock `blocks/m7business/m7business.js`

libs/utils/utils.js
bozojovicic: I moved all my changes to autoblock `blocks/m7business/m7business.js`

libs/utils/utils.js
```diff
@@ -111,6 +112,7 @@ const AUTO_BLOCKS = [
   { video: '.mp4' },
   { merch: '/tools/ost?' },
   { 'mas-autoblock': 'mas.adobe.com/studio' },
+  { m7business: '/creativecloud/business-plans.html' },
```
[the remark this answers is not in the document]
bozojovicic: But, this autoblock will be used if URL contains "/creativecloud/business-plans.html" so for both of these
https://www.adobe.com/creativecloud/business-plans.html
https://www.adobe.com/de/creativecloud/business-plans.html
it will work, right?
Instead of this `/de/` we will have `co=DE` in new URL.

bozojovicic:
This PR needs to be merged when the same redirection is available on Dexter pages, not before. That will be the case in the 2nd half of March.

libs/utils/utils.js
narcis-radu: Thank you @bozojovicic ! I have cleared my request for changes, things look good now. Thanks!

libs/blocks/m7business/m7business.js
```diff
@@ -0,0 +1,32 @@
...
+  return null;
+}
+
+export async function generateM7Link(options) {
+  const { locale } = getConfig();
+  const pageCountry = getMiloLocaleSettings(locale).country;
+  const imsCountry = await getImsCountry();
+  const country = imsCountry || pageCountry;
+
+  const m7link = new URL('https://commerce.adobe.com/store/segmentation');
```
narcis-radu, line 19: nit: we can set search params by default when building the URL.

3ch023 (approved):

libs/utils/utils.js
```diff
@@ -109,7 +110,9 @@ const AUTO_BLOCKS = [
   { 'pdf-viewer': '.pdf', styles: false },
   { video: '.mp4' },
   { merch: '/tools/ost?' },
-  { 'mas-autoblock': 'mas.adobe.com/studio' },
+  { 'mas-autoblock': 'mas.adobe.com/studio', styles: false },
```
overmyheadandbody, line 113: Since `styles` is set to `false`, is the `libs/blocks/mas-autoblock/mas-autoblock.css` empty file even needed?

libs/utils/utils.js
```diff
@@ -109,7 +110,9 @@ const AUTO_BLOCKS = [
...
+  { m7: '/creativecloud/business-plans.html', styles: false },
+  { m7: '/creativecloud/education-plans.html', styles: false },
```
overmyheadandbody, line 115: Is it certain these links will always contain the `.html` part? Milo logic adds that suffix depending on the consumer's config, so some will have it, some will not. Is it a given that these particular links always contain it? L14 from `m7.js` also doesn't check for the `.html` suffix.

overmyheadandbody (commented):
Looking good, just a couple of questions.

libs/utils/utils.js
bozojovicic: We used to have empty file `libs/blocks/mas-autoblock/mas-autoblock.css` but I removed it in this PR. That's why I set `styles: false` here

libs/utils/utils.js
bozojovicic: Yes we are absolutely sure. These are external links 
https://www.adobe.com/creativecloud/business-plans.html
and
https://www.adobe.com/creativecloud/education-plans.html
If you remove the extension you will get 404.

bozojovicic: Just tested this redirection on prod page

Screenshot 2025-03-14 at 14 49 36

libs/utils/utils.js
overmyheadandbody: Is it deleted? I still see it at https://github.com/adobecom/milo/blob/d59d2ed9ebe9f9e202568eeb8fc94cfe4660fb7e/libs/blocks/mas-autoblock/mas-autoblock.css 🤔

libs/utils/utils.js
overmyheadandbody: Great, thanks for the confirmation!

3ch023 (changes requested):
tried on test page - doesn't seem to work.
also could you create a test page in /africe to see how it works there?

npeltier (approved):

3ch023 (approved):

3ch023:
@bozojovicic added metadata & i checked my umbrella country concerns here 
https://mwpw166076m7--milo--bozojovicic.aem.page/africa/drafts/mili/m7

looks good to go

libs/blocks/m7/m7.js
```diff
@@ -0,0 +1,22 @@
...
+  m7link.searchParams.append('co', country);
+  m7link.searchParams.append('pa', paCode);
+  if (href.includes('/creativecloud/education-plans')) {
+    m7link.searchParams.append('ms', 'EDU');
+  }
+  return m7link.toString();
+}
+
+export default function init(el) {
+  el.href = generateM7Link(el.href);
```
yesil, line 21: would be nice to have try catch and lana.log as a safeguard.

yesil (approved):

libs/blocks/m7/m7.js
bozojovicic: Done.

Roycethan (changes requested):
@bozojovicic We need [&af=uc_segmentation_hide_tabs] as well to default commerce view with hidden tabs

(ex: https://commerce.adobe.com/store/segmentation?cli=adobe_com&co=US&pa=ccsn_direct_individual&cs=t&af=uc_segmentation_hide_tabs)

cc @3ch023 @npeltier @yesil

Roycethan:
@bozojovicic Similar to titan where the Users IMS country profile is not getting passed when signed in
  with noronha+photoshopuk@adobetest.com/Adobe123# (co=UK) - Was not able to make it work it here , Can you plz check this in Milo as well ?

Actual: https://commerce.adobe.com/store/segmentation?cli=creative&pa=ccsn_direct_individual&cs=t&co=US&lang=en
Expected: https://commerce.adobe.com/store/segmentation?cli=creative&pa=ccsn_direct_individual&cs=t&co=GB&lang=en

Roycethan:
> @bozojovicic Similar to titan where the Users IMS country profile is not getting passed when signed in with [noronha+photoshopuk@adobetest.com](mailto:noronha+photoshopuk@adobetest.com)/Adobe123# (co=UK) - Was not able to make it work it here , Can you plz check this in Milo as well ?
> 
> Actual: https://commerce.adobe.com/store/segmentation?cli=creative&pa=ccsn_direct_individual&cs=t&co=US&lang=en Expected: https://commerce.adobe.com/store/segmentation?cli=creative&pa=ccsn_direct_individual&cs=t&co=GB&lang=en

Thanks @bozojovicic  Post commits the issue with IMS country is fixed.
image

Roycethan (approved):

PR #3726, opened 2025-02-21
MWPW-166176 - Stop initial tabs click propagation

rgclayton:
* Stops tabs block initial click event from bubbling up.
* There should be no impact to the tabs block functionality or visually. 

**Context:**
After signing in (IMS sign in) if the tabs block is on a page and a user quickly navigates to the sign in to switch profiles, the signed in dropdown menu is closed due to tabs block loading latter on the page and firing a click event to inititalize.

Resolves: [MWPW-166176](https://jira.corp.adobe.com/browse/MWPW-166176)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/rclayton/tabs-stop-bubbling?martech=off
- After: https://tabs-stop-click-bubbling--milo--adobecom.hlx.page/drafts/rclayton/tabs-stop-bubbling?martech=off

**CC Stage URLs**
- Before: https://stage--cc--adobecom.hlx.page/creativecloud?martech=off
- After: https://stage--cc--adobecom.hlx.page/creativecloud?milolibs=tabs-stop-click-bubbling&martech=off

**BACOM URLs**
- Before: https://main--bacom--adobecom.hlx.page/drafts/rclayton/tabs-stop-bubbling?milolibs=tabs-stop-click-bubbling&martech=off
- After: https://main--bacom--adobecom.hlx.page/drafts/rclayton/tabs-stop-bubbling?milolibs=tabs-stop-click-bubbling&martech=off

**Testing notes:**
- Use browser dev tools to throttle network to 4G slow
- Load example page. If not signed into IMS sign in and reload page (Might need to be on VPN too).
- Load page and click on your signed in icon to display the drop down menu. 
- Wait a few seconds. The menu should **not** auto close. It will with the before link.
- Tabs block needs to be on the tested page.


libs/blocks/tabs/tabs.js
```diff
@@ -101,11 +101,19 @@ function getUniqueId(el, rootElem) {
...
 function configTabs(config, rootElem) {
   if (config['active-tab']) {
     const id = `#tab-${CSS.escape(config['tab-id'])}-${CSS.escape(getStringKeyName(config['active-tab']))}`;
     const sel = rootElem.querySelector(id);
-    if (sel) sel.click();
+    if (sel) {
+      sel.addEventListener('click', stopEventBubbling, false);
+      sel.click();
+      sel.removeEventListener('click', stopEventBubbling, false);
+    }
```
Ruchika4, line 116: Since we are removing the event right after the click can we use below code to trigger it once. Also for third argument `false` is the default value for event listener so I believe we can avoid passing it.

```suggestion
    if (sel) {
     sel.addEventListener('click', stopEventBubbling, { once: true });
     sel.click();
   }
```

chrischrischris, line 116: nit - In addition to @Ruchika4 change, inline one liner
```suggestion
    if (sel) {
      sel.addEventListener('click', (e) => e.stopPropagation(), { once: true });
      sel.click();
    }
```

rgclayton: Good eye! Done.

rgclayton: Thanks! Done.

rgclayton:
Nala tests are not failing for tabs on local.

Looks like errors with promo code and m@s... 

image

Ruchika4 (approved):

chrischrischris (approved):

PR #3738, opened 2025-02-25
MWPW-162956: Setting accessibleLabel as sr-only

seanchoi-dev:
Resolves: [MWPW-162956](https://jira.corp.adobe.com/browse/MWPW-162956)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.live/drafts/mili/promo-placeholders?martech=off
- After: https://MWPW-162956--milo--adobecom.aem.live/drafts/mili/promo-placeholders?martech=off

**Consumer site to test:**
- Before: https://main--cc--adobecom.aem.live/products/photoshop/plans
- After: https://main--cc--adobecom.aem.live/products/photoshop/plans?milolibs=MWPW-162956

- Before: https://main--cc--adobecom.aem.page/es/products/special-offers
- After: https://main--cc--adobecom.aem.page/es/products/special-offers?milolibs=MWPW-162956

- Before: https://main--dc--adobecom.aem.live/acrobat/features
- After: https://main--dc--adobecom.aem.live/acrobat/features?milolibs=MWPW-162956


libs/styles/styles.css
```diff
@@ -742,7 +742,7 @@ header.global-navigation + .feds-localnav {
   height: var(--global-height-navPromo);
 }
 
-.feds-sr-only {
+sr-only, .feds-sr-only {
```
yesil, line 745: this should move to global.css.js from MAS otherwise, outside Milo, `sr-only` will be shown to MAS consumers.

libs/features/mas/src/price/template.js
```diff
@@ -136,11 +137,10 @@ function renderContainer(
         taxInclusivityLabel,
         true,
     );
-
-    return renderSpan(cssClass, markup, {
+    const regularlySRLabel = accessibleLabel ? `<sr-only class="strikethrough-aria-label">${accessibleLabel}</sr-only>` : '';
```
yesil, line 140: can we instead do at the line 127:
let markup = accessibleLabel ? `<sr-only class="strikethrough-aria-label">${accessibleLabel}</sr-only>` : '';

libs/features/mas/src/inline-price.js
```diff
@@ -292,6 +294,49 @@ export class InlinePrice extends HTMLSpanElement {
         if (offers.length) {
             if (this.masElement.toggleResolved(version, offers, options)) {
                 this.innerHTML = service.buildPriceHTML(offers, options);
+
+                // Adding logic for <sr-only>Alternatively at</sr-only>
+                const htmlPattern = /<\/?[^>]+(>|$)/g;
+                async function formatLiteral(key, parameters) {
```
yesil, line 300: `formatLiteral` is already in `template.js`, please do not duplicate it.
`inline-price` is used only in the Milo implementation.
Dexter, and current Offer Selector Tool will not run this logic while rendering a price.
Please find way to move this logic to template.js

libs/features/mas/src/price/index.js
```diff
@@ -15,6 +15,9 @@ const priceOptical = createPriceTemplate({
 const priceStrikethrough = createPriceTemplate({
     displayStrikethrough: true,
 });
+const priceAlternative = createPriceTemplate({
```
yesil, line 18: instead of creating `priceAlternative`, can we

```
const priceStrikethrough = createPriceTemplate({
    displayStrikethrough: true,
    alternativePrice: true|false
});
```

`alternativePrice` could be an option set in inline-price.js based on some attributes or offer data.

Also, likely we need to add this to some of other price rendering function, maybe `pricePromoWithAnnual` ?

happy to discuss this.

yesil (changes requested):
Overall I like the direction, but I would like to consolidate all price formatting logic in `price/template.js`

libs/styles/styles.css
seanchoi-dev: But not all pages have global.css.js either.
For example: https://main--cc--adobecom.aem.live/products/photoshop/plans?milolibs=MWPW-162956
image

Do you have a better place to add it?
(Maybe I should add both places?)

libs/features/mas/src/price/index.js
seanchoi-dev: This is actually a left over from my initial code that should have removed.
However if we want to put the logic into the template.js according to your comment above, we may need this.

We can definitely discuss but I already went through the inline-price's attributes and they don't seem to be a good flag to cover all use cases.
I am not sure how to interpret offer data.

libs/features/mas/src/inline-price.js
seanchoi-dev: The reason I put the logic in inline-price is
1. I needed a place to check the price element's parent so it can decide either the price is priceStrikethrough's sibling.
2. Mariia told me all prices are inline-price.

At template.js level, I couldn't find a good way to check the 1. above.

libs/styles/styles.css
yesil: You are right, global.css.js comes with a merch card.
As you suggested, let's keep it both in styles.css and in global.css.js

libs/features/mas/src/inline-price.js
yesil: Idea is the following: keep logic in template.js, add a new flag allowing to render the markup you need.
then, in inline-price.js, you can toggle an attribute and cause a re-render with the flag value correctly set. then template.js would output the markup you expect. You can also pass values to template.js as part of options.
But all the price markup generation logic should stay in template.js.

seanchoi-dev: I was thinking very same thing initially but I could not find a good way.

Here is the logic I made currently:
1. Build a inline-price HTML via service.buildPriceHTML(offers, options).
2. See if the inline-price has more than one price and it has .price-strikethrough then add the sr-only for the sibling of the one that has .price-strikethrough
3. If it isn't 2. case, find the price's parent element (.col, merch-card, or p) and check same thing as 2. from the parent element.

Since template.js happens in 1., I can't run the logic for checking if the container (inline-price obj or its parent element) has .price-strikethrough and its sibling. In other word, the logic should run after template.js completed.

Therefore, to adapt your idea, we need **new logic for a flag to send to template.js**. And I couldn't come up with any idea. :(

yesil: AFAIK there are two cases:
1. promo price:
in template.js#createPromoPriceTemplate you know that this is a promo price and you should be able pass the right params to createPriceTemplate so that it renders as you would except.

image


3. one strikethrough and one normal price that should be considered as promo:
here you can check if  this is not a strikethrough and the closest `P` contains a strikethrough, then  you can call `this.render({actualPrice:true})`. Thus, `actualPrice` will become a flag you can use in template.js.
All the price related markup should be generated in template.js.

libs/features/mas/src/price/index.js
```diff
@@ -18,6 +18,9 @@ const priceStrikethrough = createPriceTemplate({
 const priceAnnual = createPriceTemplate({
     displayAnnual: true,
 });
+const priceAlternative = createPriceTemplate({
+    isAlternativePrice: true,
```
yesil, line 22: I wonder if we need to pass also 
```
displayOptical = false,
displayStrikethrough = false,
displayAnnual = false,
```

libs/features/mas/src/inline-price.js
```diff
@@ -292,6 +294,18 @@ export class InlinePrice extends HTMLSpanElement {
         if (offers.length) {
             if (this.masElement.toggleResolved(version, offers, options)) {
                 this.innerHTML = service.buildPriceHTML(offers, options);
+
+                // Adding logic for options.alternativePrice to add <sr-only>Alternatively at</sr-only>
+                let parentEl = this.closest('p');
+                if (!parentEl || parentEl.querySelectorAll('span[is="inline-price"]').length < 2) parentEl = this.closest('.col') ?? this.closest('merch-card');
```
yesil, line 300: `this.closest('.col') ` is a milo selector and MAS doesn't know about Milo. I would avoid adding Milo specific selectors here. I find even `this.closest('merch-card')` dangerous because some cards may have multiple prices of different plan types.
I think we should stick to `p`.

If we have to support the `.col` selector, I propose to add a new attribute to inline-price at https://github.com/adobecom/milo/blob/stage/libs/features/mas/src/inline-price.js#L87-L95

and set the attribute from merch.js where it knows about Milo specific. It can set the attribute.

yesil (commented):
Thanks @seanchoi-dev. Overall it looks very good to me, except for the tests.
would be nice to add test cases for this new feature.

libs/features/mas/src/price/index.js
seanchoi-dev: We don't need since they are false in default. 
image

libs/features/mas/src/inline-price.js
seanchoi-dev: - The logic is: the parent element is `this.closest('p')` If it is NOT what we're looking for trying `.col` then `merch-card`. So, the .col and merch-card is basically backup plan and I NEED them since there are use case like (rarely) each inline-price has its own `p` parent in a .col. OR don't even have any `p` for their parent.
image (https://main--cc--adobecom.aem.page/es/products/photoshop/plans)

I think having backup plan isn't a dangerous move and without this, it won't cover all use cases at least for Milo.

- If you still suggesting having new attribute to detect milo specific flag, I can do that but I think it is not much beneficial logically and it will make the code more complecated.
It is like
Having backup plan for Milo specific use case VS Having attribute for milo specific flag.
I assume using attribute will be like:
1. Add `data-platform="milo"` into `observedAttributes`
2. Add it in merch.js. 
3. Test to make sure the merch.js's job is always finished before the checking parent's logic in inline-price.js
4. Use the `data-platform="milo"` as flag and if it is milo, consider .col and merch-card as backup plan for its parent element.

Even if we have the attribute they will be still backup plan of P. Only benefit I can think of for having the attribute is have that backup plan only if it is milo but this part isn't real benefit because if there is no matching parentEl, it will return immediately anyway.

libs/features/mas/src/price/index.js
yesil: but since you pass `{isAlterntivePrice}` afaik every other parameter will be set to undefined.

libs/features/mas/src/inline-price.js
yesil: since merch-card is managed by us, we know what to expect in terms of markup, for `p` and `.col` I was thinking something like in `merch.js`:

then you can use the property `a11yParent`  (actually we don't need an attribute to keep it simple)
image

if `a11yParent` is not set, we can fallback it to `p` element as in your code.

I think that would cover both Milo specific use case, and keep M@S relatively agnostic yet configurable with good defaults.

seanchoi-dev: Assigning an object to merch at this stage doesn't work. 
I tested following:
image
I think the object becomes `undefined` when it gets built in MAS.
But setting dataset works. I will use that instead.

libs/blocks/merch/merch.js
```diff
@@ -759,6 +759,10 @@ export default async function init(el) {
   const isCta = searchParams.get('type') === 'checkoutUrl';
   const merch = await (isCta ? buildCta : buildPrice)(el, searchParams);
   const service = await initService();
+  if (searchParams.get('type') === 'price') {
+    const a11yParent = el.closest('.col, merch-card, p, div');
```
yesil, line 763: isn't `p` the most comment use case?  IMO the most common one should be at the first place in the selector

seanchoi-dev: I was thinking P should be more as backup but I think you're right.

seanchoi-dev: Actually, this is kinda pointless.
When this is built. the element isn't fully built. Even if I am checking .col or merch-card, they aren't there yet and their classes aren't ready either. Therefore their nodeName will be always either P or DIV. Which works for most cases but this isn't definitely a bullet proof code.

I wrote a test code already but I think I have to revisit this some hours later.

libs/features/mas/src/price/index.js
3ch023: same as for templates above, no?
i guess it works fine

3ch023 (approved):

libs/features/mas/src/price/index.js
seanchoi-dev: I tested and they are still `false` as default value. So, I think it's good as it is.
image
image

yesil (approved):
Looks very good to me, thanks for addressing all the feedbacks :)

Roycethan (changes requested):
[~seanchoi] Can you plz review this , I see Optical pricing used for regular price and for this template when used the screen reader doesn't read "alternatively at" or it doesn't have sr-only tag.
ex: https://main--cc--adobecom.aem.live/creativecloud/buy/students?milolibs=MWPW-162956#

Roycethan (changes requested):
@seanchoi-dev Post fix although it works for US and locales , 'AU' locale with Annual pricing sas still this problem , 
ex: https://main--cc--adobecom.aem.live/au/creativecloud/buy/students?milolibs=MWPW-162956#
You can see both Marquee and Bottom card here.. 
image

Roycethan:
@seanchoi-dev Alternative price tag set on Annual template now, Can you also plz check this case for alternative price where its not set ? 
https://main--dc--adobecom.aem.live/acrobat/features?milolibs=MWPW-162956
image

Roycethan:
> @seanchoi-dev Alternative price tag set on Annual template now, Can you also plz check this case for alternative price where its not set ? https://main--dc--adobecom.aem.live/acrobat/features?milolibs=MWPW-162956 image

This is fixed.

Roycethan:
@seanchoi-dev 
I see a Strikethrough pricing - where placeholder-failed and price is not getting loaded in your branch : https://main--cc--adobecom.aem.live/creativecloud/buy/students?milolibs=MWPW-162956
Not found any perticular reason but its loading  in https://main--cc--adobecom.aem.live/creativecloud/buy/students
image

seanchoi-dev:
> @seanchoi-dev I see a Strikethrough pricing - where placeholder-failed and price is not getting loaded in your branch : https://main--cc--adobecom.aem.live/creativecloud/buy/students?milolibs=MWPW-162956 Not found any perticular reason but its loading in https://main--cc--adobecom.aem.live/creativecloud/buy/students image

Fixed.

Roycethan:
> https://main--cc--adobecom.aem.live/creativecloud/buy/students?milolibs=MWPW-162956

Thanks, Fixed

Roycethan (approved):

PR #3741, opened 2025-02-26
Mwpw 164462 marquee container

zagi25:
* Added `marquee-container` block, this is a dummy block currently only used for MEP targeting
* Added simple `marquee-container` selector to personalization (MEP)
* Excluded `marquee-container` from masonry, to prevent `section-metadata` from messing up the layout if it executes before `marquee-container` is removed.

Resolves: [MWPW-164462](https://jira.corp.adobe.com/browse/MWPW-164462)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/ratko/marquee-container/demo/marquee-container-demo?martech=off
- After: https://mwpw-164462-marquee-container--milo--adobecom.hlx.page/drafts/ratko/marquee-container/demo/marquee-container-demo?martech=off


chrischrischris (changes requested):
This doesn't seem like the best thing for performance.  It's a small hit, but means we're loading 2 extra files that do nothing on every page load.

It seems that this could easily be achieved using section-metadata instead?  Really this is just being used to be able to label a marquee right?  Instead if the author makes sure to only place the content to be swapped in one section then it can be targeted by MEP in the same way as this.  Or maybe I'm missing something?

zagi25:
@chrischrischris we did performance validation [report](https://main--milo--adobecom.hlx.page/drafts/ratko/marquee-container/performance/) , and there was no performance hit when comparing one block to the multiple blocks with the addition of `marquee-container`. 
Consonant wanted an easier and more precise way to target a section that acts as `marquee`  using MEP. 
Also it is possible that in the future there will be a need for some extra functionality and some specific styling for the blocks inside the `marquee-container` section (e.g. different `font-size` for `text` block), we can just add that to the `marquee-container.js` and `marquee-container.css` without affecting the rest of the page.

overmyheadandbody:
Hey @chrischrischris! As @zagi25 already mentioned, we did look into the performance impact of this approach. The issue with the metadata approach is that you can't target that with MEP, as a specific class gets added through decoration logic, it's not there to begin with. Using an nth section approach can also lead to issues, as not all pages have this sort of marquee as the first element, thus we need to be sure we're targeting the correct section containing these block combinations.

This combination marquee will be used on a subset of pages and usually in the context of personalization, where we'd need to swap the entire section with something else. If you have alternative solutions, we're all ears!

The initial proposal was for a completely new block that could combine multiple blocks, but that approach would've led to a lot more overhead than the current proposal from a code, maintenance, performance and authoring perspective.

> Really this is just being used to be able to label a marquee right?
Not entirely. We want to target marquee-like sections that don't contain any marquee/hero-marquee blocks. Authors are already using this pattern across multiple pages, we're trying to provide the least invasive way for them to be able to target those marquee-like sections through MEP.

chrischrischris (commented):
Thanks for the explanations.

I still think it may be better to add support in MEP to target sections by specifying property values ... not sure though.

chrischrischris (approved):

libs/utils/utils.js
```diff
@@ -1413,7 +1414,7 @@ export function partition(arr, fn) {
 const preloadBlockResources = (blocks = []) => blocks.map((block) => {
   if (block.classList.contains('hide-block')) return null;
   const { blockPath, hasStyles, name } = getBlockData(block);
-  if (['marquee', 'hero-marquee'].includes(name)) {
+  if (['marquee', 'hero-marquee', 'marquee-container'].includes(name)) {
```
mokimo, line 1417: Instead of preloading, can't we just not load the block & it's CSS? We can still use it as a marker, but IMO it doesn't make sense to load files for something we don't use

libs/blocks/marquee-container/marquee-container.js
```diff
@@ -0,0 +1,3 @@
+export default function init(el) {
```
mokimo, line 1: We already support this in the `loadBlock` function in utils, by adding a `hide-block` class... does anything speak against always appending that class so that we don't need to load empty files?

```
// utils.js
  if (block.classList.contains('hide-block')) {
    block.remove();
    return null;
  }
```

libs/utils/utils.js
overmyheadandbody: We're not using the CSS file right now, but we will get Consonant updates on how a few blocks should look like when inside this combination marquee.

Nevertheless, @zagi25 is looking into @chrischrischris' suggestion of trying to target a section in MEP based on its Section Metadata properties.

libs/features/personalization/personalization.js
```diff
@@ -448,6 +449,10 @@ function modifySelectorTerm(termParam) {
     term = updateEndNumber(endNumber, term);
     return term;
   }
+  if (containerSelectors.includes(startText)) {
```
robert-bogos, line 452: If we just check for `'marquee-container'` maybe we shouldn't use an array 🤔
```suggestion
  if (startText === 'marquee-container') {
```
P.S. You'll also save 1 line like this

libs/features/personalization/personalization.js
```diff
@@ -448,6 +449,10 @@ function modifySelectorTerm(termParam) {
...
+    term = `main > div:nth-child(${endNumber || 'n'} of main > div:has([class*="${startText}"])`;
+    return term;
```
robert-bogos, line 454: In here I think you can just return the expression directly, there's no utility in caching it.
```suggestion
    return `main > div:nth-child(${endNumber || 'n'} of main > div:has([class*="${startText}"])`;
```

test/blocks/marquee-container/marquee-container.test.js
```diff
@@ -0,0 +1,17 @@
...
+import { expect } from '@esm-bundle/chai';
+
+document.body.innerHTML = await readFile({ path: './mocks/body.html' });
+describe('Marquee container', () => {
+  before(async () => {
+    const module = await import('../../../libs/blocks/marquee-container/marquee-container.js');
+    const containers = document.body.querySelectorAll('.marquee-container');
+    containers.forEach((container) => {
+      module.default(container);
+    });
```
robert-bogos, line 11: This can be simpler 😊
```suggestion
    document.body.querySelectorAll('.marquee-container').forEach((container) => module.default(container));
```

test/blocks/marquee-container/marquee-container.test.js
```diff
@@ -0,0 +1,17 @@
...
+  });
+  it('Removes marquee-container block', async () => {
+    const container = document.querySelector('.marquee-container');
+    expect(container).to.equal(null);
```
robert-bogos, line 15: Same here 😊
```suggestion
    expect(document.querySelector('.marquee-container')).to.equal(null);
```

test/blocks/section-metadata/section-meta.test.js
```diff
@@ -56,6 +56,14 @@ describe('Section Metdata', () => {
     expect(sec.classList.contains('masonry-layout')).to.be.true;
   });
 
+  it('Exclude marquee-container from masonry', async () => {
+    const sec = document.querySelector('.section.masonrysec');
+    const sm = sec.querySelector('.section-metadata');
+    await init(sm);
+    const container = sec.querySelector('.marquee-container');
+    expect(container.classList.contains('grid-full-width')).to.be.false;
```
robert-bogos, line 64: Same here
```suggestion
    expect(sec.querySelector('.marquee-container').classList.contains('grid-full-width')).to.be.false;
```

vgoodric (changes requested):
Can we have a meeting on this or start a Slack conversation to better explain the use case for this feature?

vgoodric:
So I put together a demo of how I would tackle this with existing functionality:
1. copied the folder with the demo files for this PR
2. removed the new block from my version
3. added a variant to the section metadata blocks to give them a pseudo id
image
4. Updated the selectors to use "has".  Example: `section:has(.marquee-container-1)`

Sample page: https://main--milo--adobecom.hlx.page/drafts/mepdev/2024/q4/marquee-container/marquee-container-demo
You can do a side by side with https://mwpw-164462-marquee-container--milo--adobecom.hlx.page/drafts/ratko/marquee-container/demo/marquee-container-demo?martech=off (I went 1 by 1 through the variants)
Page source file: https://adobe.sharepoint.com/:w:/r/sites/adobecom/_layouts/15/Doc.aspx?sourcedoc=%7BA91AB2D5-6BBB-432B-93E9-3A03E6E864AB%7D&file=marquee-container-demo.docx&action=default&mobileredirect=true
Manifest source file: https://adobe.sharepoint.com/:x:/r/sites/adobecom/_layouts/15/Doc.aspx?sourcedoc=%7B3EC6D95E-BC38-4E0B-B571-DE35326FBD0E%7D&file=personalization.xlsx&action=default&mobileredirect=true

Main advantage to this system is you could apply it to anything, not just marquee content.  If you have a section you want to easily target, we already use "has" and target a unique block in the section. But if there's nothing that unique about the contents and/or you want to attach a more durable selector, adding a variant that's unique to something in the section will do the trick. 

As for content specific styling on a section, any new fragment can and likely will have it's own section-metadata and that can add the specific styling.

zagi25:
New PR raised for this issue https://github.com/adobecom/milo/pull/3779

PR #3752, opened 2025-02-27
[MWPW-166184] - Pill tab color change

DKos95:
This PR adjusts colors for pill tabs to be in line with the updated design.

Resolves: [MWPW-166184](https://jira.corp.adobe.com/browse/MWPW-166184)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/docs/library/kitchen-sink/tab-block?martech=off
- After: https://pill-tab-color--milo--adobecom.aem.page/docs/library/kitchen-sink/tab-block?martech=off

### DIFF
--- a/libs/blocks/tabs/tabs.css
+++ b/libs/blocks/tabs/tabs.css
@@ -7,19 +7,13 @@
   --tabs-active-text-color: #2c2c2c;
   --tabs-bg-color: #f1f1f1;
   --tabs-active-bg-color: #fff;
-  --tabs-list-bg-gradient: linear-gradient(
-    to bottom, 
-    rgba(136, 136, 136, 0) 0%,
-    rgba(136, 136, 136, 0) 60%,
-    rgba(136, 136, 136, 0.03) 80%,
-    rgba(136, 136, 136, 0.08) 100%
-  );
-  --tabs-pill-bg-color: #e1e1e1;
-  --tabs-pill-bg-color-hover: #dadada;
-  --tabs-pill-bg-color-active: #c6c6c6;
-  --tabs-mobile-stacked-bg-color: #242424;
-  --tabs-mobile-stacked-text-color: #ffffff;
+  --tabs-list-bg-gradient: linear-gradient(rgb(255 255 255 / 0%) 60%, rgb(245 245 245 / 80%));
+  --tabs-pill-bg-color: #292929;
+  --tabs-pill-bg-color-hover: #e6e6e6;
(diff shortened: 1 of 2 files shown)


libs/blocks/tabs/tabs.css
```diff
@@ -321,17 +327,16 @@
...
 .tabs[class*='pill'] div[role="tablist"] button:focus {
-  color: var(--tabs-pill-text-color);
-  background: var(--tabs-pill-bg-color-hover);
+  color: var(--tabs-pill-text-focus-color);
+  background: var(--tabs-pill-bg-focus);
 }
 
 .tabs[class*='pill'] div[role="tablist"] button:focus-visible {
   outline: 2px solid #4145ca;
-  outline-offset: 2px;
```
meganthecoder, line 330: I'm surprised that the accessibility team wants you to get rid of the offset since I'm in the middle of adding it to the accordion at their request. Do you know if this is a one-off or if it's going to be a milo-wide change?

libs/blocks/tabs/tabs.js
```diff
@@ -65,6 +65,7 @@ function changeTabs(e) {
     window.location.assign(redirectionUrl);
     return;
   }
+  this.blur();
```
meganthecoder, line 68: It'd be good to have this called out in the PR description or title since it affects more the just the pill variant.

meganthecoder (approved):
Looks good! Just a couple small comments.

robert-bogos (approved):

libs/blocks/tabs/tabs.css
DKos95: Nice catch, the design we got for pill tabs has no outline offset that's why I removed it, it's in one of the last comments on the ticket, not sure if its milo wide. I will bring this up in a team meeting to discuss wether it's good for accessibility or not considering you are adding it currently in your task, thanks!

libs/blocks/tabs/tabs.js
DKos95: Thanks for pointing this out, I did a little bit more digging and found that this isn't good practice for accessibility to switch focus off the selected tab, so I changed the css a bit to do what I wanted which was essentially that the focus color shouldn't have precedence over the selected color when you select a tab, this was asked in one of our meetings.

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-166184

PR #3753, opened 2025-02-27
feat(MWPW-168628): Add support for loading CaaS alpha releases via query parameter

sanrai:
### Summary  
This PR introduces a new query parameter value (`alpha`) that allows loading the alpha version of CaaS assets directly from GitHub Pages. This enables automated testing of pre-release CaaS code before it is merged and deployed.  

### Changes  
- Check for the `alpha` query parameter value in `loadCaasFiles`.  
- If present, update `cssFile` and `jsFile` to point to the latest alpha assets on GitHub Pages.  
- This allows testing new CaaS code without requiring a full release.  

### Why?  
Currently, there's no easy way to test pre-PR CaaS code before release. By adding this query parameter, we can pull the alpha version of CaaS and validate changes via automation before merging into production.  

### Testing  
- Append `?caasver=alpha` to the URL and verify that the CaaS alpha assets are loaded.  
- Ensure the existing functionality remains unchanged when the query parameter is not present.  

Resolves: [MWPW-168628](https://jira.corp.adobe.com/browse/MWPW-168628)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/sanrai/foo?martech=off
- After: https://MWPW-168628--milo--adobecom.aem.page/drafts/sanrai/foo?martech=off

### DIFF
--- a/libs/blocks/caas/utils.js
+++ b/libs/blocks/caas/utils.js
@@ -211,6 +211,12 @@ export const loadCaasFiles = async () => {
     jsFile = `http://${host}.corp.adobe.com:5000/dist/main.js`;
   }
 
+  // for caas alpha releases
+  if (version === 'alpha') {
+    cssFile = 'https://adobecom.github.io/caas/dist/app.css';
+    jsFile = 'https://adobecom.github.io/caas/dist/main.source.js';
+  }
+
   loadStyle(cssFile);
   await loadScript(`https://www.adobe.com/special/chimera/caas-libs/${version}/react.umd.js`);
   await loadScript(`https://www.adobe.com/special/chimera/caas-libs/${version}/react.dom.umd.js`);


libs/blocks/caas/utils.js
```diff
@@ -211,6 +211,13 @@ export const loadCaasFiles = async () => {
     jsFile = `http://${host}.corp.adobe.com:5000/dist/main.js`;
   }
 
+  // for caas beta releases
+  const beta = searchParams?.get('caasbeta');
+  if (beta) {
```
cmiqueo, line 216: Not a big deal, but since params are passed as strings, you may want to change this to:
`if (beta === 'true') `

cmiqueo (approved):

sheridansunier (approved):

shkhan91 (approved):

PR #3759, opened 2025-02-28
MWPW-167748 Fix focus in georouting modal with tabs

meganthecoder:
* Fix focus in georouting modal with tabs

Resolves: [MWPW-167748](https://jira.corp.adobe.com/browse/MWPW-167748)

**Test URLs:**
- After: https://geo-focus--milo--meganthecoder.aem.page/?martech=off

Bacom:
- https://main--bacom--adobecom.aem.live/?milolibs=geo-focus--milo--meganthecoder&akamaiLocale=IN

### DIFF
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -3,7 +3,7 @@
 import { createTag, getMetadata, localizeLink, loadStyle, getConfig } from '../../utils/utils.js';
 import { decorateSectionAnalytics } from '../../martech/attributes.js';
 
-const FOCUSABLES = 'a:not(.hide-video), button:not([disabled], .paddle), input, textarea, select, details, [tabindex]:not([tabindex="-1"])';
+const FOCUSABLES = 'a:not(.hide-video), button:not([disabled], .locale-modal-v2 .paddle), input, textarea, select, details, [tabindex]:not([tabindex="-1"])';
 const CLOSE_ICON = `<svg aria-hidden="true" xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 20 20">
   <g transform="translate(-10500 3403)">
     <circle cx="10" cy="10" r="10" transform="translate(10500 -3403)" fill="#707070"/>
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -3,7 +3,7 @@
 import { createTag, getMetadata, localizeLink, loadStyle, getConfig } from '../../utils/utils.js';
 import { decorateSectionAnalytics } from '../../martech/attributes.js';
 
-const FOCUSABLES = 'a:not(.hide-video), button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])';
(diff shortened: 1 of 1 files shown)


vhargrave (commented):
@meganthecoder I can't even reproduce the original issue. When I open the georouting and press tab it immediately starts tabbing within the modal. Am I just blindly missing something?

DKos95 (approved):

DKos95:
> @meganthecoder I can't even reproduce the original issue. When I open the georouting and press tab it immediately starts tabbing within the modal. Am I just blindly missing something?

I think I had the same issue to notice it because I was clicking on the url in the task and it is missing the locale param(?akamaiLocale=IN) then when tabbing it goes to the languages, but if you have a local param it goes to the logo when tabbing and then when you add milolibs=geo-focus--milo--meganthecoder along with the locale param it seems fixed to me.

- https://main--bacom--adobecom.aem.live/?akamaiLocale=IN
- https://main--bacom--adobecom.aem.live/?milolibs=geo-focus--milo--meganthecoder&akamaiLocale=IN

JasonHowellSlavin (approved):

spadmasa:
Hi @meganthecoder , As part of this ticket we have https://jira.corp.adobe.com/browse/MWPW-168432 was checking on the fix, I am on India Connected to Bangalore VPN, able to get geo popup by default.When tested with this url still see the issue. Can you please advise
https://main--cc--adobecom.hlx.live/creativecloud?milolibs=geo-focus--milo--meganthecoder
image

vhargrave:
> Hi @meganthecoder , As part of this ticket we have https://jira.corp.adobe.com/browse/MWPW-168432 was checking on the fix, I am on India Connected to Bangalore VPN, able to get geo popup by default.When tested with this url still see the issue. Can you please advise https://main--cc--adobecom.hlx.live/creativecloud?milolibs=geo-focus--milo--meganthecoder image

I have the same issue as well
@meganthecoder i took a brief look and what's interesting is that the CC modal is appearing into view after the focus is set, whereas for bacom it's appearing before. Since focus isn't set for elements that are not visible, I think that might be part of the issue for CC. Bacom works fine.

JasonHowellSlavin (approved):

libs/blocks/modal/modal.js
```diff
@@ -3,7 +3,7 @@
 import { createTag, getMetadata, localizeLink, loadStyle, getConfig } from '../../utils/utils.js';
 import { decorateSectionAnalytics } from '../../martech/attributes.js';
 
-const FOCUSABLES = 'a:not(.hide-video), button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])';
+const FOCUSABLES = 'a:not(.hide-video), button:not([disabled], .paddle), input, textarea, select, details, [tabindex]:not([tabindex="-1"])';
```
vhargrave, line 6: What do you think of making this select logic a little more specific, just so you're not disabling the focus for .paddles other modal use cases. `button:not([disabled], .locale-modal-v2 .paddle)`

vhargrave:
@meganthecoder unit tests are failing as well for some reason btw.

libs/blocks/modal/modal.js
meganthecoder: @vhargrave Updated!

vhargrave (approved):

PR #3761, opened 2025-03-03
[Loc][MWPW-168684] Single Language Rollout

maagrawal16:
* Added Single Language Rollout.
* Fixed bug for [MWPW-168679](https://jira.corp.adobe.com/browse/MWPW-168679)

Resolves: [MWPW-168684](https://jira.corp.adobe.com/browse/MWPW-168684)

**Test URLs:**
- Before: https://milostudio-stage--milo--adobecom.aem.page?martech=off
- After: https://mwpw-168684-single-language-rollout--milo--maagrawal16.aem.page?martech=off

### DIFF
--- a/libs/blocks/locui-create/locui-create.js
+++ b/libs/blocks/locui-create/locui-create.js
@@ -33,11 +33,12 @@ function Create() {
         const searchParams = new URLSearchParams(window.location.search);
         const projectKey = searchParams.get('projectKey');
         const workflow = searchParams.get('workflow');
+        const language = searchParams.get('language');
         const projectInitByUrl = getProjectByParams(searchParams);
 
         /* Fetch draft project if project key is found in url params */
         if (projectKey) {
-          setUserWorkflowType(workflow);
+          setUserWorkflowType(language ? 'promoteRollout' : workflow);
           const error = await fetchDraftProject(projectKey);
           if (error) {
             setToast({ type: 'error', message: error });
--- a/libs/blocks/locui-create/locui-create.js
+++ b/libs/blocks/locui-create/locui-create.js
@@ -33,12 +33,11 @@ function Create() {
         const searchParams = new URLSearchParams(window.location.search);
         const projectKey = searchParams.get('projectKey');
         const workflow = searchParams.get('workflow');
-        const language = searchParams.get('language');
         const projectInitByUrl = getProjectByParams(searchParams);
 
         /* Fetch draft project if project key is found in url params */
         if (projectKey) {
-          setUserWorkflowType(language ? 'promoteRollout' : workflow);
+          setUserWorkflowType(workflow);
           const error = await fetchDraftProject(projectKey);
           if (error) {
             setToast({ type: 'error', message: error });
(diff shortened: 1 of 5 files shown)


nkthakur48:
@maagrawal16 Looks like there are a bunch of `eslint` errors. Could you please have a look?

maagrawal16:
> @maagrawal16 Looks like there are a bunch of `eslint` errors. Could you please have a look?

I have fixed the mentioned errors, now lint check is passing.

libs/blocks/locui-create/locui-create.js
```diff
@@ -33,11 +33,12 @@ function Create() {
         const searchParams = new URLSearchParams(window.location.search);
         const projectKey = searchParams.get('projectKey');
         const workflow = searchParams.get('workflow');
+        const language = searchParams.get('language');
         const projectInitByUrl = getProjectByParams(searchParams);
 
         /* Fetch draft project if project key is found in url params */
         if (projectKey) {
-          setUserWorkflowType(workflow);
+          setUserWorkflowType(language ? 'promoteRollout' : workflow);
```
saurabhsircar11, line 41: Do we need to add this condition?, workflow is already promoteRollout in this case coming from search params

maagrawal16: We need to include this check because, in the single language rollout case, the workflow parameter is not provided only the language parameter is being passed from the details page.

maagrawal16: Okay, I rechecked in Varun's PR as well. It is getting passed. In that case I will remove this check! Thanks for pointing it out.

libs/blocks/locui-create/store.js
```diff
@@ -223,17 +226,26 @@ export async function fetchDraftProject(projectKey) {
     );
     const resJson = await response.json();
     if (response.ok) {
+      const projectNameProfix = lang ? `-${lang}` : '';
```
nkthakur48, line 229: Shouldn't this be `projectNameSuffix`?

libs/blocks/locui-create/store.js
```diff
@@ -223,17 +226,26 @@ export async function fetchDraftProject(projectKey) {
...
+      let languages;
+      if (lang) {
+        languages = getLanguageDetails(lang);
+      } else if (userWorkflowType.value === 'promoteRollout') {
```
nkthakur48, line 233: Since we have multiple instances in the file where 'promoteRollout` is used within conditional, it's better to move it as a constant

nkthakur48 (changes requested):
@maagrawal16 Could you please take care of review comments

libs/blocks/locui-create/store.js
maagrawal16: yes, I'll change that.

libs/blocks/locui-create/store.js
maagrawal16: Added this change.

libs/blocks/locui-create/store.js
```diff
@@ -223,17 +226,26 @@ export async function fetchDraftProject(projectKey) {
...
+        languages = resJson?.languages.map((language) => ({
```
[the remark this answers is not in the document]
maagrawal16: I have moved the logic in a separate function.

nkthakur48 (approved):

saurabhsircar11 (approved):

PR #3779, opened 2025-03-05
MWPW-164462 [MEP] Add ability for MEP to select sections by section-metadata anchor value 

zagi25:
* Expand `anchor` `section-metadata` property to MEP

`anchor` property adds `id` to the section
Screenshot 2025-03-05 at 14 37 45
Section is targeted with `id` selector
Screenshot 2025-03-05 at 14 37 53

Resolves: [MWPW-164462](https://jira.corp.adobe.com/browse/MWPW-164462)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/ratko/marquee-container/demo/marquee-container-demo?martech=off
- After: https://mwpw-164462-marquee-container2--milo--adobecom.aem.page/drafts/ratko/marquee-container/demo/marquee-container-demo?martech=off

[Manifest](https://adobe.sharepoint.com/:x:/r/sites/adobecom/_layouts/15/doc2.aspx?sourcedoc=%7BB12B9C30-FBF1-45F9-8689-835C919160A6%7D&file=personalization.xlsx&action=default&mobileredirect=true)

**NOTE:** Manifest option `param-five=true` looks broken because heading element is target and swapped (just wanted to show off targeting element inside specific section)

### DIFF
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -37,7 +37,6 @@ export const PERSONALIZATION_TAGS = {
 const PERSONALIZATION_KEYS = Object.keys(PERSONALIZATION_TAGS);
 /* c8 ignore stop */
 
-const MEP_ID = 'mep-id';
 const CLASS_EL_DELETE = 'p13n-deleted';
 const CLASS_EL_REPLACE = 'p13n-replaced';
 const COLUMN_NOT_OPERATOR = 'not ';
@@ -433,8 +432,6 @@ function modifySelectorTerm(termParam) {
     'div', 'a', 'p', 'strong', 'em', 'picture', 'source', 'img', 'h',
     'ul', 'ol', 'li',
   ];
-  const hasMepId = term.startsWith(MEP_ID);
-  term = hasMepId ? term.replace(`${MEP_ID}-`, '') : term;
   const startTextMatch = term.match(/^[a-zA-Z/./-]*/);
   const startText = startTextMatch ? startTextMatch[0].toLowerCase() : '';
   const startTextPart1 = startText.split(/\.|:/)[0];
@@ -451,7 +448,6 @@ function modifySelectorTerm(termParam) {
     term = updateEndNumber(endNumber, term);
     return term;
   }
-  if (hasMepId) return `main > div:has([${MEP_ID}*="${term}"])`;
 
   if (!startText.startsWith('.')) term = `.${term}`;
   if (endNumber) {
@@ -501,7 +497,7 @@ export function modifyNonFragmentSelector(selector, action) {
 
 function getSelectedElements(sel, rootEl, forceRootEl, action) {
   const root = forceRootEl ? rootEl : document;
-  const selector = sel.startsWith(MEP_ID) ? sel.split(' ').join('-').trim() : sel.trim();
+  const selector = sel.trim();
   if (!selector) return {};
 
   if (getSelectorType(selector) === 'fragment') {
@@ -565,14 +561,14 @@ export const deleteMarkedEls = (rootEl = document) => {
     .forEach((el) => el.remove());
 };
 
-export function addMepIdToSectionMetadata(rootEl = document) {
+export function addSectionIds(rootEl = document) {
   const metadataBlocks = rootEl.querySelectorAll('.section-metadata');
-  metadataBlocks.forEach((metadataBlock) => {
-    [...metadataBlock.children].forEach((child) => {
-      const col1 = child.children[0]?.textContent.toLowerCase().trim();
-      const col2 = child.children[1]?.textContent.toLowerCase().trim();
-      if (!col1 || !col2 || col1 !== MEP_ID) return;
-      metadataBlock.setAttribute(MEP_ID, col2);
+  metadataBlocks.forEach((block) => {
+    [...block.children].forEach((row) => {
+      const col1 = row.children[0]?.textContent.toLowerCase().trim();
+      const col2 = row.children[1]?.textContent.toLowerCase().trim().split(' ').join('');
+      if (!col1 || !col2 || col1 !== 'anchor') return;
+      block.parentElement.setAttribute('id', col2);
     });
   });
 }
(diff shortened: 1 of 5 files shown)


libs/features/personalization/personalization.js
```diff
@@ -561,13 +561,26 @@ export const deleteMarkedEls = (rootEl = document) => {
     .forEach((el) => el.remove());
 };
 
+export function addSectionAnchors(rootEl = document) {
+  const metadataBlocks = rootEl.querySelectorAll('.section-metadata');
+  metadataBlocks.forEach((block) => {
```
robert-bogos, line 566: nit:
```suggestion
  rootEl.querySelectorAll('.section-metadata').forEach((block) => {
```

robert-bogos (approved):

narcis-radu:
@zagi25 - is this still needed after https://github.com/adobecom/milo/pull/3714 has been implemented? I am able to see the anchor ID in the _before_ URL.

zagi25:
@narcis-radu , `personalization` is executed before `section-metadata` block, I just extended your solution so that `id` is added as part of `personalization` (if personalization is present) and then MEP can target section with `id` selector

narcis-radu:
@zagi25 - I understand, makes sense. We would only use that ID to target personalization, it's not needed for scrolling purposes.

narcis-radu (approved):

libs/features/personalization/personalization.js
```diff
@@ -561,13 +561,25 @@ export const deleteMarkedEls = (rootEl = document) => {
     .forEach((el) => el.remove());
 };
 
+export function addSectionAnchors(rootEl = document) {
+  rootEl.querySelectorAll('.section-metadata').forEach((block) => {
+    [...block.children].forEach((row) => {
+      const col1 = row.children[0]?.textContent.toLowerCase().trim();
+      const col2 = row.children[1]?.textContent.toLowerCase().trim().split(' ').join('');
```
chrischrischris, line 568: Why is the whitespace being removed?  I'd think either it gets replaced by `-` or we only use the first term and anything after whitespace is ignored.

zagi25: great catch, I would go with whitespace replaced with `-`, @narcis-radu I guess I should do the same in the [section-metadata.js](https://github.com/adobecom/milo/blob/stage/libs/blocks/section-metadata/section-metadata.js#L60) ?

narcis-radu: @zagi25 - yes

libs/features/personalization/personalization.js
```diff
@@ -561,13 +561,25 @@ export const deleteMarkedEls = (rootEl = document) => {
     .forEach((el) => el.remove());
 };
 
+export function addSectionAnchors(rootEl = document) {
+  rootEl.querySelectorAll('.section-metadata').forEach((block) => {
+    [...block.children].forEach((row) => {
+      const col1 = row.children[0]?.textContent.toLowerCase().trim();
+      const col2 = row.children[1]?.textContent.toLowerCase().trim().replaceAll(' ', '-');
+      if (!col1 || !col2 || col1 !== 'anchor') return;
+      block.parentElement.setAttribute('id', col2);
```
overmyheadandbody, line 570: We'd end up running similar logic [here](https://github.com/adobecom/milo/blob/stage/libs/blocks/section-metadata/section-metadata.js#L60) if PZN is enabled, but at the same time I'm not sure it's worth adding logic in that file to check whether an ID already exists. So just a note for now

test/features/personalization/personalization.test.js
```diff
@@ -377,6 +377,13 @@ describe('Functional Test', () => {
     expect(document.querySelector('meta[property="og:title"]').content).to.equal('New Title');
     expect(document.querySelector('meta[property="og:image"]').content).to.equal('https://adobe.com/path/to/image.jpg');
   });
+
+  it('will add id to the section div', async () => {
+    addSectionAnchors(document);
+    const sectionWithId = document.querySelector('#marquee-container');
+    expect(sectionWithId).to.exist;
+    expect(sectionWithId.getAttribute('id')).to.equal('marquee-container');
```
overmyheadandbody, line 385: Nit: since we're querying for the id already, is there any situation where this could fail?

overmyheadandbody (approved):

libs/features/personalization/personalization.js
chrischrischris: section-metadata adds the `section-anchor` class and there's no perf impact so I think it's fine.

libs/features/personalization/personalization.js
```diff
@@ -561,13 +561,25 @@ export const deleteMarkedEls = (rootEl = document) => {
     .forEach((el) => el.remove());
 };
 
+export function addSectionAnchors(rootEl = document) {
+  rootEl.querySelectorAll('.section-metadata').forEach((block) => {
+    [...block.children].forEach((row) => {
+      const col1 = row.children[0]?.textContent.toLowerCase().trim();
+      const col2 = row.children[1]?.textContent.toLowerCase().trim().replaceAll(' ', '-');
```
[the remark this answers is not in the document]
vgoodric: I guess we could move from a string check for a space to a pattern for whitespace?

vgoodric, line 568:
```suggestion
      const col2 = row.children[1]?.textContent.toLowerCase().trim().split(/\s+/).join('-');
```
couldn't get replaceAll to work with a regex, but maybe I'm just doing it wrong lol

overmyheadandbody: Should work
```
'abc def  ghi   jkl    mno     p'.replaceAll(/\s+/g, '-');
```

zagi25: Updated just in case, but it looks like EDS transforms multiple whitespaces to just one 
Screenshot 2025-03-07 at 11 30 02
Screenshot 2025-03-07 at 11 30 21
@chrischrischris

chrischrischris (approved):

vgoodric (approved):

PR #3786, opened 2025-03-07
MWPW-165461: Merch card collection autoblock

st-angelo-adobe:
Resolves: [MWPW-165461](https://jira.corp.adobe.com/browse/MWPW-165461)

**Test URLs:**
- After: https://mwpw-165461--milo--adobecom.aem.page/?martech=off

Plans Page:
mwpw-165461--milo--adobecom.aem.page/merch/mas/plans?martech=off

Docs Page:
mwpw-165461--milo--adobecom.aem.page/libs/features/mas/docs/plans.html?martech=off


libs/blocks/mas-autoblock/mas-autoblock.js
```diff
@@ -1,5 +1,7 @@
 import { createTag } from '../../utils/utils.js';
 import '../../deps/mas/merch-card.js';
+import '../../deps/mas/merch-card-collection.js';
```
yesil, line 3: this should be loaded on the fly before rendering a merch-card-collection, otherwise it will impact the performance.

libs/features/mas/src/merch-card-collection.js
```diff
@@ -197,13 +225,65 @@ export class MerchCardCollection extends LitElement {
...
         this.sidenav = document.querySelector('merch-sidenav');
+        this.hydrateFromFragment();
     }
 
     disconnectedCallback() {
         super.disconnectedCallback();
         this.stopDeeplink?.();
     }
 
+    async hydrateFromFragment() {
```
yesil, line 236: can we cover this method with tests.

libs/features/mas/src/merch-card-collection.js
```diff
@@ -122,6 +132,24 @@ export class MerchCardCollection extends LitElement {
             ${this.footer}`;
     }
 
+    checkReady() {
+        let ticks;
+        return new Promise((resolve) => {
+            const checkReadyInterval = setInterval(() => {
+                const aemFragment = this.querySelector('aem-fragment');
```
yesil, line 139: when a merch-card-collection from MAS Studio is consumed, is there a case `aem-fragment` would appear later?
suggestion:
mark `checkReady` async:
check `aem-fragment` does not exist,  return early
Promise.race  `aem-fragment.updateComplete` with a timeout Promise.
witch `setInterval` configuration below, we might lose up to 99ms. during the initialisation.

libs/blocks/mas-autoblock/mas-autoblock.js
yesil: likely you will need to load the following dependencies as well:

https://github.com/adobecom/milo/blob/stage/libs/blocks/merch-card-collection/merch-card-collection.js#L242-L253

st-angelo-adobe: Right. One issue is from what I understand `aem-fragment` is "packaged" with `merch-card.js`. So whichever autoblock type it is, `merch-card.js` should be imported right? But yes, the collection one can be imported only when needed.

3ch023:
thinking to just combine it with https://github.com/adobecom/milo/pull/3807

libs/blocks/mas-autoblock/mas-autoblock.js
```diff
@@ -1,16 +1,44 @@
-import { createTag } from '../../utils/utils.js';
+import { createTag, getConfig } from '../../utils/utils.js';
 import '../../deps/mas/merch-card.js';
-import '../../deps/mas/merch-quantity-select.js';
 import { initService } from '../merch/merch.js';
 
+// Should probably be removed (?)
+const DEFAULT_PLACEHOLDERS = {
```
3ch023, line 6: i wonder if these should belong to collection payload coming from Odin or just be Milo placeholders..
i'd discuss it in voice in dev call

libs/blocks/mas-autoblock/mas-autoblock.js
```diff
@@ -22,32 +50,162 @@ export function getTagName(el) {
...
+      break;
+  }
+
+  el.replaceWith(element);
+  await checkReady(control);
+
+  switch (tagName) {
+    case 'merch-card-collection': {
+      /* Placeholders */
+      const placeholders = control.data?.placeholders || DEFAULT_PLACEHOLDERS;
```
3ch023, line 179: can this be done before?
el.replaceWith(element);
await checkReady(control);

if yes we could remove bulky switch contructions and just have an 
```
if (tagName === 'merch-card-collection') {
    if (options.sidenav) {
        const container = createTag('div', null, control);
        element = container;
    }
    /* Placeholders */
    const placeholders = control.data?.placeholders || DEFAULT_PLACEHOLDERS;
    ...
}
```

would be a bit easier to read

libs/blocks/mas-autoblock/mas-autoblock.js
```diff
@@ -22,32 +50,162 @@ export function getTagName(el) {
...
+}
+
+/**
+ * @param {string} fragment
+ * @param {string} tagName
+ */
+function getTagAttributesAndContent(fragment, tagName) {
+  let attributes;
+  const html = createTag('aem-fragment', { fragment });
+  switch (tagName) {
```
3ch023, line 96: nit:
this also could be just
```
if (tagName === 'merch-card') {
    attributes = { consonant: '' };
}
```
3 lines instead of 7 and easier to read

libs/blocks/mas-autoblock/mas-autoblock.js
npeltier: right. Ideally we should have a mas placeholders set in collection payload that will get replaced with the right values at service time. Tricky part this would mean quite some change in merch-card-collection & merch-sidenav that need to work with catalog w/o mas too.

libs/blocks/mas-autoblock/mas-autoblock.js
```diff
@@ -22,32 +50,162 @@ export function getTagName(el) {
...
 
 export default async function init(el) {
-  const fragment = getFragmentId(el);
-  if (!fragment) return;
-  await createCard(el, fragment);
+  const options = getOptions(el);
+  if (!options.fragment) return;
+  const tagName = getTagName(el);
+  await loadControl(tagName);
+  await createControl(el.closest('.content'), options, tagName);
```
3ch023, line 210:
```suggestion
await loadDependencies(tagName);
await createMasElements(el.closest('.content'), options, tagName);
```

3ch023: when i read 'control' i get no idea of what is it, mas-commerce-service or merch-card-collection or swc libs..

libs/features/mas/build.mjs
```diff
@@ -7,7 +7,7 @@ const defaults = {
     bundle: true,
     format: 'esm',
     minify: true,
-    sourcemap: false,
+    sourcemap: true,
```
3ch023, line 10: gotcha :D

3ch023: maybe we add 
`npm run build:dev` that will set sourcemap variable to true?
then we don't need to change it in file all the time..
feel free to add, otherwise i will do in one of my next PRs

libs/features/mas/docs/src/plans.md
```diff
@@ -1,22 +1,25 @@
...
+        <a class="studio-link" target="_blank" href="https://main--mas--adobecom.hlx.live/studio.html?#path=nala&query=616273eb-3aad-462a-a6d7-6f6857973b77"> Open in Studio</a>
       </div>
       <div>
         <merch-card><aem-fragment fragment="8373b5c2-69e6-4e9c-befc-b424dd33469b"></aem-fragment></merch-card>
-        <a target="_blank" href="https://main--mas--adobecom.hlx.live/studio.html?#path=nala&query=8373b5c2-69e6-4e9c-befc-b424dd33469b"> Open in Studio</a>
+        <a class="studio-link" target="_blank" href="https://main--mas--adobecom.hlx.live/studio.html?#path=nala&query=8373b5c2-69e6-4e9c-befc-b424dd33469b"> Open in Studio</a>
       </div>
   </div>
+  <div id="plans-collection">
+    <a target="_blank" href="https://mwpw-165478--cc--adobecom.hlx.page/drafts/commerce/plans?milolibs=mwpw-165461--milo--adobecom">Plans collection page ↗</a>
```
3ch023, line 23: I would put it on the top, before cards, for people to find it easier.
and instead of 'Plans collection page' maybe call it 'Test Plans collection page' to make it obvious its not a live page?

libs/features/mas/docs/src/plans.md
```diff
@@ -1,22 +1,25 @@
...
       </div>
       <div>
         <merch-card><aem-fragment fragment="1736f2c9-0931-401b-b3c0-fe87ff72ad38"></aem-fragment></merch-card>
-        <a target="_blank" href="https://main--mas--adobecom.hlx.live/studio.html?#path=nala&query=1736f2c9-0931-401b-b3c0-fe87ff72ad38"> Open in Studio</a>
+        <a class="studio-link" target="_blank" href="https://main--mas--adobecom.hlx.live/studio.html?#path=nala&query=1736f2c9-0931-401b-b3c0-fe87ff72ad38"> Open in Studio</a>
       </div>
       <div>
         <merch-card><aem-fragment fragment="616273eb-3aad-462a-a6d7-6f6857973b77"></aem-fragment></merch-card>
-        <a target="_blank" href="https://main--mas--adobecom.hlx.live/studio.html?#path=nala&query=616273eb-3aad-462a-a6d7-6f6857973b77"> Open in Studio</a>
+        <a class="studio-link" target="_blank" href="https://main--mas--adobecom.hlx.live/studio.html?#path=nala&query=616273eb-3aad-462a-a6d7-6f6857973b77"> Open in Studio</a>
```
3ch023, line 15: pls replace all 'hlx.live' with 'aem.live'

libs/features/mas/src/aem-fragment.js
```diff
@@ -159,11 +159,11 @@ export class AemFragment extends HTMLElement {
     }
 
     async activate(masCommerceService, resolve, reject) {
-      this.#service = masCommerceService;
-      const flushCache = !this.#author;
-      this.refresh(flushCache)
-        .then((result) => resolve(result))
-        .catch((e)=> reject(e));
+        this.#service = masCommerceService;
+        // const flushCache = !this.#author;
```
3ch023, line 163: commented stuff can be removed

libs/features/mas/src/merch-card-collection.js
```diff
@@ -6,18 +6,28 @@ import { EVENT_MERCH_SIDENAV_SELECT } from './constants.js';
...
     EVENT_MERCH_CARD_COLLECTION_SORT,
     EVENT_MERCH_CARD_COLLECTION_SHOWMORE,
+    EVENT_AEM_ERROR,
+    EVENT_AEM_LOAD,
 } from './constants.js';
 import { TABLET_DOWN } from './media.js';
 import { styles } from './merch-card-collection.css.js';
 import { getSlotText } from './utils.js';
+import './mas-commerce-service';
+import { testCollectionPayload } from './merch-card-collection-test.js';
```
3ch023, line 16: this one is temporary?

libs/features/mas/src/variants/plans.css.js
```diff
@@ -62,4 +70,4 @@ merch-card[variant="plans"] [slot="quantity-select"] {
         grid-template-columns: repeat(4, var(--consonant-merch-card-plans-width));
     }
 }
-`;
+`;
```
3ch023, line 73: missing line break

libs/blocks/mas-autoblock/mas-autoblock.js
st-angelo-adobe: Yes, actually I'll delete this. I added them for the demo, because otherwise I'd have no labels. 
I spoke with Ilyas and he did say the placeholders will come in the collection payload.

@npeltier Do you mean this page? https://main--cc--adobecom.hlx.page/products/catalog This won't be affected, I believe. Because there, the "placeholders" come from the docs, and are populated in the `sidenav` block in cc, which will remain untouched.

libs/features/mas/src/merch-card-collection.js
st-angelo-adobe: Yes, until Ilyas commits his IO PR

libs/features/mas/build.mjs
st-angelo-adobe: Was gonna remove that before the wrap up. 😁  But I'll do what you proposed

libs/blocks/mas-autoblock/mas-autoblock.js
st-angelo-adobe: True, for me switch just seemed more readable, but sure, it seems like I need to do extra stuff just for collecton. I'll just turn them into ifs.

But it can't be done, the collection needs to be hydrated before I can populate placeholders, bcs the placeholders will be in the payload. So if the tag is `merch-card-collection`, I have to do something both before and after I `replaceWith`.

libs/blocks/mas-autoblock/mas-autoblock.js
```diff
@@ -22,32 +33,147 @@ export function getTagName(el) {
...
+
+    masElement.requestUpdate();
   }
 }
 
 export default async function init(el) {
-  const fragment = getFragmentId(el);
-  if (!fragment) return;
-  await createCard(el, fragment);
+  const options = getOptions(el);
```
seanchoi-dev, line 174: We need something like to include following PR's change.
https://github.com/adobecom/milo/pull/3873

```suggestion
  const options = getOptions(el);
  options.fragment = options.fragment || options.query
```

seanchoi-dev: I actually created a PR for this.
https://github.com/adobecom/milo/pull/3884

st-angelo-adobe: Yeah, I saw your PR, but didn't know it was already merged. Thanks, fixed

libs/features/mas/src/merch-card-collection.js
```diff
@@ -205,13 +232,96 @@ export class MerchCardCollection extends LitElement {
...
         this.stopDeeplink?.();
     }
 
+    async hydrate() {
+        if (this.hydrating) return false;
+
+        const aemFragment = this.querySelector('aem-fragment');
+        if (!aemFragment) return;
+
+        this.hydrating = true;
```
npeltier, line 249: i would move this just after first line

libs/blocks/mas-autoblock/mas-autoblock.js
```diff
@@ -22,32 +34,147 @@ export function getTagName(el) {
...
+  if (!options.fragment) return;
```
yesil, line 176: can you add `options.fragment ??= options.query;`

in mas studio, going forward MAS studio will copy the fragment id with `#query=fragmentId` but on Milo side, the attribute name should be `fragment`.

st-angelo-adobe: Actually, `option.fragment` is populated with either the `query` or `fragment` param, depending on which is populated. So we should be good here.

libs/features/mas/src/merch-card-collection.js
st-angelo-adobe: Yeah, but then if there's no aemFragment, it will be stuck on hydrating forever. I could set it back to false and then return, but it seemed better to me this way.

3ch023 (approved):

npeltier (approved):

seanchoi-dev (approved):

PR #3787, opened 2025-03-07
MWPW-169143: AH Try/Buy width and height fixes

st-angelo-adobe:
Resolves: [MWPW-169143](https://jira.corp.adobe.com/browse/MWPW-169143)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.live/docs/library/kitchen-sink/merch-card?martech=off
- After: https://mwpw-169143--milo--adobecom.hlx.live/docs/library/kitchen-sink/merch-card?martech=off

Adobe Home Gallery:
mwpw-169143--milo--adobecom.hlx.page/libs/features/mas/docs/adobe-home.html?martech=off


Roycethan:
@st-angelo-adobe In safari browser,  single plan type - image is not looking right...
image
https://mwpw-169143--milo--adobecom.aem.page/libs/features/mas/docs/adobe-home.html?martech=off

Compare against: https://main--milo--adobecom.aem.page/libs/features/mas/docs/adobe-home.html?martech=off

cc @Axelcureno  @3ch023

st-angelo-adobe:
@Roycethan Hm, I can't reproduce it, but the thing is the merch-card now takes the full width and height of the container. So it's really up to the container to be the correct size.

Roycethan:
> @Roycethan Hm, I can't reproduce it, but the thing is the merch-card now takes the full width and height of the container. So it's really up to the container to be the correct size. Let's sync about this.

As discussed in call, its happening in Mac-Sonoma (14.4) version - to repro this version can be used

st-angelo-adobe:
@Roycethan I'm not convinced anything has to be done here. That specific page is just part of the docs, and this is a very niche issue, that seems to only occur on a version of Mac.
Fact of the matter is, the card should take the full height, and if Safari renders the container in a weird way, then that needs to be addressed by the css of the client (for instance by the AH team in this case). 

Yes, for the docs it's obviously our job to do that. But the question is do we want to spend some time to debug and fix this on this specific version? cc: @3ch023 @npeltier @yesil

st-angelo-adobe:
@Roycethan merging the latest stage did the trick

PR #3789, opened 2025-03-07
MWPW-169416 [MEP] stop treating remove action differently in preview

vgoodric:
* No longer has a different behavior in preview when removing content or a fragment

Resolves: [MWPW-169416](https://jira.corp.adobe.com/browse/MWPW-169416)

The goal of this PR is to remove the difference between preview and the MEP button being off (prod without MEP param or preview with `mepButton=off `param).  

1. In the before link, the first text block still takes up less of the page width because the section-metadata is still working. You can also see the section-metadata in the DOM. In the after and MEP button off links, the block is full width and the section-metadata block is not in the DOM.
2. In the before link, highlight changes shows there is a deleted fragment and it is still in the DOM. In the after and MEP button off links it is fully removed.

**Test URLs:**
- Before: https://main--cc--adobecom.aem.page/drafts/mepdev/fragments/2025/q2/meptrueremove/
- Before with MEP button off: https://main--cc--adobecom.aem.page/drafts/mepdev/fragments/2025/q2/meptrueremove/?mepButton=off
- After: https://main--cc--adobecom.aem.page/drafts/mepdev/fragments/2025/q2/meptrueremove/?milolibs=meptrueremove
- Psi-check: https://meptrueremove--milo--adobecom.aem.page/?martech=off

Marked high priority because it makes the QA for a test more difficult.


markpadbe (approved):
To avoid confusion, would it be worth adding some text to the Highlight changes checkbox text. For example:

Screenshot 2025-03-07 at 2 07 19 PM

Otherwise, the change looks good to me.

ivanvatadobe (approved):

PR #3796, opened 2025-03-10
Fix(Mwpw-168438): The play() request was interrupted by a call to pause() bug fix.

sharath-kannan:
* Safe pause functionality is introduced to avoid "The play() request was interrupted by a call to pause()." error.
* Reference : https://developer.chrome.com/blog/play-request-was-interrupted

Resolves: [MWPW-168438](https://jira.corp.adobe.com/browse/MWPW-168438)

**Test URLs:**
- Before:  https://main--milo--adobecom.hlx.page/drafts/ramuntea/accessibility/video-use-cases?martech=off
- After: https://mwpw-168438--milosh--sharath-kannan.hlx.page/drafts/ramuntea/accessibility/video-use-cases?martech=off


libs/utils/decorate.js
```diff
@@ -286,6 +286,15 @@ export function addAccessibilityControl(videoString, videoAttrs, indexOfVideo, t
   `;
 }
 
+function SafeVideoPause(video) {
+  if (video.autoplay) { video.playPromise = video.play(); }
+  if (video.playPromise !== undefined) {
+    video.playPromise.then(() => {
```
sharmrj, line 292: nit: there's no need for the if check
```javascript
video.playPromise?.then(...)
```

PR #3804, opened 2025-03-12
MWPW-169654 Optimizing Body Hiding for better Indexing and SEO

sharmrj:
As a Google bot, I need to access and read the full HTML of acom pages during the initial crawl, when JavaScript is disabled. Googlebot primarily crawls and indexes pages in two phases:

1.First pass (HTML-first indexing) – Googlebot attempts to index content without executing JavaScript, relying solely on the server-rendered HTML. If critical content is missing at this stage, it may not get indexed properly.

2.Second pass (JavaScript rendering) – Googlebot may revisit the page later with JavaScript enabled to process dynamically loaded content, but this can take time and is not guaranteed for all pages.

For optimal indexing, key content should be fully available in the server-rendered HTML during the first crawl to ensure visibility in search results.

Google bots currently see a blank page during indexing because the body is hidden using CSS. Since JavaScript is disabled when bots access the HTML, the content remains inaccessible.

This PR introduces a line to support any consuming project adding the following in their head.html _instead of_ having a style `body { display:  none }`:
```html
<script>document.documentElement.style.display = 'none';</script>
```

Resolves: [MWPW-169654](https://jira.corp.adobe.com/browse/MWPW-169654)

**Test URLs:**
- After: https://seo-body-unhide--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -1460,7 +1460,10 @@ async function processSection(section, config, isDoc) {
   await Promise.all(loadBlocks);
 
   delete section.el.dataset.status;
-  if (isDoc && firstSection) await loadPostLCP(config);
+  if (isDoc && firstSection) {
+    document.documentElement.style.removeProperty('display');
+    await loadPostLCP(config);
+  }
   delete section.el.dataset.idx;
   return section.blocks;
 }


amitbikram (approved):

JasonHowellSlavin:
I think that this may have unintended consequences in regards to cumulative layout shift. 

If the script tag in the head is necessary to prevent that and will allow Google to parse, can PRs for consumers be opened as a requirement of this ticket?

JasonHowellSlavin (changes requested):
Requesting changes until CLS resolution

sharmrj:
I can see the concern over CLS.
AFAICT the style to hide the body is removed only after the first section should be visible (we do it right before kicking off `loadPostLCP`). 
We didn't see any FOUC or CLS in the testing we did on some CC pages and some DC pages (for example, on DC we raised https://github.com/adobecom/dc/pull/1045 and tested it with content overrides on live pages. There's also https://github.com/adobecom/cc/pull/681). 
Individual consumers can choose whether or not to comnsume the change by changing the `<style>body { display: none; } </style>` in their `head.html` with `<script>document.documentElement.style.display='none';</script>`.  If they do not opt in, this PR will have zero effect on them.

JasonHowellSlavin (approved):

JasonHowellSlavin:
> I can see the concern over CLS. AFAICT the style to hide the body is removed only after the first section should be visible (we do it right before kicking off `loadPostLCP`).
> 
> We didn't see any FOUC or CLS in the testing we did on some CC pages and some DC pages (for example, on DC we raised [adobecom/dc#1045](https://github.com/adobecom/dc/pull/1045) and tested it with content overrides on prod pages. There's also [adobecom/cc#681](https://github.com/adobecom/cc/pull/681)).
> 
> Individual consumers can choose whether or not to consume the change by swapping the `<style>body { display: none; } </style>` in their `head.html` for `<script>document.documentElement.style.display='none';</script>`. If they do not opt in, this PR will have zero effect on them.

Sounds good to me. We can keep an eye on metrics. The performance dashboards will notify us if there is any issue. Thank you for going into that detail!

salonijain3:
No engineering changes are needed.

PR #3805, opened 2025-03-12
MWPW-164491 [Plans] Wide merch card

bozojovicic:
Plans wide and super-wide card with "What's included" component.
Unfortunately wide cards can be properly tested only inside of the card collection, which are not done yet, so I cannot provide any good test page in Milo for this yet.

Resolves: [MWPW-164491](https://jira.corp.adobe.com/browse/MWPW-164491)

**Test URLs:**
- After: https://mwpw164491wide--milo--bozojovicic.aem.live/?martech=off


libs/features/mas/src/merch-card.js
```diff
@@ -8,6 +8,8 @@ import {
 
 import './global.css.js';
 import './aem-fragment.js';
+import './merch-mnemonic-list.js';
+import './merch-whats-included.js';
```
yesil, line 12: I wonder what is the impact of including `merch-whats-included.js in` bytes.
if we render only on clicks, we could also load it on the fly.

yesil (approved):

libs/features/mas/src/utils.js
```diff
@@ -35,12 +35,16 @@ export function createTag(tag, attributes = {}, content = null, is = null) {
     return element;
 }
 
+export function matchMobile() {
+  return window.matchMedia('(max-width: 767px)');
```
npeltier, line 39:
```suggestion
  return window.matchMedia(MOBILE_LANDSCAPE);
```

libs/features/mas/src/merch-card.js
npeltier: 4 bytes :)?

npeltier (approved):
lgtm

libs/features/mas/src/merch-card.js
yesil: not sure, we need to keep in mind that many cards on adobe.com do not need this, especially on page load.

afmicka:
@bozojovicic 
1. layout for Stock checkbox and Secure transaction in figma is different for wide card without what's included and with it. You have it the same on both. Screenshots to compare studio and figma for wide card. The same is also if i remove the quantity selector from the card. 
- Wide card in studio: 
Screenshot 2025-03-28 at 14 30 43

- Figma: 
Screenshot 2025-03-28 at 14 12 30
Screenshot 2025-03-28 at 14 11 41

2. Callout text seems to be put in the footer for super wide card. If not callout, what is it in figma? In studio is left where it was. 
Screenshot 2025-03-28 at 14 28 11

3. When previewed in milo (with autoblock), whats included icons are not starting in the new row and are not one bellow another as in studio and figma.  Also for a default (singe) size, those boxes are moved to the right and not aligned anymore for both with and without whats included 
https://mwpw164491wide--milo--bozojovicic.aem.page/drafts/mili/document
Screenshot 2025-03-28 at 14 42 45

4. Space is removed before Quantity selector. Was that intentional? I see that space is some figma examples
https://mwpw164491wide--milo--bozojovicic.aem.page/libs/features/mas/docs/plans.html?martech=off
Screenshot 2025-03-28 at 13 56 20
Screenshot 2025-03-28 at 13 56 25

bozojovicic:
@afmicka 
1. I totally missed this in Figma and Lucy did not mention it in the ticket. I asked her in MWPW-164491 what to do about this case. Waiting ...
2. I agreed with Lucy before to exclude this requirement from this ticket. She created new one for that change MWPW-169832
3. My bad. I fixed it.
4. Lucy asked me if I can reduce some empty space where possible. But this case is really problematic. I reverted my change for that space above the quantity selector. Now it should be as before.

Axelcureno (approved):

afmicka:
@bozojovicic when the card is authored as wide or superwide but viewed as single in milo, the Stock checkbox is not visible
https://mwpw164491wide--milo--bozojovicic.aem.page/drafts/mili/document
Screenshot 2025-04-03 at 10 29 42

Screenshot 2025-04-03 at 10 39 06

PR #3809, opened 2025-03-12
[MWPW-169392][Preflight][Due diligence] Integrate full accessibility report in Preflight

skumar09:
* Integrate Axe Core Accessibility Test into Preflight
* Develop a custom accessibility-check function/method for rules that the axe-core library does not validate when analyzing or testing within the browser DOM. 

Note: 
- Currently, this code is still a work in progress

Resolves: [MWPW-169392](https://jira.corp.adobe.com/browse/MWPW-169392)

**Test URLs:**
- After: https://a11y-preflight--milo--skumar09.hlx.page/drafts/nala/preflight/media-small/?martech=off

Test Scenarios
- 1] Accessibility test for the page is PASS
- 
image

-2] Accessibility test for the page is FAIL ( Image Alt Text is null or incorrect value ] 
- 
image
image

-3] Accessibility test for the page FAIL ( Color Contrast Issue )
image
image


libs/blocks/preflight/panels/accessibility.js
```diff
@@ -1,132 +1,219 @@
...
-  </div>`;
+    <div id="preflight-panel" class="preflight-general-content">
+      <div class="preflight-columns">
+        <!-- Summary Column -->
+        <div class="preflight-column">
+          ${loading && html`<p>Running Accessibility Test...</p>`}
+          ${!loading
+          && testResults
+          && !testResults.error
+          && html`
```
mokimo, line 89: You should be able to simplify this:

```
if(loading) return html`.. loadingMarkup`
if(error) return html`.. errorMarkup`
return html`...normal state`
```

if this doesn't work, alternatively you can do
```
const loadingMarkup = loading ? html`` : ""
const ...
return html`${loadingMarkup} ${errorMarkup} ${normalMarkup}`
```

libs/blocks/preflight/panels/accessiblity-helper.js
```diff
@@ -0,0 +1,152 @@
...
+  if (checks.includes('altText')) {
+    violations.push(...checkImageAltText(elements));
+  }
+
+  if (checks.includes('keyboard')) {
+    violations.push(...checkKeyboardNavigation(elements));
+  }
+
+  return violations;
+}
```
mokimo, line 150: I think you can trim this down by ensuring getFilteredElements always returns an array & moving the checks portion into the functions and returning undefined otherwise.

Also rather than passing around so many parameters, it might be easier to have something such as `const config = {}` at the top of the file rather than trying to pass different parameters if we only use it once anyway
```suggestion
// Main function to run custom accessibility checks
async function customAccessibilityChecks(config = {}) {
  const {
    checks = ['altText', 'keyboard'],
    include = ['body'],
    exclude = [],
  } = { ...defaultConfig, ...config };
  return [
    ...getFilteredElements(include, exclude)
    .filter(element => checkImageAltText([element, checks]) || checkKeyboardNavigation([element, checks]))
    ]
}
```

mokimo (commented):
Cool, I think some things can be trimmed down, the old panel can be removed completely if we don't use it anymore - we have git history in the case that we need it again.

I love the concept, let's brush up the code a bit and make it more clear and concise before getting this in though!

libs/blocks/preflight/panels/accessibility.js
skumar09: @mokimo , 
- Thank you for the review! I've incorporated the review comments; please review and advise
- I have updated the structure and made it more modular by creating a separate folder for accessibility and its corresponding custom checks

libs/blocks/preflight/panels/accessibility/accessibility-custom-checks.js
```diff
@@ -0,0 +1,35 @@
...
+      return [];
+    }
+
+    const violations = checkFunctions.flatMap((checkFn) => checkFn(elements, config));
+
+    return violations;
+  } catch (error) {
+    window.lana.log(`Error running custom accessibility checks: ${error.message}`);
+    return [];
+  }
```
mokimo, line 32: The try catch seems a bit overly defensive, if everything is defined you should not need that. There's no harm in keeping this in though.

libs/blocks/preflight/panels/accessibility/accessibility.js
```diff
@@ -0,0 +1,236 @@
+import { html, useState, useEffect } from '../../../../deps/htm-preact.js';
+import '../../../../deps/axe.min.js';
+import customAccessibilityChecks from './accessibility-custom-checks.js';
+
+const axeCoreConfig = {
+  include: [['body']],
+  exclude: [['.preflight'], ['header']],
```
mokimo, line 7: Should we not run accessibility checks for the navigation as well?

libs/blocks/preflight/panels/accessibility/check-image-alt-text.js
```diff
@@ -0,0 +1,61 @@
...
+export default function checkImageAltText(elements = [], config = {}) {
+  const { checks = [] } = config;
+
+  // Skip this check if 'altText' isn't enabled in the config
+  if (!checks.includes('altText')) {
+    return [];
+  }
+
+  const violations = [];
+  const images = elements.filter((el) => el.tagName && el.tagName.toLowerCase() === 'img');
```
mokimo, line 18: nit:
```suggestion
  const images = elements.filter((el) => el.tagName?.toLowerCase() === 'img');
```

libs/blocks/preflight/panels/accessibility/check-image-alt-text.js
```diff
@@ -0,0 +1,61 @@
...
+        impact: 'serious',
+        id: 'image-alt',
+        help: 'Add an alt attribute that describes the image.',
+        helpUrl: 'https://www.w3.org/WAI/tutorials/images/',
+        nodes: [{
+          target: [getUniqueSelector(img)],
+          html: img.outerHTML,
+        }],
+      });
+    } else if (/^\d+$/.test(alt.trim()) || alt.trim().length < 3) {
```
mokimo, line 45: I'm usually a fan of trying to avoid chaining if/else statements, similar to using early returns.
If/else statements they usually get added on and long-term turn into huge if/if-else/if-else/else statements that are very hard to decipher
```js
if (condition) {
// ...
// continue / break / return
}
```

libs/blocks/preflight/panels/accessibility/check-keyboard-navigation.js
```diff
@@ -0,0 +1,70 @@
...
+  focusableElements.forEach((el) => {
+    const styles = window.getComputedStyle(el);
+
+    const hasVisibleOutline = styles.outlineStyle !== 'none' && parseFloat(styles.outlineWidth) > 0;
+
+    const hasBoxShadow = styles.boxShadow !== 'none' && styles.boxShadow !== '';
+
+    const hasFocusIndicator = hasVisibleOutline || hasBoxShadow;
+
+    if (!hasFocusIndicator) {
```
mokimo, line 54: nit, early returns are preferred 
```suggestion
    if (hasFocusIndicator) return
```

mokimo (approved):
A few nitpicks, however this is great! 🚀

libs/blocks/preflight/panels/accessibility/helper.js
```diff
@@ -0,0 +1,32 @@
...
+}
+
+/**
+ * Returns elements included by selectors, excluding specified selectors.
+ * @param {string[]} includeSelectors - css selectors to include.
+ * @param {string[]} excludeSelectors - css selectors to exclude.
+ * @returns {HTMLElement[]} - Filtered DOM elements.
+ */
+export function getFilteredElements(includeSelectors = ['body'], excludeSelectors = []) {
+  const includedElements = includeSelectors.flatMap((selector) => Array.from(document.querySelectorAll(`${selector} *`)));
```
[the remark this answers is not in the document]
skumar09: Updated include, so consistent with excluded

libs/blocks/preflight/panels/accessibility/check-color-contrast.js
```diff
@@ -0,0 +1,116 @@
...
+// Parses an RGB color string into an array of RGB values.
+function parseCssRgb(rgbStr) {
+  if (!rgbStr) return null;
+  const match = rgbStr.match(/rgba?\((\d+),\s*(\d+),\s*(\d+)/);
+  return match ? match.slice(1, 4).map(Number) : null;
+}
+
+// Checks if an RGBA color string is fully transparent.
+function isTransparent(rgbStr) {
+  return rgbStr.includes('rgba') && rgbStr.trim().endsWith(', 0)');
```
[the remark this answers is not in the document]
skumar09: Updated the function also to cover the modern notation

libs/blocks/preflight/panels/accessibility/check-color-contrast.js
```diff
@@ -0,0 +1,116 @@
...
+    if (!fgColor || !bgColor) {
+      // console.warn(`[${index}] Skipped: Unable to compute colors for element.`);
+      return;
+    }
+
+    const contrast = contrastRatio(fgColor, bgColor);
+    const styles = window.getComputedStyle(el);
+    const fontSize = parseFloat(styles.fontSize) || 0;
+    const fontWeight = parseInt(styles.fontWeight, 10) || 400;
+    const isLargeText = fontSize >= 24 || (fontSize >= 18.66 && fontWeight >= 700);
```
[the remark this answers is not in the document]
skumar09: These values are based on [WCAG 2.1 guidelines](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html), 
- WCAG specifies font size in points(pt), browser we get in pixels, so we need to convert points to pixies
- 1pt ≈ 1.333px, 14pt × 1.333 = 18.662px (rounded to 18.66), 18pt × 1.333 = 23.994px (rounded to 24)
- WCAG defines "large text" as 18pt (24px) for regular text and 14pt (18.66px) for bold text.

mokimo:
I'm also getting the use `await axe.run()` error, so let's try to have a look on what's causing that before merging this

libs/blocks/preflight/panels/accessibility/helper.js
```diff
@@ -0,0 +1,32 @@
...
+ * @returns {HTMLElement[]} - Filtered DOM elements.
+ */
+export function getFilteredElements(includeSelectors = ['body'], excludeSelectors = []) {
+  const includedElements = includeSelectors.flatMap((selector) => Array.from(document.querySelectorAll(`${selector},  ${selector} *`)));
+
+  if (excludeSelectors.length === 0) return includedElements;
+
+  const excludedElements = excludeSelectors.flatMap((selector) => Array.from(document.querySelectorAll(`${selector}, ${selector} *`)));
+
+  return includedElements.filter((el) => !excludedElements.includes(el));
```
vhargrave, line 31: I'm not sure this will work properly. 
As I've understood it, Array.includes() uses strict equality (===) for comparison. When comparing DOM nodes, this will only return true if you're comparing the exact same node reference. Even if two nodes represent the same element in the DOM, they will be considered different if they were obtained from different querySelectorAll calls.

Here's an example: 
```
const div = document.querySelector("div");
const divs1 = Array.from(document.querySelectorAll("div"));
const divs2 = Array.from(document.querySelectorAll("div"));
console.log(divs1[0] === divs2[0]); // false, even though they’re the same DOM element
console.log(divs1.includes(divs2[0])); // false
```

Maybe you can try something like this: 
```
const includedElements = includeSelectors.flatMap((selector) => 
  Array.from(document.querySelectorAll(`${selector}, ${selector} *`))
);
if (excludeSelectors.length === 0) return includedElements;
const excludedElements = excludeSelectors.flatMap((selector) => 
  Array.from(document.querySelectorAll(`${selector}, ${selector} *`))
);
return includedElements.filter((el) => 
  !excludedElements.some((ex) => el.isSameNode(ex))
);
```

libs/blocks/preflight/panels/accessibility/accessibility.js
```diff
@@ -0,0 +1,226 @@
...
+
+/**
+ * Runs the accessibility test using axe-core and custom checks.
+ * @returns {Promise<Object>} Test results object (pass/fail + violations) or error.
+ */
+async function runAccessibilityTest() {
+  try {
+    const results = await window.axe.run(axeCoreConfig);
+
+    const customViolations = await customAccessibilityChecks(customChecksConfig);
```
vhargrave, line 28: you could probably run your custom checks at the same time as window.axe.run right? That would make this a little more efficient.

vhargrave (changes requested):
It’d be awesome to have some unit tests for this.

Also, could you share a working 'after' URL? The current link doesn’t seem to work.

When I check the accessibility tab in a project, I’m running into this error:
image

libs/blocks/preflight/panels/accessibility/accessibility.js
```diff
@@ -0,0 +1,226 @@
+import { html, useState, useEffect } from '../../../../deps/htm-preact.js';
+import '../../../../deps/axe.min.js';
+import customAccessibilityChecks from './accessibility-custom-checks.js';
+
+const axeCoreConfig = {
+  include: [['body']],
+  exclude: [['.preflight'], ['header'], ['.global-navigation']],
```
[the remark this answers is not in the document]
skumar09: I need to check with the tattva team on that,

libs/blocks/preflight/panels/accessibility/helper.js
skumar09: @vhargrave: Good catch! (`.includes` would fail in this case due to strict equality checks on different node references], I have updated the changes

libs/blocks/preflight/panels/accessibility/accessibility.js
```diff
@@ -0,0 +1,226 @@
...
+  exclude: [['.preflight'], ['header'], ['.global-navigation'], ['footer'], ['.global-footer']],
+};
+
+/**
+ * Runs the accessibility test using axe-core and custom checks.
+ * @returns {Promise<Object>} Test results object (pass/fail + violations) or error.
+ */
+async function runAccessibilityTest() {
+  try {
+    const results = await window.axe.run(axeCoreConfig);
```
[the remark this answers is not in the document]
skumar09: @vhargrave :  I guess from Axe-Core v4.0+ onwards, it supports `Promises` and `axe.run()` rejects on errors, so try catch is catching errors here, 
- I did a quick test by passing an invalid selector in include[], and the error was caught)
- `LANA Msg:  Error running accessiblity test:, Error: No elements found for include in page Context `

libs/blocks/preflight/panels/accessibility/accessibility.js
skumar09: I tried below but noticed an inconsistency in results, which may be due to timing or DOM state changes
```
    const [results, customViolations] = await Promise.all([
      window.axe.run(axeCoreConfig),
      customAccessibilityChecks(customChecksConfig),
    ]);
```

vhargrave: alright, was worth a try at least 💪

rgclayton (changes requested):
Santosh and I had a call to talk about this.

We decided we do not want to remove the previous functionality related to capturing all images on the page and displaying their alt text. 

We'll need to work with @overmyheadandbody  to either combine that feature with the work he's doing related to blurry images or figure out how best to organize everything in preflight with out removing current functionality.

PR #3813, opened 2025-03-13
MWPW-169732 Fetch resources from adobe.com/federal instead of https://main--federal--adobecom.aem.live/federal

sharmrj:
* Added some logic to determine whether to use `www.adobe.com` or `www.stage.adobe.com` for the standalone gnav 
* There's some defensive logic in case the env is neither stage nor prod.

Resolves: [MWPW-169732](https://jira.corp.adobe.com/browse/MWPW-169732)

**Test URLs:**
- After: https://standalone-prod-content--milo--adobecom.aem.page/?martech=off

qa url: https://adobecom.github.io/nav-consumer/navigation.html?env=stage&navbranch=standalone-prod-content

### DIFF
--- a/libs/navigation/navigation.js
+++ b/libs/navigation/navigation.js
@@ -112,9 +112,13 @@ export default async function loadBlock(configs, customLib) {
     import('../utils/utils.js'),
   ]);
   const paramConfigs = getParamsConfigs(configs);
-  const origin = env === 'prod'
-    ? 'https://www.adobe.com/federal'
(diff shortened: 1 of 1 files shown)


bandana147 (approved):

mokimo (approved):
You could also just use `utils.getFederatedContentRoot()`

bandana147:
Marking it high priority as Adobe home is seeing huge error rates on prod. 
Screenshot 2025-03-13 at 3 07 44 PM

PR #3816, opened 2025-03-13
MWPW-169072: Handle messages from iframe

zagi25:
* Added event listener in `iframe` block for messages from Manage Plan iframe
* Added `hide-close-button` style to the modal
* Added event listener for default modal close event

Resolves: [MWPW-169072](https://jira.corp.adobe.com/browse/MWPW-169072)

**Test URLs:**
- After: https://mwpw-169072-cancellation-modal--milo--adobecom.aem.page/?martech=off

**Federal Test URLs:**
- https://main--federal--adobecom.hlx.page/de/federal/footer/fragments/footer-copyright#cancel
https://main--federal--adobecom.hlx.page/de/federal/footer/fragments/footer-copyright?milolibs=mwpw-169072-cancellation-modal#cancel


libs/blocks/iframe/iframe.js
```diff
@@ -1,13 +1,57 @@
...
   el.classList.remove('iframe');
   const classes = [...el.classList].join(' ');
 
-  if (!url) return;
+  let url = null;
+  try {
+    url = new URL(linkHref);
+  } catch (e) {
+    return;
+  }
```
mokimo, line 48: According to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#href the href should always be a valid url, so I think we can just do:
```suggestion
  if(!linkHref) return
  let url = new URL(linkHref)
```

libs/blocks/iframe/iframe.js
```diff
@@ -1,13 +1,57 @@
...
+export function handleIFrameEvents({ data }) {
+  let parsedMsg = null;
+  try {
+    parsedMsg = JSON.parse(data);
+  } catch (error) {
+    return;
+  }
+  const { app } = parsedMsg;
+  if (app === 'ManagePlan') handleManagePlanEvents(parsedMsg);
+}
```
mokimo, line 36: Nit: let's move everything into the try block
```suggestion
export function handleIFrameEvents({ data }) {
  try {
    const parsedMsg = JSON.parse(data);
    if (parsedMsg.app === 'ManagePlan') handleManagePlanEvents(parsedMsg);
  } catch (error) {
    return;
  }
}
```

mokimo (approved):

libs/blocks/modal/modal.css
```diff
@@ -388,6 +389,10 @@
   .dialog-modal.tall-video .milo-video {
     --modal-width-var: 35vw;
   }
+
+  .dialog-modal.manage-plan-cancel {
+    width: 1280px;
```
overmyheadandbody, line 394: Should there also be a `max-width: 100%`? What would happen at `1200px` (the breakpoint above which this rule get applied), could the modal overflow the page? 🤔

overmyheadandbody (commented):
I just had a quick look over the code, I'll get back with eventual notes after having a look over the functionality.

libs/blocks/modal/modal.css
zagi25: It doesn't overflow, there is other `max-width` rule, but I think you're right it is better to keep it as wide as possible until the break point.

overmyheadandbody (approved):

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-169072

PR #3826, opened 2025-03-17
MWPW-169166: Check autoblock source

zagi25:
Autoblocks are only loaded from:
* same origin
* `adobe.com` or `.adobe.com` 
* auto block origins (e.g. instagram, youtube)
* `pdf` has no restrictions as long as it has `pdf` extension in the url ( I didn't restrict this because of this [pdf-viewer.js](https://github.com/adobecom/milo/blob/stage/libs/blocks/pdf-viewer/pdf-viewer.js#L49) )

Resolves: [MWPW-169166](https://jira.corp.adobe.com/browse/MWPW-169166)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/ratko/mwpw-169166/autoblocks?martech=off
- After: https://mwpw-169166-autoblocks-source--milo--adobecom.aem.page/drafts/ratko/mwpw-169166/autoblocks?martech=off


libs/utils/utils.js
```diff
@@ -633,9 +633,18 @@ export function decorateImageLinks(el) {
...
 }
 
+export function isTrustedAutoBlock(autoBlock, url) {
+  if (!url.href.includes(autoBlock)) return false;
+  const urlHostname = url.hostname.replace('www.', '');
+  return urlHostname === window.location.hostname
+    || urlHostname.endsWith('.adobe.com')
+    || urlHostname === 'adobe.com'
+    || urlHostname === autoBlock
+    || (autoBlock === '.pdf' && url.pathname.endsWith(autoBlock));
```
overmyheadandbody, line 643: Would a similar condition be needed for `.mp4`? I'm not certain that auto-block condition would ever be met 🤔

overmyheadandbody (commented):
This looks good. Thinking of some edge-cases, what if CC tries to load a fragment from Federal? Would the conditions be met in that case?

libs/utils/utils.js
zagi25: Based on this documentation https://milo.adobe.com/docs/authoring/links all`aem/hlx` links are converted to the relative links, and to add fragment from different project, link must be absolute adobe domain e.g. https://adobe.com/fragments/some-fragment so those will fall under `adobe.com / endsWith('.adobe.com')` . 
Do federal fragments behave differently than this ? 

`.mp4` are only loaded from the `codeRoot` [video.js](https://github.com/adobecom/milo/blob/stage/libs/blocks/video/video.js#L14), so the code will always look for video in projects share point or I am not fully understanding of how `codeRoot` config property works ?

overmyheadandbody: Re: Federal - you're right. I have a quick test [page](https://mwpw-169166-autoblocks-source--milo--adobecom.aem.page/drafts/ramuntea/fetching-federal-fragment), the Federal fragment is included correctly. What starts out as `https://main--federal--adobecom.aem.page/federal/footer/fragments/footer-acom-column-1 ` in the document becomes `/federal/footer/fragments/footer-acom-column-1` on the page, which then meets the first condition of `urlHostname === window.location.hostname`.

Re: Videos - same as above would apply, yes.

overmyheadandbody (approved):

libs/utils/utils.js
mokimo: Yeah federal links are also relative, somewhere in the pipeline we assemble them based on spotting `/federal/*...` within the path, not sure if that leads to any timing issues? I think autoblocks should always be assembled before that

zagi25:
@mokimo For fragments no timing issue, federal url is later assembled in [fragment.js](https://github.com/adobecom/milo/blob/stage/libs/blocks/fragment/fragment.js#L97). But I see that `getFederatedUrl` and `decorateLinks` (which calls `decorateAutoBlock`) are imported together [here](https://github.com/adobecom/milo/blob/stage/libs/blocks/global-navigation/utilities/utilities.js#L8), I can also add check for `aem/hlx.page/live` domain just to be safe if somewhere some URL is transformed to federal before autoblock is assembled ?
cc: @overmyheadandbody

PR #3833, opened 2025-03-19
[MWPW-169532] - iframe title added

DKos95:
This fixes the issue where the Iframe was missing an accessible name.

Resolves: [MWPW-169532](https://jira.corp.adobe.com/browse/MWPW-169532)

**QA OTHER ORIGIN URLs:**
- Before: https://main--cc--adobecom.aem.page/drafts/dusan/iframe-testing?martech=off
- After: https://main--cc--adobecom.aem.page/drafts/dusan/iframe-testing?milolibs=iframe-title&martech=off

**QA SAME ORIGIN URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/dusan/iframe-same-origin-testing?martech=off
- After: https://iframe-title--milo--adobecom.hlx.page/drafts/dusan/iframe-same-origin-testing?martech=off

**QA VIDEO URLs:**
- Before: https://iframe-title--milo--adobecom.hlx.page/drafts/dusan/iframe-video-testing?martech=off
- After: https://main--milo--adobecom.hlx.page/drafts/dusan/iframe-video-testing?martech=off


libs/blocks/iframe/iframe.js
```diff
@@ -1,4 +1,5 @@
 import { createTag } from '../../utils/utils.js';
+import { getMetadata } from '../section-metadata/section-metadata.js';
```
mokimo, line 2: I'm not a fan of building those dependency chains, they aren't good for performance... However, having an iFrame as LCP is probably already bad as well.

Still, I see that `section-metadata.js` is used across 14 blocks, I wonder if it makes sense to move ... maybe only the `getMetadata` function into utils? @overmyheadandbody @vhargrave @robert-bogos @zagi25 any opinion on this?

mokimo (approved):

libs/blocks/iframe/iframe.js
zagi25: I don't have a strong reason against it. But if we were to move it, we would need to come up with a new name https://github.com/adobecom/milo/blob/stage/libs/utils/utils.js#L197

zagi25:
About the AC in the ticket, maybe we can check if the origin is the same (adobe.com page calls adobe.com iframe which should be the case on prod), and pick up the attributes after the fragment is loaded ? And also have title override in the metadata as a fallback

libs/blocks/iframe/iframe.js
vhargrave: @mokimo It’d make it more reusable, but the trade-off is it’d load on every page, even for users who don’t need it. Plus, it’s not usually tied to LCP elements, so I’m not sure it’s critical to have it load early. For those reasons I probably would leave it where it is, or put it in a file that can load a little later.

vhargrave (commented):
Hey @zagi25 , I noticed you brought up something similar, and I’m on the same page. It might be worth checking if the iframe’s on the same origin first—grabbing the header that way could be a smooth option. I was also thinking about using a fetch request for pages on different domains. It’s got a solid shot at pulling the header, though if it relies on JS, we might miss it. Still, I think this in combination with checking the domain will cover most use cases and save our authors a lot of time. And if neither works, maybe the modal metadata approach could be a fallback. What do you think?

And finally, if all else fails you could do the modal metadata approach then. What do you think? 

```
fetch(url)
    .then(response => response.text())
    .then(html => {
      const parser = new DOMParser();
      const header = doc.querySelector('h1, h2, h3, h4, h5, h6')?.textContent || 'No header found';
      console.log(header);
  })
```

DKos95:
> Hey @zagi25 , I noticed you brought up something similar, and I’m on the same page. It might be worth checking if the iframe’s on the same origin first—grabbing the header that way could be a smooth option. I was also thinking about using a fetch request for pages on different domains. It’s got a solid shot at pulling the header, though if it relies on JS, we might miss it. Still, I think this in combination with checking the domain will cover most use cases and save our authors a lot of time. And if neither works, maybe the modal metadata approach could be a fallback. What do you think?
> 
> And finally, if all else fails you could do the modal metadata approach then. What do you think?
> 
> ```
> fetch(url)
>     .then(response => response.text())
>     .then(html => {
>       const parser = new DOMParser();
>       const doc = parser.parseFromString(html, 'text/html');
>       const header = doc.querySelector('h1, h2, h3, h4, h5, h6')?.textContent || 'No header found';
>       console.log(header);
>   })
> ```

Thanks for the suggestion, I tried this approach but I still ended up getting origin issues.

vhargrave:
> > Hey @zagi25 , I noticed you brought up something similar, and I’m on the same page. It might be worth checking if the iframe’s on the same origin first—grabbing the header that way could be a smooth option. I was also thinking about using a fetch request for pages on different domains. It’s got a solid shot at pulling the header, though if it relies on JS, we might miss it. Still, I think this in combination with checking the domain will cover most use cases and save our authors a lot of time. And if neither works, maybe the modal metadata approach could be a fallback. What do you think?
> > And finally, if all else fails you could do the modal metadata approach then. What do you think?
> > ```
> > fetch(url)
> >     .then(response => response.text())
> >     .then(html => {
> >       const parser = new DOMParser();
> >       const doc = parser.parseFromString(html, 'text/html');
> >       const header = doc.querySelector('h1, h2, h3, h4, h5, h6')?.textContent || 'No header found';
> >       console.log(header);
> >   })
> > ```
> 
> Thanks for the suggestion, I tried this approach but I still ended up getting origin issues.

makes sense. I don't know why I thought that a normal fetch would send less headers and not encounters CORS. That's just JSONP for example.

vhargrave (approved):

libs/blocks/adobetv/adobetv.js
```diff
@@ -12,12 +13,16 @@ export default function init(a) {
       anchorTag: a,
     });
   } else {
+    const sectionMetadata = a.parentElement.parentElement.parentElement?.querySelector('.section-metadata');
+    const metadataTitle = sectionMetadata ? getMetadata(sectionMetadata)?.title?.text : null;
+    const title = metadataTitle || 'Adobe Video Publishing Cloud Player';
+
```
overmyheadandbody, line 19: I'm not quite certain this is the way to go here. My two thoughts:
* `a.parentElement.parentElement.parentElement` seems a bit error-prone. We could instead rely on something like `a.closest('.section:has(.section-metadata)')`
* even more important, we can have multiple adobetv videos in a single section. I have a [test page](https://main--milo--adobecom.aem.page/drafts/ramuntea/accessibility/iframe-video-title) borrowed from a production use-case. Judging by this logic, all adobetv videos from the 4 Editorial Cards row will have the same title. Looking at this a while back, my conclusion was that this is yet another case where we could use the pipe pattern to define the title, such as `https://video.tv.adobe.com/v/3431684?ui=2&hidetitle=1 | My Iframe title`, so we can attach titles to individual videos

DKos95: Nice catch and thanks for the feedback!

libs/blocks/adobetv/adobetv.js
```diff
@@ -12,12 +12,16 @@ export default function init(a) {
       anchorTag: a,
     });
   } else {
+    const textContent = a?.textContent?.trim();
+    const titleFromText = textContent?.includes('|') ? textContent.split('|')[1].trim() : null;
+    const title = titleFromText || 'Adobe Video Publishing Cloud Player';
```
zagi25, line 17: NIT: you can remove the above ternary
```suggestion
    const title = textContent.split('|')?.[1]?.trim() || 'Adobe Video Publishing Cloud Player';
```

libs/blocks/iframe/iframe.js
```diff
@@ -47,6 +48,21 @@ export default function init(el) {
 
   const iframe = createTag('iframe', { src: linkHref, allowfullscreen: true });
   const embed = createTag('div', { class: `milo-iframe ${classes}` }, iframe);
+  const { parentElement } = el;
+
+  iframe.onload = () => {
+    if (new URL(iframe.src).origin !== window.location.origin) {
+      const metaDataElement = parentElement.querySelector('.section-metadata');
```
zagi25, line 55: @DKos95, @overmyheadandbody , @vhargrave can we also use `|` approach as in `adobetv` for overriding title via authoring ( or adding title if iframe comes from different origin) ? Then there will be no need to import code from `section-metadata.js` right ? 
Also based on the description

AC: important: Is it possible to provide an authorable over-ride? We have seen time and time again that there are going to be special cases along the way that don't produce the desire results with an automated solution.

It looks like override should be applied always if present

zagi25 (approved):

libs/blocks/vimeo/vimeo.js
```diff
@@ -73,7 +88,14 @@ export default async function init(a) {
   const embedVimeo = () => {
     const url = new URL(a.href);
     const videoid = url.pathname.split('/')[url.hostname === 'player.vimeo.com' ? 2 : 1];
-    const liteVimeo = createTag('lite-vimeo', { videoid });
+    const isTextLink = !a.textContent.includes('http');
+    const title = isTextLink ? a.textContent : 'Content from Vimeo';
+    const textAfterPipe = title?.split('|')[1]?.trim();
```
overmyheadandbody, line 93: Same here - https://github.com/adobecom/milo/blob/stage/libs/utils/utils.js#L802-L808

overmyheadandbody (commented):
Since we have similar logic across a few places, maybe we can centralize it a bit. But I'll let you decide whether you thinks this makes sense or we should consider it for another PR.

libs/blocks/adobetv/adobetv.js
```diff
@@ -12,12 +12,14 @@ export default function init(a) {
       anchorTag: a,
     });
   } else {
+    const title = a.getAttribute('aria-label') || 'Adobe Video Publishing Cloud Player';
```
overmyheadandbody, line 15: Nit: since we're not using this, we could just define it on L22

libs/blocks/vimeo/vimeo.js
```diff
@@ -5,15 +5,31 @@ import { createIntersectionObserver, createTag, getConfig, isInTextNode, loadLin
...
+  async fetchVideoTitle() {
+    if (this.title !== LiteVimeo.defaultTitle) return null;
+
+    try {
+      const response = await fetch(`https://vimeo.com/api/oembed.json?url=https://vimeo.com/${this.videoId}`);
+      const data = await response.json();
+      if (data.title && this.iframeEl) this.iframeEl.title = data.title;
+      return null;
+    } catch (error) {
+      return null;
```
overmyheadandbody, line 29: Nit: I think ESlint would accept simple `return` statements, as long as they're consistent throughout the function.

libs/blocks/youtube/youtube.js
```diff
@@ -19,6 +21,19 @@ class LiteYTEmbed extends HTMLElement {
     this.needsYTApiForAutoplay = navigator.vendor.includes('Apple') || this.isMobile;
   }
 
+  async fetchVideoTitle() {
+    if (this.playLabel !== LiteYTEmbed.defaultTitle) return null;
```
overmyheadandbody, line 25: I'm not sure I follow the logic between Vimeo and YouTube. On your [test page](https://adobe.sharepoint.com/:w:/r/sites/adobecom/_layouts/15/Doc.aspx?sourcedoc=%7B4F570997-5378-4427-83F5-3E5F88D554DE%7D&file=iframe-video-testing.docx&action=default&mobileredirect=true), Vimeo gets the fetched title, while YouTube doesn't. YouTube's remains "Youtube display text"; both seem to be authored in the same manner, a link with custom text. Am I missing something?

overmyheadandbody (commented):
I have a few notes, the Vimeo vs YT behavior is the more important one. I also noticed the video test URLs are reversed in the PR description, that confused me a bit when I opened them 😅

libs/blocks/iframe/iframe.js
```diff
@@ -48,6 +50,15 @@ export default function init(el) {
   const iframe = createTag('iframe', { src: linkHref, allowfullscreen: true });
   const embed = createTag('div', { class: `milo-iframe ${classes}` }, iframe);
 
+  iframe.onload = () => {
+    if ((new URL(iframe.src).origin !== window.location.origin) && ariaLabel) {
```
mirafedas, line 54: Just wanted to double check: if the first condition - (new URL(iframe.src).origin !== window.location.origin) - is true, but there is no ariaLabel on the anchor, on the line 59 we will get an error due to CORS. Is it guaranteed that an anchor will always have some ariaLabel?

robert-bogos (approved):

hit11757 (approved):
Validated. Test details https://jira.corp.adobe.com/browse/MWPW-169532

PR #3835, opened 2025-03-19
[MWPW-168529] New Assets tab in Preflight

overmyheadandbody:
## Summary

This proposes a new Assets tab for the Preflight tool to enable authors to easily check whether the image assets on a page are meeting our [asset governance](https://milo.adobe.com/docs/authoring/asset-governance) recommendations. In short, the images need to be uploaded at 2x (or 1.5x for full width images) the size of the area they will be used in.

There are two layers to how (mis)matches are communicated, detailed below.

## In the Preflight modal

This clearly separates the images into two categories: those that match and those that don't match the asset governance rules. A multiplication factor is outlined, along with uploaded image size, the size at which it is displayed and its recommended size (in case of a mismatch).

Screenshot 2025-03-19 at 17 04 58

## As an image overlay

Once the Preflight modal is closed, the image details are added over each individual image, so an author can easily isolate images that need to be updated. Note that placeholder images are used in the screenshot, the original size is already embedded in the image.

Screenshot 2025-03-19 at 17 16 20

## Other notes

- the logic will run only if the viewport width is larger than `1200px`, since this is the current breakpoint at which we switch to a desktop view. The multiplication factor for full width images is calculated based on a `1920px` viewport width. Certain blocks allow the definition of per-device assets, a future iteration of the Assets logic would be helpful to consider those cases as well;
- the image Accessibility logic has also been updated to be in line with the new Assets approach. A status area will now be overlayed over the images with details around its `alt` text. This replaces the previous approach of having a glowing yellow border around decorative images. The update also ensures the `alt` text is clearly visible, which wasn't happening before;
- only images nested inside a `picture` element from the `main` area are considered for the Assets checks. These are typically the ones served through Sharepoint and the content pipeline, although there are exceptions to this rule;
- the details overlay is semi-transparent until the image is hovered, to allow an author to better see the original asset. This is especially useful for small images;
- the logic allows for a 5% tolerance when calculating the multiplication factor;
- as a QA note, unrelated to this PR, a user needs to wait for the page to load before being able to open the Preflight tab.

Resolves: [MWPW-168529](https://jira.corp.adobe.com/browse/MWPW-168529)

**Test URLs:**
- Before: https://main--milo--overmyheadandbody.aem.page/drafts/ramuntea/preflight-assets-tab?martech=off
- After: https://assets-preflight-v2--milo--overmyheadandbody.aem.page/drafts/ramuntea/preflight-assets-tab?martech=off

**Other test pages:**
- All images too small: https://main--cc--adobecom.hlx.page/products/audition?martech=off&georouting=off&milolibs=assets-preflight-v2--milo--overmyheadandbody
- Correct & incorrect images, lots of `alt` text: https://main--cc--adobecom.hlx.page/products/photoshop-lightroom?martech=off&georouting=off&milolibs=assets-preflight-v2--milo--overmyheadandbody
- All images correct - https://main--bacom--adobecom.aem.page/support/adobe-partners?martech=off&georouting=off&milolibs=assets-preflight-v2--milo--overmyheadandbody

### DIFF
--- a/libs/blocks/preflight/panels/accessibility.js
+++ b/libs/blocks/preflight/panels/accessibility.js
@@ -1,4 +1,5 @@
 import { html, signal, useEffect } from '../../../deps/htm-preact.js';
+import { createTag } from '../../../utils/utils.js';
 
 const DEF_DESC = 'Checking...';
 const decorativeImages = signal([]);
@@ -46,14 +47,39 @@ async function checkAlt() {
   if (!images) return;
 
   images.forEach((img) => {
+    if (img.classList.contains('accessibility-control')) return;
     const alt = img.getAttribute('alt');
     let parent = '';
 
     if (img.closest('header')) parent = 'gnav';
     if (img.closest('main')) parent = 'main-content';
     if (img.closest('footer')) parent = 'footer';
+
+    let pictureMetaElem;
+    const picture = img.closest('picture');
+
+    if (picture) {
+      pictureMetaElem = picture.querySelector('.picture-meta');
+      if (!pictureMetaElem) {
+        pictureMetaElem = createTag('div', { class: 'picture-meta' });
+        picture.insertBefore(pictureMetaElem, img.nextSibling);
+      }
+    } else {
+      pictureMetaElem = createTag('div', { class: 'picture-meta no-picture-tag' });
+      img.parentNode.insertBefore(pictureMetaElem, img.nextSibling);
+    }
+
+    let a11yMessage;
+
     if (alt === '') {
-      img.dataset.altCheck = 'decorative';
+      img.dataset.altCheck = 'Decorative';
+
+      a11yMessage = createTag(
+        'div',
+        { class: 'picture-meta-a11y is-decorative' },
+        img.dataset.altCheck,
+      );
+
       decorativeImages.value = [...decorativeImages.value,
         {
           src: img.getAttribute('src'),
@@ -62,6 +88,12 @@ async function checkAlt() {
         }];
     }
     if (alt) {
+      a11yMessage = createTag(
+        'div',
+        { class: 'picture-meta-a11y has-alt' },
+        `Alt: ${alt}`,
+      );
+
       altTextImages.value = [...altTextImages.value,
         {
           src: img.getAttribute('src'),
@@ -69,6 +101,8 @@ async function checkAlt() {
           parent,
         }];
     }
+
+    pictureMetaElem.append(a11yMessage);
     img.dataset.pageLocation = parent;
   });
   result.description = 'All images listed below. Please validate each alt text has been set appropriately. Decorative images have been highlighted in yellow on the page.';
@@ -99,7 +133,7 @@ function ImageGroups({ group }) {
     ${imgArray.value.length > 0 && html`
     <div class="access-image-grid ${setFilterView.value}">
       <div class="access-image-grid-item filter">
-        Filter images by: 
+        Filter images by:
         <select onChange=${(e) => filterGrid(e)} class="image-filter">
         ${filterOptions.map((option) => html`<${dropdownOptions} option=${option} />`)}
         </select>
--- a/libs/blocks/preflight/panels/accessibility.js
+++ b/libs/blocks/preflight/panels/accessibility.js
@@ -42,12 +42,11 @@ function dropdownOptions(props) {
 async function checkAlt() {
   if (altResult.value.checked) return;
   // If images are not scoped, tracking pixel/images are picked up.
-  const images = document.querySelectorAll('header img, main img, footer img');
+  const images = document.querySelectorAll(':is(header, main, footer) img:not(.accessibility-control)');
   const result = { ...altResult.value };
   if (!images) return;
 
   images.forEach((img) => {
-    if (img.classList.contains('accessibility-control')) return;
     const alt = img.getAttribute('alt');
     let parent = '';
 
--- a/libs/blocks/preflight/panels/assets.js
+++ b/libs/blocks/preflight/panels/assets.js
@@ -5,6 +5,7 @@ function isViewportTooSmall() {
   return !window.matchMedia('(min-width: 1200px)').matches;
 }
 
+const maxFullWidth = 1920;
 const imagesWithMismatch = signal([]);
 const imagesWithMatch = signal([]);
 const checksPerformed = signal(false);
@@ -29,6 +30,7 @@ async function checkImageDimensions() {
         img.setAttribute('loading', 'eager');
         return new Promise((resolve) => {
           img.addEventListener('load', resolve);
+          img.addEventListener('error', resolve);
         });
       }
       return Promise.resolve();
@@ -38,12 +40,14 @@ async function checkImageDimensions() {
   // Filter the loaded images to ensure they are visible, not an icon, and not an SVG
   const images = allImages.filter((img) => img.checkVisibility()
     && !img.closest('.icon-area')
-    && !img.src.endsWith('.svg'));
+    && !img.src.includes('.svg'));
 
   if (!images.length) return;
 
-  const maxFullWidth = 1920;
   const viewportWidth = document.documentElement.clientWidth;
+
+  // Apply overflow class to body to prevent scrolling issues during image analysis
+  document.body.classList.add('preflight-assets-analysis');
 
   for (const img of images) {
     // Get the original dimensions of the uploaded image; fallback to the natural dimensions
@@ -104,6 +108,9 @@ async function checkImageDimensions() {
 
     pictureMetaElem.append(assetMessage);
   }
+
+  // Remove overflow class from body after analysis is complete
+  document.body.classList.remove('preflight-assets-analysis');
 }
 
 function AssetsItem({ title, description }) {
--- a/libs/blocks/preflight/panels/assets.js
+++ b/libs/blocks/preflight/panels/assets.js
@@ -1,0 +1,111 @@
+import { html, signal, useEffect } from '../../../deps/htm-preact.js';
+
+const imagesWithMismatch = signal([]);
+const imagesWithMatch = signal([]);
+const checksPerformed = signal(false);
+
+const groups = [
+  { title: 'Images with dimension mismatch', imgArray: imagesWithMismatch },
+  { title: 'Images with matching dimensions', imgArray: imagesWithMatch },
+];
+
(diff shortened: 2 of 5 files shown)


libs/blocks/preflight/panels/assets.js
```diff
@@ -0,0 +1,188 @@
+import { html, signal, useEffect } from '../../../deps/htm-preact.js';
+import { createTag } from '../../../utils/utils.js';
+
+function isViewportTooSmall() {
```
narcis-radu, line 4: why not
```js
const isViewportTooSmall = () => !window.matchMedia('(min-width: 1200px)').matches;
```

libs/blocks/preflight/panels/assets.js
```diff
@@ -0,0 +1,188 @@
...
+
+  const allImages = [...document.querySelectorAll('main picture img')];
+  if (!allImages.length) return;
+
+  // Force load all images
+  await Promise.all(
+    allImages.map((img) => {
+      if (!img.complete) {
+        img.setAttribute('loading', 'eager');
+        return new Promise((resolve) => {
```
narcis-radu, line 30: suggestion
```js
return new Promise(resolve => img.addEventListener('load', resolve));
```

libs/blocks/preflight/panels/assets.js
```diff
@@ -0,0 +1,188 @@
...
+  );
+
+  // Filter the loaded images to ensure they are visible, not an icon, and not an SVG
+  const images = allImages.filter((img) => img.checkVisibility()
+    && !img.closest('.icon-area')
+    && !img.src.endsWith('.svg'));
+
+  if (!images.length) return;
+
+  const maxFullWidth = 1920;
```
narcis-radu, line 45: Can we define this at the top of the script?

narcis-radu (approved):
This is great, thank you! I have some nits, but everything looks good.

libs/blocks/preflight/panels/assets.js
overmyheadandbody: This is mostly because the Preflight logic in general uses `function` definitions, I wanted to stick to that pattern.

robert-bogos (approved):

libs/blocks/preflight/panels/assets.js
```diff
@@ -0,0 +1,188 @@
...
+
+async function checkImageDimensions() {
+  if (checksPerformed.value || viewportTooSmall.value) return;
+  checksPerformed.value = true;
+
+  const allImages = [...document.querySelectorAll('main picture img')];
+  if (!allImages.length) return;
+
+  // Force load all images
+  await Promise.all(
```
vhargrave, line 26: If any image fails to load, the Promise.all() will reject and the entire image checking process will stop. It probably would make sense to catch this type of case.

libs/blocks/preflight/panels/assets.js
```diff
@@ -0,0 +1,188 @@
...
+        });
+      }
+      return Promise.resolve();
+    }),
+  );
+
+  // Filter the loaded images to ensure they are visible, not an icon, and not an SVG
+  const images = allImages.filter((img) => img.checkVisibility()
+    && !img.closest('.icon-area')
+    && !img.src.endsWith('.svg'));
```
vhargrave, line 41: I'd be a bit careful for malformed .svg files that have query params or even a hash at the end. 
image

libs/blocks/preflight/panels/assets.js
```diff
@@ -0,0 +1,188 @@
...
+
+  const maxFullWidth = 1920;
+  const viewportWidth = document.documentElement.clientWidth;
+
+  for (const img of images) {
+    // Get the original dimensions of the uploaded image; fallback to the natural dimensions
+    const naturalWidth = img.getAttribute('width') ? parseInt(img.getAttribute('width'), 10) : img.naturalWidth;
+    const naturalHeight = img.getAttribute('height') ? parseInt(img.getAttribute('height'), 10) : img.naturalHeight;
+    // Get the display dimensions of the image
+    const displayWidth = img.offsetWidth;
```
vhargrave, line 53: Not a big deal, but this would be more efficient if you just got the getBoundingClientRect() to avoid layout thrashing. 
Getting the offsetWidth and offsetHeight separately causes two different DOM calculations. 

```
const rect = img.getBoundingClientRect();
const displayWidth = rect.width;
const displayHeight = rect.height;
```

libs/blocks/preflight/panels/accessibility.js
```diff
@@ -46,14 +47,39 @@ async function checkAlt() {
   if (!images) return;
 
   images.forEach((img) => {
+    if (img.classList.contains('accessibility-control')) return;
```
vhargrave, line 50: it would be a little more efficient to just not select these images in the first place above - 
`const images = document.querySelectorAll('header img:not(.accessibility-control), main img:not(.accessibility-control), footer img:not(.accessibility-control)');`

libs/blocks/preflight/panels/assets.js
```diff
@@ -0,0 +1,188 @@
...
+const groups = [
+  { title: 'Images with dimension mismatch', imgArray: imagesWithMismatch },
+  { title: 'Images with matching dimensions', imgArray: imagesWithMatch },
+];
+
+async function checkImageDimensions() {
+  if (checksPerformed.value || viewportTooSmall.value) return;
+  checksPerformed.value = true;
+
+  const allImages = [...document.querySelectorAll('main picture img')];
```
vhargrave, line 22: I’ve noticed that headers sometimes include images too. It might be a good idea to check those as well. What’s your take on that?

vhargrave (commented):
This looks great! I was thinking it might benefit from a bit of test coverage too. What do you think?

libs/blocks/preflight/panels/accessibility.js
overmyheadandbody: This was an omission for the initial iteration of the logic, but I can patch it up, thanks. I'll go with a `':is(header, main, footer) img:not(.accessibility-control)'` to keep things succint.

libs/blocks/preflight/panels/assets.js
overmyheadandbody: Right now the big pain points revolve around marquee and aside images, which are also the ones that are instantly visible to the user without performing any additional action. Moreover, the browser doesn't have good layout data for hidden areas, so we'd have to force display any such areas in order to perform calculations. For the MVP, I think it's ok to ignore the `header` section.

libs/blocks/preflight/panels/assets.js
overmyheadandbody: This leads to an ESLint error `Return values from promise executor functions cannot be read`, so it's more straightforward to use the current option.

libs/blocks/preflight/panels/assets.js
overmyheadandbody: Good catch, I'll use an `includes` instead.

libs/blocks/preflight/panels/assets.js
overmyheadandbody: Yes, makes sense 👍

libs/blocks/preflight/panels/assets.js
overmyheadandbody: This is quite debatable, we've had requests in the past to move away from that approach, as it was leading to noticeable reflow issues. This is the first article I could find, but there are a bunch more - https://toruskit.com/blog/how-to-get-element-bounds-without-reflow. One could argue that we can move to using an IntersectionObserver, but I think that's overkill for this particular context of non-customer facing logic.

libs/blocks/preflight/panels/assets.js
overmyheadandbody: True! Instead of a `catch`, I'll add an error listener for the image itself, should be more straightforward.

overmyheadandbody (commented):
Thanks for the feedback! I made some notes and some changes 🙂

libs/blocks/preflight/panels/assets.js
vhargrave: Sounds good. If this is going to get built out then , then we'll probably also want to some logic for images hidden behind tabs and other such cases.

libs/blocks/preflight/panels/assets.js
vhargrave: That's cool, I didn't consider an intersection observer as well. 
yes, this is definitely a nit. :)

libs/blocks/preflight/panels/assets.js
overmyheadandbody: Yes, that's a great point! I'll create a due diligence to iterate over this solution and also include this caveat that we'll need to consider hidden section as well in the future or at least surface hidden images that could not be analyzed.

vhargrave (approved):
looks good!

overmyheadandbody:
@vhargrave - I also created https://jira.corp.adobe.com/browse/MWPW-170378 to follow up with Unit Tests for this functionality. Right now just the [Performance tab](https://github.com/adobecom/milo/tree/stage/test/blocks/preflight) has tests, but if Preflight is to grow, we'll need to ensure its overall stability as well.

robert-bogos (changes requested):
There seems to be a discrepancy between the actual display size of one of the images and what Preflight is reading. This is also affecting the recommended size calculations.

At first sight, it seems to be related to Preflight removing the browser scrollbar, but it is not obvious why it is only affecting this specific image.

Screenshot 2025-03-27 at 09 21 20

Screenshot 2025-03-27 at 09 22 57

overmyheadandbody:
@robert-bogos - thanks for noticing this! It is somewhat of an edge-case for blocks using the `.grid-width-N` class when the user's system is set to display scrollbars (opt-in on Mac, always on on Windows). This is because opening the Preflight modal removes the `body` scrollbars, thus changes the viewport dimensions; which shouldn't cause issues for most sections, but the `.grid-width-N` ones are an outlier.

I've added logic to set the scrollbars to `auto` while images sizes are calculated, to ensure we're getting the actual size a user would get.

This really would have been difficult to catch due to the multiple factors involved, thanks for bringing it up!

robert-bogos:
Nice! Thank you for handling this 🙏🏻

robert-bogos (approved):

zagi25 (approved):

SilviuLCF (approved):
Testing details https://jira.corp.adobe.com/browse/MWPW-168529

PR #3844, opened 2025-03-20
Automated aria labels

vhargrave:
Adds a feature which automatically adds aria labels to CTAs when possible in the delayed.js after page load. This feature does not add an aria label if there already is one. 
**Important**: full configuration is done for the US texts on CC & DC. Full translations and configuration for all other consumers & locales are still in the works. This will be a config change in an excel though, and no longer a code change :)

Technical Documentation for this is [here](https://wiki.corp.adobe.com/display/WP4/Proposal)

This is the implementation of a prototype which was previously done here - https://jira.corp.adobe.com/browse/MWPW-165540
This feature has been vetted with multiple aria stakeholders and our aria expert Taryn Sikorski. 
Crawls were previously done over multiple milo consumers (CC, DC, Bacom & Express) and a proposal was created for all the ctas that were picked up and missing aria labels, for how aria-labels could be defined. This proposal has been accepted and this is the implementation thereof. 

Resolves: [MWPW-168985](https://jira.corp.adobe.com/browse/MWPW-168985)

**Test URLs:**
- After: https://vhargrave-automated-aria-labels--milo--adobecom.aem.page/?martech=off

**look at the primary CTAs in the hero-marquee**
- Before: https://main--cc--adobecom.aem.live/products/photoshop-lightroom/edit-photos/?martech=off
- After: https://main--cc--adobecom.aem.live/products/photoshop-lightroom/edit-photos&milolibs=vhargrave-automated-aria-labels

**look at the CTAs in the hero-marquee**
- Before: https://main--cc--adobecom.aem.live/ch_de/products/photoshop-lightroom/edit-photos/?martech=off
- After: https://main--cc--adobecom.aem.live/ch_de/products/photoshop-lightroom/edit-photos&milolibs=vhargrave-automated-aria-labels

**look at the learn more buttons**
- Before: https://main--cc--adobecom.aem.live/careers/teams?martech=off
- After: https://main--cc--adobecom.aem.live/careers/teams?milolibs=vhargrave-automated-aria-labels&milolibs=vhargrave-automated-aria-labels

**look at try for free button**
- Before: https://main--cc--adobecom.hlx.live/products/substance3d/discover/3d-files-formats?martech=off
- After: https://main--cc--adobecom.hlx.live/products/substance3d/discover/3d-files-formats?milolibs=vhargrave-automated-aria-labels

**look at start free trial button**
- Before: https://main--dc--adobecom.aem.live/acrobat/business/resources?martech=off
- After: https://main--dc--adobecom.aem.live/acrobat/business/resources?milolibs=vhargrave-automated-aria-labels

**look at the learn more buttons**
- Before: https://main--bacom--adobecom.aem.live/customer-success-stories/lenovo-case-study?martech=off
- After: https://main--bacom--adobecom.aem.live/customer-success-stories/lenovo-case-study?milolibs=vhargrave-automated-aria-labels

There are thousands more test cases. To find them, go to https://wiki.corp.adobe.com/display/WP4/Data, download cc or dc zip folders, and look at the us files for these projects. There you can filter for where proposed aria label is defined. These are the test cases where this code should also be setting aria labels. 
I've confirmed that I'm getting matches for CC already on


libs/utils/automated-aria.js
```diff
@@ -0,0 +1,263 @@
...
+  if (!block) return '';
+
+  let text = '';
+  let foundHeader = false;
+
+  const traverse = (node) => {
+    if (foundHeader) return;
+
+    if (node.nodeType === Node.ELEMENT_NODE) {
+      if (/^h[1-2]$/i.test(node.tagName)) {
```
chrischrischris, line 13: nit: set regex as const outside traverse

libs/utils/automated-aria.js
```diff
@@ -0,0 +1,263 @@
...
+    const parts2 = cn2.split('-');
+    const hasPrefixMatch = hasCommonPrefix && parts1[0] === parts2[0];
+    // eslint-disable-next-line max-len
+    const hasSuffixMatch = hasCommonSuffix && parts1[parts1.length - 1] === parts2[parts2.length - 1];
+    return parts1.length > 1 && parts2.length > 1 && (hasPrefixMatch || hasSuffixMatch);
+  }));
+};
+
+// Helper to check if divs have a common pattern (prefix or suffix)
+const hasCommonPattern = (divs, usePrefix) => {
```
chrischrischris, line 51: nit: use object for bool param
```suggestion
const hasCommonPattern = (divs, { usePrefix = false } = {}) => {
```

chrischrischris (approved):

libs/utils/automated-aria.js
```diff
@@ -0,0 +1,263 @@
...
+  const checkFn = (parts) => getPatternPart(parts, usePrefix);
+  const classNames = divs.flatMap((div) => Array.from(div.classList));
+  const patterns = new Set(classNames
+    .map((cn) => cn.split('-'))
+    .filter((parts) => parts.length > 1)
+    .map(checkFn));
+  return patterns.size > 0 && divs.some((div1) => divs
+    .some((div2) => div1 !== div2 && divsSharePattern(div1, div2, usePrefix, !usePrefix)));
```
sharmrj, line 59: Similar to my comment about `divsSharePattern`, this can be simplified. There also seems to be some duplication of logic between what this function is trying to do and what `divsSharePattern` is trying to do. We can generalize

```javascript
const getPartSet = (div) => {
  const xs = [...div.classList];
  const part = new Set();
  for (const x in xs) {
    const s = x.split('-');
    if (s.length === 1) continue; // don't count classes without '-'. 
                                  // If either set is empty then the result
                                  // is guaranteed to be disjoint
    const index = usePrefix ? 0 : s.length - 1;
    part.add(s[index]);
  }
  return part
};

const partSets = divs.map(getPartSet);
// sets from a collection are pairwise disjoint iff 
// the size of their union equals the sum of their sizes
const union = partSets.reduce((accumulator, partSet) => accumulator.union(partSet), new Set());
const sumOfParts = partSets.reduce((acc, x) => acc + x.size, 0);
return sumOfParts === union.size
```

This obviates, afaict, the need for `divsSharePattern`.

sharmrj (commented):
I still haven't wrapped my brain around the whole thing. Apologies if I misunderstood the purpose of certain functions.

libs/utils/automated-aria.js
```diff
@@ -0,0 +1,263 @@
...
+  const div2Classes = Array.from(div2.classList);
+  return div1Classes.some((cn1) => div2Classes.some((cn2) => {
+    const parts1 = cn1.split('-');
+    const parts2 = cn2.split('-');
+    const hasPrefixMatch = hasCommonPrefix && parts1[0] === parts2[0];
+    // eslint-disable-next-line max-len
+    const hasSuffixMatch = hasCommonSuffix && parts1[parts1.length - 1] === parts2[parts2.length - 1];
+    return parts1.length > 1 && parts2.length > 1 && (hasPrefixMatch || hasSuffixMatch);
+  }));
+};
```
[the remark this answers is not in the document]
vhargrave: Hey @sharmrj !
Thanks for the suggestion! I do see the appeal of your approach (avoiding nested loops and leverages Sets for a cleaner theoretical complexity (O(n + m) vs. O(n * m))). I think it would really shine if there were large class lists, however, I’m hesitant to adopt it here for several reasons. 
1. isDisjointFrom is a very new method which [older browser versions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/isDisjointFrom#browser_compatibility) don't support. This means I would need to add a polyfill for this method.
2. The original code is concise, readable, and performs well for our typical case—small class lists where the nested loop is sub-millisecond and not a bottleneck. Without `isDisjointFrom`, your version adds extra functions and lines, increasing maintenance overhead for negligible real-world gain, especially since this is all happening in the delayed.js method. I’d rather keep it simple unless profiling shows a need to optimize.

libs/utils/automated-aria.js
vhargrave: Here kind of the same comment as above. I highly appreciate the suggestions, and do see the advantages. The reasons why I'm hesistant to pull through on them though are: 
Set.union is a new method, so I'd need a polyfill for older browser support - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/union

Performance gains are very small vs maintainability tradeoffs

The original hasCommonPattern and divsSharePattern combo is tailored to ensure distinct sibling divs share patterns, which aligns with findBlockContainers’ intent—your approach detects any overlap but might not guarantee pairwise matches between different divs because it checks for overall duplication across the collection rather than confirming that at least two distinct divs share a specific prefix or suffix, as required for identifying repeating container patterns among siblings
For example, if one div has classes `card-foo card-bar` (two classes with prefix card), your code sees card twice in that div’s Set, and the sum of sizes exceeds the union, flagging overlap—even though it’s within one div, not between two different ones.
The original, however, only counts a match if `card-foo` is on one div and `card-bar` is on a different div.

libs/utils/automated-aria.js
```diff
@@ -0,0 +1,263 @@
...
+    textBeforeHeader,
+  );
+};
+
+/**
+ * Asynchronously adds ARIA labels to all eligible CTAs on the page.
+ * Fetches configuration and product names, then applies labels based on locale.
+ */
+export default async function addAriaLabels() {
+  const selector = getConfig().ariaLabelCTASelector || '.con-button:not([aria-label])';
```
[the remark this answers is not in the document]
vhargrave: Very good point. I'll add some code to handle that case too thanks!

libs/utils/automated-aria.js
sharmrj: Your point is well received, but a small point of order. My code cannot see card twice in that div's `Set` because `Set`s don't allow duplicate entries. In the example you give, that `<div class="card-foo card-bar">` will be processed into `Set(1) {'card'}`. The size of the union will be equal to the sum of sizes with that one div. The two snippets are identical in behavior.
Your point about needing polyfills is correct.

mokimo (commented):
Another super complex topic you take on & providing a lot of value in terms of accessibility, lovely! 🥳 

Similar to the feedback on the other PR - the code is very hard to digest. I don't want to limit this review to just a very high level "the feature works" check on the provided links - but ideally be able to give proper input on the code.

Common things that I spot are all related to "code-complexity"
- Variable namings
```js
// You have parts, "cns" based on splits, prefixes, suffixes, only if there is a length and/or match... 
// Without a deep dive, I have no idea what this does, or why things are the way they are
const parts1 = cn1.split('-');
const parts2 = cn2.split('-');
const hasPrefixMatch = hasCommonPrefix && parts1[0] === parts2[0];
// eslint-disable-next-line max-len
const hasSuffixMatch = hasCommonSuffix && parts1[parts1.length - 1] === parts2[parts2.length - 1];
return parts1.length > 1 && parts2.length > 1 && (hasPrefixMatch || hasSuffixMatch);
```
_It's okay and often necessary to have complex conditions - but it's your job to make this as easy on whoever needs to step into this code in the future and understand it as possible. "divShareClassPattern" is already great, since the function tells me everything I need without looking into it and thus **hides implementation complexity**. However I should also quickly able to understand everything thats inside of it_
- Comments: Those are necessary given the overall complexity, code however, should be self-documenting. So those are a byproduct of "Variable namings & complexity"
- A lot of functions, parameters, if/else statements... Could you shuffle code around to make this less complex?

libs/utils/automated-aria.js
```diff
@@ -0,0 +1,263 @@
...
+    hasCommonPattern([div, otherDiv], true) // has common prefix
+        || hasCommonPattern([div, otherDiv], false) // has common suffix
+  )));
+
+  // Require at least two matching siblings to qualify as containers
+  if (patternDivs.length > 1) {
+    const deepestContainers = processNestedContainers(patternDivs, findBlockContainers);
+    // Return containers only if multiple found, otherwise fallback to [container]
+    return deepestContainers.length > 1 ? deepestContainers : [container];
+  }
```
mokimo, line 97: A simple trick to move code closer to where it belongs:
```js
function processNestedContainers (patternDivs, findBlockContainers) {
  const patternDivs = directDivs.filter((div) => directDivs.some((otherDiv) => div !== otherDiv && (
    hasCommonPattern([div, otherDiv], true) // has common prefix
        || hasCommonPattern([div, otherDiv], false) // has common suffix
  )));
 if(!patternDivs) return
 if(deepestContainer) return [deepestContainer]
 return [container]
}

const nestedContainer = processNestedContainers(patternDivs, findBlockContainers);
if (nestedContainer) return nestedContainer
```

If you rinse&repeat this across all functions, the code ideally becomes A LOT slimmer, if/else cases mostly vanish, you avoid passing around too many parameters ... refactoring to move code closer to where it's needed and embedding it directly where it is used can do a lot of wonders

libs/utils/automated-aria.js
```diff
@@ -0,0 +1,263 @@
...
+
+  // Filter divs that share a common pattern (prefix or suffix)
+  const patternDivs = directDivs.filter((div) => directDivs.some((otherDiv) => div !== otherDiv && (
+    hasCommonPattern([div, otherDiv], true) // has common prefix
+        || hasCommonPattern([div, otherDiv], false) // has common suffix
+  )));
+
+  // Require at least two matching siblings to qualify as containers
+  if (patternDivs.length > 1) {
+    const deepestContainers = processNestedContainers(patternDivs, findBlockContainers);
```
mokimo, line 94: Do you need to pass in `findBlockContainers` ? Can't you just call the function

mokimo, line 94: I wonder if you could just pass in `container` and even regenerate `Array.from(container.children).filter((el) => el.tagName === 'DIV')` 🤔

libs/utils/automated-aria.js
vhargrave: Agree - I've spent some time going through the file to make it more readable and understandable. I think it's better, but let me know :)

libs/utils/automated-aria.js
vhargrave: Yes, unfortunately it's necessary. The issue is findDeepestRepeatingContainers is called by findBlockContainers and must therefore appear before. But findDeepestRepeatingContainers also calls findBlockContainers since it's recursive, and findBlockContainers appears after.

vhargrave: It could work but I'm not sure what benefit we'd gain. What made you think of it?

libs/scripts/delayed.js
```diff
@@ -62,6 +62,11 @@ export const loadGoogleLogin = async (getMetadata, loadIms, loadScript, getConfi
   initGoogleLogin(loadIms, getMetadata, loadScript, getConfig);
 };
 
+export const loadAriaAutomation = async () => {
+  const { default: addAriaLabels } = await import('../utils/automated-aria.js');
```
mokimo, line 66: Just one Q, does *every* consumer need that? I wonder if say bacom, blog, internal consumers, partner portals need to load this.
Maybe we should add a config for consumers to opt-into this feature and ensure CC/DC/HP/Bacom use it?

mokimo (approved):

libs/scripts/delayed.js
vhargrave: @mokimo It's a really good question. Originally I actually even built this with every consumer having different configs, but after discussing it with Taryn we decided to make it global. There are a couple of reasons for this: 
1. The button texts we've chosen are clearly ambiguous and need an aria label applied. The aria labels that are applied are simple, and should pretty much always be making things better instead of worse. 
2. This is to help get all our sites ready by the deadline. I doubt that many of the consumers will get the memo and get this done by then otherwise. 
3. It takes away more effort from our consumers figuring out what texts should have which rules. This was something that I noticed would probably be a little difficult for them, which is why Taryn and I defined the rules for them. 
4. This is a temporary solution. For future pages @overmyheadandbody & I agreed with GWP that they'll be correctly and manually setting aria-labels if necesary.

mokimo: Just want to be a bit careful of shipping features that aren't interesting for everyone

vhargrave: Maybe just to make you feel a little safer, this will only add aria labels to ctas that contain texts, which Taryn and I have clearly defined as being too ambiguous. You can see the list of texts [here](https://main--federal--adobecom.aem.page/federal/assets/data/cta-aria-label-config.json). It will also never change aria labels for ctas that already have existing aria-labels.

vhargrave: do you have an example that I can check out?

mokimo: I'm still firm about not everyone needing this - but as the majority needs this and there is no (real) downside to it, I don't mind it either. So we can consider this resolved

libs/utils/automated-aria.js
```diff
@@ -0,0 +1,289 @@
...
+    parts.add(segments[partIndex]);
+  }
+}
+
+// Extracts either prefixes or suffixes from a div's class names into a Set
+const extractClassParts = (div, extractPrefix) => {
+  const classNames = Array.from(div.classList);
+  const parts = new Set();
+  for (const className of classNames) {
+    addClassPartToSet(className, parts, extractPrefix);
```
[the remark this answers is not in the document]
vhargrave: 100% yes, I deliberately refactored to a separate method to make the code a little more readable. I think it's better to leave like that since I think the readability trumps the simplicity of inlining in this case.

libs/utils/automated-aria.js
```diff
@@ -0,0 +1,289 @@
...
+// Main function to check for shared class patterns
+const hasSharedClassPattern = (divs, { extractPrefix = false } = {}) => {
+  if (divs.length < 2) return false; // Need at least 2 divs to share anything
+  const partToDivIndices = mapPartsToDivIndices(divs, extractPrefix);
+  return Array.from(partToDivIndices.values()).some((divIndices) => divIndices.size > 1);
+};
+
+function addDeepestContainers(div, findRepeatingContainers, deepestContainers) {
+  const nestedContainers = findRepeatingContainers(div);
+  if (nestedContainers.length === 1 && nestedContainers[0] === div) {
```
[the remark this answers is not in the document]
vhargrave: sure!

narcis-radu (approved):

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-168985

PR #3860, opened 2025-03-25
MWPW-170175: Dynamic Nav Reset

JasonHowellSlavin:
* Adds a "reset" functionality to the dynamic nav, allowing pages to reset the experience entirely
* Adds the dispatching of a custom event when the global-navigation is done loading to address code execution order issue with status.js 

Resolves: [MWPW-MWPW-170175](https://jira.corp.adobe.com/browse/MWPW-MWPW-170175)

**Test URLs:**
- Before: https://main--milo--jasonhowellslavin.hlx.page/drafts/slavin/dynamic-nav/bacom-sample-on?martech=off
- After: https://dynamic-nav-reset--milo--jasonhowellslavin.hlx.page/drafts/slavin/dynamic-nav/bacom-sample-on?martech=off

Bacom
- Before: https://main--bacom--adobecom.hlx.page/drafts/slavin/dynamic-nav/aem-rtcdp-entry?martech=off
- After: https://main--bacom--adobecom.hlx.page/drafts/slavin/dynamic-nav/aem-rtcdp-entry?milolibs=dynamic-nav-reset&martech=off

### DIFF
--- a/libs/features/dynamic-navigation/dynamic-navigation.js
+++ b/libs/features/dynamic-navigation/dynamic-navigation.js
@@ -23,21 +23,25 @@ function dynamicNavGroupMatches(groupMetaData) {
 }
 
 export default function dynamicNav(url, key) {
+  const storageSource = 'gnavSource';
+  const storageKey = 'dynamicNavKey';
+  const storageGroup = 'dynamicNavGroup';
+
   if (foundDisableValues()) return url;
   const metadataContent = getMetadata('dynamic-nav');
   const dynamicNavGroup = getMetadata('dynamic-nav-group');
 
   if (metadataContent === 'reset') {
-    window.sessionStorage.removeItem('gnavSource', url);
-    window.sessionStorage.removeItem('dynamicNavKey', key);
-    window.sessionStorage.removeItem('dynamicNavGroup', dynamicNavGroup);
+    window.sessionStorage.removeItem(storageSource, url);
+    window.sessionStorage.removeItem(storageKey, key);
+    window.sessionStorage.removeItem(storageGroup, dynamicNavGroup);
     return url;
   }
 
   if (metadataContent === 'entry') {
-    window.sessionStorage.setItem('gnavSource', url);
-    window.sessionStorage.setItem('dynamicNavKey', key);
-    if (dynamicNavGroup) window.sessionStorage.setItem('dynamicNavGroup', dynamicNavGroup);
+    window.sessionStorage.setItem(storageSource, url);
+    window.sessionStorage.setItem(storageKey, key);
+    if (dynamicNavGroup) window.sessionStorage.setItem(storageGroup, dynamicNavGroup);
     return url;
   }
 
@@ -45,7 +49,7 @@ export default function dynamicNav(url, key) {
     if (!dynamicNavGroupMatches(dynamicNavGroup)) return url;
   }
 
-  if (metadataContent !== 'on' || key !== window.sessionStorage.getItem('dynamicNavKey')) return url;
+  if (metadataContent !== 'on' || key !== window.sessionStorage.getItem(storageKey)) return url;
 
-  return window.sessionStorage.getItem('gnavSource') || url;
+  return window.sessionStorage.getItem(storageSource) || url;
 }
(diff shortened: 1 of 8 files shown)


libs/features/dynamic-navigation/status.js
```diff
@@ -138,4 +141,17 @@ export default async function main() {
...
+  const { dynamicNavKey } = getConfig();
+
+  if (document.querySelector('.feds-nav-wrapper')) {
+    loadWidget(dynamicNavKey);
+    return;
+  }
+
+  window.addEventListener('feds:navLoaded', () => {
+    loadWidget(dynamicNavKey);
+  });
```
sharmrj, line 156: We don't usually encourage adding events to the global navigation code (For example: #2583). 

Wouldn't this functionally be the same as calling this `main` function at the end of the dynamic nav code? The way I understand it, you don't need to wait until the global-navigation loads, just until the dynamic nav finishes its work, which is at the very beginning of global-navigation when we do `getGnavSource()`. So you could do

```javascript
// Inside getGnavSource
url = dynamicNav(url, dynamicNavKey);
queueMicrotask(loadStatus);
//...
```

bandana147: I agree with @sharmrj, please check if it can be done without adding the event.

JasonHowellSlavin: Hi Raghav, 

The status widget is currently appending to `document.querySelector('.feds-nav-wrapper')`. Even though it is in the `loadDeffered` [function of utils.js](https://github.com/adobecom/milo/blob/stage/libs/utils/utils.js#L1335), I did notice some times when there would be a race condition where the gnav had not yet loaded, and so the status widget did not append. 

I don't really want to import status.js anywhere else because it is only really for use in preview and live environments (for instance, I would not want to import into global-navigation.js).

If the preference is to not use events in global-navigation (and I can forsee some conditions where the event might even fire before everything is loaded), maybe it is better to change where and how the dynamic-nav-status widget appends itself to the page and use CSS to put it where it may be less obtrusive as opposed to appending it to the global-navigation proper. What do you think? 

Always appreciate your feedback! 

Thank you,
Jason

JasonHowellSlavin: I've removed the customEvent and have removed the dependency of status.js on global-navigation loading.

sharmrj: > I did notice some times when there would be a race condition where the gnav had not yet loaded, and so the status widget did not append

Since the gnav init function is an`async` function, it returns a `Promise<Gnav>` (`Gnav` is the name of our Class). You can allocate that and pass it through to `loadDeferred` somehow (either explicitly or implicitly) and `await` it there

sharmrj (approved):

libs/features/dynamic-navigation/dynamic-navigation.js
```diff
@@ -27,6 +27,13 @@ export default function dynamicNav(url, key) {
   const metadataContent = getMetadata('dynamic-nav');
   const dynamicNavGroup = getMetadata('dynamic-nav-group');
 
+  if (metadataContent === 'reset') {
+    window.sessionStorage.removeItem('gnavSource', url);
+    window.sessionStorage.removeItem('dynamicNavKey', key);
+    window.sessionStorage.removeItem('dynamicNavGroup', dynamicNavGroup);
```
meganthecoder, line 33: Nit: Do you want to add the keys as variables if you're using them more than once?

test/features/dynamic-nav/status.test.js
```diff
@@ -172,6 +182,24 @@ describe('Dynamic Nav Status', () => {
       expect(info.storedSource).to.equal('/test');
     });
 
+    it('displays the correct information to the user for the enabled state "reset"', () => {
+      document.querySelector('meta[name="dynamic-nav"]').setAttribute('content', 'reset');
+      document.querySelector('meta[name="gnav-source"]').setAttribute('content', 'https://main--milo--adobecom.hlx/test');
```
meganthecoder, line 187: Probably doesn't matter but could update the link to `.aem.`

meganthecoder (approved):

Brandon32 (approved):

robert-bogos (approved):

PR #3870, opened 2025-03-26
MWPW-166214 Ability to author annualized price independently (AU offers only)

bozojovicic:
* Add your
* Specific
* Features or fixes

Resolves: [MWPW-166214](https://jira.corp.adobe.com/browse/MWPW-166214)

**Test URLs:**
- After: https://mwpw166214au--milo--bozojovicic.aem.live/?martech=off


libs/blocks/merch/merch.js
```diff
@@ -683,6 +683,7 @@ export async function getPriceContext(el, params) {
   const displayPerUnit = params.get('seat');
   const displayRecurrence = params.get('term');
   const displayTax = params.get('tax');
+  const displayAnnual = getConfig().locale?.prefix === '/au' && (params.get('annual') || !getMetadata('invert-au-abm-annual'));
```
yesil, line 686: can we check in tomorrow's tech call whether the FF `invert-au-abm-annual` is needed?
my understanding was that GWP was favorable to inverting the default logic in AU, and only let annual price to be rendered explicitly with `&annual=true`

bozojovicic: We can ask them tomorrow. I will demo this anyway. But if I invert this for the whole /au and then let them to add `&annual=true` where they need the annual price that will require lots of authoring. As far as I know they need this annual price next to the ABM almost everywhere. Australian legal reasons.

yesil: we should ask, but in parallel, can I suggest  FF name: `mas-enable-price-annual`  and it would be set only for /au tree.  The reason is that later we will have more FFs and we should follow a pattern for naming them.
We can discuss naming convetion, mine is just a suggestion.

yesil: we can set `mas-enable-price-annual` before PR merge so that it will keep working as before.
In parallel, we can ask GWP to adopt `&annual=true` notation going forward.

bozojovicic: As we agreed on yesterday's meeting we will not use any metadata flag. I updated the PR description and added the label "do not merge". Waiting for GWP now.

libs/features/mas/src/price.js
```diff
@@ -88,7 +90,7 @@ export function Price({ literals, providers, settings }) {
                 method = priceAnnual;
                 break;
             default:
-                if (options.country === 'AU' && offers[0].planType === 'ABM') {
+                if (options.country === 'AU' && offers[0].planType === 'ABM' && options.displayAnnual) {
```
npeltier, line 93: `options.country === 'AU'` test should just be deferred to `options.displayAnnual`, right?

npeltier (changes requested):

libs/features/mas/src/price.js
bozojovicic: @npeltier Yes. Right! This was originally requested for AU only but now when this feature is disabled by default, it can be used anywhere. I removed that "AU" condition.

npeltier (approved):

3ch023 (approved):

libs/deps/mas/merch-card-collection.js
yesil, line 1: why this file is modified?

yesil (approved):

Roycethan:
@bozojovicic When i compare against live vs your branch - i know by default the annual will be off , but why theres difference in strikethrough price display - it renders first in live and in your branch it renders second
https://main--cc--adobecom.aem.live/au/products/special-offers
vs
https://main--cc--adobecom.aem.live/au/products/special-offers?milolibs=mwpw166214au--milo--bozojovicic

image

bozojovicic:
@Roycethan conflicts resolved. Now regarding the order, I'm pretty sure this order here is correct when the annual price is displayed https://stage--cc--adobecom.aem.live/au/products/special-offers# that's what AU guys requested.
Now when annual price is disabled some new template is used "price alternative" where the ST price comes after ABM price. I will have to ask Nick about this.

Roycethan:
> MWPW-166214

Ok Let me know on this , Rest looks fine.. ill add the label once confirmed. Thanks

robert-bogos (approved):

Roycethan:
@bozojovicic  This is verified and regressed, whenever ready to be merged, plz resolve the conflicts and add the label thanks

PR #3875, opened 2025-03-28
[MWPW-169395] - Table aria labels missing

DKos95:
This resolves the issue where aria-labels were missing from Table CTA's, the main problem here was that Table.js and Merch.js are being executed in parallel and in the Table.js we were earlier using innerHtml to change the Table when the user resized or filtered the the tables in mobile mode, this caused the aria-labels that were being set in Merch.js to be set on the old table which was being replaced with a new table in Table.js, also columns were being removed from the DOM in curtain situations which was an additional problem causing the removal of aria-labels.

**Dev Notes**
Leaving this part in`.table .hide-mobile {
    display: none !important;
  }`. We would need to list at least 5 classes with high css specificity to override the display properties of some elements that need to be hidden, seems to be a safer option here and less lines.

**QA Notes**
It would be a good idea to test all the functionalities of the table since with this PR we are no longer using innerHtml which was at the core of the tables functionalities, especially filtering on mobile and the resizing of the screen.
[List of Pages where Tables are present](https://github.com/user-attachments/files/19504755/Some.pages.with.tables.md)

Resolves: [MWPW-169395](https://jira.corp.adobe.com/browse/MWPW-169395)

**Test URLs:**
- After: https://table-aria-labels--milo--adobecom.aem.page/docs/library/kitchen-sink/table?martech=off

**CC Merch Test URLs**
- Before: https://main--cc--adobecom.hlx.page/cc-shared/fragments/merch/products/photoshop/compare-plans/table/individual/default?martech=off
- After: https://main--cc--adobecom.hlx.page/cc-shared/fragments/merch/products/photoshop/compare-plans/table/individual/default?milolibs=table-aria-labels&martech=off

- Before: https://main--cc--adobecom.hlx.page/cc-shared/fragments/merch/products/photoshop/compare-plans/table/edu/default?martech=off
- After: https://main--cc--adobecom.hlx.page/cc-shared/fragments/merch/products/photoshop/compare-plans/table/edu/default?milolibs=table-aria-labels&martech=off

### DIFF
--- a/libs/blocks/table/table.js
+++ b/libs/blocks/table/table.js
@@ -9,6 +9,8 @@ const DESKTOP_SIZE = 900;
 const MOBILE_SIZE = 768;
 const tableHighlightLoadedEvent = new Event('milo:table:highlight:loaded');
 let tableIndex = 0;
+let isExpandEventsAssigned = false;
+
 const isMobileLandscape = () => (window.matchMedia('(orientation: landscape)').matches && window.innerHeight <= MOBILE_SIZE);
 function defineDeviceByScreenSize() {
   const screenWidth = window.innerWidth;
@@ -264,6 +266,22 @@ function handleExpand(e) {
   }
 }
 
+function setExpandEvents(el) {
+  if (isExpandEventsAssigned) return;
+
+  el.querySelectorAll('.icon.expand').forEach((icon) => {
+    isExpandEventsAssigned = true;
+    icon.parentElement.classList.add('point-cursor');
+    icon.parentElement.addEventListener('click', () => handleExpand(icon));
+    icon.parentElement.setAttribute('tabindex', 0);
+    icon.parentElement.addEventListener('keydown', (e) => {
+      if (e.key === ' ') e.preventDefault();
+
+      if (e.key === 'Enter' || e.key === ' ') handleExpand(icon);
+    });
+  });
+}
+
 function handleTitleText(cell) {
   if (cell.querySelector('.table-title-text')) return;
   const textSpan = createTag('span', { class: 'table-title-text' });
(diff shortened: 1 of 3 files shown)


libs/blocks/table/table.js
```diff
@@ -483,64 +501,65 @@ function applyStylesBasedOnScreenSize(table, originTable) {
...
         rows.forEach((row) => {
-          row.append(row.querySelector('.col:last-child').cloneNode(true));
+          const selectedColumn = row.querySelector(`.col-${selectedCol}`);
+          if (!selectedColumn) {
+            return;
+          }
+
+          selectedColumn.classList.remove('force-last');
+          selectedColumn.classList.remove('rounded-left', 'rounded-right');
+          const clone = selectedColumn.cloneNode(true);
```
overmyheadandbody, line 552: Could this cloning not lead back to the same issue with event removal?

overmyheadandbody (commented):
Looking good for the most part, I also checked this on https://table-aria-labels--milo--adobecom.hlx.page/drafts/sarchibeque/table/add-on?martech=off, where the logic for the special labels was originally added.

I just have a couple of notes for now.

libs/blocks/table/table.js
```diff
@@ -483,64 +501,66 @@ function applyStylesBasedOnScreenSize(table, originTable) {
...
+      });
+
+      if (filters[0] === filters[1]) {
+        const selectedCol = filters[0] + 1;
         rows.forEach((row) => {
-          row.append(row.querySelector('.col:last-child').cloneNode(true));
+          const selectedColumn = row.querySelector(`.col-${selectedCol}`);
+          if (!selectedColumn) {
+            return;
+          }
```
mokimo, line 548: We usually do one line returns 
```suggestion
          if (!selectedColumn) return;
```

libs/blocks/table/table.js
```diff
@@ -575,22 +595,44 @@ function applyStylesBasedOnScreenSize(table, originTable) {
...
     table.querySelector('.row-heading .col-1').style.display = 'flex';
   }
 
   if (deviceBySize === 'MOBILE' || (isMerch && deviceBySize === 'TABLET')) {
+    removeClones();
     mobileRenderer();
   } else {
-    table.innerHTML = originTable.innerHTML;
-    reAssignEvents(table);
+    removeClones();
```
mokimo, line 617: You could also put this on L612, then you don't need to redo the action if it doesn't need to be in the if/else clause but always happens

libs/blocks/table/table.js
```diff
@@ -575,22 +595,44 @@ function applyStylesBasedOnScreenSize(table, originTable) {
...
+    const element = table.querySelector('.row-heading');
+
+    if (element) {
+      const columns = [...element.children];
+      columns.forEach(({ children }) => {
+        [...children].forEach((row) => {
+          row.style.removeProperty('height');
+        });
+      });
+    }
```
mokimo, line 628: You should be able to avoid the if clause and shorten this a little
```suggestion
    [...(table.querySelector('.row-heading') || [])]
    .forEach({children}) => [...children].forEach(row => row.style.removeProperty('height'))
```

mokimo (approved):
Few code style suggestions, but looks good already. Feel free to take up the suggestions where you prefer.

overmyheadandbody (approved):

zagi25 (approved):

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-169395

PR #3881, opened 2025-03-31
Mwpw 170384 lost info 320

zagi25:
* Prevent video from overflowing inside modal
* Put chat button behind modal curtain when modal is open

Resolves: [MWPW-170384](https://jira.corp.adobe.com/browse/MWPW-170384)

Test overflow by zooming in the page or on very wide screen.
Chat button is only present on CC page. 

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/ratko/mwpw-170384-modal-overflow/modal-overflow?martech=off
- After: https://mwpw-170384-lost-info-320--milo--adobecom.aem.page/drafts/ratko/mwpw-170384-modal-overflow/modal-overflow?martech=off

**Example CC page from the ticket**
- Before: https://main--cc--adobecom.aem.live/products/firefly
- After: https://main--cc--adobecom.aem.live/products/firefly?milolibs=mwpw-170384-lost-info-320


libs/styles/iframe.css
```diff
@@ -26,6 +26,18 @@
...
 
+@media (min-width: 600px) {
+  .dialog-modal:has(.milo-video) {
+    width: 100%;
+  }
+
+  .dialog-modal .milo-video {
+    padding-bottom: 0;
+    min-width: unset;
+    height: min(calc(80vw * 0.5625), calc((100vh - 6px) - 2em));
```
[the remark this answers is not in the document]
zagi25: You're right, it was hard for me to figure everything out so I should've made it more clear, may bad. I added variables, now it should be more clearer what is what. 
About `--iframe-video-aspect-ratio: 0.5625;`, I defined this based on [iframe.css](https://github.com/adobecom/milo/blob/stage/libs/styles/iframe.css#L6), here `padding-bottom` is used to set height based on width.

libs/blocks/modal/modal.css
```diff
@@ -439,3 +441,7 @@
 .disable-scroll {
   overflow: hidden;
 }
+
+body:has(.dialog-modal) #adbMsgClientWrapper #adbmsgContainer #adbmsgCta {
+  z-index: 99;
```
mokimo, line 446: Why do we need z-index: 99? We need to be a little careful about all the z-indexes in milo, we don't have a good system to manage them.

zagi25: Because jarvis button has `z-index: 9999999;` , and it will appear on top of modal curtain and in some cases (zoom, smaller screens) it will overlap the modal content. Modal and modal curtain have `z-index` values of `102` and `101` ( I made a mistake, will bump it up to 100 if you don't have any further objections), modal div is removed when modal is closed so this should only be applied when modal is opened.
[modal.js](https://github.com/adobecom/milo/blob/stage/libs/blocks/modal/modal.js#L56)

mokimo: Can we move all the modal related z-indexes into variables? If not on a project level, at least on a block level we should be quickly able to see what should appear on top

libs/styles/iframe.css
```diff
@@ -1,3 +1,7 @@
+:root {
+  --iframe-video-aspect-ratio: 0.5625;
```
overmyheadandbody, line 2: I'm a bit reluctant to assume a 16:9 aspect ratio for all videos inside modals. We have 9:16 videos (which seem to be covered by the `tall-video` modal metadata style), square videos, etc. What would be the outcome for other aspect ratios? A modal larger than the video? Or would it have any other effect?

overmyheadandbody: I don't have a prod example, but I generated something quickly with Express, so you can use it on a page 😅 

https://github.com/user-attachments/assets/e8ecd2d5-8fe6-42cc-8ec8-a00bab1cede5

zagi25: For square videos it will be the same way as for wide videos on ultra wide screen, modal becomes bigger then the video and MPC fills it with background. 
Screenshot 2025-04-09 at 13 02 34

NadiiaSokolova (approved):
Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-170384

PR #3892, opened 2025-04-01
[MWPW-170623] - iframe video control fix

DKos95:
This resolves the issue where at around 1790px and onward screen width, the iframe controls start going down behind the scene where the user can't see or interact with them.

Resolves: [MWPW-170623](https://jira.corp.adobe.com/browse/MWPW-170623)

**Test URLs:**
- Before: https://iframe-video-controls--milo--adobecom.hlx.page/drafts/dusan/iframe-video-control-overflow
- After: https://main--milo--adobecom.hlx.page/drafts/dusan/iframe-video-control-overflow

### DIFF
--- a/libs/styles/iframe.css
+++ b/libs/styles/iframe.css
@@ -6,6 +6,13 @@
   padding-bottom: 56.25%;
 }
 
+@media (min-width: 1790px) {
+  .milo-video,
+  .milo-iframe {
+    padding-bottom: 47.25%;
+  }
+}
+
 .milo-video iframe,
 .milo-iframe iframe {
   border: 0;


mokimo (approved):
Is this supposed to be in draft?

libs/styles/iframe.css
```diff
@@ -6,6 +6,13 @@
   padding-bottom: 56.25%;
 }
 
+@media (min-width: 1790px) {
+  .milo-video,
+  .milo-iframe {
+    padding-bottom: 47.25%;
```
rgclayton, line 12: I almost wonder if adjusting the `min-width` on [line 27](https://github.com/adobecom/milo/blob/9414e399e52834bebcf7a20650f99b3e236f0be6/libs/styles/iframe.css#L28) would be more straight forward and predictable.

DKos95:
> Is this supposed to be in draft?

Yes, yesterday at the end of the day after I put up the PR, Ratko noticed that this story may be a duplicate of one that he is doing, it seems that his PR also fixes this so I put it in draft mode until we communicate this through Milo-Core, but in all likelihood we will need only his PR.

Ratko's PR - https://github.com/adobecom/milo/pull/3881
My Page with Ratko's solution - https://mwpw-170384-lost-info-320--milo--adobecom.hlx.page/drafts/dusan/iframe-video-control-overflow#test

PR #3896, opened 2025-04-03
MWPW-169814 | Change box shadow and button to close for split notification

aishwaryamathuria:
Change box shadow for new split notification variant created for https://jira.corp.adobe.com/browse/MWPW-169814
Clicking a button in notification authored with #_close will close the notification block.

Resolves: [MWPW-169814](https://jira.corp.adobe.com/browse/MWPW-169814) | [MWPW-170780](https://jira.corp.adobe.com/browse/MWPW-170780)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/mathuria/promobar/creativecloud?martech=off
- After: https://mwpw-170780--milo--adobecom.aem.page/drafts/mathuria/promobar/creativecloud?martech=off

### DIFF
--- a/libs/blocks/notification/notification.css
+++ b/libs/blocks/notification/notification.css
@@ -17,7 +17,7 @@
   --icon-size-xl: 64px;
   --pill-radius: 16px;
   --pill-shadow: 0 3px 6px #00000029;
-  --split-shadow: 0px 0px 16px 0px rgba(0, 0, 0, 0.16);
+  --split-shadow: 0px 6px 16px 0px rgba(0, 0, 0, 0.16);
 
   display: flex;
   inline-size: 100%;
--- a/libs/blocks/notification/notification.css
+++ b/libs/blocks/notification/notification.css
@@ -17,6 +17,7 @@
   --icon-size-xl: 64px;
   --pill-radius: 16px;
   --pill-shadow: 0 3px 6px #00000029;
+  --split-shadow: 0px 0px 16px 0px rgba(0, 0, 0, 0.16);
 
   display: flex;
   inline-size: 100%;
@@ -303,6 +304,10 @@
   inline-size: calc(100% - var(--spacing-m));
   margin-inline: auto;
   box-shadow: var(--pill-shadow);
+}
+
+.notification.split.pill {
+  box-shadow: var(--split-shadow);
 }
 
 .notification.pill .foreground .action-area {
--- a/libs/blocks/notification/notification.css
+++ b/libs/blocks/notification/notification.css
@@ -17,7 +17,7 @@
   --icon-size-xl: 64px;
   --pill-radius: 16px;
   --pill-shadow: 0 3px 6px #00000029;
-  --split-shadow: 0px 6px 16px 0px rgba(0, 0, 0, 0.16);
+  --split-shadow: 0 6px 16px 0 rgb(0 0 0 / 0.16);
 
   display: flex;
   inline-size: 100%;
--- a/libs/blocks/notification/notification.js
+++ b/libs/blocks/notification/notification.js
@@ -137,7 +137,7 @@ async function decorateLockup(lockupArea, el) {
 }
 
 function decorateSplitList(el, listContent) {
-  const closeEvent = "#_evt-close";
+  const closeEvent = '#_evt-close';
   const listContainer = createTag('div', { class: 'split-list-area' });
   listContent?.querySelectorAll('li').forEach((item) => {
     const listItem = createTag('div', { class: 'split-list-item' });
--- a/libs/blocks/notification/notification.js
+++ b/libs/blocks/notification/notification.js
@@ -86,14 +86,18 @@ function wrapCopy(foreground) {
   });
 }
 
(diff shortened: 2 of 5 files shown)


libs/blocks/notification/notification.js
```diff
@@ -142,6 +146,11 @@ function decorateSplitList(listContent) {
       ? item
       : item.nextElementSibling;
     const btn = createTag('div', {}, textli.lastElementChild);
+    const btnA = btn.querySelector('a');
+    if (btnA.href.includes('#_evt-close')) {
```
narcis-radu, line 150: We can use a const for the event name

narcis-radu, line 150: Need to make sure the button exists, `btnA?.href.includes` should do the trick

narcis-radu: Would be great to add some unit testing

aishwaryamathuria: @narcis-radu I have made the required changes and added the unit test for close functionality.

aishwaryamathuria: @narcis-radu  I have moved the event name to a const variable.

libs/blocks/notification/notification.css
```diff
@@ -17,6 +17,7 @@
   --icon-size-xl: 64px;
   --pill-radius: 16px;
   --pill-shadow: 0 3px 6px #00000029;
+  --split-shadow: 0px 6px 16px 0px rgba(0, 0, 0, 0.16);
```
overmyheadandbody, line 20: No need for pixel values for 0; you can use more modern color notation:
```suggestion
  --split-shadow: 0 6px 16px 0 rgb(0 0 0 / 0.16);
```

I'd also suggest we make things consistent, I see L19 has a hex declaration, `#00000029` is equivalent to `rgba(0, 0, 0, 0.16)`, we should stick to a single pattern.

test/blocks/notification/notification.test.js
```diff
@@ -102,5 +102,10 @@ describe('notification', async () => {
       const stacks = splits[0].querySelectorAll('.split-list-item').length;
       expect(stacks).to.equal(2);
     });
+    it('closes the notiification', () => {
```
overmyheadandbody, line 105:
```suggestion
    it('closes the notification', () => {
```

test/blocks/notification/notification.test.js
```diff
@@ -102,5 +102,10 @@ describe('notification', async () => {
...
+      const event = new CustomEvent('click', { bubbles: true });
```
overmyheadandbody, line 106: `click` should be a regular event, I see there's already logic to click in this file - https://github.com/adobecom/milo/blob/d8ff7a402d01d7b4d32ccdac420ade2a15d41b3c/test/blocks/notification/notification.test.js#L71

libs/blocks/notification/notification.js
```diff
@@ -142,6 +147,11 @@ function decorateSplitList(listContent) {
       ? item
       : item.nextElementSibling;
     const btn = createTag('div', {}, textli.lastElementChild);
+    const btnA = btn.querySelector('a');
+    if (btnA?.href.includes(closeEvent)) {
+      btnA.href = closeEvent;
```
overmyheadandbody, line 152: Since `preventDefault` will be called, why is this step needed?

overmyheadandbody (commented):
Unrelated to these changes, the sticky pill is quite jittery when the page is being scrolled. I tried a few quick ideas to solve it, but didn't find the culprit. Maybe this can be looked into as well, albeit maybe as part of another story.

salonijain3 (approved):

libs/blocks/notification/notification.js
aishwaryamathuria: @overmyheadandbody The step is only to clean the url and improve readability as a random or constant url will be used by GWP for this use case and this will remove the same and replace with hash only.

libs/blocks/notification/notification.css
aishwaryamathuria: @overmyheadandbody I have made the suggested change

test/blocks/notification/notification.test.js
aishwaryamathuria: @overmyheadandbody I have made the suggested change

test/blocks/notification/notification.test.js
aishwaryamathuria: @overmyheadandbody I have made the suggested change

spadmasa:
validation done on pr on below urls
https://mwpw-170780--milo--adobecom.aem.page/drafts/mathuria/promobar/creativecloud?martech=off
https://main--cc--adobecom.hlx.page/mx/drafts/souj/mweb/splitvariant11?milolibs=mwpw-170780

https://main--cc--adobecom.hlx.page/cc-shared/fragments/merch/products/firefly/sticky-banner/create-with-firefly-phone?milolibs=mwpw-170780
image

https://github.com/user-attachments/assets/0decafb8-1067-45d2-9249-1af00a5188f1

image

PR #3897, opened 2025-04-03
MWPW-170748 [Sev2][Catalog] Controls have same name but different …

bozojovicic:
… actions - action menu

Fix for A11Y issue on catalog page for action menu.

Resolves: [MWPW-170748](https://jira.corp.adobe.com/browse/MWPW-170748)

**Test URLs:**
- After: https://mwpw170748--milo--bozojovicic.aem.live/?martech=off


libs/features/mas/src/variants/catalog.js
```diff
@@ -47,7 +47,7 @@ export class Catalog extends VariantLayout {
                         tabindex="0"
                         aria-expanded="false"
                         role="button"
-                    >Action Menu</div>
+                    >Device availability and creative categories - ${this.card.querySelector('.card-heading')?.textContent}</div>
```
yesil, line 50: Instead, I recommend to add `action-menu-label` attribute to `merch-card` web component, which should be set by `merch-card` block to a Milo placeholder, normally it should be replaced by Milo properly.
suggestion for the name: `catalog-card-action-title` with `Device availability and creative categories` by default.

and use `this.card.title` to read the title which should save you the querySelector.

We will have to do the same for cards coming from MAS Studio but it can be addressed in a different ticket.

bozojovicic: In MWPW-170748 Fred asked to rename the PH key to `{{action-menu}}`. He already renamed it in sharepoint.

3ch023 (approved):

yesil (approved):

PR #3898, opened 2025-04-03
Mwpw 169999 header obscures content

zagi25:
Issue: When navigating a page by using `Tab` or `Shift + Tab` , focused element that is hidden by gnav, sticky element, cookie banner etc. remains hidden.
Screenshot 2025-04-03 at 15 29 25

* Element gets scrolled to the middle of the screen when focused by `Tab` or `Shift + Tab`
* Cookie banner get's position sticky, when there is visible focus on the page and focus is not inside cookie banner, so that when page is scrolled to the end using `Tab` all elements in the footer can be focused and are not covered by the cookie banner ( This was done with the assumption that if the cookie banner is present it is always the last element in the body )

I tried setting `scroll-padding-top` and `scroll-padding-bottom` to the exact height of the elements that are hiding other content, but this requires JS and it will bring unnecessary complexity trying to cover all the use cases.

Resolves: [MWPW-169999](https://jira.corp.adobe.com/browse/MWPW-169999) , [MWPW-170165](https://jira.corp.adobe.com/browse/MWPW-170165)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/ratko/mwpw-169999-header-obscures-content/dev?martech=off
- After: https://mwpw-169999-header-obscures-content--milo--adobecom.aem.page/drafts/ratko/mwpw-169999-header-obscures-content/dev?martech=off

**Homepage URLs:**
- Before: https://main--homepage--adobecom.aem.page/homepage/index-loggedout
- After: https://main--homepage--adobecom.aem.page/homepage/index-loggedout?milolibs=MWPW-169999-header-obscures-content

**CC URLs:**
- Before: https://main--cc--adobecom.aem.page/creativecloud
- After: https://main--cc--adobecom.aem.page/creativecloud?milolibs=MWPW-169999-header-obscures-content

**QE NOTE:** Please close the [MWPW-170165](https://jira.corp.adobe.com/browse/MWPW-170165) with the reason `fixed by`  [MWPW-169999](https://jira.corp.adobe.com/browse/MWPW-169999) if this also resolves the issue described in that ticket.


libs/styles/styles.css
```diff
@@ -682,6 +682,22 @@ div[class*='-up'] .con-block.has-bg.no-spacing { padding: 0; }
...
+html:focus-within {
+  scroll-padding-top: 50vh;
+  scroll-padding-bottom: 50vh;
+}
+
+html:focus-within #onetrust-consent-sdk:not(:focus-within) {
+  position: sticky;
+  bottom: 0;
+  /* #onetrust-banner-sdk z-index value */
+  z-index: 2147483645;
```
mokimo, line 694: That's what, 2 below the javascript max 32 bit int value? 😆

libs/styles/styles.css
```diff
@@ -682,6 +682,22 @@ div[class*='-up'] .con-block.has-bg.no-spacing { padding: 0; }
     local('TrebuchetMS');
 }
 
+html:focus-within {
+  scroll-padding-top: 50vh;
+  scroll-padding-bottom: 50vh;
+}
```
mokimo, line 688: That feels pretty intrusive, I don't think we should overwrite how the browser shifts focus. The page jumps around a lot and the adobe page would not behave "as expected" when tabbing through other pages on the web (non adobe)
https://github.com/user-attachments/assets/7734d0e7-0097-4f4b-8fd1-60c86af34014

zagi25: I understand what you are saying. I tried using the JS in combination with CSS variables, to set a CSS variable to the height of overlapping element (cookie banner, sticky element etc.) so that page will scroll only if focused element is behind something . Using JS requires to cover all cases separately because elements don't have fixed height and I tried to make it simpler using CSS only approach. I can give JS a try once again today to cover cookie banner and sticky cases and we will see if there is some other use case that needs to be covered. If you have some better suggestion I am all ears.

libs/styles/styles.css
zagi25: To be honest, I just copied the value to make things consistent didn't thought about it too much

mokimo, line 694: Jokes aside, why exactly do we need this? Its not directly clear to me

zagi25:
@mokimo I pushed solution using JS. 

- Value of scroll padding is taken from sticky sections / cookie banner height
- Values are updated on resize, sticky section closing and hiding of the cookie banner 

I left some comments for easier review, will remove them later. I added the code to accessibility file, if you think it should be added somewhere else let me know (same goes for calling of the function)

libs/styles/styles.css
zagi25: Screenshot 2025-04-07 at 11 17 34
I added this so that when page is navigated using tab, and cookie banner is not closed, cookie banner below the footer so user can see all focused footer links

libs/scripts/accessibility.js
```diff
@@ -0,0 +1,125 @@
+const CONSENT_EVENTS = ['adobePrivacy:PrivacyConsent', 'adobePrivacy:PrivacyCustom', 'adobePrivacy:PrivacyReject'];
```
overmyheadandbody, line 1: I agree with @mokimo that the initial solution strays away from the intended browser behavior, but I also agree with @zagi25's initial statement that doing it via JS would require a lot of logic, as it is made clear in this file. How about we adapt the behavior just for the edge cases we're seeing. Here's my idea:

```
function adjustElementPosition(element) {
  const rect = element.getBoundingClientRect();
  const viewportHeight = window.innerHeight;
  const oneQuarterViewport = viewportHeight / 4;

  if (rect.top <= oneQuarterViewport) {
    element.classList.add('top-quarter-focus');
    element.scrollIntoView({ behavior: 'smooth', block: 'end' });
  }

  element.addEventListener('blur', () => {
    element.classList.remove('top-quarter-focus');
  }, { capture: true, once: true });
}

document.addEventListener('focusin', (e) => {
  adjustElementPosition(e.target);
});
```

And the CSS would simply be

```
.top-quarter-focus {
  scroll-margin-top: 20vh;
}
```

This would, of course, need to be adapted for the bottom part of the page as well.

I think having complex logic that tries to consider all possible positions and layout rules would be unmaintainable in the long run. My suggestion is more of a 80/20 approach, I'd like to see if we can solve most cases with simple logic and see if and where we need to fill the gap afterwards.

overmyheadandbody (approved):
I quite like this approach and it has worked well during my testing. Just a few small nits, but looking great otherwise and much more straightforward than the previous implementation 👌

mokimo (approved):
The experience is a lot better when tabbing, glad this went through a few iterations and we really managed to improve it!

libs/styles/styles.css
```diff
@@ -682,6 +682,21 @@ div[class*='-up'] .con-block.has-bg.no-spacing { padding: 0; }
...
 
+html:focus-within {
+  scroll-behavior: smooth;
+}
+
+body:has(footer:focus-within) #onetrust-consent-sdk:not(:has(#ot-cookie-button.show)) {
+  position: sticky;
+  bottom: 0;
+  /* #onetrust-banner-sdk z-index value */
+  z-index: 2147483645;
```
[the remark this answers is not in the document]
zagi25: Don't know if this will ever happen, but without z-index if user scrolls while focus is in `footer` 
Screenshot 2025-04-09 at 16 01 34

DKos95 (approved):

hit11757:
@narcis-radu I found a bug mentioned in the ticket https://jira.corp.adobe.com/browse/MWPW-169999 Just waiting for @zagi25 to make the changes, and then I’ll re-test

zagi25:
@overmyheadandbody @mokimo 
Attempt number 125 :D 
Because iframe steals the focus, I decided to add a fallback with `scroll-padding`.
Idea is that when user tabs and `focusin` event is not fired, `scroll-padding` is set based on height of know overlapping elements (gnav, sticky-top, sticky-bottom, cookie banner) and browser will handle tabbing while the focus is inside an iframe. As soon as we regain control of focus, old solution will be used. 

cc: @hit11757

libs/utils/utils.js
```diff
@@ -1512,6 +1512,9 @@ function decorateDocumentExtras() {
 
 async function documentPostSectionLoading(config) {
   decorateFooterPromo();
+  import('../scripts/accessibility.js').then((accessibility) => {
```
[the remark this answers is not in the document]
zagi25: Yes, that is the idea that all functionality that is needed to support accessibility lives in this file (small problem is that we don't know if there will be additional accessibility related functionality). If you think there is a better place for this let me know, I was thinking putting this into `delayed.js` before I created `accessibility.js`

zagi25:
@overmyheadandbody , yes having `scroll-padding-top/bottom : 25vh` will cover most of the cases, but we still need to set it only if there is a visible focus e.g. `html:focus-within` because it can mess up an `anchor` feature. So there are two options:

1. Leave the functionality that checks if some element covers focused element, and use `scroll-padding-top/bottom: 25vh` only when focus escapes to `iframe`.
2. Only declare `html:focus-with { scroll-padding-top/bottom: 25vh }`, this is less complex then the first option, but if there are no sticky elements, it will scroll the page even if the focused element is visible. 

I am more leaning to the first option because it is precise when we have a control over the focus event but also it has fallback for when the focus goes into the `iframe`.

DKos95 (approved):

hit11757 (approved):
Testing details can be found in the ticket https://jira.corp.adobe.com/browse/MWPW-169999

PR #3899, opened 2025-04-03
MWPW-170799: Remove external.js and refactor imports.

yesil:
This PR addresses a regression that inflates `merch-card.js` size.

* removed abandoned TWP variant and related code / test files.
* made `mas-commerce-service` sync from async by embedding price literals in mas.js as a variable, and avoid using `await` 
* Refactored variant styles, so that they can be registered per bundle (acom bundle will not contain ccd/ah variants)

before (78kb):
image

after (60.7kb):
image

Resolves: [MWPW-170799](https://jira.corp.adobe.com/browse/MWPW-170799)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.live/drafts/ilyas/merch-card?martech=off
- After: https://mwpw-170799--milo--adobecom.aem.live/drafts/ilyas/merch-card?martech=off


libs/features/mas/src/aem-fragment.js
```diff
@@ -147,22 +102,65 @@ export class AemFragment extends HTMLElement {
...
+        }
     }
 
-    async activate(masCommerceService, resolve, reject) {
-        this.#service = masCommerceService;
-        this.refresh(false)
-            .then((result) => resolve(result))
-            .catch((e)=> reject(e));
+    async activate() {
+        const flushCache = !this.#author;
```
3ch023, line 162: i remember there was the reason why we changed it just a week ago in Angelo's PR
https://github.com/adobecom/milo/pull/3786/files#diff-36078108e571d0861359f4696fe62484ed743f80e52d1e082c7c37b858b08955R163

did you had to bring it back because something wasn't working?

libs/features/mas/src/checkout-mixin.js
```diff
@@ -1,18 +1,18 @@
-import { ignore } from './external.js';
+import { ignore } from '@dexter/tacocat-core';
```
3ch023, line 1: I suggest to remove this import..
ignore function looks like this:
```
const ignore = () => {
    /* do nothing */
};
```

imo no point to import it from other package, maybe we could remove dependency on tacocat-core at some moment

3ch023: or if you see value in ignore function i'd redefine it in utilities.. but its so simple it could be just defined in this file on the fly as well

libs/features/mas/src/checkout.js
```diff
@@ -15,16 +10,23 @@ import { toOfferSelectorIds, toQuantity } from './utilities.js';
  * generate Checkout configuration
  */
 export function Checkout({ providers, settings }) {
+    // Memoize settings for frequently used values
```
3ch023, line 13: typo? Memoize

libs/features/mas/src/checkout.js
```diff
@@ -15,16 +10,23 @@ import { toOfferSelectorIds, toQuantity } from './utilities.js';
...
+    const {
+        checkoutClientId,
+        checkoutWorkflow: defaultWorkflow,
+        checkoutWorkflowStep: defaultWorkflowStep,
+        country: defaultCountry,
+        language: defaultLanguage,
+        promotionCode: defaultPromotionCode,
+        quantity: defaultQuantity,
+        env,
```
3ch023, line 22: we didn't have env and landscape before, how are the used in checkout? just wondering what changed

libs/features/mas/src/literals.js
```diff
@@ -1,5 +1,7 @@
 import { Defaults } from './defaults.js';
-import { equalsCaseInsensitive } from './external.js';
+import { equalsCaseInsensitive } from '@dexter/tacocat-core';
```
3ch023, line 2: i live it to your preferrence but imo we should remove whole @dexter/tacocat-core package.
you could copy over required methods (i am not sure if all of them are indeed need to be 'reused') to utilities and remove unneccessary package dependency. we only use tacocat-core for basic js functions, but it brings useless code as well

libs/features/mas/src/mas-commerce-service.js
```diff
@@ -93,12 +90,7 @@ export class MasCommerceService extends HTMLElement {
 
         // Fetch price literals
         const literals = { price: {} };
-        try {
-            literals.price = await getPriceLiterals(
-                settings,
-                config.commerce.priceLiterals,
-            );
-        } catch (e) {}
+        literals.price = getPriceLiterals(settings);
```
3ch023, line 93: cool cool catch

libs/features/mas/src/aem-fragment.js
yesil: might be due to merging, I'll double check thanks.

libs/features/mas/src/checkout-mixin.js
yesil: I'll remove it

libs/features/mas/src/mas-element.js
```diff
@@ -26,6 +25,7 @@ const StateEventType = {
 };
 
 export class MasElement {
+    #service;
     changes = new Map();
     connected = false;
     dispose = ignore;
```
3ch023, line 31: `dispose = ignore;` can be removed?

libs/features/mas/src/checkout.js
3ch023: haha i go google whats that :D

3ch023 (approved):

bozojovicic (approved):

afmicka:
@yesil we already talked offline but for visibility could you please confirm that we are not going to support merch-block for plans card? On your branch the card is missing
https://mwpw-170799--milo--adobecom.aem.live/drafts/ilyas/merch-card?martech=off
Screenshot 2025-04-14 at 17 15 33

afmicka:
there might have been a glitch for some time, i am seeing the "missing" card on your branch too now. Will check once more in the morning

PR #3901, opened 2025-04-04
Adding dynamic PR template for Gnav components

bandana147:
* Including Gnav and footer's variants in the test Urls for all clients if the PR touches any files under these 2 folders:
```
libs/blocks/global-navigation/**
libs/blocks/global-footer/**
```

Resolves: [MWPW-170257](https://jira.corp.adobe.com/browse/MWPW-170257)

**Test URLs:**
- After: https://gnav-dynamic-pr--milo--adobecom.aem.page/?martech=off

PR generated using this template - https://github.com/adobecom/milo/pull/3903


mokimo:
@overmyheadandbody @narcis-radu @robert-bogos wdyt, should we have this for *every* PR with more general pages?
See the template in action https://github.com/adobecom/milo/pull/3903

sharmrj (approved):

.github/workflows/dynamic-pr-template.yml
```diff
@@ -0,0 +1,145 @@
...
+jobs:
+  update-template:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v39
```
mokimo, line 21: tj-actions/eslint was recently affected by a supply chain attack - would be great to pin this version similar to checkout. See https://github.com/adobecom/milo/blob/stage/.github/workflows/README.md

bandana147: Thanks. I will check this and update the tag.

sonawanesnehal3 (approved):

overmyheadandbody:
Yes, @mokimo, I love it! We had the approach of using the query param, but this is much better, also considering that a lot of contributors don't want to invest too much time in writing the PR description. We shouldn't have a 1:1 with the Gnav template, though. I'll open a story for us to look into

spadmasa:
Validated sample urls provided
https://jira.corp.adobe.com/browse/MWPW-170257

PR #3909, opened 2025-04-06
Vhargrave/mwpw 169933 language based links

vhargrave:
Allow for language based links in milo

Resolves: [MWPW-169933](https://jira.corp.adobe.com/browse/MWPW-169933)

**Test URLs:**
- After: https://vhargrave-MWPW-169933-language-based-links--milo--adobecom.aem.page/?martech=off

- Before: http://main--milo--adobecom.aem.page/de/de/
- After: http://vhargrave-mwpw-169933-language-based-links--milo--adobecom.aem.page/de/de/

**News - will be the first consumer using this new paradigm**
- Before: https://language-site--news--adobecom.aem.page/ko/drafts/ruchika/testpage
- After: https://language-site--news--adobecom.aem.page/ko/drafts/ruchika/testpage?milolibs=vhargrave-MWPW-169933-language-based-links

- Before: https://language-site--news--adobecom.aem.page/en/gb/
- After: https://language-site--news--adobecom.aem.page/en/gb/?milolibs=vhargrave-MWPW-169933-language-based-links

** CC & other consumers should not be affected by this change **
- Before: http://main--cc--adobecom.aem.live/products/photoshop-lightroom
- After: http://main--cc--adobecom.aem.live/products/photoshop-lightroom?milolibs=vhargrave-MWPW-169933-language-based-links

- Before: http://main--bacom--adobecom.hlx.page/
- After: http://main--bacom--adobecom.hlx.page/?milolibs=vhargrave-MWPW-169933-language-based-links

- Before: http://main--bacom--adobecom.hlx.page/de/
- After: http://main--bacom--adobecom.hlx.page/de/?milolibs=vhargrave-MWPW-169933-language-based-links

### DIFF
--- a/libs/utils/languages.js
+++ b/libs/utils/languages.js
@@ -1,0 +1,22 @@
+export default {
+  en: {
+    ietf: 'en',
+    tk: 'hah7vzn.css',
+    regions: [
+      { region: 'us', ietf: 'en-US', tk: 'hah7vzn.css' },
+      { region: 'gb', ietf: 'en-GB', tk: 'hah7vzn.css' },
+      { region: 'apac', ietf: 'en', tk: 'hah7vzn.css' },
+    ],
+  },
+  de: {
+    ietf: 'de',
+    tk: 'hah7vzn.css',
+    languageBased: false,
+    regions: [
+      // { region: 'at', ietf: 'de-AT', tk: 'hah7vzn.css' },
+      { region: 'ch', ietf: 'de-CH', tk: 'hah7vzn.css' },
+      { region: 'de', ietf: 'de-DE', tk: 'hah7vzn.css' },
+      // { region: 'lu', ietf: 'de-LU', tk: 'hah7vzn.css' }
+    ],
+  },
+};
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -157,6 +157,9 @@ export const SLD = PAGE_URL.hostname.includes('.aem.') ? 'aem' : 'hlx';
 const PROMO_PARAM = 'promo';
 let isMartechLoaded = false;
 
+let localeToLanguageMap;
+let siteLanguages;
+
 export function getEnv(conf) {
   const { host } = window.location;
   const query = PAGE_URL.searchParams.get('env');
@@ -198,6 +201,32 @@ export function getLocale(locales, pathname = window.location.pathname) {
   return locale;
 }
 
+export function getLanguage(languages, locales, pathname = window.location.pathname) {
+  const defaultLang = { ietf: 'en', tk: 'hah7vzn.css', prefix: '/en' };
+  if (!languages) {
+    return defaultLang;
+  }
+  const split = pathname.split('/');
+  const baseSplit = [LANGSTORE, PREVIEW].includes(split[1]) ? 1 : 0;
+  const languageString = split[baseSplit + 1];
+  const region = split[baseSplit + 2];
+
+  const language = languages[languageString];
+  if (language && region && language.regions) {
+    const [matchingRegion] = language.regions.filter((r) => r.region === region);
+    language.region = matchingRegion;
+  }
+
+  // if no language, allow for support of locale based routing still
+  if (!language || (language.languageBased === false && !language.region)) {
+    return getLocale(locales, pathname);
+  }
+
+  language.prefix = `/${languageString}${region ? `/${region}` : ''}`;
+  language.languageBased = true;
+  return language;
+}
+
 export function getMetadata(name, doc = document) {
   const attr = name && name.includes(':') ? 'property' : 'name';
   const meta = doc.head.querySelector(`meta[${attr}="${name}"]`);
@@ -236,7 +265,13 @@ export const [setConfig, updateConfig, getConfig] = (() => {
       config = { env: getEnv(conf), ...conf };
       config.codeRoot = conf.codeRoot ? `${origin}${conf.codeRoot}` : origin;
       config.base = config.miloLibs || config.codeRoot;
-      config.locale = pathname ? getLocale(conf.locales, pathname) : getLocale(conf.locales);
+      if (conf.languages) {
+        config.locale = pathname
+          ? getLanguage(conf.languages, conf.locales, pathname)
+          : getLanguage(conf.languages, conf.locales);
+      } else {
+        config.locale = pathname ? getLocale(conf.locales, pathname) : getLocale(conf.locales);
+      }
       config.autoBlocks = conf.autoBlocks ? [...AUTO_BLOCKS, ...conf.autoBlocks] : AUTO_BLOCKS;
       config.signInContext = conf.signInContext || {};
       config.doNotInline = conf.doNotInline
@@ -306,6 +341,39 @@ export const getFederatedUrl = (url = '') => {
   return url;
 };
 
+function hasLanguageLinks(area) {
+  const targetDomains = [
+    // don't add milo too. It's a special case because of tools, merch, etc.
+    'news.adobe.com',
+    '--news--adobecom.',
+  ];
+
+  return Array.from(area.querySelectorAll('a[href]')).some((link) => {
+    try {
+      const url = new URL(link.href);
+      return targetDomains.some((domain) => url.hostname.includes(domain));
+    } catch (e) {
+      return false;
+    }
+  });
+}
+
+async function loadLanguageConfig() {
+  try {
+    const config = await fetch(`${getFederatedContentRoot()}/federal/assets/data/languages-config.json`);
+    const configJson = await config.json();
+
+    localeToLanguageMap = configJson['locale-to-language-map'].data;
+    siteLanguages = configJson['site-languages'].data.map((site) => {
+      site.domainMatches = site.domainMatches.split(/[\n,]+/).map((t) => t.trim()).filter(Boolean);
+      site.languages = site.languages.split(/[\n,]+/).map((t) => t.trim()).filter(Boolean);
+      return site;
+    });
+  } catch (e) {
+    window.lana?.log('Failed to load language-config.json:', e);
+  }
+}
+
 let fedsPlaceholderConfig;
 export const getFedsPlaceholderConfig = ({ useCache = true } = {}) => {
   if (useCache && fedsPlaceholderConfig) return fedsPlaceholderConfig;
@@ -370,17 +438,40 @@ export function localizeLink(
     const extension = getExtension(path);
     const allowedExts = ['', 'html', 'json'];
     if (!allowedExts.includes(extension)) return processedHref;
-    const { locale, locales, prodDomains } = getConfig();
-    if (!locale || !locales) return processedHref;
+    const { locale, locales, languages, prodDomains } = getConfig();
+    if (!locale || !(locales || languages)) return processedHref;
     const isLocalizable = relative || (prodDomains && prodDomains.includes(url.hostname))
       || overrideDomain;
     if (!isLocalizable) return processedHref;
     const isLocalizedLink = path.startsWith(`/${LANGSTORE}`)
       || path.startsWith(`/${PREVIEW}`)
+      || Object.keys(languages).some((lang) => lang !== '' && (path.startsWith(`/${lang}/`)))
       || Object.keys(locales).some((loc) => loc !== '' && (path.startsWith(`/${loc}/`)
         || path.endsWith(`/${loc}`)));
     if (isLocalizedLink) return processedHref;
-    const urlPath = `${locale.prefix}${path}${url.search}${hash}`;
+
+    let { prefix } = locale;
+    if (locale.languageBased && !url.hostname.includes('localhost')) {
+      const site = siteLanguages?.find(
+        (s) => s.domainMatches.some((d) => url.hostname.includes(d)),
+      );
+      if (!site) {
+        const mappedLocale = localeToLanguageMap.find((m) => `/${m.languagePath}` === locale.prefix);
+        if (mappedLocale) prefix = `/${mappedLocale.locale}`;
+      }
+    } else if (siteLanguages) {
+      const site = siteLanguages?.find(
+        (s) => s.domainMatches.some((d) => url.hostname.includes(d)),
+      );
+      if (site) {
+        const language = site.languages.find((l) => `/${l}` === locale.prefix);
+        if (!language) {
+          const mappedLocale = localeToLanguageMap.find((m) => `/${m.languagePath}` === locale.prefix);
+          if (mappedLocale) prefix = `/${mappedLocale.locale}`;
+        }
+      }
+    }
+    const urlPath = `${prefix}${path}${url.search}${hash}`;
     return relative ? urlPath : `${url.origin}${urlPath}`;
   } catch (error) {
     return href;
@@ -1480,6 +1571,9 @@ export async function loadArea(area = document) {
     appendSuffixToTitles();
   }
   const config = getConfig();
+  if ((!localeToLanguageMap && !siteLanguages) && (hasLanguageLinks(area) || config.languages)) {
+    await loadLanguageConfig();
+  }
 
   if (isDoc) {
     decorateDocumentExtras();
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -210,11 +210,13 @@ export function getLanguage(languages, locales, pathname = window.location.pathn
   const baseSplit = [LANGSTORE, PREVIEW].includes(split[1]) ? 1 : 0;
   const languageString = split[baseSplit + 1];
   const region = split[baseSplit + 2];
+  let regionPath = '';
 
   const language = languages[languageString];
   if (language && region && language.regions) {
     const [matchingRegion] = language.regions.filter((r) => r.region === region);
     language.region = matchingRegion;
+    regionPath = matchingRegion ? `/${region}` : '';
   }
 
   // if no language, allow for support of locale based routing still
@@ -222,7 +224,7 @@ export function getLanguage(languages, locales, pathname = window.location.pathn
     return getLocale(locales, pathname);
   }
 
-  language.prefix = `/${languageString}${region ? `/${region}` : ''}`;
+  language.prefix = `/${languageString}${regionPath}`;
   language.languageBased = true;
   return language;
 }
@@ -341,7 +343,7 @@ export const getFederatedUrl = (url = '') => {
   return url;
 };
 
-function hasLanguageLinks(area) {
+export function hasLanguageLinks(area) {
   const targetDomains = [
     // don't add milo too. It's a special case because of tools, merch, etc.
     'news.adobe.com',
@@ -358,7 +360,7 @@ function hasLanguageLinks(area) {
   });
 }
 
-async function loadLanguageConfig() {
+export async function loadLanguageConfig() {
   try {
     const config = await fetch(`${getFederatedContentRoot()}/federal/assets/data/languages-config.json`);
     const configJson = await config.json();
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -152,6 +152,11 @@ const TARGET_TIMEOUT_MS = 4000;
 const LANGSTORE = 'langstore';
 const PREVIEW = 'target-preview';
 const PAGE_URL = new URL(window.location.href);
+const LANGUAGE_BASED_DOMAINS = [
+  // don't add milo too. It's a special case because of tools, merch, etc.
+  'news.adobe.com',
+  '--news--adobecom.',
+];
 export const SLD = PAGE_URL.hostname.includes('.aem.') ? 'aem' : 'hlx';
 
 const PROMO_PARAM = 'promo';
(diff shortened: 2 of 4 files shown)


libs/utils/languages.js
```diff
@@ -0,0 +1,22 @@
+export default {
```
mokimo, line 1: Will all of adobe.com share the same languageBased-map? If yes this belongs into utils.
If consumers will always set this themselves - move this into our scripts.

Adding extra requests for something that is required for LCP is a performance/code-maintenance tradeoff where performance wins.

libs/utils/utils.js
```diff
@@ -152,11 +152,19 @@ const TARGET_TIMEOUT_MS = 4000;
 const LANGSTORE = 'langstore';
 const PREVIEW = 'target-preview';
 const PAGE_URL = new URL(window.location.href);
+const LANGUAGE_BASED_DOMAINS = [
+  // don't add milo too. It's a special case because of tools, merch, etc.
+  'news.adobe.com',
+  '--news--adobecom.',
+];
```
mokimo, line 159: Can we rely on a flag set by consumers? Bleeding consumer implementation details back into milo is an anti-pattern. (Which I'm guilty of myself)

If consumers set this in their scripts, they can onboard to language based by adding a feature-config in their project without any PR to milo.

libs/utils/utils.js
```diff
@@ -198,6 +206,34 @@ export function getLocale(locales, pathname = window.location.pathname) {
   return locale;
 }
 
+export function getLanguage(languages, locales, pathname = window.location.pathname) {
+  const defaultLang = { ietf: 'en', tk: 'hah7vzn.css', prefix: '/en' };
+  if (!languages) {
+    return defaultLang;
+  }
+  const split = pathname.split('/');
+  const baseSplit = [LANGSTORE, PREVIEW].includes(split[1]) ? 1 : 0;
```
mokimo, line 215: What's split, could we be more specific? `urlPath`? Similar, baseSplit might be more `locOffset` ?

libs/utils/utils.js
```diff
@@ -198,6 +206,34 @@ export function getLocale(locales, pathname = window.location.pathname) {
...
+  const languageString = split[baseSplit + 1];
+  const region = split[baseSplit + 2];
```
mokimo, line 217: If this was called `const region = urlPath[locOffset +2]`  that'd be a lot more clear what it means. Reading `split[baseSplit+2]` is not intuitive at all.

As a good mind-model & pattern, reading any single line of code in isolation should still provide you with most of the context that you need to understand what it does.

libs/utils/utils.js
```diff
@@ -198,6 +206,34 @@ export function getLocale(locales, pathname = window.location.pathname) {
   return locale;
 }
 
+export function getLanguage(languages, locales, pathname = window.location.pathname) {
```
mokimo, line 209: Should we use `getConfig()` instead of passing parameters?

libs/utils/utils.js
```diff
@@ -306,6 +348,38 @@ export const getFederatedUrl = (url = '') => {
...
+      // do nothing
+    }
+  }
+  return false;
+}
+
+export async function loadLanguageConfig() {
+  const parseList = (str) => str.split(/[\n,]+/).map((t) => t.trim()).filter(Boolean);
+  try {
+    const response = await fetch(`${getFederatedContentRoot()}/federal/assets/data/languages-config.json`);
```
mokimo, line 368: Verify my assumption: since this is in federal, every single consumer will use the same config?

mokimo (commented):
A few suggestions and questions. I personally struggle a bit with two things:
- Variable naming
- A lot of if/else cases and conditions

Variable naming should be crystal clear and multiple conditions can be moved into intentionally named variables too. This starts to make sense when you have 3+ conditions or 2x2 conditions.

```js
// Option-1: hard to understand
if(language && map[part] && (localhost !== false) && !isWhatever) {if() {} else {}}


const isLangBased = language && map[part]
const isProd = (localhost === false) && !isWhatever
// Option-2: "Ah, it's language based and we only care for it on production"
if (isLangBased && isProd) 
```

To extend this, the if/else could also be extracted to a helper with early returns or be avoided otherwise.

libs/utils/languages.js
vhargrave: Hey Okan! 
Individual consumers will be having their own languageBased map. It's like the locales object, but for langauge based sites. 
I can put this into the scripts.js, however since the locales live in a separate file, I figured it made sense to do the same here thing. Thoughts?
Either way, this will only affect milo website performance and not consumer sites consuming milo.

libs/utils/utils.js
vhargrave: I really wish we could. 
There are two use cases to consider in my opinion. News.adobe.com is about to be a language based site, and so when you're on news.adobe.com, we need to know to convert the links to language based links. 
When you're on business.adobe.com though and you have links pointing to news.adobe.com, when they get localized they need to be localized directly. That means if I'm on business.adobe.com/ch_de/, my links pointing to news.adobe.com, need to point to news.adobe.com/de/ch/ or at the very least news.adobe.com/de/

libs/utils/utils.js
vhargrave: sure, the naming isn't great I agree. I'll rename 👍

libs/utils/utils.js
vhargrave: yes, that's my idea at the moment. I think until adobe.com is fully langauge based we need a config to let sites know which ones are language based already so that the hybrid of locale and lang links works properly. At some point this config will go away though 💪

libs/utils/languages.js
mokimo: Consumers have everything together: https://github.com/adobecom/bacom/blob/stage/scripts/scripts.js
We should also move this, not for performance within milo itself - but consumers mostly just copy our setup.
We don't want something with performance penalties accidentally getting copied as is

libs/utils/utils.js
mokimo: I see, unfortunate. 🫤

libs/utils/utils.js
sharmrj: I'd argue passing parameters makes the code easier to reason about in isolation. (It's also easier to test)

sukamat (commented):
@vhargrave Great work! And thanks for squeezing this in.

As discussed offline, mentioning few things here as well:
* "en" as a folder will exist in CMS (DA/SharePoint) but it will only host the region folders under it. There will be no pages authored directly under `/en/`. This setup only applies to English and no other languages as of today. Some of the currently proposed locale/path maps have been documented in the [wiki here](https://wiki.corp.adobe.com/display/WP4/Language-First+Path+Mapping+For+Adobe.com). News only supports a very small subset.
* The configurable setup in `languages-config.json` is great! This will help in onboarding sites as and when they move.
* Although the link-rewriter works for the interim needs of News where the needs are more specific and the scope is much smaller, this need to be further extended to cater for Express in Q3 that is currently being planned to be the next Langauge-First site and other sites that will be follow later this year and next year. Some of this is captured in [MWPW-169347](https://jira.corp.adobe.com/browse/MWPW-169347). From what I see, this will most probably be an extension of what you have implemented. This might come in Q3 based on whether Express gets prioritized for Q3 for the lang-first move (cc: @narcis-radu)

Couple of currently known issues: 
* Changes to the languageBased: false flag currently applied to de
* One issue that was called out by @Ruchika4 was with the footer links having `//`. For example, check the footer on this [page](https://language-site--news--adobecom.aem.page/en/apac/news?milolibs=vhargrave-MWPW-169933-language-based-links). Seems to happen only when milolibs param is added.

libs/utils/languages.js
vhargrave: alrighty, I'll move it.

libs/utils/languages.js
```diff
@@ -0,0 +1,22 @@
...
+  en: {
+    ietf: 'en',
```
narcis-radu, line 3: do we have use-cases where ietf is different than the lang ?

libs/utils/languages.js
```diff
@@ -0,0 +1,22 @@
...
+    tk: 'hah7vzn.css',
+    regions: [
+      { region: 'us', ietf: 'en-US', tk: 'hah7vzn.css' },
```
narcis-radu, line 6: would it be possible to rely on the lang based typekit by default and only set a custom one if needed? This way we would not need to set the same typekit for every region.

libs/utils/utils.js
narcis-radu: What about languages that news.adobe.com does not support, e.g. business.adobe.com/it -> news.adobe.com/it ?

libs/utils/utils.js
```diff
@@ -236,7 +272,13 @@ export const [setConfig, updateConfig, getConfig] = (() => {
       config = { env: getEnv(conf), ...conf };
       config.codeRoot = conf.codeRoot ? `${origin}${conf.codeRoot}` : origin;
       config.base = config.miloLibs || config.codeRoot;
-      config.locale = pathname ? getLocale(conf.locales, pathname) : getLocale(conf.locales);
+      if (conf.languages) {
+        config.locale = pathname
+          ? getLanguage(conf.languages, conf.locales, pathname)
```
narcis-radu, line 277: is `pathname` check needed? I see your `getLanguage` method has `pathname = window.location.pathname` by default.

libs/utils/utils.js
```diff
@@ -306,6 +348,38 @@ export const getFederatedUrl = (url = '') => {
   return url;
 };
 
+export function hasLanguageLinks(area) {
+  for (const link of area.querySelectorAll('a[href]')) {
```
narcis-radu, line 352: Why not use something like ``area.querySelectorAll(`a[href*="${your-domain}"]`)``

libs/utils/utils.js
narcis-radu: >When you're on business.adobe.com though and you have links pointing to news.adobe.com, when they get localized they need to be localized directly.

Can't that be solved through the mapping used in `loadLanguageConfig` ? https://main--federal--adobecom.aem.page/federal/assets/data/languages-config.json

libs/utils/utils.js
```diff
@@ -370,17 +449,33 @@ export function localizeLink(
...
+    const site = siteLanguages?.find((s) => s.domainMatches.some((d) => url.hostname.includes(d)));
+
+    if (locale.languageBased && url.hostname !== 'localhost') {
+      const hasValidLanguage = site?.languages.some((l) => locale.prefix.includes(l));
+      if (!site || !hasValidLanguage) {
+        prefix = mapPrefixToLocale(prefix);
+      }
+    } else if (site) {
+      const siteHasLanguage = site.languages.some((l) => `/${l}` === prefix);
+      if (!siteHasLanguage) {
```
narcis-radu, line 474: nit: this can be a one liner

libs/utils/utils.js
```diff
@@ -152,11 +152,19 @@ const TARGET_TIMEOUT_MS = 4000;
 const LANGSTORE = 'langstore';
 const PREVIEW = 'target-preview';
 const PAGE_URL = new URL(window.location.href);
+const LANGUAGE_BASED_DOMAINS = [
+  // don't add milo too. It's a special case because of tools, merch, etc.
+  'news.adobe.com',
+  '--news--adobecom.',
```
sukamat, line 158: @vhargrave Will the languages-config.json be pulled in only when the request comes through `'news.adobe.com' and '--news--adobecom.'` ? If not, it will be nice to make this configurable as well?

vhargrave:
@sukamat yes, I believe this is expected, but let's discuss: 
--news--adobecom gets converted to a relative link, because all of these types of links get converted into main--cc-adobe.com links directly by EDS. You can see that happening on the plain.html page - https://main--cc--adobecom.aem.page/ch_de/drafts/sukamat/links.plain.html. That's why this "strange" behaviour is happening. 

For the second part, I really like the idea. I don't think it's possible though. Even if I set the metadata in the federal project for example, it doesn't get applied to every single site. @mokimo maybe you have an idea?

libs/utils/languages.js
vhargrave: Good point! - For the lang probably not and for the lang-region very rarely

vhargrave: I'll remove the unnecessary usages.

libs/utils/languages.js
vhargrave: definitely!

sukamat:
@vhargrave 
What I am saying is, the News URLs linked in Bacom or CC page does not have to be catered at all. 

Because:
* The EDS behavior makes the author **NOT** add cross-site `aem.page` URLs. This is the case today as well. For cross-site linking, they always add the absolute URLs
* And if absolute URLs are added to the Word doc, AFAIK, Milo does not modify those links and keeps them as-is.

And for detecting which sites are language-first site, the metadata.xlsx file of just the onboarded site need to be updated; in this case it will be News. Other sites metadata need not be touched. Correct? Or am I missing something here?

libs/utils/utils.js
vhargrave: yes, the pathname check is needed for use cases like in the region-nav .
@Ruchika4 will be using it so she uses the same logic as utils for getting the lang & region from a url that she passes as a parameter.

libs/utils/utils.js
vhargrave: This was a great idea. Unfortuntately I won't be able to use it as I realized that I had to cover some more cases for these links (adobe.com/de/express for example) and the logic has now changed to accomodate that.

vhargrave:
> @vhargrave What I am saying is, the News URLs linked in Bacom or CC page does not have to be catered at all.
> 
> Because:
> 
> * The EDS behavior makes the author **NOT** add cross-site `aem.page` URLs. This is the case today as well. For cross-site linking, they always add the absolute URLs
> * And if absolute URLs are added to the Word doc, AFAIK, Milo does not modify those links and keeps them as-is (as seen in my example above).
> 
> And for detecting which sites are language-first site, the metadata.xlsx file of just the onboarded site need to be updated; in this case it will be News. Other sites metadata need not be touched. Correct? Or am I missing something here?

@sunilkamat aha, I see what you're saying. Yes, there's one thing that I think is being missed. If you add a domain as a prodDomain in your config, then the links do get localized for that domain. I've added similar behaviour for language links as well.

libs/utils/utils.js
```diff
@@ -308,6 +351,40 @@ export const getFederatedUrl = (url = '') => {
   return url;
 };
 
+export function hasLanguageLinks(container, paths = LANGUAGE_BASED_PATHS) {
+  if (!container) return false;
+  const links = container.querySelectorAll('a');
```
narcis-radu, line 356: I remember @overmyheadandbody did some work to remove multiple iterations over links when transforming them. Would this have any effect on the work?

libs/utils/utils.js
```diff
@@ -200,6 +207,36 @@ export function getLocale(locales, pathname = window.location.pathname) {
...
+    if (matchingRegion.ietf)language.ietf = matchingRegion.ietf;
+    if (matchingRegion.tk) language.tk = matchingRegion.tk;
+    regionPath = matchingRegion ? `/${region}` : '';
+  }
+
+  // if no language, allow for support of locale based routing still
+  // when removing this, account for the root /en/ not existing and instead being just /
+  if (!language || (language.languageBased === false && !language.region)) {
+    const locale = getLocale(locales, pathname);
+    if (locale.prefix === '') locale.language = 'en';
```
narcis-radu, line 230: can we define e constant for default language and use it across the code?

libs/utils/utils.js
vhargrave: It's certainly regressive, but I think for now the performance impact of iterating over links checking for language based sites is smaller than loading in the whole language map. At some point this might change though, and then I think we just load in the language map on every page load until every site is language based.

mokimo (commented):
Almost there

libs/utils/utils.js
```diff
@@ -372,17 +447,34 @@ export function localizeLink(
...
+      const languageInUseBySite = site.languages.find((l) => `${l}` === mappedLanguageFromPrefix.languagePath);
+      if (languageInUseBySite) {
+        prefix = languageInUseBySite === DEFAULT_LANG ? '' : `/${languageInUseBySite}`;
+      }
+    } else if (locale.language && !relative && !site?.languages.some((l) => (l === DEFAULT_LANG ? '' : `/${l}`) === prefix)) {
+      // On a language based site like news but target url is not language based and needs a locale
+      const mappedLocaleFromLanguage = localeToLanguageMap?.find((m) => `/${m.languagePath}` === prefix);
+      prefix = mappedLocaleFromLanguage ? `${mappedLocaleFromLanguage.locale === '' ? '' : '/'}${mappedLocaleFromLanguage.locale}` : prefix;
+    }
+    const urlPath = `${prefix}${path}${url.search}${hash}`;
```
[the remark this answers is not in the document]
vhargrave: yes, we can. I'll do that.

libs/utils/utils.js
```diff
@@ -372,17 +447,34 @@ export function localizeLink(
...
+    const site = siteLanguages?.find((s) => s.pathMatches.some((d) => isPathMatch(d, url.href)));
+
+    // On a locale based site, but target url has language based paths
+    if (!locale.language && site && localeToLanguageMap) {
+      const mappedLanguageFromPrefix = localeToLanguageMap?.find((m) => `${m.locale === '' ? '' : '/'}${m.locale}` === prefix);
+      const languageInUseBySite = site.languages.find((l) => `${l}` === mappedLanguageFromPrefix.languagePath);
+      if (languageInUseBySite) {
+        prefix = languageInUseBySite === DEFAULT_LANG ? '' : `/${languageInUseBySite}`;
+      }
+    } else if (locale.language && !relative && !site?.languages.some((l) => (l === DEFAULT_LANG ? '' : `/${l}`) === prefix)) {
```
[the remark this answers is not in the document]
vhargrave: This version is more readable agree, but it's less performant. I guess since the .some check that's running is very light though I'll take your more readable approach. Thanks!

libs/utils/utils.js
```diff
@@ -372,17 +468,20 @@ export function localizeLink(
...
     const isLocalizable = relative || (prodDomains && prodDomains.includes(url.hostname))
       || overrideDomain;
     if (!isLocalizable) return processedHref;
     const isLocalizedLink = path.startsWith(`/${LANGSTORE}`)
       || path.startsWith(`/${PREVIEW}`)
-      || Object.keys(locales).some((loc) => loc !== '' && (path.startsWith(`/${loc}/`)
-        || path.endsWith(`/${loc}`)));
+      || (languages && Object.keys(languages).some((lang) => lang !== '' && (path.startsWith(`/${lang}/`))))
+      || (locales && Object.keys(locales).some((loc) => loc !== '' && (path.startsWith(`/${loc}/`)
+        || path.endsWith(`/${loc}`))));
```
mokimo, line 480: Should we also extract this into variables? The isLocalizedLink chain is quite long, it's not immediately obvious all the cases it tries to check

libs/utils/utils.js
```diff
@@ -357,6 +432,27 @@ function getExtension(path) {
   return pageName.includes('.') ? pageName.split('.').pop() : '';
 }
 
+function getPrefixBySite(locale, url, relative) {
```
mokimo, line 435: I personally find the naming in this function still un-intuitive... if you or another reviewer has some ideas, that's one area where I still see potential, that'll make our lives easier in the future to understand what's going on.

As you mentioned, it's a rather small but very impactful piece of our core logic so everything that happens here should be crystal clear and working with locales, languages, maps becomes very confusing.

mokimo (approved):
Coolio 🚀

libs/utils/utils.js
vhargrave: Naming is really hard to get right, I agree. I also don't think it's perfect, but also don't have better ideas so am open to suggestions.

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-169278

libs/utils/utils.js
```diff
@@ -154,11 +154,19 @@ const TARGET_TIMEOUT_MS = 4000;
 const LANGSTORE = 'langstore';
 const PREVIEW = 'target-preview';
 const PAGE_URL = new URL(window.location.href);
+const LANGUAGE_BASED_PATHS = [
+  // don't add milo too. It's a special case because of tools, merch, etc.
+  'news.adobe.com',
```
rgclayton, line 159: Late to the game, but blog.adobe.com is also language based.

afmicka:
@vhargrave @mokimo @NadiiaSokolova this PR has been merged with failing Nala tests. These changes are preventing pages in ROW locales (at least de) to load and now it is breaking all the PRs that have synced with the latest stage. Please take a look and provide the fix.

The line in question: 
https://github.com/adobecom/milo/commit/59c81da92ed9be324704483497ca6c4b670da580#diff-28218f239d8d9c4d41e1924854440b50964619e56058fdd0ae243ff5ff6c616dR221

Open the page in incognito: 
https://stage--milo--adobecom.aem.live/de/drafts/nala/features/promotions/promo-page-filter

Screenshot 2025-04-11 at 09 35 10

PR #3942, opened 2025-04-10
MWPW-170824: Fix image label overflow

zagi25:
* Fix image label overflow

Testing: Open preflight tool from sidekick, wait for image labels to show up

Resolves: [MWPW-170824](https://jira.corp.adobe.com/browse/MWPW-170824)

**Test URLs:**
- After: https://mwpw-170824-preflight-label-overflow--milo--adobecom.aem.page/?martech=off

**CC URL**:
- Before: https://main--cc--adobecom.hlx.page/products/substance3d
- After: https://main--cc--adobecom.hlx.page/products/substance3d?milolibs=MWPW-170824-preflight-label-overflow

### DIFF
--- a/libs/blocks/preflight/preflight.css
+++ b/libs/blocks/preflight/preflight.css
@@ -667,7 +667,7 @@ body.preflight-assets-analysis {
 
 picture:has(.picture-meta) {
   position: relative;
-  display: block;
+  display: block !important;
 }
 
 .picture-meta {


libs/blocks/preflight/preflight.css
```diff
@@ -667,7 +667,7 @@ body.preflight-assets-analysis {
 
 picture:has(.picture-meta) {
   position: relative;
-  display: block;
+  display: block !important;
```
mokimo, line 670: As wise man @overmyheadandbody usually states, let's avoid the `!important` - that should be the absolute last thing we consider if all fails

zagi25: Totally agree. This is the CSS that causes the issue, don't know why it is there to me it seems that it doesn't do anything, my thought was that it is not wise to remove it and I decided to override it with `!important` just in case there is other similar rules. 
Screenshot 2025-04-10 at 11 41 37
I have similar issue on current ticket where image disappears, because image is set using `position: absolute` for unknown reason and I am forced to do some overrides because I am not sure that we want to change this rules and do the block testing

mokimo: Removing the CSS that causes it is definitely preferred. If you don't find a use-case on why it exists to begin with, let's remove it

zagi25:
Will be fixed by another PR.

PR #3943, opened 2025-04-10
MWPW-170822: Fix broken label

zagi25:
* Fix broken preflight image label

Testing: Open preflight tool via sidekick and wait for image labels to load

Resolves: [MWPW-170822](https://jira.corp.adobe.com/browse/MWPW-170822)

**Test URLs:**
- After: https://mwpw-170822-preflight-broken-label--milo--adobecom.aem.page/?martech=off

**BACOM URL:**
- Before: https://main--bacom--adobecom.hlx.page/
- After: https://main--bacom--adobecom.hlx.page/?milolibs=MWPW-170822-preflight-broken-label

### DIFF
--- a/libs/blocks/preflight/preflight.css
+++ b/libs/blocks/preflight/preflight.css
@@ -686,6 +686,7 @@ picture:has(.picture-meta) {
   line-height: 1.2;
   z-index: 9;
   opacity: 0.7;
+  height: auto !important;
 }
 
 .picture-meta.no-picture-tag {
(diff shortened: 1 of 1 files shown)


libs/blocks/preflight/preflight.css
```diff
@@ -686,6 +686,7 @@ picture:has(.picture-meta) {
   line-height: 1.2;
   z-index: 9;
   opacity: 0.7;
+  height: auto !important;
```
mokimo, line 689: As wise man @overmyheadandbody usually states, let's avoid the `!important` - that should be the absolute last thing we consider if all fails

mokimo: If we can get away slightly refactoring what currently exists, that'd be preferred

PR #3954, opened 2025-04-11
MWPW-169204: create QS from the link

3ch023:
Add a LOC-robust Quantity Selector approach:
Instead of adding QS as unsorted list it can be added as a Link:
HREF:
#qs&min=1&max=10&step=1&default=2&max-input=180
Link TITLE: QS title (translatable)
image

For backwards compatibility old approach will keep working for some time & will be removed once GWP does content updates. Check https://mwpw-169204--milo--adobecom.aem.live/drafts/mariia/pr/qs1?martech=off to see 'old' approach still working

Resolves: https://jira.corp.adobe.com/browse/MWPW-169204

**Test URLs:**
- Before: https://main--milo--adobecom.aem.live/drafts/mariia/pr/qs?martech=off
- After: https://mwpw-169204--milo--adobecom.aem.live/drafts/mariia/pr/qs?martech=off

### DIFF
--- a/libs/blocks/merch-card/merch-card.js
+++ b/libs/blocks/merch-card/merch-card.js
@@ -111,7 +111,7 @@ function extractQuantitySelect(el, merchCard) {
     quantitySelectConfig.remove();
     [attributes.min, attributes.max, attributes.step, attributes['default-value'], attributes['max-input']] = values;
   }
-  if (!attributes.min || !attributes.max) {
+  if (!attributes.min || !attributes.max || !attributes.step) {
     return null;
   }
   import('../../deps/mas/merch-quantity-select.js');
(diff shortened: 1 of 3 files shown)


libs/blocks/merch-card/merch-card.js
```diff
@@ -702,7 +711,7 @@ export default async function init(el) {
     merchCard.appendChild(footer);
 
     if (MULTI_OFFER_CARDS.includes(cardType)) {
-      const quantitySelect = extractQuantitySelect(el);
+      const quantitySelect = extractQuantitySelect(el, merchCard);
```
yesil, line 714: in case there is an unexpected exception, I think it should log a lana error and silently ignore.

yesil (approved):
left a nice to have commnet.

bozojovicic (approved):
Approved but please wrap extractQuantitySelect() with try{}catch

PR #3956, opened 2025-04-11
Fix(Mwpw-169955):share blocks copy to clipboard tooltip repositioned

sharath-kannan:
* Re-positioned copy to clipboard tooltip

Resolves: [MWPW-169955](https://jira.corp.adobe.com/browse/MWPW-169955)

**Test URLs:**
- After: https://mwpw-169955--milosh--sharath-kannan.aem.page/?martech=off

https://main--cc--adobecom.hlx.page/jp/products/photoshop/generative-fill?milolibs=mwpw-169955--milosh--sharath-kannan


libs/blocks/share/share.css
```diff
@@ -146,3 +147,13 @@ main > .section.center .share.inline p.icon-container {
...
+
+@media screen and (min-width: 600px) {
+  .share .copy-to-clipboard::before {
+    left:50%;
+  }
+
+  .share .copy-to-clipboard::after {
+    left:50%;
+  }
+}
```
meganthecoder, line 159: Would be good to clean up spacing here, a space after the colons and newline at end of file
```suggestion
@media screen and (min-width: 600px) {
  .share .copy-to-clipboard::before {
    left: 50%;
  }

  .share .copy-to-clipboard::after {
    left: 50%;
  }
}

```

meganthecoder (commented):
I think it looks better if you keep the pointer (::after) centered on the button even when the dialog is pushed over, but that's just my personal preference.

Also, did you check this out in other languages, especially RTL?

sharath-kannan:
> I think it looks better if you keep the pointer (::after) centered on the button even when the dialog is pushed over, but that's just my personal preference.
> 
> Also, did you check this out in other languages, especially RTL?

Yes, that seems right. i will change the pointer to be in the centre.
Good catch with RTL, I have considered that now.

libs/blocks/share/share.css
```diff
@@ -5,6 +5,7 @@
   margin-right: auto;
   text-align: center;
   padding: 64px 0;
+  overflow: hidden;
```
nishantka, line 8: I think we should add a note for character limit in share block for authoring this placeholder value as it looks like the char limit will be reduced after this change and overflowing content will be hidden.

nishantka (approved):

meganthecoder (approved):

libs/blocks/share/share.css
sharath-kannan: That might not be necessary since these are going to be constant values in every locale and we have considered the locale with longest constant value and hence the rest should work fine.

spadmasa:
Validation done on the below urls 
https://main--cc--adobecom.hlx.live/jp/products/photoshop/generative-fill?milolibs=mwpw-169955--milosh--sharath-kannan
image
image
image
image
https://main--cc--adobecom.hlx.live/il_he/products/photoshop/generative-fill?milolibs=mwpw-169955--milosh--sharath-kannan
image
image
https://main--cc--adobecom.hlx.live/ua/products/photoshop/generative-fill?milolibs=mwpw-169955--milosh--sharath-kannan
image
image

bandana147 (approved):

PR #3959, opened 2025-04-11
MWPW-168449: [A11Y 2025] [NAV] | Link does not have discernible link text

hkuraware:
* Fixed A11y issue.

Resolves: [MWPW-168449](https://jira.corp.adobe.com/browse/MWPW-168449)

**Test URLs:**
- After: https://mwpw-168449-a11y-2025--milo--hkuraware.aem.page/?martech=off


hkuraware:
This can be fixed with authoring change as discussed with @Deva309 .
cc: @salonijain3

PR #3964, opened 2025-04-14
Merging Tags: PersV2 and Hybrid-pers

akanshaa-18:
Merging Tags: PersV2 and Hybrid-pers
Resolves: https://jira.corp.adobe.com/browse/ENB-7894

Test URLs: https://main--cc--adobecom.aem.page/drafts/akansha/photoshop?milolibs=ENB-7894--milo--akanshaa-18&personalization-v2=on&target=on

Before: https://main--milo--adobecom.aem.live/?martech=off
After: https://enb-7894--milo--akanshaa-18.aem.live/?martech=off


swamu (approved):
LGTM 👍

pooja-mittal:
Validation is completed, good to go live.

libs/martech/helpers.js
```diff
@@ -564,21 +562,19 @@ export const loadAnalyticsAndInteractionData = async (
...
+    if (filteredPayload.length) {
+      sendPropositionDisplayRequest(filteredPayload, env, requestPayload);
     }
+    const alloyData = {
+      destinations: getPayloadsByType(targetRespJson, 'activation:pull'),
+      propositions: resultPayload,
+      inferences: getPayloadsByType(targetRespJson, 'rtml:inferences'),
+      decisions: [],
+    };
+    window.dispatchEvent(new CustomEvent('alloy_sendEvent', { detail: alloyData }));
```
sharmrj, line 575: instead of dispatching a custom event that you then listen for elsewhere, you could call the code in the event listener here directly. You'd want to do this to reduce the amount of indirection in your code; it's easier to jump to a function definition, or even to just have the code inline than to wonder about eventlisteners. and so on.

It also forces you to make the flow of data explicit. If you have to keep the event to take advantage of some data in a closure, then you may have to look into fixing the deeper data flow issue.

I realize that the event listener is in personlization and a refactor like this is beyond the scope for this ticket so you probably should add a short comment explaining where this event is likely being listened to, at least.

libs/utils/utils.js
```diff
@@ -1323,7 +1322,7 @@ async function checkForPageMods() {
       const now = performance.now();
       performance.mark('interaction-start');
       const data = await loadAnalyticsAndInteractionData(
-        { locale, env: getEnv({})?.name, calculatedTimeout, hybridPersEnabled },
+        { locale, env: getEnv({})?.name, calculatedTimeout },
```
sharmrj, line 1325: you may or may not want a fallback environment if this turns out null.

sharmrj (approved):
lgtm

libs/martech/helpers.js
```diff
@@ -520,10 +520,8 @@ export const loadAnalyticsAndInteractionData = async (
   const CURRENT_DATE = new Date();
   const localTime = CURRENT_DATE.toISOString();
   const timezoneOffset = CURRENT_DATE.getTimezoneOffset();
-  if (hybridPersEnabled) {
-    window.hybridPers = true;
-  }
-  const hitType = hybridPersEnabled ? 'pageView' : 'propositionFetch';
+  window.hybridPers = true;
```
amitbikram, line 523: Is it required?

amitbikram (changes requested):

libs/martech/helpers.js
akanshaa-18: Hi @amitbikram 
Yes `window.hybridPers = true;` is required because it is used in the launch script to suppress the launch code from executing.

amitbikram (approved):

PR #3977, opened 2025-04-15
[MWPW-170108] - Notification a11y

DKos95:
This resolves the issue where the notification container was missing a role and an aria-label, the aria-label was changed for the close button, added an announcement for the screen reader "Banner closed" after closing the banner and focusing on the next element after finishing the announcement.

Update:
After a meeting with Taryn, it was decided that when closing the banner the focus after the closing announcement would shift to the previous sections selected/checked element like an active tab/element if it has one and if not then the last focusable element in the previous section.

Resolves: [MWPW-170108](https://jira.corp.adobe.com/browse/MWPW-170108)

**Previous section last focusable element Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/dusan/notification-close-testing?martech=off
- After: https://notification-a11y--milo--adobecom.aem.page/drafts/dusan/notification-close-testing?martech=off

**Previous section active tab Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/dusan/notification-close-two?martech=off
- After: https://notification-a11y--milo--adobecom.aem.page/drafts/dusan/notification-close-two?martech=off

**Sticky Banner use case Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/dusan/notification-close-four
- After: https://notification-a11y--milo--adobecom.aem.page/drafts/dusan/notification-close-four

**Use case when previous section doesn't have focusable element or selected/checked element, so we go to it's previous section and do this over and over until we find something focusable in one of the previous sections and if no sections have a focusable element, as a last resort we focus the headers last focusable element - Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/dusan/notification-close-three?martech=off
- After: https://notification-a11y--milo--adobecom.aem.page/drafts/dusan/notification-close-three?martech=off

**DC Test URLs:**
- Before: https://main--dc--adobecom.hlx.page/dc-shared/fragments/generative-ai-pdf/promobar-try-demo-desktop
- After: https://main--dc--adobecom.hlx.page/dc-shared/fragments/generative-ai-pdf/promobar-try-demo-desktop?milolibs=notification-a11y

### DIFF
--- a/libs/blocks/notification/notification.css
+++ b/libs/blocks/notification/notification.css
@@ -340,6 +340,16 @@
   display: none;
 }
 
+.visibility-hidden {
+  position: absolute;
+  width: 0;
+  height: 0;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  white-space: nowrap;
+  z-index: -1;
+}
+
 .notification.pill .foreground .text > :not(.action-area) {
   padding-inline-end: var(--spacing-xxs);
   inline-size: calc(100% - var(--spacing-xxs));
--- a/libs/blocks/notification/notification.css
+++ b/libs/blocks/notification/notification.css
@@ -341,7 +341,9 @@
 }
 
 .notification-visibility-hidden {
-  position: absolute;
+  position: fixed;
+  top: 0;
+  left: 0;
   width: 0;
   height: 0;
   overflow: hidden;
--- a/libs/blocks/notification/notification.css
+++ b/libs/blocks/notification/notification.css
@@ -340,7 +340,7 @@
   display: none;
 }
 
-.visibility-hidden {
+.notification-visibility-hidden {
   position: absolute;
   width: 0;
   height: 0;
--- a/libs/blocks/notification/notification.js
+++ b/libs/blocks/notification/notification.js
@@ -101,6 +101,14 @@ function addCloseAction(el, btn) {
     liveRegion.textContent = 'Banner closed';
     document.body.appendChild(liveRegion);
     liveRegion.focus();
+    let isSticky = false;
+    let rect;
+    el.closest('.section').classList.forEach((cls) => {
+      if (cls.includes('sticky')) {
+        isSticky = true;
+        rect = el.closest('.section').getBoundingClientRect();
+      }
+    });
     el.style.display = 'none';
     el.closest('.section')?.classList.add('close-sticky-section');
     if (el.classList.contains('focus')) {
@@ -125,6 +133,12 @@ function addCloseAction(el, btn) {
           ? focusableElements[focusableElements.length - 1]
           : null;
       };
+
+      if (isSticky) {
+        const elementAtPosition = document.elementFromPoint(rect.left, rect.top);
+        const stickySection = !elementAtPosition.classList.contains('section') ? elementAtPosition.closest('.section') : elementAtPosition;
+        focusTarget = findFocusableInSection(stickySection);
+      }
 
       let currentSection = el.closest('.section').previousElementSibling;
       while (currentSection && !focusTarget) {
--- a/libs/blocks/notification/notification.js
+++ b/libs/blocks/notification/notification.js
@@ -90,15 +90,13 @@ function addCloseAction(el, btn) {
   btn.addEventListener('click', (e) => {
     if (btn.nodeName === 'A') e.preventDefault();
 
-    // Create accessibility live region
     const liveRegion = createTag('div', {
       class: 'notification-visibility-hidden',
       'aria-live': 'assertive',
       'aria-atomic': 'true',
       role: 'status',
       tabindex: '-1',
-    });
-    liveRegion.textContent = 'Banner closed';
+    }, 'Banner closed');
     document.body.appendChild(liveRegion);
     liveRegion.focus();
     let isSticky = false;
@@ -141,18 +139,16 @@ function addCloseAction(el, btn) {
         focusTarget = findFocusableInSection(stickySection);
       }
 
-      let currentSection = el.closest('.section').previousElementSibling;
+      let currentSection = el.closest('.section')?.previousElementSibling;
       while (currentSection && !focusTarget) {
         focusTarget = findFocusableInSection(currentSection);
         if (!focusTarget) currentSection = currentSection.previousElementSibling;
       }
 
-      if (!focusTarget) {
-        const header = document.querySelector('header');
-        if (header) {
-          const headerFocusable = [...header.querySelectorAll(focusableSelector)];
-          focusTarget = headerFocusable[headerFocusable.length - 1];
-        }
+      const header = document.querySelector('header');
+      if (!focusTarget && header) {
+        const headerFocusable = [...header.querySelectorAll(focusableSelector)];
+        focusTarget = headerFocusable[headerFocusable.length - 1];
       }
 
       liveRegion?.remove();
--- a/libs/blocks/notification/notification.js
+++ b/libs/blocks/notification/notification.js
@@ -91,7 +91,7 @@ function addCloseAction(el, btn) {
     if (btn.nodeName === 'A') e.preventDefault();
 
     const liveRegion = createTag('div', {
-      class: 'visibility-hidden',
+      class: 'notification-visibility-hidden',
       'aria-live': 'assertive',
       'aria-atomic': 'true',
       role: 'status',
--- a/libs/blocks/notification/notification.js
+++ b/libs/blocks/notification/notification.js
@@ -104,14 +104,12 @@ function addCloseAction(el, btn) {
     let isSticky = false;
     let rect;
     const sectionElement = el.closest('.section');
-    if (sectionElement) {
-      sectionElement.classList.forEach((cls) => {
-        if (cls.includes('sticky')) {
-          isSticky = true;
-          rect = sectionElement.getBoundingClientRect();
-        }
-      });
+
+    if (sectionElement?.className.includes('sticky')) {
+      isSticky = true;
+      rect = sectionElement.getBoundingClientRect();
     }
+
     el.style.display = 'none';
     el.closest('.section')?.classList.add('close-sticky-section');
     if (el.classList.contains('focus')) {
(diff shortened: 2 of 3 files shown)


libs/blocks/notification/notification.css
```diff
@@ -340,6 +340,16 @@
   display: none;
 }
 
+.visibility-hidden {
```
narcis-radu, line 343: The class name is general enough that might affect other elements if the code gets loaded. Can we make it more specific, like `notification-visibility-hidden` or `notification-hidden-wrapper`

libs/blocks/notification/notification.js
```diff
@@ -89,14 +89,33 @@ function wrapCopy(foreground) {
...
+      tabindex: '-1',
+    });
+    liveRegion.textContent = 'Banner closed';
+    el.closest('.section')?.appendChild(liveRegion);
+    liveRegion.focus();
     el.style.display = 'none';
     el.closest('.section')?.classList.add('close-sticky-section');
     document.dispatchEvent(new CustomEvent('milo:sticky:closed'));
+
+    setTimeout(() => {
```
zagi25, line 107: What is the purpose of this timeout ? When I tab into notification, close it and continue to tab, timeout will return the focus to the next element after the notification 

https://github.com/user-attachments/assets/b7bbc1bb-69d2-44e8-bd68-1c872516ec07

DKos95: @zagi25 That is the intention, that the user continues where he left off, without this after the announcement he would have to traverse the whole page to continue where he left off just because he closed the banner.

zagi25: Understandable, but if the user closes the banner with the keyboard, the most plausible case is that he tabbed into the banner, closed it and continued to tab and I think we shouldn't interfere in that case.  

Another issue with this, if I close the banner when I am on top of the page, page will scroll to the footer (next focusable element after notification in the dom)

https://github.com/user-attachments/assets/f88ece37-9cb5-4583-9614-611104bf457c

**You need to wait for the footer to load to test this.**

Possible solution for this is to use [elementFromPoint](https://developer.mozilla.org/en-US/docs/Web/API/Document/elementFromPoint) , to check which element/section is behind the banner and find the first focusable element. 

Also when banner is closed, because you focus 'Banner Closed' on L102, page will scroll immediately to dummy element. (also visible in the video above)

DKos95: > Understandable, but if the user closes the banner with the keyboard, the most plausible case is that he tabbed into the banner, closed it and continued to tab and I think we shouldn't interfere in that case.

But that is the problem, that he can't continue without us focusing on something, because when he closes the banner we need to focus a element to pronounce it was closed, then focus on something after that.

> Another issue with this, if I close the banner when I am on top of the page, page will scroll to the footer (next focusable element after notification in the dom)

It's either we focus the next, previous element, not remove the invisible element and stay on the it or let the screenreader start from the top again when we remove the invisible element we used for announcing the banner was closed.

> Also when banner is closed, because you focus 'Banner Closed' on L102, page will scroll immediately to dummy element. (also visible in the video above)

We need to focus on an element for it to read out that the banner is closed, don't see a way around this, am open to suggestions.

DKos95: > Also when banner is closed, because you focus 'Banner Closed' on L102, page will scroll immediately to dummy element. (also visible in the video above)

Found a solution for this issue, and it is to make the invisible announcement fixed so it will always be in screen view.

`.notification-visibility-hidden { position: fixed;
  top: 0;
  left: 0;`

libs/blocks/notification/notification.js
```diff
@@ -89,14 +89,78 @@ function wrapCopy(foreground) {
...
+      tabindex: '-1',
+    });
+    liveRegion.textContent = 'Banner closed';
+    document.body.appendChild(liveRegion);
+    liveRegion.focus();
+    let isSticky = false;
+    let rect;
+    const sectionElement = el.closest('.section');
+    if (sectionElement) {
+      sectionElement.classList.forEach((cls) => {
```
zagi25, line 108: If you just need to check if element contains class with `sticky` e.g. `sticky-top`, `sticky-bottom`, `sticky-whatever`, you can simplify this:
`sectionElement.className.includes('sticky')`

libs/blocks/notification/notification.js
```diff
@@ -89,14 +89,78 @@ function wrapCopy(foreground) {
...
+      while (currentSection && !focusTarget) {
+        focusTarget = findFocusableInSection(currentSection);
+        if (!focusTarget) currentSection = currentSection.previousElementSibling;
+      }
+
+      if (!focusTarget) {
+        const header = document.querySelector('header');
+        if (header) {
+          const headerFocusable = [...header.querySelectorAll(focusableSelector)];
+          focusTarget = headerFocusable[headerFocusable.length - 1];
```
zagi25, line 152:
```suggestion
          focusTarget = headerFocusable.at(-1);
```

libs/blocks/notification/notification.js
```diff
@@ -89,14 +89,78 @@ function wrapCopy(foreground) {
...
+        }
+      }
+
+      liveRegion?.remove();
+      if (focusTarget) focusTarget.focus();
```
zagi25, line 157: To prevent scrolling the element into view you can use
```suggestion
      if (focusTarget) focusTarget.focus({ preventScroll: true });
```

libs/blocks/notification/notification.js
```diff
@@ -89,14 +89,78 @@ function wrapCopy(foreground) {
...
+      let focusTarget;
+
+      const findFocusableInSection = (section) => {
+        if (!section) return null;
+
+        const selectedElement = section.querySelector(selectedSelector);
+        if (selectedElement) return selectedElement;
+
+        const focusableElements = [...section.querySelectorAll(focusableSelector)];
+        return focusableElements.length > 0
```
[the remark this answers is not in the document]
DKos95: Array.at() is not supported in Safari 15, iOS Safari 15.0-15.1 eslint

libs/blocks/notification/notification.js
DKos95: Array.at() is not supported in Safari 15, iOS Safari 15.0-15.1 eslint

libs/blocks/notification/notification.js
```diff
@@ -89,14 +89,76 @@ function wrapCopy(foreground) {
 function addCloseAction(el, btn) {
   btn.addEventListener('click', (e) => {
     if (btn.nodeName === 'A') e.preventDefault();
+
+    // Create accessibility live region
```
JasonHowellSlavin, line 93: can probably remove this comment, we try to keep as many as we can out and the variable name is good

libs/blocks/notification/notification.js
```diff
@@ -89,14 +89,76 @@ function wrapCopy(foreground) {
...
+    const liveRegion = createTag('div', {
+      class: 'notification-visibility-hidden',
+      'aria-live': 'assertive',
+      'aria-atomic': 'true',
+      role: 'status',
+      tabindex: '-1',
+    });
+    liveRegion.textContent = 'Banner closed';
```
JasonHowellSlavin, line 101:
```suggestion
    const liveRegion = createTag('div', {
      class: 'notification-visibility-hidden',
      'aria-live': 'assertive',
      'aria-atomic': 'true',
      role: 'status',
      tabindex: '-1',
    }, 'Banner closed');
```

libs/blocks/notification/notification.js
```diff
@@ -89,14 +89,76 @@ function wrapCopy(foreground) {
...
+        const elementAtPosition = document.elementFromPoint(rect.left, rect.top);
+        const stickySection = elementAtPosition.closest('.section');
+        focusTarget = findFocusableInSection(stickySection);
+      }
+
+      let currentSection = el.closest('.section').previousElementSibling;
+      while (currentSection && !focusTarget) {
+        focusTarget = findFocusableInSection(currentSection);
+        if (!focusTarget) currentSection = currentSection.previousElementSibling;
+      }
```
JasonHowellSlavin, line 144: Is there a scenario possible where while iterating through the focusables that a null `section` param gets passed into `findFocusabeInSection` and the currentSection has not `previousElementSibling`?

libs/blocks/notification/notification.js
```diff
@@ -89,14 +89,76 @@ function wrapCopy(foreground) {
...
+
+      if (!focusTarget) {
+        const header = document.querySelector('header');
+        if (header) {
+          const headerFocusable = [...header.querySelectorAll(focusableSelector)];
+          focusTarget = headerFocusable[headerFocusable.length - 1];
+        }
+      }
```
JasonHowellSlavin, line 152: nit: flatten
```suggestion
      if (!focusTarget && document.querySelector('header')) {
          const headerFocusable = [...header.querySelectorAll(focusableSelector)];
          focusTarget = headerFocusable[headerFocusable.length - 1];
      }
```

JasonHowellSlavin (commented):
Are there some tests that can be updated to reflect the new behavior here?

libs/blocks/notification/notification.js
DKos95: Hello Jason, this use case should be covered with the `if (!section) return null;` inside the function you mentioned `findFocusabeInSection`.

DKos95:
> Are there some tests that can be updated to reflect the new behavior here?

In my opinion this can be done partially with this solution so I don't know if it should be done if we can't test that the next element is focused. The reason being is that we have a timeout here for 2000 which is the exact time the announcer needs to read out that the banner is closed and then we switch focus, anything less and it would cut off the announcer, and when running unit tests from my knowledge at least, we aren't able to put a timeout of 2000 or more to wait and check if the next element has been focused since the test will fail automatically. I am obviously open to suggestions if I am missing something, thanks for the review @JasonHowellSlavin!

JasonHowellSlavin (approved):
Sounds good. We should see if we take a hit on code coverage. If so, you may be able to make a test for the `findFocusableInSection` code at least by pulling it out of the set timeout. The rest, as I understand, you need the timeout for it to work properly so testing may be an issue. 

Good work!

zagi25 (approved):

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-170108

PR #3982, opened 2025-04-16
MWPW-167550 [Plans] Swap places for strikethrough price

bozojovicic:
If price is displayed with promo code the strikethrough price should be displayed first. I checked milo, cc and dc code for CSS selectors like `.price-strikethrough{}` to see if we maybe have some left/right padding/margin which could be broken with this swap but I found none so I think we are safe with this change.

Resolves: [MWPW-167550](https://jira.corp.adobe.com/browse/MWPW-167550)

Test https://mwpw167550stswap--milo--bozojovicic.aem.live/drafts/bozo/pr/stswap?martech=off

**Test URLs:**
- After: https://mwpw167550stswap--milo--bozojovicic.aem.live/?martech=off


seanchoi-dev:
Few questions,

1. How about prices that as its own object without promo code?
https://main--cc--adobecom.aem.page/products/special-offers
image

2. Would RTL pages be ok with this?

npeltier (approved):
this will require quite some regression i guess

st-angelo-adobe (approved):

bozojovicic:
@seanchoi-dev 1. such prices are manually authored like that, as 2 separated prices, so GWPs will have manually re-author them
2. On Titan/Dexter we used the same order on both LTR and RTL. I asked Nick Lam about RTL on Milo and he said we can continue using the same order.

afmicka:
@bozojovicic  The order of the prices is not changed in OST, should it be? 
Screenshot 2025-04-29 at 11 05 43


Also, these changes affect all card variants, like CCD, AH.. need to check in their figmas if thats ok.

PR #3995, opened 2025-04-17
MWPW-171023: Fix seo body size check

zagi25:
* Exclude non content text

Resolves: [MWPW-171023](https://jira.corp.adobe.com/browse/MWPW-171023)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/narcis/applauncher?martech=off
- After: https://mwpw-171023-seo-body-size--milo--adobecom.aem.page/drafts/narcis/applauncher?martech=off
- Before: https://main--milo--adobecom.aem.page/drafts/ratko/mwpw-171023-seo/seo?martech=off
- After: https://mwpw-171023-seo-body-size--milo--adobecom.aem.page/drafts/ratko/mwpw-171023-seo/seo?martech=off

### DIFF
--- a/libs/blocks/preflight/panels/seo.js
+++ b/libs/blocks/preflight/panels/seo.js
@@ -112,11 +112,9 @@ async function checkDescription() {
 async function checkBody() {
   const nonContentEls = '#preflight, .picture-meta, aem-sidekick';
   const result = { ...bodyResult.value };
-  let documentInnerText = document.documentElement.innerText;
-  document.querySelectorAll(nonContentEls).forEach((el) => {
-    documentInnerText = documentInnerText.replace(el.innerText, '');
-  });
-  const { length } = documentInnerText.trim().replaceAll('\n', '');
+  const bodyClone = document.body.cloneNode(true);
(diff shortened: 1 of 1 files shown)


libs/blocks/preflight/panels/seo.js
```diff
@@ -110,8 +110,13 @@ async function checkDescription() {
 }
 
 async function checkBody() {
+  const nonContentEls = '#preflight, .picture-meta, aem-sidekick';
   const result = { ...bodyResult.value };
-  const { length } = document.documentElement.innerText;
+  let documentInnerText = document.documentElement.innerText;
```
qiyundai, line 115: This seems to work well.

I personally nit would be that this type of manipulation feels safer/cleaner performed on a clone DOM. And instead of replacing nonContentEl's innerText, we can just remove these elements to avoid any unexpected innerText behavior.

`
const clone = document.body.cloneNode(true);
clone.querySelectorAll(nonContentEls).forEach(el => el.remove());
const { length } = clone.innerText.replace(/\n/g, '').trim();
`

qiyundai (approved):
LGTM in general. Left one non-blocking nit.

robert-bogos (approved):

PR #4002, opened 2025-04-17
MWPW-165835: Do not load Milo if page is prerendered

hparra:
Do not load Milo if `<html data-prerendered="true">`.

Related: [MWPW-165835](https://jira.corp.adobe.com/browse/MWPW-165835)

**Test URLs:**
- After: https://ssr--milo--hparra.aem.page/?martech=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -1463,7 +1463,6 @@ export async function loadDeferred(area, blocks, config) {
     const { default: loadDNStatus } = await import('../features/dynamic-navigation/status.js');
     loadDNStatus();
   }
-  console.log('Done loading the deferred');
 }
 
 function initSidekick() {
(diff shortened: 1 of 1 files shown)


libs/utils/utils.js
```diff
@@ -1598,6 +1598,11 @@ async function processSection(section, config, isDoc) {
 
 export async function loadArea(area = document) {
   const isDoc = area === document;
+  const isPageLoadOk = document.getElementById('page-load-ok-milo') !== null;
+
+  if (isDoc && isPageLoadOk) {
+    return;
+  }
```
overmyheadandbody, line 1605: Since this isn't re-used and `utils` is (or at least should be) a place of brevity, how about
```suggestion
  if (isDoc && !!document.getElementById('page-load-ok-milo')) return;
```

hparra: @overmyheadandbody resolved in eca2049

overmyheadandbody (approved):

PR #4006, opened 2025-04-17
MWPW-169639: Fix SEO visibility of CAAS card collections for search engine crawlers

sanrai:
This PR addresses an SEO issue where card collections using lazy loading aren't being properly indexed by search engines. Google's crawler executes JavaScript but doesn't trigger the scroll events needed to load our lazy-loaded CAAS components at mobile viewport widths.

### Changes
- Added a isLikelyBot() function to detect search engine crawlers and headless browsers
- Modified the CAAS initialization to bypass lazy loading for search bots and small viewports
- This ensures card content is properly indexed while maintaining performance benefits of lazy loading for regular users

### Testing
- Tested with Puppeteer simulating Googlebot at 412px width
- Verified card collections load without requiring scroll events
- Confirmed regular desktop users still benefit from lazy loading

### Implementation Details
The solution leverages the existing "no-lazy" flag in the CAAS initialization logic, triggering immediate loading for search bots rather than waiting for scroll events.

Resolves: [MWPW-169639](https://jira.corp.adobe.com/browse/MWPW-169639)

**Test URLs:**
- After: https://MWPW-169639-2--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/libs/blocks/caas/caas.js
+++ b/libs/blocks/caas/caas.js
@@ -76,8 +76,22 @@ const loadCaas = async (a) => {
   initCaas(state, caasStrs, block);
 };
 
+const isSearchBot = function () {
+  const ua = navigator.userAgent || '';
+  const lower = ua.toLowerCase();
+
+  // 1. Headless / WebDriver contexts
+  if (navigator.webdriver) return true;
+  if (/headlesschrome/.test(ua)) return true;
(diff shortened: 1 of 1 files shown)


shkhan91 (approved):
+1

JackySun9 (approved):

PR #4020, opened 2025-04-22
MWPW-170901 [M@S] Incorrect WCS call in langstore

bozojovicic:
For langstore pages `getConfig()` returns the object where language, country, prefix have value "en" (or some other language) and then WCS calls have incorrect values for country and locale and prices are not loaded. Also on langstore pages the links `/tools/ost` are localized to `/en/tools/ost` due to the same issue. I did not have better ideas to prevent this than simple : if href ==`/tools/ost` -> do not localize link.

In `merch.js` I changed the code to return the default country for langstore languages. For `en` it will be `US`, for `ar` it will be `MENA`, for all other languages it will find the first locale for that language and will use its country.

**Test links**
CC :
`main--cc--adobecom.aem.page/langstore/en/products/illustrator/campaign/pricing?milolibs=mwpw170901wcs--milo--bozojovicic`
`main--cc--adobecom.aem.page/langstore/da/products/photoshop/plans?milolibs=mwpw170901wcs--milo--bozojovicic`
`main--cc--adobecom.aem.page/langstore/ar/products/illustrator/plans?milolibs=mwpw170901wcs--milo--bozojovicic`
Milo :
`mwpw170901wcs--milo--bozojovicic.aem.page/langstore/en/drafts/anu/photoshop`
`mwpw170901wcs--milo--bozojovicic.aem.page/langstore/en/drafts/mili/commerce-loc-fragment`
`mwpw170901wcs--milo--bozojovicic.aem.page/langstore/el/drafts/nala/features/commerce/commerce-loc`
`mwpw170901wcs--milo--bozojovicic.aem.page/langstore/zh-hant/drafts/nala/features/commerce/commerce-loc`
`mwpw170901wcs--milo--bozojovicic.aem.page/langstore/uk/drafts/nala/features/commerce/commerce-loc`

Resolves: [MWPW-170901](https://jira.corp.adobe.com/browse/MWPW-170901)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.live/drafts/bozo/pr/students?martech=off
- After: https://mwpw170901wcs--milo--bozojovicic.aem.live/drafts/bozo/pr/students?martech=off


libs/blocks/merch/merch.js
```diff
@@ -128,13 +128,42 @@ const GeoMap = {
...
+    return 'UA';
+  }
+  if (GeoMap[language]) {
+    return GeoMap[language].split('_', 2)[0];
+  }
+  for (const [, value] of Object.entries(GeoMap)) {
+    if (value.toLowerCase().endsWith(`_${language}`)) {
+      return value.split('_', 2)[0];
+    }
+  }
```
npeltier, line 151: so in the end you depend on the order of JSON object entries (which is not guaranteed) to return the correct language, correct?
In that case, i think i prefer a default language to locale map, setting en->us_EN, fr->fr_FR, etc...
then you can do something like `GeoMap[language] || LangDefaultMap[language.split('_', 2)[0]] `

npeltier (changes requested):

libs/blocks/merch/merch.js
bozojovicic: @npeltier  I created `LanguageMap` for mappings where language code is not equal to the country code : 
`en -> US, el -> GR, ja -> JP` etc.
For others I simply us language code as country code if it exists in `GeoMap`
There are also some strange language variants in the langstore like es-419, pt-PT that I need to split and then do the same.

npeltier (approved):
lgtm

yesil (approved):
nit: LanguageMap, GeoMap  look like class names, not sure if they are following our naming convention(whatever it is :) )

PR #4026, opened 2025-04-23
MWPW-170612 update description for stage to main

skholkhojaev:
The pr-manually-merge Workflow automatically adds the manually merged Pr in the description of `[Release] Stage to Main` and avoids the merge that happen with `github-actions[bot]`

The Merge-To-Stage Worfklow now only updates the description after the intial creation of the PR and then pr-manually-merge Workflow will take over the task for updating the description for each manually merged PRs

Resolves: [MWPW-170612](https://jira.corp.adobe.com/browse/MWPW-170612)

**Test URLs:**
- After: https://MWPW-170612-UpdateDescription-Stage-Main--milo--skholkhojaev.aem.page/?martech=off


.github/workflows/merge-to-stage.yaml
```diff
@@ -37,4 +42,4 @@ jobs:
           github-token: ${{ steps.milo-pr-merge-token.outputs.token }}
           script: |
             const main = require('./.github/workflows/merge-to-stage.js')
-            main({ github, context })
+            main({ github, context })
```
mokimo, line 45: Nit, we always have file endings

.github/workflows/merge-to-stage.js
```diff
@@ -249,4 +269,4 @@ if (process.env.LOCAL_RUN) {
   });
 }
 
-module.exports = main;
+module.exports = main;
```
mokimo: Nit, we always have file endings

.github/workflows/merge-to-stage.js
```diff
@@ -217,6 +217,26 @@ const main = async (params) => {
...
+            pull_number: stageToMainPR.number,
+            body,
+          });
+          console.log(`Updated Stage to Main PR #${stageToMainPR.number} with merged PR #${mergedPR.number}`);
+        } else {
+          await openStageToMainPR();
+        }
+      }
+      return;
+    }
```
mokimo, line 239: We already do this in this file, here: https://github.com/adobecom/milo/blob/stage/.github/workflows/merge-to-stage.js#L229-L239

The idea is to have this similar to the slack webhook you contributed and work outside of the `merge-to-stage` flow, but whenever something is manually merged as well

mokimo (changes requested):

skholkhojaev:
discussed it wtih @mokimo and we decided we should centrilize all "manual merging" workflows so now everytime a manual happens a slacknotification will be sent out + description update in one single workflow

.github/workflows/pr-manual-merge.js
```diff
@@ -0,0 +1,86 @@
...
+        console.error("Error sending Slack notification:", error.message);
+        console.log("Continuing with PR update...");
+    }
+    
+    if (['stage', 'main'].includes(base.ref)) {
+        await updateStageToMainPR(github, context, pull_request);
+    }
+}
+
+async function updateStageToMainPR(github, context, mergedPR) {
```
mokimo, line 44: How does this work together with the current process and code in merge-to-stage, does it work together?

skholkhojaev: no it shouldn't duplicate, becasue the manual-merge workflow only runs when a merge is closed, and if it detects the url of the merge in the description (line 57) it'll skip and will log that the PR is already exists

mokimo: Wouldn't there be a race condition though?
Merge-to-stage-process runs, merges a PR, doesn't "yet" update the description
this process runs, L57 isn't triggered yet

.github/workflows/pr-manual-merge.js
```diff
@@ -0,0 +1,79 @@
...
+async function main({ github, context } = {}) {
+    if (!github || !context) {
+        throw new Error("GitHub context is missing. Ensure you are running in the correct environment.");
+    }
+
+    if (process.env.LOCAL_RUN) {
+        console.log("Local run detected. Loading local configurations...");
+        const localConfigs = getLocalConfigs();
+        github = localConfigs.github;
+        context = localConfigs.context;
```
mokimo, line 12: I'm not a fan of re-assigning parameters. Also if this was an object and u'd change something within an object, you'd change the original.
IMO Either use new variables or simply pass them into main.

.github/workflows/pr-manual-merge.js
```diff
@@ -0,0 +1,79 @@
+const { slackNotification, getLocalConfigs } = require('./helpers.js');
+
+async function main({ github, context } = {}) {
```
mokimo, line 3: This _should_ work, I'm not certain though. Would only assign the values from getLocalConfigs when github/context are otherwise undefined
```suggestion
async function main({ github = getLocalConfigs().github, context = getLocalConfigs().context } = {}) {
```

.github/workflows/pr-manual-merge.js
```diff
@@ -0,0 +1,79 @@
...
+
+    try {
+        await slackNotification(
+            `${prefix} <${html_url}|#${number}: ${title}>.`,
+            process.env.MILO_RELEASE_SLACK_WH
+        );
+    } catch (error) {
+        console.error("Error sending Slack notification:", error.message);
+        console.log("Continuing with PR update...");
+    }
```
mokimo, line 37:
```suggestion
      await slackNotification(
          `${prefix} <${html_url}|#${number}: ${title}>.`,
          process.env.MILO_RELEASE_SLACK_WH
      ).catch(e => console.error("Error sending Slack notification:", e.message))
```

.github/workflows/pr-manual-merge.js
```diff
@@ -0,0 +1,79 @@
...
+    
+    if (['stage', 'main'].includes(base.ref)) {
```
mokimo, line 39: We don't want to update the description for the stage PR, if we merged something into main. It's not "part" of a batch, but already on prod.

.github/workflows/pr-manual-merge.js
```diff
@@ -0,0 +1,79 @@
...
+
+    await github.rest.pulls.update({
+      owner,
+      repo,
+      pull_number: stageToMain.number,
+      body,
+    });
+
+    console.log(`Updated Stage to Main PR #${stageToMain.number} with manually merged PR #${mergedPR.number}`);
+  } catch (error) {
```
mokimo, line 69: can something go wrong? Even if yes, then the action fails which is also fine. I'd remove the try/catch

.github/workflows/pr-manual-merge.js
```diff
@@ -0,0 +1,71 @@
...
+};
+
+async function main({ github, context } = {}) {
+    if (!github || !context) {
+        throw new Error("GitHub context is missing. Ensure you are running in the correct environment.");
+    }
+    if (process.env.LOCAL_RUN) {
+        console.log("Local run detected. Loading local configurations...");
+        const { github: localGithub, context: localContext } = getLocalConfigs();
+        return main({ github: localGithub, context: localContext });
```
mokimo, line 14: You're running into a loop here, do the params not work?

1. you call main
2. it's a local run
3. you call main, as its a local run
4. its a local run
5. you call main as its a local run
6. its a local run
7. ....

.github/workflows/pr-manual-merge.js
```diff
@@ -0,0 +1,64 @@
...
+        : ':rocket: Production release:';
+    console.log(`Sending notification for PR #${number}: ${title}`);
+    await slackNotification(
+      `${prefix} <${html_url}|#${number}: ${title}>.`,
+      process.env.MILO_RELEASE_SLACK_WH
+    ).catch(e => console.error("Error sending Slack notification:", e.message))
+    
+    if (baseRef === 'stage') {
+        await updateStageToMainPR(github, context, pull_request);
+    }
```
mokimo, line 29: Nit: I would pull this into the `updateStageToMainPR` method as `if(baseRef !== 'stage') return`

It's firstly a good strategy on avoiding nesting if/else cases for more complex methods, secondly, if any other method would call `updateStageToMainPR` - you would need to duplicate this check across all instances where a method is called.
So it's always better to put checks closer to where they belong

.github/workflows/pr-manual-merge.js
```diff
@@ -0,0 +1,64 @@
...
+  console.log("Updating PR's description");
+
+  await github.rest.pulls.update({
+    owner,
+    repo: repoName,
+    pull_number: stageToMain.number,
+    body,
+  });
+
+  console.log(`Updated Stage to Main PR #${stageToMain.number} with manually merged PR #${mergedPR.number}`);
```
mokimo, line 55:
```suggestion
  console.log(`Updated the description of the Stage to Main PR #${stageToMain.number} with manually merged PR #${mergedPR.number}`);
```

.github/workflows/pr-manual-merge.js
```diff
@@ -0,0 +1,64 @@
...
+}
+
+if (process.env.LOCAL_RUN) {
+    console.log("Local run detected. Loading local configurations...");
+    const { github, context } = getLocalConfigs();
+    main({ github, context });
```
mokimo, line 61: Do you need to even pass this in here, or just call main and use the default params? I'd probably _either_ rely on the defaults _or_ pass something in.

mokimo (approved):
Coolio, a few small nits - but this is great 🚀

mokimo:
Heya, since this PR has been open since a while - could you leave some feedback on https://github.com/orgs/adobecom/discussions/4407 why/what is blocking this from proceeding?

overmyheadandbody (approved):
This looks ok, but difficult to tell whether it covers all use-cases. Has it been tested on a fork or locally to ensure everything still works as expected?

skholkhojaev:
it has been tested on my own forked repo

PR #4028, opened 2025-04-23
MWPW-169224-Increase max PR count/batch pre&post-RCP

skholkhojaev:
Added a feature that will change the maximum number of PRs a batch can have dependig if we are currently in Pre or Post RCP.
For the period of 7 days each batch will be able to contain +3 PRs than usual

Resolves: [MWPW-169224](https://jira.corp.adobe.com/browse/MWPW-169224)

**Test URLs:**
- After: https://MWPW-169224-Increase-maxPRcount-PrePost-RCP--milo--skholkhojaev.aem.page/?martech=off


.github/workflows/helpers.js
```diff
@@ -90,6 +90,22 @@ const isWithinRCP = ({ offset = 0, excludeShortRCP = false } = {}) => {
   return false;
 };
 
+const isWithinPrePostRCP = ({ offset = 7 } = {}) => {
+  const now = new Date();
+  return RCPDates.some(({ start, end }) => {
+    if (now >= start && now <= end) {
+      return true; // return true if current time is within the RCP period
```
skholkhojaev, line 97: not sure if this this matters if it return `true` or `false` since there will be no merges happening during RCP

.github/workflows/helpers.js
```diff
@@ -90,6 +90,22 @@ const isWithinRCP = ({ offset = 0, excludeShortRCP = false } = {}) => {
   return false;
 };
 
+const isWithinPrePostRCP = ({ offset = 7 } = {}) => {
```
overmyheadandbody, line 93: I think this could be simplified. Say we have an RCP that lasts between the 14th and the 21st. I'm not sure we need to concern ourselves with what happens between those two dates, we're just interested in having a larger batch size between the 7th and the 28th. This would eliminate the first conditional and trim down the final return statement.

overmyheadandbody (commented):
Good progress! Just a few notes.

.github/workflows/merge-to-stage.js
```diff
@@ -2,13 +2,14 @@ const {
...
   isWithinRCP,
+  isWithinPrePostRCP,
   pulls: { addLabels, addFiles, getChecks, getReviews },
 } = require('./helpers.js');
 
 // Run from the root of the project for local testing: node --env-file=.env .github/workflows/merge-to-stage.js
 const PR_TITLE = '[Release] Stage to Main';
 const REQUIRED_APPROVALS = process.env.REQUIRED_APPROVALS ? Number(process.env.REQUIRED_APPROVALS) : 2;
-const MAX_MERGES = process.env.MAX_PRS_PER_BATCH ? Number(process.env.MAX_PRS_PER_BATCH) : 8;
+const BASE_MAX_MERGES = process.env.MAX_PRS_PER_BATCH ? Number(process.env.MAX_PRS_PER_BATCH) : 9;
```
mokimo, line 12: If you just add the condition up here, you don't need to change anything else in this file
```
const BASE_MAX_MERGES = process.env.MAX_PRS_PER_BATCH ? Number(process.env.MAX_PRS_PER_BATCH) : 9;
const MAX_MERGES = BASE_MAX_MERGES + (isWithinPrePostPC() ? 3 : 0)
```

.github/workflows/merge-to-stage.js
```diff
@@ -218,7 +224,7 @@ const main = async (params) => {
     existingPRCount = body.match(/https:\/\/github\.com\/adobecom\/milo\/pull\/\d+/g)?.length || 0;
     console.log(`Number of PRs already in the batch: ${existingPRCount}`);
 
-    if (mergeLimitExceeded()) return console.log(`Maximum number of '${MAX_MERGES}' PRs already merged. Stopping execution`);
+    if (mergeLimitExceeded()) return console.log('Maximum number of PRs already merged. Stopping execution');
```
mokimo, line 227: As suggested, let's revert this change... Now that we add extra logic it's also even more important to know how many PRs we can even merge - hence that logging that is pretty useful

mokimo (approved):
Small improvement suggestion, otherwise this looks good!

robert-bogos (approved):

SilviuLCF (approved):
NO QA needed

PR #4030, opened 2025-04-23
MWPW-171967 Support for placeholders-stage

chrischrischris:
When on non-prod sites, placeholder code will also look for a placeholders-stage.json and overlay those values on top of the placeholders.json values.

NOTE: This only adds support for stage placeholders for regular placeholders.  Feds and Pallet support will have to be done in another PR

**Implementation Details:**
In utils.js `decoratePlaceholders` used to fetch a single path to the placeholders file and initiate a fetch request that's passed into placeholders.js `decoratePlaceholderArea` function.  This is to start the load of the placeholders file before loading placeholders.js

Now `decoratePlaceholders` will check if on a stage env and start the fetch of both the placeholders.json and placeholders-stage.json.  Those are combined in a Promise.all  and passed to `decoratePlaceholderArea`.  Since the placeholder files get combined, only the prod placeholder path is passed in.

In placeholders.js `fetchPlaceholder` when the `placeholderRequest` is resolved we now check if the return value is an array, which means that we have 2 files that need to be combined.  Since prod is first and -stage is 2nd, when stage is processed it adds/overrides values from placeholders.json.

Resolves: [MWPW-171967](https://jira.corp.adobe.com/browse/MWPW-171967)

Pages below use the milo /placeholders.json and /placeholders-stage.json files.

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/cpeyer/test?martech=off
- After: https://stage-placeh--milo--adobecom.aem.page/drafts/cpeyer/test?martech=off

### DIFF
--- a/libs/features/placeholders.js
+++ b/libs/features/placeholders.js
@@ -11,7 +11,7 @@ const getPlaceholdersPath = (config, sheet) => {
 
 const parsePlaceholderJson = async (resp, placeholders) => {
   const json = resp.ok ? await resp.json() : { data: [] };
-  if (json.data.length === 0) return;
+  if (!json.data.length) return;
   json.data.forEach((item) => {
     window.mph[item.key] = item.value;
     placeholders[item.key] = item.value;
@@ -27,10 +27,8 @@ const fetchPlaceholder = (path, placeholderRequest) => new Promise(
     const placeholders = {};
 
     if (Array.isArray(resp)) {
-      // Process sequentially, later ones override earlier ones
-      for (const r of resp) {
-        await parsePlaceholderJson(r, placeholders);
-      }
+      // Overlay placeholders
+      for (const r of resp) await parsePlaceholderJson(r, placeholders);
     } else {
       await parsePlaceholderJson(resp, placeholders);
     }
--- a/libs/features/placeholders.js
+++ b/libs/features/placeholders.js
@@ -9,22 +9,46 @@ const getPlaceholdersPath = (config, sheet) => {
   return `${path}${query}`;
 };
 
-const fetchPlaceholders = async ({ config, sheet, placeholderRequest, placeholderPath }) => {
-  const path = placeholderPath || getPlaceholdersPath(config, sheet);
+const parsePlaceholderJson = async (resp, placeholders) => {
+  const json = resp.ok ? await resp.json() : { data: [] };
+  if (json.data.length === 0) return;
+  json.data.forEach((item) => {
+    window.mph[item.key] = item.value;
+    placeholders[item.key] = item.value;
+  });
+};
+
+const fetchPlaceholder = (path, placeholderRequest) => new Promise(
   // eslint-disable-next-line no-async-promise-executor
-  fetchedPlaceholders[path] = fetchedPlaceholders[path] || new Promise(async (resolve) => {
+  async (resolve) => {
     const resp = await placeholderRequest || await customFetch(
       { resource: path, withCacheRules: true },
     ).catch(() => ({}));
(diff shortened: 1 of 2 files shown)


rgclayton:
Adding a `.live` link to verify placeholders-stage.json is not showing up/working on production.

https://stage-placeh--milo--adobecom.aem.live/drafts/cpeyer/test

rgclayton (approved):

libs/features/placeholders.js
```diff
@@ -9,22 +9,46 @@ const getPlaceholdersPath = (config, sheet) => {
   return `${path}${query}`;
 };
 
-const fetchPlaceholders = async ({ config, sheet, placeholderRequest, placeholderPath }) => {
-  const path = placeholderPath || getPlaceholdersPath(config, sheet);
+const parsePlaceholderJson = async (resp, placeholders) => {
+  const json = resp.ok ? await resp.json() : { data: [] };
+  if (json.data.length === 0) return;
```
narcis-radu, line 14: nit: `if (!json.data.length) return;`

libs/features/placeholders.js
```diff
@@ -9,22 +9,46 @@ const getPlaceholdersPath = (config, sheet) => {
...
-    if (json.data.length === 0) { resolve({}); return; }
     const placeholders = {};
-    json.data.forEach((item) => {
-      window.mph[item.key] = item.value;
-      placeholders[item.key] = item.value;
-    });
+
+    if (Array.isArray(resp)) {
+      // Process sequentially, later ones override earlier ones
+      for (const r of resp) {
```
narcis-radu, line 31: nit: we can go without curly braces

libs/utils/utils.js
```diff
@@ -1069,17 +1069,30 @@ const findReplaceableNodes = (area) => {
 };
 
 let placeholderRequest;
+function getPlaceholderPaths(config) {
+  const rootPath = `${config.locale?.contentRoot}/placeholders`;
+  const placeholderPaths = [`${rootPath}.json`];
+  if (config.env.name !== 'prod') {
```
narcis-radu, line 1075: nit: `if (config.env.name !== 'prod') placeholderPaths.push(`${rootPath}-stage.json`);`

libs/utils/utils.js
```diff
@@ -1069,17 +1069,30 @@ const findReplaceableNodes = (area) => {
 };
 
 let placeholderRequest;
+function getPlaceholderPaths(config) {
+  const rootPath = `${config.locale?.contentRoot}/placeholders`;
+  const placeholderPaths = [`${rootPath}.json`];
```
narcis-radu, line 1074: we should add some unit testing here

narcis-radu (commented):
Looks good, would be great to have unit testing added.

libs/utils/utils.js
chrischrischris: I think this func is pretty simple and I'd prefer to not `export` the function?

narcis-radu (approved):

libs/utils/utils.js
narcis-radu: I'm fine, thanks!

vhargrave (commented):
working very nicely! 
I've noticed though that for the use case of `disable-geo-placeholders` being enabled, that the stage sheet is not fetched. 
I think we should probably handle that use case as well right? Imo it could also be a fast follow. 

I have an example of your page , but with the metadata here - https://main--milo--adobecom.aem.page/drafts/vhargrave/test

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-171967

PR #4035, opened 2025-04-24
[MWPW-170305] - Dialog attributes

DKos95:
This resolves the issue where the dialog container was missing multiple necessary accessability attributes.

Resolves: [MWPW-170305](https://jira.corp.adobe.com/browse/MWPW-170305)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/dusan/dialog-role-aria-label?martech=off
- After: https://dialog-role-title--milo--adobecom.aem.page/drafts/dusan/dialog-role-aria-label?martech=off

**CC Test URLs:**
- Before: https://main--cc--adobecom.aem.page/products/catalog
- After: https://main--cc--adobecom.aem.page/products/catalog?milolibs=dialog-role-title

### DIFF
--- a/libs/blocks/iframe/iframe.js
+++ b/libs/blocks/iframe/iframe.js
@@ -55,7 +55,7 @@ export default function init(el) {
 
     if (new URL(iframe.src).origin !== window.location.origin) {
       if (ariaLabel) iframe.title = ariaLabel;
-      if (dialogModal) dialogModal.setAttribute('aria-label', ariaLabel || iframe.title);
+      dialogModal?.setAttribute('aria-label', iframe.title);
       return;
     }
 
--- a/libs/blocks/iframe/iframe.js
+++ b/libs/blocks/iframe/iframe.js
@@ -60,8 +60,8 @@ export default function init(el) {
     }
 
     const sameOriginText = ariaLabel || iframe.contentWindow.document.querySelector('h1, h2, h3, h4, h5, h6')?.textContent;
-    iframe.title = sameOriginText;
-    if (dialogModal) dialogModal.title = sameOriginText;
+    if (sameOriginText) iframe.title = sameOriginText;
+    if (dialogModal && sameOriginText) dialogModal.title = sameOriginText;
   };
 
   el.insertAdjacentElement('afterend', embed);
--- a/libs/blocks/iframe/iframe.js
+++ b/libs/blocks/iframe/iframe.js
@@ -51,12 +51,17 @@ export default function init(el) {
   const embed = createTag('div', { class: `milo-iframe ${classes}` }, iframe);
 
   iframe.onload = () => {
+    const dialogModal = iframe.closest('.dialog-modal');
+
     if (new URL(iframe.src).origin !== window.location.origin) {
       if (ariaLabel) iframe.title = ariaLabel;
+      if (dialogModal) dialogModal.title = ariaLabel || iframe.title;
       return;
     }
 
-    iframe.title = ariaLabel || iframe.contentWindow.document.querySelector('h1, h2, h3, h4, h5, h6')?.textContent;
+    const sameOriginText = ariaLabel || iframe.contentWindow.document.querySelector('h1, h2, h3, h4, h5, h6')?.textContent;
+    iframe.title = sameOriginText;
+    if (dialogModal) dialogModal.title = sameOriginText;
   };
 
   el.insertAdjacentElement('afterend', embed);
--- a/libs/blocks/iframe/iframe.js
+++ b/libs/blocks/iframe/iframe.js
@@ -51,17 +51,20 @@ export default function init(el) {
   const embed = createTag('div', { class: `milo-iframe ${classes}` }, iframe);
 
   iframe.onload = () => {
-    const dialogModal = iframe.closest('.dialog-modal');
+    try {
+      const dialogModal = iframe.closest('.dialog-modal');
+      if (new URL(iframe.src).origin !== window.location.origin) {
+        if (ariaLabel) iframe.title = ariaLabel;
+        dialogModal?.setAttribute('aria-label', iframe.title);
+        return;
+      }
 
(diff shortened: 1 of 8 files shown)


libs/blocks/vimeo/vimeo.js
```diff
@@ -79,7 +82,7 @@ class LiteVimeo extends HTMLElement {
 }
 
 export default async function init(a) {
-  if (isInTextNode(a)) return;
+  if (isInTextNode(a) || !a) return;
```
overmyheadandbody, line 85: Would it be better to first check for `a` before applying a function to it?
```suggestion
  if (!a || isInTextNode(a)) return;
```

libs/utils/utils.js
```diff
@@ -754,6 +754,13 @@ export function decorateImageLinks(el) {
   });
 }
 
+export const setDialogAndIframeTitle = ({ element, title }) => {
+  const dialogModal = element.closest('.dialog-modal');
```
overmyheadandbody, line 758: Let's play it safe
```suggestion
  const dialogModal = element?.closest('.dialog-modal');
```

libs/utils/utils.js
```diff
@@ -754,6 +754,13 @@ export function decorateImageLinks(el) {
   });
 }
 
+export const setDialogAndIframeTitle = ({ element, title }) => {
```
overmyheadandbody, line 757: It might be that the naming is misleading. You're calling the function with an `element` that is an `iframe`, but that's not reflected here. The method can be called for any element, right? I'd suggest adapting the naming. Additionally, is there any other place where this could live? `utils` is used by everyone, so we try to limit its functionality to the core needs of all consumers.

libs/utils/utils.js
```diff
@@ -754,6 +754,13 @@ export function decorateImageLinks(el) {
...
+  const dialogModal = element.closest('.dialog-modal');
+
+  if (element && title) element.title = title;
+  if (dialogModal && title) dialogModal.setAttribute('aria-label', title);
```
overmyheadandbody, line 761: How about this to keep it short:
```suggestion
  if (!element || !title) return;
  element.title = title;
  element.closest('.dialog-modal')?.setAttribute('aria-label', title);
```

overmyheadandbody (commented):
Looks good, but I do have a few small notes

libs/utils/utils.js
DKos95: @overmyheadandbody I don't see a shared file for the video files, we could create a util file for youtube, vimeo and adobetv, don't see another option to be honest.

libs/utils/utils.js
```diff
@@ -754,6 +754,12 @@ export function decorateImageLinks(el) {
   });
 }
 
+export const setDialogAndElementAttributes = ({ element, title }) => {
```
overmyheadandbody, line 757: I remember @zagi25 started an `accessibility.js` file, but I can't find it right now. Was I dreaming? 😅 If that file were to exist, this method would fit perfectly in there.

overmyheadandbody (approved):
Looking good! Just one more comment on the location of the new shared method, maybe you can reach out to @zagi25 and check if that file is still a possibility. If not, we can go with this as-is

libs/utils/utils.js
zagi25: File is not there yet because PR is not merged https://github.com/adobecom/milo/pull/3898. I agree that this will perfectly fit into that file, I would just import it after request is done e.g.
```
      const response = await fetch(`https://www.youtube.com/oembed?url=http://www.youtube.com/watch?v=${this.videoId}&format=json`);
      const data = await response.json();
      const { setDialogAndElementAttributes } = await import ('../../accessibility.js');
     setDialogAndElementAttributes({element: this.iframeEl, title: data.title});
```
Same for vimeo and adobetv
cc: @DKos95

aishwaryamathuria (approved):

libs/scripts/accessibility.js
```diff
@@ -0,0 +1,8 @@
+export const setDialogAndElementAttributes = ({ element, title }) => {
+  if (!element || !title) return;
+  element.title = title;
+  element.closest('.dialog-modal')?.setAttribute('aria-label', title);
+};
+
+export default function init() {
+}
```
overmyheadandbody, line 8: I'm guessing this is in preparation for @zagi25's changes? A bit odd at the moment to have this empty 🙂

libs/blocks/iframe/iframe.js
```diff
@@ -51,12 +51,17 @@ export default function init(el) {
...
     if (new URL(iframe.src).origin !== window.location.origin) {
       if (ariaLabel) iframe.title = ariaLabel;
+      dialogModal?.setAttribute('aria-label', iframe.title);
       return;
     }
 
-    iframe.title = ariaLabel || iframe.contentWindow.document.querySelector('h1, h2, h3, h4, h5, h6')?.textContent;
+    const sameOriginText = ariaLabel || iframe.contentWindow.document.querySelector('h1, h2, h3, h4, h5, h6')?.textContent;
+    if (sameOriginText) iframe.title = sameOriginText;
+    if (dialogModal && sameOriginText) dialogModal.setAttribute('aria-label', sameOriginText);
```
overmyheadandbody, line 64: In `modal.js`, these sort of checks are wrapped in a `try/catch`. Should this be done here as well to avoid the dreaded `Uncaught SecurityError` being logged in the console?

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-170305

libs/scripts/accessibility.js
DKos95: Yeah the thing is that if this isn't the default one then I would have to call my function in a different manner where it is being called, so when Ratko makes his changes he would have to change my code also, this just seemed more straightforward and future proof.

PR #4037, opened 2025-04-24
chore: Configure Trusted Hosts for Sidekick

andreituicu:
After: https://trusted-hosts--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/tools/sidekick/config.json
+++ b/tools/sidekick/config.json
@@ -1,6 +1,7 @@
 {
   "project": "Milo",
   "host": "milo.adobe.com",
+  "trustedHosts": ["milo.adobe.com", "*--milo--adobecom.aem.page"],
   "libraries": [
     {
       "text": "Blocks",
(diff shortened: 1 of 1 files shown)


tools/sidekick/config.json
```diff
@@ -1,6 +1,7 @@
 {
   "project": "Milo",
   "host": "milo.adobe.com",
+  "trustedHosts": ["milo.adobe.com", "*--milo--adobecom.aem.page"],
```
narcis-radu, line 4: @andreituicu, we need `*--milo--adobecom.aem.live` as well to support https://github.com/adobecom/milo/blob/stage/tools/sidekick/config.json#L108-L113.

andreituicu: Sure! Normally, relative paths should work OOTB, because the current site is always a trusted host, but if this repo serves as some kind of template, then it makes sense to have this too.

narcis-radu (approved):

PR #4044, opened 2025-04-25
MWPW-170615 [Sev2][Catalog] Group of checkboxes not associated with …

bozojovicic:
…group label - type

Fix for a11y issue on catalog page for Types(Desktop/Mobile/Web) checkbox group
ARIA group/labelledby technique needs to be applied, which was implemented before
but a bug is introduced in the meanwhile, which is fixed now.

Resolves: [MWPW-170615](https://jira.corp.adobe.com/browse/MWPW-170615)

Test on `stage--cc--adobecom.hlx.page/products/catalog?milolibs=mwpw170615a11y--milo--bozojovicic`

**Test URLs:**
- After: https://mwpw170615a11y--milo--bozojovicic.aem.live/?martech=off


libs/features/mas/src/sidenav/merch-sidenav-checkbox-group.js
```diff
@@ -53,7 +53,7 @@ export class MerchSidenavCheckboxGroup extends LitElement {
         this.prepend(h3El);
 
         [...this.children].forEach((el) => {
-          if (el.id && el.id !== id) {
+          if (!el.id) {
```
yesil, line 56: I think, we need to group them together rather than having a separate group per checkbox

image

From GPT  :)
```
From an accessibility (a11y) standpoint, a single checkbox should not have role="group". Here's why:
role="group" is used to group multiple related elements together under a shared label (for example, a group of checkboxes for "Preferred Contact Methods": Email, Phone, SMS).
```

bozojovicic: You are absolutely right. And that's also how they described it but I was blind.
```
<p id="contact-label">Preferred contact method</p>
<div role="group" aria-labelledby="contact-label">
  <label for="email">Email</label><input type="radio" id="email" name="contact">
  <label for="phone">Phone</label><input type="radio" id="phone" name="contact">
</div>
```
Fixed. We already have a group so I added `role="group" aria-labelledby="sidenav-checkbox-group-title"` only there in the root element.

seanchoi-dev (approved):
lgtm

mirafedas (approved):
@bozojovicic lgtm, please resolve conflicts :)

yesil (approved):

PR #4054, opened 2025-04-28
[MWPW-171327] Table sticky heading

DKos95:
This resolves the issue where table content wasn't able to be seen from the sticky table heading, the solution is to turn off the sticky functionality if the heading occupies a curtain percentage of the users viewport height.

Resolves: [MWPW-171327](https://jira.corp.adobe.com/browse/MWPW-171327)

**Test URLs:**
- After: https://table-sticky--milo--adobecom.aem.page/docs/library/kitchen-sink/table?martech=off

**CC Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/cc-shared/fragments/merch/products/lightroom/compare-plans/table/individual/default
- After: https://main--cc--adobecom.hlx.page/cc-shared/fragments/merch/products/lightroom/compare-plans/table/individual/default?milolibs=table-sticky


libs/blocks/table/table.js
```diff
@@ -461,6 +461,14 @@ async function handleScrollEffect(table) {
   const highlightRow = table.querySelector('.row-highlight');
   const headingRow = table.querySelector('.row-heading');
 
+  setTimeout(() => {
```
robert-bogos, line 464: Why the use of setTimeout() in here? 🤔

DKos95: Without the timeout the heading height returns 0, I tried on multiple places to see if I can get it without a timeout but with no success.

libs/blocks/table/table.js
```diff
@@ -461,6 +461,14 @@ async function handleScrollEffect(table) {
...
+    if (!isStickyHeader(table)) {
+      return;
+    }
+
+    if (headingRow.offsetHeight / window.innerHeight >= 0.3) table.classList.add('cancel-sticky');
+  }, 0);
```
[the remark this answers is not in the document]
DKos95: headingRow.offsetHeight returns 0 without a timeout, tried to get it to work without one in multiple places, but to no success.

libs/blocks/table/table.js
```diff
@@ -657,10 +666,12 @@ export default function init(el) {
       setTooltipPosition(el);
     };
     handleResize();
+    handleStickyHeader(el);
```
overmyheadandbody, line 669: Since resizing alters the behavior, this would be a good fit for the already existing `handleResize` method. Moving it there also removes the need to re-call this on L674, since `handleResize` is already called there.

overmyheadandbody (commented):
Looks good, just a nit.

libs/blocks/table/table.js
DKos95: I had this idea also but it depends on the outcome that we want. If we want someone to be able to come to the page and to zoom in and for the sticky table heading to not be sticky anymore we need to call it like this because handleResize doesn't get triggered when you zoom in all the time because of this clause in the resize listener `if (deviceBySize === defineDeviceByScreenSize()) return;`.

DKos95: For example when testing on the table kitchen sink on 1920x1080 and zooming in 3 times to 150% the heading height is at the cutoff spot but handleResize doesn't get triggered and the heading stays sticky.

overmyheadandbody: Oh, I see, then this is a-ok, thanks for the explanation.

libs/blocks/table/table.js
```diff
@@ -598,6 +598,15 @@ function applyStylesBasedOnScreenSize(table, originTable) {
   setRowStyle();
 }
 
+function handleStickyHeader(el) {
+  if (!el.classList.value.includes('sticky')) return;
+
+  setTimeout(() => {
+    const headingRow = el.querySelector('.row-heading');
+    if (headingRow.offsetHeight / window.innerHeight >= 0.3) el.classList.add('cancel-sticky');
```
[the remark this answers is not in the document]
DKos95: Agreed on both points!

libs/blocks/table/table.js
```diff
@@ -598,6 +598,16 @@ function applyStylesBasedOnScreenSize(table, originTable) {
   setRowStyle();
 }
 
+function handleStickyHeader(el) {
+  if (!el.classList.value.includes('sticky')) return;
+
+  setTimeout(() => {
+    const headingRow = el.querySelector('.row-heading');
+    if (headingRow.offsetHeight / window.innerHeight < 0.45) el.classList.remove('cancel-sticky');
+    if (headingRow.offsetHeight / window.innerHeight >= 0.45) el.classList.add('cancel-sticky');
```
overmyheadandbody, line 607: Micro-nit, how about
```suggestion
    el.classList.toggle('cancel-sticky', !(headingRow.offsetHeight / window.innerHeight < 0.45));
```

zagi25 (approved):

SilviuLCF (approved):
Verified , testing details[ ](https://jira.corp.adobe.com/browse/MWPW-171327)

PR #4056, opened 2025-04-28
MWPW-171000 [MMM][MEP] Send updates to MMM without MEP button

vgoodric:
The primary goal of this PR is to make the MMM data more accurate. To achieve this, we only save on prod pages.  We will still always save if the MEP button is being used because we assume they are an internal user. However, we will also sample a minimal sample rate of every user on a MEP-enabled page (1 in 1,000 users).  To achieve this:
* The saveToMmm function is now called in utils.js instead of personalization.js because if we are impacting real users, we will now call this function postLCP. The current implementation is internal user only so we called it pre-LCP
* saveToMmm now wrapped in a req to be prod or a random number chosen user
* 2 functions (saveToMmm, parseMepConfig) from preview.js are now in a new file: mmm.js, so users saving to MMM only need to download the code needed for saving
* deleted parsePageAndUrl function because we will no longer try to figure out the prod equivalent URL.
* added localhost as an excluded string to not save
* to make QA possible, if you spoof a prod environment, it still enters the code to save, but left in the code that checks for exclusion strings in the URL. But I also moved the code lower and output the object we would have saved to console. 

Resolves: [MWPW-171000](https://jira.corp.adobe.com/browse/MWPW-171000)

QA instructions:
1. Git is kind of terrible with comparing what I've done to the moved functions. I recommend using something like https://text-compare.com/ and comparing the 2 moved functions separately: saveToMmm, parseMepConfig
2. It's also difficult because we want to test this before we put it on prod, but one of the changes is to only save on prod.  That's why the QA link has the prod spoof.  But we still don't want to clutter our database.  So I've added an output to the console so we can see what would have been sent without impacting the database. So below when you're checking to see "if it saved", you're looking for console output that it only did not save because you are on a rejected URL. To check if you saved in before links you can check the network call
3. Because the sample rate is so small, if you want to test if it saves when you qualify by RNG, you can override utils line 1405 and change the sample rate for the recommended scenarios

**Test URLs (when I say on prod I also mean spoofing prod):**
1. Does it always save on prod when there is a MEP button? (good to keep the low sample rate)
- Before: https://main--cc--adobecom.aem.live/products/illustrator?env=prod&mep
- After: https://main--cc--adobecom.aem.live/products/illustrator?env=prod&mep&milolibs=mmmsend
2. Does it save when on prod without the MEP button when you qualify for the sample rate? (good to change to .5 sample rate)
- Before: https://main--cc--adobecom.aem.live/products/illustrator?env=prod
- After: https://main--cc--adobecom.aem.live/products/illustrator?env=prod&milolibs=mmmsend
3. Does it never save when there is no MEP? (good to change to .99 sample rate)
- Before: https://main--cc--adobecom.aem.live/products/illustrator?mep=off
- After: https://main--cc--adobecom.aem.live/products/illustrator?mep=off&milolibs=mmmsend
2. Does it never save when not on prod? (good to change to .99 sample rate)
- Before: https://main--cc--adobecom.aem.live/products/illustrator?milolibs=mmmsend
- After: https://main--cc--adobecom.aem.live/products/illustrator

- Psi-check: https://mmmsend--milo--adobecom.aem.page/?martech=off


ivanvatadobe (approved):
Test cases pass for me. 
https://jira.corp.adobe.com/browse/MWPW-172543

PR #4057, opened 2025-04-29
MWPW-171967 Handle non-json

chrischrischris:
Also will not attempt to fetch placeholders-stage.json unless `placeholders-stage` metadata is set to `on`

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/cpeyer/test?martech=off
- After: https://171967-stage-pl-fix--milo--adobecom.aem.page/drafts/cpeyer/test?martech=off

### DIFF
--- a/libs/features/placeholders.js
+++ b/libs/features/placeholders.js
@@ -12,13 +12,12 @@ const getPlaceholdersPath = (config, sheet) => {
 const parsePlaceholderJson = async (resp, placeholders) => {
   try {
     const json = resp.ok ? await resp.json() : { data: [] };
-    if (!json.data.length) return;
-    json.data.forEach((item) => {
+    json.data?.forEach((item) => {
       window.mph[item.key] = item.value;
       placeholders[item.key] = item.value;
     });
   } catch (e) {
-    console.log('Error parsing placeholder json', e);
+    window.lana.log(`Error parsing placeholder json: ${e.message}`, { tags: 'placeholders', errorType: 'e' });
   }
 };
 
--- a/libs/features/placeholders.js
+++ b/libs/features/placeholders.js
@@ -10,12 +10,16 @@ const getPlaceholdersPath = (config, sheet) => {
 };
 
 const parsePlaceholderJson = async (resp, placeholders) => {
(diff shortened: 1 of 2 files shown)


seanchoi-dev (approved):

libs/features/placeholders.js
```diff
@@ -10,12 +10,16 @@ const getPlaceholdersPath = (config, sheet) => {
...
-  });
+  try {
+    const json = resp.ok ? await resp.json() : { data: [] };
+    if (!json.data.length) return;
+    json.data.forEach((item) => {
+      window.mph[item.key] = item.value;
+      placeholders[item.key] = item.value;
+    });
+  } catch (e) {
+    console.log('Error parsing placeholder json', e);
```
narcis-radu, line 21: Can we send errors to lana instead of console?

libs/features/placeholders.js
```diff
@@ -10,12 +10,16 @@ const getPlaceholdersPath = (config, sheet) => {
...
-  const json = resp.ok ? await resp.json() : { data: [] };
-  if (!json.data.length) return;
-  json.data.forEach((item) => {
-    window.mph[item.key] = item.value;
-    placeholders[item.key] = item.value;
-  });
+  try {
+    const json = resp.ok ? await resp.json() : { data: [] };
+    if (!json.data.length) return;
+    json.data.forEach((item) => {
```
overmyheadandbody, line 16: Tiny nit, since this file is used by everyone:
```suggestion
    json.data?.forEach((item) => {
```
If `json.data` is an empty array, then `forEach` wouldn't be executed. Saves one line of code, but not the end of the world if it stays like this either.

overmyheadandbody (commented):
Agree with @narcis-radu on substituting the console statement, otherwise looking good. Has the strategy of using this new metadata field been discussed with GWP, are they aware of this requirement moving forward?

sukamat (approved):

overmyheadandbody (approved):

narcis-radu (approved):

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-171967

PR #4059, opened 2025-04-29
[MWPW-171958] - video pause/play a11y

DKos95:
This resolves the issue where an aria-label was missing from the pause/play button.

Resolves: [MWPW-171958](https://jira.corp.adobe.com/browse/MWPW-171958)

**Test URLs:**
- After: https://video-element-name-missing--milo--adobecom.aem.page/?martech=off

**CC Test URLs:**
- Before: https://main--cc--adobecom.hlx.page/cc-shared/fragments/products/substance3d/painter/aside-non-destructive
- After: https://main--cc--adobecom.hlx.page/cc-shared/fragments/products/substance3d/painter/aside-non-destructive?milolibs=video-element-name-missing

### DIFF
--- a/libs/utils/decorate.js
+++ b/libs/utils/decorate.js
@@ -278,7 +278,7 @@ export function addAccessibilityControl(videoString, videoAttrs, indexOfVideo, t
   }
   return `
     <div class='video-container video-holder'>${videoString}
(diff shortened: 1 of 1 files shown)


narcis-radu (approved):

robert-bogos (approved):

overmyheadandbody (approved):
I'm not certain we need both `title` and `aria-label`, but I checked and there's no announcement duplication, so this should be fine.

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-171958

PR #4088, opened 2025-05-07
MWPW-172916 Pep prompt is seen loaded and redirected when geo popup is on and not closed

sharmrj:
* Sometimes georouting loads after PEP starts loading, which causes the georouting modal to show up on top of a running pep dialogue. 
* This was technically introduced when we stopped awaiting Georouting in `loadPostLCP` but the bug did not manifest until the global navigation was optimized to the point that it became likely that PEP would load before Georouting was finished
* The solution is to save the promise for loading georouting and await it inside pep.
* The promise is cleaned up with `delete` since it's not used anywhere else. 

Resolves: [MWPW-172816](https://jira.corp.adobe.com/browse/MWPW-172916)

**Test URLs:**
- After: https://georouting-pep--milo--adobecom.aem.page/?martech=off

**QA**: https://main--cc--adobecom.hlx.page/creativecloud/animation/testdoc/pepphsp1?skipPepEntitlements=true&lanadebug&georouting=on&milolibs=georouting-pep


libs/features/webapp-prompt/webapp-prompt.js
```diff
@@ -367,6 +367,9 @@ export class AppPrompt {
 
 export default async function init(config) {
   try {
+    const miloConfig = getConfig();
+    await miloConfig.georouting.loadedPromise;
+    delete miloConfig.georouting;
     const appPrompt = new AppPrompt(config);
```
bandana147, line 373: This would fail if georouting is off since the promise is initialized only if georouting is on.

sharmrj: Good catch. It's fixed now.

bandana147 (approved):

spadmasa:
Validation done on the pr on below urls
https://main--cc--adobecom.hlx.page/creativecloud/animation/testdoc/pepphsp1?skipPepEntitlements=true&lanadebug&georouting=on&milolibs=georouting-pep
Pep modal is seen loading only when the geo modal is closed
image
image
image


https://main--cc--adobecom.hlx.page/creativecloud/animation/testdoc/pepphsp?skipPepEntitlements=true&lanadebug&georouting=off&milolibs=georouting-pep
image
image

detailed results in https://jira.corp.adobe.com/browse/MWPW-172916

PR #3592, opened 2025-01-30
Fix(Mwpw-165662):text container mover out of bounds

sharath-kannan:
* display changed to block for mobile screens.
Resolves: [MWPW-165662](https://jira.corp.adobe.com/browse/MWPW-165662)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/sharathkannan/rtl?martech=off
- After: https://mwpw-165662--milosh--sharath-kannan.hlx.page/drafts/sharathkannan/rtl?martech=off

PR #4097, opened 2025-05-07
[MWPW-170714] - Section metadata grid

DKos95:
This resolves the issue where when on mobile the spaces between the cards is missing, and the section metadata two-up from the resolution 1070px to 1199px ended up making 3 items instead of 2 per row.

Resolves: [MWPW-170714](https://jira.corp.adobe.com/browse/MWPW-170714)

**PR Note:**
#Example2 looks to be intended functionality, which is explained in the ticket comment https://jira.corp.adobe.com/browse/MWPW-170714#comment-47764490, will need more discussion with Consonant and won't be addressed in this PR, while #Example1 will be addressed.

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/klee/jira-ticket/mwpw-170714-section-metadata-gap-ups?martech=off
- After: https://section-metadata-grid--milo--adobecom.aem.page/drafts/klee/jira-ticket/mwpw-170714-section-metadata-gap-ups?martech=off

**Test URLs for the previous solution that caused this issue:**
- Before: https://main--milo--adobecom.aem.page/drafts/sharathkannan/rtl?martech=off
- After: https://section-metadata-grid--milo--adobecom.aem.page/drafts/sharathkannan/rtl?martech=off

### DIFF
--- a/libs/blocks/section-metadata/section-metadata.css
+++ b/libs/blocks/section-metadata/section-metadata.css
@@ -319,11 +319,13 @@ main > .section[class*='-up'] > .content {
   }
 }
 
-@media screen and (min-width: 1200px) {
+@media screen and (min-width: 1070px) {
   .section.two-up {
     grid-template-columns: repeat(2, minmax(0, 1fr));
   }
+}
 
+@media screen and (min-width: 1200px) {
   .section.three-up {
     grid-template-columns: repeat(3, minmax(0, 1fr));
   }
--- a/libs/blocks/section-metadata/section-metadata.css
+++ b/libs/blocks/section-metadata/section-metadata.css
@@ -269,6 +269,9 @@ main > .section[class*='-up'] > .content {
     display: block;
   }
 
+  .section[class*='grid-width-'] > *:not(:first-child) {
+    margin-top: var(--spacing-m);
+  }
 }
 
 @media screen and (min-width: 600px) and (max-width: 1200px) {
--- a/libs/blocks/section-metadata/section-metadata.css
+++ b/libs/blocks/section-metadata/section-metadata.css
@@ -266,11 +266,8 @@ main > .section[class*='-up'] > .content {
   }
 
   .section[class*='grid-width-'] {
-    display: block;
-  }
-
-  .section[class*='grid-width-'] > *:not(:first-child) {
-    margin-top: var(--spacing-m);
+    display: flex;
+    flex-direction: column;
   }
 }
 
@@ -326,6 +323,10 @@ main > .section[class*='-up'] > .content {
 }
 
 @media screen and (min-width: 1200px) {
+  .section.two-up {
+    grid-template-columns: repeat(2, minmax(0, 1fr));
+  }
+
   .section.three-up {
     grid-template-columns: repeat(3, minmax(0, 1fr));
   }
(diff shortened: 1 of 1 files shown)


libs/blocks/section-metadata/section-metadata.css
```diff
@@ -316,11 +319,13 @@ main > .section[class*='-up'] > .content {
   }
 }
 
-@media screen and (min-width: 1200px) {
+@media screen and (min-width: 1070px) {
```
overmyheadandbody, line 322: I'm not confident in this change. This is the only such media query in our code, so there must be something else at play here. A bit more investigation might be needed here - looking at the code, it seems that the `.#-up` pattern was only supposed to work on desktop, so it could be that the request from the ticket is not a bug, it's an enhancement that should come as a new requirement and get Consonant approval. Can you look through older PRs to check when the `.#-up` pattern was introduced and which request it came from?

libs/blocks/section-metadata/section-metadata.css
```diff
@@ -269,6 +269,9 @@ main > .section[class*='-up'] > .content {
     display: block;
   }
 
+  .section[class*='grid-width-'] > *:not(:first-child) {
+    margin-top: var(--spacing-m);
+  }
```
overmyheadandbody, line 274: This seems straightforward enough, but it fixes other suboptimal code. As mentioned in the story, the issue was introduced by https://github.com/adobecom/milo/pull/3592/files, which completely removes the grid settings, in turn removing the gap. That fix is likely wrong. Instead of switching the `display` value, we could use something like `justify-content: center`. Here's the original page for that previous issue, if you want to test - https://main--milo--adobecom.hlx.page/drafts/sharathkannan/rtl?martech=off

overmyheadandbody (commented):
Although these fixes do address the requirements, the issue seems to come from previous faulty code and feature misalignment. I provided a couple of comments to track these down and make sure we implement a more sustainable approach.

libs/blocks/section-metadata/section-metadata.css
DKos95: `justify-content: center` makes it look kind of stretched on mobile. I've found that `display:flex;flex-direction:column;` makes it look more like `display:block;`.

libs/blocks/section-metadata/section-metadata.css
```diff
@@ -266,9 +266,9 @@ main > .section[class*='-up'] > .content {
   }
 
   .section[class*='grid-width-'] {
```
zagi25, line 268: This works, but wouldn't removing this rule ( `display: block` ) have the same effect ? Because `display: grid` is defined for wider screens ?

DKos95: That would the issue for this ticket, but would cause issues for the previous tickets solution that caused this issue.
On mobile it would be scewed to one side like said in their ticket https://section-metadata-grid--milo--adobecom.aem.page/drafts/sharathkannan/rtl?martech=off.

DKos95: Talking about the issue mentioned in this ticket to be the cause of this. https://jira.corp.adobe.com/browse/MWPW-165662

zagi25 (approved):

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-170714

PR #4103, opened 2025-05-08
Improve carousel accessibility

zagi25:
Issues: Missing slide content announcement when navigating carousel and select slide indicator missing correct state. 
* Added `aria-live` [docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-live)
* Update `aria-live` with slide text content, so it gets announced to the user
* Set correct `aria-label` attributes to the slide indicators
* Set `aria-current` attribute to the active indicator

Resolves: [MWPW-171960](https://jira.corp.adobe.com/browse/MWPW-171960) and [MWPW-172184](https://jira.corp.adobe.com/browse/MWPW-172184)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/ratko/mwpw-171960-carousel-dots/carousel?martech=off
- After: https://accessibility-carousel--milo--adobecom.aem.page/drafts/ratko/mwpw-171960-carousel-dots/carousel?martech=off

### DIFF
--- a/libs/blocks/carousel/carousel.js
+++ b/libs/blocks/carousel/carousel.js
@@ -67,7 +67,7 @@ function decorateLightboxButtons() {
   return [expandBtn, closeBtn];
 }
 
-function decorateSlideIndicators(slides, jumpTo) {
+function decorateSlideIndicators(slides) {
   const indicatorDots = [];
 
   for (let i = 0; i < slides.length; i += 1) {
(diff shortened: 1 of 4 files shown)


JasonHowellSlavin (approved):

libs/blocks/carousel/carousel.js
```diff
@@ -179,6 +178,13 @@ function setIndicatorMultiplyer(carouselElements, activeSlideIndicator, event) {
   }
 }
 
+function updateAriaLive(ariaLive, slide) {
```
DKos95, line 181: NIT: Does this really need to be a function?

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-172184

PR #4106, opened 2025-05-08
[MWPW-170341] - table bold text break fix

DKos95:
This fixes the issue where bold text breaks into new line.

Resolves: [MWPW-170341](https://jira.corp.adobe.com/browse/MWPW-170341)

**Test URLs:**
- After: https://table-bold-break--milo--adobecom.aem.page/docs/library/kitchen-sink/table?martech=off

**CC Test URLs:**
- Before: https://main--cc--adobecom.aem.page/drafts/vmorel/table-bold-link-break
- After: https://main--cc--adobecom.aem.page/drafts/vmorel/table-bold-link-break?milolibs=table-bold-break


DKos95:
> Just fix line to be < 100 length.

Hey Chris, thanks for the initial review, I appreciate it. I have however since changed the logic up a bit since I did see some issues in the kitchen sink example, for this reason I will be re requesting a review, thanks again!

zagi25:
Maybe this can be fixed using CSS only, what if
```
.table .col:has(> strong) {
  display: block;
}
```
this should target divs that have only plain text and strong, it shouldn't mess up the other cases but I am not 100% sure.

libs/blocks/table/table.js
```diff
@@ -368,6 +368,15 @@ function handleSection(sectionParams) {
       sectionRowTitle.setAttribute('scope', 'row');
     }
   }
+
+  rowCols.forEach((col) => {
+    const strong = col.querySelector('strong');
+    if (strong && Array.from(col.childNodes).some(
+      (node) => node.nodeType === Node.TEXT_NODE && node.textContent.trim()
+      && node.parentNode === col && strong.parentNode.tagName === 'DIV',
+    )) col.replaceChildren(createTag('p', {}, [...col.childNodes]));
```
overmyheadandbody, line 377: Do we really need all these checks? EDS returns a `div` if there's a single line of text (`<div>My authored text</div>`) or `p` for multiple lines of text (`<p>My 1st authored text</p><p>My 2nd authored text</p>`). The latter is not an issue for our use-case. The former is, should the text contain additional tags, such as `strong`, `em` (not part of the story, but a possibility), etc. So to simplify the checks and cover all use-cases, what if we wrap the content in a `p` if `col` does have a child, but it is not a `p` or `.icon` already?

overmyheadandbody (commented):
I have a suggestion that could reduce checks and extend this beyond `strong` alone, might be worth checking out

overmyheadandbody:
@zagi25 - I had the same thought initially, but there are exceptions, like:
Screenshot 2025-05-09 at 12 15 57

I tried a few other CSS-only ideas, but didn't get to a perfect solution. Maybe there is one, I'd be happy if we could solve it that way, otherwise a JS approach might be needed.

Also, tables are getting a redesign soon, the code we have is beyond patching in my opinion, I'd love to get bandwidth to re-write it entirely, if possible.

libs/blocks/table/table.js
DKos95: The thing is that we want to create a wrapper only if there is strong tag and other text next to it and that they aren't already in a p, because then the break will occur. For example the first table in the table block kitchen sink has strong in the heading but it doesn't have text next to it but a picture so we don't need to apply it there, so we need to also check the surroundings of the strong element, if we were to apply the p wrapper in that situation also then that first example in the kitchen sink wouldn't look the same, the text would be next to the picture.

DKos95: Updated PR to expand the tags that are queried as indeed they are causing the same kind of problem as bold is.

libs/blocks/table/table.js
```diff
@@ -368,6 +374,15 @@ function handleSection(sectionParams) {
       sectionRowTitle.setAttribute('scope', 'row');
     }
   }
+
+  rowCols.forEach((col) => {
+    const formattingTags = col.querySelector('strong, em, mark, b, i, u, del, ins, code, small, sub, sup, cite');
+    if (formattingTags && Array.from(col.childNodes).some(
```
zagi25, line 380: This works well and I understand why this needs to be done. But to add on @overmyheadandbody comment, could we maybe simplify this check by using `children` instead `childNodes` ? 
```
const textModifiers = ['STRONG', 'EM', 'MARK', 'B', 'I', 'U', 'DEL', 'INS', 'CODE', 'SMALL', 'SUB', 'SUP', CITE']
if (Array.from(col.children).every((child) => textModifiers.includes(child.tagName)) {
    col.replaceChildren(createTag('p', {}, [...col.childNodes]));
}
```
This way, `p` wrapper is created only if div contains 'text modifier' tags as children. This should work also for `strong` + `image` case that you mentioned, right ?

libs/blocks/table/table.js
```diff
@@ -368,6 +368,14 @@ function handleSection(sectionParams) {
       sectionRowTitle.setAttribute('scope', 'row');
     }
   }
+
+  rowCols.forEach((col) => {
+    const textModifiers = ['STRONG', 'EM', 'MARK', 'B', 'I', 'U', 'DEL', 'INS', 'CODE', 'SMALL', 'SUB', 'SUP', 'CITE'];
```
overmyheadandbody, line 373: I'm only aware of these styles that are applied through the EDS pipeline (excluding headings, links and all the obvious ones) - https://main--milo--adobecom.aem.page/drafts/ramuntea/word-styles. So `strong`, `em`, `u`, `code`, `del`, `sub`, `sup`. Were you able to get a `mark`, `b`, `i`, `u`, `ins`, `small` or `cite`? If not, we can probably remove those references.

overmyheadandbody (commented):
I've noticed a side-effect of this - the headers get bolder due to now being wrapped in a different parent.
Screenshot 2025-05-20 at 12 41 18

Previous markup was `<div><strong>`, now it's `<div><p><strong>` which ends up applying a `font-weight: bolder` to the column heading.

libs/blocks/table/table.js
```diff
@@ -375,6 +375,15 @@ function handleSection(sectionParams) {
       sectionRowTitle.setAttribute('scope', 'row');
     }
   }
+
+  rowCols.forEach((col) => {
+    const formattingTags = col.querySelector('strong, em, del, code, sub, sup');
```
overmyheadandbody, line 380: Could it be that the `col` has a `strong` inside it, but it's already nested deeper in the DOM? Maybe something like  `.querySelector(':scope > :is(strong, em, del, code, sub, sup)');` would address this issue. And that could eliminate the need to have the `formattingTags.parentNode.tagName === 'DIV'` check as well

PR #4109, opened 2025-05-08
[MWPW-173161] - Refactor preflight to export checks

robert-bogos:
## Description
This PR is refactoring the preflight logic to export the methods used for checking the page.

## Related Issue
Resolves: [MWPW-173161](https://jira.corp.adobe.com/browse/MWPW-173161)

## Test URLs
**Milo:**
- Before: https://stage--milo--adobecom.aem.page/?martech=off
- After: https://vhargrave-general-preflight-enhancements--milo--adobecom.aem.page/?martech=off


mokimo (commented):
I wonder, should we offer a layer in between such as an `preflightApi.js` that imports all of the individual checks?

It might be more obvious that any change is a breaking change in an exposed API and that anything in the individual files is an implementation detail.

That would also make unit tests easier for the API part of it and clearly outline dangerous actions such as `it('verifies preflightApi.seo.checkX exists')` ... if we'd mess up or rename any imported methods, this test would instantly highlight a breaking change

robert-bogos:
> I wonder, should we offer a layer in between such as an `preflightApi.js` that imports all of the individual checks?
> 
> It might be more obvious that any change is a breaking change in an exposed API and that anything in the individual files is an implementation detail. It also makes it easier to decomission the individual files and gives us more liberty to change those - while keeping the api file intact
> 
> That would also make unit tests easier for the API part of it and clearly outline dangerous actions such as `it('verifies preflightApi.seo.checkX exists')` ... if we'd mess up or rename any imported methods, this test would instantly highlight a breaking change

@mokimo I personally like the idea, the whole thing would be easier to grasp if we centralise the imports in a single file. On the other hand, this change would require a small refactor. 
Any thoughts? @narcis-radu @overmyheadandbody

narcis-radu:
I like @mokimo's suggestion, let's implement it.

narcis-radu (approved):

robert-bogos:
Failing NALA tests should be solved once [this PR](https://github.com/adobecom/milo/pull/4201) is merged.

PR #4114, opened 2025-05-09
Mwpw 170378 preflight assets unit test

skholkhojaev:
Unit Test for Preflight Assets Tab

**Test URLs:**
- After: https://MWPW-170378-Preflight-Assets-Unit-Test--milo--skholkhojaev.aem.page/?martech=off


robert-bogos (commented):
Code looks good 👍
Just please rebase your branch to bring it up to date; it should fix the failing linter check.

Also you could change the named import of [the Panel component](https://github.com/adobecom/milo/blob/vhargrave/general-preflight-enhancements/test/blocks/preflight/panels/performance/performance.test.js#L19) into a default import. That should fix the unit tests check. I know is out of scope, but since we have this PR for unit tests, it’s good to fix it here.

PR #4121, opened 2025-05-09
MWPW-164457 Add injectblock metadata support

chrischrischris:
Adds support for `injectblock` metadata property, which when specified takes a comma delimited list of blocks to add to the page.  The blocks are added after initial page render to not affect performance, and a generic block with no variation is added to the bottom of the page.

PR also has:
* a few slight ui tweaks to graybox
* Graybox flyout is only open if there is a custom title or description

Resolves: [MWPW-164457](https://jira.corp.adobe.com/browse/MWPW-164457)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/cpeyer/temp/insertblock?martech=off
- After: https://injectblock--milo--adobecom.aem.page/drafts/cpeyer/temp/insertblock?martech=off

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -1552,6 +1552,11 @@ function decorateDocumentExtras() {
 }
 
 async function documentPostSectionLoading(config) {
+  const injectBlock = getMetadata('injectblock');
+  if (injectBlock) {
+    import('./injectblock.js').then((module) => module.default(injectBlock));
+  }
+
   decorateFooterPromo();
   if (getMetadata('seotech-structured-data') === 'on' || getMetadata('seotech-video-url')) {
     import('../features/seotech/seotech.js').then((module) => module.default(
--- a/libs/blocks/graybox/graybox.css
+++ b/libs/blocks/graybox/graybox.css
@@ -24,11 +24,16 @@
   top: 50%;
   left: 50%;
   transform: translate(-50%, -50%);
-  width: calc(100% - 16px);
+  width: calc(100% - 15px);
   height: calc(100vh - 16px);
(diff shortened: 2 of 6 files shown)


colloyd (approved):
Looks like there's some eslint recommendations, but it looks good to me otherwise.

sukamat (approved):

JackySun9 (approved):

libs/utils/utils.js
```diff
@@ -1552,6 +1552,11 @@ function decorateDocumentExtras() {
 }
 
 async function documentPostSectionLoading(config) {
+  const injectBlock = getMetadata('injectblock');
+  if (injectBlock) {
+    import('./injectblock.js').then((module) => module.default(injectBlock));
+  }
```
sukamat, line 1558: Temporarily blocking the PR based on the offline discussion.

As the metadata will only be added on a graybox origin site, if a user is browsing graybox via Akamai domain (*.graybox.adobe.com) and the page does not exist, Akamai does a 404 fallback and pulls the page from the source site origin. This page will not have the `injectBlock` metadata.

This will make the block injection work on pages that are authored in the graybox content repo and not on all pages  served on the graybox Akamai domain.

We might need a different approach if the requirement is to inject the graybox block on all pages serves from the graybox Akamai domain.

sukamat (changes requested):

sukamat (approved):

libs/utils/utils.js
sukamat: Requirement confirmed. It will be graybox pages only.

PR #4129, opened 2025-05-12
[MWPW-173157] More accurate aria-label for custom play button

overmyheadandbody:
This uses the video poster's `alt` value to provide a more accurate `aria-label` for the play button.

I've based this on the fact that a custom icon is the 3rd part of an image link definition, making it highly likely that authors will set an `alt` value, which is the 2nd part, see [here](https://github.com/adobecom/milo/blob/stage/libs/utils/utils.js#L740). Even if they don't, they can go in and add one to solve the accessibility issue, if requested. This strategy has also been suggested by our accessibility partners.

Even though the `alt` value could've been passed to the method as an argument, it might be that other consumers use the method directly, thus it was safer to adjust the method itself.

Resolves: [MWPW-173157](https://jira.corp.adobe.com/browse/MWPW-173157)

**Test URLs:**
- Before (PSI): https://main--milo--overmyheadandbody.aem.page?martech=off
- After (PSI): https://custom-play-label--milo--overmyheadandbody.aem.page?martech=off
- Before (Original issue): https://main--dc--adobecom.aem.page/acrobat/complete-pdf-solution?martech=off
- After (Original issue): https://main--dc--adobecom.aem.page/acrobat/complete-pdf-solution?milolibs=custom-play-label--milo--overmyheadandbody&martech=off


mokimo (approved):

DKos95 (approved):

SilviuLCF (approved):
Verified , testing details 
https://jira.corp.adobe.com/browse/MWPW-173157

sukamat (approved):

PR #4139, opened 2025-05-13
MWPW-173130: Preserve regular links in merch-card footers

st-angelo-adobe:
Resolves: [MWPW-173130](https://jira.corp.adobe.com/browse/MWPW-173130)

Fix M@S lana regression, where custom sample rate set by `<mas-commerce-service>` is overriden by code in mas.js setting it back to 1.
It used to work in mas-v0.4.0 because the order was different, first code in mas.js was executed and afterwards mas-commerce-service.

**Test URLs:**
- After: https://mwpw-173130--milo--adobecom.aem.page/?martech=off

Test URL (difference can be seen for `bdf40d06-5914-4f1f-aa10-77c5676fe671`):
mwpw-173130--milo--adobecom.aem.page/libs/features/mas/docs/ccd.html


3ch023 (approved):
@yesil i noticed that 'Register Now' became 'a' instead of 'button', is this ok or we need it to be button, just without commerce attributes?

@st-angelo-adobe once we get answer to the above, to make sure we don't re-introduce this issue, could you add a test in masccd.test.js to verify 'Register now' pls

libs/features/mas/src/hydrate.js
```diff
@@ -324,15 +324,21 @@ export function processUptLinks(fields, merchCard) {
...
-    const spectrumCta = CheckoutButton.createCheckoutButton({}, cta.innerHTML);
-    spectrumCta.setAttribute('tabindex', 0);
+function createSpectrumCssButton(cta, aemFragmentMapping, isOutline, variant, isCheckout) {
+    let button = cta;
+    if (isCheckout) {
+        const CheckoutButton = customElements.get('checkout-button');
+        button = CheckoutButton.createCheckoutButton({}, cta.innerHTML);
+    }
+    else {
+        button.innerHTML = `<span style="pointer-events: none;">${button.textContent}</span>`
```
yesil, line 334: for regular links and buttons, we don't need `style="pointer-events: none;"`

libs/features/mas/src/hydrate.js
```diff
@@ -405,15 +417,15 @@ export function processCTAs(fields, merchCard, aemFragmentMapping, variant) {
         const footer = createTag('div', { slot }, fields.ctas);
 
         const ctas = [...footer.querySelectorAll('a')].map((cta) => {
-            const checkoutLinkStyle =
-                CHECKOUT_STYLE_PATTERN.exec(cta.className)?.[0] ?? 'accent';
+            const isCheckout = cta.hasAttribute('data-wcs-osi');
```
yesil, line 420: To be on the safe side can we check `data-wcs-osi` has also a value.

yesil:
@3ch023 since what they really author are links, I'm fine with them being rendered as `a`.
@st-angelo-adobe however, I recommend to remove the class `accent` once the real classes are added.

yesil (approved):
can you also add a unit test in hydate.js for this specific use case?

3ch023:
> @3ch023 since what they really author are links, I'm fine with them being rendered as `a`. @st-angelo-adobe however, I recommend to remove the class `accent` once the real classes are added.

okie, i was just checking since you worked with CCD team and maybe they have some dependency on it being 'button' in code
otherwise good from my side

Roycethan:
@st-angelo-adobe  Plz resolve conflicts here thanks

PR #4140, opened 2025-05-14
Standalone Gnav Link transformation support for hosts other than stage.adobe.com

bandana147:
* Adds support to add the current rendered host to stage domain map
* Moving top nav wrapper variable to styles.css as it is overriding standalone gnav's smaller height variable

Resolves: [MWPW-173409](https://jira.corp.adobe.com/browse/MWPW-173409)

**Test URLs:**
- After: https://gnav--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/libs/styles/styles.css
+++ b/libs/styles/styles.css
@@ -2,6 +2,7 @@
 
  :root {
   --global-height-nav: 64px;
+  --feds-height-nav: 63px;
   --global-height-breadcrumbs: 33px;
   /* stylelint-disable-next-line custom-property-pattern */
   --global-height-navPromo: 72px;
--- a/libs/blocks/global-navigation/base.css
+++ b/libs/blocks/global-navigation/base.css
@@ -1,7 +1,6 @@
 /* Custom properties (variables) */
 :root {
   /* Top navigation - box model */
-  --feds-height-nav: 63px;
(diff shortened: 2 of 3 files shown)


sharmrj (approved):

libs/styles/styles.css
```diff
@@ -2,6 +2,7 @@
 
  :root {
   --global-height-nav: 64px;
+  --feds-height-nav: 63px;
```
salonijain3, line 5: what is the reason for moving the style to libs instead of global navigaiton?

bandana147: As base.css gets loaded later than navigation.css, it was overriding the smaller height variable of feds-topnav-wrapper for standalone gnav.

Deva309 (approved):

nishantka (approved):

prativas22:
Verified the standalone GNAV link formation supports stage domain mapping PR; the result looks good
https://adobecom.github.io/nav-consumer/?env=stage&authoringpath=/federal/learn&navbranch=gnav&newNav=true
image
image
image
image
image

https://adobecom.github.io/nav-consumer/navigation.html?env=stage&theme=dark&navbranch=gnav&authoringpath=/federal/dev/gnavtest/locgnav&locale=jp
image
image
image

PR #4142, opened 2025-05-14
Remove libraries flag

mokimo:
The libraries flag does not match the [sidekick schema](https://tools.aem.live/sidekick/config.schema.json) and is not used within the config service anymore. Thus we should remove it from Github as well. [MWPW-Ticket](https://jira.corp.adobe.com/browse/MWPW-173458)

**Test URLs:**
- After: https://remove-unused-flag--milo--adobecom.aem.page/?martech=off


robert-bogos (approved):

overmyheadandbody (approved):

mokimo:
I'm self-applying the `ready for stage` label as this is cleanup work and the config service already does not have this flag anymore!

SilviuLCF (approved):
No QA needed, ready to STAGE

PR #4155, opened 2025-05-15
Mwpw 170301 focus outside modal

zagi25:
Issue: Focus will leave modal when navigating with arrows / tab, this was fixed by https://github.com/adobecom/milo/pull/4035 . Another issue is that cookie banner will steal the focus from modal when it loads, resulting in ability to navigate the page while the modal is open. 
I consulted Taryn via Slack about the best possible solution, and it was decided that hiding the cookie banner while the modal is open is the best approach. 

* Hide cookie banner while the modal is open to prevent focus stealing
* Return focus to cookie banner if the modal was open via deep link (with modal hash in the url)
* Banner is shown again after closing the modal

Easiest way to force cookie banner to appear:
- Connect to VPN (e.g. dublin or any country that requires cookie banner to be present)
- Clear cookies
- Clear session storage
- Reload the page

Resolves: [MWPW-170301](https://jira.corp.adobe.com/browse/MWPW-170301)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/ratko/mwpw-170384-modal-overflow/modal-overflow?martech=off
- After: https://mwpw-170301-focus-outside-modal--milo--adobecom.aem.page/drafts/ratko/mwpw-170384-modal-overflow/modal-overflow?martech=off

**DC URLs:**
- Before: https://main--dc--adobecom.aem.page/acrobat/generative-ai-pdf
- After: https://main--dc--adobecom.aem.page/acrobat/generative-ai-pdf?milolibs=mwpw-170301-focus-outside-modal


libs/blocks/modal/modal.js
```diff
@@ -294,6 +300,7 @@ export default function init(el) {
   if (delayedModal(el) || window.location.hash !== modalHash || document.querySelector(`div.dialog-modal${modalHash}`)) return null;
   if (dialogLoadingSet.has(modalHash?.replace('#', ''))) return null; // prevent duplicate modal loading
   const details = findDetails(window.location.hash, el);
+  if (details) isDeepLink = true;
```
mokimo, line 303: Not sure I understand this correct, is it always a deepLink when there are details? Details sounds quite generic, so that any other modal could also have them or not.

Is there anything more specific within the details we should check?

zagi25: you're right, I moved `isDeepLink` to `getModal`, to make sure that it is updated when the modal is open. From what I saw, `init` function is only called when there is `hash` in the url when page is loaded. So I would say that is safe to assume that if there is a hash in the url when page is loaded and hash corresponds to modal `id`, modal was open via deep link.

mokimo (approved):

libs/blocks/modal/modal.js
```diff
@@ -294,7 +301,7 @@ export default function init(el) {
   if (delayedModal(el) || window.location.hash !== modalHash || document.querySelector(`div.dialog-modal${modalHash}`)) return null;
   if (dialogLoadingSet.has(modalHash?.replace('#', ''))) return null; // prevent duplicate modal loading
   const details = findDetails(window.location.hash, el);
-  return details ? getModal(details) : null;
+  return details ? getModal(details, undefined, true) : null;
```
overmyheadandbody, line 304: > I nit, therefore I am dev 😛 

Personal preference, but I hate passing `undefined` as an argument. I also prefer having objects for anything more than 2 params. I understand a lot of other method depend on calling this, so changing the params is not an option. But how about appending a new property to `details` to specify that this is a deep link modal?

Not a showstopper, more of a cosmetic improvement.

overmyheadandbody (approved):

libs/blocks/modal/modal.js
zagi25: I like this approach better also, updated. Thanks

robert-bogos (approved):

SilviuLCF (approved):
Verified, testing details https://jira.corp.adobe.com/browse/MWPW-170301

PR #4159, opened 2025-05-15
MWPW-169410: Add download icon to sidekick

zagi25:
* Add download icon

Screenshot 2025-05-15 at 13 57 56

Resolves: [MWPW-169410](https://jira.corp.adobe.com/browse/MWPW-169410)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/ratko/mwpw-169410-download-icon/download-icon?martech=off
- After: https://mwpw-169410-download-icon--milo--adobecom.aem.page/drafts/ratko/mwpw-169410-download-icon/download-icon?martech=off


mokimo (approved):

libs/img/icons/icons.svg
```diff
@@ -133,4 +133,10 @@
     <desc>X Logo</desc>
     <path fill="currentcolor" d="M714.163 519.284L1160.89 0H1055.03L667.137 450.887L357.328 0H0L468.492 681.821L0 1226.37H105.866L515.491 750.218L842.672 1226.37H1200L714.137 519.284H714.163ZM569.165 687.828L521.697 619.934L144.011 79.6944H306.615L611.412 515.685L658.88 583.579L1055.08 1150.3H892.476L569.165 687.854V687.828Z" />
   </symbol>
+  <symbol id="download" viewBox="0 0 18 18">
+    <title>Download</title>
```
overmyheadandbody, line 137: I'm not convinced with assigning this `title`. On your test page, using the screen reader, this leads to a "Download Download" announcement. However, should the icon be used outside of a link (would this even be a use-case?), then the title would be very helpful. I see mixed strategies in this file - company names are likely ok to have a `title`, others may lead to announcement duplication. Maybe we call it `Download icon` [0]? What do you think?

[0] This also brings up a question on localizing these strings, but that's a problem for another day 😅

overmyheadandbody (approved):

libs/img/icons/icons.svg
zagi25: Great point, I would say we can safely remove title and description so that only link text gets announced, `Download icon` needs to be inside the link so something can be downloaded right ? 🤔

narcis-radu (approved):

PR #4170, opened 2025-05-16
MWPW-173608 [M@S] Mnemonics filed disappear from card preview when …

bozojovicic:
… editing card

@st-angelo-adobe I had to revert this code. When card has some tag from `mas:product_code` it is OK. After editing in mas the card will be first rendered without mnemonics and then it will check the tag and it will render the card again to add the analytics attributes, this time with mnemonics.
But when card does not have any tags from `mas:product_code` the card mnemonics will be removed and there will be no 2nd rendering and mnemonics are gone. 
To keep it simple I did what we do for all other slot tags, we have them there all the time, no matter if the content is authored/set or not.

Tested on https://main--mas--adobecom.hlx.live/studio.html?milolibs=mwpw173608--milo--bozojovicic#page=content&path=sandbox&query=aec092ef-d5b5-4271-8b6f-4bbd535fcc56

Resolves: [MWPW-173608](https://jira.corp.adobe.com/browse/MWPW-173608)

**Test URLs:**
- After: https://mwpw173608--milo--bozojovicic.aem.live/?martech=off


markovicpredrag1209 (approved):

3ch023 (approved):

st-angelo-adobe:
@bozojovicic There's a problem. The reason why I did the weird change to begin with is that the slot itself is affected by `gap`. 

As you will see here: https://mwpw173608--milo--bozojovicic.aem.live/libs/features/mas/docs/plans.html, cards with no icons will have an extra `8px` space at the top, which should not happen.
I believe there's a `:has-assigned` pseudo selector for empty slots, but it doesn't work on Chrome (or it didn't last time I checked). Using that we could've been able to keep the slot, but style it so gap doesn't apply.

Sooo... I guess there's an issue.
cc: @3ch023 @markovicpredrag1209

bozojovicic:
@st-angelo-adobe I think it is OK now. Hydrate code is executed after the `render()` method and if there are no mnemonics it will remove that non-used slot.

st-angelo-adobe (approved):

PR #364, opened 2022-12-20
MWPW-116989 -  Consonant card horizontal block

Sartxi:
As an SEO marketer, I want multiple options for single action cards. [Card Horizontal, Card Horizontal (tile)] - *supports 1-3up

Resolves: [MWPW-116989](https://jira.corp.adobe.com/browse/MWPW-116989)

**Test URLs:**
- Before: https://main--milo--adobecom.hlx.page/drafts/sarchibeque/card-horizontal?martech=off
- After: https://card-horizontal--milo--adobecom.hlx.page/drafts/sarchibeque/card-horizontal?martech=off

PR #4180, opened 2025-05-19
MWPW-173425: Allow container to expand based on line height

zagi25:
Issue: When text spacing is applied via extension, `card-horizontal` content text overflows. 

* Removed container `max-height`, to let container expand when line spacing is applied (height is determined by [.card-image](https://github.com/adobecom/milo/blob/stage/libs/blocks/card-horizontal/card-horizontal.css#L108))
* Removed `margin` from `.card-content` because it increases height of the container when there is no `max-height`
* `-webkit-line-clamp: 5` in combination with [align-items](https://github.com/adobecom/milo/blob/stage/libs/blocks/card-horizontal/card-horizontal.css#L70) will keep content height as before

Screenshots: 
Screenshot 2025-05-19 at 10 35 19
Screenshot 2025-05-19 at 10 35 27

[Text spacing extension](https://chromewebstore.google.com/detail/text-spacing-editor/amnelgbfbdlfjeaobejkfmjjnmeddaoj)

Resolves: [MWPW-173425](https://jira.corp.adobe.com/browse/MWPW-173425)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/ratko/mwpw-173425-card-horizontal/card-horizontal?martech=off
- After: https://mwpw-173425-card-text-cutoff--milo--adobecom.aem.page/drafts/ratko/mwpw-173425-card-horizontal/card-horizontal?martech=off
- Before: https://main--milo--adobecom.aem.page/docs/library/kitchen-sink/card-horizontal?martech=off
- After: https://mwpw-173425-card-text-cutoff--milo--adobecom.aem.page/docs/library/kitchen-sink/card-horizontal?martech=off

**CC URLs:**: 
- Before: https://main--cc--adobecom.aem.page/fr/products/premiere/features
- After: https://main--cc--adobecom.aem.page/fr/products/premiere/features?milolibs=mwpw-173425-card-text-cutoff

### DIFF
--- a/libs/blocks/card-horizontal/card-horizontal.css
+++ b/libs/blocks/card-horizontal/card-horizontal.css
@@ -72,7 +72,6 @@
   border-radius: 4px;
   gap: var(--spacing-s);
   border: 1px solid #EAEAEA;
-  max-height: 130px;
   max-width: 600px;
   margin: 0 auto;
 }
@@ -90,7 +89,6 @@
 
 .card-horizontal .foreground .card-block .card-content {
   padding-right: var(--spacing-s);
-  margin: 16px 0;
   /* stylelint-disable-next-line value-no-vendor-prefix */
   display: -webkit-box;
   -webkit-line-clamp: 5;


mokimo (approved):
Is there any context or git history why we needed a max height to begin with?

zagi25:
> Is there any context or git history why we needed a max height to begin with?

@mokimo From what I see it was added at the beginning https://github.com/adobecom/milo/pull/364/files, tbh don't know why 😅 Line clamp prevents content from increasing container height.

libs/blocks/card-horizontal/card-horizontal.css
```diff
@@ -90,7 +89,6 @@
 
 .card-horizontal .foreground .card-block .card-content {
   padding-right: var(--spacing-s);
-  margin: 16px 0;
```
overmyheadandbody, line 93: I'm not so convinced with this one. I see the image height is static, so that shouldn't impact the layout, right? However, we might see font changes in the near future, which would make the card text collapse with the borders. Is there any way we can avoid this? Maybe setting some paddings on inner elements?

zagi25: @overmyheadandbody When I remove `max-height` margin will increase the height of the container (few pixels more than 130px, not much but because of the image it is visible) if there is enough text to fill out 5 lines. 
The only way I can think of that will cause the text collapse with the borders is if the font-size drastically changes and line-clamp changes. Something like this 
Screenshot 2025-05-21 at 03 16 12
Paddings don't do much, I can maybe keep the margin but decrease it so it doesn't affect container height ?

overmyheadandbody: Ok, let's leave it as is for now and figure out edge cases later.

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-173425

PR #4202, opened 2025-05-20
ServiceNow CMR Fixes - [MWPW=173724]

amauch-adobe:
Fixes these issues:
* ServiceNow Registry InstanceID that was retired has now been changed to correct ID
* Back-off timer has been updated to have an hour timeout since in high peak times the Kafka queue for the SNOW Change Request API may take a long time to complete requests. 
* GitHub action only runs when merges are closed against the production branch.  This now includes branches from forks being directly merged into the production branch for hot-fixes.
* Fixed API payload values for planned start and end time to be whole integers rather than timestamp floats.

Enhancements:
* Leveraging Python F-strings where needed
* Cleaned up functions to follow linting rules
* Updated release summary information for Change Requests to include PR URLs.
* Updated print statements to provide clarity for instructions for finding Change Requests (CRs) and/or debugging when a CR isn't found.

Resolves: [MWPW-173724](https://jira.corp.adobe.com/browse/MWPW-173724)

**Test URLs:**
- After: https://servicenow-cmr-MWPW-173724--milo--adobecom.aem.page/?martech=off


amauch-adobe:
Successfully tested in this action:  https://github.com/amauch-adobe/milo/actions/runs/15148657536
This PR: https://github.com/amauch-adobe/milo/pull/48

amauch-adobe:
@mokimo and @narcis-radu I went and looked over all of the failures from Nala and they were all based on these three situations:

1. Timeouts in loading pages.  I checked all the pages uses my feature branch url and the pages load just fine in Chrome.
2. Failures due the carousel block test now putting the starting index as 0 rather then Null, the test needs to be updated to reflect this index change.
3. Accessibility test failure on the review block. This failure only happened on Firefox it passed on the other browsers.

In all situations none of them are or should be affected by my code changes where everything in this PR is a github action workflow.  So I believe these are all prior issues in Nala. I believe, in my opinion, we are good to go.

.github/workflows/servicenow.py
```diff
@@ -226,20 +234,24 @@ def get_cmr_id_operation():
...
     sys.exit(1)
   else:
-    print("CMR closure was successful: ", response.status_code)
+    print(f"CMR closure was successful: {response.status_code}")
     print(response.text)
 
-  print("Change Management Request in ServiceNow was successful.")
-  print ("You can find the change record in ServiceNow https://adobe.service-now.com/now/change-launchpad/homepage, by searching for this ID: ", cmr_id)
+  print("Change Management Request has been sent to the queue and is being processed.")
+  print(f"You can find the change record in ServiceNow https://adobe.service-now.com/now/change-launchpad/homepage, by searching for this ID: {cmr_id}")
```
mokimo, line 253: Should we also send a slack message both for when this succeeds and fails?

.github/workflows/servicenow.yaml
```diff
@@ -31,9 +33,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+    - uses: actions/checkout@v4
     - name: Set up Python 3.x, latest minor release
-      uses: actions/setup-python@42375524e23c412d93fb67b49958b491fce71c38
+      uses: actions/setup-python@v5
```
mokimo, line 38: There was a supply chain attack a few months ago based on which we switched to commit SHAs based on https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

mokimo:
@amauch-adobe Rebasing to stage should help to fix those tests!

.github/workflows/servicenow.py
amauch-adobe: Yeah, I'll add that in.

.github/workflows/servicenow.yaml
amauch-adobe: Ah, I didn't know that, I will add back the SHAs.

JackySun9 (approved):

amauch-adobe:
The rebasing has caused issues, I'm pulling in things a fresh.

PR #4203, opened 2025-05-20
Accessibility carousel

zagi25:
Issues: Missing slide content announcement when navigating carousel.

- Added aria-live [docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-live)
- Update aria-live with slide text content or image alt if image is present and there is no text, so it gets announced to the user.
- Set `aria-hidden=true` to non visible slides so they are hidden from the reader
- Set `tabindex=-1` to focusable elements inside slide that has `aria-hidden=true`
- Update `aria-hidden` and `tabindex` on resize because of `show-` variant (on screens with width < 900 only one slide will be shown)
- Hide slide indicators ( this was decided by consonant and is a temporary)
- Updated nala test and disable anything related to indicators

**NOTE:** Some slides only contain videos, but videos don't have any information that can be used for `aria-live` not sure what to do in this case.

Resolves: [MWPW-171960](https://jira.corp.adobe.com/browse/MWPW-171960)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/ratko/mwpw-171960-carousel-dots/carousel?martech=off?martech=off
- After: https://accessibility-carousel--milo--adobecom.aem.page/drafts/ratko/mwpw-171960-carousel-dots/carousel?martech=off?martech=off
- Before: https://main--milo--adobecom.aem.page/docs/library/kitchen-sink/carousel?martech=off
- After: https://accessibility-carousel--milo--adobecom.aem.page/docs/library/kitchen-sink/carousel?martech=off
- Before: https://main--milo--adobecom.aem.page/drafts/rclayton/carousel-videos?martech=off
- After: https://accessibility-carousel--milo--adobecom.aem.page/drafts/rclayton/carousel-videos?martech=off

**CC URLs:**
- Before: https://main--cc--adobecom.aem.page/creativecloud/all-apps
- After: https://main--cc--adobecom.aem.page/creativecloud/all-apps?milolibs=accessibility-carousel


libs/blocks/carousel/carousel.js
```diff
@@ -179,6 +180,46 @@ function setIndicatorMultiplyer(carouselElements, activeSlideIndicator, event) {
...
+    if (img) ariaLive.textContent = img.getAttribute('alt');
+  }
+}
+
+function setAriaHiddenAndTabIndex(carouselElements, activeSlide) {
+  const { el: carouselBlock, slides } = carouselElements;
+  const active = activeSlide ?? carouselBlock.querySelector('.carousel-slide.active');
+  const activeSlideIndex = slides
+    .findIndex((ele) => active.isSameNode(ele));
+  const IndexOfShowClass = [...carouselBlock.classList].findIndex((ele) => ele.includes('show-'));
```
overmyheadandbody, line 201: Nit: we usually define constants/variables to start with lower case letters. Same comment for `NoOfVisibleSlides`. Any reason why that would not be ok? I realize this was here before, but it's probably a good time to standardize.

overmyheadandbody (commented):
Looks good for the most part! I've added a few comments; I'm hoping @rgclayton can help with some feature testing, as I'm not very familiar with all the carousel use-cases.

libs/blocks/carousel/carousel.js
```diff
@@ -179,6 +180,47 @@ function setIndicatorMultiplyer(carouselElements, activeSlideIndicator, event) {
...
+  let text = '';
+  slide.querySelectorAll(':scope > :not(.section-metadata').forEach((el, index) => {
+    text += `${index ? ' ' : ''}${el.textContent.trim()}`;
+  });
+  if (text) {
+    ariaLive.textContent = text;
+  } else {
+    const img = slide.querySelector('img[alt]');
+    const video = slide.querySelectorAll('video[title], iframe[title]');
+    ariaLive.textContent = img?.getAttribute('alt') || video?.getAttribute('title') || '';
```
mokimo, line 193: Since you have `querySelectorAll` here, you can't call `video?.getAttribute('title')` on a nodeList... or am I missing something?

So IMO you wanted to do
```suggestion
    const el = slide.querySelector('img[alt], video[title], iframe[title]');
    ariaLive.textContent = el?.getAttribute('alt') || el?.getAttribute('title') || '';
```

libs/blocks/carousel/carousel.js
```diff
@@ -179,6 +180,47 @@ function setIndicatorMultiplyer(carouselElements, activeSlideIndicator, event) {
...
+  }
+}
+
+function setAriaHiddenAndTabIndex(carouselElements, activeSlide) {
```
mokimo, line 197: are we sure, that both carouselElements and activeSlide are always defined?

libs/blocks/carousel/carousel.js
```diff
@@ -179,6 +180,47 @@ function setIndicatorMultiplyer(carouselElements, activeSlideIndicator, event) {
...
+    let tabIndex = -1;
+    if (index < noOfVisibleSlides) {
+      tabIndex = 0;
+      slide.removeAttribute('aria-hidden');
+    } else {
+      slide.setAttribute('aria-hidden', true);
+    }
+    slide.querySelectorAll(FOCUSABLE_SELECTOR).forEach((focusableElement) => { focusableElement.setAttribute('tabindex', tabIndex); });
+  });
+}
```
mokimo, line 222: We can get more concise with the naming and structure ... An untested suggestion:
```suggestion
function setAriaHiddenAndTabIndex({ el: block, slides }, activeEl) {
    const active = activeEl ?? block.querySelector('.carousel-slide.active');
    const activeIdx = slides.findIndex(el => el === active);
    const isWide = window.matchMedia('(min-width: 900px)').matches;
    const showClass = [...block.classList].find(cls => cls.startsWith('show-'));
    const visible = isWide && showClass ? +showClass.split('-')[1] : 1;
    const ordered = activeIdx > 0 ? [...slides.slice(activeIdx), ...slides.slice(0, activeIdx)] : slides;
    ordered.forEach((slide, i) => {
      const isVisible = i < visible;
      slide.toggleAttribute('aria-hidden', !isVisible);
      slide.querySelectorAll(FOCUSABLE_SELECTOR).forEach(el =>
        el.setAttribute('tabindex', isVisible ? 0 : -1)
      );
    });
  }
```

libs/blocks/carousel/carousel.js
zagi25: https://github.com/adobecom/milo/blob/stage/libs/blocks/carousel/carousel.js#L435, 
yes, `carouselElements` is defined at the beginning, and there should always be one active slide.

libs/blocks/carousel/carousel.js
zagi25: great catch, thank you !

mokimo (approved):

zagi25:
I commented `no-buttons` css rules, because if indicators are hidden and there is no buttons, it is not possible to use carousel. When indicators issue is resolved, `no-buttons` variant should be made available again.

PR #4204, opened 2025-05-21
Mwpw 169219 a11y aside

zagi25:
Issue: Content from aside desktop is not available on aside tablet/mobile. (more info [MWPW-167522](https://jira.corp.adobe.com/browse/MWPW-167522)) 

* Add tooltip icon with aside `textContent` to tablet and mobile if tablet/mobile `textContent` is different from desktop.

Resolves: [MWPW-169219](https://jira.corp.adobe.com/browse/MWPW-169219)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/ratko/mwpw-169219-a11y-aside/aside-tooltip?martech=off
- After: https://mwpw-169219-a11y-aside--milo--adobecom.aem.page/drafts/ratko/mwpw-169219-a11y-aside/aside-tooltip?martech=off
- Before: https://main--milo--adobecom.aem.page/drafts/ratko/mwpw-169219-a11y-aside/page?martech=off
- After: https://mwpw-169219-a11y-aside--milo--adobecom.aem.page/drafts/ratko/mwpw-169219-a11y-aside/page?martech=off
- Before: https://main--milo--adobecom.aem.page/docs/library/kitchen-sink/aside-promo?martech=off
- After: https://mwpw-169219-a11y-aside--milo--adobecom.aem.page/docs/library/kitchen-sink/aside-promo?martech=off
- Before: https://main--milo--adobecom.aem.page/ae_ar/drafts/ratko/aside?martech=off
- After: https://mwpw-169219-a11y-aside--milo--adobecom.aem.page/ae_ar/drafts/ratko/aside?martech=off


DKos95:
1. Last after link is broken.
2. When tabbing to a tooltip icon and clicking Escape to close it, I can't open it again with either Enter or Space, I think this would count as an a11y issue.
3. Focus box is a little bit skewed, not sure if this is also an a11y issue. If it is I would recommend positioning the icon centrally in it's parent div and then adding a minus margin that is equal to the space that is on the left side from the icon to the parent container, this would make it be in a vertical line with the content above like is now but with a better focus box.
Screenshot 2025-05-21 at 13 04 15

DKos95 (commented):
1. Last after link is broken.
2. When tabbing to a tooltip icon and clicking Escape to close it, I can't open it again with either Enter or Space, I think this would count as an a11y issue.
3. Focus box is a little bit skewed, not sure if this is also an a11y issue. If it is I would recommend positioning the icon centrally in it's parent div and then adding a minus margin that is equal to the space that is on the left side from the icon to the parent container, this would make it be in a vertical line with the content above like is now but with a better focus box.
Screenshot 2025-05-21 at 13 04 15

libs/blocks/aside/aside.css
```diff
@@ -595,11 +595,36 @@
   align-items: center;
 }
 
+.aside.promobar.popup .mobile-up.promo-text:has(.milo-tooltip) {
+  position: relative;
+}
+
+.aside.promobar.popup .promo-text .milo-tooltip {
+  font-size: 18px;
```
narcis-radu, line 603: shouldn't we use a token here for font?

libs/blocks/aside/aside.js
```diff
@@ -111,7 +111,43 @@ function checkViewportPromobar(foreground) {
   if (childCount < 3) addPromobar(children[childCount - 1], foreground);
 }
 
-function combineTextBocks(textBlocks, iconArea, viewPort, variant) {
+function toolTipPosition(container) {
+  const isRtl = document.documentElement.getAttribute('dir') === 'rtl';
+  const isTablet = container.classList.contains('tablet-up');
+  const isMobile = container.classList.contains('mobile-up');
+  if ((isRtl && isTablet) || (isMobile && !isRtl)) {
```
narcis-radu, line 118: nit: one line, no need for {}

libs/blocks/aside/aside.css
```diff
@@ -595,11 +595,28 @@
...
 }
 
+.aside.promobar.popup .mobile-up.promo-text:has(.milo-tooltip) {
+  position: relative;
+}
+
+.aside.promobar.popup .mobile-up.promo-text .milo-tooltip {
+  position: absolute;
+  bottom: var(--spacing-xs);
+  left: calc(var(--spacing-xs) - 4px);
```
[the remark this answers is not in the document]
zagi25: @overmyheadandbody I decided to go with adding styles for `rtl`, because `.action-area` might not be present and I feel there would be some other complications.
RTL page: https://mwpw-169219-a11y-aside--milo--adobecom.aem.page/ae_ar/drafts/ratko/aside?martech=off

libs/blocks/aside/aside.js
```diff
@@ -111,7 +111,43 @@ function checkViewportPromobar(foreground) {
...
+  const toolTipIcons = [];
+  [...foreground.querySelectorAll(':scope > div:not(.desktop-up)')].forEach((viewPortEl) => {
+    const childContent = viewPortEl.querySelector('.text-area');
+    const childContentText = childContent.textContent.trim();
+    if (childContentText === desktopContentText) return;
+    const appendTarget = viewPortEl.querySelector('.text-area').lastElementChild;
+    const iconWrapper = createTag('em');
+    iconWrapper.style.display = 'none';
+    iconWrapper.textContent = `${toolTipPosition(viewPortEl)}|${desktopContentText}`;
+    const tooltipSpan = createTag('span', { class: 'icon icon-tooltip' });
```
narcis-radu, line 135: Should the tooltip be visible on focus? I see this is the behaviour for tooltips in table - https://main--milo--adobecom.aem.page/docs/library/kitchen-sink/table

libs/blocks/aside/aside.css
zagi25: @narcis I wanted to but there is:
```
:root:lang(ar) {
  --body-font-family: adobe-arabic, myriad-arabic, 'Adobe Clean', adobe-clean, 'Trebuchet MS', sans-serif;
  --font-size-multiplier: 1.25;
}
```
` font-size: var(--type-heading-s-size);`

Screenshot 2025-05-21 at 18 11 12
Screenshot 2025-05-21 at 18 11 02

resulting in tooltip icon looking bigger on ar lang pages. If this is not a problem I can use a variable for font size and adjust tooltip icon accordingly.

mokimo: Should we always exclude icons within the `:root:lang(ar)` rule? I can imagine two scenarios:
- Font is bigger, hence icons should be equally bigger
- While the font is bigger, we actually never want to change the icons

libs/blocks/aside/aside.js
```diff
@@ -111,7 +111,41 @@ function checkViewportPromobar(foreground) {
   if (childCount < 3) addPromobar(children[childCount - 1], foreground);
 }
 
-function combineTextBocks(textBlocks, iconArea, viewPort, variant) {
+function toolTipPosition(container) {
```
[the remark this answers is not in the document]
zagi25: Tooltip position is set based on the direction word before the pipe `right|Tooltiptext`, this is usually done via authoring. Setting the correct position with only CSS would mean copying all `icons.css` rules connected to tooltip position and adjusting it for our use case.

libs/blocks/aside/aside.css
overmyheadandbody: Icons such as the tooltip are graphics, so their size should be consistent across locales. The fact that we're defining their size based on font rules is a dev choice, but ultimately the two shouldn't be tied together in my view. Table is a block that often uses the tooltip, I see it sets a `16px` maximum height. So, if anything, I'd say the size should be standardized to be consistent across blocks.

overmyheadandbody (approved):

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-169219

PR #4214, opened 2025-05-21
Mwpw 173540 preflight preview publish

skholkhojaev:
## Description
opned a PR for two tickets since both were small fixes

I've modified the `findLinks()` function so that it now stores URLs without query strings and hash fragments. Because, when comparing them they had the same origin but different query strings, hence why the function had no problems treating them as "unique" Links. Thats why these links would appear multiple times in the general tab.

Resolves: [MWPW-173540](https://jira.corp.adobe.com/browse/MWPW-173540)
Resolves: [WPW-173770](https://jira.corp.adobe.com/browse/MWPW-173770)

**Test URLs:**
- After: https://MWPW-173540-Preflight-Preview-Publish--milo--skholkhojaev.aem.page/?martech=off

to Test the page: https://main--bacom--adobecom.aem.page/sg/products/adobe-analytics/content-analytics/boost-campaign-performance?milolibs=MWPW-173540-Preflight-Preview-Publish--milo--skholkhojaev


libs/blocks/preflight/preflight.css
```diff
@@ -226,6 +226,8 @@ p.preflight-date-wrapper .disabled-publish {
 
 .preflight-group-row a {
   color: #FFF;
+  word-break: break-all;
```
mokimo, line 229: When tested with caas links, this still looks a bit funky. See [here](https://main--bacom--adobecom.aem.page/sg/products/adobe-analytics/content-analytics/boost-campaign-performance?milolibs=mwpw-173540-preflight-preview-publish--milo--skholkhojaev)

image

mokimo (commented):
We should ensure links have a max length

libs/blocks/preflight/preflight.css
mokimo: Its not a blocker, but I don't think theres any limit on link sizes, so if a link was potentially 3x longer than that CaaS one, we'd have issues again

skholkhojaev: so would recommend that we have a limit andthe links that exceed the limit to just not show the rest, have a horizontally scrollable text or show the beginning and the end of the URL and have Ellipsis in the middle?

mokimo: Both Ellipsis in the middle & not showing the rest is good. Since @rgclayton mentioned this was initially added so authors have a quick overview over the links, I'd vote for just cutting out the end on really long links

skholkhojaev: Screenshot 2025-05-22 at 10 01 12

does this look good?

libs/blocks/preflight/panels/general.js
```diff
@@ -72,8 +72,12 @@ function findLinks(selector) {
   return [...document.body.querySelectorAll(selector)]
     .reduce((links, el) => {
       const url = getUrl(el);
-      if (!hrefs.includes(url.href)) {
-        hrefs.push(url.href);
+      const baseUrl = url.origin + url.pathname;
```
overmyheadandbody, line 75: This definitely works, but how about using template literals instead?

libs/blocks/preflight/panels/general.js
```diff
@@ -72,8 +72,12 @@ function findLinks(selector) {
...
+      if (url.pathname.includes('/tools/caas')) {
```
overmyheadandbody, line 76: We might have additional paths to exclude in the future. How about storing such paths somewhere and checking against them here?

libs/blocks/preflight/panels/general.js
```diff
@@ -72,8 +72,12 @@ function findLinks(selector) {
...
+        return links;
+      }
```
overmyheadandbody, line 78: A Milo-specific nit: for brevity, when there's a single declaration inside a conditional, we drop the brackets and have everything on one line.

libs/blocks/preflight/panels/general.js
```diff
@@ -72,8 +72,12 @@ function findLinks(selector) {
...
+      if (!hrefs.includes(baseUrl)) {
+        hrefs.push(baseUrl);
```
overmyheadandbody, line 80: This could be a use-case for a `Set`. Maybe you can look into it and see if it makes sense to use it instead of checking in an array.

libs/blocks/preflight/preflight.css
```diff
@@ -226,6 +226,12 @@ p.preflight-date-wrapper .disabled-publish {
 
 .preflight-group-row a {
```
overmyheadandbody, line 227: This will target more links than intended, including the one that leads to the source document. How about you apply the changes more selectively to the relevant link?

overmyheadandbody (commented):
Good job! I've added a few comments with some ideas on how to make this even more robust, but the functionality is looking pretty good, congrats!

libs/blocks/preflight/preflight.css
```diff
@@ -134,6 +134,16 @@ p.preflight-content-heading-edit {
...
 }
 
+.preflight-group-row p:first-child a{
+  color: #FFF;
+  display: inline-block;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  max-width: 30em;
+  width: 100%;
```
overmyheadandbody, line 144: I'd like to challenge this a bit more - the `a` element is not the actual issue here. You're thus forced to make some assumptions about its maximum width, which are more of an estimation. Since `grid` is used to arrange the elements, maybe you can check what other CSS declarations you can change so you don't have to make any changes to the `a` element itself.

overmyheadandbody (commented):
Cool! But let's go one step further and check what else could be causing the layout issues, should the anchor content be too long.

libs/blocks/preflight/preflight.css
```diff
@@ -134,6 +134,16 @@ p.preflight-content-heading-edit {
   cursor: pointer;
 }
 
+.preflight-group-row p:first-child a{
+  color: #FFF;
```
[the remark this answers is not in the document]
skholkhojaev: Right i overlooked that part, will remove it

libs/blocks/preflight/preflight.css
skholkhojaev: I've changed `grid-template-columns: 1fr` to `grid-template-columns: minmax(0, 1fr)` to let the column shrink below its minimum content size and the rest of the content that is not in the column will be ellipsis.

robert-bogos (approved):
Nice! Looks good from my side.
P.S. Great that you implemented all the feedback!

hit11757 (approved):
Validated. Testing details https://jira.corp.adobe.com/browse/MWPW-173770

PR #4225, opened 2025-05-22
MWPW-173888 + MWPW-173919: Appending extra options for relative urls & multiple cta reopen fix

st-angelo-adobe:
Resolves: [MWPW-173888](https://jira.corp.adobe.com/browse/MWPW-173888), [MWPW-173919](https://jira.corp.adobe.com/browse/MWPW-173919)

GWP staged content with 3in1 modals on Stage. Planned go live 5th June.
Warriors implemented feature flag "mas-ff-3in1", to enable GWP publish 3in1 modals before 5th June. Until metadata flag is removed/switched to "on", 3in1 modals will fallback to be shown as legacy Dexter Modals.

However we've identified 2 issues with fallback mechanism:

### Issue 1
deeplinking with ms=e and cs=t is not working on fallback legacy modals
### Issue 2
if there are multiple modals with same hash, then opening the page with hash will result in hanging modal and all of them opening at the same time.

This PR is blocking GWP from publishing content, so we will be requesting approval to merge it during RCP.

**Test URLs:**
- After: https://mwpw-173888--milo--adobecom.aem.page/?martech=off

Can be tested by overriding the source directly on these urls 
https://www.adobe.com/cc-shared/fragments/drafts/mariia/3in1
https://www.stage.adobe.com/cc-shared/fragments/drafts/mariia/3in1


libs/blocks/merch/merch.js
```diff
@@ -535,11 +537,15 @@ export function appendExtraOptions(url, extraOptions) {
       );
     }
   });
+  if (isRelative) {
```
3ch023, line 540: might be that we don't even need this part

3ch023 (approved):
This will need thourough regression of 3in1 and dexter fallback modals.
plus it might affect non-fallback legacy modals.

impact: appendExtraOptions never worked for Dexter modals, but now it will.
Dexter modals should ignore unknown params, but they might show up in url.

bozojovicic (approved):

libs/blocks/merch/merch.js
st-angelo-adobe: True, initially I just left the full url, but then idk.. I thought maybe I should leave things going out of that method as is
But I can simplify and return a full url

bozojovicic (approved):

npeltier (approved):

PR #4254, opened 2025-05-27
[MWPW-170178] Table highlight a11y

overmyheadandbody:
This is an attempt to address the accessibility issues related to tables that make use of highlights. What it tries to address:
* stop announcing highlight content as the first row; these are visually tied to a column heading, thus highlights should not be announced individually;
* announce highlight text along with heading content when navigating through table cells, as that is what the visual representation suggests;
* make sure the highlight row isn't counted, so that the heading row is announced as the first one;
* reduce possibilities of new layout regressions being introduced.

This is likely a suboptimal fix, as the table structure would need to be completely rebuilt to account for all accessibility requirements; that redesign is currently being discussed, but in the interim, this should bring the table closer to being EAA-compliant.

### Screenshots
Notice that the VoiceOver announcement from the 'After' capture contains the highlight text as well.

Before | After
--- | ---
Screenshot 2025-05-27 at 15 59 16 | Screenshot 2025-05-27 at 15 57 54

Resolves: [MWPW-170178](https://jira.corp.adobe.com/browse/MWPW-170178)

**Test URLs:**
- Before: https://main--milo--overmyheadandbody.aem.page/docs/library/kitchen-sink/table?martech=off
- After: https://table-highlight--milo--overmyheadandbody.aem.page/docs/library/kitchen-sink/table?martech=off

### DIFF
--- a/libs/blocks/table/table.css
+++ b/libs/blocks/table/table.css
@@ -131,6 +131,12 @@
   visibility: hidden;
 }
 
+.table .highlight-text {
+  overflow: hidden;
+  width: 0;
+  height: 0;
+}
+
 /* heading */
 .table .col-heading.hidden {
   border-top: none !important;
--- a/libs/blocks/table/table.js
+++ b/libs/blocks/table/table.js
@@ -39,8 +39,8 @@ function handleHeading(table, headingCols) {
     if (!elements.length) {
       col.innerHTML = `<p class="tracking-header">${col.innerHTML}</p>`;
     } else {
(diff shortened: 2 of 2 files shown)


colloyd (approved):

rgclayton (approved):

mokimo (approved):

SilviuLCF (approved):
verified , testing details https://jira.corp.adobe.com/browse/MWPW-170178

PR #4255, opened 2025-05-28
Centralized DA importer hosted within milo

mokimo:
### Description
Based on the workflow that the [helix-website](https://github.com/adobe/helix-website/blob/main/.github/workflows/log-publish.yml) uses - I've rewritten the DA nodejs based importer in a way that it can be **[imported from milo](https://github.com/adobecom/da-playground/blob/main/.github/workflows/track-publish.yaml#L166-L170)** into a consuming repo/or a separate milo workflow just needs to have a separate `.yaml` file passing the right environment variables.

I've [stress-tested ](https://github.com/adobecom/da-playground/actions/runs/15276808974)this new importing workflow which imported >4000 articles within 45 minutes...
Actions can [run for up to 6 hours](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/troubleshooting-workflows/actions-limits) which means that we should be fine even with bigger events being published.

I also refactored the messaging so that we get a Slack notification keeping us in the loop about what gets imported into which projects
```
Importing 46 published documents from 906 log entries. Log Link: URL
• Importer fetched logs from: adobecom/bacom
• Importing into: adobecom/da-playground

Succcessful: 46 paths | Failed: 0 paths.
• Importer fetched logs from: adobecom/bacom
• Importing into: adobecom/da-playground
```

Resolves: [MWPW-173767](https://jira.corp.adobe.com/browse/MWPW-173767)

**Test URLs:**
- After: https://centralized-importer--milo--adobecom.aem.live/?martech=off


narcis-radu:
@mokimo - I want to make sure I understand your message:
>I've [stress-tested ](https://github.com/adobecom/da-playground/actions/runs/15276808974)this new importing workflow which imported 4500 articles within 45 minutes...
I am seeing the following in the poll-logs:
```
{ result: { success: 4158, error: 0, errorPaths: [] } }
```

mokimo:
> @mokimo - I want to make sure I understand your message:
I've [stress-tested ](https://github.com/adobecom/da-playground/actions/runs/15276808974)this new importing workflow which imported 4500 articles within 45 minutes...
I am seeing the following in the poll-logs:
{ result: { success: 4158, error: 0, errorPaths: [] } }

I've edited the message to feature the exact number. It imported exactly as many articles as I've expected it to and we did not run into any rate-limiting!

.github/workflows/import/daFetch.js
```diff
@@ -0,0 +1,107 @@
...
+    },
+    body: params,
+  });
+
+  if (!response.ok) {
+    throw new Error('Failed to retrieve IMS token');
+  }
+
+  const data = await response.json();
+  token = data.access_token;
```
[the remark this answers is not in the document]
mokimo: If we'd return a token, we would create a new IMS token for every request. On ~5k articles, we don't need to create 5000 tokens, but can re-use the one we initially create.

.github/workflows/import/converters.js
```diff
@@ -0,0 +1,168 @@
...
+    gridTable: mdast2hastGridTablesHandler(),
+  };
+  const hast = mdast2hast(mdast, { handlers, allowDangerousHtml: true });
+  return raw(hast);
+}
+
+function removeImageSizeHash(dom) {
+  const imgs = dom.window.document.querySelectorAll('[src*="#width"]');
+  imgs.forEach((img) => {
+    img.setAttribute('src', img.src.split('#width')[0]);
```
overmyheadandbody, line 144: Hopefully these URLs are always formatted correctly. Otherwise, using a regex to `replace` the width hash and its value might be a safer option.

.github/workflows/import/index.js
```diff
@@ -0,0 +1,277 @@
...
+    if (ROLLING_IMPORT_ENABLE_DEBUG_LOGS)
+      console.log(
+        'Failed to import resource to DA ' +
+          toOrg +
+          '/' +
+          toRepo +
+          ' | destination: ' +
+          url.pathname +
+          ' | error: ' +
+          e.message
```
overmyheadandbody, line 271: 🤷 
```suggestion
        `Failed to import resource to DA ${toOrg}/${toRepo} | destination: ${url.pathname} | error: ${e.message}`
```

.github/workflows/import/poll-logs.js
```diff
@@ -0,0 +1,270 @@
...
+
+function getWorkflowRunUrl() {
+  if (GITHUB_SERVER_URL && toRepo && GITHUB_RUN_ID) {
+    return `${GITHUB_SERVER_URL}/${toOrg}/${toRepo}/actions/runs/${GITHUB_RUN_ID}`;
+  }
+  return null;
+}
+
+const slackNotification = (text) => {
+  console.log(text);
```
overmyheadandbody, line 73: Is this still needed? If so, should it be prefixed with a mention that the message is intended for Slack?

.github/workflows/import/poll-logs.js
```diff
@@ -0,0 +1,270 @@
...
+        .filter((entry) => entry.route === 'live')
+        .flatMap((log) => [
+          log.path,
+          ...(Array.isArray(log.paths) ? log.paths : []),
+        ])
+        .filter(Boolean)
+    )
+  );
+  if(LOCAL_RUN) console.log("Live paths found: ", livePaths.length);
+  if (!LOCAL_RUN)
```
overmyheadandbody, line 196: Maybe this could work as well
```suggestion
  if(LOCAL_RUN) console.log("Live paths found: ", livePaths.length);
  else
```

.github/workflows/import/poll-logs.js
mokimo: The console.log will output it to the action-logs, so yes this is intentional so that if one goes into the action - we have better debug logs

.github/workflows/import/converters.js
mokimo: Keeping it as is for parity with nexter https://github.com/da-sites/nexter/blob/1932d72cac6d74d457bad14869ffa0cd85fe335e/nx/utils/converters.js#L136

.github/workflows/import/converters.js
```diff
@@ -0,0 +1,168 @@
...
+      ...table.querySelectorAll(':scope > tbody > tr, :scope > thead > tr'),
+    ];
+    const nameRow = rows.shift();
+    const divs = rows.map((row) => {
+      const cols = row.querySelectorAll(':scope > td, :scope > th');
+      // eslint-disable-next-line no-shadow
+      const divs = [...cols].map((col) => {
+        const { innerHTML } = col;
+        const div = dom.window.document.createElement('div');
+        div.innerHTML = innerHTML;
```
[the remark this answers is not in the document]
mokimo: Also here for parity with nexter: https://github.com/da-sites/nexter/blob/main/nx/utils/converters.js#L43

mokimo:
@overmyheadandbody FYI most of the comments you made are related to nexter core code, I'm trying to keep parity apart necessary changes to make it work on nodejs vs in-browser.

Reason being is, if we detect a bug in nexter -> we'll fix it there and only afterwards here.
If we detect a bug in the auto importer, it should always be contributed back to nexter for the greater DA-community.

PR #4286, opened 2025-06-02
Set the correct path from which the workflow is running from

mokimo:
### Description
Small improvement for https://github.com/adobecom/milo/pull/4255 to ensure we generate the right link to the workflow. Initially I had planned for the worfklows to be executed in each consumer repo, however for governance purposes milo makes more sense as central hub.

Resolves: [MWPW-173767](https://jira.corp.adobe.com/browse/MWPW-173767)

**Test URLs:**
- After: https://rolling-importer-fixes--milo--mokimo.aem.live/?martech=off

### DIFF
--- a/.github/workflows/import/poll-logs.js
+++ b/.github/workflows/import/poll-logs.js
@@ -52,7 +52,7 @@ if (!LOCAL_RUN)
 
 const queue = new PQueue({ concurrency: 10 });
 const FROM_PARAM = LOCAL_RUN
-  ? getISOSinceXDaysAgo(1)
+  ? (LAST_RUN_ISO || getISOSinceXDaysAgo(1))
   : encodeURIComponent(LAST_RUN_ISO || getISOSinceXDaysAgo(1));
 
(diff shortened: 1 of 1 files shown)


skholkhojaev:
shouldn't ESLint detect the missing semicolon or am I im missing something? 🤔

mokimo:
@skholkhojaev no it's ignored by default... Not sure why, I don't think .github/* should be ignored by default

> ⚠️ [eslint] reported by [reviewdog](https://github.com/reviewdog/reviewdog) 🐶
File ignored by default. Use a negated ignore pattern (like "--ignore-pattern '!<relative/path/to/filename>'") to override.

mokimo:
Also, there's no semicolon missing @skholkhojaev ... [multi line ternaries don't have semicolons](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_operator#conditional_chains)

robert-bogos (approved):

PR #4293, opened 2025-06-03
DOTCOM-145774 Default values for commerce flags

bozojovicic:
We accidentally merged the code, one year ago, that changes the default state for `displayOldPrice` and `displayPerUnit` and that caused some problems since many prices have manually added "per licence" text next to the price. GWP said they need this code to be enabled again to fix the content.
The same counts for default values for `displayTax` and `forceTaxExclusive` where, to make things more interesting, we have different values for different countries and different customer/market segments, as presented in the **table** here (2nd sheet) https://adobe-my.sharepoint.com/:x:/r/personal/rakaur_adobe_com/_layouts/15/doc2.aspx?sourcedoc=%7B1CE2C121-3004-4F19-A8AC-A3AB4A10E5AA%7D&file=vat-label-issues.xlsx&wdLOR=cEE9D3057-2BCE-4D1A-BC1A-C6E5BE3D1E5A&action=default&mobileredirect=true

This time, we introduced feature flag `mas-ff-defaults` for these 3 changes.
It needs to be set to `true` to enable these changes, so once it is enabled for some page or locale, all changes will be enabled
When GWPs finish with all changes on the whole content we can remove this FF from the code.
At the moment, there are no changes on pages where FF is not set.

**Test page for all 3 changes** : `mwpw151238--milo--bozojovicic.aem.page/drafts/bozo/prices`
The change for parameters `seat` and `old` can be tested on any locale while the VAT label can be tested only on locales where VAT labels are displayed according to the table in the link above.

The change in OST code will have to be made to reflect these default values in OST checkboxes. But that can be done only after this PR is merged. Actually the PR with these changes is still open :) https://git.corp.adobe.com/wcms/tacocat.js/pull/631

Resolves: [MWPW-151238](https://jira.corp.adobe.com/browse/MWPW-151238)
Resolves: [MWPW-151445](https://jira.corp.adobe.com/browse/MWPW-151445)
Resolves: [MWPW-143069](https://jira.corp.adobe.com/browse/MWPW-143069)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.live/drafts/bozo/prices?martech=off
- After: https://mwpw151238--milo--bozojovicic.aem.live/drafts/bozo/prices?martech=off


3ch023:
@bozojovicic could you add your flag here https://github.com/adobecom/milo/wiki/Feature-Flags---Request-Parameters-Registry?

3ch023 (changes requested):
could you pls make https://mwpw151238--milo--bozojovicic.aem.page/drafts/bozo/prices  PSI-check friendly and add it in /before and /after links?

also, let's add nala tests.
there is an easier and harder option :) easier - we could move your test page to /drafts/nala/features/commerce and just add checks for these changes. we woudl also need a page in a specific locale to test that use-case.
harder option is to review existing nala tests for cta/prices and see if we have all defaults covered - and add those too.

libs/features/mas/src/constants.js
```diff
@@ -166,3 +166,5 @@ export const TEMPLATE_PRICE = 'price';
 export const TEMPLATE_PRICE_STRIKETHROUGH = 'price-strikethrough';
 export const TEMPLATE_PRICE_ANNUAL = 'annual';
 export const TEMPLATE_PRICE_LEGAL = 'legal';
+
+export const FF_DEFAULT_FLAG = 'mas-ff-default-flag';
```
yesil, line 170: the flag term is repeated.
how about `mas-ff-defaults` ?

libs/features/mas/src/inline-price.js
```diff
@@ -92,7 +180,7 @@ export class InlinePrice extends HTMLSpanElement {
             'data-display-annual',
             'data-perpetual',
             'data-promotion-code',
-            'data-tax-exclusive',
+            'data-force-tax-exclusive',
```
yesil, line 183: why change this? if you change this, you need to update inline-price.md as well.

libs/features/mas/src/inline-price.js
```diff
@@ -255,15 +290,24 @@ export class InlinePrice extends HTMLSpanElement {
...
         if (!options.wcsOsi.length) return false;
 
-        /*
-        Commented out until issues in content with manually added tax labels are resolved
-
-        if (!this.masElement.dataset.displayTax) {
-            // set default value for displayTax if not set neither in OST nor in price URL
-            options.displayTax =
-                (await this.resolveDisplayTax(service, options)) || false;
+        const ffDefaultFlag = getMetadata(FF_DEFAULT_FLAG) === 'true';
```
yesil, line 293: the metadata logic belongs to Milo, not to MAS. The feature flag should be resolved earlier when the commerce is enabled first by merch.js and set on `mas-commerce-service`  somehow.
We should avoid moving Milo concept in MAS.
fyi: @3ch023 @npeltier

libs/features/mas/src/settings.js
```diff
@@ -28,6 +29,7 @@ function getLocaleSettings({
 }
 
 function getSettings(config = {}) {
+    const ffDefaultflag = getMetadata(FF_DEFAULT_FLAG) === 'true';
```
yesil, line 32: please move metadata logic to merch.js

yesil (changes requested):

libs/features/mas/src/inline-price.js
bozojovicic: Actually first I tried it like that  andI was totally not happy with the solution. And now I see what I did wrong. That flag should be set in `mas-commerce-service`, I better don't say what I did. :D Will fix it.

libs/features/mas/src/inline-price.js
bozojovicic: Because `data-tax-exclusive` is wrong. We don't use that data attribute. We have `data-force-tax-exclusive` when `exclusive` flag is set to `true`. I will fix inline-price.md.

libs/features/mas/src/settings.js
3ch023: @bozojovicic you just call `getParameter(FF_DEFAULT_FLAG)` , it will work

libs/features/mas/src/inline-price.js
```diff
@@ -71,13 +78,95 @@ const DISPLAY_TAX_MAP = {
...
+// and excluded for Business and Universities. This is the map of TaxExclusive values for other countries
+const TAX_EXCLUDED_MAP = {
+  ['ZA_en']: [false, false, Defaults.forceTaxExclusive, Defaults.forceTaxExclusive],
+  ['AT_de']: [false, false, false, true],
+  ['NG_en']: [false, false, Defaults.forceTaxExclusive, Defaults.forceTaxExclusive],
+  ['AU_en']: [false, false, false, false],
+  ['JP_ja']: [false, false, false, false],
+  ['NZ_en']: [false, false, false, false],
+  ['TH_en']: [false, false, false, false],
+  ['TH_th']: [false, false, false, false],
```
npeltier, line 96: aren't those somehow redundant?

npeltier (approved):
i find the logic a bit verbose, but approving for now, feel free to make it leaner :)

mokimo:
Heya, since this PR has been open since a while - could you leave some feedback on https://github.com/orgs/adobecom/discussions/4407 why/what is blocking this from proceeding?

libs/features/mas/src/inline-price.js
```diff
@@ -71,13 +78,95 @@ const DISPLAY_TAX_MAP = {
...
 };
+const TAX_EXCLUDED_MAP_INDEX = [INDIVIDUAL, BUSINESS, STUDENT, UNIVERSITY];
+const TAX_EXCLUDED_MAP_OTHER = {
+  [INDIVIDUAL]: false,
+  [BUSINESS]: true,
+  [STUDENT]: false,
+  [UNIVERSITY]: true,
+};
```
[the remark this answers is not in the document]
bozojovicic: Done

3ch023 (approved):

PR #4311, opened 2025-06-04
MWPW-170229: Show action scroller nav on resize

zagi25:
This resolves an issue where navigation button in `action-scroller` doesn't show up after resize.

Resolves: [MWPW-170229](https://jira.corp.adobe.com/browse/MWPW-170229)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/ratko/mwpw-169845/tabs-scroller?martech=off
- After: https://mwpw-170229-scroller-resize--milo--adobecom.aem.page/drafts/ratko/mwpw-169845/tabs-scroller?martech=off
- Before: https://main--milo--adobecom.aem.page/docs/library/kitchen-sink/action-items?martech=off
- After: https://mwpw-170229-scroller-resize--milo--adobecom.aem.page/docs/library/kitchen-sink/action-items?martech=off

### DIFF
--- a/libs/blocks/action-scroller/action-scroller.js
+++ b/libs/blocks/action-scroller/action-scroller.js
@@ -99,6 +99,13 @@ function handleNavigation(el) {
   return buttons;
 }
 
+const allActionScrollers = [];
+function handleResize() {
+  allActionScrollers.forEach(({ scroller, buttons }) => {
+    handleBtnState(scroller, buttons);
+  });
+}
+
 export default function init(el) {
   const hasNav = el.classList.contains(NAV);
   const actions = el.parentElement.querySelectorAll('.action-item');
@@ -109,6 +116,8 @@ export default function init(el) {
   el.replaceChildren(items, ...buttons);
   if (hasNav) {
     items.addEventListener('scroll', () => handleBtnState(items, buttons));
+    allActionScrollers.push({ scroller: items, buttons });
+    window.addEventListener('resize', handleResize);
     handleBtnState(items, buttons);
   }
 }


libs/blocks/action-scroller/action-scroller.js
```diff
@@ -99,6 +99,13 @@ function handleNavigation(el) {
   return buttons;
 }
 
+const allActionScrollers = [];
+function handleResize() {
+  allActionScrollers.forEach(({ scroller, buttons }) => {
+    handleBtnState(scroller, buttons);
+  });
+}
```
overmyheadandbody, line 107: I'm not sure I understand why we need to iterate all action scrollers 🤔 And say that we have 3 action scrollers on the page, the resize event would be attached 3 times, thus running the logic 3 times for each action scroller, right?

Would it be an option to have the (debounced) resize listener and the scroll one in the `handleBtnState` method directly?

zagi25: I don't know if this will be a real world case, but I create this array just in case that there are multiple `action-scroller` blocks present on the page (like on my test page), and to decide if navigation should be present we need to go over all the action-scrollers (some have more elements some have less so we need to check all of them).

If  `window.addEventListener('resize', handleResize);` is called multiple times, listener will only be attached once because callback function has the same reference as already attached one, and it will be executed once per firing of `resize` event. 
```
If the function or object is already in the list of event listeners for this target, 
the function or object is not added a second time.
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener
```

About the debounce, you mean to debounce the callback so it is not called every time `resize` is fired but after some time or you meant to debounce the attachment to prevent multiple calls to `addEventListener` ? 

Resize callback is called only once per firing: 
https://github.com/user-attachments/assets/411ab631-edb9-4dde-aa48-3b50e771ea71

overmyheadandbody: Debouncing the callback, so we don't call it each time the resize event fires, which can occur quite often depending on how the window is resized.

zagi25: Updated, `scroll` and `resize` events are now debounced. I've chosen conditional import  of `action.js` just because of the pattern that we follow `only import when you need something`, and because there is non-navigation variant, `action.js` is not needed all the time. If you think that this is slight overkill because `action.js` is small file, I can import it normally to make a code little bit cleaner.

overmyheadandbody: No, I think this is the proper approach 👌

DKos95 (approved):

rgclayton (approved):
Nice work.

Maybe something for future thought related to having many action-item scrollers in tabs (I don't see this as a real use case right now) would be to use something like `requestAnimationFrame` within `handleResize` to batch DOM writes and reduce layout updates.

hit11757 (approved):
Validated. Testing details can be found in the ticket https://jira.corp.adobe.com/browse/MWPW-170229

PR #4353, opened 2025-06-10
MWPW-174782 and MWPW-174694

suhjainadobe:
* Add your
* Specific
* Features or fixes

Resolves: [MWPW-174782](https://jira.corp.adobe.com/browse/MWPW-NUMBER)

After: https://mweb-errors--milo--suhjainadobe.aem.page/drafts/suhjain/m2-folder/carousel/photoshop-page?mep=/drafts/suhjain/ace1052.json&martech=off&georouting=off

### DIFF
--- a/libs/mep/ace1052/carousel/carousel.css
+++ b/libs/mep/ace1052/carousel/carousel.css
@@ -11,8 +11,7 @@
   --carousel-button-border-color: #e1e1e1;
   --carousel-button-border-hover-color: #cacaca;
   --carousel-button-border-focus-color: #378ef0;
-  --carousel-indicator-background-color: var(--carousel-button-border-hover-color);
-  --carousel-indicator-non-active-background-color: #8F8F8F;
+  --carousel-indicator-background-color: #8F8F8F;
   --carousel-indicator-active-background-color: #242424;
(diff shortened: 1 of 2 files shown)


libs/mep/ace1052/carousel/carousel.css
```diff
@@ -12,6 +12,7 @@
   --carousel-button-border-hover-color: #cacaca;
   --carousel-button-border-focus-color: #378ef0;
   --carousel-indicator-background-color: var(--carousel-button-border-hover-color);
+  --carousel-indicator-non-active-background-color: #8F8F8F;
```
overmyheadandbody, line 15: Hopefully this becomes the new default, right? So I'd suggest the value be updated for `--carousel-indicator-background-color`. Is there a reason why that wouldn't work? I'd also suggest you add an mweb comment here as well, so we know why this was updated compared to the prod code.

suhjainadobe: A new variable was added only because i was not sure, if these changes are to done for the prod. But as it won't affect the revert, Ill make the changes to the --carousel-indicator-background-color itself.

overmyheadandbody (approved):

mokimo (approved):

PR #4368, opened 2025-06-11
[Fire-Drill] This will have to get reverted by ryan

mokimo:
### Description
As part of offering a robust team of capable engineers, we are conducting "fire-drills" to emergency revert something from production.

Team members that have not recently gone through this exercise in a real case, will get a simplified case where they'll have to go through the steps to remove code from production outside of a normal automated batched-release.

Resolves: No MWPW

**Test URLs:**
- After: https://ryan-fire-drill--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/test/blocks/fire-drill-ryan/README.md
+++ b/test/blocks/fire-drill-ryan/README.md
@@ -1,0 +1,1 @@
+Let's revert this after this is on production.


test/blocks/fire-drill-ryan/README.md
```diff
@@ -0,0 +1 @@
+Let's revert this after this is on production.
```
robert-bogos, line 1: nit: Could be shorter. 
JK 😂
```suggestion
Let's revert this after is on production.
```

robert-bogos (approved):

PR #4371, opened 2025-06-11
MWPW-174199 M@S Plans Page: traditional Milo modals do not open from …

bozojovicic:
…M@S generated merch cards

When MAS card/collection is ready, `decorateLinks()` is called for that card/collection element where all `a[href]` elements are processed. If some of these links have URL with hash value, it will be transformed into the Milo modal link. Working example for both MAS cards and collections below.

Problem is reported for the card on STE tab on plans page
https://main--cc--adobecom.aem.live/creativecloud/plans?milolibs=mwpw174199plansmodals--milo--bozojovicic
but I fixed it for both individual cards and card collections which can be tested on PSI link below.

Resolves: [MWPW-174199](https://jira.corp.adobe.com/browse/MWPW-174199)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.live/drafts/bozo/pr/plansmodal?martech=off
- After: https://mwpw174199plansmodals--milo--bozojovicic.aem.live/drafts/bozo/pr/plansmodal?martech=off


3ch023 (changes requested):
lgtm and works as expected for See Terms link.
Noticed missing requirement thought - we need to make it work also with Dexter modals, because this is where What's Included content is located.

bozojovicic:
@3ch023 It actually works for both, milo modals and dexter modals but it doesn't work on `(aem|hlx).page`
On pages https://main--cc--adobecom.aem.live/creativecloud/plans?milolibs=mwpw174199plansmodals--milo--bozojovicic (STE tab) and on my test page https://mwpw174199plansmodals--milo--bozojovicic.aem.live/drafts/bozo/pr/plansmodal?martech=off you can see it in action (You will have to configure ModHeader to really see the Dexter page)
Now, what is the problem with `aem.page`? This method here https://github.com/adobecom/milo/blob/stage/libs/utils/utils.js#L932 redirects https://www.adobe.com/creativecloud/whats-included/plans/edu-all-apps-whats-included.html to https://main--cc--adobecom.aem.page/creativecloud/whats-included/plans/edu-all-apps-whats-included.html and then we have 404. It is configured to do this. The object `stageDomainsMap` has values 
```
{
    "www.stage.adobe.com": {
        "www.adobe.com(?!\\/*\\S*\\/(mini-plans|plans-fragments\\/modals|genuine(\\.html)?\\/?)\\S*)": "origin",
    },
    "--cc--adobecom.(hlx|aem).live": {
        "www.adobe.com": "www.adobe.com",
    },
    "--cc--adobecom.(hlx|aem).page": {
        "www.adobe.com(?!\\/*\\S*\\/(mini-plans|plans-fragments\\/modals|genuine(\\.html)?\\/?)\\S*)": "origin",
    }
}
```
which basically says redirect www.adobe.com to origin (main--cc--adobecom.aem.page) on aem.page and www.stage.adobe.com but leave it as it is on aem.live.
This method is called from loadArea -> decorateSections -> decorateLinks and I don't see how I can change it to fix our problem and not affect anything else. Should/can we leave it like this? It seems it works like this for a long time already.

bozojovicic:
@3ch023 I presented this change on Thursday's demo meeting and Geri says that GWPs are aware of the problem with Dexter modals on aem.page and it is OK for them since they are testing all this on stage.adobe.com where it should work since we have all these Dexter modals also on stage.adobe.com so I think we are fine.

PR #4396, opened 2025-06-13
MWPW-173815: Merch addon TRIAL fix & style update

st-angelo-adobe:
Resolves: [MWPW-173815](https://jira.corp.adobe.com/browse/MWPW-173815), [MWPW-176314](https://jira.corp.adobe.com/browse/MWPW-176314)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/nala/features/commerce/plans?martech=off
- After: https://mwpw-173815--milo--adobecom.aem.page/drafts/nala/features/commerce/plans?martech=off


libs/features/mas/src/merch-addon.js
```diff
@@ -63,31 +63,65 @@ export default class MerchAddon extends LitElement {
...
 
     static styles = css`
         :host {
-            display: flex;
-            gap: 9px;
-            align-items: start;
-            cursor: pointer;
+            --merch-addon-gap: 10px;
+            --merch-addon-checkbox-size: 12px;
+            --merch-addon-checkbox-border: 2px solid rgb(117, 117, 117);
```
Axelcureno, line 84: This won't work on dark mode, consider using a spectrum var.

Axelcureno (dismissed):
I know plans will be in acom which doesn't support dark mode, but we should still test addon in dark mode and make sure it renders properly (at least the checkbox should be legible

libs/features/mas/src/merch-addon.js
st-angelo-adobe: I checked the Figma, this was slightly off, so I changed it, but overall this is what the exact shade is in the Figma.
If as you said below, we don't need to worry about dark mode in acom, I'd just leave it like this, there are several other areas where we hardcode color values.

libs/features/mas/src/merch-addon.js
```diff
@@ -18,8 +18,8 @@ export default class MerchAddon extends LitElement {
     getOsi(planType, offerType) {
       const offerTypeOptions = {
         'TRIAL': ['TRIAL'],
-        'BASE': ['BASE', 'PROMOTION'],
-        'PROMOTION': ['PROMOTION', 'BASE']
+        'BASE': ['BASE', 'PROMOTION', 'TRIAL'],
+        'PROMOTION': ['PROMOTION', 'BASE', 'TRIAL']
```
3ch023, line 22: could you add this use case to unit tests pls

3ch023 (changes requested):
hi, looks good but found 1 small issue. if i select/unselect checkbox in modal and close modal - card checkbox is not updated.
it needs to communicate back to the card similar way as implemented on addon.
@mirafedas and @markovicpredrag1209 know more details on it

bozojovicic (changes requested):
Sorry but I see several issues here. We have the method `adaptForMobile()` in https://github.com/adobecom/milo/blob/stage/libs/features/mas/src/variants/plans.js#L86 which places the addon checkbox in the right place for Default/wide/super-wide cards. But after the "stock checkbox" is renamed, the selector used in that method should be renamed as well. So all you need to do is to change `#stock-checkbox` with `slot[name=addon]` in 2 places in that method. You can revert other changes. This method also handles situations in Mas studio when you change the card size from Default to Wide and back. And of course it places it in the right place on Mobile screen. On your test page I see it is in the wrong place on mobile.

On more thing, did you move the callout to the footer on super-wide card? I don't see that AC in the ticket. Can you prepare some test page with this example? With longer callout text this can look pretty bad. I tried to reject this change but it seems that now they asked this from you.

st-angelo-adobe:
@3ch023 Regarding the unckeck in the modal part, for some offers it seems like the 3in1 close event doesn't have the data that's needed. 
@markovicpredrag1209 said he would investigate this further 🙏 , so I think for this task we can leave this out (it's not an AC either).

3ch023 (approved):

bozojovicic (approved):

st-angelo-adobe:
1. @markovicpredrag1209 said there is this Jira open for the commerce team, seems to be an issue on their side for some of the offers: https://jira.corp.adobe.com/browse/ECOMM-152112
2. Fixed
3. Fixed (?) but check this: https://mwpw-173815--milo--adobecom.aem.page/drafts/astatescu/plans/individual-addon-placement. Doesn't really work with the quantity selector as well, and that would be harder to fix. I'll ask Lucy if the scenario is relevant and if they should still be aligned then (probably yes but still)
4. I see that from time to time in different pages.. I'm guessing it's because the font hasn't loaded with all weights maybe? Either way I don't think this has ever affected production, I would skip it here, and maybe look into it more when it can be reproduced (I could not reproduce it)

@afmicka

afmicka:
@st-angelo-adobe it diturb quantity selector alignment without stock in the cards. Business tab.
https://main--cc--adobecom.aem.live/creativecloud/plans?milolibs=mwpw-173815
Screenshot 2025-07-14 at 16 25 21

without your changes:
Screenshot 2025-07-14 at 16 26 14

st-angelo-adobe:
@afmicka Fixed. I moved the checkbox below the selector. Basically the checkboxes need to always be aligned at the bottom, the selectors not necessarily - in case there's no stock addon I mean.

But according to Lucy, the addon + selector is not a real requirement right now anyway.

afmicka:
@st-angelo-adobe these changes messed up AI Assistanc on DC page 
https://main--dc--adobecom.aem.live/acrobat/features?milolibs=mwpw-173815
Screenshot 2025-07-15 at 09 31 56

st-angelo-adobe:
@afmicka True, hadn't considered genai.

The plans checkbox is custom, so I modified the `merch-addon` to be "custom" from the get go, but broke the rest. I returned to the old implementation and just gave the possibility to override with a custom checkbox - now being used by plans.

afmicka:
@st-angelo-adobe @3ch023 @npeltier this looks ok to me. Could someone please dismiss @Axelcureno change request so that this can be merged to the current Stage to Main batch?

PR #4397, opened 2025-06-13
Exclude zero impact count

skholkhojaev:
Exclude zero-impact PRs from count

**Test URLs:**
- After: https://exclude-zero-impact-count--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/.github/workflows/merge-to-stage.js
+++ b/.github/workflows/merge-to-stage.js
@@ -140,8 +140,10 @@ const merge = async ({ prs, type }) => {
           merge_method: 'squash',
         });
       }
-      existingPRCount++;
-      console.log(`Current number of PRs merged: ${existingPRCount}`);
+      if (type !== LABELS.zeroImpact) {
+        existingPRCount++;
+      }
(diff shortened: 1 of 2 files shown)


mokimo (commented):
This has a soft dependency on https://github.com/adobecom/milo/pull/4026
Soft as in, we can already merge this - but should account for PRs having a certain prefix in the description.


Further down in this file, L219 we count the merge limit if the batch already exists:
```js
    // Regex to be adapted to account for zero impact PRs
    existingPRCount = body.match(/https:\/\/github\.com\/adobecom\/milo\/pull\/\d+/g)?.length || 0;
    // Or if it's easier body.match(/https:\/\/github\.com\/adobecom\/milo\/pull\/\d+/g)?.filter(match => !match.includes("[ZERO-IMPACT]:")).length 
    console.log(`Number of PRs already in the batch: ${existingPRCount}`);
```

PR #4026 will need to prefix zero-impact PRs so that we can recognize them here. So like `[ZERO-IMPACT]: PR Title` and  then we can ensure in regex that we don't match anything that has this prefix

mokimo (approved):

PR #4405, opened 2025-06-16
MWPW-174568: [Preflight] Metadata on hover overlaps with navigation bar

skholkhojaev:
Since the navigation bar had a `z-index: 10`, i lowered the `z-index:` of the Metadata on hover so that it wouldnt show on top of the navigation bar when scrolling up

Resolves: [MWPW-174568](https://jira.corp.adobe.com/browse/MWPW-174568)

**Test URLs:**
- Before: https://main--bacom--adobecom.aem.page/sg/products/genstudio-for-performance-marketing
- After: https://mwpw-174568-metadata-overlaping--milo--skholkhojaev.aem.page/ 

**TEST Link:**
-  https://main--bacom--adobecom.aem.page/sg/products/genstudio-for-performance-marketing?milolibs=MWPW-174568-Metadata-overlaping--milo--skholkhojaev

### DIFF
--- a/libs/blocks/preflight/preflight.css
+++ b/libs/blocks/preflight/preflight.css
@@ -324,7 +324,7 @@ span.preflight-time {
 .tooltip:hover::after {
   opacity: 1;
   visibility: visible;
-  z-index: 9;
+  z-index: 3;
 }
 
 /* SEO */
@@ -822,7 +822,7 @@ img:hover ~ .picture-meta {
   position: fixed;
   border: 2px solid red;
   background-color: white;
-  z-index: 103;
+  z-index: 9;
   overflow: hidden;
   pointer-events: none;
   display: block;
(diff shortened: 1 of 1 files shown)


libs/blocks/preflight/preflight.css
```diff
@@ -742,7 +742,7 @@ picture:has(.picture-meta) {
 picture:hover .picture-meta,
 img:hover ~ .picture-meta {
   opacity: 1;
-  z-index: 19;
+  z-index: 9;
```
mokimo, line 745: Gave it a [quick check](https://github.com/search?q=repo%3Aadobecom%2Fmilo%20z-index&type=code), seems like chart/list.css has a z-index of 10 for buttons? Ah our z-index strategy is a mess.
Further below in the file, we have a z-index of 103... should that be 9/10 as well?

mokimo: Can you also change the z-index for the performance LCP modal?

PR #4413, opened 2025-06-16
Rolling importer fixes

mokimo:
### Description
Due to network flakiness, a few requests can timeout leading to articles not being successfully imported. We can have some very simplified "retry" logic here, to retry once, which should bring down failures to close to 0. It's manual effort for me to import any straddlers, hopefully can be avoided with this already.

Also added a from-to date for easier tracking of the exact calls we make to the AEM APIs and allowing to set more params locally to simulate a workflow run for debugging purposes.

**Test URLs:**
- After: https://rolling-importer-fixes--milo--mokimo.aem.page/?martech=off


.github/workflows/import/localPathsToImport.js
```diff
@@ -0,0 +1,3 @@
+export default [
```
mokimo, line 1: Compared to localPathsToImport, which actually, when set, would import those specific paths.

overmyheadandbody: Would it make sense to have these notes as a comment in the file itself, so whoever runs the script knows which to use?

.github/workflows/import/poll-logs.js
```diff
@@ -207,26 +208,31 @@ const saveLivePaths = (livePaths) => {
 
 async function main() {
   await getImsToken();
-  const entries = await fetchLogsForSite(
+  const TO_PARAM = process.env.LAST_RUN_ISO_TO || new Date().toISOString();
+  if(localPathsToImport.length) console.log("Importing paths from local environment");
```
overmyheadandbody, line 212: OCD:
```suggestion
  if (localPathsToImport.length) console.log("Importing paths from local environment");
```

.github/workflows/import/poll-logs.js
```diff
@@ -237,20 +243,36 @@ async function main() {
             console.log(
               `Progress: Success: ${result.success} | Error: ${result.error}`
             );
-            
         })
-        .catch(() => {
+        .catch((e) => {
```
overmyheadandbody, line 247: The linter inside of me is asking: doesn't this lead to a "defined but never used" type of linter error?

.github/workflows/import/localPathsToImport.js
narcis-radu: I'm still not clear about the difference between the two files. I see localPathsToImport is used to read from it, while importingPaths is used to write in it.

.github/workflows/import/poll-logs.js
mokimo: Removed

.github/workflows/import/poll-logs.js
mokimo: Ughh. For some reason the linter doesn't scan the dotfiles such as `.github` ... I don't think we can overwrite that. Thanks for the manual linting hawk eyes though 😁

.github/workflows/import/localPathsToImport.js
mokimo: I tried to add a more elaborate explanation

PR #4439, opened 2025-06-19
[MWPW-174802] - Carousel items align height

DKos95:
This PR adds an option to align your items by height by adding `align-height` class to the carousel. 

Resolves: [MWPW-174802](https://jira.corp.adobe.com/browse/MWPW-174802)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/rclayton/carousel-videos?martech=off
- After: https://carousel-item-height-align--milo--adobecom.aem.page/drafts/rclayton/carousel-videos?martech=off

**Kitchen sink Test URLs:**
- Before: https://main--milo--adobecom.aem.page/docs/library/kitchen-sink/carousel
- After: https://carousel-item-height-align--milo--adobecom.aem.page/docs/library/kitchen-sink/carousel

**Example where `align-height` is applied to 1st but not to 2nd carousel**
- Before: https://main--milo--adobecom.aem.page/drafts/dusan/carousel-align-height-feature
- After: https://carousel-item-height-align--milo--adobecom.aem.page/drafts/dusan/carousel-align-height-feature

### DIFF
--- a/libs/blocks/carousel/carousel.js
+++ b/libs/blocks/carousel/carousel.js
@@ -513,13 +513,9 @@ export default function init(el) {
     });
 
     videoData.forEach(({ video, foreground, videoHeight }) => {
-      if (foreground) {
-        video.style.height = `${videoHeight}px`;
-        video.style.maxHeight = `${videoHeight}px`;
-      } else {
-        video.style.height = `${maxHeight}px`;
-        video.style.maxHeight = `${maxHeight}px`;
-      }
+      const finalHeight = foreground ? videoHeight : maxHeight;
+      video.style.height = `${finalHeight}px`;
+      video.style.maxHeight = `${finalHeight}px`;
     });
   }
 


zagi25:
It looks like the update doesn't cover all cases
Screenshot 2025-06-20 at 11 19 48

stereoke:
> This feature doesn't look to be an opt in feature. Is that as planned? This will potentially change the look of any live page if it's not opt in.
> 
> cc: @narcis-radu @overmyheadandbody

I'm going to suggest that our Design Publisher take this question for confirmation in the authoring community. My thought is that: opt in makes sense if we’re “enhancing” the block. But if we’re solving a “problem” that GWP encounters…then change the default.

Let me confirm before we merge.

libs/blocks/carousel/carousel.js
```diff
@@ -496,6 +496,41 @@ export default function init(el) {
   nextPreviousContainer.append(...nextPreviousBtns, controlsContainer);
   el.append(nextPreviousContainer);
 
+  function normalizeVideoHeights() {
+    const videos = el.querySelectorAll('video');
+    if (videos.length === 0) return;
```
rgclayton, line 501: Nit. This could probably shortened to `if (!videos) return`;

libs/blocks/carousel/carousel.js
```diff
@@ -496,6 +496,41 @@ export default function init(el) {
...
+        video.style.maxHeight = `${videoHeight}px`;
+      } else {
+        video.style.height = `${maxHeight}px`;
+        video.style.maxHeight = `${maxHeight}px`;
+      }
+    });
+  }
+
+  setTimeout(normalizeVideoHeights, 100);
+
```
rgclayton, line 533: Nit: I was thinking this could be a bit more efficient/simple. Possibly something like this

```suggestion
function normalizeVideoHeights() {
  const videos = el.querySelectorAll('video');
  if (videos.length === 0) return;

  const videoData = [];
  let maxHeight = 0;

  // Collect data and calculate maxHeight
  videos.forEach((video) => {
    const foreground = video.closest('.editorial-card')?.querySelector('.foreground');
    const videoHeight = video.offsetHeight;
    const foregroundHeight = foreground ? foreground.offsetHeight : 0;
    const totalHeight = videoHeight + foregroundHeight;
    videoData.push({ video, foreground, videoHeight, totalHeight });
    if (totalHeight > maxHeight) maxHeight = totalHeight;
  });

  // Apply heights
  videoData.forEach(({ video, foreground, videoHeight }) => {
    if (foreground) {
      video.style.height = `${videoHeight}px`;
      video.style.maxHeight = `${videoHeight}px`;
    } else {
      video.style.height = `${maxHeight}px`;
      video.style.maxHeight = `${maxHeight}px`;
    }
  });
}
```

rgclayton (approved):
Nice work @DKos95. Just a couple nits from me. I think we should wait to hear what @stereoke reports back about either adding a new variant for this or go with changing the default.

libs/blocks/carousel/carousel.js
DKos95: Did give it a try but if there are no videos an empty Node list gets returned, so if we would do `if (!videos) return` it would pass, however `!videos.length` should work.

libs/blocks/carousel/carousel.js
DKos95: Thanks for the optimization advice!

DKos95:
> Nice work @DKos95. Just a couple nits from me. I think we should wait to hear what @stereoke reports back about either adding a new variant for this or go with changing the default.

Hey @rgclayton, Danielle has notified me that this would be best if it could be an opt-in feature. I have added a feature where when you add `align-height` as a carousel class the code will execute and since the context has changed a lot in this regard I think it's best to re request a review from you.

rgclayton (approved):
The new variant update looks good. Everything worked as expected on testing pages.

libs/blocks/carousel/carousel.js
```diff
@@ -496,6 +496,34 @@ export default function init(el) {
   nextPreviousContainer.append(...nextPreviousBtns, controlsContainer);
   el.append(nextPreviousContainer);
 
+  function normalizeVideoHeights() {
```
overmyheadandbody, line 499: If the window is resized, the video parents might also get resized, thus the normalized video height might no longer apply. Should this be run on resize (debounced) as well?

libs/blocks/carousel/carousel.js
```diff
@@ -496,6 +497,36 @@ export default function init(el) {
...
+      } else {
+        video.style.height = `${maxHeight}px`;
+        video.style.maxHeight = `${maxHeight}px`;
+      }
+    });
+  }
+
+  if (el.classList.contains('align-height')) setTimeout(normalizeVideoHeights, 100);
+
+  window.addEventListener('resize', debounce(() => { if (el.classList.contains('align-height')) normalizeVideoHeights(); }));
```
overmyheadandbody, line 528: Default debounce timeout is `300ms`, while the `setTimeout` from L526 is `100ms`, should these be consolidated?

libs/blocks/carousel/carousel.js
```diff
@@ -496,6 +497,36 @@ export default function init(el) {
...
+    });
+
+    videoData.forEach(({ video, foreground, videoHeight }) => {
+      if (foreground) {
+        video.style.height = `${videoHeight}px`;
+        video.style.maxHeight = `${videoHeight}px`;
+      } else {
+        video.style.height = `${maxHeight}px`;
+        video.style.maxHeight = `${maxHeight}px`;
+      }
```
overmyheadandbody, line 522: Micro-optimization idea:
```suggestion
      const finalHeight = foreground ? videoHeight : maxHeight;
      video.style.height = `${finalHeight}px`;
      video.style.maxHeight = `${finalHeight}px`;
```

overmyheadandbody (commented):
Looking good overall, just some final nits + the failing PSI checks, then I will happily approve this

libs/blocks/carousel/carousel.js
DKos95: 100ms is how much time is needed for all videos to be rendered and available in `normalizeVideoHeights` function, while the 300ms for debounce is how much time needs to pass after user stops resizing for this to trigger, I don't see a reason why we would need to make them the same since when the user resizes the videos will already be loaded on the screen, the timers have different purposes.

overmyheadandbody (approved):
I see the PSI check is still failing, some updates might be needed to the test page(s).

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-174802

PR #4450, opened 2025-06-23
[MWPW-174959] - Table border fix

DKos95:
Resolves: [MWPW-174959](https://jira.corp.adobe.com/browse/MWPW-174959)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/dusan/table-in-tab-expand?martech=off
- After: https://table-border-fix--milo--adobecom.aem.page/drafts/dusan/table-in-tab-expand?martech=off

**Table Kitchen sink Test URLs:**
- After: https://table-border-fix--milo--adobecom.aem.page/docs/library/kitchen-sink/table?martech=off

### DIFF
--- a/libs/blocks/table/table.css
+++ b/libs/blocks/table/table.css
@@ -149,22 +149,24 @@
   border-radius: 0 !important;
 }
 
-.table .col-heading:first-child {
+.table .left-round {
   border-top-left-radius: var(--border-radius);
-}
-
-[dir="rtl"] .table .col-heading:first-child {
-  border-top-left-radius: 0;
-  border-top-right-radius: var(--border-radius);
-}
-
-.table .col-heading:last-child {
-  border-top-right-radius: var(--border-radius);
-}
-
-[dir="rtl"] .table .col-heading:last-child {
   border-top-right-radius: 0;
+}
+
+.table .right-round {
+  border-top-right-radius: var(--border-radius);
+  border-top-left-radius: 0;
+}
+
+[dir="rtl"] .table .left-round {
+  border-top-right-radius: var(--border-radius);
+  border-top-left-radius: 0;
+}
+
+[dir="rtl"] .table .right-round {
   border-top-left-radius: var(--border-radius);
+  border-top-right-radius: 0;
 }
 
 .table .row-heading .col-heading.col.no-rounded,
@@ -396,7 +398,7 @@
 /* rtl */
 [dir="rtl"] .table .col {
   border-right: 1px var(--border-color) solid;
-  border-left: 1px solid transparent;
+  border-left: 1px var(--border-color) solid;
 }
 
 [dir="rtl"] .table.merch .col {
--- a/libs/blocks/table/table.css
+++ b/libs/blocks/table/table.css
@@ -663,23 +663,13 @@
   .table.merch {
     margin: 0 30px;
   }
-
-  .table .left-round {
-    border-top-left-radius: var(--border-radius);
-    border-top-right-radius: 0;
-  }
   
-  .table .right-round {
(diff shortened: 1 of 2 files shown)


libs/blocks/table/table.css
```diff
@@ -664,6 +664,26 @@
...
+  
+  .table .right-round {
+    border-top-right-radius: var(--border-radius);
+    border-top-left-radius: 0;
+  }
+  
+  [dir="rtl"] .table .left-round {
+    border-top-right-radius: var(--border-radius);
+    border-top-left-radius: 0;
+  }
```
overmyheadandbody, line 680: Why not combine these, since they declare the same thing?
```suggestion
  .table .right-round,
  [dir="rtl"] .table .left-round {
    border-top-right-radius: var(--border-radius);
    border-top-left-radius: 0;
  }
```
Same for the other two new declarations.

libs/blocks/table/table.js
```diff
@@ -517,7 +517,7 @@ function applyStylesBasedOnScreenSize(table, originTable) {
       const filters = Array.from(table.parentElement.querySelectorAll('.filter')).map((f) => parseInt(f.value, 10));
       const rows = table.querySelectorAll('.row');
 
-      table.querySelectorAll('.hide-mobile, .force-last').forEach((col) => { col.classList.remove('hide-mobile', 'force-last'); });
+      table.querySelectorAll('.hide-mobile, .force-last, .rounded-left, .rounded-right').forEach((col) => { col.classList.remove('hide-mobile', 'force-last', 'rounded-left', 'rounded-right'); });
```
overmyheadandbody, line 520: I'm not sure I understand the purpose of `.rounded-left` and `.rounded-right`. I see you've removed the code that removes those classes, but I don't get where they're added to begin with. Are they used in authoring? Is their removal still useful here?

libs/blocks/table/table.js
```diff
@@ -531,8 +531,17 @@ function applyStylesBasedOnScreenSize(table, originTable) {
       }
 
       rows.forEach((row) => {
+        const firstFilterCol = row.querySelector(`.col-${filters[0] + 1}`);
         const secondFilterCol = row.querySelector(`.col-${filters[1] + 1}`);
-        if (secondFilterCol) secondFilterCol.classList.add('force-last');
+
+        if (firstFilterCol && firstFilterCol.classList.contains('col-heading')) {
```
overmyheadandbody, line 537: Is optional chaining not an option?
```suggestion
        if (firstFilterCol?.classList.contains('col-heading')) {
```
Same for L541

overmyheadandbody (commented):
Always confusing to look at tables 😅 I just have a few notes

libs/blocks/table/table.js
overmyheadandbody: I get the intent, but I still don't get the actual removed classes. You're removing `.rounded-left` and `.rounded-right`, but you never add these classes, nor are they defined in CSS. You do define `.left-round` and `.right-round`, however. Why would we need both `.rounded-left` and `.left-round`? If they are in fact both needed, then a better naming convention should be used, to make the purpose of each clear.

DKos95: Ah sorry I didn't see the difference in the class names at first, thanks for noticing. I also updated the PR to cover some additional use cases.

DKos95:
> Looks good overall, but on the Kitchen Sink I did notice that the first table has rounded corners on the `main` branch, square ones on the feature branch. That table already has a funky layout, but it might be worth looking into.

Hey @overmyheadandbody thank you for noticing this, for me this was only happening on mobile. I have since updated the solution to cover this situation, thanks!

PR #4458, opened 2025-06-24
Table flaky/failing tests fix

DKos95:
Fixes flaky/failing table tests introduced in https://github.com/adobecom/milo/pull/4374

**Test URLs:**
- After: https://table-flaky-tests-fix--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/libs/blocks/table/table.js
+++ b/libs/blocks/table/table.js
@@ -256,6 +256,8 @@ function handleExpand(e) {
 }
 
 function setExpandEvents(el) {
+  if (el.dataset.expandEvents) return; // prevent multiple calls in unit tests
+  el.dataset.expandEvents = 'true';
   el.querySelectorAll('.icon.expand').forEach((icon) => {
     icon.parentElement.classList.add('point-cursor');
     icon.parentElement.addEventListener('click', () => handleExpand(icon));
--- a/libs/blocks/table/table.js
+++ b/libs/blocks/table/table.js
@@ -256,8 +256,6 @@ function handleExpand(e) {
 }
 
 function setExpandEvents(el) {
-  if (el.dataset.expandEvents) return; // prevent multiple calls in unit tests
-  el.dataset.expandEvents = 'true';
   el.querySelectorAll('.icon.expand').forEach((icon) => {
     icon.parentElement.classList.add('point-cursor');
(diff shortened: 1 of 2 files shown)


libs/blocks/table/table.js
```diff
@@ -245,6 +245,8 @@ function handleExpand(e) {
 }
 
 function setExpandEvents(el) {
+  if (el.dataset.expandEvents) return; // prevent multiple calls in unit tests
```
mokimo, line 248: If the unit tests are failing, shouldn't we adapt them - rather than the table?

DKos95: That would be ideal and I did try but to no avail, as soon as I fix this, another issue pops up in the table.test for which initializing the table multiple times is needed, so this is the easiest solution and this was already implemented before in another manner but we removed it because it was causing an issue with the variable solution in the PR that I linked.

DKos95: If the team is ok with me spending more time on this I am up for it, but I thought it would be best to fix the failing tests because they are on stage and we already had a report for this in the milo-dev channel.

test/blocks/table/table.test.js
```diff
@@ -12,7 +12,7 @@ setConfig(conf);
 const config = getConfig();
 
 describe('table and tablemetadata', () => {
-  beforeEach(() => {
+  before(() => {
```
mokimo, line 15: To me this looks like it builds up test-state and is an anti-pattern. If we depend on whatever _one_ test has done - it's likely wrong.
Maybe we should just wipe the `innerHTML` and recreate the markup on each run - to properly isolate each test?

mokimo: I think it should stay a beforeEach and l7 should be moved into the `beforeEach` block

mokimo: PS: If the tests already been written in a way, that they depend on whatever the test before has done - Good luck 😆

DKos95: Tried adding 2 tables on 1 page and added a console log but it only attached 2 times https://table-flaky-tests-fix--milo--adobecom.aem.page/drafts/dusan/check-multiple-tables.

overmyheadandbody (approved):

PR #4465, opened 2025-06-25
MWPW-174164: Collection header & UI fixes

st-angelo-adobe:
### NOTE: CAN'T be merged before https://github.com/adobecom/cc/pull/758 is merged

Extracted the collection header functionality into a `merch-card-collection-header` component.

How it works:
- It's instantiated by the collection autoblock
- `{variant}-container` class became 2 classes `collection-container {variant}`, with `collection-container` being a `grid` which displays the sidenav on the side, and the header and collection stacked - any variant specific style changes in the future can be done via the `{variant}` class
- The header itself is a `grid` with a layout secified with `grid-template-areas` and `grid-template-columns` - the areas are `search`, `filter`, `sort`, `result`, `custom`
- `grid-template-areas` and `grid-template-columns` can be set with css variables for each screen size, `mobile`, `tablet`, `desktop`, overriding some defaults, for example:

```
merch-card-collection-header.plans {
    --merch-card-collection-header-columns: 1fr fit-content(100%);
    --merch-card-collection-header-areas: "result filter";
}

@media screen and ${DESKTOP_UP} {
  merch-card-collection-header.plans {
      --merch-card-collection-header-columns: fit-content(100%);
      --merch-card-collection-header-areas: "custom";
  }
}
```
- `custom` is basically a custom area in the header, configured in the variant class as a static property, `collectionOptions` - this was introduced for maximum flexibility (for plans, it's used in the desktop size for the filter label - previously it was implemented in the `result` area - viable, but I went a step further)
- To hide areas for certain screen sizes, also use `collectionOptions` - see `plans.js` as an example

Resolves: [MWPW-174164](https://jira.corp.adobe.com/browse/MWPW-174164), [MWPW-175520](https://jira.corp.adobe.com/browse/MWPW-175520), [MWPW-175406](https://jira.corp.adobe.com/browse/MWPW-175406), [MWPW-175330](https://jira.corp.adobe.com/browse/MWPW-175330)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/npeltier/plans?martech=off
- After: https://mwpw-174164--milo--adobecom.aem.page/drafts/npeltier/plans?martech=off


libs/features/mas/src/variants/catalog.css.js
```diff
@@ -1,43 +1,27 @@
...
-    }
+.collection-container.catalog merch-sidenav {
+    --merch-sidenav-gap: 10px;
+}
 
-    .three-merch-cards.catalog,
-    .four-merch-cards.catalog {
-        grid-template-columns: repeat(3, var(--consonant-merch-card-catalog-width));
+@media screen and ${TABLET_UP} {
+    .collection-container.catalog {
```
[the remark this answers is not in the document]
st-angelo-adobe: Let's discuss in tomorrow's call.

libs/blocks/merch-card-collection-autoblock/merch-card-collection-autoblock.js
```diff
@@ -4,25 +4,22 @@ import { postProcessAutoblock } from '../merch/autoblock.js';
 import '../../deps/mas/merch-card.js';
 import '../../deps/mas/merch-quantity-select.js';
 
-const COLLECTION_AUTOBLOCK_TIMEOUT = 5000;
+const DEPS_TIMEOUT = 5000;
```
yesil, line 7: 5s might be a bit short with really slow connections.

libs/features/mas/src/constants.js
```diff
@@ -82,6 +82,10 @@ export const EVENT_MERCH_SEARCH_CHANGE = 'merch-search:change';
 
 export const EVENT_MERCH_CARD_COLLECTION_SORT = 'merch-card-collection:sort';
 
+export const EVENT_MERCH_CARD_COLLECTION_LITERALS_CHANGED = 'merch-card-collection:literals-changed';
```
yesil, line 85: nit: merch-card-collection specific stuff could stay in merch-card-collection.js and be exported/imported from there.

yesil (approved):

libs/features/mas/src/constants.js
st-angelo-adobe: I think it's better for these to be in a common place, because `merch-card-collection-header.js` needs them, but the collection needs the header, so they end up importing each other. The merch card ones are also in `constants.js`, so these can stay here as well imo.

st-angelo-adobe:
> @st-angelo-adobe is plans branch as target intentional?

@yesil It was at first, but I guess we will merge the `plans` branch as is, so I changed this one to target `stage`.

afmicka:
@st-angelo-adobe @3ch023 Issues seen with this PR:
Catlog page:
https://mwpw-174164--cc--adobecom.aem.live/products/catalog?martech=off&milolibs=mwpw-174164

1. Even though it was required for plans page on mobile view to close the Filter modal when any filter option is closed, that is not desired behaviour for catalog page. Filters on catalog do NOT close and it has multiple choice and has more levels within categories. Right now, if user clicks lets say on Marketing and Commerce, the modal closes immediately and if user want to filter deeper down within the category, needs to open the modal again. The same if user wants to add additional filer for the Device type, will have to open the modal the third time :)  (this might have been fixed while validating and fixing this comment but needs checking)

2. when there is no results in search, the view breaks down on mobile 
Screenshot 2025-07-11 at 13 28 56

3. The font for search results is much smaller on mobile and spacing between search field and filter is much smaller. Was that required? 
Screenshot 2025-07-11 at 13 28 24
Screenshot 2025-07-11 at 13 28 38

Plans page:
https://mwpw-174164--cc--adobecom.aem.live/creativecloud/plans?martech=off&milolibs=mwpw-174164#filter=all
1. On tablet/mobile view  on prod the filters button changes the color if any filter is selected (indicating that there is a selection). Also the tab is dark color. The filter modal itself should not have the "close" button (as it has no function with the requirement to close automatically when selection is made). Also there should not be label "Categories" in the filters modal
Screenshot 2025-07-11 at 13 52 51

prod: Screenshot 2025-07-11 at 13 52 58
this PR: 
Screenshot 2025-07-11 at 13 53 31

2. I think we need to communicate the tablet view with GWP. We do have the example in the figma but the comment said "t should drop to tablet width like on the current a.com Plans page business tab" which currently means all card become wide and are show
Screenshot 2025-07-11 at 14 02 51
n one blow another 

cc: @Roycethan

st-angelo-adobe:
@afmicka @3ch023 Actually for most of these we should follow the Figma: https://www.figma.com/design/32yMtkQsTbmKBkECAlnvVy/Plans-and-Catalog?node-id=366-99318&m=dev

I doubt we need to take into account the live page for styling, so I will handle the following:
**Catalog:**
- The filter modal autoclose on select problem (actually already fixed)
- The layout breakdown for no search result in mobile
**Plans**
- I will remove the close button on the filters modal, it's indeed not present in the Figma: https://www.figma.com/design/32yMtkQsTbmKBkECAlnvVy/Plans-and-Catalog?node-id=571-22781&m=dev

Roycethan:
@st-angelo-adobe Looks like the spacing is not per figma between Filters and Sort buttons:
Expect:
image

Actual:
image

PR #4493, opened 2025-06-30
MWPW-175097: Enable preview for stage.adobe.com

markovicpredrag1209:
* enable preview on stage.adobe.com

can only be verified via browser overrides, since its only reproducible on www.stage
(https://www.stage.adobe.com/cc-shared/fragments/drafts/mariia/document)

Resolves: [MWPW-175097](https://jira.corp.adobe.com/browse/MWPW-175097)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.live/drafts/nala/features/commerce/plans?martech=off&georouting=off
- After: https://MWPW-175097--milo--markovicpredrag1209.aem.live/drafts/nala/features/commerce/plans?martech=off&georouting=off

### DIFF
--- a/libs/blocks/merch/merch.js
+++ b/libs/blocks/merch/merch.js
@@ -671,7 +671,7 @@ export async function getCheckoutAction(offers, options, imsSignedInPromise, el)
 
 export function setPreview(attributes) {
   const { host } = window.location;
-  if (host.includes(`${SLD}.page`) || host.origin === 'https://www.stage.adobe.com') {
+  if (host.includes(`${SLD}.page`) || host === 'www.stage.adobe.com') {
     attributes.preview = 'on';
   }
 }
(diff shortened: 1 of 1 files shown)


libs/blocks/merch/merch.js
```diff
@@ -671,7 +671,7 @@ export async function getCheckoutAction(offers, options, imsSignedInPromise, el)
 
 export function setPreview(attributes) {
   const { host } = window.location;
-  if (host.includes(`${SLD}.page`) || host.origin === 'https://www.stage.adobe.com') {
+  if (host.includes(`${SLD}.page`) || host === 'www.stage.adobe.com') {
```
mokimo, line 674: Is this used on business.adobe.com as well? Might as well cut the www. part

3ch023: bacom doesn't use merch prices afaik.
however i still like the idea to cut the www. part, it shouldn't hurt and we would be future proof in case bacom will start using our pricing

mokimo: Ye I agree with CodeQL as well (& in fact, already assumed codeQL would complain), though you could use something like `endsWith(.stage.adobe.com)` 
See https://codeql.github.com/codeql-query-help/python/py-incomplete-url-substring-sanitization/

PR #4496, opened 2025-07-01
[MWPW-172587] Preflight check status

skholkhojaev:
Now Preflight will be ran automatically when the page is loaded (once the `utils.js` is loaded).

After Preflight is done running in the background, it will save its result as cache and when the User clicks on `tools`>`preflight` to open up the Preflight Panel then the checks first look for already existing caches and if none were found they will then be ran manually.

Also added a notification bar to inform the Athors of any failling checks, The Athors can click on the `review` button to open up the Preflight Panel and get more details.

Resolves: [MWPW-172587](https://jira.corp.adobe.com/browse/MWPW-172587)
**Test URLs:**
- After: https://Preflight-CheckStatus--milo--skholkhojaev.aem.page/?martech=off


skholkhojaev:
> There is no ASO, let's not write any code for something we don't have yet

@narcis-radu @robert-bogos 
I am down to remove the ASO for now until its implemnted but i want to hear your confirmation and opinion about it.

Because we talked about it in our sync that there should be a "config type" to determain which Prefligh will be used and a "coverter" that would standardized the results.

mokimo (commented):
Let's also hear what Robert & Narcis have to say. But as an alternative, you could also raise the ASO specific implementation towards our ASO branch.

That way you can split this into "get the initial ticket & feature running"
and "make it compatible with the new feature (in it's own branch)"

... Having smaller PRs also makes this easier and more concise to review, focussing on the core bit - which is to guide authors with page health checks

libs/blocks/preflight/panels/assets.js
```diff
@@ -17,7 +21,19 @@ const viewportTooSmall = signal(isViewportTooSmall());
...
+    const cachedResults = await executePreflightChecks();
+    if (cachedResults?.assets) {
+      checks = cachedResults.assets;
+    } else {
+      checks = runChecks(window.location.pathname, document);
+    }
+  } catch (error) {
+    console.log('Failed to get cached preflight results, running checks directly:', error);
+    checks = runChecks(window.location.pathname, document);
+  }
```
mokimo, line 36: You've duplicated this logic (more than twice...) that's a very good sign it should get centralized somewhere.

We already have the "preflightApi" layer, so ideally, u'd call `preflightApi.getResults`

libs/blocks/preflight/panels/assets.js
```diff
@@ -17,7 +21,19 @@ const viewportTooSmall = signal(isViewportTooSmall());
  * Runs asset checks and updates signals with the results.
  */
 async function getResults() {
-  const checks = runChecks(window.location.pathname, document);
+  let checks;
```
mokimo, line 24: You can pretty much keep the whole code as it is and just call
`const { isViewportTooSmall, runChecks } = await preflightApi.getResults();`

libs/utils/sidekick-decorate.js
```diff
@@ -1,9 +1,124 @@
...
+
+  observer.observe(document, {
+    attributes: true,
+    childList: true,
+    subtree: true,
+    attributeFilter: ['open'],
+  });
+
+  return observer;
+}
```
mokimo, line 121: all of this belongs in its own file

skholkhojaev:
overall thankyou for the review i'll go and change them @mokimo 👍

libs/utils/sidekick-decorate.js
```diff
@@ -1,9 +1,124 @@
...
+      await createPreflightNotification();
+    }
+  } catch (error) {
+    console.warn('Failed to check preflight status:', error);
+  }
+}
+
+function createSidekickVisibilityObserver() {
+  const observer = new MutationObserver(async () => {
+    const sidekick = document.querySelector('aem-sidekick');
```
narcis-radu, line 82: Suggestion: create a method that returns the sidekick (if available) and reuse it:
* https://github.com/adobecom/milo/pull/4496/files#diff-8fca44a004d5ac9b5fd40efac647ac0deb02ada1314e83f34593176290dcbb53R12
* https://github.com/adobecom/milo/pull/4496/files#diff-8fca44a004d5ac9b5fd40efac647ac0deb02ada1314e83f34593176290dcbb53R82
* https://github.com/adobecom/milo/pull/4496/files#r2180167699

libs/utils/utils.js
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable no-console */
+
```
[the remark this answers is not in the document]
skholkhojaev: ✅

libs/blocks/preflight/checks/preflightExecutor.js
```diff
@@ -0,0 +1,36 @@
...
+  preflightResults = getPreflightResults(url, area).then((results) => {
+    preflightResults = results.runChecks;
+    return results.runChecks;
+  });
+
+  return preflightResults;
+};
+
+export function hasPreflightFailures(results = null) {
+  const cache = results || preflightResults;
```
mokimo, line 22: don't u already cache it in the preflight API, why here as well?
also why do you need a method to executePreflightChecks if the API already has a getResults method

libs/blocks/preflight/panels/assets.js
```diff
@@ -17,7 +16,8 @@ const viewportTooSmall = signal(isViewportTooSmall());
  * Runs asset checks and updates signals with the results.
  */
 async function getResults() {
-  const checks = runChecks(window.location.pathname, document);
+  const results = await getPreflightResults(window.location.pathname, document);
+  const checks = results.runChecks.assets || [];
```
mokimo, line 20: looks a lot neater 👍

libs/blocks/preflight/panels/seo.js
```diff
@@ -90,12 +88,18 @@ async function getResults() {
     linksResult,
   ];
 
-  const checks = runChecks(window.location.pathname);
+  const results = await getPreflightResults(window.location.pathname, document);
+  const runChecks = results.runChecks.seo || [];
+
+  // Check if runChecks is available before proceeding
```
mokimo, line 94: why would it not be available? ur not doing this in other files

libs/utils/sidekick-decorate.js
```diff
@@ -122,6 +123,9 @@ export default async function stylePublish(sk) {
   if (stylePublishCalled) return;
   stylePublishCalled = true;
 
+  createSidekickVisibilityObserver();
```
mokimo, line 126: this should move into the checkPreflightAndShowNotification method imo

libs/utils/preflight-notification.js
```diff
@@ -0,0 +1,99 @@
...
+  if (sk) {
+    sk.dispatchEvent(new CustomEvent('custom:preflight', { bubbles: true }));
+  }
+}
+
+async function createPreflightNotification() {
+  const existingNotification = document.querySelector('.milo-preflight-overlay');
+  if (existingNotification) {
+    existingNotification.remove();
+  }
```
robert-bogos, line 17: nit: Usually, in Milo, we write single-statement if block in one line. Makes the code neater
```suggestion
  if (existingNotification) existingNotification.remove();
```

robert-bogos:
> > don't u already cache it in the preflight API, why here as well?
> > also why do you need a method to executePreflightChecks if the API already has a getResults method
> 
> the `preflightExecutor.js` was initially written with the assumption of ASO Api, but since I removed it, there isnt really a purpose for the file anymore besides two function which 100% could be handled over to `preflightApi.js`
> 
> im thinking of deleting the file and making things much simpler until we have ASO implemented.

IMO, we only have to gain by gradually refactoring Preflight to accommodate ASO. It saves us time later, and while a case can be made that ASO might not happen, I think it’s clear by now that it will.

Since there’s no immediate time pressure, we can advance now, unlike the possible tight deadlines when ASO integration becomes necessary. Also, since both Preflight and ASO are evolving in parallel, would help to build bridges (e.g., preflightExecutor.js) to keep them aligned.

robert-bogos:
After further discussion, the best way forward seems to be keeping `preflightApi.js` as the determiner of how the checks logic runs. So @skholkhojaev, you can go ahead and remove `preflightExecutor.js`. I’ll modify `preflightApi.js` later to accommodate the ASO integration when the time comes.

libs/blocks/preflight/checks/preflightApi.js
```diff
@@ -53,3 +55,46 @@ export default {
...
+    };
+  })();
+
+  preflightResults = await checks;
+
+  return {
+    isViewportTooSmall: isViewportTooSmall(),
+    runChecks: preflightResults,
+  };
+}
```
mokimo, line 88: Formatting is horrible, but I'd probably do it like this:
```suggestion
const runChecks = async () => {
      const assets = await Promise.all(runChecksAssets(url, area));
      const performance = await Promise.all(runChecksPerformance(url, area));
      const seo = await Promise.all(runChecksSeo({ url, area }));
      return {assets, performance, seo}
}

export async function getPreflightResults(url, area, useCache = true) {
  if (useCache) {
cachedChecks = checks || await runChecks()
  return {
    isViewportTooSmall: isViewportTooSmall(),
    runChecks: cachedChecks,
  };
  }
   const res = await runChecks()
    return {
      isViewportTooSmall: isViewportTooSmall(),
      runChecks: res,
    }
}
```

libs/blocks/preflight/checks/preflightApi.js
```diff
@@ -53,3 +55,46 @@ export default {
...
+  if (!preflightResults) return false;
+
+  const runChecks = preflightResults.runChecks || preflightResults;
+  const allResults = [
+    ...(runChecks.assets || []),
+    ...(runChecks.performance || []),
+    ...(runChecks.seo || []),
+  ];
+  return allResults.some((result) => result.status === 'fail');
+}
```
mokimo, line 100: That's wrong, you need to reliably return your results, you say it doesn't have any failures, just because u didn't run any checks.

libs/utils/preflight-notification.js
```diff
@@ -0,0 +1,106 @@
+import { hasPreflightFailures, getPreflightResults } from '../blocks/preflight/checks/preflightApi.js';
+import { loadStyle, getConfig } from './utils.js';
+
+let preflightNotificationDismissed = false;
+let observerCreated = false;
+
+function openPreflightPanel() {
+  const sk = document.querySelector('aem-sidekick, helix-sidekick');
+  if (sk) {
+    sk.dispatchEvent(new CustomEvent('custom:preflight', { bubbles: true }));
```
mokimo, line 10: is the preflight not open by default within the `sidekick.js` ? 🤔 
Do you decide to open the preflight for the user now? I don't quite get it

skholkhojaev: yes because if the user clicks on the `review` button, the preflight panel should be opened for them,

```
  const reviewLink = overlay.querySelector('.preflight-review-link');
  reviewLink.addEventListener('click', (e) => {
    e.preventDefault();
    e.stopPropagation();
    openPreflightPanel(); // will open the preflight
    overlay.remove();
  });
```

or am I misunderstanding your question?

skholkhojaev: to open up the preflight panel when the User clicks the `review` button 😅,

libs/utils/sidekick-decorate.js
```diff
@@ -122,6 +124,14 @@ export default async function stylePublish(sk) {
   if (stylePublishCalled) return;
   stylePublishCalled = true;
 
+  const config = getConfig();
+  config.onFooterReady = () => {
```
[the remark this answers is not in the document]
mokimo: I don't see the function being defined in bacom or CC
https://github.com/search?q=repo%3Aadobecom%2Fbacom%20onFooterReady&type=code - are you certain it's getting called?

libs/utils/preflight-notification.js
mokimo: Oooh now I get it, yeah makes sense! Thanks for the context

libs/utils/sidekick-decorate.js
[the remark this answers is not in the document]
skholkhojaev: the consumers import the `libs` from milo and run files such as `global-footer.js` which then that ends up running the function, and results in our callback rolling
https://github.com/adobecom/milo/blob/ae7abcc0781acc6e9856e6a972ed66515709aeb2/libs/blocks/global-footer/global-footer.js#L143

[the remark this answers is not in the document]
mokimo: Ah nvm, you assign it to be function... now I get it. That poses two questions though:

- Do you potentially accidentally overwrite this function for standalone consumers? Since it's not an array, but a single function.
- What about timing, the footer is async, could it be that the function is called before u define it and thus it never gets executed? What ensures that this doesn't happen?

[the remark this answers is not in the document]
mokimo: > tattva confirmed that there wont be any overwritting and they will test it once im done with my changes 👍
https://cq-dev.slack.com/archives/C06LCG55U3C/p1752151757191959?thread_ts=1752050245.486109&cid=C06LCG55U3C

Perfect, great

As for the race condition, I think it's unlikely to happen, but not impossible, since we never wait for the sidekick to load. E.g. if you'd do something such as this - to fake a very slow network response - you could observe that your callback is never executed.
```js
  setTimeout(() => {
    initSidekick();
  }, 10000);
```

libs/blocks/preflight/checks/preflightApi.js
```diff
@@ -56,45 +56,40 @@ export default {
...
-    const assets = await Promise.all(runChecksAssets(url, area));
-    const performance = await Promise.all(runChecksPerformance(url, area));
-    const seo = await Promise.all(runChecksSeo({ url, area }));
-    return {
-      assets,
-      performance,
-      seo,
-    };
-  })();
+  const res = useCache ? (checks = runChecks(url, area)) : runChecks(url, area);
```
mokimo, line 81: Since you already handle the caching part above, I'd re-start with it here. You already have a code-path for the cache scenario.

libs/utils/sidekick-decorate.js
```diff
@@ -127,7 +127,10 @@ export default async function stylePublish(sk) {
   const config = getConfig();
   config.onFooterReady = () => {
     window.milo.deferredPromise.then(() => {
-      checkPreflightAndShowNotification();
+      const isPublishDisabled = document.querySelector('aem-sidekick')?.shadowRoot?.querySelector('plugin-action-bar')?.shadowRoot?.querySelector('sk-action-button.publish[disabled]');
+      if (!isPublishDisabled) {
+        checkPreflightAndShowNotification();
+      }
```
mokimo, line 133: We can still run the checks also if publishing is disabled no?

libs/blocks/preflight/checks/preflightApi.js
skholkhojaev: is this what you mean?

```
export async function getPreflightResults(url, area, useCache = true) {
  let runChecksResult;

  if (useCache && checks) {
    runChecksResult = await checks;
  } else {
    const res = useCache ? (checks = runChecks(url, area)) : runChecks(url, area);
    runChecksResult = await res;
  }
```

mokimo: - If cache, for the first URL cache the checks
- If no cache, always run checks

libs/utils/sidekick-decorate.js
skholkhojaev: ah yes i didnt consider that part 😅,
thankyou for pointing that out

libs/utils/preflight-notification.js
```diff
@@ -48,6 +48,14 @@ async function createPreflightNotification() {
   document.body.appendChild(overlay);
 }
 
+async function showNotificationIfAllowed(hasFailures) {
+  const isPublishButtonDisabled = document.querySelector('aem-sidekick')?.shadowRoot?.querySelector('plugin-action-bar')?.shadowRoot?.querySelector('sk-action-button.publish[disabled]');
+  if (!hasFailures || preflightNotificationDismissed || isPublishButtonDisabled) {
+    return;
+  }
+  await createPreflightNotification();
```
mokimo, line 56: Same question as before, shouldn't we create the notification even if the publish button is disabled?

skholkhojaev: nono, we talked with Narcis about it here: https://adobe-mwp.slack.com/archives/C08GJP18528/p1752083684191999?thread_ts=1752064030.446019&cid=C08GJP18528

the authros will get another tooltip showed to them if the publish button is not available, we dont want preflight distracting them from that

libs/utils/preflight-notification.js
```diff
@@ -0,0 +1,107 @@
...
+      if (!notification && !preflightNotificationDismissed) {
+        const { hasFailures } = await getPreflightResults(window.location.href, document);
+        await showNotificationIfAllowed(hasFailures);
+      }
+    } else {
+      if (notification) {
+        notification.remove();
+      }
+      preflightNotificationDismissed = false;
+    }
```
mokimo, line 82: If else, and especially nested if else cases are always hard to read. Try to avoid nesting if's... there's rarely a good case to do so.

You already handle the remeoving notification case.
```suggestion
    if(sidekick.getAttribute('open') !== "open" || notificationWasDismissed) return
    const { hasFailures } = await getPreflightResults(window.location.href, document);
    await showNotificationIfAllowed(hasFailures);
```

libs/utils/preflight-notification.js
```diff
@@ -0,0 +1,107 @@
...
+    attributes: true,
+    childList: true,
+    subtree: true,
+    attributeFilter: ['open'],
+  });
+
+  return observer;
+}
+
+export default async function checkPreflightAndShowNotification() {
```
mokimo, line 95: I'd probably simplify this to just show.. your within the file 'preflight-notification' so that's what you'd show. The functionality is implied by the file your in IMO
```suggestion
export default async function show() {
```

libs/utils/preflight-notification.js
```diff
@@ -0,0 +1,107 @@
...
+  if (!observerCreated) {
+    createSidekickVisibilityObserver();
```
mokimo, line 97: naming-nit: `createSidekickVisibilityObserver()` -> `observeSidekick`

libs/utils/preflight-notification.js
```diff
@@ -0,0 +1,107 @@
...
+  const closeBtn = overlay.querySelector('.notification-close');
+  closeBtn.addEventListener('click', () => {
+    overlay.remove();
+    preflightNotificationDismissed = true;
+  });
+
+  document.body.appendChild(overlay);
+}
+
+async function showNotificationIfAllowed(hasFailures) {
```
mokimo, line 51: IMO This function can get killed and move into "show"

libs/utils/preflight-notification.js
```diff
@@ -0,0 +1,107 @@
...
+    e.preventDefault();
+    e.stopPropagation();
+    openPreflightPanel();
+    overlay.remove();
+  });
+
+  const closeBtn = overlay.querySelector('.notification-close');
+  closeBtn.addEventListener('click', () => {
+    overlay.remove();
+    preflightNotificationDismissed = true;
```
mokimo, line 45: nit naming suggestion: "wasDismissed"

libs/utils/preflight-notification.js
```diff
@@ -0,0 +1,107 @@
...
+    attributeFilter: ['open'],
+  });
+
+  return observer;
+}
+
+export default async function checkPreflightAndShowNotification() {
+  if (!observerCreated) {
+    createSidekickVisibilityObserver();
+    observerCreated = true;
```
mokimo, line 98: I'd probably rather have something such as `sidekickObserver = new MutationObserver`.... you don't need the extra variable, you can just see if you already created the observer once.

libs/utils/preflight-notification.js
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  const { hasFailures } = await getPreflightResults(window.location.href, document);
+  const existingNotification = document.querySelector('.milo-preflight-overlay');
+  if (existingNotification) {
+    existingNotification.remove();
+  }
+  await showNotificationIfAllowed(hasFailures);
```
mokimo, line 106: IMO you can remove this and just move the functionality to `show` ... that's probably clear enough

libs/utils/sidekick-decorate.js
```diff
@@ -122,6 +124,14 @@ export default async function stylePublish(sk) {
   if (stylePublishCalled) return;
   stylePublishCalled = true;
 
+  const config = getConfig();
+  config.onFooterReady = () => {
+    window.milo.deferredPromise.then(() => {
+      checkPreflightAndShowNotification();
+    });
+  };
+  updateConfig(config);
```
mokimo, line 133:
```suggestion
  const config = getConfig();
  config.onFooterReady = () => {
    await window.milo.deferredPromise
    checkPreflightAndShowNotification()
  };
  updateConfig(config);
```

mokimo (commented):
Was about to approve 😁 But a few more nits-cleanup and naming suggestions!

libs/styles/preflight-notification.css
```diff
@@ -0,0 +1,73 @@
+.milo-preflight-overlay {
+  position: fixed;
+  bottom: 120px;
+  left: 50%;
+  transform: translateX(-50%);
+  z-index: 10000;
```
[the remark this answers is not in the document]
skholkhojaev: thats true, i changed it to `101`

libs/styles/preflight-notification.css
```diff
@@ -0,0 +1,73 @@
+.milo-preflight-overlay {
+  position: fixed;
```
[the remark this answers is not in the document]
skholkhojaev: i could try to do it like that, but but since the want for hte notification bar was to have it "somewhere around" the publish button i dont think it would make much of a difference would it?

libs/utils/preflight-notification.js
```diff
@@ -0,0 +1,107 @@
...
+        </span>
+        <button class="notification-close">×</button>
+      </div>
+    </div>
+  `;
+
+  const reviewLink = overlay.querySelector('.preflight-review-link');
+  reviewLink.addEventListener('click', (e) => {
+    e.preventDefault();
+    e.stopPropagation();
```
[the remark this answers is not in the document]
skholkhojaev: thinking about it it give no additional benefit for this usecase.

libs/styles/preflight-notification.css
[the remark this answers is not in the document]
mokimo: What's the thought process here... would it make sense to start some kind of global variables in the styles.css to use? E.g. "modal-z-index" so that all modals share the same ..?

You don't need to make a huge refactoring but maybe it'd make sense to get this under control and at least start some variables. 

If we have an entry such as this:
```css
  --z-index-gnav: 100;
  --z-index-modals: 101;
  --z-index-x: 105;
```
it gets a lot more obvious on what item is placed where and for what reason.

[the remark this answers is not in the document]
mokimo: Nothing that's particularly your fault, but let's be good pathfinders and improve this for future use-cases

libs/utils/preflight-notification.js
```diff
@@ -0,0 +1,95 @@
...
+  });
+
+  return observer;
+}
+
+export default async function show() {
+  if (!observerCreated) {
+    sidekickObserver();
+    observerCreated = true;
+  }
```
mokimo, line 83:
```suggestion
export default async function show() {
  createObserver()
```

libs/utils/preflight-notification.js
```diff
@@ -0,0 +1,95 @@
...
+  closeBtn.addEventListener('click', () => {
+    overlay.remove();
+    wasDismissed = true;
+  });
+
+  document.body.appendChild(overlay);
+}
+
+function sidekickObserver() {
+  const observer = new MutationObserver(async () => {
```
mokimo, line 50:
```suggestion
function createObserver() {
  if(sidekickObserver) return
  sidekickObserver = new MutationObserver(async () => {
```

libs/utils/preflight-notification.js
```diff
@@ -0,0 +1,95 @@
...
+  });
+
+  observer.observe(document, {
+    attributes: true,
+    childList: true,
+    subtree: true,
+    attributeFilter: ['open'],
+  });
+
+  return observer;
```
mokimo, line 76:
```suggestion
```

libs/styles/preflight-notification.css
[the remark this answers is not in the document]
skholkhojaev: i totally agree, should i try an add it into this PR?

skholkhojaev:
@mokimo done ✅ 
if you want could you take a one last look to see if its alright

libs/styles/styles.css
```diff
@@ -10,6 +10,10 @@
   --feds-totalheight-nav: calc(var(--feds-height-nav, --global-height-nav) + var(--feds-height-breadcrumbs, --global-height-breadcrumbs));
   --feds-localnav-height: 40px;
 
+  /* Global Z-Index Variables */
+  --modal-z-index: 102;
+  --modal-z-index-curtain: 101;
```
mokimo, line 15: could you also pull the other z-index values here as variables, which are defined further down in the styles.css?

mokimo (approved):

libs/blocks/modal/modal.css
```diff
@@ -1,8 +1,6 @@
 :root {
   --modal-focus-color: #109cde;
   --modal-max-height: calc((100vh - 6px) - 2em);
-  --modal-z-index: 102;
```
robert-bogos, line 4: Why removing this? I see is still used on [line 25 of this file](https://github.com/adobecom/milo/pull/4496/files#diff-bad812e28c6ccc634b5b6ba319037036e8e6b1d59fbcff1df04cc269bf984660R25) 🤔

skholkhojaev: spoke with Okan and I removed those two lines and added them in the [Styles.css](https://github.com/adobecom/milo/pull/4496/files#diff-c685e2a5191e8c907b9958b7f267d04cc48fdce1491da2c38461955f5d2bd913) 😁

mokimo: While they are modal specific, given our "z-index issue" grows bigger as the time passes, IMO the tradeoff to keep them in the global styles for visibility is worth there... Esp. since z-index is very much globally interacting with all blocks in one way or another

libs/styles/preflight-notification.css
```diff
@@ -0,0 +1,72 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  gap: 12px;
+}
+
+.notification-message {
+  flex: 1;
+  white-space: normal;
+}
```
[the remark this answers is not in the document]
skholkhojaev: great, thankyou for noticing 😅

robert-bogos (changes requested):
I see that the SEO checks take a lot more to execute, looks like they are waiting for the Links check to be done. I think we should keep the previous behaviour  where SEO results came instantly and only the Links results later.

Also, the image labels applied to the page seem to render without styles for ~1 second.

robert-bogos (commented):
@skholkhojaev I see you have some conflicts

libs/blocks/preflight/visual-metadata.js
```diff
@@ -0,0 +1,54 @@
...
+      picture.insertBefore(container, element.nextSibling);
+    }
+  } else {
+    container = createTag('div', { class: 'picture-meta preflight-decoration no-picture-tag' });
+    element.parentNode.insertBefore(container, element.nextSibling);
+  }
+
+  const metadataEl = createTag('div', { class: `picture-meta-a11y preflight-decoration ${status}` }, message);
+  container.appendChild(metadataEl);
+}
```
robert-bogos, line 54: Trailing newline is missing here
```suggestion
}

```

libs/utils/sidekick.js
```diff
@@ -36,4 +38,4 @@ export default function init({ createTag, loadBlock, loadScript, loadStyle }) {
 
   // Color code publish button
   stylePublish(sk);
-}
+}
```
robert-bogos, line 41: Trailing newline also missing here
```suggestion
}

```

robert-bogos (commented):
Just some last nits 😊

mokimo:
FYI @robert-bogos merging without approving//dismissing your change requests, makes this look a bit weird in the Github search. Got confused why this was merged with changes requested 🤔 
image

PR #4497, opened 2025-07-01
Fix(Mwpw-173688): Non-MPC videos honor o.s level accessibility setting.

sharath-kannan:
* Autoplay videos honor o.s level reduce motion setting.

Resolves: [MWPW-173688](https://jira.corp.adobe.com/browse/MWPW-173688)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/ramuntea/accessibility/video-use-cases?martech=off
- After: https://mwpw-173688--milosh--sharath-kannan.aem.page/drafts/ramuntea/accessibility/video-use-cases?martech=off

### DIFF
--- a/libs/utils/decorate.js
+++ b/libs/utils/decorate.js
@@ -18,6 +18,7 @@ let videoLabels = {
   playIcon: 'Play icon',
   hasFetched: false,
 };
+const isReducedMotion = window.matchMedia('(prefers-reduced-motion: reduce)') === true || window.matchMedia('(prefers-reduced-motion: reduce)').matches === true;
 let videoCounter = 0;
 
 export function decorateButtons(el, size) {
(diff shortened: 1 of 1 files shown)


libs/utils/decorate.js
```diff
@@ -18,6 +18,7 @@ let videoLabels = {
   playIcon: 'Play icon',
   hasFetched: false,
 };
+const isReducedMotion = window.matchMedia('(prefers-reduced-motion: reduce)') === true || window.matchMedia('(prefers-reduced-motion: reduce)').matches === true;
```
robert-bogos, line 21: I'm not sure I understand the change in here 🤔
window.matchMedia() is always having an object as the return value, which makes  `window.matchMedia('(prefers-reduced-motion: reduce)') === true` redundant. 
Please point it out if I'm missing something.

nkthakur48 (approved):

PR #4501, opened 2025-07-04
MWPW-169236 [Merch Cards] Support for Volume Discount Flex Promos

bozojovicic:
For merch cards with prices with promos where the offer has minProductQuantity > 1 I added the support for these promotions based on selected value in the quantity selector. If such price with promo is not in the card or quantity selector is not there, nothing will change. If they are both there and the selected quantity is >= minProductQuantity then :

- price is displayed with promotion (with old price)
- CTA URL contains `apc` parameter with the promo code if that CTA is authored with that promo code

If the selected quantity is < minProductQuantity then :

- price is displayed without promotion (no old price)
- CTA URL loses `apc` parameter and `data-promotion-code` attribute (if that CTA is authored with that promo code)

I added unit tests where all this is simulated.

Resolves: [MWPW-169236](https://jira.corp.adobe.com/browse/MWPW-169236)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/bozo/qs-promo?martech=off
- After: https://mwpw169236--milo--bozojovicic.aem.page/drafts/bozo/qs-promo?martech=off


libs/features/mas/src/merch-card.js
```diff
@@ -372,6 +377,29 @@ export class MerchCard extends LitElement {
         for (const link of allLinks) {
             link.dataset.quantity = event.detail.option;
         }
+
+        if (this.prices.length === 0) return;
+
+        const price = this.prices[0];
+        if (this.quantityBasedPromo?.resolved && this.quantityBasedPromo?.promotionCode && price) {
+            if (event.detail.option >= this.quantityBasedPromo.minProductQuantity) {
```
yesil, line 385: I think the whole promo min quantity logic should move in checkout-mixin.
I don't understand the intent here.  Promo code should remain on checkout links irrespective of the quantity.  It belongs checkout page to apply or not the promo based on the quantity.

yesil (changes requested):
IMO, the only minQuantity logic we need to have is promo price rendering in inline-price.
checkout links should be good as is. 
Happy to discuss these in a call.

PR #4530, opened 2025-07-09
MWPW-176184 [MEP] Inline fragments created by MEP should not have a paragraph wrap

vgoodric:
Detailed Description: When trying to replace a button with an inline fragment, it sometimes puts the new button on a second line.

Cause: Paragraph elements do not support a block-level element nested inside of it. But it does allow you to insert one via javascript.  However, if you try to update innerHTML to have a nested paragraph in a paragraph, the browser will THEN fix the malformed HTML.  The result, inline fragments can be inserted in a paragraph, but will be moved if the HTML is processed for placeholders. 
URL: https://main--cc--adobecom.aem.page/drafts/mepdev/fragments/2025/q3/buttononnewlinebug/reduced
Steps to Reproduce:
1. Go to URL

Expected Results: buttons in yellow blade (section 2) should be side by side
Actual Results: buttons are stacked and offset

Resolves: [MWPW-NUMBER](https://jira.corp.adobe.com/browse/MWPW-176184)

**Test URLs:**
- Before: https://main--cc--adobecom.aem.page/drafts/mepdev/fragments/2025/q3/buttononnewlinebug/reduced
- After: https://main--cc--adobecom.aem.page/drafts/mepdev/fragments/2025/q3/buttononnewlinebug/reduced?milolibs=buttononnewlinebug
- https://buttononnewlinebug--milo--adobecom.aem.page/?martech=off


jpratt2 (approved):
I approve; I used both the screenshot tool and my own visual inspection, and I found no problems resulting from this code (using a milolibs param on several aem.live pages),

PR #4536, opened 2025-07-10
MWPW-175671: Adapt quantity selector popover placement based on current scroll 

st-angelo-adobe:
Resolves: [MWPW-175671](https://jira.corp.adobe.com/browse/MWPW-175671)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/npeltier/plans?martech=off&plan=team
- After: https://mwpw-175671--milo--adobecom.aem.page/drafts/npeltier/plans?martech=off&plan=team


libs/features/mas/src/merch-quantity-select.js
```diff
@@ -166,6 +166,15 @@ export class MerchQuantitySelect extends LitElement {
...
         this.closed = !this.closed;
+        if (!this.closed) this.adjustPopoverPlacement();
+        else this.highlightedIndex = this.options.indexOf(this.selectedValue);
+    }
+
+    adjustPopoverPlacement() {
+        const popover = this.shadowRoot.querySelector('.popover');
+        if (popover.getBoundingClientRect().bottom > window.innerHeight)
+            popover.setAttribute('placement', 'top');
+        else popover.setAttribute('placement', 'bottom');
```
JasonHowellSlavin, line 177: since these are functions, could we utilize early return to remove the `else`s. It is a pattern we typically seem to follow in Milo.

JasonHowellSlavin (approved):

libs/features/mas/src/merch-quantity-select.js
st-angelo-adobe: I removed an if by changing the logic slightly, but in `adjustPopoverPlacement` I'd just leave it as is, it's actually fewer lines this way.

PR #4548, opened 2025-07-14
MWPW-175655: Disabling automatic adding of legal text to card description prices

markovicpredrag1209:
* Removing description prices from adjust legal, since it will be manually added through studio ost
* After merging this PR, cards will need to be reauthored

Resolves: [MWPW-175655](https://jira.corp.adobe.com/browse/MWPW-175655)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/pedja/ost-legal?martech=off
- After: https://MWPW-175655--milo--markovicpredrag1209.aem.page/drafts/pedja/ost-legal?martech=off


mirafedas (approved):

afmicka:
Discussed offline regarding above comment, this price will need re-authoring with the new template. Does not affect any live card. cc: @Roycethan fyi when validating Stage to Main batch

PR #4552, opened 2025-07-15
MWPW-174582: retrying to fix id lookup

sheridansunier:
ID lookup tool broken when primary CTA is missing. We need to make sure that we are picking up a cta when authors do not explicitly author one to ensure the ID lookup tool will work correctly downstream.

Resolves: [MWPW-174582](https://jira.corp.adobe.com/browse/MWPW-174582)

Test URLs:

After: https://mwpw-174582-2--milo--sheridansunier.aem.page/?martech=off

### DIFF
--- a/tools/send-to-caas/send-utils.js
+++ b/tools/send-to-caas/send-utils.js
@@ -364,7 +364,6 @@ const parseCardMetadata = () => {
 function checkCtaUrl(s, options, i) {
   if ((s?.trim() === '' || s === undefined) && i > 1) return '';
   const url = (s?.trim() !== '' && s !== undefined) ? s : (options.prodUrl || window.location.origin + window.location.pathname);
-  if (url.includes('/tools/send-to-caas/bulkpublisher.html')) return '';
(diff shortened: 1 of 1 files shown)


tools/send-to-caas/send-utils.js
```diff
@@ -362,8 +362,8 @@ const parseCardMetadata = () => {
 };
 
 function checkCtaUrl(s, options, i) {
-  if (s?.trim() === '') return '';
-  const url = s || options.prodUrl || window.location.origin + window.location.pathname;
+  if ((s?.trim() === '' || s === undefined) && i > 1) return '';
+  const url = (s?.trim() !== '' && s !== undefined) ? s : (options.prodUrl || window.location.origin + window.location.pathname);
```
chrischrischris, line 366: nit
```suggestion
  const url = s?.trim() !== '' ? s : (options.prodUrl || window.location.origin + window.location.pathname);
```

chrischrischris (approved):

cmiqueo (approved):

PR #4556, opened 2025-07-16
[MWPW-176388] - Iframe dialog automated a11y attributes

DKos95:
Resolves: [MWPW-176388](https://jira.corp.adobe.com/browse/MWPW-176388)

**Test URLs:**
- After: https://iframe-dialog-a11y-attributes--milo--adobecom.aem.page/?martech=off

**CC Test URLs:**
- Before: https://main--cc--adobecom.aem.live/products/catalog
- After: https://main--cc--adobecom.aem.live/products/catalog?milolibs=iframe-dialog-a11y-attributes

### DIFF
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -98,10 +98,7 @@ function isElementInView(element) {
   );
 }
 
-const returnDialogLabel = (title) => {
-  if (title) return `Modal: ${title}`;
-  return 'Modal';
-};
+const returnDialogLabel = (title) => `Modal: ${title}`;
 
 function getCustomModal(custom, dialog) {
   const { miloLibs, codeRoot } = getConfig();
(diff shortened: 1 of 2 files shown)


libs/blocks/modal/modal.js
```diff
@@ -134,6 +135,7 @@ export async function getModal(details, custom) {
   const dialog = createTag('div', { class: 'dialog-modal', id, role: 'dialog', 'aria-modal': true });
   const loadedEvent = new Event('milo:modal:loaded');
 
+  if (custom && !custom?.title) custom.title = findDetails(window.location.hash, null)?.title;
```
mokimo, line 138: If custom?.title is available, custom is Too. You already check for `custom?.`
```suggestion
  if (!custom?.title) custom.title = findDetails(window.location.hash, null)?.title;
```

mokimo (approved):

libs/blocks/modal/modal.js
DKos95: But we have an ! in front of that, won't it enter the if clause then if custom is null and cause an error because we are assigning title to null?

mokimo: Oh your right, disregard the comment!

chrischrischris (approved):

overmyheadandbody (approved):

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-176388

PR #4568, opened 2025-07-18
[MILO][Preflight] Text overlapping in the Preflight Accessibility tab

skumar09:
* It fixes the Text is overlapping in the Preflight Accessibility tab when the right column has multiple entries

Resolves: [MWPW-176564](https://jira.corp.adobe.com/browse/MWPW-176564)

**Test URLs:**
- After: https://text-overlap--milo--adobecom.aem.page/?martech=off

Fixes following Issue
Issue: 
image

Test URL: https://main--cc--adobecom.aem.page/products/illustrator?milolibs=text-overlap--milo--skumar09&mep=off&martech=off

### DIFF
--- a/libs/blocks/preflight/preflight.css
+++ b/libs/blocks/preflight/preflight.css
@@ -861,7 +861,7 @@ img:hover ~ .asset-meta,
   display: grid;
   grid-template-columns: 1fr 1fr;
   gap: 48px;
-  height: 250px;
+  height: 280px;
 }
 
 .preflight-accessibility-item {
@@ -1012,8 +1012,8 @@ img:hover ~ .asset-meta,
 .preflight-full-width {
   grid-column: 1 / -1;
   width: 100%;
-  margin-top: 34px;
-  padding-top: 36px;
+  margin-top: 24px;
+  padding-top: 16px;
   box-shadow: 0 2px 4px rgba(0,0,0,0.05);
   border-top: 2px rgba(0, 0, 0, 0.2);
 }
--- a/libs/blocks/preflight/preflight.css
+++ b/libs/blocks/preflight/preflight.css
@@ -1012,8 +1012,8 @@ img:hover ~ .asset-meta,
 .preflight-full-width {
   grid-column: 1 / -1;
   width: 100%;
-  margin-top: 24px;
-  padding-top: 16px;
+  margin-top: 34px;
+  padding-top: 36px;
   box-shadow: 0 2px 4px rgba(0,0,0,0.05);
   border-top: 2px rgba(0, 0, 0, 0.2);
 }


mokimo (approved):

skholkhojaev (approved):

overmyheadandbody (changes requested):
I have a couple of issues with this:
* please update the test link to actually use your feature branch, in this case https://text-overlap--milo--skumar09.aem.page/?martech=off. Right now this leads to an erroneous page. @mokimo, @skholkhojaev, even if the change seems trivial, it's always a good idea to have a look whether the functionality works as expected;
* the margin and padding values are not the root cause of this issue, it's likely this - https://github.com/adobecom/milo/blob/2b4cf6ecd8d0ec7bd08c0dae2e487890b402c40b/libs/blocks/preflight/preflight.css#L864. We should address that rather than patch here.

skumar09:
> I have a couple of issues with this:
> 
> * please update the test link to actually use your feature branch, in this case https://text-overlap--milo--skumar09.aem.page/?martech=off. Right now this leads to an erroneous page. @mokimo, @skholkhojaev, even if the change seems trivial, it's always a good idea to have a look whether the functionality works as expected;
> * the margin and padding values are not the root cause of this issue, it's likely this - https://github.com/adobecom/milo/blob/2b4cf6ecd8d0ec7bd08c0dae2e487890b402c40b/libs/blocks/preflight/preflight.css#L864
>   . We should address that rather than patch here.



> I have a couple of issues with this:
> 
> * please update the test link to actually use your feature branch, in this case https://text-overlap--milo--skumar09.aem.page/?martech=off. Right now this leads to an erroneous page. @mokimo, @skholkhojaev, even if the change seems trivial, it's always a good idea to have a look whether the functionality works as expected;
> * the margin and padding values are not the root cause of this issue, it's likely this - https://github.com/adobecom/milo/blob/2b4cf6ecd8d0ec7bd08c0dae2e487890b402c40b/libs/blocks/preflight/preflight.css#L864
>   . We should address that rather than patch here.

Hi @overmyheadandbody : thanks for the review and suggestion, i have added the test url and updated the fix, please review.

hit11757 (approved):
Validated. Testing details https://jira.corp.adobe.com/browse/MWPW-176564

libs/blocks/preflight/preflight.css
```diff
@@ -861,7 +861,7 @@ img:hover ~ .asset-meta,
   display: grid;
   grid-template-columns: 1fr 1fr;
   gap: 48px;
-  height: 250px;
+  height: 280px;
```
overmyheadandbody, line 864: I'm still not sure why we need a fixed value here instead of letting things flow naturally, but I guess this does address the issue for now.

overmyheadandbody (commented):
Removing the request changes to move this forward.

PR #4578, opened 2025-07-21
FaaS a11y enhancements

overmyheadandbody:
This tries to address multiple FaaS form accessibility issues found on Milo pages. Since FaaS doesn't have extensive Consonant specs, Marketo forms have been used as a source of inspiration, with the aim of having a consistent UI across forms on adobe.com pages. There are additional accessibility issues for the FaaS forms, which will be addressed in a subsequent contribution.

Note that VPN is needed to access test pages containing FaaS forms.

The most complex part of this is the error announcements, especially considering that we're consuming FaaS, not building it in Milo itself. A few notes on this:
- when an error appears due to another field being focused, the next form field will be announced first, then the recent error. Although not ideal, this is acceptable from an a11y perspective. We can nevertheless look into further improving it in the future. I've confirmed this with our a11y specialists and also cross-referenced with other forms considered accessible. Also note that I've also tried to make the live aria be `assertive`, but this leads to the message never being announced in this context, as the announcement of the newly focused field takes precedence;
- when a field that already has an error gets focused, its error message should be announced;
- keep in mind some fields are initially hidden and become visible based on the user's options;
- once submitting with errors, the first fields with an error should be focused and its error announced;

| Before | After |
| Screenshot 2025-07-18 at 16 52 57 | Screenshot 2025-07-18 at 16 52 46 |

Resolves:
* Select lacking accessible focus indicator: [MWPW-167924](https://jira.corp.adobe.com/browse/MWPW-167924)
* Submit button contrast issues: [MWPW-167919](https://jira.corp.adobe.com/browse/MWPW-167919)
* Submit button focus state: [MWPW-167921](https://jira.corp.adobe.com/browse/MWPW-167921)
* Error text contrast color: [MWPW-167916](https://jira.corp.adobe.com/browse/MWPW-167916)
* Error message announcement: [MWPW-167918](https://jira.corp.adobe.com/browse/MWPW-167918)

**Test URLs:**
- Before (PSI): https://main--milo--overmyheadandbody.aem.page?martech=off
- After (PSI): https://faas-a11y--milo--overmyheadandbody.aem.page?martech=off
- Before (CC): https://main--cc--adobecom.aem.live/creativecloud/business/teams/request-information?georouting=off
- After (CC): https://main--cc--adobecom.aem.live/creativecloud/business/teams/request-information?georouting=off&milolibs=faas-a11y--milo--overmyheadandbody


libs/blocks/faas/faas.js
```diff
@@ -12,6 +64,31 @@ const loadFaas = async (a) => {
   if (faas && faas.closest('.dialog-modal')) {
     faas.querySelector('.faas').classList.add('column2');
   }
+
+  // Accessibility logic
+  // Set up a MutationObserver to wait for the form to be added to the DOM
+  const formObserver = new MutationObserver(() => {
```
mokimo, line 70: I thought we don't do comments in milo? You could name it `accessibilityComplianceFormObserver` if you want to bring that point across IMO

libs/blocks/faas/faas.css
```diff
@@ -46,24 +59,28 @@ a[href*='/tools/faas#'], a[href*='/tools/faas?'] {
...
-  color: #fff;
-  font-weight: 700;
-  padding: 7.5px 13.1px;
+  background: var(--color-accent);
+  border-radius: 16px;
+  border: 2px solid var(--color-accent);
+  height: 32px;
+  color: var(--color-white);
+  display: inline-block;
+  font-family: var(--body-font-family);
```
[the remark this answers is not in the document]
overmyheadandbody: True, will make the change.

libs/blocks/faas/faas.js
```diff
@@ -12,6 +61,23 @@ const loadFaas = async (a) => {
...
+    const validationObserver = createValidationObserver();
+
+    [...faasForm.querySelectorAll('.row')].forEach((row) => {
+      validationObserver.observe(row, { attributes: true, attributeFilter: ['class'] });
+    });
+
+    formObserver.disconnect();
+  });
+
+  formObserver.observe(faas, { childList: true, subtree: true });
```
rgclayton, line 80: This is a clever approach. Nice work! TBH, it sort of melts my brain.

SilviuLCF:
> Hey @overmyheadandbody. While testing this on page, I ran into an interesting use case. Essentially, while using the Mac OS enabled voice over accessibility tool, I accidentally hit the enter key while trying to select an option in the first dropdown, on CC test page you listed. This marks ALL required fields with their error message all at once. When I tab through the fields after that the error messages are not announced.
> 
> Here's a screen recording.
> 
>  faas-screen-reader-test_1.mp4

HI Ryan, 

Tried to replicate your scenario, and wasn't able.  Same for Rares

Attached is a short video trying to replicate

libs/blocks/faas/faas.js
rgclayton: Interesting. Ok. If it's working as expected. I just wanted to call out and record what I saw.

DKos95 (approved):

SilviuLCF (approved):
verified, testing details in corresponding jira tickets

PR #4595, opened 2025-07-23
MWPW-176866 Placeholders within fragments are not resolved anymore

vgoodric:
In a previous PR, we started passing inline fragments through loadArea like non-inline fragments. However, this function normally also processes placeholders found in the fragment. This causes fed placeholders in gnav to break.  We need a solution that keeps the processing of nested fragments while not processing placeholders like before.

Resolves: [MWPW-176866](https://jira.corp.adobe.com/browse/MWPW-176866)

**Test URLs:**
- After: https://inlinefragfedplaceholders--milo--adobecom.aem.page/?martech=off

QA links:
Needs to still have a blue button
Before: https://main--cc--adobecom.aem.page/drafts/mepdev/fragments/2025/q3/pznfailswithinline/inline-no-mep
After: https://main--cc--adobecom.aem.page/drafts/mepdev/fragments/2025/q3/pznfailswithinline/inline-no-mep?milolibs=inlinefragfedplaceholders
image

Needs to start resolving placeholder again
Before: https://main--homepage--adobecom.aem.page/homepage/index-loggedout
After: https://main--homepage--adobecom.aem.page/homepage/index-loggedout?milolibs=inlinefragfedplaceholders
image

### DIFF
--- a/libs/blocks/fragment/fragment.js
+++ b/libs/blocks/fragment/fragment.js
@@ -52,7 +52,7 @@ const insertInlineFrag = async (sections, a, relHref) => {
   const promises = [];
   fragChildren.forEach((child) => {
     child.setAttribute('data-path', relHref);
-    promises.push(loadArea(child));
+    promises.push(loadArea(child, false));
   });
   await Promise.all(promises);
 };
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -1149,8 +1149,8 @@ function getPlaceholderPaths(config) {
 }
 
 let placeholderRequest;
-export async function decoratePlaceholders(area, config) {
-  if (!area) return;
+export async function decoratePlaceholders(area, config, processPlaceholders = true) {
+  if (!area || !processPlaceholders) return;
   const nodes = findReplaceableNodes(area);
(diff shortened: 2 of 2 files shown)


vgoodric:
@mokimo Yeah, the assumption that inline would only be small snippets is the in fact reason we did not decorate it.  But the problem is trying to put things like cards in a fragment.  Anything that needs to be side by side with other content and have the fragment's content adjust in height to its "siblings".  If you don't make the fragment inline, the other cards are more like uncles or cousins in the DOM.  They're nested too deep.  So authors want to start using inline with any side-by-side content.  This results in fragments containing entire blocks using inline.

sharmrj (approved):

spadmasa:
Validation on the pr done on the below url
https://main--homepage--adobecom.aem.page/homepage/index-loggedout?milolibs=inlinefragfedplaceholders
image
https://main--cc--adobecom.aem.page/drafts/mepdev/fragments/2025/q3/pznfailswithinline/inline-no-mep?milolibs=inlinefragfedplaceholders
image

mokimo:
fyi @vgoodric had to be reverted :(

PR #4596, opened 2025-07-24
MWPW-175973 [Accessibility] Filter component accessibility issue on Blog page

suhjainadobe:
[Accessibility] Filter component accessibility issue on Blog page

-When navigating with the Tab key, the filter component gets skipped
-The Tab key also skips the Reset and Apply buttons
-Focus should be looped within the dropdown once it’s open
-If I click on one of the filter tabs, I can tab through the dropdown options, but I can’t select a checkbox using Enter
-If one filter tab is selected, I can't move to another one using the arrow keys navigation

Resolves: [MWPW-175973](https://jira.corp.adobe.com/browse/MWPW-NUMBER)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page
- After: https://article-feed--milo--suhjainadobe.aem.page

- Before: https://main--blog--adobecom.aem.page/en/drafts/suhani/creativity
- After: https://main--blog--adobecom.aem.page/en/drafts/suhani/creativity?milolibs=article-feed--milo--suhjainadobe


overmyheadandbody (commented):
Thanks for contributing this, Suhani! I have a first round of comments and I'll reach out to a11y to check whether the proposed functionality matches their expectations.

I see that a couple of checks are failing, they would need some looking into to get this green.

libs/blocks/article-feed/article-feed.css
```diff
@@ -525,4 +530,4 @@ main a.article-card:any-link {
   }
 }
 
-/** END **/
+/** END **/
```
overmyheadandbody, line 533: This comment can probably be removed, but the empty line at the end of the file should be kept.

libs/blocks/article-feed/article-feed.js
```diff
@@ -289,19 +362,52 @@ function buildFilterOption(itemName, type) {
...
   button.classList.add('filter-button');
   button.id = `${type}-filter-button`;
-  button.setAttribute('aria-haspopup', true);
-  button.setAttribute('aria-expanded', false);
-  button.setAttribute('role', 'button');
+  button.setAttribute('tabindex', '0');
+  button.setAttribute('aria-haspopup', 'true');
+  button.setAttribute('aria-expanded', 'false');
+  button.setAttribute('aria-controls', `${type}-filter-panel`);
   button.textContent = tax.getCategoryTitle(type);
```
overmyheadandbody, line 372: This could also be defined with `createTag` to keep consistency

overmyheadandbody (approved):
There are some failing lint checks which need to be addressed before being able to merge this. Otherwise this looks good, definitely more complex than I had originally expected

PR #4597, opened 2025-07-24
MWPW-176777: Safari Fix

markovicpredrag1209:
- Fixes a bug where onceSettled from inline-price is not available yet on safari

Resolves: [MWPW-176777](https://jira.corp.adobe.com/browse/MWPW-176777)

**Test URLs:**
- After: https://MWPW-176777--milo--adobecom.aem.page/?martech=off


3ch023 (approved):

afmicka:
@markovicpredrag1209 this looks good to me. The only thing noticing is that font for ABM and Addon look different. Possibly Secure transaction as well but not sure for that one. It has been mentioned in the jira's description. If something can be easily done here for it, let's add it. Otherwise we can proceed with this as is and deal with font separately.
Screenshot 2025-07-25 at 12 34 00

cc: @Roycethan @3ch023

PR #4600, opened 2025-07-24
MWPW-176945 [Milo] Tabs block fires multiple click events on load when nested

markpadbe:
When the tabs block is nested within other tabs blocks, multiple click events are fired unexpectedly on load. This appears to be because the changeTab click handler gets added to nested tabs more than once. This PR addresses the issue by removing the handler before adding it.  

URL: https://www.adobe.com/creativecloud/pro.html
Steps to Reproduce:
1. load the above url and open the network tabs in the browser
2. filter for "collect" and reload if necessary
3. you should see more than 5 events where launch is the initiator(actual)
4. override the tab.js file in your browser using the file in this pr
5. reload
6. you should only see one event where launch is the initiator(expected)

actual: 
Screenshot 2025-07-24 at 4 04 39 PM

expected:
Screenshot 2025-07-24 at 4 13 38 PM

Resolves: [MWPW-176945](https://jira.corp.adobe.com/browse/MWPW-176945)

**Test URLs:**
- After: https://fixtabsmultieventbug--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/libs/blocks/tabs/tabs.js
+++ b/libs/blocks/tabs/tabs.js
@@ -164,6 +164,7 @@ function initTabs(elm, config, rootElem) {
     });
   });
   tabs.forEach((tab) => {
+    tab.removeEventListener('click', changeTabs);
     tab.addEventListener('click', changeTabs);
   });
   if (config) configTabs(config, rootElem);


rgclayton:
The update looks simple enough and I am not seeing any bad side effects. However, I'm not seeing any difference on the example page with the updated code.

I see the same results when filtering for "collect" in the network tab with and with out the code update.

Attaching a screen recording of what I'm seeing to get some clarification. 

I'm assuming testing the feature branch against `aem.live `page would have the same results but when I test I don't get any reference for target being the initiator. https://main--cc--adobecom.aem.live/creativecloud/pro?milolibs=fixtabsmultieventbug


https://github.com/user-attachments/assets/0ac4125e-fcd8-4750-a13b-e46c555dc122

PR #4606, opened 2025-07-28
MWPW-177021 The Gnav Bundling process fails because it tries to import the entirety of merch

sharmrj:
* Add your
* Specific
* Features or fixes

Resolves: [MWPW-177021](https://jira.corp.adobe.com/browse/MWPW-177021)

**Test URLs:**
- After: https://custom-merch-import-plugin--milo--adobecom.aem.page/?martech=off


npeltier:
@sharmrj that looks wrong to me: what if merch libs change? you do need to re run your stuff, no? can't you rather raise a discussion naming your issues, and we look at it together?

sharmrj:
> @sharmrj that looks wrong to me: what if merch libs change? you do need to re run your stuff, no? can't you rather raise a discussion naming your issues, and we look at it together?

@npeltier I've raised that discussion here: https://github.com/orgs/adobecom/discussions/294

To be clear, our gnav bundling isn't for everyone. The default way of consuming gnav is to be a Milo Consumer. The second way of consuming it is by dynamic import (no bundling). Bundling is for specific consumers that need versions and don't want to keep up with Milo's daily release schedule. So we'll only be running the build process when we do a release.
We've written this script precisely so that if merch changes we should still be able to bundle it automatically (so long as our underlying assumptions don't change).

PR #4631, opened 2025-07-31
MWPW-169932 adding font-display:swap for iphone safari to stopping fl…

suhjainadobe:
…ickering on page load

Milo pages flashing occurs after initial load on Mobile Safari because the font that are rendering on safari have font-display as optional. And due to browser behaviour, when the adobe-fonts replace the backup fonts, the text flickers. It is not as smooth as other browsers.

Resolves: [MWPW-169932](https://jira.corp.adobe.com/browse/MWPW-NUMBER)

**Test URLs:**
- Before: https://main--milo--suhjainadobe.aem.live/drafts/suhjain/m2-folder/carousel/photoshop-page
- After: https://font-swap--milo--suhjainadobe.aem.live/drafts/suhjain/m2-folder/carousel/photoshop-page


libs/utils/fonts.js
```diff
@@ -18,7 +18,13 @@ export default function loadFonts(locale, loadStyle) {
   const tkSplit = locale.tk.split('.');
   if (tkSplit[1] === 'css') {
     return new Promise((resolve) => {
-      loadStyle(`https://use.typekit.net/${locale.tk}`, resolve);
+      const isSafari = navigator.userAgent.includes('Safari') && !navigator.userAgent.includes('Chrome');
+      const PHONE_SIZE = window.screen.width < 550 || window.screen.height < 550;
+      if (isSafari && locale.tk === 'hah7vzn.css' && PHONE_SIZE) {
```
robert-bogos, line 23: can we rely on `'hah7vzn.css'` ? Feels brittle

suhjainadobe: We have to replace only thah7vzn.css typekit and this typekit is used in multiple locales. Thus, to apply condition on the locales itself might be a overkill. 
In that case as well, we will have to check if that locale has hah7vzn.css as typekit, only then replace the typekit. Would appreciate if there any other suggestions for this.

libs/utils/fonts.js
```diff
@@ -18,7 +18,13 @@ export default function loadFonts(locale, loadStyle) {
   const tkSplit = locale.tk.split('.');
   if (tkSplit[1] === 'css') {
     return new Promise((resolve) => {
-      loadStyle(`https://use.typekit.net/${locale.tk}`, resolve);
+      const isSafari = navigator.userAgent.includes('Safari') && !navigator.userAgent.includes('Chrome');
+      const PHONE_SIZE = window.screen.width < 550 || window.screen.height < 550;
```
mokimo, line 22: Where does the 550px value come from? 🤔 Isn't mobile usually 600px

libs/utils/fonts.js
```diff
@@ -18,7 +18,13 @@ export default function loadFonts(locale, loadStyle) {
   const tkSplit = locale.tk.split('.');
   if (tkSplit[1] === 'css') {
     return new Promise((resolve) => {
-      loadStyle(`https://use.typekit.net/${locale.tk}`, resolve);
+      const isSafari = navigator.userAgent.includes('Safari') && !navigator.userAgent.includes('Chrome');
```
mokimo, line 21: can the user agent be both safari AND chrome?

suhjainadobe: Yes, the user agent string can include both "Safari" and "Chrome", and this is common in browsers like Google Chrome.
Chromes user agent: 
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/115.0.0.0 Safari/537.36

libs/utils/fonts.js
mokimo: Im not sure how personalization got the 550px value either 🤔

overmyheadandbody: I'd also be more comfortable having the standard `600` value as the mobile one. So depending on your findings for iPad, maybe we can standardize this with our media queries. Looks good otherwise!

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details https://jira.corp.adobe.com/browse/MWPW-169932

PR #4639, opened 2025-08-01
[MWPW-174763] - columns table a11y update

DKos95:
- Updates column table to have adequate a11y attributes
- Tests added for column table

Resolves: [MWPW-174763](https://jira.corp.adobe.com/browse/MWPW-174763)

**Column kitchen sink Test URLs:**
- Before: https://main--milo--adobecom.aem.page/docs/library/kitchen-sink/columns?martech=off
- After: https://columns-table-a11y--milo--adobecom.aem.page/docs/library/kitchen-sink/columns?martech=off

**Column table without headers Test URLs:**
- Before: https://main--cc--adobecom.aem.live/privacy/policy-cn
- After: https://main--cc--adobecom.aem.live/privacy/policy-cn?milolibs=columns-table-a11y

**One table column header, one table row header Test URLs:**
- Before: https://main--cc--adobecom.aem.live/trust/transparency/content-policies/content-rights/private-info
- After: https://main--cc--adobecom.aem.live/trust/transparency/content-policies/content-rights/private-info?milolibs=columns-table-a11y

**Column table with only images Test URLs:**
- Before: https://main--cc--adobecom.aem.live/trust/compliance/esig-identity-providers?mep=off
- After: https://main--cc--adobecom.aem.live/trust/compliance/esig-identity-providers?mep=off&milolibs=columns-table-a11y

**Double column header Test URLs:**
- Before: https://main--cc--adobecom.aem.live/creativecloud/file-types/image/comparison/raster-vs-vector?mep=off
- After: https://main--cc--adobecom.aem.live/creativecloud/file-types/image/comparison/raster-vs-vector?mep=off&milolibs=columns-table-a11y

**Column header and row header in one table Test URLs:**
- Before: https://main--bacom--adobecom.aem.live/customers/consulting-services/premier-support
- After: https://main--bacom--adobecom.aem.live/customers/consulting-services/premier-support?milolibs=columns-table-a11y

- Before: https://main--milo--adobecom.aem.page/drafts/dusan/column-table-row-header
- After: https://columns-table-a11y--milo--adobecom.aem.page/drafts/dusan/column-table-row-header

**Column header Test URLs:**
- Before: https://main--milo--adobecom.aem.page/docs/library/blocks/columns?martech=off
- After: https://columns-table-a11y--milo--adobecom.aem.page/docs/library/blocks/columns?martech=off

- Before: https://main--cc--adobecom.aem.live/trust/compliance/compliance-list?mep=off
- After: https://main--cc--adobecom.aem.live/trust/compliance/compliance-list?mep=off&milolibs=columns-table-a11y

- Before: https://main--cc--adobecom.aem.live/trust/transparency/transparency-2020
- After: https://main--cc--adobecom.aem.live/trust/transparency/transparency-2020?milolibs=columns-table-a11y

- Before: https://main--cc--adobecom.aem.live/uk/creativecloud/illustration/discover/digital-art?mep=off
- After: https://main--cc--adobecom.aem.live/uk/creativecloud/illustration/discover/digital-art?mep=off&milolibs=columns-table-a11y

**Column header table and no row header table Test URLs:**
- Before: https://main--cc--adobecom.aem.live/trust/transparency/content-policies/harmful-content/regulated-goods?mep=off
- After: https://main--cc--adobecom.aem.live/trust/transparency/content-policies/harmful-content/regulated-goods?mep=off&milolibs=columns-table-a11y

- Before: https://main--milo--adobecom.aem.page/drafts/dusan/column-table-strong-inside
- After: https://columns-table-a11y--milo--adobecom.aem.page/drafts/dusan/column-table-strong-inside

### DIFF
--- a/libs/blocks/columns/columns.js
+++ b/libs/blocks/columns/columns.js
@@ -12,8 +12,8 @@ const isRowHeader = (cols, col, cdx, rows) => {
   const hasColumnHeaderInFirstRow = correspondingFirstRowCol?.getAttribute('role') === 'columnheader';
 
   return columnsWithStrongOrH.length === 1
-  && columnsWithStrongOrH[0] === col
-  && !hasColumnHeaderInFirstRow;
+    && columnsWithStrongOrH[0] === col
+    && !hasColumnHeaderInFirstRow;
 };
 
 const isColumnHeader = (rdx, containsHTag, meaningfulChildren) => (


libs/blocks/columns/columns.js
```diff
@@ -1,19 +1,64 @@
...
+      const containsHTag = meaningfulChildren[0]?.tagName?.match(/^H[1-6]$/);
+      if (containsHTag) meaningfulChildren[0].setAttribute('role', 'paragraph');
+
+      if (rdx > 0) {
+        const columnsWithStrongOrH = Array.from(cols).filter((rowCol) => {
+          const rowColMeaningfulChildren = Array.from(rowCol.childNodes).filter(
+            (node) => node.nodeType !== Node.TEXT_NODE || node.textContent.trim() !== '',
+          );
+          return rowColMeaningfulChildren.some((child) => child.tagName?.match(/^(H[1-6]|STRONG)$/))
+            || (rowCol.childNodes.length === 1 && rowCol.childNodes[0].tagName === 'STRONG');
```
robert-bogos, line 37: Do we need both `rowColMeaningfulChildren.some((child) => child.tagName?.match(/^(H[1-6]|STRONG)$/))` and `(rowCol.childNodes.length === 1 && rowCol.childNodes[0].tagName === 'STRONG')` for the return statement? 🤔
They seem to check for the same thing

DKos95: You are indeed right Robert, these were some additional checks left over after improving the solution a couple of times, we can just do `rowColMeaningfulChildren.some((child) => child.tagName?.match(/^(H[1-6]|STRONG)$/))` because later we check if this has a corresponding column header, if it does not have a column header then it's a row header and vice versa, thanks a lot for the review!

libs/blocks/columns/columns.js
```diff
@@ -1,19 +1,63 @@
...
+      }
+
+      if ((!rdx && meaningfulChildren.length === 1
+        && !meaningfulChildren.some((child) => child.classList?.contains('image-link'))) || containsHTag) {
+        col.classList.add('column-title');
+        col.setAttribute('role', 'columnheader');
+        return;
+      }
+
+      if (col.innerText) col.setAttribute('role', 'cell');
```
mokimo, line 60: IMO there's some opportunities here to extract some or most of this logic into helper functions, this is currently quite hard to read

libs/blocks/columns/columns.js
```diff
@@ -1,19 +1,63 @@
...
+
+    if (isTable && Array.from(cols).some((col) => col.hasChildNodes())) row.setAttribute('role', 'row');
+
     cols.forEach((col, cdx) => {
       col.classList.add('col', `col-${cdx + 1}`);
       if (!isTable) return;
-      if (!rdx && !cdx && !col.innerText) col.classList.add('empty-table-heading');
-      const isTableRowTitle = !cdx && col.childNodes.length === 1 && col.childNodes[0].tagName === 'STRONG';
-      if (isTableRowTitle) col.classList.add('row-title');
+      if (!rdx && !cdx && !col.innerText) {
```
overmyheadandbody, line 20: Maybe also calling `.trim()` on `innerText` would be safer if an author would accidentally leave in some spaces/returns?

overmyheadandbody (commented):
I agree with @mokimo that this is a bit difficult to read and could benefit from a bit of reorganization to avoid deep nesting.

libs/blocks/columns/columns.js
```diff
@@ -1,19 +1,74 @@
+const hasMeaningfulChildren = (col) => Array.from(col.childNodes).filter(
```
overmyheadandbody, line 1: The `has` makes me think this returns a Boolean, when in fact it returns an Array, so maybe `getMeaningfulChildren` would be more accurate?

libs/blocks/columns/columns.js
```diff
@@ -1,19 +1,74 @@
...
+  (node) => node.nodeType !== Node.TEXT_NODE || node.textContent.trim() !== '',
+);
+
+const isRowHeader = (cols, col, cdx, rows) => {
+  const columnsWithStrongOrH = Array.from(cols).filter((rowCol) => {
+    const rowColMeaningfulChildren = hasMeaningfulChildren(rowCol);
```
narcis-radu, line 7: nit: not really needed, you can just do `(hasMeaningfulChildren(rowCol)).some`

libs/blocks/columns/columns.js
```diff
@@ -1,19 +1,74 @@
...
+    el.classList.add('columns-table');
+    el.classList.remove('table');
+    el.setAttribute('role', 'table');
+  }
+
   rows.forEach((row, rdx) => {
     row.classList.add('row', `row-${rdx + 1}`);
     const cols = row.querySelectorAll(':scope > div');
+
+    if (isRow(cols, isTable)) row.setAttribute('role', 'row');
```
narcis-radu, line 61: In the ticket there is a recommendation to use `rowgroup` as a wrapper for all `row` elements, but I cannot see it implemented. Is that optional?

overmyheadandbody (commented):
Looks good overall, it's tough to handle all of the use-cases for this. I have a few notes that might be worth looking into:
* on https://main--bacom--adobecom.aem.live/customers/consulting-services/premier-support?milolibs=columns-table-a11y, the first column on the first row should ideally also be marked as a `cell`. Running the screen reader will announce "Expert Success" as "Column 1 of 3", even though it's the second; same for https://columns-table-a11y--milo--adobecom.aem.page/drafts/dusan/column-table-row-header, which also has empty `rowheader` cells, those should be marked as `cell` as well. However, the fact that the empty cells are not marked as such on https://main--cc--adobecom.aem.live/privacy/policy-cn?milolibs=columns-table-a11y actually plays in our favor, I'm not sure there's an easy way to account for both situations;
* a bit of an edge-case, but maybe a `.innerText.trim()` check should be used on the whole table for cases like https://main--cc--adobecom.aem.live/trust/compliance/esig-identity-providers?mep=off&milolibs=columns-table-a11y. I'm not sure the columns on this page should be marked as a table, as there are no text entries, no column or row headers, it seems to me that the table variant is used just for UI purposes

libs/blocks/columns/columns.js
```diff
@@ -1,19 +1,86 @@
...
+const applyCellIfRowColHasText = (cols, col) => {
+  if (Array.from(cols).some((otherCol) => otherCol.innerText.trim())) col.setAttribute('role', 'cell');
+};
+
 export default function init(el) {
   const rows = el.querySelectorAll(':scope > div');
   const isTable = el.classList.contains('table');
+
+  if (isTable) {
+    el.classList.add('columns-table');
```
zagi25, line 62: NIT: You should be able to use [replace](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/replace) here, also role can be set here once instead of checking and setting it in `applyAccessibilityAttributes`
```suggestion
    el.classList.replace('table', 'columns-table');
    el.setAttribute('role', 'table');
```

libs/blocks/columns/columns.js
```diff
@@ -1,19 +1,86 @@
...
+  && columnsWithStrongOrH[0] === col
+  && !hasColumnHeaderInFirstRow;
+};
+
+const isColumnHeader = (rdx, containsHTag, meaningfulChildren) => (
+  (!rdx && meaningfulChildren.length === 1
+    && !meaningfulChildren.some((child) => child.classList?.contains('image-link'))) || containsHTag
+);
+
+const applyTableRow = (row) => { if (!row.getAttribute('role')) row.setAttribute('role', 'row'); };
```
zagi25, line 25: NIT: Can this be changed to e.g. `setElRole(el, role)` or something like that ? Because I see that on multiple lines there are `.setAttribute('role', ...)` and this pretty much does the same thing. Or maybe even better, function can accept array of arrays (or some smarter param structure)
```
function setElsRole(elements) {
   elements.forEach(([el, role]) => {
       if(el.getAttribute('role')) return;
       el.setAttribute('role', role);
}

if (isColumnHeader(rdx, containsHTag, meaningfulChildren)) {
    setElsRole([col, 'columnheader'], [row, 'row'])
    return;
}
  
Something like this, so repetition can be avoided

File: libs/blocks/columns/columns.js [hunk 9]
```diff
@@ -1,19 +1,86 @@
...
+  (node) => node.nodeType !== Node.TEXT_NODE || node.textContent.trim() !== '',
+);
+
+const isRowHeader = (cols, col, cdx, rows) => {
+  const columnsWithStrongOrH = Array.from(cols)
+    .filter((rowCol) => getMeaningfulChildren(rowCol)
+      .some((child) => child.tagName?.match(/^(H[1-6]|STRONG)$/)));
+
+  const firstRow = rows[0];
+  const firstRowCols = firstRow.querySelectorAll(':scope > div');
```
zagi25, line 11: NIT: If `firstRow` is only used here, there is no need for variable
```suggestion
  const firstRowCols = row[0].querySelectorAll(':scope > div');
```

libs/blocks/columns/columns.js
DKos95: We didn't use this for the regular table block and it didn't show up in reports and if we need it here we need it there also, @overmyheadandbody  will bring this up during the meeting with the accessibility team and we can take it from there, this probably wouldn't be a small change considering we would be adding another element in there and css would also have to be adapter. Very nice catch!

DKos95:
> Looks good overall, it's tough to handle all of the use-cases for this. I have a few notes that might be worth looking into:
> 
> * on https://main--bacom--adobecom.aem.live/customers/consulting-services/premier-support?milolibs=columns-table-a11y, the first column on the first row should ideally also be marked as a `cell`. Running the screen reader will announce "Expert Success" as "Column 1 of 3", even though it's the second; same for https://columns-table-a11y--milo--adobecom.aem.page/drafts/dusan/column-table-row-header, which also has empty `rowheader` cells, those should be marked as `cell` as well. However, the fact that the empty cells are not marked as such on https://main--cc--adobecom.aem.live/privacy/policy-cn?milolibs=columns-table-a11y actually plays in our favor, I'm not sure there's an easy way to account for both situations;
> * a bit of an edge-case, but maybe a `.innerText.trim()` check should be used on the whole table for cases like https://main--cc--adobecom.aem.live/trust/compliance/esig-identity-providers?mep=off&milolibs=columns-table-a11y. I'm not sure the columns on this page should be marked as a table, as there are no text entries, no column or row headers, it seems to me that the table variant is used just for UI purposes

All concerns should be addressed, thanks for the input!

libs/blocks/columns/columns.js
```diff
@@ -1,19 +1,84 @@
...
+    .filter((rowCol) => getMeaningfulChildren(rowCol)
+      .some((child) => child.tagName?.match(/^(H[1-6]|STRONG)$/)));
+
+  const firstRowCols = rows[0].querySelectorAll(':scope > div');
+  const correspondingFirstRowCol = firstRowCols[cdx];
+  const hasColumnHeaderInFirstRow = correspondingFirstRowCol?.getAttribute('role') === 'columnheader';
+
+  return columnsWithStrongOrH.length === 1
+  && columnsWithStrongOrH[0] === col
+  && !hasColumnHeaderInFirstRow;
```
overmyheadandbody, line 16: Formatting nit, we usually indent multi-line condition checks
```suggestion
  return columnsWithStrongOrH.length === 1
    && columnsWithStrongOrH[0] === col
    && !hasColumnHeaderInFirstRow;
```

overmyheadandbody (approved):

PR #4643, opened 2025-08-01
MWPW-177279 [Nala]: Update price checks in tests

afmicka:
Currently CCLO launch is ongoing changing price values. These actions are happening occasionally and in these tests in not relevant to check exact price value but rather that price is rendered with correct currency and literals. Updating all commerce tests to test only that to prevent future unnecessary failures.

Resolves: [MWPW-177279](https://jira.corp.adobe.com/browse/MWPW-177279)

**Test URLs:**
- After: https://MWPW-177279--milo--afmicka.aem.live/?martech=off

### DIFF
--- a/nala/blocks/merch/three-in-one.test.js
+++ b/nala/blocks/merch/three-in-one.test.js
@@ -119,10 +119,10 @@ test.describe('ThreeInOne Block test suite', () => {
         await expect(cta).toHaveAttribute(key, value);
       }
       await cta.waitFor({ state: 'visible' });
-      
+
       // Wait for entitlements to load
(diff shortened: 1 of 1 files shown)


afmicka:
> looks great to me @afmicka , ideally we should have one constant AMERICAN_PRICE_PATTERN FRENCH_PRICE_PATTERN we reuse allover the place

We might but we also have prices with /mo and /yr. Let me check and update with constants in respective files. Thanks

npeltier (approved):

npeltier:
@afmicka we are good. As discussed offline, we could probably move all those patterns in one single central file later, but let's unblock nala for now :-)

PR #4645, opened 2025-08-01
[MWPW-177282] Use Akamai header to return user country

narcis-radu:
* Use Akamai header to retrieve user country instead of API call
* Cache result to avoid API calls performed by geo-routing and privacy

Resolves: [MWPW-177282](https://jira.corp.adobe.com/browse/MWPW-177282)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.live/drafts/narcis/default?martech=off
- After: https://geo--milo--adobecom.aem.live/drafts/narcis/default?martech=off

Note: The header is set by Akamai and cannot be tested on `aem.page` or `aem.live`. Feature testing should be performed on Stage and regression on `aem.live`. On `aem.page|live` the user country should be retrieved using API calls.


robert-bogos (approved):

overmyheadandbody (approved):

SilviuLCF (approved):
Verified, testing details https://jira.corp.adobe.com/browse/MWPW-177282

PR #4661, opened 2025-08-05
MWPW-177379 [MEP][MILO] Coordinate with launch updates for UK

vgoodric:
* Add consent checks to disable martech if needed
* Account for user's country

Resolves: [MWPW-177379](https://jira.corp.adobe.com/browse/MWPW-177379)

**Test URLs:**
- After: https://martechconsent--milo--adobecom.aem.page/?martech=off


libs/utils/utils.js
```diff
@@ -1391,7 +1391,71 @@ export function enablePersonalizationV2() {
   const enablePersV2 = getMepEnablement('personalization-v2');
   return !!enablePersV2 && isSignedOut();
 }
+export function getCookie(key) {
+  const cookie = document.cookie.split(';')
+    .map((x) => decodeURIComponent(x.trim()).split(/=(.*)/s))
+    .find(([k]) => k === key);
+  return cookie ? cookie[1] : null;
+}
+export async function determineCountry(searchParams = Object.fromEntries(PAGE_URL.searchParams)) {
```
narcis-radu, line 1400: Hey @vgoodric , have you seen the changes introduced in https://github.com/adobecom/milo/pull/4645? `setCountry` is scheduled to run first thing, the method will set the user country in `akamai` session storage object. I don't think this is needed, we should reuse the existing code.

libs/utils/utils.js
```diff
@@ -1391,7 +1391,71 @@ export function enablePersonalizationV2() {
...
+  const { akamaiLocale } = searchParams;
+  if (akamaiLocale) return akamaiLocale;
+
+  const { _satellite } = window;
```
narcis-radu, line 1404: I don't agree with this being in utils, it is not needed for all consumers.

libs/utils/utils.js
```diff
@@ -1391,7 +1391,71 @@ export function enablePersonalizationV2() {
...
+      country = code;
+    });
+  });
+  return country;
+}
+async function getCountry() {
+  const country = await determineCountry();
+  return country?.toLowerCase();
+}
+export async function getMartechConsent() {
```
narcis-radu, line 1432: This is martech specific, it should not be in utils, we cannot turn it off.

libs/utils/utils.js
```diff
@@ -1391,7 +1391,71 @@ export function enablePersonalizationV2() {
...
+export async function determineCountry(searchParams = Object.fromEntries(PAGE_URL.searchParams)) {
+  const { akamaiLocale } = searchParams;
+  if (akamaiLocale) return akamaiLocale;
+
+  const { _satellite } = window;
+  const serverTiming = _satellite?.getVar('serverTiming');
+  const serverTimingCountry = serverTiming?.geo;
+  if (serverTimingCountry) return serverTimingCountry;
+
+  let fedsLocation = sessionStorage.getItem('feds_location');
```
narcis-radu, line 1409: I would not rely on this value, any change in Privacy would basically require a change here as well.

libs/utils/utils.js
vgoodric: we do need to decide on C0002 consent to decide which MEP features to fire.  Names refactored to make the purpose more clear.  Please review.

libs/utils/utils.js
vgoodric: removed

libs/utils/utils.js
```diff
@@ -1391,38 +1391,82 @@ export function enablePersonalizationV2() {
...
+    .find(([k]) => k === key);
+  return cookie ? cookie[1] : null;
+}
+export function normCountry(country) {
+  if (!country) return null;
+  let c = country?.toLowerCase();
+  if (c === 'uk') c = 'gb';
+  return c.split('_')[0];
+}
+export async function getC0002(optin, akamaiLocale) {
```
narcis-radu, line 1406: We are running these methods for everyone, regardless of the user location, privacy initialization or consent categories. My suggestion is to write the code in a way that code is executed (and even imported) only if needed.

narcis-radu, line 1406: The method is a bit confusing: it returns a country as well as whether consent is provided. We should rewrite like
```
const country = await getCountry(); // promise that returns value from session storage or tries to fetch from geo2 API.
if (country !== 'gb') return;
const hasConsentedForC0002 = hasProvidedConsent() && getConsent().includes('C0002); //has provided consent -> `OptanonAlertBoxClosed` is defined and it is a proper date, `getConsent` -> returns the `groups=` value from `OptanonConsent` cookie and C0002:1 is there.

File: libs/utils/utils.js [hunk 6]
```diff
@@ -1391,38 +1391,82 @@ export function enablePersonalizationV2() {
...
+      .reduce((consentByPurpose, categoryPair) => {
+        const [key, value] = categoryPair.split('=');
+        consentByPurpose[key] = value;
+        return consentByPurpose;
+      }, {});
+    if (consent?.general === 'in') return { country, hasC0002: true };
+    if (consent?.general) return { country, hasC0002: false };
+  }
+  const { adobePrivacy } = window;
+  if ((adobePrivacy?.hasUserProvidedConsent() || adobePrivacy?.hasUserProvidedCustomConsent())) {
```
narcis-radu, line 1431: This is not needed if you're checking the cookies directly. If you need to wait for the consent signal, this is not the proper way to implement. Details are here: https://wiki.corp.adobe.com/display/adobedotcom/Consent+Management+Platform+Onboarding#ConsentManagementPlatformOnboarding-Usage

libs/utils/utils.js
```diff
@@ -1391,38 +1391,82 @@ export function enablePersonalizationV2() {
...
+  if (optin) {
+    const allowed = {
+      on: true,
+      off: false,
+      unset: country ? !explicitConsentCountries.includes(country) : true,
+    };
+    return { country: normCountry(country), hasC0002: allowed[optin] };
+  }
+
+  const kndctrCookie = getCookie('kndctr_9E1005A551ED61CA0A490D45_AdobeOrg_consent');
```
narcis-radu, line 1419: I am not familiar with this cookie, haven't seen it used before.

narcis-radu (changes requested):

PR #4675, opened 2025-08-06
[MWPW-176965] - faas form error announcement order

DKos95:
Resolves: [MWPW-176965](https://jira.corp.adobe.com/browse/MWPW-176965)

**PSI URLs:**
- After: https://faas-a11y-voiceover--milo--adobecom.aem.page/?martech=off

**Form Test URLs:**
- Before: https://main--cc--adobecom.aem.live/creativecloud/business/teams/request-information?georouting=off
- After: https://main--cc--adobecom.aem.live/creativecloud/business/teams/request-information?georouting=off&milolibs=faas-a11y-voiceover

### DIFF
--- a/libs/blocks/faas/faas.js
+++ b/libs/blocks/faas/faas.js
@@ -47,8 +47,11 @@ const createValidationObserver = () => new MutationObserver((mutations) => {
       field.setAttribute('aria-describedby', `${id}_em_`);
     });
 
-    const errorMessage = rowElement.querySelector('.errorMessage');
-    updateLiveRegion(errorMessage?.textContent);
+    const errorMessage = rowElement.querySelector('.errorMessage')?.textContent;
+
+    if (document.activeElement.parentElement.querySelector('.errorMessage')?.textContent
+        !== errorMessage) return;
+    updateLiveRegion(errorMessage);
   });
 });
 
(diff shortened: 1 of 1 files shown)


overmyheadandbody (changes requested):
This disables the original functionality of announcing errors when they pop up. Say I am focused on the First Name input, I don't add any text and tab to Last Name. Previously, the Last Name field would be announced, followed by the newly visible First Name error. That error no longer gets announced with this updated logic. It might be that a check for when the form is submitted be put in place to disable announcement of the last error from the form.

DKos95:
> Looking at this latest iteration, I'm noticing one other small issue - the error from the last focused item still gets announced after submit. I focused on Business e-mail, submitted the form without filling in any data, I got an announcement for the error from the first field and then for the one arising from the Business e-mail one. Is that intentional? At first glance, I'd say just the first error should be announced.

Well if we are on Business e-mail field and submit the form shouldn't we tell the user that the field in which he was when he hit submit cannot be empty, that seems logical to me? The reason it's announced later and not first is because of existing form functionality where when you submit the form and there is an error it focuses you on the first field that is required and empty, so because of this the screenreader announces that elements error first and then the second elements error on which you were when you hit enter.

SilviuLCF (approved):
Verified , testing details 

https://jira.corp.adobe.com/browse/MWPW-176965

PR #4687, opened 2025-08-07
MWPW-177105: Fixed plans education card layout

st-angelo-adobe:
Resolves: [MWPW-177105](https://jira.corp.adobe.com/browse/MWPW-177105)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/nala/features/commerce/plans2?martech=off
- After: https://mwpw-177105--milo--adobecom.aem.page/drafts/nala/features/commerce/plans2?martech=off


libs/features/mas/src/utils.js
```diff
@@ -81,3 +81,8 @@ export function discoverService(getConfig, { once = false } = {}) {
 export function getService() {
   return document.getElementsByTagName(MAS_COMMERCE_SERVICE)?.[0];
 }
+
+export function getOuterHeight(element) {
+    const style = window.getComputedStyle(element);
+    return element.offsetHeight + parseFloat(style.marginTop) + parseFloat(style.marginBottom);
+}
```
yesil, line 88: EOF

libs/features/mas/src/variants/plans.js
```diff
@@ -156,12 +157,65 @@ export class Plans extends VariantLayout {
...
+
+        /* Wait for legal prices to be ready */
+        const legals = this.card.querySelectorAll('[is="inline-price"][data-template="legal"]');
+        const legalSettledPromises = [];
+        for (const legal of legals) {
+            legalSettledPromises.push(legal.onceSettled());
+        }
+        await Promise.all(legalSettledPromises);
+        
+        const intersectionObs = new IntersectionObserver(([entry]) => {
```
yesil, line 184: Suggestion to move the intersection observer to module scope and let it observe only education cards.

libs/features/mas/src/variants/plans.js
```diff
@@ -156,12 +157,65 @@ export class Plans extends VariantLayout {
...
+
+    async postCardUpdateHook() {
         this.adaptForMedia();
         this.adjustTitleWidth();
-        this.adjustLegal();
         this.adjustAddon();
         this.adjustCallout();
+        if (!this.legalAdjusted) {
+            await this.adjustLegal();
+            await this.adjustEduLists();
```
yesil, line 217: `adjustEduLists` should not be dependent on `legalAdjusted`.

libs/features/mas/src/variants/plans.js
st-angelo-adobe: It only observes education cards already, at the start of the method there is this: `if (this.card.variant !== 'plans-education') return;`

I think maybe it's ok if we leave it like this, since I'm disconnecting it right away. If it was module scope I'd have to deal with multiple entries. I think it's simpler this way, but if there's a stronger argument for module scope I can do it.

bozojovicic (changes requested):
Spacer component now creates unwanted empty space in cards on mobile. The rest seems to be ok.

bozojovicic (approved):

afmicka:
@st-angelo-adobe  is there any reason why your changes would remove bold style from Refine your search text on catalog page? 
https://main--cc--adobecom.aem.live/products/catalog?milolibs=mwpw-177105
Comparing with stage milolibs and yours, i get this
Screenshot 2025-09-17 at 11 30 22
Screenshot 2025-09-17 at 11 30 18

PR #4688, opened 2025-08-07
Fix memory issues when fetching nested circular fragments

mokimo:
### Description

I've noticed that the DC rolling import flow failed due to running out of memory. https://github.com/adobecom/milo/actions/runs/16802946956/job/47588605400 

I remember a similar PR safeguard against circular fragment references which lead to issues on the FE code, so let's also safeguard it within the importer, in case a fragment is accidentally referring to itself causing a loop.

Resolves: [MWPW-174328](https://jira.corp.adobe.com/browse/MWPW-174328)

**Test URLs:**
- After: https://dont-crash-on-circular-fragments--milo--mokimo.aem.page/?martech=off

### DIFF
--- a/.github/workflows/import/index.js
+++ b/.github/workflows/import/index.js
@@ -166,9 +166,9 @@ function safeguardMetadataImages(dom) {
 const map = {}
 async function importUrl(aemPath, importedMedia) {
   const url = new URL(importFrom + aemPath.replace('.md', ''));
-    // AVOID Re-fetching the same resource (e.g. nested circular fragments)
-    if(map[importFrom + aemPath.replace('.md', '')]) return
-    map[importFrom + aemPath.replace('.md', '')] = true
+  // AVOID Re-fetching the same resource (e.g. nested circular fragments)
+  if(map[importFrom + aemPath.replace('.md', '')]) return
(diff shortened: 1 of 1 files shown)


TsayAdobe (approved):

.github/workflows/import/index.js
```diff
@@ -163,8 +163,12 @@ function safeguardMetadataImages(dom) {
   }
 }
 
+const map = {}
 async function importUrl(aemPath, importedMedia) {
   const url = new URL(importFrom + aemPath.replace('.md', ''));
+  // AVOID Re-fetching the same resource (e.g. nested circular fragments)
+  if(map[importFrom + aemPath.replace('.md', '')]) return
+  map[importFrom + aemPath.replace('.md', '')] = true
```
overmyheadandbody, line 171: Although this is a workflow file, I see it still has the ending semicolons, so maybe
```suggestion
  if(map[importFrom + aemPath.replace('.md', '')]) return;
  map[importFrom + aemPath.replace('.md', '')] = true;
```
It might also be worth defining `importFrom + aemPath.replace('.md', '')` as a constant to avoid some repetition.

overmyheadandbody (approved):
Sir Nit-a-lot is at it again :D

PR #4689, opened 2025-08-07
MWPW-177328 rework postProcessAutoblock

npeltier:
- all sub function now only work on cards (beside tabs analytics work moved to separate function),
- all sub function start once one card checkReady is called,
- fix few async / await,
- fix unit tests

Resolves: [MWPW-177328](https://jira.corp.adobe.com/browse/MWPW-177328)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.live/drafts/npeltier/plans?martech=off
- After: https://MWPW-177328--milo--npeltier.aem.live/drafts/npeltier/plans?martech=off


libs/blocks/merch/autoblock.js
```diff
@@ -55,49 +51,54 @@ export function handleCustomAnalyticsEvent(eventName, element) {
   }
 }
 
-export async function enableAnalytics(el, self) {
+export function cleanupTabsAnalytics(el) {
```
yesil, line 54: the fact that merch autoblock has tabs logic looks weird to me. the analytics logic should be on the tab side IMO.

test/blocks/merch-card-collection-autoblock/merch-card-collection-autoblock.test.js
```diff
@@ -13,6 +13,10 @@ const satellite = { track: sinon.spy() };
 const originalFetch = window.fetch;
 describe('merch-card-collection autoblock', () => {
   describe('init method', () => {
+    // Create mock mas-commerce-service element
+    const mockService = document.createElement('mas-commerce-service');
+    mockService.readyPromise = Promise.resolve();
```
yesil, line 18: readyPromise is not needed anymore.

yesil (approved):
left a comment for thought because I don't have enough understanding of tabs related requirements.
otherwise LGTM.

libs/blocks/merch/autoblock.js
3ch023: its a hack, yes :(
unfortunately its not so easy to fix tabs analytics since some teams already rely on the way it is working right now.
but for plans page we need a specific behaviour

afmicka:
@npeltier @3ch023 
1. Collection container lost its "daa-lh" when no filter is selected. Is it expected?
Screenshot 2025-08-08 at 14 41 01

main: 
Screenshot 2025-08-08 at 14 41 09

The rest looks good to me. @Roycethan these changes are touching accessibility, second pair of eyes might be good in that area. If the above is not an issue, please put the labels. I will be on PTO next week

npeltier:
> Collection container lost its "daa-lh" when no filter is selected. Is it expected?

@afmicka if no filter (esp. all filter) is selected, then yes, i would not expect daa-lh to equals "all--cat", but this is theorical example as we'll always end up having a selected filter

Roycethan:
> Can we plz see why analytics events are not getting fired for CTA's/Tabs/links ? May or may not related to this PR but this need to be working for go-live so better to find the reason.... Fired for Left filter categories and Addon checkbox Not getting fired for : Select CTA's / Modal links on cards Plans Tabs See Plans & Details link on card Links on the top and bottom sections like Generative credits,Leran more etc...

Since its also occuring on stage/main urls ill create a separate ticket for now and mark this Ready for stage 
Plz investigate this: https://jira.corp.adobe.com/browse/MWPW-177976 
Thanks

PR #4697, opened 2025-08-08
Section up to list

zagi25:
Issue: sections with `-up` class, containing `action-item` and `icon-block`, visually looks like a list but are not marked as such.

* Add list roles to sections with `-up` class containing `action-item` and `icon-block` excluding sections that have headings
* Add list roles to `action-scroller`
* Add tests

Resolves: [MWPW-177447](https://jira.corp.adobe.com/browse/MWPW-177447), [MWPW-177446](https://jira.corp.adobe.com/browse/MWPW-177446)
**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/ratko/mwpw-177446-to-list/to-list?martech=off
- After: https://section-up-to-list--milo--adobecom.aem.page/drafts/ratko/mwpw-177446-to-list/to-list?martech=off
- Before: https://main--milo--adobecom.aem.page/docs/library/kitchen-sink/action-items?martech=off
- After: https://section-up-to-list--milo--adobecom.aem.page/docs/library/kitchen-sink/action-items?martech=off
- Before: https://main--milo--adobecom.aem.page/docs/library/kitchen-sink/icon-block?martech=off
- After: https://section-up-to-list--milo--adobecom.aem.page/docs/library/kitchen-sink/icon-block?martech=off

**DC URLs:**
- Before: https://main--dc--adobecom.aem.page/de/acrobat
- After: https://main--dc--adobecom.aem.page/de/acrobat?milolibs=section-up-to-list
- Before: https://main--dc--adobecom.aem.page/fr/acrobat/about-adobe-pdf
- After: https://main--dc--adobecom.aem.page/fr/acrobat/about-adobe-pdf?milolibs=section-up-to-list


libs/scripts/accessibility.js
```diff
@@ -97,6 +97,18 @@ export const setDialogAndElementAttributes = ({ element, title }) => {
   element.closest('.dialog-modal')?.setAttribute('aria-label', title);
 };
 
+export function setListRoles() {
+  const selector = ':is(.section[class*="-up"]:has(.action-item, .icon-block), .scroller):not(:has(h1, h2, h3, h4, h5, h6))';
```
overmyheadandbody, line 101: Looking at the selector, I'm thinking it might need a bit more protection - are the action-item/icon-block elements usually direct children of the section? Judging by the the `[...parent.children]` on L104, I'd assume so. If so, maybe a `>` child combinator could make things more specific, to avoid applying this rule to sections that happen to have a single action item _somewhere_. Speaking of 'a single action item', should we check that there are multiple in order to apply the rule? Something as simple as `:nth-child(2)` within the `:has` braces could account for this as well.

overmyheadandbody (commented):
Some notes on the selector to make it a bit more specific, otherwise looking good 👌

overmyheadandbody (approved):
Looks good, I agree that this can live here and not necessarily in `accessibility.js`

PR #4710, opened 2025-08-12
[MWPW-178053] Show copy tooltip when its content is hovered

overmyheadandbody:
This allows the copy button tooltip to be visible when its content is being hovered, required for accessibility purposes.

I've decided to completely remove the `pointer-events` declarations, as their only purpose was to disallow executing the copy logic when the tooltip content was clicked. This seems like an edge case that has no real implications, thus the simplest solution was to remove these. An easy alternative would be to use `layerY` ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/layerY)) to figure out if the click occurred on the element itself or one of its pseudo-elements, but this is a non-standard feature and implementing it doesn't bring enough advantages to grant its use. I'm open to other opinions, though 🙂 

Resolves: [MWPW-178053](https://jira.corp.adobe.com/browse/MWPW-178053)

**Test URLs:**
- Before: https://main--cc--adobecom.aem.page/it/creativecloud/file-types/image/vector?martech=off
- After: https://main--cc--adobecom.aem.page/it/creativecloud/file-types/image/vector?milolibs=share-copy-a11y--milo--overmyheadandbody&martech=off

### DIFF
--- a/libs/blocks/share/share.css
+++ b/libs/blocks/share/share.css
@@ -85,8 +85,7 @@
 
 .share .copy-to-clipboard::before {
   content: attr(data-copy-to-clipboard);
-  display: block;        
-  pointer-events: none;
+  display: block;
   position : absolute;
   bottom: 100%;
   left: -45%;
@@ -119,7 +118,6 @@
 .share .copy-to-clipboard::after {
   content: '';
   display: block;
-  pointer-events: none;
   position : absolute;
   bottom: 100%;
   left: 50%;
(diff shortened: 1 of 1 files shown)


zagi25 (approved):
This seems to be duplicate of [MWPW-176923](https://jira.corp.adobe.com/browse/MWPW-176923) ( it looks like this ticket was created before PR went to main). But I would say that it works better without `pointer-events`.

PR #4716, opened 2025-08-13
MWPW-178057 Tab state is not retained when navigating back

bozojovicic:
Save the tab selection in session storage and recover it on the next page load (page refresh, navigating back) if the meta parameter `remember` is authored/set on tabs component to value `'on'`. We can set this parameter on plans page so other pages will not be affected by this change.

Resolves: [MWPW-178057](https://jira.corp.adobe.com/browse/MWPW-178057)

Test page `main--cc--adobecom.aem.live/creativecloud/plans?milolibs=mwpw178057tabstate--milo--bozojovicic`
but `remember` is not set yet

On the page bellow we have multiple tabs on the same page. State is saved in sessionStorage in items `/drafts/bozo/pr/tabs/mini-compare-tab-state` and `/drafts/bozo/pr/tabs/other-tabs-tab-state`

**Test URLs:**
- Before: https://main--milo--adobecom.aem.live/drafts/bozo/pr/tabs?martech=off
- After: https://mwpw178057tabstate--milo--bozojovicic.aem.live/drafts/bozo/pr/tabs?martech=off


libs/blocks/tabs/tabs.js
```diff
@@ -57,6 +57,19 @@ export function getRedirectionUrl(linkedTabsList, targetId) {
   return currentUrl;
 }
 
+const loadActiveTabFromStorage = () => {
+  const activeTab = localStorage.getItem('active-tab');
```
yesil, line 61: Is the requirement to make the tab selection persistent across visits? if it is only for the current navigation, we should be using sessionStorage.

libs/blocks/tabs/tabs.js
```diff
@@ -57,7 +57,53 @@ export function getRedirectionUrl(linkedTabsList, targetId) {
   return currentUrl;
 }
 
-function changeTabs(e) {
+const loadActiveTabFromStorage = () => {
+  const activeTab = sessionStorage.getItem('active-tab');
```
yesil, line 61: when there are multiple tab containers in the page, this can be confusing, I think the key should be contructed from the console.

I would be more cautious here, and require a name, to make it persistent.
basically GWP should author the name in the block.

libs/blocks/tabs/tabs.js
```diff
@@ -57,7 +57,28 @@ export function getRedirectionUrl(linkedTabsList, targetId) {
...
+const generateStorageName = (tabId) => {
+  const { pathname } = window.location;
+  return `${pathname}/${tabId}-tab-state`;
+};
+
+const loadActiveTab = (config) => {
+  if (config['remember-selected-tab'] !== 'on') return 0;
+
+  const tabId = config['tab-id'];
+  return sessionStorage.getItem(generateStorageName(tabId));
```
yesil, line 69: fallback to 0 if it doesn't exist in sessionStorage?

libs/blocks/tabs/tabs.js
```diff
@@ -57,7 +57,28 @@ export function getRedirectionUrl(linkedTabsList, targetId) {
...
 }
 
-function changeTabs(e) {
+const generateStorageName = (tabId) => {
+  const { pathname } = window.location;
+  return `${pathname}/${tabId}-tab-state`;
+};
+
+const loadActiveTab = (config) => {
+  if (config['remember-selected-tab'] !== 'on') return 0;
```
yesil, line 66: We need to check if they prefer an explicit long token name such as `remember-selected-tab` which might be challenging to remember, or a shorthand version which can be adapted later in other places, aka : `remember`

libs/blocks/tabs/tabs.js
bozojovicic: This is the method call.
```
const activeTabIndex = loadActiveTab(config);
if (activeTabIndex) config['active-tab'] = activeTabIndex;
```
Zero is not the valid tab index, it starts from 1.
Return value 0 just means that the value configured in the document for "active-tab" will be used and sessionStorage will be ignored.

PR #4717, opened 2025-08-13
[MWPW-177996] Use heading for share block title

overmyheadandbody:
This ensures that the text above the share icons from the Share block is announced as a heading. An option would be to actually use a heading element, but that leads to different styling needs, so I've chosen to address the issue via attributes on the existing element.

An interesting addition is the `getPrevHeadingLevel` method, which tries to find the heading level of the first `h` tag preceding a given element. This could prove useful in the future, since Adobe Accessibility rules state that the heading hierarchy should be clear on the page (i.e. no jumping from `h2` to `h4` or viceversa).

LE: After @zagi25's note, I noticed a few other small details to address:
* the `tracking-header` class was applied inconsistently, sometimes it didn't even include the heading text;
* the code intentionally replaced authored tags with a paragraph for headings. During the Accessibility work, it became clear that this pattern should be avoided and authors should have control over the tags they use;
* the logic was a bit convoluted, I rewrote it to include only the useful use-cases.

Resolves: [MWPW-177996](https://jira.corp.adobe.com/browse/MWPW-177996)

**Test URLs:**
- Before: https://main--milo--overmyheadandbody.aem.page/drafts/ramuntea/share-scenarios?martech=off
- After: https://share-heading-a11y--milo--overmyheadandbody.aem.page/drafts/ramuntea/share-scenarios?martech=off
- Before (original issue): https://main--cc--adobecom.aem.page/es/creativecloud/design/discover/golden-ratio?martech=off
- After (original issue): https://main--cc--adobecom.aem.page/es/creativecloud/design/discover/golden-ratio?milolibs=share-heading-a11y--milo--overmyheadandbody&martech=off

### DIFF
--- a/libs/blocks/share/share.css
+++ b/libs/blocks/share/share.css
@@ -45,8 +45,8 @@
   gap: 18px;
 }
 
-.share p {
-  padding-bottom: 8px;
+.share :is([role="heading"], h1, h2, h3, h4, h5, h6) {
+  padding-block-end: 8px;
 }
 
 .share.dark p {
--- a/libs/blocks/share/share.js
+++ b/libs/blocks/share/share.js
@@ -60,39 +60,34 @@ function getPrevHeadingLevel(block) {
   return prevHeadingTag.replace('h', '');
 }
 
+function toSentenceCase(str) {
+  if (!str || typeof str !== 'string') return '';
+  /* eslint-disable-next-line no-useless-escape */
+  return str.toLowerCase().replace(/(^\s*\w|[\.\!\?]\s*\w)/g, (c) => c.toUpperCase());
+}
+
 export default async function decorate(block) {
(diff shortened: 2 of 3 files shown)


zagi25 (approved):

overmyheadandbody:
@zagi25 - thanks for noticing! Fair enough, I see the spacing in the Figma file as well, although I suspect it was added after the fact. Nevertheless, I've updated the CSS declaration. There's still some quirkiness though, as everything is based on browser defaults using `em`, so it might be that this needs to be properly revisited in the future to ensure consistency across browsers and account for users' browser settings.

DKos95 (approved):

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-177996

PR #4718, opened 2025-08-13
[MWPW-167915] - faas autocomplete preserve

DKos95:
On PROD (Firefox) https://www.adobe.com/creativecloud/business/teams/request-information.html there is this dependency file https://tag.demandbase.com/shared/forms_45169bf05c.min.js that changes autocomplete to off on faas forms, since we can't change that file we have to create a mechanism to preserve these attributes if they are changed.

**QA Note**
Since this dependency file is only available on PROD we can't directly test this issue with that file changing the autocomplete to off, but we can emulate this behavior by applying a script in the browser console and trying to change the autocomplete ourselves to off and seeing if they are preserved, the following script can be used to test this:

`document.querySelectorAll('.faas-form input:not([type="hidden"]), .faas-form textarea').forEach(el => el.setAttribute('autocomplete', 'off'));`

Resolves: [MWPW-167915](https://jira.corp.adobe.com/browse/MWPW-167915)

**PSI Test URLs:**
- After: https://faas-autocomplete-demandbase--milo--adobecom.aem.page/?martech=off

**Test URLs:**
- Before: https://main--cc--adobecom.aem.live/creativecloud/business/teams/request-information?georouting=off
- After: https://main--cc--adobecom.aem.live/creativecloud/business/teams/request-information?georouting=off&milolibs=faas-autocomplete-demandbase

### DIFF
--- a/libs/blocks/faas/faas.js
+++ b/libs/blocks/faas/faas.js
@@ -75,14 +75,11 @@ const createValidationObserver = () => new MutationObserver((mutations) => {
 const createAutocompleteObserver = () => (
   new MutationObserver((mutations) => {
     mutations.forEach((mutation) => {
-      if (mutation.type !== 'attributes') return;
-
       const inputElement = mutation.target;
-
-      if (inputElement.getAttribute('autocomplete') !== 'off') return;
+      if (mutation.type !== 'attributes' || inputElement.getAttribute('autocomplete') !== 'off') return;
 
       const originalValue = inputElement.dataset.originalAutocomplete;
-      if (originalValue === undefined) return;
+      if (originalValue === '') return;
       inputElement.setAttribute('autocomplete', originalValue);
     });
   })
--- a/libs/blocks/faas/faas.js
+++ b/libs/blocks/faas/faas.js
@@ -72,6 +72,33 @@ const createValidationObserver = () => new MutationObserver((mutations) => {
   });
 });
 
+const createAutocompleteObserver = () => (
+  new MutationObserver((mutations) => {
+    mutations.forEach((mutation) => {
+      if (mutation.type !== 'attributes') return;
+
+      const inputElement = mutation.target;
+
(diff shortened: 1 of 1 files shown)


mokimo (approved):
Wasn't this fixed as per @overmyheadandbody comment?

> This seems to have been fixed, autocomplete is set to organization.

Seems a bit weird to me to overwrite whatever the faas script is doing not sure who maintains that, ideally it should be fixed there...

I don't see any blockers on merging per-se though, hence I'll approve. Would be great if Rares can have some eyes on this as he recently worked on this block

DKos95:
> Wasn't this fixed as per @overmyheadandbody comment?
> 
> > This seems to have been fixed, autocomplete is set to organization.
> 
> Seems a bit weird to me to overwrite whatever the faas script is doing not sure who maintains that, ideally it should be fixed there...
> 
> I don't see any blockers on merging per-se though, hence I'll approve. Would be great if Rares can have some eyes on this as he recently worked on this block

It probably seemed that it was fixed but that was because it happens on some browsers and also depends on how fast you are to check because this script takes a while to execute and change the autocomplete.


https://github.com/user-attachments/assets/0689d5ca-b232-454f-8543-0abfe0c16f8d

robert-bogos (approved):
code looks good 👍

overmyheadandbody:
@zagi25 might have a good point with this being intentional, I hadn't noticed it before, but it does make sense. I'm not sure how to best treat this, though. I will bring this up with Taryn to see if we can make the field accessible, while still allowing for the custom autocomplete logic.

overmyheadandbody (changes requested):
Blocking for now, I think Ratko makes a great point, I'd like to confirm the scenario before moving forward with this. I'll have an answer by Thursday evening on this.

SilviuLCF (approved):
Verified , testing details 
https://jira.corp.adobe.com/browse/MWPW-167915

PR #4719, opened 2025-08-13
MWPW-178287: Fix horizontal scroll in plans collection

Axelcureno:
Fixes horizontal scroll in plans collection. 

Resolves: [MWPW-178287](https://jira.corp.adobe.com/browse/MWPW-178287)

Plans URL: `https://main--cc--adobecom.aem.page/creativecloud/plans?milolibs=MWPW-178287`

**Test URLs:**
- After: https://MWPW-178287--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/libs/deps/mas/merch-card.js
+++ b/libs/deps/mas/merch-card.js
@@ -1495,7 +1495,7 @@ merch-card-collection:has([slot="subtitle"]) merch-card {
     }
 
     .collection-container.plans:has(merch-sidenav) {
-        width: 90vw;
+        width: fit-content;
         position: relative;
         left: 50%;
         transform: translateX(-50vw);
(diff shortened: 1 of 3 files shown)


libs/deps/mas/merch-card.js
```diff
@@ -1495,7 +1495,7 @@ merch-card-collection:has([slot="subtitle"]) merch-card {
     }
 
     .collection-container.plans:has(merch-sidenav) {
-        width: 100vw;
+        width: 90vw;
```
st-angelo-adobe, line 1498: This works but the real issue is the padding, that isn't being included in the `100vw` calculation, which means the horizontal scroll is basically the two (left & right) paddings overflowing. If we increase the padding, chances are we'd have to modify this value again, which is not good.

I think just adding `box-sizing: border-box` will do the trick.

st-angelo-adobe (changes requested):

seanchoi-dev (approved):

st-angelo-adobe (approved):

PR #4739, opened 2025-08-19
mWeb Ps Journey | Section Metdata - Expand functionality

suhjainadobe:
mWeb Ps Journey | Section Metdata - Expand functionality

 * Add a see-more-feature button in the section-metadata, which will appear when we have more than 3 blocks inside the section. 
 *Initially, only three blocks are visible, and clicking the button expands the section to reveal the remaining blocks.

Resolves: [https://jira.corp.adobe.com/browse/MWPW-177520](https://jira.corp.adobe.com/browse/MWPW-NUMBER)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/suhjain/m2-folder/editorialcards/document
- After: https://mwpw-177520--milo--suhjainadobe.aem.page/drafts/suhjain/m2-folder/editorialcards/document

**Testing Notes**
1: Initially, only three blocks are visible, and clicking the button expands the section to reveal the remaining blocks.
2: After we click the button, the button shold get removed for that session.
3: It is applicable for all the blocks in the section.


libs/blocks/section-metadata/section-metadata.js
```diff
@@ -1,4 +1,10 @@
 import { handleFocalpoint } from '../../utils/decorate.js';
+import { createTag, getConfig } from '../../utils/utils.js';
+import { decorateDefaultLinkAnalytics } from '../../martech/attributes.js';
+import { replaceKey } from '../../features/placeholders.js';
+
+const replacePlaceholder = async (key) => replaceKey(key, getConfig());
```
overmyheadandbody, line 6: I still don't quite understand why this isn't fetched from the Federated content, something like:
```suggestion
import { createTag, getConfig, getFedsPlaceholderConfig } from '../../utils/utils.js';
import { decorateDefaultLinkAnalytics } from '../../martech/attributes.js';
import { replaceKey } from '../../features/placeholders.js';

const replacePlaceholder = async (key) => replaceKey(key, getFedsPlaceholderConfig());
```
Trying it locally yields the expected results.

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -71,6 +77,38 @@ export const getMetadata = (el) => [...el.childNodes].reduce((rdx, row) => {
...
 
+async function handleShowMoreButton(section) {
+  const showMoreBtn = section.querySelector('.show-more-button button');
+  if (!showMoreBtn) return;
+  const iconSpan = createTag('span', {
+    class: 'show-more-icon',
+    'aria-hidden': 'true',
+  }, `${ADD_MORE_ICON}`);
+
+  showMoreBtn.innerHTML = `${await replacePlaceholder('see-more-features')}`;
```
overmyheadandbody, line 88: Kodiak will mark this as a security issue that could easily be avoided by attaching the placeholder where the button is created in `handleCollapseSection`.

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -71,6 +77,38 @@ export const getMetadata = (el) => [...el.childNodes].reduce((rdx, row) => {
   return rdx;
 }, {});
 
+async function handleShowMoreButton(section) {
+  const showMoreBtn = section.querySelector('.show-more-button button');
```
overmyheadandbody, line 81: You could add the button as an argument. My rule of thumb is to avoid querying for elements we create, as much as possible, since we should already have references to them.

overmyheadandbody (commented):
Looking good for the most part, but I do have a few comments with some improvement ideas

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -71,6 +77,37 @@ export const getMetadata = (el) => [...el.childNodes].reduce((rdx, row) => {
...
+    section.classList.add('show-all');
+    section.querySelector('.show-more-button').remove(); // Remove the entire show-more-button div
+  });
+}
+
+async function handleCollapseSection(text, section) {
+  if (!text || !section) return;
+  const blocks = section.querySelectorAll('div:not(:last-child)');
+  const existingShowMoreButton = section.querySelector('.show-more-button');
+  if (text === 'on' && blocks.length > 3 && !existingShowMoreButton) {
```
zagi25, line 99: Can we maybe avoid unnecessary indentation like this: 
```suggestion
  const blocks = section?.querySelectorAll('div:not(:last-child)');
  const existingShowMoreButton = section?.querySelector('.show-more-button');
  if (text !== 'on' || !blocks.length <= 3 || existingShowMoreButton) return;
```

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -95,5 +132,7 @@ export default async function init(el) {
   if (metadata.masonry) handleMasonry(metadata.masonry.text, section);
   if (metadata.delay) handleDelay(metadata.delay.text, section);
   if (metadata.anchor) handleAnchor(metadata.anchor.text, section);
+  // eslint-disable-next-line max-len
+  if (metadata['collapse-ups-mobile']) await handleCollapseSection(metadata['collapse-ups-mobile'].text, section);
```
zagi25, line 136: If `text` is only used to check for `on` value, I would say it is not necessary as a param 
```suggestion
  if (metadata['collapse-ups-mobile']?.text === 'on') await handleCollapseSection(section);
```

zagi25 (commented):
Maybe placeholder value should be updated so it doesn't contains dashes ? 
Screenshot 2025-09-03 at 15 05 43

libs/blocks/section-metadata/section-metadata.css
```diff
@@ -239,6 +270,22 @@ main > .section[class*='-up'] > .content {
   .section[class*='grid-width-'] {
     display: block;
   }
+
+  .section:has(.con-block ~ .show-more-button),
+  .section.show-all:has(.con-block) {
+    margin-inline: 30px;
+    padding: var(--spacing-xs);
+    margin-bottom: 24px;
```
zagi25, line 278: Should there be a `margin-top` also ? 
Screenshot 2025-09-03 at 15 11 35
Screenshot 2025-09-03 at 15 11 30

libs/blocks/section-metadata/section-metadata.css
```diff
@@ -257,13 +304,31 @@ main > .section[class*='-up'] > .content {
...
     padding-right: calc((100vw - 600px) / 2);
   }
+
+  .section .show-more-button {
+    display: none;
+  }
+
+  .section:has(.con-block ~ .show-more-button),
+  .section.show-all:has(.con-block) {
+    background: none !important;
```
zagi25, line 330: Instead of using `!important`. Can we handle this in code ? We can possibly skip [this](collapse-ups-mobile) if there is `collapse-ups-mobile`. Same goes for L314

libs/blocks/section-metadata/section-metadata.css
```diff
@@ -226,6 +226,37 @@ main > .section[class*='-up'] > .content {
...
+  border-radius: 8px;
+}
+
+.section .show-more-button button {
+  background-color: var(--color-gray-100);
+  border: none;
+  color: #070707;
+  text-decoration: none;
+  cursor: pointer;
+  font-size: 18px;
```
overmyheadandbody, line 244: Why not `--type-body-m-size`. When an eventual redesign comes, it will be easier to make global updates, rather than having to edit each file.

libs/blocks/section-metadata/section-metadata.css
```diff
@@ -226,6 +226,37 @@ main > .section[class*='-up'] > .content {
...
+  font-style: normal;
```
overmyheadandbody, line 245: This is the default value of this property, is this actually needed?

libs/blocks/section-metadata/section-metadata.css
```diff
@@ -226,6 +226,37 @@ main > .section[class*='-up'] > .content {
...
+  font-weight: 700;
+  line-height: 22px;
+  padding: var(--spacing-xxs) var(--spacing-s);
+  border-radius: 40px;
```
overmyheadandbody, line 249: I'm not sure I get how `background-color` and `border-radius` interact. The current experience has an authored background color for the section, while the button's color is hardcoded. They now happen to match, but that might not always be the case. So my questions are:
* is the expectation for them to always match? If so, the background color should be inherited from the section. Also, the border radius doesn't achieve much, as the button never contrasts its parent
* is the expectation that the colors can be different and that the button needs to have the current gray color? Something like:

Image

libs/blocks/section-metadata/section-metadata.css
overmyheadandbody: I'd actually be curious as to why a bottom margin would be needed in the first place, as sections can benefit from variants that can set particular margins. Hardcoding this here would disable that capability, which sounds undesirable.

libs/blocks/section-metadata/section-metadata.css
overmyheadandbody: Additionally, why is it needed for 2 breakpoints? Disabling above `600px` will disable it above `720px` as well, right?

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -71,6 +77,38 @@ export const getMetadata = (el) => [...el.childNodes].reduce((rdx, row) => {
...
+  const showMoreButton = createTag('div', { class: 'show-more-button' });
+  const button = createTag('button', {}, '');
+  button.innerHTML = seeMoreText;
+
+  const iconSpan = createTag('span', {
+    class: 'show-more-icon',
+    'aria-hidden': 'true',
+  }, `${ADD_MORE_ICON}`);
+  button.appendChild(iconSpan);
+  button.setAttribute('aria-label', seeMoreText);
```
overmyheadandbody, line 91: This should also work and is a bit more brief and clear. `aria-label` should not be needed on the button, since text is already nested inside.
```suggestion
  const button = createTag('button', {}, seeMoreText);

  const iconSpan = createTag('span', {
    class: 'show-more-icon',
    'aria-hidden': 'true',
  }, `${ADD_MORE_ICON}`);
  button.appendChild(iconSpan);
```

overmyheadandbody (commented):
I have a few notes and I see Ratko made some good points as well. Additionally, I see NALA tests are failing, please try rebasing with `stage` and/or rerun the tests so all checks pass.

libs/blocks/section-metadata/section-metadata.css
suhjainadobe: The requirement was that the background color should be authorable, and the button color and radius should follow according to the figma. Ill re-confirm this again with Joel.

libs/blocks/section-metadata/section-metadata.js
```diff
@@ -1,4 +1,10 @@
 import { handleFocalpoint } from '../../utils/decorate.js';
+import { createTag, getFedsPlaceholderConfig } from '../../utils/utils.js';
+import { decorateDefaultLinkAnalytics } from '../../martech/attributes.js';
+import { replaceKey } from '../../features/placeholders.js';
```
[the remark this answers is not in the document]
suhjainadobe: I am now dynamically importing the placeholder file only when it is needed, so that it does not load on every section-metadata call. It is needed as we cannot author the text for button on section-metadata.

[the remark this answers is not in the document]
JasonHowellSlavin: Nice! Can the same be done for `decorateDefaultLinkAnalytics` since it is also only invoked if the `if (metadata['collapse-ups-mobile']?.text === 'on')` conditional is true?

spadmasa:
Issue is seen when the fragment is embedded and tried , once its fixed will give heads up on the pr
https://main--cc--adobecom.aem.live/drafts/souj/mweb/expandnew?milolibs=mwpw-177520--milo--suhjainadobe

PR #4743, opened 2025-08-20
MWPW-178111 [Plans] Quantity selector - accessibility issue

bozojovicic:
Fix A11Y issues in Quantity selector component.

Button is there but it is not accessible since Taryn said that the QS component should have only one accessible field. 
The arrow icon is changed in the button when the popup is expanded/collapsed.
Popup is expanded upon user's click anywhere on the quantity selector, or when the text field gets focus and then user presses the arrow down key.
Then user can use arrows to navigate between options 1-10 and can use Enter/Tab/Space to select the value. User can also enter any numerical value in the field.
The selected value is now marked with the Checkmark icon.
User can use Escape to hide the dropdown without changing the selected value.
Type of the field is now "text" instead of "number" and validation is done manually, as Taryn suggested.

Screenshot 2025-08-20 at 17 23 24

Screen reader now reads all the important info, the component name, the highlighted value, the selected value, collapsed/expanded state etc.

Test link : `main--cc--adobecom.aem.live/creativecloud/plans?milolibs=mwpw178111a11y--milo--bozojovicic`

Resolves: [MWPW-178111](https://jira.corp.adobe.com/browse/MWPW-178111)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.live/drafts/bozo/pr/qs?martech=off
- After: https://mwpw178111a11y--milo--bozojovicic.aem.live/drafts/bozo/pr/qs?martech=off


3ch023 (approved):

yesil (approved):
I can enter even 1000+, is there an upper bound that can be set?

Roycethan:
> @Roycethan
> 
> 1. I messed this up with fixes for other problems. Fixed now.
> 2. Ok but we need some ability to open it on mouse click. So now if we click on text field the popup is not displayed. If we click on button it is displayed/hidden. Actually that's what we have on live now so it should be good now.
> 3. and 4) This is where the problem starts. According to Taryn the QS should be one component only. First she proposed to use the ordinary HTML select component. When I explained why that is not possible she proposed to use the text field only and remove the button. But that's not what is requested in figma. So she proposed a compromise, to make it look like this component https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-both/ as much as possible. That's why the button is not accessible with tab. It's just displayed and if you click on it the popup will show/hide.
>    I added `role="presentation"` to that button. Is it OK now? If not, can you tell me what tool are you using to detect A11Y issues so I can try to fix that "There's no focus on the interactive element" problem.



Thanks @bozojovicic  1 and 2 fixed . 
For 3 and 4 i think if theres no problems you can fix by:
Adding **tabindex=0** attribute value and [**:focus](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus) and [:focus-visible**](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) CSS pseudo-classes.
image
Theres good docs available at https://developer.mozilla.org/en-US/docs/Web/Accessibility/Guides/Understanding_WCAG/Keyboard?utm_source=devtools&utm_medium=a11y-panel-checks-keyboard#Interactive_elements_must_be_focusable

#5: Also in ticket Taryn expects Screen readers to hear  "Select number of licences in quantity selector in Business tab"
Can you plz check if "Business tab" part also can be added  for screen readers so that they know thay are in this tab while selecting the Quantity.


cc @3ch023

Roycethan:
> @Roycethan there is no need to add tabindex=0 to the button. Button is focusable/tabbable by default. The problem is that Taryn doesn't want it to be tabbable. She said that QS should be one component only and in this way it is made from 2 components. I explained that QS component is much more complicated than ordinary select box and it has to be like this. Anyway, now the button is focusable/tabbable and according to [this](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Guides/Understanding_WCAG/Keyboard#clickable_elements_must_be_focusable_and_should_have_interactive_semantics) we made it correctly. If Taryn complains again we will have to groom all this again.
> 
> For #5 I can't do anything. All the necessary aria attributes are already there and screen reader announces for the CTA/links that we are in Business tab but for QS it doesn't do that. I tried everything that I could find.

@bozojovicic  Srry but Axe tool is complaining now:
Ill share here on what its complaining:
image

Ensure an element's role supports its ARIA attributes
Element Location:
#\38 2cf8b24-980f-4dae-a8a6-df84b7f15c60 > div[slot="quantity-select"] > merch-quantity-select,.picker-button[aria-controls="qsPopover"][aria-labelledby="qsLabel"]
<button class="picker-button" aria-controls="qsPopover" aria-labelledby="qsLabel" aria-activedescendant="" aria-expanded="false">
                    <div class="picker-button-fill open"></div>
                </button>

To solve this problem, you need to fix the following:
ARIA attribute is not allowed: aria-activedescendant=""
https://dequeuniversity.com/rules/axe/4.10/aria-allowed-attr?application=AxeChrome

bozojovicic:
@Roycethan I fixed it

Roycethan:
> @Roycethan I fixed it

Thanks @bozojovicic  for the fix ... Plz pull in the latest from stage , so that Nala can go through...

PR #4745, opened 2025-08-20
 Incorrect scroll position after closing a modal with duplicate hash values

suhjainadobe:
When a modal is closed, the page scrolls to an unintended position if the same modal hash is used earlier in the page.

Resolves: [MWPW-169954](https://jira.corp.adobe.com/browse/MWPW-NUMBER)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/suhjain/hero-marquee-bg-bottom-mobile
- After: https://scroll-position--milo--suhjainadobe.aem.page/drafts/suhjain/hero-marquee-bg-bottom-mobile

Click on buy-now and free-trial buttons on the first and last section. The page should retain the scroll position.


libs/utils/utils.js
```diff
@@ -888,6 +888,9 @@ export function decorateAutoBlock(a) {
         a.dataset.modalHash = url.hash;
         a.href = url.hash;
         a.className = `modal link-block ${[...a.classList].join(' ')}`;
+        a.addEventListener('click', () => {
```
zagi25, line 891: Instead of adding click listener, maybe you can add something like: 

`document.activeElement.dataset.isModalTrigger = 'true';` [here](https://github.com/adobecom/milo/blob/stage/libs/blocks/modal/modal.js#L159)

`document.activeElement` gives you the link that was clicked to open the modal. And this will also cover all modal openings, so there is no need for change in `merch.js`

zagi25: No, idea is that when user clicks/open modal with keyboard, that link/button is focused and is current activeElement in the document. And as soon as modal.js code is called, we can mark that element as trigger because nothing else will be focused in the meantime. I just tried it briefly so there are possible edge-cases and I didn't think about deeplink modals, but this way there is no need to update other parts of the codebase (e.g. `merch.js`)

libs/blocks/modal/modal.js
```diff
@@ -49,6 +49,12 @@ export function sendAnalytics(event) {
   }
 }
 
+function focusTriggerElement(modalId) {
```
zagi25, line 52: This looks good, but I think we need to keep 
```
document.querySelector(`[data-modal-hash="#${mod.id}"]`)?.focus();
and 
document.querySelector(`a[data-modal-id="${id}"].con-button`)?.focus();
```
 in addition to the `trigger` logic.
Because if the modal is open via deep link, `trigger` attribute will not be set properly and modal button will not be focused after closing.

robert-bogos (approved):
code looks good 👍

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details [MWPW-169954](https://jira.corp.adobe.com/browse/MWPW-169954)

PR #4760, opened 2025-08-25
MWPW-178730: add fallback for structuredClone for older safari versions

markovicpredrag1209:
* Fixes a bug where inline price is not showing because structuredClone is not available for older Safari version

Resolves: [MWPW-178730](https://jira.corp.adobe.com/browse/MWPW-178730)

**Test URLs:**
- After: https://MWPW-178730--milo--markovicpredrag1209.aem.live/?martech=off
-
Simulator recording attached

https://github.com/user-attachments/assets/f649d352-6cc7-4f03-84ce-aa64dfc26457


libs/blocks/merch/merch.js
```diff
@@ -468,6 +468,13 @@ export async function polyfills() {
       isSupported = true;
     },
   });
+  
+  if (!globalThis.structuredClone) {
+    globalThis.structuredClone = function(obj) {
```
yesil: I think this is not needed here.

libs/features/mas/src/price.js
```diff
@@ -24,11 +24,19 @@ import { toOfferSelectorIds, toQuantity } from './utilities.js';
...
+            }
+            return JSON.parse(JSON.stringify(obj));
+        };
+        
         let options = {
             country: settings.country,
             language: settings.language,
             locale: settings.locale,
-            literals: structuredClone(literals.price),
+            literals: deepClone(literals.price),
```
yesil, line 39: Here, we are actually copying a simple object without any nesting.
so, `{...literals.price}` should be sufficient.

yesil (changes requested):

PR #4766, opened 2025-08-27
MWPW-178718: Fix geo and normal modal

zagi25:
Issue: If the modal is open via deep link (page is loaded with modal hash), and if georouting is enabled on the page, geo-popup will show up on top of the modal.
Screenshot 2025-08-27 at 15 26 41

* Hide the geo-popup while the modal is open
* Hide the geo-popup curtain if the modal doesn't have `curtain-off` class
* Focus the geo-popup when the modal is closed
* Focus the cookie banner when the geo-popup is closed
* Focus the cookie banner when the modal is closed and there is no geo-popup

Resolves: [MWPW-178718](https://jira.corp.adobe.com/browse/MWPW-178718)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/ratko/mwpw-178718-geo-popup-modal/geo-modal?martech=off
- After: https://mwpw-178718-geo-plus-modal--milo--adobecom.aem.page/drafts/ratko/mwpw-178718-geo-popup-modal/geo-modal?martech=off

**NOTE:** Test page has two modals `#modal` and `#curtain-off`. 
Testing instructions: 
- Connect to the VPN
- Open the test page with one of the modal hashes in incognito window. 
- Georouting can be disabled with `georouting=off`


mokimo (approved):
Didn't test this, but from a code perspective, LGTM

libs/blocks/modal/modal.js
```diff
@@ -49,6 +49,18 @@ export function sendAnalytics(event) {
   }
 }
 
+function focusAfterModalClose(modal) {
+  const { id } = modal;
+  const isGeoPopup = id === 'locale-modal-v2';
```
narcis-radu, line 54: We should use a constant for `locale-modal-v2`.
```suggestion
  const LOCALE_MODAL_ID = 'locale-modal-v2';
  ...
  const isGeoPopup = id === LOCALE_MODAL_ID;
```

narcis-radu, line 54: Not sure we need `id` here, it's only used once. Code can be rewritten as `modal?.id === LOCALE_MODAL_ID`

libs/blocks/modal/modal.js
```diff
@@ -49,6 +49,18 @@ export function sendAnalytics(event) {
...
+  const onetrustBanner = (isDeepLink || isGeoPopup) && document.querySelector('#onetrust-banner-sdk');
+  const geoPopupFocus = !isGeoPopup && document.querySelector('.dialog-modal#locale-modal-v2')?.querySelector('a.con-button');
```
narcis-radu, line 56:
```suggestion
  const geoPopupFocus = !isGeoPopup && document.querySelector('.dialog-modal#locale-modal-v2 a.con-button');
```

libs/blocks/modal/modal.js
```diff
@@ -49,6 +49,18 @@ export function sendAnalytics(event) {
...
+  const toFocus = geoPopupFocus || onetrustBanner || null;
```
narcis-radu, line 57: not sure why we'd need `null`

libs/blocks/modal/modal.js
```diff
@@ -158,7 +166,7 @@ function addIframeKeydownListener(iframe, dialog) {
 export async function getModal(details, custom) {
   if (!((details?.path && details?.id) || custom)) return null;
   const { id, deepLink } = details || custom;
-  isDeepLink = deepLink;
+  if (id !== 'locale-modal-v2') isDeepLink = deepLink;
```
narcis-radu, line 169: we should use `const LOCALE_MODAL_ID`

robert-bogos (approved):
Code looks good. Just make sure to resolve the conflicts

mokimo (commented):
Have not tested the solution, but from a code perspective this looks good to me.

mokimo (approved):

hit11757 (approved):
Validated. Testing details https://jira.corp.adobe.com/browse/MWPW-178718

PR #4782, opened 2025-08-29
MWPW-178561 [Sev2][Catalog] Filters - State: Selected state of the el…

bozojovicic:
…ement is missing or incorrect

Test page https://main--cc--adobecom.aem.live/products/catalog?milolibs=mwpw178561sidenav--milo--bozojovicic&martech=off&georouting=off

Resolves: [MWPW-178561](https://jira.corp.adobe.com/browse/MWPW-178561)

**Test URLs:**
- After: https://mwpw178561sidenav--milo--bozojovicic.aem.live/?martech=off


Roycethan:
@bozojovicic  Screen reader still not read the filter selected as "Selected..."
Another thing Axe noticed is this:
To solve this problem, you need to fix the following:
ARIA attribute is not allowed: aria-selected="true"
I think its not supported since role="listitem" for the element
probably role="tab" should support aria-selected="true"  or ria-selected="true" should be removed. plz expore the options

cc @3ch023

PR #4783, opened 2025-08-29
[MWPW-174755] Update carousel nav colors for a11y

overmyheadandbody:
This adapts the colors of the Carousel's navigation arrows for Accessibility purposes. There is an outstanding question around hover state colors when a dark theme is used; currently, a `#000` value is used for the arrow background, while the new Figma specs suggest no change is needed. I've continued using the previous value, a new PR will be needed, should the intention actually be to not have any.

Resolves: [MWPW-174755](https://jira.corp.adobe.com/browse/MWPW-174755)

**Test URLs:**
- Before: https://main--milo--overmyheadandbody.aem.page/drafts/ramuntea/carousel-arrow-a11y?martech=off
- After: https://carousel-arrow-a11y--milo--overmyheadandbody.aem.page/drafts/ramuntea/carousel-arrow-a11y?martech=off

### DIFF
--- a/libs/blocks/carousel/carousel.css
+++ b/libs/blocks/carousel/carousel.css
@@ -1,26 +1,22 @@
 .carousel {
-  /* Scope variables to carousel */
   --carousel-slide-button-height: 40px;
   --carousel-slide-button-width: var(--carousel-slide-button-height);
   --carousel-slide-button-height-desktop: 40px;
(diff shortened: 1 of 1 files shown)


mokimo (approved):

hit11757 (approved):
Validated. Testing details https://jira.corp.adobe.com/browse/MWPW-174755

PR #4794, opened 2025-09-02
MWPW-179291: Lang First POC

shkhan91:
* Add GEO IP-based country detection for langFirst localization in CaaS API
* Implement automatic country fallback when country is not specified in URL path
* Enhance getCountryAndLang function to use Akamai GEO IP service for better localization
* Add graceful error handling and fallback to US when GEO IP lookup fails

Resolves: [MWPW-179291](https://jira.corp.adobe.com/browse/MWPW-179291)

**Test URLs:**
- After: https:/langfirst-poc--milo--adobecom.aem.page/?martech=off

**Testing:**
- Test langFirst URLs without country (e.g., `/es/products`) to verify GEO IP country detection
- Test langFirst URLs with country (e.g., `/es/mx/products`) to ensure existing functionality works
- Test with different IP locations to verify country detection accuracy
- Test error scenarios when GEO IP service is unavailable

### DIFF
--- a/libs/blocks/caas/utils.js
+++ b/libs/blocks/caas/utils.js
@@ -533,7 +533,6 @@ export async function getCountryAndLang({ autoCountryLang, country, language }) 
           countryStr = geoCountry.toUpperCase();
         }
       } catch (error) {
-        // eslint-disable-next-line no-console
         countryStr = 'xx'; // fallback
       }
     }


libs/blocks/caas/utils.js
```diff
@@ -523,6 +523,19 @@ export function getCountryAndLang({ autoCountryLang, country, language }) {
     const pathArr = pageConfigHelper()?.pathname?.split('/') || [];
     const langStr = LANGS[pathArr[1]] ?? LANGS[''] ?? 'en';
     let countryStr = LOCALES[pathArr[2]] ?? 'xx';
+
+    // If no country found in URL path, try GEO IP lookup
+    if (countryStr === 'xx') {
```
robert-bogos, line 528: Is the 'xx' approach necessary? 🤔
```suggestion
    let countryStr = LOCALES[pathArr[2]];

    // If no country found in URL path, try GEO IP lookup
    if (countryStr) {
```

libs/blocks/caas/utils.js
```diff
@@ -523,6 +523,19 @@ export function getCountryAndLang({ autoCountryLang, country, language }) {
...
+      try {
+        const { getAkamaiCode } = await import('../../features/georoutingv2/georoutingv2.js');
+        const geoCountry = await getAkamaiCode(true);
+        if (geoCountry) {
+          countryStr = geoCountry.toUpperCase();
+        }
```
robert-bogos, line 534: nit 😊
```suggestion
        if (geoCountry) countryStr = geoCountry.toUpperCase();
```

libs/blocks/caas/utils.js
```diff
@@ -523,6 +523,19 @@ export function getCountryAndLang({ autoCountryLang, country, language }) {
...
+      } catch (error) {
+        countryStr = 'xx'; // fallback
```
robert-bogos, line 536: if we give up on the 'xx' approach, this would also need changing

robert-bogos (commented):
Also, please have a look at the unit tests. They seem to fail

PR #4805, opened 2025-09-04
Adding dnt support for language picker

bandana147:
* Previously, every URL with a hash value was assumed to be a modal. Now, instead, we are excluding the _dnt hash value to support DNT.

Resolves: [MWPW-179744](https://jira.corp.adobe.com/browse/MWPW-179744)

**Test URLs:**
- After: https://regionpicker-dnt--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/libs/blocks/global-footer/global-footer.js
+++ b/libs/blocks/global-footer/global-footer.js
@@ -347,7 +347,7 @@ class Footer {
 
     // Note: the region picker currently works only with Milo modals/fragments;
     // in the future we'll need to update this for non-Milo consumers
-    if (url.hash !== '') {
+    if (url.hash !== '' && url.hash !== '#_dnt') {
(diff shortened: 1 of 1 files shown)


libs/blocks/global-footer/global-footer.js
```diff
@@ -347,7 +347,7 @@ class Footer {
 
     // Note: the region picker currently works only with Milo modals/fragments;
     // in the future we'll need to update this for non-Milo consumers
-    if (url.hash !== '') {
+    if (url.hash !== '' && url.hash !== '#_dnt') {
```
sharmrj, line 350: This probably ought to be `includes` since (I think) it's possible for a link to have multiple `#_{something}` values.

nishantka (approved):

PR #4807, opened 2025-09-05
Refactor icon library to adhere to the DA standard

mokimo:
### Description
DA uses 'key, icon' while milo uses 'name, url' .. let's switch to what DA does so that we can re-use the federal icon sheet. For backwards compatibility, I added both versions to the sheet, so this PR shouldn't break anything and be forward-compatible.

Resolves: [MWPW-180007](https://jira.corp.adobe.com/browse/MWPW-180007)

### Visuals
Tested this on DA and Sharepoint using the new format using a local overwrite for the json // icon code.
image
Screenshot 2025-09-05 at 10 17 01

### Actions needed after merge
- Remove name/url/dnt sheet from the federal sheet
- add the right federal sheet to the DA project configs.

**Test URLs:**
- After: https://refactor-icon-library--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/libs/blocks/library-config/lists/icons.js
+++ b/libs/blocks/library-config/lists/icons.js
@@ -9,24 +9,24 @@ export default async function iconList(content, list, query) {
   if (!fedIconList) {
     fedIconList = await fetchIconList(content[0].path);
     if (!fedIconList?.length) throw new Error('No icons returned from fetchIconList');
-    fedIconList.forEach((icon) => {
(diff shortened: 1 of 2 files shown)


mokimo:
> Is the plan to then remove the redundant fields from the icons sheet?

@overmyheadandbody Yes, it's in the description:
> **Actions needed after merge**
Remove name/url/dnt sheet from the federal sheet
add the right federal sheet to the DA project configs.

I'll remove the old fields as a follow up once this is merged, they need to stay in the transition avoid breaking anything both before and after this is merged.

overmyheadandbody:
Thanks @mokimo! I looked at the story to see if it's mentioned, so QEs can also validate this, and didn't see a mention of it. But I admit I didn't scroll down so much in the PR description to see that, my bad 🙈

PR #4827, opened 2025-09-09
MWPW-175209: Get placeholders from other consumer sites.

seanchoi-dev:
- New getPlaceholderRoot function in libs/features/placeholders.js that provides flexible placeholder root URL resolution
- Comprehensive unit test coverage with 6 test cases covering all scenarios

Resolves: [MWPW-175209](https://jira.corp.adobe.com/browse/MWPW-175209)

**Test URLs:**
- After: https://mwpw-175209--milo--adobecom.aem.page/?martech=off

**Other Test URLs:**
- Before: https://main--homepage--adobecom.aem.page/kr/homepage/drafts/seanchoi/index-loggedout?martech=off
- After: https://main--homepage--adobecom.aem.page/kr/homepage/drafts/seanchoi/index-loggedout?martech=off&milolibs=mwpw-175209

### DIFF
--- a/libs/features/placeholders.js
+++ b/libs/features/placeholders.js
@@ -3,8 +3,24 @@ import { customFetch, getConfig, getMetadata } from '../utils/utils.js';
 const fetchedPlaceholders = {};
 window.mph = {};
 
+export const getPlaceholderRoot = (config) => {
+  const placeholderConstentRoot = getMetadata('placeholders-content-root');
+  if (!placeholderConstentRoot) return config.locale?.contentRoot;
+
+  let origin = 'https://www.adobe.com';
+  const placeholderOrigin = getMetadata('placeholders-origin');
+  if (config.env.name !== 'prod' && placeholderOrigin) {
+    const originParts = window.location.origin.split('--');
+    if (originParts.length === 3) {
+      origin = `${originParts[0]}--${placeholderOrigin}--${originParts[2]}`;
+    }
+  }
+  return `${origin}${config.locale?.prefix || ''}/${placeholderConstentRoot}`;
+};
+
 const getPlaceholdersPath = (config, sheet) => {
-  const path = `${config.locale.contentRoot}/placeholders.json`;
+  const placeholderRoot = getPlaceholderRoot(config);
+  const path = `${placeholderRoot}/placeholders.json`;
   const query = sheet !== 'default' && typeof sheet === 'string' && sheet.length ? `?sheet=${sheet}` : '';
   return `${path}${query}`;
 };
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -1153,11 +1153,12 @@ const findReplaceableNodes = (area) => {
   return nodes;
 };
 
-function getPlaceholderPaths(config) {
-  const root = `${config.locale?.contentRoot}/placeholders`;
-  const paths = [`${root}.json`];
-  if (config.env.name !== 'prod'
-    && getMetadata('placeholders-stage') === 'on') paths.push(`${root}-stage.json`);
+function getPlaceholderPaths(config, placeholderRoot) {
+  const paths = [`${placeholderRoot}/placeholders.json`];
+  // Add stage-specific placeholder if placeholders-stage is on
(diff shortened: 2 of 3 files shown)


libs/features/placeholders.js
```diff
@@ -3,8 +3,24 @@ import { customFetch, getConfig, getMetadata } from '../utils/utils.js';
...
+
+  let origin = 'https://www.adobe.com';
+  const placeholderOrigin = getMetadata('placeholders-origin');
+  if (config.env.name !== 'prod' && placeholderOrigin) {
+    const originParts = window.location.origin.split('--');
+    if (originParts.length === 3) {
+      origin = `${originParts[0]}--${placeholderOrigin}--${originParts[2]}`;
+    }
+  }
+  return `${origin}${config.locale?.prefix || ''}/${placeholderConstentRoot}`;
```
3ch023, line 18: typo: constent instead of content?

libs/features/placeholders.js
```diff
@@ -3,8 +3,24 @@ import { customFetch, getConfig, getMetadata } from '../utils/utils.js';
 const fetchedPlaceholders = {};
 window.mph = {};
 
+export const getPlaceholderRoot = (config) => {
+  const placeholderConstentRoot = getMetadata('placeholders-content-root');
+  if (!placeholderConstentRoot) return config.locale?.contentRoot;
+
+  let origin = 'https://www.adobe.com';
```
3ch023, line 10:
```suggestion
  let origin = window.location.origin;
```

this will ensure placeholders are pulled from 'https://www.stage.adobe.com' when on stage

seanchoi-dev: There is placeholder-stage in case stage is on.
Using prod URL as default is more making sense since the file isn't affected by code.

libs/utils/utils.js
```diff
@@ -1167,11 +1168,11 @@ export async function decoratePlaceholders(area, config) {
   const nodes = findReplaceableNodes(area);
   if (!nodes.length) return;
   area.dataset.hasPlaceholders = 'true';
-  const phPaths = getPlaceholderPaths(config);
+  const { decoratePlaceholderArea, getPlaceholderRoot } = await import('../features/placeholders.js');
```
chrischrischris, line 1171: If there are replaceable nodes then we have to load placeholders.js.  I think we should move `getPlaceholderPaths` and the rest of this `decoratePlaceholders` function into placeholders.js in order to reduce code in utils.js.

libs/features/placeholders.js
chrischrischris: If `placeholders-repo` is not defined, but `placeholders-content-root` is, then there will be unexpected behavior that may confuse authors.  Both are required for this to work, so what do you think about defining both values in one metadata field?  Something like:
`placeholders-override: reponame,placeholder-content-root`
`placeholders-override : cc,cc-shared`

Also in the future there may be other consumer projects on non-www subdomains that want to use this, but for now I think we can leave as www.adobe.com.

chrischrischris (changes requested):

mokimo:
Why are you not using a federal placeholder sheet instead? The concept exists specifically to re-use placeholders.
I feel like homepage consuming from CC is an anti-pattern & instead you should move to a sheet in federal and consume that for CC and Homepage

libs/utils/utils.js
seanchoi-dev: I agree, but I think that's a little out of scope for this ticket.
Migrating decoratePlaceholders will require some investigation and thorough testing for various use cases.

mokimo:
@3ch023 & @seanchoi-dev is everything clear now? Based on the ticket, here's what needs to happen next

**GWP** - move cc placeholders & publish the new-cc/hp placeholders within federal
**Engineering:**
- This PR needs to add consumer config option to use placeholders from within the federal repository, set within the consumer scripts (see https://github.com/adobecom/milo/pull/4827#discussion_r2378534356)
- PR on consumer to set `config.locale?.federalPlaceholderPath:` to the new location for both CC/Homepage within their consumer scripts

After the new placeholders are consumed, the old ones in CC/HP can be deleted

PR #4853, opened 2025-09-15
[MWPW-175987] - Mweb section2 lcp fix

DKos95:
This PR aims to give authors the ability to improve LCP performance on mobile devices where the first section doesn't have an image that is considered to be an LCP candidate while the second section does. By applying the class `lcp-deferred` to the first section that doesn't have and LCP image we know not to treat it as the LCP section but the next section after it. The more ideal situation would be for us to automate this but unfortunately this is not possible considering that at the moment where we need to choose which section is LCP there is no structure on the page since sections haven't been rendered yet, and we can't write a query to check if the first section has a background or asset image without extracting render logic for each Marquee block which would add a considerable level of complexity and dependency in the utils file which we don't want.

**Performance analysis page**
https://main--milo--adobecom.aem.page/drafts/dusan/lcp-section2-analysis

Resolves: [MWPW-175987](https://jira.corp.adobe.com/browse/MWPW-175987)

**PSI Test URLs:**
- After: https://mweb-section2-lcp-fix--milo--adobecom.aem.page/?martech=off

**CC Test URLs:**
- Before: https://main--cc--adobecom.aem.page/drafts/dusan/lcp-section2-normal-text
- After: https://main--cc--adobecom.aem.page/drafts/dusan/lcp-section2-normal-text?milolibs=mweb-section2-lcp-fix

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -1777,21 +1777,21 @@ export async function loadArea(area = document) {
 
   const areaBlocks = [];
   let lcpSectionId = null;
-  let isMarqueeAndNoMedia = null;
+  let isLcpDeferred = null;
 
   for (const section of sections) {
-    if (!section.idx && window.matchMedia('(max-width: 768px)').matches) {
-      isMarqueeAndNoMedia = section.el.querySelector('.hero-marquee.no-media, .marquee.no-media, .quiz-marquee.no-media');
+    if (!section.idx && window.matchMedia('(max-width: 600px)').matches) {
+      isLcpDeferred = section.el.querySelector('.lcp-deferred');
     }
 
-    if (isMarqueeAndNoMedia && section.idx === 1) {
-      section.el.querySelectorAll('img')?.forEach((img) => img.setAttribute('loading', 'eager'));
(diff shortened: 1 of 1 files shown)


mokimo (changes requested):
This solution was rejected in the past... this is a content issue and (back then) it was deemed not the right path to introduce more complexity to cater around content that specifically sways away from performance recommendations.

The LCP algorithm is _currently_ based on a set of certain rules, but this can change at any point if chrome decides so, or AI traffic can change the game. If you now force authors to re-author 1000s of pages, this is just not maintainable.

If that authors decide that section one is supposed to get skipped - but some re-authoring happens, this will get forgotten, and you now have authors negatively impact performance. They'd also all need to be trained specifically on how to achieve better performance gains.

As this was not acceptable 1-2 years ago due to many reasons, as in, putting non-standard specific metadata attributes into the authors hands to steer performance and loading patterns on a page, I don't see why that would be acceptable now.

**My proposal would be sticking to the content based solutions** that you've outlined in your due diligence report and closing this as 'won't fix'

PR #4879, opened 2025-09-18
[MWPW-180618] - action item alignement fix

DKos95:
This fixes alignment issues for the action item. On some pages like https://www.adobe.com/nl/nonprofits.html the class `center` is missing and would need to be applied for this fix to work as it is applied on the test pages supplied in the ticket and here bellow in the Test URLs section.

Resolves: [MWPW-180618](https://jira.corp.adobe.com/browse/MWPW-180618)

**PSI Test URLs:**
- After: https://action-item-align-fix--milo--adobecom.aem.page/?martech=off

**Test URLs:**
- Before: https://main--cc--adobecom.aem.live/drafts/klee/test-action-items
- After: https://main--cc--adobecom.aem.live/drafts/klee/test-action-items?milolibs=action-item-misalignment

### DIFF
--- a/libs/blocks/action-item/action-item.css
+++ b/libs/blocks/action-item/action-item.css
@@ -2,6 +2,10 @@
   display: flex;
 }
 
+.action-item .static {
+  width: -webkit-fill-available;
+}
+
 .action-item img {
   display: block;
   max-height: 240px;


libs/blocks/action-item/action-item.css
```diff
@@ -2,6 +2,10 @@
   display: flex;
 }
 
+.action-item .static {
+  width: -webkit-fill-available;
```
overmyheadandbody, line 6: On [this page](https://www.adobe.com/uk/nonprofits.html) for example, the issue can be reproduced on Firefox as well, where `-webkit-` would likely not have an effect. It might be that something like
```
.action-item.center {
  justify-content: center;
}
```
solves the issue. Alternatively, setting the width to `100%` should be applicable everywhere.

DKos95: Thanks for pointing that out, should be ok on FF now also.

PR #4920, opened 2025-09-25
[MWPW-180345] - Carousel dot aria labels

DKos95:
- Styling update in accordance with figma design
- Jump to functionality removed - https://jira.corp.adobe.com/browse/MWPW-180345#comment-50141255
- Indicator dots can't be tabbed to or clicked on
- After consulting with a11y team it has been decided that aria live announcement need an update so that when indicator dots are visible, after text content of slide has been announced the screenreader will also announce 'Slide x of y'

Resolves: [MWPW-180345](https://jira.corp.adobe.com/browse/MWPW-180345)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/docs/library/kitchen-sink/carousel
- After: https://carousel-dot-aria-labels--milo--adobecom.aem.page/docs/library/kitchen-sink/carousel


libs/blocks/carousel/carousel.js
```diff
@@ -361,7 +358,7 @@ function moveSlides(event, carouselElements, jumpToIndex) {
   }
 
   activeSlideIndicator.classList.add('active');
-  if (jumpTo) activeSlideIndicator.setAttribute('tabindex', 0);
+  activeSlideIndicator.setAttribute('aria-current', 'location');
```
zagi25, line 361: Just to confirm, last time I worked on this, I got the information from Taryn that indicators should have `tabindex=0` and `aria-label`, reason being that indicators are visible and user should be able to focus them with tab. Is this not the case anymore ?

libs/blocks/carousel/carousel.js
```diff
@@ -26,12 +26,27 @@ const KEY_CODES = {
...
+
+function decorateNextPreviousBtns(slides, currentIndex = 0) {
+  const totalSlides = slides ? slides.length : 0;
+
   const previousBtn = createTag(
     'button',
     {
       class: 'carousel-button carousel-previous is-delayed',
-      'aria-label': 'Previous slide',
+      'aria-label': getPreviousAriaLabel(currentIndex, totalSlides),
```
zagi25, line 49: Should there be a `getNextAriaLabel` also for next button ?

libs/blocks/carousel/carousel.js
```diff
@@ -26,12 +26,27 @@ const KEY_CODES = {
 };
 const FOCUSABLE_SELECTOR = 'a, :not(.video-container, .pause-play-wrapper) > video';
 
-function decorateNextPreviousBtns() {
+function getPreviousAriaLabel(currentIndex, totalSlides) {
+  return currentIndex === 0 && totalSlides > 0
+    ? `Previous slide, slide ${currentIndex + 1} of ${totalSlides}`
```
zagi25, line 31: Why this is only applied when index === 0 ?
Also, should it say 'Previous slide, slide 1 of 3' if we are on the first slide ? Should it not indicate previous slide (last index) ?

libs/blocks/carousel/carousel.js
DKos95: No because the index gets announced when the user actually clicks it, the issue was for the first initial slide when the user tabs to it he doesn't know on which slide he is on, so when he tabs on the left arrow he will be told that he is on the first slide if he is on the first slide, in any other case if he uses either left or right arrow the new index gets announced, this was of course all checked with Taryn.

libs/blocks/carousel/carousel.js
DKos95: The issue was that when the user tabs to this initially he doesn't know on which slide he is on currently, so this is announced only if he is on the first slide, if he uses the left or right arrow the new index will be announced, so the issue was only with initially letting the user know on which slide he is on, this was checked with Taryn and she said it is compliable.

libs/blocks/carousel/carousel.js
```diff
@@ -235,17 +235,36 @@ function setIndicatorMultiplyer(carouselElements, activeSlideIndicator, event) {
...
+function updateAriaLive(ariaLive, slide, carouselElements) {
   let text = '';
   slide.querySelectorAll(':scope > :not(.section-metadata').forEach((el, index) => {
     text += `${index ? ' ' : ''}${el.textContent.trim()}`;
   });
+
+  const { slides, el: block } = carouselElements;
+
+  if (![...block.classList].find((cls) => cls.startsWith('show-'))) {
+    const currentSlideIndex = slides.findIndex((s) => s === slide);
```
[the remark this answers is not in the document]
DKos95: I agree for the `slide.dataset.index` part I didn't notice that, thanks!
Regarding `carouselElements` argument, it is being passed also to be able to take out block and check the classes on the block.

libs/blocks/carousel/carousel.js
```diff
@@ -361,7 +347,7 @@ function moveSlides(event, carouselElements, jumpToIndex) {
   }
 
   activeSlideIndicator.classList.add('active');
```
narcis-radu, line 349: I noticed a small issue, where the dots are not indicating the active slide, not sure if it's related to this particular code change. Here's how to reproduce:
1. Enable mobile view
2. Go to URL and scroll to the last carousel
3. Press the right arrow 6 times to scroll

Desired behaviour: first dot is highlighted (carousel starting from first slide)
Actual behaviour: no dots are highlighted

overmyheadandbody: Good catch! This might have been added at the beginning of the year - https://github.com/adobecom/milo/blob/4515f0c581fe7a9acd613449af04df9c6490dcb2/libs/blocks/carousel/carousel.css#L216

DKos95: I will be changing all instances where 'multiplyer' is used to 'multiplier' since only the latter is correct spelling.

zagi25 (approved):

PR #4923, opened 2025-09-25
MWPW-179866 Brand Concierge - Responsive Input

meganthecoder:
* Update Brand Concierge input to be responsive (grow vertically)

Resolves: [MWPW-179866](https://jira.corp.adobe.com/browse/MWPW-179866)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/methomas/brand-concierge?martech=off
- After: https://bc-responsive--milo--adobecom.aem.page/drafts/methomas/brand-concierge?martech=off


libs/blocks/brand-concierge/brand-concierge.js
```diff
@@ -37,6 +37,13 @@ function updateModalHeight() {
   modal.style.height = `${modalHeight}px`;
 }
 
+function updateInputHeight(el) {
+  const fieldInput = el.querySelector('.bc-input-field textarea');
+  if (!fieldInput) return;
+  fieldInput.style.height = 'auto';
+  fieldInput.style.height = `${fieldInput.scrollHeight}px`;
+}
```
overmyheadandbody, line 45: Maybe an unpopular opinion, but instead of constantly running some resize logic, why not let the browser handle it via a `contenteditable`? Something like [this](https://jsbin.com/rabosasaju/edit?html,css,output).

libs/blocks/brand-concierge/brand-concierge.css
```diff
@@ -209,10 +209,18 @@
   line-height: 1.5;
   outline: none;
   width: 100%;
+  resize: none;
+  overflow-wrap: break-word;
+  word-wrap: break-word;
+  white-space: pre-wrap;
```
overmyheadandbody, line 215: Is the `white-space` declaration coming from a requirement? I'd assume it'd be cleaner without it in certain scenarios. Trying a Cmd + Return already submits the result, so I'm guessing this might be here for when someone pastes something in?

libs/blocks/brand-concierge/brand-concierge.css
```diff
@@ -209,10 +209,18 @@
   line-height: 1.5;
   outline: none;
   width: 100%;
+  resize: none;
+  overflow-wrap: break-word;
+  word-wrap: break-word;
```
overmyheadandbody, line 214: Aren't these sort of "synonyms"? One or the other should do the trick, did you notice situations where both have an effect?

JackySun9 (approved):

PR #4926, opened 2025-09-29
MWPW-174852 [M@S] Enable default merch settings on MAS driven pages

bozojovicic:
Enable `mas-ff-defaults` in MAS cards and collections.

Resolves: [MWPW-174852](https://jira.corp.adobe.com/browse/MWPW-174852)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.live/fr/drafts/bozo/cards?martech=off
- After: https://mwpw174852defaults2--milo--bozojovicic.aem.live/fr/drafts/bozo/cards?martech=off


yesil (approved):

Roycethan:
@bozojovicic Can you plz check why these studio tests are failing ?
strikethrough pricing is not rendered on the card although its authored....looks like the price class is missing 'price-strikethrough' and 'price-alternative'
https://main--mas--adobecom.aem.live/studio.html?milolibs=mwpw174852defaults2--milo--bozojovicic#page=content&path=nala&query=3b1fb0f1-b74e-4e8f-81ad-1744012b1935

Roycethan:
> https://main--mas--adobecom.aem.live/studio.html?milolibs=mwpw174852defaults2--milo--bozojovicic#page=content&path=nala&query=3b1fb0f1-b74e-4e8f-81ad-1744012b1935

So @bozojovicic  What if this case of one offer with displayOldPrice='false' case exist on production...doesn't it breaks (no strikethrough rendered) when we merge this PR ?

Need confirmation and some analysis on this usecase , before i give green light/merge...

cc @3ch023

bozojovicic:
@Roycethan this change might affect only pages where we use some MAS cards or collections. Currently that's only us/en plans page. If we have prices with promo code where the parameter "old" is not set then we should set it "&old=true" to the merch link of that price before this PR is merged. Can you check or maybe we should ask Fred to check?
If everything is OK with this PR maybe we could mark it as verified and then "ready for stage" once Fred checks the prices.
CC @3ch023

PR #4929, opened 2025-09-29
MWPW-180338 [Plans] Issues with deep-linked tabs and modals

bozojovicic:
On plans page we can have multiple modals with the same modal ID but on different tabs. 
When we open that modal, go to commerce page and then go back, the modal from the active tab needs to be opened.

Resolves: [MWPW-180338](https://jira.corp.adobe.com/browse/MWPW-180338)

**Test URLs:**
- After: https://mwpw180338modal--milo--bozojovicic.aem.live/?martech=off


mirafedas:
@bozojovicic When I open the modal in the STE tab, click Continue, then go back, close the modal, go to the first tab and open a Firefly modal, click Continue and go back - it should redirect me to the first tab and open the Firefly modal, instead it redirects to the STE tab and doesn't open a modal. I sent you the screencast is DM (too big to attach it here), please check.
Also please make sure to test together with a delayed modal and a georouting modal.

PR #4935, opened 2025-09-30
MWPW-181306 | Exporting internal functions from preflight for content insights

aishwaryamathuria:
The PR adds support for exporting some internal functions from preflight tool for milo studio content insights.
Following is a small summary of the changes:
- **Preflight core function:** Minor changes were made to export the existing core function for better reusability.
- **assets.js:** Added an export to open all modals. This does not affect current preflight functionality but will be used in Studio code.
- **seo.js:** Expanded the link filter to exclude a few more URLs. This won’t affect existing preflight behaviour but ensures both preflight and Studio can share the same list when needed.
- **contentInsights.js:** Since inline code execution is restricted on some Adobe domains (e.g., [adobe.com](http://adobe.com/), Acrobat sites) and direct imports aren’t allowed in Adobe I/O deployments, we created a new file for content insights imports. This file is only loaded in utils after LCP if a specific URL parameter is set (done only in content insights automation). It does not impact the existing site.

Resolves: [MWPW-181306](https://jira.corp.adobe.com/browse/MWPW-181306)

**Test URLs:**


libs/blocks/preflight/checks/assets.js
```diff
@@ -198,6 +199,29 @@ export function isViewportTooSmall() {
   return !window.matchMedia('(min-width: 1200px)').matches;
 }
 
+export async function openAllModals(area = document) {
+  const links = await area.querySelectorAll('main a[data-modal-hash]');
```
zagi25, line 203: Why is this awaited ? `querySelectorAll` is not an async function.

libs/blocks/preflight/checks/assets.js
```diff
@@ -198,6 +199,29 @@ export function isViewportTooSmall() {
...
+  // eslint-disable-next-line no-unused-vars
+  for (const [index, link] of links.entries()) {
+    try {
+      // Scroll into view if necessary
+      link.scrollIntoView({ behavior: 'instant', block: 'center' });
+      link.getAttribute('href');
```
zagi25, line 209: what is the point of getting `href` if it is not used ?

libs/blocks/preflight/checks/seo.js
```diff
@@ -246,6 +246,30 @@ function compareResults(result, link) {
...
+      && !link.href.startsWith('#') // Is not a local link
+      && !link.href.startsWith('https://#') // Is not a local link
+      && !link.href.includes('local') // Is not a local link
+      && !link.href.includes('bookmark://') // Ignore bookmarks
+      && !link.closest('.preflight') // Is not inside preflight
+      && !knownBadUrls.some((url) => url === link.hostname) // Is not a known bad url
+    ) {
+      return true;
+    }
+    return false;
```
zagi25, line 269: You should be able to just return this, no need for `if`
```suggestion
      return link.href // Has an href tag
      && !link.href.includes('tel:')  // Is not a phone number
      && !link.href.includes('mailto:')  // Is not an email address
      && !link.href.startsWith('#') // Is not a local link
      && !link.href.startsWith('https://#') // Is not a local link
      && !link.href.includes('local') // Is not a local link
      && !link.href.includes('bookmark://') // Ignore bookmarks
      && !link.closest('.preflight') // Is not inside preflight
      && !knownBadUrls.some((url) => url === link.hostname) // Is not a known bad url
```

zagi25 (commented):
Can you please resolve all eslint errors, it is hard to review it like this. Thank you.

libs/blocks/preflight/checks/assets.js
```diff
@@ -198,6 +199,29 @@ export function isViewportTooSmall() {
   return !window.matchMedia('(min-width: 1200px)').matches;
 }
 
+export async function openAllModals(area = document) {
```
robert-bogos, line 202: If this is only needed for Studio, why not have this logic on the Studio side? 🤔

libs/blocks/preflight/panels/accessibility.js
```diff
@@ -106,6 +106,11 @@ async function checkAlt() {
   });
   result.description = 'All images listed below. Please validate each alt text has been set appropriately. Decorative images have been highlighted in yellow on the page.';
   altResult.value = { ...result, checked: true };
+  
+  return {
+    "decorativeImages": decorativeImages.value,
+    "altTextImages": altTextImages.value
```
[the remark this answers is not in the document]
aishwaryamathuria: I have made this change

aishwaryamathuria:
> Can you please resolve all eslint errors, it is hard to review it like this. Thank you.

The comments have been removed for readability.

libs/utils/utils.js
```diff
@@ -1633,6 +1631,14 @@ function initSidekick() {
   }
 }
 
+function initContentInsights() {
```
narcis-radu, line 1634: Are there any other options to init? It doesn't make sense to add the code into `utils` which is loaded by all consumers.

aishwaryamathuria: @narcis-radu Content insights code is executed on prod instance and adobe.com sites (like adobe.com/acrobat) do not allow inline script execution hence this script can only be executed to get the preflight content from the site if it is already loaded with it. Hence we had to place it in utils (to be loaded in similar way as preflight)
The additional code and content insights instances will load only if the query param is set and will not impact page performance or LCP.

libs/blocks/preflight/checks/seo.js
```diff
@@ -246,33 +246,20 @@ function compareResults(result, link) {
...
-
   const links = [...area.querySelectorAll('a')]
     .filter((link) => {
       if (
         link.href
+        && !link.href.includes('tel:')
+        && !link.href.includes('mailto:')
+        && !link.href.startsWith('#')
+        && !link.href.startsWith('https://#')
+        && !link.href.includes('bookmark://')
```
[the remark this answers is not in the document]
raga-adbe-gh: Currently these are sent to Spidy and reported as errors. There are some pages these was found while running content insights
Examples - 
mailto in https://www.adobe.com/legal/terms.html
bookmark in https://www.adobe.com/pt/privacy/policy.html
cc: @aishwaryamathuria

robert-bogos: One more thing I would do here is to add a comment for the future us to remember why we did this. 
Also, when we move away from Spidy to something that works better for these use cases, we might miss faulty URLs because they get filtered out
```suggestion
// Added extra checks because Spidy misidentifies these URL schemes as faulty. Can be removed once we stop using Spidy.
        && !link.href.includes('tel:')
        && !link.href.includes('mailto:')
        && !link.href.startsWith('#')
        && !link.href.startsWith('https://#')
        && !link.href.includes('bookmark://')
        //
```

libs/blocks/preflight/checks/seo.js
```diff
@@ -246,6 +246,26 @@ function compareResults(result, link) {
...
+        && !link.href.includes('tel:')
+        && !link.href.includes('mailto:')
+        && !link.href.startsWith('#')
+        && !link.href.startsWith('https://#')
+        && !link.href.includes('local')
+        && !link.href.includes('bookmark://')
+        && !link.closest('.preflight')
+        && !knownBadUrls.some((url) => url === hostname);
+    }).map((link) => link.href);
+}
```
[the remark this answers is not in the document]
robert-bogos: Makes sense! Thank you for explaining 🙏🏻

libs/blocks/preflight/checks/seo.js
aishwaryamathuria: @robert-bogos I have made the recommended change

robert-bogos (approved):
Thank you for implementing the feedback 🙏🏻

mokimo (approved):
Gave it a birds-eye view and LGTM.

PR #4947, opened 2025-10-03
MWPW-172104 [Kodiak][R2]: DOMXSS in [github.com/adobecom/cc-sandbox]

bozojovicic:
There is a service in CC that returns HTML which needs to be injected into the page but first it needs to be sanitized (all potentially dangerous tags and attributes removed).

I had to do all this manually since https://developer.mozilla.org/en-US/docs/Web/API/HTML_Sanitizer_API is not yet fully available.

Test page https://mwpw172104xss--cc--bozojovicic.aem.live/uk/offers/promo-terms?milolibs=mwpw172104sanitize2--milo--bozojovicic&locale=fr_FR&service_providers=PROMO_TERMS&country=FR&api_key=PlatformDocs&offer_id=00AD23F3606079A216964E5F3DAB08E7&promo_context=initial_purchase&campaignEnd=2025/6/3#

Resolves: [MWPW-172104](https://jira.corp.adobe.com/browse/MWPW-172104)

**Test URLs:**
- After: https://mwpw172104sanitize2--milo--bozojovicic.aem.live/?martech=off

### DIFF
--- a/libs/utils/sanitizeHtml.js
+++ b/libs/utils/sanitizeHtml.js
@@ -1,7 +1,13 @@
 function stringToHTML(str) {
-  const parser = new DOMParser();
-  const doc = parser.parseFromString(str, 'text/html');
-  return doc.body || document.createElement('body');
+  const emptyBody = document.createElement('body');
+  if (!str?.trim()) return emptyBody;
+  try {
+    const parser = new DOMParser();
+    const doc = parser.parseFromString(str, 'text/html');
+    return doc.body || emptyBody;
+  } catch (err) {
+    return emptyBody;
+  }
 }
 
 function removeScripts(html) {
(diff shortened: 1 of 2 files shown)


libs/utils/sanitizeHtml.js
```diff
@@ -0,0 +1,37 @@
...
+  const val = value.replace(/\s+/g, '').toLowerCase();
+  if (['src', 'href', 'xlink:href'].includes(name)) {
+    // eslint-disable-next-line no-script-url
+    if (val.includes('javascript:') || val.includes('data:text/html')) return true;
+  }
+  if (name.startsWith('on')) return true;
+  return false;
+}
+
+function removeAttributes(elem) {
```
mirafedas, line 22: Nit: we can make it one line shorter if you prefer:
```
function removeAttributes({ attributes }) {
  [...attributes].forEach(({ name, value } ) => {
    if (isPossiblyDangerous(name, value)) {
      elem.removeAttribute(name);
    }
  });
}

File: libs/utils/sanitizeHtml.js [hunk 2]
```diff
@@ -0,0 +1,37 @@
+function stringToHTML(str) {
+  const parser = new DOMParser();
```
mirafedas, line 2: If `str` is not a TrustedHTML, the next line will throw an error. Let's add the try catch statement or checking if `str` is defined and is a proper data type

mirafedas (changes requested):

PR #4955, opened 2025-10-07
Brand Concierge Hero variant

colloyd:
* hero variant
* support for background images or gradients
* pill-card mod for the standard inline variant

Resolves: [MWPW-181227](https://jira.corp.adobe.com/browse/MWPW-181227)

**Test URLs:**
Hero Variant
- Before: https://main--milo--adobecom.aem.page/drafts/colloyd/brand-concierge/brand-concierge-hero?martech=off
- After: https://bc-hero--milo--colloyd.aem.page/drafts/colloyd/brand-concierge/brand-concierge-hero?martech=off

Inline Variant with pill-cards and input-first
- Before: https://main--milo--adobecom.aem.page/drafts/colloyd/brand-concierge/brand-concierge-404?martech=off
- After: https://bc-hero--milo--colloyd.aem.page/drafts/colloyd/brand-concierge/brand-concierge-404?martech=off


libs/blocks/brand-concierge/brand-concierge.js
```diff
@@ -88,7 +81,16 @@ async function openChatModal(initialMessage, el) {
 function decorateBackground(el, background) {
   const bgValue = background.textContent.trim();
   if (bgValue) {
+    el.classList.add('has-bg-color');
     el.style.setProperty('--brand-concierge-bg', bgValue);
+  } else {
+    const bgImage = background.querySelector('img');
+    if (bgImage) {
+      // remove query string to prevent blurry images.
+      const rawImage = bgImage.src.slice(0, bgImage.src.indexOf('?'));
```
meganthecoder, line 90: Isn't the query string added by AEM EDS to improve performance? It doesn't seem like that should be stripped out

libs/blocks/brand-concierge/brand-concierge.js
```diff
@@ -88,7 +81,16 @@ async function openChatModal(initialMessage, el) {
...
+      el.classList.add('has-bg-image');
```
meganthecoder, line 91: You should consider as an enhancement using `decorateBlockBg` and the background-by-breakpoint setup that consonant blocks have

meganthecoder (approved):
It's a little confusing that this PR is combined with the input one but I trust that you'll copy over any updates.

JackySun9 (approved):

libs/blocks/brand-concierge/brand-concierge.js
colloyd: The query string being present always left the image blurry. If this were added as a normal image, rather than a css background, I would just drop the whole picture element in. As this is being added as a css background (to keep things consistent with the use of background gradient as well as to match the figma more closely), I opted to add this fix to ensure that authors aren't left confused about blurry images. I did test this thoroughly for performance as I was concerned about the implications of doing this. I was consistently getting 98's with the image as it is, so it seems safe... as long as an author doesn't try to dump in a gigantic unoptimized image :) I'm open to making adjustments if this becomes a problem in production.

libs/blocks/brand-concierge/brand-concierge.js
colloyd: I'm open to looking into that as a possible enhancement down the road, especially if this setup ever becomes problematic.

PR #5616, opened 2026-03-19
[MWPW-190563][MEP Lingo] Fallback for regional EN sites fails when MEP‑Lingo row is used in marquee block

markpadbe:
- Fixes a bug where authored content in block/section swaps (e.g., hero-marquee) was replaced by the base fragment when the regional fragment wasn't found in the query-index
- Root cause: when the query-index finds no regional version, the link falls back to the base locale path. If that base fragment exists (which it typically does -- mep-lingo fragments are authored at the US root path, and localized copies often exist too), the fetch returns 200 OK. Because the response is successful, the existing 404 handler is bypassed and the base fragment content overwrites the authored block. This gap has existed since block/section swaps were first introduced in PR #5219.
- Adds a guard after FallbackDetected that detects when fallback (non-regional) content is being used for a block/section swap and preserves the authored content instead. Standalone mep-lingo blocks are not affected.
- **Update**: Redirect detection for production CDN
Adds a second fallback detection for block/section swaps when the production CDN (business.adobe.com) returns a 301 redirect instead of a 404 for non-existent regional fragments. The fetch API follows the redirect automatically, so resp.ok is true and the existing 404 fallback logic is bypassed — causing authored blocks to be replaced with the redirect's meaningless content.
Detects this via resp.redirected and treats it as a fallback scenario, preserving the authored block. This only affects regionalized block/section swaps and is a no-op if the CDN is later fixed to return proper 404s.
Refactored both fallback conditions (fetchedNonRegionalContent and redirectedRegionalFetch) into named variables for clarity.

Steps to QA: 
Open the Before and After links below and scroll to the bottom of the page where I've authored a second hero-maquee.
Expected(After): Hero Marquee with Title: "ES Version - Define el futuro del marketing impulsado por IA"
Actual(Before): Fallback Fragment showing marquee with title: "ES Marquee Fragment - doens't exist in LA"

Steps to QA for production: 
prod url: https://business.adobe.com/solutions/unified-customer-experience.html?akamaiLocale=ru
Override the fragment.js file. 
Expected: marquee appears on page
Actual: marquee does not not appear

Resolves: [MWPW-190563](https://jira.corp.adobe.com/browse/MWPW-190563)

**Test URLs:**
- Before: https://main--da-bacom--adobecom.aem.page/es/drafts/mepdev/fragments/base-insert-remove?mep=&mepHighlight=true&akamaiLocale=la
- After: https://main--da-bacom--adobecom.aem.page/es/drafts/mepdev/fragments/base-insert-remove?milolibs=mep-lingo-fix-us-block-swap&mep=&mepHighlight=true&akamaiLocale=la
- Psi-check: https://mep-lingo-fix-us-block-swap--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/libs/blocks/fragment/fragment.js
+++ b/libs/blocks/fragment/fragment.js
@@ -91,6 +91,12 @@ export const removeMepLingoRow = (container) => {
   mepLingoRow?.remove();
 };
 
+const preserveAuthored = (a, isBlockSwap, originalBlock, originalSection) => {
+  if (isBlockSwap) removeMepLingoRow(originalBlock);
+  else removeMepLingoRow(originalSection?.querySelector('.section-metadata'));
+  a.parentElement?.remove();
+};
+
 export default async function init(a) {
   const { decorateArea, mep, placeholders, locale } = getConfig();
   let relHref = await localizeLinkAsync(a.href, window.location.hostname, false, a);
@@ -223,13 +229,7 @@ export default async function init(a) {
 
   if (usedFallback && !isMepLingoBlock
     && ((isBlockSwap && originalBlock) || (isSectionSwap && originalSection))) {
-    if (isBlockSwap) {
-      removeMepLingoRow(originalBlock);
-      a.parentElement?.remove();
-    } else {
-      removeMepLingoRow(originalSection?.querySelector('.section-metadata'));
-      a.parentElement?.remove();
-    }
+    preserveAuthored(a, isBlockSwap, originalBlock, originalSection);
     return;
   }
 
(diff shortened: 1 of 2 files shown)


vhargrave (commented):
Feature wise it seems to be working pretty well. I'm a bit confused by the preview i'm seeing MEP though. 

the url it's showing me is not the URL i'm seeing in DA. 
Furthermore, should it not also be yellow if it fell back to the default content? 

Image

Image

vhargrave (approved):
makes sense - thanks for the explanation and here's an approval!

sukamat (approved):

JackySun9 (approved):

PR #5626, opened 2026-03-19
MWPW-189237: add ?caas-env=dev query param support to bulk publisher

sanrai:
## Summary
- Adds `?caas-env=dev` query param support to the bulk publisher
- When present, `postDataToCaaS` routes to the dev proxy endpoint (`14257-milocaasproxy-dev`) instead of prod
- Also sends `caas-env: dev` in the request header
- Allows E2E automation to target the dev environment natively without JS fetch interception

## Test plan
- Visit `bulkpublisher.html?caas-env=dev` and trigger a publish — verify the request goes to `14257-milocaasproxy-dev` with `caas-env: dev` header
- Visit without the param — verify prod endpoint and existing behaviour unchanged
- Visit with `?caas-env=stage` or any other value — verify prod endpoint is used (no URL override)

This has already been tested and is automated here: 
https://github.com/adobecom/caas/pull/427

### DIFF
--- a/tools/send-to-caas/send-utils.js
+++ b/tools/send-to-caas/send-utils.js
@@ -10,6 +10,7 @@ import {
 const CAAS_TAG_URL = 'https://www.adobe.com/chimera-api/tags';
 const HLX_ADMIN_STATUS = 'https://admin.hlx.page/status';
 const URL_POSTXDM = 'https://14257-milocaasproxy.adobeio-static.net/api/v1/web/milocaas/postXDM';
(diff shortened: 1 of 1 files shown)


cmiqueo (approved):

gmirijan (approved):

sheridansunier (approved):
Clean, minimal, well-scoped change. The dev endpoint routing is correctly gated behind an explicit opt-in param, existing behavior is fully preserved, and the E2E coverage in caas#427 covers the new branch. Two minor notes for follow-up if desired: (1) restore the blank line between `options` and `let response` for style consistency; (2) confirm no stage proxy URL is needed before merging. Nothing blocking. Approved.

PR #5755, opened 2026-04-02
[SR] Router marquee performance tweaks

rgclayton:
* Reduce inertia during fast scrolling to avoid sustained RAF CPU usage
* Pause auto rotation and video playing when scrolling the router marquee out of view. This also stop the background loading that continues to happen even when viewing content lower on the page.

For the 2nd bullet point open up dev tools network tab and watch the media.mp4's stop loading when the router marquee scrolls out of view. This was using extra cpu resources while scrolling down the page. The cpu seems to spin up less. Give it a look. 

Details around the first bullet point:
How it works:

- `deltaY > 80` — threshold for “fast” scroll (a typical slow scroll is ~30–50, fast flick is 100+). Tune as needed.
- `lerp = 0.4` — during fast scroll, Lenis converges ~5x faster to the target position, so the RAF loop idles sooner instead of easing for hundreds of ms.
- After 300ms of no fast-scroll events, lerp resets to the original smooth value.
- Why this helps CPU: Lenis’s RAF loop runs on every frame until Math.abs(velocity) < threshold. With lerp = 0.08, a fast fling can keep it running for 1–2 seconds per gesture. At lerp = 0.4, it converges in a few frames.

**Tuning knobs if needed:**
Raise 80 threshold if it kicks in too eagerly on normal scrolling
Lower 0.4 toward 1.0 for even less inertia on fast scroll (1.0 = instant, no smoothing)

Raise 300 debounce if the smooth feel is snapping back too soon during continuous fast scroll

**Test URLs:**
- Before: https://main--upp--adobecom.aem.page/homepage/drafts/ramuntea/redesign-demo?milolibs=site-redesign-foundation&martech=off
- After: https://main--upp--adobecom.aem.page/homepage/drafts/ramuntea/redesign-demo?milolibs=rm-performance-tweaks&martech=off

### DIFF
--- a/libs/c2/blocks/router-marquee/router-marquee.js
+++ b/libs/c2/blocks/router-marquee/router-marquee.js
@@ -314,6 +314,7 @@ const startAutoplay = (slides, cards, container, block) => {
   let timer = null; // timer for the autoplay
   let paused = false; // whether the autoplay is paused
   let userPaused = false; // whether the user explicitly paused via the play/pause button
+  let offscreen = false; // whether the block is outside the viewport
   let cleanupTimer = null; // cleanup timer that resets temp inline styles
   let pendingSlide = null; // the slide that is currently transitioning in
   let leaveTimer = null; // timer for restarting autoplay on block mouse leave
@@ -428,6 +429,7 @@ const startAutoplay = (slides, cards, container, block) => {
   };
 
   const advance = () => {
+    if (paused) return;
     clearTimeout(timer);
     clearFill(active);
     activate((active + 1) % cardEls.length, 1);
@@ -436,8 +438,17 @@ const startAutoplay = (slides, cards, container, block) => {
     preloadNextVideo();
   };
 
+  const cancelLeaveTimer = () => {
+    clearTimeout(leaveTimer);
+    leaveTimer = null;
+  };
+
   const pause = () => {
     clearTimeout(timer);
+    clearTimeout(cleanupTimer);
+    cleanupTimer = null;
+    cancelLeaveTimer();
+    finishSlideTransition();
     clearFill(active);
     paused = true;
     setPlayingState(false);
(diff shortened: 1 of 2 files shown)


libs/c2/blocks/router-marquee/router-marquee.js
```diff
@@ -705,4 +652,33 @@ export default function init(el) {
     loadViewportVideos(el);
     initViewportAutoplay();
   });
+
+  // Pause/resume based on scroll position. The router-marquee fills the viewport at
+  // the top of the page. Once the user has scrolled past half the viewport height,
+  // the block is considered offscreen. window.scrollY is reliable regardless of
+  // whether Lenis or native scroll is driving the animation.
```
JasonHowellSlavin, line 659: Did you want to keep these comments in? I know we tend to leave comments out if we can.

libs/c2/blocks/router-marquee/router-marquee.js
```diff
@@ -705,4 +652,33 @@ export default function init(el) {
...
+      autoplayControllers.forEach((ctrl) => ctrl.setVisible(false));
+    } else if (isVisible && visibilityPaused) {
+      visibilityPaused = false;
+      autoplayControllers.forEach((ctrl) => ctrl.setVisible(true));
+    }
+  };
+
+  // Lenis loads after blocks, so window.lenis isn't set at init() time.
+  // On the first native scroll event, swap to lenis.on() if available so
+  // subsequent visibility checks run inside Lenis's RAF loop.
```
JasonHowellSlavin, line 675: same here with these.

libs/utils/utils.js
```diff
@@ -2080,6 +2080,15 @@ async function loadPostLCP(config) {
       lerp,
       prevent: (node) => lenisPreventClasses.some((cls) => node.classList?.contains(cls)),
     });
+    // Reduce inertia during fast scrolling to avoid sustained RAF CPU usage
+    let fastScrollTimer;
+    window.addEventListener('wheel', (e) => {
+      if (Math.abs(e.deltaY) > 70) {
+        window.lenis.options.lerp = 0.6;
```
JasonHowellSlavin, line 2087: not sure here, but you mentioned above that lenis loads after blocks, so it might be good to optional chain here? Or check for `window.lenis`?

JasonHowellSlavin (commented):
Were there any tests we could add? I see this is for c2, so not sure how that is set up.

rgclayton:
> Were there any tests we could add? I see this is for c2, so not sure how that is set up.

Thanks Jason! No, no C2 block as tests yet. This is for a test that will be going on the next month. We will have to circle back to address this. That's why comments are still in place for some follow up and knowledge.

libs/c2/blocks/router-marquee/router-marquee.js
rgclayton: Yep, notes for fast follow. Thank you!

rgclayton:
Closing for now. Will revisit while test is running.

PR #4461, opened 2025-06-24
[MWPW-175255] Actionless action item a11y

overmyheadandbody:
Although the action item was not intended to be used without actual actions (links), it seems that there are cases where it is used in that manner. The arrows that move the items left or right are accessible, but axe is suggesting that the whole area needs to be focusable as well. This is what the current PR aims to implement. Nothing changes for action scrollers that do have nested links.

axe suggestion:
Screenshot 2025-06-24 at 18 17 59

Resolves: [MWPW-175255](https://jira.corp.adobe.com/browse/MWPW-175255)

**Test URLs:**
- Before: https://main--milo--overmyheadandbody.aem.live/docs/library/kitchen-sink/action-items?martech=off
- After: https://actionless-action-item--milo--overmyheadandbody.aem.live/docs/library/kitchen-sink/action-items?martech=off
- Before (original issue): https://main--da-bacom--adobecom.aem.live/products/firefly-business?martech=off
- After (original issue): https://main--da-bacom--adobecom.aem.live/products/firefly-business?martech=off&milolibs=actionless-action-item--milo--overmyheadandbody

PR #5773, opened 2026-04-06
[MWPW-190455] - Action item a11y

DKos95:
This PR addresses an issue for action-scroller where just list is read out and sometimes some list items are randomly announced.

After consultation with our a11y consultant it has been confirmed that if items aren't interactable they shouldn't be in a list.
One thing that required a change after we removed `role=list` is changing from `tabIndex=0` to `aria-hidden=true` in order to preserve the fix from this PR https://github.com/adobecom/milo/pull/4461 and for the screenreader to read out all of the alt names from images when tabbed to in this use case.
The action-scroller element is tabable by default because it has overflow which browsers automatically add to the list of tabable elements.

Resolves: [MWPW-190455](https://jira.corp.adobe.com/browse/MWPW-190455)

**PSI Test URLs:**
- After: https://action-item-a11y--milo--adobecom.aem.page/?martech=off

**Test URLs:**
- Before: https://main--da-bacom--adobecom.aem.live/products/express-business
- After: https://main--da-bacom--adobecom.aem.live/products/express-business?milolibs=action-item-a11y

### DIFF
--- a/libs/blocks/action-scroller/action-scroller.js
+++ b/libs/blocks/action-scroller/action-scroller.js
@@ -119,6 +119,7 @@ export default function init(el) {
   el.replaceChildren(items, ...buttons);
   if (hasNav) {
     handleBtnState(items, buttons);
+    if (!items.querySelectorAll('a').length) items.setAttribute('aria-hidden', 'true');
     allActionScrollers.push({ scroller: items, buttons });
     import('../../utils/action.js').then(({ debounce }) => {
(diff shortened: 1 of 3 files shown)


robert-bogos (approved):

overmyheadandbody (approved):

narcis-radu (approved):

biljana-cvijanovic (approved):
Validated. Testing details https://jira.corp.adobe.com/browse/MWPW-190455

PR #5783, opened 2026-04-07
[MWPW-191878] - Sr base card click

DKos95:
PR aims to make the whole base card clickable with hover animations being triggered when hovering any part of the base card. The approach has been vetted by an a11y consultant.

Resolves: [MWPW-191878](https://jira.corp.adobe.com/browse/MWPW-191878)
Resolves: [MWPW-192292](https://jira.corp.adobe.com/browse/MWPW-192292)

**PSI Test URLs:**
- After: https://sr-base-card-click--milo--adobecom.aem.page/?martech=off

**Test URLs:**
- Before: https://load-c2-styles--upp--adobecom.aem.page/homepage/drafts/ramuntea/redesign-demo
- After: https://load-c2-styles--upp--adobecom.aem.page/homepage/drafts/ramuntea/redesign-demo?milolibs=sr-base-card-click

### DIFF
--- a/libs/c2/blocks/base-card/base-card.css
+++ b/libs/c2/blocks/base-card/base-card.css
@@ -61,11 +61,16 @@
     .standalone-link {
       color: var(--s2a-color-gray-1000);
       text-decoration: none;
-      position: relative;
       display: inline-block;
       margin-top: var(--s2a-spacing-24);
       margin-inline-end: var(--s2a-spacing-12);
       margin-bottom: var(--s2a-spacing-4);
+
+      &::before {
+        content: '';
+        position: absolute;
+        inset: 0;
+      }
     }
   }
 
--- a/libs/c2/blocks/base-card/base-card.css
+++ b/libs/c2/blocks/base-card/base-card.css
@@ -70,6 +70,7 @@
         content: '';
         position: absolute;
         inset: 0;
+        z-index: 1;
       }
     }
   }
--- a/libs/c2/blocks/base-card/base-card.css
+++ b/libs/c2/blocks/base-card/base-card.css
@@ -47,7 +47,12 @@
       width: 100%;
       object-fit: cover;
       aspect-ratio: 4/3;
+      transition: transform 0.3s ease;
     }
+  }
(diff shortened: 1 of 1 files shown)


zagi25:
> The reason we didn't implement this at the time was because we thought we would need an inline link in the body copy for Nano Banana in this card:

Based on this message from the thread in the ticket, inline links are possible part of the card text? With this solution they won't be clickable right ?

DKos95:
> > The reason we didn't implement this at the time was because we thought we would need an inline link in the body copy for Nano Banana in this card:
> 
> Based on this message from the thread in the ticket, inline links are possible part of the card text? With this solution they won't be clickable right ?

I didn't understand it like that, I understood it as: We didn't implement whole card click because we thought we would need inline link in the body, but now since we know we don't need it we can make the whole card clickable, right?

overmyheadandbody (commented):
Would it instead be possible to switch the whole card to use an `a` tag? This would allow user text selection for text elements in the card and would also provide more relevant context to screen readers, as the whole content would be announced. The focus state when tabbing through elements would land on the whole card instead of just the single link.

DKos95:
> Would it instead be possible to switch the whole card to use an `a` tag? This would allow user text selection for text elements in the card and would also provide more relevant context to screen readers, as the whole content would be announced. The focus state when tabbing through elements would land on the whole card instead of just the single link.

The issue with that is that then the whole text of the card would be announced which our a11y consultant said is overwhelming, this tab on link instead of cards was also veted, this approach has been described as being the best of both worlds for screenreader users.

narcis-radu (approved):

robert-bogos (approved):

overmyheadandbody (approved):
Still not a fan of this, but the counterproposal is much worse, so this should be ok for now.

Deva309 (approved):

NadiiaSokolova (approved):
Verified. Ready to be merged.
Testing details [MWPW-191878](https://jira.corp.adobe.com/browse/MWPW-191878)

PR #5785, opened 2026-04-07
[SR] page performance tweaks

rgclayton:
Working through some page performance tweaks to minimize cpu spin up, layout thrashing and paint storms during scrolling.
* Reduce inertia during fast scrolling to avoid sustained RAF CPU usage
* Added `content-visibility: auto:` For sections lower on the page (like the “Explore what’s new”,  “Adobe News” or “Tools that work for you”), use content-visibility: auto. This allows the browser to skip the rendering work (layout and painting) for off-screen content until the user scrolls near it.
* Added`contain: layout`; News-cards, elastic-cards, and base-cards. This tells the browser that changes inside those boxes won’t affect the layout of the rest of the page, limiting the scope of “Reflows.”

Resolves: [MWPW-191875](https://jira.corp.adobe.com/browse/MWPW-191875)

**Test URLs:**
- Before: https://main--upp--adobecom.aem.page/homepage/drafts/ramuntea/redesign-demo?mep=off&martech=off
- After: https://main--upp--adobecom.aem.page/homepage/drafts/ramuntea/redesign-demo?milolibs=sr-page-scroll-perf&mep=off&martech=off

**Notes**
How the inertia throttling works:

- `deltaY > 80` — threshold for “fast” scroll (a typical slow scroll is ~30–50, fast flick is 100+). Tune as needed.
- `lerp = 0.4` — during fast scroll, Lenis converges ~5x faster to the target position, so the RAF loop idles sooner instead of easing for hundreds of ms.
- After `300ms` of no fast-scroll events, lerp resets to the original smooth value.

Why this helps CPU: Lenis’s RAF loop runs on every frame until Math.abs(velocity) < threshold. With lerp = 0.08, a fast fling can keep it running for 1–2 seconds per gesture. At lerp = 0.4, it converges in a few frames.

Adjusting numbers:
- Raise `90` threshold if it kicks in too eagerly on normal scrolling
- Lower `0.4` toward `1.0` for even less inertia on fast scroll (1.0 = instant, no smoothing)
- Raise `300` debounce if the smooth feel is snapping back too soon during continuous fast scroll

### DIFF
--- a/libs/c2/blocks/section-metadata/section-metadata.css
+++ b/libs/c2/blocks/section-metadata/section-metadata.css
@@ -62,10 +62,6 @@
     align-items: stretch;
     grid-auto-rows: initial;
   }
-
-  &:not(.section:first-of-type) {
-    content-visibility: auto;
-  }
 }
 
 .section[class*='grid-width-'] {
--- a/libs/c2/blocks/section-metadata/section-metadata.css
+++ b/libs/c2/blocks/section-metadata/section-metadata.css
@@ -62,6 +62,10 @@
     align-items: stretch;
     grid-auto-rows: initial;
   }
+
+  &:not(.section:first-of-type) {
+    content-visibility: auto;
+  }
 }
 
 .section[class*='grid-width-'] {
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -2095,6 +2095,9 @@ async function loadPostLCP(config) {
       loadScript(`${config.base}/deps/lenis.min.js`),
     ]);
     const lerp = parseFloat(PAGE_URL.searchParams.get('inertialFactor')) || 0.08;
+    const fsThreshold = parseFloat(PAGE_URL.searchParams.get('fsThreshold')) || 90;
+    const fsFactor = parseFloat(PAGE_URL.searchParams.get('fsFactor')) || 0.4;
+    const fsDelay = parseFloat(PAGE_URL.searchParams.get('fsDelay')) || 300;
     const lenisPreventClasses = ['dialog-modal', 'ot-sdk-container', 'global-navigation'];
     window.lenis = new window.Lenis({
       autoRaf: true,
@@ -2104,10 +2107,10 @@ async function loadPostLCP(config) {
     // Reduce inertia during fast scrolling to avoid sustained RAF CPU usage
     let fastScrollTimer;
     window.addEventListener('wheel', (e) => {
-      if (Math.abs(e.deltaY) > 90) {
-        window.lenis.options.lerp = 0.4;
+      if (Math.abs(e.deltaY) > fsThreshold) {
+        window.lenis.options.lerp = fsFactor;
         clearTimeout(fastScrollTimer);
-        fastScrollTimer = setTimeout(() => { window.lenis.options.lerp = lerp; }, 300);
+        fastScrollTimer = setTimeout(() => { window.lenis.options.lerp = lerp; }, fsDelay);
       }
     }, { passive: true });
   }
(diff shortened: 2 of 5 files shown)


robert-bogos (approved):

narcis-radu:
@rgclayton - I'm seeing this behaviour where I scroll very fast, then inertia seems to stop immediately and then resume, scroll a bit and stop again. Best way to see that is to scroll near the "news" section, the stop is a bit too abrupt. What would we need to tweak?

rgclayton:
> @rgclayton - I'm seeing this behaviour where I scroll very fast, then inertia seems to stop immediately and then resume, scroll a bit and stop again. Best way to see that is to scroll near the "news" section, the stop is a bit too abrupt. What would we need to tweak?

@narcis-radu Possibly the debounce rate. It's at 300ms right now.  I can try adding a longer time to see if that helps. 

This was part of the request, because fast scrolling sends rafRate way up and CPU goes nuts. 

Adjusting numbers:

- Raise 90 threshold if it kicks in too eagerly on normal scrolling
- Lower 0.4 toward 1.0 for even less inertia on fast scroll (1.0 = instant, no smoothing)
- Raise 300 debounce if the smooth feel is snapping back too soon during continuous fast scroll

libs/utils/utils.js
```diff
@@ -2095,12 +2095,24 @@ async function loadPostLCP(config) {
       loadScript(`${config.base}/deps/lenis.min.js`),
     ]);
     const lerp = parseFloat(PAGE_URL.searchParams.get('inertialFactor')) || 0.08;
+    const fsThreshold = parseFloat(PAGE_URL.searchParams.get('fsThreshold')) || 90;
```
narcis-radu: Very good for testing purposes. We will need to remove the query parameters before merging the code, I guess the perfect values have been found.

narcis-radu (approved):

libs/c2/blocks/section-metadata/section-metadata.css
```diff
@@ -62,6 +62,10 @@
     align-items: stretch;
     grid-auto-rows: initial;
   }
+
+  &:not(.section:first-of-type) {
+    content-visibility: auto;
+  }
```
overmyheadandbody: I will strongly advise against this. In essence, this makes some sections have 0 height until they enter the viewport, which means we could be introducing a CLS problem. Chrome has started handling it better via some size caching on repeat loads (if the additional `contain-intrinsic-size` is used as well), but not all browsers will apply the same strategy.

This also impacts how Lenis works. Its scroll position calculations will be based on a document with inaccurate height. This can produce layout jumps or unexpected inertias, which is not the intended behavior.

Intersection Observers will not work as expected for blocks below the fold. I'm not sure we can instruct everyone to stop using them, they definitely have their benefits. I'm not saying they don't work at all, but it introduces a lot of quirks and unexpected edge cases.

This can also lead to some stacking content issues, which can affect different absolute positioning or z-index declarations.

Since we're already loading section by section and we don't have the full decorated HTML page there to begin with, we're already getting the benefits of what this is trying to solve, without any of the constraints.

Overall, this introduces more issues than it solves, based on my experimentation with it. Unless there is undeniable proof that this makes a big difference, I think it should be avoided.

overmyheadandbody (changes requested):

libs/c2/blocks/section-metadata/section-metadata.css
rgclayton: @overmyheadandbody Good catch!

Your points are all valid. I'll remove this.

libs/utils/utils.js
rgclayton: Updated.

libs/c2/blocks/section-metadata/section-metadata.css
rgclayton: Removed.

overmyheadandbody (approved):
Since you've started adding `contain: layout` to more complex sections with multiple similar elements, I think it would make sense to apply this to the Explore cards as well.

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-191875

PR #5794, opened 2026-04-09
MWPW-189335 [cc][LANA] non-critical messages reported: target respons…

VaibhavMathurAdobe:
In the personalization.js code , line 1583, an object (respTime) was being passed inside the roundToQuarter() function. Because of which the roundToQuarter() function was returning NAN because it was unable to read the value or a number for respTime. 

image

Instead of the respTime object, we should pass the value of duration(which is available inside the object) to calculate the roundedResponseTime.
Using the following line for reading the duration value if present: "respTime?.duration ?? respTime"

Resolves: [[MWPW-189335](https://jira.corp.adobe.com/browse/MWPW-189335)]

**Test URLs:**

Preview: https://main--milo--vaibhavmathuradobe.aem.page/
Live: https://main--milo--vaibhavmathuradobe.aem.live/

URL used to debug the issue : https://www.adobe.com/products/flashplayer/end-of-life-alternative.html?lanadebug
 
Before the Fix : 
image
image

After the fix : 
image
image

### DIFF
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -1582,7 +1582,7 @@ async function handleMartechTargetInteraction(
     const { targetInteractionData, respTime, respStartTime } = await targetInteractionPromise;
     sendTargetResponseAnalytics(false, respStartTime, calculatedTimeout);
     if (targetInteractionData.result) {
-      const roundedResponseTime = roundToQuarter(respTime);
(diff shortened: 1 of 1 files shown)


sharg1 (approved):
LGTM

Nadeemakramn:
Validation is completed. Good to go live:

Before Change:
Target_NaN_BC
After Change:
image

mokimo (approved):

PR #5797, opened 2026-04-09
[SR] Pause Router-Marquee when out of view

rgclayton:
Added logic to pause the router marquee slide animations when it is scrolled out of view. It resumes the slide animation when scrolled back into view.

This also frees up resources for a small performance bump while scrolling through the page.

Discussed with Chip from the design team and got sign off from him.

Resolves: [MWPW-191875](https://jira.corp.adobe.com/browse/MWPW-191875)

**Test URLs:**
- Before: https://main--upp--adobecom.aem.page/homepage/drafts/ramuntea/redesign-demo?martech=off
- After: https://main--upp--adobecom.aem.page/homepage/drafts/ramuntea/redesign-demo?milolibs=sr-rm-scroll-pause&martech=off

### DIFF
--- a/libs/c2/blocks/router-marquee/router-marquee.js
+++ b/libs/c2/blocks/router-marquee/router-marquee.js
@@ -386,6 +386,7 @@ const startAutoplay = (slides, cards, container, block) => {
   let timer = null; // timer for the autoplay
   let paused = false; // whether the autoplay is paused
   let userPaused = false; // whether the user explicitly paused via the play/pause button
+  let offscreen = false; // whether the block is outside the viewport
   let cleanupTimer = null; // cleanup timer that resets temp inline styles
   let pendingSlide = null; // the slide that is currently transitioning in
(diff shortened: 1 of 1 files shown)


overmyheadandbody (approved):

robert-bogos (changes requested):
I saw a few things that need to be changed. I'll come back after I fully articulate the changes needed

robert-bogos (changes requested):
Initially I wanted to request the changes here, but I think is easier if I just open a new PR: https://github.com/adobecom/milo/pull/5823

The gist of it is that this can be achieved by decreasing the lines of code instead of increasing.
Given the new desired behaviout, a good amount of newly added and former logic can be removed as is redundant.
The scroll event listener approach is risky as is randomly causing a threshold bouncing kind of effect (only randomly reproducible). Is cleaner&safer to stay away from lenis and use an intersection observer.

robert-bogos:
closing this PR as https://github.com/adobecom/milo/pull/5823 is merged and addressing the issue

PR #5801, opened 2026-04-10
[MMM] Fix 500 Internal Server Error when toggling manifest drawer

markpadbe:
**Summary**

Toggling any page/manifest drawer on mmm page (i.e., clicking  ".mmm-trigger" button) triggers a 500 Internal Server Error from the lambda API, followed by an uncaught TypeError: Cannot destructure property 'page' of 'mepConfig' as it is null. This update prevents normalizePath from duplicating query params on external URLs while preserving the existing entitlements param-retention behavior for known origins.

- personalization.js: Introduce isKnownOrigin flag to distinguish AEM/HLX/adobe/localhost URLs from external ones. Known origins continue using pathname + search + hash (preserving the 97b9337 entitlements fix). External URLs reconstruct via new URL(path) to avoid duplicating query params.
- mmm.js: Add null guard on pageData in toggleDrawer to prevent the destructure-of-null crash when fetchData returns null.
- normalizePath.test.js: Two new tests covering external URLs with query params and #_dnt hash stripping.

**Root Cause**
#5450  added ${search} to the return in normalizePath to fix entitlement sheet parameters not being retained. This works correctly for AEM/HLX/adobe/localhost URLs where path is reassigned to just pathname, but for external URLs (like the lambda API), path is never reassigned and still contains the full URL with query params. Appending ${search} duplicates them. The server receives a malformed URL and returns 500. fetchData returns null, which propagates to getMepPopup(null, true) where destructuring null throws.

Steps to QA Using the Before and After Test URLS below: 
1. Scroll down to the first page with manifests found
2. Toggle the + to the right of the page name
Expected Results: Manifest content is retrieved and displayed
Actual Results: spinner loads and 500 error thrown in console.

Resolves: [MWPW-192357](https://jira.corp.adobe.com/browse/MWPW-192357)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/docs/authoring/features/mmm/?martech=off
- After: https://mmm-fix500error--milo--adobecom.aem.page/docs/authoring/features/mmm/?martech=off

### DIFF
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -123,11 +123,13 @@ export const normalizePath = (p, localize = true) => {
     const firstFolder = pathname.split('/')[1];
     const mepHash = '#_dnt';
 
-    if (path.startsWith(config.codeRoot)
+    const isKnownOrigin = path.startsWith(config.codeRoot)
       || path.includes('.hlx.')
       || path.includes('.aem.')
       || path.includes('.adobe.')
-      || path.includes('localhost:')) {
+      || path.includes('localhost:');
+
+    if (isKnownOrigin) {
       if (!localize
         || config.locale?.ietf === 'en-US'
         || hash?.includes(mepHash)
@@ -141,7 +143,12 @@ export const normalizePath = (p, localize = true) => {
       }
     }
     path = isFederal ? getFederatedUrl(path) : path;
-    return `${path}${search}${hash.replace(mepHash, '')}`;
+    if (isKnownOrigin) {
+      return `${path}${search}${hash.replace(mepHash, '')}`;
+    }
+    const normalizedUrl = new URL(path);
+    normalizedUrl.hash = normalizedUrl.hash.replace(mepHash, '');
+    return normalizedUrl.toString();
   } catch (e) {
     path = isFederal ? getFederatedUrl(path) : path;
     return path;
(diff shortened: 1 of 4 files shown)


AdobeLinhart (approved):

libs/features/personalization/personalization.js
```diff
@@ -124,7 +126,12 @@ export const normalizePath = (p, localize = true) => {
       }
     }
     path = isFederal ? getFederatedUrl(path) : path;
-    return `${path}${search}${hash.replace(mepHash, '')}`;
+    if (isKnownOrigin) {
+      return `${path}${search}${hash.replace(mepHash, '')}`;
```
denlight, line 147: Not sure we are addressing the case where `path` has `search` already. if it does, we are going to append `search` again, which is what is breaking mmm currently. I believe what was meant to be used is `pathname`?

denlight (changes requested):

libs/features/personalization/personalization.js
markpadbe: Good question. This line (130) only runs when isKnownOrigin is true, which means path has already been reassigned to just pathname on line 121. So appending ${search} here is re-attaching the query params to the stripped pathname, not duplicating them.

The case you're describing (where path retains the full URL) is handled just below on lines 132-134, in the !isKnownOrigin branch. There we reconstruct via new URL(path) which preserves existing query params without appending anything.

denlight (approved):

vgoodric (approved):

jpratt2 (approved):
fixed
https://mmm-fix500error--milo--adobecom.aem.page/docs/authoring/features/mmm/?martech=off

Image

PR #5815, opened 2026-04-15
MWPW-184277: Email collection phone number

zagi25:
* Add `phone country code` and `phone number` fields
* Fetch different Consent string based on `consetId` from metadata
* Bypass checking `isSubscribed` for `cs8A` consent string (this is a short term solution, because MPS doesn't support phone numbers yet, on runtime MPS request is ignored for this consent id and currently we can't check if user is subscribed via MPS API).  
* Refactor and clean up. 

Phone fields for now are only available on Brazil pages. 
Validation of number is done based on [wikipedia](https://en.wikipedia.org/wiki/Telephone_numbers_in_Brazil)

Resolves: [MWPW-184277](https://jira.corp.adobe.com/browse/MWPW-184277)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/br/drafts/ratko/email-collection/phone-number-page
- After: https://mwpw-184277--milo--adobecom.aem.page/br/drafts/ratko/email-collection/phone-number-page


libs/blocks/email-collection/email-collection.js
```diff
@@ -66,6 +69,75 @@ export const [showHideMessage, setMessageEls] = (() => {
   ];
 })();
 
+function validateInput(input) {
+  const { value, type, id } = input;
+  const emailRegex = /^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$/;
```
chrischrischris: I was curious and asked claude about the regex.  Only one in the list that might be a valid concern is the Unicode support:

> A few issues, ordered by practical impact:
> 
> 1. **Allows consecutive dots** — `a..b@foo.com` matches but is invalid (RFC 5322 forbids `..` in local part).
> 2. **Allows leading/trailing dots in local part** — `.foo@bar.com` and `foo.@bar.com` match but are invalid.
> 3. **Allows bad domain boundaries** — `-foo.com` or `foo-.com` match but violate DNS label rules (no leading/trailing hyphens).
> 4. **No Unicode / IDN support** — rejects valid international addresses like `user@münchen.de` or `用户@例子.中国`. TLD class `[a-zA-Z]` also blocks IDN TLDs in non-punycode form.
> 5. **No length bounds** — RFC 5321 caps local part at 64, domain at 255, total at 254.
> 6. **Missing RFC-valid local-part chars** — `!#$&'*/=?^_\`{|}~` are legal but excluded (rarely matters in practice).
> 7. **No IP-literal or quoted local parts** — `user@[192.168.1.1]` and `"a b"@x.com` are valid but rejected (almost never matters).
> 
> For most web forms this regex is "good enough," but if you want correctness, the pragmatic fix is: validate shape loosely, then send a confirmation email. A stricter regex quickly becomes unreadable and still won't match RFC 5322.

libs/blocks/email-collection/utils.js
```diff
@@ -259,8 +320,14 @@ export const [getFormData, setFormData] = (() => {
 
       if (!fields.email
         || !metadata.mpsSname
-        || !metadata.subscriptionName) return 'Section metadata is missing email/mps-sname/subscription-name field';
+        || (!metadata.subscriptionName && !metadata.consentId)) return 'Section metadata is missing email/mps-sname/subscription-name/consent-id field';
+
+      if ((fields['phone-number'] && !fields['phone-country-code'])
+      || (!fields['phone-number'] && fields['phone-country-code'])) {
+        return 'Phone number field is missing phone country code field';
```
chrischrischris, line 326: nit: message is off for when country code is filled but number isn't

libs/blocks/email-collection/email-collection.js
zagi25: I've added unicode support.

libs/blocks/email-collection/utils.js
zagi25: Screenshot 2026-04-28 at 11 10 09
This message comes from authored json (excel), above message is only used to notify user that form is not authored correctly.

libs/blocks/email-collection/email-collection.css
```diff
@@ -233,6 +273,10 @@
   .email-collection.no-image {
     max-width: 700px;
   }
+
+  .email-collection.no-image.small-width {
```
mokimo, line 277: **Dead CSS — `.small-width` class is never applied**

This rule is defined but no JS code ever adds `small-width` to the element. Either the class needs to be applied somewhere in `init` / `decorate`, or this rule should be removed.

libs/blocks/email-collection/email-collection.js
```diff
@@ -492,6 +571,9 @@ function decorateText(elChildren) {
 }
 
 async function checkIsSubscribed() {
+  const { consentId } = await getFormData('consent');
+  if (consentId?.startsWith('cs8A')) return false;
```
mokimo: **`cs8A` bypass is fragile and untested**

The `consentId` resolved here comes from the **second `<div>` in the fetched consent HTML file**, not from the section metadata. The test mock always returns `'consent-id'` in that position, so this bypass never fires in tests and the path has no coverage.

Also, the comparison is case-sensitive — if the HTML file returns `cs8a` (lowercase), the bypass silently fails. Consider:
```js
if (consentId?.toLowerCase().startsWith('cs8a')) return false;
```
or compare directly against `getFormData('metadata').consentId` which is the value you actually control.

libs/blocks/email-collection/email-collection.js
```diff
@@ -293,62 +413,6 @@ function setMaxHeightToForm(formContainer, restore) {
...
-    focusInvalid = focusInvalid ?? input;
-  });
-
-  focusInvalid?.blur();
-  setTimeout(() => focusInvalid?.focus(), 50);
-
-  return !isInvalidForm;
-}
-
 async function decorateConsentString() {
```
mokimo, line 418: **`subscriptionName` may be `undefined`, producing the literal string `"undefined"` in the consent HTML**

With this PR, `subscription-name` is no longer required when `consent-id` is provided. When only `consent-id` is set, `subscriptionName` will be `undefined`, and `replaceAll` will coerce it to the string `"undefined"`. If a consent document ever contains `{{subscription-name}}`, the user will see that word.

Fix:
```js
consentDiv.innerHTML = consentDiv.innerHTML.replaceAll(regex, subscriptionName ?? '');
```

libs/blocks/email-collection/email-collection.js
```diff
@@ -66,6 +69,75 @@ export const [showHideMessage, setMessageEls] = (() => {
...
+  input.classList.toggle('invalid', errorType);
+  label.classList.toggle('invalid', errorType);
+  error?.classList.toggle('hidden', !errorType);
+
+  return !!errorType;
+}
+
+async function formatPhoneNumber(input) {
+  if (!input) return;
+  const { code, format: formatNumber } = getPhoneFieldConfig();
```
mokimo: **`formatPhoneNumber` will throw if `getPhoneFieldConfig()` returns `undefined`**

Destructuring `undefined` throws a TypeError. While phone fields are currently only configured for Brazil, this `change` listener is attached unconditionally and will crash on any unsupported locale.

```js
const config = getPhoneFieldConfig();
if (!config) return;
const { code, format: formatNumber } = config;
```

libs/blocks/email-collection/email-collection.js
```diff
@@ -139,69 +226,95 @@ function decorateSelect({ data, id, placeholder, selectValue }) {
...
+    });
+    return input;
+  },
+  'phone-country-code': (params) => {
+    const { code, icon } = getPhoneFieldConfig() ?? {};
+    if (!code) return null;
+    const { labelText } = params;
+    const input = createInput({ ...params, inputValue: code });
+    const countryCodeContainer = createTag('div', { class: 'phone-country-code-container' });
+    countryCodeContainer.style.setProperty('--phone-country-ch-size', `${labelText.length}ch`);
```
mokimo, line 292: **`--phone-country-ch-size` is coupled to the label text length, not the input value**

`labelText` here is the author-provided label (e.g. `"Código de área"` → 14 chars), not the country code value (`"+55"` → 3 chars). If the label changes in translation or a different locale has a shorter label than the code requires, the layout will break.

Consider basing this on the code value instead:
```js
countryCodeContainer.style.setProperty('--phone-country-ch-size', `${code.length + 2}ch`);
```

libs/blocks/email-collection/utils.js
```diff
@@ -75,8 +83,61 @@ export const FORM_FIELDS = {
...
+      type: 'tel',
+      required: true,
+      autocomplete: 'tel',
+    },
+  },
+  'phone-country-code': {
+    tag: 'input',
+    attributes: {
+      type: 'tel',
+      required: true,
```
mokimo: **`required: true` on a disabled field shows a misleading required indicator**

The `phone-country-code` input is always disabled (auto-populated), yet `required: true` causes `label.classList.toggle('required', attributes.required)` to add the required class — showing a `*` indicator on a field the user cannot interact with. Consider removing `required: true` here since the value is always set programmatically.

libs/blocks/email-collection/email-collection.js
```diff
@@ -66,6 +69,75 @@ export const [showHideMessage, setMessageEls] = (() => {
...
+async function validateForm(form) {
+  const inputs = form.querySelectorAll('input:not([readonly]), select');
+  let isInvalidForm = false;
+  let focusInvalid;
+  const { placeholders } = await getFormData('config');
+  inputs.forEach((input) => {
+    if (!showInputError(form, input, placeholders)) return;
+    isInvalidForm = true;
+
+    /* eslint-disable */
```
mokimo, line 122: **`eslint-disable` block and `_hasChangeListener` pattern carried over unchanged**

This was already present in the removed code before the refactor. Since this block was rewritten, it's a good opportunity to clean it up — a `WeakSet` would avoid the non-standard expando property and the eslint suppression:
```js
const changeListeners = new WeakSet();
// ...
if (!changeListeners.has(input)) {
  input.addEventListener('change', () => showInputError(form, input, placeholders));
  changeListeners.add(input);
}
```

libs/blocks/email-collection/email-collection.js
```diff
@@ -66,6 +69,75 @@ export const [showHideMessage, setMessageEls] = (() => {
...
+  if ((type !== 'email' && !input.checkValidity())
+  || (type === 'email' && !value)) return 'required';
+  if (type === 'email' && !emailRegex.test(value)) return 'email';
+  if (id === 'phone-number' && !validatePhoneNumber(value)) return 'phone';
+  return '';
+}
+
+function showInputError(form, input, placeholders) {
+  const { id } = input;
+  const label = form.querySelector(`label[for=${id}]`);
```
mokimo: **Unquoted attribute selector will break for hyphenated IDs**

`label[for=${id}]` — when `id` is `phone-number` or `phone-country-code`, the unquoted value makes this an invalid CSS selector and `querySelector` will throw. This is a pre-existing bug that now affects the new phone fields.

Fix:
```js
const label = form.querySelector(`label[for="${id}"]`);
```

libs/blocks/email-collection/email-collection.css
mokimo: Non-AI: Is this added as modifier within the block?

mokimo (approved):
Sent over an agent and kept the (hopefully) rightful catches - feel free to dismiss anything that's not right.

Other then that, looks good to me.

libs/blocks/email-collection/email-collection.css
narcis-radu: Looks like a variant: https://mwpw-184277--milo--adobecom.aem.page/br/drafts/ratko/email-collection/phone-number-page#phone-number

libs/blocks/email-collection/email-collection.js
narcis-radu: I don't think this will be happening, but better to put in some sort of safety net.

narcis-radu (approved):
Looks good, but some things can be improved. Not a blocker though.

libs/blocks/email-collection/email-collection.css
zagi25: Yes, it is a new variant because of the design.

biljana-cvijanovic (approved):
Validated. Test details https://jira.corp.adobe.com/browse/MWPW-184277

PR #5823, opened 2026-04-16
[SR] Pause router marquee when scrolled out of view

robert-bogos:
Added logic to pause the router marquee autoplay when it is scrolled out of view. It resumes when scrolled back into view.
This also frees up resources for a small performance bump while scrolling through the page.

**Test URLs:**
- Before: https://main--upp--adobecom.aem.page/homepage/drafts/ramuntea/redesign-demo?martech=off
- After: https://main--upp--adobecom.aem.page/homepage/drafts/ramuntea/redesign-demo?milolibs=rm-pause-play&martech=off

### DIFF
--- a/libs/c2/blocks/router-marquee/router-marquee.js
+++ b/libs/c2/blocks/router-marquee/router-marquee.js
@@ -82,7 +82,6 @@ const STAGGER_MS = 1000;
 const STAGGER_BASE = 60;
 const STAGGER_STEP = 20;
 const EASE = 'cubic-bezier(0.42, 0, 0, 1)';
-const RESUME_DELAY = 2000;
 const SWIPE_THRESHOLD = 100;
 
 const prefersReducedMotion = () => window.matchMedia('(prefers-reduced-motion: reduce)').matches;
@@ -385,10 +384,8 @@ const startAutoplay = (slides, cards, container, block) => {
   let active = 0; // index of the current active slide
   let timer = null; // timer for the autoplay
   let paused = false; // whether the autoplay is paused
-  let userPaused = false; // whether the user explicitly paused via the play/pause button
   let cleanupTimer = null; // cleanup timer that resets temp inline styles
   let pendingSlide = null; // the slide that is currently transitioning in
-  let leaveTimer = null; // timer for restarting autoplay on block mouse leave
 
   const isMobile = () => !window.matchMedia('(min-width: 1280px)').matches;
   const isDesktopSmallVp = isMobile()
@@ -500,6 +497,7 @@ const startAutoplay = (slides, cards, container, block) => {
   };
 
   const advance = () => {
+    if (paused) return;
     clearTimeout(timer);
     clearFill(active);
     activate((active + 1) % cardEls.length, 1);
@@ -509,7 +507,9 @@ const startAutoplay = (slides, cards, container, block) => {
   };
 
   const pause = () => {
+    if (paused) return;
     clearTimeout(timer);
+    finishSlideTransition();
     clearFill(active);
     paused = true;
     setPlayingState(false);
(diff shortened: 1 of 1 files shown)


rgclayton:
Nit. I'm noticing that if I've scrolled the router-marquee out of view and resize the page to a different viewport that autoplaying of router-marquee starts back up, even though it's not in view.

DKos95 (approved):

robert-bogos:
> Nit. I'm noticing that if I've scrolled the router-marquee out of view and resize the page to a different viewport that autoplaying of router-marquee starts back up, even though it's not in view.

Yes, this an arhitecture compromise to support different content for every viewport. What happens is that each viewport has it's own slides so if you resize the page manually, hitting a new viewport, the relevant slides are rendered and they start playing automatically.

overmyheadandbody:
@robert-bogos, couldn't there be a gate checking whether the router-marquee is in view on the `change` event and set some variables based on that, that would prevent the video from playing?

If it's super-easy, maybe it's worth implementing. Otherwise, I'd argue it's not a common enough use-case that users resize their browser to hit another breakpoint, thus it won't have much impact in the real world.

libs/c2/blocks/router-marquee/router-marquee.js
```diff
@@ -705,4 +686,11 @@ export default function init(el) {
     loadViewportVideos(el);
     initViewportAutoplay();
   });
+
+  new IntersectionObserver(([entry]) => {
+    const action = entry.isIntersecting ? 'pause' : 'resume';
+    if (entry.isIntersecting || entry.boundingClientRect.top > 0) {
+      autoplayControllers.forEach((ctrl) => ctrl[action]());
+    }
+  }, { rootMargin: '0px 0px -30% 0px' }).observe(el.closest('.section')?.nextElementSibling);
```
overmyheadandbody: What if the router marquee is the only block on the page? I know it's not a use-case we'd have on our pages, but a NALA integration test or a Kitchen Sink example might have it as the only block on the page, case in which `nextElementSibling` might fail.

I'd suggest limiting this to the router marquee itself (so just `el`).

libs/c2/blocks/router-marquee/router-marquee.js
```diff
@@ -82,7 +82,6 @@ const STAGGER_MS = 1000;
 const STAGGER_BASE = 60;
 const STAGGER_STEP = 20;
 const EASE = 'cubic-bezier(0.42, 0, 0, 1)';
-const RESUME_DELAY = 2000;
```
overmyheadandbody, line 85: Is it intended that the loop never continues after a user hovers on one of the cards? A user would now have to press play, is that an updated requirement?

overmyheadandbody (commented):
This has a somewhat unintended positive side-effect: due to how the cross-section animation is set up between the router marquee and the next section, some content lower on the page was being moved around when the router marquee slides were shifting; this only occurred for smaller height viewports, when the router marquee had to adapt the height of the slides. With this implementation, that effect is gone, which is great!

However, I have a couple of questions around some other side-effects this would introduce.

robert-bogos:
> @robert-bogos, couldn't there be a gate checking whether the router-marquee is in view on the `change` event and set some variables based on that, that would prevent the video from playing?
> 
> If it's super-easy, maybe it's worth implementing. Otherwise, I'd argue it's not a common enough use-case that users resize their browser to hit another breakpoint, thus it won't have much impact in the real world.

We could add a gate checking mechanism to handle this. Is not super hard, also not super easy. I share your opinion that this is not a common enough edge-case to be worth the hustle. 
Worse case scenario, if someone does this, the only thing that will happen is that the router-marquee will autoplay once when out of view.

libs/c2/blocks/router-marquee/router-marquee.js
robert-bogos: Thank you 🙏🏻 Did not think that NALA/KS examples might place the router marquee alone on the page. I wouldn't limit the logic to the router marquee itself though, because it will require some high level logic changes on how this works.
I think a cleaner solution is to add a guard for when the nextElementSibling is not present. If the marquee is alone on the page, we wouldn't need this IntersectionObserver anyway. I've pushed a commit

libs/c2/blocks/router-marquee/router-marquee.js
robert-bogos: Yes, is intended. That constant was for resuming autoplay when the router marquee is out of view ( pointer exits). If now we want it paused for that scenario, the old logic is outdated.

overmyheadandbody (approved):

SilviuLCF (approved):

SilviuLCF:
QA validated

PR #5826, opened 2026-04-16
MWPW-191951: Detect unpublished M@S fragments in Preflight

joaquinrivero:
* Detects every M@S `<aem-fragment>` referenced on the rendered page and flags any that return non-200 from the Adobe MAS fragment API as CRITICAL failures in the Preflight aggregator.
* Adds an "Unpublished Fragments" section to the Preflight M@S panel listing each unpublished fragment ID, HTTP status, and a "Scroll to card" button.
* Decorates affected cards with a red outline, diagonal hatch background, and a "Not published" corner badge.
* Appends an additive `M@S: N unpublished fragment(s) on this page.` line to the failure toast — does not hard-block publishing.

Resolves: [MWPW-191951](https://jira.corp.adobe.com/browse/MWPW-191951)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/drafts/rivero/mas-preflight/my-document-with-unpublish-fragments
- After: https://mwpw-191951--milo--adobecom.aem.page/drafts/rivero/mas-preflight/my-document-with-unpublish-fragments

Screenshot 2026-04-30 at 3 08 53 PM
Screenshot 2026-04-30 at 3 09 58 PM

### DIFF
--- a/libs/blocks/preflight/checks/merch.js
+++ b/libs/blocks/preflight/checks/merch.js
@@ -1,0 +1,88 @@
+import { getConfig } from '../../../utils/utils.js';
+import { SEVERITY } from './constants.js';
+
+const API_BASE = 'https://www.adobe.com/mas/io/fragment';
+const API_KEY = 'wcms-commerce-ims-ro-user-milo';
+const HYDRATION_DELAY_MS = 3000;
+
+export function findFragmentElements(area = document) {
+  const els = area.querySelectorAll('aem-fragment[fragment]');
+  const entries = [];
+  els.forEach((el) => {
+    const uuid = el.getAttribute('fragment');
+    if (!uuid) return;
+    const card = el.closest('merch-card, merch-card-collection, mas-field') || el;
+    entries.push({ uuid, el, card });
+  });
+  return entries;
+}
+
+export async function checkFragmentPublished(uuid, locale) {
+  const params = new URLSearchParams({ id: uuid, api_key: API_KEY, locale });
+  const url = `${API_BASE}?${params.toString()}`;
+  try {
+    const res = await fetch(url);
+    return { uuid, httpStatus: res.status, published: res.status === 200 };
+  } catch {
+    return { uuid, httpStatus: 0, published: false };
+  }
+}
+
+function resolveLocale(locale) {
+  if (locale) return locale;
+  const ietf = getConfig()?.locale?.ietf || 'en-US';
+  return ietf.replace('-', '_');
+}
+
+export async function checkUnpublishedFragments({ area = document, locale } = {}) {
+  const resolvedLocale = resolveLocale(locale);
+  const entries = findFragmentElements(area);
+  const byUuid = new Map();
+  entries.forEach((entry) => {
+    if (!byUuid.has(entry.uuid)) byUuid.set(entry.uuid, []);
+    byUuid.get(entry.uuid).push(entry);
+  });
+
+  const uuids = [...byUuid.keys()];
+  const results = await Promise.all(
+    uuids.map((uuid) => checkFragmentPublished(uuid, resolvedLocale)),
+  );
+
+  const unpublished = results
+    .filter((r) => !r.published)
+    .map((r) => {
+      const group = byUuid.get(r.uuid);
+      return {
+        uuid: r.uuid,
+        httpStatus: r.httpStatus,
+        elements: group.map((g) => g.el),
+        cards: group.map((g) => g.card),
+      };
+    });
+
+  // eslint-disable-next-line no-console
+  console.log(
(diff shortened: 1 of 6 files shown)


narcis-radu (dismissed):
Hey @joaquinrivero! I strongly believe the direction is incorrect, and I want to make you are aware of the existing concerns and guardrails around this.
* The AEM sidekick should be used as is, without replacing functionality.
* The "Publish" button has an existing workflow, where authors are requested to confirm before the actual publishing happens. We should not alter the behaviour.
* Authors should never be blocked to publish content. The only exception is around specific configuration, where anyone outside a group is blocked from publishing.

Concerns around functionality
* The solution _might_ not always work on DA
* If for any reason the code is not working properly, authors will not be able to publish.

My recommendation is to leverage existing guardrail tools like Preflight. The tool will run in the background and proactively notify authors if there are any critical errors that need to be checked before publishing the page.

mokimo:
Definitely a +1 on this, it's a very complex matter (bulk tools, DA, Sidekick) - most of which we don't own the code for.

I assume this would break once the Sidekick sees any sort of changes, likely conflicts with other features such as the milo publish permissions which look for the publish button. I don't think it's a rabbit hole we should go down into, but look for alternatives.

The preflight platform seems like the better candidate, notify authors and we log the interactions. If an author does something disregarding recommendations of the tool, we can follow up by logging those pages (@skholkhojaev is currently working on a dashboard for displaying and flagging faulty pages)

Blainegunn:
@narcis-radu @mokimo
Thanks for the feedback — All fair points on blocking the publish button so we can move the conversation to pre-flight. I think we need to realign on what pre-flight is meant to do here.

To clarify: we're not blocking anyone from publishing. The "Ignore & Publish" escape hatch is intentional and stays. The goal is to make the risk explicit — authors must consciously bypass the warning rather than unknowingly push broken content live.

That said, now that we're actively migrating everyone to M@S, I think we need to go a step further: pre-flight should be mandatory for pages with merch content, not just advisory. The April 7 CSO is a direct example of what happens when a staged/preview-only fragment quietly ends up in a live manifest — the check should catch that before publish, not after.
I'd like to move this to a proper GitHub discussion so we can align on expectations and scope. My position:

Flag any unpublished/preview-only fragments on pages with merch content as a pre-publish requirement
Keep the escape hatch, but log it as an explicit author decision
Treat this as a CSO-prevention guardrail, not just a UX nicety.

Blainegunn (approved):
keep us safe 🙏

Roycethan (approved):

Roycethan:
@narcis-radu / @mokimo  plz drop your review as well thanks

libs/blocks/preflight/checks/merch.js
```diff
@@ -0,0 +1,88 @@
...
+      return {
+        uuid: r.uuid,
+        httpStatus: r.httpStatus,
+        elements: group.map((g) => g.el),
+        cards: group.map((g) => g.card),
+      };
+    });
+
+  // eslint-disable-next-line no-console
+  console.log(
```
narcis-radu: We could skip the console logs, maybe use LANA if something critical needs to be logged.

libs/blocks/preflight/preflight.css
```diff
@@ -1280,4 +1280,38 @@ img:hover ~ .asset-meta,
...
+.preflight-mas-unpublished::before {
+  content: 'Not published';
+  position: absolute;
+  top: -1px;
+  right: -1px;
+  z-index: 2;
+  padding: 4px 10px;
+  background: #f44;
+  color: #fff;
+  font-family: 'Adobe Clean', adobe-clean, sans-serif;
```
narcis-radu: Wondering if this is really needed, I assume the font family would get inherited together with other definitions.

libs/blocks/preflight/checks/merch.js
```diff
@@ -0,0 +1,88 @@
+import { getConfig } from '../../../utils/utils.js';
+import { SEVERITY } from './constants.js';
+
+const API_BASE = 'https://www.adobe.com/mas/io/fragment';
+const API_KEY = 'wcms-commerce-ims-ro-user-milo';
```
narcis-radu, line 5: any concerns with sharing the key?

Blainegunn: nah, `wcms-commerce-ims-ro-user-milo` it public 
it's in every commerce call we make in production in plain view

libs/blocks/preflight/checks/merch.js
joaquinrivero: Nala is now in place, thanks

Roycethan:
@narcis-radu / @mokimo pending your approvals

PR #5827, opened 2026-04-17
Centralize per-viewport content decoration for C2

overmyheadandbody:
This aims to centralize a method to decorate cross-viewport content, now that we're introducing standard authoring pattern. Blocks where this has already been applied: base-card, rich-content, explore-card. Blocks that might not currently align with the authoring pattern or more investigation is needed from responsible engineers: router-marquee (where each viewport has multiple slides), news (where we're not aware of a need to have per-viewport content; additionally, in the future its content might come from an external source), elastic-carousel (which also has a different authoring pattern at the moment).

The logic currently support both `mobile` and `mobile-viewport` keywords (same for tablet/desktop), until engineers and GWP align on using the latter, which is needed for the localization process.

**Test URLs:**
- Before: https://main--upp--adobecom.aem.page/homepage/drafts/ramuntea/redesign-demo?milolibs=site-redesign-foundation
- After: https://main--upp--adobecom.aem.page/homepage/drafts/ramuntea/redesign-demo?milolibs=vp-content-decoration

### DIFF
--- a/libs/utils/decorate.js
+++ b/libs/utils/decorate.js
@@ -559,7 +559,7 @@ export function decorateAnchorVideo({ src = '', anchorTag }) {
 
 /* NOTE: Experimental logic to substitute authored classes
    with centralized kit classes */
-export function decorateBlockKit(el, kits = []) {
+export function decorateBlockKit(el, kits = {}) {
   const kitClass = [...el.classList].find((cls) => cls.endsWith('-kit'));
   if (!kitClass || !Object.keys(kits).includes(kitClass)) return;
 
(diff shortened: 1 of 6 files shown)


DKos95 (approved):

rgclayton:
Looking good!

Here's [sample page](https://main--upp--adobecom.aem.page/homepage/drafts/rclayton/redesign-demo-rc?milolibs=vp-content-decoration&mep=off) where the first 2 explore-cards have a video on desktop and image on mobile.

libs/utils/decorate.js
```diff
@@ -568,3 +568,185 @@ export function decorateBlockKit(el, kits = []) {
...
+    const setContent = () => {
+      if (!mq.matches) return;
+      el.classList.remove(...allVariants);
+      if (variants.length) el.classList.add(...variants);
+      el.replaceChildren(...children);
+      decorateTextOverrides(el);
+    };
+
+    setContent();
+    mq.addEventListener('change', setContent);
```
rgclayton, line 733: Any concerns about this listener never being cleared?

rgclayton (approved):

zagi25 (approved):

libs/utils/decorate.js
overmyheadandbody: Not really, since we never detach blocks and all of the content is kept in closures. This also gets added only if the block has variance between viewports, which is unlikely to always be the case.

SilviuLCF (approved):
Verified, https://jira.corp.adobe.com/browse/MWPW-193314

PR #5830, opened 2026-04-20
[MWPW-191877] - SR Firefox animations

DKos95:
PR adds support for firefox animations until they implement scroll driven animations.
Plans to implement polyfil were abandoned because of bad performance in combination with lenis, while this approach gives us better performance and requires less code.
Performance was measured to be ~120fps on firefox

Resolves: [MWPW-191877](https://jira.corp.adobe.com/browse/MWPW-191877)

**PSI Test URLs:**
- After: https://ff-lenis-no-pol--milo--adobecom.aem.page/?martech=off

**Firefox Test URLs:**
- Before: https://load-c2-styles--upp--adobecom.aem.live/homepage/drafts/ramuntea/redesign-demo
- After: https://load-c2-styles--upp--adobecom.aem.live/homepage/drafts/ramuntea/redesign-demo?milolibs=ff-lenis-no-pol


libs/utils/utils.js
```diff
@@ -2113,6 +2113,10 @@ async function loadPostLCP(config) {
         fsScrollTimer = setTimeout(() => { window.lenis.options.lerp = lerp; }, fsDelay);
       }
     }, { passive: true });
+    if (!CSS.supports('animation-timeline: view()')) {
```
zagi25: Should we also check for `reduced-motion` here ?

libs/c2/scroll-animations.js
```diff
@@ -0,0 +1,553 @@
+const vh = window.innerHeight;
```
zagi25, line 1: This can change on resize right ?

zagi25:
I noticed few issues when resizing the browser, animations work pretty well on first load, but after screen is resized they start to brake. Also I believe `base-card` image should have a max width of 1920px (same goes for carousel slide)

https://github.com/user-attachments/assets/5c847860-91e9-4f83-9638-8984f7842f7c


https://github.com/user-attachments/assets/323bd77e-5217-424a-b43d-6f6b7e764232

Screenshot 2026-04-20 at 17 50 11

DKos95:
> I noticed few issues when resizing the browser, animations work pretty well on first load, but after screen is resized they start to brake. Also I believe `base-card` image should have a max width of 1920px (same goes for carousel slide)
> 
>  Screen.Recording.2026-04-20.at.17.48.24.mov 
>  Screen.Recording.2026-04-20.at.17.51.08.mov 
> Screenshot 2026-04-20 at 17 50 11

The firefox usage is low as is at 3%, when factoring in how many people will resize, I don't think it even scratches close to 0.5%, not sure if investing time into this is worth it especially with wave 2 coming and even more animations needing firefox support, adding resize to the mix to support ALL animations would make this even more complex.

Assuming 3% of user use firefox and 1-5% resize screen we get: 3% × 1–5% = ~0.03% to 0.15% of users.

libs/utils/utils.js
```diff
@@ -2113,6 +2113,11 @@ async function loadPostLCP(config) {
         fsScrollTimer = setTimeout(() => { window.lenis.options.lerp = lerp; }, fsDelay);
       }
     }, { passive: true });
+    if (!CSS.supports('animation-timeline: view()')
+      && !window.matchMedia('(prefers-reduced-motion: reduce)').matches) {
+      const { default: initScrollAnimations } = await import('../c2/scroll-animations.js');
```
chrischrischris: Does this have to be awaited or would it still work with import().then... ?

libs/c2/scroll-animations.js
```diff
@@ -0,0 +1,555 @@
...
+
+function viewRange({ elHeight, dist, total }, sType, sPct, eType, ePct) {
+  const s = sType === 'entry' ? elHeight * sPct : total * sPct;
+  const e = eType === 'entry' ? elHeight * ePct : total * ePct;
+  return Math.max(0, Math.min(1, (dist - s) / (e - s)));
+}
+
+function sectionHasStyle(section, style) {
+  if (section.classList.contains(style)) return true;
+  if (section.querySelector(`.${style}`)) return true;
```
chrischrischris: nit
```suggestion
  if (section.classList.contains(style) || section.querySelector(`.${style}`)) return true;
```

libs/c2/scroll-animations.js
```diff
@@ -0,0 +1,555 @@
+let vh = window.innerHeight;
+const STAGGER_RE = /parallax-stagger-(ltr|rtl)/;
+const STAGGER_SELECTOR = '[class*="parallax-stagger-ltr"],'
+  + '[class*="parallax-stagger-rtl"]';
+
+/* ── Scroll loop ────────────────────────────────── */
```
chrischrischris: nit: most of these comments don't seem necessary as they mostly duplicate function names

PR #5809, opened 2026-04-14
MWPW-189841 pass location differently to IMS with lingo

vhargrave:
Pass country to ims instead of locale for lingo 

Resolves: [MWPW-189841](https://jira.corp.adobe.com/browse/MWPW-189841)

**Test URLs:**
- Before: https://main--da-cc--adobecom.aem.live/fr/products/photoshop?akamaiLocale=CA&langfirst=on
- After: https://main--da-cc--adobecom.aem.live/fr/products/photoshop?akamaiLocale=CA&langfirst=on&milolibs=vhargrave-pass-country-ims
**Steps to test:** Click on the sign in button in the global navigation "Se connecter". When you're taken to the sign in, you should be seeing fr_CA in the URL and in the cookie preferences (Préférences de cookies) in the bottom right, it should also be pointing to a fr_CA link 

- Before: https://main--da-cc--adobecom.aem.live/fr/products/photoshop?langfirst=on
- After: https://main--da-cc--adobecom.aem.live/fr/products/photoshop?langfirst=on&milolibs=vhargrave-pass-country-ims
**Steps to test:** Click on the sign in button in the global navigation "Se connecter". When you're taken to the sign in, you should be seeing fr_Fr in the URL and in the cookie preferences (Préférences de cookies) in the bottom right, it should also be pointing to a fr_Fr link

- Before: https://main--da-cc--adobecom.aem.live/products/photoshop?akamaiLocale=CA&langfirst=on
- After: https://main--da-cc--adobecom.aem.live/products/photoshop?akamaiLocale=CA&langfirst=on&milolibs=vhargrave-pass-country-ims
**Steps to test:** Click on the sign in button in the global navigation. When you're taken to the sign in, you should be seeing en_US in the URL and in the cookie preferences in the bottom right, it should also be pointing to a en_US link 

**Susi Flow**
The only way to test this is to override the files from this PR in your browser. You'd then expect this link below to also take you to the fr_CA sign in page. You'd also need to add a base property in the cc scripts.js file to be 'fr' for the 'ca_fr' locale.
https://www.adobe.com/fr/products/photoshop/login.html?akamaiLocale=CA&langfirst=on

<details><summary><strong>GNav Test URLs</strong></summary>

**Gnav + Footer + Region Picker modal:**
- Acrobat: https://main--dc--adobecom.aem.live/acrobat?martech=off&milolibs=vhargrave/pass-country-ims--milo--adobecom
[...]

PR #5842, opened 2026-04-22
MWPW-192822 Add geo-ip placeholder resolution with deferred LCP support

markpadbe:
**Summary:** 

Adds Milo-level support for geo-specific placeholders via a -geo-ip key suffix. When langfirst=on is set and a user's country is detected, placeholders ending in -geo-ip resolve from a regional placeholder sheet (e.g. lu_fr/placeholders.json). To protect LCP, resolution in page sections is deferred — base values render immediately and geo values swap in asynchronously after the fetch completes. This feature is independent of DC's existing geo phone number handling — their custom code continues to work unchanged.

**What Changed:** 

- Add -geo-ip placeholder suffix for geo-specific resolution, gated by langfirst=on metadata
- Defer geo-ip resolution in decoratePlaceholderArea (LCP path) -- base values render immediately, geo values swap in asynchronously after the geo placeholder fetch completes
- Direct replaceText callers (e.g. GNAV) use the default defer: false and await geo values before rendering -- no flash of default content
- Preserve existing MEP placeholder priority -- config.placeholders overrides are checked first in the new geo-ip resolution path
- Rename getMepLingoPrefix to getGeoLocalePrefix across utils.js, fragment.js, preview.js, and README.md

**How it works:**
The geo fetch is cached per content root via ensureGeoFetch, so multiple calls share one network request. For GNAV federated content, the origin and path are derived from the caller's contentRoot URL so the fetch targets the correct federated origin.

**Note on resetGeoPlaceholderCache:**
The geo placeholder cache is a module-private Map keyed by content root. Tests must clear it between runs to prevent stale data from a lingo-enabled test leaking into a lingo-disabled test (same cache key, different lingo state). This export is not called in production code.

**Authoring note:**
Unlike the DC -geo suffix (which strips the suffix and falls back to the base key), -geo-ip keys are looked up exactly as written. If a -geo-ip key is used on the page but no matching entry exists in the base placeholder sheet, the fallback text will be the key name (e.g. "phone number geo ip"). Authors must add entries for -geo-ip keys in both the base and regional placeholder sheets.

Steps to QA Feature:
1) Open the Before and After Urls below
2) Observe the 3 placeholders shown in the image

Expected:
1. federal-lu-fr
2. Placeholder: da-cc-lu-fr
3. nested placeholder: da-cc-lu-fr

Actual:
1. feder-fr
2. Placeholder: da-cc-fr
3. nested placeholder: [da-cc-fr](tel:da-cc-fr).

Screenshot 2026-04-21 at 11 09 22 PM

Resolves: [MWPW-192822](https://jira.corp.adobe.com/browse/MWPW-192822)

**Test URLs:**
- Before: https://main--da-cc--adobecom.aem.page/fr/drafts/markp/placeholders?akamaiLocale=lu
- After: https://main--da-cc--adobecom.aem.page/fr/drafts/markp/placeholders?milolibs=phonenumber-ip&akamaiLocale=lu
- Psi: https://phonenumber-ip--milo--adobecom.aem.page/?martech=off

**Update (addressing review feedback):**
- Removed geoPlaceholderCache, ensureGeoFetch, and resetGeoPlaceholderCache — the existing fetchedPlaceholders cache already deduplicates by URL. No test-only exports remain.
- Exported getPlaceholderPaths from utils.js and reused it in getGeoPlaceholders to eliminate the duplicated placeholders-stage check.
- Replaced undefined, undefined args in replaceText calls with explicit PH_RE constant.
- Added federated GNAV test with /federal/globalnav path suffix and dedicated fixture.

**Update 2 (rebase fix + other small review updates):**

- Fixed rebase conflict with #5809: renamed getGeoLocalePrefix → getLingoRegion (returns full region object), added thin getGeoLocalePrefix wrapper for prefix-only callers.
- getPlaceholderPaths import moved to dynamic import inside getGeoPlaceholders (static import triggered circular-dep hang in WTR).
- Added lana warning log in geo placeholder fetch catch (was silent).
- Fixed test pollution in utils.test.js: defensive langfirst meta cleanup in getLingoRegion and loadIms beforeEach blocks.
- Removed extra blank lines per nit feedback.

### DIFF
--- a/libs/features/placeholders.js
+++ b/libs/features/placeholders.js
@@ -1,4 +1,4 @@
-import { customFetch, getConfig, getMetadata } from '../utils/utils.js';
+import { customFetch, getConfig, getMetadata, lingoActive, getMepLingoPrefix } from '../utils/utils.js';
 
 const fetchedPlaceholders = {};
 window.mph = {};
@@ -9,11 +9,11 @@ const getPlaceholdersPath = (config, sheet) => {
   return `${path}${query}`;
 };
 
-const parsePlaceholderJson = async (resp, placeholders) => {
+const parsePlaceholderJson = async (resp, placeholders, { updateMph = true } = {}) => {
   try {
     const json = resp.ok ? await resp.json() : { data: [] };
     json.data?.forEach((item) => {
-      window.mph[item.key] = item.value;
+      if (updateMph) window.mph[item.key] = item.value;
       placeholders[item.key] = item.value;
     });
   } catch (e) {
@@ -21,7 +21,7 @@ const parsePlaceholderJson = async (resp, placeholders) => {
   }
 };
 
-const fetchPlaceholder = (path, placeholderRequest) => new Promise(
+const fetchPlaceholder = (path, placeholderRequest, { updateMph = true } = {}) => new Promise(
   // eslint-disable-next-line no-async-promise-executor
   async (resolve) => {
     const resp = await placeholderRequest || await customFetch(
@@ -31,9 +31,9 @@ const fetchPlaceholder = (path, placeholderRequest) => new Promise(
 
     if (Array.isArray(resp)) {
       // Overlay placeholders
-      for (const r of resp) await parsePlaceholderJson(r, placeholders);
+      for (const r of resp) await parsePlaceholderJson(r, placeholders, { updateMph });
     } else {
-      await parsePlaceholderJson(resp, placeholders);
+      await parsePlaceholderJson(resp, placeholders, { updateMph });
     }
 
     resolve(placeholders);
@@ -45,16 +45,52 @@ export const fetchPlaceholders = async ({
   sheet,
   placeholderRequest,
   placeholderPath,
+  updateMph = true,
 }) => {
   const path = placeholderPath || getPlaceholdersPath(config, sheet);
 
-  fetchedPlaceholders[path] ||= fetchPlaceholder(path, placeholderRequest);
+  fetchedPlaceholders[path] ||= fetchPlaceholder(path, placeholderRequest, { updateMph });
 
   return fetchedPlaceholders[path];
 };
 
 function keyToStr(key) {
   return key.replaceAll('-', ' ');
+}
+
+const isGeoKey = (key) => key.endsWith('-geo');
+const stripGeo = (key) => key.slice(0, -4);
+
+let geoPlaceholderPromise;
+export function resetGeoPlaceholderCache() { geoPlaceholderPromise = undefined; }
+
+async function getGeoPlaceholders(sheet) {
+  if (!lingoActive()) return null;
+  const geoPrefix = await getMepLingoPrefix();
+  if (!geoPrefix) return null;
+
+  const siteConfig = getConfig();
+  const { origin } = window.location;
+  const contentRoot = siteConfig.contentRoot ?? '';
+  const geoContentRoot = `${origin}${geoPrefix}${contentRoot}`;
+  const geoConfig = { locale: { contentRoot: geoContentRoot } };
+
+  const root = `${geoContentRoot}/placeholders`;
+  const paths = [`${root}.json`];
+  if (siteConfig.env?.name !== 'prod'
+    && getMetadata('placeholders-stage') === 'on') paths.push(`${root}-stage.json`);
+
+  const placeholderRequest = Promise.all(
+    paths.map((path) => customFetch({ resource: path, withCacheRules: true }).catch(() => ({}))),
+  );
+
+  return fetchPlaceholders({
+    config: geoConfig,
+    sheet,
+    placeholderRequest,
+    placeholderPath: paths[0],
+    updateMph: false,
+  }).catch(() => ({}));
 }
 
 async function getPlaceholder(key, config, sheet) {
@@ -92,6 +128,9 @@ async function getPlaceholder(key, config, sheet) {
   };
 
   if (config.placeholders?.[key]) return config.placeholders[key];
+
+  if (isGeoKey(key)) return getPlaceholder(stripGeo(key), config, sheet);
+
   let placeholders;
 
   if (geoLocDisabled === 'on') {
@@ -108,6 +147,14 @@ async function getPlaceholder(key, config, sheet) {
   }
 
   return keyToStr(key);
+}
+
+async function resolveGeoKey(key, config, sheet) {
+  const baseKey = stripGeo(key);
+  geoPlaceholderPromise ??= getGeoPlaceholders(sheet);
+  const geoPlaceholders = await geoPlaceholderPromise;
+  if (typeof geoPlaceholders?.[baseKey] === 'string') return geoPlaceholders[baseKey];
+  return getPlaceholder(baseKey, config, sheet);
 }
 
 export async function replaceKey(key, config, sheet = 'default') {
@@ -129,6 +176,11 @@ export async function replaceKeyArray(keys, config, sheet = 'default') {
   return placeholders;
 }
 
+function isTelContext(text, matchIndex) {
+  const preceding = text.slice(Math.max(0, matchIndex - 10), matchIndex);
+  return /(?:^|["'=\s])tel:/i.test(preceding);
+}
+
 export async function replaceText(
   text,
   config,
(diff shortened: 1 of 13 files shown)


libs/features/placeholders.js
```diff
@@ -57,6 +57,56 @@ function keyToStr(key) {
...
+async function getGeoPlaceholders(config, sheet) {
+  // Dynamic to avoid circular dep with utils.js that hangs WTR test runner
+  const { lingoActive, getGeoLocalePrefix } = await import('../utils/utils.js');
+  if (!lingoActive()) return null;
+  const geoPrefix = await getGeoLocalePrefix();
+  if (!geoPrefix) return null;
+
+  const siteConfig = getConfig();
+  let geoOrigin = window.location.origin;
+
```
JasonHowellSlavin: nit: newline probably not needed

libs/features/placeholders.js
```diff
@@ -57,6 +57,56 @@ function keyToStr(key) {
...
+  const placeholderRequest = Promise.all(
+    paths.map((path) => customFetch({ resource: path, withCacheRules: true }).catch(() => ({}))),
+  );
+
+  return fetchPlaceholders({
+    config: geoConfig,
+    sheet,
+    placeholderRequest,
+    placeholderPath: paths[0],
+  }).catch(() => ({}));
```
JasonHowellSlavin: does it make sense to do any logging in the catch here?

sukamat (approved):

sukamat:
@markpadbe Thanks for getting this in. LMK once the PR is in Stage. I will clean up the test placeholders from the sheet that we added for validation.

libs/features/placeholders.js
```diff
@@ -57,6 +57,51 @@ function keyToStr(key) {
   return key.replaceAll('-', ' ');
 }
 
+const isGeoIpKey = (key) => key.endsWith('-geo-ip');
+const PH_RE = /{{(.*?)}}|%7B%7B(.*?)%7D%7D/g;
+
+async function getGeoPlaceholders(config, sheet) {
+  // Dynamic to avoid circular dep with utils.js that hangs WTR test runner
+  const { lingoActive, getGeoLocalePrefix, getPlaceholderPaths } = await import('../utils/utils.js');
```
vhargrave: this import is not necessary right since we're already importing utils.js. Can we move all this stuff to the static import?

libs/features/placeholders.js
```diff
@@ -57,6 +57,51 @@ function keyToStr(key) {
   return key.replaceAll('-', ' ');
 }
 
+const isGeoIpKey = (key) => key.endsWith('-geo-ip');
+const PH_RE = /{{(.*?)}}|%7B%7B(.*?)%7D%7D/g;
```
vhargrave: can we rename to be a little clearer that this is a phone number regex?

vhargrave (commented):
found two small things then it's good from my end - nice!

libs/features/placeholders.js
markpadbe: You're right that the dependency edge already exists, so the original "circular dep" comment was misleading. I've reworded it.

Keeping it dynamic for an empirical reason though: moving these into the static import above makes the WTR suite ~3.7× slower (42s → 157s) and pushes ~11 tests past testsFinishTimeout. Same code paths, just static vs. dynamic — reproducible across runs.

Ruled out:

- Module graph change: you're right, the placeholders.js → utils.js edge already exists from the 3-symbol static import; adding more bindings doesn't add an edge.
- The ?t=${timestamp} cache-busting re-imports of utils.js in utils.test.js (used in the lingo test blocks to force a fresh module instance): verified by replacing them with an in-place resetLingoState() helper that doesn't re-evaluate the module. Slowdown persists, so the cache-busting pattern isn't the cause.
- The new symbols being used: slowdown happens even when they're imported but never referenced.
Mechanism not yet isolated. Worth a follow-up to chase upstream and revert.

Cost of the workaround is essentially zero at runtime: utils.js is already in the browser's module cache by the time this runs (the static import at the top of the file pulled it in), so the await resolves synchronously without a real async hop.

libs/features/placeholders.js
markpadbe: Happy to rename to PLACEHOLDER_REGEX. PH_RE was a quick abbreviation, the longer name is clearer. (FYI the geo-ip scope is no longer restricted to phone numbers, so this regex really is "the placeholder regex" in the general sense, not anything number-specific.)

sukamat: What @markpadbe said. The update now basically introduces geo-ip placeholder functionality for any placeholder and not just phone-numbers. It relies on the suffix `-geo-ip` set on a placeholder.

libs/features/placeholders.js
sukamat: If the fix for this specific issue is time consuming and the current implementation does-no-harm, could we create a follow-up ticket to investigate? Mainly because some of the content updates for DC, CC, Genuine is pending on this feature to be on Stage. 

WDYT @markpadbe @vhargrave ?

vhargrave: @sunilkamat I wouldn't have felt comfortable approving tbh, because it was really weird we couldn't make the "small" change without things breaking. For me that was a potential sign that we were introducing a potential hidden bug. 
I've spent today looking at this though and have gotten to the bottom of it. @markpadbe i went ahead and just committed the changes. Hope that's ok. 

Here's the writeup : 

# WTR Suite Crash: Root Cause and Fix

The WTR crash was never about coverage instrumentation, circular dependencies, concurrency, or any of the other hypotheses from earlier sessions. It was a **one-character bug in the custom test reporter**.

## The bug

`web-test-runner.config.mjs` line 10:

```js
session.testResults.tests.forEach((test) => {
```

`testResults` is `undefined` when a test session errors out before any tests run (e.g., a module fails to load). This line throws a `TypeError`, which becomes an **uncaught exception** in the Node process. WTR's `uncaughtException` handler in `startTestRunner.js` calls `runner.stop()`, which kills the entire test suite immediately — regardless of how many files have already passed.

The fix: `session.testResults?.tests?.forEach(...)`.

## Why it only appeared with the static import change

`placeholders.js` was changed from:

```js
const { lingoActive, getGeoLocalePrefix, getPlaceholderPaths } = await import('../utils/utils.js');
```

to a static import at the top of the file. This means those names must be resolvable at **module load time**, not just at runtime.

Several HTML test files (`ost.test.html`, `upgrade.test.html`, `merch-card.polyfill.html`) use **import maps** to redirect `utils.js` to lightweight mock files. Those mocks didn't export `lingoActive`, `getGeoLocalePrefix`, or `getPlaceholderPaths`. With the dynamic import, this didn't matter — `getGeoPlaceholders` was never called in those tests. With the static import, the browser throws a `SyntaxError` at module resolution time, the session finishes with no test results, and the reporter crashes the suite.

## The full causal chain

1. Static import adds three names to the `placeholders.js` → `utils.js` import
2. Test HTML files redirect `utils.js` to mocks via import map
3. Mocks don't export the new names → browser `SyntaxError` during module resolution
4. Test session finishes with `testResults = undefined`
5. Custom reporter accesses `session.testResults.tests` → `TypeError`
6. `TypeError` hits WTR's `uncaughtException` handler → `runner.stop()` → process exits at ~120/255 files
7. Exit code 1 even with 0 reported test failures

## What was fixed

| File | Change |
|---|---|
| `web-test-runner.config.mjs` | `?.` null guard on `testResults.tests` |
| `libs/features/placeholders.js` | Static imports replace dynamic import workaround |
| `test/blocks/ost/mocks/ost-utils.js` | Added `lingoActive`, `getGeoLocalePrefix`, `getPlaceholderPaths` stubs |
| `test/blocks/merch/mocks/embed-utils.js` | Same three stubs |

The reporter fix is the important one — it's a latent bug that would crash the suite any time *any* test file fails to load its modules. The mock fixes are just keeping those files in sync with what `utils.js` now exports.

vhargrave (approved):

libs/features/placeholders.js
markpadbe: thank you!

spadmasa:
Validation done on the pr on below url

https://main--da-cc--adobecom.aem.page/fr/drafts/prats/frphonenumbertest?akamaiLocale=lu&milolibs=phonenumber-ip
image

https://main--da-cc--adobecom.aem.page/fr/drafts/souj/test-fr-phone-number?akamaiLocale=lu&milolibs=phonenumber-ip
image
https://main--da-cc--adobecom.aem.page/fr/drafts/souj/test-fr-phone-number?akamaiLocale=be&milolibs=phonenumber-ip
image

PR #5852, opened 2026-04-24
[MWPW-193332] Skills for Figma block -> DA -> Milo Block

overmyheadandbody:
This is a first stab at getting some shared Claude Code skills into Milo. They are currently aimed specifically at the Site Redesign initiative, but they contain general Milo information that will be separated at a later point, so that any future skill can make use of it. Each has a `README.md` file, so all of the details can be found there; this is also a pattern we could continue using, to ensure everyone gets a quick overview on what a skill does and what its expected input is. The animation skill is still work in progress, but thought I'd push now and get back with improvements in the future.

Resolves: [MWPW-193332](https://jira.corp.adobe.com/browse/MWPW-193332)


mokimo (approved):

narcis-radu (approved):

mokimo:
Let's merge the spike and then iterate as needed.

sharg1:
❤️ this @overmyheadandbody 👏 !!

PR #5857, opened 2026-04-27
MWPW-193754 Stage aware query-index loading

vhargrave:
Resolves: [MWPW-193754](https://jira.corp.adobe.com/browse/MWPW-193754)

## Summary
- Cross-site lingo query-indexes (from `lingo-site-mapping.json`) now load from the matching stage host instead of always hitting prod live
- On `business.stage.adobe.com`, da-cc's query-index now loads from `www.stage.adobe.com/.../query-index-preview.json` instead of `www.adobe.com/.../query-index.json`
- There's a new config for this in `lingo-site-mapping.json` named `stageHost` — sites without it fall back to prod (no behaviour change) - https://main--federal--adobecom.aem.page/federal/assets/data/lingo-site-mapping.json

## Why not `.aem.page` / `.aem.live`?
Cross-origin fetches between `*.aem.page` domains require authentication. The auth token is in an HTTP-only cookie that JavaScript cannot access, so the fetch would 401 and `processQueryIndexMap` would silently return empty paths — worse than the current prod fallback. Stage hosts share the same auth layer so cross-origin fetches work without extra tokens.

## Mapping config needed (federal repo)
Add a `stageHost` column to the `site-query-index-map` sheet in `lingo-site-mapping.json`. I've added but only previewed so far. 

**Test URLs:**
- Before: https://business.stage.adobe.com/fr/?akamaiLocale=CA
- After: https://business.stage.adobe.com/fr/?milolibs=vhargrave-env-aware-query-index&akamaiLocale=CA

- Before: https://www.stage.adobe.com/fr/acrobat/acrobat-pro.html?akamaiLocale=ca
- After: https://www.stage.adobe.com/fr/acrobat/acrobat-pro.html?akamaiLocale=ca&milolibs=vhargrave-env-aware-query-index

## How to test
- [ ] On both test links: DevTools Network shows cross-site Query Index fetched from `*.stage.adobe.com` with `-preview` suffix, and no longer from just adobe.com
- [ ] On `*.aem.page`: cross-site QI still loads from prod (graceful fallback)

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -678,24 +678,19 @@ function processQueryIndexMap(link, domain) {
 }
 const getDomainLingo = (path) => path?.split('/*')[0];
 
-const ENV_AEM_HOST_REGEX = new RegExp(`\\.${SLD}\\.(page|live)$`);
-const ENV_STAGE_HOST_REGEX = /\.stage\.adobe\.com$/;
+const ENV_STAGE_HOST_RE = /\.stage\.adobe\.com$/;
 
 export function resolveCrossSiteIndex(
-  { queryIndexWebPath, stageHost, aemRepo },
+  { queryIndexWebPath, stageHost },
   prefix,
   suffix,
   currentHost,
 ) {
   const prodHost = getDomainLingo(queryIndexWebPath);
-  const aemMatch = currentHost.match(ENV_AEM_HOST_REGEX);
   let host = prodHost;
   let sfx = '';
 
-  if (aemMatch && aemRepo) {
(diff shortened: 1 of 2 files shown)


sukamat (approved):
Bummer that this will require Akamai FQDN but this change was necessary to reduce confusion on link transformation

raga-adbe-gh (approved):

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details [MWPW-193754](https://jira.corp.adobe.com/browse/MWPW-193754)

sukamat:
@vhargrave @NadiiaSokolova If the validation is complete, could we set the required labels on this so that it gets merged to Stage?

PR #5859, opened 2026-04-27
MWPW-193845 [MEP] Add Analytics Listener for Spectra Project Reach

ivanvatadobe:
This is a Project Reach enhancement. 

MEP currently allows experience targeting based on user Line Of Business (LOB), which relied on matching MEP manifest experience column name (example: lob-cc) with the LOB returned by Spectra API on page hit (example: cc). 

Spectra API has been updated to return the LOB **name**, AND the attributed confidence **score**. This enhancement takes both the name and score values returned from the API and records them into a custom analytic object on the page. 

**Features/Notes:** 
* `lob.js` now adds the name and score value of the API response to a custom tracking object: `alloy_all.data._adobe_corpnew.event.custom`
* We have repurposed functions from email-collection block utils `awaitWindowProperty` to continue to check for `alloy_all` to exist before adding the tracking values to the custom tracking object 
* All code changes should be exclusive to `/mep/addons/lob.js` and will only fire when a page has metadata to call the function on page hit 

**Notes on code implementation**
* The API response returns two values (`spectraLob` and `spectraScore`), but we are only returning `spectraLob` as the variable to be passed onto `personalization.js`'s `init`. There is no need to have the score passed to/interact with `personalization.js`.
* We are waiting for 2 dependencies to be available when `lob.js` runs: `alloy_all` object on the browser and the Spectra API response. 
* Checking for `alloy_all` will always happen _after_ the API response. 
* If the API response is received, but `alloy_all` is not yet available, then `lob.js` will still return the `spectraLob` value without waiting for `addAlloyTracking`. However, timeouts in the form of promises are used to allow `addAlloyTracking` to fire later in execution even after `lob.js` has reached its return value. 

**Resolves**: [MWPW-193845](https://jira.corp.adobe.com/browse/MWPW-193845)

**QA instructions/Test Urls:**
* The LOB feature can be enabled by adding `lob=on` url param. 
* QA/testing requires a local override of `libs/features/mep/addons/lob.js` (copy PR version of `lob.js` onto a local version and override the content from sources menu with your local file).

Steps: 
1. Go to https://www.adobe.com/?lob=on 
2. Open console and check for the following object: `alloy_all.data._adobe_corpnew.event.custom`
3. Confirm: the custom object has an entry with _propertyName_ '**spectraLob**', and an associated LOB for _propertyValue_ (ex: 'cc', 'dc', 'smb') 
4. Confirm: custom object has an entry with _propertyName_ '**spectraScore**', and an associated _number_ for _propertyValue_ (ex: 0.528565269468545) 
5. Observe other pages that are not homepage for the same 

**Known issue:**
`getECID = getCookie(AMCV_COOKIE);` sometimes fails when loading the page for the first time in incognito mode. The code returns a value of `false`, which is intended. This is likely an edge case, but will be investigated.

Make sure to disable local overrides when done testing. 

Example analytics object `alloy_all.data._adobe_corpnew.event.custom`:
image


libs/features/mep/addons/lob.js
```diff
@@ -19,18 +19,65 @@ export async function getSpectraLOB(lastVisitedPage) {
       body: null,
     });
     const content = await rawResponse.json();
-    return content.modelLineOfBusiness?.toLowerCase();
+    content.modelLineOfBusiness = content?.modelLineOfBusiness.toLowerCase();
+    return content;
+    // return content.modelLineOfBusiness;
```
markpadbe: can we remove this comment?

libs/features/mep/addons/lob.js
```diff
@@ -19,18 +19,65 @@ export async function getSpectraLOB(lastVisitedPage) {
...
+        ?.push({ propertyName: spectraValues[key], propertyValue: value });
+    });
+  });
+}
+
 export default async function init(enablement) {
   if (enablement !== true) return enablement;
-  if (window.location.hostname.includes('.aem.')) return 'cc';
-  const consentCookieValue = getCookie('OptanonConsent');
-  if (consentCookieValue?.includes('C0002:0')) return 'cc';
```
markpadbe, line 33: is this changing how consent is handled? Also, this might be triggering some unit test failures.

libs/features/mep/addons/lob.js
```diff
@@ -19,18 +19,65 @@ export async function getSpectraLOB(lastVisitedPage) {
       body: null,
     });
     const content = await rawResponse.json();
-    return content.modelLineOfBusiness?.toLowerCase();
+    content.modelLineOfBusiness = content?.modelLineOfBusiness.toLowerCase();
```
markpadbe: nit: ?. should probably guard modelLineOfBusiness, not content, though it looks the try/catch handles either case

libs/features/mep/addons/lob.js
```diff
@@ -19,18 +19,65 @@ export async function getSpectraLOB(lastVisitedPage) {
...
+
 export default async function init(enablement) {
   if (enablement !== true) return enablement;
-  if (window.location.hostname.includes('.aem.')) return 'cc';
-  const consentCookieValue = getCookie('OptanonConsent');
-  if (consentCookieValue?.includes('C0002:0')) return 'cc';
   const lobValue = await getSpectraLOB(document.referrer);
-  return lobValue;
+  addAlloyTracking(lobValue);
+  return lobValue.modelLineOfBusiness;
```
markpadbe: if getSpectraLOB returns false, this throws and breaks MEP initialization(personalization.js:1640).                       
                                   
```suggestion
  return lobValue?.modelLineOfBusiness;
```

libs/features/mep/addons/lob.js
```diff
@@ -19,18 +19,65 @@ export async function getSpectraLOB(lastVisitedPage) {
...
+      modelLineOfBusiness: 'spectraLob',
+      modelScore: 'spectraScore',
+      experienceSelected: 'spectraExperience',
+    };
+    const dataObjString = 'data._adobe_corpnew.event.custom';
+    const existingCustomData = window.alloy_all?.get(window.alloy_all, dataObjString) || [];
+    window.alloy_all?.set(window.alloy_all, dataObjString, existingCustomData);
+    Object.entries(lobObject).forEach(([key, value]) => {
+      if (!spectraValues[key]) return;
+      window.alloy_all?.data?._adobe_corpnew?.event?.custom
```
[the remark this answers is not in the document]
markpadbe: you can disable this on line 72: // eslint-disable-next-line no-underscore-dangle

markpadbe (changes requested):

libs/features/mep/addons/lob.js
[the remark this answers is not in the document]
ivanvatadobe: done

markpadbe (approved):

jpratt2 (approved):
I can approve based on a QA review of functionality

Image

PR #5860, opened 2026-04-28
[MWPW-193628]: [Accessibility] Fixed Content is lost at 200% zoom.

hkuraware:
- Fix ribbon notification CTAs being cut off at 200% zoom (e.g. “Visit the Google Play Store”).
- Improves WCAG 1.4.4 Resize text (AA) by allowing layout to reflow/wrap instead of clipping.
- Adjusts ribbon space-between flex sizing/wrapping and prevents the title (e.g. “Lightroom”) from wrapping per-letter at higher zoom.
- Test: Chrome 1366×768 at 100% vs 200% zoom; confirm both CTAs are fully visible/clickable.

Resolves: [MWPW-193628](https://jira.corp.adobe.com/browse/MWPW-193628)

**Test URLs:**
- Before: https://main--da-cc--adobecom.aem.page/drafts/himani/sticky-banner-notification?martech=off
- After: https://main--da-cc--adobecom.aem.page/drafts/himani/sticky-banner-notification?milolibs=mwpw-193628-a11y-lightroom--milo--hkuraware&martech=off

**Dev validation:**

Before(1366px by 768px zoom 200% ):- 
before-fix

After(1366px by 768px zoom 200% ):- 
after-fix
after-fix-dimensions

### DIFF
--- a/libs/blocks/notification/notification.css
+++ b/libs/blocks/notification/notification.css
@@ -224,7 +224,6 @@
   flex-direction: column;
   justify-content: center;
   flex: 1 1 0;
-  /* Prevent the title from shrinking so far it wraps per-letter at higher zoom levels */
   min-inline-size: min-content;
   max-inline-size: 100%;
 }
--- a/libs/blocks/notification/notification.css
+++ b/libs/blocks/notification/notification.css
@@ -38,6 +38,7 @@
 
 .notification.ribbon {
   min-block-size: var(--min-block-size-ribbon);
+  overflow: visible;
 }
 
 .dark .notification,
@@ -211,7 +212,9 @@
 }
 
 .notification.ribbon.space-between .foreground .text {
-  flex-wrap: nowrap;
+  flex-wrap: wrap;
+  row-gap: var(--spacing-xs);
+  column-gap: var(--spacing-s);
   inline-size: 100%;
 }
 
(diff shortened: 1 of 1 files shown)


libs/blocks/notification/notification.css
nkthakur48: Could you check if removing `margin-inline-end: var(--spacing-s)` from `.copy-wrap` since `column-gap` on the parent now owns that horizontal spacing

drashti1712 (approved):

libs/blocks/notification/notification.css
```diff
@@ -38,6 +38,7 @@
 
 .notification.ribbon {
   min-block-size: var(--min-block-size-ribbon);
+  overflow: visible;
```
nkthakur48, line 41: I noticed that this overrides `"overflow": none` from the `.notification` base class. Could you help clarify why this was done? It's not immediately obvious from the code, so a bit of context would be really helpful!

nkthakur48 (commented):
@hkuraware Could you please have a look at review comments?

libs/blocks/notification/notification.css
hkuraware: Hi @nkthakur48 
`.notification` sets `overflow: hidden`, which is fine for the default block but it clips ribbon content when we reflow icon / title / CTAs at smaller effective widths (e.g. zoom / long labels). `.notification.ribbon` `{ overflow: visible; }` scopes the override to ribbons only so reflowed controls stay fully visible instead of being cut off at the block edge (WCAG 1.4.4 resize / reflow).

hkuraware:
Closing this PR since design change approval is needed before implementation can move forward.

PR #5869, opened 2026-04-29
[MWPW-191877] - SR Firefox animations

DKos95:
PR adds support for firefox animations until they implement scroll driven animations.
Plans to implement polyfil were abandoned because of bad performance in combination with lenis, while this approach gives us better performance and requires less code.
When user resizes on firefox animations are no longer active.
Performance was measured to be ~120fps on firefox

Resolves: [MWPW-191877](https://jira.corp.adobe.com/browse/MWPW-191877)

PSI Test URLs:
After: https://firefox-animations--milo--adobecom.aem.page/?martech=off

Firefox Test URLs:
Before: https://load-c2-styles--upp--adobecom.aem.live/homepage/drafts/ramuntea/redesign-demo
After: https://load-c2-styles--upp--adobecom.aem.live/homepage/drafts/ramuntea/redesign-demo?milolibs=firefox-animations

### DIFF
--- a/libs/c2/scroll-animations.js
+++ b/libs/c2/scroll-animations.js
@@ -1,0 +1,577 @@
+const vh = window.innerHeight;
+const STAGGER_RE = /parallax-stagger-(ltr|rtl)/;
+const STAGGER_SELECTOR = '[class*="parallax-stagger-ltr"],'
+  + '[class*="parallax-stagger-rtl"]';
+
+const scrollTasks = [];
+const cleanupTasks = [];
+
+function getDocTop(el) {
+  let top = el.offsetTop;
+  let { offsetParent } = el;
+  while (offsetParent) {
+    top += offsetParent.offsetTop;
+    offsetParent = offsetParent.offsetParent;
+  }
+  return top;
+}
+
+function getScrollMetrics(scroll, elHeight, docTop, insetTop = 0, insetBottom = 0) {
+  const dist = (scroll + vh * (1 - insetBottom)) - docTop;
+  const total = elHeight + vh * (1 - insetTop - insetBottom);
+  return { elHeight, dist, total };
+}
+
+function viewRange({ elHeight, dist, total }, sType, sPct, eType, ePct) {
+  const s = sType === 'entry' ? elHeight * sPct : total * sPct;
+  const e = eType === 'entry' ? elHeight * ePct : total * ePct;
+  return Math.max(0, Math.min(1, (dist - s) / (e - s)));
+}
+
+function sectionHasStyle(section, style) {
+  if (section.classList.contains(style) || section.querySelector(`.${style}`)) return true;
+  const smBlock = section.querySelector('.section-metadata');
+  if (!smBlock) return false;
+  const rows = [...smBlock.children];
+  const styleRow = rows.find(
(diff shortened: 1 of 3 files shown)


overmyheadandbody (approved):
Looks good overall, I'm hoping Firefox enables support outside of the feature flag soon, so we can remove the additional logic.

I just have one mention regarding the garage door, it looks to me like the heading and body text get "stuck" for quite a long time before reaching the intended spacing between them, can that be tweaked a bit?

Image

DKos95:
> Looks good overall, I'm hoping Firefox enables support outside of the feature flag soon, so we can remove the additional logic.
> 
> I just have one mention regarding the garage door, it looks to me like the heading and body text get "stuck" for quite a long time before reaching the intended spacing between them, can that be tweaked a bit?
> 
> Image

Thanks for the review, I've updated it so that the animation starts and finishes a bit earlier, hopefully this is an improvement.

libs/c2/scroll-animations.js
```diff
@@ -0,0 +1,577 @@
...
+    });
+  };
+
+  document.querySelectorAll('.elastic-carousel-container').forEach(setupContainer);
+
+  observeForNew('.elastic-carousel-container', setupContainer);
+}
+
+function initCarouselC2() {
+  const isRtl = document.documentElement.dir === 'rtl';
```
narcis-radu: This can be moved to the top, I don't see the value changing

libs/c2/scroll-animations.js
```diff
@@ -0,0 +1,577 @@
+const vh = window.innerHeight;
```
narcis-radu, line 1: Would this cause issues when resizing?

libs/c2/scroll-animations.js
```diff
@@ -0,0 +1,577 @@
...
+    + ' animation: none !important; }');
+
+  let growFrom; let revealFrom; let coverEnd; let growStart;
+
+  const updateBreakpoints = () => {
+    const w = window.innerWidth;
+    const cvh = window.innerHeight;
+    growFrom = -0.5 * cvh;
+    revealFrom = 0.2 * cvh;
+    if (w >= 1280) { growFrom = -1.1 * cvh; revealFrom = 0.4 * cvh; }
```
narcis-radu: Can we store the breakpoints in a constant?

libs/c2/scroll-animations.js
DKos95: As I said here https://github.com/adobecom/milo/pull/5830#issuecomment-4288361048, it would be very difficult to implement animations + resizing especially if we take into account all the positions the user could have scrolled to and started resizing, this would especially be tough with parallax animations like garage door as it's not just a matter of re running the logic we have on resize, and with wave two animations also coming in this would quickly become very tough to maintain (the file is 600lines already) and given the amount of people that a. use firefox and b. resize their window is ~0.03% to 0.15% of users not sure this would be worth investing time in. Those users that usually resize their screen are visually impared users or screenreader users and for that situation we just reset all animations so that the screenreader can go through everything so we comply with a11y.

narcis-radu (approved):

zagi25 (approved):

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details [MWPW-191877](https://jira.corp.adobe.com/browse/MWPW-191877)

PR #5845, opened 2026-04-22
MWPW-193436 Fix: MEP-gnav stacking and mep-body-scroll issues on c2 pages 

markpadbe:
**Summary:** 

The federal gnav promotes its nav to the browser top layer via popover="manual", causing the MEP overlay to render behind it regardless of z-index. Promote the overlay to the top layer using the same Popover API.
On C2 pages, Lenis smooth-scroll intercepts wheel events inside the MEP popup body, preventing native overflow scroll. Patch Lenis's prevent callback to exempt nodes inside the overlay.

**Updates:**

- preview.js: Create overlay with popover="manual" and call .showPopover() to promote it to the top layer. Patch window.lenis.options.prevent to allow scrolling inside .mep-preview-overlay. Call .hidePopover() on close before DOM removal.

- preview.js: Guard overlayEl?.remove() against a null overlay on close. Restore window.lenis.options.prevent to its original value when the pill is dismissed. Use showPopover?.() for safe degradation on browsers without the Popover API.

- preview.css: Replace bottom/left with inset shorthand and reset UA popover defaults (border, padding, margin, background, overflow).

Detailed Description:
................................
URL: https://main--upp--adobecom.aem.page/homepage/index-loggedout?mep=%2Fhomepage%2Ffragments%2Fmep%2Fhp-04-22-26-hp-ff-unlimited-updates-psnlz.json--all---%2Fupp-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1165v2%2Face1165v2.json--target-var1---%2Fhomepage%2Ffragments%2Floggedout%2Ftests%2F2026%2Fq2%2Fsite-redesign%2Fsite-redesign-v2-4-14-update.json--target-var1
................................
Steps to Reproduce:
1. open mep button on the Before and after links below
2. expand collapsible manifest content
3. Observe the overlap with nav bar and also attempt to scroll the mep button content
Expected Results: the mep button appears above the nav and mep button content is scrollable.
................................
Actual Results: Nav is above the mep button and mep content isn't scrollable.

Resolves: [MWPW-193436](https://jira.corp.adobe.com/browse/MWPW-193436)

**Test URLs:**
[...]

PR #5882, opened 2026-04-30
MWPW-194181[MEP] Fix flash in preview

markpadbe:
### Summary                                                                        
                                                                                 
  Regression from [MWPW-193436](https://jira.corp.adobe.com/browse/MWPW-193436) / #5845, which fixed the MEP overlay being obscured
   by the federal gnav on C2 pages. That fix used the Popover API
  (popover="manual") to promote the overlay into the browser top layer — but     
  called showPopover() immediately on creation, making the overlay visible in the
   top layer as soon as it was appended to the DOM.                              
                                                                                 
  Previously the overlay was created with style="display:none" and stayed        
  invisible until the user opened the pill. With the Popover API change, the
  overlay is shown unconditionally on init — before the pill's mep-hidden CSS has
   a chance to hide its contents — causing a brief flash.
                                                                               
  The flash appears to happen on GNAV hover because decoratePreviewMode() runs in
   Phase D (~3s after LCP). By that point the user is naturally interacting with
  the page, so the flash coincides with cursor movement and looks                
  hover-triggered. It isn't.
                                                                               
###   Fix

  Defer showPopover() to the badge click handler, called only when the pill is   
  opening (not toggling closed). The overlay stays hidden in the popover state
  until the user first opens it, at which point it's promoted to the top layer as
   intended.      
                                                                                                   
###   Scope                                                                          
  
  Only affects internal/preview users — createPreviewPill() only runs when MEP   
  preview mode is active.
                                                                                 

###   Steps to QA                                                     
  1. Open the Before and After test urls below
  2. wait 3-5 seconds after page load and watch for a flash (described in more detail in ticket)
  **Expected Results**: No modal or overlay flashes on page load or during any
  interaction.                                 
  **Actual Results:** A modal briefly appears and disappears approximately 3 seconds 
  after page load.                                   
                         
Resolves: [MWPW-194181](https://jira.corp.adobe.com/browse/MWPW-194181)

**Test URLs:**
- Before: https://business.stage.adobe.com
- After: https://business.stage.adobe.com/?milolibs=mep-fixflash
- PSI-chech: https://mep-fixflash--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/libs/features/personalization/preview.js
+++ b/libs/features/personalization/preview.js
@@ -132,9 +132,10 @@ function addDividers(node, selector) {
   });
 }
 
-function addPillEventListeners(div, onClose) {
+function addPillEventListeners(div, overlay, onClose) {
   div.querySelector('.mep-manifest.mep-badge').addEventListener('click', () => {
-    div.classList.toggle('mep-hidden');
+    const isHidden = div.classList.toggle('mep-hidden');
+    if (!isHidden) overlay?.showPopover?.();
   });
   div.querySelector('.mep-close').addEventListener('click', () => {
     const overlayEl = document.querySelector('.mep-preview-overlay');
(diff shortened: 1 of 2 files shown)


AdobeLinhart (approved):

jpratt2:
fixed. There is no flash coinciding with loading the green MEP button

jpratt2 (approved):

sukamat (approved):

PR #5891, opened 2026-05-05
[MWPW-194479] Fix merge-to-main workflow to trigger on label events from stage to main

mokimo:
### Description
The `pull_request_target` labeled trigger was already defined in the workflow but the job condition checked for `pull_request` event name instead of `pull_request_target`, so label events never actually ran the job.

This fix corrects the condition to match the trigger, improving workflow stability and execution time by allowing the merge to happen immediately when a label is applied rather than waiting for the scheduled run.

Resolves: [MWPW-194479](https://jira.corp.adobe.com/browse/MWPW-194479)

### DIFF
--- a/.github/workflows/merge-to-main.yaml
+++ b/.github/workflows/merge-to-main.yaml
@@ -17,7 +17,7 @@ jobs:
     environment: milo_pr_merge
     # Run this when manually triggered or on a schedule
     # Otherwise run this only on PRs that are merged from stage to main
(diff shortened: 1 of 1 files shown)


overmyheadandbody (approved):

zagi25 (approved):

DKos95 (approved):

mokimo:
admin merging this given the approvals/this being zero-impact & the tests are not connected and fail on stage as well.

PR #5895, opened 2026-05-05
[MWPW-194269] Add six-up layout support

overmyheadandbody:
This adds support for a six-up layout via section metadata. It also adds all the required properties to ensure six-up layouts can use the staggered animation effect. This is the first step towards having the new block suggested in Figma with 6 items on the same row.

**Test URLs:**
- Before: https://site-redesign-foundation--milo--adobecom.aem.page/drafts/ramuntea/six-up-test
- After: https://six-up-support--milo--adobecom.aem.page/drafts/ramuntea/six-up-test

### DIFF
--- a/libs/c2/blocks/news/news.js
+++ b/libs/c2/blocks/news/news.js
@@ -31,7 +31,7 @@ export default async function init(el) {
   const [head, ...tail] = rows;
   formatHeader(head);
   rows = tail;
-  const upsMap = { 2: 'two-up', 3: 'three-up', 4: 'four-up' };
+  const upsMap = { 2: 'two-up', 3: 'three-up', 4: 'four-up', 6: 'six-up' };
   // TODO: Infer parallax class from authoring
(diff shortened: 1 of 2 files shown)


zagi25 (approved):

DKos95 (approved):

ivanvatadobe:
@overmyheadandbody 

this PR adds block code directly to /libs/c2/blocks. 

our goal for testing is to add block code to /libs/mep/ace1205/<block> for the test, and have the test page call the test block on the appropriate test page via MEP manifest. 

please let me know if there is an exception needed for this PR as it is directly changing the prod code.

ivanvatadobe (dismissed):
requesting that these changes be added to `/libs/mep/ace1205` folder instead of directly to `/libs/c2` folder for test purposes.

overmyheadandbody:
@ivanvatadobe - the majority of the code is in `styles.css`, which can't be added as part of that folder. We can move (or even disregard) the changes for news, since it's likely not useful at this point anyway.

PR #5899, opened 2026-05-05
[MWPW-193582] - Optimize brand-concierge floating button scroll and layout handling

DKos95:
Eliminates all layout-triggering reads from the floating button scroll handler by caching scrollHeight, offsetHeight, and getComputedStyle in a ResizeObserver — so the scroll handler only ever reads scrollY and innerHeight. Guards all DOM writes behind prev-value checks to skip no-op updates. Adds an RAF throttle with a scrollPending guard and marks the listener { passive: true }. Also narrows transition: all on the floating button hover states to only the properties actually animating (background, border-color), preventing the browser from tracking transitions on every CSS property.

Resolves: [MWPW-193582](https://jira.corp.adobe.com/browse/MWPW-193582)

PSI Test URLs:
After: https://perf-brand-concierge-scroll-optimization--milo--adobecom.aem.page/?martech=off

Test URLs:
Before: https://load-c2-styles--upp--adobecom.aem.live/homepage/drafts/ramuntea/redesign-demo?milolibs=site-redesign-foundation
After: https://load-c2-styles--upp--adobecom.aem.live/homepage/drafts/ramuntea/redesign-demo?milolibs=perf-brand-concierge-scroll-optimization

### DIFF
--- a/libs/c2/blocks/brand-concierge/brand-concierge.css
+++ b/libs/c2/blocks/brand-concierge/brand-concierge.css
@@ -386,12 +386,12 @@
 }
 
 .bc-floating-button:hover .bc-floating-button-container {
-  transition: all 0.3s ease;
+  transition: background 0.3s ease;
   background: #fff;
 }
 
 .bc-floating-button:hover::before {
-  transition: all 0.3s ease;
+  transition: border-color 0.3s ease;
   border-color: var(--bc-floating-button-border-color-hover);
 }
 
--- a/libs/c2/blocks/brand-concierge/brand-concierge.js
+++ b/libs/c2/blocks/brand-concierge/brand-concierge.js
@@ -487,52 +487,78 @@ function decorateFloatingButton(el) {
   const mainElement = document.querySelector('main');
   const gnavElement = document.querySelector('header.global-navigation');
 
-  const hideFloatingButton = () => {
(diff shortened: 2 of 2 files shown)


colloyd (commented):
@DKos95 So there are elements to this that have pending changes, particularly we have removed reliance on the presence of gnav. I appreciate the effort put in for this, but as the values in the handler have changed, certain elements of this fix are out of date. 

Finally, we do need whatever is done in c2 to be mirrored in c1. It may be best if I handle the optimizations, so I can account for all of the many variables at play for brand concierge.  I will use the code here as a basis for my own stab at optimization and would appreciate your review of that work when I'm ready.

DKos95:
> @DKos95 So there are elements to this that have pending changes, particularly we have removed reliance on the presence of gnav. I appreciate the effort put in for this, but as the values in the handler have changed, certain elements of this fix are out of date.
> 
> Finally, we do need whatever is done in c2 to be mirrored in c1. It may be best if I handle the optimizations, so I can account for all of the many variables at play for brand concierge. I will use the code here as a basis for my own stab at optimization and would appreciate your review of that work when I'm ready.

Thanks for the context and for taking this on yourself, makes sense given all the variables at play. I'll go ahead and close this PR.

PR #5900, opened 2026-05-05
[MWPW-192608] Fix race condition in bulk publish v2 authentication

mokimo:
### Description
Fix race condition in bulk publish v2 where the sidekick `status-fetched` event fires before listeners are registered, causing users to see "open sidekick and sign in" even when the sidekick is already open and logged in.

The fix extracts listener registration into an `addListeners` helper and immediately checks `sidekick.appStore?.status` after attaching listeners to handle the case where the event already fired.

Resolves: [MWPW-192608](https://jira.corp.adobe.com/browse/MWPW-192608)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.live/tools/bulk-publish
- After: https://MWPW-192608-bulk-publish-auth-race-fix--milo--mokimo.aem.live/tools/bulk-publish


mokimo:
closing because of upper-case letters in the branch..

PR #5901, opened 2026-05-05
[MWPW-192608] Fix race condition in bulk publish v2 authentication

mokimo:
### Description
Fix race condition in bulk publish v2 where the sidekick `status-fetched` event fires before listeners are registered, causing users to see "open sidekick and sign in" even when the sidekick is already open and logged in.

The fix extracts listener registration into an `addListeners` helper and immediately checks `sidekick.appStore?.status` after attaching listeners to handle the case where the event already fired.

Resolves: [MWPW-192608](https://jira.corp.adobe.com/browse/MWPW-192608)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.live/tools/bulk-publish
- After: https://mwpw-192608--milo--mokimo.aem.live/tools/bulk-publish

### DIFF
--- a/libs/blocks/bulk-publish-v2/services.js
+++ b/libs/blocks/bulk-publish-v2/services.js
@@ -79,16 +79,25 @@ const authenticate = async (tool = null) => {
   isPushedDown();
   const setUserV6 = (event) => { tool.user = setUserData(event?.detail?.data); };
   const setUserV7 = (event) => { tool.user = setUserData(event?.detail); };
+
+  const addListeners = (sidekick) => {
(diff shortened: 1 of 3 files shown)


DKos95 (approved):
Looks good to me, haven't been able to replicate the issue again.

SilviuLCF (approved):
Verified, testing details :https://jira.corp.adobe.com/browse/MWPW-192608

overmyheadandbody (approved):

robert-bogos (approved):

PR #5913, opened 2026-05-07
[SR] Bento layout foundation

rgclayton:
Creating the foundation of the bentos layout. 
* Currently uses `section-metadata` `masonry` grid for the more simple grid layouts. Makes it easy to quickly adjust each card in the section or rearrange.
* Uses `section-metadata` `layout` property to enable`bento` layout feature. 
* Bentos is a layout feature that currently uses the `explore-card` and then applies style changes to align with new designs.

**TO DO**
* Add mobile stacking feature

Resolves: 

- [MWPW-192884](https://jira.corp.adobe.com/browse/MWPW-192884)
- [MWPW-192886](https://jira.corp.adobe.com/browse/MWPW-192886)

**Test URLs:**
- Before: https://main--upp--adobecom.aem.page/homepage/drafts/rclayton/sr-bento?milolibs=site-redesign-foundation&martech=off
- After: https://main--upp--adobecom.aem.page/homepage/drafts/rclayton/sr-bento?milolibs=sr-bentos&martech=off

Notes:
[Offers](https://www.figma.com/design/q87sUm2fvForRQzxGum5wE/Offer-%E2%80%94-A.com?node-id=12494-35020&m=dev) and [product](https://www.figma.com/design/6FQTDUAttFbukvtskDAkPS/Product-%E2%80%94-A.com?node-id=7065-10746&m=dev) bento layouts are on the sample page.

### DIFF
--- a/libs/mep/ace1205/section-metadata/section-metadata.css
+++ b/libs/mep/ace1205/section-metadata/section-metadata.css
@@ -1,0 +1,296 @@
+.section-metadata {
+  display: none;
+}
+
+.section {
+  &.has-background {
+    background-color: unset;
+  }
+
+  &.divider {
+    border-bottom: 1px solid var(--s2a-color-gray-300);
+  }
+
+  &.vertical-center {
+    align-items: center;
+  }
+
+  &.rounded-corners,
+  &.rounded-corners-top,
+  &.rounded-corners-bottom {
+    border-radius: var(--s2a-border-radius-lg);
+    overflow: clip;
+    z-index: 3;
+    margin-block: calc(-1 * var(--s2a-border-radius-lg));
+  }
+
+  &.rounded-corners-top {
+    border-radius: var(--s2a-border-radius-lg) var(--s2a-border-radius-lg) 0 0;
+    margin-bottom: initial;
+  }
+
+  &.rounded-corners-bottom {
+    border-radius: 0 0 var(--s2a-border-radius-lg) var(--s2a-border-radius-lg);
+    margin-top: initial;
+  }
+
+  &.masonry-layout {
+    display: grid;
+    gap: var(--grid-gutter);
+    grid-template-columns: repeat(var(--grid-columns), 1fr);
+
+    .grid-full-width { grid-column: span var(--grid-columns); }
+    .grid-half-width { grid-column: span 3; }
+
+    .grid-span-1,
+    .grid-span-2,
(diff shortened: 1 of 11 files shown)


libs/mep/ace1205/section-metadata/section-metadata.css
```diff
@@ -0,0 +1,296 @@
...
+  &.masonry-responsive {
+    display: grid;
+  }
+
+  &.product-grid[class*='-up'] {
+    align-items: stretch;
+    grid-auto-rows: initial;
+  }
+
+  &.bento .rich-content.grid-full-width {
```
overmyheadandbody: I'm not sure we want block-specific styles in this more general section metadata. I see Rich Content, Explore Card, etc. code in here that might be better suited in the block files themselves. Otherwise we run the risk of overcrowding a core block, for which all of our pages will pay a price, even when they're not using the bento feature.

libs/mep/ace1205/section-metadata/section-metadata.css
```diff
@@ -0,0 +1,296 @@
...
+    margin-block-end: calc(var(--s2a-spacing-lg) - var(--grid-gutter));
+  }
+
+  &.bento .rich-content.grid-full-width ~ .rich-content.grid-full-width {
+    margin-inline-start: 0;
+    margin-block-end: 0;
+  }
+
+  &.bento:not(.container):not(.wide) { padding-inline: var(--s2a-spacing-lg);}
+  &.bento:not(.container):not(.wide)[class*='combo'] { padding-bottom: var(--s2a-spacing-lg);}
```
overmyheadandbody, line 69: Are there some other options here to avoid chained `:not` selectors? We've had maintenance issues in the past and it's usually a quick solve at this stage.

ivanvatadobe (approved):
Confirming that we're only adding block changes to ACE1205, and that we don't do anything different with utils for testing purposes. This is good.

libs/mep/ace1205/section-metadata/section-metadata.css
rgclayton: I'll take a look at some options.

libs/mep/ace1205/section-metadata/section-metadata.css
rgclayton: Looking into this.

PR #5915, opened 2026-05-08
 MWPW-194203 / MWPW-194204 [MEP] Highlight CaaS + M@S content in MEP preview

markpadbe:
### Summary                                                                                                                                                   

  Adds two "Highlight ..." modes to the MEP preview panel:                                                                                                        
   
  1. **Highlight CaaS** (MWPW-194203) — CaaS-backed card collections get an outline, an "Edit in CaaS Configurator" badge, and per-card market/lang pills with    
  click-to-source navigation.                                                                                                                                   
  2. **Highlight M@S** (MWPW-194204) — M@S surfaces (collections, cards, inline fields, OST links, standalone offers) get outlines plus "Edit in M@S Studio"      
  badges that open the correct Studio entry for that specific content.                                                                                            
   
  Each mode is gated by its own checkbox in the popup, persists via URL param (`mepCaasHighlight=true` / `mepMasHighlight=true`), and has zero impact when off.   
                                                                                                                                                                
  Builds on caas-libs changes from [adobecom/caas#453](https://github.com/adobecom/caas/pull/453) (merged), which added `data-country` and `data-card-url` to card
   elements.                                                                                                                                                    
                                                                                                                                                                  
  Resolves: [MWPW-194203](https://jira.corp.adobe.com/browse/MWPW-194203), [MWPW-194204](https://jira.corp.adobe.com/browse/MWPW-194204)                          
   
  ---                                                                                                                                                             
                                                                                                                                                                
  ### Highlight CaaS — MWPW-194203                                                                                                                                
   
  - **Block badge** — `<a class="mep-caas-edit-badge">` injected as a sibling immediately before each `[data-caas-block]`, linking to the Configurator with the   
  full block config pre-loaded. Sibling placement prevents React from reconciling the anchor away.                                                              
  - **Per-card path badges** — cards from `caas-libs` carry `data-card-url` (Chimera reference field) and `data-country`. CSS `::before` pills show 🔗            
  `caas-market:` (green) or 🔗 `caas-lang:` (yellow) by country code.                                                                                             
  - **Click-to-navigate** — clicking a per-card badge area opens the card's source URL, rewritten to the current preview host. BACOM `/blog/` cards route to the
  `-blog` sibling repo.                                                                                                                                           
  - **Lazy-load support** — MutationObserver picks up late-rendered blocks. **Observer attaches on toggle-on, disconnects on toggle-off** (addresses review     
  feedback).                                                                                                                                                      
  - **Review feedback addressed** — selectors scoped to `[data-caas-block]`, empty-path badge suppression, `data-card-path` cleanup on toggle-off, observer     
  hygiene.                                                                                                                                                        
                                                                                                                                                                
  ### Highlight M@S — MWPW-194204                                                                                                                                 
                                                                                                                                                                
  - **Five surfaces** stamped with `data-mas-block`: `collection`, `card`, `inline`, `ost`, `offer`. Each gets its own outline color and "Edit in M@S Studio"     
  badge.
    - Collection (purple) → sibling badge                                                                                                                         
    - Card (orange) → real-DOM action stack (Edit Card / Edit OSI / Copy Fragment ID)                                                                           
    - Inline / OST / Offer (pink / blue / blue) → CSS `::before` pseudo-badges with click delegation to the captured Studio URL                                   
  - **Sub-collections** — `<merch-card-collection>` sidenav-filtered sub-collections get their own "Edit Sub-collection" badge. Sub-collection metadata is        
  captured at fragment load via `aem:load` (autoblock-side `attachAemLoadListener`; new `preview-mas-subcollection.js`).                                          
  - **Authoring URL capture** — `mepMasStudioUrls` WeakMap stores the original studio URL keyed by host element. Capture happens at autoblock time                
  (merch-card-autoblock, merch-card-collection-autoblock, merch.js); consumed by the preview overlay for badge `href`s.                                           
  - **Market awareness** — page market resolved from `<mas-commerce-service country>` → locale → `?akamaiLocale`. Per-host market checked against               
  `data-ims-country` (post-WCS) → `country=` query param → page fallback. Card chip flips to yellow on mismatch.                                                  
  - **M@S Markets spoofer dropdown** — lets authors test different markets without hand-editing the URL. Standalone on non-Lingo pages; gated by "Use M@S markets
  instead" checkbox alongside the Lingo dropdown.                                                                                                                 
  - **"No M@S content found" message** when the page has no M@S surfaces.                                                                                       
  - **Observer hygiene** — same pattern as CaaS: all listeners (capture-phase click, `aem:load`, restamp-click, MutationObserver) attach on `Highlight M@S`       
  toggle-on and detach on toggle-off via named handler refs.                                                                                                      
   
  ---                                                                                                                                                             
                                                                                                                                                                
  ### Implementation notes

  - The CaaS Configurator deep-link URL is ~2.5KB of base64 config — too large for a `data-*` attribute. Stored in a module-level `WeakMap` (`mepCaasConfigUrls`  
  in `caas/utils.js`), keyed by the wrapper element. CaaS removes and recreates the wrapper during init; the entry is copied during the handoff.
  - The M@S `mepMasStudioUrls` WeakMap is shared across `merch.js`, `merch-card-autoblock.js`, `merch-card-collection-autoblock.js`, and the preview overlay via  
  the new `libs/blocks/merch/mas-mep-utils.js` module. WeakMap so entries GC alongside their host elements.                                                       
  - Sub-collection capture has a strict ordering constraint: the `aem:load` listener must be attached BEFORE the container is mounted — M@S removes
  `<aem-fragment>` immediately after dispatching the event. The autoblock attaches at fragment-creation time; the preview overlay has a document-level fallback   
  for the "highlight enabled mid-session" race.                                                                                                                 
                                                                                                                                                                  
  ---                                                                                                                                                           

  ### Test URLs

  **CaaS:**
  - https://main--da-bacom--adobecom.aem.page/resources/main?akamaiLocale=sg&milolibs=mep-cardattribute&mepCaasHighlight=true
                                                                                                                                                                  
  **M@S:**
  - https://stage--cc--adobecom.aem.page/fr/creativecloud/plans?milolibs=mep-cardattribute&mepMasHighlight=true&mepMasMarket=true                                                                                                                                            
  - https://main--cc--adobecom.aem.page/products/photoshop?milolibs=mep-cardattribute&mepMasHighlight=true 
                                                                                                                                                                  
  PSI: https://mep-cardattribute--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/libs/blocks/caas/caas.js
+++ b/libs/blocks/caas/caas.js
@@ -1,4 +1,5 @@
 import {
+  mepCaasConfigUrls,
   decodeCompressedString,
   initCaas,
   loadCaasFiles,
@@ -60,6 +61,8 @@ const loadCaas = async (a) => {
   a.insertAdjacentElement('afterend', modalDiv);
 
   a.insertAdjacentElement('afterend', block);
+  block.dataset.caasBlock = '';
+  mepCaasConfigUrls.set(block, a.href);
   a.remove();
 
   const langFirst = getMetadata('langfirst');
--- a/libs/blocks/caas/utils.js
+++ b/libs/blocks/caas/utils.js
@@ -12,6 +12,9 @@ import { getLingoActive } from '../../utils/lingo-active.js';
 import { fetchWithTimeout } from '../utils/utils.js';
 import getUuid from '../../utils/getUuid.js';
 
+// CaaS deep-link URL per wrapper element, keyed off-DOM for MEP preview (URL too large for data-*).
+export const mepCaasConfigUrls = new WeakMap();
+
 export const LANGS = {
   en: 'en',
   de: 'de',
@@ -1225,11 +1228,16 @@ export const initCaas = async (state, caasStrs, el) => {
   if (!caasEl) return;
 
   const appEl = caasEl.parentElement;
+  const preservedConfigUrl = mepCaasConfigUrls.get(caasEl);
   caasEl.remove();
 
   const newEl = document.createElement('div');
   newEl.id = 'caas';
   newEl.className = 'caas-preview';
+  if (preservedConfigUrl) {
+    newEl.dataset.caasBlock = '';
+    mepCaasConfigUrls.set(newEl, preservedConfigUrl);
+  }
   appEl.append(newEl);
 
   const config = await getConfig(state, caasStrs);
--- a/libs/blocks/merch-card-autoblock/merch-card-autoblock.js
+++ b/libs/blocks/merch-card-autoblock/merch-card-autoblock.js
@@ -1,6 +1,7 @@
 import { createTag, getConfig } from '../../utils/utils.js';
 import { decorateButtons, getBlockSize } from '../../utils/decorate.js';
 import { postProcessAutoblock } from '../merch/autoblock.js';
+import { mepMasStudioUrls } from '../merch/mas-mep-utils.js';
 import {
   initService,
   getOptions,
@@ -145,6 +146,11 @@ export async function createCard(el, options) {
   seenFragments.add(options.fragment);
   const aemFragment = createTag('aem-fragment', attrs);
   const merchCard = createTag('merch-card', { consonant: '' }, aemFragment);
+  // For the "Edit Card" mep preview badge.
+  if (getConfig()?.mep?.preview) {
+    mepMasStudioUrls.set(merchCard, el.href);
+    merchCard.dataset.masBlock = 'card';
+  }
   const parent = el.parentElement;
   if (parent && parent.tagName === 'P' && parent.children.length === 1) {
     parent.replaceWith(merchCard);
@@ -162,6 +168,10 @@ async function createInline(el, options) {
   seenFragments.add(options.fragment);
   const aemFragment = createTag('aem-fragment', attrs);
   const masField = createTag('mas-field', { field: options.field }, aemFragment);
+  if (getConfig()?.mep?.preview) {
+    mepMasStudioUrls.set(masField, el.href);
+    masField.dataset.masBlock = 'inline';
+  }
   el.replaceWith(masField);
   await checkReady(masField);
   normalizeBlockFieldWrappers(masField);
--- a/libs/blocks/merch-card-collection-autoblock/merch-card-collection-autoblock.js
+++ b/libs/blocks/merch-card-collection-autoblock/merch-card-collection-autoblock.js
@@ -1,6 +1,7 @@
 import { createTag, getConfig, localizeLinkAsync } from '../../utils/utils.js';
 import { debounce } from '../../utils/action.js';
 import { postProcessAutoblock, handleCustomAnalyticsEvent } from '../merch/autoblock.js';
+import { mepMasStudioUrls } from '../merch/mas-mep-utils.js';
 import {
   initService,
   getOptions,
@@ -320,6 +321,17 @@ export async function createCollection(el, options) {
   }
   const collection = createTag('merch-card-collection', attributes, aemFragment);
   const container = createTag('div', null, collection);
+  if (getConfig()?.mep?.preview) {
+    mepMasStudioUrls.set(container, el.href);
+    container.dataset.masBlock = 'collection';
+    // Attach BEFORE the replaceWith below — M@S removes aem-fragment
+    // immediately after dispatching aem:load. Dynamic import keeps
+    // preview-only code out of the production bundle.
+    const { attachAemLoadListener } = await import(
+      '../../features/personalization/preview-mas-subcollection.js'
+    );
+    attachAemLoadListener(aemFragment, container);
+  }
   const paragraph = el.parentElement;
   const toReplace = paragraph?.tagName === 'P' && hasOnlyTargetContent(paragraph, el)
     ? paragraph
--- a/libs/blocks/merch/mas-mep-utils.js
+++ b/libs/blocks/merch/mas-mep-utils.js
@@ -1,0 +1,4 @@
+// Only populated when MEP preview is active; WeakMap so entries
+// GC with their host element.
+// eslint-disable-next-line import/prefer-default-export
+export const mepMasStudioUrls = new WeakMap();
--- a/libs/blocks/merch/merch.js
+++ b/libs/blocks/merch/merch.js
@@ -3,6 +3,7 @@ import {
   shouldAllowKrTrial, getCountry,
 } from '../../utils/utils.js';
 import { replaceKey } from '../../features/placeholders.js';
+import { mepMasStudioUrls } from './mas-mep-utils.js';
 
 // MAS Component Names
 export const COMMERCE_LIBRARY = 'commerce';
@@ -1554,6 +1555,12 @@ export default async function init(el) {
   log = service.Log.module('merch');
   if (merch) {
     log.debug('Rendering:', { options: { ...merch.dataset }, merch, el });
+    // Rebuilt merch href keeps only the OSI; stash the original for
+    // the "Edit OSI" badge.
+    if (getConfig()?.mep?.preview) {
+      mepMasStudioUrls.set(merch, el.href);
+      merch.dataset.masBlock = 'ost';
+    }
     el.replaceWith(merch);
     return merch;
   }
--- a/libs/blocks/ost/ost.js
+++ b/libs/blocks/ost/ost.js
@@ -160,10 +160,12 @@ export async function loadOstEnv() {
     }
     window.history.replaceState({}, null, `${window.location.origin}${window.location.pathname}?${searchParameters.toString()}`);
   }
+  const countryParam = searchParameters.get('country');
   const attributes = { 'allow-override': 'true' };
   if (masDefaultsEnabled) {
     attributes['data-mas-ff-defaults'] = 'on';
   }
+  if (countryParam) attributes.country = countryParam.toUpperCase();
   await initService(true, attributes);
   // Load commerce.js based on masLibs parameter
   masCommerceService = await loadMasComponent(COMMERCE_LIBRARY);
@@ -239,6 +241,7 @@ export async function loadOstEnv() {
   const repo = searchParameters.get('repo');
 
   let { country, language } = getMiloLocaleSettings();
+  if (countryParam) country = countryParam;
   const { locales } = getConfig();
   const log = Log.module('ost');
   const metadata = {};
--- a/libs/features/personalization/preview-mas-subcollection.js
+++ b/libs/features/personalization/preview-mas-subcollection.js
@@ -1,0 +1,175 @@
+// MEP "Highlight M@S Content" — sub-collection edit badges.
+//
+// A <merch-card-collection> can host sub-collections surfaced in the
+// sidenav (e.g., plans-page categories: Photo / Video / Design). When the
+// user picks one, we render an "Edit Sub-collection" badge linking into
+// M@S Studio.
+//
+// Capture is wired by the autoblock via attachAemLoadListener BEFORE
+// decoration: the M@S collection web component removes the parent
+// <aem-fragment> immediately after dispatching aem:load, so its rawData
+// can't be read after the fact.
+
+import { createTag } from '../../utils/utils.js';
+import { mepMasStudioUrls } from '../../blocks/merch/mas-mep-utils.js';
+
+export const SUB_COLLECTION_BADGE_CLASS = 'mep-mas-sub-collection-badge';
+
+// container -> Array<{ id, label, queryLabel }>
+export const mepMasSubCollections = new WeakMap();
+
+// Walks the referencesTree carried by aem:load.detail (mirrors the shape
+// in mas/web-components/src/merch-card-collection.js). Sub-collections are
+// nodes whose fieldName isn't 'cards' or 'variations'. Recurses for nested
+// catalog hierarchies.
+export function extractSubCollections(payload) {
+  if (!payload) return [];
+  const { references, referencesTree } = payload;
+  if (!Array.isArray(referencesTree) || !references) return [];
+  const out = [];
+  const seen = new Set();
+  function walk(nodes) {
+    nodes.forEach((node) => {
+      if (!node || typeof node !== 'object') return;
+      const { identifier, fieldName } = node;
+      if (!identifier || seen.has(identifier)) return;
+      if (fieldName === 'cards' || fieldName === 'variations') return;
+      const fields = references[identifier]?.value?.fields;
+      if (fields && (fields.label || fields.queryLabel)) {
+        seen.add(identifier);
+        out.push({
+          id: identifier,
+          label: fields.label || '',
+          queryLabel: fields.queryLabel || '',
+        });
+      }
+      if (Array.isArray(node.referencesTree) && node.referencesTree.length) {
+        walk(node.referencesTree);
+      }
+    });
+  }
+  walk(referencesTree);
+  return out;
+}
+
+export function getSubCollectionsFor(container) {
+  if (!container) return [];
+  return mepMasSubCollections.get(container) || [];
+}
+
+// One-shot aem:load capture — must be attached pre-decoration because M@S
+// removes the aem-fragment immediately after dispatching the event.
+export function attachAemLoadListener(aemFragment, container) {
+  if (!aemFragment || !container) return;
+  aemFragment.addEventListener('aem:load', (event) => {
+    const subCollections = extractSubCollections(event.detail);
+    if (subCollections.length) {
+      mepMasSubCollections.set(container, subCollections);
+    }
+  }, { once: true });
+}
+
+// Rewrites the parent's Studio URL to open the named sub-collection in
+// fragment-editor view. content-type stays as merch-card-collection
+// (sub-collections are themselves collections).
+export function toSubCollectionStudioUrl(parentUrl, subCollectionId) {
+  if (!parentUrl || !subCollectionId) return parentUrl;
+  try {
+    const u = new URL(parentUrl, window.location.origin);
+    const hash = (u.hash || '').replace(/^#/, '');
+    const hp = new URLSearchParams(hash);
+    if (hp.has('fragmentId')) hp.set('fragmentId', subCollectionId);
+    else if (hp.has('query')) hp.set('query', subCollectionId);
+    else if (hp.has('fragment')) hp.set('fragment', subCollectionId);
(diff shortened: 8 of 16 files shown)


sanrai (changes requested):
Architecture is solid. The WeakMap trick to stash the Configurator URL off-DOM is the right call given the payload size, and putting the anchor as a previousElementSibling to dodge React reconciliation is a nice detail.

Test this up against main--da-bacom--adobecom.aem.live/resources/main?milolibs=mep-cardattribute&mepCaasHighlight=true. 

Was able to validate:

- Block outline and "Edit in CaaS Configurator" badge inject correctly. Decoded the href and it's a valid CaaS config (Bacom source, "All resources" collection, 9 per page). Clicked through and the Configurator opens with everything pre-loaded.
- MutationObserver picks up lazy-rendered cards and reapplies the badges.
- Toggling the checkbox in the MEP popup is symmetric: body attr flips, the anchor gets removed, CSS outlines disappear immediately, and flipping back restores all of it.

A few things that should be addressed:

**1. [data-country] selectors are unscoped.** They match any element with data-country on the page, not just cards inside [data-caas-block]. If anything else on a BACOM page ever stamps that attribute (country picker, form input, etc.) it'll pick up the green/yellow pill treatment when the highlight is on. Tightening to [data-caas-block] [data-country] would scope it.

**2. data-card-path persists after toggling off.** derivePathsForCards writes the attribute, but removeCaasBadges never strips it. After you turn the highlight off, the paths are still sitting in the DOM. Not functionally a problem but it's invisible state that'll trip up debugging later. Would be nice to have a matching cleanup.

**3. MutationObserver never disconnects.** Set up once in decoratePreviewMode and runs forever. The early return on body data-attr keeps it cheap, but for an author-only preview feature I'd rather see it attach on toggle-on and detach on toggle-off. Keeps things clean for authors who never enable the feature.

**4. escapeHtml import looks unused.** It's in the new import block in preview.js but I don't see any new call sites. Probably leftover from an earlier pass. Lint will flag it.

One other thing that came up during testing, may or may not be in scope: on /resources/main every card came back with data-country="xx"` and no `data-card-url`. 


```css
body[data-mep-caas-highlight='true'] [data-country][data-card-path='']::before {
  content: none;
}
```

Happy to re-approve once 1 and the empty-path cosmetic are sorted. 2, 3, 4 are non-blocking from my side but would be good to take in this PR.

sanrai:
## Tested locally

**Loaded the deployed milolibs URL:**
https://main--da-bacom--adobecom.aem.live/resources/main?milolibs=mep-cardattribute&mepCaasHighlight=true

**Confirmed the PR's bundle is live** at `mep-cardattribute--milo--adobecom.aem.live/libs/features/personalization/preview.js` — contains the new `injectCaasBadges` and `mepCaasHighlight` symbols

**Verified the caas-libs stable bundle** already ships the `data-country` / `data-card-url` attribute logic from PR 453 (string-grep on `www.adobe.com/special/chimera/caas-libs/stable/main.min.js` confirmed both attributes present)

---

**DOM checks against the rendered page:**

- `body[data-mep-caas-highlight='true']` set correctly from the URL param
- `[data-caas-block]` stamped on the CaaS wrapper after `initCaas` recreates the element (the WeakMap-preservation handoff in `caas.js` works as designed)
- `<a class="mep-caas-edit-badge">` injected as `previousElementSibling` of the wrapper (correctly outside React's reconciliation scope)
- Anchor `href` starts with `https://milo.adobe.com/tools/caas#` with a ~2.5 KB base64 payload
- Decoded the payload via `DecompressionStream('gzip')`: 6800 chars of JSON with expected config keys (`endpoint`, `cardStyle`, `andLogicTags`, `excludeTags`, `layoutType`, `paginationEnabled`, `resultsPerPage`, etc.) and the actual Chimera endpoint URL
- Visible blue dashed outline around `[data-caas-block]`
- Yellow dashed outlines + 🔗 `caas-lang:` pills on cards with `data-country="xx"`

---

**End-to-end Configurator round-trip:** Clicked the "Edit in CaaS Configurator" badge — a new tab opened at `milo.adobe.com/tools/caas` with the Configurator UI pre-loaded with Collection Name "All resources", Source "Bacom", Results Per Page 9, Auto detect country & lang checked. All values match the source block on the parent page.

---

**MEP popup toggle interaction:**

- Opened popup via the "1 Manifest found" pill
- Confirmed new "Highlight CaaS" checkbox present, checked from URL param state
- Unchecked → `body` data-attr flipped to `false`, badge anchor removed, all CSS outlines disappeared immediately (`bodyAttr=false`, `badges=0`, `outlinedCards=0`)
- Re-checked → all artifacts reappeared (`bodyAttr=true`, `badges=1`, `outlinedCards=9`)
- Symmetric toggle behavior with no state leakage

---

**MutationObserver lazy-render handling:** Initial probe showed 0 cards (CaaS data still fetching); 5 seconds later 9 cards rendered, badges had been re-injected automatically

markpadbe:
Thanks @sanrai — heads up. I’ve expanded this PR to also cover **`MWPW-194204`** (`Highlight M@S Content`).

Both features are required for the `Lingo` launch next week, and combining them avoids parallel review cycles against the same deadline.

## Status

### CaaS review feedback — addressed in `7cde0cae4`

- Scope selectors to descendants of `[data-caas-block]`
- Suppress per-card badge when `data-card-path` is empty
- Strip `data-card-path` in `removeCaasBadges` on toggle-off
- Attach `MutationObserver` on toggle-on; disconnect on toggle-off

### M@S work — `f22d66d29`

- Mirrors the CaaS observer pattern (listeners attach/detach via the highlight toggle)
- Adds highlight surfaces: `collection`, `card`, `inline`, `OST`, `offer`
- Each surface includes its own outline and an “Edit in `M@S Studio`” entry point
- Adds sub-collection capture
- Adds `market-spoofer` dropdown
- Adds “No M@S content found” message per the ticket

Test URLs are included in the updated description. Thanks again.

sanrai (approved):

sanrai:
Followup pass since the rebase. The previous review feedback all landed cleanly and the new M@S scope checks out.

**Prior concerns addressed**

- `[data-country]` selectors now scoped to `[data-caas-block]` descendants. Confirmed across all four rules in preview.css. Country pickers / form inputs elsewhere on the page are no longer affected.
- `removeCaasBadges` now strips `data-card-path` so derived state doesn't persist after toggle-off. Comment in source documents the intent.
- MutationObserver attach/detach symmetric pair (`watchForCaasBlocks` / `unwatchForCaasBlocks`) wired into the toggle. No longer observing for the page lifetime.
- Empty-path pill suppressed via `[data-card-path='']::before { content: none; }`. Pills on `reference`-less cards no longer render as `🔗 caas-market:` with a blank path.

`escapeHtml` is genuinely used at preview.js lines 1428, 1429, 1521, 1523, 1525 in the geo-detection debug rendering.

**M@S scope — what I tested**

Loaded `https://main--cc--adobecom.aem.live/creativecloud/plans?milolibs=mep-cardattribute&mepMasHighlight=true` and walked the full surface.

- `hasMasSurfaces()` correctly detects: 210 `[data-mas-block]`, 34 `merch-card`, 180 `[data-wcs-osi]`, 3 `merch-card-collection`
- New "Highlight M@S" checkbox renders in the MEP popup, reflects URL-param state on load
- Three-tier badge taxonomy renders correctly:
  - 3 block-level `Edit Collection in M@S Studio · US` badges (one per `merch-card-collection`)
  - `Edit Card · US` orange badges on each `<merch-card>`
  - `Edit OSI · US` blue badges on each `[data-wcs-osi]` price element (54 total card+OSI badges)
  - `Copy Fragment ID` buttons compose alongside
- Badge destinations route correctly:
  - Card and Collection badges → `mas.adobe.com/studio.html` with base64 config in URL hash (~103 chars)
  - OSI badges → `milo.adobe.com/tools/ost` with query param
  - All badges carry `target="_blank"` and `rel="noopener noreferrer"`
  - All carry the market suffix (`· US`)
- Sub-collection badge composes with parent collection badge (renders adjacent, indented arrow indicator)
- Toggle off → all 57 badges and 3 outlines disappear instantly. `bodyMasHighlight` flips to `false`, no state leakage.
- Toggle on → everything re-injects. Symmetric.

**One thread worth opening**

PR scope grew between the first round and now. Title still reads "[MEP] Highlight CaaS blocks in MEP preview" but the M@S work is its own MWPW ticket (194204) and adds ~3,100 lines on top. Either the title could be updated to reflect both scopes (e.g., "Highlight CaaS and M@S blocks in MEP preview") or the M@S commit could move to its own PR. Not a blocker either way, just calling it out so it's documented.

Architecture continues to be solid. WeakMap for off-DOM URL storage mirrored cleanly between `mepCaasConfigUrls` and `mepMasStudioUrls`, body-attr-gated CSS, attach/detach observer pattern. Reuse is appropriate.

cmiqueo (approved):

markpadbe:
@sanrai @cmiqueo — heads up, additional commits
  since your approval covering a few bug fixes,     
  feedback items, and added test coverage. Re-review
   appreciated. 
  
## Update — additional commits since last review

### View in OST: pass spoofed market country  e363f5762

OST badge links now append `&country=<pageMarket>` to the OST URL so the tool opens pre-populated to the correct market.

`ost.js` was updated to read the `?country=` param and pass it to `mas-commerce-service` on init, so the checkout widget reflects the correct pricing immediately.

#### How to test

- Enable Chrome DevTools → **Sources → Overrides**, then add a local override for:


https://milo.adobe.com/tools/ost/ost.js

- Open the following URL:


https://milo.adobe.com/tools/ost?osi=r_JXAnlFI7xD6FxWKl2ODvZriLYBoSL701Kd1hRyhe8&country=BE

- Confirm:
  - Country field is pre-populated with `BE`
  - Pricing reflects Belgium

#### Why not test via badge click?

> **Note**  
> Chrome DevTools local overrides only intercept network requests when DevTools is open and active on that specific tab.

Clicking a **View in OST** badge opens a new tab where DevTools is not open, so the override does not apply and production `ost.js` (which does not yet read `?country=`) loads instead.

> **Workaround**  
> Paste the URL into the existing OST tab where DevTools is already open with the override active.  
> Once this branch ships to production, no override will be needed and badge clicks will work end-to-end.

---

### Fix clickability for pre-stamped `data-mas-block="ost"` elements  bd4eebeb6

`annotateOffers` previously skipped elements that already had `data-mas-block="ost"` stamped (for example, from a prior preview run or server-side stamping).

Those elements now receive a badge regardless of whether the attribute was pre-existing.

---

### Fix inline field badge clickability and nested badge overlap  051097d24

Changes included:

**Label rename**
- `"Edit Inline Field in M@S Studio"` → `"Edit Inline Field"`

**Nested hit-zone fix**
- When an OST/offer badge sits inside an inline-field element, `handleChildCardBadgeClick` could select the wrong candidate from `composedPath()`
- The handler now iterates all matching candidates and detects nesting via `isNestedInInline`
- Nested badges shift their pseudo-badge hit zone below the host element instead of overlapping it

This prevents the inline field badge from swallowing clicks intended for the child OST badge.

---

### Test coverage  f147b34f3

118 tests now passing. New coverage includes:

- CaaS block badge injection via `mepCaasConfigUrls`
- `getResolvedPageMarket` fallback chain:


mas-commerce-service → locale → ?akamaiLocale= → null

- Stale badge replacement (`prev.remove()`) when card market changes between passes
- `masAemLoadHandler` guard conditions and sub-collection extraction
- `derivePathsForCards` error catch path
- Clipboard copy via `handleHighlightLinkClick`
- Nested OST badge positioning (`isNestedInInline` hit-zone shift)

--- 

### Also pushed 23203c781 to clear a CodeQL data-flow false positive — no behavior change.

Axelcureno (approved):
Did a deeper pass focused on correctness, MAS-side contracts, and edge paths. Approving — findings below are non-blocking, fine as a follow-up.

### Verified clean
- `aem:load` event detail shape matches MAS's dispatch (`references` + `referencesTree`)
- No `data-mas-*` namespace collision with MAS web-components
- OST `el.href` capture timing correct — full authoring URL preserved at capture point
- Observer / listener hygiene symmetric on toggle-on/off (capture flag matched on add/remove)
- Click delegation correctly uses `composedPath()` for shadow-DOM crossing
- WeakMap handoff around `caasEl.remove()` is sound

### Medium — worth fixing

**M1. `libs/features/personalization/preview.js:977` — missing `upt-link` from OSI selector**

`MAS_OSI_SELECTOR` mirrors three of the four MAS commerce selectors from `constants.js:52–57` but drops `SELECTOR_MAS_UPT_LINK`. Upt-links carry `data-ims-country` and are grouped with the other commerce surfaces in MAS. `upt-link` instances will silently skip the "View in OST" overlay.

```js
const MAS_OSI_SELECTOR = [
  'span[is="inline-price"][data-wcs-osi]',
  'a[is="checkout-link"][data-wcs-osi]',
  'a[is="upt-link"][data-wcs-osi]',
  'button[is="checkout-button"][data-wcs-osi]',
  'sp-button[data-wcs-osi]',
].join(',');
```

**M2. `libs/features/personalization/preview.js:1339` — RTL hit-box swallows real clicks**

The click hit zone uses `rect.right - hit.w ≤ clientX ≤ rect.right`. On RTL pages (Arabic, Hebrew) `rect.right` is the text-start edge, not the visual badge edge, and the CSS `::before` doesn't mirror for RTL either.

Two failure modes on RTL pages:
1. Authors can't click the visible badge (hit zone is on the wrong side)
2. Clicks on legitimate CTAs near the geometric-right edge get `preventDefault + stopPropagation`'d and appear dead

If the Lingo launch includes RTL CC pages, worth fixing now. Otherwise follow-up.

```js
const isRtl = getComputedStyle(host).direction === 'rtl';
const xMin = isRtl ? rect.left : rect.right - hit.w;
const xMax = isRtl ? rect.left + hit.w : rect.right;
```
Plus matching `[dir="rtl"]` selectors in `preview.css` to mirror the `::before` position.

**M3. `libs/features/personalization/preview.js:1024` — `getCardMarket` misses inline-price market context**

In MAS, only `checkout-mixin` and `upt-link` set `data-ims-country` post-WCS. `inline-price` (the `<span is="inline-price">` surface) does not. A display-only card (prices, no checkout link) falls through to `pageMarket`, so the chip shows the page market even if the rendered price came back for a different country.

Not a correctness bug for the common case (almost every commerce card has a checkout link). A one-line comment is enough:

```js
// inline-price doesn't set data-ims-country (only checkout-mixin and
// upt-link do); display-only cards will fall back to pageMarket.
```

### Low — nice to have

**L1. `libs/features/personalization/preview.js:140–144` — `MAS_LABELS` keys mostly unreachable**

`buildMasBadge` is only called from the loop at `preview.js:1294`, and the non-collection surfaces all short-circuit before reaching it (lines 1266–1275 handle card via action-stack, offer/inline/ost via `::before` pseudos). The comment "label kept for symmetry" applies to all four non-collection keys, not just `offer`. Can reduce to:

```js
const MAS_LABELS = { collection: 'Edit Collection in M@S Studio' };
```
(and drop the `|| 'Edit in M@S Studio'` fallback.)

**L2. `libs/features/personalization/preview.js:1248` — orphan CSS state classes**

The badge stamps `mep-mas-edit-badge-market-${state}` where `state ∈ {match, mismatch, unknown}`, but `preview.css` only defines `-market` (base) and `-market-mismatch`. The `-match` and `-unknown` classes are applied but have no rules. Either drop the `-${state}` suffix from JS or add the missing CSS rules.

**L3. `libs/features/personalization/preview-mas-subcollection.js:217` — fieldName filter brittle to MAS model changes**

The negative-list filter (`!== 'cards' && !== 'variations'`) breaks silently if MAS adds a new top-level reference fieldName (e.g. `media`, `icons`, `addons`). A positive-list is safer — sub-collections always carry `fields.queryLabel`, standalone references don't:

```js
if (!fields?.queryLabel) return; // not a sub-collection
```

Also worth scoping the "M@S removes `<aem-fragment>` immediately" comment to collections — `<merch-card>` and `<mas-field>` don't do this.

markpadbe:
Thanks for the deep pass, @Axelcureno — pushed 49ad8948 addressing five of the six items:

- M1 — Added "a[is=\"upt-link\"][data-wcs-osi]" to MAS_OSI_SELECTOR.
- M3 — Added the documentation comment in getCardMarket.
- L1 — Simplified MAS_LABELS to just collection; confirmed buildMasBadge only sees collection (card/inline/ost/offer all early-return).
- L2 — Dropped the orphan "-match" / "-unknown" state classes; only stamp "-mismatch" now, updated the test. These were legacy from Lingo/CaaS where state was visualized; current M@S uses surface-color + country code, not state-color, so the extra classes weren’t doing anything useful.
- L3 — Switched to a positive-list filter on fields.queryLabel. Confirmed against MAS source — queryLabel is the sidenav filter key and only sub-collections carry it; cards and variations have only label.

- M2 — Took a closer look here but left this as-is. Tested on mena_ar and the badge/hit-box align as expected. getBoundingClientRect().right is always physical-viewport right (CSSOM-View spec), and the badge CSS uses physical "right: -2px", not "inset-inline-end", so both anchors land on the same physical edge regardless of dir. Changing it would shift the hit zone left while the badge stays visually on the right. Happy to add a short CSS comment calling out the intentional coordinate behavior if that helps future readers.

markpadbe:
Context on the Nala failure: the failing Nala MAS check ('@Commerce-Volume-Discount-Annual') is a known-external test fixture issue, not from this PR.                                                                                                                            
                                                                                                                                                                                                                                                                Confirmed by @ilyas in the '#merch-at-scale' Slack thread today: WCS isn't returning a promo offer for 'TDARCH_3LIC_7PT5', and the annual price template doesn't fall back gracefully to regular-price logic. Reproducible on plain stage in DevTools without any branch involvement:                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                https://main--milo--adobecom.aem.live/drafts/nala/features/commerce/prices-volume-discount-annual?georouting=off&martech=off                                                                                                                                   
   
→ both inline-price elements: `placeholder-failed`                                                                                                                                                                                                             
                                                                                                                                                                                                                                                               
The same Nala failure is currently blocking the stage→main release and showing up on other open PRs, so it's not specific to #5915. Requesting merge override; happy to retest once the WCS data / template handling fix lands.

cc: @narcis-radu @vhargrave @vivgoodrich @mokimo @rgclayton

sukamat (approved):
Looks great!  👏 
This should simplify the authoring steps for MAS & CaaS for GWP. Also very valuable for quick validations on where the content is coming from.

PR #5920, opened 2026-05-11
feat: add hreflang sitemap caching to eliminate redundant bot fetches

jsandland:
## Problem

Bots crawling locale pages (e.g. `/de/`) fetch the full sitemap XML on every single page load to extract hreflang alternate links. For large locale sitemaps like CC's `de`, that's **470KB compressed / 15MB uncompressed per page load**. A bot crawling 1k pages downloads 15GB of sitemap data and spends ~23 minutes just on sitemap fetches.

This was flagged in [da-cc PR #18](https://github.com/adobecom/da-cc/pull/18) (MWPW-188278) and tracked for follow-up as [MWPW-191067](https://jira.corp.adobe.com/browse/MWPW-191067).

## Solution

A new shared module `libs/features/hreflang/hreflang.js` that:

1. On the first page a bot visits, fetches and parses the full sitemap
2. Extracts only the URL-to-hreflang mapping (strips all XML markup overhead)
3. Caches the processed map in `sessionStorage`
4. On every subsequent page in that session, skips the fetch entirely and does a direct map lookup

**First page:** 1 fetch, parses once, stores lightweight map
**Every subsequent page:** 0 fetches, instant sessionStorage lookup

## Design decisions

- **Processed map, not raw XML** keeps sessionStorage footprint well within the 5MB browser limit even for large sitemaps
- **Locale-scoped cache keys** so `/de/sitemap.xml` and `/fr/sitemap.xml` get separate entries with no cross-locale collisions
- **Fully configurable** via `sitemapTemplate` and `sitemapOrigin` so this works for bacom, CC, acrobat, and any other property without hardcoding paths
- **`sitemapOrigin` vs `location.origin`** - fetch uses current origin (works in local dev), map lookup uses prod origin (matches `<loc>` values in sitemap)
- **Bot-only** and gated by `meta[name="hreflinksuseragents"]` so regular users are never affected

## Usage

```js
import(`${libs}/features/hreflang/hreflang.js`).then(({ appendHreflangLinks }) => {
  appendHreflangLinks({
    locales: LOCALES,
    sitemapTemplate: '/{locale}/cc-shared/assets/sitemap.xml', // CC
    sitemapOrigin: 'https://www.adobe.com',
  });
}).catch((e) => window.lana?.log(`hreflang: failed to load module - ${e.message}`, { tags: 'hreflang', severity: 'error' }));
```

## Test plan

- [x] 9 unit tests covering: UA gating, sitemap fetch, sessionStorage cache hit, page not found, fetch failure, timeout/abort, quota exceeded, invalid template
- [x] Validated locally on bacom: sitemap fetched once, 11 hreflang links injected after `<title>` on homepage
- [ ] CC to wire up and validate against `/de/cc-shared/assets/sitemap.xml`

### DIFF
--- a/libs/features/hreflang/hreflang.js
+++ b/libs/features/hreflang/hreflang.js
@@ -48,24 +48,32 @@ async function fetchSitemapMap(sitemapUrl) {
     }
     return buildHreflangMap(xmlDoc);
   } catch (e) {
-    window.lana?.log(`hreflang: error fetching sitemap ${sitemapUrl} - ${e.message}`, { tags: 'hreflang', severity: 'error' });
+    const msg = e.name === 'AbortError'
+      ? `hreflang: sitemap fetch timed out after ${FETCH_TIMEOUT_MS}ms: ${sitemapUrl}`
+      : `hreflang: error fetching sitemap ${sitemapUrl} - ${e.message}`;
+    window.lana?.log(msg, { tags: 'hreflang', severity: 'error' });
     return null;
   } finally {
     clearTimeout(timeoutId);
   }
 }
 
-function getSitemapPath(pathname, locales, sitemapTemplate) {
-  const localeMatch = locales.find((locale) => pathname.startsWith(`/${locale}/`));
-  return localeMatch ? sitemapTemplate.replace('{locale}', localeMatch) : sitemapTemplate.replace('/{locale}', '');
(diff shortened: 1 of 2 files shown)


JasonHowellSlavin (changes requested):
Sorry. We actually cannot contribute this to Milo. I will explain a bit more later today.

PR #5922, opened 2026-05-12
New grid definition for C2

overmyheadandbody:
The grid system for C2 is being adapted to support Wave 2 use-cases for the Site Redesign initiative. The default is now the `1920px` version. The `1440px` one has been removed entirely; references to it from the Global 

Resolves: [MWPW-194970](https://jira.corp.adobe.com/browse/MWPW-194970)

**Test URLs:**
- Before: https://main--upp--adobecom.aem.live/upp-shared/fragments/tests/2026/q2/ace1201/sr-homepage?milolibs=site-redesign-foundation
- After: https://main--upp--adobecom.aem.live/upp-shared/fragments/tests/2026/q2/ace1201/sr-homepage?milolibs=c2-new-grid

PR #5923, opened 2026-05-12
[SR][MWPW-192887] - Social proof

zagi25:
* Add `social-proof` block
* Add `hangOpeningQuote` to `decorate.js` (`hangOpeningQuote` should be imported in `rich-content` also). I will create a follow up PR. 

[Design](https://www.figma.com/design/q87sUm2fvForRQzxGum5wE/Offer-%E2%80%94-A.com?node-id=11675-8733&m=dev)
[Spec](https://www.figma.com/design/q87sUm2fvForRQzxGum5wE/Offer-%E2%80%94-A.com?node-id=12072-10437&m=dev)

Resolves: [MWPW-192887](https://jira.corp.adobe.com/browse/MWPW-192887)

**NOTE:** Grid is not the same as on design, waiting for #5922. Also animation is missing, per design, waiting for prototype. 

**Test URLs:**
- https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-offer?milolibs=mwpw-192887&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all


libs/mep/ace1205/social-proof/social-proof.css
```diff
@@ -0,0 +1,85 @@
...
+    gap: var(--s2a-spacing-lg);
+    padding: var(--s2a-spacing-lg) var(--s2a-spacing-lg) var(--s2a-spacing-2xl);
+    border-radius: inherit;
+
+    .foreground {
+      display: flex;
+      flex-direction: column;
+      gap: var(--s2a-spacing-md);
+      padding-inline: var(--s2a-spacing-md);
+      max-width: 460px;
```
overmyheadandbody, line 33: Should this be a fixed value or a number of columns?

overmyheadandbody (approved):

robert-bogos (approved):

narcis-radu (approved):

libs/mep/ace1205/social-proof/social-proof.css
zagi25: Per design/specs, it is a fixed value

libs/mep/ace1205/social-proof/social-proof.css
```diff
@@ -1,19 +1,42 @@
 .social-proof {
+  overflow: clip;
   border-radius: var(--s2a-border-radius-24);
+  container-type: inline-size;
+  --animation-curve: cubic-bezier(0.42, 0, 0, 1);
```
overmyheadandbody: `--parallax-easing` should be available in the `:root`, I'd argue it's better to have it in that one place, so any changes cascade.

libs/mep/ace1205/social-proof/social-proof.css
```diff
@@ -1,19 +1,42 @@
...
+  --animation-timeline: --social-proof-view;
+  --animation-range: entry 10% contain 60%;
+  --animation-fill: both;
 
   > div {
+    --stretch-target: 100cqi;
+    --stretch-peak: min(2560px, 100vw);
```
overmyheadandbody, line 11: How does the `100vw` work if you enable scrollbars? Does it cause any horizontal scroll?

libs/mep/ace1205/social-proof/social-proof.css
```diff
@@ -83,3 +112,20 @@
     padding: var(--s2a-layout-md);
   }
 }
+
+@supports (animation-timeline: view()) {
```
overmyheadandbody, line 115: An additional `@supports` might be needed for users who prefer reduced motion.

libs/mep/ace1205/social-proof/social-proof.css
zagi25: Updated, thanks!

libs/mep/ace1205/social-proof/social-proof.css
zagi25: Added, thanks!

libs/mep/ace1205/social-proof/social-proof.css
zagi25: There is no horizontal scroll because of the `overflow: clip`. But scalling block to the 100vw will stretch it bellow the vertical scroll bar.
Screenshot 2026-05-20 at 15 45 13
Screenshot 2026-05-20 at 15 45 18

I didn't find any good way to simply solve this by just using css (100% doesn't work), one thing that can be done is to set variable through js to calculate width without scroll bar, something like `document.documentElement.clientWidth;`, but we would need an resize observer for this. Calculating just the scroll bar width will also need an observer because scroll bar can disappear on resize (not most likely in our case but possible) 

I am kinda ok with block stretching below scroll bar trade off, because it is barely visible. If you don't think this is okay, I can solve it with js or even better if you know a CSS trick that can help us 😄

overmyheadandbody: I can't see an issue in practice, so maybe we don't need to worry about it. So all good from my side for now, let's see if QA finds some edge cases.

overmyheadandbody (approved):

NadiiaSokolova (approved):
Verified. Ready to be merged.
Testing details [MWPW-192887](https://jira.corp.adobe.com/browse/MWPW-192887)

PR #5925, opened 2026-05-12
MWPW-192744 [Site Redesign] [Product: Acrobat Pro] | Product Marquee

suhjainadobe:
Adding product-marquee-grid block to site-redesign

Resolves: [MWPW-192744](https://jira.corp.adobe.com/browse/MWPW-192744)

Test URLs:

Before: https://main--upp--adobecom.aem.page/homepage/drafts/suhjain/redesign-demo
After: https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-product?milolibs=pm-grid&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all


libs/mep/ace1205/product-marquee-grid/product-marquee-grid.css
```diff
@@ -0,0 +1,116 @@
...
+    flex-direction: column;
+    justify-content: space-between;
+
+    .body-md {
+      color: var(--s2a-color-content-title);
+      font-family: var(--heading-font-family);
+      font-size: var(--s2a-typography-font-size-title-4);
+      font-weight: var(--s2a-font-weight-adobe-clean-display-black);
+      letter-spacing: var(--s2a-font-letter-spacing-4xl);
+      line-height: var(--s2a-typography-line-height-title-4);
```
overmyheadandbody: These should usually come free as part of the typography class. It might be easier to change the body class to a title one via JS in this case, to keep semantics correctly attached to styles.

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.css
```diff
@@ -0,0 +1,116 @@
...
+      object-fit: contain;
+    }
+
+    .title-2 {
+      color: var(--s2a-color-content-title);
+      font-family: var(--heading-font-family);
+      font-size: var(--s2a-typography-font-size-super);
+      font-weight: var(--s2a-font-weight-adobe-clean-display-black);
+      letter-spacing: var(--s2a-typography-letter-spacing-super);
+      line-height: var(--s2a-typography-line-height-super);
```
overmyheadandbody: Same comment as before. Even better in this scenario, you could add a `title super` variant to the block and the logic should pick it up and style it corrrectly.

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.js
```diff
@@ -0,0 +1,47 @@
...
+
+function decorate(block) {
+  const col = block.children[0]?.children[0];
+  if (!col) return;
+
+  const isSvgUrl = (url) => /\.svg(\?.*)?$/i.test(url || '');
+  const iconEl = col.querySelector('p img[src*=".svg"]');
+  if (iconEl && isSvgUrl(iconEl.src)) {
+    iconEl.src = getFederatedUrl(iconEl.src);
+  }
```
robert-bogos: `isSvgUrl` seems redundant here since the query selector already looks for `'p img[src*=".svg"]'`
```suggestion
  const iconEl = col.querySelector('p img[src*=".svg"]');
  if (iconEl) iconEl.src = getFederatedUrl(iconEl.src);
```

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.css
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  .pm-promo-button {
+    align-items: center;
+    background: var(--s2a-color-background-knockout);
+    border-radius: var(--s2a-border-radius-md);
+    display: flex;
+    gap: var(--s2a-spacing-md);
+    padding-block: var(--s2a-spacing-12);
+    padding-inline: var(--s2a-spacing-24) var(--s2a-spacing-12);
```
overmyheadandbody: Is it intentional to not have these be symmetrical? If so, this needs to be reflected for RTL. Additionally, I recommend you use semantic tokens, the ones with t-shirt sizes, instead of the primitive ones. For example, I think `--s2a-spacing-sm` maps to `--s2a-spacing-12`

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.css
```diff
@@ -0,0 +1,116 @@
...
+    text-decoration: none;
+
+    .pm-promo-text {
+      color: var(--s2a-color-content-knockout);
+      flex: 1;
+      font-family: var(--body-font-family);
+      font-size: var(--s2a-typography-font-size-eyebrow);
+      font-weight: var(--s2a-font-weight-adobe-clean-bold);
+      letter-spacing: var(--s2a-typography-letter-spacing-eyebrow);
+      line-height: var(--s2a-typography-line-height-eyebrow);
```
overmyheadandbody: Might be easier to add the class via JS.

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.css
```diff
@@ -0,0 +1,116 @@
...
+    padding-block: var(--s2a-layout-lg) var(--s2a-spacing-3xl);
+    padding-inline: var(--s2a-spacing-3xl);
+
+    .pm-chiclet-row img.icon {
+      block-size: 80px;
+      inline-size: 80px;
+    }
+
+    .pm-foreground {
+      flex: 1;
```
robert-bogos: This is already declared in the base rules. I think it can be removed
```suggestion
```

robert-bogos (commented):
Looks good overall, just a few comments.
I wanted to check out the test URL but is not loading the product marquee. Might be something you'd want to look into.

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.css
```diff
@@ -0,0 +1,116 @@
+.product-marquee-grid {
+  background: var(--s2a-color-background-default);
+  display: flex;
+  flex-direction: column;
+  min-block-size: 494px;
```
narcis-radu: how would this work on small screen sizes?

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.css
narcis-radu: +1 on this

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.js
```diff
@@ -0,0 +1,47 @@
...
+
+  const ctaLink = col.querySelector('p:has(em a) em a, p:has(strong a) strong a');
+  col.querySelector('p:has(em a), p:has(strong a)')?.remove();
+
+  decorateBlockText(col, { heading: '2', body: 'md' });
+
+  const heading = col.querySelector('h1, h2, h3, h4');
```
narcis-radu: what about h5, h6 ?

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.js
```diff
@@ -0,0 +1,47 @@
+import { createTag, getFederatedUrl } from '../../../utils/utils.js';
+import { decorateBlockText, decorateViewportContent } from '../../../utils/decorate.js';
+
+const CHEVRON_SVG = '<svg width="16" height="16" viewBox="0 0 12 12" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M11.208 5.417L7.50781 1.7168C7.18554 1.39453 6.66406 1.39453 6.34179 1.7168C6.01953 2.03907 6.01953 2.56055 6.34179 2.88282L8.63281 5.17481H1.375C0.918955 5.17481 0.549805 5.54395 0.549805 6.00001C0.549805 6.45607 0.918945 6.82521 1.375 6.82521H8.63281L6.34179 9.1172C6.01953 9.43947 6.01953 9.96095 6.34179 10.2832C6.50292 10.4444 6.71386 10.5254 6.9248 10.5254C7.13574 10.5254 7.34668 10.4444 7.50781 10.2832L11.208 6.58302C11.5303 6.26075 11.5303 5.73927 11.208 5.417Z" fill="#fff"/></svg>';
```
narcis-radu, line 4: I was checking the walkthrough shared in https://jira.corp.adobe.com/browse/MWPW-192744 and noticed that clicking the "view featured offer" would open a panel. I am not seeing that functionality as part of this PR

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.css
suhjainadobe: @narcis-radu : For smaller devices as well, the height of the marquee remains the same. Only the width will change based on the smaller devices.

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.css
suhjainadobe: By adding a variant, do you mean that we should make it authorable or add it directly via JS?

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.css
suhjainadobe: I have added an additional padding on the text separately and gave same padding on the button

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.css
suhjainadobe: done

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.js
suhjainadobe: Thanks for the review! Done

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.js
suhjainadobe: Initial intention was that the authors will not author a small heading on the page. But now, we are supporting it will a class which will style it irrespective of the heading type. So added the remaining headings as well.

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.js
suhjainadobe: @narcis-radu : It was confirmed that the button will be used an anchor link which when clicked will scroll the merch cards that are present on the lower section of the page.
I have added a comment on the Jira.

suhjainadobe:
@narcis-radu : Added support for viewport greater than 1920px. Could you please review.

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.css
```diff
@@ -0,0 +1,105 @@
+/* Product Marquee — mobile-first */
+
```
robert-bogos: super nit: I don't think the comment is necessary, the convention for C2 blocks is mobile first anyway

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.css
```diff
@@ -0,0 +1,105 @@
...
+    padding: var(--s2a-spacing-sm);
+    text-decoration: none;
+
+    .pm-promo-text {
+      color: var(--s2a-color-content-knockout);
+      flex: 1;
+      padding-inline-start: var(--s2a-spacing-2xs);
+    }
+
+    .pm-promo-chevron {
```
narcis-radu, line 54: For RTL, I _think_ we might want to rotate the chevron 180deg, but not a blocker.

narcis-radu (approved):

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.js
```diff
@@ -0,0 +1,47 @@
...
+    iconEl.classList.add('icon');
+    chicletRow.append(iconEl);
+  }
+  if (heading) chicletRow.append(heading);
+
+  const foreground = createTag('div', { class: 'pm-foreground' });
+  foreground.append(chicletRow, ...bodyEls);
+
+  const promoArea = createTag('div', { class: 'pm-promo-area' });
+  if (ctaLink) {
```
robert-bogos, line 33: another nit: `const promoArea = ...` should be moved inside the if block, otherwise we could end up creating the promoArea div for no reason, when ctaLink is not present.


```suggestion
  if (ctaLink) {
    const promoArea = createTag('div', { class: 'pm-promo-area' });
```

robert-bogos (approved):

zagi25:
I am not sure if design changed but shouldn't the chevron point down ? 
Screenshot 2026-05-21 at 15 40 59
https://www.figma.com/design/6FQTDUAttFbukvtskDAkPS/Product-%E2%80%94-A.com?node-id=7500-21264&m=dev

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.css
```diff
@@ -0,0 +1,105 @@
...
+      margin-block-start: 0;
+      min-width: 260px;
+    }
+  }
+}
+
+@media (width >= 1920px) {
+  .product-marquee-grid .pm-content {
+    margin-inline: auto;
+    max-width: 1920px;
```
overmyheadandbody: This caught my attention, along with https://github.com/adobecom/milo/pull/5925/changes#diff-d685df3f9ee0b3a77ba77d6427489c61eccf7d9420846e4e0159588383a92849R77 and https://github.com/adobecom/milo/pull/5925/changes#diff-d685df3f9ee0b3a77ba77d6427489c61eccf7d9420846e4e0159588383a92849R11. These should be tied to the grid system. You can either apply some variants to the block name, `container fixed` or directly via JS, the same classes. That should handle all of the grid complexity and keep things consistent.

overmyheadandbody (dismissed):
I'd like to avoid having a non-standard grid implementation before we merge this.

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.js
suhjainadobe: We are appending promoArea inside the outer div content. Thus, I had it outside the ctaLink check

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.css
suhjainadobe: Initial thought was that for media > 768, the container or fixed class would not work, as the padding is 48px. But, now I have used container class and overridden it for the devices specifically.

suhjainadobe:
> I am not sure if design changed but shouldn't the chevron point down ? Screenshot 2026-05-21 at 15 40 59 https://www.figma.com/design/6FQTDUAttFbukvtskDAkPS/Product-%E2%80%94-A.com?node-id=7500-21264&m=dev

Thanks for pointing this out. I referred to the first canvas in the following figma. Ill confirm with Erin which chevron to use.
https://www.figma.com/design/6FQTDUAttFbukvtskDAkPS/Product-%E2%80%94-A.com?node-id=5523-5055&p=f&m=dev

libs/mep/ace1205/product-marquee-grid/product-marquee-grid.css
overmyheadandbody: I would've hoped that we don't have to stray away from the standard grid, but we can get back to this in the future, at least the base functionality is now consolidated.

zagi25:
I see that `View featured offer` button doesn't have hover/active state, is this expected ?

SilviuLCF (approved):
Ready to merge, testing details https://jira.corp.adobe.com/browse/MWPW-192744

PR #5930, opened 2026-05-13
[MWPW-184676]:-  support mobile-rider embed in Marquee block for DA Events

hnvadobe:
Context: Adobe Events (e.g., Adobe MAX) are being migrated to Document Authoring (DA). The da-events repo already has a mobile-rider block that renders live streams powered by MobileRider.

The need: Event pages use the Marquee block as the hero area. For the live stream to appear there, the Marquee block in Milo needs to support a MobileRider embed in its asset area — so authors can simply place a MobileRider link inside the Marquee block in DA and have the live stream render correctly.

Resolves: [MWPW-184676] https://jira.corp.adobe.com/browse/MWPW-184676

**Test URLs:**
- After: https://mwpw-184676-mobile-rider--milo--adobecom.aem.page/?martech=off

Event Author Page:-  https://da.live/edit#/adobecom/da-events/drafts/hnv/mobile-rider-marquee-impl
Event Preview Page:- https://main--da-events--adobecom.aem.page/drafts/hnv/mobile-rider-marquee-impl?milolibs=mwpw-184676-mobile-rider

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -149,7 +149,7 @@ const AUTO_BLOCKS = [
   { 'merch-card-autoblock': 'mas.adobe.com/studio.html', styles: false },
   { m7: '/creativecloud/business-plans', styles: false },
   { m7: '/creativecloud/education-plans', styles: false },
-  {'mobile-rider': 'mobilerider.com'},
+  { 'mobile-rider': 'mobilerider.com' },
 ];
 const DO_NOT_INLINE = [
   'accordion',
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -149,6 +149,7 @@ const AUTO_BLOCKS = [
   { 'merch-card-autoblock': 'mas.adobe.com/studio.html', styles: false },
   { m7: '/creativecloud/business-plans', styles: false },
   { m7: '/creativecloud/education-plans', styles: false },
+  {'mobile-rider': 'mobile-rider.com'},
 ];
 const DO_NOT_INLINE = [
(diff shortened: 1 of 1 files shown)


libs/utils/utils.js
```diff
@@ -149,6 +149,7 @@ const AUTO_BLOCKS = [
   { 'merch-card-autoblock': 'mas.adobe.com/studio.html', styles: false },
   { m7: '/creativecloud/business-plans', styles: false },
   { m7: '/creativecloud/education-plans', styles: false },
+  { 'mobile-rider': 'mobilerider.com' },
```
narcis-radu, line 152: While I understand the reason behind this, I don't agree with the approach since the block lives in a totally different library: https://github.com/adobecom/event-libs/tree/dev/event-libs/v1/blocks/mobile-rider. `da-events` overrides the block library, I think the functionality needs to be extended to auto-blocks as well. See initial PR for details: https://github.com/adobecom/milo/pull/4833

narcis-radu (changes requested):

libs/utils/utils.js
qiyundai: Agreed with @narcis-radu . Event Libs has its own auto blocks processing. https://github.com/adobecom/event-libs/blob/1b51de3e0fb70a71230672f328facead69880654/event-libs/v1/utils/decorate.js#L560. Since we provide the block, we should handle the decoration from the event-libs too.

hnvadobe:
Auto blocks processing will be implemented in da-vents, as suggested.

PR #5932, opened 2026-05-13
[SR] - [MWPW-192883] - Rich Content Hero Jump Variant

DKos95:
Adds rich-content jump-link hero variant with parallax double garage door animation

Resolves: [MWPW-192883](https://jira.corp.adobe.com/browse/MWPW-192883)

**Test URLs:**
- Before: https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-offer?milolibs=site-redesign-foundation&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all
- After: https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-offer?milolibs=rich-content-hero&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all


libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -40,6 +40,108 @@
...
+      flex-direction: column;
+    }
+
+    .foreground .content {
+      padding-block-start: var(--s2a-layout-sm);
+      padding-inline: var(--s2a-spacing-lg);
+
+      .title-2,
+      .eyebrow {
+        max-width: 656px;
```
overmyheadandbody, line 67: Is this meant to be fixed or does it map to a specific number of columns?

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -80,12 +182,45 @@
...
+.section:has(.rich-content.jump-link.hero) {
+  height: 584px;
+}
+
+.section.has-background:has(.rich-content.jump-link.hero) .section-background picture::after {
+  background: linear-gradient(180deg, rgba(0, 0, 0, 0.8) 0%, rgba(0, 0, 0, 0) 39%, rgba(0, 0, 0, 0) 74%, rgba(0, 0, 0, 0.8) 100%);
+}
+
+.section:not(.has-background):has(.rich-content.jump-link.hero) {
+  background: linear-gradient(180deg, rgba(0, 0, 0, 0.8) 0%, rgba(0, 0, 0, 0) 39%, rgba(0, 0, 0, 0) 74%, rgba(0, 0, 0, 0.8) 100%), var(--s2a-color-background-default);
```
overmyheadandbody, line 193: Having specific colors in code might not be a good strategy for the long term. Can authoring be leveraged in any way here?

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -40,6 +40,108 @@
...
+        height: 12px;
+        background-color: currentColor;
+        mask-image: url('./assets/arrow.svg');
+        mask-repeat: no-repeat;
+        mask-size: contain;
+      }
+    }
+
+    .jump-link-label {
+      font-family: var(--s2a-font-family-title);
```
narcis-radu: `--s2a-font-family-title` var is not defined. I think we need to use `--s2a-font-family-heading`

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -97,6 +232,10 @@
     height: 1040px;
   }
 
+  .section:has(.rich-content.jump-link.hero) {
+    height: 814px;
```
narcis-radu: Can't the height be set to `auto` here, given the `.rich-content.jump-link.hero` element has the height defined? https://github.com/adobecom/milo/pull/5932/changes#diff-a4c2a70eb76db2764bd78e2d192053573b2db476ad448df3e76be16f4ab55ebdR254

libs/mep/ace1205/rich-content/rich-content.css
DKos95: We can but this is just to override the wave 1 linear background being applied also via code, if we changed it here it would make sense to change it for the wave 1 code and make it authorable, not sure if we should invest time into that currently, but I do agree we should do it at one point.

libs/mep/ace1205/rich-content/rich-content.css
DKos95: Not sure, what do you think?

Viewport | Content width | Max-width | Approx columns
-- | -- | -- | --
768px+ | 768 − 64 = 704px | 530px | 530/704 ≈ 9/12
1024px+ | 1024 − 80 = 944px | 656px | 656/944 ≈ 8.3/12
1440px+ | 1440 − 80 = 1360px | 656px | 656/1360 ≈ 5.8/12

libs/c2/styles/styles.css
```diff
@@ -1238,6 +1238,29 @@ p {
...
+
+  .section:is(.parallax-double-garage-door, :has(.parallax-double-garage-door)) {
+    overflow: clip;
+    animation: dg-top-door ease-out both, dg-exit ease-out both;
+    animation-timeline: view(), view();
+    animation-range: entry 0% entry 70%, exit 0% exit 90%;
+    animation-composition: replace, add;
+  }
+
+  @keyframes dg-top-door {
```
rgclayton, line 1254: dg = double garage door?

rgclayton (approved):

libs/c2/styles/styles.css
DKos95: Correct

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -40,6 +40,108 @@
...
+    }
+
+    .eyebrow:has(picture, img) {
+      display: flex;
+      align-items: center;
+      gap: var(--s2a-spacing-xs);
+
+      picture,
+      img {
+        width: 24px;
```
zagi25: here we should be able to use tokens

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -40,6 +40,108 @@
...
+      background: var(--s2a-color-transparent-white-16);
+      border: 1px solid var(--s2a-color-transparent-white-12);
+      transition: background-color 0.2s ease, color 0.2s ease;
+      flex-shrink: 0;
+      color: var(--s2a-color-content-knockout);
+
+      &::after {
+        content: '';
+        display: block;
+        width: 12px;
```
zagi25: tokens ?

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -40,6 +40,108 @@
...
+        mask-size: contain;
+      }
+    }
+
+    .jump-link-label {
+      font-family: var(--s2a-font-family-heading);
+      font-size: var(--s2a-font-size-lg);
+      line-height: var(--s2a-typography-line-height-heading-4);
+      font-weight: var(--s2a-font-weight-heading);
+      letter-spacing: var(--s2a-font-letter-spacing-4xl);
```
overmyheadandbody: These seem a bit disconnected. Are these the tokens used in Figma as well? Could they be mapped closer to a single element, like `heading-4`? `--s2a-typography-line-height-heading-4` will scale depending on viewport, while `--s2a-font-letter-spacing-4xl` is a primitive and is fixed across breakpoints. This looks like an omission somehwere.

libs/mep/ace1205/rich-content/rich-content.css
overmyheadandbody: I see. Ok, I've opened https://jira.corp.adobe.com/browse/MWPW-195995 to track this.

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -40,6 +40,107 @@
...
+
+    .eyebrow:has(picture, img) {
+      display: flex;
+      align-items: center;
+      gap: var(--s2a-spacing-xs);
+
+      picture,
+      img {
+        width: var(--s2a-spacing-24);
+        height: var(--s2a-spacing-24);
```
overmyheadandbody: I'm not convinced we should attach block/inline sizes to spacing tokens. I know this was suggested in an earlier review, and overall using primitives is likely safe, since they shouldn't change. But saying width correlates to some particular spacing is likely not a good long term practice. I'd also say we should stick to a consistent pattern, not use tokens in one place and fixed values in another, like https://github.com/adobecom/milo/pull/5932/changes#diff-a4c2a70eb76db2764bd78e2d192053573b2db476ad448df3e76be16f4ab55ebdR114-R115. We don't have icon tokens, so using explicit values would be fine.

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -80,12 +181,50 @@
...
+      .eyebrow { max-width: 530px; }
+    }
+
+    .jump-links {
+      flex-direction: row;
+      gap: var(--s2a-spacing-xl);
+      align-items: center;
+    }
+
+    .jump-link-label { font-size: var(--s2a-typography-font-size-heading-4); }
```
overmyheadandbody: Why does this have to be repeated, since it's already declared in https://github.com/adobecom/milo/pull/5932/changes#diff-a4c2a70eb76db2764bd78e2d192053573b2db476ad448df3e76be16f4ab55ebdR137?

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -109,6 +248,30 @@
...
+
+        .heading-2,
+        .eyebrow { max-width: 656px; }
+      }
+
+      .jump-links { padding-block-end: var(--s2a-layout-sm); }
+
+      .jump-link-label {
+        font-size: var(--s2a-typography-font-size-heading-5);
+        line-height: var(--s2a-typography-line-height-heading-5);
```
overmyheadandbody: Switching between heading sizes across viewports doesn't look right to me. Is this what the designs suggest? Additionally, it might be a good idea that if this is actually the intent, letter-spacing should be adapted as well.

libs/mep/ace1205/rich-content/rich-content.css
DKos95: Removed.

libs/mep/ace1205/rich-content/rich-content.css
DKos95: This was a mistake on my part when looking at design, heading-5 is accurate if looking at pixel in figma but token changes haven't gone into figma so leaving heading-4 is the right call I guess.

libs/mep/ace1205/rich-content/rich-content.css
narcis-radu: We can iterate if needed.

narcis-radu (approved):

zagi25 (approved):

robert-bogos (approved):

NadiiaSokolova (approved):
Verified. Ready to be merged.
Testing details [MWPW-192883](https://jira.corp.adobe.com/browse/MWPW-192883)

PR #5934, opened 2026-05-13
MWPW-195071: adjusted cc and dc mapping to be accounted for in lingo

sheridansunier:
When you send dc or cc through the bulkpublisher, it converts the origin to "doccloud" or "hawks" respectively.  This mapping needs to be done ahead of Lingo Lookups generally to avoid potential content issues in bulk published lingo content to these two origins.

Resolves: [MWPW-195071](https://jira.corp.adobe.com/browse/MWPW-195071)

**Test URLs:**
- After: https://MWPW-195071--milo--sheridansunier.aem.page/?martech=off
 

Testing:

- using source overrides, send a sample page to stage through the bulkpublisher
 Screenshot 2026-05-13 at 4 20 54 PM

- verify that if you send [this page](https://www.adobe.com/fr/acrobat/roc/business/reports/unlock-value-genai-document-mgmt-functional-leader) without the language first tag, you get "fr" and "fr" for country/language respectively
- verify that we receive "xx" and "fr" for a page like this being sent to stage w language first.

Please leave that page sent w language first on stage to keep it in the state it was before you started doing your testing.

### DIFF
--- a/tools/send-to-caas/send-utils.js
+++ b/tools/send-to-caas/send-utils.js
@@ -198,11 +198,7 @@ export const getOrigin = (fgColor) => {
   const { project, repo } = getConfig();
   const origin = project || processRepoForFloodgate(repo, fgColor);
 
-  const mappings = {
-    cc: 'hawks',
-    dc: 'doccloud',
-  };
-  const originLC = mappings[origin.toLowerCase()] || origin;
+  const originLC = LANG_FIRST_SOURCE_MAPPINGS[origin.toLowerCase()] || origin;
   if (originLC) {
     return originLC;
   }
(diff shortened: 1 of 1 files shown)


cmiqueo (approved):

tools/send-to-caas/send-utils.js
```diff
@@ -359,7 +362,11 @@ const getCountryAndLang = async (options, origin) => {
   if (langFirst) {
     const isBulkPublisher = window.location.pathname.includes('/tools/send-to-caas/bulkpublisher');
     const fqdn = isBulkPublisher ? 'bulkpublisher' : window.location.hostname;
-    return getLanguageFirstCountryAndLang(window.location.pathname, origin, fqdn);
+    return getLanguageFirstCountryAndLang(
+      window.location.pathname,
+      LANG_FIRST_SOURCE_MAPPINGS[origin.toLowercase()] || origin,
```
sanrai: Runtime Error: Should be `origin.toLowerCase()`

sanrai (changes requested):

tools/send-to-caas/send-utils.js
sheridansunier: good catch ty

sanrai (approved):

shkhan91 (approved):

PR #5940, opened 2026-05-15
Brand Concierge - Floating Scroll Optimization and Refactor

colloyd:
* Move Floating scroll functionality out of the floating button decorate function to make it more generically useful
* Optimize the scroll handler for floating elements
* Add the ability to define custom delay for floating element anchor using a variant: floating-anchor-delay[amount]

Resolves: [MWPW-194524](https://jira.corp.adobe.com/browse/MWPW-194524), [MWPW-194260](https://jira.corp.adobe.com/browse/MWPW-194260)

**Test URLs:**
- Before: https://stage--milo--adobecom.aem.page/drafts/colloyd/brand-concierge/bc-floating-button-delay
- After: https://bc-scroll-opt--milo--colloyd.aem.page/drafts/colloyd/brand-concierge/bc-floating-button-delay

Testing notes: 
The "after" page demonstrates custom delay for floating element anchor positions as noted above. The amount of top floating delay is 100 pixels, the bottom or anchor delay is set to 450, so the button will hide at 450 above the footer.


libs/blocks/brand-concierge/brand-concierge.js
```diff
@@ -34,6 +34,99 @@ const webClientVersion = params.get('webclientversion');
...
+      if (variants.isFloatingAnchorHide || variants.floatingAnchorDelay) {
+        hideFloating();
+      } else {
+        floatingSpacer.style.cssText = `height: ${targetHeight}px; pointer-events: none; display: block;`;
+      }
+    } else {
+      showFloating();
+      target.style.bottom = '0';
+    }
+    if (variants.isHero || variants.floatingDelay || variants.floatingAnchorDelay) {
```
JasonHowellSlavin: We could probably pull this || statement into a const for readability: 
`const hasDelayVariant = variants.isHero || variants.floatingDelay || variants.floatingAnchorDelay` 
Since there are some if's below it, it might help with understanding what we are trying to trigger.

libs/blocks/brand-concierge/brand-concierge.js
```diff
@@ -34,6 +34,99 @@ const webClientVersion = params.get('webclientversion');
...
+
+  const ro = new ResizeObserver((entries) => {
+    for (const entry of entries) {
+      const size = Math.floor(entry.borderBoxSize?.[0]?.blockSize);
+      if (entry.target === el) {
+        elHeight = size ?? el.scrollHeight;
+      } else if (entry.target === mainElement) {
+        mainHeight = size ?? mainElement.scrollHeight;
+      } else if (entry.target === targetEl) {
+        targetHeight = getTargetHeight(targetEl);
```
JasonHowellSlavin: Looks like a good candidate for a switch statement. Or could just remove the `else if`s and just use if.

JasonHowellSlavin (approved):
Two nits, looks good otherwise!

JackySun9 (approved):

libs/blocks/brand-concierge/brand-concierge.js
colloyd: looks like a good candidate for having Claude do it for me :) 

This is a direct copy from the proposed changes that were given for optimization. The switch is nicer.

libs/blocks/brand-concierge/brand-concierge.js
colloyd: Done. Good suggestion.

PR #5955, opened 2026-05-19
[MWPW-195790] caas configurator tags fetch error fix

jedjedjedM:
* Fix for "Unable to fetch tags, loading backup tags" error on CaaS configurator

Resolves: [MWPW-195790](https://jira.corp.adobe.com/browse/MWPW-195790)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/?not-applicable
- After: https://MWPW-195790--milo--adobecom.aem.page/?not-applicable

### DIFF
--- a/libs/blocks/caas/utils.js
+++ b/libs/blocks/caas/utils.js
@@ -296,9 +296,12 @@ export const loadCaasFiles = async () => {
 export const loadCaasTags = async (tagsUrl) => {
   let errorMsg = '';
   if (tagsUrl) {
-    const url = tagsUrl.startsWith('https://') || tagsUrl.startsWith('http://') ? tagsUrl : `https://${tagsUrl}`;
+    const urlObj = new URL(tagsUrl.startsWith('https://') || tagsUrl.startsWith('http://') ? tagsUrl : `https://${tagsUrl}`);
+    if (urlObj.hostname !== window.location.hostname && !urlObj.searchParams.has('s')) {
+      urlObj.searchParams.set('s', window.location.host);
+    }
     try {
-      const resp = await fetchWithTimeout(url);
+      const resp = await fetchWithTimeout(urlObj.toString());
       if (resp.ok) {
         const json = await resp.json();
         return {
--- a/test/blocks/caas-config/mockFetch.js
+++ b/test/blocks/caas-config/mockFetch.js
@@ -1,13 +1,13 @@
-import { stub } from 'sinon';
+import { stub, match } from 'sinon';
 import tagsData from './tagsData.js';
 
-const CAAS_TAG_URL = 'https://www.adobe.com/chimera-api/tags';
+const CAAS_TAGS_URL_BASE = 'https://www.adobe.com/chimera-api/tags';
 
 const ogFetch = window.fetch;
 
 export const stubFetch = () => {
   window.fetch = stub();
-  window.fetch.withArgs(CAAS_TAG_URL).returns(
+  window.fetch.withArgs(match((url) => url.startsWith(CAAS_TAGS_URL_BASE))).returns(
     new Promise((resolve) => {
       resolve({
         ok: true,


cmiqueo (approved):

sanrai (commented):
The fix here works, but I think the real root cause is one line in chimera-io rather than the Configurator.

In `actions/response.js` (lines 60-69) the `vary: 'origin'` directive is nested inside the `isOriginAllowed()` conditional:

```javascript
const getCorsResponseHeaders = () => ({
    ...(requestProcessor.isOriginAllowed() && {
        'access-control-allow-methods': requestProcessor.headers.acReqMethod,
        'access-control-allow-origin': requestProcessor.headers.origin,
        'access-control-allow-headers': requestProcessor.headers.acReqHeaders,
        'access-control-allow-credentials': true,
        vary: 'origin',          // only emitted when origin is allow-listed
    }),
});
```

When a request arrives with no Origin header, the conditional short-circuits and no Vary directive ships. That response gets cached and reused for subsequent cross-origin Configurator fetches, which fail CORS.

Akamai itself is fine, its property is already keying by Origin internally. You can verify with the cache-key debug pragma:

```bash
PRAGMA='Pragma: akamai-x-cache-on, akamai-x-get-cache-key'

# CASE 1: No Origin (the poisoning condition)
curl -sI 'https://www.adobe.com/chimera-api/tags' -H "$PRAGMA" \
  | grep -iE 'x-cache-key|vary|access-control'

# Output:
#   x-cache-key: ... cid=///
#   vary: Accept-Encoding                       <-- Origin missing from Vary

# CASE 2: With Origin (the clean path)
curl -sI 'https://www.adobe.com/chimera-api/tags' -H "$PRAGMA" \
  -H 'Origin: https://milo.adobe.com' \
  | grep -iE 'x-cache-key|vary|access-control'

# Output:
#   x-cache-key: ... cid=//Origin=https%3a%2f%2fmilo.adobe.com/
#   vary: Accept-Encoding, Origin               <-- Origin is in Vary
#   access-control-allow-origin: https://milo.adobe.com
```

The fix would be hoisting `vary: 'origin'` out of the conditional so it ships on every response:

```javascript
const getCorsResponseHeaders = () => ({
    vary: 'origin',              // always emit
    ...(requestProcessor.isOriginAllowed() && {
        'access-control-allow-methods': requestProcessor.headers.acReqMethod,
        'access-control-allow-origin': requestProcessor.headers.origin,
        'access-control-allow-headers': requestProcessor.headers.acReqHeaders,
        'access-control-allow-credentials': true,
    }),
});
```

My suggestion is to ship the chimera-io fix first since it addresses the bug at the source.

If we keep this PR as a tactical bridge while that lands, every other fetch to this URL in milo needs the same suffix treatment, including any new ones added going forward. Three sibling files have the bare URL today (verified below):

```bash
# tag-selector (hosted at milo.adobe.com/tools/tag-selector per sidekick config)
curl -s --compressed 'https://main--milo--adobecom.aem.live/libs/blocks/tag-selector/tag-selector.js' \
  | grep -n 'chimera-api/tags'
# 57:  const caasTagUrl = 'https://www.adobe.com/chimera-api/tags';

# caas-import
curl -s --compressed 'https://main--milo--adobecom.aem.live/tools/caas-import/parseCaasConfig.js' \
  | grep -n 'chimera-api/tags'
# 64:    tagsUrl: 'www.adobe.com/chimera-api/tags',
# 252:    tagsUrl: 'www.adobe.com/chimera-api/tags',

# caas runtime block (runs on every consumer page)
curl -s --compressed 'https://main--milo--adobecom.aem.live/libs/blocks/caas/utils.js' \
  | grep -n 'chimera-api/tags'
# 1329:  tagsUrl: 'www.adobe.com/chimera-api/tags',
```

shkhan91 (approved):

jedjedjedM:
> The fix would be hoisting `vary: 'origin'` out of the conditional so it ships on every response:

Akamai strips this header it is not an accept-encoding value.  this is a known quirk/issue with akamai.  See:
https://techdocs.akamai.com/property-mgr/docs/rm-vary-header

Screenshot 2026-05-20 at 11 11 44 AM

In short, the `vary: origin` will not work with our current CDN configuration.

meganthecoder (approved):

jedjedjedM:
@sanrai re-worked this a bit. tag selector / configurator / collection load it with a query param if the url does not match window.location.hostname.  Missing other use-cases?

sheridansunier:
lgtm - verified using local overrides :)

sheridansunier (approved):

PR #5965, opened 2026-05-21
MWPW-195844 Hide VAT labels from strikethrough pricing on milo cards

bozojovicic:
Hide labels on ST prices when directly followed with promo price, some empty text in between is ignored.

Related MAS PR https://github.com/adobecom/mas/pull/899

Test pages https://mwpw195844--milo--bozojovicic.aem.page/drafts/bozo/st-prices?maslibs=MWPW-195844

https://main--da-cc--adobecom.aem.live/fr/cc-shared/fragments/merch/products/photoshop/mini-compare/creativecloud/individual/intro-pricing?maslibs=MWPW-195844&milolibs=mwpw195844--milo--bozojovicic

Resolves: [MWPW-195844](https://jira.corp.adobe.com/browse/MWPW-195844)

**Test URLs:**
- After: https://mwpw195844--milo--bozojovicic.aem.page/?martech=off

### DIFF
--- a/libs/blocks/merch/merch.css
+++ b/libs/blocks/merch/merch.css
@@ -135,3 +135,7 @@ a[is='checkout-link'].con-button > span {
   top: 50%; 
   transform: translate(-50%,-50%);
 }
+
+.price-strikethrough.price-labels-hidden span:is(.price-tax-inclusivity, .price-unit-type) {
+  display: none;
+}


yesil (changes requested):
this is not solving the problem at the root and it is solving only for Milo, we need to remain consistent on inline-price behaviour.

In this specific case, authors could add `&tax=false` to stirkethough price.

Image

yesil:
@bozojovicic also if they use the display old price on the promo one, they get the right experience:

for `<span is="inline-price" data-display-old-price="true" data-promotion-code="CCPROIP_50_3M_WW" data-quantity="1" data-template="price" data-wcs-osi="r_JXAnlFI7xD6FxWKl2ODvZriLYBoSL701Kd1hRyhe8"></span>`

image

bozojovicic:
Not needed. Everything is implemented in MAS. Closing this one.

PR #5969, opened 2026-05-22
[SR] Offer hero

robert-bogos:
## Description
This PR is adding the Offer Hero block

## Related Issue
Resolves: [MWPW-192882](https://jira.corp.adobe.com/browse/MWPW-192882)

## Test URL
https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-offer?milolibs=192882&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all


libs/mep/ace1205/offer-hero/offer-hero.css
```diff
@@ -0,0 +1,235 @@
+.offer-hero {
+  margin-inline: calc(50% - 50vw);
```
narcis-radu: I don't see this changing anything.

libs/mep/ace1205/offer-hero/offer-hero.css
```diff
@@ -0,0 +1,235 @@
...
+
+  .hero {
+    position: relative;
+    height: 100vh;
+    z-index: 1;
+  }
+
+  .hero-copy {
```
narcis-radu: Would it be possible to rename as `hero-content` ? This way we'll be consistent with existing blocks

libs/mep/ace1205/offer-hero/offer-hero.css
```diff
@@ -0,0 +1,235 @@
...
+    flex-direction: column;
+    align-items: center;
+    text-align: center;
+    will-change: transform, opacity;
+
+    .eyebrow {
+      color: var(--s2a-color-content-default);
+    }
+
+    :is(h1, h2, h3, h4, h5, h6) {
```
narcis-radu, line 24: If I add additional content, it will stay behind the cards. I don't expect this much content to be added, but is there a way to dynamically push the cards below?
Image

libs/mep/ace1205/offer-hero/offer-hero.css
```diff
@@ -0,0 +1,235 @@
...
+    max-width: var(--grid-max-width-default);
+    margin-inline: auto;
+    padding-inline: var(--s2a-spacing-lg);
+
+    &.two-up {
+      align-items: stretch;
+    }
+  }
+
+  .hero-card {
```
narcis-radu, line 91: cards look bad on RTL

libs/mep/ace1205/offer-hero/offer-hero.js
```diff
@@ -0,0 +1,316 @@
...
+      const videoEl = createTag('video', {
+        src: videoSrc,
+        poster: posterImg.getAttribute('src') || '',
+        playsinline: '',
+        muted: '',
+        preload: 'metadata',
+        tabindex: '-1',
+        'aria-hidden': 'true',
+      });
+      videoEl.muted = true;
```
narcis-radu, line 108: can't it be part of the `createTag` declaration?

libs/mep/ace1205/offer-hero/offer-hero.js
```diff
@@ -0,0 +1,316 @@
...
+    if (row.children.length !== 2) return;
+    const [mediaCell, textCell] = row.children;
+
+    decorateBlockText(textCell, { heading: '6', body: 'md' });
+    const heading = textCell.querySelector(':is(h1, h2, h3, h4, h5, h6)');
+    heading?.classList.add('hero-card-title');
+    const paras = [...textCell.querySelectorAll('p')];
+    paras.forEach((p) => p.classList.add('hero-card-body-text'));
+
+    const media = createTag('div', { class: 'hero-card-media' });
```
narcis-radu, line 93: Would be better to use the agreed authoring pattern, with text on the left, images on the right. Right now it is the opposite.

Image

libs/mep/ace1205/offer-hero/offer-hero.js
```diff
@@ -0,0 +1,316 @@
...
+  }
+
+  setup();
+
+  const io = new IntersectionObserver(([entry]) => {
+    if (entry.isIntersecting) {
+      if (!running) { running = true; tick(); }
+    } else {
+      running = false;
+      cancelAnimationFrame(rafId);
```
narcis-radu, line 342: You need to disconnect the resize observer when the block is out of view

libs/mep/ace1205/offer-hero/offer-hero.js
```diff
@@ -0,0 +1,316 @@
...
+    eyebrow = block.querySelector('.hero-eyebrow');
+    cards = [...block.querySelectorAll('.hero-card')];
+    tiles = [...block.querySelectorAll('.hero-card-tile')];
+    medias = [...block.querySelectorAll('.hero-card-media')];
+    videos = [...block.querySelectorAll('.hero-card-media video')];
+    cardTexts = [...block.querySelectorAll('.hero-card-title, .hero-card-body-text, .learn-more')];
+  }
+
+  refreshRefs();
+  if (tiles.length !== 4) return;
```
narcis-radu, line 173: What happens with more than 4 cards ?

libs/mep/ace1205/offer-hero/offer-hero.js
```diff
@@ -0,0 +1,316 @@
...
+        poster: posterImg.getAttribute('src') || '',
+        playsinline: '',
+        muted: '',
+        preload: 'metadata',
+        tabindex: '-1',
+        'aria-hidden': 'true',
+      });
+      videoEl.muted = true;
+      videoEl.addEventListener('loadedmetadata', () => {
+        try { videoEl.currentTime = 0.001; } catch (e) { /* */ }
```
narcis-radu, line 110: I see an issue with this, here's how to reproduce:
1. scroll down a bit and see content of cards
2. cards start playing video
3. scroll up to top
4. scroll down a bit

cards background is different

Image

Image

narcis-radu (dismissed):

libs/mep/ace1205/offer-hero/offer-hero.css
robert-bogos: True, it got left over from an earlier iteration

libs/mep/ace1205/offer-hero/offer-hero.css
robert-bogos: Done 👍

libs/mep/ace1205/offer-hero/offer-hero.css
robert-bogos: Definitely can dynamically push cards below. Made the change 👍  
FYI: if you want to test again, you need to have the lengthly content authored, not overwritten with dev tools

libs/mep/ace1205/offer-hero/offer-hero.css
robert-bogos: Did not know we need to support RTL for the test. I've added support 👍

libs/mep/ace1205/offer-hero/offer-hero.js
robert-bogos: I tried that initially, but we need to set the JS muted property to avoid autoplay being blocked by the browser. createTag only uses setAttribute, which causes autoplay to fail.

libs/mep/ace1205/offer-hero/offer-hero.js
robert-bogos: Disconnecting the observer will cause issues if the viewport changes while the block is off screen. The main issue is that animations will point to detached nodes.

libs/mep/ace1205/offer-hero/offer-hero.js
robert-bogos: If more than 4 cards they'll just go directly to the grid position.

libs/mep/ace1205/offer-hero/offer-hero.js
robert-bogos: That is expected. At least, this is how the prototype worked: cards' videos only play once, then they stay on the final frame.

libs/mep/ace1205/offer-hero/offer-hero.js
robert-bogos: Made the change to support text on the left, images on the right. 
I used a skill to create the authored doc and did not notice the text/image reversed order. 😬

libs/mep/ace1205/offer-hero/offer-hero.js
robert-bogos: I will confirm again with design 👍

libs/mep/ace1205/offer-hero/offer-hero.css
```diff
@@ -0,0 +1,238 @@
...
+    display: flex;
+  }
+
+  .hero-card-tile {
+    position: relative;
+    flex: 1;
+    display: flex;
+    flex-direction: column;
+    justify-content: flex-end;
+    background: var(--s2a-color-gray-75, #f5f5f5);
```
DKos95: Do we need fallback values for colors?

libs/mep/ace1205/offer-hero/offer-hero.css
```diff
@@ -0,0 +1,238 @@
...
+    border-radius: var(--s2a-border-radius-md);
+    box-shadow: inset 0 0 0 2px rgb(255 255 255 / 10%);
+    overflow: hidden;
+    transform-origin: top left;
+    padding-top: calc(100% * 100 / 109);
```
DKos95, line 107: I'd like to understand this calculation :)

libs/mep/ace1205/offer-hero/offer-hero.css
```diff
@@ -0,0 +1,238 @@
...
+        max-width: calc(100% - 438px);
+      }
+    }
+
+    .hero-eyebrow {
+      --hero-eyebrow-gap: var(--s2a-spacing-2xl);
+    }
+
+    .hero-card-tile {
+      padding-top: calc(100% * 123 / 152);
```
DKos95, line 237: Also would like to understand this.

libs/utils/utils.js
```diff
@@ -121,6 +121,7 @@ const C2_BLOCKS = [
   'martech-metadata',
   'modal',
   'news',
+  'offer-hero',
```
DKos95: Is this needed if the block is in mep?

libs/mep/ace1205/offer-hero/offer-hero.js
```diff
@@ -0,0 +1,322 @@
...
+    positionContent();
+    computeLayouts();
+    updateEyebrow();
+  }
+
+  setup();
+
+  const io = new IntersectionObserver(([entry]) => {
+    if (entry.isIntersecting) {
+      if (!running) { running = true; tick(); }
```
DKos95: Can we merge the two if's to not use nested ifs and then early return instead of else?

libs/mep/ace1205/offer-hero/offer-hero.css
robert-bogos: Sure! `100 / 109` is the aspect ratio taken from figma, and `padding-top: calc(100% * 100 / 109);` is applying that aspect ratio to the upper part of the card

libs/mep/ace1205/offer-hero/offer-hero.css
robert-bogos: same reasoning as above

libs/mep/ace1205/offer-hero/offer-hero.js
robert-bogos: Good suggestion 👍 I made the change

libs/utils/utils.js
robert-bogos: no, not needed at the moment. It will be needed when we move the block to the c2 folder, and I was using it to test on my own draft page.
Anyway, I'll remove it and add it back when we actually need it.

libs/mep/ace1205/offer-hero/offer-hero.js
```diff
@@ -0,0 +1,324 @@
...
+      willChangeOn = animating;
+      tiles.forEach((tile) => { tile.style.willChange = animating ? 'transform' : ''; });
+    }
+  }
+
+  function updateEyebrow() {
+    if (!eyebrow) return;
+    if (medias[0]) {
+      const gap = parseFloat(getComputedStyle(eyebrow).getPropertyValue('--hero-eyebrow-gap')) || 24;
+      eyebrow.style.top = `${medias[0].getBoundingClientRect().top - eyebrow.offsetHeight - gap}px`;
```
zagi25: If I am not mistaken this is used to move `What you need is what you get.` header right ? 
Is it possible to use `translateY` instead of `top` ? Updating `top` creates small CLS `0.0002 - 0.0016`, but if you can use translateY we should be able to eliminate it completely  and leave a little bit of performance space for animations that we can't move to compositor.

libs/mep/ace1205/offer-hero/offer-hero.js
```diff
@@ -0,0 +1,324 @@
...
+    if (running) return;
+    running = true;
+    tick();
+  }, { rootMargin: '200px 0px' });
+  io.observe(block);
+
+  let resizeTimer;
+  const reSetup = () => { clearTimeout(resizeTimer); resizeTimer = setTimeout(setup, 120); };
+  new ResizeObserver(reSetup).observe(block);
+  window.addEventListener('resize', reSetup);
```
zagi25, line 354: Why do we need both `ResizeObserver` and `resize` listener ?

zagi25:
I noticed that `Lear more` chevron is smaller than the font size
Screenshot 2026-05-26 at 11 18 05

Also, header that fades in, at the beginning of the animation, overlaps cards (if this is intended, please ignore)
Screenshot 2026-05-26 at 11 29 42

libs/mep/ace1205/offer-hero/offer-hero.css
robert-bogos: no, we don't need. Removed that 👍

robert-bogos:
> I noticed that `Lear more` chevron is smaller than the font size Screenshot 2026-05-26 at 11 18 05
> 
> Also, header that fades in, at the beginning of the animation, overlaps cards (if this is intended, please ignore) Screenshot 2026-05-26 at 11 29 42

The header overlapping thing is expected. Is because of the figma spec spacing.

Good catch on the "Learn more" chevron. Adjusted to match the font size 👍

libs/mep/ace1205/offer-hero/offer-hero.js
robert-bogos: They detect events on different scopes. `ResizeObserver` detects changes in the block's box size, `resize listener` detects viewport changes.

We need the `resize listener` for when content is swapped between breakpoints, so that animations use the new references.
We need the `ResizeObserver` because C2 CSS is loaded async. Without it, animations would run on wrong positions.

libs/mep/ace1205/offer-hero/offer-hero.js
robert-bogos: Good point! `translateY` is much better. Made the change 👍

libs/mep/ace1205/offer-hero/offer-hero.js
```diff
@@ -0,0 +1,329 @@
...
+  }
+
+  function tick() {
+    if (!running) return;
+    const progress = getProgress();
+    applyAnimation(progress);
+    updateEyebrow();
+
+    if (progress >= 0.85) setSettled(true);
+    else if (isSettled) setSettled(false);
```
DKos95: Maybe we could shorten this to `if (progress >= 0.85 || isSettled) setSettled(progress >= 0.85);`?

libs/mep/ace1205/offer-hero/offer-hero.js
```diff
@@ -0,0 +1,329 @@
...
+  }
+
+  function updateEyebrow() {
+    if (!eyebrow) return;
+    if (medias[0]) {
+      const gap = parseFloat(getComputedStyle(eyebrow).getPropertyValue('--hero-eyebrow-gap')) || 24;
+      const offset = medias[0].getBoundingClientRect().top - eyebrow.offsetHeight - gap;
+      eyebrow.style.transform = `translateY(${offset}px)`;
+    }
+    if (heroContent) {
```
DKos95: early return possible

libs/mep/ace1205/offer-hero/offer-hero.js
```diff
@@ -0,0 +1,329 @@
...
+      }
+    });
+
+    cardTexts.forEach((textEl, i) => {
+      const textProgress = clamp01((progress - (0.6 + i * 0.025)) / 0.2);
+      textEl.style.opacity = String(1 - (1 - textProgress) * (1 - textProgress));
+    });
+
+    const animating = progress > 0.001 && progress < 0.999;
+    if (animating !== willChangeOn) {
```
DKos95: early return possible

libs/mep/ace1205/offer-hero/offer-hero.js
```diff
@@ -0,0 +1,329 @@
...
+      const sx = lerp(stack.w / natural.w, 1, progress);
+      const sy = lerp(stack.h / natural.h, 1, progress);
+      const txTarget = rtl
+        ? (stack.x + stack.w) - (natural.x + natural.w)
+        : stack.x - natural.x;
+      const tx = lerp(txTarget, 0, progress);
+      const ty = lerp(stack.y - natural.y, 0, progress);
+      const rot = lerp(stack.rot, 0, progress);
+      tile.style.transform = `translate(${tx}px, ${ty}px) rotate(${rot}deg) scale(${sx}, ${sy})`;
+      if (CARD_SHADOWS[i]) {
```
DKos95: early return possible

libs/mep/ace1205/offer-hero/offer-hero.js
```diff
@@ -0,0 +1,329 @@
...
+  function setSettled(next) {
+    if (isSettled === next) return;
+    isSettled = next;
+    if (next) {
+      videoObserver = new IntersectionObserver((entries) => {
+        entries.forEach((entry) => {
+          const video = videos[medias.indexOf(entry.target)];
+          if (!video || video.ended) return;
+          if (entry.isIntersecting) video.play()?.catch(() => {});
+          else video.pause();
```
DKos95: `entry.isIntersecting ? video.play()?.catch(() => {}) : video.pause();`?

libs/mep/ace1205/offer-hero/offer-hero.js
narcis-radu: We're going to have performance issues

libs/mep/ace1205/offer-hero/offer-hero.js
robert-bogos: even better, I just did `setSettled(progress >= 0.85);`

libs/mep/ace1205/offer-hero/offer-hero.js
robert-bogos: yes, can be. Did the change

libs/mep/ace1205/offer-hero/offer-hero.js
robert-bogos: I would stick to the `if else` block in here. Ternary is mostly used for assigning values, ESLint will complain if used for function calls

libs/mep/ace1205/offer-hero/offer-hero.css
```diff
@@ -0,0 +1,252 @@
...
+      opacity: 1;
+    }
+  }
+
+  .hero-cards {
+    max-width: var(--grid-max-width-default);
+    margin-inline: auto;
+    padding-inline: var(--s2a-spacing-lg);
+
+    &.two-up {
```
overmyheadandbody: This is usually applied at the section level, meaning you're inheriting some styles. This is alos traditionally authored by GWP, I'm assuming this is set from code in this context, right? No real action items, but might be worth a double check that things won't break if the `.X-up` pattern is updated.

libs/mep/ace1205/offer-hero/offer-hero.css
```diff
@@ -0,0 +1,252 @@
...
+
+  .hero-card-tile {
+    position: relative;
+    flex: 1;
+    display: flex;
+    flex-direction: column;
+    justify-content: flex-end;
+    background: var(--s2a-color-gray-75);
+    border-radius: var(--s2a-border-radius-md);
+    box-shadow: inset 0 0 0 2px rgb(255 255 255 / 10%);
```
overmyheadandbody, line 104: `--s2a-color-transparent-white-08` or `--s2a-color-transparent-white-12`?

libs/mep/ace1205/offer-hero/offer-hero.css
```diff
@@ -0,0 +1,252 @@
...
+
+  .hero-card-body-text {
+    color: var(--s2a-color-content-body-subtle);
+  }
+
+  .learn-more {
+    color: var(--s2a-color-content-label);
+    font-weight: var(--s2a-font-weight-label);
+    font-size: var(--s2a-typography-font-size-label);
+    line-height: var(--s2a-typography-line-height-label);
```
overmyheadandbody: Would it be easier to simply apply a `.label` class via JS?

libs/mep/ace1205/offer-hero/offer-hero.js
```diff
@@ -0,0 +1,325 @@
+import { decorateBlockText, decorateViewportContent } from '../../../utils/decorate.js';
+import { createTag, getFederatedUrl } from '../../../utils/utils.js';
+
+const STACK_W = 836;
+const STACK_REF = [
+  { dx: 90.94, dy: 116.93, w: 418.531, h: 489.484, rot: -6 },
+  { dx: 202.94, dy: 0, w: 535.885, h: 666.645, rot: -2 },
+  { dx: 36.78, dy: 275.30, w: 304.543, h: 324.33, rot: -15 },
+  { dx: 417.44, dy: 272.79, w: 418.531, h: 455.136, rot: 3 },
+];
```
overmyheadandbody, line 20: This is not very human-readable. Maybe this block needs a README to explain what each value represents, so updates are easier in the future, if they're needed.

overmyheadandbody (commented):
I'm not quite sure I see the content difference between mobile and tablet. I'm worried that this introduces two identical h1 tags on the page - view-source:https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-offer.plain.html. An alternative might be needed, maybe similar to the router-marquee, where we ended up using the `visually-hidden` block to set the `h1`

robert-bogos:
> I'm not quite sure I see the content difference between mobile and tablet. I'm worried that this introduces two identical h1 tags on the page - view-source:https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-offer.plain.html. An alternative might be needed, maybe similar to the router-marquee, where we ended up using the `visually-hidden` block to set the `h1`

The video assets are different, is currently like that just to showcase some asset issues to design

DKos95 (approved):

libs/mep/ace1205/offer-hero/offer-hero.css
robert-bogos: neither matches exactly 😔

libs/mep/ace1205/offer-hero/offer-hero.js
robert-bogos: True, those came from Figma and they can be confusing. I added a comment there to explain what each does

libs/mep/ace1205/offer-hero/offer-hero.css
robert-bogos: yes, good point. Made the change 👍

libs/mep/ace1205/offer-hero/offer-hero.css
robert-bogos: There is risk for things to break if the .X-up pattern is updated in certain ways. I removed it and added the grid rules directly

overmyheadandbody (approved):
Thanks for the changes!

PR #5970, opened 2026-05-22
[SR] - [MWPW-192890] - comparison table c2 block

DKos95:
PR adds comparison-table-c2 to mep ace1205.

Resolves: [MWPW-192890](https://jira.corp.adobe.com/browse/MWPW-192890)

**Test URLs:**
- Before: https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-plans?milolibs=site-redesign-foundation&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all
- After: https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-plans?milolibs=comparison-table-c2&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all


libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,883 @@
+@supports (animation-timeline: view()) {
```
narcis-radu, line 1: Seeing some weird rendering between 900 - 920px.

Image

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,883 @@
...
+    .table-row .table-cell:nth-child(even) div {
+      background-color: transparent;
+    }
+  }
+
+  .header-content-dummy + .table-container {
+    margin-top: 0;
+  }
+
+  .table-column-header {
```
robert-bogos, line 412: I see an UI issue with the column header for small desktops (1280px to 1600px). The header cards are overlaying with the column header, hiding the text.

Image

Same issue when the table rows are collapsed.

Image

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,883 @@
...
+    div {
+      border: none;
+      border-radius: 0;
+      width: 100%;
+      padding: 0;
+      display: flex;
+      flex-direction: column;
+      justify-content: center;
+      align-items: center;
+      background: transparent !important;
```
robert-bogos: using `!important` is risky, is it necessary?

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,883 @@
...
+
+  .toggle-icon {
+    width: var(--s2a-spacing-20);
+    height: var(--s2a-spacing-20);
+    position: relative;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    flex-shrink: 0;
+    background-image: none !important;
```
robert-bogos: using `!important` is risky, is it necessary?

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,883 @@
...
+    .mobile-filter-select {
+      display: none;
+    }
+
+    .mobile-title-row::after {
+      display: none;
+    }
+
+    .sub-header-item-container:first-child {
+      padding-right: 0;
```
robert-bogos: This might be redundant as `padding: 0` is already set on line 235

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.js
```diff
@@ -0,0 +1,631 @@
...
+
+function decorate(el) {
+  el.classList.add('con-block');
+  const [headerChild, ...tableChildren] = [...el.children];
+  decorateHeader(el, headerChild);
+  decorateTables(el, tableChildren);
+}
+
+export default function init(el) {
+  decorate(el);
```
robert-bogos, line 638: Did you consider using `decorateViewportContent(el, decorate)` ?
I'm not 100% sure is needed if you don't need to support viewport specific content, but is the C2 convention for decorating the block

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.js
```diff
@@ -0,0 +1,631 @@
...
+    decorateHeaderItem({
+      headerItem,
+      headerTitles,
+      headerItemIndex,
+      el,
+      headerItemsCount: headerItems.length,
+    });
+  });
+  const headerItemHeader = createTag('div', { class: 'header-item header-item-header' });
+  headerItemHeader.appendChild(createTag('h2', { class: 'title-4' }, 'Compare plans'));
```
robert-bogos: Do we want the `'Compare plans'` string hardcoded? It might cause issue if the page is localized

robert-bogos (dismissed):

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.js
```diff
@@ -0,0 +1,631 @@
...
+
+const hasTextNode = (...nodeLists) => nodeLists.some(
+  (nodes) => nodes && [...nodes].some(
+    (node) => node.nodeType === Node.TEXT_NODE && node.textContent.trim(),
+  ),
+);
+
+const getFirstVisibleColumnIndex = (el) => {
+  const firstVisible = [...el.querySelectorAll('.header-item[data-column-index]')]
+    .find((item) => !item.classList.contains('hidden'));
```
zagi25: NIT: If I am not wrong this can be simplified like this ? 
```suggestion
  const firstVisible = el.querySelector('.header-item[data-column-index]:not(.hidden)');
```

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
DKos95: We are waiting on designers to say which token number to use since somewhere it says 4 somewhere 5

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.js
```diff
@@ -0,0 +1,631 @@
...
+  const collapsible = createTag('div', { class: 'header-item-collapsible' });
+  const collapsibleInner = createTag('div');
+  const subContainers = [...headerItem.querySelectorAll('.sub-header-item-container')].slice(1);
+  subContainers.slice(0, -1).forEach((c) => collapsibleInner.appendChild(c));
+  if (collapsibleInner.children.length) {
+    collapsible.appendChild(collapsibleInner);
+    headerItem.appendChild(collapsible);
+  }
+  const btnSection = subContainers[subContainers.length - 1];
+  if (btnSection) {
```
zagi25: NIT: early return 
`if (!btnSection) return;`

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.js
```diff
@@ -0,0 +1,631 @@
...
+    .find((item) => !item.classList.contains('hidden'));
+  return firstVisible ? parseInt(firstVisible.getAttribute('data-column-index'), 10) : -1;
+};
+
+function syncAccessibilityHeaders(el) {
+  const accessibilityHeaderRow = el.querySelector('.accessibility-header-row');
+  if (!accessibilityHeaderRow) return;
+
+  const visibleHeaderItems = [...el.querySelectorAll('.header-item:not(.hidden)')];
+  if (!visibleHeaderItems.length) return;
```
zagi25: NIT: can this be combined in one early return ? 
`if (!accessibilityHeaderRow || !visibleHeaderItems.length) return;`

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.js
DKos95: Nice catch, thanks!

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.js
DKos95: This should be a nice to have right? Also this is mostly to have specific images per viewport I think which we don't have here.

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.js
```diff
@@ -0,0 +1,631 @@
...
+      id,
+      role: 'tooltip',
+      class: 'ct-tooltip-content',
+    });
+    tooltipSpan.append(
+      createTag('span', { 'aria-label': 'Tooltip :', role: 'tooltip' }),
+      tooltipText,
+    );
+
+    const trigger = createTag('a', {
```
zagi25, line 341: Does the trigger needs to be `a` ? If click behavior is prevented later on, couldn't `span` be used, so we avoid setting listener and adding `href: '#'` ?

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.js
```diff
@@ -0,0 +1,631 @@
...
+  if (!headerContent) return;
+  const cardsContainer = el.querySelector('.header-cards-container');
+  if (!cardsContainer) return;
+
+  let wasCollapsed = false;
+  let isExpanding = false;
+  let lastScrollY = window.scrollY;
+  let cachedMobileDelta = null;
+
+  const isMobile = () => window.matchMedia('(max-width: 899px)').matches;
```
zagi25, line 532: Is this media query correct or should `767px` be used ?

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.js
```diff
@@ -0,0 +1,631 @@
...
+  new ResizeObserver(() => {
+    if (!isExpanding) syncHeaderHeight();
+  }).observe(cardsContainer);
+}
+
+function setupTooltips(el) {
+  const tooltips = el.querySelectorAll('.milo-tooltip');
+  if (!tooltips.length) return;
+
+  const mq = window.matchMedia('(max-width: 899px)');
```
zagi25, line 609: same here

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.js
DKos95: Screenshot 2026-05-26 at 12 24 52

This block is a bit special :)

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.js
DKos95: same answer as above

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.js
robert-bogos: Yeah, definitely a nice to have

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.js
DKos95: Accessability team said it needs to be an anchor based on this link tooltip https://dequeuniversity.com/library/aria/tooltip

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
DKos95: There simply isn't enough space to render buttons with text and padding that are designed unless we break button text.

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
robert-bogos: Looks good now 👍

robert-bogos (approved):

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
robert-bogos: For the buttons, you get rid of the padding and the wrapper's`overflow: hidden;` which will make them look more decent.

Screenshot 2026-05-27 at 12 20 43

DKos95: I agree they look better but I don't think this is a solution since the buttons are overflowing then, which means some design tweaks are needed.

narcis-radu: @DKos95 - let's bring this up with design. Either we break earlier, or we have a different approach.

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.js
```diff
@@ -0,0 +1,632 @@
...
+  import('../../../utils/decorate.js').then(({ decorateButtons }) => {
+    elementsToDecorate.forEach((element) => decorateButtons(element, 'button-lg'));
+    addLastContainerElements(container);
+  });
+
+  return container;
+}
+
+function decorateHeaderItem({ headerItem, headerTitles, headerItemIndex, el, headerItemsCount }) {
+  headerItem.classList.add('header-item', 'header-item-card');
```
overmyheadandbody: Why are 2 classes needed instead of just one? Is their intent different or can it be consolidated? This could help avoid confusion in the future, to make it clear where changes are needed. Could we stick with just one class?

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.js
```diff
@@ -0,0 +1,632 @@
...
+    }
+    decorateHeaderItem({
+      headerItem,
+      headerTitles,
+      headerItemIndex,
+      el,
+      headerItemsCount: headerItems.length,
+    });
+  });
+  const headerItemHeader = createTag('div', { class: 'header-item header-item-header' });
```
overmyheadandbody: Similar comment, are both classes needed? And on another note, would it be a good idea to scope these to a table-specific class? `header-item` is quite generic and could be influenced by other blocks or 3rd party code.

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.js
```diff
@@ -0,0 +1,632 @@
...
+
+  window.matchMedia('(max-width: 899px)').addEventListener('change', () => {
+    cachedMobileDelta = null;
+    syncTop();
+    syncHeaderHeight();
+  });
+
+  window.addEventListener('resize', () => {
+    removeCollapsed();
+  });
```
overmyheadandbody, line 584: Should this be debounced?

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+  :is(p, h1, h2, h3, h4, h5, h6) {
+    margin: 0;
+  }
+
+  .ct-tooltip-content {
+    display: none;
+  }
+
+  .milo-tooltip:focus + .ct-tooltip-content,
+  .milo-tooltip:hover + .ct-tooltip-content {
```
overmyheadandbody: Petty nit:

```suggestion
  .milo-tooltip:is(:focus, :hover) + .ct-tooltip-content {
```

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+    display: block;
+    position: absolute;
+    left: -10000px;
```
overmyheadandbody: Why not something like `-200vw` for those who have three monitors glued together? :D

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+    padding-bottom: var(--s2a-spacing-xs);
+    position: sticky;
+    top: calc(var(--ct-nav-height) + var(--s2a-spacing-lg));
+    z-index: 2;
+
+    &.is-collapsed {
+      align-items: stretch;
+
+      .header-item-card {
+        box-shadow: 0 4px 16px rgba(0, 0, 0, 0.07);
```
overmyheadandbody: `--s2a-color-transparent-black-08` should be close enough and will allow for easier theming if we get a dark style in the future.

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+      }
+
+      .header-item-collapsible,
+      .btn-section-wrap {
+        grid-template-rows: 0fr;
+      }
+    }
+  }
+
+  @media (max-width: 899px) {
```
overmyheadandbody:
```suggestion
  @media (width < 900px) {
```
just to keep media queries consistent

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+    flex-direction: column;
+    gap: var(--s2a-spacing-2xs);
+
+    &:first-of-type {
+      margin-inline-end: 0;
+    }
+
+    img {
+      width: var(--s2a-spacing-lg);
+      height: var(--s2a-spacing-lg);
```
overmyheadandbody: I vaguely remember sending a similar comment for a similar situation. Spacing tokens can change, especially semantic ones. Tying box sizes to spacings might not be a good option long-term, they should be particular to the block's needs.

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+
+      .price {
+        justify-content: flex-start;
+      }
+    }
+
+    > .sub-header-item-container {
+      > p:not(:has(strong, em, a, picture)) {
+        font-size: var(--s2a-typography-font-size-label);
+        font-family: var(--s2a-font-family-label);
```
overmyheadandbody: This token doesn't exist in our styles (yes, we do get them from Consonant, but they're lacking a lot of fallback mechanisms and Typekit syntax that we need, so we're just not using them). Removing it should do the trick. Otherwise you can just use `--body-font-family`

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+        font-weight: var(--s2a-font-weight-label);
+        line-height: var(--s2a-typography-line-height-label);
+        letter-spacing: var(--s2a-typography-letter-spacing-label);
+        color: var(--s2a-color-gray-600);
+      }
+
+      > p:has(strong) {
+        font-size: var(--s2a-typography-font-size-heading-4);
+        font-weight: var(--s2a-font-weight-heading);
+        line-height: var(--s2a-typography-line-height-heading-4);
```
overmyheadandbody: What about `letter-spacing`?

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+
+    &.btn-section {
+      padding: var(--s2a-spacing-20) 0 0;
+    }
+
+    .description {
+      text-align: center;
+      padding: 0;
+      font-size: var(--s2a-typography-font-size-label);
+      line-height: var(--s2a-typography-line-height-label);
```
overmyheadandbody, line 241: You could also apply a `.label` class or another class that contains `label`. Otherwise you should also add `letter-spacing`

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+        color: var(--s2a-color-gray-900);
+      }
+    }
+
+    :is(h1, h2, h3, h4, h5, h6) {
+      font-size: var(--s2a-typography-font-size-heading-5);
+      font-family: var(--heading-font-family);
+      font-weight: var(--s2a-font-weight-heading);
+      line-height: var(--s2a-typography-line-height-heading-5);
+      letter-spacing: var(--s2a-typography-letter-spacing-heading-5);
```
overmyheadandbody: You could also programatically apply a `heading-5` class via JS for the effect

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+      word-break: break-word;
+    }
+
+    :is(p, h1, h2, h3, h4, h5, h6) {
+      text-align: center;
+      word-break: break-word;
+      color: var(--s2a-color-gray-900);
+    }
+
+    @media (max-width: 899px) {
```
overmyheadandbody:
```suggestion
    @media (width < 900px) {
```

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+    p:last-child a {
+      display: flex;
+      width: 100%;
+      justify-content: center;
+      align-items: center;
+      border-radius: var(--s2a-border-radius-round);
+      padding: 14px var(--s2a-spacing-lg);
+      font-size: var(--s2a-typography-font-size-label);
+      font-weight: var(--s2a-font-weight-label);
+      line-height: var(--s2a-typography-line-height-label);
```
overmyheadandbody, line 352: `letter-spacing` is feeling abandoned 😓

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+  .header-item:not(.hidden):not(:nth-child(2)) .sub-header-item-container:not(:last-of-type) {
+    background-color: transparent;
+  }
+
+  .table-row-header {
+    text-align: center;
+    font-size: var(--s2a-typography-font-size-eyebrow);
+    font-weight: var(--s2a-font-weight-adobe-clean-bold);
+    line-height: var(--s2a-typography-line-height-eyebrow);
+    letter-spacing: var(--s2a-typography-letter-spacing-eyebrow);
```
overmyheadandbody: Maybe just apple an `eyebrow` class via JS?

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+  }
+
+  .table-cell {
+    padding: var(--s2a-spacing-md) 0;
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    justify-content: center;
+
+    &.hidden {
```
overmyheadandbody: There's a similar declaration here - https://github.com/adobecom/milo/pull/5970/changes#diff-0071798550bcaf5e7918742e99a3404ca730d2cf7bdbbbbdb57f6dc793769a09R154, maybe they can be consolidated

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+      font-size: var(--s2a-typography-font-size-label);
+      line-height: var(--s2a-typography-line-height-label);
+      color: var(--s2a-color-gray-900);
+    }
+
+    > p:nth-child(2) {
+      margin-top: var(--s2a-spacing-3xs);
+      font-size: var(--s2a-typography-font-size-caption);
+      font-weight: var(--s2a-font-weight-body);
+      line-height: var(--s2a-typography-line-height-caption);
```
overmyheadandbody, line 550: Any chance `.label` / `.caption` can be applied via JS? If not, `letter-spacing` should be applied as well.

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+    font-size: var(--s2a-font-size-xl);
+
+    svg {
+      top: unset;
+    }
+  }
+
+  .icon-checkmark svg {
+    width:  20px;
+    height:  20px;
```
overmyheadandbody:
```suggestion
    width: 20px;
    height: 20px;
```

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+  .icon-dash {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    width: 20px;
+    height: 20px;
+
+    svg {
+      width: var(--s2a-spacing-md);
+      height: var(--s2a-spacing-md);
```
overmyheadandbody: Same "dimensions should not be based on spacings" comment as earlier, better have them be particular to the block so this doesn't update when spacing tokens are re-mapped.

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+      flex-shrink: 0;
+    }
+  }
+
+  .toggle-icon {
+    width:  20px;
+    height:  20px;
```
overmyheadandbody:
```suggestion
    width: 20px;
    height: 20px;
```

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+    position: relative;
+    display: flex;
+    align-items: center;
+
+    &::before {
+      content: '';
+      width: var(--s2a-spacing-xs);
+      height: var(--s2a-spacing-xs);
```
overmyheadandbody: Better not have these tied to spacing tokens, as those could change, affecting non-spacing elements

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+  .milo-tooltip svg {
+    height: var(--s2a-spacing-sm);
+    width: var(--s2a-spacing-sm);
+  }
+
+  .header-item-collapsible,
+  .btn-section-wrap {
+    display: grid;
+    grid-template-rows: 1fr;
+    transition: grid-template-rows 300ms cubic-bezier(0.4, 0, 0.2, 1);
```
overmyheadandbody: I've seen this before as well, `cubic-bezier(0.4, 0, 0.2, 1)`. Can it:
1. be made into a custom property?
2. be standardized to use `--parallax-easing`?

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+      justify-content: flex-start;
+      align-items: flex-start;
+      margin-top: 0;
+    }
+
+    &::after {
+      content: '';
+      flex-shrink: 0;
+      width:  20px;
+      height:  20px;
```
overmyheadandbody:
```suggestion
      width: 20px;
      height: 20px;
```

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+    margin: -1px;
+    overflow: hidden;
+    clip: rect(0, 0, 0, 0);
+    white-space: nowrap;
+    border-width: 0;
+  }
+
+  .accessibility-header-cell.hidden {
+    display: none;
+  }
```
overmyheadandbody: This could likely leverage the more general `.hidden` class I've suggested earlier

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+
+  .price {
+    font-size: var(--s2a-typography-font-size-body-md);
+    font-weight: var(--s2a-font-weight-adobe-clean-bold);
+    line-height: var(--s2a-typography-line-height-body-md);
```
overmyheadandbody, line 693: Either apply a `body-md` class via JS or add `letter-spacing` as well

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+    .header-content-wrapper .header-item:first-child {
+      display: flex;
+      align-items: flex-start;
+      justify-content: flex-start;
+      padding-top: var(--s2a-spacing-lg);
+
+      :is(h1, h2, h3, h4, h5, h6) {
+        font-size: var(--s2a-typography-font-size-heading-4);
+        font-weight: var(--s2a-font-weight-adobe-clean-bold);
+        line-height: var(--s2a-typography-line-height-heading-4);
```
overmyheadandbody: `heading-4` class or `letter-spacing` addition

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+      }
+    }
+
+    .table-row-header {
+      display: inline;
+      text-align: start;
+      padding: var(--s2a-spacing-md) var(--s2a-spacing-md) var(--s2a-spacing-md) 0;
```
overmyheadandbody: This could cause discrepancies on RTL. How about

```suggestion
      padding-block: var(--s2a-spacing-md);
      padding-inline: 0 var(--s2a-spacing-md);
```

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
```diff
@@ -0,0 +1,874 @@
...
+  .header-content {
+    .description,
+    .description a {
+      color: var(--s2a-color-gray-25);
+    }
+
+    &.sticky {
+      .action-area a:not(.blue) {
+        border-color: var(--text-color);
+        color: var(--text-color);
```
overmyheadandbody: This color is not defined, should be switched to the proper S2A value

overmyheadandbody (commented):
Looking very good! I do have a few nits to look into before test rollout

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.js
DKos95: I think this needs to happen instantly, based on the prototype at least.

libs/mep/ace1205/comparison-table-c2/comparison-table-c2.css
DKos95: @narcis-radu @robert-bogos @overmyheadandbody this should be fixed now with reduced padding.

narcis-radu (approved):
Looks good. Also, the previous issue between 900 and 920px is not that visible anymore, so we can handle things after the test if needed.

biljana-cvijanovic (approved):

PR #5973, opened 2026-05-26
[SR] - [MWPW-193259] - Rich Content Video variant

DKos95:
PR adds video variant to rich-content block.

Resolves: [MWPW-193259](https://jira.corp.adobe.com/browse/MWPW-193259)

**Test URLs:**
- Before: https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-product?milolibs=site-redesign-foundation
- After: https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-product?milolibs=sr2-rich-content-video


robert-bogos (commented):
I see you have some conflicts, might want to take a look

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -294,3 +294,237 @@
     .jump-links { padding-inline: 0; }
   }
 }
+
+/* video variant */
```
rgclayton: Nit: can probably get rid of the comment. The css notes it well enough.

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -294,3 +294,237 @@
...
+      margin: 0;
+    }
+
+    .video-container {
+      position: relative;
+      box-sizing: border-box;
+      max-width: 214px;
+      margin-inline: auto;
+      aspect-ratio: 9 / 16;
+      border-radius: 22.542px;
```
rgclayton: These decimal points on odd numbers. Do they cause any visual oddities at different screen sizes? 

Is it a big difference to round to a whole number?

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -294,3 +294,237 @@
...
+      max-width: var(--grid-width-10);
+
+      .video-container {
+        max-width: unset;
+        margin-inline: unset;
+        aspect-ratio: 16 / 9;
+
+        video {
+          height: auto;
+          object-fit: unset;
```
rgclayton: Unset resolves to initial which is fill — the video stretches to fill its box, ignoring aspect ratio. The intent is clearly to cancel the mobile cover. 
Suggestion:
  use object-fit: contain (letterbox) or just remove object-fit from the mobile rule entirely if
  it's not needed there.

libs/mep/ace1205/rich-content/rich-content.js
```diff
@@ -54,7 +54,34 @@ function decorateJumpLinks(content, foreground) {
...
+    container.append(mediaCell);
+  }
+
+  if (ctaCell) {
+    decorateBlockText(ctaCell);
+    ctaCell.classList.add('cta-area');
+    container.append(ctaCell);
+  }
+
+  row.remove();
```
rgclayton, line 75: The function guards ctaCell and mediaCell individually, but always removes row, correct? If both are absent  (e.g., an empty table cell), the block renders empty with no diagnostic. 

Try moving row.remove() inside a `if (ctaCell || mediaCell)` guard, or log a warning before removing.

libs/mep/ace1205/rich-content/rich-content.js
```diff
@@ -54,7 +54,34 @@ function decorateJumpLinks(content, foreground) {
   foreground.append(nav);
 }
 
+function decorateVideoVariant(container) {
```
rgclayton, line 57: Does this bypass `decorateViewportContent()`? Meaning no viewport variant support?

Does this mean that the video variant will never support the `mobile/tablet/desktop` viewport authoring pattern? If so it would be good to leave a comment noting that for future reference and any questions that might raise as to why it's missing.

libs/mep/ace1205/rich-content/rich-content.css
rgclayton: Or better yet, there should be a token for this.

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -294,3 +294,237 @@
...
+    .video-container {
+      position: relative;
+      box-sizing: border-box;
+      max-width: 214px;
+      margin-inline: auto;
+      aspect-ratio: 9 / 16;
+      border-radius: 22.542px;
+      border: 9px solid rgba(255, 255, 255, 0.05);
+      background: rgba(255, 255, 255, 0.05);
+      backdrop-filter: blur(18.133974075317383px);
```
rgclayton: Any tokens for this?

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -294,3 +294,237 @@
...
+    }
+
+    .section:has(+ .section.parallax-video-garage-door .rich-content.video) {
+      position: relative;
+      z-index: 2;
+      margin-bottom: calc(var(--rc-pull-offset) * -1);
+      animation: rc-section-pull-up cubic-bezier(0.4, 0, 0.2, 1) both;
+      animation-duration: auto;
+      animation-timeline: view();
+      animation-range: exit 0% exit 120%;
```
rgclayton, line 497: Is the 120% correct? The animation continues after the element has fully exited the viewport?

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -294,3 +294,237 @@
...
+
+@keyframes rc-video-section-slow {
+  from { transform: translateY(0); }
+  to { transform: translateY(50vh); }
+}
+
+@supports (animation-timeline: view()) {
+  @media (prefers-reduced-motion: no-preference) {
+    body {
+      timeline-scope: --rc-video-garage;
```
rgclayton, line 472: Claude is saying

> timeline-scope on body forces the browser to track this named timeline on every style recalc, even on pages where no .parallax-video-garage-door section exists (MEP active but no video block). If a tighter ancestor exists (main, a layout wrapper), scope it there. If not, add a comment acknowledging the intentional global scope.

libs/mep/ace1205/rich-content/rich-content.css
narcis-radu: maybe `--s2a-border-radius-24`

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -294,3 +294,237 @@
...
+  .media {
+    position: relative;
+    z-index: 1;
+    width: 100%;
+
+    p {
+      margin: 0;
+    }
+
+    .video-container {
```
narcis-radu, line 317: on mobile things seem a bit off, feels like someone would need to put a lot of effort to click on the CTA.

Image

libs/mep/ace1205/rich-content/rich-content.css
DKos95: Will put --s2a-blur-16 it's closest, let's see if they complain.

libs/mep/ace1205/rich-content/rich-content.css
DKos95: It's correct, not fault.

libs/mep/ace1205/rich-content/rich-content.js
DKos95: `decorateVideoVariant` is called inside of `decorate` which is called inside of `decorateViewportContent`, also in authoring I used the pattern and it seems to be working, think we are good on this one.

libs/mep/ace1205/rich-content/rich-content.css
DKos95: Updated to check if body has a section with the class animation, nice catch and thank you!

libs/mep/ace1205/rich-content/rich-content.css
DKos95: Not really sure what can be done about this, the animation is that the bottom part gets pulled up, we can only slow it down if needed so the user can more easily scroll up in order for the bottom part to go down but I don't think design would like that.

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -294,3 +294,235 @@
...
+      position: relative;
+      box-sizing: border-box;
+      max-width: 214px;
+      margin-inline: auto;
+      aspect-ratio: 9 / 16;
+      border-radius: var(--s2a-border-radius-24);
+      border: 9px solid rgba(255, 255, 255, 0.05);
+      background: rgba(255, 255, 255, 0.05);
```
overmyheadandbody: Maybe `--s2a-color-transparent-white-04` is close enough?

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -294,3 +294,235 @@
...
+  .action-area.dark {
+    background-color: unset;
+  }
+}
+
+@media (width >= 768px) {
+  .rich-content.video {
+    .media {
+      padding-inline: unset;
+      max-width: var(--grid-width-10);
```
overmyheadandbody, line 366: Keep in mind `--grid-width-10` is only available inside a `.container`. That should usually be the case, but there is potential for an edge case here.

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -294,3 +294,235 @@
...
+      img {
+        filter: blur(30px);
+        transform: scale(1.08);
+      }
+
+      &::after {
+        content: '';
+        position: absolute;
+        inset: 0;
+        background: rgba(0, 0, 0, 0.4);
```
overmyheadandbody: Maybe `--s2a-color-transparent-black-32` or `--s2a-color-transparent-black-48`?

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -294,3 +294,235 @@
...
+
+    .rich-content.video {
+      background: transparent;
+    }
+  }
+}
+
+.rich-content.video .video-container {
+  transform-origin: center bottom;
+}
```
overmyheadandbody: Can't this be attached to the group above, since it also touches on `.rich-content.video`?

overmyheadandbody (commented):
I agree with Narcis that it seems like we're making it difficult for the user to click the CTA button, regardless if viewport. Design should be contacted to check this and/or find some alternatives.

On the performance side, we would likely want to pause the video when it's not in view, so there's no needless CPU load when other sections of the page are in view. Is this a planned update?

Otherwise this looks pretty nifty! Just a few more details to iron out.

DKos95:
> On the performance side, we would likely want to pause the video when it's not in view, so there's no needless CPU load when other sections of the page are in view. Is this a planned update?

I don't think this is requested in the ticket anywhere or at least I didn't see it. Also I am not sure how doable this is if the block has a scroll animation since it has a translate slow down so IntersectionObserve reading will be skewed a bit. I've tried to implement it but seems that the animation transform on the block is an issue I haven't found a workaround to pause the video when it visually goes off screen.

overmyheadandbody:
@DKos95 - likely not requested officially, but as we've seen for the Wave 1 performance investigations, playing videos when they're off screen can needlessly load CPU, causing lag in other areas of the page. Router marquee implemented a solution; I also remember that the video logic used to have this OOTB in C1, I'm not sure what might've changed. Not a blocker right now, but likely a follow-up to score some performance wins.

libs/mep/ace1205/rich-content/rich-content.css
DKos95: We are using a container in this variant so we should be fine, I mean if someone takes out the container class the block will break(not look like it should) regardless, right?

DKos95:
> @DKos95 -  Router marquee implemented a solution; 

Like I said we have an animation on this block which is transform and it skews what the observer reads, which the router marquee doesn't have so the difference is probably there. I do agree that we should re visit it when we have time.

libs/mep/ace1205/rich-content/rich-content.css
DKos95: Will bring this up in a meeting today I have with Jordan.

narcis-radu: @DKos95 - If there's feedback from Design, maybe we can tackle it as a new requirement and have a separate JIRA ticket. @overmyheadandbody , @SilviuLCF

narcis-radu (approved):
Looks good, but we will need additional effort to handle:
* CTA positioning - requires design input, we can hopefully treat it separately (new JIRA ticket - @SilviuLCF)
* performance - video should pause when it's out of view; we need some research here (new JIRA ticket @SilviuLCF)

SilviuLCF:
@narcis-radu @DKos95 

Ticket for performance https://jira.corp.adobe.com/browse/MWPW-196476 
Ticket for CTA positioning  https://jira.corp.adobe.com/browse/MWPW-196479

SilviuLCF (approved):
QA verified testing details  https://jira.corp.adobe.com/browse/MWPW-193259 

With follow up concerns 
Ticket for performance https://jira.corp.adobe.com/browse/MWPW-196476
Ticket for CTA positioning  https://jira.corp.adobe.com/browse/MWPW-196479

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -294,3 +294,232 @@
...
+
+.section:has(.rich-content.video) {
+  background: var(--s2a-color-background-inverse);
+
+  &.has-background {
+    .section-background {
+      overflow: hidden;
+
+      img {
+        filter: blur(30px);
```
overmyheadandbody: How about using `--s2a-blur-md`?

overmyheadandbody (approved):

PR #5978, opened 2026-05-27
[SR] Tabs Block (Wave 2)

ivanvatadobe:
This PR adds the tabs block needed for ACE1205 (Site redesign wave 2) test. 

**Features:** 
- **New base tabs** - The basic tabs and radio variant were removed- the new base version uses _tab pills_. 
- **Paddles have been completely removed for mobile**. The tabs container will now be scrollable from screen edge x axis.
- **Sliding pill indicator** - a .tab-indicator element that slides behind the buttons to show the active tab, replacing the underline/border-bottom approach. 
- **Mobile horizontal scroll** - on viewports ≤ 599px, .tabList is set to overflow: visible so the tab list expands to fit all buttons. The parent .tabs-wrapper becomes the scroll container. This feature allows a user to _scroll_ the tabs container itself across the viewport, instead of scrunching/resizing. 
Auto-scroll on click/focus - scrollTabIntoView detects whether the active tab is fully within its scroll container using getBoundingClientRect, then calls container.scrollTo() (not scrollIntoView) to center the button without scrolling any ancestor elements.
- **Text color crossfade** - careful attention has been placed on transitions of text from hover state to selected state while the sliding pill indicator transition is taking place. The copy does not disappear during transition, a smooth transition from gray to white takes place, and there is no sudden or glitchy transition for the pill indicator. 

**Features that may be used for ACE1205:** 
- Setting default selected tab via authoring block row value: `active-tab=<number>`
- Base alignment is left aligned. `center` and `right` variants are available. Mobile will always be left aligned. 

**Retained Features:**

1. Deeplinking - Add `deeplink=<value>` to a tab content's section metadata. Link to that tab via URL using `?<tabs-id>=<value>`. Example: `?plans=edu`
2. Tab URL parameter - No authoring required. Append `?tab=<tabs-id>-<tab-number>` to any URL to open a specific tab on load. Example: `?tab=plans-2`
3. Default active tab - Add `active-tab=<tab-number>` to the tabs' block config to set which tab is open on load.
4. Remember last selected tab - Add `remember=on` to the tabs' block config to persist the last selected tab across sessions via sessionStorage. URL parameters always take priority over this setting.
5. Transparent tab container - Add `style=tabs-background-transparent` to the section metadata of the section containing the tabs block. This makes the tab list background transparent, not the tab content area.
6. Custom tab button color - Add `tab-background=<color>` to the section metadata of an individual tab content section to change that tab button's background color when selected.
7. Linked tabs - Add `link=<url>` to the section metadata of an individual tab content section to navigate to a URL instead of switching tab content on the same page.
8. Aria label for tabs - Add `pretext=<aria-label text>` to tabs block config will add desired aria label to tabs container.

**TODO/Notes**
1. ~Add variant to make background transparent (currently set to #fff).~  - **done**
2. ~Make left aligned the default alignment for >599px viewport, and add `center` and `right` variants.~ - **done**
3. ~Accessibility for :active and :focus - added focus-visible outline, no :active changes.~ - **done**
4. Set tab container widths (`.contained` css class that uses old `--grid-container-width`) vs section metadata classes. **(this may not be needed, and can be removed as `.container` handles sizing?)**
5. Consider moving helper functions outside of init(). 
6. Transition - tab transition has been added to non mobile breakpoints. I need to determine how the merch cards will be added to account for actual intended content.  

**Resolves:** 
[MWPW-192897](https://jira.corp.adobe.com/browse/MWPW-192897) (basic tabs)
[MWPW-193264](https://jira.corp.adobe.com/browse/MWPW-193264) ([Old design] left aligned tabs variant)

**Test URLs:**
- Before: https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/playground/viloria/plans-cards-table?milolibs=site-redesign-foundation&martech=off
- After: https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/playground/viloria/plans-cards-table?milolibs=sr2-tabs-block-fix&martech=off&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-playground-ivan.json--all

**DEMO PAGE WITH TABS AND MERCH CARDS:** 
https://main--da-dc--adobecom.aem.page/drafts/antonio/bizpro-plans?milolibs=local&martech=off&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-playground-ivan.json--all&maslibs=mwpw-192902


libs/mep/ace1205/tabs/tabs.js
```diff
@@ -0,0 +1,375 @@
...
+  parentSection?.classList.add(`tablist-${tabId}-section`);
+
+  // Tab Content
+  const tabContentContainer = createTag('div', { class: 'tab-content-container' });
+  const tabContent = createTag('div', { class: 'tab-content' }, tabContentContainer);
+  block.append(tabContent);
+
+  // Tab List
+  const tabList = rows[0];
+  tabList.classList.add('tabList');
```
robert-bogos: is better to use kebab-case for css classes, that's the convention across  milo

libs/mep/ace1205/tabs/tabs.js
```diff
@@ -0,0 +1,375 @@
...
+        id: `tab-${tabId}-${tabName}`,
+        tabindex: (i === 0) ? '0' : '-1',
+        'aria-selected': (i === 0) ? 'true' : 'false',
+        'data-block-id': `tabs-${tabId}`,
+        'daa-state': 'true',
+        'daa-ll': `tab-${tabId}-${tabName}`,
+        'aria-controls': controlId,
+      };
+      const tabBtn = createTag('button', tabBtnAttributes);
+      tabBtn.innerText = item.textContent;
```
robert-bogos: you can pass the text content as the 3rd argument for `createTag`

```suggestion
      const tabBtn = createTag('button', tabBtnAttributes, item.textContent);
```

libs/mep/ace1205/tabs/tabs.js
```diff
@@ -0,0 +1,375 @@
...
+        tabs[tabFocus].setAttribute('tabindex', '0');
+        tabs[tabFocus].focus();
+      }
+    });
+  });
+  tabs.forEach((tab) => {
+    tab.addEventListener('click', (e) => changeTabs(e, config));
+    tab.addEventListener('focus', () => scrollTabIntoView(tab));
+  });
+  if (config) configTabs(config, rootElem); // config will always exist
```
robert-bogos: do we need an `if block` if the `config` will always exist? 🤔

libs/mep/ace1205/tabs/tabs.css
```diff
@@ -0,0 +1,238 @@
...
+      }
+
+      &:focus-visible {
+        outline: 2px solid var(--tabs-pill-border-focus-color);
+        outline-offset: -2px;
+      }
+    }
+  }
+
+  @media screen and (max-width: 599px) {
```
robert-bogos: we're using the modern syntax for media queries across C2 blocks, better change in here too for consistency

```suggestion
  @media (width < 600px)
```

libs/mep/ace1205/tabs/tabs.css
```diff
@@ -0,0 +1,238 @@
...
+    &:nth-child(3) { animation-delay: 0.2s; }
+    &:nth-child(4) { animation-delay: 0.3s; }
+    &:nth-child(5) { animation-delay: 0.4s; }
+    &:nth-child(6) { animation-delay: 0.5s; }
+    &:nth-child(7) { animation-delay: 0.6s; }
+    &:nth-child(8) { animation-delay: 0.7s; }
+
+    @media (prefers-reduced-motion: reduce) {
+      animation: none;
+    }
```
robert-bogos: this should be a top level rule, not nested within a media query. 
With the guard here, animations will still fire on mobile => a11y violation

robert-bogos (dismissed):
Requesting changes mainly for the a11y issue

libs/mep/ace1205/tabs/tabs.css
```diff
@@ -0,0 +1,238 @@
...
+    display: flex;
+    width: 100%;
+    justify-content: var(--tabs-base-alignment);
+  }
+
+  .tab-indicator {
+    position: absolute;
+    top: var(--s2a-spacing-2xs);
+    bottom: var(--s2a-spacing-2xs);
+    left: 0;
```
DKos95, line 26: This is probably an issue for rtl?

libs/mep/ace1205/tabs/tabs.css
```diff
@@ -0,0 +1,238 @@
...
+        outline-offset: -2px;
+      }
+    }
+  }
+
+  @media screen and (max-width: 599px) {
+    .tabs-wrapper {
+      width: 100vw;
+      position: relative;
+      left: 50%;
```
DKos95, line 112: Also RTL potential issue?

libs/mep/ace1205/tabs/tabs.js
```diff
@@ -0,0 +1,375 @@
...
+    }
+  }
+
+  if (config['active-tab']) {
+    const id = `#tab-${CSS.escape(config['tab-id'])}-${CSS.escape(getStringKeyName(config['active-tab']))}`;
+    const sel = rootElem.querySelector(id);
+    if (sel) {
+      sel.addEventListener('click', (e) => e.stopPropagation(), { once: true });
+      sel.click();
+    }
```
DKos95:
```suggestion
    if (!config['active-tab']) return;
    const id = `#tab-${CSS.escape(config['tab-id'])}-${CSS.escape(getStringKeyName(config['active-tab']))}`;
    const sel = rootElem.querySelector(id);
    if (!sel) return;
    sel.addEventListener('click', (e) => e.stopPropagation(), { once: true });
    sel.click();
```

libs/mep/ace1205/tabs/tabs.js
```diff
@@ -0,0 +1,375 @@
...
+          tabFocus -= 1;
+          /* c8 ignore next */
+          if (tabFocus < 0) tabFocus = tabs.length - 1;
+        }
+        tabs.forEach((t) => t.setAttribute('tabindex', '-1'));
+        tabs[tabFocus].setAttribute('tabindex', '0');
+        tabs[tabFocus].focus();
+      }
+    });
+  });
```
DKos95, line 207: Would something like this work?
```suggestion
tabLists.forEach((tabList) => {
  tabList.addEventListener('keydown', (e) => {
    if (e.key !== 'ArrowRight' && e.key !== 'ArrowLeft') return;

    const forward = e.key === (document.dir === 'rtl' ? 'ArrowLeft' : 'ArrowRight');
    tabFocus = (tabFocus + (forward ? 1 : -1) + tabs.length) % tabs.length;

    tabs.forEach((t) => t.setAttribute('tabindex', '-1'));
    tabs[tabFocus].setAttribute('tabindex', '0');
    tabs[tabFocus].focus();
  });
});

```

libs/utils/utils.js
```diff
@@ -128,6 +128,7 @@ const C2_BLOCKS = [
   'social-proof',
   'visually-hidden',
   'product-marquee-grid',
+  'tabs',
```
DKos95, line 132: Not needed since this is in mep.

libs/mep/ace1205/tabs/tabs.css
```diff
@@ -0,0 +1,238 @@
...
+  --tabs-pill-bg-color: transparent;
+  --tabs-pill-bg-color-selected: var(--s2a-color-background-knockout);
+  --tabs-pill-text-color-hover: #2c2c2c;
+  --tabs-pill-text-selected-color: var(--s2a-color-content-knockout);
+  --tabs-pill-bg-focus: var(--s2a-color-gray-700); /* keep for testing ace1205 focus feature if needed */
+  --tabs-pill-bg-color-active: var(--s2a-color-transparent-black-64); /* keep for testing ace1205 active feature if needed */
+  --tabs-pill-text-focus-color: var(--s2a-color-gray-25); /* keep for testing ace1205 focus feature if needed */
+  --tabs-pill-border-focus-color: var(--s2a-color-blue-800);
+  --m-rounded-corners-tabs: 50px;
+}
```
overmyheadandbody, line 4: Add these to `.tabs` directly. It avoid scope bleed and shortens some of the custom property names, as some no longer have to be prefixed with `tabs-`

libs/mep/ace1205/tabs/tabs.css
```diff
@@ -0,0 +1,238 @@
...
+  background-color: var(--tabs-active-bg-color);
+  z-index: 1;
+
+  &.background-transparent {
+    background: transparent;
+  }
+
+  .tab-headline { /* keep for testing ace1205 when merch is available */
+    margin-top: var(--spacing-xxl);
+    margin-bottom: var(--spacing-xl);
```
overmyheadandbody: These variables don't exist on C2 pages.

libs/mep/ace1205/tabs/tabs.css
```diff
@@ -0,0 +1,238 @@
...
+
+    button.tab-button {
+      color: var(--tabs-pill-text-color);
+      cursor: pointer;
+      font-family: var(--body-font-family);
+      font-size: var(--tabs-pill-font-size);
+      text-align: center;
+      font-style: normal;
+      font-weight: var(--s2a-font-weight-adobe-clean-bold);
+      line-height: var(--s2a-font-line-height-2xs);
```
overmyheadandbody: You most likely need to use the `.label` styles. Either apply a `.label` class to the button or use its properties, https://github.com/adobecom/milo/blob/site-redesign-foundation/libs/c2/styles/styles.css#L716-L721. This ensures consistency and proper scaling across viewports.

libs/mep/ace1205/tabs/tabs.css
```diff
@@ -0,0 +1,238 @@
...
+      transition: color 0.1s ease-in-out;
+      outline: none;
+      -webkit-tap-highlight-color: transparent;
+      white-space: nowrap;
+      width: auto;
+      position: relative;
+      z-index: 1;
+      flex: 1 0 auto;
+      background: var(--tabs-pill-bg-color);
+      height: var(--s2a-spacing-xl);
```
overmyheadandbody: Likely not a good idea to tie box dimensions to spacings, as spacings can change in future token iterations. A px value would likely be acceptable here.

libs/utils/utils.js
narcis-radu: You bring up an interesting point: you are correct, however some folks have added the newly created blocks in C2, but not ALL of them.

libs/mep/ace1205/tabs/tabs.css
overmyheadandbody: Furthermore, our mobile breakpoint is 768, that should be adapted as well.

libs/mep/ace1205/tabs/tabs.css
```diff
@@ -0,0 +1,238 @@
...
+  --tabs-base-alignment: flex-start;
+  --tabs-mobile-alignment: flex-start;
+
+  /* Tab Colors */
+  --tabs-active-text-color: #2c2c2c; /* keep for testing ace1205 */
+  --tabs-active-bg-color: var(--s2a-color-gray-25);
+  --tabs-container-background-color: var(--s2a-color-transparent-black-04);
+
+  /* Pill colors */
+  --tabs-pill-font-size: var(--s2a-typography-font-size-label);
```
narcis-radu: wondering if we really need to define new vars. `--tabs-pill-font-size` is used once; same goes for some of the other vars defined

libs/mep/ace1205/tabs/tabs.css
```diff
@@ -0,0 +1,238 @@
...
+      }
+    }
+
+    [role='tablist'] {
+      overflow: visible;
+      margin: 0 var(--s2a-spacing-20);
+    }
+  }
+
+  @media screen and (min-width: 600px) {
```
overmyheadandbody: Same breakpoint comment, both syntax and value

```suggestion
  @media (width >= 768px) {
```

libs/mep/ace1205/tabs/tabs.css
```diff
@@ -0,0 +1,238 @@
...
+    &.right .tabs-wrapper {
+      justify-content: flex-end;
+    }
+  }
+}
+
+/* Contained tabs content */
+[role='tabpanel'] > .section > .content,
+.tabs.contained .tab-content .tab-content-container {
+  width: var(--grid-container-width);
```
overmyheadandbody: No such variable for C2 pages

libs/mep/ace1205/tabs/tabs.css
```diff
@@ -0,0 +1,238 @@
...
+/* Section Metadata - keep for testing ace1205 when merch is available */
+.tabs-background-transparent {
+  .tabs,
+  .tabs [role='tablist'],
+  .tabs [role='tablist'] button.tab-button[aria-selected="true"] {
+    background: transparent;
+  }
+}
+
+@media screen and (min-width: 600px) {
```
overmyheadandbody: Same media query syntax + value comment

libs/mep/ace1205/tabs/tabs.css
```diff
@@ -0,0 +1,238 @@
...
+    &:nth-child(7) { animation-delay: 0.6s; }
+    &:nth-child(8) { animation-delay: 0.7s; }
+
+    @media (prefers-reduced-motion: reduce) {
+      animation: none;
+    }
+  }
+}
+
+@media screen and (min-width: 1200px) {
```
overmyheadandbody: Depending on intent, this should be either 1024 or 1280

libs/mep/ace1205/tabs/tabs.css
```diff
@@ -0,0 +1,238 @@
+:root {
+  /* Alignment */
+  --tabs-base-alignment: flex-start;
+  --tabs-mobile-alignment: flex-start;
+
+  /* Tab Colors */
+  --tabs-active-text-color: #2c2c2c; /* keep for testing ace1205 */
```
narcis-radu: this is not used

libs/mep/ace1205/tabs/tabs.css
```diff
@@ -0,0 +1,238 @@
...
+  --tabs-active-bg-color: var(--s2a-color-gray-25);
```
narcis-radu: no need to create new variables

zagi25:
Content/blocks inside tab are overlapping 
Screenshot 2026-05-29 at 15 42 51

libs/mep/ace1205/tabs/tabs.js
```diff
@@ -0,0 +1,375 @@
...
+  const targetId = target.getAttribute('id');
+  const redirectionUrl = getRedirectionUrl(linkedTabs, targetId);
+  /* c8 ignore next 4 */
+  if (redirectionUrl) {
+    window.location.assign(redirectionUrl);
+    return;
+  }
+  const parent = target.closest('.tab-list-container');
+  const tabsBlock = target.closest('.tabs');
+  const content = getContentElement(parent, 3);
```
zagi25: Do we need this function ? If I understand correctly we are looking for `tab-content`, then you can do `target.closest('.tabs').lastElementChild` to get the same element right ?

libs/mep/ace1205/tabs/tabs.css
ivanvatadobe: removed.

libs/mep/ace1205/tabs/tabs.css
ivanvatadobe: removed variable, and switched to using style token instead.

libs/mep/ace1205/tabs/tabs.css
ivanvatadobe: this is a good point. during dev, i kept them in as a makeshift config to prevent needing to comb through declarations manually. since there are no complicated variants, then this is probably pointless. i will cut down on variables that are not used more than once.

libs/mep/ace1205/tabs/tabs.css
ivanvatadobe: variables used only once have been removed, and their settings applied to the css rules directly via styles token or static value.

libs/mep/ace1205/tabs/tabs.css
ivanvatadobe: the class itself is also not used. i kept them in just in case we needed to use headlines for buttons. i've confirmed that we are using labels instead. i've removed the declaration.

rgclayton (approved):
Worked with Ivan to get needed updates in place to demo.

narcis-radu (approved):

NadiiaSokolova:
Verified. Only RTL issues were found, which will be covered in the follow-up ticket.

PR #5981, opened 2026-05-27
MWPW-191003 M@S Errors Visible in aem.page 

Blainegunn:
On `aem.page` users will now see more accurate errors states for m@s fragments 
Call made to `mas/io` will also have a `source` query param to help declutter logs. They values will be either `preflight` or `aem`. These will only come from `aem.page` pages where preflight can run. 

DA (aem.page) - https://main--upp--adobecom.aem.page/drafts/gunn/pre-mas-error?milolibs=MWPW-191003
SP (aem.page)  - https://mwpw-191003--milo--adobecom.aem.page/drafts/mariia/card404

Before:
Only a message for the collection; cards fail silently. 
image

After:
image

Resolves: [MWPW-191003](https://jira.corp.adobe.com/browse/MWPW-191003)

**Test URLs:**
- After: https://mwpw-191003--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/libs/blocks/merch-card-autoblock/merch-card-autoblock.js
+++ b/libs/blocks/merch-card-autoblock/merch-card-autoblock.js
@@ -71,10 +71,10 @@ export async function checkReady(masElement, fragment) {
   if (isMasErrorEnv()) {
     const uuid = fragment ?? masElement.querySelector('aem-fragment')?.getAttribute('fragment');
     if (masElement.hasAttribute('failed')) {
-      createFragmentErrorEl(uuid, 'Frag').then((el) => masElement.insertAdjacentElement('beforebegin', el));
+      createFragmentErrorEl(uuid, 'Card').then((el) => masElement.insertAdjacentElement('beforebegin', el));
     } else {
       masElement.addEventListener('aem:error', async (e) => {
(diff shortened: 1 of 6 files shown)


Roycethan (commented):
@Blainegunn 
https://main--da-cc--adobecom.aem.page/drafts/reuben/error?milolibs=MWPW-191003#
1) for first in collection (collection with only deleted card) getting "Load Error" , why not "Not found" is rendered
3) For 3rd collection (Collection with valid and deleted cards) the deleted card is removed , but the Card message "Not found" is not rendered

Blainegunn:
@Roycethan 
For (1) it only examines the top-level fragment. In this case it's an empty collection. So it's not a `404`; but there is a load error. We aren't going any deeper to determine what it is to limit how many request we make to`mas/io`.
For (2) I believe any deleted cards that were in the collection are now removed. So it would detect the issue b/c we handle on the studio side. 
image (see below). Your test collection has only one card now.

Roycethan:
@Blainegunn Plz resolve conflicts

Roycethan (approved):

TsayAdobe (approved):

PR #5986, opened 2026-05-28
[Site Re-design] [Hub Hero] [MWPW-192873] Code Complete

denlight:
Hub Hero for Site Redesign

* Add your
* Specific
* Features or fixes

Resolves: [MWPW-192873](https://jira.corp.adobe.com/browse/MWPW-192873)

**Test URLs:**

TEST URL: https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-bizpro-hub?milolibs=hub-hero&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all


libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
+/** main classes */
```
overmyheadandbody: Logic should be added to the ace1205 folder, both CSS and JS

libs/c2/blocks/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+  --start-gap: 6rem;
+  --end-gap: .5rem;
+  --hub-hero-carousel-container-width: calc((var(--slides) * var(--hub-hero-carousel-item-max-width)) + (var(--slides) * (var(--end-gap) * 2)));
+
+  --grid-col-one-offset: -132px;
+  --grid-col-two-offset: 36px;
+  --grid-col-three-offset: 146px;
+  --grid-col-four-offset: 36px;
+  --grid-col-five-offset: -132px;
+}
```
overmyheadandbody: Are these actually needed at the root level? This overwrites a bunch of core custom properties. Instead, they should be nested inside the `.hub-hero` class.

libs/c2/blocks/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+
+html[dir="rtl"] {
+  .hub-hero-carousel {
+    & .hub-hero-carousel-container {
+      & .hub-hero-carousel-item {
+        & .hub-hero-carousel-item-footer {
+          &:after {
```
overmyheadandbody: No need for nesting in these cases, where the class name is so specific. You can likely summarize this to just `.hub-hero-carousel-item-footer:after`. This also makes future edits easier, since specificity is lower and clear.

libs/c2/blocks/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+        & .hub-hero-carousel-item-header {
+          &:after {
+            transform: rotate(180deg);
+            left: 0;
+            right: unset;
+          }
+
+          .eyebrow {
+            font-size: var(--s2a-font-size-md);
+            font-weight: var(--s2a-font-weight-adobe-clean-bold);
```
overmyheadandbody: Eyebrow text should be standardized already, applying the class should take care of all of the typography rules. Why is this override needed?

libs/c2/blocks/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+      & .hub-hero-carousel-item {
+        & .hub-hero-carousel-item-footer {
+          &:after {
+            transform: rotate(180deg);
+            left: 0;
+            right: unset;
+          }
+        }
+
+        & .hub-hero-carousel-item-header {
```
overmyheadandbody: Same comment here. Instead of the whole nesting chain, you can just use `.hub-hero-carousel-item-header`. I'll stop adding this comment, but this applies to other parts of the file as well.

Also, assuming that this was needed, you don't need the `&` prefix, that's implicit in this case.

libs/c2/blocks/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+  padding-top: var(--hub-hero-top-spacing);
+  position: sticky;
+  top: 0;
+  z-index: 0;
+  animation: hubHeroHeaderFade var(--paralax-easing) both;
+  animation-timeline: --hub-hero;
+  animation-range: contain 0% contain 15%;
+  height: var(--hub-hero-header-height);
+
+  div {
```
overmyheadandbody: This sounds a bit risky, as it's likely there could be other div elements added in the future. Can it get a class or be made more strict via some direct child selectors or something else?

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+    top: var(--hub-hero-top-spacing);
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    text-align: center;
+    gap: var(--s2a-spacing-lg);
+
+    .eyebrow {
+      font-weight: var(--s2a-font-weight-adobe-clean-bold);
+      font-size: var(--s2a-typography-font-size-eyebrow);
```
overmyheadandbody: Same as before, eyebrows should already be standardized, I'm not sure why it's needed to stray away from those system values - https://github.com/adobecom/milo/blob/site-redesign-foundation/libs/c2/styles/styles.css#L708-L714

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+    }
+
+    p {
+      margin-bottom: 0;
+      max-width: 574px;
+    }
+
+    h1 {
+      color: var(--S2AC-Palette-static-black, black);
```
overmyheadandbody: This isn't a valid variable, where is it coming from? Maybe `--s2a-color-gray-1000` fits the bill

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+      font-size: var(--s2a-typography-font-size-heading-1);
+      font-weight: var(--s2a-font-weight-adobe-clean-bold);
+      line-height: var(--s2a-font-line-height-76);
```
overmyheadandbody: These should already be standardized via https://github.com/adobecom/milo/blob/site-redesign-foundation/libs/c2/styles/styles.css#L650-L654. Overrides will break the typography system, making things inconsistent

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+    line-height: var(--s2a-typography-line-height-eyebrow);
+    position: relative;
+    cursor: pointer;
+    border: 0;
+
+    &::before {
+      content: "";
+      background-image: url('https://www.adobe.com/cc-shared/assets/img/product-icons/svg/acrobat-pro-64.svg');
+      position: absolute;
+      left: 12px;
```
overmyheadandbody: Is this reversed for RTL?

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+    }
+
+    &::after {
+      content: "";
+      background-image: url(../../assets/img/arrow-bottom.svg);
+      position: absolute;
+      padding: var(--s2a-spacing-xs);
+      height: 14px;
+      width: 14px;
+      right: 12px;
```
overmyheadandbody: Does this reverse on RTL?

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+.hub-hero-image-grid-container {
+  display: flex;
+  flex-direction: row;
+  z-index: 1;
+  justify-content: center;
+  animation: hubHeroGridGapShrink var(--paralax-easing) both;
+  animation-timeline: --hub-hero;
+  animation-range: var(--grid-gap-animation-range);
+
+  .hub-hero-image-grid-container-col {
```
overmyheadandbody: Is the additional nesting specificity really needed? Otherwise this could be moved outside its parent, since its name already contains all the context.

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+    }
+  }
+}
+
+.hub-hero-carousel-header {
+  position: absolute;
+  top: -14%;
+  width: 100%;
+  text-align: center;
+  font-size: var(--s2a-typography-font-size-title-2);
```
overmyheadandbody: This token no longer exists. Furthermore, if this needs to be styled as an h2, it could make use of the `heading-2` class

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+  animation-range: var(--header-animation-range);
+}
+
+.hub-hero-carousel {
+  height: var(--hub-hero-carousel-container-height-desktop);
+  position: sticky;
+  top: var(--carousel-items-top);
+  overflow: visible;
+  margin: auto;
+  left: 0;
```
overmyheadandbody, line 228: Reversed on RTL?

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+  right: 0;
+  z-index: 1;
+
+  .hub-hero-carousel-container {
+    display: flex;
+    position: absolute;
+    justify-content: center;
+    left: 50%;
+    transform: translateX(-50%);
```
overmyheadandbody: RTL?

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+
+    .hub-hero-carousel-item {
```
overmyheadandbody: Nesting likely not needed since the name is so clear, this would also make specificity clearer

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+      max-width: var(--max-image-size);
+      min-width: var(--min-image-size);
+      background-color: var(--s2a-color-gray-75);
+      border-radius: var(--s2a-border-radius-md);
+      cursor: pointer;
+      transition: background-color .2s var(--paralax-cubic);
+      contain: layout;
+
+      .hub-hero-carousel-item-header p {
+        font-size: var(--s2a-font-size-lg);
```
overmyheadandbody: Typography usually scales depending on viewport, this might not be the best value. Is this supposed to be a body text? Eyebrow? Caption? Label? We have classes for those that you could apply to get typography declarations OOTB

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+          display: block;
+        }
+      }
+
+      .hub-hero-carousel-item-footer {
+        display: flex;
+        flex-direction: column;
+        font-size: var(--s2a-font-size-md);
+        font-weight: var(--s2a-font-weight-adobe-clean-bold);
+        line-height: var(--s2a-typography-line-height-eyebrow);
```
overmyheadandbody: You can likely apply an `eyebrow` class to get all the correct values.

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+        --icon-size: 24px;
+
+        &::after {
+          content: "+";
+          position: absolute;
+          padding: var(--s2a-spacing-xs);
+          height: var(--icon-size);
+          width: var(--icon-size);
+          right: var(--s2a-spacing-lg);
```
overmyheadandbody: RTL?

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+          background-color: var(--s2a-color-transparent-white-16);
+          opacity: 0;
+          font-size: var(--s2a-font-size-2xl);
+          display: flex;
+          align-items: center;
+          justify-content: center;
+        }
+
+        :is(h1, h2, h3, h4, h5, h6) {
+          /* TODO: remove eyebrow styles if eyebrow class is applied directly to footer headline */
```
overmyheadandbody: This would likely be a better approach, yes

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+          overflow: hidden;
+        }
+      }
+    }
+  }
+}
+
+
+/** custom media rule 1 */
+@media (768px <=width < 1495px) {
```
overmyheadandbody:
```suggestion
@media (768px <= width < 1495px) {
```
We don't have a 1495 breakpoint, where is this coming from?

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+  .hub-hero-carousel {
+    .hub-hero-carousel-container {
+      transition: transform .3s var(--paralax-cubic);
+
+      &.stick-left {
+        transform: translateX(-48%);
+        left: 50%;
```
overmyheadandbody: RTL?

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+      }
+    }
+  }
+}
+
+
+/** custom media rule 1 */
+@media (768px <=width < 1495px) {
+  .hub-hero-carousel {
+    .hub-hero-carousel-container {
```
overmyheadandbody: Nesting likely not needed.

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+      transition: transform .3s var(--paralax-cubic);
+
+      &.stick-left {
+        transform: translateX(-48%);
+        left: 50%;
+      }
+
+      &.stick-right {
+        transform: translateX(calc(100vw - 100% - var(--s2a-spacing-sm))) !important;
```
overmyheadandbody: Is `!important` really needed? This sounds like a specificity issue, that could be solved via the earlier comments of flattening selectors.

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+      }
+
+      .hub-hero-carousel-item {
+        width: calc(100vw / calc(var(--slides) - 1.5));
+      }
+
+      .hub-hero-carousel-item.hovered,
+      .hub-hero-carousel-item:focus-visible {
+        max-width: 33vw !important;
```
overmyheadandbody: Same `!important` comment, we do our best to avoid these.

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+      }
+    }
+  }
+}
+
+/** custom media rule - 3 */
+@media (width >=1940px) {
```
overmyheadandbody: Also a non-standard breakpoint, could this be turned into `1920`?

libs/c2/blocks/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+      .hub-hero-carousel-item:focus-visible {
+        max-width: 33vw !important;
+      }
+    }
+  }
+}
+
+/** custom media rule - 3 */
+@media (width >=1940px) {
+  :root {
```
overmyheadandbody: Same as the `:root` comment above, this should sit in a hub-specific class to avoid unintentional overrides affecting the whole page.

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+  .hub-hero-carousel {
+    height: var(--hub-hero-carousel-container-height-tablet);
+
+    .hub-hero-carousel-container {
+      max-height: var(--hub-hero-carousel-container-height-tablet);
+    }
+  }
+}
+
+/** tablet up */
```
overmyheadandbody: Media query order seems a bit off, I see desktop above, but now I see tablet, later mobile-only. We usually follow a mobile-first approach. That makes the file easier to read and also point at the clear inheritance rules that are used.

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+      animation: carouselGapShrink var(--paralax-easing) both;
+      animation-timeline: --hub-hero;
+      animation-range: var(--grid-gap-animation-range);
+
+      .hub-hero-carousel-item-container {
+        position: relative;
+      }
+
+      .hub-hero-carousel-item {
+        transition: max-width .3s var(--paralax-cubic), width .3s var(--paralax-cubic), background-color .3s var(--paralax-cubic);
```
overmyheadandbody: Keep in mind your `--paralax-cubic` is the default `--parallax-easing`, this will need to be updated once the custom properties are nested inside the hub element.

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+
+        &.hovered,
+        &:focus-visible {
+          will-change: max-width, width;
+          background-color: var(--s2a-color-gray-1000);
+          max-width: 555px !important;
+          width: calc((100vw/calc(var(--slides))) * var(--hub-hero-carousel-item-hover-grow-index)) !important;
```
overmyheadandbody: Can specificity be edited to avoid `!important`?

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+        .hub-hero-carousel-item-container {
+          position: relative;
+          background-color: var(--s2a-color-gray-1000);
+          border-radius: var(--s2a-border-radius-md);
+          animation: elasticItemsShrink var(--paralax-easing) both;
+          animation-timeline: --hub-hero;
+          animation-range: entry 50% exit 100%;
+          overflow-x: clip;
+
+          .hub-hero-carousel-item-footer::after {
```
overmyheadandbody: Again, the level of nesting is way too high, this can be used without any indenting, like a simple class declaration without any nesting inside a lot of parents.

libs/c2/blocks/hub-hero/hub-hero.js
```diff
@@ -0,0 +1,310 @@
...
+
+  const grid = handleGridImages(gridImages);
+  const elasticCarousel = handleCarousel(carouselImages);
+  elasticCarousel.prepend(carouselHeader);
+  el.replaceChildren();
+
+  el.append(hero, grid, elasticCarousel);
+
+
+  document.head.querySelector('style[id="ims-body-style"]').remove();
```
overmyheadandbody: This can cause issues on lower environments where IMS doesn't load, it should have some optional chaining before calling `remove`

libs/c2/blocks/hub-hero/hub-hero.js
```diff
@@ -0,0 +1,310 @@
...
+import { processTrackingLabels } from '../../../martech/attributes.js';
+
+const leaveTimeouts = new WeakMap();
+let hoverTracked = false;
+const rewindIntervals = new WeakMap();
+const slideLeaveTimeouts = new WeakMap();
+
+const isSvgUrl = (url) => /\.svg(\?.*)?$/i.test(url || '');
+const isRtl = () => document.documentElement.getAttribute('dir') === 'rtl';
+const isMobile = () => window.innerWidth <= 768;
```
overmyheadandbody: `matchMedia` is more performant and a standard in the codebase

overmyheadandbody (dismissed):
This looks great visually, but some tweaks are needed to get this ready for test rollout.

The thing I'm worried about the most is the custom properties added to `:root`, which override the regular ones, like the parallax easing and plenty more. Hub-specific variables should be nested in the hub to-most element.

Additionally, I'm still noticing lateral scroll on the page due to this block, it would be great to get that addressed.

libs/c2/blocks/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+  --grid-col-three-offset: 146px;
+  --grid-col-four-offset: 36px;
+  --grid-col-five-offset: -132px;
+}
+
+html[dir="rtl"] {
+  .hub-hero-carousel {
+    & .hub-hero-carousel-container {
+      & .hub-hero-carousel-item {
+        & .hub-hero-carousel-item-footer {
```
DKos95: & should be not needed here since there is a space between it and the class

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1282 @@
...
+      line-height: var(--s2a-font-line-height-76);
+      word-wrap: break-word;
+    }
+  }
+
+  .action-area .con-button {
+    background-color: black;
+    color: white;
+    padding: var(--s2a-spacing-lg) 64px var(--s2a-spacing-lg) 54px;
+    border-radius: 12px;
```
DKos95: --s2a-border-radius-12

libs/mep/ace1205/hub-hero/hub-hero.js
```diff
@@ -0,0 +1,310 @@
...
+
+const rewindVideo = (video) => {
+  stopRewind(video);
+  video.pause();
+  const startSystemTime = Date.now();
+  const startVideoTime = video.currentTime;
+  const intervalRewind = setInterval(() => {
+    if (video.currentTime === 0) {
+      stopRewind(video);
+      video.load();
```
DKos95, line 30: early return in order to not use else

libs/c2/blocks/hub-hero/hub-hero.js
```diff
@@ -0,0 +1,310 @@
...
+  slides.forEach((slide, index) => {
+    const video = slide.querySelector('video');
+    if (!video) return;
+
+    const nextSlide = slides[index + 1];
+
+    // Play when this slide enters view — but not if the next slide is already covering it
+    const slideObserver = new IntersectionObserver(
+      ([entry]) => {
+        if (!isMobile()) return;
```
DKos95:
```suggestion
        if (!isMobile() || !entry.isIntersecting) return;
```

And then bellow we don't need two ifs just one ?

libs/mep/ace1205/hub-hero/hub-hero.js
```diff
@@ -0,0 +1,310 @@
...
+
+    if (!nextSlide) return;
+
+    // Rewind when the next slide starts covering this one;
+    // play again when it uncovers (user scrolls back up)
+    const nextSlideObserver = new IntersectionObserver(
+      ([entry]) => {
+        if (!isMobile()) return;
+        if (entry.isIntersecting) {
+          rewindVideo(video);
```
DKos95, line 69: early return

libs/c2/blocks/hub-hero/hub-hero.js
```diff
@@ -0,0 +1,310 @@
...
+
+  if (isRtl()) {
+    container.classList.toggle('stick-right', slideIndex === 1);
+    container.classList.toggle('stick-left', slideIndex === 5);
+  } else {
+    container.classList.toggle('stick-left', slideIndex === 1);
+    container.classList.toggle('stick-right', slideIndex === 5);
+  }
+
+  if (!hoverTracked) {
```
DKos95: early return in order to not nest things in if.

libs/utils/utils.js
```diff
@@ -118,6 +118,7 @@ const C2_BLOCKS = [
   'faq',
   'global-footer',
   'global-navigation',
+  'hub-hero',
```
DKos95: Once you move the block to mep this is not needed anymore

libs/c2/blocks/hub-hero/hub-hero.js
```diff
@@ -0,0 +1,310 @@
...
+  removeHovered(slideEl.closest('.hub-hero-carousel'));
+  slideEl.classList.add('hovered');
+
+  if (isRtl()) {
+    container.classList.toggle('stick-right', slideIndex === 1);
+    container.classList.toggle('stick-left', slideIndex === 5);
+  } else {
+    container.classList.toggle('stick-left', slideIndex === 1);
+    container.classList.toggle('stick-right', slideIndex === 5);
+  }
```
DKos95:
```suggestion
const rtl = isRtl();
container.classList.toggle('stick-left', rtl ? slideIndex === 5 : slideIndex === 1);
container.classList.toggle('stick-right', rtl ? slideIndex === 1 : slideIndex === 5);
```

We can shorten this to 3 lines instead of 7

libs/c2/blocks/hub-hero/hub-hero.js
denlight: unwanted debugging artefact. removing

libs/mep/ace1205/hub-hero/hub-hero.js
denlight: if we are already in the if, else shouldn't be used, right ?

libs/mep/ace1205/hub-hero/hub-hero.js
denlight: else shouldn't fire, not sure why we need an early return, help me understand :)

libs/mep/ace1205/hub-hero/hub-hero.js
```diff
@@ -0,0 +1,301 @@
...
+import { processTrackingLabels } from '../../../martech/attributes.js';
+
+const leaveTimeouts = new WeakMap();
+let hoverTracked = false;
+const rewindIntervals = new WeakMap();
+const slideLeaveTimeouts = new WeakMap();
+
+const isSvgUrl = (url) => /\.svg(\?.*)?$/i.test(url || '');
+const isRtl = () => document.documentElement.getAttribute('dir') === 'rtl';
+const isMobile = () => window.matchMedia('(min-width: 768px)');
```
rgclayton: isMobile() returns a MediaQueryList object, not a boolean. Every call-site (e.g. lines 52, 68) treats it as truthy — it will always be truthy, even on mobile. 

**Suggestion:** window.matchMedia('(min-width: 768px)').matches, or better, a stored MediaQueryList with a .matches check.

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1266 @@
...
+
+    &::after {
+      content: "";
+      background-image: url('https://www.adobe.com/dc-shared/fragments/tests/2026/q2/ace1205/fragments/assets/arrow-bottom.svg');
+      position: absolute;
+      padding: var(--s2a-spacing-xs);
+      height: 14px;
+      width: 14px;
+      right: 12px;
+      border-radius: 6px;
```
rgclayton: Design/engineering should be using one of the current tokens for this. 

- --s2a-border-radius-4
- --s2a-border-radius-8

libs/mep/ace1205/hub-hero/hub-hero.js
```diff
@@ -0,0 +1,301 @@
...
+  // assign unique aria-label to the first slide
+  if (index === 0) ariaLabel = `${getCarouselName(link)}, carousel. ${ariaLabel}`;
+
+  const slideEl = createTag('a', {
+    class: 'hub-hero-carousel-item',
+    tabindex: 0,
+    href: link?.href,
+    'data-index': index + 1,
+    role: 'link',
+    ...(isMobile() && {
```
rgclayton, line 208: Since isMobile() always returns truthy (see comment above related to this), the aria-roledescription and aria-label are always applied. 

Is the intent to omit on desktop? Screen readers on desktop still benefit from slide labels.

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1263 @@
...
+  60% {
+    padding: calc(var(--s2a-spacing-lg)*2 + var(--s2a-typography-line-height-eyebrow)) 0 calc(var(--s2a-spacing-xl)*2 + var(--s2a-typography-line-height-eyebrow)) 0;
+    pointer-events: all;
+  }
+
+  100% {
+    padding: calc(var(--s2a-spacing-lg)*2 + var(--s2a-typography-line-height-eyebrow)) 0 calc(var(--s2a-spacing-xl)*2 + var(--s2a-typography-line-height-eyebrow)) 0;
+    pointer-events: all;
+  }
+}
```
robert-bogos: missed the trailing newline

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1263 @@
+/** main classes */
```
robert-bogos: IMO, these comments should be removed. They're just adding noise to the file

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1263 @@
...
+      left: 12px;
+      height: 32px;
+      width: 32px;
+      background-size: contain;
+      background-repeat: no-repeat;
+    }
+
+    &::after {
+      content: "";
+      background-image: url('https://www.adobe.com/dc-shared/fragments/tests/2026/q2/ace1205/fragments/assets/arrow-bottom.svg');
```
robert-bogos: do we want hardcoded svg urls in css? I think is safer to hold this locally, similar to how is done for other C2 blocks

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1263 @@
...
+          }
+
+        }
+      }
+    }
+  }
+}
+
+
+/** animations */
```
robert-bogos: we should have a `prefers-reduced-motion` guard for all these keyframes. Otherwise, we'll have issues with EAA (a11y)

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1263 @@
...
+
+    h1 {
+      color: var(--S2AC-Palette-static-black, black);
+      word-wrap: break-word;
+    }
+  }
+
+  .action-area .con-button {
+    background-color: black;
+    color: white;
```
robert-bogos: we should use s2a css variables for consistency.
I see you're already using `--s2a-color-gray-1000` and `--s2a-color-gray-25` in this file, should be the same for these 2 lines.

robert-bogos (changes requested):

libs/mep/ace1205/hub-hero/hub-hero.js
DKos95: Correct that is my whole point, you don't need to use else if early return in if, it prevents needless nesting and easier to read and maintain in future.

libs/mep/ace1205/hub-hero/hub-hero.js
DKos95: same comment as above, just a matter of not really needing else if early return, easier to read, maintain and prevents nesting code inside and also if in the future we need to add an if clause we don't need to nest them in each other, just a matter of coding style for maintainability.

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1263 @@
...
+  --end-gap: .5rem;
+  --hub-hero-carousel-container-width: calc((var(--slides) * var(--hub-hero-carousel-item-max-width)) + (var(--slides) * (var(--end-gap) * 2)));
+
+  --grid-col-one-offset: -132px;
+  --grid-col-two-offset: 36px;
+  --grid-col-three-offset: 146px;
+  --grid-col-four-offset: 36px;
+  --grid-col-five-offset: -132px;
+
+  max-width: 1920px;
```
narcis-radu: shouldn't this follow the grid? `--grid-max-width-wide`

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1263 @@
...
+    border-radius: 12px;
+    font-size: var(--s2a-typography-font-size-eyebrow);
+    line-height: var(--s2a-typography-line-height-eyebrow);
+    position: relative;
+    cursor: pointer;
+    border: 0;
+
+    &::before {
+      content: "";
+      background-image: url('https://www.adobe.com/cc-shared/assets/img/product-icons/svg/acrobat-pro-64.svg');
```
narcis-radu: I think the approach is incorrect here. The SVG should be authorable, not something fixed.
If we're talking about a generic SVG, it needs to be part of the block, not served from the absolute URL.

libs/mep/ace1205/hub-hero/hub-hero.css
narcis-radu: I second that!

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1263 @@
...
+          }
+
+          .hub-hero-carousel-item-footer p,
+          .hub-hero-carousel-item-header p {
+            color: var(--s2a-color-gray-25);
+          }
+
+          .hub-hero-carousel-item-footer {
+
+            h1,
```
narcis-radu: you've used `:is(h1, h2, h3, h4, h5, h6)` before, you can be consistent

libs/mep/ace1205/hub-hero/hub-hero.css
denlight: That is a good callout, @robert-bogos! What animations should be affected and what exactly should I change in inside prefers-reduced-motion media rule? Is it mainly to kind of slow down the animations or is there a specific list of reqs? Thanks

libs/mep/ace1205/hub-hero/hub-hero.js
denlight: Gooood catch!

libs/mep/ace1205/hub-hero/hub-hero.js
denlight: ahhhhh. thanks!

libs/mep/ace1205/hub-hero/hub-hero.js
denlight: this was introduced by @overmyheadandbody on Elastic Carousel. Not sure why. @overmyheadandbody , could you help us understand? Thanks

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1263 @@
...
+  transform: rotate(180deg);
+  left: 0;
+  right: unset;
+}
+
+html[dir="rtl"] .hub-hero-carousel-item-header::after {
+  transform: rotate(180deg);
+  left: 0;
+  right: unset;
+}
```
overmyheadandbody:
```suggestion
[dir="rtl"] {
  .hub-hero-carousel-item-footer::after,
  .hub-hero-carousel-item-header::after {
    transform: rotate(180deg);
    left: 0;
    right: unset;
  }
}
```

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1263 @@
...
+
+.hub-hero {
+  --hub-hero-height: 4200px;
+  --hub-hero-header-height: 375px;
+  --max-image-size: 388px;
+  --min-image-size: 164px;
+  --grid-img-width: 20vw;
+  --paralax-cubic: cubic-bezier(0.42, 0, 0, 1);
+  --paralax-easing: linear;
```
overmyheadandbody: This will cause confusion with the already existing variable that is used across the system - https://github.com/adobecom/milo/blob/site-redesign-foundation/libs/c2/styles/styles.css#L429. Please use the global `--parallax-easing` instead of the `--parallax-cubic`. This will enable us to easily make system-wide updated in the future. You can either define a hub-hero-specific variable for `linear` or just use the keyword directly where needed.

libs/mep/ace1205/hub-hero/hub-hero.css
overmyheadandbody: This is now `--grid-max-width-default` and should be used

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1263 @@
...
+  word-wrap: break-word;
+}
+
+.hub-hero-header .action-area .con-button {
+  background-color: black;
+  color: var(--s2a-color-gray-25);
+  padding: var(--s2a-spacing-lg) 64px var(--s2a-spacing-lg) 54px;
+  border-radius: var(--s2a-border-radius-sm);
+  font-size: var(--s2a-typography-font-size-eyebrow);
+  line-height: var(--s2a-typography-line-height-eyebrow);
```
overmyheadandbody, line 90: Is `letter-spacing` correct? Otherwise, it should be added as well. I also know that CTAs have been refactored as a whole, so these might not even be needed.

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1263 @@
...
+  position: relative;
+  cursor: pointer;
+  border: 0;
+}
+
+.hub-hero-header .action-area .con-button::before {
+  content: "";
+  background-image: url('https://www.adobe.com/cc-shared/assets/img/product-icons/svg/acrobat-pro-64.svg');
```
overmyheadandbody, line 98: This was brought up in another area, but should be applied here as well - https://github.com/adobecom/milo/pull/5986#discussion_r3322994566

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1263 @@
...
+  left: 12px;
+  height: 32px;
+  width: 32px;
+  background-size: contain;
+  background-repeat: no-repeat;
+}
+
+.hub-hero-header .action-area .con-button::after {
+  content: "";
+  background-image: url('https://www.adobe.com/dc-shared/fragments/tests/2026/q2/ace1205/fragments/assets/arrow-bottom.svg');
```
overmyheadandbody, line 109: This was brought up in another area, but should be applied here as well - https://github.com/adobecom/milo/pull/5986#discussion_r3322994566

libs/mep/ace1205/hub-hero/hub-hero.css
overmyheadandbody: This still stands, if the variable is still needed, it should be `--s2a-typography-font-size-heading-2`. Alternatively, you can apply a `.heading-2` class to the element via JS

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1263 @@
...
+  right: 0;
+  z-index: 1;
+}
+
+.hub-hero-carousel-container {
+  display: flex;
+  position: absolute;
+  justify-content: center;
+  left: 50%;
+  transform: translateX(-50%);
```
overmyheadandbody, line 239: RTL?

libs/mep/ace1205/hub-hero/hub-hero.css
overmyheadandbody: Design team should be consulted for the preferred path. We need to eliminate animations altogether. So the question is: do we still show the initial mosaic and after some regular scroll the elastic-cards-type block is displayed? Or is the elastic-cards-type block shown to begin with if the user prefers reduced motion. This should be asked on the designed channel for clarification.

robert-bogos: Some general guidelines can be found on the [w3 website](https://www.w3.org/WAI/WCAG22/Understanding/animation-from-interactions.html). 
The core idea is that motion and size/layout shifts should not happen when reduced-motion is on

libs/mep/ace1205/hub-hero/hub-hero.js
overmyheadandbody: Elastic cards was using `const isMobile = () => window.innerWidth <= 768;`, so `isMobile` did return the correct value. Previously, `isMobile` was defined as just `const isMobile = () => window.matchMedia('(min-width: 768px)');` in this PR, which, as Ryan said, is always truthy, since it's a function. With the recent addition of `.matches`, this is now correct.

overmyheadandbody (commented):
I'm still concerned by the fact that there is scroll on the X-axis, that needs to be addressed. Most of the comments can be implemented at a later time, but the scrolling is something that impacts page quality.

libs/mep/ace1205/hub-hero/hub-hero.css
rgclayton: I see @denlight started a conversation [in Slack](https://adobe-mwp.slack.com/archives/C09C8DLTNJZ/p1780421083825889) around this. Thank you!

rgclayton:
> I'm still concerned by the fact that there is scroll on the X-axis, that needs to be addressed. Most of the comments can be implemented at a later time, but the scrolling is something that impacts page quality.

I completely agree with this. Can this get solved ASAP to merge in, then circle back to the rest of the suggested fixes?

Doing a quick look @denlight 

It looks like adding o`verflow: clip` to `.hub-hero-carousel` and `.hub-hero-image-grid-container` solves for this. Can you do sanity test on with this and confirm if the horizontal scroll is removed and nothing in your block breaks?

I'll leave comments on the exact lines in code for clarity.

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1263 @@
...
+/** image grid */
+.hub-hero-image-grid-container {
+  display: flex;
+  flex-direction: row;
+  z-index: 1;
+  justify-content: center;
+  animation: hubHeroGridGapShrink var(--paralax-easing) both;
+  animation-timeline: --hub-hero;
+  animation-range: var(--grid-gap-animation-range);
+}
```
rgclayton, line 134: (1 of 2)
Quick testing shows that adding `overflow: clip;` to `.hub-hero-image-grid-container` fixes the horizontal scrolling [noted here](https://github.com/adobecom/milo/pull/5986#issuecomment-4606671138)

libs/mep/ace1205/hub-hero/hub-hero.css
```diff
@@ -0,0 +1,1263 @@
...
+.hub-hero-carousel {
+  height: var(--hub-hero-carousel-container-height-desktop);
+  position: sticky;
+  top: var(--carousel-items-top);
+  overflow: visible;
+  margin: auto;
+  left: 0;
+  right: 0;
+  z-index: 1;
+}
```
rgclayton, line 232: (2 of 2)
Quick testing shows that adding `overflow: clip;` to `.hub-hero-carousel` fixes the horizontal scrolling [noted here](https://github.com/adobecom/milo/pull/5986#issuecomment-4606671138)

overmyheadandbody (approved):
Most of the feedback has been addressed, so approving for now to move into the VQA phase. However, we should open some follow-up stories to address remaining concerns like moving from absolute to relative paths for icons, potential RTL issues and the fact that on mobile the logo ticker is covered by the cards (took the screenshot while having a look at RTL, but issue is reproducible on LTR as well):

Image

libs/mep/ace1205/hub-hero/hub-hero.css
robert-bogos: I see this was fixed, but there are other hardcoded SVGs in the same file. We should fix all of them

robert-bogos (approved):
Approving so that this can go to VQA, but let's create follow up tickets for the feedback that was not addressed. 
From my side, what's left is:
- removing SVG relative paths from the CSS file
- handle the a11y issues coming from the missing `prefers-reduced-motion` guard

DKos95 (approved):

biljana-cvijanovic (approved):
I approve the PR, but I’ve identified several issues that need to be addressed. All of them will be documented in Jira tickets.

PR #5991, opened 2026-05-28
[MWPW-194269] Quick actions block

nishantka:
This PR adds the Quick Actions block for Site Redesign Wave 2.

[Figma Link](https://www.figma.com/design/6FQTDUAttFbukvtskDAkPS/Product-%E2%80%94-A.com?node-id=6254-103017&m=dev)

**Test URLs:**
- After: https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-product?milolibs=quick-actions-block&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all&martech=off

- Before: https://main--da-playground--adobecom.aem.page/drafts/nishantkaush/test-block-6up
- After: https://main--da-playground--adobecom.aem.page/drafts/nishantkaush/test-block-6up?milolibs=quick-actions-block&mep=%2Fdrafts%2Fnishantkaush%2Ftest-pzn-quick.json--+all&martech=off


libs/mep/ace1205/quick-actions/quick-actions.js
```diff
@@ -0,0 +1,62 @@
...
+
+function decorate(block) {
+  const header = decorateSectionHeader(block);
+  const tileRows = [...block.children].filter((row) => row.children.length >= 2);
+  const nUp = N_UP_MAP[tileRows.length] || 'six-up';
+  const grid = createTag('div', { class: `quick-actions-grid ${nUp}` });
+
+  tileRows.forEach((row) => grid.append(buildTile(row)));
+
+  block.innerHTML = '';
```
overmyheadandbody: Maybe `replaceChildren()`, to not have this falsely flagged as a security issue?

libs/mep/ace1205/quick-actions/quick-actions.css
```diff
@@ -0,0 +1,236 @@
+.quick-actions {
+  position: relative;
+
+  .quick-actions-header {
+    grid-column: 1 / -1;
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    gap: var(--s2a-spacing-24);
+    padding-block-end: var(--s2a-spacing-24);
```
DKos95: Probably better to use `--s2a-spacing-lg`

libs/mep/ace1205/quick-actions/quick-actions.css
```diff
@@ -0,0 +1,236 @@
+.quick-actions {
+  position: relative;
+
+  .quick-actions-header {
```
overmyheadandbody: Since the class name is already prefixed with the block name, this doesn't need to be nested, as it likely needlessly increases specificity, making future updates potentially more complicated than they need to be. Same comment for the rest of the file

libs/mep/ace1205/quick-actions/quick-actions.css
```diff
@@ -0,0 +1,236 @@
...
+    &:focus-visible {
+      outline: var(--s2a-border-width-2) solid var(--s2a-color-focus-ring-default);
+      outline-offset: var(--s2a-border-width-2);
+    }
+  }
+
+  .quick-actions-media {
+    position: absolute;
+    inset: 0;
+    inset-block-start: var(--s2a-spacing-8);
```
DKos95: Probably better to use `--s2a-spacing-xs`

libs/mep/ace1205/quick-actions/quick-actions.css
```diff
@@ -0,0 +1,236 @@
...
+  .quick-actions-tile-footer {
+    position: absolute;
+    inset-block-end: 0;
+    inset-inline: 0;
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    gap: var(--s2a-spacing-8);
+    padding-block-end: var(--s2a-spacing-16);
+    padding-inline: var(--s2a-spacing-16);
```
DKos95: Sasme comment as above for variable use

libs/mep/ace1205/quick-actions/quick-actions.css
```diff
@@ -0,0 +1,236 @@
...
+  .quick-actions-tile-label {
+    display: -webkit-box;
+    -webkit-box-orient: vertical;
+    -webkit-line-clamp: 2;
+    overflow: hidden;
+    min-inline-size: 0;
+    font-family: var(--heading-font-family);
+    font-size: var(--s2a-font-size-lg);
+    line-height: var(--s2a-font-line-height-xs);
+    letter-spacing: var(--s2a-font-letter-spacing-neg-0_2);
```
DKos95: `--s2a-font-letter-spacing-5xl`

libs/mep/ace1205/quick-actions/quick-actions.css
```diff
@@ -0,0 +1,236 @@
...
+
+  .quick-actions-header {
+    grid-column: 1 / -1;
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    gap: var(--s2a-spacing-24);
+    padding-block-end: var(--s2a-spacing-24);
+
+    :is(h1, h2, h3, h4) {
```
overmyheadandbody: Heading should already have the correct typography applied, so font-weight might be redundant. On another note, why stop at `h4` and not include h5/h6 as well?

libs/mep/ace1205/quick-actions/quick-actions.css
```diff
@@ -0,0 +1,236 @@
...
+@media (width < 768px) {
+  .quick-actions .quick-actions-grid {
+    grid-template-columns: repeat(2, 1fr);
+    gap: var(--s2a-spacing-2xs);
+  }
+}
+
+@media (width >= 768px) {
+  .quick-actions .quick-actions-header {
+    padding-block-end: var(--s2a-spacing-32);
```
DKos95: `--s2a-spacing-xl`

libs/mep/ace1205/quick-actions/quick-actions.css
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  .quick-actions .quick-actions-tile {
+    block-size: 290px;
+    border-radius: var(--s2a-border-radius-24);
+  }
+
+  .quick-actions .quick-actions-tile-footer {
+    padding-block-end: var(--s2a-spacing-32);
+    padding-inline: var(--s2a-spacing-32);
```
DKos95: `--s2a-spacing-xl`

libs/utils/utils.js
```diff
@@ -109,6 +109,7 @@ const C1_BLOCKS = [
 ];
 
 const C2_BLOCKS = [
+  'quick-actions',
```
DKos95: Since block is in mep this is not needed.

libs/mep/ace1205/quick-actions/quick-actions.css
overmyheadandbody: You could likely apply a `label` class that scales typography correctly - https://github.com/adobecom/milo/blob/site-redesign-foundation/libs/c2/styles/styles.css#L716-L721

DKos95 (approved):
All NIT's that I found, looking good!

libs/mep/ace1205/quick-actions/quick-actions.css
```diff
@@ -0,0 +1,236 @@
...
+@media (width >= 768px) {
+  .quick-actions .quick-actions-header {
+    padding-block-end: var(--s2a-spacing-32);
+  }
+
+  .quick-actions .quick-actions-header :is(h1, h2, h3, h4) {
+    font-size: var(--s2a-font-size-4xl);
+    line-height: var(--s2a-font-line-height-lg);
+    letter-spacing: var(--s2a-font-letter-spacing-2xl);
+  }
```
overmyheadandbody: Applying the right `.heading-X` class would likely get you all of the typography rules you need without having to override via CSS

libs/mep/ace1205/quick-actions/quick-actions.css
```diff
@@ -0,0 +1,236 @@
...
+  .quick-actions .quick-actions-tile-footer {
+    padding-block-end: var(--s2a-spacing-24);
+    padding-inline: var(--s2a-spacing-24);
+  }
+
+  .quick-actions .quick-actions-tile-label {
+    font-size: var(--s2a-font-size-20);
+    line-height: var(--s2a-font-line-height-20);
+    letter-spacing: var(--s2a-font-letter-spacing-neg-0_48);
+  }
```
overmyheadandbody: Applying `.label` would likely scale these correctly OOTB, same comment for the same section in `@media (width >= 2560px)`

libs/mep/ace1205/quick-actions/quick-actions.js
rgclayton: +1

libs/mep/ace1205/quick-actions/quick-actions.js
```diff
@@ -0,0 +1,62 @@
...
+  decorateBlockText(header, { heading: '4', body: 'md', button: 'lg' });
+  return header;
+}
+
+function buildTile(tileRow) {
+  const labelCell = tileRow.children[0];
+  const mediaCell = tileRow.children[1];
+  const labelLink = labelCell?.querySelector('a');
+
+  const tile = createTag('a', { class: 'quick-actions-tile', href: labelLink?.href || '#' });
```
rgclayton: Adding a `#` if there is no `href` scrolls the page to the top, which will be confusing for keyboard users. Consider omitting the <a> entirely when there's no link, or using href: labelLink.href and skipping the fallback.

libs/mep/ace1205/quick-actions/quick-actions.css
nishantka: The .heading-4 is already being added by `decorateBlockText`. But we still need to override for >=768px as we need these values as per design which are provided by heading-4 at 1280px+.

libs/c2/styles/styles.css
```diff
@@ -390,6 +390,7 @@
   --s2a-typography-font-size-body-lg: var(--s2a-font-size-md);
   --s2a-typography-font-size-eyebrow: var(--s2a-font-size-md);
   --s2a-typography-font-size-label: var(--s2a-font-size-sm);
+  --s2a-typography-font-size-label-lg: var(--s2a-font-size-lg);
```
zagi25: Form what I know we shouldn't add properties with `--s2a` prefix as they will be overwritten by the token consumption script ? Am I right @overmyheadandbody ?

libs/mep/ace1205/quick-actions/quick-actions.js
```diff
@@ -0,0 +1,60 @@
...
+  const labelCell = tileRow.children[0];
+  const mediaCell = tileRow.children[1];
+  const labelLink = labelCell?.querySelector('a');
+
+  const tile = createTag('a', { class: 'quick-actions-tile', ...(labelLink && { href: labelLink.href }) });
+
+  const mediaPic = mediaCell?.querySelector('picture');
+  const mediaImg = mediaPic?.querySelector('img') ?? mediaCell?.querySelector('img');
+  if (mediaImg) {
+    mediaImg.alt = '';
```
zagi25: Do we need to remove this alt in code ? From what I know alt is added via authoring so if there is no need for alt authors will not add it

libs/mep/ace1205/quick-actions/quick-actions.js
```diff
@@ -0,0 +1,60 @@
...
+
+  return tile;
+}
+
+const N_UP_MAP = { 2: 'two-up', 3: 'three-up', 4: 'four-up' };
+
+function decorate(block) {
+  const header = decorateSectionHeader(block);
+  const tileRows = [...block.children].filter((row) => row.children.length >= 2);
+  const nUp = N_UP_MAP[tileRows.length] || 'six-up';
```
zagi25, line 48: Shouldn't this be added via authoring ?

libs/c2/styles/styles.css
overmyheadandbody: Indeed, this will disappear at the next token update, the `--s2a-` prefixed tokens need to match the Consonant library exactly. If you have a special need, it's likely specific to a block, thus you could nest it in that appropriate class name.

libs/c2/styles/styles.css
```diff
@@ -720,6 +723,12 @@ p {
   letter-spacing: var(--s2a-typography-letter-spacing-label);
 }
 
+.label-lg {
+  font-size: var(--s2a-typography-font-size-label-lg);
+  line-height: var(--s2a-typography-line-height-label-lg);
+  letter-spacing: var(--s2a-typography-letter-spacing-label-lg);
+}
```
overmyheadandbody: Same token/standardization comment here, these should be standardized from the system perspective, otherwise they should be added at the block level depending on individual needs. and of course this applies to all the other `label-lg` overrides.

libs/mep/ace1205/quick-actions/quick-actions.js
```diff
@@ -0,0 +1,60 @@
...
+  if (mediaImg) {
+    mediaImg.alt = '';
+    (mediaPic ?? mediaImg).classList.add('quick-actions-media');
+    tile.append(mediaPic ?? mediaImg);
+  }
+
+  if (labelLink) {
+    const footer = createTag('div', { class: 'quick-actions-tile-footer' });
+    const chevron = createTag('span', { class: 'quick-actions-chevron', 'aria-hidden': 'true' });
+    chevron.innerHTML = CHEVRON_SVG;
```
overmyheadandbody:
```suggestion
    const chevron = createTag('span', { class: 'quick-actions-chevron', 'aria-hidden': 'true' }, CHEVRON_SVG);
```

libs/mep/ace1205/quick-actions/quick-actions.css
```diff
@@ -0,0 +1,220 @@
...
+.quick-actions-tile:focus-visible {
+  outline: var(--s2a-border-width-2) solid var(--s2a-color-focus-ring-default);
+  outline-offset: var(--s2a-border-width-2);
+}
+
+.quick-actions-media {
+  position: absolute;
+  inset: 0;
+  inset-block-start: var(--s2a-spacing-xs);
+  inset-block-end: 52px; /* padding-block-end(16px) + 2 × line-height(18px) */
```
overmyheadandbody: Are those values coming from tokens? If so, could this be substituted for a `calc`?

libs/mep/ace1205/quick-actions/quick-actions.css
```diff
@@ -0,0 +1,220 @@
...
+    gap: var(--s2a-spacing-2xs);
+  }
+}
+
+@media (width >= 768px) {
+  .quick-actions-header {
+    padding-block-end: var(--s2a-spacing-xl);
+  }
+
+  .quick-actions-header :is(h1, h2, h3, h4) {
```
overmyheadandbody: The list of headings should also be completed here, adding `h5` / `h6`

libs/mep/ace1205/quick-actions/quick-actions.css
```diff
@@ -0,0 +1,220 @@
...
+[dir="rtl"] .quick-actions-chevron {
+  transform: scaleX(-1);
+}
+
+[dir="rtl"] .quick-actions-tile:hover .quick-actions-chevron {
+  transform: scaleX(-1) translate3d(var(--s2a-spacing-4), 0, 0);
+}
+
+@media (width < 768px) {
+  .quick-actions-grid[class*="-up"] {
```
overmyheadandbody: Could this not rely on the existing `.X-up` pattern, to keep the structure consistent?

libs/mep/ace1205/quick-actions/quick-actions.css
```diff
@@ -0,0 +1,220 @@
...
+
+@media (width >= 768px) {
+  .quick-actions-header {
+    padding-block-end: var(--s2a-spacing-xl);
+  }
+
+  .quick-actions-header :is(h1, h2, h3, h4) {
+    font-size: var(--s2a-font-size-4xl);
+    line-height: var(--s2a-font-line-height-lg);
+    letter-spacing: var(--s2a-font-letter-spacing-2xl);
```
overmyheadandbody: I will once more advise against this. Designers should be consulted, as there shouldn't be any reason to stray away from the typography rules we have in place. A `.heading-4` should look the same everywhere it's used, otherwise it defeats the purpose of a system. It's likely this is an omission on the design side, if other values are used.

libs/mep/ace1205/quick-actions/quick-actions.css
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  .quick-actions-tile-footer {
+    padding-block-end: var(--s2a-spacing-xl);
+    padding-inline: var(--s2a-spacing-xl);
+  }
+
+  .quick-actions-chevron svg {
+    inline-size: var(--s2a-spacing-2xs);
+    block-size: var(--s2a-spacing-xs);
```
overmyheadandbody: Tying sizes to spacings is likely not a good idea, as they represent two different dimensions. When spacings are updated, the size of the icon will be unintentionally impacted.

libs/mep/ace1205/quick-actions/quick-actions.js
nishantka: The n-up class here serves a different purpose than a typical authored layout variant. Authors don't control column count per viewport — as per design the block always renders as 2 columns on mobile, 3 on tablet, and 6 on desktop regardless of how many tiles are authored. This mapping is added as a backup to provide better mapping if there are less than 6 tiles authored. Eg. for 4 tiles in desktop -> instead of showing 4 tiles with last 2 empty columns - it would show it in 4-up.

libs/c2/styles/styles.css
nishantka: Done

libs/mep/ace1205/quick-actions/quick-actions.js
nishantka: Done

libs/c2/styles/styles.css
nishantka: Done

libs/mep/ace1205/quick-actions/quick-actions.js
nishantka: Done

libs/mep/ace1205/quick-actions/quick-actions.css
nishantka: Done

libs/mep/ace1205/quick-actions/quick-actions.css
nishantka: Done

libs/mep/ace1205/quick-actions/quick-actions.css
nishantka: Done

libs/mep/ace1205/quick-actions/quick-actions.css
nishantka: Done

libs/mep/ace1205/quick-actions/quick-actions.css
nishantka: I have removed this as I have got the updated figma which shows heading-4 for tablet view also. Although, the value in figma is different which I'll communicate to design team.

narcis-radu (approved):
Small visual issues with alignment, otherwise code looks good.

Image

nishantka:
> Small visual issues with alignment, otherwise code looks good.
> 
> Image

@narcis-radu The image alignment will be fixed by adding appropriate css (`object-fit: cover`) once we have the final assets.

nishantka:
> > Small visual issues with alignment, otherwise code looks good.
> > Image
> 
> @narcis-radu The image alignment will be fixed by adding appropriate css (`object-fit: cover`) once we have the final assets.

I have used the latest assets present in the Asset folder and applied the required css. There is still cropping around 500px to 768px which is similar to what we see in [Figma](https://www.figma.com/design/6FQTDUAttFbukvtskDAkPS/Product-%E2%80%94-A.com?node-id=7311-20078&m=dev) for min-max width designs. I'll confirm with design team if this will be fixed by new assets.

libs/mep/ace1205/quick-actions/quick-actions.css
```diff
@@ -0,0 +1,155 @@
...
+  block-size: var(--s2a-typography-line-height-heading-6);
+  display: flex;
+  align-items: center;
+  color: var(--s2a-color-content-default);
+  transition: transform 0.5s var(--parallax-easing);
+}
+
+.quick-actions-chevron svg {
+  inline-size: 4.5px;
+  block-size: 7.5px;
```
overmyheadandbody: Any chance these could be rounded up/down? Subpixel rendering and eventual animations could lead to some issues down the line.

overmyheadandbody (approved):

robert-bogos (approved):

libs/mep/ace1205/quick-actions/quick-actions.css
nishantka: Sure @overmyheadandbody. I've updated it to 5px X 8px to match base card link chevrons.

NadiiaSokolova (approved):
Verified. Ready to be merged.
Testing details [MWPW-194269](https://jira.corp.adobe.com/browse/MWPW-194269)

PR #5995, opened 2026-05-28
Brand Concierge - C2 dark variant

colloyd:
* Dark variant for C2, support for basic dark variant also.
* Contains other work that is not yet on stage (didn't make rcp) including:
Analytics for the welcome page suggested prompts
Optimization of the floating element scroll handler
Set a custom bottom fade out threshold using the floating-anchor-delay-[amount] variant

Resolves: [MWPW-192877](https://jira.corp.adobe.com/browse/MWPW-192877)

**Test URLs:**
- Before: https://stage--milo--adobecom.aem.page/drafts/colloyd/brand-concierge/bc-floating-button-c2-dark
- After: https://bc-c2-dark--milo--colloyd.aem.page/drafts/colloyd/brand-concierge/bc-floating-button-c2-dark


rgclayton (commented):
Cody and I chatted about adding to `/mep/ace1205/` folder.

colloyd:
> Cody and I chatted about adding to `/mep/ace1205/` folder.

@rgclayton I've added it to the specified folder, does it look right now?

robert-bogos (commented):
Is this work already implemented/tested on C1 and just copied over to C2 using this PR?
`+3,480` lines of code are not easy to review 😅

DKos95:
This PR https://github.com/adobecom/milo/pull/5907 also adds modal.js and modal.css into mep and I do see there are differences between them so either of you will probably encounter conflicts depending on which one reaches the base branch first, to what degree the conflicts will be I am not sure but it can prolong either PR, just a heads up so you aren't surprised.

colloyd:
> This PR #5907 also adds modal.js and modal.css into mep and I do see there are differences between them so either of you will probably encounter conflicts depending on which one reaches the base branch first, to what degree the conflicts will be I am not sure but it can prolong either PR, just a heads up so you aren't surprised.

@DKos95  I will pull the unchanged modal from this PR - but it will have failed eslint tests as a result. As long as everyone is aware and it can be merged anyway.

colloyd:
> Is this work already implemented/tested on C1 and just copied over to C2 using this PR? `+3,480` lines of code are not easy to review 😅

@robert-bogos Yes, there are elements to this that are based on work in this PR that did not make it to stage ahead of RCP, but there are elements that are required here, like updated analytics handling, scroll handler optimization and new options for hiding the floating button that are required for this test. I've also pulled all of the c1 stuff, so this is c2 only.

https://github.com/adobecom/milo/pull/5963

All of that said, a lot of this is test files, so unless tests fail (their not, only lint because I've pulled modal back out) you only need to really look at the 4 block files. This process of adding content to `/mep/ace1205/` does make review harder, because you can't easily see what has changed :(

You can look at the c2 files in the PR above to see the base changes. Aside from those, it's really just these added to the css: https://github.com/adobecom/milo/pull/5995/changes#diff-fc2cae5ddbfe2fdc9a8f6f02ab4be73b137d3e5a6cefac237e273775d72383b3R475-R605

libs/mep/ace1205/brand-concierge/brand-concierge.css
```diff
@@ -0,0 +1,980 @@
...
+  --bc-input-border-size: 2px;
+  --bc-floating-button-height: 48px;
+  --bc-floating-button-spacing: 24px;
+  --bc-input-border-color: rgba(0 0 0 / 12%);
+  --bc-floating-button-border-color: rgba(0 0 0 / 12%);
+  --bc-floating-button-border-color-hover: rgba(0 0 0 / 22%);
+  --bc-button-dimensions: 32px;
+  --bc-button-radius: 50%;
+  --bc-button-color: #292929;
+  --bc-button-hover-color: #131313;
```
DKos95, line 37: These values exist in styles.css already, we should probably use them here to be consistent.

libs/mep/ace1205/brand-concierge/brand-concierge.css
```diff
@@ -0,0 +1,980 @@
...
+  --bc-input-font-size: 16px;
+  --bc-input-line-height: 1.5;
+  --bc-input-radius: 26px;
+  --bc-input-height: 52px;
+  --bc-input-border-size: 2px;
+  --bc-floating-button-height: 48px;
+  --bc-floating-button-spacing: 24px;
+  --bc-input-border-color: rgba(0 0 0 / 12%);
+  --bc-floating-button-border-color: rgba(0 0 0 / 12%);
+  --bc-floating-button-border-color-hover: rgba(0 0 0 / 22%);
```
DKos95, line 33: `rgba(0 0 0 / 12%)` already exists in styles.css as `-s2a-color-transparent-black-12`, `rgba(0 0 0 / 22%)` should be close enough to `  --s2a-color-transparent-black-24: rgb(0 0 0 / 24%);`

libs/mep/ace1205/brand-concierge/brand-concierge.css
```diff
@@ -0,0 +1,980 @@
...
+  --bc-button-dimensions: 32px;
+  --bc-button-radius: 50%;
+  --bc-button-color: #292929;
+  --bc-button-hover-color: #131313;
+  --bc-keyboard-focus-color: #5574F7;
+  --bc-dark-text-color: #fff;
```
DKos95, line 39: Already exists in styles.css

libs/mep/ace1205/brand-concierge/brand-concierge.css
```diff
@@ -0,0 +1,980 @@
...
+  --bc-dark-text-color-secondary: #aba8a6;
+  --bc-dark-border-color: rgba(255 255 255 / 20%);
+  --bc-dark-border-color-secondary: rgba(255 255 255 / 30%);
+  --bc-dark-input-background: #131313;
+  --bc-dark-button-color: rgba(255 255 255 / 80%);
+  --bc-dark-button-hover-color: rgba(255 255 255 / 90%);
+
+  box-sizing: border-box;
+  padding: 40px 24px 12px;
```
DKos95, line 48: Similar thing, we should use tokens 
```suggestion
  padding: var(--s2a-spacing-2xl) var(--s2a-spacing-lg) var(--s2a-spacing-sm);
```

libs/mep/ace1205/brand-concierge/brand-concierge.css
```diff
@@ -0,0 +1,980 @@
...
+  }
+}
+
+.bc-header {
+  color: var(--bc-header-color);
+  display: flex;
+  flex-direction: column;
+  gap: 4px;
+  text-align: center;
+  margin-bottom: 32px;
```
DKos95, line 74:
```suggestion
  margin-bottom: var(--s2a-spacing-xl);
```

libs/mep/ace1205/brand-concierge/brand-concierge.css
```diff
@@ -0,0 +1,980 @@
...
+.bc-header-subtitle {
+  font-size: var(--bc-header-subtitle-font-size);
+  line-height: var(--bc-header-subtitle-line-height);
+  margin: 0;
+}
+
+.bc-prompt-cards {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 12px;
```
DKos95, line 93:
```suggestion
  gap: var(--s2a-spacing-sm);
```

libs/mep/ace1205/brand-concierge/brand-concierge.css
```diff
@@ -0,0 +1,980 @@
...
+  margin-bottom: 40px;
```
DKos95, line 94:
```suggestion
  margin-bottom: var(--s2a-spacing-2xl);
```

libs/mep/ace1205/brand-concierge/brand-concierge.css
```diff
@@ -0,0 +1,980 @@
...
+}
+
+.prompt-card-button {
+  background: #fff;
+  border: 1px solid var(--bc-card-border-color);
+  border-radius: 12px;
```
DKos95, line 100:
```suggestion
  border-radius: var(--s2a-border-radius-12);
```

libs/mep/ace1205/brand-concierge/brand-concierge.css
```diff
@@ -0,0 +1,980 @@
...
+.prompt-card-image {
+  display: none;
+}
+
+.prompt-card-text {
+  color: var(--bc-card-text-color);
+  display: grid;
+  grid-template-columns: 16px auto;
+  font-size: var(--bc-card-font-size);
+  gap: 8px;
```
DKos95, line 125:
```suggestion
  gap: var(--s2a-border-radius-sm);
```

libs/mep/ace1205/brand-concierge/brand-concierge.css
```diff
@@ -0,0 +1,980 @@
...
+  line-height: var(--bc-card-line-height);
+  margin: 0;
+  padding: 12px 16px;
```
DKos95, line 128:
```suggestion
  padding: var(--s2a-spacing-sm) var(--s2a-spacing-md);
```

libs/mep/ace1205/brand-concierge/brand-concierge.css
```diff
@@ -0,0 +1,980 @@
...
+  text-align: left;
+}
+
+.prompt-card-text .card-icon {
+  align-self: flex-start;
+  margin-top: 4px;
```
DKos95, line 134:
```suggestion
  margin-top: var(--s2a-spacing-2xs);
```

libs/mep/ace1205/brand-concierge/brand-concierge.css
```diff
@@ -0,0 +1,980 @@
...
+}
+
+.bc-input-field,
+.bc-floating-button {
+  margin-bottom: 40px;
+  position: relative;
+}
+
+.input-first .bc-input-field {
+  margin-bottom: 24px;
```
DKos95, line 151:
```suggestion
  margin-bottom: var(--s2a-spacing-lg);
```

libs/mep/ace1205/brand-concierge/brand-concierge.css
```diff
@@ -0,0 +1,980 @@
...
+}
+
+.prompt-card-text p {
+  color: var(--bc-card-text-color);
+  margin: 0;
+}
+
+.bc-input-field,
+.bc-floating-button {
+  margin-bottom: 40px;
```
DKos95, line 146:
```suggestion
  margin-bottom: var(--s2a-spacing-2xl);
```

libs/mep/ace1205/brand-concierge/brand-concierge.css
```diff
@@ -0,0 +1,980 @@
...
+}
+
+.input-first .bc-input-field {
+  margin-bottom: 24px;
+}
+
+.bc-input-field-container,
+.bc-floating-button-container {
+  align-items: center;
+  background: #fff;
```
DKos95, line 157: Probably

```suggestion
  background: var(--s2a-color-background-default);
```

libs/mep/ace1205/brand-concierge/brand-concierge.css
```diff
@@ -0,0 +1,980 @@
...
+  border-radius: calc(var(--bc-input-radius) - var(--bc-input-border-size));
+  box-shadow: 0 4px 12px 0 rgba(0 0 0 / 6%);
+  display: flex;
+  min-height: var(--bc-input-height);
+  margin-bottom: 40px;
```
DKos95, line 162:
```suggestion
  margin-bottom: var(--s2a-spacing-2xl);
```

libs/mep/ace1205/brand-concierge/brand-concierge.css
```diff
@@ -0,0 +1,980 @@
...
+  padding: 0 8px;
+  position: relative;
+  gap: 12px;
```
DKos95, line 165:
```suggestion
  gap: var(--s2a-spacing-sm);
```

libs/mep/ace1205/brand-concierge/brand-concierge.css
```diff
@@ -0,0 +1,980 @@
...
+}
+
+.bc-input-field::before,
+.bc-floating-button::before {
+  border: var(--bc-input-border-size) solid var(--bc-input-border-color);
+  border-radius: var(--bc-input-radius);
+  content: '';
+  backdrop-filter: blur(10px);
+  height: 100%;
+  left: 50%;
```
DKos95, line 176: Potential rtl issue?

libs/mep/ace1205/brand-concierge/brand-concierge.js
```diff
@@ -0,0 +1,725 @@
...
+    if (hasDelay) {
+      if (window.scrollY > topDelay && threshold <= mainHeight) {
+        showFloating();
+      }
+      if (window.scrollY < topDelay
+        || (variants.floatingAnchorDelay && threshold > mainHeight - anchorDelay)) {
+        hideFloating();
+      }
+    }
+  };
```
DKos95, line 118: Would something like this also work, should be a bit more compact?
```suggestion
const handleScroll = (target) => {
  const threshold = window.scrollY + window.innerHeight - mainTop;
  const topDelay = variants.floatingDelay ? variants.floatingDelayAmount : elHeight;
  const pastMain = threshold > mainHeight;
  const shouldHideAnchor = variants.isFloatingAnchorHide || variants.floatingAnchorDelay;

  target.style.bottom = pastMain ? `${threshold - mainHeight}px` : '0';

  if (!pastMain) showFloating();
  else if (shouldHideAnchor) hideFloating();
  else floatingSpacer.style.cssText = `height: ${targetHeight}px; pointer-events: none; display: block;`;

  if (!hasDelay) return;

  if (window.scrollY > topDelay && !pastMain) showFloating();
  if (window.scrollY < topDelay || (variants.floatingAnchorDelay && threshold > mainHeight - anchorDelay))   hideFloating();
};

```

libs/mep/ace1205/brand-concierge/brand-concierge.js
```diff
@@ -0,0 +1,725 @@
...
+        resolve(true);
+      } else if (Date.now() - startTime >= timeout) {
+        resolve(false);
+      } else {
+        setTimeout(check, interval);
+      }
+    };
+    check();
+  });
+}
```
DKos95, line 176:
```suggestion
function waitForCondition(checkFn, timeout = 5000, interval = 100) {
  return new Promise((resolve) => {
    const start = Date.now();
    (function check() {
      if (checkFn()) resolve(true);
      else if (Date.now() - start >= timeout) resolve(false);
      else setTimeout(check, interval);
    })();
  });
}

```

libs/mep/ace1205/brand-concierge/brand-concierge.css
colloyd:
```suggestion
  --bc-button-color: #292929;
  --bc-button-hover-color: #131313;
```

Right, so this code does have to work properly for both c1 and c2 - most of the implementations are actually c1. To keep sanity in working with these, we are trying to keep most of these files consistent between both, hence all of the local variables instead of relying on styles.css that are inconsistent between c1 and c2.

libs/mep/ace1205/brand-concierge/brand-concierge.css
colloyd: See the other comment: this code does have to work properly for both c1 and c2 - most of the implementations are actually c1. To keep sanity in working with these, we are trying to keep most of these files consistent between both, hence all of the local variables instead of relying on styles.css that are inconsistent between c1 and c2.

libs/mep/ace1205/brand-concierge/brand-concierge.css
colloyd: Same as above: this code does have to work properly for both c1 and c2 - most of the implementations are actually c1. To keep sanity in working with these, we are trying to keep most of these files consistent between both, hence all of the local variables instead of relying on styles.css that are inconsistent between c1 and c2.

libs/mep/ace1205/brand-concierge/brand-concierge.css
colloyd: same as above: this code does have to work properly for both c1 and c2 - most of the implementations are actually c1. To keep sanity in working with these, we are trying to keep most of these files consistent between both, hence all of the local variables instead of relying on styles.css that are inconsistent between c1 and c2.

libs/mep/ace1205/brand-concierge/brand-concierge.css
colloyd: same as above: this code does have to work properly for both c1 and c2 - most of the implementations are actually c1. To keep sanity in working with these, we are trying to keep most of these files consistent between both, hence all of the local variables instead of relying on styles.css that are inconsistent between c1 and c2.

libs/mep/ace1205/brand-concierge/brand-concierge.css
colloyd: same as above: this code does have to work properly for both c1 and c2 - most of the implementations are actually c1. To keep sanity in working with these, we are trying to keep most of these files consistent between both, hence all of the local variables instead of relying on styles.css that are inconsistent between c1 and c2.

libs/mep/ace1205/brand-concierge/brand-concierge.css
colloyd: same as above: this code does have to work properly for both c1 and c2 - most of the implementations are actually c1. To keep sanity in working with these, we are trying to keep most of these files consistent between both, hence all of the local variables instead of relying on styles.css that are inconsistent between c1 and c2.

libs/mep/ace1205/brand-concierge/brand-concierge.js
colloyd: There are a lot of use cases that this has to support - interaction between different variants. Also, this whole JS file will soon be refactored, with several of this sort of helper function being moved out into a sort of bc-utils file and imported. I'm not terribly concerned with code being concise, given all of the factors. Perhaps we can reduce that as part of that larger refactor.

libs/mep/ace1205/brand-concierge/brand-concierge.js
colloyd: Same situation with this - it will be moved out in an impending refactor. I'm hesitant to make any changes to what we know works with the runway for the test shrinking rapidly.

libs/mep/ace1205/brand-concierge/brand-concierge.css
colloyd: No. This renders the gradient border around the input field. RTL should not have any impact as this is essentially a decorative element.

rgclayton (approved):

libs/mep/ace1205/brand-concierge/brand-concierge.css
```diff
@@ -0,0 +1,980 @@
...
+
+.brand-concierge.has-bg-image {
+  background: var(--brand-concierge-bg) 50% / cover no-repeat;
+}
+
+.brand-concierge.floating-button-only {
+  padding: 0;
+
+  /* TODO: likely not all pages will be dark, this needs more granular control */
+  main:has(&) {
```
narcis-radu, line 63: @colloyd - I'm concerned about this, authors will not be able to work around this limitation. Do you have a demo page for this particular use-case? I was not able to see the `floating-button-only` on https://bc-c2-dark--milo--colloyd.aem.page/drafts/colloyd/brand-concierge/bc-floating-button-c2-dark

narcis-radu (dismissed):

libs/mep/ace1205/brand-concierge/brand-concierge.css
colloyd: @narcis-radu The use case has never been present on site redesign - the button has always been hidden on site redesign using `floating-anchor-hide`. This css has been there for a while (I didn't actually write this) and I've left it as a catchall for this situation. 

Further, with this new placement meant to be at the bottom, I've introduced a custom threshold at which the button is hidden so it doesn't run over the inline element. See the description for: `floating-anchor-delay-[amount]`

As for a demo page, mine was broken when I moved all of this to `/mep/ace1205/`, since my changes are no longer in base c2.

colloyd: I've made the selector much more specific to only the case where a floating-button-only is not hidden.

narcis-radu: ~~I see, this has been added by @JasonHowellSlavin some time ago.~~
~~@JasonHowellSlavin - we're hardcoding background, I don't think this is good.~~

colloyd: Actually, it was added by @overmyheadandbody when we first introduced brand-concierge to c2.

overmyheadandbody: I've intentionally stayed away from this block, as I don't have the knowledge of its inner-workings, so I'm not sure where this could come from. I see it existed before any C2 work, https://github.com/colloyd/milo/blame/0f48d3bca455e54a5a6243576c5b2ea2e31f8a45/libs/blocks/brand-concierge/brand-concierge.css#L56-L58, introduced by @JasonHowellSlavin. Any tips as to where I made some changes here?

colloyd: @overmyheadandbody it was done here: https://github.com/adobecom/milo/pull/5640

PR #6000, opened 2026-05-29
[SR][MWPW-193871] Side by side

zagi25:
* Add `side-by-side` C2 block
* Fix `viewportplay` 
* Add different `play/pause` button decoration in `decorate.js` for C2 pages

Resolves: https://jira.corp.adobe.com/browse/MWPW-193871

**QE NOTE:** Since this touches global decorate logic for videos, please test videos on C1 pages. Everything should work and look same as before except `viewportplay` videos, before they were played on load because of the `autoplay`, now they should only play if video is in viewport. Also please check `router-marquee`, to me it looks like this doesn't affect router-marquee play/pause buttons, but it is better to double check. 

**Test URLs:**
- Before: https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-bizpro-hub?milolibs=site-redesign-foundation&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all
- After: https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-bizpro-hub?milolibs=mwpw-193871-side-2-side&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all

**Rich content video:**
- After: https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-product?milolibs=mwpw-193871-side-2-side&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all

**C1 video page:**
- Before: https://main--milo--adobecom.aem.page/drafts/ratko/mwpw-193871-side-2-side/document1
- After: https://mwpw-193871-side-2-side--milo--adobecom.aem.page/drafts/ratko/mwpw-193871-side-2-side/document1


libs/blocks/video/video.css
```diff
@@ -164,15 +164,86 @@ video {
...
 
-/* C2 foundation overrides */
-:root:has(meta[name="foundation"][content="c2"]) {
-  .video-container .pause-play-wrapper {
-    .offset-filler {
-      background: var(--s2a-color-gray-800);
+button.play-pause-button {
+  width: 32px;
+  height: 32px;
+  border-radius: var(--s2a-border-radius-4);
```
DKos95: Probably better to use `--s2a-border-radius-xs`

libs/blocks/video/video.css
```diff
@@ -164,15 +164,86 @@ video {
...
+  align-items: center;
+  justify-content: center;
+  position: absolute;
+  z-index: 1;
+  right: var(--s2a-spacing-md);
+  bottom: var(--s2a-spacing-md);
+
+  svg path {
+    fill: var(--s2a-color-gray-1000);
+    stroke: var(--s2a-color-gray-1000);
```
DKos95: There are a couple of semantic tokens that use this color, can we use one of them?

libs/blocks/video/video.css
```diff
@@ -164,15 +164,86 @@ video {
...
+    outline-offset: 1px;
+  }
+
+  &:active {
+    border: var(--s2a-border-width-sm) solid var(--s2a-color-transparent-white-24);
+    background: var(--s2a-color-background-knockout);
+
+    svg path {
+      fill: var(--s2a-color-gray-25);
+      stroke: var(--s2a-color-gray-25);
```
DKos95: Same comment as above.

libs/blocks/video/video.css
```diff
@@ -164,15 +164,86 @@ video {
...
+  }
+}
+
+.dark button.play-pause-button {
+  border: var(--s2a-border-width-sm) solid var(--s2a-color-transparent-white-12);
+  background: var(--s2a-color-transparent-white-16);
+
+  svg path {
+    fill: var(--s2a-color-gray-25);
+    stroke: var(--s2a-color-gray-25);
```
DKos95: Same comment as above.

libs/blocks/video/video.css
```diff
@@ -164,15 +164,86 @@ video {
...
+  &:hover {
+    background: var(--s2a-color-transparent-white-12);
+  }
+
+  &:active {
+    background: var(--s2a-color-content-knockout);
+
+    svg path {
+      fill: var(--s2a-color-gray-1000);
+      stroke: var(--s2a-color-gray-1000);
```
DKos95: Same comment as above.

libs/mep/ace1205/side-by-side/side-by-side.css
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  .card-overlay {
+    --parallax-stagger-index: 0.5;
+
+    position: relative;
+    overflow: hidden;
+    min-height: 305px;
+    aspect-ratio: 33/37;
+    border: var(--s2a-border-width-2) solid var(--s2a-color-transparent-white-12);
```
DKos95: Maybe `--s2a-border-width-md`?

libs/mep/ace1205/side-by-side/side-by-side.css
```diff
@@ -0,0 +1,135 @@
...
+      min-width: 233px;
+    }
+
+    .foreground {
+      padding: var(--s2a-spacing-md);
+      display: flex;
+      flex-direction: column;
+      gap: var(--s2a-spacing-xs);
+      max-width: 400px;
+      color: var(--s2a-color-gray-1000);
```
DKos95: Also can be a semantic token.

libs/utils/utils.js
```diff
@@ -125,6 +125,7 @@ const C2_BLOCKS = [
   'rich-content',
   'router-marquee',
   'section-metadata',
+  'side-by-side',
```
DKos95: Not needed since block is in mep.

DKos95 (approved):
Only some minor NIT's found, looking good!

libs/blocks/video/video.css
zagi25: I will leave this as is, because it is needed to override semantic token that is used in `light` variant of the button

libs/blocks/video/video.css
```diff
@@ -164,15 +164,81 @@ video {
...
-:root:has(meta[name="foundation"][content="c2"]) {
-  .video-container .pause-play-wrapper {
-    .offset-filler {
-      background: var(--s2a-color-gray-800);
+button.play-pause-button {
+  width: 32px;
+  height: 32px;
+  border-radius: var(--s2a-border-radius-xs);
+  border: var(--s2a-border-width-sm) solid var(--s2a-color-transparent-black-08);
+  background: var(--s2a-color-transparent-black-04);
```
overmyheadandbody: `--s2a` variables are added to the page only if the `c2` `foundation` is set via metadata. So these might not yield the expected results. Am I missing something? Otherwise we might need two different sets of styles or fallbacks for each S2A variable. This, of course, applies to the whole PR

libs/blocks/video/video.css
```diff
@@ -164,15 +164,81 @@ video {
...
+    background: var(--s2a-color-content-knockout);
+
+    svg path {
+      fill: var(--s2a-color-gray-1000);
+      stroke: var(--s2a-color-gray-1000);
+    }
+  }
+}
+
+@media (width >= 1280px) {
```
overmyheadandbody: Also, keep in mind C1 and C2 have different breakpoints.

libs/mep/ace1205/side-by-side/side-by-side.css
```diff
@@ -0,0 +1,134 @@
...
+    .media {
+      position: absolute;
+      inset: 0;
+    }
+
+    .media::after {
+      content: '';
+      position: absolute;
+      inset: 0;
+      background: linear-gradient(144deg, rgba(0, 0, 0, 0.92) 6.95%, rgba(0, 0, 0, 0.00) 27.98%);
```
overmyheadandbody: Any chance we can sprinkle in some tokens? `--s2a-color-gray-1000` instead of `rgba(0, 0, 0, 0.00)`, maybe `--s2a-color-transparent-black-88` instead of `rgba(0, 0, 0, 0.92)`. Alternatively, maybe we can at least switch to the more modern notation of `rgb(0 0 0 / 0%)`

libs/mep/ace1205/side-by-side/side-by-side.css
```diff
@@ -0,0 +1,134 @@
...
+      }
+    }
+  }
+}
+
+
+
+@media (width >= 768px) {
+  .side-by-side {
+    grid-template-columns: 1.81fr 1fr;
```
overmyheadandbody, line 95: Where does the `1.81` come from?

overmyheadandbody (commented):
My main concern is with the S2A variables in `video.css`, hopefully I'm just ignorant of a mechanism that ties it together correctly.

libs/blocks/video/video.css
zagi25: `button` should only appear on the C2 pages, C1 pages will get old CTA/button with `a` tag and `img` etc. check `addAccessibilityControl` function. Or do we want to apply this structure with old style for C1 pages also ?

zagi25: I can add `:root:has(meta[name="foundation"][content="c2"])` but this will add more specificity and complicate things if anything needs to be overwritten on the block level

libs/mep/ace1205/side-by-side/side-by-side.css
zagi25: It was calculated by Claude by comparing left and right column widths from the design.

libs/blocks/video/video.css
overmyheadandbody: Oh, I see, thanks for the explanation! I'd still suggest the previous separation via `:root:has(meta[name="foundation"][content="c2"]) {`, if it doesn't cause any issues. That would help clearly separate things for future reviews/updates or for when we're ready to pull the plug on C1

narcis-radu:
@zagi25 - with the latest code change I'm seeing some issues in the rich content video
Screenshot 2026-06-01 at 12 24 00

SilviuLCF (approved):
QA verified https://jira.corp.adobe.com/browse/MWPW-193871

zagi25:
> @zagi25 - with the latest code change I'm seeing some issues in the rich content video Screenshot 2026-06-01 at 12 24 00

@narcis-radu fixed, thanks!

narcis-radu (approved):

libs/mep/ace1205/side-by-side/side-by-side.css
overmyheadandbody: This will be a mystery in a few months from now when a change will be needed, but it's also not a strong enough reason to postpone this PR

PR #6002, opened 2026-05-29
[MWPW-194266] - Sticky CTA

Deva309:
### Sticky CTA Block.

Block Authoring:
Screenshot 2026-05-29 at 6 56 01 PM

Resolves: [MWPW-194266](https://jira.corp.adobe.com/browse/MWPW-194266)

**Test URL:**
https://main--upp--adobecom.aem.page/homepage/drafts/devashish/floating-cta?milolibs=sticky-cta


libs/c2/assets/img/promo-arrow-right-black.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="12" height="12" viewBox="0 0 12 12" fill="none">
+<path d="M11.208 5.417L7.508 1.717C7.186 1.395 6.664 1.395 6.342 1.717 6.02 2.039 6.02 2.561 6.342 2.883L8.633 5.175H1.375C.919 5.175.55 5.544.55 6 .55 6.456.919 6.825 1.375 6.825H8.633L6.342 9.117C6.02 9.44 6.02 9.961 6.342 10.283 6.503 10.444 6.714 10.525 6.925 10.525 7.136 10.525 7.347 10.444 7.508 10.283L11.208 6.583C11.53 6.261 11.53 5.739 11.208 5.417Z" fill="black"/>
```
overmyheadandbody: Since the icon is identical except the color, we could just adjust the colors via CSS, right? It doesn't look like we need two different icons.

libs/mep/ace1205/floating-cta/floating-cta.js
```diff
@@ -0,0 +1,50 @@
...
+
+  const linkEl = linkPara.querySelector('a');
+  const sourceText = (linkEl ? linkEl.textContent : linkPara.textContent).trim();
+  const [ctaText, ariaLabel = ctaText] = sourceText.split('|').map((s) => s.trim());
+  const ctaHref = linkEl?.getAttribute('href') || '#';
+
+  const cta = createTag('a', { href: ctaHref, class: 'promo-cta', 'aria-label': ariaLabel });
+  cta.appendChild(img);
+  cta.append(ctaText);
+  cta.appendChild(createTag('span', { class: 'icon-button arrow', 'aria-hidden': 'true' }));
```
overmyheadandbody:
```suggestion
  const arrow = createTag('span', { class: 'icon-button arrow', 'aria-hidden': 'true' });
  const cta = createTag('a', { href: ctaHref, class: 'promo-cta', 'aria-label': ariaLabel }, [img, ctaText, arrow]);
```

libs/c2/styles/styles.css
```diff
@@ -942,6 +942,66 @@ p {
   }
 }
 
+/* Icon Button Styles */
+.icon-button {
+  align-items: center;
+  background-color: var(--s2a-color-transparent-white-16);
+  block-size: var(--s2a-spacing-xl);
```
overmyheadandbody: Better have this explicit, as spacings could change in future token package updates. Same for `inline-size: var(--s2a-spacing-xl);` below.

libs/c2/styles/styles.css
```diff
@@ -942,6 +942,66 @@ p {
...
+  border: var(--s2a-border-width-sm) solid var(--s2a-color-transparent-white-12);
+  border-radius: 6px; /* No token */
+  box-sizing: border-box;
+  display: inline-flex;
+  flex-shrink: 0;
+  inline-size: var(--s2a-spacing-xl);
+  justify-content: center;
+  transition: background-color 0.2s ease, border-color 0.2s ease, filter 0.2s ease;
```
overmyheadandbody: Maybe add a `will-change` in this case?

libs/c2/styles/styles.css
```diff
@@ -942,6 +942,66 @@ p {
...
+  gap: var(--s2a-spacing-sm);
+  letter-spacing: var(--s2a-typography-letter-spacing-eyebrow);
+  line-height: var(--s2a-typography-line-height-eyebrow);
+  padding: var(--s2a-spacing-sm);
+  text-decoration: none;
+  white-space: nowrap;
+  img {
+    block-size: var(--s2a-spacing-xl);
+    flex-shrink: 0;
+    inline-size: var(--s2a-spacing-xl);
```
overmyheadandbody: Same dimensions x spacing token conflict, better have them explicit

libs/mep/ace1205/floating-cta/floating-cta.js
```diff
@@ -0,0 +1,50 @@
...
+
+  const [imgPara, linkPara] = contentDiv.querySelectorAll('p');
+  if (!imgPara || !linkPara) return;
+
+  const img = imgPara.querySelector('img');
+  if (!img) return;
+
+  const relativeSrc = img.getAttribute('src');
+  if (relativeSrc?.startsWith('/')) {
+    img.src = getFederatedUrl(relativeSrc);
```
zagi25, line 17: I think we don't need this, svg added via link should have a full link. Also if I am not wrong this will limit usage of the project specific svgs

libs/mep/ace1205/floating-cta/floating-cta.css
```diff
@@ -0,0 +1,44 @@
...
+      &:has(.floating-cta) {
+        margin-block: var(--s2a-spacing-xl) calc(-1 * var(--s2a-spacing-xl));
+      }
+    }
+  }
+}
+
+.floating-cta {
+  display: flex;
+  justify-content: center;
```
zagi25, line 19: This can be moved bellow under `.floating-cta`

libs/utils/utils.js
```diff
@@ -116,6 +116,7 @@ const C2_BLOCKS = [
   'elastic-carousel',
   'explore-card',
   'faq',
+  'floating-cta',
```
DKos95: Not needed since block is in mep

libs/c2/assets/img/promo-arrow-right-black.svg
Deva309: @overmyheadandbody Now, I am using the SVG directly as HTML content, which makes it easy to customize the color. However, since this icon is part of the promo CTA (with the arrow) and may be reused in other places. 
Do you think maintaining these icons in `icons.js` is the good approach, or if it would be better to keep them in a `utils.js`?

libs/mep/ace1205/floating-cta/floating-cta.js
Deva309: Done.

libs/c2/styles/styles.css
Deva309: Done

libs/c2/styles/styles.css
Deva309: Done

libs/c2/styles/styles.css
Deva309: Done

libs/mep/ace1205/floating-cta/floating-cta.js
Deva309: It works with full url too. I tested both authoring types, and it works fine because the method checks for /federal/ and converts it to a federal URL. Otherwise, it returns the original full URL.

libs/mep/ace1205/floating-cta/floating-cta.css
Deva309: Done

libs/utils/utils.js
Deva309: Done

libs/mep/ace1205/floating-cta/floating-cta.css
```diff
@@ -0,0 +1,41 @@
+.section {
+  &:has(.floating-cta) {
+    bottom: 0;
+    position: sticky;
+    z-index: 4;
+  }
+  &.rounded-corners-bottom,
```
narcis-radu, line 7: shouldn't we also target `rounded-corners-top` ?

narcis-radu (approved):

libs/mep/ace1205/floating-cta/floating-cta.css
Deva309: As per prototype, it is always sticky to the bottom of a section so applied for `rounded-corners-bottom` only but in future if needed we can add this as well.

libs/c2/styles/styles.css
```diff
@@ -942,6 +942,65 @@ p {
   }
 }
 
+/* Icon Button Styles */
+.icon-button {
+  align-items: center;
+  background-color: var(--s2a-color-transparent-white-16);
+  block-size: var(--s2a-spacing-32);
```
overmyheadandbody: This was brought up before, but I see it hasn't changed. Width/height should not be based on spacing tokens, these can be explicit unit values. Same for L955

libs/c2/styles/styles.css
```diff
@@ -942,6 +942,65 @@ p {
...
+  font-size: var(--s2a-typography-font-size-eyebrow);
+  font-weight: var(--s2a-font-weight-eyebrow);
+  gap: var(--s2a-spacing-sm);
+  letter-spacing: var(--s2a-typography-letter-spacing-eyebrow);
+  line-height: var(--s2a-typography-line-height-eyebrow);
+  padding: var(--s2a-spacing-sm);
+  text-decoration: none;
+  white-space: nowrap;
+  img {
+    block-size: var(--s2a-spacing-32) ;
```
overmyheadandbody:
```suggestion
    block-size: var(--s2a-spacing-32);
```
And the same comment around using spacing tokens for width/height, it should be avoided. Since I see this repeating, you can just scope a variable with the desired size to the block itself, like
```
.icon-button {
  --icon-size: 32px;
  ...
}
```

libs/mep/ace1205/floating-cta/floating-cta.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  const linkEl = linkPara.querySelector('a');
+  const sourceText = (linkEl ? linkEl.textContent : linkPara.textContent).trim();
+  const [ctaText, ariaLabel = ctaText] = sourceText.split('|').map((s) => s.trim());
+  const ctaHref = linkEl?.getAttribute('href') || '#';
+
+  const arrow = createTag('span', { class: 'icon-button', 'aria-hidden': 'true' });
+  arrow.innerHTML = icons.arrowRightWhite;
```
overmyheadandbody: This could cause a false positive from Kodiak, you could likely add this through `createTag`

overmyheadandbody (approved):

libs/mep/ace1205/floating-cta/floating-cta.css
```diff
@@ -0,0 +1,41 @@
...
+    will-change: opacity, clip-path, transform;
+    &.active {
+      clip-path: inset(0% round var(--s2a-border-radius-12));
+      opacity: 1;
+      transform: translateY(0);
+      transition:
+        opacity 0.38s cubic-bezier(0.25, 0.46, 0.45, 0.94),
+        clip-path 0.38s cubic-bezier(0.25, 0.46, 0.45, 0.94),
+        transform 0.45s cubic-bezier(0.34, 1.56, 0.64, 1);
+    }
```
robert-bogos, line 43: We should add a guard to prevent these animations if reduced-motion is on

libs/mep/ace1205/floating-cta/floating-cta.js
```diff
@@ -0,0 +1,50 @@
...
+  const customHide = metadata['custom-hide']?.text[0];
+  if (!customHide) return;
+
+  // Intersection Observer on section selected by custom-hide in section metadata
+  const customHideSection = document.querySelector(customHide);
+  if (!customHideSection) return;
+  const observer = new IntersectionObserver(
+    ([entry]) => {
+      cta.classList.toggle('active', !entry.isIntersecting);
+      cta.setAttribute('aria-hidden', String(entry.isIntersecting));
```
robert-bogos: We should also set `tabindex="-1"` to take it out of the keyboard navigation, if we want it hidden from screen readers

robert-bogos (commented):
some a11y issues

libs/c2/styles/styles.css
Deva309: Done

libs/c2/styles/styles.css
Deva309: Done

libs/mep/ace1205/floating-cta/floating-cta.js
Deva309: Done

libs/mep/ace1205/floating-cta/floating-cta.css
Deva309: Done

libs/mep/ace1205/floating-cta/floating-cta.js
Deva309: Done

robert-bogos (approved):
Thank you for taking care of the feedback ❤️

zagi25 (approved):

SilviuLCF (approved):
Ready to be merged , testing details 
https://jira.corp.adobe.com/browse/MWPW-194266

PR #6019, opened 2026-06-02
Fixing region modal fragment's max height for ios issue

bandana147:
* Added max height of fragment for ios devices to 100vdh instead of vh

Resolves: [MWPW-197330](https://jira.corp.adobe.com/browse/MWPW-197330)

**Test URLs:**
- Before: https://main--upp--adobecom.aem.page/de/homepage/index-loggedout
- After: https://main--upp--adobecom.aem.page/de/homepage/index-loggedout?milolibs=footer-fixes

### DIFF
--- a/libs/c2/blocks/region-nav/region-nav.css
+++ b/libs/c2/blocks/region-nav/region-nav.css
@@ -82,7 +82,7 @@
 }
 
 .dialog-modal:has(.region-nav) > .fragment {
-  max-height: calc((100dvh - 6px) - 2em);
+  max-height: calc((100svh - 6px) - 2em);
 }
 
 @media (min-width: 600px) {
--- a/libs/c2/blocks/region-nav/region-nav.css
+++ b/libs/c2/blocks/region-nav/region-nav.css
@@ -77,9 +77,12 @@
   color: var(--s2a-color-gray-1000);
 }
 
-
 .dialog-modal .region-nav a:hover {
   text-decoration: underline;
+}
+
+.dialog-modal:has(.region-nav) > .fragment {
+  max-height: calc((100dvh - 6px) - 2em);
 }
 
 @media (min-width: 600px) {


Deva309 (approved):

nishantka (approved):

sonawanesnehal3 (approved):

libs/c2/blocks/region-nav/region-nav.css
```diff
@@ -77,11 +77,14 @@
   color: var(--s2a-color-gray-1000);
 }
 
-
 .dialog-modal .region-nav a:hover {
   text-decoration: underline;
 }
 
+.dialog-modal:has(.region-nav) > .fragment {
+  max-height: calc((100dvh - 6px) - 2em);
```
rgclayton: Because `dvh` calculates layout changes in real-time, the browser re-evaluates dimensions as toolbars disappear. On content-heavy pages, this can cause small, visual layout shifts or minor performance hits during scrolling. If you want a layout that remains completely rigid, consider utilizing `svh` instead.

[Check the note](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/length#dynamic_viewport_units) on MDN under dvh:

> Note: While the dynamic viewport size can give you more control and flexibility, using viewport-percentage units based on the dynamic viewport size can cause the content to resize while a user is scrolling a page. This can lead to degradation of the user interface and cause a performance hit.

bandana147: Thanks for the review, suggested changes have been implemented.

SushmaNayak630:
Validation on the PR : https://github.com/adobecom/milo/pull/6019

Able to scroll  the Langnav completely to the bottom and can access the last language options, working fine  in all 4 locales

[https://www.stage.adobe.com/de/milolibs=footer-fixes#langnav](https://www.stage.adobe.com/de/#langnav)

https://github.com/user-attachments/assets/86c48b76-81e9-4762-97eb-499be4569975

narcis-radu:
@bandana147 - do we need this for wave 2 or Production? It looks like Production, so the PR should be against STAGE.

bandana147:
> @bandana147 - do we need this for wave 2 or Production? It looks like Production, so the PR should be against STAGE.

Ok, i will create a new PR from stage then.

bandana147:
Will create a new PR from stage.

PR #6027, opened 2026-06-03
MWPW-197413: Add RPS/worker limit for Nala execution

afmicka:
EDS started enforcing 200 rps limit which often causes huge amount of 429 errors in Nala executions on PRs on github. 
We introduce throttle per worker in order not to exceed the limit for now, capping to the MAX of total 180 rps.
- limitation is added only for MAS Nala tests
- leaving current setup with 7 workers on CI, 3 workers locally. The rps limited per worker is 180/number_of_workers
- introduces `nala-test.js` to inject `page.route()` into every scoped test. Playwright does not have global hook to do it from playwright.config.js or globalSetup

Resolves: [MWPW-197413](https://jira.corp.adobe.com/browse/MWPW-197413)

**Test URLs:**
- After: https://MWPW-197413--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/nala/libs/eds-throttle.js
+++ b/nala/libs/eds-throttle.js
@@ -17,7 +17,8 @@ export function resolveEdsMaxRps() {
   if (process.env.NALA_EDS_THROTTLE_DISABLED === '1') return 0;
   if (process.env.NALA_EDS_MAX_RPS !== undefined && process.env.NALA_EDS_MAX_RPS !== '') {
     const v = Number.parseInt(process.env.NALA_EDS_MAX_RPS, 10);
-    return Number.isFinite(v) && v > 0 ? v : 0;
+    if (Number.isFinite(v) && v > 0) return v;
+    console.warn(`[NALA] NALA_EDS_MAX_RPS="${process.env.NALA_EDS_MAX_RPS}" is not a positive integer — ignoring, falling back to worker-derived default.\n`);
   }
   const workers = Number.parseInt(process.env.NALA_WORKER_COUNT ?? '1', 10);
   const n = Number.isFinite(workers) && workers > 0 ? workers : 1;
@@ -75,15 +76,16 @@ export function logEdsThrottleOnce(edsMaxRps) {
 }
 
 /**
- * Register a route handler that paces EDS-bound requests on a Playwright page.
+ * Register a route handler that paces EDS-bound requests on a Playwright browser context.
+ * Covers all pages in the context, including those created with context.newPage().
  * Called automatically from the nala-test.js base fixture for every test.
- * @param {import('@playwright/test').Page} page
+ * @param {import('@playwright/test').BrowserContext} context
  */
-export async function installEdsThrottleOnPage(page) {
+export async function installEdsThrottleOnContext(context) {
   const edsMaxRps = resolveEdsMaxRps();
   if (edsMaxRps <= 0) return;
   logEdsThrottleOnce(edsMaxRps);
-  await page.route('**/*', async (route) => {
+  await context.route('**/*', async (route) => {
     if (isEdsEdgeHost(route.request().url())) {
       await throttleEdsGap(edsMaxRps);
     }
--- a/nala/libs/eds-throttle.js
+++ b/nala/libs/eds-throttle.js
@@ -14,28 +14,28 @@
 const EDS_SAFE_TOTAL_RPS = 180;
 
 export function resolveEdsMaxRps() {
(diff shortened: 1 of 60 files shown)


nala/libs/eds-throttle.js
```diff
@@ -0,0 +1,92 @@
...
+ */
+
+/** Combined RPS budget with headroom under the 200 rps/hostname tenant limit. */
+const EDS_SAFE_TOTAL_RPS = 180;
+
+export function resolveEdsMaxRps() {
+  if (process.env.NALA_EDS_THROTTLE_DISABLED === '1') return 0;
+  if (process.env.NALA_EDS_MAX_RPS !== undefined && process.env.NALA_EDS_MAX_RPS !== '') {
+    const v = Number.parseInt(process.env.NALA_EDS_MAX_RPS, 10);
+    return Number.isFinite(v) && v > 0 ? v : 0;
```
honstar: If someone sets `NALA_EDS_MAX_RPS=foo` (or `0`, or `-5`), `resolveEdsMaxRps()` returns `0`, which is the same path as the explicit disable flag — the throttle quietly turns off. A typo'd env var produces the exact failure mode this PR is meant to prevent. Suggest falling back to the worker-derived default and logging a warning, or throwing.

nala/libs/nala-test.js
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Usage — replace the @playwright/test import in test files:
+ *   import { test, expect } from '../../libs/nala-test.js';
+ */
+
+import { test as base, expect } from '@playwright/test';
+import { installEdsThrottleOnPage } from './eds-throttle.js';
+
+export const test = base.extend({
+  page: async ({ page }, use) => {
```
honstar: The fixture wraps the default `page`, so any extra pages/contexts created inside a test (`context.newPage()`, `browser.newContext()`, etc.) bypass the throttle. Not sure if this an actual use-case.

honstar (changes requested):

afmicka:
@honstar could you please take another look if the comments are addressed properly?

honstar (approved):

skumar09:
@afmicka : 
- **I really like the approach this PR takes** to address the EDS throttling issue and the **framework-level safeguard it provides.**
- **One thought:** before introducing request-level throttling, could we evaluate reducing the worker count for the `MAS CI runs (for example, from 7 to 2–3 workers) `and monitor the results for a few days? Since the issue is currently scoped to MAS tests, a worker-count adjustment in the Playwright execution/configuration would be a relatively small change compared to introducing **routing/throttling** logic.

- **Example**
```
{
  name: 'mas-chromium',
  workers: 2, // example
  testMatch: isCI ? masFeatures : undefined,
  use: {
    ...devices['Desktop Chrome'],
    userAgent: USER_AGENT_DESKTOP,
    channel: 'chrome',
  },
}

```
- If reducing concurrency does not sufficiently address the EDS rate-limit issue, then the throttling solution would definitely make sense as the next step. In any case, I like having these throttling utilities available in NALA libs for future use cases as well.

cc: @honstar , @narcis-radu

afmicka:
@skumar09 We can certainly try. It will slow down the execution but maybe not for much. 

I have made the change to only reduce the number of workers for MAS tests but keeping the implementation for easy enablement in the future. 
In order to apply throttle only the `@playwright/test` import needs to be replaced with `nala-test.js` in the test files, the rest is being handled.

For now, I am keeping the implementation available in NALA libs but not using it.

afmicka:
@skumar09 i see Nala failing still on stage and here with updated branch. These failures should not be related to the 429. Do you know what is the cause?

Axelcureno (approved):
LGTM — throttle approach is sound, and the per-worker budget holds given the MAS matrix runs isolated on its own runner.

One thing: `eds-throttle.js` ships pure exported helpers (`resolveEdsMaxRps`, `isEdsEdgeHost`, `throttleEdsGap`, `logEdsThrottleOnce`) with no unit tests — they're trivially testable (env parsing, host matching, gap math) and would lock in the budget logic. Worth adding before merge.

skumar09:
@afmicka  : Manually merging the PR into the stage branch, (as it will unblock other PRs where M@S tests are failing)
- The Nala tests currently failing on the Chromium browser are passing on both Firefox and WebKit.
- I also validated the failing tests locally on Chrome/Chromium, and they passed successfully.
      - The failures appear to be isolated to the GitHub Actions Chromium run rather than an issue with the changes in this PR.
    
cc: @narcis-radu @Axelcureno

PR #6034, opened 2026-06-03
[Site Redesign] - Use C2 modal only for C2 pages

Deva309:
This PR is to make sure we use C2 modal only for C2 pages.
We need to remove below selected rows from menifest as well:
Screenshot 2026-06-03 at 6 01 05 PM

**Test URL:**
https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-bizpro-hub?milolibs=c2-modal&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all#tour-marketing


narcis-radu (approved):
this looks good to me, but we'll need some validation that all existing modals are still working.

colloyd:
@narcis-radu @Deva309 I've checked the provided link and the brand concierge modal seems to work properly.

colloyd (approved):

libs/mep/ace1205/hub-hero/hub-hero.js
```diff
@@ -213,6 +213,9 @@ const buildSlide = ({ slide, index, slidesTotal }) => {
     'daa-ll': `${processTrackingLabels(heading?.textContent)}-${index + 1}--${processTrackingLabels(heading?.textContent)}`,
   }, content);
 
+  if (link?.dataset?.modalHash) slideEl.dataset.modalHash = link.dataset.modalHash;
+  if (link?.dataset?.modalPath) slideEl.dataset.modalPath = link.dataset.modalPath;
```
overmyheadandbody, line 217: These are identical, right? So one can be removed.

overmyheadandbody (approved):
Just a small comment

libs/mep/ace1205/hub-hero/hub-hero.js
Deva309: Both are required for the modal to function. `modal.js` reads them separately as below:
```
const { modalHash, modalPath } = el.dataset;
// modalHash → used to match against window.location.hash
if (window.location.hash !== modalHash ...) return null;

// modalPath → used to load the fragment content
const path = a?.dataset.modalPath || await localizeLinkAsync(...);
```
The lines look structurally similar but operate on different attributes for different purposes.

SilviuLCF (approved):
verified https://jira.corp.adobe.com/browse/MWPW-197513

PR #6044, opened 2026-06-04
[SR] - Rich content eyebrow enhance

DKos95:
**Test URLs:**
- Before: https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-offer?milolibs=site-redesign-foundation
- After: https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-offer?milolibs=rich-content-enhance-eyebrow

### DIFF
--- a/libs/mep/ace1205/rich-content/rich-content.css
+++ b/libs/mep/ace1205/rich-content/rich-content.css
@@ -28,15 +28,19 @@
       max-width: 680px;
     }
 
-    p:has(picture) {
-      display: flex;
-      align-items: center;
-      justify-content: center;
-      gap: var(--s2a-spacing-xs);
+    .eyebrow {
+      color: var(--s2a-color-content-knockout);
 
-      picture, img {
-        width: 24px;
-        height: 24px;
+      &:has(picture) {
+        display: flex;
+        align-items: center;
+        justify-content: center;
+        gap: var(--s2a-spacing-xs);
+  
+        picture, img {
+          width: 24px;
+          height: 24px;
+        }
       }
     }
   }
--- a/libs/mep/ace1205/rich-content/rich-content.css
+++ b/libs/mep/ace1205/rich-content/rich-content.css
@@ -28,7 +28,7 @@
       max-width: 680px;
     }
 
-    .eyebrow :has(picture) {
+    .eyebrow:has(picture) {
       color: var(--s2a-color-content-knockout);
       display: flex;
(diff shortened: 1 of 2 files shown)


libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -28,15 +28,19 @@
       max-width: 680px;
     }
 
-    p:has(picture) {
-      display: flex;
-      align-items: center;
-      justify-content: center;
-      gap: var(--s2a-spacing-xs);
+    .eyebrow {
+      color: var(--s2a-color-content-knockout);
```
overmyheadandbody, line 32: The eyebrow has a color defined, `--s2a-color-content-eyebrow`, where's the disconnect between that color and the one used here? Is it certain that all eyebrows should have this color, regardless of context? This currently resolves to white, which is unlikely to be the global behavior we need.

DKos95: Rich Content Jump - https://www.figma.com/design/q87sUm2fvForRQzxGum5wE/Offer-%E2%80%94-A.com?node-id=12214-24418&m=dev
Rich Content Garage Door with eyebrow - https://www.figma.com/design/q87sUm2fvForRQzxGum5wE/Offer-%E2%80%94-A.com?node-id=67-9329&m=dev

They both have white as the color, `--s2a-color-content-eyebrow` is not close to that.

DKos95: @overmyheadandbody after more digging I think the best option is to gate this behind `.eyebrow:has(picture)` since these ones have white as the color, thanks for the heads up!

overmyheadandbody: Great, thanks for looking into it! In these sorts of scenarios I do wonder whether not using the `eyebrow` class and going with CSS properties instead might be a better idea, since the picture is likely added programatically to the eyebrow, it's not a universally-supported use-case.

overmyheadandbody (approved):

libs/mep/ace1205/rich-content/rich-content.css
DKos95: Actually it's added via authoring.

zagi25 (approved):

PR #6060, opened 2026-06-05
[SR] MWPW-194269: Quick Action aspect - VAQ

nishantka:
Update aspect ratio for tiles in Quick Action block

**Test URLs:**
- After: https://quick-action-aspect--milo--adobecom.aem.page/?martech=off

QA:
https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-product?milolibs=quick-action-aspect&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all


libs/mep/ace1205/quick-actions/quick-actions.css
```diff
@@ -117,6 +118,17 @@ img.quick-actions-media {
...
 @media (width < 768px) {
+  .quick-actions {
+    padding-block-end: var(--s2a-spacing-2xl);
+  }
+
+  .quick-actions-tile {
+    aspect-ratio: auto;
+    /* mobile needs same aspect ratio till min height 180px, then only width should reduce
+       calc find 48:50 ratio — (grid width - 1 col gap) / 2 cols × 50/48 — with minimum height 180px */
+    block-size: max(180px, calc((100cqw - var(--s2a-spacing-xs)) / 2 * 50 / 48));
```
zagi25: We maybe don't need this calculation. I noticed that both image and footer are `position: absolute`, is there a reason for this ? I think things can be simplified if image is not position absolute and has `min-height: 180px` along with width and height 100%, and it seems that some of aspect ratios are not needed as grid handles everything. I might be wrong but maybe it is worth a try.

nishantka: @zagi25 Thanks! I see the given assets do have correct intrinsic ratio themselves. So, we can use it to render tiles. I have given min-width for tile as 0 as grid cell tries to maintain the content width which was cutting off from right.

robert-bogos (approved):

zagi25 (approved):

DKos95 (approved):

PR #6064, opened 2026-06-05
[MWPW-197698] Serve larger image renditions for big screens

narcis-radu:
The code will allow serving larger image renditions for big screens. The code is behind a feature flag - `larger-images`, which needs to be set to `on`.

Resolves: [MWPW-197698](https://jira.corp.adobe.com/browse/MWPW-197698)

**Test URLs:**
- After: https://img-compression--milo--adobecom.aem.page/?martech=off

Demo URL: https://img-compression--milo--adobecom.aem.page/drafts/narcis/image-size
* new source is added with `media=(min-width: 1920px)` and `src="img.type?width=3000&format=webply"`
* no change if screen size is < 1920px
* image with 3000 width should be loaded if screen size > 1920px


DKos95 (approved):

libs/utils/utils.js
```diff
@@ -1439,6 +1439,27 @@ export function decorateImageLinks(el) {
   });
 }
 
+export function decoratePictures(area) {
+  area.querySelectorAll('picture').forEach((picture) => {
+    if (picture.classList.contains('large-image-decorated')) return;
+    const sources = picture.querySelectorAll('source');
+    const image = picture.querySelector('img');
+    if (!sources.length || !image) return;
+    const imageWidth = image.width || image.naturalWidth;
```
robert-bogos:
```suggestion
    const imageWidth = image.naturalWidth || image.width;
```
if we're aiming for the original width of the image I think these should be reversed. `image.width` is giving the rendered width , but `image.naturalWidth` is the original width of the image

mokimo: Isn't this always 0 as we lazyload the images?

overmyheadandbody: We're interested in the attribute, which comes from EDS, not the actual rendered width

overmyheadandbody: Tagging along :D I think it would make sense to use `image.getAttribute("width")` (and then parse it as a number), since that's what gives us the resolution at which it was uploaded in EDS. We can make a fair assumption that anything over 2000px is meant to be crisp. Otherwise, as @mokimo suggested, we could run into lazy-loading issues

narcis-radu: what @overmyheadandbody said

narcis-radu: The width comes from EDS directly see, `view-source:https://img-compression--milo--adobecom.aem.page/drafts/narcis/image-size`.

overmyheadandbody (approved):

zagi25 (approved):

robert-bogos (approved):

mokimo (approved):

mokimo:
A nit that doesn't neccessarily need to be addressed in this PR, but we should have unit tests for any new code added in utils!

SilviuLCF (approved):
Verified, testing details https://jira.corp.adobe.com/browse/MWPW-197698

PR #6073, opened 2026-06-08
[MWPW-197855] Fix sidekick library scroll — can't reach bottom of block list

mokimo:
### Description
When an author opens the Library sidekick and scrolls through the available blocks, the last block(s) are cut off and unreachable.

**Root cause:** `.con-container` had `height: calc(100% - 36px)`, accounting only for the 36px title row. When the search bar is visible (blocks, templates, icons), the header grows to ~72px, so the container overflows the 360px parent by ~36px — clipped by `overflow: hidden` on `.sk-library`.

**Fix:** Make `.sk-library` a flex column and set `.con-container` to `flex: 1; min-height: 0`. The list area now naturally fills the remaining space regardless of header height. Also removed the `margin-bottom: 39px` workaround that was patching the symptom.

Resolves: [MWPW-197855](https://jira.corp.adobe.com/browse/MWPW-197855)

## Before
image

## After
image

### DIFF
--- a/libs/blocks/library-config/library-config.css
+++ b/libs/blocks/library-config/library-config.css
@@ -3,6 +3,8 @@ Common styles for the block library
 */
 
 .sk-library {
+  display: flex;
+  flex-direction: column;
   width: 360px;
   height: 360px;
   background: #303030;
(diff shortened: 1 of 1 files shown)


cod17828 (approved):

rgclayton (approved):

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-197855

robert-bogos (approved):

PR #6083, opened 2026-06-08
[SR] - Hover list vqa

nishantka:
Addressing VQA Feedback:
- Sticky Heading in Tablet
- Modified numbering + spacing + alignment
- Height of list items updated
- hover effect removed from desktop and tablet

**Test URLs:**
- After: https://hover-list-vqa--milo--adobecom.aem.page/?martech=off

QA:
Before:
https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-product?milolibs=site-redesign-foundation&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all

After:
https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-product?milolibs=hover-list-vqa&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all

### DIFF
--- a/libs/mep/ace1205/hover-list/hover-list.js
+++ b/libs/mep/ace1205/hover-list/hover-list.js
@@ -48,6 +48,21 @@ function hideMedia(media) {
   });
 }
 
+function setupStickyBoundary(headline, list) {
+  const tabletMQ = window.matchMedia('(width >= 768px) and (width < 1280px)');
+  const wrapper = headline.parentElement;
+
+  const update = () => {
+    if (!tabletMQ.matches) { wrapper.style.height = ''; return; }
+    const items = [...list.querySelectorAll('.hover-list-item')];
+    const first4Height = items.slice(0, 5).reduce((sum, el) => sum + el.offsetHeight, 0);
+    wrapper.style.height = `${first4Height + headline.offsetHeight}px`;
+  };
+
+  new ResizeObserver(update).observe(list);
+  tabletMQ.addEventListener('change', update);
+}
+
 function addCursorFollower(list) {
   const cursor = { x: 0, y: 0, vx: 0, hasPrev: false };
   let activeItem = null;
@@ -191,10 +206,13 @@ function decorate(block) {
     list.append(item);
   });
 
+  const headlineWrapper = createTag('div', { class: 'hover-list-headline-wrapper' });
+  headlineWrapper.append(headline);
   addCursorFollower(list);
   const listCol = createTag('div', { class: 'hover-list-col' });
   listCol.append(list);
-  block.replaceChildren(headline, listCol);
+  block.replaceChildren(headlineWrapper, listCol);
+  setupStickyBoundary(headline, list);
 }
 
 export default function init(el) {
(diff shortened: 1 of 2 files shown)


DKos95 (commented):
The first item seems to not have any text rendered, it happens also with the base branch but maybe this is a good time to fix this if it is an issue at all?

Screenshot 2026-06-08 at 15 46 22

DKos95:
> The first item seems to not have any text rendered, it happens also with the base branch but maybe this is a good time to fix this if it is an issue at all?
> 
Seems this is an authoring issue.

libs/mep/ace1205/hover-list/hover-list.js
```diff
@@ -48,6 +48,21 @@ function hideMedia(media) {
...
+function setupStickyBoundary(headline, list) {
+  const tabletMQ = window.matchMedia('(width >= 768px) and (width < 1280px)');
+  const wrapper = headline.parentElement;
+
+  const update = () => {
+    if (!tabletMQ.matches) { wrapper.style.height = ''; return; }
+    const items = [...list.querySelectorAll('.hover-list-item')];
+    const secondLastBoundary = items.slice(0, items.length - 2).reduce((sum, el) => sum + el.getBoundingClientRect().height, 0);
+    wrapper.style.height = `${secondLastBoundary + headline.getBoundingClientRect().height}px`;
+  };
```
DKos95, line 62: Would something like this work? 

Reasons to optimize:
- getBoundingClientRect() — triggers layout reads; you only need box height, not viewport position
- [...querySelectorAll(...)] — allocates a new array every resize
- .slice().reduce() — another temporary array + callback overhead

```suggestion
  const update = () => {
    if (!tabletMQ.matches) { wrapper.style.height = ''; return; }
    const items = list.children;
    let height = headline.offsetHeight;
    for (let i = 0, stop = items.length - 2; i < stop; i += 1) height += items[i].offsetHeight;
    wrapper.style.height = `${height}px`;
  };
```

nishantka:
> > The first item seems to not have any text rendered, it happens also with the base branch but maybe this is a good time to fix this if it is an issue at all?
> 
> Seems this is an authoring issue.

This is an issue with cloning content from different viewports. It doesn't clone the missing content from smaller viewport if it contains only text. That will be addressed in a separate PR.

libs/mep/ace1205/hover-list/hover-list.js
nishantka: I have updated the code to make it simpler but we need to use `el.getBoundingClientRect().height` to avoid 1-2px differences between the sizes that makes the upper border line misaligned and visually incorrect.

DKos95 (approved):

rgclayton (approved):

biljana-cvijanovic (approved):
Validated, ready to merge

PR #6103, opened 2026-06-09
MWPW-197933: Block merch free-trial links from decorating in gnav

zagi25:
Issue: Merch checkout CTA is not hidden despite code detecting that it should be hidden. Reason is, `shouldBlockFreeTrialLinks` recognizes that button should be hidden, recognizes that it is a checkout CTA and adds `hidden-osi-trial-link` class that should hide CTA with `display:none` but it returns `false` and this allows [global-navigation.js](https://github.com/adobecom/milo/blob/stage/libs/blocks/global-navigation/global-navigation.js#L139) to continue to decorate CTA and in the process to remove `hidden-osi-trial-link` which hides the CTA. 

* return `true` if checkout CTA should be hidden

Resolves: [MWPW-197933](https://jira.corp.adobe.com/browse/MWPW-197933)

**Test URLs:**
- Before: https://main--milo--adobecom.aem.page/kr/drafts/ratko/kr-free-trial-links
- After: https://mwpw-197933-hide-kr-ft--milo--adobecom.aem.page/kr/drafts/ratko/kr-free-trial-links

**EDU URLs:**
- Before: https://main--edu--adobecom.aem.page/kr/education/students/creativecloud
- After: https://main--edu--adobecom.aem.page/kr/education/students/creativecloud?milolibs=mwpw-197933-hide-kr-ft

### DIFF
--- a/libs/blocks/merch/merch.css
+++ b/libs/blocks/merch/merch.css
@@ -28,8 +28,8 @@ a[is='checkout-link'].con-button > span {
   min-width: 66px;
 }
 
-.hidden-osi-trial-link {
-  display: none;
+a[data-hide-kr-free-trial="true"] {
+  display: none !important;
 }
 
 .price-unit-type:not(.disabled)::before,
--- a/libs/blocks/merch/merch.js
+++ b/libs/blocks/merch/merch.js
@@ -1501,7 +1501,7 @@ export async function buildCta(el, params) {
     const prefix = getConfig()?.locale?.prefix;
     if (!(prefix === '/kr' && cta.value[0]?.offerType === OFFER_TYPE_TRIAL)) return;
     if (shouldAllowKrTrial(el, prefix)) {
-      cta.classList.remove('hidden-osi-trial-link');
+      cta.removeAttribute('data-hide-kr-free-trial');
       return;
     }
     cta.remove();
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -568,8 +568,8 @@ export const shouldBlockFreeTrialLinks = (link) => {
   }
 
   if (link.dataset.wcsOsi) {
-    link.classList.add('hidden-osi-trial-link');
-    return true;
+    link.dataset.hideKrFreeTrial = 'true';
+    return false;
   }
 
   const parent = link.parentElement;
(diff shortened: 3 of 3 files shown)


DKos95:
For the life of me I can't remember the reason why this was done like it is, there must have been a reason, but there is merch.js code that also removes this class that we added way back then https://github.com/adobecom/milo/blob/stage/libs/blocks/merch/merch.js#L1504

DKos95:
It's possible that this link wasn't populated at this moment and so we added a class to hide it then left it to merch.js to decide if it should be removed or not, this is my best guess.

SilviuLCF (approved):
QA verified https://jira.corp.adobe.com/browse/MWPW-197933

robert-bogos (approved):

mokimo (approved):

overmyheadandbody (approved):

DKos95 (changes requested):
This was added because we wanted to deffer the decision to merch.js since free trial can be written in a number of ways that we don't know, checking the .offerType in merch.js [here](https://github.com/adobecom/milo/blob/8018dda11641f7a39472d04f27227b31a0f63274/libs/blocks/merch/merch.js#L1502) is a 100% chance to remove a free trial type link for kr, the reason why we check it there is because it is available only there when the button is settled, here in utils.js the offerType is not available, if we were to do it like this then we would be depending on knowing what the exact text of the button is and that we cannot be sure about.

DKos95 (approved):
Approving since this solves issue from the ticket, but a follow up ticket should be created to make sure the merch.js code is executed since we can't be 100% sure of text content in buttons.

zagi25:
Current logic works, problem occurs because in global navigation (and some other places connected to nav/footer), CTA is [cloned](https://github.com/adobecom/milo/blob/stage/libs/blocks/global-navigation/global-navigation.js#L138), because of this, this [fallback](https://github.com/adobecom/milo/blob/stage/libs/blocks/merch/merch.js#L1507) in merch.js loses reference to CTA and it can't remove it if free-trial  in merch link has text that we don't expect (Dusan explained in the [comment](https://github.com/adobecom/milo/pull/6103#pullrequestreview-4465980693)).
I've updated code to use attribute instead of class, but regardless merch fallback is useless in this case because it doesn't have reference to the actual CTA.

narcis-radu (approved):

SilviuLCF (approved):
Verified https://jira.corp.adobe.com/browse/MWPW-197933

PR #6105, opened 2026-06-09
[SR] rich content block content widths update

rgclayton:
* Updating default body width per breakpoint with new widths
* Adding new variant to set max-width on `heading` `body` (`narrow` and `narrow-xs`)

**Test URLs:**
- After: https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-product?milolibs=sr-rc-content-widths&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all

New heading max width variant added to these `rich-content` block on Product page.
image

image

### DIFF
--- a/libs/c2/styles/styles.css
+++ b/libs/c2/styles/styles.css
@@ -624,7 +624,7 @@ a {
   }
 }
 
-h1, h2, h3, h4, h5, h6, [class*="heading-"]:not(.heading-max-width) {
+h1, h2, h3, h4, h5, h6, [class*="heading-"] {
   margin: 0;
   font-family: var(--heading-font-family);
   font-weight: var(--s2a-font-weight-heading);
--- a/libs/c2/styles/styles.css
+++ b/libs/c2/styles/styles.css
@@ -624,7 +624,7 @@ a {
   }
 }
 
-h1, h2, h3, h4, h5, h6, [class*="heading-"] {
+h1, h2, h3, h4, h5, h6, [class*="heading-"]:not(.heading-max-width) {
   margin: 0;
   font-family: var(--heading-font-family);
   font-weight: var(--s2a-font-weight-heading);
--- a/libs/mep/ace1205/rich-content/rich-content.css
+++ b/libs/mep/ace1205/rich-content/rich-content.css
@@ -221,8 +221,12 @@
     --rc-sweep-offset: 230px;
   }
 
-  .rich-content .content [class*="body-"]:not(.action-area) {
-    min-width: 560px;
+  .rich-content.heading-max-width .content [class*="heading-"]:not(.action-area) {
+    max-width: 680px;
+  }
+
+  .rich-content .foreground .content [class*="body-"]:not(.action-area) {
+    max-width: 415px;
     width: calc((100% / var(--grid-columns)) * 8);
   }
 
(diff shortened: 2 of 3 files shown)


DKos95 (approved):

libs/c2/styles/styles.css
```diff
@@ -624,7 +624,7 @@ a {
   }
 }
 
-h1, h2, h3, h4, h5, h6, [class*="heading-"] {
+h1, h2, h3, h4, h5, h6, [class*="heading-"]:not(.heading-max-width) {
```
overmyheadandbody: Why is the `:not` needed here? It increases specificity, but the declarations inside don't touch width. They do come with some valuable rules that would be disregarded when the variant is used, though.

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -221,8 +221,12 @@
...
-  .rich-content .content [class*="body-"]:not(.action-area) {
-    min-width: 560px;
+  .rich-content.heading-max-width .content [class*="heading-"]:not(.action-area) {
+    max-width: 680px;
+  }
+
+  .rich-content .foreground .content [class*="body-"]:not(.action-area) {
+    max-width: 415px;
     width: calc((100% / var(--grid-columns)) * 8);
   }
```
overmyheadandbody: This essentially limits all body text for all Rich Content blocks. This could have unintended consequences and should be a documented system decision before introducing - https://sr-rc-content-widths--milo--adobecom.aem.page/drafts/ramuntea/rich-content-widths. Same comment for https://github.com/adobecom/milo/pull/6105/changes#diff-a4c2a70eb76db2764bd78e2d192053573b2db476ad448df3e76be16f4ab55ebdR345-R347

libs/utils/decorate.js
```diff
@@ -260,12 +260,14 @@ function applyTextOverrides(el, override, targetEl) {
 }
 
 const textOverridesConfig = isC2 ? ['heading-', 'body-', 'button-'] : ['-heading', '-body', '-detail'];
+const headingSizePattern = /^heading-(1|2|3|4|5|6|font|max|super)$/;
```
overmyheadandbody: Where do `font` / `max` come from?

overmyheadandbody (changes requested):
This will introduce some undesired side-effects for body text and uses some patterns that are likely irrelevant. Could this change be applied surgically where it's needed for now? There is a desire to standardize headings to be limited to `ch` widths, but that should be done at a system level and with ample time to QA.

libs/c2/styles/styles.css
rgclayton: Because it gets picked up as a heading size and then removed from the block level. It then tries to apply `heading-max` as the size to the heading.

rgclayton:
> This will introduce some undesired side-effects for body text and uses some patterns that are likely irrelevant. Could this change be applied surgically where it's needed for now? There is a desire to standardize headings to be limited to `ch` widths, but that should be done at a system level and with ample time to QA.

@overmyheadandbody 
The ask was to apply it to all body sizes for rich-content. Details are in the Slack thread, you should be tagged for visibility.

We can ask about adding a new variant for this similar to `heading-max-width`

libs/utils/decorate.js
rgclayton: `max` is there to filter out. I can update to remove `super`.

libs/c2/styles/styles.css
rgclayton: Double checked on this. I can remove it. The problem i was running into is corrected with the update to filter in `decorate.js`.

overmyheadandbody: Why not rename the variant to be something other than `.heading-`?

overmyheadandbody (approved):
This is much cleaner. We'll still have to revisit this for rollout, but this is now scoped enough to not cause system issues. Thanks for the update!

NadiiaSokolova:
Verified. No issues found.

PR #6122, opened 2026-06-10
[SR] - Fix video not pausing when out of viewport

DKos95:
Fixes bug where videos are not paused when scrolled past, this has been an issue in c1 also.

Resolves: [MWPW-196476](https://jira.corp.adobe.com/browse/MWPW-196476)

**C1 Test URLs:**
- Before: https://main--milo--adobecom.aem.page/docs/library/kitchen-sink/media
- After: https://rich-content-video-performance--milo--adobecom.aem.page/docs/library/kitchen-sink/media

**C2 Test URLs:**
- After: https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-product?milolibs=rich-content-video-performance&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all

**Homepage Test URLs:**
- Before: https://load-c2-styles--upp--adobecom.aem.live/homepage/drafts/ramuntea/redesign-demo?milolibs=site-redesign-foundation
- After: https://load-c2-styles--upp--adobecom.aem.live/homepage/drafts/ramuntea/redesign-demo?milolibs=rich-content-video-performance

### DIFF
--- a/libs/utils/decorate.js
+++ b/libs/utils/decorate.js
@@ -450,23 +450,24 @@ function getVideoIntersectionObserver() {
   if (!window?.videoIntersectionObs) {
     window.videoIntersectionObs = new window.IntersectionObserver((entries) => {
       entries.forEach((entry) => {
-        const { intersectionRatio, target: video } = entry;
+        const { isIntersecting, intersectionRatio, target: video } = entry;
         const isHaveLoopAttr = video.getAttributeNames().includes('loop');
         const { playedOnce = false } = video.dataset;
         const isUserPaused = video.hasAttribute(USER_PAUSED_ATTR);
         const isPlaying = video.currentTime > 0 && !video.paused && !video.ended
           && video.readyState > video.HAVE_CURRENT_DATA;
+        const playInViewport = video.hasAttribute('data-play-viewport');
         const canPlay = !isUserPaused && (isHaveLoopAttr || !playedOnce) && !isPlaying;
 
-        if (intersectionRatio <= 0.8) {
+        if (!isIntersecting) {
           if (isPlaying && (!playedOnce && !isUserPaused)) syncPausePlayIcon(video);
           video.pause();
-        } else if (canPlay) {
+        } else if (canPlay && (!playInViewport || intersectionRatio > 0.8)) {
           video.play();
           syncPausePlayIcon(video, { type: 'playing' });
         }
       });
-    }, { threshold: [0.8] });
+    }, { threshold: [0, 0.8] });
   }
   return window.videoIntersectionObs;
 }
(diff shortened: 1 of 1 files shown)


libs/utils/decorate.js
```diff
@@ -441,40 +441,45 @@ export function handleObjectFit(bgRow) {
...
           && video.readyState > video.HAVE_CURRENT_DATA;
+        const playInViewport = video.hasAttribute('data-play-viewport');
+        const canPlay = !isUserPaused && (isHaveLoopAttr || !playedOnce) && !isPlaying;
 
-        if (intersectionRatio <= 0.8) {
+        if (!isIntersecting) {
           if (isPlaying && (!playedOnce && !isUserPaused)) syncPausePlayIcon(video);
           video.pause();
-        } else if (!isUserPaused && (isHaveLoopAttr || !playedOnce) && !isPlaying) {
+        } else if (canPlay && (!playInViewport || intersectionRatio > 0.8)) {
```
rgclayton, line 465: For plain autoplay videos, `playInViewport` is false, so `!playInViewport` is always true, making `intersectionRatio > 0.8 `irrelevant. The threshold: [0] addition now fires the callback the moment any pixel of the video enters the viewport — and at that point the play condition evaluates to just canPlay, meaning the video resumes immediately at near-zero visibility.


```suggestion
  } else if (canPlay && intersectionRatio > 0.8) {
```

libs/utils/decorate.js
```diff
@@ -441,40 +441,45 @@ export function handleObjectFit(bgRow) {
   });
 }
 
+function shouldUseViewportObserver(video) {
+  if (!video || video.hasAttribute('data-hoverplay') || video.controls) return false;
+  return video.hasAttribute('data-play-viewport') || video.hasAttribute('autoplay');
```
rgclayton, line 446: Is the intent for this to change the behavior for every autoplay video? Just asking because this could cause unexpected regressions anywhere autoplay is used without `data-play-viewport`.

DKos95: I don't see a good reason why we need authors to apply `data-play-viewport`, if the video is paused when not in the users viewport then the user isn't losing out on anything just gaining performance, and we are not leaning on authors to apply this and we are giving them one less thing to think about, not sure why this is an opt in initialy to be honest?

NadiiaSokolova (approved):
Verified.

libs/utils/decorate.js
rgclayton: I'm pretty sure opt in is coming from c1 world where default was click to play. So, autoplay was set as opt in. Maybe this needs revisited in c2?

rgclayton (approved):

PR #6125, opened 2026-06-10
[Site Redesign] - Added support of using merch/ost link within Sticky CTA block

Deva309:
### Added support of using merch/ost link within `Sticky CTA` block.

**Test URL:**
https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-offer?milolibs=checkout-link-issue

### DIFF
--- a/libs/mep/ace1205/floating-cta/floating-cta.js
+++ b/libs/mep/ace1205/floating-cta/floating-cta.js
@@ -2,6 +2,28 @@ import { createTag, getFederatedUrl } from '../../../utils/utils.js';
 import { getMetadata } from '../section-metadata/section-metadata.js';
 import icons from '../../../c2/assets/icons.js';
 
+function waitForCheckoutLink(linkPara, timeoutMs = 10000) {
+  const existing = linkPara.querySelector('a');
+  if (existing?.isCheckoutLink) return Promise.resolve(existing);
+
+  return new Promise((resolve, reject) => {
+    let timeoutId;
+    const observer = new MutationObserver(() => {
+      const link = linkPara.querySelector('a');
+      if (link?.isCheckoutLink) {
+        clearTimeout(timeoutId);
+        observer.disconnect();
+        resolve(link);
+      }
+    });
+    observer.observe(linkPara, { childList: true, subtree: true });
+    timeoutId = setTimeout(() => {
+      observer.disconnect();
+      reject(new Error('Timed out waiting for checkout link to upgrade'));
+    }, timeoutMs);
+  });
+}
+
 function applyCustomHide(el, ctaEl) {
   const sectionMeta = el.parentElement?.querySelector('.section-metadata');
   if (!sectionMeta) return;
@@ -15,11 +37,12 @@ function applyCustomHide(el, ctaEl) {
     ([entry]) => {
       const isHidden = entry.isIntersecting;
       ctaEl.classList.toggle('active', !isHidden);
-      ctaEl.setAttribute('aria-hidden', String(isHidden));
       if (isHidden) {
         ctaEl.setAttribute('tabindex', '-1');
+        ctaEl.setAttribute('aria-hidden', String(isHidden));
       } else {
         ctaEl.removeAttribute('tabindex');
+        ctaEl.removeAttribute('aria-hidden');
       }
     },
     {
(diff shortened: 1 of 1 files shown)


libs/mep/ace1205/floating-cta/floating-cta.js
```diff
@@ -18,38 +46,24 @@ export default function init(el) {
   }
 
   const linkEl = linkPara.querySelector('a');
+  if (linkEl?.isCheckoutLink) {
+    await linkEl.onceSettled();
+    const [ctaText] = linkEl.textContent.trim().split('|').map((s) => s.trim());
```
zagi25: Can checkout link also have ariaLabel as on line 62 ? 
 `const [ctaText, ariaLabel = ctaText] = sourceText.split('|').map((s) => s.trim());`

zagi25 (approved):

libs/mep/ace1205/floating-cta/floating-cta.js
Deva309: @zagi25 merch link already have aria-label. So just passing cta text only.

libs/mep/ace1205/floating-cta/floating-cta.js
```diff
@@ -53,3 +29,38 @@ export default function init(el) {
...
+
+  const relativeSrc = img.getAttribute('src');
+  if (relativeSrc?.startsWith('/')) {
+    img.src = getFederatedUrl(relativeSrc);
+  }
+
+  const linkEl = linkPara.querySelector('a');
+  const arrow = createTag('span', { class: 'icon-button', 'aria-hidden': 'true' }, icons.arrowRightWhite);
+  if (linkEl?.isCheckoutLink) {
+    await linkEl.onceSettled();
```
rgclayton: Does this need error handling?

If the promise rejects (network failure, invalid SKU, etc.), the async init will throw an unhandled rejection. The block will be left in a broken intermediate state (raw merch link visible, no promo-cta styling).

libs/mep/ace1205/floating-cta/floating-cta.js
```diff
@@ -53,3 +29,38 @@ export default function init(el) {
...
+  if (!img) return;
+
+  const relativeSrc = img.getAttribute('src');
+  if (relativeSrc?.startsWith('/')) {
+    img.src = getFederatedUrl(relativeSrc);
+  }
+
+  const linkEl = linkPara.querySelector('a');
+  const arrow = createTag('span', { class: 'icon-button', 'aria-hidden': 'true' }, icons.arrowRightWhite);
+  if (linkEl?.isCheckoutLink) {
```
rgclayton: Does this need a retry if undefined?

isCheckoutLink is a property set by the merch/OST initializer on the anchor element. If merch hasn't processed the link yet when init runs, this is undefined → falls through to the regular path, which creates a plain <a> from the link's raw text. The checkout link is then silently rendered incorrectly with no retry.

libs/mep/ace1205/floating-cta/floating-cta.js
Deva309: @rgclayton 
Addressed in the latest commit.

The merch path is now wrapped in a try/catch. On WCS rejection (or if the link never upgrades within 10s), we log to LANA with severity error and call el.remove() so the floating CTA disappears instead of rendering in a broken intermediate state.

libs/mep/ace1205/floating-cta/floating-cta.js
Deva309: @rgclayton 
Addressed in the latest commit.

1. Detection now handles both timing scenarios: `linkEl.classList.contains('merch')` OR `linkEl.isCheckoutLink`. The merch class is added synchronously by `decorateAutoBlock` inside `decorateLinksAsync`, which `decorateSection` awaits BEFORE any block init runs — so the class is guaranteed to be on the anchor regardless of whether `merch.init` has executed yet.

2. Added a `waitForCheckoutLink()` helper that uses a `MutationObserver` on the linkPara to re-acquire the upgraded checkout-link anchor after `merch.init` calls `el.replaceWith(merch)`. This eliminates the stale-reference issue since `merch.init` and `floating-cta.init` run in parallel via Promise.all.

If the upgrade doesn't happen within 10s, we log to LANA and call el.remove() instead of silently falling through to the plain-link path.

SilviuLCF (approved):
QA verified https://jira.corp.adobe.com/browse/MWPW-198219

overmyheadandbody (approved):

libs/mep/ace1205/floating-cta/floating-cta.js
```diff
@@ -2,48 +2,47 @@ import { createTag, getFederatedUrl } from '../../../utils/utils.js';
...
+  const existing = linkPara.querySelector('a');
+  if (existing?.isCheckoutLink) return Promise.resolve(existing);
 
-  const arrow = createTag('span', { class: 'icon-button', 'aria-hidden': 'true' }, icons.arrowRightWhite);
-  const cta = createTag('a', { href: ctaHref, class: 'promo-cta', 'aria-label': ariaLabel, tabindex: '-1' }, [img, ctaText, arrow]);
-  el.replaceChildren(cta);
+  return new Promise((resolve, reject) => {
+    let timeoutId;
+    const observer = new MutationObserver(() => {
+      const link = linkPara.querySelector('a');
```
zagi25, line 12: nit: Probably not needed because of `existing`

libs/mep/ace1205/floating-cta/floating-cta.js
```diff
@@ -2,48 +2,47 @@ import { createTag, getFederatedUrl } from '../../../utils/utils.js';
...
+      if (link?.isCheckoutLink) {
```
zagi25, line 13: nit: early return

zagi25 (approved):

PR #6137, opened 2026-06-11
[SR] Offer hero - a11y fixes

robert-bogos:
## Description
This PR is fixing the a11y issues raised in the following tickets:
- https://jira.corp.adobe.com/browse/MWPW-198058
- https://jira.corp.adobe.com/browse/MWPW-198057
- https://jira.corp.adobe.com/browse/MWPW-198056

## Test URL
- Before: https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-offer?milolibs=site-redesign-foundation&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all

- After: https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-offer?milolibs=sr-offer-hero&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all


libs/mep/ace1205/offer-hero/offer-hero.js
```diff
@@ -364,6 +366,16 @@ function initAnimation(block) {
   }, { rootMargin: '200px 0px' });
   io.observe(block);
 
+  let scrollOrigin;
+  window.addEventListener('scroll', () => {
+    if (!document.activeElement?.matches('a.hero-card-tile')) return;
+    scrollOrigin ??= window.scrollY;
+    if (scrollOrigin - window.scrollY > FOCUS_REMOVE_THRESHOLD) {
+      document.activeElement.blur();
```
overmyheadandbody, line 374: Does this reset the focused element on the page? Say I've tabbed through 12 elements and then this logic runs, would tabbing take the user back to the first element on the page? I think losing the focus progress could be an issue if that's the case

SilviuLCF (approved):
QA verified 
Testing details: 
https://jira.corp.adobe.com/browse/MWPW-198058
https://jira.corp.adobe.com/browse/MWPW-198057
https://jira.corp.adobe.com/browse/MWPW-198056

libs/mep/ace1205/offer-hero/offer-hero.js
```diff
@@ -364,6 +366,16 @@ function initAnimation(block) {
   }, { rootMargin: '200px 0px' });
   io.observe(block);
 
+  let scrollOrigin;
+  window.addEventListener('scroll', () => {
```
overmyheadandbody, line 370: The operation is cheap enough, but using `requestAnimationFrame` or some debounce might reduce a tiny bit of load on the CPU.

libs/mep/ace1205/offer-hero/offer-hero.css
```diff
@@ -97,12 +97,17 @@
     display: flex;
   }
 
+  .hero-card:focus-within {
+    outline: 2px solid var(--s2a-color-blue-800);
```
overmyheadandbody, line 101: This will differ from the browser's default, which is what we're using usually. Can the default be kept and the override avoided?

overmyheadandbody (approved):

DKos95 (approved):
Approving, but do agree with Rares regarding `document.activeElement.blur();`, it may cause another a11y issue, should be checked with Taryn but I am fairly curtain we shouldn't change tabbing order.

zagi25 (approved):

libs/mep/ace1205/offer-hero/offer-hero.js
robert-bogos: I've also thought about this, but the extra complexity does not justify the very subtle performance gain

libs/mep/ace1205/offer-hero/offer-hero.js
robert-bogos: No, focus progress in not lost 👍

robert-bogos:
> Approving, but do agree with Rares regarding document.activeElement.blur();, it may cause another a11y issue, should be checked with Taryn but I am fairly curtain we shouldn't change tabbing order.

Yes, I will check with Taryn about that. Anyway, is better now then before

libs/mep/ace1205/offer-hero/offer-hero.css
robert-bogos: I've looked a bit into it and the override can be avoided. Opened a new PR for this since this one was merged
https://github.com/adobecom/milo/pull/6138

PR #6156, opened 2026-06-12
MWPW-193615 makes usage of pzn & mask in autoblocks

npeltier:
Resolves: consuming part of [MWPW-193615](https://jira.corp.adobe.com/browse/MWPW-193615) where mas autoblocks can now have pzn & mask attributes that will be forwarded to mas/io (depends on https://github.com/adobecom/mas/pull/958 to be working, but will not break anything if put before).

a later enhancement for fancier authoring integration should be planned. For the moment, attributes will have to be added by hand.

Test URLs:
- Before: https://main--milo--adobecom.aem.live/drafts/nala/features/commerce/plans?martech=off
- After: https://mwpw-193615--milo--adobecom.aem.live/drafts/nala/features/commerce/plans?martech=off

https://mwpw-193615--milo--adobecom.aem.page/drafts/npeltier/photoshop?maslibs=mwpw-193615
https://mwpw-193615--milo--adobecom.aem.live/drafts/npeltier/photoshop?maslibs=mwpw-193615&mas-io-url=https://14257-merchatscale-npeltier.adobeioruntime.net/api/v1/web/MerchAtScale&mas.preview=off

### DIFF
--- a/libs/blocks/mas-compare-chart-autoblock/mas-compare-chart-autoblock.js
+++ b/libs/blocks/mas-compare-chart-autoblock/mas-compare-chart-autoblock.js
@@ -2,6 +2,7 @@ import { createTag, getConfig, loadStyle } from '../../utils/utils.js';
 import {
   getOptions,
   initService,
+  createAemFragment,
   loadMasComponent,
   MAS_MERCH_CARD,
   overrideOptions,
@@ -56,11 +57,7 @@ function loadCompareChartStyles() {
 }
 
 export async function createCompareChart(el, options) {
-  const attrs = { fragment: options.fragment };
-  if (seenFragments.has(options.fragment)) attrs.loading = 'cache';
-  seenFragments.add(options.fragment);
-
-  const aemFragment = createTag('aem-fragment', attrs);
+  const aemFragment = createAemFragment(options, seenFragments);
   const compareChart = createTag(MAS_COMPARE_CHART, { consonant: '' }, aemFragment);
   const paragraph = el.parentElement;
   paragraph.parentElement.className = 'mas-compare-chart-container';
--- a/libs/blocks/merch-card-collection-autoblock/merch-card-collection-autoblock.js
+++ b/libs/blocks/merch-card-collection-autoblock/merch-card-collection-autoblock.js
@@ -4,6 +4,7 @@ import { postProcessAutoblock, handleCustomAnalyticsEvent } from '../merch/autob
 import { mepMasStudioUrls } from '../merch/mas-mep-utils.js';
 import {
   initService,
+  createAemFragment,
   getOptions,
   MEP_SELECTOR,
   overrideOptions,
@@ -309,7 +310,7 @@ function paintStPriceRed(collection, locale) {
 }
 
 export async function createCollection(el, options) {
-  const aemFragment = createTag('aem-fragment', { fragment: options.fragment });
+  const aemFragment = createAemFragment(options);
   // Get MEP overrides if available
   const { mep, locale } = getConfig();
   const mepFragments = mep?.inBlock?.[MEP_SELECTOR]?.fragments || {};
--- a/libs/blocks/merch/merch.js
+++ b/libs/blocks/merch/merch.js
@@ -1551,16 +1551,45 @@ export function overrideOptions(fragment, options) {
   return options;
 }
 
+export function createAemFragment(options, seenFragments = null) {
+  const { fragment, pzn, mask } = options;
+  const cacheKey = `${fragment}|${pzn || ''}|${mask || ''}`;
+  const attrs = Object.fromEntries(
+    Object.entries({ fragment, pzn, mask }).filter(([, v]) => v != null),
+  );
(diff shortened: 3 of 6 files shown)


Axelcureno (commented):
(replaced by inline comments below)

libs/blocks/merch/merch.js
```diff
@@ -1551,16 +1551,45 @@ export function overrideOptions(fragment, options) {
...
+
+      case 'fragment':
+      case 'query':
+        options.fragment = value;
+        break;
+
+      case 'mask':
+      case 'pzn':
+      case 'field':
+        options[key] = value;
```
Axelcureno, line 1584: `mask`/`pzn` arrive percent-decoded from the link hash, and mas#958 interpolates them unencoded into the IO endpoint (`endpoint += '&mask=' + …`), so an authored `mask=foo%26locale%3Dde_DE` injects extra query params into the /fragment request. Best fixed with `encodeURIComponent` in #958 while it's still open; an allowlist here wouldn't hurt as defense-in-depth.

libs/blocks/merch-card-collection-autoblock/merch-card-collection-autoblock.js
```diff
@@ -309,7 +310,7 @@ function paintStPriceRed(collection, locale) {
 }
 
 export async function createCollection(el, options) {
-  const aemFragment = createTag('aem-fragment', { fragment: options.fragment });
+  const aemFragment = createAemFragment(options);
```
Axelcureno, line 313: `mask` on the collection's `aem-fragment` stays a silent no-op even after mas#958 (io overlays card-model roots only) — intentional forward-compat, or worth omitting here?

libs/blocks/merch/merch.js
```diff
@@ -1551,16 +1551,45 @@ export function overrideOptions(fragment, options) {
   return options;
 }
 
+export function createAemFragment(options, seenFragments = null) {
+  const { fragment, pzn, mask } = options;
+  const cacheKey = `${fragment}|${pzn || ''}|${mask || ''}`;
+  const attrs = Object.fromEntries(
+    Object.entries({ fragment, pzn, mask }).filter(([, v]) => v != null),
```
Axelcureno: nit: `fragment` is stamped first, so mas#958's `getFetchInfo(this.cacheKey())` runs before `pzn`/`mask` are set and fetch telemetry gets keyed under the bare id — ordering `{ pzn, mask, fragment }` here fixes it.

libs/blocks/merch/merch.js
```diff
@@ -1551,16 +1551,45 @@ export function overrideOptions(fragment, options) {
   return options;
 }
 
+export function createAemFragment(options, seenFragments = null) {
```
Axelcureno, line 1554: nit: `createAemFragment` and the new `getOptions` arms are only covered via autoblock init — a couple of direct unit tests in merch.test.js would be cheap.

Axelcureno (commented):
Checked this against mas#958 — attribute names line up exactly, and on current MAS main `pzn`/`mask` are inert, so merging this first is safe as described. The `createAemFragment` refactor preserves prior behavior at all four call sites, including the shared `seenFragments` set. Details inline.

libs/blocks/merch-card-collection-autoblock/merch-card-collection-autoblock.js
npeltier: yep, for now no mask for collections nor comp chart, but it is likely to happen

libs/blocks/mas-compare-chart-autoblock/mas-compare-chart-autoblock.js
```diff
@@ -56,11 +57,7 @@ function loadCompareChartStyles() {
 }
 
 export async function createCompareChart(el, options) {
-  const attrs = { fragment: options.fragment };
-  if (seenFragments.has(options.fragment)) attrs.loading = 'cache';
```
yesil, line 60: @npeltier removing cache param intentionally ?

npeltier: no no, it's managed centrally

yesil (approved):

Axelcureno (approved):
Re-reviewed — all four points from last round are handled. Cache key now includes `pzn`/`mask` (with tests proving same-fragment-different-overlay stays uncached), and the attr ordering puts the overlays before `fragment` so telemetry keys correctly. The injection concern I raised is closed on the mas#958 side: `mask`/`pzn` are now validated against `^[a-zA-Z0-9_-]+$` and `mask` goes into a path segment, not the query string. `createAemFragment` cleanly dedupes the four call sites. LGTM 👍

Roycethan (commented):
Feedbacks are given here:
https://github.com/adobecom/mas/pull/958#pullrequestreview-4502015015
Apart from that non-regressions covered and keeping this open for addressing the feedbacks.

Roycethan (approved):

PR #6162, opened 2026-06-15
[MWPW-198599] Preflight: make IMS token optional for log capture

mokimo:
### Description
The preflight `captureMetrics` function currently bails out when no IMS access token is present, meaning unauthenticated users never send preflight logs.

This makes the token optional:
- Gate only on `imsClientId` being present (still required for the `clientId` query param)
- Skip `getProfile()` and leave `email: ''` when no token is available
- Omit the `Authorization` header from the fetch when unauthenticated

Resolves: [MWPW-198599](https://jira.corp.adobe.com/browse/MWPW-198599)


zagi25 (approved):

skholkhojaev (approved):

SilviuLCF (approved):
QA verified https://jira.corp.adobe.com/browse/MWPW-198599

PR #6178, opened 2026-06-16
Warn on duplicate fragment references within a page

shkhan91:
A fragment URL can be referenced more than once on the same page (typically via accidental copy-paste during authoring). There is currently no feedback mechanism to alert the author or developer that a duplicate exists. Loading behavior must remain unchanged; the goal is purely observability — a console warning that surfaces the duplicate so it can be corrected.

### DIFF
--- a/libs/blocks/fragment/fragment.js
+++ b/libs/blocks/fragment/fragment.js
@@ -5,6 +5,36 @@ import {
 
 const fragMap = {};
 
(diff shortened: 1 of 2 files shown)


mokimo:
Should we instead add this in the preflight? https://github.com/adobecom/milo/tree/main/libs/blocks/preflight

That is our tool to surface (probable) mistakes to an author i think a "warn" type check within the general section would make sense
image

PR #6181, opened 2026-06-16
MWPW-198810[MEP] Fix highlight badge text and click region for promo manifest

markpadbe:
## Description
                                                                                                                                                                                                                                                                                                                        
  Fixes two MEP highlight-badge bugs that surface on promo-manifest                                                                                                                                                                                                                                                     
  replacements:                                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                        
  1. **Gnav promo badge was blank (only the 🔗 icon showed).** Inline                                                                                                                                                                                                                                                   
     fragment loads set `data-manifest-id` and `data-path` on the loaded
     children but never wrote `data-manifest-display`, so the CSS rendered                                                                                                                                                                                                                                              
     the icon with empty text. Now the inline branch of                                                                                                                                                                                                                                                               
     `updateFragDataProps` writes the display string alongside the id.                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                      
  2. **Merch-card badge read "manifest.json: html" instead of the                                                                                                                                                                                                                                                       
     replacement fragment path.** The merch-card propagation in                                                                                                                                                                                                                                                       
     `addHighlightData` copied `data-manifest-id` from the parent                                                                                                                                                                                                                                                       
     `.fragment` but not its `data-path`, so the fallback defaulted to                                                                                                                                                                                                                                                  
     "html". Now the parent's path is carried onto the merch-card.
                                                                                                                                                                                                                                                                                                                        
  Also makes the badge click handler position-aware: it now reads the                                                                                                                                                                                                                                                   
  `::before` badge's computed `top`/`left` and resolves them against the                                                                                                                                                                                                                                                
  host's true containing block (walking up when the host is                                                                                                                                                                                                                                                             
  `position: static`, e.g., gnav promos which `preview.css` excludes from                                                                                                                                                                                                                                             
  `position: relative`). Without this, clicks on those badges never                                                                                                                                                                                                                                                     
  registered because the hit-test assumed the badge sat at the host's                                                                                                                                                                                                                                                   
  top-left.                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                        
  All changes are gated to MEP preview mode.   

 Use the Before and After URLs below for each step.
                                                                                                                                                                                                                                                                                                                        
  1. Open the gnav dropdown and observe the MEP badge on the promo card.                                                                                                                                                                                                                                              
  2. Scroll down to the merch-card tabs and observe the MEP badge on the
     right card.                                                                                                                                                                                                                                                                                                        
   
  **Expected (After URL):** both badges show the manifest name and the                                                                                                                                                                                                                                                  
  fragment path, and are clickable (open the replacement fragment in a                                                                                                                                                                                                                                                
  new tab).                                                                                                                                                                                                                                                                                                             
   
  **Actual (Before URL):** gnav shows an empty badge with only the 🔗                                                                                                                                                                                                                                                   
  icon; the merch-card shows a non-clickable badge with the promo                                                                                                                                                                                                                                                     
  manifest name followed by "html" instead of the fragment path.
                                                                  
Resolves: [MWPW-198810](https://jira.corp.adobe.com/browse/MWPW-198810)

**Test URLs:**
- Before: https://main--da-cc--adobecom.aem.page/products/photoshop?mepHighlight=true
- After: https://main--da-cc--adobecom.aem.page/products/photoshop?mepHighlight=true&milolibs=mep-fix-promo-badges
- PSI-check: https://mep-fix-promo-badges--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/libs/features/personalization/personalization.js
+++ b/libs/features/personalization/personalization.js
@@ -198,7 +198,11 @@ const COMMANDS_KEYS = {
 };
 
 function addIds(el, manifestId, targetManifestId) {
-  if (manifestId) el.dataset.manifestId = manifestId;
+  if (manifestId) {
+    el.dataset.manifestId = manifestId;
+    const { path } = el.dataset;
+    el.dataset.manifestDisplay = path ? `${manifestId}: ${path}` : `${manifestId}: html`;
+  }
   if (targetManifestId) el.dataset.adobeTargetTestid = targetManifestId;
 }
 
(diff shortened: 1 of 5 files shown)


jpratt2 (approved):
Approved.
I see the full path to the fragment and it links to the fragment. 

Image

AdobeLinhart (approved):

PR #6182, opened 2026-06-17
Passing convertStageLinks to gnav code

bandana147:
* Added support for stage link conversion

Resolves: [MWPW-198846](https://jira.corp.adobe.com/browse/MWPW-198846)

**Test URLs:**
- Before: https://www.stage.adobe.com/?
- After: https://www.stage.adobe.com/?fedsbranch=stage-links&milolibs=gnav-stage-links

### DIFF
--- a/libs/c2/blocks/global-navigation/global-navigation.js
+++ b/libs/c2/blocks/global-navigation/global-navigation.js
@@ -64,7 +64,9 @@ export default async function init(el) {
       commands: [...commands, ...gnavMepCommands],
       handleCommands: personalizationHandler,
     },
-    convertStageLinks,
+    convertStageLinks: ({ anchors, hostname, href }) => {
+      convertStageLinks({ anchors, config, hostname, href });
+    },
   }).catch((error) => {
     window.lana?.log?.('Failed to initialize federal global navigation', {
       error,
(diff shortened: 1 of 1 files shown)


sonawanesnehal3 (approved):

sharmrj (commented):
since the second param is just the miloconfig, do you think we ought to bind it before we pass it along?

```javascript

const convertStageLinks_ = (anchors, hostname, href) => convertStageLinks(anchors, getConfig(), hostname, href);
```

That way we don't need a function with an arcane `unknown` property in https://github.com/adobecom/federal/pull/143/changes#diff-08c9dc9c052a1e94c8f768d446c5415dc093e911ab0c705c252804866b30f583R54

Deva309 (approved):

bandana147:
> since the second param is just the miloconfig, do you think we ought to bind it before we pass it along?
> 
> ```js
> const convertStageLinks_ = (anchors, hostname, href) => convertStageLinks(anchors, getConfig(), hostname, href);
> ```
> 
> That way we don't need a function with an arcane `unknown` property in https://github.com/adobecom/federal/pull/143/changes#diff-08c9dc9c052a1e94c8f768d446c5415dc093e911ab0c705c252804866b30f583R54

Done

sharmrj (approved):

spadmasa:
Validation on the pr
https://www.stage.adobe.com/jp/?fedsbranch=stage-links&milolibs=gnav-stage-links
image
image
image

PR #6200, opened 2026-06-18
MWPW-199051 [MEP] Save results of Rainfocus API to config.mep

vgoodric:
- Updated the event initialization function to store fetched event details in the configuration object. This change enhances the management of event data and ensures that registration status can be accessed from the config.
- Added a corresponding test to verify the new behavior.

Resolves: [MWPW-199051](https://jira.corp.adobe.com/browse/MWPW-199051)

**Test URLs:**
- After: https://mepeventssavedata--milo--adobecom.aem.page/?martech=off

 **Update — event-code + Milo-page registration**

  * Fix the `registered` audience to match on `event?.isRegistered` rather than the presence of a resolved event 
  * Switch the addon and `loadMepAddons` from `event-id` to `event-code` — the addon keys
    on the event *property* code (e.g. max2025), not the per-event UUID. 
  * Fix registration on Milo pages: read enablement from lowercase DA metadata
    (`signedin`/`userid`) as well as camelCase, fail open to IMS `getProfile()` when no
    `userId` meta is present, and guard `loadIms()`/`getProfile()` so a missing IMS session
    can no longer blank the page. 
  
  **Update — persistent registration cache**

  * Add a persistent localStorage cache (registered 24h / not-registered 3 min, mirroring
    FEDS `cacheTime`), keyed by event-code + userId. Persists only
    `{ isRegistered, inPersonAttendee }` — never the RF authToken/userKey.
  * Normalize Rainfocus's empty `{}` response to `{ isRegistered: false }`.
  * Redirect-cookie cache invalidation is committed but gated off behind
    `REDIRECT_INVALIDATION_ENABLED`, pending confirmation of the cookie name/domain (see
    ticket comment). Cache self-heals via the 3-min TTL until then.

**Update — redirect-cookie invalidation enabled (FEDS-style)**

  * Enabled the redirect-cookie invalidation (removed the `REDIRECT_INVALIDATION_ENABLED` gate). When
  `feds_<eventCode>_registeredByRedirect` is present, the user is treated as registered for the pre-LCP paint (no blocking RF call), the
  flag is cleared, and the result is written through to the cache.
  * Confirmed from the VEAL `post-reg-redirector-generic` source that the cookie is set `value=true`, `domain=.adobe.com`, `path=/`, 24h,
  not HttpOnly — so it's readable, and the flag is cleared with a matching `domain=.adobe.com`. No VEAL-side changes needed.
  * The cookie name is event-specific; a mismatch degrades to the normal cache/RF path (never a wrong result).


danielnoliva (approved):

libs/features/personalization/personalization.js
```diff
@@ -1676,6 +1676,7 @@ export async function init(enablements = {}) {
   } else {
     for (const [key, promise] of Object.entries(promises)) promises[key] = await promise;
     config.mep = {
+      ...(config.mep || {}),
       updateFragDataProps,
```
markpadbe, line 1688: The blanket `...(config.mep || {})` spread is what's failing CI — it preserves `blocks`, `fragments`,           
  `commands`, `entitlementMap`, etc. across `init()` calls, but the `consolidateObjects` and `consolidateArray` calls at lines 1446–1449 assume a fresh `config.mep` each run.                                                                          
                                                                                                                
  ```suggestion
      const previousEventDetails = config.mep?.eventDetails;
      config.mep = {
        ...(previousEventDetails && { eventDetails: previousEventDetails }),                                      
        updateFragDataProps,
  ```                                                                                                             
                                                                                                                
  Verified locally: `actions.test.js`, `mepGeolocation.test.js`, and the new `event.test.js` assertion all pass (28/28) with this change.

markpadbe (changes requested):

libs/features/personalization/personalization.js
```diff
@@ -1084,7 +1084,7 @@ async function getPersonalizationVariant(
     if (userEntitlements?.includes(name)) return true;
     const { lob, event } = config.mep.promises;
     if (lob && lob === name.split('lob-')[1]?.toLowerCase()) return true;
-    if (name === 'registered' && event) return true;
+    if (name === 'registered' && event.isRegistered) return true;
```
markpadbe: To avoid a `TypeError` if a manifest references the `registered` tag on a page without the `event-id` enabled: 
```suggestion
    if (name === 'registered' && event?.isRegistered) return true;
```

markpadbe (approved):

PR #6201, opened 2026-06-18
MEP Next v1

AdobeLinhart:
This is a complete rewrite of the MEP button. 

The primary objective is to clean up the code base, decouple functions and maintain readability for future edits while implementing new features. All overlay code should live in the mep-overlay folder with shared functions eventually being moved into mep-next.js as needed.

**GENERAL**
 - Each card should expand/collapse.
 - Expansion should retain states per session.
 - All options should update based on corresponding URL params upon load in.
 
**AUTHENTICATION**
Logged in
- Should display the Actions tab content.

Logged out
- Should display a Login to AEM Sidekick card for the actions tab content

**MEP TAB**
- Toggles the drawer.

**DRAWER**
Mep Logo
- Should link to MMM.
	
**Close Icon**
- Toggles the drawer.
	
Tabs
- Should switch between Actions and Summary.

**ACTIONS**
MEP Manifest List
- Should list manifests running on page.
- Should list details about the manifest when expanded.
- Should have a select with the experiments which updates the Preview button on change.
	
Highlight Toggles
- Should have toggles for MEP, CaaS, M@S and Other Fragments.
- Each toggle lists the amount of "updates" on the page with the element count tracked by a mutation observer.
- Toggles should highlight the corresponding amount of changes on the page and update the Preview button on change.
	
Toggles
- Preview Linke MepButton=off should update the Preview button and turn mep off on preview.
- If avaiable, show toggle for 7 day MMM manifest data.
- 7 day MMM manifest data should append available manifests to the list when toggled on following the normal manifest requirements.
	
Spoof Geo
- Should always display Top Markets.
- Should disable MEP Lingo and Lingo M@S if unavaiable.
- Should have a select with available Geo options per radio button.
- Should update Geo spoof in real time upon change.
- Upon reload, should default to the first available select that matches the akamaiLocale parameter, otherwise Top Markets "none".

Load Manifest
- Accepts a manifest URL provided by MEP sidekick.
- Should update the preview button.
	
Preview button
- Should load in with default active manifests for the href.
- Should update with a param for any changed input options.
- Should preview options selected upon click.
	
**SUMMARY**
Should have Page, Consent, Lingo, M@S and CaaS cards with the following data.

Page 
- Manifests Found
- Foundation
- Theme
- Target Integration
- Personalization

Consent
- Level 2 | Performance
- Level 4 | Advertising

Lingo
- Mep Lingo Updates
- Lang First | Lingo
- Geo Folder
- Country Cookie
- User Country
- Geo + User

Resolves: [MWPW-195019](https://jira.corp.adobe.com/browse/MWPW-195019)

**Test URLs:**
(Any Milo page, compare old MEP button to new MEP button.)

- Before: https://main--da-cc--adobecom.aem.page/creativecloud
- After: https://main--da-cc--adobecom.aem.page/creativecloud?milolibs=mep-next-v1&mepnext=on

Psi: https://mep-next-v1--milo--adobecom.aem.page/?martech=off


libs/features/mep/mep-next/mep-overlay/mep-overlay-highlight.js
```diff
@@ -0,0 +1,300 @@
...
+    el.dataset[prop] = manifestName;
+    if (prop === 'manifestId') el.dataset.manifestDisplay = `${manifestName}: html`;
+  });
+}
+
+function setHighlightData(manifests) {
+  const { mep } = getConfig();
+  if (!mep?.experiments) return;
+
+  manifests.forEach(({ selectedVariant, manifest }) => {
```
markpadbe: manifests is undefined here and mep.experiments is always at least [] (initialized in personalization.js), so the guard never short-circuits and this throws. That aborts init() before setDefaultFragments() runs — which is why the default fragment badges do not currently render.  

```suggestion
function setHighlightData() {
  const { mep } = getConfig();
  if (!mep?.experiments) return;

  mep.experiments.forEach(({ selectedVariant, manifest }) => {
```

libs/features/mep/mep-next/mep-overlay/mep-overlay-highlight.js
```diff
@@ -0,0 +1,300 @@
...
+  if (visibleChildren.length === 0) return e.clientX >= 0 && e.clientX < badgeWidth;
+  return visibleChildren.some((child) => {
+    const { top, left } = child.getBoundingClientRect();
+    const relX = e.clientX - left;
+    const relY = e.clientY - top;
+    return relY >= (topOffset - badgeHeight) && relY < topOffset && relX >= 0 && relX < badgeWidth;
+  });
+}
+
+function setBadgeHandlers() {
```
markpadbe: setBadgeHandlers here duplicates setBadgeEventListeners above (same selector, same ::before visibility check, same window.open on click) -- both end up attached to document.body. Today it's masked by the setHighlightData crash aborting init() before this line runs, so only one handler is registered. Once that's fixed, both will fire on every badge click and open the fragment in two tabs(tested locally to confirm). Want to pick one and drop the other before this lands.

libs/features/mep/mep-next/mep-overlay/mep-overlay.js
```diff
@@ -0,0 +1,561 @@
...
+      return;
+    }
+
+    const cards = buildActionsContent(svgData, pageId);
+    contentEl.replaceChildren(...cards);
+    drawerEl.appendChild(buildFooter());
+    await Promise.all(cards.map((c) => c.ready).filter(Boolean));
+    setDefaultValues();
+    setPreviewButton();
+  }, { envs: ['prod', 'stage'] });
```
markpadbe: The following were marked TEMP in the original mep-next-v1-auth-updates commits: the envs: ['prod',
  'stage'] expansion (your line 418), the mock-auth override block, and the console.log here. Each had a "remove before merge" comment explaining it was test scaffolding. Those comments got dropped during integration but the code stayed. Can we tighten envs back to ['prod'] and drop the mock-auth block + console.log before this is merged?

libs/features/mep/mep-next/mep-overlay/mep-overlay-highlight.js
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    const { top, left } = fragment.getBoundingClientRect();
+    const clickX = e.clientX - left;
+    const clickY = e.clientY - top;
+    if (isInBadgeArea(clickX, clickY, topOffset, badgeWidth, badgeHeight)) handleBadgeClick();
+  });
+}
+
+export function getPageUpdates(label) {
```
markpadbe: The body + subtree pattern itself isn't new — caasObserver and masObserver in mep-caas.js/mep-mas.js (carried over from preview.js) already use it, but they have an unwatch* teardown and an enable-guard inside the callback. What's new here is that getPageUpdates creates a fresh observer per call and gets invoked 4× from CARD_DATA.actions (MEP, Caas, M@S, Other Fragments) -- 4 perpetual document-wide observers with no teardown and no enable guard. Can they reuse the existing watchers, or scope to the drawer count nodes and disconnect when the drawer closes?

libs/features/mep/mep-next/mep-overlay/mep-overlay.js
```diff
@@ -0,0 +1,561 @@
...
+      const selectEl = document.querySelector('select.mep-spoof-geo');
+      if (!selectEl) return;
+
+      await populateGeoSelect(selectEl, 'spoof-geo-lingo-mas');
+      selectEl.value = mepAkamaiLocale;
+    }
+  };
+
+  let debounceTimer;
+  const observer = new MutationObserver((mutations) => {
```
markpadbe: This is a second M@S observer doing what watchForMasContent (in mep-mas.js) already does -- same body + subtree, same M@S targets. Can the summary refresh hook into the existing watcher instead of starting a parallel one?

libs/features/mep/mep-next/mep-overlay/mep-overlay-logic.js
```diff
@@ -0,0 +1,534 @@
+import { mepMasSubCollections } from '../mep-mas-subcollection.js';
+import { HIGHLIGHT_KEYS } from './mep-overlay-highlight.js';
+import { getMarketConfig, marketsLangForLocale } from '../../../../utils/market.js';
+import {
+} from '../mep-caas.js';
```
markpadbe: nit empty import.

libs/features/mep/mep-next/mep-next.js
```diff
@@ -1780,7 +1226,7 @@ export async function saveToMmm() {
 }
 export default async function decoratePreviewMode() {
```
markpadbe: After utils.js switched the preview entry to mep-overlay-highlight.js + mep-overlay.js, decoratePreviewMode here is no longer reachable in production -- and most of the helpers only it calls go with it:
                                                                                                                       
  Dead (delete):                                                                                                       
  - createPreviewPill                                                                                                
  - addHighlightData (duplicate of setHighlightData in mep-overlay-highlight.js)                                       
  - markDefaultFragments (duplicate of setDefaultFragments)                     
  - addFragmentBadgeClickHandlers (duplicate of the badge-click handler in mep-overlay-highlight.js)                 
                                                                                                                       
  Not replicated in the new overlay -- needs porting before delete:                                                     
  - adjustBadgesForZeroHeightSections and adjustStackedBadges. These write --badge-top-offset, which the new overlay   
  reads (mep-overlay-highlight.js:221, mep-overlay-highlight.css:196) but never writes. They handle badges in zero-height sections and stacked display: contents badges -- both currently broken on the new path (false-green tests mask this since they exercise decoratePreviewMode).                                                                  
                                                     
  Duplicated and still in use -- maybe consolidate to one home:                                                             
  - API_URLS, parseMepConfig, parsePageAndUrl (also defined in mep-overlay-logic.js; reached by mmm.js through this    
  file, so the surviving copy needs to live somewhere both paths can import from)                                      
                                                                                                                       
  Live, unchanged: getMepPopup, saveToMmm, escapeHtml, mep-next.css (for mmm.js/personalization.js).                   
                                                                                                                       
  Can decoratePreviewMode + the dead helpers be deleted, the two adjust-badge functions ported to the new overlay init, and the duplicates consolidated?

test/features/mep/mep-next/mep-next.test.js
markpadbe, line 1: This file's 118 tests are all run against decoratePreviewMode from mep-next.js, which is no longer the production preview path — utils.js now loads mep-overlay-highlight.js + mep-overlay.js. The live mep-overlay-* path has zero coverage, so green CI here doesn't tell us whether the new overlay works.
Can the tests be retargeted at the live entry points before this merges?

markpadbe (changes requested):

markpadbe:
A couple CI items:                                                                                            
                                                                                                                       
  - ESLint: import init from './image-video-link.js'; at the top of libs/utils/utils.js is unused and shadows local init in 7+ places. Looks like an accidental IDE auto-import -- one-line delete.
  - Nala: failing across MAS, MEP, and MILO. The MEP button suite (mep-button.{spec,page,test}.js) will probably need a full rewrite (let's discuss with @jpratt2) -- every selector targets the old preview-pill/popup, replaced by the FAB/drawer, and there's new surface (auth gate, 4-way toggles, spoof-geo, card state) with no coverage. mep-actions/mep-order don't touch the UI so hopefully unaffected. Some unrelated MILO failures (e.g., brand-concierge) should re-triage cleanly after a rebase onto current stage.

libs/utils/utils.js
```diff
@@ -1,3 +1,5 @@
+import init from './image-video-link.js';
```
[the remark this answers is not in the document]
markpadbe: This is the apparent auto-import I mentioned in the top-level CI comment. Every no-shadow violation in the ESLint check references "line 1 column 8", so deleting this line should clear the remaining ESLint failures in one shot.

libs/features/mep/mep-next/mep-overlay/mep-overlay-highlight.js
AdobeLinhart: Was due to a last minute refactor. Should be fixed.

libs/utils/utils.js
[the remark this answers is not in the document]
AdobeLinhart: Removed import init from './image-video-link.js';

libs/features/mep/mep-next/mep-overlay/mep-overlay.js
AdobeLinhart: It's only temporary and will be removed once no longer aware.

libs/features/mep/mep-next/mep-overlay/mep-overlay-highlight.js
AdobeLinhart: Unintentional. Should be fixed.

libs/features/mep/mep-next/mep-overlay/mep-overlay-logic.js
AdobeLinhart: Updated.

libs/features/mep/mep-next/mep-next.js
AdobeLinhart: Viv said she wants a completely unaltered mep-next.js file for the time being. It's hopefully reverted to the last good state since your rename. We'll need to fix it otherwise.

test/features/mep/mep-next/mep-next.test.js
AdobeLinhart: Claude has generated completely new unit tests that have 100% file coverage as per a discussion with Viv. We may need to sync on what to do with the old files.

libs/features/mep/mep-next/mep-overlay/mep-overlay-highlight.js
AdobeLinhart: Definitely don't need four observers. Hopefully that is fixed. The page may still be updating when the drawer is opened, which likely rules out open/close. We also don't know how long the updates will take. Not sure if there's a solution to have it disconnect if it doesn't know when the updates are finished.

libs/features/mep/mep-next/mep-overlay/mep-overlay.js
AdobeLinhart: Ran this through Claude. There was some minor cleanup for my observers. Otherwise regarding redundancies and separation of concerns: 

image

image

libs/features/mep/mep-next/mep-overlay/mep-overlay.js
markpadbe: Right, it was always supposed to be only temporary. I was just highlighting that you removed the flag without removing the risk. but  this comment is the flag now. so, we just need to make sure the temp code is removed before this is merged.

libs/features/mep/mep-next/mep-overlay/mep-overlay.js
markpadbe: Agreed on what Claude surfaced — the three callbacks are complementary, and the toggle-gated observers serve a different purpose than the always-on one. Both land.             
   
  Two narrower things still worth tightening:                                                                                                                                     
                                                                                                                                                                                
  - The always-on setMasObserver runs all three refreshes on every body mutation regardless of drawer state. When the drawer is closed nothing's visible, so the work is wasted. A
   drawerEl?.matches(':popover-open') short-circuit at the top of the callback addresses that without breaking late-hydration capture — same in-callback gate pattern that masObserver / caasObserver already use.                                                                                                                                         
  - new MutationObserver(...).observe(...) has no variable binding, so it can never be .disconnect()'d. Binding it costs nothing and unblocks future teardown.

libs/features/mep/mep-next/mep-overlay/mep-overlay-highlight.js
markpadbe: The 4→1 collapse addresses the core multiplication concern.                                                                                                        

One followup: ran the tests locally and mep-overlay-highlight.test.js:223 now fails (expected '0 Page Updates' to equal '1 Page Updates') because the refresh trigger moved to setMasObserver in mep-overlay.js, which this test file doesn't initialize. The companion test at :246 passes-by-accident for the same reason — it asserts the initial value is unchanged, which holds any time the observer doesn't fire. Easiest fix is probably to call the now-exported refreshPageUpdateCounts directly in both tests.

libs/features/mep/mep-next/mep-next.js
markpadbe: I'm not really sure what you mean here — can you (or Viv) spell out what counts as "unaltered" specifically?

libs/features/mep/mep-next/mep-overlay/mep-overlay-highlight.js
AdobeLinhart: Looks like a few of the tests broke. Should be good now. Although perhaps unit tests should be the last piece after a full QA in case code is reworked. It'd be less time consuming to fix them after everything is locked in.

libs/features/mep/mep-next/mep-next.js
AdobeLinhart: I phrased that poorly.

The discussion with viv regarded editing decoratePreviewMode to remove dead code (createPreviewPill, addFragmentBadgeClickHandlers, etc), for example. Her instructions were to copy the function, apply the updates and leave the original unchanged. Perhaps there was a miscommunication.

Technically there is a heap of dead code in mep-next.js. My understanding is that'll be cleaned up at a later date with MMM. As for consolidation and duplicates, agree.

It's likely adjustBadgesForZeroHeightSections and adjustStackedBadges were lost in one of the iterations. They should now be consolidated as adjustBadgePositions and tied them into the badge mutation observers rather than utilizing a timeout. We'll need to verify that they're still working as intended.

Will get back to you on the duplicate consolidation.

AdobeLinhart: Badge stacking update (old/new):

image

image

jpratt2 (approved):
This passes my QA, see https://jira.corp.adobe.com/browse/MWPW-201871

jpratt2:
I did a quick review and the latest changes are good to go
https://main--da-cc--adobecom.aem.page/creativecloud?milolibs=mep-next-v1&mepnext=on

markpadbe (approved):

PR #6207, opened 2026-06-22
perf: add 5s timeout + dedup to geo2 fetch, remove redundant fallback…

mokimo:
… call

geo2.adobe.com/json/ was fetched with no timeout (browser TCP timeout ~30s) and no deduplication — concurrent callers each started an independent fetch. resolveDetectedMarketCountry() compounded this by making a second serial geo2 attempt on failure, meaning a blocked geo2 could cost 60s total.

Changes:
- geo.js: wrap fetch in AbortController with 5s timeout; cache the in-flight promise so concurrent calls share one fetch instead of each issuing a duplicate request
- utils.js: remove the redundant getAkamaiCode() fallback block from resolveDetectedMarketCountry() — getCountry() already tried geo2 and returned null, an immediate retry will have the same result

lana.js loading late is a symptom, not a cause: it lazy-loads on the first window.lana.log() call, which fires when geo2 times out.

* Add your
* Specific
* Features or fixes

**Test URLs:**
- After: https://feature-perf-geo-timeout--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/libs/utils/geo.js
+++ b/libs/utils/geo.js
@@ -1,24 +1,43 @@
 /* eslint-disable no-console */
 
+const GEO_TIMEOUT_MS = 5000;
+
+// Shared in-flight promise so concurrent callers don't issue duplicate fetches.
+let inflight = null;
+
 /**
  * Gets the Akamai country code from the geo2.adobe.com service.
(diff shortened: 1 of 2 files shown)


mokimo:
Closed in favor of https://github.com/adobecom/milo/pull/6210

PR #6209, opened 2026-06-22
perf: defer and deprioritize Lingo Wave 2 cross-site index fetches

mokimo:
### Description

On Lingo-active pages (e.g. \`/fr/creativecloud\`), \`lingo-site-mapping.json\` (0.9 KB) resolves in ~70 ms and immediately triggers 6 cross-site query-index fetches (Wave 2) while Wave 1's 45 KB + 26 KB primary/base indexes are still mid-download. On a 4G connection this bandwidth race delays Wave 1 completion by ~220 ms, pushing back the LCP image fetch start by the same amount.

**Three changes in `utils.js`:**

**1. `fetchOptions` param on `processQueryIndexMap`**
Adds an optional `fetchOptions = {}` argument passed straight to `fetch()`, so call-sites can independently control priority.

**2. `priority: 'low'` on lingo-site-mapping fetch**
The site-mapping file has no LCP dependency — it only triggers cross-site index discovery. Deprioritising it frees bandwidth for LCP-critical hero CSS/JS during the most sensitive render window.

**3. Await Wave 1 before firing Wave 2**
Inside the `lingoSiteMapping` IIFE, added:
```js
await Promise.all(
  [queryIndexes[siteId]?.pathsRequest, baseQueryIndex?.pathsRequest].filter(Boolean),
);
```
before the Wave 2 `forEach`. Wave 2 now starts only after the 45 KB + 26 KB Wave 1 files have landed, eliminating the bandwidth race entirely on slow connections. Wave 2 calls also pass `{ priority: 'low' }`.

Wave 1 is unchanged — fires immediately at default priority, required for LCP-section link localization.

**Before (4G):** Wave 2 fires at ~t+70 ms (site-mapping resolves), competing with Wave 1 still downloading its 71 KB.
**After (4G):** Wave 2 fires at ~t+295 ms (after Wave 1 lands), no overlap.

Resolves: [MWPW-TBD](https://jira.corp.adobe.com/browse/MWPW-TBD)

**Test URLs (da-cc fr/creativecloud):**
- Before: https://main--da-cc--adobecom.aem.live/fr/creativecloud
- After: https://main--da-cc--adobecom.aem.live/fr/creativecloud?milolibs=mokimo-lingo-wave2-defer

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6208, opened 2026-06-22
[MWPW-TBD] Perf: parallelize personalization manifests and start lingo indexes earlier

mokimo:
### Description

Quick-win performance improvements for the personalization + lingo pipeline.

**1. `getManifestConfig` serial loop → `Promise.all`** (`personalization.js:1429`)
Each manifest JSON fetch was blocking the next. Now all fetches fire in parallel.
Estimated gain: 300–800 ms on pages with 3+ manifests.

**2. `categorizeActions` serial loop → `Promise.all`** (`personalization.js:1440`)
`categorizeActions` is effectively synchronous but was wrapped in a serial `for-await` loop. Now runs via `Promise.all` — eliminates microtask waterfall.

**3. `getEntitlementMap` in-flight dedup** (`personalization.js:1017`)
When `getManifestConfig` calls run concurrently via `Promise.all`, each calls `getPersonalizationVariant` → `getEntitlementMap`. Without dedup, N concurrent callers each issued a separate XLG tags fetch. Fixed by caching the in-flight promise on `config.mep.entitlementMapFetch`.

**4. `loadLingoIndexes` before `decorateDocumentExtras`** (`utils.js:2757`)
Sections (`.section` class) are now set up before `decorateDocumentExtras()` is awaited, so query-index fetches start while header/meta decoration runs.
Estimated gain: overlaps ~100–300 ms of `decorateDocumentExtras` latency.

**5. Query index URL dedup** (`utils.js:747`)
`loadQueryIndexes` was calling `processQueryIndexMap` twice with the same URL on prod/live when the locale prefix resolves to the same path for both the primary and base index. Now shares one in-flight fetch object.

**6. `market.js` modulepreload before `preview.js` import** (`personalization.js:1722`)
`preview.js` has a static `import` of `market.js`, creating a serial fetch waterfall. A `modulepreload` hint for `market.js` added right before the dynamic `import('./preview.js')` call lets both load in parallel.

Resolves: [MWPW-TBD](https://jira.corp.adobe.com/browse/MWPW-TBD)

**Test URLs (da-cc fr/creativecloud):**
- Before: https://main--da-cc--adobecom.aem.live/fr/creativecloud
- After: https://main--da-cc--adobecom.aem.live/fr/creativecloud?milolibs=mokimo-perf-quick-wins

**milo aem.page:**
- Before: https://main--milo--adobecom.aem.page/
- After: https://mokimo-perf-quick-wins--milo--adobecom.aem.page/

[...]

PR #6206, opened 2026-06-22
[MWPW-TBD] Optimize personalization module loading — eliminate serial fetch waterfall

mokimo:
### Description

On pages with personalization enabled (e.g. `adobe.com/fr/creativecloud.html`), the personalization module chain loaded fully serially:

| Step | Module | Start | Delay |
|---|---|---|---|
| 1 | `personalization.js` | t=499ms | — |
| 2 | `sanitizeHtml.js` (static import, discovered at parse) | t=540ms | +41ms serial |
| 3 | `promo-utils.js` (dynamic import fired in `init()`) | t=584ms | +44ms serial |

**~130ms of avoidable serial latency** before personalization could apply changes to the DOM.

#### Fix 1 — `modulepreload` for personalization deps (`utils.js`)

Added `modulepreload` hints for `personalization.js` and (when promo is enabled) `promo-utils.js` in `checkForPageMods()`, placed **before** the `getCountry()` await. Both fetches now start during country detection rather than after it, and are in the browser module cache by the time `await import()` fires.

#### Fix 2 — Inline `sanitizeHtml.js` into `personalization.js`

`sanitizeHtml.js` had exactly one consumer (`personalization.js`) and was always required via a static import — making it a guaranteed serial fetch. Moving the logic directly into `personalization.js` eliminates the file and the extra request entirely. `sanitizeHtmlBody` is exported from `personalization.js` for testability; the standalone file is deleted.

**Result:** all three modules now start fetching in parallel.

```
Before: personalization.js(40ms) → sanitizeHtml.js(43ms) → promo-utils.js(48ms) ≈ 130ms serial
After:  all three start together (during getCountry await) ≈ 50ms parallel
```

To observe: DevTools → Network → filter JS → waterfall for `personalization.js` / `promo-utils.js` on any personalized page.

---

**Test URLs (milo branch vs main on the observed page):**
- Before: https://main--da-cc--adobecom.aem.live/?milolibs=https://main--milo--adobecom.aem.page/libs
- After: https://main--da-cc--adobecom.aem.live/?milolibs=https://mokimo-perf-personalization-preload--milo--adobecom.aem.page/libs

**Test URLs (milo standalone preview):**
- Before: https://main--milo--adobecom.aem.page/
- After: https://mokimo-perf-personalization-preload--milo--adobecom.aem.page/

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Before
Personalization -> sanitize-html -> promo-utils
image

After
[...]

PR #6210, opened 2026-06-22
perf: improve lingo performance — personalization, geo, lingo, XLG

mokimo:
### Before
Multiple sequential render blocking calls
image

### After
Parallelize and prefetch needed code
image

### Description

Combined quick-win performance improvements targeting LCP on Lingo-active pages (e.g. `/fr/creativecloud`). Rolls up and supersedes the four individual PRs listed below — they can be closed once this merges.

**Supersedes:**
- #6206 — Personalization module serial waterfall
- #6207 — geo2 timeout + dedup
- #6208 — Parallel personalization manifests, earlier lingo indexes
- #6209 — Lingo Wave 2 defer + priority hints

---

#### 1. Personalization module `modulepreload` + inline `sanitizeHtml` (#6206)

`personalization.js → sanitizeHtml.js → promo-utils.js` loaded serially (~130ms). Added `modulepreload` hints before the `getCountry()` await so all three start in parallel. Inlined `sanitizeHtml.js` (single-consumer static import) to eliminate that extra request entirely.

**Gain:** ~80ms off personalization apply time.

---

#### 2. `geo2` 5s timeout + in-flight dedup (#6207)

`geo2.adobe.com` had no timeout (browser TCP timeout ~30s) and no dedup — concurrent callers each issued a separate fetch. `resolveDetectedMarketCountry()` compounded this with a serial retry on failure. Added `AbortController` with 5s timeout, cached the in-flight promise, and removed the redundant fallback call.

**Gain:** eliminates pathological 30–60s stall; no impact on typical fast-path.

---

#### 3. Parallel personalization manifest fetches + `getEntitlementMap` dedup (#6208)

`getManifestConfig` iterated manifests serially; `categorizeActions` used a serial `for-await` loop. Both converted to `Promise.all`. When manifests run concurrently, each was calling `getEntitlementMap` independently — fixed with an in-flight promise cache on `config.mep.entitlementMapFetch`. Also moved `loadLingoIndexes` before `decorateDocumentExtras` to overlap lingo index fetches with header decoration, and deduped identical primary/base query-index URLs.

**Gain:** 300–800ms on pages with 3+ manifests.

---

#### 4. Lingo Wave 2 defer + `priority: 'low'` (#6209)

`lingo-site-mapping.json` (0.9 KB) resolves in ~70ms and immediately fired 6 cross-site query-index fetches (Wave 2) while Wave 1's 45 KB + 26 KB primary/base files were still downloading. On 4G this bandwidth race delayed Wave 1 completion by ~220ms.

Three changes: `fetchOptions` param on `processQueryIndexMap`; `priority: 'low'` on the site-mapping fetch; await Wave 1 (`Promise.all` of primary + base) before firing Wave 2's `forEach`. Wave 2 calls also pass `priority: 'low'`.

Wave 1 unchanged — fires immediately at default priority for LCP-section link localization.

**Gain:** ~220ms bandwidth relief on 4G; Wave 2 no longer races Wave 1.

---

#### 5. XLG preload URL alignment (new)

`getXLGListURL` returns `https://www.adobe.com/federal/assets/data/mep-xlg-tags.json`. The `rel=preload` link used that raw URL; `fetchData` passes it through `normalizePath` which resolves it via `getFederatedUrl` to the environment-specific federated root (e.g. `main--federal--adobecom.aem.live` on preview). Two different resolved URLs → browser cache miss → two round trips. Wrapping the preload argument in `normalizePath()` aligns both calls to the same URL in every environment.

**Gain:** eliminates one redundant XLG JSON fetch on non-prod environments.

---

Resolves: [MWPW-TBD](https://jira.corp.adobe.com/browse/MWPW-TBD)

**Test URLs (da-cc fr/creativecloud):**
- Before: https://main--da-cc--adobecom.aem.live/fr/creativecloud
- After: https://main--da-cc--adobecom.aem.live/fr/creativecloud?milolibs=mokimo-perf-combined


libs/utils/sanitizeHtml.js
markpadbe, line 1: Hey @mokimo, nice perf wins on #6210!                                              
                                                                                     
  One thing to flag before you merge: deleting `libs/utils/sanitizeHtml.js`          
  breaks `adobecom/cc` (and `cc-graybox`), which dynamically imports it in           
  `universal-promo-terms.js:148`:                                                    
                                                                                     
  ```js                                                                              
  const { sanitizeHtml } = await import(`${miloLibs}/utils/sanitizeHtml.js`);        
   ```               
  Two problems: the path 404s, and only sanitizeHtmlBody survived -- sanitizeHtml     
  (firstChild variant) is gone, and they have different return shapes. The block
  ships on /offers/promo-terms, so this would break that page in prod the moment     
  milo deploys.                                                                      
   
  Potential updates: keep sanitizeHtml.js in place as a thin shim with both        
  exports (your perf win is unaffected since personalization no longer imports it),
  or land a coordinated PR to cc/cc-graybox first to inline their own helper, then   
  delete from milo. Either works -- wanted to make sure it was on your radar before
  merge.

libs/features/personalization/personalization.js
```diff
@@ -1426,22 +1467,17 @@ export async function applyPers({ manifests }) {
...
   let results = [];
 
-  for (const experiment of experiments) {
-    const result = await categorizeActions(experiment, config);
-    if (result) results.push(result);
-  }
-  results = results.filter(Boolean);
+  results = (await Promise.all(
+    experiments.map((exp) => categorizeActions(exp, config)),
+  )).filter(Boolean);
```
markpadbe, line 1446: Not blocking, just flagging: this Promise.all preserves last-write-wins for
  config.mep.replacepage / updateframework only because categorizeActions has a  
  synchronous body despite the async keyword. The PR turns that into an implicit
  invariant -- a future await inside categorizeActions would quietly break execution
  order.                                                                             
   
  Might be worth a code comment noting the constraint, or dropping async so a future await becomes a clear signal. Your call.

markpadbe (changes requested):

libs/features/personalization/personalization.js
```diff
@@ -1722,7 +1758,15 @@ export async function init(enablements = {}) {
...
     if (manifests?.length) await applyPers({ manifests });
-    if (config.mep?.preview) await import('./preview.js').then(({ saveToMmm }) => saveToMmm());
+    if (config.mep?.preview) {
+      loadLink(`${config.base}/utils/market.js`, { rel: 'modulepreload', crossorigin: 'anonymous' });
+      // Flatten the preview.js → caas/utils.js → {lingo-active, getUuid} discovery chain
+      loadLink(`${config.base}/utils/lingo-active.js`, { rel: 'modulepreload', crossorigin: 'anonymous' });
+      loadLink(`${config.base}/utils/getUuid.js`, { rel: 'modulepreload', crossorigin: 'anonymous' });
+      import('./preview.js').then(({ saveToMmm }) => saveToMmm()).catch((e) => {
+        log(`MEP save error: ${e.toString()}`);
+      });
```
markpadbe, line 1735: Observability gap: the new .catch only logs to console, where the old await let errors hit the outer try/catch and window.lana?.log. So saveToMmm failures in editor sessions would go silent in monitoring.
                                                                                                        
potential one-line fix:                                                                        
                                                                                                        
 ``` 
.catch((e) => {                                                                                       
    log(`MEP save error: ${e.toString()}`);
    window.lana?.log(`MEP save error: ${e.toString()}`);
  });   
```

libs/utils/sanitizeHtml.js
mokimo: ah.. would be cleaner to delete this on the consumer, but from a perf perspective, it should be fine to just preload the asset, so it happens in parallel to the other things - restored it while keeping the perf benefit

libs/features/personalization/personalization.js
mokimo: Added a comment hope thats fine for now, great to have you given it an in depth look and sharing the context!

libs/features/personalization/personalization.js
mokimo: Added!

vhargrave (approved):
very nice! :)

sukamat (approved):

markpadbe (approved):

PR #6225, opened 2026-06-24
VULN-35559 Fix XSS via URI fragment in Marketo configurator

jsandland:
## Summary
- Sanitize all string values loaded from the URL hash in `getHashConfig()` before they enter state or localStorage
- Uses the browser's HTML parser (`el.innerHTML = value; return el.textContent`) to strip tags and event handlers
- Prevents `form.cta.override` and `form.success.content` XSS payloads from reaching `window.mcz_marketoForm_pref`, which Marketo's `forms2.min.js` injects as raw HTML

Resolves: [VULN-35559](https://jira.corp.adobe.com/browse/VULN-35559)

## Attack vector (fixed)
A crafted URL fragment containing JSON with form.cta.override: `">` caused Marketo's external JS to inject the payload as raw HTML, executing arbitrary JS in the milo.adobe.com origin. The attack used window.name to carry an unbounded exfil payload across the navigation, exfiltrating the victim's Adobe IMS access token (bearer token valid for API calls against Adobe IMS), full name, email, account details, and all cookies from adobe.com. A valid access token allows the attacker to impersonate the victim against Adobe APIs for the duration of the token's validity, which is effectively full account takeover for any signed-in Adobe user who visits the crafted URL.

## How to validate

1. Set up an httpworkbench.com instance and configure the httpworkbench hostname / delivery page response body as described in VULN-35559, setting the collector URL to your instance
2. Sign in to your Adobe account in the test browser

Prod (unfixed):

3.  Visit the httpworkbench hostname / delivery page -- it will redirect to the PoC URL from VULN-35559
4. Check your httpworkbench instance -- you will see a POST containing your cookies, sessionStorage, and Adobe IMS access token

Feature Branch (fixed):
5. Update the delivery page POC URL in workbench to point to https://xss-attack--milo--adobecom.aem.page/tools/marketo?preview=1 keeping the hash from the PoC URL in VULN-35559 unchanged
6. Visit the httpworkbench hostname / delivery page again
7. Check your httpworkbench instance -- no POST is received, only the initial GET from the delivery page load

## Test plan
- [x] Unit tests added for `sanitizeConfigValue` and `sanitizeHashConfig` covering the reported payload, svg/onload variant, plain text preservation, and non-string passthrough
- [x] Verified on `https://xss-attack--milo--adobecom.aem.page/tools/marketo` with full PoC hash -- exfil POST no longer fires
- [x] Verified prod still vulnerable without fix (confirms test is meaningful)
- [x] All existing marketo-config tests pass

### DIFF
--- a/libs/blocks/marketo-config/context.js
+++ b/libs/blocks/marketo-config/context.js
@@ -21,6 +21,20 @@ function deepCopy(obj) {
   return JSON.parse(JSON.stringify(obj));
 }
 
+export const sanitizeConfigValue = (value) => {
+  if (typeof value !== 'string') return value;
+  const el = document.createElement('span');
+  el.innerHTML = value;
+  return el.textContent;
+};
+
+export const sanitizeHashConfig = (config) => {
+  if (!config || typeof config !== 'object') return config;
+  return Object.fromEntries(
+    Object.entries(config).map(([key, value]) => [key, sanitizeConfigValue(value)]),
+  );
+};
+
 /* c8 ignore next 7 */
 const getHashConfig = () => {
   const { hash } = window.location;
@@ -28,7 +42,7 @@ const getHashConfig = () => {
   window.location.hash = '';
 
   const encodedConfig = hash.startsWith('#') ? hash.substring(1) : hash;
-  return parseEncodedConfig(encodedConfig);
+  return sanitizeHashConfig(parseEncodedConfig(encodedConfig));
 };
 
 const getInitialState = (defaultState, lsKey) => {
--- a/libs/blocks/marketo-config/context.js
+++ b/libs/blocks/marketo-config/context.js
@@ -23,9 +23,7 @@ function deepCopy(obj) {
 
 export const sanitizeConfigValue = (value) => {
   if (typeof value !== 'string') return value;
-  const el = document.createElement('span');
-  el.innerHTML = value;
-  return el.textContent;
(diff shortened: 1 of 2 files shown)


mokimo (approved):

libs/blocks/marketo-config/context.js
```diff
@@ -21,14 +21,28 @@ function deepCopy(obj) {
   return JSON.parse(JSON.stringify(obj));
 }
 
+export const sanitizeConfigValue = (value) => {
+  if (typeof value !== 'string') return value;
+  const el = document.createElement('span');
+  el.innerHTML = value;
+  return el.textContent;
```
overmyheadandbody: I'm not certain this is the full picture. I created a quick demo here - https://jsbin.com/geberezihe/edit?html,js,output

The return value might be correct, but a malicious script would still execute. The unit tests should check that custom scripts are not executed during the sanitization.

Brandon32: @jsandland You should look at new DOMParser

jsandland: I will explore. Thanks @overmyheadandbody and @Brandon32.

Brandon32 (approved):

overmyheadandbody (approved):

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-200538

PR #53, opened 2022-06-22
Column table

seanchoi-dev:
* Enabling `table` option for column block.
* Setting up variables.css to share milo CSS variables with other helix sites.

Resolves: [MWPW-111569](https://jira.corp.adobe.com/browse/MWPW-111569)

**Test URLs:**
- Before: https://column-fix--business-website--adobe.hlx.page/blog/basics/business-case
- After: https://column-fix--business-website--adobe.hlx.page/blog/basics/business-case?milolibs=column-table

PR #6227, opened 2026-06-25
CaaS auto-publish: extract payload builder into leaf module + standalone bundle

shkhan91:
## What & why

Milo is static-only (AEM Edge Delivery) and cannot host a backend endpoint, but the CaaS card→XDM business logic must be reusable from a backend service (`milo-caas`) that auto-publishes pages to CaaS on a schedule. This PR extracts that logic into a **self-contained leaf module** and produces a **small standalone bundle** that `milo-caas` fetches (pinned by SHA-256) and runs in a `vm` sandbox — keeping a single source of truth here in Milo.

This PR is **library/tooling only** — it does not change any page rendering and is behavior-preserving for the existing browser "Send to CaaS" / bulk-publish paths.

* Add `buildCaasXdmPayload()` (plus `hasCardMetadata` / `isDisabledOnPage`) in a **new self-contained leaf** `tools/send-to-caas/caas-payload-core.js` that does **not** import `libs/utils/utils.js` or `libs/blocks/caas/utils.js` (those transitively drag in Milo's bootstrap, a 706KB tag taxonomy, and lazy chunks — which makes the module impossible to bundle standalone). The only `libs/` import is the leaf-safe `getUuid.js`.
* `send-utils.js` now **re-exports** the payload surface from the leaf and keeps only the browser-only pieces (`postDataToCaaS`, `isPagePublished`). It **injects** the 706KB `caas-tags` fallback lazily via `setConfig({ getCaasTagsFallback })`, so the leaf stays free of it and the browser never eagerly downloads it when the network tag fetch succeeds.
* Refactor `caasAutoPublish()` to call `buildCaasXdmPayload()` (removes the duplicated inline `setConfig → loadCaasTags → getCardMetadata → getCaasProps` sequence). No behavior change.
* Add `build/caas-payload-builder.js` entry + `build:caas-payload-builder` npm script (microbundle UMD, `--external none`, mirroring the existing `build:gnav-profile` pattern) producing a single **~20KB** committed bundle at `tools/send-to-caas/caas-payload-builder.js`, served from `milo.adobe.com`.

## Companion PR

This is **Part 1 of 2**. The backend service that consumes the bundle lives in `milo-caas`:
- milo-caas: https://git.corp.adobe.com/wcms/milo-caas/pull/53 (da-log-processor: scheduled CaaS auto-publish from DA logs)

Together: `milo-caas` polls the DA/Franklin admin log per tenant → fetches the page HTML → runs **this** bundle to build the XDM payload → posts to Chimera (preview → CaaS stage, publish → CaaS prod).

## Design notes for reviewers

* **Why a leaf + bundle instead of importing `send-utils.js` directly?** A standalone import of `send-utils.js` pulls in `utils.js` (2850 lines, top-level `window`/`document` side effects) and its lazy graph (e.g. `preflight-notification` ~170KB). The leaf isolates the pure payload logic so the bundle is ~20KB and `vm`-safe.
* **Why is the `caas-tags` fallback injected rather than imported?** microbundle UMD cannot code-split, so a dynamic import breaks the single-file build, while a static import would bloat the bundle to ~404KB and eager-load 706KB in the browser. Injecting it keeps the bundle tiny and the browser lazy. The leaf reaches the fallback only when the network tag fetch fails.
* **Card identity:** `contentId`/`entityId`/`url` derive from the `prodUrl` string — unchanged here, but this is why the consuming service applies per-tenant `.html` rules.

## Testing

* `tools/send-to-caas` + `blocks/caas` unit tests: **271 passing** (`auto-publish`, `bulk-publish-to-caas`, `send-to-caas`, `caas/utils`). No regressions.
* Bundle verified as a single 20KB UMD file with no code-split chunks; smoke-tested end-to-end from the `milo-caas` `vm`+`linkedom` loader against the real bundle (tags resolve, `caasProps` correct, `.html` drives `contentId`/`url`).

Resolves: [MWPW-117684](https://jira.corp.adobe.com/browse/MWPW-117684)

**Test URLs:**
- After: https://da-caas-autopublish--milo--shkhan91.aem.page/?martech=off


sheridansunier:
## Review findings

Good architecture overall — the leaf extraction and UMD bundle approach is clean. Found 3 confirmed bugs and 2 plausible ones, all in new code.

---

### 🔴 1. UMD bundle has no way to inject `getCaasTagsFallback` — hard crash on tags network failure

**File:** `tools/send-to-caas/caas-payload-builder.js`

The bundle entry only imports from `caas-payload-core.js` and never calls `setConfig`, so `getConfig().getCaasTagsFallback` is always `undefined` in the `milo-caas` vm context. When the chimera tags API has a transient failure, `loadCaasTags` exits with `tags = undefined`, then `getTag` calls `Object.values(undefined)` → `TypeError` — crashing every page in the batch with no structured error response.

**Fix options:**
- Export `setConfig` from `caas-payload-builder.js` so the backend can inject a Node-compatible fallback before calling `buildCaasXdmPayload`
- Or wire a fallback directly in the leaf using a conditional `require`/dynamic import that the bundler can tree-shake in browser contexts

---

### 🔴 2. Concurrent `buildCaasXdmPayload` calls race on shared mutable config

**File:** `tools/send-to-caas/caas-payload-core.js` (~line 1134) + `libs/blocks/bulk-publish-v2/auto-publish-hook.js` (~line 103)

`runAutoPublishForJob` launches all per-path `publish()` calls concurrently via `Promise.all`. Each call eventually reaches `buildCaasXdmPayload`, which writes `setConfig({ doc: dom, pageUrl, ... })` into a module-level IIFE closure, then immediately suspends at `await loadCaasTags()`. A second path's `setConfig` can overwrite `doc`/`pageUrl` before the first path's `getCardMetadata` reads `getConfig().doc` — page A's XDM payload is silently built from page B's DOM.

**Fix:** Thread config as a call-time parameter rather than mutating shared state, or process pages serially in `runAutoPublishForJob`.

---

### 🟠 3. `matchesUrl('/products/**', '/products')` returns `false` — bare directory path never matches its own `**` rule

**File:** `tools/send-to-caas/auto-publish.js` (~line 57)

`pattern.slice(0, -2)` for `'/products/**'` yields `'/products/'`. `'/products'.startsWith('/products/')` is `false`, so a publish for the index page `/products` is silently skipped with `'no-matching-rule'` even when the config intends to include it.

**Fix:**
```js
if (pattern.endsWith('**')) {
  const prefix = pattern.slice(0, -2);
  return path === prefix || path.startsWith(prefix);
}
```

---

### 🟡 4. `host` built as `'repo.owner'` instead of actual Helix hostname — corrupts image URLs

**File:** `libs/blocks/bulk-publish-v2/auto-publish-hook.js` (~line 93)

`host: owner ? \`${repo}.${owner}\` : ''` produces e.g. `'bacom.adobecom'`. `addHost()` uses `https://${host}${path}` for slash-relative image URLs, so any card with a root-relative `cardimage` path gets `https://bacom.adobecom/media/img.jpg` (non-resolvable) written to CaaS.

**Fix:** Use the full Helix hostname: `host: new URL(job.origin).hostname` (e.g. `'main--bacom--adobecom.aem.live'`).

---

### 🟡 5. Module-level `new URL(window.location.href)` crashes if vm has no window shim

**File:** `tools/send-to-caas/caas-payload-core.js` (~line 43)

This executes at import time. If the `milo-caas` vm context doesn't provide a valid `window.location.href`, the module fails to load before any function is callable. If `href` is `'about:blank'` (a common default shim), `lingoActive()` silently returns `false` for all pages. Worth confirming the vm setup covers this explicitly.

sheridansunier (approved):

sheridansunier:
Hey @shkhan91 — while reviewing this PR I noticed commit f4729b6 ("Reconcile Path A card identity with the milo-caas poller") includes a lot of files unrelated to its stated scope:

- `libs/blocks/chimera/{caas.css,caas.js,utils.js}` + `test/blocks/chimera/utils.test.js` — a ~1,300-line fork of the existing `libs/blocks/caas/` block. It isn't registered anywhere in `libs/utils/utils.js`, so it's currently dead/unreachable code.
- `.agents/skills/build-block-from-figma/*`, `build-content-from-figma/*`, `build-scroll-animation/*`, `source-command-build-c2-block/SKILL.md`
- `.codex/hooks.json`, `.scout/scout.config.yaml`, `AGENTS.md`

None of these existed on `stage` before this commit, and nothing later in the PR touches them — looks like they got swept in via a `git add -A`/`git commit -a` alongside unrelated work sitting in the same working directory.

Only about 7 files (~60-70 lines) in the commit actually relate to the stated identity-reconciliation fix: `tools/send-to-caas/{auto-publish.js,caas-payload-core.js,caas-payload-builder.js,send-utils.js}`, `libs/blocks/bulk-publish-v2/auto-publish-hook.js`, and `test/tools/send-to-caas/auto-publish.test.js`.

Could you rebase this commit to drop the unrelated additions (or replace it with a clean commit containing just the relevant files)? Happy to help if useful — just don't want the chimera fork or agent-tooling config landing on stage as a side effect of this fix.

Thanks!

sheridansunier:
Since I have push access to the base repo and this PR has "allow edits from maintainers" enabled, I went ahead and pushed a fix rather than waiting: d7a079aa6 removes the ~30 unrelated files from f4729b628 (chimera fork, agent-skill tooling, AGENTS.md, .codex/.scout config). It's a plain commit on top, not a rebase/force-push, so nothing in the existing history changed. Tests (274 passing) and lint are clean on the result.

sheridansunier:
Heads up: I force-pushed a rewritten history to this branch (previous tip was d7a079aa6).

Why: this branch had forked from `stage` before `MWPW-194951` (Add auto-detect Lingo to CaaS bulk publisher, #6142) landed, so it was missing 145 commits' worth of `stage` history. I rebased onto the current `stage` tip, reconciling the CaaS auto-publish logic against the newer Lingo auto-detect matrix (mainly in `getBulkPublishLangAttr`/`getCountryAndLang`/`isLingoLangFirstPath`), and added:

- `rule.autoDetectLingo` support, so a fully Lingo-enabled site (e.g. bacom) gets the same automatic per-URL Language-First detection in auto-publish that the bulk publisher's "Auto-detect Lingo" checkbox provides, instead of relying on a single manual `languageFirst` flag.
- `initBulkPublisherLingoMapping` in the leaf module (`caas-payload-core.js`) — it keeps its own `lingoSiteMappingPromise` separate from `libs/blocks/caas/utils.js`'s, so the bulk publisher needs this version to actually pre-warm/reset the cache `isLingoLangFirstPath`/`getBulkPublishLangAttr` read from.
- Tests covering the autoDetectLingo matrix (mapping-driven LFL, root-locale LFL, unmapped-locale fallback, and confirming autoDetectLingo has no effect on news — that still requires a manual `languageFirst` flag).

Content-wise nothing from the original commits was dropped — I verified the diff against the old tip file-by-file before pushing. 304 CaaS-related tests passing, lint clean. Apologies for the disruption if you had local work on the old tip; let me know if anything needs reconciling on your end.

jedjedjedM:
Summary: The review surfaced 8 findings, most notably two real backward-compatibility/correctness regressions worth fixing before merge:

1. Lingo site-mapping host silently changed (milo.adobe.com → www.adobe.com) — affects the existing bulk-publish caller.
2. Floodgate color dropped in the new auto-publish path — can produce mismatched/duplicate CaaS card identities for floodgated sites.
3. CaaS tag-fallback wiring lost for the new auto-publish path — transient network errors now hard-fail every page instead of degrading gracefully.
4. Plus a possible crash (mergeRetriedResources missing an array guard), a content-type-tag detection gap, a per-page tag re-fetch inefficiency, a matcher-semantics mismatch vs. its documented "mirrors" claim, and a fragile one-shot-fire design in the Lit component.


- tools/send-to-caas/caas-payload-core.js — Lingo host regression (line 395), missing tag-fallback wiring (line 692), per-page tag re-fetch (line 674), hasContentTypeTag ignoring primarytag (line 1307)
- libs/blocks/bulk-publish-v2/auto-publish-hook.js — floodgate color not derived/forwarded (line 101)
- libs/blocks/bulk-publish-v2/components/job-process.js — missing Array.isArray guard in mergeRetriedResources (line 78), fragile one-shot-fire design (line 76)
- tools/send-to-caas/auto-publish.js — matchesUrl anchoring mismatch vs. its documented parity claim (line 57)

sheridansunier:
Re #2 (floodgate) — dug into this further. It's a real gap, not something Shahbaz intentionally deferred as a fast-follow. Checked MWPW-117684 and both PR descriptions: floodgate is never mentioned anywhere, including milo-caas #53's own "Out of scope (follow-up PR)" section, which does explicitly call out unpublish/delete as deferred. So the practice of flagging known deferred scope was there — this one just wasn't considered (git blame traces the gap to the original hook implementation, commit 6329c00957).

Filed it as a separate follow-up rather than adding more scope to this already-long-open PR — floodgated content can still go through the manual bulk publisher in the meantime: [MWPW-207191](https://jira.corp.adobe.com/browse/MWPW-207191)

Also pushed a fix for #1 (Lingo host regression) in 7224d984 — restores the `milo.adobe.com` fetch to match `stage`'s behavior exactly.

sheridansunier:
Addressed the tag-fallback wiring gap (finding #3/#6):

- `findTag()` now guards against `tags` being `undefined` (was assuming it was always populated).
- `getTag()`'s Events-subtree fallback lookup uses optional chaining (`caasTags?.events?.tags`) instead of a bare property chain that would throw if `caasTags` was ever undefined.
- `loadCaasTags()` is now memoized — once a fetch succeeds it's a no-op on later calls, and on a failed re-fetch it keeps whatever tags were already cached rather than clobbering them with `undefined`.

This closes the gap for the milo-caas server/vm-sandbox path specifically: a transient network failure on the tag-taxonomy fetch previously left `caasTags` undefined and crashed the next `getTag()` lookup for every subsequent page in that run. The browser bulk-publish-v2 path was already covered transitively (via `send-utils.js`'s `getCaasTagsFallback` injection), so this was scoped to the backend path.

Pushed as `6008eaad6`. `send-to-caas` (62/62) and `bulk-publish-v2` (16/16) suites pass with no regressions.

Two related gaps are real but out of scope for this PR (pre-existing, not introduced here) — filed as follow-ups rather than expanding this PR further:
- [MWPW-207201](https://jira.corp.adobe.com/browse/MWPW-207201) — milo-caas has no offline tag-taxonomy fallback for a *cold-start* first-fetch failure (no prior cache to fall back to yet).
- [MWPW-207202](https://jira.corp.adobe.com/browse/MWPW-207202) — `da-log-processor` has no error-rate alerting on top of its per-page try/catch, so a sustained failure mode would currently only surface in logs.

sheridansunier:
Addressed two more of the 8 findings from the review above:

- **#4** (possible crash) — `mergeRetriedResources` now guards against `jobStatus.data.resources` not being an array (returns `[]` instead of throwing). Added a test.
- **#7** (matcher-semantics mismatch) — `matchesUrl`'s doc comment falsely claimed it mirrors `libs/tools/utils/publish.js`'s matcher. That matcher does an unanchored substring check (`webPath.includes(...)`) and would over-match (e.g. `/other/foo-launch` against `/foo/**`) — too loose for auto-publish routing rules. Corrected the comment rather than the code; `matchesUrl`'s stricter anchored behavior was already right.

Pushed as `1107883b2`. `send-to-caas` + `bulk-publish-v2` suites (113/113) pass, lint clean.

**#5** (`hasContentTypeTag` ignoring `primarytag`) is a real gap but touches a function shared across `caas-payload-core.js`, `send-utils.js`, `auto-publish.js`, and the built bundle `wcms/milo-caas` pins by checksum — filed as [MWPW-207292](https://jira.corp.adobe.com/browse/MWPW-207292) rather than expanding this PR further, same triage as #2/#3's residual gaps (MWPW-207191/207201/207202).

**#8** (fragile one-shot-fire design in `job-process.js`) — leaving as-is; the existing inline comments already document the rationale (waiting on the 503-retry queue to settle before firing), and it's not a correctness bug, just a design tradeoff Jed flagged as worth knowing about.

That's all 8 findings triaged: 3 fixed in code (#1, #3/#6, and now #4/#7), 1 fixed in doc comment only (#7), 3 filed as follow-ups (#2, #5, plus the two from the tag-fallback fix — MWPW-207201/207202).

sheridansunier:
Heads up — found (not from Jed's review) that the committed bundle was stale: `6008eaad6` and `7224d984` changed `caas-payload-core.js` but nobody re-ran `npm run build:caas-payload-builder`, so the tag-fallback-memoization and Lingo-host fixes were silently missing from `tools/send-to-caas/caas-payload-builder.js` — the actual artifact `milo-caas` fetches and checksum-verifies at runtime. Rebuilt and pushed as `0712dbfb3`.

SHA-256 changes `43b51d1a...` → `4062b167...`. **This means `wcms/milo-caas`'s `MILO_PAYLOAD_BUILDER_SHA256` deploy config will need bumping to the new hash once this merges** — flagging so it doesn't get missed at rollout.

jedjedjedM:
this for sure need to be addresses:

_tools/send-to-caas/caas-payload-core.js_
  ●  1326 [correctness] The new shared getProdUrl() always returns a scheme-qualified URL, but the untouched legacy bulk-publish-to-caas.js still builds prodUrl without a scheme, and both feed the same getUuid()-based contentId/entityId hash.

these were also flagged.  not sure about criticality:

_l_ibs/blocks/bulk-publish-v2/auto-publish-hook.js
  ●  115 [correctness] runAutoPublishForJob() only serializes CaaS-publish calls within a single job, but caas-payload-core.js's buildCaasXdmPayload() writes per-page request state into a shared module-level config singleton with no cross-job coordination.

_tools/send-to-caas/caas-payload-core.js_
  ●  581 [correctness] The setConfig/getConfig IIFE's initial config object reads the bare `document` global at module-evaluation time, inconsistent with the file's own lazy-window.location pattern used elsewhere specifically to be vm-sandbox-safe.
  ●  784 [correctness] getOrigin() falls back to a bare, unguarded window.location.hostname reference when repo/project can't be resolved, with no typeof-window guard anywhere in the file.
  ● 1121 [correctness] The primarytag prop calls getTag(s) with only one argument, leaving `errors` undefined; getTag's errors.push(tagName) then throws. The standalone bundle imports directly from this leaf and never registers the getCaasTagsFallback that send-utils.js sets up for browser callers.
  ●  813 [correctness] getImagePathMd does getConfig.pageUrl.origin — a property access on the getConfig function reference itself rather than on getConfig()'s return value — so .pageUrl is undefined and .origin throws.
  ●  687 [correctness] loadCaasTags() gained a new `if (tags) return;` memoization guard absent from the pre-PR send-utils.js, which always re-fetched tags on every call.

_tools/send-to-caas/auto-publish.js_
  ●   70 [correctness] resolveRule's specificity sort compares only raw url.length; when an exact-match rule and a wildcard rule tie in length, the stable sort falls back to original array order instead of preferring the exact match, inverting the documented precedence.

_libs/blocks/bulk-publish-v2/components/job-process.js_
  ●  215 [correctness] The new mergeRetriedResources matches retry-queue entries by webPath || path, while the pre-existing getJob(path) method's own retry lookup matches only by .path.

jedjedjedM:
The code review for PR #6227 finished. Here are the findings, most severe first:

Correctness bugs

1. job-process.js:94 — retry() overwrites this.queue with only the still-503 subset each round, dropping resources that already succeeded. Since mergeRetriedResources (line 81) only looks at what's left in this.queue, a page that succeeds on an early retry round gets silently left with its stale 503 status in the final job data — and is then excluded from CaaS auto-publish even though it actually published fine.
2. caas-payload-core.js:289 — getPageLocale()'s locale list is hardcoded to [] instead of being derived from the real page config (as libs/blocks/caas/utils.js does). This makes localizeCtaUrl's caaslocaleinject feature a permanent no-op — CTA URLs on localized pages (e.g. /de/) stop getting their locale prefix injected, a real regression not called out as breaking.
3. caas-payload-core.js:813 — getImagePathMd has getConfig.pageUrl.origin (missing the () call), which throws a TypeError. This bug pre-existed in the deleted send-utils.js, but this PR now wires it into the automatic fire-and-forget auto-publish path (job-process.js's updated() hook), so it fires on every successful bulk-publish job with a root-relative badge/card image instead of only when someone manually ran the legacy tool.
4. job-process.js:81 — this.jobStatus.data.resources is dereferenced without optional chaining, unlike the identical line two above it (this.jobStatus?.data?.resources?.filter(...)). If jobStatus.data is ever absent/malformed, this throws before the intended .catch(() => {}) fire-and-forget guard is even reached.
5. caas-payload-core.js:701 — The local tag-taxonomy fallback is now opt-in via setConfig({ getCaasTagsFallback }), but that's only ever called from send-utils.js. The standalone bundle (caas-payload-builder.js) imports caas-payload-core.js directly, bypassing that call — so if milo-caas doesn't independently configure the fallback and a tag fetch fails transiently, tags stays undefined for the rest of the process's lifetime, silently excluding the whole batch from auto-publish.

Reuse / drift risks

6. caas-payload-builder.js:1 — A generated bundle is committed and consumed by milo-caas via checksum, but nothing in CI verifies it's rebuilt from caas-payload-core.js — a future source edit can silently ship a stale bundle.
7. caas-payload-core.js:120 — Large data tables (LANGS, LOCALES) and other functions are hand-copied from libs/blocks/caas/utils.js with a "keep in sync" comment instead of a shared module — future edits to the original will silently drift.
8. auto-publish.js:59 — matchesUrl/resolveRule reimplement wildcard rule matching that already exists in libs/tools/utils/publish.js's userCanPublishPage, for the same config-file convention.
9. auto-publish-hook.js:33 — CONFIG_PATH and the "non-empty rules" opt-in check are duplicated verbatim between this file and auto-publish.js.
10. auto-publish.js:83 — fetchPageDom duplicates page-fetch-and-parse logic already in bulk-publish-to-caas.js's getPageDom, with signatures already diverging.

sheridansunier:
Addressed the correctness findings from the two most recent reviews (2026-09-14 and 2026-09-16). Pushed as `0131c3b03`.

**Fixed:**

- **job-process.js retry()** — was overwriting `this.queue` with only the still-503 subset each round, dropping entries that already succeeded/exhausted retries. `mergeRetriedResources` then fell back to their stale original 503, silently excluding a successfully-retried page from CaaS auto-publish. `getJob()`'s own retry lookup also only matched `.path`, unlike every other lookup in the file (webPath-or-path) — the UI status icon and the auto-publish gate could disagree on the same resource. Both fixed; added `applyRetryUpdate` + tests.
- **job-process.js `updated()`** — missing optional chaining on `this.jobStatus.data.resources`, inconsistent with the identical line two above it.
- **Card identity scheme mismatch** (the one flagged "for sure needs addressing") — `buildCaasXdmPayload` was feeding the same schemed `pageUrl` into both `config.pageUrl` (needed schemed, for graybox/image URL parsing) and `options.prodUrl` (hashed into contentId/entityId). The untouched legacy `bulk-publish-to-caas.js` tool hashes a scheme-less prodUrl, so the same page could hash to two different CaaS card identities depending on which publisher touched it. Now strips the scheme only at the hash input; `pageUrl`/`getProdUrl`'s schemed contract is untouched elsewhere (graybox/image resolution still need a real parseable URL). Added a regression test that would fail against the previous behavior.
- **vm-safety**: `setConfig`/`getConfig`'s initial config object read the bare `document` global at module-eval time — would throw on import in a vm sandbox with no `document` shim, defeating the whole point of the leaf module. Also fixed the dangling `this.doc` (should've been `config.doc`, never actually reachable before since it's dead code). `doc` now resolves lazily.
- **getPageLocale** — `pageLocales` was hardcoded to `[]`, making `localizeCtaUrl`'s `caaslocaleinject` feature a permanent no-op. Falls back to `Object.keys(LOCALES)` (the classic table already used as source-of-truth elsewhere in this file) since the leaf has no access to a site's actual configured locale list.
- **getImagePathMd** — `getConfig.pageUrl.origin` was missing the `()` call *and* treating a string as a URL object; threw on every root-relative card/badge image in the bulk-publish path. Fixed via the file's existing `prefixHttps` helper.
- **primarytag** — called `getTag(s)` with no `errors` array, so `errors.push(tagName)` threw whenever the metadata value didn't match any tag in the taxonomy (a plausible authoring typo). Now passes `[]`.
- **getOrigin** — bare `window.location.hostname` access with no guard, reachable when repo/project resolve to `''` (e.g. `auto-publish-hook.js`'s repo-parse-failure fallback) — would throw in a vm with no `window` global. Guarded.
- **resolveRule** — specificity sort compared raw pattern length, but a wildcard's `**` is match syntax, not literal path characters. A same-raw-length wildcard rule could tie an exact-match rule and, being a stable sort, win purely on array order. Now ranks by effective (dir-only) length and breaks ties in favor of exact matches. Added the exact repro case as a test (`/a/**` vs `/a/bc` in both orders).
- Deduplicated `CONFIG_PATH` between `auto-publish-hook.js` and `auto-publish.js` (minor reuse item).

Rebuilt `caas-payload-builder.js` from the updated source (SHA-256 `4062b167...` → `4841d971...`) — `wcms/milo-caas`'s `MILO_PAYLOAD_BUILDER_SHA256` will need bumping again on merge.

**Deliberately not changed** (matching the triage pattern from earlier in this thread — filing rather than expanding scope further):

- `auto-publish-hook.js:115` — no cross-job coordination on the shared config singleton (only within-job is serialized). You flagged uncertainty on criticality yourself; a real fix means a lock/mutex across concurrent `runAutoPublishForJob` calls, which is a bigger, riskier change than this pass warrants. Happy to file a ticket if you think it's worth prioritizing.
- `loadCaasTags` memoization "regression" vs pre-PR always-refetch — this was the deliberate fix for the original review's #3/#6 (crash-on-transient-failure + per-page refetch). `buildCaasXdmPayload`'s own doc comment calls the non-browser host "the milo-caas **scheduled** service," which suggests short-lived/periodic process invocations rather than a long-running daemon — so the staleness risk you're flagging is likely bounded to a single scheduled run in practice. Let me know if that assumption is wrong for the actual milo-caas deployment model.
- No CI check that the bundle is rebuilt from source — real risk (this thread has now hit it twice), but adding that check is a separate, larger change than this pass. Can file a follow-up ticket if wanted.
- LANGS/LOCALES hand-copied duplication, and the `matchesUrl`/`resolveRule` vs `publish.js` reimplementation, and `fetchPageDom` vs `getPageDom` duplication — all accepted/documented tradeoffs already (leaf-module constraint, deliberately-stricter-matcher, and diverging signatures respectively), no action taken.
- Incidental (not from your review): found `getCardImageAltText` throws `new URL(null)` for any card-metadata page with **no** image at all (no cardimage, no `<main>` img, no og:image) — pre-existing in the deleted `send-utils.js`, not introduced by this PR. Worked around it in the new tests; flagging here rather than fixing since it's out of scope, same as the other pre-existing gaps already filed as follow-ups.

`send-to-caas` + `bulk-publish-v2` + `caas` suites (368/368) pass, lint clean.

jedjedjedM:
libs/blocks/bulk-publish-v2/components/job-process.js:97 / libs/blocks/bulk-publish-v2/auto-publish-hook.js:113 (known limitation — tracking ask, not a blocker)

▎ Known limitation, not new: caas-payload-core.js's module-level config singleton is only serialized within a job (auto-publish-hook.js's sequential for...of), not coordinated across concurrent jobs — so two jobs running at once can still clobber each other's doc/host/pageUrl mid-flight. A real fix needs a lock/mutex across concurrent runAutoPublishForJob calls, which is bigger and riskier than this pass warrants, and the current sequential-within-a-job behavior is the deliberate (if costly) mitigation for it. Agree this shouldn't block merge — can you file a follow-up ticket for the cross-job coordination so it's tracked?

---

tools/send-to-caas/caas-payload-core.js:1022 (correctness)

▎ localizeCtaUrl derives the current locale from window.location.pathname/origin, but every sibling prop (country/lang/etc.) correctly uses options.prodUrl per the warning comment already in this file. In the bulk-publish/auto-publish flow and the milo-caas vm context, window.location is the tool's own page, not the page being processed — so CTA URL localization ends up keyed to the wrong page. Should this use options.prodUrl like the rest of the props?

---

tools/send-to-caas/caas-payload-core.js:48 (correctness)

▎ getMetadata(name, doc = document) defaults to the ambient global document instead of getConfig().doc, unlike getMetaContent right next to it, which correctly reads from getConfig().doc. Single-arg callers (lingoActive, localizeCtaUrl's caaslocaleinject read, the floodgate/country/entityid/lang props) will read the wrong document in the bulk-publish/auto-publish browser flow and in the milo-caas vm/backend context where the ambient document may not even be the injected page.

---

tools/send-to-caas/caas-payload-core.js:292 (correctness)

▎ pageLocales now falls back to the entire global LOCALES table (~200 entries) instead of the site's actual configured locale subset (pageConfigHelper().locales in the original). This overcorrects the documented no-op bug into an over-broad match — a site whose real locales are a strict subset of LOCALES will now treat any path segment that happens to exist anywhere in the global table as a valid locale, causing wrong CTA-URL locale-prefix injection (or incorrectly skipping it). This is distinct from the accepted LOCALES/LANGS duplication tradeoff — it's a behavior change in the fallback logic itself. Can this fall back to the site's configured locale list instead of the full global table?

---

tools/send-to-caas/caas-payload-core.js:1359 (correctness)

▎ getCaasIds(prodUrl, floodgatecolor) hashes prodUrl as-is (no scheme stripping), but buildCaasXdmPayload hashes a scheme-stripped prodUrl when computing the same ids at publish time. The docstring claims these are computed "exactly as on publish" — if milo-caas calls getCaasIds(getProdUrl(...)) (which returns a schemed URL) for card removal, the resulting contentId/entityId won't match what was published, and the wrong (or no) card gets targeted for removal. Should this strip the scheme the same way buildCaasXdmPayload does?

---

tools/send-to-caas/auto-publish.js:145 (correctness)

▎ When a matched /.milo/caas/config.json rule omits host (a shape this code explicitly allows via rule.host || host), the card's production identity falls back to the aem.page/aem.live hostname the job fired from, rather than the site's real production host — diverging from whatever prod host the milo-caas backend computes for the same page. Should a rule missing host be validated/rejected instead of silently falling back to the staging hostname?

---

libs/blocks/bulk-publish-v2/components/job-process.js:140 (correctness — minor)

▎ The retry-status lookup here uses item.webPath || item.path (coalesce, then compare), but the resource lookup two lines up uses src.path === path || src.webPath === path (compare-either). For an item with both fields set, if path matches item.path but not item.webPath, the resource lookup finds it but the retry lookup doesn't — so a successful retry's updated status never overwrites the stale one shown in the UI. Worth making these two lookups consistent.

---

tools/send-to-caas/send-utils.js:4 (breaking change — flag per repo convention)

▎ LANG_FIRST_SOURCE_MAPPINGS was a public named export of send-utils.js pre-PR; it's not re-exported here anymore (now lives only in caas-payload-core.js). This is a breaking change for any external code importing it directly from send-utils.js, and it isn't called out in the PR description alongside the "behavior-preserving" claim — can we either re-export it or note the removal explicitly?

sheridansunier:
Addressed the findings from the 2026-09-23 review. Pushed as `e038cd9e` + `11b1f7c3`.

**Fixed:**

- **`localizeCtaUrl` used `window.location`** — now derives the current-page path/URL-base from `options.prodUrl` when set, same rule `getCountryAndLang` already documents/applies a few lines up for the same reason (in bulk-publish/auto-publish, `window.location` is wherever the *tool* runs, not the page being processed).
- **`getMetadata` defaulted to the ambient `document`** — now defaults to `getConfig().doc`, matching `getMetaContent` right below it.
- **`pageLocales` over-corrected to the full global table** — `send-utils.js` now injects the browser's real `getConfig().locales` (mirrors `libs/blocks/caas/utils.js`'s own `pageConfigHelper().locales` read, same module-eval timing) so bulk-publish/send-to-caas use the site's actual configured locale subset; the milo-caas vm/backend context has no live page config to inject and keeps the broader table fallback as the best available option there.
  - Caught a real bug in my own first pass on this: I initially injected the bare `getConfig().locales` value, so a genuinely non-localized site (where that's legitimately `undefined`) was indistinguishable from "never injected at all" and fell through to the full-table fallback anyway — meaning a non-localized site could get a spurious locale prefix injected into a CTA URL. Fixed by injecting `|| {}` instead, and added `test/tools/send-to-caas/caas-payload-locales.test.js` — there was no existing coverage anywhere for `caaslocaleinject`/this fallback, across any of the three publish paths.
- **`getCaasIds` didn't strip the scheme** — now strips it the same way `buildCaasXdmPayload` does before hashing. Confirmed against `wcms/milo-caas`: `computeCaasIds(code, pageUrl)` (the removal path) and `buildCaasXdmPayload`'s own hash were computing different ids for the same page when `pageUrl` was schemed — this closes that gap.
- **`job-process.js`'s retry lookup** — now matches `path === path || webPath === path` (compare-either), consistent with the resource lookup two lines above it, instead of coalesce-then-compare.
- Bonus, not separately called out but same root cause as the export list: **`LANG_FIRST_SOURCE_MAPPINGS`** was a public named export of `send-utils.js` pre-refactor and got dropped when it moved into the leaf — re-exported.

**Not changed** (2 items) — both because the fix would be a bigger behavior/architecture change than this pass warrants:

- Cross-job coordination lock — agreed this shouldn't block merge. Will file a follow-up ticket for the cross-job coordination if that works, rather than scope-creeping this PR further.
- `auto-publish.js`'s `rule.host` fallback — nearly every existing rule fixture in `test/tools/send-to-caas/auto-publish.test.js` omits `rule.host` and relies on the caller-supplied `host` fallback today. Rejecting/validating it outright would silently stop CaaS auto-publish for any currently-configured site that doesn't set `rule.host`, which is a bigger blast radius than a review-comment-driven fix should take on unilaterally. Also worth a follow-up rather than folding into this PR.

**Verification:**

- `send-to-caas` + `blocks/caas` + `bulk-publish-v2` suites: 371/371 passing (up from 368 — added the new locale-fallback coverage).
- Rebuilt `tools/send-to-caas/caas-payload-builder.js` (the source changes above touch `caas-payload-core.js`, and the committed bundle would otherwise go stale again per the earlier lesson in this thread). New SHA-256: `b488cfaa15b5e5122f60f18371d255e6b0faae71ac5268840565fc1f839d62d0` (was `4841d9711199abf8cbf0607196ce91e71fbe1862e53e9f632ef73286f9ee648`) — `wcms/milo-caas`'s `MILO_PAYLOAD_BUILDER_SHA256` deploy config will need bumping to the new hash once this merges, same as before.
- Ran the rebuilt bundle through `milo-caas`'s real, unmodified `actions/shared/payload-builder-loader.js` (not a mock) end to end: publish/delete card-identity parity, `autoDetectLingo` resolution for `/fr/`, root, and `/uk/` (correctly non-LFL), CTA locale injection, and the tag-fetch-failure guard all hold against the actual bundle.
- Ran the full milo suite 4x with these changes and once at the pre-fix commit as a baseline (all on the same port, since `codeRoot`/`miloLibs` resolution is port-sensitive in this test setup). Baseline: 4742 passed, 1 failed. With-fix runs: 4742–4810 passed, 1–4 failed. Every failure across all 5 runs, including the baseline, is either a `global-navigation`/`Navigation component` test (external cookie-consent/IMS network flake, different specific sub-test each run) or, in one run only, two `libs/utils/utils.js` IMS/link-localization tests that use hardcoded real-clock `setTimeout(100)` races unrelated to any file this PR touches — didn't reproduce in 3 other identical runs or in isolation. Zero failures, in any run, in any CaaS/send-to-caas/bulk-publish-v2 test.

jedjedjedM:
1. caas-payload-core.js:1114 (correctness) — the window.location → options.prodUrl fix for localizeCtaUrl was only wired up for cta1url. cta2url still calls it with no options, so CTA-2 locale injection is still keyed to the bulk-publish tool's own page rather than the page being processed — the exact bug class this commit claims to have closed, just not for both CTA slots.

2. send-utils.js:51 (correctness) — the new locale-sourcing works for the paths tested, but the sidekick "Send to CaaS" entry point always loads send-to-caas.js from a hardcoded https://milo.adobe.com/... URL, so its relative libs/utils/utils.js import resolves to a separate, never-configured module instance. On a real localized site reached via the sidekick button, getMiloConfig().locales is undefined there, so caaslocaleinject silently stays a no-op — even though the fix description implies this path is covered.

PR #6237, opened 2026-07-01
MWPW-193755 country param

vhargrave:
Allow for country param to be used as default now instead of both it and akamaiLocale.

Resolves: [MWPW-193755](https://jira.corp.adobe.com/browse/MWPW-193755)

**Test URLs:**
- Before: https://main--da-bacom--adobecom.aem.live/?country=fr
- After: https://main--da-bacom--adobecom.aem.live/?country=fr&milolibs=vhargrave-country-param

Steps to test: 
1. Join the basel vpn
2. Go to the before link - notice how the language banner wants to take you to the german experience
3. Go to the after link
4. Notice how the language banner now wants to take you to french experience. That shows the country param now working for this use case too.

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -964,12 +964,15 @@ function setCountry() {
   sessionStorage.setItem('feds_location', JSON.stringify({ country: country.toUpperCase() }));
 }
 
+export function getCountryFromParams(searchParams) {
+  const validate = (v) => (typeof v === 'string' && /^[a-zA-Z]{2,6}$/.test(v) ? v : null);
+  return validate(searchParams.get('country')) || validate(searchParams.get('akamaiLocale'));
+}
+
 export async function getCountry(skipFallback = false) {
   if (isBot()) return null;
 
-  const rawAkamai = PAGE_URL.searchParams.get('akamaiLocale');
-  const akamaiLocale = /^[a-zA-Z]{2,6}$/.test(rawAkamai) ? rawAkamai : null;
-  const country = akamaiLocale || sessionStorage.getItem('akamai');
+  const country = getCountryFromParams(PAGE_URL.searchParams) || sessionStorage.getItem('akamai');
   if (country || skipFallback) return country?.toLowerCase();
 
   try {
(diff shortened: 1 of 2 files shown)


zagi25 (approved):

libs/utils/utils.js
```diff
@@ -964,12 +964,13 @@ function setCountry() {
   sessionStorage.setItem('feds_location', JSON.stringify({ country: country.toUpperCase() }));
 }
 
-export async function getCountry(skipFallback = false) {
+export async function getCountry(skipFallback = false, searchParams = PAGE_URL.searchParams) {
```
narcis-radu, line 967: I'm trying to find an use-case where `getCountry` is called with different search params. I image we'd always use `searchParams`, so not sure if we really need to pass that as argument.

vhargrave: @narcis-radu it's not necessary at all for the productive code yes. I added for the unit tests. I know we try to heavily limit code for tests, but adding a param seemed small enough to get away with. Was my thinking anyway.

narcis-radu (approved):

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details [MWPW-193755](https://jira.corp.adobe.com/browse/MWPW-193755)

sukamat:
@vhargrave Thanks for the update.

Just before the break, Joel and I were discussing on certain use-cases and the country param/cookie came up. 

Order of preference that Joel was asking for, especially given that akamaiLocale is an internal testing query-param:

```
?country > country cookie > ?akamaiLocale > default
```

1. Country query-param
2. Country cookie
3. akamaiLocale query-param
4. default


Not sure if Joel logged a ticket for this but calling this out to make sure the above is catered (if not already).

sukamat:
@vhargrave Here are the other tickets Joel had created for what I mentioned above: 
* [MWPW-199975](https://jira.corp.adobe.com/browse/MWPW-199975) 
* [MWPW-199499](https://jira.corp.adobe.com/browse/MWPW-199499)

markpadbe (approved):

PR #6238, opened 2026-07-01
[MWPW-199397] - Rich content jump links hover addition

DKos95:
Resolves: [MWPW-199397](https://jira.corp.adobe.com/browse/MWPW-199397)

**Test URLs:**
- Before: https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-offer?milolibs=stage&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all
- After: https://main--da-dc--adobecom.aem.page/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-offer?milolibs=rich-contnet-jump-hover&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all

### DIFF
--- a/libs/mep/ace1205/rich-content/rich-content.css
+++ b/libs/mep/ace1205/rich-content/rich-content.css
@@ -121,6 +121,16 @@
         margin: 0;
         padding: 0;
 
+        &:has(.jump-link-anchor:focus-visible) .jump-link-anchor:not(:hover, :focus-visible) {
+          color: var(--s2a-color-transparent-white-48);
+
+          .jump-link-badge {
+            background: var(--s2a-color-transparent-white-16);
+            border-color: var(--s2a-color-transparent-white-12);
+            color: inherit;
+          }
+        }
+
         @media (hover: hover) {
           &:has(.jump-link-anchor:hover) .jump-link-anchor:not(:hover, :focus-visible) {
             color: var(--s2a-color-transparent-white-48);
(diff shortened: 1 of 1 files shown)


libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -119,6 +120,18 @@
...
         padding: 0;
+
+        @media (hover: hover) {
+          &:has(.jump-link-anchor:hover) .jump-link-anchor:not(:hover, :focus-visible) {
+            color: var(--s2a-color-transparent-white-48);
+
+            .jump-link-badge {
+              background: var(--s2a-color-transparent-white-16);
+              border-color: var(--s2a-color-transparent-white-12);
+              color: var(--s2a-color-transparent-white-48);
```
zagi25: Super nit:
```suggestion
              color: inherit;
```

zagi25 (approved):

overmyheadandbody (commented):
We usually use the same styles when an item is focused, could that be implemented as part of this PR as well? Otherwise looking 👌

DKos95:
> We usually use the same styles when an item is focused, could that be implemented as part of this PR as well? Otherwise looking 👌

If it is needed I don't have a problem implementing it, but the request in the ticket is to make it similar to nav, on nav when an item is focused via keyboard the effect is not happening.

overmyheadandbody (approved):
I'd say the missing focus requirement is an omission, I received this feedback for the FAQ block, so I'm assuming it's a practice we want to standardize, which I agree with

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-199397

PR #6239, opened 2026-07-01
[MWPW-195995] - Rich content authorable gradient

DKos95:
PR gives authors the ability to set gradient on rich-content per breakpoint.

Resolves: [MWPW-195995](https://jira.corp.adobe.com/browse/MWPW-195995)

**Test URLs:**
- Before: https://main--da-dc--adobecom.aem.page/drafts/dusan/rich-content-gradient-2?milolibs=stage&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all
- After: https://main--da-dc--adobecom.aem.page/drafts/dusan/rich-content-gradient-2?milolibs=rich-content-authorable-gradient&mep=%2Fdc-shared%2Ffragments%2Ftests%2F2026%2Fq2%2Face1205%2Fdc1205-base-page-dc.json--all

**HP test URLs:**
- Before: https://load-c2-styles--upp--adobecom.aem.live/homepage/drafts/ramuntea/redesign-demo?milolibs=stage
- After: https://load-c2-styles--upp--adobecom.aem.live/homepage/drafts/ramuntea/redesign-demo?milolibs=rich-content-authorable-gradient


zagi25:
From the authoring perspective, does it make sense to move gradient value to `section-metadata` or to the column next to the text ? Background image is already added via `section-metadata`, and I don't expect `rich-content` block to have an image on the side, so I guess we can also use right column for the gradient.  
Screenshot 2026-07-01 at 15 30 54

libs/utils/decorate.js
```diff
@@ -709,7 +709,63 @@ function resolveInheritance(rows, previousContent) {
   });
 }
 
+const HERO_GRADIENT_KEY = 'hero-gradient';
+const HERO_GRADIENT_VARIANTS = [
+  ['jump-link', '--rc-hero-jump-link-gradient'],
```
zagi25: Since for the both variants, gradient is applied to the same element, why do we need two specific variables ?

DKos95:
> From the authoring perspective, does it make sense to move gradient value to `section-metadata` or to the column next to the text ? Background image is already added via `section-metadata`, and I don't expect `rich-content` block to have an image on the side, so I guess we can also use right column for the gradient.>

I don't know if it makes sense moving it to section-metadata since it's specific to only rich-content.
On moving it to the column next to text I thought about that, but what if they want to have a background color and gradient, would it maybe be confusing, this way they are kind of separate?

libs/utils/decorate.js
DKos95: You are right, no need, thanks.

zagi25:
> > From the authoring perspective, does it make sense to move gradient value to `section-metadata` or to the column next to the text ? Background image is already added via `section-metadata`, and I don't expect `rich-content` block to have an image on the side, so I guess we can also use right column for the gradient.>
> 
> I don't know if it makes sense moving it to section-metadata since it's specific to only rich-content. On moving it to the column next to text I thought about that, but what if they want to have a background color and gradient, would it maybe be confusing, this way they are kind of separate?

@DKos95 , I guess if they want background color they can apply gradient directly to it ? Also since background image is added via `section-metadata`, color can be added the same way and gradient can be put on top right ?

DKos95:
> @DKos95 , I guess if they want background color they can apply gradient directly to it ? Also since background image is added via `section-metadata`, color can be added the same way and gradient can be put on top right ?

Ah I wasn't aware of the ability to set a solid color with linear-gradient and have shadows like that, thanks for the awesome feedback! 

Updated the code to make it authorable similar to what we have with background-color in the 2nd column and I do agree it looks a lot nicer 🥳 .

Screenshot 2026-07-02 at 13 37 41

libs/utils/decorate.js
```diff
@@ -790,14 +847,17 @@ function applyViewportContent(el, viewports) {
  * - block — the element to decorate (viewport container or el itself)
  * - root  — for checking base classes on detached containers */
 export function decorateViewportContent(el, decorateFn) {
+  const hasHeroGradient = supportsHeroGradient(el);
```
narcis-radu: nit: `hasHeroGradient` is a bit misleading in the context, since you're only checking whether it _could_ support a gradient.

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -201,33 +201,42 @@
   margin-block-end: 0;
 }
 
-.section.has-background:has(.rich-content.hero) .section-background picture::after {
+.section:has(.rich-content.hero:not(.jump-link)) {
```
narcis-radu: I assume https://github.com/adobecom/milo/pull/6239/changes#diff-5e0dc32bc992cde90812fb7ffe7c8f222b0379dd7da7db614bed65312db5d10dR62 and this are different strictly because of the ongoing test and production page.

libs/utils/decorate.js
```diff
@@ -709,7 +709,53 @@ function resolveInheritance(rows, previousContent) {
   });
 }
 
+const HERO_GRADIENT_PROP = '--rc-hero-gradient';
+
+function isCssGradient(value) {
+  return /^(repeating-)?(linear|radial|conic)-gradient\(.+\)$/i.test(value?.trim());
+}
+
+function supportsHeroGradient(el) {
```
narcis-radu: is there a way to keep all the rich-content gradient specific logic outside of `decorate`? The logic seems pretty specific to one use-case.

libs/mep/ace1205/rich-content/rich-content.css
DKos95: ace1205 has two separate default gradients from desgin:

.hero:not(.jump-link) → solid black edge fade (standard hero)
.jump-link.hero → softer transparent-black fade (jump-link hero)

libs/utils/decorate.js
DKos95: I've created a utils for rich content and put the shared functions there so we can use them both for mep and the regular rich content block so we don't repeat ourselves, don't think it's possible to not touch decorate.js at all though, but the rich-content specific logic is moved out as much as possible now I think.

narcis-radu (approved):
Will need testing, but looks good to me.

libs/mep/ace1205/rich-content/rich-content.css
overmyheadandbody: Personally, I'm not a big fan of `:not`, especially when it's likely new variants will pop up. It increases specificity and it makes overrides more difficult in the future. Can it be avoided? Can the jump link variant define a transparent `--rc-hero-gradient` instead?

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -201,33 +201,42 @@
...
+  --rc-hero-gradient: linear-gradient(180deg, var(--s2a-color-transparent-black-80) -0.02%, var(--s2a-color-transparent-black-50) 39.99%, var(--s2a-color-transparent-black-20) 59.99%, var(--s2a-color-transparent-black-80) 100%);
 }
 
 .section.has-background:has(.rich-content.jump-link.hero) .section-background picture::after {
-  background: linear-gradient(180deg, var(--s2a-color-transparent-black-80) -0.02%, var(--s2a-color-transparent-black-50) 39.99%, var(--s2a-color-transparent-black-20) 59.99%, var(--s2a-color-transparent-black-80) 100%);
+  content: '';
+  position: absolute;
+  background: var(--rc-hero-gradient);
+  inset: 0;
+  z-index: 1;
```
overmyheadandbody: Is this really needed, since the `::after` pseudo-element is added after the `source` / `img` ones?

libs/utils/decorate.js
```diff
@@ -786,18 +813,24 @@ function applyViewportContent(el, viewports) {
...
+  const viewports = parseViewportContent(el, metadataHooks);
   if (viewports.hasViewportVariations) {
     Object.values(viewports.content).forEach(({ container }) => {
       decorateFn(container, el);
     });
-    applyViewportContent(el, viewports);
+    applyViewportContent(el, viewports, globalMetadata, metadataHooks);
   } else {
     decorateFn(el, el);
+    if (metadataHooks?.apply) metadataHooks.apply(globalMetadata, el);
```
overmyheadandbody:
```suggestion
    metadataHooks?.apply?.(globalMetadata, el);
```

overmyheadandbody (changes requested):
I stopped reviewing when I got to the utils file and all the changes in decorate.js. I believe we can find a better way to do this with a smaller blast radius and lower headroom. I've started some investigation on this, but I need more time to have a POC. The benefits of the feature are outweighed by the code additions at this point, in my opinion. I'll get back to this tomorrow.

overmyheadandbody:
Hey @DKos95! Here's what I was thinking of: https://github.com/adobecom/milo/compare/stage...rc-overlay-author, you can preview it at https://main--da-dc--adobecom.aem.page/drafts/dusan/rich-content-gradient-2?milolibs=rc-overlay-author. It makes no changes to `decorate.js`, has a lower code footprint, doesn't add new files. Is it missing any of your acceptance criteria?

libs/mep/ace1205/rich-content/rich-content.css
```diff
@@ -201,33 +201,46 @@
...
 
-.section:not(.has-background):has(.rich-content.hero) {
-  background: linear-gradient(180deg, #000 10.13%, var(--s2a-color-transparent-black-00) 70.89%, var(--s2a-color-transparent-black-00) 75.53%, #000 95%), var(--s2a-color-background-default);
+.section:not(.has-background):has(.rich-content.hero:not(.jump-link)) {
+  background: var(--rc-hero-overlay), var(--s2a-color-background-default);
 }
 
 .section:has(.rich-content.jump-link.hero) {
   height: 584px;
+  --rc-hero-overlay: linear-gradient(180deg, var(--s2a-color-transparent-black-80) -0.02%, var(--s2a-color-transparent-black-50) 39.99%, var(--s2a-color-transparent-black-20) 59.99%, var(--s2a-color-transparent-black-80) 100%);
```
overmyheadandbody: As a rule, we have custom properties first, then declarations, to keep things organized.

overmyheadandbody (approved):
Thanks for adapting this! A few of my previous comments still stand, but otherwise looking good!

DKos95:
> Thanks for adapting this! A few of my previous comments still stand, but otherwise looking good!

Thanks for helping reduce code, wasn't aware of that Observer technique with returned viewports, that alone saved 50 lines probably!

zagi25 (approved):

overmyheadandbody (approved):

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details [MWPW-195995](https://jira.corp.adobe.com/browse/MWPW-195995).

PR #6240, opened 2026-07-01
Resolve unintended click behavior in editorial card

jsmitrah26:
Resolves: [MWPW-199911](https://jira.corp.adobe.com/browse/MWPW-199911)

**Description**
 
Hi While exploring the Adobe Blog components in Milo, I came across an interaction issue in the Editorial Card footer. Since this was a good opportunity to contribute, I implemented a fix for the issue.
 
When the Editorial Card footer contains only a text-based call-to-action, the entire footer area becomes interactive instead of just the intended link. This change limits the clickable area to the call-to-action element, ensuring consistent interaction behavior across all footer variants.
 
 
**Steps to Reproduce**
 
1. Add an Editorial Card (default or open variant).
2. Add a text-based call-to-action in the card footer.
 
**Expected:** Only the CTA text is clickable.

**Actual:** The entire footer area is clickable.
 
**Test URLs**

**Before:**  https://main--mitrah-milo-hub--jsmitrah26.aem.live/

**After:** https://main--mitrah-milo-hub--jsmitrah26.aem.live/?milolibs=fix-editorial-cta-click--milo--jsmitrah26

### DIFF
--- a/libs/blocks/editorial-card/editorial-card.css
+++ b/libs/blocks/editorial-card/editorial-card.css
@@ -55,6 +55,10 @@
   font-size: var(--type-body-s-size);
 }
 
+.editorial-card .card-footer > div > a:only-child { 
+  align-self: end; 
+}
+
 .editorial-card.static-links-copy .foreground a:not([class*="button"]) {
   color: inherit;
   text-decoration: underline;
--- a/libs/blocks/editorial-card/editorial-card.js
+++ b/libs/blocks/editorial-card/editorial-card.js
@@ -57,9 +57,6 @@ const decorateForeground = async (el, rows) => {
     } else if (i === (rows.length - 1)) {
       row.classList.add('card-footer');
       if (!row.textContent.trim()) row.classList.add('empty');
-      row.querySelectorAll('div > a:only-child').forEach((link) => {
-        link.parentElement.classList.add('action-area');
-      });
     } else {
       row.classList.add('extra-row');
     }
--- a/libs/blocks/editorial-card/editorial-card.js
+++ b/libs/blocks/editorial-card/editorial-card.js
@@ -57,6 +57,9 @@ const decorateForeground = async (el, rows) => {
     } else if (i === (rows.length - 1)) {
       row.classList.add('card-footer');
       if (!row.textContent.trim()) row.classList.add('empty');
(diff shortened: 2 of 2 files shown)


narcis-radu:
Hi @jsmitrah26 - do you have a real life example of the issue?

jsmitrah26:
Hi @narcis-radu, I haven't found a production page using this exact configuration yet. I identified the behavior while exploring the Editorial Card component and reproduced it using a demo page to validate the fix.

vhargrave (approved):
Approved — thanks for the contribution! 🎉

The fix is correct. The footer's inner `div` is a flex **column**, so a lone link stretches to the full width (default `align-items: stretch`) — that's why the whole footer was clickable. Adding `.action-area` flips it to a **row** so the link only takes up its own width.

One friendly suggestion (non-blocking): this could be done in pure CSS without the JS, which is a bit cleaner:

```css
.editorial-card .card-footer > div > a:only-child { align-self: end; }
```

`align-self: end` overrides the column's `align-items: stretch` for just that lone link, so it shrinks to its content and right-aligns (matching the footer's `text-align: end`) — same result, no JS. Using `.action-area` also works, though that class is used elsewhere in Milo mainly for button CTAs (see `decorateButtons`), so the CSS-only route keeps things a little more focused. Totally your call.

Other optional notes:
- A quick unit test for the footer link no longer stretching would be nice — the current suite doesn't cover it.
- A bare link in `.foreground > div` (also a flex column) would have the same issue, if you want to cover it too. Out of scope here, just flagging.

jsmitrah26:
Hi @vhargrave , thanks for the review and the suggestions! I've updated the implementation to follow the CSS-only approach as suggested and have pushed the latest changes, Thanks!

JasonHowellSlavin (approved):

jsmitrah26:
Hi @vhargrave , I don't have the required permissions to add the Ready for Stage label. Could you please add it? Thanks!

vhargrave:
> Hi @vhargrave , I don't have the required permissions to add the Ready for Stage label. Could you please add it? Thanks!

@jsmitrah26 one of our QAs will take a look and assign the label 🙌

jsmitrah26:
@vhargrave  Great, thanks for letting me know!

PR #6252, opened 2026-07-06
[MWPW-198999] - Added Offer hero updates for CPro offer page

Deva309:
Added some updates related to `offer-hero` block on CPro offer page.

Resolves: [MWPW-198999](https://jira.corp.adobe.com/browse/MWPW-198999)

**Test URLs:**
- After: https://offer-hero-updates--milo--adobecom.aem.page/?martech=off

**QA:**
https://main--da-cc--adobecom.aem.page/cc-shared/fragments/tests/2026/q3/ace1209/fragments/cpro-offer?milolibs=offer-hero-updates&mep=%2Fcc-shared%2Ffragments%2Ftests%2F2026%2Fq3%2Face1209%2Face1209-cpro-redesign.json--all

### DIFF
--- a/libs/c2/styles/styles.css
+++ b/libs/c2/styles/styles.css
@@ -497,6 +497,7 @@
   --s2a-color-content-strong: var(--s2a-color-gray-600);
   --s2a-color-content-subheading: var(--s2a-color-content-knockout);
   --s2a-color-content-subtle: var(--s2a-color-transparent-white-64);
+  --s2a-color-content-body-subtle: var(--s2a-color-content-subtle);
   --s2a-color-content-heading: var(--s2a-color-content-knockout);
   --s2a-color-focus-ring-default: var(--s2a-color-blue-500);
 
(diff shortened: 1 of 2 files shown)


libs/c2/styles/styles.css
```diff
@@ -497,6 +497,7 @@
   --s2a-color-content-strong: var(--s2a-color-gray-600);
   --s2a-color-content-subheading: var(--s2a-color-content-knockout);
   --s2a-color-content-subtle: var(--s2a-color-transparent-white-64);
+  --s2a-color-content-body-subtle: var(--s2a-color-content-subtle);
```
overmyheadandbody: This is already defined - https://github.com/adobecom/milo/blob/e7443b590e7feeb8673cc48865e700b98d41768c/libs/c2/styles/styles.css#L354. Not only that, but all of the `--s2a-*` tokens come from the Consonant library and should always be in sync. As soon as there is a new release on the Consonant side, any changes manually made by developers will get lost.

This should be removed.

overmyheadandbody (commented):
Looking good; however, it would be best to remove manually added S2A tokens from the styles file.

libs/c2/styles/styles.css
Deva309: @overmyheadandbody 
I have removed this, but since the already defined is picking up the wrong color, we will need to redefine it inside the `.dark` theme to ensure the correct color is applied.

narcis-radu (approved):

DKos95 (approved):
Looks good to me.

overmyheadandbody (approved):

SilviuLCF (approved):
Verified , testing details MWPW-198999

PR #6258, opened 2026-07-07
Fix: last column not visible in 2x2 tables on mobile and tablet views

jsmitrah26:
Resolves: [MWPW-199959](https://jira.corp.adobe.com/browse/MWPW-199959)

**Description**

As part of exploring the Adobe Blog components in Milo, I identified an issue with the Table block's responsive column display. I investigated the root cause, implemented a fix, and contributed the changes.

When a Table block contains exactly two columns, the first column becomes hidden on mobile and tablet viewports, leaving only one column visible. This change automatically preserves both columns for two-column tables, ensuring they remain visible and accessible across smaller viewports while keeping the existing behavior unchanged for tables with more than two columns.

**Steps to Reproduce**

1. Add a Table block with exactly two columns.
2. View the page on a mobile or tablet viewport.

**Expected:** Both columns remain visible.

**Actual:** The first column is hidden, leaving only one column visible.

**Test URLs**

**Before:** https://main--mitrah-milo-hub--jsmitrah26.aem.live/table-block

**After:** https://main--mitrah-milo-hub--jsmitrah26.aem.live/table-block?milolibs=table-preserve-column--milo--jsmitrah26

### DIFF
--- a/libs/blocks/table/table.css
+++ b/libs/blocks/table/table.css
@@ -692,6 +692,10 @@
     margin: 0;
   }
 
+  .table.preserve-columns .section-row-title {
+    grid-column: auto !important;
+  }
+
   .table:not(.merch) .row .section-head-title,
   .table:not(.merch) .row .section-row-title {
     border-right: 1px solid var(--border-color);
@@ -701,10 +705,23 @@
     display: block;
   }
 
+  .table.preserve-columns .col-heading.col-1,
+  .table.preserve-columns .row-highlight .col-highlight.col-1 {
+    display: flex !important;
+  }
+
   .table .section-head .col:not(.section-head-title),
   .table:not(.merch) .col-heading.col-1,
   .table:not(.merch) .row-highlight .col-highlight.col-1:not(:only-child) {
     display: none;
+  }
+
+  .table.preserve-columns .row-heading .col-heading:first-child {
+    border-top-left-radius: var(--border-radius);
+  }
+
+  .table.preserve-columns .row-heading .col-heading:last-child {
+    border-top-left-radius: 0;
   }
 
   .table:not(.merch) .row-highlight:has(> :only-child) {
--- a/libs/blocks/table/table.css
+++ b/libs/blocks/table/table.css
@@ -692,8 +692,8 @@
     margin: 0;
   }
 
-  .table.preserve-columns .section-row-title {
-    grid-column: auto !important;
+  .table.preserve-columns .row .section-row-title {
+    grid-column: auto;
   }
 
   .table:not(.merch) .row .section-head-title,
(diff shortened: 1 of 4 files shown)


libs/blocks/table/table.css
```diff
@@ -692,6 +692,10 @@
     margin: 0;
   }
 
+  .table.preserve-columns .section-row-title {
+    grid-column: auto !important;
```
zagi25: Can selector be adjusted so we can avoid `!important` ?

libs/blocks/table/table.css
```diff
@@ -701,12 +705,25 @@
     display: block;
   }
 
+  .table.preserve-columns .col-heading.col-1,
+  .table.preserve-columns .row-highlight .col-highlight.col-1 {
+    display: flex !important;
```
zagi25: Same as above.

libs/blocks/table/table.js
```diff
@@ -477,6 +477,9 @@ function applyStylesBasedOnScreenSize(table, originTable) {
   const deviceBySize = defineDeviceByScreenSize();
 
   const setRowStyle = () => {
+    if (table.classList.contains('preserve-columns')) {
+      return;
+    }
```
zagi25: nit;
```suggestion
    if (table.classList.contains('preserve-columns')) return;
```

libs/blocks/table/table.js
```diff
@@ -670,6 +676,13 @@ export default function init(el) {
     expandSection = handleSection(sectionParams);
   });
 
+  const firstRow = el.querySelector('.row-1');
+  const columnCount = firstRow?.children.length || 0;
+
+  if (columnCount <= 2) {
+    el.classList.add('preserve-columns');
+  }
```
zagi25: nit:
```suggestion
  if (columnCount <= 2) el.classList.add('preserve-columns');
```

libs/blocks/table/table.js
```diff
@@ -670,6 +676,13 @@ export default function init(el) {
     expandSection = handleSection(sectionParams);
   });
 
+  const firstRow = el.querySelector('.row-1');
+  const columnCount = firstRow?.children.length || 0;
```
zagi25: nit:
```suggestion
  const columnCount = el.querySelector('.row-1')?.children.length || 0;
```

jsmitrah26:
Thanks for the review, @zagi25 !

I've addressed all the feedback and pushed the latest changes, including the CSS specificity update and the suggested JS improvements. Thanks!

zagi25:
@jsmitrah26 has this been run by Consonant or anyone else ? Although it does fix the issue it changes look of the table, currently rows are stacked, with your fix they stay as columns. 
Before: 
Screenshot 2026-07-10 at 14 58 13
After:
Screenshot 2026-07-10 at 14 58 03

jsmitrah26:
Hi @zagi25 , Thanks for pointing this out. This hasn't been reviewed by the Consonant team yet. 

While exploring the Table block, I noticed this issue specifically with two-column tables, where the mobile/tablet column selector isn't available, resulting in one column being hidden with no way to access it.

That's why I decided to work on a fix for this specific case. I also discussed this with my colleague [VenkateshMitrah](https://github.com/VenkateshMitrah), who is a vendor worker on Adobe. He pointed me to the related tickets ([MWPW-194499](https://jira.corp.adobe.com/browse/MWPW-194499) and [MWPW-188525](https://jira.corp.adobe.com/browse/MWPW-188525)) and explained that they are currently on hold until the site redesign.

PR #6266, opened 2026-07-09
[MWPW-195995] - Rich content ace1209 linear gradient

DKos95:
Extension PR of https://github.com/adobecom/milo/pull/6239 in order to provide authorable linear gradient for ace1209 rich content.

Resolves: [MWPW-195995](https://jira.corp.adobe.com/browse/MWPW-195995)

**Test URLs:**
- Before: https://main--da-cc--adobecom.aem.page/cc-shared/fragments/tests/2026/q3/ace1209/fragments/cpro-offer?milolibs=site-redesign-foundation&mep=%2Fcc-shared%2Ffragments%2Ftests%2F2026%2Fq3%2Face1209%2Face1209-cpro-redesign.json--all
- After: https://main--da-cc--adobecom.aem.page/cc-shared/fragments/tests/2026/q3/ace1209/fragments/cpro-offer?milolibs=rc-ace1209-linear-gradient&mep=%2Fcc-shared%2Ffragments%2Ftests%2F2026%2Fq3%2Face1209%2Face1209-cpro-redesign.json--all

### DIFF
--- a/libs/mep/ace1209/rich-content/rich-content.css
+++ b/libs/mep/ace1209/rich-content/rich-content.css
@@ -242,7 +242,7 @@
 }
 
 .section:has(.rich-content.jump-link.hero) {
-  --rc-hero-overlay: linear-gradient(180deg, var(--s2a-color-transparent-black-80) -0.02%, var(--s2a-color-transparent-black-50) 39.99%, var(--s2a-color-transparent-black-20) 59.99%, var(--s2a-color-transparent-black-80) 100%);
+  --rc-hero-overlay: linear-gradient(180deg, var(--s2a-color-transparent-black-80) -0.02%, var(--s2a-color-transparent-black-48) 39.99%, var(--s2a-color-transparent-black-24) 59.99%, var(--s2a-color-transparent-black-80) 100%);
   height: 584px;
 }
 
--- a/libs/mep/ace1209/rich-content/rich-content.css
+++ b/libs/mep/ace1209/rich-content/rich-content.css
@@ -226,33 +226,33 @@
   margin-block-end: 0;
 }
 
+.section:has(.rich-content.hero) {
+  --rc-hero-overlay: linear-gradient(180deg, var(--s2a-color-gray-1000) 10.13%, var(--s2a-color-transparent-black-00) 70.89%, var(--s2a-color-transparent-black-00) 75.53%, var(--s2a-color-gray-1000) 95%);
+}
+
 .section.has-background:has(.rich-content.hero) .section-background picture::after {
   content: '';
   position: absolute;
-  background: linear-gradient(180deg, #000 10.13%, var(--s2a-color-transparent-black-00) 70.89%, var(--s2a-color-transparent-black-00) 75.53%, #000 95%);
+  background: var(--rc-hero-overlay);
   inset: 0;
-  z-index: 1;
 }
 
 .section:not(.has-background):has(.rich-content.hero) {
-  background: linear-gradient(180deg, #000 10.13%, var(--s2a-color-transparent-black-00) 70.89%, var(--s2a-color-transparent-black-00) 75.53%, #000 95%), var(--s2a-color-background-default);
+  background: var(--rc-hero-overlay), var(--s2a-color-background-default);
 }
 
 .section:has(.rich-content.jump-link.hero) {
+  --rc-hero-overlay: linear-gradient(180deg, var(--s2a-color-transparent-black-80) -0.02%, var(--s2a-color-transparent-black-50) 39.99%, var(--s2a-color-transparent-black-20) 59.99%, var(--s2a-color-transparent-black-80) 100%);
   height: 584px;
 }
 
-.section.has-background:has(.rich-content.jump-link.hero) .section-background picture::after {
-  background: linear-gradient(180deg, var(--s2a-color-transparent-black-80) -0.02%, var(--s2a-color-transparent-black-50) 39.99%, var(--s2a-color-transparent-black-20) 59.99%, var(--s2a-color-transparent-black-80) 100%);
-}
-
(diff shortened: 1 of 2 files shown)


zagi25 (approved):

mokimo (commented):
## Ultrareview — PR #6266

**Size:** +33/-12 across 2 files. Base: `site-redesign-foundation` (extension of #6239). Not oversize.

### Blockers
None.

### Suggestions
- `rich-content.css` — the old `.section:not(.has-background):has(.rich-content.jump-link.hero)` rule is deleted outright rather than folded into `--rc-hero-overlay`. Before this PR, a jump-link hero section **without** a background image got its own gradient (0%/39%/74%/100%, fully transparent mid-section). After this PR that case falls back to `.section:has(.rich-content.jump-link.hero)`'s `--rc-hero-overlay` (-0.02%/39.99%/59.99%/100%, using `-black-48`/`-black-24` — partially opaque instead of transparent in the middle), and at `≥768px` it now also picks up the desktop override that previously only applied to the has-background variant. That's a real visual change, not just a mechanical refactor. Couldn't confirm on the linked test URLs (401, Adobe SSO). Worth a manual check that unifying these two cases is intentional — the PR description only covers the authorable-gradient feature, not this.
- `rich-content.js:114-117` — `hero-overlay-source` detection (no image + non-empty text) repurposes an existing content cell as a raw-CSS-value input with no format validation. Low risk (invalid values just fall back to the `:has()` default), but undocumented for whoever authors these fragments next.

### Nice-to-haves
- The `MutationObserver` added in `init()` is never disconnected. Not a real leak given block lifetime, just noting it.

### Test plan
Test URLs provided (good), but require Adobe SSO — recommend someone with access specifically checks a `jump-link.hero` fragment *without* a background image at mobile + desktop widths to confirm the overlay-unification above is intentional.

mokimo (approved):

overmyheadandbody (approved):

NadiiaSokolova (approved):
Verified. Ready to be merged.
Testing details [MWPW-195995](https://jira.corp.adobe.com/browse/MWPW-195995).

libs/mep/ace1209/rich-content/rich-content.js
```diff
@@ -107,6 +109,12 @@ function decorate(block, root = block) {
   content?.classList.add('content');
   foreground?.classList.add('foreground');
   decorateText(content);
+
+  const bgCell = foreground?.children[1];
+  if (bgCell && !bgCell.querySelector('picture, img') && bgCell.textContent.trim()) {
+    bgCell.classList.add('hero-overlay-source');
+  }
+
```
rgclayton, line 119: Does this need to be scoped to the `hero` variant? Currently this will run on every `rich-content` block.

It looks like any variant whose "foreground" row happens to have a second cell containing plain text (no image) will have that text silently hidden and repurposed as a CSS gradient value, instead of rendering as authored content.

libs/mep/ace1209/rich-content/rich-content.css
```diff
@@ -226,33 +226,33 @@
...
 }
 
-.section.has-background:has(.rich-content.jump-link.hero) .section-background picture::after {
-  background: linear-gradient(180deg, var(--s2a-color-transparent-black-80) -0.02%, var(--s2a-color-transparent-black-50) 39.99%, var(--s2a-color-transparent-black-20) 59.99%, var(--s2a-color-transparent-black-80) 100%);
-}
-
-.section:not(.has-background):has(.rich-content.jump-link.hero) {
-  background: linear-gradient(180deg, var(--s2a-color-transparent-black-80) 0%, var(--s2a-color-transparent-black-00) 39%, var(--s2a-color-transparent-black-00) 74%, var(--s2a-color-transparent-black-80) 100%), var(--s2a-color-background-default);
+.rich-content .hero-overlay-source {
+  display: none;
```
rgclayton, line 250: Same comment as [here](https://github.com/adobecom/milo/pull/6266/changes#r3560602206)

libs/mep/ace1209/rich-content/rich-content.js
```diff
@@ -120,6 +128,19 @@ function decorate(block, root = block) {
   decorateJumpLinks(content, foreground);
 }
 
+function applyHeroOverlay(el) {
+  const section = el.closest('.section');
+  if (!section) return;
+  const source = el.querySelector('.hero-overlay-source');
+  if (source) section.style.setProperty(HERO_OVERLAY_PROP, source.textContent.trim());
+  else section.style.removeProperty(HERO_OVERLAY_PROP);
+}
```
rgclayton, line 139: Nit: is it worth considering validating `source` before `setProperty` is called?

Maybe a light validation (regex-check for `linear-gradient()` before calling `setProperty`, falling back to `removeProperty` on failure.

overmyheadandbody: For the gradient, I am to blame for it not being included, it was part of the original code. My thinking: we're not checking validity of colors anywhere else in our blocks or section metadata logic, thus I don't see a reason why we should start here. Additionally, if the authored value is not valid, the CSS would simply not apply, so it wouldn't lead to any issues. Lastly, even though we're using a linear gradient now, it doesn't mean that will always be the case. RGBA / HSL / eight value HEX syntax / radial gradient / etc. are all valid options, I don't think we should make this so rigid.

libs/mep/ace1209/rich-content/rich-content.js
DKos95: Agreed that it should restricted to hero, thanks for the feedback Ryan!

libs/mep/ace1209/rich-content/rich-content.css
DKos95: Agreed and restricted.

rgclayton (approved):

PR #6272, opened 2026-07-10
[Site redesign] - Enhance the Tour modal to show scroll to the browser window

Deva309:
### Enhance the Tour modal to show scroll to the browser window instead of within the modal content.

Resolves: [MWPW-199390](https://jira.corp.adobe.com/browse/MWPW-199390)

**Test URLs:**
- After: https://tour-modal-scroll--milo--adobecom.aem.page/?martech=off

**QA:**
https://main--da-cc--adobecom.aem.page/cc-shared/fragments/tests/2026/q3/ace1209/fragments/cpro-hub?milolibs=tour-modal-scroll&mep=%2Fcc-shared%2Ffragments%2Ftests%2F2026%2Fq3%2Face1209%2Face1209-cpro-redesign.json--all#legal-tour

### DIFF
--- a/libs/mep/ace1209/tour/tour.css
+++ b/libs/mep/ace1209/tour/tour.css
@@ -4,6 +4,11 @@
   position: relative;
   display: flex;
   flex-direction: column;
+  pointer-events: auto;
+  margin: var(--tour-section-gap) auto;
+  background-color: var(--s2a-color-background-default);
+  border-radius: var(--s2a-border-radius-md);
+  clip-path: inset(0 round var(--s2a-border-radius-md));
 }
 
 .tour-header {
@@ -165,25 +170,20 @@
 /* Tour Modal Styles */
 .dialog-modal:has(.tour) {
   --tour-drag-y: 0px;
-  max-width: 696px;
-  width: 100dvw;
-  margin-top: 20px;
-  border-bottom-left-radius: 0;
-  border-bottom-right-radius: 0;
-  max-height: 100dvh;
-  height: calc(100dvh - 20px);
+  --tour-section-gap: var(--s2a-spacing-20);
+  --tour-modal-max-width: 696px;
+  width: 100svw;
+  max-width: 100svw;
+  height: 100svh;
+  max-height: 100svh;
+  background-color: transparent;
+  margin: 0;
+  border-radius: 0;
   transform: translateY(var(--tour-drag-y));
   will-change: transform;
   & > .fragment {
     max-height: 100%;
-    /* Firefox */
-    scrollbar-width: none;
-    /* IE and old Edge */
-    -ms-overflow-style: none;
-    /* Chrome, Safari, Edge (Chromium), Opera */
-    &::-webkit-scrollbar {
-      display: none;
-    }
+    pointer-events: none;
   }
 
   &.is-dragging {
@@ -223,27 +223,69 @@
     backdrop-filter: blur(var(--s2a-blur-32));
     will-change: backdrop-filter;
   }
+
+  .dialog-close {
+    /* If top spacing changes then update in JS as well */
+    top: calc(var(--tour-section-gap) + var(--s2a-spacing-md));
+    right: calc(50% - calc(var(--tour-modal-max-width) / 2) + var(--s2a-spacing-md));
+  }
 }
 
 [dir='rtl'] {
   .dialog-modal:has(.tour) {
     .dialog-close {
       right: auto;
-      left: var(--s2a-spacing-md);
+      left: calc(50% - calc(var(--tour-modal-max-width) / 2) + var(--s2a-spacing-md));
     }
   }
 }
 
-/* Below 696px the modal is full-width; above it the width is fixed at 696px */
+/* Below (--tour-modal-max-width) the modal is full-width,
+  above it the width is fixed at (--tour-modal-max-width) */
 @media (max-width: 696px) {
   .dialog-modal:has(.tour) {
-    max-width: 100dvw;
+    max-width: 100svw;
+    .dialog-close {
+      right: var(--s2a-spacing-md);
+    }
   }
+
   .tour-row-image-center {
     img {
       max-height: 110px;
       width: auto;
     }
+  }
+
+  [dir='rtl'] {
+    .dialog-modal:has(.tour) {
+      .dialog-close {
+        right: auto;
+        left: var(--s2a-spacing-md);
+      }
+    }
+  }
+}
+
+@supports (timeline-scope: none) {
+  .dialog-modal:has(.tour) {
+    timeline-scope: --tour-scroll;
+  }
+
+  .dialog-modal:has(.tour) > .fragment {
+    scroll-timeline-name: --tour-scroll;
+    scroll-timeline-axis: block;
+  }
+
+  .dialog-modal:has(.tour) > .dialog-close {
+    animation: tour-close-track linear both;
+    animation-timeline: --tour-scroll;
+    animation-range: 0px var(--tour-section-gap);
+  }
+
+  @keyframes tour-close-track {
+    from { top: calc(var(--tour-section-gap) + var(--s2a-spacing-md)); }
+    to { top: var(--s2a-spacing-md); }
   }
 }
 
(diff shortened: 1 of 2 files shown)


libs/mep/ace1209/tour/tour.css
```diff
@@ -223,28 +223,70 @@
...
+
+  [dir='rtl'] {
+    .dialog-modal:has(.tour) {
+      .dialog-close {
+        right: auto;
+        left: var(--s2a-spacing-md);
+      }
+    }
+  }
+}
```
overmyheadandbody, line 267: Instead of re-declaring for RTL, you could use `inset-inline-end`

libs/mep/ace1209/tour/tour.js
```diff
@@ -3,6 +3,39 @@ import icons from '../../../c2/assets/icons.js';
...
+  const modal = el.closest('.dialog-modal');
+  if (!modal) return;
+
+  requestAnimationFrame(() => {
+    const fragment = modal.querySelector('.fragment');
+    const closeBtn = modal.querySelector('.dialog-close');
+    if (!fragment || !closeBtn) return;
+
+    const style = getComputedStyle(modal);
+    const cardGap = parseFloat(style.getPropertyValue('--tour-card-gap'));
```
overmyheadandbody: `--tour-card-gap` doesn't seem to exist, is this a leftover from some previous naming pattern?

libs/mep/ace1209/tour/tour.css
```diff
@@ -165,25 +170,20 @@
...
   --tour-drag-y: 0px;
-  max-width: 696px;
-  width: 100dvw;
-  margin-top: 20px;
-  border-bottom-left-radius: 0;
-  border-bottom-right-radius: 0;
-  max-height: 100dvh;
-  height: calc(100dvh - 20px);
+  --tour-section-gap: var(--s2a-spacing-20);
+  --tour-modal-max-width: 696px;
```
overmyheadandbody, line 173: Could this and https://github.com/adobecom/milo/blob/05ea91ee91eaa53cebdcc29f3f2f4ac09c5b3a7f/libs/mep/ace1209/tour/tour.css#L3 be combined, since it's the same value?

libs/mep/ace1209/tour/tour.css
```diff
@@ -223,28 +223,70 @@
     backdrop-filter: blur(var(--s2a-blur-32));
     will-change: backdrop-filter;
   }
+
+  .dialog-close {
```
overmyheadandbody: There are already declarations for this at https://github.com/adobecom/milo/blob/05ea91ee91eaa53cebdcc29f3f2f4ac09c5b3a7f/libs/mep/ace1209/tour/tour.css#L197, they should be combined to avoid confusion

overmyheadandbody (commented):
Looking good functionality-wise, but I have a few notes to be looked into

libs/mep/ace1209/tour/tour.js
```diff
@@ -3,6 +3,39 @@ import icons from '../../../c2/assets/icons.js';
 
 const SWIPE_CLOSE_THRESHOLD = 100;
 
+function addScrollCloseSync(el) {
+  if (CSS.supports('timeline-scope', 'none')) return;
+  const modal = el.closest('.dialog-modal');
+  if (!modal) return;
+
+  requestAnimationFrame(() => {
+    const fragment = modal.querySelector('.fragment');
```
zagi25, line 13: Will modal always contain fragment, maybe it is better to query for `tour` ?

libs/mep/ace1209/tour/tour.js
```diff
@@ -3,6 +3,39 @@ import icons from '../../../c2/assets/icons.js';
...
+  });
+}
+
+function addOutsideClickClose(el) {
+  const modal = el.closest('.dialog-modal');
+  if (!modal) return;
+
+  modal.addEventListener('click', (e) => {
+    if (!el.contains(e.target) && !e.target.closest('.dialog-close')) {
+      modal.querySelector('.dialog-close')?.click();
```
zagi25: Instead of clicking element via JS, can we use this method https://github.com/adobecom/milo/blob/stage/libs/c2/blocks/modal/modal.js#L73 ?

libs/mep/ace1209/tour/tour.js
```diff
@@ -3,6 +3,39 @@ import icons from '../../../c2/assets/icons.js';
...
+  if (!modal) return;
+
+  requestAnimationFrame(() => {
+    const fragment = modal.querySelector('.fragment');
+    const closeBtn = modal.querySelector('.dialog-close');
+    if (!fragment || !closeBtn) return;
+
+    const style = getComputedStyle(modal);
+    const cardGap = parseFloat(style.getPropertyValue('--tour-card-gap'));
+    const spacingMd = parseFloat(style.getPropertyValue('--s2a-spacing-md'));
```
zagi25: This spacing can be changed in the future, can we check for element padding/margin (whatever `--s2a-spacing-md` in this case represents) ?

libs/mep/ace1209/tour/tour.css
Deva309: @overmyheadandbody 
Here overriding styles coming from `modal.css`. So, overridden in the same way only.

libs/mep/ace1209/tour/tour.js
Deva309: Forgot to update this to `--tour-section-gap` already defined within `tour.css`.

libs/mep/ace1209/tour/tour.css
Deva309: Done

libs/mep/ace1209/tour/tour.css
Deva309: Done

libs/mep/ace1209/tour/tour.js
Deva309: yes modal always have fragment element.

libs/mep/ace1209/tour/tour.js
Deva309: Done

libs/mep/ace1209/tour/tour.js
Deva309: Done

libs/mep/ace1209/tour/tour.css
overmyheadandbody: Even though this works in a sort of "hoisting" fashion, it might be best to declare the variable before it's first used on L3. Not a blocker, but it's a pattern we use in our styles.

overmyheadandbody (approved):

zagi25 (approved):

libs/mep/ace1209/tour/tour.css
Deva309: Fixed 👍🏻

rgclayton (approved):

libs/mep/ace1209/tour/tour.js
```diff
@@ -1,8 +1,41 @@
...
+  if (!modal) return;
+
+  requestAnimationFrame(() => {
+    const fragment = modal.querySelector('.fragment');
+    const closeBtn = modal.querySelector('.dialog-close');
+    if (!fragment || !closeBtn) return;
+
+    const style = getComputedStyle(el);
+    const sectionGap = parseFloat(style.getPropertyValue('--tour-section-gap'));
+    const initialTop = parseFloat(getComputedStyle(closeBtn).top);
```
rgclayton, line 19: Nit:
Either value could possibly yield `NaN` if `getComputedStyle` returns "auto" in some edge case.
Maybe adding a condition to make a failure more explicit.

`if (Number.isNaN(sectionGap) || Number.isNaN(initialTop)) return;)`

Deva309: Fixed👍🏻

NadiiaSokolova (approved):
Verified. Ready to be merged.
Testing details [MWPW-199390](https://jira.corp.adobe.com/browse/MWPW-199390)

PR #6294, opened 2026-07-14
MWPW-200482 - Remove JIRA Epic field from .kodiak config

tusc-adbe:
Removing JIRA Epic field from .kodiak/config.yaml


thedoc31 (approved):
Approved via MWPW-200482 batch review

PR #6295, opened 2026-07-14
[MWPW-199341] Reduce motion effect when applied not seen working on carousel

SamSawant876:
## Changes:

- Replaced the index-based isSlideVisible() check with getOverlapRatio(), which compares a video's getBoundingClientRect() against the carousel wrapper's clipped (overflow: hidden) bounds.
 - Added VISIBILITY_THRESHOLD (0.5) — a slide's video only stays exempt from pausing if at least half of it is still visible in the wrapper, so a hinting layout's partial "peek" of a neighboring slide doesn't count as fully in view.
 - checkSlideForVideo() / pauseVideosOutOfView() now run against every slide instead of only the previously-active one, so a video started on a peeking/inactive slide is no longer invisible to the pause check.
 - Added waitForSlideTransition() so the visibility check runs once the slide track's transform transition actually settles (with a TRANSITION_FALLBACK_MS timeout as a safety net), guarded to the track's own transform property so a child slide's own opacity transition can't resolve it early.
 - Added test coverage for the visibility-ratio thresholds and the boundary-path pause call.

## Expected video behaviour:

Reduce motion on: Videos do not auto-play on arrow navigation, user must click play.
User clicks play: Video plays while its slide remains sufficiently visible in the carousel viewport.
User swipes / navigates away (either direction): Video's slide is no longer sufficiently visible and is paused.
Navigation blocked at a disable-circular-nav boundary: A playing video that's actually out of view still gets paused, even though the active slide didn't change.
Slide returns to view: Video stays paused, it does not auto-resume. User must click play again.

## Test URLs:

Before: https://main--da-cc--adobecom.aem.live/products/firefly/features/image-to-video
After: https://main--da-cc--adobecom.aem.live/products/firefly/features/image-to-video?milolibs=mwpw-199341-firefly-carousal-fix

### DIFF
--- a/libs/blocks/carousel/carousel.js
+++ b/libs/blocks/carousel/carousel.js
@@ -306,14 +306,63 @@ function handleLightboxButtons(lightboxBtns, el, slideWrapper) {
   });
 }
 
-function checkSlideForVideo(activeSlide) {
-  const video = activeSlide.querySelector('video');
-  /* c8 ignore start */
-  if (video?.played.length > 0 && !video?.paused) {
-    video.pause();
-    syncPausePlayIcon(video);
-  }
-  /* c8 ignore end */
+// Hinting layouts intentionally show a peek of the neighboring slide, so a
+// slide can still be clipping the wrapper's edge by a meaningful amount even
+// once it's no longer the one you care about. Require majority visibility
+// rather than any nonzero overlap before treating it as still "in view".
+const VISIBILITY_THRESHOLD = 0.5;
+
+function getOverlapRatio(rect, wrapperRect) {
+  const overlapWidth = Math.max(0, Math.min(rect.right, wrapperRect.right)
+    - Math.max(rect.left, wrapperRect.left));
+  const overlapHeight = Math.max(0, Math.min(rect.bottom, wrapperRect.bottom)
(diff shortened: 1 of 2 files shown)


libs/blocks/carousel/carousel.js
```diff
@@ -306,14 +306,63 @@ function handleLightboxButtons(lightboxBtns, el, slideWrapper) {
...
+// Waits for the slide track's own transform transition to finish (with a
+// timed fallback, since edge cases may never fire a matching transitionend)
+// before running the visibility check, since positions aren't final until
+// the transition settles. Guarded to the track's own transform so a child
+// slide's opacity transition (carousel.css .carousel-slide) can't resolve
+// this early via event bubbling.
+// Margin above the CSS transition (25ms reflow delay + .6s transform,
+// see carousel.css .carousel-slides.is-ready) in case transitionend
+// never fires for this move.
+const TRANSITION_FALLBACK_MS = 700;
```
aishwaryamathuria: Can you remove huge comments and if required compress them?

aishwaryamathuria (approved):

suhjainadobe (approved):

spadmasa:
PR is verified and details are added in ticket

PR #6296, opened 2026-07-14
MWPW-200568 Parallelize initService script loading in merch block

TsayAdobe:
* initService in merch.js loaded four independent resources sequentially, creating an unnecessary waterfall on every page that uses the merch block.
* Replaced sequential await chain with Promise.all([loadMasComponent, fragmentClient, getLocaleSettings, getValidatedMarket])

Resolves: [MWPW-200568](https://jira.corp.adobe.com/browse/MWPW-200568)

**Test URLs:**
- After: https://MWPW-200568--milo--adobecom.aem.page/?martech=off

Change on https://www.adobe.com/fr/creativecloud.html with content overrides
Before:
BeforeBR

After
AfterPR

### DIFF
--- a/libs/blocks/merch/merch.js
+++ b/libs/blocks/merch/merch.js
@@ -1116,27 +1116,25 @@ export async function initService(force = false, attributes = {}) {
   });
   initService.promise = initService.promise
     ?? polyfills().then(async () => {
-      await loadMasComponent(COMMERCE_LIBRARY);
-
-      // Load fragment-client.js when maslibs is present
-      const fragmentClientUrl = getFragmentClientUrl();
-      if (fragmentClientUrl) {
-        const { loadScript: loadScriptUtil } = await import('../../utils/utils.js');
-        try {
-          await loadScriptUtil(fragmentClientUrl, 'module');
-        } catch (e) {
-          log?.error('Failed to load fragment-client.js:', e);
-        }
-      }
-
-      const { language, locale, country } = await getLocaleSettings(miloLocale);
       const useGeoMarket = isMasGeoDetectionEnabled();
+
+      // Load all independent resources in parallel
+      const fragmentClientUrl = getFragmentClientUrl();
+      const [, , { language, locale, country }, validatedMarket] = await Promise.all([
+        loadMasComponent(COMMERCE_LIBRARY),
+        fragmentClientUrl
+          ? loadScript(fragmentClientUrl, 'module').catch((e) => {
+            log?.error('Failed to load fragment-client.js:', e);
+          })
+          : Promise.resolve(),
+        getLocaleSettings(miloLocale),
+        useGeoMarket
+          ? import('../../utils/market.js').then(({ getValidatedMarket }) => getValidatedMarket())
+          : Promise.resolve(null),
+      ]);
+
       let countryFromMarket = country;
-      if (useGeoMarket) {
-        const { getValidatedMarket } = await import('../../utils/market.js');
-        const validatedMarket = await getValidatedMarket();
-        if (validatedMarket) countryFromMarket = validatedMarket.toUpperCase();
-      }
+      if (useGeoMarket && validatedMarket) countryFromMarket = validatedMarket.toUpperCase();
       let service = document.head.querySelector('mas-commerce-service');
       if (!service) {
         setPreview(attributes);
--- a/libs/blocks/merch/merch.js
+++ b/libs/blocks/merch/merch.js
@@ -1119,6 +1119,7 @@ export async function initService(force = false, attributes = {}) {
 
       // Load all independent resources in parallel
       const fragmentClientUrl = getFragmentClientUrl();
+      const localeSettingsPromise = getLocaleSettings(miloLocale);
       const [, , { language, locale, country }, validatedMarket] = await Promise.all([
         loadMasComponent(COMMERCE_LIBRARY),
         fragmentClientUrl
@@ -1126,7 +1127,7 @@ export async function initService(force = false, attributes = {}) {
             log?.error('Failed to load fragment-client.js:', e);
           })
           : Promise.resolve(),
-        getLocaleSettings(miloLocale),
+        localeSettingsPromise,
         useGeoMarket
           ? import('../../utils/market.js').then(({ getValidatedMarket }) => getValidatedMarket())
           : Promise.resolve(null),
--- a/libs/blocks/merch/merch.js
+++ b/libs/blocks/merch/merch.js
@@ -3,7 +3,6 @@ import {
   shouldAllowKrTrial, getCountry,
 } from '../../utils/utils.js';
 import { replaceKey } from '../../features/placeholders.js';
-import { mepMasStudioUrls } from './mas-mep-utils.js';
 
 // MAS Component Names
 export const COMMERCE_LIBRARY = 'commerce';
@@ -1621,6 +1620,7 @@ export default async function init(el) {
     // Rebuilt merch href keeps only the OSI; stash the original for
     // the "Edit OSI" badge.
     if (getConfig()?.mep?.preview) {
+      const { mepMasStudioUrls } = await import('./mas-mep-utils.js');
       mepMasStudioUrls.set(merch, el.href);
       merch.dataset.masBlock = 'ost';
     }
--- a/libs/blocks/merch-card-autoblock/merch-card-autoblock.js
+++ b/libs/blocks/merch-card-autoblock/merch-card-autoblock.js
@@ -1,7 +1,6 @@
 import { createTag, getConfig } from '../../utils/utils.js';
 import { decorateButtons, getBlockSize } from '../../utils/decorate.js';
 import { postProcessAutoblock } from '../merch/autoblock.js';
-import { mepMasStudioUrls } from '../merch/mas-mep-utils.js';
 import {
   initService,
   createAemFragment,
@@ -152,6 +151,7 @@ export async function createCard(el, options) {
   const merchCard = createTag('merch-card', { consonant: '' }, aemFragment);
   // For the "Edit Card" mep preview badge.
   if (getConfig()?.mep?.preview) {
+    const { mepMasStudioUrls } = await import('../merch/mas-mep-utils.js');
     mepMasStudioUrls.set(merchCard, el.href);
     merchCard.dataset.masBlock = 'card';
   }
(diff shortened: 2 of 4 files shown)


vhargrave:
@TsayAdobe thank you very much for this contribution 🚀 
Is there nothing else that can be done to speed up M@S ? This seems like it will only help with like 200-300 milliseconds whereas the performance hit is around 2 seconds when it loads in. 
This is a major hit for lingo right now and if I'm interpreting this right and we're only getting minor savings then there will be more discussions on if we can roll out M@S for the English pages as well as part of the major August launch
cc: @Blainegunn @sukamat

TsayAdobe:
@vhargrave I also converted `mas-mep-utils.js` from a static to a dynamic import -- it was blocking merch.js from executing despite only being needed in MEP preview environments. Further gains beyond this would require architectural changes (e.g., server-side prerender to emit preload hints and collapse the discovery waterfall).

libs/blocks/merch/merch.js
```diff
@@ -1116,27 +1115,25 @@ export async function initService(force = false, attributes = {}) {
...
       const fragmentClientUrl = getFragmentClientUrl();
-      if (fragmentClientUrl) {
-        const { loadScript: loadScriptUtil } = await import('../../utils/utils.js');
-        try {
-          await loadScriptUtil(fragmentClientUrl, 'module');
-        } catch (e) {
-          log?.error('Failed to load fragment-client.js:', e);
-        }
-      }
+      const [, , { language, locale, country }, validatedMarket] = await Promise.all([
```
Axelcureno, line 1123: These two look independent but aren't. Old order:

```
await getLocaleSettings()  → getCountry() → getAkamaiCode() → writes sessionStorage.akamai
await getValidatedMarket() → getCountry() → reads it ✅
```

The sequential await was doing load-bearing work — priming the cache before the second read. In parallel, both call `getCountry()` before either has written it. `getAkamaiCode` (`geo.js:7`) has no in-flight dedup and is `cache: 'no-cache'`, so nothing collapses them.

I ran this locally against the branch — sequential gives 1 `geo2.adobe.com` fetch, parallel gives 2. So: two concurrent geo fetches per geo-enabled page load. Related risk I did *not* reproduce, but the code path is there — `getValidatedMarket()` resolving on a null country falls back to `currLang.defaultMarket || 'us'` (`market.js:38`), which would be a wrong-country price render on a slow or failing geo call.

Keeps most of the win — chain just the market call off the settings that prime the cache:

```js
const localeSettingsPromise = getLocaleSettings(miloLocale);
const [, , { language, locale, country }, validatedMarket] = await Promise.all([
  loadMasComponent(COMMERCE_LIBRARY),
  fragmentClientUrl ? loadScript(fragmentClientUrl, 'module').catch(...) : Promise.resolve(),
  localeSettingsPromise,
  useGeoMarket
    ? localeSettingsPromise.then(() => import('../../utils/market.js')
        .then(({ getValidatedMarket }) => getValidatedMarket()))
    : Promise.resolve(null),
]);
```

`loadMasComponent` / `loadScript` / `getLocaleSettings` are genuinely independent, so that's where the real gain was anyway. Alternative: memoize the in-flight promise in `getAkamaiCode` — fixes the double-fetch for every caller, since `getCountry` and `resolveDetectedMarketCountry` both funnel through it.

Axelcureno (changes requested):
Nice win on the waterfall — the `loadScript` swap and the mep-utils dynamic import both look right.

One thing to fix before merge: with geo detection on, `getLocaleSettings` and `getValidatedMarket` aren't actually independent. The sequential await was priming the akamai cache for the second call. In `Promise.all` they both hit `geo2.adobe.com` — I reproduced it locally, **1 fetch on stage vs 2 on this branch**. Details inline.

### On tests

There are none here, and the existing ones can't catch this. Every geo test either pre-seeds `sessionStorage.akamai` (`merch.test.js:285`, `market.test.js:172/217/235/273`) or stubs fetch to reject — so `getCountry` short-circuits at `utils.js:952` and never hits the network. The suite only exercises the cache-hit path, which is exactly what the sequential await guaranteed and this PR removes.

Something like this fails on the branch, passes on stage:

```js
it('does not double-fetch geo when geo detection is on', async () => {
  sessionStorage.removeItem('akamai');  // force the network path
  const fetchStub = sinon.stub(window, 'fetch');
  fetchStub.withArgs(sinon.match(/geo2\.adobe\.com/)).resolves({
    ok: true, json: () => Promise.resolve({ country: 'DE' }),
  });
  // set mas-geo-detection meta to 'on'
  await initService(true);
  expect(fetchStub.withArgs(sinon.match(/geo2\.adobe\.com/)).callCount).to.equal(1);
});
```

Also worth a `countryFromMarket` assertion with the cache cold, and a fragment-client test that `initService` still resolves when `loadScript` rejects — that `.catch()` is load-bearing now (without it one failed script rejects the whole `Promise.all`) and has no coverage today.

Axelcureno:
@vhargrave you're right that this PR alone won't move the needle much — so we dug into where the ~2s actually goes. We traced the full load path and measured plans.html live. TL;DR: **the data fetches are fast (40–140ms each); the time is serialized orchestration.** On the measured page, the first merch data request can't fire until ~2.9s, then takes 60ms.

**Where the time goes** (measured, warm cache):
- 0→1.4s: Milo bootstrap incl. personalization gate (`checkForPageMods` blocks all sections)
- 1.4→2.9s: sequential section loop + block→service→bundle→element chain — commerce service is ready at 1.9s, but the first fragment fetch waits until 2.9s on section progression alone
- 2.9→4.3s: fragment fetch → per-OSI price fetch, strictly serial (and price batching was lost in the Milo→MAS migration — one request per price element today)

**What MAS team is fixing** (tickets filed):
- Restore WCS price batching ([MWPW-200883](https://jira.corp.adobe.com/browse/MWPW-200883))
- Bundle refactor — commerce engine currently ships twice, all 24 variants always ship; ~148KB → ~80KB gz ([MWPW-200884](https://jira.corp.adobe.com/browse/MWPW-200884))
- HTTP caching for fragment/price fetches ([MWPW-200885](https://jira.corp.adobe.com/browse/MWPW-200885))
- Fragment prefetch API ([MWPW-200886](https://jira.corp.adobe.com/browse/MWPW-200886))
- Detach card readiness from section flow in `merch-card-autoblock` (ours) — today a slow card holds every later section at `display:none` for up to 5s; after the fix, sections below merch reveal ~1–2.5s earlier. Would appreciate Milo review since it touches the reveal model.

**Two small asks of Milo core** (block-level code runs too late for both):
1. Early `preconnect`/`modulepreload` for MAS origins at `loadArea` time when the page has merch links — or bless the metadata-preload pattern plans.html already uses by hand.
2. A post-MEP, pre-section hook so we can prefetch fragment data (with correct MEP-overridden IDs) at ~1.4s instead of ~2.9s. This is the single biggest lever we found (~1.3–1.5s).

**Projection:** MAS fixes alone get cards to ~3s; with the two hooks, **~1.8–2.2s — within ~0.5s of the personalization floor**, which is the practical best for any block on a personalized page. Longer-term ideas (parallel section loading with ordered reveal, scoped personalization) would help every block, not just merch — happy to share the full RCA and measurements if there's interest.

cc @Blainegunn @sukamat @TsayAdobe @reil-adobe @3ch023 @npeltier

TsayAdobe:
I added static hints for `merch.js` and `merch-card-autoblock.js` to preload MAS libs.

https://www.adobe.com/fr/creativecloud.html
Before
creativecloud_before
After
creativecloud_after

https://www.adobe.com/fr/products/photosho.html
Before
photoshop_before
After
photoshop_after

cc @Blainegunn @sukamat @vhargrave  @Axelcureno  @reil-adobe @3ch023 @npeltier

vhargrave:
> @vhargrave you're right that this PR alone won't move the needle much — so we dug into where the ~2s actually goes. We traced the full load path and measured plans.html live. TL;DR: **the data fetches are fast (40–140ms each); the time is serialized orchestration.** On the measured page, the first merch data request can't fire until ~2.9s, then takes 60ms.
> 
> **Where the time goes** (measured, warm cache):
> 
> * 0→1.4s: Milo bootstrap incl. personalization gate (`checkForPageMods` blocks all sections)
> * 1.4→2.9s: sequential section loop + block→service→bundle→element chain — commerce service is ready at 1.9s, but the first fragment fetch waits until 2.9s on section progression alone
> * 2.9→4.3s: fragment fetch → per-OSI price fetch, strictly serial (and price batching was lost in the Milo→MAS migration — one request per price element today)
> 
> **What MAS team is fixing** (tickets filed):
> 
> * Restore WCS price batching ([MWPW-200883](https://jira.corp.adobe.com/browse/MWPW-200883))
> * Bundle refactor — commerce engine currently ships twice, all 24 variants always ship; ~148KB → ~80KB gz ([MWPW-200884](https://jira.corp.adobe.com/browse/MWPW-200884))
> * HTTP caching for fragment/price fetches ([MWPW-200885](https://jira.corp.adobe.com/browse/MWPW-200885))
> * Fragment prefetch API ([MWPW-200886](https://jira.corp.adobe.com/browse/MWPW-200886))
> * Detach card readiness from section flow in `merch-card-autoblock` (ours) — today a slow card holds every later section at `display:none` for up to 5s; after the fix, sections below merch reveal ~1–2.5s earlier. Would appreciate Milo review since it touches the reveal model.
> 
> **Two small asks of Milo core** (block-level code runs too late for both):
> 
> 1. Early `preconnect`/`modulepreload` for MAS origins at `loadArea` time when the page has merch links — or bless the metadata-preload pattern plans.html already uses by hand.
> 2. A post-MEP, pre-section hook so we can prefetch fragment data (with correct MEP-overridden IDs) at ~1.4s instead of ~2.9s. This is the single biggest lever we found (~1.3–1.5s).
> 
> **Projection:** MAS fixes alone get cards to ~3s; with the two hooks, **~1.8–2.2s — within ~0.5s of the personalization floor**, which is the practical best for any block on a personalized page. Longer-term ideas (parallel section loading with ordered reveal, scoped personalization) would help every block, not just merch — happy to share the full RCA and measurements if there's interest.
> 
> cc @Blainegunn @sukamat @TsayAdobe @reil-adobe @3ch023 @npeltier

Hi @Axelcureno - first off thank you, appreciate all the difference attack vectors and , yes we're definitely interested in your proposal of preconnecting MAS origins - especially if there are merch elements within LCP. Let's have a chat so we can see about getting that added.

Roycethan (approved):

joaquinrivero (approved):

Axelcureno (approved):
LGTM — thanks!

PR #6302, opened 2026-07-15
MWPW-195996: Flex card configurations updates

cmiqueo:
## Summary
- Replaces `flexCardHideFooter` option with `flexCardShowDateOnFooter` to control date visibility on the flex card footer
- Adds `hidden` and `productName` options to `detailsTextOption`, and a `default` option to `flexCardTextAlign`
- Adds `getProducts` helper to fetch CaaS product tags when `detailsTextOption` is set to `productName`, passed through `getConfig` as `products`

## Jira Ticket
Resolves: [MWPW-195996](https://jira.corp.adobe.com/browse/MWPW-195996)

## Changes
- `libs/blocks/caas-config/caas-config.js` — updated `flexCardTextAlign`/`detailsTextOption` option labels; replaced "Hide Footer (CTA)" checkbox with "Show Date on Footer"
- `libs/blocks/caas/utils.js` — added `getProducts`, wired `products` into `getConfig` output, renamed `flexCardHideFooter` → `flexCardShowDateOnFooter` in config and default state
- `test/blocks/caas-config/expectedConfigs/defaultConfig.js`, `test/blocks/caas/utils.test.js` — updated expected configs to include `products: {}`

## Test plan
- [x] Verify CaaS Configurator UI shows "Show Date on Footer" checkbox and new text-align/details-text options
- [x] Verify flex card footer date renders correctly when the new checkbox is enabled
- [x] Verify `detailsTextOption: 'productName'` fetches and displays product names on cards
- [x] Unit tests added/updated

**Test URLs**
After: [https://MWPW-195996--milo--adobecom.aem.page/?martech=off](https://www.google.com/search?q=https://MWPW-195996--milo--adobecom.aem.page/%3Fmartech%3Doff)


sanrai (approved):
Approving. A couple of things to double-check:

**Double-check these**

- `defaultState` is missing `flexCardShowDateOnFooter: false`. Every sibling flex prop has an explicit default; this one currently relies on `undefined` being falsy. Adding it keeps things consistent and safe if the include-guard logic ever changes.
- The old `flexCardHideFooter: false` is still in `defaultState` but nothing reads it anymore (grepped the repo, it's dead after this rename). Let's remove it as part of the change.

**Tests**

- The test changes only assert `products: {}` on the default configs. Could we add coverage for the new logic itself: the `detailsTextOption === 'productName'` branch calling `getProducts` and populating `products`, and `flexCardShowDateOnFooter: true` producing the `flexCard` block with `showDateOnFooter: true`? The renderer side is tested in `caas`, but milo's own `getProducts`/conditional-fetch path is currently untested.

sheridansunier:
**Concern: `flexCardTextAlign` default value inconsistency**

There's a three-way mismatch introduced by adding `default: 'Default'` to the `flexCardTextAlign` options:

1. **`defaultState.flexCardTextAlign` is still `'text-left'`** — so every fresh config opens with 'Left' pre-selected in the dropdown instead of 'Default'.

2. **The guard in `utils.js` still checks `!== 'text-left'`** — so selecting the new 'Default' option (value `'default'`) makes `'default' !== 'text-left'` = true, fires the guard, and emits `flexCard.textAlign: 'default'` into the output config. The renderer has no CSS for that value.

3. **Selecting 'Left' is a silent no-op** — since `'text-left'` is still the guard's noop sentinel, explicitly choosing 'Left' produces no output and the renderer falls back to its own default.

Suggested fix — two options:
- **(a) Keep `'text-left'` as the true default**: don't add a new `'default'` key; just relabel `'text-left'` back to 'Default' or leave it as 'Left' and drop the new entry.
- **(b) Migrate to `'default'` as the noop**: update `defaultState.flexCardTextAlign` to `'default'` AND change the guard to `!== 'default'`, and ensure the renderer treats `'default'` as a no-op (or omit the key entirely when it equals `'default'`).

sheridansunier (approved):
conditional approval: you need to address the unit test failures :P

PR #6316, opened 2026-07-17
[MWPW-199159] Lingo web performance: re-apply #6210 with iconography.css fix

vhargrave:
Re-applies the reverted Lingo web-performance PR #6210, with the iconography regression that caused the revert fixed at the root cause.

## Background
#6210 (_perf: improve lingo performance — personalization, geo, lingo, XLG_) was merged to `stage` and reverted ~3h later by #6267 because `iconography.css` stopped being applied — icons broke on almost all stage pages ([milo-dev thread](https://adobe-mwp.slack.com/archives/C03HAU5SU2E/p1783598532069229)). This PR restores #6210 and fixes that regression properly.

## Root cause of the icon regression
`loadLink()` (`libs/utils/utils.js`) deduped existing links by `href` **alone**, ignoring `rel`. #6210 added `rel=preload, as=style` hints for `iconography.css` and `breakpoint-theme.css` in `preloadBlockResources()` (for `marquee`/`hero-marquee`). Those stylesheets are applied later via `loadStyle()` (hero-marquee, notification, editorial-card, aside). Because a `<link>` with that href already existed (the preload), the later `loadStyle()` → `loadLink()` found it and no-oped instead of creating the applying `rel=stylesheet` link — so the CSS downloaded but was never applied.

## Changes
<details>
<summary>more details</summary>

- **Re-apply #6210 in full**, including the two CSS preloads — clean revert-of-the-revert onto current `stage`. All personalization / geo / lingo / XLG optimizations preserved unchanged.
- **fix (root cause): make `loadLink()` rel-aware.** A `rel=stylesheet` request now dedups only against an existing **stylesheet** link, so a preload no longer shadows the `loadStyle()`. The block's `loadStyle()` creates a fresh `<link rel=stylesheet>` (fires `load` reliably in every engine — no rel-swap, no dependency on preload `load` semantics) that reuses the preloaded bytes via HTTP cache — the canonical `preload → stylesheet` pattern ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/preload)). Non-stylesheet `rel` requests keep href-only dedup, unchanged. This keeps Okan's non-render-blocking early fetch *and* applies the CSS — rather than discarding the optimization.
- **perf:** priority-aware Lingo query-index loading. Wave 1 (primary + base) and any cross-site index flagged `fetchPriority` in `lingo-site-mapping.json` (today only `upp`) are fetched first; the remaining cross-site indexes wait behind a barrier and then fetch at `{ priority: 'low' }` so they don't compete during the LCP window. `lingo-site-mapping.json` itself fetches at default priority so the flagged indexes are discovered promptly.

</details>

## Design note (why the root-cause fix, not a rel-swap)
<details>
<summary>more details</summary>

An alternative — swap the preload link's `rel` to `stylesheet` on load — was designed and rejected after a devil's-advocate review: it depends on `rel=preload as=style` firing `load` in every engine (unverified on Safari/WebKit; our suite is Chromium-only), and if it doesn't, the awaited `loadStyle` callback can never fire → block decoration **hangs** (worse than the bug). The rel-aware-dedup fix has no such dependency: worst case is a benign double-fetch (perf, not correctness); icons always apply. Cross-repo audit: no adobecom consumer preloads a CSS URL then `loadStyle`s it, so the shared-fn change is behavior-preserving for every non-broken caller.

</details>

## Verification
- **Tests:** full `test/**` run green (3702 passed at last full run; the sole failure is a **pre-existing floating-promise flake** — `fetchAndProcessPlainHtml with MEP`, `utilities.test.js:42`, asserts inside a non-awaited `.then()`, proven non-deterministic by isolation and outside the CSS/`loadLink` path). New/added coverage for every risk-bearing change: the `loadLink` preload-dedup fix (a `loadStyle` for a preloaded href creates a distinct applying stylesheet, not shadowed), the `fetchPriority: "yes"` query-index branch, `geo.js` `getAkamaiCode` (concurrent-caller dedup, post-failure re-fetch, 5s abort→reject), and `categorizeActions` last-write-wins ordering under `Promise.all`. Lint clean.
- **Visual / DOM** — da-cc `/creativecloud`, `?milolibs=main` vs `?milolibs=MWPW-199159`, confirmed on this build: the branch shows `iconography.css` and `breakpoint-theme.css` each as **two links — a `rel=preload as=style` (not applied) plus an applied `rel=stylesheet` (`link.sheet` non-null)** — the canonical preload→stylesheet pair; main has the single applied stylesheet. **0 unstyled icons on both** (9 visible icons each). Confirms the preload optimization is preserved *and* the CSS applies.
- Did a test on https://main--da-cc--adobecom.aem.live/fr/creativecloud to ensure the query indexes still loaded , but in a different order. Looked great. 

Resolves: [MWPW-199159](https://jira.corp.adobe.com/browse/MWPW-199159)

**Test URLs:**
- After: https://MWPW-199159--milo--adobecom.aem.page/?martech=off

**Icon / perf comparison (consumer page):**
- Before: https://main--da-cc--adobecom.aem.live/creativecloud?milolibs=main
- After: https://main--da-cc--adobecom.aem.live/creativecloud?milolibs=MWPW-199159

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -2681,8 +2681,10 @@ const preloadBlockResources = (blocks = []) => blocks.map((block) => {
   if (['marquee', 'hero-marquee'].includes(name)) {
     const { base } = getConfig();
     loadLink(`${base}/utils/decorate.js`, { rel: 'preload', as: 'script', crossorigin: 'anonymous' });
-    loadLink(`${base}/styles/iconography.css`, { rel: 'preload', as: 'style' });
-    loadLink(`${base}/styles/breakpoint-theme.css`, { rel: 'preload', as: 'style' });
+    // Do NOT rel=preload iconography.css / breakpoint-theme.css here. They are
+    // applied later via loadStyle(), and loadLink() dedupes by href regardless
+    // of rel — a preload link shadows that loadStyle() so the stylesheet is
+    // fetched but never applied, breaking icons. (Caused the #6210 → #6267 revert.)
   }
   loadLink(`${blockPath}.js`, { rel: 'preload', as: 'script', crossorigin: 'anonymous' });
   return hasStyles && new Promise((resolve) => { loadStyle(`${blockPath}.css`, resolve); });
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -677,14 +677,14 @@ function isLocalizedPath(path, locales) {
     || legacyLocalePath;
 }
 
-function processQueryIndexMap(link, domain) {
+function processQueryIndexMap(link, domain, fetchOptions = {}) {
   const result = {
     pathsRequest: null,
     requestResolved: false,
     domains: [domain],
   };
 
-  result.pathsRequest = fetch(`${link}?limit=30000`)
+  result.pathsRequest = fetch(`${link}?limit=30000`, fetchOptions)
     .then((response) => response.json())
     .then((json) => json.data?.map((d) => (d.path ?? d.Path)?.replace(/\.html$/, '')) ?? [])
     .catch((error) => {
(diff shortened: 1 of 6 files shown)


mokimo (approved):

libs/utils/utils.js
```diff
@@ -1165,7 +1174,10 @@ export function localizeLink(
 export function loadLink(href, {
   id, as, callback, crossorigin, rel, fetchpriority,
 } = {}) {
-  let link = document.head.querySelector(`link[href="${href}"]`);
+  const selector = rel === 'stylesheet'
+    ? `link[href="${href}"][rel="stylesheet"]`
+    : `link[href="${href}"]`;
```
DKos95, line 1179: NIT:
```suggestion
const selector = `link[href="${href}"]${rel === 'stylesheet' ? '[rel="stylesheet"]' : ''}`;
```

DKos95 (approved):
Looks good to me, found just one NIT.

NadiiaSokolova (approved):
Verified. Ready for Stage.

PR #6319, opened 2026-07-17
[MWPW-199324] - merch moment rich-content update.

DKos95:
Resolves: [MWPW-199324](https://jira.corp.adobe.com/browse/MWPW-199324)

**Test URLs:**
- Before: https://main--da-cc--adobecom.aem.page/cc-shared/fragments/tests/2026/q3/ace1209/fragments/cpro-offer?milolibs=site-redesign-foundation&georouting=off&mep=%2Fcc-shared%2Ffragments%2Ftests%2F2026%2Fq3%2Face1209%2Face1209-cpro-redesign.json--all
- After: https://main--da-cc--adobecom.aem.page/cc-shared/fragments/tests/2026/q3/ace1209/fragments/cpro-offer?milolibs=rc-merch-moment-w3&georouting=off&mep=%2Fcc-shared%2Ffragments%2Ftests%2F2026%2Fq3%2Face1209%2Face1209-cpro-redesign.json--all

### DIFF
--- a/libs/mep/ace1209/rich-content/rich-content.css
+++ b/libs/mep/ace1209/rich-content/rich-content.css
@@ -215,6 +215,49 @@
       }
     }
   }
+
+  &.parallax-garage-door-reveal:has(.rich-content.merch-moment) {
+    .foreground {
+      animation-range: cover -10% cover 90%;
+    }
+  }
+
+  &:has(.rich-content.merch-moment) {
+    display: grid;
+    grid-template-rows: auto auto;
+    box-sizing: border-box;
+    height: auto;
+
+    .rich-content.merch-moment {
+      grid-row: 1;
+    }
+
+    p:has(span[is="inline-price"]) {
+      font-weight: var(--s2a-font-weight-adobe-clean-bold);
+    }
+
+    &.has-background .section-background {
+
+      &,
+      & > * {
+        position: relative;
+        inset: unset;
+      }
+
+      img {
+        width: 100%;
+        height: auto;
+        aspect-ratio: 375 / 434;
+        object-fit: cover;
+        object-position: center top;
+        animation: unset;
+      }
+    }
+  }
+
+  &.dark:has(.rich-content.merch-moment) {
+    background: var(--s2a-color-gray-1000);
+  }
 }
 
 .section.bento .rich-content.grid-full-width {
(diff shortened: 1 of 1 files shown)


mokimo (approved):

libs/mep/ace1209/rich-content/rich-content.css
```diff
@@ -251,6 +294,14 @@
 }
 
 @media (width >= 768px) {
+  .section:has(.rich-content.merch-moment).has-background .section-background img {
+    width: 100%;
+    max-width: 1006px;
```
overmyheadandbody: Is this value (and the one at L441) called out explicitly in the tech specs? The effect looks great for this particular scenario, I just wonder if these values fit other images as well or why we can't base them on the grid max sizes

DKos95: I think we can safely remove the 1006px value as the intentions from design aren't clear enough if it should be constrained or to stretch so it's debatable, but as for the L441 value I have tried to achieve this via authoring with the container class and I think it can't be done with the grid system, at big resolutions like 2560 the difference is noticable and in the figma design it is clearly limited to that 1612 width at that resolution.

libs/mep/ace1209/rich-content/rich-content.css
```diff
@@ -375,12 +425,21 @@
...
+  }
+
   .rich-content.jump-link.hero {
     padding-block-start: var(--s2a-layout-xl);
   }
 }
 
 @media (width >= 1440px) {
+  .section:has(.rich-content.merch-moment).has-background .section-background img {
+    max-width: 1612px;
```
overmyheadandbody, line 440: Then maybe a `grid-width-8` is close enough? This is fine for the test, but it's something that needs to be settled moving forward.

overmyheadandbody (approved):

zagi25 (approved):

libs/mep/ace1209/rich-content/rich-content.css
DKos95: I stand corrected, `grid-width-10` would be good for 1920 onwards as it would give us 1600 image width, but if we put `container grid-width-10` for `desktop-viewport` it will target resolutions from 1280+ and that would mean that 1280 res would have a padding on each side where as in the design the image is end to end.

NadiiaSokolova (approved):
Verified. Ready to be merged.
Testing details [MWPW-199324](https://jira.corp.adobe.com/browse/MWPW-199324)

PR #6320, opened 2026-07-20
fix(modal): don't force focus ring on CTA when modal opened by mouse click

qiyundai:
- Fixes an unwanted blue focus ring appearing on modal CTAs (e.g. a video modal's "Watch full session" button) when the modal is opened by a mouse click. `getModal()` in `libs/blocks/modal/modal.js` was unconditionally calling the initial focus with `focusVisible: true`, forcing the ring regardless of input modality — browser support differences for that focus option explain why only ~50% of users saw it.
- The ring is now only suppressed when the *confirmed* trigger for this modal (the element already tracked via `data-modal-hash`/`data-modal-id`) was itself activated via mouse/pointer (i.e. doesn't match `:focus-visible`). Deep links, delayed/auto-shown modals, and any open with no confirmed trigger element keep the ring visible by default, so keyboard and assistive-technology users never lose their focus indicator.
- Added unit tests in `test/blocks/modals/modals.test.js` covering the keyboard-trigger, mouse-trigger, and no-confirmed-trigger (deep-link/auto-shown) cases.
- Scope: only the shared `libs/blocks/modal/modal.js` implementation was changed; the C2 and MEP-experiment copies of the modal were intentionally left untouched.

Resolves: [MWPW-194884](https://jira.corp.adobe.com/browse/MWPW-194884)

**Test URLs:**
- Before: https://main--da-events--adobecom.aem.page/events
- After: https://main--da-events--adobecom.aem.page/events?milolibs=fix-modal-focus-visible-mouse-click

### DIFF
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -205,17 +205,12 @@ export async function getModal(details, custom) {
   if (isConfirmedTrigger) {
     document.activeElement.dataset.isModalTrigger = 'true';
   }
-  // Default to a visible focus ring on the modal's initial focus target — the safe
-  // fallback for deep links, delayed/auto-shown modals, and any open with no confirmed
-  // trigger click, so keyboard/AT users never lose their focus indicator. Only suppress
-  // it when the confirmed trigger for this modal was itself reached by mouse/pointer
-  // (i.e. doesn't match :focus-visible), which is the reported bug.
   let openedViaKeyboard = true;
   if (isConfirmedTrigger) {
     try {
       openedViaKeyboard = document.activeElement.matches(':focus-visible');
     } catch (e) {
-      // :focus-visible unsupported in matches() — keep the safe default
+      openedViaKeyboard = true;
     }
   }
 
--- a/libs/blocks/modal/modal.js
+++ b/libs/blocks/modal/modal.js
@@ -200,11 +200,23 @@ export async function getModal(details, custom) {
   const { id, deepLink } = details || custom;
   if (id !== LOCALE_MODAL_ID) isDeepLink = deepLink;
   const activeElementData = document.activeElement.dataset;
-  if (!isDeepLink
-    && (activeElementData.modalHash === `#${id}`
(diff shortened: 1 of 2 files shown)


mokimo (approved):
**Approve**

Small, focused fix (+63/-5, 2 files: `libs/blocks/modal/modal.js`, `test/blocks/modals/modals.test.js`) against base `stage`.

Verified `openedViaKeyboard` (modal.js:203-215, 289) correctly gates `focusVisible` only for confirmed-trigger opens, leaves the Tab-cycle focus-trap logic untouched, and fails open to showing the ring for unsupported browsers. Matches an existing precedent (`libs/mep/ace1205/hub-hero/hub-hero.js`). Tests genuinely exercise the confirmed-trigger fixture wiring (`data-modal-hash="#milo"` in `mocks/body.html`). All CI green.

Non-blocking note: PR intentionally leaves C2/MEP-experiment modal copies unpatched — worth checking for a fast-follow ticket, not a blocker.

mokimo (approved):
Verified `openedViaKeyboard` (modal.js:203-215,289) correctly gates `focusVisible` only for confirmed-trigger opens, leaves the Tab-cycle focus-trap logic untouched, fails open to showing the ring for unsupported browsers, and matches an existing precedent (`libs/mep/ace1205/hub-hero/hub-hero.js`). Small (+63/-5, 2 files), base `stage`. Tests genuinely exercise the confirmed-trigger fixture wiring (`data-modal-hash="#milo"` in mocks/body.html), not just mocked past. All CI green (unit, Nala E2E across Milo/MAS/MEP x3 browsers, lint, CodeQL).

Non-blocking note: PR intentionally leaves the C2 and MEP-experiment modal copies unpatched — worth checking whether a fast-follow ticket exists for those, since they may share the same unconditional-ring behavior.

SilviuLCF (approved):
QA verified https://jira.corp.adobe.com/browse/MWPW-194884

narcis-radu (approved):

PR #6228, opened 2026-06-25
[sr] bentos wave 3

rgclayton:
* Added inset border around bento boxes
* Display icon/nemonic (supports multiple)
* Spacing updates within bentos:
  * XL, L : From 48px –> 32px
  * M: From 32px –> 24px
  * S : 24px (same)
* Global fix for `stand-alone `link icon to work with dark mode  

Resolves: 
* [MWPW-199377](https://jira.corp.adobe.com/browse/MWPW-199377)
* [MWPW-198997](https://jira.corp.adobe.com/browse/MWPW-198997)

**Test URLs:**
- Before: https://main--da-cc--adobecom.aem.page/cc-shared/fragments/tests/2026/q3/ace1209/fragments/cpro-hub?milolibs=site-redesign-foundation&mep=%2Fcc-shared%2Ffragments%2Ftests%2F2026%2Fq3%2Face1209%2Face1209-cpro-redesign.json--all
- After: https://main--da-cc--adobecom.aem.page/cc-shared/fragments/tests/2026/q3/ace1209/fragments/cpro-hub?milolibs=bentos-w3&mep=%2Fcc-shared%2Ffragments%2Ftests%2F2026%2Fq3%2Face1209%2Face1209-cpro-redesign.json--all

PR #6322, opened 2026-07-20
[sr] Mobile bento support

rgclayton:
* Adds opt in mobile stacking feature (`stack-mobile`). Mobile only feature.
* Removed border per design ask - originally added in [this PR](https://github.com/adobecom/milo/pull/6228)
* Bento aspect ratio update to 1 / 1.1
* Supports sticky title when authored with a rich-content block as the first element in a bento section.

Resolves: [MWPW-199414](https://jira.corp.adobe.com/browse/MWPW-199414)

**Test URLs:**
- Before: https://main--da-cc--adobecom.aem.page/cc-shared/fragments/tests/2026/q3/ace1209/fragments/cpro-hub?milolibs=site-redesign-foundation&mep=%2Fcc-shared%2Ffragments%2Ftests%2F2026%2Fq3%2Face1209%2Face1209-cpro-redesign.json--all
- After: https://main--da-cc--adobecom.aem.page/cc-shared/fragments/tests/2026/q3/ace1209/fragments/cpro-hub?milolibs=sr-mobile-bento&mep=%2Fcc-shared%2Ffragments%2Ftests%2F2026%2Fq3%2Face1209%2Face1209-cpro-redesign.json--all

image

### DIFF
--- a/libs/features/bento-stack.js
+++ b/libs/features/bento-stack.js
@@ -40,43 +40,62 @@ function clearMeasurements(section) {
   section.style.removeProperty('--gnav-offset');
 }
 
+export function contentReady(cards) {
+  return cards.every((card) => card.querySelector('.explore-card-content'));
+}
+
+// Sibling blocks (e.g. explore-card) decorate in parallel with this module's own
+// (lazy-imported) load, so a card's content wrapper may not exist yet on first run —
+// wait for it (bounded, so a malformed card can't stall the feature forever) rather
+// than measuring the raw, undecorated card and locking that height in permanently.
+export function whenReady(cards, cb, deadline = performance.now() + 1000) {
+  if (contentReady(cards) || performance.now() >= deadline) {
+    cb();
+    return;
+  }
+  requestAnimationFrame(() => whenReady(cards, cb, deadline));
+}
+
 export default function initBentoStack(section) {
   const cards = getCards(section);
-  if (!cards.length) return;
-  setIndices(section, cards);
-  // If this module fails to load/run, the class is never added and the
-  // bento degrades to its normal static layout
-  section.classList.add('bento-stack-ready');
+  if (!cards.length || section.classList.contains('bento-stack-ready')) return;
 
-  const mq = window.matchMedia(MOBILE);
-  let frame = 0;
-  let measuring = false;
-  let ro;
-  const update = () => {
-    // Self-heal: an MEP replaceInner() can detach this section and re-init on a
-    // fresh one. Drop the orphaned observer + listener so they neither pin the
-    // detached nodes nor keep firing on a dead section.
-    if (!section.isConnected) {
-      ro?.disconnect();
-      mq.removeEventListener('change', update);
-      return;
-    }
-    // (clearing/setting --card-height resizes the observed content).
-    if (measuring) return;
-    cancelAnimationFrame(frame);
-    frame = requestAnimationFrame(() => {
-      measuring = true;
-      if (mq.matches) measure(section, cards);
-      else clearMeasurements(section);
-      requestAnimationFrame(() => { measuring = false; });
-    });
-  };
+  whenReady(cards, () => {
+    setIndices(section, cards);
+    // If this module fails to load/run, the class is never added and the
+    // bento degrades to its normal static layout
+    section.classList.add('bento-stack-ready');
 
-  update();
+    const mq = window.matchMedia(MOBILE);
+    let frame = 0;
+    let measuring = false;
+    let ro;
+    const update = () => {
+      // Self-heal: an MEP replaceInner() can detach this section and re-init on a
+      // fresh one. Drop the orphaned observer + listener so they neither pin the
+      // detached nodes nor keep firing on a dead section.
+      if (!section.isConnected) {
+        ro?.disconnect();
+        mq.removeEventListener('change', update);
+        return;
+      }
+      // (clearing/setting --card-height resizes the observed content).
+      if (measuring) return;
+      cancelAnimationFrame(frame);
+      frame = requestAnimationFrame(() => {
+        measuring = true;
+        if (mq.matches) measure(section, cards);
+        else clearMeasurements(section);
+        requestAnimationFrame(() => { measuring = false; });
+      });
+    };
 
-  ro = new ResizeObserver(update);
-  cards.forEach((card) => ro.observe(naturalEl(card)));
-  const title = section.querySelector(':scope > .rich-content');
-  if (title) ro.observe(title);
-  mq.addEventListener('change', update);
+    update();
+
+    ro = new ResizeObserver(update);
+    cards.forEach((card) => ro.observe(naturalEl(card)));
+    const title = section.querySelector(':scope > .rich-content');
+    if (title) ro.observe(title);
+    mq.addEventListener('change', update);
+  });
 }
--- a/libs/features/bento-stack.js
+++ b/libs/features/bento-stack.js
@@ -1,4 +1,4 @@
-const MOBILE = '(max-width: 767px)';
+const MOBILE = '(width < 768px)';
 
 export function getCards(section) {
   return [...section.querySelectorAll(':scope > .explore-card')];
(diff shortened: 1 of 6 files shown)


libs/mep/ace1209/rich-content/rich-content.css
```diff
@@ -381,8 +361,13 @@
 }
 
 @media (width >= 1440px) {
+  .rich-content.indent {
+    margin-block-end: 0;
+  }
```
overmyheadandbody: Isn't this already inherited from https://github.com/adobecom/milo/pull/6322/changes#diff-547cb21a4c0facdd7c96a9b877aba9c3178e494425359797b88fda171f80a40fR292-R294 ?

libs/mep/ace1209/rich-content/rich-content.css
```diff
@@ -562,6 +547,44 @@
...
 
+@keyframes stack-title-stagger {
+  0% {
+    opacity: 0;
+    transform: translate3d(0, var(--parallax-translate-y), 0);
+  }
+
+  100% {
+    opacity: 1;
+    transform: translate(0, 0);
```
overmyheadandbody, line 554: I see a mix between `translate3d` and translate, maybe we can stick to one or the other?

libs/mep/ace1209/section-metadata/section-metadata.css
```diff
@@ -2,6 +2,115 @@
...
+  100% {
+    transform: translateY(calc(var(--stack-offset) - var(--offset-adjuster)));
+    opacity: 1;
+  }
+}
+
+@keyframes stack-section-exit {
+  0% { transform: translate(0, 0); }
+  100% { transform: translate3d(0, calc(-1 * var(--section-exit-y)), 0); }
+}
```
overmyheadandbody, line 53: Let's stick to just `translate` for consistency, as nowadays this alone pushes things to the compositor layer.

libs/mep/ace1209/section-metadata/section-metadata.css
```diff
@@ -2,6 +2,115 @@
...
+.section.bento.stack-mobile {
+  --phone-h: 775px;
+  --phone-vh: calc(var(--phone-h) / 100);
+  --gnav-offset: 72px;
+
+  /* Single knob for design fine-tuning — everything pins at --stack-pin-top. */
+  --bento-stack-top-gap: 40px;
+  --stack-pin-top: calc(var(--gnav-offset) + var(--bento-stack-top-gap));
+  --title-height: 24px;
+  --title-card-gap: var(--s2a-spacing-24, 24px);
```
overmyheadandbody: We know the token exists, so the fallback likely doesn't bring anything useful.

libs/mep/ace1209/section-metadata/section-metadata.css
```diff
@@ -2,6 +2,115 @@
...
+
+  /* To re-enable, set this back to 40px; the stack-section-exit animation
+     + --section-vt timeline are kept intact */
+  --section-exit-y: 0;
+  --paralax-easing: cubic-bezier(0.42, 0, 0, 1);
```
overmyheadandbody: This is already defined in our root. However, there's a small typo here - it should be `parallax`, double `l`. This could then be removed and the value would stay centralized in our main styles, so any change would cascade when needed.

libs/mep/ace1209/section-metadata/section-metadata.css
```diff
@@ -2,6 +2,115 @@
...
+  --bento-surface-middle: #edebeb;
+  --bento-surface-deepest: #e3e2e2;
+  --bento-shadow:
+    0 86.137px 34.455px #00000012,
+    0 35.986px 14.394px #0000000d,
+    0 19.24px 7.696px #0000000a,
+    0 10.786px 4.314px #0000000a,
+    0 5.728px 2.291px #00000008,
+    0 2.384px 0.953px #00000005;
+  --bento-shadow-none: 0 0 0 #0000;
```
overmyheadandbody: Any chance we can use some S2A variables?

libs/mep/ace1209/section-metadata/section-metadata.css
```diff
@@ -287,6 +399,89 @@
...
+    }
+
+    > .explore-card .explore-card-content::after {
+      content: '';
+      position: absolute;
+      inset: 0;
+      z-index: 1;
+      border-radius: inherit;
+      pointer-events: none;
+      background: #000;
```
overmyheadandbody: `--s2a-color-gray-1000` maybe?

libs/mep/ace1209/section-metadata/section-metadata.js
```diff
@@ -204,4 +204,9 @@ export default async function init(el) {
   if (metadata.layout) handleStyle(metadata.layout.text, section);
   if (metadata.images) handleImages(metadata.images?.text[0], section);
   handleStickyFocus(section);
+  if (section?.matches('.bento.stack-mobile')) {
+    import('../../../features/bento-stack.js')
+      .then(({ default: initBentoStack }) => initBentoStack(section))
+      .catch((e) => window.lana?.log(`bento-stack init failed: ${e}`, { tags: 'bento-stack', severity: 'info' }));
```
overmyheadandbody: I'd suggest this be moved to a `handleBentoStacking` method, to follow the same pattern as above.

libs/features/bento-stack.js
```diff
@@ -0,0 +1,82 @@
...
+
+function clearMeasurements(section) {
+  section.style.removeProperty('--card-height');
+  section.style.removeProperty('--title-height');
+  section.style.removeProperty('--gnav-offset');
+}
+
+export default function initBentoStack(section) {
+  const cards = getCards(section);
+  if (!cards.length) return;
```
overmyheadandbody: Maybe it's a good idea to check whether `bento-stack-ready` has already been added, to avoid the possibility of re-attaching the Observer

overmyheadandbody (commented):
Looking good! I have a few suggestions

libs/mep/ace1209/section-metadata/section-metadata.css
rgclayton: Good catch!

libs/mep/ace1209/section-metadata/section-metadata.css
rgclayton: I updated `--bento-shadow-none` to use a s2a token. The others do not have a close related value due to the transparency. The jumps are too big in the s2a tokens.

libs/mep/ace1209/section-metadata/section-metadata.css
rgclayton: Updated.

libs/mep/ace1209/rich-content/rich-content.css
rgclayton: Updated.

libs/mep/ace1209/rich-content/rich-content.css
rgclayton: Updated.

libs/mep/ace1209/section-metadata/section-metadata.css
rgclayton: Updated.

libs/mep/ace1209/section-metadata/section-metadata.css
rgclayton: Updated.

libs/mep/ace1209/section-metadata/section-metadata.js
rgclayton: Updated.

libs/features/bento-stack.js
rgclayton: Updated.

rgclayton:
Thanks for the feedback @overmyheadandbody!

I got all of it implemented. I also discovered an edge case race condition that I solved for. Claude suggested I add some unit tests to help with that testing so I went ahead and added those to harden the code better.

overmyheadandbody (approved):
Thanks for all the changes, Ryan, looking great!

SilviuLCF (approved):
QA verified , testing details https://jira.corp.adobe.com/browse/MWPW-199414

libs/mep/ace1209/rich-content/rich-content.css
```diff
@@ -562,6 +543,44 @@
...
+
+    position: sticky;
+    top: var(--stack-pin-top);
+    z-index: 1;
+    margin: 0;
+    line-height: 1;
+    will-change: opacity, transform, translate;
+    animation:
+      stack-title-stagger var(--paralax-easing) both,
+      stack-title-card-match var(--paralax-easing) both;
```
zagi25: Typo, I don't see an example of this on the test page, but animation won't work because of the undefined CSS variable. 
```suggestion
      stack-title-stagger var(--parallax-easing) both,
      stack-title-card-match var(--parallax-easing) both;
```

zagi25 (approved):
Approving, since rest of the code looks good, but we should not merge before typo is fixed.

libs/mep/ace1209/rich-content/rich-content.css
rgclayton: Good catch! Thank you!

rgclayton: Updated.

PR #6337, opened 2026-07-23
[SR] Section hairline gap

rgclayton:
This issue is showing up more on the new Cpro pages due to the dark background throughout the page.
* Added -0.5px inset to top and bottom of `.section-background` to remove hairline gap.

Resolves: [MWPW-201576](https://jira.corp.adobe.com/browse/MWPW-201576)

**Test URLs:**
- Before: https://main--da-cc--adobecom.aem.page/cc-shared/fragments/tests/2026/q3/ace1209/fragments/cpro-hub?milolibs=site-redesign-foundation&mep=%2Fcc-shared%2Ffragments%2Ftests%2F2026%2Fq3%2Face1209%2Face1209-cpro-redesign.json--all
- After: https://main--da-cc--adobecom.aem.page/cc-shared/fragments/tests/2026/q3/ace1209/fragments/cpro-hub?milolibs=sr-section-gap&mep=%2Fcc-shared%2Ffragments%2Ftests%2F2026%2Fq3%2Face1209%2Face1209-cpro-redesign.json--all

Before:
image

After:
image

### DIFF
--- a/libs/mep/ace1209/section-metadata/section-metadata.css
+++ b/libs/mep/ace1209/section-metadata/section-metadata.css
@@ -353,7 +353,11 @@ body:has(header.local-nav) .section.bento.stack-mobile {
 .section-background {
   z-index: 0;
 
-  &,
+  /* Bleed -0.5px into the section background top/bottom to remove hairline gap created by
+     sub-pixel compositor (from parallax/view-timeline animations during smooth-scroll) */
+  position: absolute;
+  inset: -0.5px 0;
+
   & > * {
     position: absolute;
     inset: 0;
--- a/libs/mep/ace1209/section-metadata/section-metadata.css
+++ b/libs/mep/ace1209/section-metadata/section-metadata.css
@@ -130,6 +130,9 @@ body:has(header.local-nav) .section.bento.stack-mobile {
     overflow: clip;
     z-index: 3;
     margin-block: calc(-1 * var(--s2a-border-radius-lg));
+    /* Bleed 1px into the section top/bottom to remove hairline gap created by
+     sub-pixel compositor */
+    inset: 1px 0;
   }
 
   &.rounded-corners-top {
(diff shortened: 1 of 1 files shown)


libs/mep/ace1209/section-metadata/section-metadata.css
```diff
@@ -353,7 +353,11 @@ body:has(header.local-nav) .section.bento.stack-mobile {
 .section-background {
   z-index: 0;
 
-  &,
+  /* Bleed -0.5px into the section background top/bottom to remove hairline gap created by
+     sub-pixel compositor (from parallax/view-timeline animations during smooth-scroll) */
+  position: absolute;
+  inset: -0.5px 0;
```
overmyheadandbody: Since we're trying to avoid some sub-pixel rendering issues, why not go with a full `1px` here?

overmyheadandbody (approved):
This is a step in the right direction! However, when dev tools are open, there's still a hairline on scroll. Looking into it, it seems that removing the `rounded-corners-bottom` variant solves it. Would it be worth looking into why that is?

libs/mep/ace1209/section-metadata/section-metadata.css
rgclayton: Just trying to minimize impact. 0.5px was correcting the issue. I can easily round up to 1px.

rgclayton: I adjusted to a whole pixel. 

I also added an inset for the rounded corners section to mitigate when dev tools are open. Give it another test.

overmyheadandbody: Thanks for the update, Ryan! The rounded corners hairline is still there, unfortunately, I could only solve it with a negative `1px` top margin. We can get back to this in the future, as it seems it only occurs when Dev Tools are open, though.

zagi25: I was only able to reproduce this in responsive mobile mode. To me it looks like https://github.com/adobecom/milo/pull/6337/changes#diff-94ea2acee86eb8447786f052cb81a537578636fbee2c4951a281842465221f34R135 cancels out other inset. 
In both screenshots I deleted globe block and this creates two hairlines, if above rule is disabled it seems that both hairlines are gone 🤷‍♂️
Screenshot 2026-07-24 at 16 09 56
Screenshot 2026-07-24 at 16 10 08

rgclayton: I'll remove the new rounded corner inset i added yesterday. I agree that one was getting a little tricky and I too was only seeing it in dev mode with zoomed out settings (75%).

zagi25 (approved):

narcis-radu (approved):

NadiiaSokolova (approved):
Verified. Ready to be merged.

PR #6340, opened 2026-07-23
[SR] Tabs Block (Wave 3)

ivanvatadobe:
This PR adds the tabs block needed for ACE1209 (Site redesign wave 3) test. 

Note: this is the first push to `site-redesign-foundation`, and **_I would like to push this PR through immediately._** There are still items to implement, but I want to at least have a draft version of the wave 3 tabs in our shared demo pages. 

**Features:** 
- **Add variant: radio** - radio variant is re-enabled
- **Add variant: quiet** - basic text with blue underline as clickable tabs
- **Add dark mode for base variant (pill)**
- **Add dark mode for radio variant**
- **Add dark mode for quiet variant** 

**TODO/Notes**
1. confirm spacing for top level tabs when nested (section spacing is not applied below the parent tab).
2. confirm left/right mode works as intended from ace1205.
3. implement autoscroll to center selected radio tab in lower breakpoints (this is already available for base pill). 

**Resolves:** 
[MWPW-198998](https://jira.corp.adobe.com/browse/MWPW-198998) 

**Test URLs:**
- Before: https://main--da-cc--adobecom.aem.page/cc-shared/fragments/tests/2026/q3/ace1209/playground/viloria/tabs-layers-poc-radio-dark?milolibs=site-redesign-foundation&martech=off
- After: https://main--da-cc--adobecom.aem.page/cc-shared/fragments/tests/2026/q3/ace1209/playground/viloria/tabs-layers-poc-pill-dark?milolibs=sr3-tabs-block-radio&martech=off

**DEMO PAGES WITH TABS BY TYPE AND MODE** 

**base pill - default**
https://main--da-cc--adobecom.aem.page/cc-shared/fragments/tests/2026/q3/ace1209/playground/viloria/tabs-layers-poc-pill-light?milolibs=sr3-tabs-block-radio

**base pill - dark**
https://main--da-cc--adobecom.aem.page/cc-shared/fragments/tests/2026/q3/ace1209/playground/viloria/tabs-layers-poc-pill-dark?milolibs=sr3-tabs-block-radio

**radio - default**
https://main--da-cc--adobecom.aem.page/cc-shared/fragments/tests/2026/q3/ace1209/playground/viloria/tabs-layers-poc-radio-light?milolibs=sr3-tabs-block-radio

**radio - dark**
https://main--da-cc--adobecom.aem.page/cc-shared/fragments/tests/2026/q3/ace1209/playground/viloria/tabs-layers-poc-radio-dark?milolibs=sr3-tabs-block-radio

### DIFF
--- a/libs/mep/ace1209/tabs/tabs.css
+++ b/libs/mep/ace1209/tabs/tabs.css
@@ -35,12 +35,6 @@
       padding: var(--s2a-spacing-2xs) var(--s2a-spacing-lg);
     }
 
-    .tab-list-container::before {
-      content: attr(data-pretext);
-      color: var(--s2a-color-content-default);
-      white-space: nowrap;
-    }
-
     button.tab-button {
       display: flex;
       align-items: center;
--- a/libs/mep/ace1209/tabs/tabs.css
+++ b/libs/mep/ace1209/tabs/tabs.css
@@ -24,22 +24,28 @@
         display: flex;
         align-items: center;
         min-width: 0;
-        gap: var(--s2a-spacing-lg);
-        background: var(--s2a-color-transparent-black-04);
+        gap: var(--s2a-spacing-md);
+        background: var(--s2a-color-background-subtle);
         border-radius: var(--pill-rounded-corners-setting);
-        padding: var(--s2a-spacing-2xs) var(--s2a-spacing-lg);
+        padding: var(--s2a-spacing-sm) var(--s2a-spacing-md);
+      }
+
+      .tab-list-label {
+        color: var(--s2a-color-content-subtle);
+        white-space: nowrap;
+        margin: 0;
       }
 
       button.tab-button {
         display: flex;
         align-items: center;
-        gap: var(--s2a-spacing-2xs);
+        gap: var(--s2a-spacing-xs);
         height: auto;
         flex: 0 0 auto;
         padding: 0;
         background: transparent;
         border: none;
-        color: var(--s2a-color-content-default);
+        color: var(--s2a-color-content-subtle);
         cursor: pointer;
         font-family: var(--body-font-family);
         text-decoration: none;
@@ -50,9 +56,9 @@
         &::before {
           content: '';
           box-sizing: border-box;
-          width: 14px;
-          height: 14px;
-          min-width: 14px;
+          width: 20px;
+          height: 20px;
+          min-width: 20px;
           border: 2px solid var(--s2a-color-gray-400);
           border-radius: 100%;
           transition: border 150ms;
@@ -62,9 +68,13 @@
           border-color: var(--pill-bg-color-selected);
         }
 
-        &[aria-checked="true"]::before {
-          border-width: 5px;
-          border-color: var(--pill-bg-color-selected);
+        &[aria-checked="true"] {
+          color: var(--s2a-color-content-default);
+
+          &::before {
+            border-width: 6px;
+            border-color: var(--pill-bg-color-selected);
+          }
         }
 
         &:focus-visible {
--- a/libs/mep/ace1209/tabs/tabs.css
+++ b/libs/mep/ace1209/tabs/tabs.css
@@ -190,7 +190,25 @@
         background: var(--s2a-color-background-inverse);
         color: var(--s2a-color-content-inverse);
       }
+
+      button.tab-button:hover:not([aria-selected="true"]) {
+        color: var(--s2a-color-gray-100);
+      }
+
+      .tab-indicator {
+        background: var(--s2a-color-background-inverse);
+      }
     }
+  }
+
+  /*
+   * Plain pill is the only variant that keeps a solid block-level background
+   * in light mode (radio/quiet already force it transparent unconditionally
+   * above). Nothing overrode it for dark, so it stayed at the light value —
+   * matching radio/quiet's transparent treatment for now.
+   */
+  &.dark:not(.radio):not(.quiet) {
+    background: transparent;
   }
 
   .tabs-wrapper {
--- a/libs/mep/ace1209/tabs/tabs.css
+++ b/libs/mep/ace1209/tabs/tabs.css
@@ -19,58 +19,58 @@
 
     [role='radiogroup'] {
       overflow: visible;
-    }
 
-    .tab-list-container {
-      display: flex;
-      align-items: center;
-      min-width: 0;
-      gap: var(--s2a-spacing-lg);
-      background: var(--s2a-color-transparent-black-04);
-      border-radius: var(--pill-rounded-corners-setting);
-      padding: var(--s2a-spacing-2xs) var(--s2a-spacing-lg);
-    }
-
-    button.tab-button {
-      display: flex;
-      align-items: center;
-      gap: var(--s2a-spacing-2xs);
-      height: auto;
-      flex: 0 0 auto;
-      padding: 0;
-      background: transparent;
-      border: none;
-      color: var(--s2a-color-content-default);
-      cursor: pointer;
-      font-family: var(--body-font-family);
-      text-decoration: none;
-      outline: none;
-      -webkit-tap-highlight-color: transparent;
-      white-space: nowrap;
-
-      &::before {
-        content: '';
-        box-sizing: border-box;
-        width: 14px;
-        height: 14px;
-        min-width: 14px;
-        border: 2px solid var(--s2a-color-gray-400);
-        border-radius: 100%;
-        transition: border 150ms;
+      .tab-list-container {
+        display: flex;
+        align-items: center;
+        min-width: 0;
+        gap: var(--s2a-spacing-lg);
+        background: var(--s2a-color-transparent-black-04);
+        border-radius: var(--pill-rounded-corners-setting);
+        padding: var(--s2a-spacing-2xs) var(--s2a-spacing-lg);
       }
 
-      &:hover::before {
-        border-color: var(--pill-bg-color-selected);
-      }
+      button.tab-button {
+        display: flex;
+        align-items: center;
+        gap: var(--s2a-spacing-2xs);
+        height: auto;
+        flex: 0 0 auto;
+        padding: 0;
+        background: transparent;
+        border: none;
+        color: var(--s2a-color-content-default);
+        cursor: pointer;
+        font-family: var(--body-font-family);
+        text-decoration: none;
+        outline: none;
+        -webkit-tap-highlight-color: transparent;
+        white-space: nowrap;
 
-      &[aria-checked="true"]::before {
-        border-width: 5px;
-        border-color: var(--pill-bg-color-selected);
-      }
+        &::before {
+          content: '';
+          box-sizing: border-box;
+          width: 14px;
+          height: 14px;
+          min-width: 14px;
+          border: 2px solid var(--s2a-color-gray-400);
+          border-radius: 100%;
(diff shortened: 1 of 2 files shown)


rgclayton:
Clicking any of the parent tabs (photo, design, video) causes the page to jump up.

libs/mep/ace1209/tabs/tabs.css
```diff
@@ -13,6 +15,157 @@
...
+          background: transparent;
+          color: var(--s2a-color-content-default);
+          border-bottom-color: var(--s2a-color-blue-800);
+        }
+      }
+    }
+  }
+
+  &.dark:not(.radio):not(.quiet) > .tabs-wrapper {
+    [role='tablist'] {
```
rgclayton, line 241: This and on line 165 — the condition `:not(.radio):not(.quiet)` is duplicated across two separate rule blocks; could be consolidated (e.g., a shared class or comment noting they must stay in sync) to reduce future drift risk.

rgclayton:
Looks like some alignment issues on mobile compared [to the design](https://www.figma.com/design/brgxjoNmioT7z6zFk1QHbR/Plans-%E2%80%94-A.com?node-id=2918-20714&m=dev)
image

image

libs/mep/ace1209/tabs/tabs.js
```diff
@@ -271,39 +266,48 @@ const init = async (block) => {
...
       const tabBtn = createTag('button', tabBtnAttributes, item.textContent);
       const btnWrapper = createTag('div', { class: 'btn-wrapper' });
       btnWrapper.append(tabBtn);
       tabListContainer.append(btnWrapper);
 
       const tabContentAttributes = {
-        id: `tab-panel-${tabId}-${tabName}`,
-        role: 'tabpanel',
+        id: controlId,
+        ...(isRadio ? {} : { role: 'tabpanel' }),
```
zagi25, line 309: **Blocking (radio visual regression):** radio panels get no role here, so they render as `<div class="tabpanel">`. But the layout rules in `tabs.css:295` and `:357-359` still key off `[role='tabpanel']` (section positioning, content centering, background z-index), and content `.section`s are appended into these panels — so the radio variant won't receive them → visual regression. Fix: either keep `role='tabpanel'` on radio panels, or update those two CSS selectors to `.tabpanel` (matching the change you already made at line 377).

libs/mep/ace1209/tabs/tabs.js
```diff
@@ -271,39 +266,48 @@ const init = async (block) => {
...
+        id: btnId,
         tabindex: (i === 0) ? '0' : '-1',
-        'aria-selected': (i === 0) ? 'true' : 'false',
+        [isRadio ? 'aria-checked' : 'aria-selected']: (i === 0) ? 'true' : 'false',
         'data-block-id': `tabs-${tabId}`,
         'daa-state': 'true',
-        'daa-ll': `tab-${tabId}-${tabName}`,
-        'aria-controls': controlId,
+        'daa-ll': btnId,
+        ...(isRadio ? { 'data-control-id': controlId } : { 'aria-controls': controlId }),
```
zagi25, line 300: **Radio a11y:** radio buttons use `data-control-id` instead of `aria-controls`, and the panels are role-less (line 308), so there's no programmatic tab→panel association — a screen-reader user isn't told that selecting a radio revealed content elsewhere, and the `aria-labelledby` on the role-less panel (line 310) is effectively ignored by AT. Somewhat inherent to modeling a tab/disclosure widget as a radiogroup — worth tracking with the radio TODOs.

libs/mep/ace1209/tabs/tabs.js
```diff
@@ -191,8 +183,8 @@ function configTabs(config, rootElem) {
 }
 
 function initTabs(elm, config, rootElem) {
-  const tabs = elm.querySelectorAll('[role="tab"]');
-  const tabLists = elm.querySelectorAll('[role="tablist"]');
+  const tabs = elm.querySelectorAll(':scope > .tabs-wrapper [role="tab"], :scope > .tabs-wrapper [role="radio"]');
+  const tabLists = elm.querySelectorAll(':scope > .tabs-wrapper [role="tablist"], :scope > .tabs-wrapper [role="radiogroup"]');
   let tabFocus = 0;
   tabLists.forEach((tabList) => {
     tabList.addEventListener('keydown', (e) => {
```
zagi25, line 190: **Radio a11y (keyboard):** for `role=radiogroup`, APG expects arrow keys to move focus *and* check the radio (selection follows focus), and to honor Up/Down as well as Left/Right. Here arrows only move roving focus; `aria-checked` changes on click/Space/Enter, and only Left/Right are handled. Functional, but not the standard radiogroup interaction.

libs/mep/ace1209/tabs/tabs.js
```diff
@@ -271,39 +266,48 @@ const init = async (block) => {
...
-  if (tabListLabel) tabList.setAttribute('aria-label', tabListLabel);
 
-  const tabListItems = rows[0].querySelectorAll(':scope li');
+  const tabListLabelEl = tabListContainer.querySelector(':scope > p');
+  if (tabListLabelEl && isRadio) {
+    tabListLabelEl.classList.add('tab-list-label', 'label');
+    tabListLabelEl.id = `tab-list-label-${tabId}`;
+    tabList.setAttribute('aria-labelledby', tabListLabelEl.id);
+  } else {
+    tabListLabelEl?.remove();
```
zagi25, line 280: Nit (behavior change): for non-radio variants a stray `<p>` in the tab-list container is now removed; previously it would have survived. Looks intentional — just flagging.

libs/mep/ace1209/tabs/tabs.css
```diff
@@ -1,5 +1,7 @@
+/* stylelint-disable no-descending-specificity */
```
zagi25, line 1: Nit: blanket-disabling `no-descending-specificity` for the whole file is a smell — the ordering could be resolved instead of globally disabled.

zagi25 (commented):
_Reviewed by Claude (Claude Code, Opus 4.8), run by @zagi25._

Solid, net-positive change for a flag-gated MEP experiment (dark/pill base + `radio` + `quiet`; migration off hardcoded hexes and `-knockout` tokens to valid semantic `--s2a-*` tokens — verified, no typos). The base/quiet variants meet the WAI-ARIA tabs pattern.

**One issue to fix before the radio variant carries real content** (inline on `tabs.js:308`): radio panels lose `role='tabpanel'` but the CSS at `tabs.css:295` / `:357-359` still targets `[role='tabpanel']` → visual regression.

**Radio-variant a11y deviations** (inline): selection doesn't follow focus + Up/Down unsupported; no programmatic panel association; and the radiogroup's accessible name depends on an authored `<p>` label or `config.pretext` — none if neither is present. Somewhat inherent to the radiogroup modeling; suggest tracking against the existing radio TODOs rather than blocking.

**Minor:** the legacy `.tabs-background-transparent` hook (`tabs.css:316-321`) only targets `[role='tablist']`, so it won't make the radio `role=radiogroup` transparent (edge case).

**Security:** no new risk — no unsafe sinks; the one dynamic `#`-selector uses a sanitized, code-generated id.
**Tests:** none added — consistent with the `ace*` experiment convention (a gap, not a regression given it's flag-gated).

libs/mep/ace1209/tabs/tabs.js
ivanvatadobe: this is intentional. the `<p>` determines the label for radio buttons. if radio is not enabled, then the p is deleted.

libs/mep/ace1209/tabs/tabs.js
ivanvatadobe: adding to todo. i will reach out for a general guidance on which solution to implement regarding `tabpanel` vs `region`.

libs/mep/ace1209/tabs/tabs.js
ivanvatadobe: updated using `.tabpanel` css selector option.

libs/mep/ace1209/tabs/tabs.js
ivanvatadobe: adding to todo. closely related with decision for item about (re `tabpanel` vs `region`). 

current behavior: 
left/right arrows - moves focus (does not activate) to tab in current tabgroup. 
down/up - does not change between tabgroups
tab - switches to next group/exists the tabgroups altogether when reaching the last child tab items
cmd+tab - switches back to the previous tabgroup
enter - activates focused tab in tabground

libs/mep/ace1209/tabs/tabs.css
ivanvatadobe: added `.pill` class to base variant to simplify selectors and removed the dupes.

ivanvatadobe:
> Looks like some alignment issues on mobile compared [to the design](https://www.figma.com/design/brgxjoNmioT7z6zFk1QHbR/Plans-%E2%80%94-A.com?node-id=2918-20714&m=dev) image
> 
> image

added styling to fix this issue.

ivanvatadobe:
@zagi25 

i met with @narcis-radu to discuss the three issues not yet addressed: 

1. APG expects arrow keys to move focus and check the radio (selection follows focus), and to honor Up/Down as well as Left/Right.
2. aria-labelledby on the role-less panel (line 310) is effectively ignored by AT.
3. page jump issue (also an issue in production on /acrobat.html). 

i am going to reach out to Taryn to discuss the accessibility issues- if she requests changes, then it will be listed as a follow-up, not required for the test unless absolutely necessary. 

I'm also going to investigate the page jump issue, but i don't believe that's a deal breaker for getting this pr approved for further testing on our test pages. 

please review the other changes when ready so i can get this to `site-redesign-foundation`.

rgclayton:
> Clicking any of the parent tabs (photo, design, video) causes the page to jump up.

I had this happen today while testing. It happened clicking on a nested tab. 

I also wanted to verify if the merch cards in the nested tabs are all the same? When i click any of those tabs the content doesn't change, and the animation doesn't run. Is that expected?

ivanvatadobe:
> > Clicking any of the parent tabs (photo, design, video) causes the page to jump up.
> 
> I had this happen today while testing. It happened clicking on a nested tab.
> 
> I also wanted to verify if the merch cards in the nested tabs are all the same? When i click any of those tabs the content doesn't change, and the animation doesn't run. Is that expected?

no... this isn't expected. the only issue (which is also replicable in production right now in c1) is the page jumping a bit when you click on the tabs when the page is not fully loaded yet (the bottom of the page takes a _very_ long time to load. 

the animations work as intended/accepted in the previous wave for me.

rgclayton (approved):

zagi25 (approved):

DKos95:
I would suggest asking the accessibility team if the radio buttons are big enough to pass a11y, as far as I remember 24x24px is the minimum dimensions a clickable object can be and currently the radio buttons are 20x20.

narcis-radu (approved):

overmyheadandbody (approved):
Visual nit: the "g" in "Design" gets cut off at the bottom on desktop

Image

ivanvatadobe:
> Visual nit: the "g" in "Design" gets cut off at the bottom on desktop
> 
> Image

i added css fixes for this issue.

SilviuLCF (approved):
Verified, testing details https://jira.corp.adobe.com/browse/MWPW-198998

PR #6342, opened 2026-07-24
MWPW-201353: add unit tests for c2 rich-content block

SilviuLCF:
## Summary
- Adds unit tests for the `rich-content` c2 block (`libs/c2/blocks/rich-content`), which previously had no coverage in `test/blocks`
- Covers: foreground/content decoration, opening-quote hanging into a `.opening-quote` span, hero-overlay source detection + section `--rc-hero-overlay` variable, image-cell guard (no overlay source), and first-paragraph heading promotion
- Follows conventions from existing block tests (e.g. `test/blocks/brand-concierge`, `test/blocks/base-card`)

Resolves: https://jira.corp.adobe.com/browse/MWPW-201353

## Test plan
- [x] `npx wtr --config ./web-test-runner.config.mjs "./test/blocks/rich-content/rich-content.test.js" --node-resolve --port=2000` — 7/7 passing
- [x] `npx eslint test/blocks/rich-content/rich-content.test.js` — clean

### DIFF
--- a/test/blocks/rich-content/mocks/bg-image.html
+++ b/test/blocks/rich-content/mocks/bg-image.html
@@ -1,0 +1,12 @@
+<main>
+  <div class="section">
+    <div class="rich-content">
+      <div>
+        <div>
+          <h2>Heading over media</h2>
+        </div>
+        <div><picture><img src="" alt="background" /></picture></div>
+      </div>
+    </div>
+  </div>
+</main>
--- a/test/blocks/rich-content/mocks/default.html
+++ b/test/blocks/rich-content/mocks/default.html
@@ -1,0 +1,13 @@
+<main>
+  <div class="section">
+    <div class="rich-content">
+      <div>
+        <div>
+          <h2>“Great things ahead</h2>
+          <p>Some supporting copy.</p>
+        </div>
+        <div>rgba(0, 0, 0, 0.5)</div>
+      </div>
+    </div>
+  </div>
+</main>
--- a/test/blocks/rich-content/mocks/no-quote.html
+++ b/test/blocks/rich-content/mocks/no-quote.html
@@ -1,0 +1,12 @@
+<main>
+  <div class="section">
+    <div class="rich-content">
+      <div>
+        <div>
+          <h2>Great things ahead</h2>
+          <p>Some supporting copy.</p>
+        </div>
+      </div>
+    </div>
+  </div>
+</main>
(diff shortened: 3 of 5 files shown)


zagi25 (approved):
Approving — assertions match stage's `rich-content.js` behavior (foreground/content classes, opening-quote hang, text-only hero-overlay source + the image-cell guard, first-paragraph heading promotion). Addition-only, zero-impact.

One heads-up (not a change request on this PR): these tests assert the hero-overlay logic (`hero-overlay-source` class + overlay variable) that lives on `stage`. The `site-redesign-foundation` branch has a diverged copy of `rich-content.js` that removed that logic — so if that version reaches stage, tests 4 & 5 here would break (test 4 failing, test 5 passing for the wrong reason). Worth reconciling the two branches before the redesign work merges. Not blocking this PR against its current base.

mokimo (approved):
Reviewed per Milo conventions (test-only PR).

**Verified against source** — pulled `libs/c2/blocks/rich-content/rich-content.js` (pre-existing block, not touched here) and traced each assertion against the real implementation:
- `decorate()`'s foreground/content class wiring, `hangOpeningQuote`'s `\p{Pi}` regex + span-prepend, the `hero-overlay-source` guard (`querySelector('picture, img')` + non-empty text), and `promoteParagraphHeading`'s body→heading class swap all match the mocks and expectations exactly (traced the `body-*` class assignment through `decorateBlockText` in `libs/utils/decorate.js` to confirm the promote-heading mock's untagged `<p>`s get tagged before the promotion check runs).
- `no-quote.html` is correctly reused for both the "no opening quote" and "leaves paragraphs alone" cases since it has an `h2` and no second cell.

**Conventions**
- Directory/file layout (`test/blocks/rich-content/rich-content.test.js` + `mocks/`) matches `test/blocks/base-card` and `test/blocks/brand-concierge` exactly, as claimed in the description.
- `libs/c2/blocks/rich-content` has no CODEOWNERS entry, but that's a pre-existing gap on the block itself (added before this PR) — not something this test-only diff introduces, and no other `test/blocks/*` directory has a CODEOWNERS line either, so nothing to flag here.
- No new deps, no console.log/debugger, no CSS-in-JS, base branch is `stage`. Well under size thresholds (+132/-0, 5 files, `zero-impact` label is accurate).
- Test plan in the description includes the actual run (7/7 passing) and a clean eslint pass — good enough for a unit-test-only change; no aem.page test URL needed since no page/markup behavior changed.

No blockers, no suggested changes. Nice fill of a real coverage gap.

PR #6346, opened 2026-07-24
[MWPW-187555] - Comparison table a11y

DKos95:
Change has been vetted by the accessibility team.

Resolves: [MWPW-187555](https://jira.corp.adobe.com/browse/MWPW-187555)

**Test URLs:**
- Before: https://stage--milo--adobecom.aem.page/docs/library/kitchen-sink/comparison-table
- After: https://comparison-table-a11y--milo--adobecom.aem.page/docs/library/kitchen-sink/comparison-table

### DIFF
--- a/libs/blocks/comparison-table/comparison-table.js
+++ b/libs/blocks/comparison-table/comparison-table.js
@@ -472,15 +472,22 @@ function setupResponsiveHiding(el) {
 
 function setAccessibilityLabels(el) {
   import('../../features/placeholders.js').then(({ replaceKeyArray }) => {
-    replaceKeyArray(['choose-table-column', 'empty-table-cell', 'not-a-feature'], getConfig())
-      .then(([ariaLabel, emptyText, notAFeatureText]) => {
+    replaceKeyArray(['choose-table-column', 'empty-table-cell', 'not-a-feature', 'primary-feature'], getConfig())
+      .then(([ariaLabel, emptyText, notAFeatureText, primaryFeatureText]) => {
         [...el.querySelectorAll('.mobile-filter-select')].forEach((element) => element.setAttribute('aria-label', ariaLabel));
 
         [...el.querySelectorAll('.table-cell div')].forEach((cellDiv) => {
           const closeIcon = cellDiv.querySelector('.icon-close');
           if (closeIcon) {
-            closeIcon.setAttribute('role', 'img');
-            closeIcon.setAttribute('aria-label', notAFeatureText);
+            closeIcon.setAttribute('aria-hidden', 'true');
+            cellDiv.appendChild(createTag('span', { class: 'sr-only' }, notAFeatureText));
+            return;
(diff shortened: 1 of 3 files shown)


libs/blocks/comparison-table/comparison-table.js
```diff
@@ -472,18 +472,32 @@ function setupResponsiveHiding(el) {
...
+        [...el.querySelectorAll('.table-cell div')].forEach((cellDiv) => {
+          const closeIcon = cellDiv.querySelector('.icon-close');
+          if (closeIcon) {
+            closeIcon.setAttribute('aria-hidden', 'true');
+            cellDiv.appendChild(createTag('span', { class: 'sr-only' }, notAFeatureText));
+            return;
+          }
+          const checkmarkIcon = cellDiv.querySelector('.icon-checkmark');
+          if (checkmarkIcon) {
+            checkmarkIcon.querySelector('title')?.remove();
```
zagi25, line 489: Does title needs to be removed from `closeIcon` also ?

narcis-radu: It doesn't look like the close icon has a `title`, but we can be defensive if needed and add the same logic.

narcis-radu (approved):

libs/blocks/comparison-table/comparison-table.js
DKos95: Remove should be on both because both are getting either sr 'not a feature' or 'primary feature'.

mokimo (approved):
Fixes a real, a11y-vetted issue in `libs/blocks/comparison-table/comparison-table.js`'s `setAccessibilityLabels`: checkmark icons carried a redundant native SVG `<title>` with no distinct screen-reader text, and close/empty cells were indistinguishable to assistive tech. Fix hides both icon types via `aria-hidden`, removes the redundant `<title>`, adds distinct `sr-only` text per case.

+48/-13 across 3 files, not oversize, single concern, base `stage`. CODEOWNERS already covers `libs/blocks/comparison-table/`. No new deps, no console.*, no inline CSS. New tests directly assert the resulting DOM; verified the new placeholder keys (`not-a-feature`, `primary-feature`) resolve through the existing dash-to-space fallback in `libs/features/placeholders.js`, matching the pattern already used for `choose-table-column`/`empty-table-cell`. Confirmed the mock `body.html` still leaves other empty cells in place, so the pre-existing empty-cell fallback path stays test-covered. CI green, already approved by @narcis-radu.

Non-blocking nit: the key name `primary-feature` sits oddly next to the file's existing `COLUMN_TYPES.PRIMARY" concept — worth a glance that the live content sheet's copy doesn't get confused with that.

NadiiaSokolova (approved):
Verified. Ready for Stage.
Testing details [MWPW-187555](https://jira.corp.adobe.com/browse/MWPW-187555)

PR #6332, opened 2026-07-22
[MWPW-200955] Sync ace1205 to c2/blocks

overmyheadandbody:
This brings over all of the Wave 2 Site Redesign work for the ACE1205 A/B test into the regular `libs/c2/blocks` folder, in preparation for rollout.

New blocks that have been copy/pasted from the test folder: tabs, comparison-table-c2, side-by-side, pdf-space, faq, floating-cta, product-marquee-grid, split-aside-grid, hub-hero, plans-hero, tour, quick-actions, offer-hero, logo-ticker, hover-list, social-proof.

Blocks that got updates in the test folder and already existed in `libs/c2/blocks`: explore-card, rich-content, section-metadata, brand-concierge, elastic-carousel, news. 🚨 This is where the most review and QA attention is needed. The homepage (linked below) is a good example of a page that will get updates to blocks already on production, it should be tested thoroughly.

Resolves: [MWPW-200955](https://jira.corp.adobe.com/browse/MWPW-200955)

**Test URLs:**
Hub
* ACE1205: https://main--da-dc--adobecom.aem.live/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-bizpro-hub
* libs/c2/blocks: https://main--da-dc--adobecom.aem.live/drafts/ramuntea/ace1205/ace1205-bizpro-hub?milolibs=ace1205-to-stage

Offer
* ACE1205: https://main--da-dc--adobecom.aem.live/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-offer
* libs/c2/blocks: https://main--da-dc--adobecom.aem.live/drafts/ramuntea/ace1205/ace1205-offer?milolibs=ace1205-to-stage

Product
* ACE1205: https://main--da-dc--adobecom.aem.live/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-product
* libs/c2/blocks: https://main--da-dc--adobecom.aem.live/drafts/ramuntea/ace1205/ace1205-product?milolibs=ace1205-to-stage

Plans
* ACE1205: https://main--da-dc--adobecom.aem.live/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-plans
* libs/c2/blocks: https://main--da-dc--adobecom.aem.live/drafts/ramuntea/ace1205/ace1205-plans?milolibs=ace1205-to-stage

Homepage
* ACE1205: https://main--upp--adobecom.aem.live/homepage/index-loggedout
[...]

PR #6347, opened 2026-07-24
[MWPW-201841] - Fix comparison-table-c2 sticky header covered by gnav; 

DKos95:
The sticky `.header-cards-container` computed its top offset from the nav element inside header, but that inner nav isn't the sticky one — the outer header element is. This made `--ct-nav-height` always resolve to `0px`, so the comparison table header stacked underneath (and got hidden by) the sticky global nav on scroll.

Replaced the hand-rolled offset calculation with the shared `getGnavHeight()` utility (`libs/blocks/global-navigation/utilities/utilities.js`), already used by `table.js` and `sticky-section.js` for the same purpose. This also picks up correct handling for the mobile local-nav and promo-banner cases that the previous logic didn't account for.

**Resolves:** [MWPW-201841](https://jira.corp.adobe.com/browse/MWPW-201841)

**Test URLs:**

- Before: https://main--da-dc--adobecom.aem.live/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-plans?milolibs=stage
- After: https://main--da-dc--adobecom.aem.live/dc-shared/fragments/tests/2026/q2/ace1205/fragments/ace1205-plans?milolibs=c2-comparison-table-header

🤖 Generated with Claude Code

### DIFF
--- a/libs/mep/ace1205/comparison-table-c2/comparison-table-c2.js
+++ b/libs/mep/ace1205/comparison-table-c2/comparison-table-c2.js
@@ -1,5 +1,6 @@
 import { createTag, getConfig, loadStyle } from '../../../utils/utils.js';
 import { getMetadata as getSectionMetadata } from '../section-metadata/section-metadata.js';
+import { getGnavHeight } from '../../../blocks/global-navigation/utilities/utilities.js';
 
 const COLUMN_TYPES = { PRIMARY: 'primary' };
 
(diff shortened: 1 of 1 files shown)


zagi25 (approved):
Should this change also be part of the https://github.com/adobecom/milo/pull/6332 ? Or later added via PR to stage ?

SilviuLCF (approved):
QA verified, testing details https://jira.corp.adobe.com/browse/MWPW-201841

narcis-radu (approved):

PR #6352, opened 2026-07-27
MWPW-201354: add unit tests for c2 router-marquee block

SilviuLCF:
## Summary
- Adds unit tests for the `router-marquee` c2 block (`libs/c2/blocks/router-marquee`), which previously had no coverage in `test/blocks`
- Covers: per-breakpoint viewport building with clone fallback, controls + card tablist, slide region decoration with a11y state (`is-active`/`aria-hidden`/`inert`), title/eyebrow/body/CTA promotion, nav card build (federated icon, label, `aria-label`, href), block/slide/card analytics attributes, and `starting-marquee` slide reorder
- Follows conventions from existing block tests (e.g. `test/blocks/carousel`, `test/blocks/elastic-carousel`)

Scope note: covers deterministic DOM decoration; the autoplay/timer/swipe/IntersectionObserver/hover behaviors are intentionally not asserted (timer- and layout-driven, flaky in unit tests), consistent with the existing carousel tests.

Resolves: https://jira.corp.adobe.com/browse/MWPW-201354

## Test plan
- [x] `npx wtr --config ./web-test-runner.config.mjs "./test/blocks/router-marquee/router-marquee.test.js" --node-resolve --port=2000` — 7/7 passing
- [x] `npx eslint test/blocks/router-marquee/router-marquee.test.js` — clean

### DIFF
--- a/test/blocks/router-marquee/mocks/default.html
+++ b/test/blocks/router-marquee/mocks/default.html
@@ -1,0 +1,35 @@
+<main>
+  <div class="section">
+    <div class="router-marquee">
+      <div>
+        <div>mobile</div>
+      </div>
+      <div>
+        <div>
+          <p>Eyebrow one</p>
+          <h2>Slide one title</h2>
+          <p>Body copy for slide one.</p>
+          <p><em><strong><a href="https://www.adobe.com/buy">Buy now</a></strong></em> <em><a href="https://www.adobe.com/learn">Learn more</a></em></p>
+          <p><img src="/federal/icons/card-one.svg" alt="card one icon" /></p>
+          <p><a href="https://www.adobe.com/slide-one">Slide one</a></p>
+        </div>
+        <div>
+          <picture><img src="" alt="background one" /></picture>
+        </div>
+      </div>
+      <div>
+        <div>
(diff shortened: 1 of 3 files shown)


zagi25 (approved):
Approving — assertions all trace correctly against the block source (per-viewport build, controls/a11y decoration, nav-card federated icon URL, analytics attributes, starting-marquee reorder). Two non-blocking suggestions for a follow-up, not needed to merge:

1. **Teardown.** `init()` → `startAutoplay()` wires a `setTimeout(advance, 5000)`, a `requestAnimationFrame`, `resize`/`mouseover`/`focusin`/`touch` listeners, and Intersection/Mutation observers that are never disconnected, and there's no `afterEach`. Benign today (timers fire against detached DOM), but a flakiness risk as the suite grows. An `afterEach` that clears the body and disconnects the observers would tighten isolation.
2. **Fragile assertion.** `expect(body.textContent).to.not.contain('Slide one')` passes partly by case-coincidence (lowercase "one" still remains in the body copy). Asserting on the specific element/label would be more robust.

Neither blocks — approving.

narcis-radu (approved):

PR #6354, opened 2026-07-27
MWPW-201350: add unit tests for c2 modal block

SilviuLCF:
## Summary
- Adds unit tests for the `modal` c2 block (`libs/c2/blocks/modal`), which previously had no coverage in `test/blocks`
- Covers: `getHashParams` parsing, `findDetails` resolution (passed anchor + DOM lookup), `delayedModal` desktop/mobile behavior, custom-content `getModal` (dialog role/aria/class/content, close button `daa-ll`, curtain, scroll lock), `closeModal` teardown with `closeCallback`, and the `init` hash-mismatch guard
- Follows conventions from the existing non-c2 `test/blocks/modals` suite

Scope note: focuses on the custom-content path and pure helpers; the fragment-fetch path (`getPathModal`) and module-level `hashchange` orchestration are intentionally not asserted (network/history-driven, flaky in unit tests), mirroring how the existing `modals` suite exercises custom modals.

Resolves: https://jira.corp.adobe.com/browse/MWPW-201350

## Test plan
- [x] `npx wtr --config ./web-test-runner.config.mjs "./test/blocks/modal/modal.test.js" --node-resolve --port=2000` — 9/9 passing
- [x] `npx eslint test/blocks/modal/modal.test.js` — clean

### DIFF
--- a/test/blocks/modal/mocks/body.html
+++ b/test/blocks/modal/mocks/body.html
@@ -1,0 +1,4 @@
+<a id="c2-modal-link" href="https://milo.adobe.com/path/foo"
+   data-modal-path="/path/foo"
+   data-modal-hash="#foo"
+   aria-label="Foo label">Foo</a>
--- a/test/blocks/modal/modal.test.js
+++ b/test/blocks/modal/modal.test.js
@@ -1,0 +1,128 @@
+/* eslint-disable no-underscore-dangle */
+import { readFile, setViewport } from '@web/test-runner-commands';
+import { expect } from '@esm-bundle/chai';
+import sinon from 'sinon';
+import { waitForRemoval } from '../../helpers/waitfor.js';
+import { setConfig, createTag } from '../../../libs/utils/utils.js';
+
+setConfig({ codeRoot: '/libs' });
+document.body.innerHTML = await readFile({ path: './mocks/body.html' });
+
+const {
+  default: init,
+  getModal,
+  findDetails,
+  getHashParams,
+  delayedModal,
+} = await import('../../../libs/c2/blocks/modal/modal.js');
+
+const cleanup = () => {
+  document.querySelectorAll('.dialog-modal, .modal-curtain').forEach((n) => n.remove());
+  document.documentElement.classList.remove('disable-scroll');
+  window.location.hash = '';
+};
+
+describe('Modal (c2)', () => {
+  afterEach(() => {
+    sinon.restore();
+    cleanup();
+  });
+
+  describe('getHashParams', () => {
+    it('returns an empty object for empty input', () => {
+      expect(getHashParams()).to.deep.equal({});
+    });
+
+    it('parses the hash and delay (seconds to ms)', () => {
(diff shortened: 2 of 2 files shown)


zagi25 (approved):
Approving — clean addition-only coverage for the c2 modal block. Traced the assertions against `modal.js` and they match actual behavior (getHashParams ms conversion, findDetails, delayedModal viewport gating, custom-content getModal a11y + close analytics, closeCallback teardown, the init hash guard). No blockers — thanks for adding this.

mokimo (approved):
## PR #6354 — MWPW-201350: add unit tests for c2 modal block
**Size:** +132/-0 across 2 files. Base: `stage`. Oversize: N.

Test-only addition (no `libs/` change), so no CODEOWNERS gap. Verified every assertion in `test/blocks/modal/modal.test.js` against `libs/c2/blocks/modal/modal.js`:

- `getHashParams('#delayed-modal:delay=2')` → `{hash, delay: 2000}` matches the `:`-split/`delay=`×1000 logic exactly.
- `findDetails` — both the passed-anchor and DOM-lookup (`a[data-modal-hash="#foo"]`) paths resolve `id`/`path`/`title` correctly against the new `mocks/body.html` fixture.
- `delayedModal` — the desktop-breakpoint test pre-seeds `sessionStorage.setItem('shown:#delayed', ...)` specifically so the internal `setTimeout(() => window.location.replace(...))` branch is skipped, avoiding an actual navigation inside the unit test. Nice touch — the returned `true`/rewritten `href`/`dataset.modalHash` assertions still exercise the real desktop/mobile branching.
- `getModal` (custom-content path) — dialog `role`/`aria-modal`/`aria-label`/class/content, `close` button `daa-ll` (via `contain`, robust to the `analyticsEventName` prefix), curtain insertion, and `disable-scroll` all line up with `getCustomModal`/`getModal`.
- `closeModal` teardown — `closeCallback` invocation, `#c2-close` removal, curtain removal, and `disable-scroll` cleanup are asserted via `waitForRemoval`, matching `closeModal`'s async cleanup order.
- `init` hash-mismatch guard returns `null` as implemented (`window.location.hash !== modalHash`).

Structure follows the existing `test/blocks/modals/modals.test.js` conventions (`readFile` mock body, `sinon`, `waitfor` helpers). Scope note in the description (skipping `getPathModal`/module-level `hashchange`) is reasonable — those are network/history-driven and already excluded by the sibling suite for the same reason.

### Blockers
- None.

### Suggestions
- `test/blocks/modal/modal.test.js:44` — `expect(details.isHash).to.equal(window.location.hash === '#foo')` re-derives the exact same expression the source computes (`hash === window.location.hash`) rather than asserting a concrete expected boolean. It still happens to catch a hardcoded-`true` regression (since `location.hash` is `''` at that point), but a literal `.to.be.false` (with a comment on why) would make the intent obvious without mirroring the implementation.

### Nice-to-haves
- None beyond the above.

### Test plan
- PR description includes the exact `wtr` invocation and result (9/9 passing) plus a clean `eslint` run — sufficient for a test-only change; no live test URL needed since there's no runtime/UI behavior change.

PR #6357, opened 2026-07-27
MWPW-201351: add unit tests for c2 modal-metadata block

SilviuLCF:
## Summary
- Adds unit tests for the `modal-metadata` c2 block (`libs/c2/blocks/modal-metadata`), which previously had no coverage in `test/blocks`
- The c2 block is a pass-through re-export of the shared `libs/blocks/modal-metadata` init, so these tests also cover that previously-untested shared logic
- Covers: the no-modal guard, `style` class application, `curtain: off` toggle, the `tall-video` background guard, and background picture handling for tall-video modals
- Follows conventions from existing block tests (e.g. `test/blocks/section-metadata`)

Resolves: https://jira.corp.adobe.com/browse/MWPW-201351

## Test plan
- [x] `npx wtr --config ./web-test-runner.config.mjs "./test/blocks/modal-metadata/modal-metadata.test.js" --node-resolve --port=2000` — 6/6 passing
- [x] `npx eslint test/blocks/modal-metadata/modal-metadata.test.js` — clean

### DIFF
--- a/test/blocks/modal-metadata/mocks/default.html
+++ b/test/blocks/modal-metadata/mocks/default.html
@@ -1,0 +1,18 @@
+<main>
+  <div class="dialog-modal" id="test-modal">
+    <div class="modal-metadata">
+      <div>
+        <div>style</div>
+        <div>center, grid width 2</div>
+      </div>
+      <div>
+        <div>curtain</div>
+        <div>off</div>
+      </div>
+      <div>
+        <div>background</div>
+        <div><picture><img src="" alt="background" /></picture></div>
+      </div>
+    </div>
+  </div>
+</main>
--- a/test/blocks/modal-metadata/mocks/no-modal.html
+++ b/test/blocks/modal-metadata/mocks/no-modal.html
@@ -1,0 +1,14 @@
+<main>
+  <div class="section">
+    <div class="modal-metadata">
+      <div>
+        <div>style</div>
+        <div>center</div>
+      </div>
+      <div>
+        <div>curtain</div>
+        <div>off</div>
+      </div>
+    </div>
+  </div>
+</main>
--- a/test/blocks/modal-metadata/mocks/tall-video.html
+++ b/test/blocks/modal-metadata/mocks/tall-video.html
@@ -1,0 +1,10 @@
+<main>
+  <div class="dialog-modal tall-video" id="tall-video-modal">
+    <div class="modal-metadata">
+      <div>
+        <div>background</div>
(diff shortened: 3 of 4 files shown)


mokimo (approved):
## PR #6357 — MWPW-201351: add unit tests for c2 modal-metadata block
**Size:** +105/-0 across 4 files (1 test + 3 HTML mocks). Base: `stage`. Not oversize.

Test-only change (`test/blocks/modal-metadata/`), no source touched — no CODEOWNERS/registration implications.

I traced each assertion against the actual implementation (`libs/blocks/modal-metadata/modal-metadata.js`, re-exported as-is by `libs/c2/blocks/modal-metadata/modal-metadata.js`, plus `handleStyle`/`handleBackground`/`getMetadata` in `libs/blocks/section-metadata/section-metadata.js`) rather than just reading the test file in isolation:

- `no-modal.html` — block isn't under `.dialog-modal`, so `init` returns at the `el.closest('.dialog-modal')` guard (line 5) before touching the DOM. Test correctly asserts nothing changes.
- `default.html` style row `"center, grid width 2"` → `handleStyle` splits on `', '` then replaces spaces with `-`, yielding classes `center` and `grid-width-2`. Matches the test's exact expected class names (a case that's easy to get wrong if you guess at the naming instead of reading `handleStyle`).
- `curtain: off` → `metadata.curtain?.text === 'off'` adds `curtain-off`; correctly asserted both present (default.html) and absent (tall-video.html, which has no curtain row).
- Background is only ever applied inside the `modal.classList.contains('tall-video')` branch — the `default.html` case (no `tall-video` class) correctly asserts no `has-background`/`section-background` picture, and `tall-video.html` correctly asserts both appear via `handleBackground` moving the `<picture>` into the modal.

Follows the existing `test/blocks/section-metadata` convention (`readFile` + mocks/*.html), CI is green (unit, lint, Nala all pass), and this closes a real gap — `modal-metadata` had zero coverage despite being referenced from `card`, `faas-modal`, `marquee`, `quote`, etc.

### Blockers
None.

### Suggestions
None — small enough that no simplification is warranted.

### Nice-to-haves
- Could add a mock/case for a `tall-video` modal with a background *color* (not just a picture) to exercise `applyBackground`'s color path in `section-metadata.js`, but that's arguably section-metadata's own test responsibility, not this block's.

### Test plan
- PR description shows local `wtr` run (6/6 passing) and clean eslint — consistent with the green CI here.

Approving.

zagi25 (approved):

PR #6358, opened 2026-07-27
MWPW-201355: add unit tests for c2 visually-hidden block

SilviuLCF:
## Summary
- Adds a smoke test for the `visually-hidden` c2 block (`libs/c2/blocks/visually-hidden`), which previously had no coverage in `test/blocks`
- This is a CSS-only block: its JS `init` is an intentional no-op (all styling lives in `visually-hidden.css`), so the test verifies `init` is a callable no-op that returns `undefined` and leaves the block markup untouched
- Provides line coverage for the block and a test folder consistent with the other c2 blocks

Resolves: https://jira.corp.adobe.com/browse/MWPW-201355

## Test plan
- [x] `npx wtr --config ./web-test-runner.config.mjs "./test/blocks/visually-hidden/visually-hidden.test.js" --node-resolve --port=2000` — 3/3 passing
- [x] `npx eslint test/blocks/visually-hidden/visually-hidden.test.js` — clean

### DIFF
--- a/test/blocks/visually-hidden/mocks/default.html
+++ b/test/blocks/visually-hidden/mocks/default.html
@@ -1,0 +1,7 @@
+<main>
+  <div class="section">
+    <div class="visually-hidden">
+      <div><p>Screen-reader only text</p></div>
+    </div>
+  </div>
+</main>
--- a/test/blocks/visually-hidden/visually-hidden.test.js
+++ b/test/blocks/visually-hidden/visually-hidden.test.js
@@ -1,0 +1,25 @@
+import { readFile } from '@web/test-runner-commands';
+import { expect } from '@esm-bundle/chai';
+
+import init from '../../../libs/c2/blocks/visually-hidden/visually-hidden.js';
+
+// visually-hidden is a CSS-only block; its JS init is an intentional no-op.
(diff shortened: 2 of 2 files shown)


mokimo (approved):
Reviewed per Milo conventions (CODEOWNERS, c2 structure, size, simplification).

**Size:** +32/-0 across 2 files (both new). Base: `stage`. Not oversize.

Verified against the actual block source (`libs/c2/blocks/visually-hidden/visually-hidden.js` is literally `export default function init() {}`) — the test's claim that `init` is an intentional no-op is accurate, not just asserted in the PR description.

### Blockers
None.

### Notes
- No new `libs/` directory added (test-only change against an existing c2 block), so no CODEOWNERS update is needed.
- Mock/test structure (`mocks/default.html` + `readFile` + `@esm-bundle/chai`) matches the existing convention used by other block tests (e.g. `test/blocks/aside`).
- Closes a real coverage gap — `visually-hidden` previously had zero test coverage under `test/blocks`.

### Nice-to-haves
- Could fold the "exports a function" and "no-op returns undefined" assertions into one `it`, but three focused assertions read fine as-is — not worth churning.

LGTM, approving.

zagi25 (approved):

PR #6360, opened 2026-07-27
MWPW-202052: fix c2 news crash on standalone link in non-quiet variant

SilviuLCF:
## Problem
`libs/c2/blocks/news/news.js:47` passed an empty-string token to `classList.add` for **non-quiet** blocks:

```js
linkEl.classList.add('standalone-link', 'label', `${el.classList.contains('quiet') ? 'quiet' : ''}`);
```

`DOMTokenList.add('')` throws `SyntaxError: The token provided must not be empty`. Since `quiet` is an opt-in modifier, a plain `news` block is **non-quiet by default**, so any default block containing a link-only (standalone) paragraph crashes mid-`init`, halting decoration for that block. The non-quiet standalone link is a designed, styled state (`libs/c2/styles/styles.css:613-643` styles base `a.standalone-link` independently of `quiet`).

## Fix
```js
linkEl.classList.add('standalone-link', 'label');
if (el.classList.contains('quiet')) linkEl.classList.add('quiet');
```

## How to test
- Non-quiet `news` block with a link-only item (`<p><a href="...">Read more</a></p>`): no error thrown, link gets `standalone-link label` (underlined + arrow), all following items decorate correctly.
- Regression — `news quiet` block with a link-only item: link gets `standalone-link label quiet` (no underline), unchanged from before.
- Rest of the block (headline/eyebrow, svg icon federated rewrite, `news-items` up-count, headline/body classification) unaffected.

Existing authored usage (e.g. `https://main--upp--adobecom.aem.page/homepage/drafts/msale/2026/sr/06-18-sr-ff-ps-index-loggedout`) uses the `news quiet` variant, which is why the crash has been latent — the non-quiet default path is the one that was broken.

## Notes
- Fix scoped to the canonical block. The same buggy line is duplicated in MEP experiment copies (`libs/mep/ace1201/news/news.js:47`, `libs/mep/ace1205/news/news.js:47`) — flagged in the ticket for a separate follow-up to avoid touching running experiments here.
- Unit tests for the block will be added in a follow-up PR (MWPW-201352) after this fix is deployed.

Resolves: https://jira.corp.adobe.com/browse/MWPW-202052

TEST URL for REGRESSION https://main--upp--adobecom.aem.page/homepage/drafts/msale/2026/sr/06-18-sr-ff-ps-index-loggedout?milolibs=news-standalone-link-non-quiet 

### DIFF
--- a/libs/c2/blocks/news/news.js
+++ b/libs/c2/blocks/news/news.js
@@ -44,7 +44,8 @@ export default async function init(el) {
       if (indx === 0) content.classList.add('news-item-headline');
       else if (isLinkOnlyContent(content, linkEl)) {
         content.classList.add('news-item-link');
-        linkEl.classList.add('standalone-link', 'label', `${el.classList.contains('quiet') ? 'quiet' : ''}`);
+        linkEl.classList.add('standalone-link', 'label');
+        if (el.classList.contains('quiet')) linkEl.classList.add('quiet');
       } else content.classList.add('news-item-body');
     });
   });


mokimo (approved):
Reviewed per Milo PR conventions (size, base branch, c2/blocks placement, CODEOWNERS, simplification, test plan).

**Root cause verified.** `DOMTokenList.add()` throws `SyntaxError` on an empty-string token per the DOM spec, and the non-quiet default path does pass `''` (the template-literal ternary `${el.classList.contains('quiet') ? 'quiet' : ''}` at the old `news.js:47`) — confirmed by reading the file directly. Since `quiet` is opt-in, any default (non-quiet) `news` block with a link-only paragraph would hit `init` and throw, halting decoration for that block instance. That matches the PR's own hypothesis/evidence.

**Fix is correct and minimal:** conditionally adding the `quiet` class only when present, instead of passing an empty-string token, is the smallest possible change (+2/-1, one file) and preserves the existing quiet-variant behavior exactly.

**Convention checks:**
- Base branch: `stage` — correct.
- Size: 1 file, +2/-1 — not oversize.
- No new directory → no CODEOWNERS entry needed.
- Already in `libs/c2/blocks/` (correct c2 location, unchanged).
- No CSS/dependency/console.log changes.
- Test URL provided for both the fix and the quiet-variant regression case.
- CI green across all suites (MILO/MAS/MEP, all three browsers, eslint, CodeQL).

**Note (non-blocking, already flagged in the PR description):** the same buggy line is duplicated in `libs/mep/ace1201/news/news.js:47` and `libs/mep/ace1205/news/news.js:47`. Intentionally left untouched here to avoid touching running MEP experiments — correct call for this PR's scope, but worth confirming a follow-up ticket actually exists for those two copies.

Approving.

zagi25 (approved):

narcis-radu (approved):

PR #6364, opened 2026-07-28
MWPW-202151: validate maslibs param (VULN-36379)

Axelcureno:
Fixes the milo side of [VULN-36379](https://jira.corp.adobe.com/browse/VULN-36379) (HackerOne #3786786): `?maslibs=` was interpolated unvalidated into remote-script origins.

- Add `getValidatedMasLibsUrl()` in utils: `branch` / `branch--repo` / `branch--repo--owner` pattern + `.aem.live` hostname assertion; invalid → null → default origin
- `getMasLibsBaseUrl()` delegates to it and returns null on www.adobe.com (prod guard: maslibs only works on non-prod hosts)
- `getMasDepUrl()` refactored onto the validator; dead `getMasBase()` deleted
- ost.js `getMasLibsBase()` falls back to mas.adobe.com on invalid input instead of throwing
- MAS-side fix is a separate PR on adobecom/mas (same branch name); the mas bundle bump into `libs/deps/mas` lands separately via release-mas-js after that PR merges

Verify: on the After URL append `?maslibs=evil.com` — the network tab shows no request to evil.com and the default bundle loads. `?maslibs=<branch>` on aem.live still loads the branch dist.

Resolves: [MWPW-202151](https://jira.corp.adobe.com/browse/MWPW-202151)

**Test URLs:**
- After: https://mwpw-202151--milo--adobecom.aem.page/?martech=off


seanchoi0 (approved):
Clean fix — validation logic is sound and closes the vuln correctly. The URL encoding, protocol manipulation, and injection paths are all blocked. CI is all green. Approving, with a couple of maintenance notes worth considering:

- Both test mocks (`embed-utils.js`, `ost-utils.js`) copy `getValidatedMasLibsUrl` verbatim rather than importing it from `utils.js`. If the function needs a security patch down the line, the mocks will drift and tests will keep passing against the stale copy. Worth switching to a direct import.

- Milo and MAS now each have their own copy of `getValidatedMasLibsUrl` that diverge on one thing: MAS accepts an `extension` param and passes `'page'` when on `.aem.page` hosts; Milo hardcodes `.aem.live`. No breakage today since these are separate load paths, but worth aligning if they ever converge.

test/blocks/merch/mocks/embed-utils.js
```diff
@@ -58,3 +58,24 @@ export const shouldAllowKrTrial = stub();
 export const lingoActive = () => false;
 export const getGeoLocalePrefix = () => Promise.resolve(null);
 export const getPlaceholderPaths = () => [];
+
+const MASLIBS_PATTERN = /^([a-z0-9]+(-[a-z0-9]+)*)(--([a-z0-9]+(-[a-z0-9]+)*)){0,2}$/;
+const MASLIBS_MAX_LENGTH = 100;
+
+/** Mirrors getValidatedMasLibsUrl in libs/utils/utils.js (VULN-36379). */
+export function getValidatedMasLibsUrl(masLibs) {
```
zagi25, line 66: Low / non-blocking: the validator body is now hand-copied into two test mocks (`test/blocks/merch/mocks/embed-utils.js` and `test/blocks/ost/mocks/ost-utils.js`) in addition to the real `libs/utils/utils.js`. The "Mirrors getValidatedMasLibsUrl" comment is honest, but if the real validator is ever tightened (e.g. the regex or segment count), these copies won't follow and the block tests would keep passing against stale logic while only `getValidatedMasLibsUrl.test.js` catches the change. Since this is a pure, side-effect-free function, the mocks could re-export the real implementation instead of duplicating it. Fine to keep as-is if it matches existing mock conventions here.

zagi25 (approved):
Solid, well-scoped security fix. The single `getValidatedMasLibsUrl()` validator is the right shape, and the whole caller chain degrades safely:

- **Host-escape is genuinely closed.** The charset allowlist (`[a-z0-9]` + `-` with `--` separators, ≤3 segments) means no `.`, `/`, `@`, `:`, `#`, or `%` can reach the interpolated host, so `https://<value>.aem.live` can only ever resolve to an `aem.live` subdomain. The `new URL(...)` parse + `endsWith('.aem.live')` check is good defense-in-depth on top of that.
- **No ReDoS.** The length cap is evaluated before the regex via `||` short-circuit, so overlong input never reaches the pattern; the pattern itself has disjoint boundaries (alphanumeric vs `-`), so no catastrophic backtracking either.
- **Prod guard** on `getMasLibsBaseUrl()` (ignore maslibs on `www.adobe.com`) closes the exact vector called out in the ticket.
- **Callers handled correctly:** `getMasLibs()`/`getFragmentClientUrl()` already null-check, `getMasDepUrl()` falls back to `main--mas--adobecom`, and `ost.js getMasLibsBase()`'s new `?.`/`??` fallback prevents the throw that deleting the old happy-path would otherwise cause.
- **Dead-code removal is safe:** `getMasBase()` has no remaining callers anywhere in the repo besides the test import this PR also removes.
- **Tests** cover the hostile-payload matrix, malformed shapes, punycode, overlong values, and the prod guard across utils/merch/ost.

Scope note (not a gap in this PR): the ticket also mentions `mas-io-url` and the MAS-side `getFragmentClientUrl()` — per the PR description those land in a separate `adobecom/mas` PR, so they're correctly out of scope here.

Nits below are low / optional.

_Review by Claude._

mirafedas (approved):

PR #6374, opened 2026-07-30
MWPW-202190: add unit tests for c2 faq block

SilviuLCF:
## Summary
- Adds unit tests for the `faq` c2 block (`libs/c2/blocks/faq`), which previously had no coverage in `test/blocks`
- Covers: accordion build (one `details.faq-item` per question row inside `.faq-list`), first-item open state + `aria-expanded`, trigger/panel layout (`.faq-question.heading-5` + `.faq-icon`, `.faq-panel.body-md > .faq-panel-inner`), open/close `daa-ll` analytics labels, heading-less row skipping, `toggle` behavior (aria-expanded + label prefix swap), entry-animation classes, and the `.seo` `FAQPage` ld+json (present when `.seo`, absent otherwise)
- Includes a regression guard that authored answer text stays inert inside the ld+json `<script>` (no `</script>` breakout)
- Follows conventions from existing block tests (e.g. `test/blocks/brand-concierge`, `test/blocks/base-card`)

Resolves: https://jira.corp.adobe.com/browse/MWPW-202190

## Test plan
- [x] `npx wtr --config ./web-test-runner.config.mjs "./test/blocks/faq/faq.test.js" --node-resolve --port=2000` — 12/12 passing
- [x] `npx eslint test/blocks/faq/faq.test.js` — clean


test/blocks/faq/faq.test.js
```diff
@@ -0,0 +1,152 @@
...
+    const block = document.querySelector('.faq');
+    init(block);
+
+    const triggers = block.querySelectorAll('.faq-trigger');
+    // first is open on load, so its label is prefixed close- (the action if clicked)
+    expect(triggers[0].getAttribute('daa-ll')).to.match(/^close-1--/);
+    expect(triggers[1].getAttribute('daa-ll')).to.match(/^open-2--/);
+  });
+
+  it('skips rows that have no heading', async () => {
```
zagi25, line 74: Low / non-blocking: consider a case where **every** row lacks a heading, asserting the resulting `.faq-list` is empty — currently only a mixed valid/invalid case is covered, so the all-invalid branch of the `rows.filter(...)` / `items.filter(Boolean)` logic isn't exercised.

zagi25 (approved):
**Code review — MWPW-202190: unit tests for c2 faq block**

Test-only PR matching the ticket's scope (follows the brand-concierge pattern; no production code touched). The reviewer validated the tests by actually running them against the real block source (checked out `faq.js` at its `stage` ancestor, applied this diff, ran `wtr` → **10/10 pass**).

- Assertions are meaningful — structure, `open`/`aria-expanded` state, `daa-ll` tracking-label prefixes, toggle behavior, multi-paragraph answers, SEO on/off gating — not no-throw checks. Mocks model the real block›row›cell DOM shape, nothing is over-mocked, and `afterEach(removeLd)` correctly cleans the `document.head` ld+json between tests.
- **`seo-xss.html` is good, non-tautological security coverage.** Verified empirically: `setSEO()` builds the ld+json via `createTag('script', …, JSON.stringify(schema))` → `insertAdjacentHTML('beforeend', …)` on an already-created `<script>` element, which does **not** trigger the `</script>` tokenizer breakout — the payload (literal `</script>` + crafted ``) lands as inert text, no `` inserted, JSON parses intact. The test documents real, verified-safe behavior rather than masking a bug.

FYI (out of diff scope, not blocking): `faq.js` keeps `SEO_SCHEMA` as a module-level mutable singleton — safe only because `setSEO()` is fully synchronous; a latent footgun if it ever goes async. Worth a future ticket, not this one.

One low coverage nit inline.

_Review by Claude._

SilviuLCF:
Thanks for the review! Both findings addressed in 81a95d12d (test-only, no production change):

- **All-invalid coverage nit** — added `builds an empty list when no row has a heading` (mock `no-headings.html`): asserts the `.faq-list` is present with 0 `.faq-item`, exercising the all-invalid branch of the `rows.filter(...)` / `items.filter(Boolean)` logic.
- **`SEO_SCHEMA` module-level singleton** — added `emits an independent schema per block for multiple seo blocks` (mock `seo-multi.html`): inits two `.seo` blocks (2 vs 3 questions) and asserts each emitted ld+json contains only its own questions, guarding against state leaking through the shared singleton (would catch e.g. a switch from reassigning to accumulating `mainEntity`).

Now 12/12 passing, eslint clean. The underlying suggestion to make `SEO_SCHEMA` non-shared in the source is still worth a follow-up ticket (out of scope for this test-only PR).

mokimo (approved):
LGTM — and I want to specifically call out the security regression test as good work, not just a check-box.

Verified by running the suite for real (not just reading it): 12/12 passing against the unmodified `libs/c2/blocks/faq/faq.js`.

The `'keeps the ld+json payload inert (no </script> breakout)'` test is asserting on a genuinely real risk — `setSEO()` builds the FAQPage JSON-LD from author-supplied FAQ answer text via `createTag('script', ..., JSON.stringify(SEO_SCHEMA))`, and `createTag`'s string-content branch uses `el.insertAdjacentHTML('beforeend', html)` (`libs/utils/utils.js`), not `textContent`. `JSON.stringify` doesn't escape `<`/`/`, so a literal `</script>` in authored answer text is a legitimate thing to worry about. I independently reproduced the exact scenario in a real Chromium instance (Playwright) using the real `createTag` implementation and the real entity-decoded payload from `seo-xss.html`, and confirmed empirically that `insertAdjacentHTML('beforeend', ...)` on a **detached** `<script>` element does *not* trigger script-data-state breakout in Chromium — the JSON lands as inert text, the test's `ld.childElementCount === 0` and `document.querySelector('img') === null` assertions hold for real, not just on paper. Good instinct to write this as a locked-in regression guard rather than trusting it silently.

Also notably solid: the dedicated test for the shared module-level `SEO_SCHEMA` singleton (`emits an independent schema per block for multiple seo blocks`, using `seo-multi.html`) — this guards against a real state-leak class of bug (the object is mutated and reused across `setSEO()` calls), and the `afterEach(removeLd)` cleanup correctly prevents `<script>` tags from a previous test in the file leaking into a later one's `document.head.querySelector(...)` assertions. That's the kind of test-hygiene detail that's easy to miss and would have caused flaky failures.

One non-blocking nit: PR description says "10/10 passing," actual file has 12 `it(...)` blocks (confirmed 12/12 by the real run). Worth a quick description update, same as a couple of the sibling PRs in this batch.

SilviuLCF:
Thank you for review, updated also the description

PR #6375, opened 2026-07-30
MWPW-202189: add unit tests for c2 comparison-table-c2 block

SilviuLCF:
## Summary
- Adds unit tests for the `comparison-table-c2` c2 block (`libs/c2/blocks/comparison-table-c2`), which previously had no coverage in `test/blocks`
- Covers: `con-block`/header decoration, header cards + `--ct-card-count`, primary column & cell marking, table shell (`role=table`, toggle `button[aria-expanded]`, `.accessibility-header-row`), row header/cell decoration (`role=rowheader`/`role=cell`, `data-column-index`, `.cell-content`), empty (`-`) cell marking, header toggle open/close behavior, underline tooltip triggers (`label|pos|text` → `.milo-tooltip` + `.ct-tooltip-content`; plain `<u>` → `span.dotted-underline`), and `+++` table-group splitting
- Follows conventions from existing block tests (e.g. `test/blocks/table`, `test/blocks/base-card`)

Scope note: covers the deterministic DOM decoration (assertions verified against the real decorated output first). The scroll-collapse, responsive column-hiding, and async placeholder/tooltip-loading behaviors are intentionally not asserted (timer/observer/matchMedia/network-driven, flaky in unit tests), consistent with the other c2 suites.

Resolves: https://jira.corp.adobe.com/browse/MWPW-202189

## Test plan
- [x] `npx wtr --config ./web-test-runner.config.mjs "./test/blocks/comparison-table-c2/comparison-table-c2.test.js" --node-resolve --port=2000` — 10/10 passing
- [x] `npx eslint test/blocks/comparison-table-c2/comparison-table-c2.test.js` — clean


test/blocks/comparison-table-c2/comparison-table-c2.test.js
```diff
@@ -0,0 +1,135 @@
+import { readFile } from '@web/test-runner-commands';
+import { expect } from '@esm-bundle/chai';
+import { setConfig } from '../../../libs/utils/utils.js';
+
+import init from '../../../libs/c2/blocks/comparison-table-c2/comparison-table-c2.js';
+
+setConfig({
+  locales: { '': { ietf: 'en-US', tk: 'hah7vzn.css' } },
+  contentRoot: '/test/blocks/comparison-table-c2/mocks',
```
zagi25, line 10: Low / non-blocking: `init()` triggers `setAccessibilityLabels()`, which fetches `${contentRoot}/placeholders.json` (via `replaceKeyArray`). No such file exists under `mocks/`, so it resolves through an unmocked 404 → empty-data fallback in `placeholders.js` — harmless (tests don't flake), but it's an unmocked background network call, and the aria-label / `sr-only` empty-cell text it adds is never asserted. Consider a minimal `mocks/placeholders.json` (or stub `fetch`) if you want deterministic, asserted coverage of that path.

zagi25 (approved):
**Code review — MWPW-202189: unit tests for c2 comparison-table-c2 block**

Test-only PR (3 new files, no source edits) that fulfils the ticket. I traced all 8 tests against the real 646-line block source (`libs/c2/blocks/comparison-table-c2/comparison-table-c2.js` on `stage`, fetched via the API since the local worktree is on another branch):

- Every assertion is genuinely behavioral — card count, `data-column-index` values, primary-column detection (`primary` marker row), empty-cell `-` sentinel, tooltip attribute parsing, and the toggle click round-trip all match what the real code produces; none are tautological or mirroring an over-mocked stub. Mocks are authored in the block's real DOM shape (`-`, `+++`, `primary`), not contrived to force a pass.

Low-severity, non-blocking notes:

- **Coverage gaps** (not required by the ticket, worth a follow-up): `setupResponsiveHiding` (mobile column hide/reorder), `setupCollapsingHeader` (scroll collapse), the `section-metadata` `expand` values (`'all'` / index list) in `isExpandedSection`, and the `headerItemsCount > 3` mobile filter-select path are all untested branches.
- **No listener teardown**: each `it()` calls `init()` on a fresh element, re-registering `matchMedia`/`scroll`/`resize` listeners on `window` with no cleanup — but this matches the existing `brand-concierge` test convention (the pattern the ticket names), so it's not a regression.

One unmocked-fetch note inline.

_Review by Claude._

SilviuLCF:
Thanks for the review! Addressed the inline note (and the related filter-select coverage gap) in 2c528bcfe (test-only, no source change):

- **Unmocked `placeholders.json` fetch** — added `mocks/placeholders.json` (`choose-table-column`, `empty-table-cell`), so `setAccessibilityLabels` resolves deterministically instead of through a 404 fallback. Added `adds placeholder-driven screen-reader text to empty cells`, which waits for and asserts the empty cell's `.sr-only` text (`"Not included"`).
- **`headerItemsCount > 3` filter-select path** (from the coverage-gaps note) — added `mocks/three-columns.html` and `builds a labelled mobile filter select per column when there are 3+ columns`, asserting one `.mobile-filter-select` per column and their async placeholder-driven `aria-label`s.

Now 10/10 passing, eslint clean. The remaining `setupResponsiveHiding` / `setupCollapsingHeader` / `isExpandedSection` expand-values gaps stay as noted follow-ups (scroll/matchMedia-driven, out of scope for this suite).

mokimo (approved):
LGTM, and this is the strongest suite of the batch. Verified by running it directly: 10/10 passing against unmodified `libs/c2/blocks/comparison-table-c2/comparison-table-c2.js`.

I hand-traced the two least-obvious assertions against the actual decoration logic in `comparison-table-c2.js` to make sure they weren't coincidentally-correct guesses:
- "marks the primary column card and its cells" — confirmed `data-column-index="2"` for the primary card and exactly 2 `.primary-cell` matches is the correct derived result of `decorateTableToggleButton`'s `arePrimaryColumns` bookkeeping plus `setupCellAttributes`, given the `default.html` fixture's "primary" marker cell.
- The tooltip test correctly distinguishes the 3-part `label|position|text` syntax (→ `.milo-tooltip` trigger + separate `.ct-tooltip-content`) from the label-only `<u>` syntax (→ plain `span.dotted-underline`), matching `decorateUnderlineTooltipTriggers` exactly.

Proper async handling throughout: `setAccessibilityLabels` resolves placeholders via a dynamic import + promise chain that `init()` doesn't await, and the tests correctly use the shared `waitFor` helper (`test/helpers/waitfor.js`) rather than a race-prone fixed delay — same pattern as `test/blocks/table`. The `setConfig({ contentRoot: '/test/blocks/comparison-table-c2/mocks', ... })` + `placeholders.json` fixture setup matches the established convention (same pattern in `test/blocks/table`, `test/blocks/merch/three-in-one.test.js`).

Only a non-blocking nit: PR description claims "8/8 passing," actual file has 10 tests (confirmed by the real run: 10/10). Worth a description update.

SilviuLCF:
Thank you for review, i've updated the description

PR #6377, opened 2026-07-30
MWPW-201352: add unit tests for c2 news block

SilviuLCF:
## Summary
- Adds unit tests for the `news` c2 block (`libs/c2/blocks/news`), which previously had no coverage in `test/blocks`
- Covers: single-row no-op, header eyebrow formatting (`.headline > .headline-text > .eyebrow`, original cell removed), header svg icon move with federated `src` rewrite, count-based `news-items` up class (two-up/three-up), item `foreground` cell + content classification (`news-item-headline`/`news-item-body`), and standalone-link handling in both non-quiet and quiet variants
- The non-quiet standalone-link path used to throw (`classList.add('')`); that was fixed in MWPW-202052 (PR #6360, merged), and this suite locks in the corrected behavior
- Follows conventions from existing block tests (e.g. `test/blocks/base-card`, `test/blocks/brand-concierge`)

Resolves: https://jira.corp.adobe.com/browse/MWPW-201352

## Test plan
- [x] `npx wtr --config ./web-test-runner.config.mjs "./test/blocks/news/news.test.js" --node-resolve --port=2000` — 11/11 passing
- [x] `npx eslint test/blocks/news/news.test.js` — clean

### DIFF
--- a/test/blocks/news/mocks/default.html
+++ b/test/blocks/news/mocks/default.html
@@ -1,0 +1,28 @@
+<main>
+  <div class="section">
+    <div class="news">
+      <div>
+        <div><p>Latest news</p></div>
+      </div>
+      <div>
+        <div>
+          <h3>Item one headline</h3>
+          <p>Body copy for item one.</p>
+          <p><a href="https://www.adobe.com/one">Read more</a></p>
+        </div>
+      </div>
+      <div>
+        <div>
+          <h3>Item two headline</h3>
+          <p>Body copy for item two.</p>
+        </div>
+      </div>
+      <div>
+        <div>
+          <h3>Item three headline</h3>
+          <p>Body copy for item three.</p>
+        </div>
+      </div>
+    </div>
+  </div>
+</main>
--- a/test/blocks/news/mocks/four-up.html
+++ b/test/blocks/news/mocks/four-up.html
@@ -1,0 +1,13 @@
+<main>
+  <div class="section">
+    <div class="news">
+      <div>
+        <div><p>Latest news</p></div>
+      </div>
+      <div><div><h3>Item one</h3><p>Body one.</p></div></div>
+      <div><div><h3>Item two</h3><p>Body two.</p></div></div>
+      <div><div><h3>Item three</h3><p>Body three.</p></div></div>
+      <div><div><h3>Item four</h3><p>Body four.</p></div></div>
+    </div>
+  </div>
+</main>
--- a/test/blocks/news/mocks/header-icon.html
+++ b/test/blocks/news/mocks/header-icon.html
@@ -1,0 +1,24 @@
+<main>
+  <div class="section">
+    <div class="news">
+      <div>
+        <div>
+          <picture><img src="/federal/icons/news-icon.svg" alt="news icon" /></picture>
+          <p>Latest news</p>
+        </div>
+      </div>
+      <div>
+        <div>
+          <h3>Item one headline</h3>
+          <p>Body copy for item one.</p>
+        </div>
+      </div>
+      <div>
+        <div>
+          <h3>Item two headline</h3>
+          <p>Body copy for item two.</p>
+        </div>
+      </div>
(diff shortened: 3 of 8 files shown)


zagi25 (approved):
**Code review — MWPW-201352: unit tests for c2 news block**

Test-only PR (1 test file + 4 mock fixtures, no source edits) that fulfils the ticket and follows the brand-concierge pattern.

- Assertions are meaningful, not no-throw checks: item wrapping + count-based `up` class, foreground/content classification, standalone-link handling (quiet and non-quiet), and the mocks are minimal and representative of real authoring. `readFile` + `@esm-bundle/chai`, one `describe` per concern; no sinon needed since the block does no network I/O.
- Verified green: the unit-test CI job (`Running tests (20.x)`) and all three MILO browser jobs pass on this PR, and the `news.js` standalone-link class logic on this branch is the safe form (`classList.add('standalone-link', 'label')` then a guarded `.add('quiet')`), so the non-quiet path the tests exercise is correct.

Low-severity, non-blocking coverage gaps (not required by the ticket, worth a follow-up):
- No test for the `four-up` / `six-up` item-count classes (only 2-up / 3-up are covered).
- No test for the `isSvgUrl` false branch (a non-SVG icon `src` should be left unrewritten).

_Review by Claude._

SilviuLCF:
Thanks for the review! Covered both non-blocking gaps in f50394cc4 (test-only):

- **four-up / six-up counts** — added `four-up.html` (4 items) and `six-up.html` (6 items) with tests asserting the corresponding `news-items` count classes.
- **`isSvgUrl` false branch** — added `header-png-icon.html` and `leaves a non-svg header icon src unrewritten`, which asserts a `.png` header icon keeps its original `src` (not passed through `getFederatedUrl`) while still receiving the `icon` class and moving into the headline.

Now 11/11 passing, eslint clean.

mokimo (commented):
Verified by running the suite directly: 11/11 passing against unmodified `libs/c2/blocks/news/news.js`.

Tests are solid and non-trivial — traced through `formatHeader`'s DOM surgery by hand (the header row's original cell wrapper gets emptied by relocating `headlineText`/`headlinePicture` out of it, then `row.firstElementChild.remove()` cleans up the now-empty leftover wrapper) and confirmed the assertions match actual output, not assumptions. The federated-svg-only rewrite (`.svg` rewritten, `.png` left as-authored) is a real, meaningful edge case and is correctly tested with two near-identical mocks (`header-icon.html` vs `header-png-icon.html`) that isolate exactly that one variable. The regression note about the standalone-link fix (MWPW-202052, `classList.add('')` crash) is a good call-out and the quiet/non-quiet standalone-link split is genuinely exercised, not just asserted once.

Non-blocking nits:
- PR description says "8/8 passing" in both the summary and the test plan, but the file actually has 11 `it(...)` blocks and the real run confirms 11/11. Worth updating the description so it doesn't undersell/mislead reviewers about coverage.
- The `header-png-icon.html` mock triggers a real (harmless) 404 network request during the run (`federal/icons/news-icon.png` doesn't resolve to anything servable) — wtr surfaces this as a warning, not a failure, and it doesn't affect the assertions, but a mock image with a real fixture path (or a documented reason it's intentionally unresolvable) would be tidier.

SilviuLCF:
Thanks @mokimo!

- **Test count in description:** the body already reads 11/11 (summary + test plan) — the earlier 8/8 was pre the review commit that added the four-up/six-up/non-svg tests, so it's now accurate.
- **header-png-icon 404:** added a comment in the mock (c3d31be0f) documenting that the `.png` src is an intentionally-unresolvable fixture path — the test asserts a non-svg icon src is left as-authored (not federated-rewritten), so the image is never loaded, only its `src` read. The 404 wtr surfaces is harmless/expected.

Still 11/11 passing, eslint clean.

zagi25 (approved):

mokimo (approved):

PR #6378, opened 2026-07-30
MWPW-202200: add unit tests for c2 social-proof block

SilviuLCF:
## Summary
- Adds unit tests for the `social-proof` c2 block (`libs/c2/blocks/social-proof`), which previously had no coverage in `test/blocks`
- Covers: standard-variant decoration (`.foreground`, `heading-3`, `.content` wrapping non-heading children), opening-quote hanging (`.hang-opening-quote`), background color applied from the second cell (+ cell removal), media-cell relocation into the first row (+ media row removal), the `<blockquote>` → `<figure>`/`<figcaption class="body-lg">` variant, and the empty-block no-op
- Follows conventions from existing block tests (e.g. `test/blocks/base-card`, `test/blocks/rich-content`)

Resolves: https://jira.corp.adobe.com/browse/MWPW-202200

## Test plan
- [x] `npx wtr --config ./web-test-runner.config.mjs "./test/blocks/social-proof/social-proof.test.js" --node-resolve --port=2000` — 6/6 passing
- [x] `npx eslint test/blocks/social-proof/social-proof.test.js` — clean

### DIFF
--- a/test/blocks/social-proof/social-proof.test.js
+++ b/test/blocks/social-proof/social-proof.test.js
@@ -1,0 +1,92 @@
+import { readFile } from '@web/test-runner-commands';
+import { expect } from '@esm-bundle/chai';
+
+import init from '../../../libs/c2/blocks/social-proof/social-proof.js';
+
+describe('Social Proof', () => {
+  it('decorates the standard variant: foreground, heading level 3 and content wrapper', async () => {
+    document.body.innerHTML = await readFile({ path: './mocks/default.html' });
+    const block = document.querySelector('.social-proof');
+    init(block);
+
+    const foreground = block.querySelector('.foreground');
+    expect(foreground).to.exist;
+
+    const heading = foreground.querySelector('h3');
+    expect(heading.classList.contains('heading-3')).to.be.true;
+
+    // non-heading children are wrapped in a .content div; the heading stays outside it
+    const content = foreground.querySelector(':scope > .content');
+    expect(content).to.exist;
+    expect(content.querySelectorAll('p').length).to.equal(2);
+    expect(content.querySelector('h3')).to.be.null;
+  });
+
+  it('hangs the opening quote of the heading into its own span', async () => {
+    document.body.innerHTML = await readFile({ path: './mocks/default.html' });
+    const block = document.querySelector('.social-proof');
+    init(block);
+
+    const quote = block.querySelector('.foreground h3 .hang-opening-quote');
+    expect(quote).to.exist;
+    expect(quote.textContent).to.equal('“');
+  });
+
+  it('applies the background color from the second cell and removes that cell', async () => {
+    document.body.innerHTML = await readFile({ path: './mocks/default.html' });
+    const block = document.querySelector('.social-proof');
+    init(block);
+
+    const firstRow = block.children[0];
+    expect(firstRow.style.backgroundColor).to.equal('rgb(245, 245, 245)');
+    // the color cell is consumed, not left in the DOM
+    expect([...firstRow.children].some((c) => c.textContent.trim() === '#f5f5f5')).to.be.false;
+  });
+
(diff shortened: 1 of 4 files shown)


test/blocks/social-proof/social-proof.test.js
```diff
@@ -0,0 +1,92 @@
...
+    const spans = caption.querySelectorAll('span');
+    expect(spans.length).to.equal(2);
+    expect(spans[0].textContent).to.equal('Jane Doe');
+    expect(spans[1].textContent).to.equal('CEO, Acme');
+
+    // caption paragraphs are moved out of the blockquote
+    expect(blockquote.querySelectorAll('p').length).to.equal(1);
+  });
+
+  it('does nothing when the block has no rows', async () => {
```
zagi25, line 84: Med / non-blocking coverage gap: consider a case for a row with a foreground but **no media row**. The source's `decorate()` does `const mediaCell = secondRow?.children[0]; mediaCell.classList.add('media');` — when `secondRow` is absent, `mediaCell` is `undefined` and this throws a `TypeError`. Every mock here includes a media row, so this crash path stays untested. It's a pre-existing bug in `social-proof.js` (out of scope for a test-only ticket), but a fixture that documents/guards it would be worth adding.

zagi25 (approved):
**Code review — MWPW-202200: unit tests for c2 social-proof block**

Test-only PR (4 new files, no source edits), and it matches the ticket exactly. I traced each assertion against the real block source (`libs/c2/blocks/social-proof/social-proof.js` on `stage`, fetched via `git show` since the local worktree is on another branch):

- Heading / content-wrap, opening-quote hang (`\p{Pi}` matches the curly quote), background-color (`#f5f5f5` → `rgb(245, 245, 245)` serialization + bg-cell removal), media-move, blockquote/figure decoration, and the empty-block early-return are all asserted correctly and meaningfully — no tautological no-throw-only checks, no over-mocking, and `init()` is correctly called synchronously (the author adapted the brand-concierge pattern rather than copy-pasting its async `init`).
- No security surface: static HTML fixtures + local decoration utilities only; no network/`fetch`, no `eval`, no injection beyond fixed fixtures.

One non-blocking coverage suggestion inline (a pre-existing, out-of-scope source crash path with no regression coverage).

_Review by Claude._

SilviuLCF:
Thanks for the review! Good catch on the no-media-row crash — agreed it's a real pre-existing bug in `social-proof.js` (`mediaCell.classList.add('media')` throws a `TypeError` when `secondRow` is absent), and out of scope for this test-only ticket.

Leaving this PR as-is; I'll track the source fix (a guard around `mediaCell`, same shape as the news standalone-link fix) plus its regression test as a separate follow-up ticket rather than documenting the crash here.

mokimo (approved):
LGTM. Verified by running the suite directly: 6/6 passing against unmodified `libs/c2/blocks/social-proof/social-proof.js`, matching the PR's test plan.

The tests exercise real, non-trivial DOM transforms rather than mock echoes: background-color extraction + cell removal, media-cell relocation into the first row (with a `block.children.length === 1` check proving the second row is actually consumed, not just decorated), and the `<blockquote>` → `<figure>`/`<figcaption class="body-lg">` restructuring, including that caption spans preserve order (`Jane Doe`, `CEO, Acme`) and that caption paragraphs are moved out of the blockquote. `init(block)` is called without `await`, which is correct — `init` is synchronous in this block, no async operation is silently unawaited.

One thing worth a non-blocking note: `decorate()`'s `mediaCell.classList.add('media')` (line ~44 of `social-proof.js`) will throw if a block ever has a first row but no second row — none of the four mocks exercise that shape (only "no rows at all" is tested, which correctly early-returns before reaching that line). This is a pre-existing block behavior, not something introduced by this PR, so not blocking here — just flagging in case a future content author trips over it and needs a repro test.

PR #6379, opened 2026-07-30
MWPW-202191: add unit tests for c2 floating-cta block

SilviuLCF:
## Summary
- Adds unit tests for the `floating-cta` c2 block (`libs/c2/blocks/floating-cta`), which previously had no coverage in `test/blocks`
- Covers: `a.promo-cta` build (img + arrow `span.icon-button` + `tabindex=-1`, block content replaced), label split on `|` into CTA text + `aria-label`, no-pipe fallback (`aria-label` = CTA text), federated svg `src` rewrite, plain-text fallback when there is no anchor (uses `textContent`, `#` href), and the missing-link-paragraph early return
- Follows conventions from existing block tests (e.g. `test/blocks/base-card`, `test/blocks/explore-card`)

Scope note: covers the deterministic non-merch CTA build. The merch/checkout path (`waitForCheckoutLink` + `onceSettled`) and the `applyCustomHide` IntersectionObserver are intentionally not asserted (async / commerce / observer-driven, flaky in unit tests), consistent with the other c2 suites.

Resolves: https://jira.corp.adobe.com/browse/MWPW-202191

## Test plan
- [x] `npx wtr --config ./web-test-runner.config.mjs "./test/blocks/floating-cta/floating-cta.test.js" --node-resolve --port=2000` — 6/6 passing
- [x] `npx eslint test/blocks/floating-cta/floating-cta.test.js` — clean

### DIFF
--- a/test/blocks/floating-cta/floating-cta.test.js
+++ b/test/blocks/floating-cta/floating-cta.test.js
@@ -1,0 +1,73 @@
+import { readFile } from '@web/test-runner-commands';
+import { expect } from '@esm-bundle/chai';
+
+import init from '../../../libs/c2/blocks/floating-cta/floating-cta.js';
+
+describe('Floating CTA', () => {
+  it('builds a promo-cta anchor from the image and link, replacing the block content', async () => {
+    document.body.innerHTML = await readFile({ path: './mocks/default.html' });
+    const block = document.querySelector('.floating-cta');
+    await init(block);
+
+    expect(block.children.length).to.equal(1);
+    const cta = block.querySelector(':scope > a.promo-cta');
+    expect(cta).to.exist;
+    expect(cta.getAttribute('href')).to.equal('https://www.adobe.com/buy');
+    expect(cta.getAttribute('tabindex')).to.equal('-1');
+    expect(cta.querySelector('img')).to.exist;
+    expect(cta.querySelector('span.icon-button[aria-hidden="true"]')).to.exist;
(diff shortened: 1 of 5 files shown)


mokimo (approved):
LGTM. Verified by actually running the suite (`wtr test/blocks/floating-cta/floating-cta.test.js`): 6/6 passing against the unmodified `libs/c2/blocks/floating-cta/floating-cta.js`, matching the PR's test plan exactly.

Spot-checked the meaningful assertions:
- The federated-URL rewrite assertion (`https://main--federal--adobecom.aem.page/federal/icons/cta.svg`) matches the actual default `getFederatedContentRoot()` resolution, not a made-up value — cross-checked against the same string used in `test/blocks/global-navigation/global-navigation.test.js`.
- The pipe-split (`cta text | aria-label`), no-pipe fallback, text-only fallback (`#` href), and missing-link-paragraph early-return all correspond to real branches in `init()`, each exercised by a distinct mock fixture rather than reusing one fixture for everything.
- The scope note explicitly excluding the merch/checkout path (`waitForCheckoutLink`) and `applyCustomHide`'s `IntersectionObserver` is the right call — those are async/observer-driven and consistent with how sibling suites scope out non-deterministic behavior.

No fabricated/tautological assertions, no missing awaits, no leftover `console.log`/`debugger`. No CODEOWNERS entry needed (test-only addition to an existing block dir under `test/blocks/`, which isn't itself a CODEOWNERS-governed path).

zagi25 (approved):

PR #6380, opened 2026-07-30
MWPW-202194: add unit tests for c2 logo-ticker block

SilviuLCF:
## Summary
- Adds unit tests for the `logo-ticker` c2 block (`libs/c2/blocks/logo-ticker`), which previously had no coverage in `test/blocks`
- Covers: track build (single `.logo-ticker-track[role=img]` replacing the block content), duplicated logo sets (two `.logo-ticker-set`s, the second `aria-hidden`, logos cloned not moved), `aria-label` sourced from the second block row, the no-second-row case (no label), and the no-logos no-op

Scope note: covers the deterministic track construction. `syncTrackMetrics`/`.is-static` and the `ResizeObserver` are layout/observer-driven (depend on rendered widths) and intentionally not asserted; they run without error in the tests. Consistent with the other c2 suites.

Resolves: https://jira.corp.adobe.com/browse/MWPW-202194

## Test plan
- [x] `npx wtr --config ./web-test-runner.config.mjs "./test/blocks/logo-ticker/logo-ticker.test.js" --node-resolve --port=2000` — 5/5 passing
- [x] `npx eslint test/blocks/logo-ticker/logo-ticker.test.js` — clean

### DIFF
--- a/test/blocks/logo-ticker/logo-ticker.test.js
+++ b/test/blocks/logo-ticker/logo-ticker.test.js
@@ -1,0 +1,61 @@
+import { readFile } from '@web/test-runner-commands';
+import { expect } from '@esm-bundle/chai';
+
+import init from '../../../libs/c2/blocks/logo-ticker/logo-ticker.js';
+
+describe('Logo Ticker', () => {
+  it('builds a single track (role=img) that replaces the block content', async () => {
+    document.body.innerHTML = await readFile({ path: './mocks/default.html' });
+    const block = document.querySelector('.logo-ticker');
+    init(block);
+
+    expect(block.children.length).to.equal(1);
+    const track = block.querySelector(':scope > .logo-ticker-track');
+    expect(track).to.exist;
+    expect(track.getAttribute('role')).to.equal('img');
+  });
+
+  it('duplicates the logos into two sets, hiding the second from assistive tech', async () => {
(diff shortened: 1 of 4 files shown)


mokimo (approved):
LGTM. Confirmed against `libs/c2/blocks/logo-ticker/logo-ticker.js`:
- `SET_COUNT = 2` duplication, `aria-hidden` only on the second set, and the exact icon count (`3 logos × 2 sets = 6`) are all correctly derived from source, not just copied from the mock.
- The `aria-label` sourced from `el.children[1].textContent` is asserted to be present when there's a second row and correctly asserted absent (`hasAttribute('aria-label')` false) in `no-aria.html`, which only has one row — matches the `el.children.length >= 2` gate in `buildTrack`.
- The no-logos early return (`if (!logos.length) return;`) is verified by asserting `.logo-ticker-track` is `null`, not just "block unchanged" — a precise assertion that would actually fail if the guard were removed.

No shallow or tautological assertions. `syncTrackMetrics`'s `is-static` toggle (based on real layout metrics) isn't tested, which is reasonable given `@web/test-runner`'s real-browser environment makes deterministic layout assertions brittle — not blocking.

zagi25 (approved):

narcis-radu (approved):

PR #6382, opened 2026-07-30
MWPW-202201: add unit tests for c2 split-aside-grid block

SilviuLCF:
## Summary
- Adds unit tests for the `split-aside-grid` c2 block (`libs/c2/blocks/split-aside-grid`), which previously had no coverage in `test/blocks`
- Covers: block structure (`.split-aside-grid-items` / `-controls` / `-stack` / `.aria-live-container`), per-slide decoration (`data-slide-index`, toggle button, `.foreground`), content decoration (`heading-5`, `.content-container`, standalone-link marking **and** the negative inline-link branch), media moved into the stack, the dot `tablist` (first `aria-current`/`is-active`), the initial active-slide state from `applyRotation(0)` (item active, toggle expanded, media `data-slot=0`/`aria-hidden`), and the `Slide N` heading fallback

Scope note: covers deterministic decoration + the initial active state (viewport-independent). The swipe/drag pointer gestures, desktop/mobile matchMedia rebinding, stack fly/snap animations, and Resize/Intersection observers are intentionally not asserted (timer/pointer/observer/matchMedia-driven, flaky in unit tests); they run without error here.

Resolves: https://jira.corp.adobe.com/browse/MWPW-202201

## Test plan
- [x] `npx wtr --config ./web-test-runner.config.mjs "./test/blocks/split-aside-grid/split-aside-grid.test.js" --node-resolve --port=2000` — 7/7 passing
- [x] `npx eslint test/blocks/split-aside-grid/split-aside-grid.test.js` — clean

### DIFF
--- a/test/blocks/split-aside-grid/mocks/default.html
+++ b/test/blocks/split-aside-grid/mocks/default.html
@@ -1,0 +1,34 @@
+<main>
+  <div class="section">
+    <div class="split-aside-grid">
+      <div>
+        <div>
+          <h3>First slide</h3>
+          <p>Body copy with an <a href="https://www.adobe.com/inline">inline link</a> in a sentence.</p>
+          <p><a href="https://www.adobe.com/one">Learn more</a></p>
+        </div>
+        <div>
+          <picture><img src="" alt="media one" /></picture>
+        </div>
+      </div>
+      <div>
+        <div>
+          <h3>Second slide</h3>
+          <p>Body copy for the second slide.</p>
+        </div>
+        <div>
+          <picture><img src="" alt="media two" /></picture>
+        </div>
+      </div>
+      <div>
+        <div>
+          <h3>Third slide</h3>
+          <p>Body copy for the third slide.</p>
+        </div>
+        <div>
+          <picture><img src="" alt="media three" /></picture>
+        </div>
+      </div>
+    </div>
+  </div>
+</main>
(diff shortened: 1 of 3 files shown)


mokimo (approved):
LGTM — this is the strongest of the five. Traced every assertion against `libs/c2/blocks/split-aside-grid/split-aside-grid.js`:
- `decorateItem`'s toggle button, `content-container` wrapping, and `markStandaloneLinks`' text-match heuristic (link text === parent text → `standalone-link label`, else left alone) are both correctly verified, including the negative case (an inline link inside a sentence must NOT get `standalone-link`).
- The "Slide N" fallback title when a slide has no heading is verified against `no-heading.html`, matching `decorateItem`'s `content?.querySelector(':is(h1..h6)')?.textContent || `Slide ${index+1}`` fallback exactly.
- Dot list `role="tablist"`, `aria-current="location"` only on the active dot, and `data-slot`/`aria-hidden` wiring after the initial `applyRotation(0, true)` call in `syncBindings()` are all correctly asserted — this requires tracing through `setupBlock` → `syncBindings` → `applyRotation`, which the author clearly did.

No shallow or tautological tests. Swipe/drag/keyboard-navigation and the mobile-vs-desktop `bindMobile`/`bindDesktop` branching aren't covered, but that's consistent with this being a decoration/structure test, not an interaction test — not blocking.

zagi25 (approved):

PR #6383, opened 2026-07-30
MWPW-202192: add unit tests for c2 hover-list block

SilviuLCF:
## Summary
- Adds unit tests for the `hover-list` c2 block (`libs/c2/blocks/hover-list`), which previously had no coverage in `test/blocks`
- Covers: `container` class + `.hover-list-headline-wrapper` / `.hover-list-col > ol.hover-list-items` structure, first-row headline decoration (`h2.heading-2`), per-row list items (`.hover-list-copy` with `.hover-list-number.eyebrow` "N." + `.hover-list-text.heading-5`), picture rows wrapped in a `.hover-list-media[popover=manual]`, the no-picture negative branch, and the empty-block no-op

Scope note: covers the deterministic decoration. The cursor-follower animation (mousemove / rAF / popover), sticky-boundary ResizeObserver, and matchMedia desktop/tablet behavior are intentionally not asserted (pointer/timer/observer-driven, flaky in unit tests); they run without error here.

Resolves: https://jira.corp.adobe.com/browse/MWPW-202192

## Test plan
- [x] `npx wtr --config ./web-test-runner.config.mjs "./test/blocks/hover-list/hover-list.test.js" --node-resolve --port=2000` — 6/6 passing
- [x] `npx eslint test/blocks/hover-list/hover-list.test.js` — clean

### DIFF
--- a/test/blocks/hover-list/hover-list.test.js
+++ b/test/blocks/hover-list/hover-list.test.js
@@ -1,0 +1,82 @@
+import { readFile } from '@web/test-runner-commands';
+import { expect } from '@esm-bundle/chai';
+
+import init from '../../../libs/c2/blocks/hover-list/hover-list.js';
+
+describe('Hover List', () => {
+  it('marks the block as a container and builds headline + list columns', async () => {
+    document.body.innerHTML = await readFile({ path: './mocks/default.html' });
+    const block = document.querySelector('.hover-list');
+    init(block);
+
+    expect(block.classList.contains('container')).to.be.true;
+
+    const children = [...block.children];
+    expect(children[0].classList.contains('hover-list-headline-wrapper')).to.be.true;
+    expect(children[1].classList.contains('hover-list-col')).to.be.true;
+    expect(block.querySelector('.hover-list-col > ol.hover-list-items')).to.exist;
+  });
+
+  it('decorates the headline from the first row', async () => {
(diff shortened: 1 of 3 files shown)


mokimo (approved):
LGTM. Verified each assertion against `libs/c2/blocks/hover-list/hover-list.js`'s `decorate()`:
- Headline flattening from `rows[0].children[0]` and the `heading-2` class (via `decorateBlockText(headingCol, {heading:'2'})`, `libs/utils/decorate.js:132-155`) are both correctly asserted.
- Per-row item structure (`.hover-list-copy` → number + text), the `popover="manual"` media wrapper, and the picture-count-per-row cases (2 pictures / 1 picture / 0 pictures → no media element) all match source behavior exactly, including the row with an empty media `<div></div>` correctly producing no `.hover-list-media`.
- The empty-block no-op case asserts absence of both the headline wrapper and the items list, matching the early `if (!rows.length) return;` guard.

No fabricated or tautological assertions found — every check traces to a real conditional in `decorate()`. The interactive cursor-follower/animation code (`addCursorFollower`, RAF loop) isn't covered, but that's inherently timing/pointer-driven and out of scope for a DOM-structure unit test; reasonable to leave to Nala/manual QA.

zagi25 (approved):

PR #6385, opened 2026-07-30
MWPW-202196: add unit tests for c2 plans-hero block

SilviuLCF:
## Summary
Adds unit tests for the c2 Milo block `plans-hero` (`libs/c2/blocks/plans-hero/plans-hero.js`). `init(el)` adds the `container` class and runs `decorateViewportContent`, whose `decorate` reshapes the first row into a `.plans-hero-media` div (moving the media-cell `<picture>`) and a `.plans-hero-content` div (the text cell, decorated via `decorateBlockText` to `heading-2` / `body-md`).

Coverage:
- default (text + media): container class, row replaced by media + content divs, picture moved (not duplicated), heading `heading-2`, body paragraphs `body-md`
- no media picture: empty `.plans-hero-media` still produced, text still decorated
- empty row (no text cell): fallback empty content div branch
- no rows: `init` is a no-op reshape, adds `container` only, does not throw

## Scope
TEST-ONLY. Nothing under `libs/` was modified. Deterministic DOM-decoration assertions only; timer/rAF/observer/matchMedia/network behaviors are not asserted, but `init` must not throw. 9 tests, all green; eslint clean.

Resolves: https://jira.corp.adobe.com/browse/MWPW-202196

## Test plan
- [x] `npx wtr --config ./web-test-runner.config.mjs "./test/blocks/plans-hero/plans-hero.test.js" --node-resolve` — 9 passed
- [x] `npx eslint test/blocks/plans-hero/plans-hero.test.js` — clean

### DIFF
--- a/test/blocks/plans-hero/mocks/default.html
+++ b/test/blocks/plans-hero/mocks/default.html
@@ -1,0 +1,16 @@
+<main>
+  <div class="section">
+    <div class="plans-hero">
+      <div>
+        <div>
+          <h2>Plans hero heading</h2>
+          <p>Some supporting body copy for the plans hero.</p>
+          <p><a href="https://www.adobe.com/plans">Buy now</a></p>
+        </div>
+        <div>
+          <picture><img src="/media/plans-hero.png" alt="plans hero" /></picture>
+        </div>
+      </div>
+    </div>
+  </div>
+</main>
--- a/test/blocks/plans-hero/mocks/empty-row.html
+++ b/test/blocks/plans-hero/mocks/empty-row.html
@@ -1,0 +1,7 @@
+<main>
+  <div class="section">
+    <div class="plans-hero">
+      <div></div>
+    </div>
+  </div>
+</main>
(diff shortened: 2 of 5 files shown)


mokimo (approved):
LGTM. Smallest of the five blocks and the tests are appropriately scoped to match. Verified:

- "replaces the row with a media div and a content div" correctly checks `:scope > div > div > picture` is null, proving the picture was actually moved up (not left nested under the original row wrapper) — a real structural assertion, not just an existence check.
- The `empty-row.html` case correctly exercises the `textCell ?? createTag('div', {class:'plans-hero-content'})` fallback in `plans-hero.js` when the row has zero children — good edge-case coverage.
- `no-rows.html` correctly asserts the block still gets the `container` class (added unconditionally in `init()`, before `decorate()`'s early-return guard runs) but nothing else changes.

One small non-blocking nit: `test/blocks/plans-hero/plans-hero.test.js:47` asserts `expect(bodyParas).to.have.lengthOf.at.least(1)` where the mock (`mocks/default.html`) authors exactly one body paragraph — an exact `.to.have.lengthOf(1)` would be tighter and just as easy to write. Not blocking.

zagi25 (approved):

PR #6386, opened 2026-07-30
MWPW-202198: add unit tests for c2 quick-actions block

SilviuLCF:
## Summary
Adds unit tests for the c2 Milo `quick-actions` block (`libs/c2/blocks/quick-actions/quick-actions.js`). 12 tests across 4 mocks covering deterministic DOM decoration:

- Section header promotion: first row moved into `.quick-actions-header`, source row removed, heading decorated at level 4, body copy preserved.
- Grid/tile construction: one `a.quick-actions-tile` per tile row, grid carries `quick-actions-grid parallax-stagger-ltr`, block ends with only header + grid.
- Tile href taken from the label link.
- Media handling: `<picture>` gets `quick-actions-media`; bare `` (no picture) fallback gets `quick-actions-media`.
- Tile footer: `.quick-actions-tile-label.heading-6` with trimmed text plus an aria-hidden chevron containing the SVG.
- Responsive n-up: exactly one of `two-up`/`three-up`/`six-up` applied.
- Negative/fallback branches: no header when first cell lacks a heading, no href/footer for link-less tiles, rows with fewer than two cells ignored, empty block renders an empty grid without throwing.

## Scope note
TEST-ONLY — nothing under `libs/` was modified. Per scope, timer/rAF/observer/matchMedia/pointer/network behaviors are not asserted; only deterministic DOM output is verified and `init` must not throw. The 404 image requests in the test log are harmless mock media fetches.

Resolves: https://jira.corp.adobe.com/browse/MWPW-202198

## Test plan
- [x] `npx wtr --config ./web-test-runner.config.mjs "./test/blocks/quick-actions/quick-actions.test.js" --node-resolve --port=2095` → 12 passed, 0 failed
- [x] `npx eslint test/blocks/quick-actions/quick-actions.test.js` → clean

### DIFF
--- a/test/blocks/quick-actions/mocks/default.html
+++ b/test/blocks/quick-actions/mocks/default.html
@@ -1,0 +1,24 @@
+<main>
+  <div class="section">
+    <div class="quick-actions">
+      <div>
+        <div>
+          <h2>Quick actions</h2>
+          <p>Get things done fast with Adobe Express.</p>
+        </div>
+      </div>
+      <div>
+        <div><p><a href="https://www.adobe.com/express/feature/pdf/convert">Convert to PDF</a></p></div>
+        <div><picture><img src="/media_convert.png" alt="Convert to PDF" /></picture></div>
+      </div>
+      <div>
+        <div><p><a href="https://www.adobe.com/express/feature/pdf/compress">Compress PDF</a></p></div>
+        <div><picture><img src="/media_compress.png" alt="Compress PDF" /></picture></div>
+      </div>
+      <div>
+        <div><p><a href="https://www.adobe.com/express/feature/pdf/merge">Merge PDF</a></p></div>
+        <div><picture><img src="/media_merge.png" alt="Merge PDF" /></picture></div>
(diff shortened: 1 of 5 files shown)


mokimo (approved):
LGTM. Cross-checked every assertion against `libs/c2/blocks/quick-actions/quick-actions.js` and all four mocks:

- "adds media class to bare img when no picture is present" correctly reflects `(mediaPic ?? mediaImg).classList.add(...)` picking the bare `` when there's no `<picture>` wrapper (fallback.html's "Remove background" row).
- "renders no href and no footer for a tile without a label link" correctly accounts for `...(labelLink && {href: labelLink.href})` spreading `null` as a no-op — a subtle JS behavior the test gets exactly right.
- Empty-block test confirms `decorateSectionHeader` returns `null` safely and the grid still renders with 0 tiles, matching the `[header, grid].filter(Boolean)` composition in source.

Nothing blocking, nothing to add.

zagi25 (approved):

PR #6387, opened 2026-07-30
MWPW-202195: add unit tests for c2 offer-hero block

SilviuLCF:
## Summary
Adds unit test coverage for the c2 Milo block `offer-hero`. Tests the deterministic DOM decoration performed by `decorate()`:
- Hero row wrapped in `.hero > .hero-content`; authored heading promoted to `h1.heading-1`.
- App-icon eyebrow built from the authored svg image, `alt` cleared, and `src` rewritten to a federated URL.
- Hero body (`body-lg`) and CTA (`con-button`) decoration.
- Heading row moved into `.what-included` as an `h3.hero-eyebrow`.
- One `.hero-card` per valid card row, with card title (`hero-card-title`), body text (`hero-card-body-text`), media (`hero-card-media`), and learn-more link (`.learn-more.label` with chevron svg and `tabindex="-1"`).
- Linked cards become anchor tiles with `href` + `data-tracking-label`; unlinked cards become div tiles.
- Branch coverage: hero without app-icon, card without link, card without heading, invalid (non-two-cell) row skipped, and the empty/no-rows early-return no-op.

## Scope note
Test-only — nothing under `libs/` is modified. Coverage is limited to the deterministic DOM decoration. Animation behavior driven by timers, requestAnimationFrame, IntersectionObserver/ResizeObserver, matchMedia, and scroll/pointer events is intentionally not asserted (env-dependent); `init` is only verified to run without throwing.

Resolves: https://jira.corp.adobe.com/browse/MWPW-202195

## Test plan
- [x] `npx wtr` — 15/15 passing
- [x] `npx eslint test/blocks/offer-hero/offer-hero.test.js` — clean

### DIFF
--- a/test/blocks/offer-hero/mocks/default.html
+++ b/test/blocks/offer-hero/mocks/default.html
@@ -1,0 +1,60 @@
+<main>
+  <div class="section">
+    <div class="offer-hero">
+      <div>
+        <div>
+          <p class="eyebrow">Special offer</p>
+          <p><img src="/federal/assets/icons/app-icon.svg" alt="app icon"></p>
+          <h2 id="save-today">Save big on the entire collection today</h2>
+          <p>Unlock every creative app at one low introductory price.</p>
+          <p><strong><a href="/buy-now">Buy now</a></strong></p>
+        </div>
+      </div>
+      <div>
+        <div>
+          <h3>What's included</h3>
+        </div>
+      </div>
+      <div>
+        <div>
+          <h4>Photoshop</h4>
+          <p>Edit and composite photos on desktop and mobile.</p>
+          <p><a href="/apps/photoshop">Learn more</a>.</p>
+        </div>
+        <div>
+          <picture><img src="/media/photoshop.png" alt="Photoshop artwork"></picture>
+        </div>
+      </div>
+      <div>
+        <div>
+          <h4>Illustrator</h4>
+          <p>Create vector graphics and illustrations.</p>
+          <p><a href="/apps/illustrator">Learn more</a>.</p>
+        </div>
+        <div>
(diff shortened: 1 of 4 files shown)


mokimo (approved):
LGTM — thorough and accurate. Highlights that were specifically checked:

- The federated-URL assertion (`icon.src` equal to `'https://main--federal--adobecom.aem.page/federal/assets/icons/app-icon.svg'`) is correct given `getFederatedUrl`'s localhost-detection branch (`libs/utils/utils.js:476-483` + `getFederatedContentRoot`, lines 441-467) — this requires understanding that the source URL must contain `/federal/` and that web-test-runner serves on `localhost`, which is not obvious; the author got it right.
- "skips a row that does not have exactly two cells" (in `variants.html`) correctly verifies the skipped row is left un-removed (source returns *before* `row.remove()` for non-2-cell rows), and asserts via `block.textContent` rather than a brittle selector.
- The learn-more/chevron/tabindex behavior and the `.hero-card-text-wrapper` sibling-not-child check for `learnMore` in `offer-hero.js` (lines ~568-585) are both correctly verified.

One coverage gap worth flagging (non-blocking, not a correctness issue): `initAnimation()` in `libs/c2/blocks/offer-hero/offer-hero.js:154-379` — roughly 225 lines of scroll-driven pile animation, IntersectionObserver wiring, and resize handling — has zero test coverage. That's reasonable to leave for now (this behavior is CSS-scroll-timeline + real-viewport dependent and hard to unit test meaningfully), but worth a follow-up ticket if this logic gets touched again, since today a regression there would ship silently.

zagi25 (approved):

PR #6388, opened 2026-07-30
MWPW-202199: add unit tests for c2 side-by-side block

SilviuLCF:
## Summary
Adds unit tests for the c2 Milo block `side-by-side` (`libs/c2/blocks/side-by-side/side-by-side.js`). 12 tests, all green, covering deterministic DOM decoration produced by `init(el)`:

- Builds exactly two cards with `card-overlay` / `card-stacked` variants in order.
- Each card pairs a `.media` cell followed by a `.foreground` cell.
- Media cell retains its `picture`/`img`.
- Foreground text decoration: `heading-6` on headings, `title-6` on the `p:has(strong)` label, `body-md` on body copy.
- Inline (`em > a`) link is promoted to a `con-button` inside an `action-area`; standalone plain link is left untouched (opposite branch).
- Overlay card gets `dark` when the block is not dark; does NOT when the block already has `dark` (opposite branch).
- Video media path decorates without throwing and keeps the `<video>` in `.media`.
- No-op branches: single-row block and empty block leave the DOM unchanged / create no cards.

~94% line coverage of the block (74/79).

## Scope note
TEST-ONLY change — nothing under `libs/` was modified. Per scope, timer/rAF/IntersectionObserver/matchMedia/pointer/network behaviors are not asserted (the uncovered lines are inside the IntersectionObserver callback, which only fires on real intersection); tests assert only deterministic DOM decoration and that `init` does not throw.

Resolves: https://jira.corp.adobe.com/browse/MWPW-202199

## Test plan
- [x] `npx wtr --config ./web-test-runner.config.mjs "./test/blocks/side-by-side/side-by-side.test.js" --node-resolve` — 12 passed, 0 failed
- [x] `npx eslint test/blocks/side-by-side/side-by-side.test.js` — clean

### DIFF
--- a/test/blocks/side-by-side/mocks/dark.html
+++ b/test/blocks/side-by-side/mocks/dark.html
@@ -1,0 +1,30 @@
+<main>
+  <div class="section">
+    <div class="side-by-side dark">
+      <div>
+        <div>
+          <picture>
+            <img src="/media/overlay.png" alt="Overlay media" width="800" height="600" />
+          </picture>
+        </div>
+        <div>
+          <picture>
+            <img src="/media/stacked.png" alt="Stacked media" width="800" height="600" />
+          </picture>
+        </div>
+      </div>
+      <div>
+        <div>
+          <p><strong>Overlay heading</strong></p>
+          <h3>Overlay title</h3>
+          <p>Overlay body copy.</p>
+        </div>
+        <div>
+          <p><strong>Stacked heading</strong></p>
+          <h3>Stacked title</h3>
+          <p>Stacked body copy.</p>
+        </div>
+      </div>
+    </div>
+  </div>
+</main>
--- a/test/blocks/side-by-side/mocks/default.html
+++ b/test/blocks/side-by-side/mocks/default.html
@@ -1,0 +1,32 @@
+<main>
+  <div class="section">
+    <div class="side-by-side">
+      <div>
+        <div>
+          <picture>
+            <img src="/media/overlay.png" alt="Overlay media" width="800" height="600" />
+          </picture>
+        </div>
+        <div>
+          <picture>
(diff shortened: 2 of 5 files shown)


mokimo (approved):
LGTM, this one is particularly well done. A few assertions specifically checked because they looked like they could be wrong and weren't:

- "promotes an inline (em > a) link into a con-button action area" correctly relies on the shared `decorateButtons`/`decorateBlockText` utility (`libs/utils/decorate.js:56-84`), not something local to `side-by-side.js` — confirmed the `em a` selector and the `.action-area` class land on the right element after the `<em>` wrapper is removed.
- "does not add dark to the overlay card when block is already dark" correctly exercises the `if (!block.classList.contains('dark'))` guard in `side-by-side.js` — a real edge case, not a restatement of the happy path.
- No-op branches (`single-row.html`, fully empty block) both assert `init(block)` doesn't throw and leaves DOM untouched, matching the early-return guards in `decorate()`.

Nothing blocking. Nice use of `beforeEach` (fresh DOM per test) here vs `before` in the sibling hub-hero PR — this is the safer default.

zagi25 (approved):

LGTM — test-only, CI green, real branch coverage (dark-guard, `em > a` button vs. plain-link, no-op guards). Approving.

Non-blocking and not specific to this PR: the `heading-6`/`title-6` assertions pass because `isC2` is `false` under the test harness (no `foundation=c2` meta set at module load). In production (`isC2 === true`), `decorateBlockText` toggles an `eyebrow` class and the `body-md → title-6` swap no-ops — so those assertions describe a path that doesn't run in prod. This affects the whole c2 suite, so it's a shared-fixture opportunity (set `foundation=c2` meta) rather than anything to fix here.

PR #6389, opened 2026-07-30
MWPW-202193: add unit tests for c2 hub-hero block

SilviuLCF:
## Summary
Adds unit tests for the c2 Milo block `hub-hero` (`libs/c2/blocks/hub-hero/hub-hero.js`). This is test-only; nothing under `libs/` was changed. 15 tests across two mocks cover the deterministic DOM decoration performed by `init(el)`:

- Block is rebuilt into three children in order: `.hub-hero-header`, `.hub-hero-image-grid-container`, `.hub-hero-carousel`.
- Hero header CTA is decorated into an `a.promo-cta` with the arrow icon and image, using the text after `|` as the `aria-label` (alias) and falling back to the CTA text when no alias is authored.
- Image grid builds 5 columns and clones the media of authored slides 2 and 4 into columns 2 and 4.
- Carousel exposes `aria-roledescription="carousel"` and an `aria-label` taken from the first slide link name, falling back to `Adobe Cards` when unnamed; the carousel header is prepended.
- Per-slide decoration: 4 authored slides + 1 injected placeholder; continuous `data-index` (1-4) skipping the placeholder; header/media/footer containers with the add icon; `aria-labelledby` wired to generated eyebrow/heading ids; `daa-ll` tracking label.
- Opposite/negative branches: modal link slide (`role="button"` + `data-modal-hash`/`data-modal-path`) vs standard link slide (`role="link"`, no modal data); video asset slide (`preload=none`, `muted`, `tabindex=-1`, no `controls`, `<source type=video/mp4>`) vs picture asset slide.

## Scope
Only the deterministic DOM decoration is asserted. Intentionally NOT asserted, because they are driven by runtime environment/behavior rather than initial decoration: IntersectionObserver-based mobile autoplay/rewind, `MutationObserver` cleanup, `requestAnimationFrame`-based slide offset measurement (`setCarouselSlideOffsets`), `matchMedia` mobile checks, hover/focus/mouseleave/wheel/scroll event handlers, video playback and rewind timers, `AbortController`-based preload upgrade, and analytics `sendAnalytics` calls. `init` is still executed in every test and must run without throwing. The block has no empty/no-rows early return, so no such no-op case exists to assert.

Resolves: https://jira.corp.adobe.com/browse/MWPW-202193

## Test plan
- [x] `npx wtr --config ./web-test-runner.config.mjs "./test/blocks/hub-hero/hub-hero.test.js" --node-resolve --port=2091` — 15/15 passing
- [x] `npx eslint test/blocks/hub-hero/hub-hero.test.js` — clean

### DIFF
--- a/test/blocks/hub-hero/hub-hero.test.js
+++ b/test/blocks/hub-hero/hub-hero.test.js
@@ -1,0 +1,162 @@
+import { readFile } from '@web/test-runner-commands';
+import { expect } from '@esm-bundle/chai';
+
+import init from '../../../libs/c2/blocks/hub-hero/hub-hero.js';
+
+const loadBlock = async (path) => {
+  document.body.innerHTML = await readFile({ path });
+  const block = document.querySelector('.hub-hero');
+  await init(block);
+  return block;
+};
+
+describe('hub-hero block', () => {
+  describe('top-level structure (default.html)', () => {
+    let block;
+    before(async () => {
+      block = await loadBlock('./mocks/default.html');
+    });
+
+    it('replaces the block content with header, grid and carousel in order', () => {
+      const kids = [...block.children];
+      expect(kids).to.have.lengthOf(3);
+      expect(kids[0].classList.contains('hub-hero-header')).to.be.true;
+      expect(kids[1].classList.contains('hub-hero-image-grid-container')).to.be.true;
+      expect(kids[2].classList.contains('hub-hero-carousel')).to.be.true;
+    });
+
+    it('decorates the hero header CTA as a promo-cta with an alias aria-label', () => {
+      const cta = block.querySelector('.hub-hero-header a.promo-cta');
+      expect(cta).to.exist;
+      // text before the pipe is the label, text after is the aria-label
+      expect(cta.getAttribute('aria-label')).to.equal('Explore all Adobe products');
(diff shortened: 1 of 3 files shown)


mokimo (approved):
LGTM. Traced every assertion against `libs/c2/blocks/hub-hero/hub-hero.js` and both mocks — they hold up, including the subtler ones:

- The placeholder-slide index math (`hub-hero.js`'s `index > 2 ? index - 1 : index` adjustment) is verified correctly in "assigns continuous data-index values across the real slides" — this is exactly the kind of off-by-one a shallow test would get wrong, and it's right.
- "builds a 5-column image grid and clones slide media into columns 2 and 4" correctly targets `cols[1]`/`cols[3]` (0-based) matching the `colMap = {1:0,2:1,4:3,5:4}` 1-based nth-child mapping in source.
- Fallback branches (no pipe in CTA/carousel-name text) are covered via `mocks/fallbacks.html`.

No blocking issues. One non-blocking style note: the two `describe` blocks use mocha `before()` (not `beforeEach()`) to decorate a block shared across several `it`s — safe here since none of the assertions mutate the DOM, but worth keeping in mind if a future test in the same block is added that does mutate state (switch to `beforeEach` at that point).

zagi25 (approved):

PR #6393, opened 2026-07-30
MWPW-201399 Preload MAS fragments

TsayAdobe:
As more pages integrate with MAS, prefetching MAS fragment content is important for performance. This PR adds an early prefetch for MAS fragment URLs referenced on the marquee: as soon as a MAS-fragment autoblock link (merch-card, merch-card-collection, mas-compare-chart) is recognized during autoblock decoration, its fragment URL is preloaded — before the block itself would otherwise request it.

Since the locale/country derivation needed to build a matching URL depends on merch.js's GeoMap, a copy of that lookup table was added to utils.js (MAS_GEO_MAP).

Resolves: [MWPW-201399](https://jira.corp.adobe.com/browse/MWPW-201399)

**Test URLs:**
- After: https://MWPW-201399--milo--adobecom.aem.page/?martech=off

The following tests was done with content overrides on www.adobe.com. In the file `/mas/libs/merch-cards.js`, `credentials:"omit"` should be replaced with `credentials:"same-origin"`.  Then replace the file `/libs/utils/utils.js` with the one in this PR. Environment settings: CPU: **4X slowdown**, Network: **Fast 4G**, and Disable network cache.
* https://www.adobe.com/fr/products/premiere.html?akamaiLocale=be&mep=off
(when mep is on, the marquee is replaced and it doesn't use a MAS fragment)
- Before:
The call `/mas/io/fragment` is usually taking about 200ms to 800ms blocking the LCP.
Before-Premiere-BE

- After:
The fragment is preloaded and the LCP is done earlier.
After-Premiere-BE

### DIFF
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -1495,7 +1495,7 @@ const MAS_FRAGMENT_AUTOBLOCKS = [
 
 const MAS_MEP_SELECTOR = 'mas';
 const MAS_FRAGMENT_API = 'https://www.adobe.com/mas/io/fragment';
-const MAS_FRAGMENT_PRELOAD_API_KEY = 'wcms-commerce-ims-ro-user-milo-cc';
+const DEFAULT_MAS_FRAGMENT_API_KEY = 'wcms-commerce-ims-ro-user-milo';
 
 // Kept in sync with GeoMap in blocks/merch/merch.js.
 const MAS_GEO_MAP = {
@@ -1625,7 +1625,8 @@ function preloadMasFragment(a) {
 
   // Mirrors aem-fragment.js: country is only a separate param when it's not already implied
   // by the locale suffix (e.g. locale=fr_CH + country=CH would be redundant).
-  let endpoint = `${MAS_FRAGMENT_API}?id=${fragmentId}&api_key=${MAS_FRAGMENT_PRELOAD_API_KEY}&locale=${locale}`;
+  const apiKey = getConfig().commerce?.['wcs-api-key'] ?? DEFAULT_MAS_FRAGMENT_API_KEY;
+  let endpoint = `${MAS_FRAGMENT_API}?id=${fragmentId}&api_key=${apiKey}&locale=${locale}`;
   if (country && !locale.endsWith(`_${country}`)) endpoint += `&country=${country}`;
 
   loadLink(endpoint, { rel: 'preload', as: 'fetch', crossorigin: 'anonymous', type: 'application/json' });
--- a/libs/utils/utils.js
+++ b/libs/utils/utils.js
@@ -1487,6 +1487,150 @@ export function isTrustedAutoBlock(autoBlock, url) {
     || (autoBlock === '.pdf' && url.pathname.endsWith(autoBlock));
 }
 
+const MAS_FRAGMENT_AUTOBLOCKS = [
+  'mas-compare-chart-autoblock',
+  'merch-card-collection-autoblock',
+  'merch-card-autoblock',
+];
+
+const MAS_MEP_SELECTOR = 'mas';
+const MAS_FRAGMENT_API = 'https://www.adobe.com/mas/io/fragment';
+const MAS_FRAGMENT_PRELOAD_API_KEY = 'wcms-commerce-ims-ro-user-milo-cc';
+
+// Kept in sync with GeoMap in blocks/merch/merch.js.
+const MAS_GEO_MAP = {
+  ar: 'AR_es',
+  be_en: 'BE_en',
+  be_fr: 'BE_fr',
+  be_nl: 'BE_nl',
+  br: 'BR_pt',
+  ca: 'CA_en',
+  ch_de: 'CH_de',
+  ch_fr: 'CH_fr',
+  ch_it: 'CH_it',
+  cl: 'CL_es',
+  co: 'CO_es',
+  la: 'DO_es',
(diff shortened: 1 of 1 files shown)


libs/utils/utils.js
```diff
@@ -1487,6 +1487,151 @@ export function isTrustedAutoBlock(autoBlock, url) {
...
+  'merch-card-collection-autoblock',
+  'merch-card-autoblock',
+];
+
+const MAS_MEP_SELECTOR = 'mas';
+const MAS_FRAGMENT_API = 'https://www.adobe.com/mas/io/fragment';
+const DEFAULT_MAS_FRAGMENT_API_KEY = 'wcms-commerce-ims-ro-user-milo';
+
+// Kept in sync with GeoMap in blocks/merch/merch.js.
+const MAS_GEO_MAP = {
```
yesil, line 1501: @TsayAdobe I find it risky to duplicate mas GEO in a different file.
the fragment must be fetched with the right parameters once Milo initializes fully with the Lingo logic.

maybe you can move this logic around `isLcpSection` and import merch.js to let commerce initialize and then use it to read country/locale/api key etc.

yesil (changes requested):
fyi: @npeltier

narcis-radu (changes requested):
A lot of MAS related logic that's going to be loaded by all Milo consumers, regardless of whether they use MAS or not. The code should not live inside `utils.js`, and should be loaded only when / if needed.

vhargrave:
@TsayAdobe just an idea because I know we need this fix , is there a super lightweight file we could load from m@s as quickly as possible if a m@s autoblock is in LCP, which holds the mas geo information and kicks off the fragment load and everything else, so that there's still just one file in M@S with that information, but we're also kicking the load off as fast as we can ? 
perf

libs/utils/utils.js
TsayAdobe: @yesil For performance purposes, we should move more merch initialization services into Milo’s utils.js. We can then preload merch scripts and prefetch fragments. We can externalize mas GEO in utils.js and have the merch scripts import it.

How often do we update a fragment’s ID or country in a way that makes prefetching ineffective? If that happens less than 10% of the time, I think prefetching is still a net win.

TsayAdobe:
> A lot of MAS related logic that's going to be loaded by all Milo consumers, regardless of whether they use MAS or not. The code should not live inside `utils.js`, and should be loaded only when / if needed.

@narcis-radu This needs to run synchronously inside decorateAutoBlock, before any async work, so the preload fires in the same tick a MAS link is recognized — that's what lets it win the race against the block's own fetch. Deferring it to a dynamic import would push it past that tick and defeat the purpose. 

On pages that don't use MAS, the cost is just the early-return check against MAS_FRAGMENT_AUTOBLOCKS — no request, no locale/geo work runs.

TsayAdobe:
> @TsayAdobe just an idea because I know we need this fix , is there a super lightweight file we could load from m@s as quickly as possible if a m@s autoblock is in LCP, which holds the mas geo information and kicks off the fragment load and everything else, so that there's still just one file in M@S with that information, but we're also kicking the load off as fast as we can ? 

@vhargrave Lightweight doesn't solve this — for a file this small, the cost is almost entirely the network round-trip (DNS + connection + request/response), not transfer time, so shrinking the payload barely helps. Any separately-hosted file still costs a full extra RTT we don't pay today. utils.js is already part of Milo's bootstrap and loads on every page regardless, so folding the geo data into it adds zero extra requests — we're just parsing a few more KB of a file already in flight. That's why this needs to live in utils.js, not a separate file, even a tiny one.

vhargrave:
> > @TsayAdobe just an idea because I know we need this fix , is there a super lightweight file we could load from m@s as quickly as possible if a m@s autoblock is in LCP, which holds the mas geo information and kicks off the fragment load and everything else, so that there's still just one file in M@S with that information, but we're also kicking the load off as fast as we can ?
> 
> @vhargrave Lightweight doesn't solve this — for a file this small, the cost is almost entirely the network round-trip (DNS + connection + request/response), not transfer time, so shrinking the payload barely helps. Any separately-hosted file still costs a full extra RTT we don't pay today. utils.js is already part of Milo's bootstrap and loads on every page regardless, so folding the geo data into it adds zero extra requests — we're just parsing a few more KB of a file already in flight. That's why this needs to live in utils.js, not a separate file, even a tiny one.

@TsayAdobe fair - alright, two comments from my end then
1. if you can clean up the comments 
2. if you can rewrite this a little so that the merch stuff is only preloaded if a price is in LCP (essentially only if a price is in the first section). To do that you could pass the section to processLinkDecoration , and then to the decorateAutoBlock functions and check if the section index is zero. Something like that. 

If you do that then this looks good to me .

libs/utils/utils.js
yesil: I agree but we should use merch.js to initialize it.
Maybe even before `decorateAutoBlock`, you can check if there is merch in the LCP block, load merch.js first, init commerce and and use it to preload the fragment.

narcis-radu:
Thanks, @TsayAdobe. Would it be possible to check whether M@S is in use before `utils.js` even loads? That way we'd know upfront if the extra code is actually needed. Not sure if this is feasible, but it seems worth exploring before we commit to loading additional code in `utils.js`.
https://github.com/adobecom/milo-college/blob/main/scripts/scripts.js#L109

PR #6399, opened 2026-08-03
[Site Redesign] [CPro Offer] | App Roller Carousel

suhjainadobe:
[Site Redesign] [CPro Offer] | App Roller Carousel

Resolves: [[MWPW-199005](https://jira.corp.adobe.com/browse/MWPW-199005)](https://jira.corp.adobe.com/browse/MWPW-NUMBER)

**Test URLs:**
- Before:https://main--da-cc--adobecom.aem.page/cc-shared/fragments/tests/2026/q3/ace1209/fragments/cpro-offer
- After: https://main--da-cc--adobecom.aem.page/cc-shared/fragments/tests/2026/q3/ace1209/fragments/cpro-offer?milolibs=roller-carousel-design&mep=%2Fcc-shared%2Ffragments%2Ftests%2F2026%2Fq3%2Face1209%2Face1209-cpro-redesign.json--all#legal-tour

### DIFF
--- a/libs/mep/ace1209/roller-carousel/roller-carousel.css
+++ b/libs/mep/ace1209/roller-carousel/roller-carousel.css
@@ -1,0 +1,385 @@
+.roller-carousel {
+  background: var(--s2a-color-background-knockout);
+  color: var(--s2a-color-content-knockout);
+
+  .rcc-scroll-wrapper {
+    position: relative;
+    z-index: 1;
+  }
+
+  .rcc-sticky {
+    position: sticky;
+    top: 0;
+    height: 100dvh;
+    overflow: hidden;
+  }
+
+  .rcc-bg {
+    position: sticky;
+    top: 0;
+    height: 100dvh;
+    margin-bottom: -100dvh;
+    z-index: 0;
+    overflow: hidden;
+    pointer-events: none;
+  }
+
+  .rcc-bg-slide {
+    position: absolute;
+    inset: 0;
+    opacity: 0;
+    transition: opacity 0.7s ease;
+
+    &.is-active {
+      opacity: 1;
+    }
+
+    picture {
+      position: absolute;
+      inset: -150px;
+      display: block;
+    }
+
+    img {
+      width: 100%;
+      height: 100%;
+      object-fit: cover;
+      filter: blur(145px);
+    }
+  }
+
+  .rcc-bg-overlay {
+    position: absolute;
+    inset: 0;
+    background: var(--s2a-color-transparent-black-64);
+  }
+
+  .rcc-content {
+    position: relative;
+    z-index: 1;
+    display: flex;
+    flex-direction: column;
+    height: 100%;
+    padding: var(--s2a-layout-md) var(--s2a-spacing-lg) 0;
+    box-sizing: border-box;
+  }
+
+  .rcc-left {
+    display: flex;
+    flex-direction: column;
+    flex: 1;
+    min-height: 0;
+    gap: var(--s2a-spacing-3xl);
+  }
+
+  .rcc-header {
+    display: flex;
+    flex-direction: column;
+    gap: var(--s2a-spacing-xs);
+    flex-shrink: 0;
+  }
+
+  .rcc-eyebrow {
+    font-family: var(--body-font-family);
+    font-size: var(--s2a-typography-font-size-eyebrow);
+    line-height: var(--s2a-typography-line-height-eyebrow);
+    font-weight: var(--s2a-font-weight-eyebrow);
+    letter-spacing: var(--s2a-typography-letter-spacing-eyebrow);
+    color: var(--s2a-color-content-knockout);
+    margin: 0;
+    padding: 0;
+  }
+
+  .rcc-heading {
+    font-family: var(--heading-font-family);
+    font-size: var(--s2a-typography-font-size-heading-2);
+    line-height: var(--s2a-typography-line-height-heading-2);
+    letter-spacing: var(--s2a-typography-letter-spacing-heading-2);
+    color: var(--s2a-color-content-knockout);
+    margin: 0;
+    padding: 0;
+  }
+
+  .rcc-carousel {
+    display: flex;
+    flex-direction: column;
+    flex: 1;
+    min-height: 0;
+    position: relative;
+  }
+
+  .rcc-category-wrapper {
+    display: flex;
+    flex-direction: column;
+    gap: var(--s2a-spacing-xs);
+    flex-shrink: 0;
+  }
+
+  .rcc-category {
+    font-family: var(--heading-font-family);
+    font-size: var(--s2a-typography-font-size-heading-6);
+    line-height: var(--s2a-typography-line-height-heading-6);
+    font-weight: var(--s2a-font-weight-heading);
+    letter-spacing: var(--s2a-typography-letter-spacing-heading-6);
+    color: var(--s2a-color-content-knockout);
+    white-space: nowrap;
+    margin: 0;
+  }
+
+  .rcc-divider {
+    height: 1px;
+    background: rgba(255 255 255 / 10%);
+  }
+
+  .rcc-list-wrapper {
+    flex: 1;
+    overflow: hidden;
+    min-height: 0;
+    position: relative;
+  }
+
+  .rcc-list {
+    position: absolute;
+    left: 0;
+    right: 0;
+    margin: 0;
+    padding: 0;
+    list-style: none;
+    will-change: transform;
+  }
+
+  .rcc-item {
+    font-family: var(--heading-font-family);
+    font-size: var(--s2a-typography-font-size-heading-2);
+    line-height: var(--s2a-typography-line-height-heading-2);
+    letter-spacing: var(--s2a-typography-letter-spacing-heading-2);
+    color: var(--s2a-color-content-knockout);
+    opacity: 0.12;
+    transition: opacity 0.3s ease;
+    cursor: default;
+    user-select: none;
+    white-space: nowrap;
+
+    &.is-active {
+      opacity: 1;
+    }
+  }
+
+  .rcc-media-wrapper {
+    position: absolute;
+    right: var(--s2a-spacing-lg);
+    bottom: var(--s2a-spacing-lg);
+    z-index: 2;
+    aspect-ratio: 650 / 766;
+    height: clamp(265px, 100dvh - 382px, 360px);
+    max-width: 70vw;
+    border-radius: var(--s2a-border-radius-sm);
+    overflow: hidden;
+  }
+
+  .rcc-media-icon {
+    position: absolute;
+    top: var(--s2a-spacing-md);
+    left: var(--s2a-spacing-md);
+    width: 24px;
+    height: 24px;
+    z-index: 1;
+
+    img {
+      width: 100%;
+      height: 100%;
+      object-fit: contain;
+    }
+  }
+
+  .rcc-media-slide {
+    position: absolute;
+    inset: 0;
+    opacity: 0;
+    transition: opacity 0.7s ease;
+
+    &.is-active {
+      opacity: 1;
+    }
+
+    > picture {
+      display: block;
+      width: 100%;
+      height: 100%;
+
+      img {
+        width: 100%;
+        height: 100%;
+        object-fit: cover;
+      }
+    }
+  }
+
+  &.rcc-reflow {
+    > .rcc-header {
+      position: relative;
+      z-index: 1;
+      display: flex;
+      flex-direction: column;
+      gap: var(--s2a-spacing-xs);
+      padding: var(--s2a-layout-sm) var(--s2a-spacing-lg) var(--s2a-spacing-3xl);
+    }
+
+    .rcc-content {
+      padding-top: 0;
+    }
+
+    .rcc-sticky {
+      top: var(--s2a-layout-sm);
+      height: calc(100dvh - var(--s2a-layout-sm));
+    }
+  }
+
+  &.rcc-reduced-motion {
+    position: relative;
+
+    .rcc-bg {
+      position: absolute;
+      inset: 0;
+      height: auto;
+      margin: 0;
+    }
+
+    .rcc-rm-content {
+      position: relative;
+      z-index: 1;
+      display: flex;
+      flex-direction: column;
+      gap: var(--s2a-spacing-3xl);
+      max-width: 759px;
+      padding: var(--s2a-layout-sm) var(--s2a-spacing-lg);
+      box-sizing: border-box;
+    }
+
+    .rcc-rm-list {
+      display: flex;
+      flex-direction: column;
+    }
+
+    .rcc-rm-group {
+      display: flex;
+      flex-direction: column;
+      margin: 0;
+      padding: 0;
+      list-style: none;
+    }
+
+    .rcc-rm-category {
+      margin: var(--s2a-spacing-3xl) 0 var(--s2a-spacing-md);
+
+      &:first-child {
+        margin-top: 0;
+      }
+    }
+
+    .rcc-rm-item {
+      appearance: none;
+      border: 0;
+      background: none;
+      padding: 0;
+      text-align: start;
+      cursor: pointer;
+      font-family: var(--heading-font-family);
+      font-size: var(--s2a-typography-font-size-heading-2);
+      line-height: var(--s2a-typography-line-height-heading-2);
+      letter-spacing: var(--s2a-typography-letter-spacing-heading-2);
+      color: var(--s2a-color-content-knockout);
+
+      &:focus-visible {
+        outline: 2px solid var(--s2a-color-content-knockout);
+        outline-offset: 4px;
+      }
+    }
+  }
+}
+
+@media (width >= 768px) {
+  .roller-carousel {
+    .rcc-content {
+      padding-top: var(--s2a-layout-lg);
+    }
+
+    .rcc-item {
+      font-size: var(--s2a-typography-font-size-heading-1);
+      line-height: var(--s2a-typography-line-height-heading-1);
+      letter-spacing: var(--s2a-typography-letter-spacing-heading-1);
+    }
+
+    .rcc-media-wrapper {
+      bottom: var(--s2a-spacing-lg);
+      height: clamp(252px, 100dvh - 384px, 560px);
+      border-radius: var(--s2a-border-radius-md);
+    }
+
+    .rcc-media-icon {
+      top: var(--s2a-spacing-lg);
+      left: var(--s2a-spacing-lg);
+      width: 32px;
+      height: 32px;
+    }
+  }
+}
+
+@media (width >= 1280px) {
+  .roller-carousel {
+    .rcc-content {
+      flex-direction: row;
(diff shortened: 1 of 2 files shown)


libs/mep/ace1209/roller-carousel/roller-carousel.js
```diff
@@ -0,0 +1,330 @@
...
+
+const SCROLL_PER_APP = 200;
+const M_BREAKPOINT = 768;
+const L_BREAKPOINT = 1280;
+const MIN_ROLLER_ROOM = 120;
+
+function prepPic(picture) {
+  if (!picture) return null;
+  const clone = picture.cloneNode(true);
+  clone.querySelectorAll('img').forEach((img) => img.removeAttribute('loading'));
```
narcis-radu, line 12: I'm wondering why the `loading` attribute gets removed here. Don't we want the images to be lazy loaded?

libs/mep/ace1209/roller-carousel/roller-carousel.js
```diff
@@ -0,0 +1,330 @@
...
+      const catWrap = createTag('div', { class: 'rcc-category-wrapper rcc-rm-category' });
+      const catLabel = createTag('h3', { class: 'rcc-category' });
+      catLabel.textContent = currentCategory;
+      catWrap.append(catLabel, createTag('div', { class: 'rcc-divider', 'aria-hidden': 'true' }));
+      list.append(catWrap);
+      group = createTag('ul', { class: 'rcc-rm-group', 'aria-label': currentCategory });
+      list.append(group);
+    }
+    const li = createTag('li');
+    const btn = createTag('button', { class: 'rcc-rm-item', type: 'button' });
```
narcis-radu: For reduced motion, we're showing app names as `buttons`, but I don't see any action associated.

libs/mep/ace1209/roller-carousel/roller-carousel.js
```diff
@@ -0,0 +1,330 @@
...
+    const icon = pics.length > 1 ? pics[0] : null;
+    const picture = pics.length > 1 ? pics[1] : (pics[0] ?? null);
+    if (name) apps.push({ category: currentCategory, name, picture, icon });
+  });
+
+  return { eyebrowEl, headingEl, apps };
+}
+
+function buildReducedMotion(block, eyebrowEl, headingEl, apps) {
+  block.classList.add('rcc-reduced-motion');
```
narcis-radu, line 102: for reduced motion, the background is eagerly loaded

libs/mep/ace1209/roller-carousel/roller-carousel.js
```diff
@@ -0,0 +1,330 @@
...
+      catWrap.append(catLabel, createTag('div', { class: 'rcc-divider', 'aria-hidden': 'true' }));
+      list.append(catWrap);
+      group = createTag('ul', { class: 'rcc-rm-group', 'aria-label': currentCategory });
+      list.append(group);
+    }
+    const li = createTag('li');
+    const btn = createTag('button', { class: 'rcc-rm-item', type: 'button' });
+    btn.textContent = app.name;
+    li.append(btn);
+    group.append(li);
```
zagi25, line 132: **Potential `TypeError` (blocking).** `group` is only assigned inside `if (app.category && app.category !== currentCategory)`. If the first app has an empty category (content not starting with a single-column category row — `parseContent` initializes `currentCategory = ''`), `group` is still `null` here and this throws, breaking the reduced-motion render. Suggest a fallback before appending:
```js
if (!group) {
  group = createTag('ul', { class: 'rcc-rm-group' });
  list.append(group);
}
group.append(li);
```

libs/mep/ace1209/roller-carousel/roller-carousel.js
zagi25: **Non-functional interactive control (blocking, a11y).** These `<button>`s have `cursor:pointer` and a `:focus-visible` outline but no click handler, and the reduced-motion background is fixed to `apps[0]`. A focusable button that does nothing is a WCAG issue and confusing for keyboard/AT users. Either wire the button up to swap the active bg/state, or render non-interactive text (`<span>`/`<li>`) instead.

libs/mep/ace1209/roller-carousel/roller-carousel.js
```diff
@@ -0,0 +1,330 @@
...
+
+  window.requestAnimationFrame(updatePosition);
+  new ResizeObserver(updatePosition).observe(listWrapper);
+
+  evaluateReflow();
+  window.addEventListener('resize', () => {
+    evaluateReflow();
+    window.requestAnimationFrame(updatePosition);
+  }, { passive: true });
+  window.addEventListener('scroll', () => {
```
zagi25, line 310: **rAF-throttle the scroll handler.** This runs `updatePosition()` (interleaved `getBoundingClientRect` reads + `style` writes) plus `evaluateReflow()` synchronously on every scroll event — forced reflow on a hot path. The `resize` handler just above already wraps `updatePosition` in `requestAnimationFrame`; mirror that here with a `ticking` guard:
```js
let ticking = false;
window.addEventListener('scroll', () => {
  if (ticking) return;
  ticking = true;
  window.requestAnimationFrame(() => {
    updatePosition();
    evaluateReflow();
    ticking = false;
  });
}, { passive: true });
```

libs/mep/ace1209/roller-carousel/roller-carousel.js
```diff
@@ -0,0 +1,330 @@
...
+    buildReducedMotion(block, eyebrowEl, headingEl, apps);
+    return;
+  }
+
+  const refs = buildRoller(block, eyebrowEl, headingEl, apps);
+  optimizeImageLoading(block, refs.bg, refs.media);
+  initScroll(block, refs, apps);
+}
+
+export default function init(el) {
```
zagi25, line 338: `decorateViewportContent` invokes `decorate()` once per viewport container when the content has viewport variations (see `libs/utils/decorate.js`). That means `initScroll` runs multiple times, each adding its own `window` `scroll`/`resize` listeners and a `ResizeObserver` on a detached subtree for the page lifetime. The `if (!wrapRect.height) return;` guard avoids visible breakage, but the duplicate global listeners leak. Can you confirm this block is never authored with viewport variations, or gate `initScroll` to the displayed container?

libs/mep/ace1209/roller-carousel/roller-carousel.js
```diff
@@ -0,0 +1,330 @@
...
+  const bg = buildBg(apps);
+  const content = createTag('div', { class: 'rcc-content' });
+  const left = createTag('div', { class: 'rcc-left' });
+  const header = buildHeader(eyebrowEl, headingEl);
+
+  const carousel = createTag('div', { class: 'rcc-carousel' });
+  const categoryWrapper = createTag('div', { class: 'rcc-category-wrapper' });
+  const categoryLabel = createTag('span', { class: 'rcc-category' });
+  categoryLabel.textContent = apps[0].category;
+  const divider = createTag('div', { class: 'rcc-divider', role: 'separator', 'aria-hidden': 'true' });
```
zagi25: Minor: `{ role: 'separator', 'aria-hidden': 'true' }` — `aria-hidden` removes the node from the a11y tree, so `role="separator"` is dead. Drop the role (the reduced-motion divider correctly uses `aria-hidden` alone).

libs/mep/ace1209/roller-carousel/roller-carousel.css
```diff
@@ -0,0 +1,385 @@
...
+    font-weight: var(--s2a-font-weight-heading);
+    letter-spacing: var(--s2a-typography-letter-spacing-heading-6);
+    color: var(--s2a-color-content-knockout);
+    white-space: nowrap;
+    margin: 0;
+  }
+
+  .rcc-divider {
+    height: 1px;
+    background: rgba(255 255 255 / 10%);
```
zagi25, line 112: Minor: divider uses a hardcoded `rgba(255 255 255 / 10%)` while everything else in the file uses `--s2a-*` tokens. Prefer a token for theming consistency.

zagi25 (commented):
_🤖 Reviewed with Claude Code; posted on my behalf._

Thanks @suhjainadobe — nicely decomposed block, and it tracks the existing `ace1209` conventions well. A couple of user-facing correctness/a11y items I'd like resolved before approval, plus some perf/polish follow-ups. Inline comments have specifics.

**Blocking:**
1. **Reduced-motion path can crash.** In `buildReducedMotion`, `group` stays `null` until a non-empty category is seen. If content doesn't start with a category row (`parseContent` starts `currentCategory = ''`), `group.append(li)` throws a `TypeError` and the block fails for reduced-motion users. The roller path degrades gracefully in the same case, so it's an asymmetric crash.
2. **Reduced-motion buttons don't do anything.** They're `<button>`s with pointer/focus styling but no handler, and the background is fixed to `apps[0]`. That's an interactive control with no effect (WCAG). Either wire them up or render non-interactive elements.

**Non-blocking follow-ups:**
- Scroll handler isn't rAF-throttled (the resize path already is) — layout thrash on a hot path.
- `decorateViewportContent` runs `decorate()` once per viewport variation, so `initScroll` attaches duplicate `window` listeners + `ResizeObserver`s if this block is ever authored with viewport variations. Worth confirming intended usage.
- Minor: hardcoded divider color vs `--s2a-*` tokens; contradictory `role="separator"` + `aria-hidden` on the roller divider.

No security concerns — all DOM is built via `createTag`/`textContent`, no HTML injection surface.

libs/mep/ace1209/roller-carousel/roller-carousel.css
```diff
@@ -0,0 +1,385 @@
...
+  }
+
+  .rcc-header {
+    display: flex;
+    flex-direction: column;
+    gap: var(--s2a-spacing-xs);
+    flex-shrink: 0;
+  }
+
+  .rcc-eyebrow {
```
narcis-radu, line 82: using `.eyebrow` via JS is also an option.

libs/mep/ace1209/roller-carousel/roller-carousel.css
```diff
@@ -0,0 +1,385 @@
...
+    white-space: nowrap;
+
+    &.is-active {
+      opacity: 1;
+    }
+  }
+
+  .rcc-media-wrapper {
+    position: absolute;
+    right: var(--s2a-spacing-lg);
```
narcis-radu: I see `RTL` is not fully handled, maybe a TODO should be added.

libs/mep/ace1209/roller-carousel/roller-carousel.css
```diff
@@ -0,0 +1,385 @@
...
+    top: var(--s2a-spacing-md);
+    left: var(--s2a-spacing-md);
+    width: 24px;
+    height: 24px;
+    z-index: 1;
+
+    img {
+      width: 100%;
+      height: 100%;
+      object-fit: contain;
```
narcis-radu, line 168: contain or cover ?

libs/mep/ace1209/roller-carousel/roller-carousel.css
```diff
@@ -0,0 +1,385 @@
...
+    }
+
+    .rcc-rm-item {
+      appearance: none;
+      border: 0;
+      background: none;
+      padding: 0;
+      text-align: start;
+      cursor: pointer;
+      font-family: var(--heading-font-family);
```
narcis-radu: if we don't use buttons, is there an use for `font-family` ?

libs/mep/ace1209/roller-carousel/roller-carousel.css
```diff
@@ -0,0 +1,385 @@
...
+    color: var(--s2a-color-content-knockout);
+    margin: 0;
+    padding: 0;
+  }
+
+  .rcc-heading {
+    font-family: var(--heading-font-family);
+    font-size: var(--s2a-typography-font-size-heading-2);
+    line-height: var(--s2a-typography-line-height-heading-2);
+    letter-spacing: var(--s2a-typography-letter-spacing-heading-2);
```
overmyheadandbody: Same comment as @narcis-radu's above - this could simply use a `heading-2` class to keep things consistent and reduce some code repetition.

libs/mep/ace1209/roller-carousel/roller-carousel.css
```diff
@@ -0,0 +1,385 @@
...
+    gap: var(--s2a-spacing-xs);
+    flex-shrink: 0;
+  }
+
+  .rcc-category {
+    font-family: var(--heading-font-family);
+    font-size: var(--s2a-typography-font-size-heading-6);
+    line-height: var(--s2a-typography-line-height-heading-6);
+    font-weight: var(--s2a-font-weight-heading);
+    letter-spacing: var(--s2a-typography-letter-spacing-heading-6);
```
overmyheadandbody: A `heading-6` class would achieve the same result

libs/mep/ace1209/roller-carousel/roller-carousel.css
overmyheadandbody: `inset-inline-end` should be preferred, as we aim to use logical properties in our CSS

libs/mep/ace1209/roller-carousel/roller-carousel.css
```diff
@@ -0,0 +1,385 @@
...
+      appearance: none;
+      border: 0;
+      background: none;
+      padding: 0;
+      text-align: start;
+      cursor: pointer;
+      font-family: var(--heading-font-family);
+      font-size: var(--s2a-typography-font-size-heading-2);
+      line-height: var(--s2a-typography-line-height-heading-2);
+      letter-spacing: var(--s2a-typography-letter-spacing-heading-2);
```
overmyheadandbody, line 261: Apply a `heading-2` class instead

libs/mep/ace1209/roller-carousel/roller-carousel.css
```diff
@@ -0,0 +1,385 @@
...
+
+    .rcc-media-wrapper {
+      bottom: var(--s2a-spacing-lg);
+      height: clamp(252px, 100dvh - 384px, 560px);
+      border-radius: var(--s2a-border-radius-md);
+    }
+
+    .rcc-media-icon {
+      top: var(--s2a-spacing-lg);
+      left: var(--s2a-spacing-lg);
```
overmyheadandbody: Prefer logical inset properties

libs/mep/ace1209/roller-carousel/roller-carousel.css
```diff
@@ -0,0 +1,385 @@
+.roller-carousel {
+  background: var(--s2a-color-background-knockout);
+  color: var(--s2a-color-content-knockout);
+
+  .rcc-scroll-wrapper {
```
overmyheadandbody, line 5: Is there a real need for nesting in this file? The CSSOM would be a bit more efficient without the additional nesting

libs/mep/ace1209/roller-carousel/roller-carousel.css
suhjainadobe: it would be contain as tt scales the image to fit inside the box while preserving its full aspect ratio, so nothing gets cropped

libs/mep/ace1209/roller-carousel/roller-carousel.js
suhjainadobe: done

libs/mep/ace1209/roller-carousel/roller-carousel.js
suhjainadobe: Confirmed with the design team, for all the viewports, there will always be one authoring.

libs/mep/ace1209/roller-carousel/roller-carousel.js
suhjainadobe: The reason , button was added initially was because they were supposed to be interactive so we the user can click them and change the foreground image. But, as confirmed with the design team, we are no longer showing the foreground image at all on reduce motion. Thus removed the button.

suhjainadobe: The reason , button was added initially was because they were supposed to be interactive so we the user can click them and change the foreground image. But, as confirmed with the design team, we are no longer showing the foreground image at all on reduce motion. Thus removed the button.

libs/mep/ace1209/roller-carousel/roller-carousel.css
suhjainadobe: done!

libs/mep/ace1209/roller-carousel/roller-carousel.css
suhjainadobe: done

libs/mep/ace1209/roller-carousel/roller-carousel.css
suhjainadobe: done!

libs/mep/ace1209/roller-carousel/roller-carousel.css
suhjainadobe: done

libs/mep/ace1209/roller-carousel/roller-carousel.css
suhjainadobe: The reason to add nesting is mainly for &.rcc-reduced-motion and &.rcc-reflow classes. Written flat, the reflow/reduced-motion blocks would each have to repeat the .roller-carousel.rcc-reflow prefix on every child rule. Also, nesting inside the bracket made it easier to edit.

libs/mep/ace1209/roller-carousel/roller-carousel.js
suhjainadobe: Yes narcis, as only background image would be loading for reduce-motion, it will be eager-loaded.

libs/mep/ace1209/roller-carousel/roller-carousel.css
suhjainadobe: removed it

libs/mep/ace1209/roller-carousel/roller-carousel.css
suhjainadobe: Actually, 10% transparent white is not there, thus added the hardcoded value

libs/mep/ace1209/roller-carousel/roller-carousel.css
suhjainadobe: done

libs/mep/ace1209/roller-carousel/roller-carousel.js
suhjainadobe: Removing loading resets Milo's default lazy so the block can set its own strategy. Lazy doesn't fit here — all slides are stacked in the sticky first viewport and cross-fade on scroll, so they'd either load anyway or pop-in mid-transition. Instead optimizeImageLoading loads the first bg/media slide eagerly (bg = fetchpriority: high, it's the LCP) and drops the rest to fetchpriority: low so they preload without competing with first paint.

narcis-radu (approved):

zagi25 (approved):

overmyheadandbody (approved):

SilviuLCF (approved):
Verified , testing details https://jira.corp.adobe.com/browse/MWPW-199005

PR #6412, opened 2026-08-04
 MWPW-202632 [MEP] ACE1225 upload-marquee: poster image → video swap + CLS fix

markpadbe:
### What
  The media cell shows a poster image that swaps to the video once it buffers — **same slot, no layout change**. Authors put the mp4 URL in the image's **alt text** using milo's standard convention:
  `https://…/clip.mp4#_autoplay | caption`.

  ### Why
  The current code (from #6392) authors a separate `<picture>` + `.mp4` link and does manual DOM surgery to wire the poster, which is fragile — poster landing in the wrong place, video not loading. This
  replaces it with milo's native video/poster pipeline, and consolidates on the same `alt="url | caption"` convention the other fragments already use so authors don't have to remember which field each block
  expects.

  ### Changes
  - **Poster→video** — relies on milo's built-ins: `decorateImageLinks` (global) converts the alt-pipe image into an `<a href="…mp4" data-video-poster="…">` before the block runs, then `decorateAnchorVideo`
  (shared) builds the native `<video poster>`. `decorateUploadVideos` is now just glue (find anchor, ensure hash, lift out of `<p>`, hand off); drops the bespoke picture extraction.
  - **Ordering** — run `decorateUploadVideos` before `decorateContentRow` so the media-priority pass can find the `<video>`.
  - **CLS fix** — add the pause/play control styles `video.css` supplies but a `useBlockCode` block doesn't load (absolute positioning + reserved 40×40 `offset-filler`). Stops the async icon SVGs from
  reflowing the media.

  ### Authoring note
  The alt **must** contain a `|` and a full `https://` URL — both are how `decorateImageLinks` gates conversion. Missing either = static poster, no video, no error.

  ### Scope
  ace1225 MEP test code only (`libs/mep/ace1225/`), delivered via `useBlockCode`. **No shared/core files touched.**

  Resolves: [MWPW-202632](https://jira.corp.adobe.com/browse/MWPW-202632)
  - Test URL: https://main--da-cc--adobecom.aem.page/cc-shared/fragments/tests/2026/q3/ace1225/ps-index/z-ace1225-index-new-b?milolibs=mep-ace1225-vid-poster&mep=%2Fproducts%2Fphotoshop.json--desktop---%2Fdrafts%2Fmarkp%2Funity-quick-check.json
 

  - PSI-check: https://mep-ace1225-vid-poster--milo--adobecom.aem.page/?martech=off

### DIFF
--- a/libs/mep/ace1225/upload-marquee/upload-marquee.css
+++ b/libs/mep/ace1225/upload-marquee/upload-marquee.css
@@ -159,6 +159,11 @@
   background: var(--color-gray-800);
 }
 
+.upload-marquee-block .video-container .pause-play-wrapper:focus-visible {
+  outline: 2px solid var(--um-accent-focus);
+  outline-offset: 2px;
+}
+
 .upload-marquee-block .video-container .pause-play-wrapper img.accessibility-control {
   width: 40px;
   height: 40px;
--- a/libs/mep/ace1225/upload-marquee/upload-marquee.css
+++ b/libs/mep/ace1225/upload-marquee/upload-marquee.css
@@ -128,6 +128,45 @@
 .upload-marquee-block .upload-marquee-right .media-container .video-container {
   width: 100%;
   height: 100%;
+  position: relative;
+}
+
+/* Pause/play control: absolute + reserved 40x40 box so the async icon SVGs don't shift the
+   media (CLS). Mirrors libs/blocks/video/video.css, which a useBlockCode block doesn't load. */
+.upload-marquee-block .video-container .pause-play-wrapper {
+  position: absolute;
+  bottom: 2%;
+  top: auto;
+  inset-inline-end: 2%;
+  inset-inline-start: auto;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  margin: 0;
+  padding: 3px;
+  border-radius: 50%;
+  z-index: 3;
+  cursor: pointer;
+}
+
+.upload-marquee-block .video-container .pause-play-wrapper .offset-filler {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  width: 40px;
+  height: 40px;
+  border-radius: inherit;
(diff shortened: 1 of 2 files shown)


harryboddu (approved):

jpratt2:
Code review from Claude
image

jpratt2 (approved):
I see a video running in the left half of the marquee at https://main--da-cc--adobecom.aem.page/cc-shared/fragments/tests/2026/q3/ace1225/ps-index/z-ace1225-index-new-b?milolibs=mep-ace1225-vid-poster&mep=%2Fproducts%2Fphotoshop.json--desktop---%2Fdrafts%2Fmarkp%2Funity-quick-check.json
approved

markpadbe:
> Code review from Claude image

 1. :focus-visible on .pause-play-wrapper — great catch, and you're right that the useBlockCode block never loads video.css so the focus ring was missing. Fixed in ba1b28e7, mirroring the --um-accent-focus /
  :focus-visible pattern the block already uses for .action-button and .drop-zone. Agreed this was the one worth blocking on.
  2. Posterless-video robustness — good flag. Since your review I refactored away the data-title bridge entirely and consolidated on milo's standard alt="url#_autoplay | caption" convention, so the poster now
  comes through decorateImageLinks → decorateAnchorVideo natively rather than my own getImgSrc fallback. The "no webp source → no poster" edge is now just milo's baseline video behavior, so I left it to the
  shared pipeline rather than special-casing it in this temporary block.
  3. Trim the authored URL — also resolved by the same refactor; decorateImageLinks does the source.trim() itself, so there's no longer a raw dataset.title going into href.
  4. Leftover empty <p> — likewise moot after dropping the bridge; in the authored content that paragraph also carries the body copy, so it's never left empty.

  Net: went with the built-in alt-pipe path so authors have one convention across fragments and we carry less ace1225-only code to unwind at RTP. Thanks again

